PR #1085, opened 2019-03-21
Add example NCBI apiKey specification

olgabot:
Documentation-level changes for 19.03-edge release


pditommaso:
Thanks for this. I was not mentioned in the doc, because I'm not yet 100%  sure to keep all this config options. Also you were mentioning that the api key can be specified in the `~/.ncbi` file. It would make sense to support this. Is there doc about that?

ps. the correct syntax is `Channel.fromSRA(ids, apiKey:"0123456789abcdef")` (colon, not equals)

pditommaso:
Ok, merge manually.

PR #1108, opened 2019-04-06
Update docker docs for "Memory limited without swap" error

olgabot:
As mentioned [here](https://github.com/alesssia/YAMP/issues/7), this error has been plaguing a few people:

```
Command error:
WARNING: Your kernel does not support swap limit capabilities or the cgroup is not mounted. Memory limited without swap.
```

I could find the solution in the [YAMP wiki](https://github.com/alesssia/YAMP/wiki/How-to-use-Docker) but not in the Nextflow documentation so here it is.


pditommaso:
Outdated (sorry for the lack of feedback, this got completely lost)

PR #1113, opened 2019-04-12
Adds note in documentation for operators that consume channels

sven1103:
Operators like `println` consume channels, if combined with other operators during channel creation. I added a small note in the docs, that highlights this for:

- print
- println
- subscribe
- set

Any remarks welcome :)


pditommaso:
Excellent! thanks.

Issue #1088, opened 2019-03-26
wr as new Nextflow backend


sb10:
## New feature

I develop [wr](https://github.com/VertebrateResequencing/wr) which is a workflow runner like Nextflow, but can also just be used as a backend scheduler. It can schedule to LSF and OpenStack right now.

The benefit to Nextflow users of going via wr instead of using Nextflow’s existing LSF or Kubernetes support is:

1. wr makes more efficient use of LSF: it can pick an appropriate queue, use job arrays, and “reuse” job slots. In a simple test I did, Nextflow using wr in LSF mode was 2 times faster than Nextflow using its own LSF scheduler.
2. wr’s OpenStack support is incredibly easy to use and set up (basically a single command to run), and provides auto scaling up and down. Kubernetes, by comparison, is really quite complex to get working on OpenStack, doesn’t auto scale, and wastes resources with multiple nodes needed even while no workflows are being operated on. I was able to get Nextflow to work with wr in OpenStack mode (but the shared disk requirement for Nextflow’s state remains a concern).

## Usage scenario 

Users with access to LSF or OpenStack clusters who want to run their Nextflow workflows efficiently and easily.

## Suggest implementation 

Since I don’t know Java well enough to understand how to implement this “correctly”, I wrote a simple bsub emulator in wr, which is what my tests so far have been based on. I submit the Nextflow command as a job to wr, turning on the bsub emulation, and configure Nextflow to use its existing LSF scheduler. While running under the emulation, Nextflow’s bsub calls actually call wr.

Of course the proper way to do this would be have Nextflow call wr directly (either the wr command line, or it’s REST API). The possibly tricky thing with regard to having it work in OpenStack mode is having it tell wr about OpenStack-specific things like what image to use, what hardware flavour to use, pass details on how to mount S3 etc. (the bsub emulation handles all of this).

Here's what I did for my LSF test...

echo_1000_sleep.nf:
```
#!/usr/bin/env nextflow

num = Channel.from(1..1000)

process echo_sleep {
  input:
  val x from num

	output:
	stdout result

  "echo $x && sleep 1"
}

result.subscribe { println it }

workflow.onComplete {
    println "Pipeline completed at: $workflow.complete"
    println "Execution status: ${ workflow.success ? 'OK' : 'failed' }"
}
```

nextflow.config:
```
process {
  executor='lsf'
  queue='normal'
	memory='100MB'
}
```

install wr:
```
[...]

PR #1114, opened 2019-04-15
wr executor

sb10:
Work in progress for an executor that uses wr as a scheduler. See https://github.com/nextflow-io/nextflow/issues/1088

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/WrExecutor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/WrExecutor.groovy
@@ -286,7 +286,7 @@ class WrTaskHandler extends TaskHandler implements BatchHandler<String,Map> {
         bash.build()
 
         WrFileCopyStrategy copyStrategy = bash.copyStrategy as WrFileCopyStrategy
-        String wrapperPath = copyStrategy.wrPath(wrapperFile)
+        String wrapperPath = copyStrategy.wrWorkPath(wrapperFile)
 
         // submit the task
         jobId = client.add("/bin/bash $wrapperPath", task, copyStrategy)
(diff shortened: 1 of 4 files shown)


modules/nextflow/src/main/groovy/nextflow/executor/WrExecutor.groovy
```diff
@@ -0,0 +1,664 @@
...
+		log.debug "[wr] cacertpath=$path"
+        return path
+    }
+
+    /**
+     * @return {@code false} whenever the containerization is managed by the executor itself
+     */
+    boolean isContainerNative() {
+        return false
+    }
```
pditommaso: Does `wr` manage the container execution on its own? In that case, it should be `true`.

modules/nextflow/src/main/groovy/nextflow/executor/WrExecutor.groovy
```diff
@@ -0,0 +1,664 @@
...
+@Slf4j
+@CompileStatic
+class WrTaskHandler extends TaskHandler implements BatchHandler<String,Map> {
+
+    // static final private String WORK_DIR = '/work'
+
+    final List<String> COMPLETE_STATUSES = ['complete', 'buried', 'deleted']
+
+    final List<String> STARTED_STATUSES = ['delayed', 'reserved', 'running'] + COMPLETE_STATUSES
+
```
pditommaso: Makes sense to declare these fields also `public static`

modules/nextflow/src/main/groovy/nextflow/executor/WrExecutor.groovy
```diff
@@ -0,0 +1,664 @@
...
+@Slf4j
+@CompileStatic
+class WrExecutor extends Executor {
+
+	static private String token
+	static private String defaultManagerDir
+    static private String endpoint
+	static private String cacertPath
+	static private WrRestApi client
+
```
pditommaso: It looks there's a problem with tab/space indentation. Please use 4 spaces in place of tabs. More details [here](https://github.com/nextflow-io/nextflow/blob/master/CONTRIBUTING.md).

modules/nextflow/src/main/groovy/nextflow/executor/WrExecutor.groovy
sb10: As in doing something if the user wishes to run a docker container? No, wr doesn't yet have native container support. I'm hoping BashWrapperBuilder will put something appropriate in the wrapper script if this boolean returns false.

pditommaso: Then it's fine and yes, BashWrapperBuilder should manage the container as usual.

sb10:
After I `make compile` I end up with:

```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   docs/conf.py
	modified:   modules/nextflow/src/main/groovy/nextflow/Const.groovy
	modified:   nextflow
```

I did not modify these files myself,
Is this expected? Is there a way to avoid these modifications being reported by git? And to avoid conflicts with them etc?

modules/nextflow/src/main/groovy/nextflow/executor/WrExecutor.groovy
sb10: I made that change, thanks.

modules/nextflow/src/main/groovy/nextflow/executor/WrExecutor.groovy
sb10: Should now be fixed

pditommaso:
The `compile` automatically changes the timestamp and build number, but those changes should not be commited.

sb10:
Superseded by #1148

PR #1125, opened 2019-04-23
Implements the support for GPU resources on K8s

pditommaso:
This commit implements the `gpu` directive that allows the specification of GPU resources as shown below: 

```
gpu request: 5, limit: 10,type: 'nvidia'
```

or 

``` 
gpu 10, type: 'nvidia'
``` 

In the above case `request` == `limit` == 10. When omitted the `type`defaults to `nvidia`. 
  

So far only the K8s executed applies this setting.


bentsherman:
@pditommaso Is there a way for me to test this branch, like with `NXF_VER`?

pditommaso:
Yes, use version `19.04.0-edge`. I've merged into master as an experimental directive.

Issue #1145, opened 2019-05-08
Improve the "reliability" of the weblog


olifly:
## Usage scenario 

We're using the weblog feature to integrate with other services we're running.  The weblog today is rather unreliable when it comes to ensuring that it's listener receives all messages that it should emit.  This "new feature" would enhance the reliability of the weblog (and other TraceObservers).

## Suggest implementation 

The main culprit is that the weblog is emitting messages asynchronously using an Agent.  When the nextflow Session class finishes (successfully, aborted or is cancelled) it doesn't ensure that all observers that is has registered have finished emitting messages/events/traces/etc before it terminates.

The implementation would add an await() method to the TraceObserver interface that observers can override to ensure that they have finished emitting all events they have received.

The Session object would then make sure that before it finishes, to loop through all of its observers and call await().

This implementation would fix issue https://github.com/nextflow-io/nextflow/issues/1010

pditommaso:
More than happy to review a PR for this issue. Not sure to understand what this would bring more that then #1010.

olifly:
@pditommaso The workaround shown in #1010 (using a sleep in workflow.onComplete) doesn't assure we get onFlowError messages when there are syntax errors in the script being executed for example.  This would also fix that

pditommaso:
Provided that #1010 that should be fixed and the `sleep` is not supposed to be a valid patch, still not understanding what this issue would bring more.

pditommaso:
Solved by #1010.

Issue #1010, opened 2019-01-31
Nextflow doesn't wait for WebLogObserver to transmit all http messages before exiting.


olifly:
### Expected behavior and actual behavior

Expected: When executing nextflow with the `-with-weblog [url]` parameter then nextflow should send (taken from [https://www.nextflow.io/docs/latest/tracing.html#weblog-via-http](url))

> detailed workflow execution metadata and runtime statistics to a HTTP endpoint.

Nextflow should ensure that all trace messages have been sent successfully to the HTTP endpoint before shutting down.

Actual:  Nextflow sends HTTP messages asynchronously to the HTTP endpoint. When it exits it doesn't check if all the HTTP messages have been sent causing the web endpoint to potentially miss out on HTTP messages (especially if the HTTP endpoint is slow to respond).

_sidte note: The weblog only accepts HTTP status code 200 as success, everything else is a failure. This is less than optimal since  HTTP POST endpoints often reply with other 2XX statuses (201 - Created for example).  Accepting all 2XX status codes as success would be a preferable change._

### Steps to reproduce the problem

1. Run the following command: `./nextflow -log output.log run hello -with-weblog https://httpbin.org/post && grep -c "code 200" output.log`

This should count how many times the WebObserver receives status code 200 from the HTTP endpoint when executing the standard hello nextflow script.  The count depends on how many times nextflow manages to send HTTP messages before terminating (on my computer it fluctuates between 3 and 4).

2. Save the following nf script as hello-wait.nf

```
#!/usr/bin/env nextflow
echo true

cheers = Channel.from 'Bonjour', 'Ciao', 'Hello', 'Hola'

process sayHello {
  input: 
    val x from cheers
  script:
    """
    echo '$x world!'
    """
}

workflow.onComplete = {
    println "Pipeline complete"
    sleep(4000)
}
```

3. Execute the following command: `./nextflow -log output.log run hello-wait.nf -with-weblog https://httpbin.org/post && grep -c "code 200" output.log`
 
Inserting the wait at the end should give nextflow enough time to send all queued up HTTP messages before terminating (the count should be 13)

### Environment 

* Nextflow version: nextflow version 19.01.0-edge.5050 
* Java version: 1.8.0_144
* Operating system: macOS Mojave 10.14.2 (18C54)

Best Regards,
Ólafur Haukur Flygenring

apeltzer:
@sven1103 you might be interested in this !

sven1103:
[...]

PR #1153, opened 2019-05-15
Session ensures it awaits on all observers before exiting.

olifly:
Session calls await on all observers before destroy/cancel/abort allowing them to finish emitting messages (https://github.com/nextflow-io/nextflow/issues/1145) . Fix for issue: https://github.com/nextflow-io/nextflow/issues/1010

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -200,7 +200,7 @@ class Session implements ISession {
 
     private int poolSize
 
-    private List<TraceObserver> observers = Collections.emptyList()
+    protected List<TraceObserver> observers = Collections.emptyList()
 
     private Closure errorAction
 
@@ -676,6 +676,8 @@ class Session implements ISession {
         terminated = true
         monitorsBarrier.awaitCompletion()
         log.debug "Session await > all barriers passed"
+        observers *.await()
+        log.debug "Observers await > all observers have finished"
     }
 
     void destroy() {
@@ -693,6 +695,9 @@ class Session implements ISession {
             execService = null
             log.trace "Session > executor shutdown"
 
+            //wait for all observers to finish processing
+            observers *.await()
+
             // -- close db
             cache?.close()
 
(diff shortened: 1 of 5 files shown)


pditommaso:
Hi Ólafur, thanks for submitting this patch. I got that the problem is the missing `await`, hower it looks to me that the solution is overkilling. It's enough to add it in the `onFlowComplete` method and indeed the `TraceFileObserver` has already it https://github.com/nextflow-io/nextflow/blob/651af95a59a8d7325ba46109ff325c247de509a8/modules/nextflow/src/main/groovy/nextflow/trace/TraceFileObserver.groovy

I would propose to fix the weblog in the same way.

olifly:
Hi.
That was my first choice of handling the problem, but doing that does not solve the problem of also missing onFlowError messages.  Adding weblogAgent.await to onFlowError would, again, solve that error, but then we have other observers (and maybe others in the future) using async/Agents and they will have to also make sure to await/sync on their onFlowComplete/onFlowError methods to make sure that all messages they receive are processed before nextflow terminates. It also makes those two events in effect synchronous instead of keeping them async until the time the Session needs them to finish up their work.

This fix tackles this problems in a more general way. Making sure that the Session waits for all observers to finish their processing before exiting.  That way observers that are doing async stuff can put their sync code in the await command and observers not doing async don't have to implement it.

Having the "sync/finish your work" command be called "await" on the TraceObserver trait is maybe a bit misleading since it suggests perhaps that all TraceObservers are using Agents.

pditommaso:
Let me check.

pditommaso:
I've double check this, and the `onFlowComplete` is invoked by the [`shutdown`](https://github.com/nextflow-io/nextflow/blob/2322fbcf364761605314bd4b61a660531da55211/modules/nextflow/src/main/groovy/nextflow/Session.groovy#L749) which in turn is invoked when the execution session is destroy 

https://github.com/nextflow-io/nextflow/blob/2322fbcf364761605314bd4b61a660531da55211/modules/nextflow/src/main/groovy/nextflow/Session.groovy#L689

This means that `onFlowComplete` is invoked even after an error condition, therefore, the `await` there should be enough.

pditommaso:
Merge manually (partially) #1010.

PR #1158, opened 2019-05-21
 [DOCS] Add tip for error strategy directive

apeltzer:


pditommaso:
I've realised that this example is already in the [dynamic res section](https://www.nextflow.io/docs/latest/process.html#dynamic-computing-resources). I'll add a link to it in the tip

pditommaso:
Merged manually [bff6c168].

Issue #1107, opened 2019-04-05
S3 Download Fails when using AWS Batch


sivkovic:
### Expected behavior and actual behavior

Execution fails occasionally when using AWS Batch because of broken connection during download of the files. Even if it fails, command is still run and shows the error input file is missing. It should probably have a retry in this case, and return different error if it fails multiple times.

### Steps to reproduce the problem

It is not 100% reproducible since it is network related, but I experienced this error from time to time.

### Program output 

Logs from CloudWatch:
```
download failed: s3://path/to/dbsnp_138.b37.vcf to ./dbsnp_138.b37.vcf ("Connection broken: ConnectionResetError(104, 'Connection reset by peer')", ConnectionResetError(104, 'Connection reset by peer'))
```

### Environment 

* Nextflow version: 19.0.1
* Java version: Java(TM) SE Runtime Environment (build 1.8.0_191-b12)
* Operating system: macOS

pditommaso:
Not sure to understand what is generating that error message. Include a code snippet and the relative `.nextflow.log` file.

sivkovic:
This error comes from `awscli` while downloading files when run from `command.run`
It happens to random process. The log below is from Mutect2 which I got from CloudWatch. If I run nextflow again with -resume most of the time it finish successfully.
```
nxf-scratch-dir ip-10-0-0-247:/nxf.8TrRzUl1QX
download failed: s3://igenomes/Homo_sapiens/GATK/GRCh37/Sequence/WholeGenomeFasta/human_g1k_v37_decoy.fasta to ./human_g1k_v37_decoy.fasta ("Connection broken: ConnectionResetError(104, 'Connection reset by peer')", ConnectionResetError(104, 'Connection reset by peer'))
download failed: s3://igenomes/Homo_sapiens/GATK/GRCh37/Annotation/GATKBundle/dbsnp_138.b37.vcf to ./dbsnp_138.b37.vcf ("Connection broken: ConnectionResetError(104, 'Connection reset by peer')", ConnectionResetError(104, 'Connection reset by peer'))
13:20:27.443 INFO NativeLibraryLoader - Loading libgkl_compression.so from jar:file:/gatk/gatk-package-4.1.0.0-local.jar!/com/intel/gkl/native/libgkl_compression.so
13:20:27.683 INFO Mutect2 - ------------------------------------------------------------
13:20:27.683 INFO Mutect2 - The Genome Analysis Toolkit (GATK) v4.1.0.0
[...]

PR #1164, opened 2019-05-30
s3 download retry

sivkovic:
Hi Paolo,
Since the https://github.com/nextflow-io/nextflow/commit/26d3755677479065d59172227345a42e0f98c512 didn't solve our problem https://github.com/nextflow-io/nextflow/issues/1107 I implemented a retry mechanism when downloading files from s3. We tested it with 200 TN samples (15536 processes). This way we are able to use 'optimal' for instance types for our compute environment and having a single queue for all processes. Also what we detected as really beneficial in this kind of setup is that wait time for instances is shorter and cost is lower since we are not limiting AWS Batch scheduler to just some instance types. If you think this is beneficial I can also add configurable number of retries or any cleanup you think is needed.
Best,


modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/AwsBatchFileCopyStrategy.groovy
```diff
@@ -93,7 +93,7 @@ class AwsBatchFileCopyStrategy extends SimpleFileCopyStrategy {
     @Override
     String stageInputFile( Path path, String targetName ) {
         // third param should not be escaped, because it's used in the grep match rule
-        "downloads+=(\"nxf_s3_download s3:/${Escape.path(path)} ${Escape.path(targetName)}\")"
+        "downloads+=(\"nxf_retry nxf_s3_download s3:/${Escape.path(path)} ${Escape.path(targetName)}\")"
     }
```
pditommaso: It should be an opt-in feature. Enabled when the attempts > 0.

modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/S3Helper.groovy
```diff
@@ -43,6 +43,40 @@ class S3Helper {
             unset IFS
         }
         
+        nxf_retry() {
+            local max_attempts=\${ATTEMPTS-4}
+            local timeout=\${TIMEOUT-1}
```
pditommaso: The values for these settings should be provided by: 
  * `aws.batch.maxTransferAttempts` 
  * `aws.batch.maxTransferTime` (or transferTimeout?)

modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/S3Helper.groovy
```diff
@@ -43,6 +43,40 @@ class S3Helper {
...
+            local attempt=0
+            local exitCode=0
+            echo "Downloading" \$2
+
+            while [[ \$attempt < \$max_attempts ]]
+            do
```
pditommaso: Number comparison in Bash syntax is `(( x < y ))`. See https://stackoverflow.com/a/18668580/395921

modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/S3Helper.groovy
```diff
@@ -43,6 +43,40 @@ class S3Helper {
             unset IFS
         }
         
+        nxf_retry() {
```
pditommaso: Let's call it `nxf_s3_retry`.

modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/S3Helper.groovy
```diff
@@ -43,6 +43,40 @@ class S3Helper {
...
+            local max_attempts=\${ATTEMPTS-4}
+            local timeout=\${TIMEOUT-1}
+            local attempt=0
+            local exitCode=0
+            echo "Downloading" \$2
+
+            while [[ \$attempt < \$max_attempts ]]
+            do
+              if "\$@"
```
pditommaso: Ummm, this must be tested with paths containing one or more blanks.

modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/S3Helper.groovy
```diff
@@ -43,6 +43,40 @@ class S3Helper {
...
+                then
+                  return 0
+              else
+                exitCode=\$?
+              fi
+              echo \$exitCode
+              if [[ \$exitCode == 0 ]]
+              then
+                echo "Download completed" \$2
```
pditommaso: I would avoid debugging messages on the stdout, it can break task expected output. Let's only keep messages that report an error on the `stderr` channel.

modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/S3Helper.groovy
sivkovic: I would maybe name it `aws.batch.sleepBetweenAttempts` instead of `aws.batch.maxTransferTime` if you agree.

modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/S3Helper.groovy
sivkovic: Fixed

modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/S3Helper.groovy
sivkovic: I renamed it.

modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/S3Helper.groovy
sivkovic: I actually tried it previously with s3 paths containing blanks and it didn't work. Should the paths be escaped in input files?

modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/S3Helper.groovy
sivkovic: Removed all echos.

modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/AwsBatchFileCopyStrategy.groovy
sivkovic: Enabled only when you specify more then 1 attempt.

modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/S3Helper.groovy
pditommaso: What about instead `aws.batch.delayBetweenAttempts` ?

sivkovic: 👍

modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/S3Helper.groovy
sivkovic: Just exporting the `IFS=$'\n'` same as you did in `nxf_s3_upload`, solved the problem.
I tested paths without blanks, having blank spaces in directory name, and having blank spaces in filenames, and all of them worked.

pditommaso: Want to see passing the tests [this](https://github.com/nextflow-io/nextflow/blob/754da16fc4cd15dd1392f65fc1969b264b9cd949/tests/complex-names.nf) :D

sivkovic: Is this the expected result?
```
[warm up] executor > awsbatch
WARN: Config setting `executor.awscli` has been deprecated -- Use instead `aws.batch.cliPath`
executor >  awsbatch (2)
[14/a9897e] process > foo [100%] 1 of 1 ✔
[e0/afd4d6] process > bar [100%] 1 of 1 ✔
Hello world!

Completed at: 03-Jun-2019 16:43:08
Duration    : 2m 31s
CPU hours   : (a few seconds)
Succeeded   : 2
```

pditommaso: Looks promising, tho still failing the unit tests. Tip: run `make test` locally before pushing the commit. It will check the unit tests (tho not the integration tests)

sivkovic: Oh, I missed some tests. I'll fixed them and let you know. Thanks for tip.

sivkovic: I fixed all tests related to AWS Batch, and when I run `make test` all the tests pass, but for some reason `GooglePipelinesTaskHandlerTest` fails on travis. Do you have any idea what could be the problem. I run `GooglePipelinesTaskHandlerTest` locally and all of them are passing.

pditommaso: Seems related to this changes https://github.com/nextflow-io/nextflow/commit/331c1bc4ba3e3afc314a880caab5382f35404b61. But weird it happens only on travis. Have you updated recently your branch with the master? I think you are missing the last commit.

sivkovic: That was the problem. I didn't update it with master for a while. Travis tests are passing now.

modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/AwsBatchFileCopyStrategy.groovy
```diff
@@ -93,7 +93,12 @@ class AwsBatchFileCopyStrategy extends SimpleFileCopyStrategy {
...
     String stageInputFile( Path path, String targetName ) {
         // third param should not be escaped, because it's used in the grep match rule
-        "downloads+=(\"nxf_s3_download s3:/${Escape.path(path)} ${Escape.path(targetName)}\")"
+        if (opts.maxTransferAttempts > 1) {
+            "downloads+=(\"nxf_s3_retry nxf_s3_download s3:/${Escape.path(path)} ${Escape.path(targetName)}\")"
+        }
+        else {
+            "downloads+=(\"nxf_s3_download s3:/${Escape.path(path)} ${Escape.path(targetName)}\")"
+        }
     }
```
pditommaso: Instead of using an `if` branch, the code would result more readable writing something like: 

```
def cmd = opts.maxTransferAttempts > 1 ? "nxf_s3_retry nxf_s3_download" : "nxf_s3_download"
"downloads+=(\"$cmd s3:/${Escape.path(path)} ${Escape.path(targetName)}\")"
```

sivkovic: Sure. I changed also. Also I thought it might be worth adding unit test for this, but now some test from HttpFilesTests failed after I added the new one. Locally they are still passing. Do you have any idea why this is happening now?

pditommaso: I think it was a temporary network hiccup, relaunching it the build tests are green

pditommaso:
The PR look ok, thanks a lot. To proceed with the merge I need a commit with your sign-off.

pditommaso:
Thanks. One last thing, could you please add a entry in the documentation regarding the new setting options. See [config.rst](https://github.com/nextflow-io/nextflow/blob/53c94270a4772e74639ab50eb380ee40b327f459/docs/config.rst#L512) line 512.

pditommaso:
OK, merged on *testing* branch. Thanks a lot for your contribution!

Issue #1190, opened 2019-06-14
Support for custom machine types when running at GCP


mozack:
## New feature

Support for custom machine types when running at GCP.

When cloud.instanceType = "custom", use the cpus and memory directives from the process definition to request a custom machine type with the specified resources.

## Usage scenario 

This will allow more efficient (i.e. less expensive) usage of resources at GCP

## Suggest implementation 

Please see pull request

mozack:
One additional note: Resource requests must match those allowable by GCP.

i.e.
"Above 1, the instance vCPU count must be even, such as 2, 4, 6, 8, 10, and so on."
"By default, the memory per vCPU of a custom machine type must be between 0.9 GB and 6.5 GB per vCPU, inclusive."
"The total memory for a custom machine type must be a multiple of 256 MB"

See: https://cloud.google.com/compute/docs/instances/creating-instance-with-custom-machine-type

This PR does not support extended memory.

pditommaso:
The custom instance type needs to be created beforehand or it's managed automatically by the GCP api?

mozack:
It is managed automatically by the GCP API.  No need for any additional setup.

When a request is issued, instead of passing along a standard instance name (i.e. _**n1-standard-1**_), we instead pass a custom name specifying the # cpus and memory (i.e. **_custom-2-8192_** for 2 cpus and 8GB), with the cpus and memory pulled from the process directives.

We are implementing a pipeline where some processes require 1 cpu and others require 18 cpus, so this is an important enhancement for us.

pditommaso:
This is interesting, would not make sense to use this approach as the default then? is there any drawback?

mozack:
We could make this the default, but still support ```cloud.instanceType``` if that's desirable.  I am aware of 2 potential drawbacks:

1) Potential loss of backwards compatibility for projects that don't currently have a ```cloud.instanceType``` value specified.  Even this case should still be OK in most cases as we will simply use the process directives instead of n1-standard-1.

2) Projects with cpu/memory directives specified that don't meet the allowable resources for GCP will fail.

Both of those should be easily correctable by the end user.

pditommaso:
[...]

PR #1191, opened 2019-06-14
Added support for custom GCP instance types using the cpus and memory…

mozack:
… directives from the process definition.

See issue #1190

Signed-off-by: Lisle E. Mose <lisle.mose@gmail.com>


mozack:
Withdrawing this pull request and will replace with another.

PR #1198, opened 2019-06-20
GCP dynamic machine type support

mozack:
Added support for custom GCP instance types using the cpus and memory directives from the process definition.
Default to 1GB of RAM per cpu at GCP when memory directive not specified
Added optional machineType process directive which can be used to directly specify the GCP machine type (overrides the process directives)

See issue #1190

Signed-off-by: Lisle E. Mose <lisle.mose@gmail.com>
Signed-off-by: Lisle E. Mose <lmose@email.unc.edu>

### DIFF
--- a/docs/google.rst
+++ b/docs/google.rst
@@ -276,7 +276,6 @@ Create a ``nextflow.config`` file in the project root directory. The config must
 
 * Google Pipelines as Nextflow executor i.e. ``process.executor = 'google-pipelines'``.
 * The Docker container images to be used to run pipeline tasks e.g. ``process.container = 'biocontainers/salmon:0.8.2--1'``.
-* The Google Compute Engine machine instance type i.e. ``cloud.instanceType = 'n1-standard-1'``.
 * The Google Cloud `project` to run in e.g. ``google.project = 'rare-lattice-222412'``.
 * The Google Cloud `region` or `zone`. You need to specify either one, **not** both. Multiple regions or zones can be
   specified by separating them with a comma e.g. ``google.zone = 'us-central1-f,us-central-1-b'``.
@@ -304,7 +303,7 @@ machine type with the specified compute resources.  If ``memory`` is not specifi
 
 The process ``machineType`` directive may optionally be used to specify a predifined Google Compute Platform `machine type <https://cloud.google.com/compute/docs/machine-types>`_
 If specified, this value overrides the ``cpus`` and ``memory`` directives.
-If the ``cpus`` and ``memory`` directives are used, the values must comply with the allowed custom machine type `specifications <https://cloud.google.com/compute/docs/instances/creating-instance-with-custom-machine-type#specifications>`_  Extended memory is not directly supported, however high memory or cpu predefined
+If the ``cpus`` and ``memory`` directives are used, the values must comply with the allowed custom machine type `specifications <https://cloud.google.com/compute/docs/instances/creating-instance-with-custom-machine-type#specifications>`_ .  Extended memory is not directly supported, however high memory or cpu predefined
 instances may be utilized using the ``machineType`` directive
 
 Examples::
@@ -312,12 +311,20 @@ Examples::
     process custom_instance_using_cpus_and_memory {
         cpus 8
         memory '40 GB'
+
+        """
+        <Your script here>
+        """
     }
 
     process predefined_instance_overrides_cpus_and_memory {
         cpus 8
         memory '40 GB'
         machineType 'n1-highmem-8'
+
+        """
+        <Your script here>
+        """
     }
 
 Pipeline execution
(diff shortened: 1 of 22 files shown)


modules/nf-google/src/main/nextflow/cloud/google/pipelines/GooglePipelinesTaskHandler.groovy
```diff
@@ -109,11 +108,34 @@ class GooglePipelinesTaskHandler extends TaskHandler {
...
+    }
+
+    String getProcessMachineType(String cloudInstanceType, String taskMachineType, int cpus, MemoryUnit memory) {
+
+        String machineType = taskMachineType ?: cloudInstanceType
+
+        if (machineType == null) {
+            long megabytes = memory != null ? memory.mega : cpus*1024
+            machineType = 'custom-' + cpus + '-' + megabytes
         }
```
pditommaso: While it makes sense to have the memory requirement to be related to the number of cpus, nextflow does not do this for other executors. For consistency, it would be better to fallback to 1Gb irrespective the number of cpus. The user should be responsible to provide a specific value.

modules/nf-google/src/main/nextflow/cloud/google/pipelines/GooglePipelinesTaskHandler.groovy
```diff
@@ -109,11 +108,34 @@ class GooglePipelinesTaskHandler extends TaskHandler {
...
+
+        log.trace "[GPAPI] Task: $task.name - Instance Type: $machineType"
+
+        return machineType
+    }
+
+    String getProcessMachineType(String cloudInstanceType, String taskMachineType, int cpus, MemoryUnit memory) {
+
+        String machineType = taskMachineType ?: cloudInstanceType
+
```
pditommaso: A message should be printed when using `cloud.instanceType` warning the user that the setting is deprecated should use `process.machineType` instead.

pditommaso: You can use `log.warn1` for this, see for example [here](https://github.com/nextflow-io/nextflow/blob/b54dc046a0b0238bae6fbcc52d231bd77cf4b9f4/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy#L554)` 

The `warn1` is an extension of `warn` that print only the very first occurrence of the same message.

modules/nf-google/src/main/nextflow/cloud/google/pipelines/GooglePipelinesTaskHandler.groovy
mozack: This makes sense in the abstract and I can move forward with this if you like as I agree the user should specify the desired memory requirements.  

However, the GCP API by default requires at least 0.9 GB per CPU.  So, anything above 1 cpus will simply fail when the API request is issued under this scenario.

Let me know if you'd still like me to proceed with this change.

pditommaso: OK, in that case, I agree that makes sense to accommodate the requirements of the platform. Please put a note in the google pipelines docs regarding this.

mozack: Updated docs.

modules/nf-google/src/main/nextflow/cloud/google/pipelines/GooglePipelinesTaskHandler.groovy
mozack: Added warn1 message

mozack:
> Also, please `machineType` in the list of [valid directives](https://github.com/nextflow-io/nextflow/blob/987382accf2eb6a79bc212151d4ad8700ce93261/modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy#L46).

This was in the original PR.  Let me know if you are seeing otherwise.

> Finally, it would be great if you could update the docs for [process directives(https://github.com/nextflow-io/nextflow/blob/bff6c168eac75b5234e3616bdf64173a25257b61/docs/process.rst#L1150) and [google pipelines](https://github.com/nextflow-io/nextflow/blob/d57348700a08123cba28e75ab793cccaf725cb43/docs/google.rst#L252) accordingly.

Docs have been updated.

mozack:
Please let me know if you see any other issues.  The JDK 11 integration failure does not appear to be related to this PR.

docs/google.rst
```diff
@@ -302,6 +296,36 @@ Example::
...
+
+        """
+        <Your script here>
+        """
+    }
+
+    process predefined_instance_overrides_cpus_and_memory {
+        cpus 8
+        memory '40 GB'
+        machineType 'n1-highmem-8'
```
pditommaso: I think it would be clear just putting `machineType` here, to stress that's an alternative setting. I can do if you agree.

mozack: Sounds good, thanks!

PR #1216, opened 2019-07-05
Add megabyte unit to slurm memory config

rernst:
The admins on my HPC have configured SLURM with the default_gbytes option set. This will mean that the --mem is interpreted as gigabyte instead of the default megabyte. This behavior can be overruled by supyling a unit. To be compatible with both configurations it is best that nextflow supplies a unit (M) when setting memory requirements.

See https://slurm.schedmd.com/sbatch.html under --mem.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/SlurmExecutor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/SlurmExecutor.groovy
@@ -61,7 +61,7 @@ class SlurmExecutor extends AbstractGridExecutor {
             // be stored, just collected). In both cases memory use is based upon the job's
             // Resident Set Size (RSS). A task may exceed the memory limit until the next periodic
             // accounting sample. -- https://slurm.schedmd.com/sbatch.html
-            result << '--mem' << task.config.getMemory().toMega().toString()
+            result << '--mem' << task.config.getMemory().toMega().toString() + 'M'
         }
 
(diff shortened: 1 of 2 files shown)


pditommaso:
Could you provide some reference documentation as support of this change?

rernst:
> Could you provide some reference documentation as support of this change?

I have updated my first message with a bit more information. Would you like to describe the 'fix' in the code as well?

pditommaso:
I see. Thanks for that. it's critical to understand if the `M` unit has been always supported of it's a new feature. Potential this can break other installations.

rernst:
I don't think this will break other installations as the SLURM documentation clearly states that it is possible to supply a unit to the mem option (see below). I also think it is better / clearer to supply an unit as it makes sure that nextflow jobs are always interpret in the correct way regardless of any (future) configuration possibilities. 

From the SLURM documentation (https://slurm.schedmd.com/sbatch.html):
> --mem=<size[units]>
Specify the real memory required per node. Default units are megabytes unless the SchedulerParameters configuration parameter includes the "default_gbytes" option for gigabytes. Different units can be specified using the suffix [K|M|G|T].

pditommaso:
It looks a fair assumption https://twitter.com/Fwippy/status/1148387362319155200

rernst:
Thanks for effort @twitter and approving the pull request!

pditommaso:
Thanks finding this out and contributing to the project!

Issue #1223, opened 2019-07-11
K8s sometimes throws error "invalid pod status" when submitting many pods


bentsherman:
### Expected behavior and actual behavior

When submitting many jobs using k8s executor, some pods may not be scheduled immediately due to insufficient resources. In this case, nextflow should be able to wait while the pod is scheduled. However, recently I've noticed that if I increase the parallelism such that many pods are pending, nextflow throws an error which I traced to this line of code:

https://github.com/nextflow-io/nextflow/blob/b49e6d74dae782def6be2b087d06fdd8fa6b9291/modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy#L260

Here is the status object that nextflow tried to parse:
```
      "status": {
          "phase": "Pending",
          "qosClass": "Guaranteed"
      }
```

Looking at the code I can see that nextflow was expecting to see a `conditions` array which is missing. I suppose the assumption was that this conditions array would be initialized when the phase was set to `Pending`, however my experience suggests otherwise.

Although I'm not an expert on k8s internals, I think we should simply modify this `podState` function to not fail if the `conditions` array is missing:
```groovy
        if( status?.phase == 'Pending' ) {
            if( status.conditions instanceof List ) {
                final allConditions = status.conditions as List<Map>
                final cond = allConditions.find { cond -> cond.type == 'PodScheduled' }
                if( cond.reason == 'Unschedulable' ) {
                    def message = "K8s pod cannot be scheduled"
                    if( cond.message ) message += " -- $cond.message"
                    //def cause = new K8sResponseException(resp)
                    log.warn1(message)
                }
            }
            // undetermined status -- return an empty response
            return Collections.emptyMap()
        }

        throw new K8sResponseException("K8s invalid pod status (missing container status)", resp)
```

### Steps to reproduce the problem

Run a workflow which submits many more pods than the capacity of the k8s cluster, such that several pods are pending.

### Environment 

* Nextflow version: 19.04.0
* Java version: 1.8
* Operating system: Linux

olifly:
This bug had been cropping up fairly regularly for us and I had made a quick fix, but hadn't submitted it upstream til now: https://github.com/nextflow-io/nextflow/pull/1242

bentsherman:
Fixed in 19.07.0.

PR #1242, opened 2019-07-22
Fix: "invalid pod status" when running scrips with the kubernetes executor.

olifly:
Fix for issue: https://github.com/nextflow-io/nextflow/issues/1223

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
@@ -259,7 +259,7 @@ class K8sClient {
             return Collections.emptyMap()
         }
 
-        throw new K8sResponseException("K8s invalid pod status (missing container status)", resp)
+        throw new K8sResponseException("K8s Can't determine pod state from status returned by K8s API ", resp)
     }
 
     protected void checkInvalidWaitingState( Map waiting, K8sResponseJson resp ) {
--- a/modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
@@ -244,14 +244,16 @@ class K8sClient {
             return state
         }
 
-        if( status?.phase == 'Pending' && status.conditions instanceof List ) {
-            final allConditions = status.conditions as List<Map>
-            final cond = allConditions.find { cond -> cond.type == 'PodScheduled' }
-            if( cond.reason == 'Unschedulable' ) {
-                def message = "K8s pod cannot be scheduled"
-                if( cond.message ) message += " -- $cond.message"
-                //def cause = new K8sResponseException(resp)
-                log.warn1(message)
+        if( status?.phase == 'Pending' ){
+            if( status.conditions instanceof List ) {
+                final allConditions = status.conditions as List<Map>
+                final cond = allConditions.find { cond -> cond.type == 'PodScheduled' }
+                if( cond.reason == 'Unschedulable' ) {
+                    def message = "K8s pod cannot be scheduled"
+                    if( cond.message ) message += " -- $cond.message"
+                    //def cause = new K8sResponseException(resp)
+                    log.warn1(message)
+                }
             }
             // undetermined status -- return an empty response
             return Collections.emptyMap()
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
```diff
@@ -244,14 +244,16 @@ class K8sClient {
...
+            if( status.conditions instanceof List ) {
+                final allConditions = status.conditions as List<Map>
+                final cond = allConditions.find { cond -> cond.type == 'PodScheduled' }
+                if( cond.reason == 'Unschedulable' ) {
+                    def message = "K8s pod cannot be scheduled"
+                    if( cond.message ) message += " -- $cond.message"
+                    //def cause = new K8sResponseException(resp)
+                    log.warn1(message)
+                }
             }
```
pditommaso: I'm not understanding why splitting the `if` should be different.

olifly: @pditommaso - You need a little more of the code that's surrounding this to see why than github is showing in this pull request.

The issue is that Kubernetes will sometimes return the state of a Pod so that phase == 'Pending', but it returns a null value for the conditions since there are not conditions yet (the conditions key is not set in the Map returning a null).  Kubernetes will eventually return the conditions List as soon as there are some conditions in it.

Before this pull request:

```
if( status?.phase == 'Pending' && status.conditions instanceof List ) {
            final allConditions = status.conditions as List<Map>
            final cond = allConditions.find { cond -> cond.type == 'PodScheduled' }
            if( cond.reason == 'Unschedulable' ) {
                def message = "K8s pod cannot be scheduled"
                if( cond.message ) message += " -- $cond.message"
                //def cause = new K8sResponseException(resp)
                log.warn1(message)
            }
            // undetermined status -- return an empty response
            return Collections.emptyMap()
        }

        throw new K8sResponseException("K8s invalid pod status (missing container status)", resp)
    }
```

That is, when this issue crops up (the phase is Pending, but there are no conditions yet (a null value instead of List,) then Nextflow will skip all of the if statement logic and throw a K8SReponseException aborting the run.

After this pull request:

```
if( status?.phase == 'Pending' ){
            if( status.conditions instanceof List ) {
                final allConditions = status.conditions as List<Map>
                final cond = allConditions.find { cond -> cond.type == 'PodScheduled' }
                if( cond.reason == 'Unschedulable' ) {
                    def message = "K8s pod cannot be scheduled"
                    if( cond.message ) message += " -- $cond.message"
                    //def cause = new K8sResponseException(resp)
                    log.warn1(message)
                }
            }
            // undetermined status -- return an empty response
            return Collections.emptyMap()
        }

        throw new K8sResponseException("K8s invalid pod status (missing container status)", resp)
```

That is, if phase is Pending, but there are no conditions, then we return an undetermined status.

Only if we reach the end of the procedure and haven't been able to determine the status do we return an K8sResponseException.

I see that I've not changed the message the final K8sResponseException contains so that it's midleading.  I'll fix that and update this pull request with it

pditommaso: Got it thanks

modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
```diff
@@ -244,20 +244,22 @@ class K8sClient {
...
             // undetermined status -- return an empty response
             return Collections.emptyMap()
         }
 
-        throw new K8sResponseException("K8s invalid pod status (missing container status)", resp)
+        throw new K8sResponseException("K8s Can't determine pod state from status returned by K8s API ", resp)
     }
```
pditommaso: Frankly, I think the previous err message was a bit more informative.

olifly: @pditommaso The previous message did not reflect why the Exception was being thrown anymore.  The problem is not missing container status, but rather that K8s returned a set of phase and conditions that the code can not determine the pod state (end of function without any resolvement).

I can revert the change to the message, or insert a more informative message of your choice :) 


_I'm really bad at constructing helpful error message btw :)_

pditommaso: Fair enough, what about `K8s undermined status conditions for pod $podName` ?

olifly: > Fair enough, what about `K8s undermined status conditions for pod $podName` ?

Sounds good to me.  Will update pull request.

pditommaso:
Thanks for this contribution.

Issue #860, opened 2018-09-07
AWS Batch: nextflow does not respect AWS_PROFILE variable


dtenenba:
## Bug report 

(If you a reporting a bug please follow this template replacing the text between parentheses with the requested information)

### Expected behavior and actual behavior

I have multiple AWS profiles in my `~/.aws/credentials` and `~/.aws/config` files, and none are
set as the default, by design (this reduces the possibility of doing something in the wrong account by accident). When I want to work with one profile, I do:

```
export AWS_PROFILE=profilename
```

in the shell I'm working in. This always works fine with the AWS CLI, boto3, etc. 
It does not work with nextflow (see error text in next section).

If I back up my `~/.aws` files and create one that has only a single (default) profile,
the job starts fine.

`AWS_PROFILE` is documented here: 
https://docs.aws.amazon.com/cli/latest/userguide/cli-environment.html

### Program output 

```
$ nextflow run main.nf -w s3://bucket-name-censored/dtenenba/nextflow-tests/
N E X T F L O W  ~  version 0.31.1
Launching `main.nf` [jovial_mestorf] - revision: 2522c5d504
ERROR ~ Access Denied (Service: Amazon S3; Status Code: 403; Error Code: AccessDenied; Request ID: 71E3656AC03CA663; S3 Extended Request ID: PX3KdFIF219oli9mWNLHqOHucyl9/SUwSsLNXPdtdtHshxGVO7mXqxhrMV8JfaTdw5SqxK79i90=)

 -- Check '.nextflow.log' file for details

```

### Steps to reproduce the problem

1. Back up and then remove `~/.aws/config` and `~/.aws/credentials`
2. run `aws configure --profile=nondefault`. Provide valid AWS credentials/region.
3. run `export AWS_PROFILE=nondefault`
4. run `aws s3 ls` just to prove that the AWS CLI respects the `AWS_PROFILE` variable.
5. run any nextflow job that uses AWS batch, you'll see output similar to the above.

### Environment 

* Nextflow version: 0.31.1
* Java version: 1.8.0_101
* Operating system: macOS 10.13.6

pditommaso:
This a kind of duplicate of #423. I was not aware of `AWS_PROFILE` variable, it should be quite straighforward to support it.

PR #1248, opened 2019-07-25
Enable use AWS profiles other than `default`

nkrumm:
Hi Nextflow team--

This PR addresses #860 (respecting AWS_DEFAULT_PROFILE env variable).

AWS has actually two different environmental variables that can specify the profile to be used. Historically `AWS_DEFAULT_PROFILE` was used but current documentation seems to suggest `AWS_PROFILE` is more correct: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html. 

The current PR first looks for `AWS_PROFILE` in the env, then `AWS_DEFAULT_PROFILE` and then falls back to `default` if none are set.

(Edits: separated this PR from PR for AWS temporary session credentials)


pditommaso:
Hi, thanks a lot for your PR. Just a few things: 

* I would propose to keep this for the temporary session and open a separate PR for the default profile (of the other way around as you prefer). 
* Please rename `sessionKey` to `sessionToken` to make it more readable. 
* Have a look to the [DCO](https://github.com/apps/dco) and sign it accordingly 

> I am not a java developer at all and so much of this may not be the "right way"

No problem, code looks very good.

nkrumm:
Hi @pditommaso, thanks for taking a look. 

This PR now just address #1248 (I will open a second PR for the temporary session credentials). Agree that `sessionKey` should be `sesionToken`, so that is updated now. I also added a few bits of documentation.

Tests aren't passing right now, but I'm not sure that it's related to this PR?

PR #1253, opened 2019-07-30
Tiny docs fixes

robsyme:
Before this commit, the example showed the foo() process being passed the output of bar(). I might be mistaken, but the example looks like the output of foo() should be passed to bar(<here>) rather than the other way around - foo() doesn't take any input.

While we're looking at this section, we might as well show the bar() process doing something with the input file.


pditommaso:
Nice, thanks Rob.

Issue #1255, opened 2019-08-01
K8sTaskhandler never sets it status to TaskStatus.RUNNING


olifly:
### Expected behavior and actual behavior

**Expected:** When a pod starts running a task the K8sTaskHandler should change the state to _RUNNING_.

**Actual:** The K8sTaskHandler checkIfRunning never sets the status to Running, but just returns true if the task is running.

This behaviour is not consistent with the behaviour of other task handlers and has the side effect of letting the TaskMonitor send process_started message to all observations each time it check task statuses until the tasks complete. 

### Steps to reproduce the problem

* Run a workflow with the k8s executor with weblog enabled to see all of the process_started messages sent for every task that is running each time the TaskMonitor check the status of the tasks (default poll time is 5 seconds for the k8s executor.

* Check the nextflow log when running a workflow with the k8s executor.  All of the tasks that are listed by _dumpPendingTasks_ have SUBMITTED status even though they are running.

### Environment 

* Nextflow version: 19.07.0
* Java version: Java 8
* Operating system: macOS

### Additional context

(Add any other context about the problem here)

pditommaso:
Thanks for the patch.

PR #1256, opened 2019-08-01
K8s Taskhandler updates its status correctly to 'RUNNING'

olifly:
fix for issue: https://github.com/nextflow-io/nextflow/issues/1255


modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sResponseException.groovy
```diff
@@ -32,6 +32,7 @@ class K8sResponseException extends Exception {
 
     K8sResponseException(K8sResponseJson response) {
         super(msg0(response))
+        this.response = response
     }
```
pditommaso: Can you add a test for this change?

olifly: @pditommaso - No problem :)

PR #1257, opened 2019-08-01
Docs fix: collectFile `size` vs `skip`

robsyme:
The docs previously suggested the `size` parameter would control how many lines were skipped, but that parameter doesn't exist. I think that that `skip` parameter does this job instead.


pditommaso:
You are right! thanks!

pditommaso:
DLS2 stuff is merged on master, consider pushing PR directly on that branch in the future.

robsyme:
Will do. Thanks Paolo

Issue #1283, opened 2019-08-20
K8S - volumeClaim pod directive/config creates multiple volumes for a single pvc


olifly:
**Expected behaviour:** 

Multiple *volumeClaim* pod directives/config for the same pvc, but mounted at different paths and/or have different subpaths should create 1 volumeMount per directive, but only only 1 volume per pvc name.

Given the following volumeClaim directives:

```
[volumeClaim: 'first', mountPath: '/work']
[volumeClaim: 'first', mountPath: '/work2', subPath: '/bar']
[volumeClaim: 'second', mountPath: '/data', subPath: '/foo']
[volumeClaim: 'second', mountPath: '/data2', subPath: '/fooz']
``` 

should result in the following pod mounts:

```
volumeMounts:[
    [name:'vol-1', mountPath:'/work'],
    [name:'vol-1', mountPath:'/work2', subPath: '/bar'],
    [name:'vol-2', mountPath:'/data', subPath: '/foo'],
    [name:'vol-2', mountPath:'/data2', subPath: '/fooz']
]
```

and the following volumes:

```
volumes:[
    [name:'vol-1', persistentVolumeClaim:[claimName:'first']],
    [name:'vol-2', persistentVolumeClaim:[claimName:'second']] ]
```

**Actual behaviour:**

The behaviour as it is today results in the following volumeMounts and volumes:

```
volumeMounts:[
    [name:'vol-1', mountPath:'/work'],
    [name:'vol-2', mountPath:'/work2', subPath: '/bar'],
    [name:'vol-3', mountPath:'/data', subPath: '/foo'],
    [name:'vol-4', mountPath:'/data2', subPath: '/fooz']
]
```
```
volumes:[
    [name:'vol-1', persistentVolumeClaim:[claimName:'first']],
    [name:'vol-2', persistentVolumeClaim:[claimName:'first']],
    [name:'vol-3', persistentVolumeClaim:[claimName:'second']],
    [name:'vol-4', persistentVolumeClaim:[claimName:'second']] ]
```

*Having multiple volumes for the same pvc results in the pod not being scheduled.*

### Steps to reproduce the problem

Run a workflow using the k8s executor that has processes configured to use the same volumeClaim at different paths and/or subpaths.

### Environment 

* Nextflow version: 19.07.0
* Java version: 1.8
* Operating system: macOS Mojave

pditommaso:
Solved by #1284

PR #1284, opened 2019-08-20
K8S - One volume per pvc name

olifly:
Only create a volume once for every pvc name (specifying the same pvc…in many volumes results in the pod not being able to be schedules by k8s).

Fix for issue: https://github.com/nextflow-io/nextflow/issues/1283


olifly:
@pditommaso - Any chance you could take a look at this PR? :) 

Have other bugfixes to submit, but want to submit/discuss/approve/reject one at a time :)

pditommaso:
Hi Olafur, thanks for submitting this PR. Unfortunately not yet, this week we are very busy with the NF conference in Barcelona (you will got a mention!). I'll check next week or more likely the following one.

olifly:
> Sorry for the late reply. However, I'm not understanding why there could be a duplicate claim name. Could you please elaborate a bit your use case?

Sure :)

We have a PVC for a huge NFS volume that we are mounting at different paths/subpaths in order to limit access to data and provide consistant paths to data between runs

Example claims are:

```
[volumeClaim: 'pvc-data-nfs', mountPath: '/data', subPath: '<path to the data on the nfs being processes>'],
[volumeClaim: 'pvc-data-nfs', mountPath: '/work', subPath: '<path on the nfs for the nextflow work folder'],
[volumeClaim: 'pvc-data-nfs', mountPath: '/project', subPath: '<path on the nfs for the project we're publishing result into>']
```

Here we are mounting the same pvc, but at different mountPaths for different subPaths.  This does not work before the fix since nextflow will create three volumes referencing the same pvc, and that is something that Kubernetes doesn't like.

After the fix only one volume will be created.

olifly:
> I see, makes sense. Can I just a minor change to use dot property notation instead of array syntax for consistency? eg `it.persistentVolumeClaim.claimName` instead of `it['persistentVolumeClaim']['claimName']` of `volume.name` instead of `volume['name']`. Tx!

No problem though it means removing the @CompileStatic annotation from the class.

pditommaso:
> No problem though it means removing the @CompileStatic annotation from the class or adding @CompileDynamic to the method 

Ouch, I was not thinking at that. What if instead we use the claim name to create the volume handle. it would become much cleaner. eg 

```
            final name = "vol-${entry.claimName}"
            final claim = [name: name, mountPath: entry.mountPath ]
```

What do you think?

olifly:
> > No problem though it means removing the @CompileStatic annotation from the class or adding @CompileDynamic to the method
> 
> Ouch, I was not thinking at that. What if instead we use the claim name to create the volume handle. it would become much cleaner. eg
> 
> ```
>             final name = "vol-${entry.claimName}"
>             final claim = [name: name, mountPath: entry.mountPath ]
> ```
> 
> What do you think?

It makes the code much clearer yes although I will have to change a few additional unit tests because the name of volumes referencing volumeClaims will change :)

We still require @CompileDynamic on the build function or use bracket syntax because we still need to check if the volume has already been defined.
```
volumes.find {it['name'] == name}
or
volumes.find {it.name == name}
```

pditommaso:
(Sorry, travelling. I'll check asap)

pditommaso:
OK, I've merged this manually with a few changes. Thanks a contributing and a sorry for the late reply.

PR #1323, opened 2019-09-25
Adding ENDPOINT support for cn regions.

lch14forever:
Signed-off-by: Chenhao Li <lichenhao.sg@gmail.com>


pditommaso:
Thanks a lot! It should be enough. Have tested it to verify it solves the problem?

lch14forever:
I have tested it within cn-northwest-1 and was able to spin up a cluster with this. cn-north-1 should also work as Beijing is also in the csv file.

PR #1324, opened 2019-09-26
Fix #1195

### LINKED ISSUE #1195: Environment activation fails with recent versions of Conda

If the user tries to execute a Nextflow pipeline that contains processes which run in conda environments, the pipeline fails with:

```
Command wrapper:
  .command.run: line 266: activate: No such file or directory
```

* Ensure that you do not already have a conda environment active.
* Ensure that a recent version of the `conda` utility is installed (tested with 4.6.12 and 4.6.14)
* Try to run any Nextflow pipeline that contains a process with a `conda` directive.

### Environment 

$ nextflow info
  Version: 19.04.1 build 5072
  Modified: 03-05-2019 12:29 UTC (14:29 CEST)
  System: Linux 4.4.0-116-generic
  Runtime: Groovy 2.5.6 on OpenJDK 64-Bit Server VM 1.8.0_212-8u212-b03-0ubuntu1.16.04.1-b03
  Encoding: UTF-8 (UTF-8)

$ conda --version
conda 4.6.12

### Additional context

The Anaconda project has recently been working on changing the way in which conda environments are activated and deactivated. The `activate` and `deactivate` shell scripts are no longer put in the user's path by default — *only* the `conda` utility itself is in the user's path.

Activating a conda environment is now no longer done by sourcing the `activate` shell script, but with `conda activate`.

Nextflow constructs a `.command.run` script that tries to activate the conda environment using `source activate`, which fails due to the `activate` script not being in the path.

It should be possible to fix by replacing `source activate` with `conda activate`.

grst:
This addresses #1195. Works for me on SGE now. 

It still requires some input & discussion: 

- I apologize if this looks ugly, it is the first piece of Groovy I wrote and I'm afraid it looks too much like Java :/

- I am unsure how to adapt the test for that. In python, one can mock a shell command output, but I didn't find anything similar for Groovy. Probably it's best to delegate the `conda info --json` to a configuration option and stub that from the test. I am, however, too unfamiliar with the code-base to understand where that should go. `CondaConfig.groovy`?

- This approach would not work if the worker nodes and the head node don't share the same path to the conda installation. In that case, running `conda info` would need to be delegated to the worker nodes. I'm afraid that this would make everything more complicated. This would, of course, still require that the conda binary is in the `PATH` there.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -22,7 +22,6 @@ import java.nio.file.Path
 import groovy.transform.CompileStatic
 import groovy.transform.PackageScope
 import groovy.util.logging.Slf4j
-import groovy.json.JsonSlurper
 import nextflow.container.ContainerBuilder
 import nextflow.container.DockerBuilder
 import nextflow.container.ShifterBuilder
@@ -316,19 +315,7 @@ class BashWrapperBuilder {
     }
 
     private String getCondaActivateSnippet() {
-        if ( condaEnv ) {
-            // Get path to conda installation using `conda info`
-            def sout = new StringBuilder()
-            Process proc = "conda info --json".execute()
-            proc.waitForProcessOutput(sout, System.err)
-            JsonSlurper jsonSlurper = new JsonSlurper()
-            Map condaInfo = (Map) jsonSlurper.parseText(sout.toString())
-            // Build absolute path to activate script
-            def activateScript = "${condaInfo.conda_prefix}/bin/activate"
-            return "# conda environment\nsource ${Escape.path(activateScript)} ${Escape.path(condaEnv)}\n"
-        } else {
-            return null
-        }
+        condaEnv ? "# conda environment\nsource activate ${Escape.path(condaEnv)}\n" : null
     }
 
     protected String getTraceCommand(String interpreter) {
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -315,7 +315,9 @@ class BashWrapperBuilder {
     }
 
     private String getCondaActivateSnippet() {
-        condaEnv ? "# conda environment\nsource activate ${Escape.path(condaEnv)}\n" : null
+        // Extract the "conda_prefix" entry from JSON
+        def condaPrefixSubCmd = 'conda info --json | sed  -n \'s/^.*"conda_prefix":\\s*"\\(.\\+\\)".*$/\\1/p\''
+        condaEnv ? "# conda environment\nsource \$($condaPrefixSubCmd)/bin/activate ${Escape.path(condaEnv)}\n" : null
     }
 
     protected String getTraceCommand(String interpreter) {
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -22,6 +22,7 @@ import java.nio.file.Path
 import groovy.transform.CompileStatic
 import groovy.transform.PackageScope
 import groovy.util.logging.Slf4j
+import groovy.json.JsonSlurper
 import nextflow.container.ContainerBuilder
 import nextflow.container.DockerBuilder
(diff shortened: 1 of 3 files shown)


grst:
Hi, 

thanks for following up. No, that was not done on purpose, I just don't know the code-base well enough. 

I agree that it would be perfect to execute it once on each worker node. That would ensure that it still works in the (probably rare) case that the login node and the worker nodes don't share the file system and have different paths to the conda binary.

pditommaso:
Is should be even easier than. The `conda info` command should be added in the template generated by nextflow.

grst:
Just to get it right: 
I would still put it here
https://github.com/nextflow-io/nextflow/blob/0fcb0ed890055f2cf3d7be213496da3759f20dec/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy#L318

but not execute the `conda info` from within groovy, but let it execute in the bash script?

pditommaso:
Likely, but it is supposed to help? Running `conda info` I see 

```
$ conda info

     active environment : None
       user config file : /Users/pditommaso/.condarc
 populated config files : /Users/pditommaso/.condarc
          conda version : 4.5.11
    conda-build version : 2.0.6
         python version : 2.7.12.final.0
       base environment : /Users/pditommaso/miniconda2  (writable)
           channel URLs : https://conda.anaconda.org/conda-forge/osx-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://conda.anaconda.org/bioconda/osx-64
                          https://conda.anaconda.org/bioconda/noarch
                          https://repo.anaconda.com/pkgs/main/osx-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/free/osx-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/r/osx-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/pro/osx-64
                          https://repo.anaconda.com/pkgs/pro/noarch
          package cache : /Users/pditommaso/miniconda2/pkgs
                          /Users/pditommaso/.conda/pkgs
       envs directories : /Users/pditommaso/miniconda2/envs
                          /Users/pditommaso/.conda/envs
               platform : osx-64
             user-agent : conda/4.5.11 requests/2.18.4 CPython/2.7.12 Darwin/18.5.0 OSX/10.14.4
                UID:GID : 501:20
             netrc file : None
           offline mode : False
```

A second problem is that it adds ~ 1 sec of overhead to run it to each task, not so good.

grst:
have a look at
```
conda info --json | grep conda_prefix
```
it's not shown without `--json`.

pditommaso:
Then also `conda info -a | grep CONDA_ROOT` work, isn't it?

grst:
I guess so. Interesting that the names are inconsistent.

rpetit3:
Hi @pditommaso and @grst 

Are there plans to take this PR forward? If not, what would be recommended way forward? I would like to help out.

I have a uncommon use case where this would be useful. See PR https://github.com/bioconda/bioconda-recipes/pull/19402

Thanks!

pditommaso:
I'm lost on which is the current *portable* alternative to former `source active <recipe>` command. Any suggestion from conda gurus @johanneskoester @rpetit3 ?

grst:
At least in Snakemake, it's done like that: 
(1) get the `conda_prefix` from `conda info --json`: [conda.py#L357](https://github.com/snakemake/snakemake/blob/937c9a4b923fecb79522c4c687f09a10640a7a5a/snakemake/deployment/conda.py#L357)
(2) run `source ${conda_prefix}/bin/activate ${env}`: [conda.py#L414-L425](https://github.com/snakemake/snakemake/blob/937c9a4b923fecb79522c4c687f09a10640a7a5a/snakemake/deployment/conda.py#L414-L425)

pditommaso:
Thanks for having a look at this. OK, therefore this is the way to go. The goal is to refactor your PR, that is 1 and 2 commands as Bash one-liner. Willing to give it a try?

grst:
The regex for parsing the conda info is a bit of a monster, but it appears to work now. 
At least for me, the `activate` command is found.

pditommaso:
> The regex for parsing the conda info is a bit of a monster, but it appears to work now.

Yep, and above all does not work on Mac. I guess `awk` could make a better job here.

pditommaso:
This looks better 

```
conda info --json | awk '/conda_prefix/ { gsub(/"|,/, "", $2); print $2 }'
```

pditommaso:
Merged this PR manually. Thanks for contributing.

Issue #1289, opened 2019-08-23
Broken container environment when enabling metrics


cimendes:
### Expected behavior and actual behavior

When updating to the lastest release (19.07.0), all [flowcraft's](https://github.com/assemblerflow/flowcraft/) generated pipelines suddenly stopped working. Regardless of the process, they all terminate with bash errors (ps, uname, grep and date not found)

After some time, I traced the error to the `nxf_container_env()` in the `.command.run` file. All flowcraft generated pipelines use the same [nextflow.config](https://raw.githubusercontent.com/assemblerflow/flowcraft/master/flowcraft/generator/templates/nextflow.config) file, that includes the env scope. 
For some reason, the paths defined there show up with escaped quotes. 

```nxf_container_env() {
cat << EOF
export PYTHONPATH=\"\$baseDir/templates:\$PYTHONPATH\"
export PATH=\"/home/ines/temp/flowcraft_test/v19/bin:\$baseDir/templates:\$PATH\"
EOF
}
```
When manually removing those "\", i'm able to execute the `.command.run` with no issues. 
So far any alteration to alter this through the nextflow.config fileby changing how the paths are declared has failed. :/

I would expect for them to show up in the `.nextflow.run` file as
```
nxf_container_env() {
cat << EOF
export PYTHONPATH="\$baseDir/templates:\$PYTHONPATH"
export PATH="/home/ines/temp/flowcraft_test/v19/bin:\$baseDir/templates:\$PATH"
EOF
}
````

### Steps to reproduce the problem

I've created a simple 1 component (fastqc) nextflow pipeline for testing purposes, and is available here: https://github.com/cimendes/nextflow_issue 
Test data is not included

### Program output 

from the `.command.log`:
```
/home/ines/temp/flowcraft_test/v19/work/78/2250444327559d6f2b9e8e8c8df4db/.command.run: line 186: uname: command not found
/home/ines/temp/flowcraft_test/v19/work/78/2250444327559d6f2b9e8e8c8df4db/.command.run: line 138: grep: command not found
/home/ines/temp/flowcraft_test/v19/work/78/2250444327559d6f2b9e8e8c8df4db/.command.run: line 139: grep: command not found
/home/ines/temp/flowcraft_test/v19/work/78/2250444327559d6f2b9e8e8c8df4db/.command.run: line 200: date: command not found
Command 'ps' required by nextflow to collect task metrics cannot be found
```

### Environment 

* Nextflow version: 19.07.0.5106
[...]

PR #1327, opened 2019-10-02
shifterimg workaround for ensuring image pulls

t0rrant:
For some unknown reason shifterimg returns before finishing,
it should wait, even from a non-interactive shell.
Added some code from the previous helper function

Signed-off-by: Manuel Torrinha <manuel.torrinha@tecnico.ulisboa.pt>

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/container/ShifterBuilderTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/container/ShifterBuilderTest.groovy
@@ -59,6 +59,12 @@ class ShifterBuilderTest extends Specification {
         then:
         cli ==  '''
                 shifterimg pull ubuntu:14
+                shifterimg lookup ubuntu:14
+                while ! shifterimg lookup ubuntu:14; do
+                    sleep 5
+                    STATUS=$(shifterimg -v pull ubuntu:14| awk -F: '$0~/"status":/{gsub("[\", ]","",$2);print $2}')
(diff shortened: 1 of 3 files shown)


t0rrant:
WIP
running tests locally

t0rrant:
Fixing #1289 may affect the outcome of these tests

modules/nextflow/src/test/groovy/nextflow/container/ShifterBuilderTest.groovy
```diff
@@ -57,26 +57,44 @@ class ShifterBuilderTest extends Specification {
         when:
         def cli = new ShifterBuilder('ubuntu:14').build().getRunCommand()
         then:
-        cli ==  '''
+        cli.stripIndent().trim() ==  '''
                 shifterimg pull ubuntu:14
```
pditommaso: This is not a good idea. The command string should be produced without indentation instead of removing it to make the test pass! 😬

t0rrant: Of course you are right :smile:
The truth is I was just having trouble with testing what indentation comes out of the getRunCommand, and I thought that in this case indentation was not really an issue.

I will submit a new change shortly :)

pditommaso:
Better, but the build fails now https://travis-ci.org/nextflow-io/nextflow/builds/595571233

t0rrant:
yes, I see that I still have issues with indentation, now testing locally, have 2 more spaces from the generated command than I was expecting. Probably have to use `stripIndent` in the command generation as well, will try to generate a full batch file in a pipeline to make sure it is what we need.

PR #1341, opened 2019-10-21
Repository error 404 handling

piotr-faba-ardigen:
Adds dedicated error handling for 404 error. 
Use case:
Perform `nextflow pull` on repository that exists (and credentials in scm configuration file are correct), but the API is different for some reason.

```
nextflow pull test_proj/test_repo -hub test_hub
```

Then the current message displayed to user is:

```
Cannot find `test_proj` -- Make sure exists a BitBucketServer repository at this address `https://bitbucket.server.com/scm/test_proj/test_repo`
```

This is misleading when dealing with unsupported repository provider, as we know this repository exists, so the information is of no use in such case.

The new message indicates clearer what the source of error is:
```
API Uri not found. The uri that was used for the query: https://bitbucket.biscrum.com/rest/api/1.0/projects/test_proj/repos/test_repo/raw/main.nf
```


pditommaso:
OK, I've merge this manually [02cf0a55]. Thanks!

PR #1382, opened 2019-11-18
Handle ProviderMismatchException in CacheHelper

lindenb:
I had a java.nio.file.ProviderMismatchException which is not catched in CacheHelper because IOException is not a parent class of ProviderMismatchException

The following PR had a catch so the error will be displayed.

```
Nov-18 14:54:18.058 [Actor Thread 20] ERROR nextflow.processor.TaskProcessor - Error executing process > 'mergeControls (merge contols)'

Caused by:
  java.nio.file.ProviderMismatchException

java.nio.file.ProviderMismatchException: null
	at sun.nio.fs.UnixPath.toUnixPath(UnixPath.java:200)
	at sun.nio.fs.UnixFileSystemProvider.getFileAttributeView(UnixFileSystemProvider.java:115)
	at sun.nio.fs.LinuxFileSystemProvider.getFileAttributeView(LinuxFileSystemProvider.java:68)
	at sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:144)
	at sun.nio.fs.LinuxFileSystemProvider.readAttributes(LinuxFileSystemProvider.java:99)
	at java.nio.file.Files.readAttributes(Files.java:1737)
	at nextflow.util.CacheHelper.hashFile(CacheHelper.java:216)
	at nextflow.util.CacheHelper.hasher(CacheHelper.java:167)
	at nextflow.util.CacheHelper.hasher(CacheHelper.java:159)

```

### DIFF
--- a/modules/nf-commons/src/main/nextflow/util/CacheHelper.java
+++ b/modules/nf-commons/src/main/nextflow/util/CacheHelper.java
@@ -216,8 +216,11 @@ public class CacheHelper {
         try {
             attrs = Files.readAttributes(path, BasicFileAttributes.class);
         }
-        catch(IOException | ProviderMismatchException e) {
+        catch(IOException e) {
             log.debug("Unable to get file attributes file: {} -- Cause: {}", FilesEx.toUriString(path), e.toString());
+        }
+        catch(ProviderMismatchException e) {
+            log.warn("File system is unable to get file attributes file: {} -- Cause: {}", FilesEx.toUriString(path), e.toString());
         }
 
         if( mode==HashMode.DEEP && attrs!=null && attrs.isRegularFile() )
--- a/modules/nf-commons/src/main/nextflow/util/CacheHelper.java
+++ b/modules/nf-commons/src/main/nextflow/util/CacheHelper.java
@@ -220,6 +220,7 @@ public class CacheHelper {
             log.debug("Unable to get file attributes file: {} -- Cause: {}", FilesEx.toUriString(path), e.toString());
         }
         catch(ProviderMismatchException e) {
+            // see https://github.com/nextflow-io/nextflow/pull/1382
             log.warn("File system is unable to get file attributes file: {} -- Cause: {}", FilesEx.toUriString(path), e.toString());
         }
 
--- a/modules/nf-commons/src/main/nextflow/util/CacheHelper.java
+++ b/modules/nf-commons/src/main/nextflow/util/CacheHelper.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.ProviderMismatchException;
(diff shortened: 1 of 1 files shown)


modules/nf-commons/src/main/nextflow/util/CacheHelper.java
```diff
@@ -215,7 +216,7 @@ static private Hasher hashFile( Hasher hasher, Path path, HashMode mode ) {
         try {
             attrs = Files.readAttributes(path, BasicFileAttributes.class);
         }
-        catch(IOException e) {
+        catch(IOException | ProviderMismatchException e) {
             log.debug("Unable to get file attributes file: {} -- Cause: {}", FilesEx.toUriString(path), e.toString());
         }
```
pditommaso: Thanks for sending this PR. It makes sense to capture this, tho maybe it could be useful to report a warning to the user because it's a quite unusual error and can signal something bad. What it's causing this error on your side?

lindenb: @pditommaso 

> What it's causing this error on your side?

I whish I knew :cry:

lindenb: @pditommaso so instead of `log.debug` , you wish I use` log.warn` for this kind of error ?

pditommaso: Yes, `log.warn` even tho this can spam the program output. 

> I whish I knew

Usually, this error happens when composing two `Path` objects belonging to different file systems. For example: 

```
s3File = file('s3://data/there')
localFile = file('/data/here')
localFile.resolve(localFile)
```

pditommaso: note: `file(something)` is just a wrapper around java `Paths.get(something)`

lindenb: @pditommaso 
> Usually, this error happens when composing two Path objects belonging to different file systems.

yes, it could be my problem but it's always the very same file among other files that are created the same way. For now, I solved this problem by making a copy of the file instead of of using the symllink created by NF...

lindenb:
pushed with signed-off

pditommaso:
Sorry to bother, but it should be as the following other it's not recognised, eg 

```
Signed-off-by: Random J Developer <random@developer.example.org>
```

lindenb:
@pditommaso pushed with the `Signed-off-by:` token in my commit message:

https://github.com/lindenb/nextflow/commit/95e81537e3397e84fc9e4799f33383b45edf1a63

is it a wrong syntax ?

pditommaso:
That should be fine, but seems it's only in your fork. Maybe you need to update the PR otherwise the bot don't see it.

lindenb:
I updated the PR

pditommaso:
For some reason, it was still not happy .. anyhow, no pb. Thanks Pierre!

PR #1389, opened 2019-11-26
Feature: Kubernetes emptyDir pod directive/config

olifly:
Adds support for a pod config/directive that configures an emptyDir volume mount for processes (https://kubernetes.io/docs/concepts/storage/volumes/#emptydir).

Extremely useful for provisioning in memory temporary filesystem for processes.


pditommaso:
Sorry for the late reply. In principle, in-mem scratch storage should already be possible using `process.scratch = 'ramdisk'`. 

https://github.com/nextflow-io/nextflow/blob/21088ad3e5890d4ecdb0f8e17b75f6d5b9e99ba6/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy#L129-L131

What's the advantage of K8s `emptyDir`?

lepsalex:
For our (ICGC-ARGO) use, having `emptyDir` would allow us to direct temporary file I/O to the disk local to the container instead of going through our network storage. Given the `readWriteMany` requirement we are using NFS backed volumes and having the `emptyDir` directive available would help us increase throughput on certain workflows by as much as 60%. I guess this PR is a little out of date now but it would be immensely useful to us, and I imagine to anyone running Nextflow in Kubernetes to have this option available.

pditommaso:
Still not getting the benefits to NF built-in feature `process.scratch = 'ramdisk'`.

lepsalex:
When using NF in Kubernetes we need a way of mounting a directory from the node's local disk drive to the pod. In the docs you mention that the `scratch` directive is useful for exactly the reason I am stating. The difference is that when NF creates the pod definition that gets launched in Kubernetes we need a way of specifying the local disk mount (in addition to the NFS-backed persistent volumes).

Also, we are specifically interested in mounting physical disk drives local to the Kubernetes Node not RAM as disk.

`emptyDir` would allow us to do that in addition to having the ability to "auto-cleanup" as its contents is cleared once the process pod is removed.

olifly:
@pditommaso - Hi and sorry for the really REALLY  late reply.  Been busy the last months implementing linear regression scripts in nextflow.

EmptyDir gives you advantages over process.scratch = 'ramdisk'.


Since emptyDir is a kubernetes declaration mean that the kubernetes scheduler is aware of the requirement is can take that into consideration when dispatching a pod to a node.

The kubernetes scheduler will also terminate the pod if it exceeds the memory limit specified.

Using a file emptyDir allows the nextflow pods to exchange data with sidecars that are injected into them, for example to consume secrets provided by the hashicorp vault injector.

In short, using emptyDir in kubernetes in sead of process.scratch plays better with the kubernetes encironment and scheduler :)

Best Regards,
Ólafur.

bentsherman:
@pditommaso I concur with @olifly, `emptyDir` has some nice k8s-specific properties that you don't get from `process.scratch`.

I was considering whether we should rewire `process.scratch` to use `emptyDir` behind the scenes in the case of k8s, but I think they are two different ways to acquire scratch space. `emptyDir` has more options and occurs at a different level than `process.scratch`. Better I think to just incorporate this feature as is. Allowing users to specify an `emptyDir` is no different in principle from allowing a `config`, `secret`, or `volumeClaim`.

bentsherman:
Closing in favor of #2998

PR #1409, opened 2019-12-09
Channel from sql

helios:
The initial idea is to have a channel that emits the results generated by a query on a RDBMS, in my case it is PostgreSQL. Each "row" corresponds to a Map object, converted at runtime. I decided to drop for now the ability to access to the internal data structures offered by groovy-sql for semplicity and sability.

In a Nextflow script 

```
server="dbServer/IP"
database="myDb"
query = "SELECT * FROM myTable"
// myTable has 4 columns (as example) library_id, path to fastq file, md5sum, sha256sum;

connectionInfo = ['dbUrl': "jdbc:postgresql://${server}/${database}",
                  'dbUser': 'guest',
                  'dbPassword': 'password',
                  'dbDriver': 'org.postgresql.Driver']


//Example with closure using just 2 of the 4 columns, the resuls here is the aggregation by library_id in order to get for all the library_id(s) all the fastqs associated to each of them
Channel.fromDBSql(connectionInfo, query) {
    [it.library_id, it.path]
}
.groupTuple()

// use the plain output
Channel.fromDBSql(connectionInfo, query)
.subscribe {
println it
}
```

The user can "grab" the connector library for database directly from the CLI 
`NXF_GRAB='org.postgresql:postgresql:42.2.8' nextflow run main.nf`


modules/nextflow/src/main/groovy/nextflow/database/DBSql.groovy
```diff
@@ -0,0 +1,87 @@
...
+
+import groovy.util.logging.Slf4j
+import groovy.sql.Sql
+import groovyx.gpars.dataflow.DataflowQueue
+import groovyx.gpars.dataflow.DataflowWriteChannel
+import nextflow.Channel
+import nextflow.extension.CH
+
+@Slf4j
+class DBSql {
```
pditommaso: It would be better to annotate this as `@CompileStatic` https://stackoverflow.com/a/29611007/395921

pditommaso:
Thanks for taking this initiative. This is a feature we discussed a lot with @fstrozzi and @emi80, it would be nice to have it implemented. 

One thing that would be nice to have is to hide the database configuration and keep it in the nextflow config file. In this way, we could have an abstraction for the datasource along the same way for executors. 

For example: 

```nextflow

Channel
  .fromSql('select * from foo', datasource: 'something')
  .somethingElse()
```

then in the config, it could be possible to have 

```
datasources {
   something {
    driver = 'org.postgresql.Driver'
    url = 'etc' 
    user = 'foo'
    password = 'secret' 
  }
}
```
 
We could even have a `default` datasource that is implicitly used when it's not specified in the `fromSql` method. 

Final note: what is the closure is doing in your example?

helios:
> 
> 
> Thanks for taking this initiative. This is a feature we discussed a lot with @fstrozzi and @emi80, it would be nice to have it implemented.

Thanks! 

> 
> One thing that would be nice to have is to hide the database configuration and keep it in the nextflow config file. In this way, we could have an abstraction for the datasource along the same way for executors.
> 
> For example:
> 
> ```nextflow
> Channel
>   .fromSql('select * from foo', datasource: 'something')
>   .somethingElse()


I will change `fromDBSql` to `fromSql` which is more "readable"


> then in the config, it could be possible to have
> 
> ```
> datasources {
>    something {
>     driver = 'org.postgresql.Driver'
>     url = 'etc' 
>     user = 'foo'
>     password = 'secret' 
>   }
> }
> ```
> 
> We could even have a `default` datasource that is implicitly used when it's not specified in the `fromSql` method.

I really like this suggestion. How do I access the config? Digging into the source code I found `Global.session.config.`CONFIG_KEY_NAME so something like `Global.session.config.datasources.something` is ok ? 
Where do you validate the configuration ?

> 
> Final note: what is the closure is doing in your example?

unluky example without a context. I just updated the original message

modules/nextflow/src/main/groovy/nextflow/database/DBSql.groovy
helios: Got it.

pditommaso:
>  How do I access the config? Digging into the source code I found Global.session.config.CONFIG_KEY_NAME so something like Global.session.config.datasources.something is ok ?
Where do you validate the configuration ?

Yes, the `Global.session.config` allows you to access the resolved configuration object. Then you can access each entry by the key like any other `Map` object. The `navigate` method allows you to traverse the nested keys returning a default when missing, for example:

https://github.com/nextflow-io/nextflow/blob/e523c51cd4f29e7ceff07a1c096246bd36632345/modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/AwsOptions.groovy#L80-L84

helios:
> > How do I access the config? Digging into the source code I found Global.session.config.CONFIG_KEY_NAME so something like Global.session.config.datasources.something is ok ?
> > Where do you validate the configuration ?
> 
> Yes, the `Global.session.config` allows you to access the resolved configuration object. Then you can access each entry by the key like any other `Map` object. The `navigate` method allows you to traverse the nested keys returning a default when missing, for example:
> 
> https://github.com/nextflow-io/nextflow/blob/e523c51cd4f29e7ceff07a1c096246bd36632345/modules/nextflow/src/main/groovy/nextflow/cloud/aws/batch/AwsOptions.groovy#L80-L84

With https://github.com/helios/nextflow/blob/channel_from_sql/modules/nextflow/src/main/groovy/nextflow/database/DBSql.groovy#L59 the parameters for the connection can be placed into `nextflow.config` under 
```
datasources {
   'default' { 
        dbUser = '...'
        dbUrl = '...'
        dbPassword = '...'
        dbDriver = '...'
    }
}
```
I still need a validation method, in the meantime

> For example:
> 
> ```Channel
>  .fromSql('select * from foo', datasource: 'something')
>  .somethingElse()
> ```

works

helios:
I had some problem about portability about missing JDBC libraries. I am going to include some of the as `gradle` dependencies.

helios:
> I had some problem about portability on missing JDBC libraries. I am going to include some of the as `gradle` dependencies.

Addes JDBC drivers for PostgreSQL; SQLite; MySQL; MariaDB; Ignite

Ignite was a requirement for another database, honestly I did not debug which one required it because I was getting a `SEVERE: Failed to resolve default logging config file: config/java.util.logging.properties` without it.

pditommaso:
Umm, not sure that makes sense to include all JDBC driver libs, at least not in the default distribution.

helios:
> 
> 
> Umm, not sure that makes sense to include all JDBC driver libs, at least not in the default distribution.

Which are the possible solution, feasible at the moment ?
My first option would be to use a dynamic way to load the library using `Grab` but I failed. If you have some experience on that I really would like to implement a dynamic approach.

pditommaso:
I need to investigate how to make a system that allow users to plug their own extensions and libs. For now you can keep as it is.

pditommaso:
Closing this PR, since it has been integrated in the branch https://github.com/nextflow-io/nextflow/tree/nf-sqldb. With some luck it should be out relatively soon 😅 Thanks for your patience!

helios:
Thanks for accepting the idea, maybe a Christmas gift ;)

Issue #1408, opened 2019-12-09
Task parameters ordering may invalidate task cache


emorice:
### Expected behavior and actual behavior

Adding a blank line between processes definition (not inside process definitions), thus changing nothing but line numbers in the main script, causes the processes after to be flagged as changed, under some specific circumstances. It would be expected that the hash of these tasks do not change and `-resume` simply get them from the cache.

### Steps to reproduce the problem

Get a recent nextflow release :
```
wget https://get.nextflow.io -O nextflow
chmod +x nextflow
```
and create an example pipeline `test.nf` containing exactly (do not change a single character):
```
process A {
	output:
		file 'a' into chan
	"""
		echo 'foo' > a
	"""
}

prog = ''
var = 'c'

process B {
        output:
                file('b') into result


        """
	echo $prog xxx > b ; touch $var

        """
}
```
Process A is actually irrelevant, but I left it as a control process to highlight the problem with B.
Then run the pipeline two or three time:

Run 1, initial run, behave as expected:
```
./nextflow run test.nf -resume
```
and runs both tasks:
```
N E X T F L O W  ~  version 19.10.0
Launching `test.nf` [suspicious_koch] - revision: e33cd08ab1
WARN: It appears you have never run this project before -- Option `-resume` is ignored
executor >  local (2)
[d3/9eddc1] process > A [100%] 1 of 1 ✔
[5a/6f88e3] process > B [100%] 1 of 1 ✔
```
Optional: Run 2, exact rerun, behaves as expected:
```
./nextflow run test.nf -resume
```
both processes are cached and recovered:
```
N E X T F L O W  ~  version 19.10.0
Launching `test.nf` [marvelous_shannon] - revision: e33cd08ab1
[d3/9eddc1] process > A [100%] 1 of 1, cached: 1 ✔
[5a/6f88e3] process > B [100%] 1 of 1, cached: 1 ✔
```

Run 3, single blank line change, fails to hit cache:
Change the `test.nf` file and insert a single blank line after the `var` declaration, so that it now looks like this:
```
process A {
	output:
		file 'a' into chan
	"""
		echo 'foo' > a
	"""
}

prog = ''
var = 'c'


process B {
        output:
                file('b') into result


        """
	echo $prog xxx > b ; touch $var

        """
}
```
and run again:
```
./nextflow run test.nf -resume
```
process A is correctly recovered, process B isn't though nothing has changed but line numbers:
```
N E X T F L O W  ~  version 19.10.0
[...]

PR #1419, opened 2019-12-12
Task hash independence from variable order

emorice:
Proposed fix and unit test for #1408


emorice:
Seems ok.

pditommaso:
Thanks for this pull request. I've merged manually with some changes.

PR #1474, opened 2020-01-27
Update google.rst to clarify location v. region

moschetti:
Add details and links which clarify the difference between location (Cloud Life Sciences API location) and region (Compute Engine region).


docs/google.rst
```diff
@@ -456,6 +460,7 @@ Example::
 
     google {
         project = 'your-project-id'
+        location = 'europe-west2'
         zone = 'europe-west1-b'
```
pditommaso: Then, the `location` could be omitted from the example snippet. Do you agree?

docs/google.rst
```diff
@@ -572,6 +577,7 @@ For example::
...
```
pditommaso: Same as before.

docs/google.rst
moschetti: agreed. I'll remove the example snippets but leave the definition part and submit an updated pr

pditommaso: thanks!

Issue #540, opened 2017-11-30
Process prolog and epilog


pditommaso:
The goal of this enhancement is to add two new process directives, namely: `prolog` and `epilog` that would make it possible to add to process task a common *script* prefix and suffix. 

These are different from [beforeScript](https://www.nextflow.io/docs/latest/process.html#beforescript) and [afterScript](https://www.nextflow.io/docs/latest/process.html#afterscript) directives that are designed mainly for custom task configuration and, above all, are execute in the task wrapper context, that can be a separated environment from the task execution one (in particular when using containers).  

Instead `prolog` content should be prepended to the user script just after the shebang declaration and `epilog` should appended to the user task script.

pditommaso:
This can be a little harder than expect. The basic idea was to prepend/append the user script with the `prolog` and `epilog` snippets when defined. However, provided the `prolog` and `epilog` should be BASH code, that would corrupt non-BASH user commands. 

It cannot either included in the command launcher ie. `.command.run`, because that would not allow to execute it when the user command is executed in a container image.

ODiogoSilva:
> It cannot either included in the command launcher ie. .command.run, because that would not allow to execute it when the user command is executed in a container image.

This is due to the issue of not adding the `bin` directory to the path when using a container image? Is there any particular reason to prevent both from being using in a process?

In my case, I need to execute some bash scripts (which are in the `bin` directory) in the work directory before (and after) some processes (I originally though that the `epilog` and `prolog` were meant to allow the execution of such scripts, which wouldn't have impact on the user commands/templates). This is regardless of the processes running in a container image or not. In fact, I made a small tweak in nextflow to allow the usage of both `bin` and container images in the same process and it has been working fine. Maybe I'm missing some use cases where they would conflict somehow?

pditommaso:
No, this happens because it's the `.command.run` that launch the container, therefore it cannot execute the prolog and the epilog. 

One solution, that I would like to avoid it to use an intermediate wrapper that contains the `prolog`, run the command and finally executes the `epilog` 

A better alternative, could be to embed the user `prolog` and `epilog` in a couple of variables in the `.command.run` via heredoc declaration, then append the container execution command line, but there could be tricky side effects on special characters expansion or in the max length of the command line.

ewels:
Bump

abhi18av:
[...]

PR #1511, opened 2020-02-26
Execute beforeScript directive after setting the task environment.

manuelesimi:
This PR makes it possible to execute a script inside the bin folder in the beforeScript directive. The change is consistent with the behavior of the afterScript directive that already allows that.


pditommaso:
Inclined to not merge this. The `beforeScript` it's not meant to allow access to the task env 9for example when using the container execution, this it's not even possible. 

Moreover, changing it can break existing code that use it to load the singularity runtime. 

What you are looking for it's meant to be implemented by #540.

manuelesimi:
I see your point. However, afterScript allows to access the task environment... it's a bit odd the different behavior between the two directives. Of course, I don't have a big picture of the impact this change would have if merged. 

prolog and epilog would fit my needs, but as far as I understand, they are not available yet.

PR #1578, opened 2020-04-23
fixes for accelerator docs

KevinSayers:
Fixed some minor issues with the accelerator directive docs. The anchor from the TOC of directives was linking to Google docs due to the link in the tip box.


pditommaso:
Nice! Thanks Kevin.

PR #1589, opened 2020-04-30
Fix bugs in GA4GH TES support

tskir:
As part of the [CINECA](https://www.cineca-project.eu/) project we plan to use Nextflow with [GA4GH TES](https://github.com/ga4gh/task-execution-schemas), a simple standardised execution schema. This pull request fixes three critical problems in Nextflow's current support of TES:

###  1. Specify `type` field explicitly for all standard inputs and outputs
* Nextflow creates several standard inputs and outputs for each submitted task.
* For all of them, the `type` field is always set to `null`.
* TES allows this parameter to be either _FILE_ or _DIRECTORY_, but [does not allow it to be null](https://github.com/ga4gh/task-execution-schemas/blob/9cc12b0c215a7f54fdeb0d0598ebc74fa70eb2a7/openapi/task_execution.swagger.yaml#L261).
* This PR fixes it by explicitly setting the type of all inputs and outputs as `FILE`.

### 2. Update docs to mention that specifying a Docker image name is mandatory when running with TES
* If Docker image is not specified, the `executor.image` field in Nextflow's request will be `null`, hence the request will [fail to comply](https://github.com/ga4gh/task-execution-schemas/blob/9cc12b0c215a7f54fdeb0d0598ebc74fa70eb2a7/openapi/task_execution.swagger.yaml#L217) with TES specification.
* This PR fixes it by mentioning that specifying a Docker image is mandatory when running with TES.

### 3. Clarify that the backend URL must not include a trailing slash
* If the URL is specified with a trailing slash, e. g. `http://back.end.com/`, Nextflow will append `/v1/tasks` to it, and the resulting URL will be `http://back.end.com//v1/tasks` with double slashes.
* Some major implementations of TES, for example [TESK](https://github.com/EMBL-EBI-TSI/TESK), will treat that as a non-normalized URL, and the task submission will fail with a somewhat cryptic error message.
* This PR fixes it by mentinoning this behaviour in the docs.

### DIFF
--- a/docs/executor.rst
+++ b/docs/executor.rst
@@ -334,9 +334,9 @@ To use this feature define the following variables in the workflow launching env
     export NXF_EXECUTOR_TES_ENDPOINT='http://back.end.com'
     
 
-Then you will be able to run your workflow over TES using the usual Nextflow command line, i.e.::
+Then you will be able to run your workflow over TES using the usual Nextflow command line. Be sure to specify the Docker image to use, i.e.::
 
-    nextflow run rnaseq-nf
+    nextflow run rnaseq-nf -with-docker alpine
 
 .. note:: If the variable ``NXF_EXECUTOR_TES_ENDPOINT`` is omitted the default endpoint is ``http://localhost:8000``.
 
--- a/docs/executor.rst
+++ b/docs/executor.rst
@@ -334,7 +334,11 @@ To use this feature define the following variables in the workflow launching env
     export NXF_EXECUTOR_TES_ENDPOINT='http://back.end.com'
     
 
-Then you will be able to run your workflow over TES using the usual Nextflow command line. Be sure to specify the Docker image to use, i.e.::
+It is important that the endpoint is specified without the trailing slash; otherwise, the resulting URLs will be not
+normalized and the requests to TES will fail.
+
+Then you will be able to run your workflow over TES using the usual Nextflow command line. Be sure to specify the Docker
+image to use, i.e.::
 
     nextflow run rnaseq-nf -with-docker alpine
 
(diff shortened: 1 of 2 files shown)


pditommaso:
Thanks a lot for submitting this PR. While the rationale of these changes I fear would not help to solve the main compatibility issue of NF with the current version of TES specs. 

Nextflow input/output declarations leave intentionally unspecified the type of path object that can be either a `file` or a `directory` which it's resolved at runtime.  

I've raised years ago [this issue](https://github.com/ga4gh/task-execution-schemas/issues/76) along with the support for [glob pattern](https://github.com/ga4gh/task-execution-schemas/issues/77). 

With these, the Nextflow support for TES is inherently broken.

tskir:
@pditommaso I realise that this is only a start of the path, and that it will not be possible to achieve full support of TES in Nextflow without ultimately modifying TES itself in some way.

However, what we're looking for right now in CINECA is to start running at least the simpler workflows. I feel with that with this PR we're going from the state where TES support is fully broken, as all task creation requests do not comply with the schema, to the one where we support at least some simpler workflows which will not be using directories as either inputs or outputs.

pditommaso:
Fair enough, it sounds a good plan.

lvarin:
I can check it, but not now. At the moment I have not too much time to dedicate to this, sorry. 

It was some time ago, so my memory can be clouded, but one thing I remember that we needed to do a workaround for, was file parameters. In order to pass files to TESK, we needed to use the `file://`protocol and share a volume in the k8s deployment. Normally you will use `http://` or `ftp://` for "sending" files to TESK. I am not sure this helps.

Anyways, the Slack channel is a good way to contact the TESK developers, and I saw your message there already.

PR #1615, opened 2020-05-30
Clone/Pull submodules recursively

pachiras:
Hello,

I found `run` command fails to download the codes of submodule of submodule (the grand daughter module) from the repository. `pull` and `clone` commands even fail to download the codes of submodule (the daughter module).

For most of the nextflow users need the submodules downloaded recursively when they `run` their scripts. So this is my attempt to achieve it.

version 20.05.0-SNAPSHOT build 5335
created 10-05-2020 21:32 UTC (11-05-2020 06:32 JDT)


modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
```diff
@@ -589,6 +591,7 @@ class AssetManager {
             clone
                 .setURI(cloneURL)
                 .setDirectory(localPath)
+                .setCloneSubmodules(true)
```
pditommaso, line 594: Not sure about this. Modules are updates separately, see 

https://github.com/nextflow-io/nextflow/blob/a34b1d2a3786e04c55dc94906394d4079b40c911/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy#L952-L969

pachiras: OK. But without this line, I couldn't pull the grand-daughter module, which I need. The daughter modules were OK, though.

pditommaso: Not sure how to help. We need to keep the clone of submodules as opt-in feature. My understanding is that your proposal makes it the default. My suggestion is to avoid a nested substructure in your submodules.

pachiras:
Hi, Paolo,

I have added `-recurse-submodules` options to `run/clone/pull` sub commands, as the `git` command does. Nextflow's behaviors on those commands should not be affected without the option. Please consider about it.

pachiras:
Thank you for your comment, Paolo.
I'll try to change the code as you said.
By the way our server crashed today. It could take some time for its recovery.
Please wait for a few days.

pditommaso:
No hurry!

pachiras:
Hi,

Sorry to be late, but I have changed the code to control the recursive behavior from manifest scope. I hope it is acceptable. Thanks.

pditommaso:
Nice thanks!

PR #1624, opened 2020-06-09
Fix ProviderConfig to support local repositories

### LINKED ISSUE #1492: Unable to work with local bare repos or bad documentation
When I make a local bare repo as explained in documentation, I can run from it only once.

When I run a workflow from a local bare repository it works only first time. Next time I run it says 
`Can't find any configured provider for git server `XXX` -- Make sure to have specified it in your `scm` file. For details check https://www.nextflow.io/docs/latest/sharing.html#scm-configuration-file`

```
mkdir /tmp/deleteme
cd /tmp/deleteme
git init --bare bare.git
git clone /tmp/deleteme/bare.git repo
cd repo
echo -e '#!/usr/bin/env nextflow\nprocess dummy {"sleep 1"}' > main.nf
git commit main.nf -m "dummy"
git push
cd ..
rm repo -rf
nextflow run file:/tmp/deleteme/bare.git
####THIS RUNS OK WITH SAFE WARNINGS####
nextflow run file:/tmp/deleteme/bare.git
####THIS COMPLAINS:
####Can't find any configured provider for git server `/tmp/deleteme` -- Make sure to have specified it in your `scm` file. For details check https://www.nextflow.io/docs/latest/sharing.html#scm-configuration-file
```

### Program output 

```
N E X T F L O W  ~  version 19.10.0
Can't find any configured provider for git server `/tmp/deleteme` -- Make sure to have specified it in your `scm` file. For details check https://www.nextflow.io/docs/latest/sharing.html#scm-configuration-file
```

### Environment 

* Nextflow version: 19.10.0 build 5170
* Java version: openjdk version "1.8.0_232"
* Operating system: Linux Debian 9.11
* Bash version: GNU bash, versión 4.4.12(1)-release (x86_64-pc-linux-gnu)

### Realted to
#95 Add support for local bare git repositories

emi80:
This should fix #1492

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/scm/ProviderConfig.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/ProviderConfig.groovy
@@ -110,9 +110,14 @@ class ProviderConfig {
         def p = result.indexOf('://')
         if( p != -1 )
             result = result.substring(p+3)
+        // a local file url (e.g. file:///path/to/repo or /path/to/repo)
+        // so we need to return the full path as the domain
+        if ( result.startsWith('/') )
+            return result
+        // a server url so we look for the domain without subdirectories
         p = result.indexOf('/')
         if( p != -1 )
-            result = result.substring(0,p)
+            result = result.substring(0, p)
         return result
     }
 
--- a/modules/nextflow/src/test/groovy/nextflow/scm/ProviderConfigTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/scm/ProviderConfigTest.groovy
@@ -102,7 +102,25 @@ class ProviderConfigTest extends Specification {
     def 'should return provider attributes' () {
 
         when:
-        def config = new ProviderConfig('custom',[platform: 'github', server:'http://local.host'])
+        def config = new ProviderConfig('local',[platform: 'file', path:'/local/repo'])
+        then:
+        config.name == 'local'
+        config.platform == 'file'
+        config.domain == '/local/repo'
+        config.server == null
+        config.endpoint == null
+
+        when:
+        config = new ProviderConfig('local',[platform: 'file', path:'file:///local/repo'])
+        then:
+        config.name == 'local'
+        config.platform == 'file'
+        config.domain == '/local/repo'
+        config.server == null
+        config.endpoint == null
+
+        when:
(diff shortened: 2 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/scm/ProviderConfig.groovy
```diff
@@ -111,8 +111,8 @@ class ProviderConfig {
         if( p != -1 )
             result = result.substring(p+3)
         p = result.indexOf('/')
-        if( p != -1 )
-            result = result.substring(0,p)
+        if( p > 0 )
+            result = result.substring(0, p)
         return result
```
pditommaso, line 116: I think the logic should be a more explicit ie. if it contains `://` it does the same as now, otherwise it should start with `/` ?

Also add a test here 

https://github.com/seqeralabs/nextflow/blob/ae04fc374cafb822af844d3b3479f2d52f6b7c40/modules/nextflow/src/test/groovy/nextflow/scm/ProviderConfigTest.groovy#L72-L100

emi80: About tests. they are [here](https://github.com/nextflow-io/nextflow/blob/local-repo/modules/nextflow/src/test/groovy/nextflow/scm/ProviderConfigTest.groovy#L104:L120) already

emi80: About the logic, in case of a local repository can the domain start with `file://`? If that's the case, we have to handle this too and making it explicitly might maybe end up being too verbose...

pditommaso: Don't remember this, but it could make sense to handle it.

emi80: An possiility would be:
```groovy
def result = server ?: path
def p = result.indexOf('://')
if( p != -1 )
    result = result.substring(p+3)
if ( result.startsWith('/') )
    return result
p = result.indexOf('/')
if( p != -1 )
    result = result.substring(0, p)
return result
```

pditommaso: Cool, please add some comment to explain the mess behind e.g. remote url, local url, abs paths. Tx!

PR #1633, opened 2020-06-16
Add support for selecting minimum CPU platform on GCP

hnawar:
This PR adds support to specify the minimum CPU platform used by a run when using the Google Cloud Life Sciences executor


pditommaso:
Thanks for submitting this PR. I was thinking this would make more sense at process (job) level instead of pipeline wide. What do you think?

hnawar:
I thought about that. There is currently no additional cost for specifying Skylake, Cascade lake is only available in the N2 instances so they would need to specify a different machineType all together.

The setting specifies the minimum CPU Platform, and when not specified you are still likely to end up on Skylake.

It was mainly easier for me to do it on the profile level rather than on process level, and the benefit of adding it to process level was not worth it.

I'll check with Andrew to get his opinion

hnawar:
I had a quick sanity check with our team, and I think pipeline level config should be sufficient.
Let me know if you are happy with that

pditommaso:
Ok, that's reasonable. If so it's needed to add the reading of that value from the nextflow config file. 

See for example here: 

https://github.com/nextflow-io/nextflow/blob/bc9db8d85622e77d13a95e3c94d9609ae4059368/modules/nf-google/src/main/nextflow/cloud/google/lifesciences/GoogleLifeSciencesConfig.groovy#L109-L117

Also, uni tests should be added to corresponding changes. It would be nice also to add an entry in the docs for google LS

https://github.com/nextflow-io/nextflow/blob/bc9db8d85622e77d13a95e3c94d9609ae4059368/modules/nf-google/src/main/nextflow/cloud/google/lifesciences/GoogleLifeSciencesConfig.groovy#L109-L117

Tx!

hnawar:
Thanks,
I added the cpuPlatform to the ConfigGroovy.
I ran some small test after compiling from scratch and verified it is passed properly to the API.
This is how it looks like in my profile to specify Skylake.
google.lifeSciences.cpuPlatform = 'Intel Skylake'

I've added it to the documentation as an example as it will be the most common option for most users.

I'll look at how to add unit test as well, but can sure use some help on that.

docs/google.rst
```diff
@@ -106,6 +106,7 @@ google.region                                  The Google *region* where the com
 google.zone                                    The Google *zone* where the computation is executed in Compute Engine VMs. Multiple zones can be provided separating them by a comma. Do not specify if a region is provided. See  `available Compute Engine regions and zones <https://cloud.google.com/compute/docs/regions-zones/>`_
 google.location                                The Google *location* where the job executions are deployed to Cloud Life Sciences API. See  `available Cloud Life Sciences API locations <https://cloud.google.com/life-sciences/docs/concepts/locations>`_ (default: the same as the region or the zone specified).
 google.enableRequesterPaysBuckets              When ``true`` uses the configured Google project id as the billing project for storage access. This is required when accessing data from *reqester pays enabled* buckets. See `Requester Pays on Google Cloud Storage documentation  <https://cloud.google.com/storage/docs/requester-pays>`_ (default: ``false``)
+google.lifeSciences.cpuPlatform                Set the minimum CPU Platform e.g `'Intel Skylake'` (default: none).
```
pditommaso, line 109: Nice. It would be great to include also a link to the Google documentation so that users can find avail values easily.

pditommaso:
Thanks for this changes. For the tests,it could be enough t o verify the `cpuPlatform` is taken correctly for the config, for example look here 

https://github.com/nextflow-io/nextflow/blob/bc9db8d85622e77d13a95e3c94d9609ae4059368/modules/nf-google/src/test/nextflow/cloud/google/lifesciences/GoogleLifeSciencesConfigTest.groovy#L231-L242

and then it's included in the request, look for example here 

https://github.com/nextflow-io/nextflow/blob/a481b42b152ad002eab01f8df8823e877d0e8e5b/modules/nf-google/src/test/nextflow/cloud/google/lifesciences/GoogleLifeSciencesTaskHandlerTest.groovy#L162-L182

and 

https://github.com/nextflow-io/nextflow/blob/a481b42b152ad002eab01f8df8823e877d0e8e5b/modules/nf-google/src/test/nextflow/cloud/google/lifesciences/GoogleLifeSciencesHelperTest.groovy#L133-L206

Also, please, make sure so sign-off the commit to fulfill the DCO bot requirement. Thanks!

hnawar:
I have added the unit tests, signed the commits with DCO and Travis  build passed.

hnawar:
I have also staged changes to add support for specifying disk type per process. I can push these changes as well and put all in one PR if this is more convenient or submit a separate pull request

docs/google.rst
```diff
@@ -106,6 +106,7 @@ google.region                                  The Google *region* where the com
 google.zone                                    The Google *zone* where the computation is executed in Compute Engine VMs. Multiple zones can be provided separating them by a comma. Do not specify if a region is provided. See  `available Compute Engine regions and zones <https://cloud.google.com/compute/docs/regions-zones/>`_
 google.location                                The Google *location* where the job executions are deployed to Cloud Life Sciences API. See  `available Cloud Life Sciences API locations <https://cloud.google.com/life-sciences/docs/concepts/locations>`_ (default: the same as the region or the zone specified).
 google.enableRequesterPaysBuckets              When ``true`` uses the configured Google project id as the billing project for storage access. This is required when accessing data from *reqester pays enabled* buckets. See `Requester Pays on Google Cloud Storage documentation  <https://cloud.google.com/storage/docs/requester-pays>`_ (default: ``false``)
```
pgrosu, line 108: s/reqester/requester

pditommaso: Thanks for spotting it. Fixed.

pditommaso:
As for specifying disk type per process, free free to draft a separate PR for it. However, I would like to keep a portable model for this, i.e. able to support also other computing services.

pditommaso:
Any suggestion to retrieve the list of [avail cpu platforms](https://cloud.google.com/compute/docs/instances/specify-min-cpu-platform#specifications) for a given zone using the Java API? 

Can't find an immediate way using the `com.google.api.services.compute.Compute` client.

PR #1643, opened 2020-06-22
Add support for specifying disk type for Google Cloud executor

hnawar:


hnawar:
#1643

hnawar:
I think this can be easily extended to other platforms who can support this option.
The diskType is picked by TaskConfig.groovy and ProcessConfig.groovy it will be ignored by other executors who don't have the option.
If in the future options need to be more generic then perhaps more generic names for the disk types can be used and each executor can translate to the platform equivalent

pditommaso:
Thanks for this PR, however, I'm not sure we should introduce a directive only to specify the disk type. I was thinking on the possibility to extend the current the existing `disk` directive to add the support for an optional `type` attribute eg 

```
disk 100.GB, type: 'ssd'
```

Need to check what's the impact on the existing code, but IMO the resulting syntax would be a bit more consistent.

Issue #1647, opened 2020-06-23
accelerator directive limited to GPUs in k8s


mes5k:
## New feature

The [accelerator directive](https://www.nextflow.io/docs/latest/process.html?highlight=accelerator#accelerator) is limited to using GPUs in k8s. Other hardware resources that aren't GPUs (e.g. fpga) either can't be accessed or need to be erroneously identified as GPUs. This is because of [this line](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy#L409).

## Usage scenario 

Not sure how TPUs are identified, but we definitely don't want to identify [DRAGEN fpgas](https://www.illumina.com/products/by-type/informatics-products/dragen-bio-it-platform.html) as GPUs. 

## Suggest implementation 

This is tricky depending on how much backwards compatibility we want to maintain.

First option would just be to delete [this line](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy#L409). This would obviously break pipelines that depend on this feature.

A second option would be to add another key to the accelerator map, maybe `exact_type`, that if specified overrides the `type` key.  

A third option would be to add a `dont_make_assumptions` boolean to the accelerator map that allows `type` to be used directly instead of doing all of [this stuff](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy#L406-L409).

Please weigh in on which option you'd prefer and I'll submit a patch!

pditommaso:
Can you help to make an assessment of possible values of accelerator type identifiers supported by K8s? At the time it was implemented I remember only GPU was documented and therefore that suffix was hardcoded.

mes5k:
Right, so the k8s device plugins you linked to are exactly what we're using to make the DRAGEN fpga a schedule-able resource in k8s. Device plugins are named [according to this](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/#device-plugin-registration).  We were thinking of using `illumina.com/dragen-fpga` or something similar.

[...]

PR #1649, opened 2020-06-24
Fix for feature 1647 that allows non gpu k8s accelerator resources.

mes5k:
See https://github.com/nextflow-io/nextflow/issues/1647 for details.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
@@ -403,11 +403,13 @@ class PodSpecBuilder {
 
         def type = accelerator.type ?: 'nvidia.com'
 
-        // If '/' is missing then assume we're using GPU and update as necessary.
-        if ( !type.contains('/') ) {
-            if( !type.contains('.') ) type += '.com'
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
```diff
@@ -396,17 +396,31 @@ class PodSpecBuilder {
...
+    @CompileDynamic
+    String getAcceleratorType(AcceleratorResource accelerator) {
+
+        def type = accelerator.type ?: 'nvidia.com'
+
+        // If '/' is missing then assume we're using GPU and update as necessary.
+        if ( !type.contains('/') ) {
+            if( !type.contains('.') ) type += '.com'
+            type += '/gpu'
+        }
```
pditommaso, line 410: Can i suggest: 

```
if( type.contains('/') ) 
  return type 

if( !type.contains('.') ) type += '.com'
type += '/gpu'
```

sorry :)

mes5k: No problem!

mes5k:
DCO should now be signed off. At least the check looks like it's passing.

Issue #1654, opened 2020-06-26
Invalid GLS path composition in staging directories


hnawar:
## Bug report 

(Please follow this template replacing the text between parentheses with the requested information)

### Expected behavior and actual behavior

Using path as an output for a folder and as an input for the next step. The folder should be copied to the VMof the next process
instead the piepline fails with an error
### Steps to reproduce the problem

(Provide a test case that reproduce the problem either with a self-contained script or GitHub repository)
This happened in step 3 of scflow
https://github.com/combiz/nf-core-scflow

### Program output 

+ cd /[redacted]/fe/0cd4677e6b0d3fe3cd00b1ef31c1d4
+ gsutil -m -q cp gs://[redacted]/[redacted]/fe/0cd4677e6b0d3fe3cd00b1ef31c1d4/.command.run .
+ bash .command.run nxf_stage
CommandException: No URLs matched: gs://[redacted]/[redacted]/c5/a34546b98fea695400c94b695ed7a5/tipif_sce//
CommandException: 1 file/object could not be transferred.

### Environment 

* Nextflow version: 20.04.1
* Java version: [?]
* Operating system: [macOS, Linux, etc] linux
* Bash version: (use the command `$SHELL --version`)

### Additional context
Google Life Science Executor

(Add any other context about the problem here)

pditommaso:
Solved e33fff98.

PR #1655, opened 2020-06-26
Remove trailing / when localising folder with Google Life Sciences

hnawar:
This is to address issue #1654 which can cause a pipeline to fail if a dir is set as an input and has a trailing / causing the path in the resulting gsutil cp command in staging to have // at the end of the source path


pditommaso:
Hi, thanks for this PR. Was the patch tested in a real pipeline to be sure it fixes #1654 and has no side effects?

hnawar:
Hi Paolo,
I did some further testing and this is not yet ready to merge.
Here is what I know so far
For path with trailing / the staging gsutil cp command end up with trailing // throwing an error
When this is removed it works, but the unstagin command for paths with trailing / end up creating an additional subfolder e.g.
output:
path 'folder/'
the unstaging is no gsutil cp -r folder gs://<workdir>/folder/ 
This creates gs://<workdir>/folder/folder
removing this triggers an error somewhere else probably in the mv command added for directories
I've done few more edits locally, and will run additional tests.

hnawar:
This is my test pipeline
`#!/usr/bin/env nextflow



process createFolder {

    output:
    	path 'gsfolder/' into folders
	path 'gsfolder2' into folders2
        path 'test4.txt' into file1

    """
    mkdir gsfolder
    echo 'Hello 1'>>gsfolder/test1.txt
    echo 'Hello 2'>>gsfolder/test2.txt
    mkdir gsfolder2
    echo 'Hello 3'>>gsfolder2/test3.txt
    echo 'Hello 4'>>test4.txt
    """
}


process listFolder {

    input:
    	path test_folder from folders
        path test_folder2 from folders2
        path test4 from file1

    output:
    	stdout result

    """
    ls $test_folder
    ls $test_folder2
    cat $test4
    """
}



`

hnawar:
Removing the renaming of the folder in the FileCopyStrategy and removing the folder name from destination seems to solve the issue.
But I'm not sure why the renaming was introduced
I will test with more standard piplines and see how that works

hnawar:
The test pipeline and ranseq both ran successfully.
Do you recommend any other pipeline for testing?

hnawar:
Also updated test pipeline to have subfolder
`
#!/usr/bin/env nextflow

process createFolder {
    output:
        path 'gsfolder/' into folders
        path 'gsfolder2' into folders2
        path 'test4.txt' into file1

    """
    mkdir -p gsfolder/sub
    echo 'Hello 1'>>gsfolder/test1.txt
    echo 'Hello 2'>>gsfolder/sub/test2.txt
    mkdir gsfolder2
    echo 'Hello 3'>>gsfolder2/test3.txt
    echo 'Hello 4'>>test4.txt
    """
}

process listFolder {
    input:
        path test_folder from folders
        path test_folder2 from folders2
        path test4 from file1

    output:
        stdout result

    """
    ls -R $test_folder
    ls $test_folder2
    cat $test4
    """
}



`

pditommaso:
Does the above snippet isolate the issue? If so, can you make it self checking so that I'll include it in the integration tests e.g.

```
[ \$(cat $test4) = "Hello 4" ] || false
```

hnawar:
Hi Paolo,
The above test does isolate the issue with the two folders with and without the trailing /
I added a file as path to ensure files are not impacted by the changes.
I've made sure that this pipeline fails on the master branch and passes with the branch.

Hatem

hnawar:
I added back the lines removed but added 1 line to handle trailing / in the output.
The trailing / was causing directories to be copied to a subfolder with the same name
e.g. gsfolder/ would be copied to gsfolder/gsfolder
I put back the mv statement but adjusted the if condition to use the sorePath.name instead of the .toString() it seems to work fine, an alternate would be also check for trailing / and remove it before the if statement.
The test-subdirs , test-complexpaths and test-readpairs now all pass

pditommaso:
OK, thanks a lot for this patch! merged manually.

PR #1664, opened 2020-07-05
Fix Java version detection in launch.sh

tskir:
The main version check above (`version_check="^(1.8|9|10|11|12)"`) works fine, but when deciding whether to append additional args for Java 9+, missing parentheses meant that Java 1.8.0_212 also matches the regex and is being considered 9+, causing an error due to incompatible arguments.


pditommaso:
Nice. thanks.

PR #1666, opened 2020-07-05
Include GA4GH support into the standalone binary (make pack)

tskir:
When I tried building a modified Nextflow version, I noticed that the standalone binary generated by `make pack` command works fine but does not include support for GA4GH/TES functionality. This PR fixes it.

Signed-off-by: Kirill Tsukanov <tsukanoffkirill@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,12 @@ pack:
 	BUILD_PACK=1 ./gradlew packAll
 
 #
+# Create self-contained distribution package, including GA4GH support and associated dependencies
+#
+packGA4GH:
+	BUILD_PACK=1 ./gradlew -PGA4GH packAll
+
+#
 # Upload NF launcher to nextflow.io web site
 #
 deploy:
--- a/README.md
+++ b/README.md
@@ -214,6 +214,8 @@ The self-contained runnable Nextflow packages can be created by using the follow
 make pack
 ```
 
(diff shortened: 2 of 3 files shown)


tskir:
OK, thanks for letting me know! I suspected there was a reason it wasn't included, just didn't know why exactly.

If you don't mind keeping this PR open for a day or two, I'll look into it and reimplement as a separate Gradle target

pditommaso:
No hurry

tskir:
@pditommaso I implemented the changes which we discussed (sorry it took me twenty days instead of two!) — now GA4GH support is not included in the self-contained binary by default. There is a separate make target for it, and inside Gradle it is controlled via an optional project property. I think it's less cumbersome than including a whole new Gradle target, which would inevitably cause some significant code duplication.

I additionally tested that `make pack` and `make packGA4GH` generate different binaries, and the latter one is slightly larger and is the only one to actually include GA4GH support, as expected.

Please let me know what you think.

tskir:
P. S. On my system, the size of binaries is:
* `make pack` = 52.3 MB
* `make packGA4GH` = 53.5 MB

So the size increase when including GA4GH support seems to be minimal. So, out of curiosity, what is the reason for not including it by default? Are some of the dependencies which GA4GH support brings along unstable, or are there e.g. some license limitations?

Issue #1684, opened 2020-07-21
k8s executor can miss running state, messing up the timeline


mes5k:
## Bug report 

When using the `k8s` executor with very fast tasks (< 1 second) the run timeline report often doesn't include start/stop data for those tasks. 

### Expected behavior and actual behavior

Because the `k8s` TaskHandler polls the k8s api to determine task state, it's possible for a task to complete within the polling interval. This means the TaskHandler status can skip directly from SUBMITTED to COMPLETED without ever setting the RUNNING state. The result is that TaskHandler.startTimeMillis never gets set. This in turn messes up the run timeline report.

### Steps to reproduce the problem

Run a pipeline with very fast tasks  (< 1 second) and look at the run timeline report.

### Additional context

Will be attaching a patch shortly...

PR #1686, opened 2020-07-23
Fixs k8s timeline start  Try 2

mes5k:
Trying again with a single, signed commit.  Sigh.

This fix addresses #1684.

The patch is pretty simple. It has the K8sTaskHandler check whether the TaskHandler.startTimeMillis is equal to 0 when the task is completing. If it is, then is sets the TaskHandler.startTimeMillis to be the startedAt value found in the Pod terminated message payload.


mes5k:
Ok, I think I finally got it right with this one:  https://github.com/nextflow-io/nextflow/pull/1687.

Sorry for cluttering your PRs.

PR #1687, opened 2020-07-23
Fix k8s timeline start TRY 3

mes5k:
Trying again with a single, signed commit. Sigh.  AGAIN

This fix addresses #1684.

The patch is pretty simple. It has the K8sTaskHandler check whether the TaskHandler.startTimeMillis is equal to 0 when the task is completing. If it is, then is sets the TaskHandler.startTimeMillis to be the startedAt value found in the Pod terminated message payload.


pditommaso:
It should be enough to push or force push BTW 😆

pditommaso:
Was thinking if could not make sense to populate the start time in any case with the timestamp from the pod to avoid any potential inconsistency.

mes5k:
So is your concern that when things work normally and we set `startTimeMillis` [here](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy#L270) that the timestamp isn't precise enough because the handler only polls once a second?

It would be easy enough to remove the `if` statement if that is all that you're thinking.  Or maybe you have something else in mind?

pditommaso:
I was more thinking to use the `startedAt` and `finishedAt` ts retuned by the pod as default either for symmetry and consistency of the collected time.

mes5k:
Ok, that makes sense.  What I will do is rename `updateStartTime` to `updateTimes`, remove the `if` check, and set both `startTimeMillis` and `completeTimeMillis` regardless.  I'll also move when I call `updateTimes` to happen after `status = TaskStatus.COMPLETED`.  And update tests.  Sound about right?

pditommaso:
Instead of having a method with side effects (ie. modifying the timestamp field) IMO it's preferable having one that returns the ts giving the pod state, e.g 

```
  void getStartTime(Map terminated) {
        try {
                final time = DateTimeFormatter.ISO_INSTANT.parse(terminated.startedAt as String)
                return Instant.from(time).toEpochMilli()
        } catch( Exception e ) {
            log.debug "Failed attempted update of startTimeMillis: ${startTimeMillis} from startedAt in: '${terminated.toString()}'", e
            return System.currentTimeMillis() 
        }
    }
```

Much easier to test

mes5k:
Sure, no prob.

modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy
```diff
@@ -273,6 +275,30 @@ class K8sTaskHandler extends TaskHandler {
...
+     */
+    void updateTimestamps(Map terminated) {
+        try {
+            startTimeMillis = getEpochMilli(terminated.startedAt as String)
+            completeTimeMillis = getEpochMilli(terminated.finishedAt as String)
+        } catch( Exception e ) {
+            log.debug "Failed updating timestamps '${terminated.toString()}'", e
+            startTimeMillis = System.currentTimeMillis()
+            completeTimeMillis = System.currentTimeMillis()
+        }
```
pditommaso, line 299: It would be a bit better to keep in two separate try/catch statements. But ok, I can do.

pditommaso: How this sounds to you ? 

```
    void updateTimestamps(Map terminated) {
        try {
            startTimeMillis = getEpochMilli(terminated.startedAt as String)
        } catch( Exception e ) {
            log.debug "Failed updating start timestamp - pod startedAt=${terminated.startedAt}", e
            if( !startTimeMillis )
                startTimeMillis = System.currentTimeMillis()
        }

        try {
            completeTimeMillis = getEpochMilli(terminated.finishedAt as String)
        } catch( Exception e ) {
            log.debug "Failed updating completion timestamp - pod finishedAt=${terminated.finishedAt}", e
            if( !completeTimeMillis )
                completeTimeMillis = System.currentTimeMillis()
        }
    }
```

mes5k: The reason I like one try/catch is to maintain consistency with the timestamps. If you have two try/catch blocks, and `startedAt` fails, but `finishedAt` succeeds, then `startTimeMillis` _could_ have a greater timestamp than `completeTimeMillis`. With just one try/catch, if either fails then we at least know that the timestamps will be consistent.  Make sense?

But I do like that you're checking `if (!startTimeMillis)` in the catch block.  I'll add that to my patch.

pditommaso:
Yeah, I've realized later the consistency thing. Are both are set it none.

Yes, it would be better to set the timestamp on error only if not already
set

PR #1689, opened 2020-07-29
Passing exact memory to executors in GiB

shukapoo:
While integrating TESK (https://github.com/EMBL-EBI-TSI/TESK), A GA4GH TES compliant executor based on Kubernetes, me & my colleague discovered Nextflow is passing user specified memory to executors by a Byte shift operation in `toGiga` function which passes rounded of memory in `Long` format to executor.

This works likes a charm for Memory>1GB, but fails if memory less than 1 GB is supplied (In that scenario, memory requirement of 0GB is passed to the executor which causes error). Additionally exact memory requirement is never passed to executor as memory gets rounded off to nearest integer.

As a workaround we modified `toGiga` function to provide exact memory to the executor, this change also lead to changing return type of function from `Long` to `Double`. This new changes is successfully tested with GA4GH TES (TESK) & Kubernetes executors.

If you think passing memory value in `Double` format would not be digested by some executors & would like to stick to `Long` format, we can think of having some memory handling logic like this:https://github.com/nextflow-io/nextflow/commit/4b9ad005d991631deda6d20021fc65f5cff6a656
 in `getResources` function of `TesTaskHandler`.


shukapoo:
Thanks Paolo ,

So if we understood, we should change the method `getResources` in class `TesTaskHandler` to not use `toGiga` from `MemoryUnit` but write our own in `TesTaskHandler` ?

This suggestion really makes sense, however my colleague @lvarin also suggested another way:  to have this new method in `MemoryUnit` instead and call it `toGigaDouble `or something like that?

So which way you prefer more @pditommaso? We can make a new sanitised version of PR with the way you propose.

pditommaso:
There still some open decisions to be taken on the evolution of the resource allocation model, which may include the inclusion of the `iB` unit. 

For now, I would favour a change limited to the ga4gh package. I don't see any limitation in taking the `MemoryUnit.bytes` and reformatting as you need with a local helper method.

pditommaso:
I think this can be closed in favour of #1696, isn't it?

shukapoo:
Correct, thanks for merging #1696 , this is an duplicate of it & thus can be closed

PR #1696, opened 2020-08-05
Passing exact memory to executors in GiB

lvarin:
This is a new attempt for:
https://github.com/nextflow-io/nextflow/pull/1689

Instead of changing the toGiga function, we are defining a new toGiga function in the `TesTaskHandler` class instead.

### DIFF
--- a/modules/nf-ga4gh/src/main/nextflow/ga4gh/tes/executor/TesTaskHandler.groovy
+++ b/modules/nf-ga4gh/src/main/nextflow/ga4gh/tes/executor/TesTaskHandler.groovy
@@ -214,13 +214,8 @@ class TesTaskHandler extends TaskHandler {
     }
 
     private Double toGiga(MemoryUnit size) {
-        try {
-            // 1073741824 = 1GB
-            return ((double)size.getBytes())/1073741824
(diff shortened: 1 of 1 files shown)


modules/nf-ga4gh/src/main/nextflow/ga4gh/tes/executor/TesTaskHandler.groovy
```diff
@@ -204,14 +205,23 @@ class TesTaskHandler extends TaskHandler {
...
 
+    private Double toGiga(MemoryUnit size) {
+        try {
+            // 1073741824 = 1GB
+            return ((double)size.getBytes())/1073741824
+        } catch (NullPointerException npe) {
+            // If no size is provided, return null, so no memory is sent
+            return null
+        }
+    }
```
pditommaso, line 225: Just a tip, exception handing as flow control is very inefficient. What about: 

```
return size != null ? ((double)size.bytes)/1073741824 : null
```

lvarin: I just changed it.

It is indeed a good tip, thanks

PR #1711, opened 2020-08-18
Fix fromSRA

sstadick:
It looks as though the ebi endpoint changed and is now adding an extra
field to the result of the query for the FASTQ file path. EX:

```text
run_accession\tfastq_ftp
SRR1448774\tftp.sra.ebi.ac.uk/vol1/fastq/SRR144/004/SRR1448774/SRR1448774.fastq.gz
```

This causes the SraExplorer.getFastqUrl method to fail. I have added
some flexibility in parsing the returned TSV from ebi to pick the
`fastq_ftp` by header. I have also updated the tests.

Note that the SRA tests only run if you have a token for NCBI. These
tests are skipped normally.

Additionally, the test `should explor sra` was failing due to how the
output was being checked. I think that it was assuming an older version
of a target class. I've updated that with what I think is a sensible
check.

Signed-off-by: Seth Stadick <sstadick@gmail.com>

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/datasource/SraExplorer.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/datasource/SraExplorer.groovy
@@ -37,7 +37,7 @@ import nextflow.util.Duration
 /**
  * Query NCBI SRA database and returns the retrieved FASTQs to the specified
  * target channel. Inspired to SRA-Explorer by Phil Ewels -- https://ewels.github.io/sra-explorer/
- * 
+ *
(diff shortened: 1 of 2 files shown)


sstadick:
I'll add that Groovy is not a language I use often (ever really) so feel free to point out / change anything that is not idiomatic.

abhi18av:
@sstadick,  were you able to build and test this one locally?

pditommaso:
Thanks a lot for this PR! I've merged manually taking a slightly different approach. See a45decfb.

PR #1712, opened 2020-08-18
WIP: add charliecloud support

### LINKED ISSUE #1707: Using Charliecloud containers
Is it possible to use Charliecloud containers and if so is there any documentation on how to use it?

tbugfinder:
fixes #1707 
I'm opening this PR so that community is aware of WIP of Charliecloud
.

### DIFF
--- a/docs/charliecloud.rst
+++ b/docs/charliecloud.rst
@@ -1,0 +1,68 @@
+.. _charliecloud-page:
+
+************************
+Charliecloud containers
+************************
+
+Nextflow integration with `Charliecloud <https://hpc.github.io/charliecloud>`_ technology allows you to write self-contained
+and truly reproducible computational pipelines.
+
+By using this feature any process in a Nextflow script can be transparently executed into a Podman container. This may
+be extremely useful to package the binary dependencies of a script into a standard and portable format that can be 
+executed on any platform supporting the Podman engine.
+
+.. note::
+    This feature requires Nextflow version ``20.01.0`` or later.
+
+.. warning::
+    This is an incubating feature. The use in production environment is not recommended.
+
+Prerequisites
+==============
+
+You will need Podman installed on your execution environment e.g. your computer or a distributed cluster, depending
+on where you want to run your pipeline. Running in rootless mode requires appropriate OS configuration. Due to current
+Podman limits using cpuset for cpus and memory such is only possible using sudo.
+
+
+How it works
+=============
+
+You won't need to modify your Nextflow script in order to run it with Podman. Simply specify the docker image from
+where the containers are started by using the ``-with-charliecloud`` command line option. For example::
+
+  nextflow run <your script> -with-charliecloud [OCI container image]
+
+Every time your script launches a process execution, Nextflow will run it into a charliecloud container created by using the
+specified image. In practice Nextflow will automatically wrap your processes and run them by executing the ``ch-run``
+command with the image you have provided.
+
+.. note:: A OCI container image can contain any tool or piece of software you may need to carry out a process execution. Moreover the
+  container is run in such a way that the process result files are created in the hosting file system, thus
+  it behaves in a completely transparent manner without requiring extra steps or affecting the flow in your pipeline.
+
+If you want to avoid entering the Podman image as a command line parameter, you can define it in the Nextflow configuration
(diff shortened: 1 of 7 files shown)


pditommaso (changes requested):
Thanks a lot for this initiative. Please sign the commit to make the DCO bot happy. https://github.com/nextflow-io/nextflow/blob/master/CONTRIBUTING.md#contributing-code-changes

phue:
Hi @tbugfinder,
I just came across this PR and I would also be very interested in charliecloud support in nextflow. Therefore I would be willing to help with this.

Can you elaborate on what is needed in this PR ?
I noticed that newer versions of charliecloud  also support pulling images from docker registries without root.
I understand that it would require methods for pulling (`ch-grow pull`) and running (`ch-run`) charliecloud containers.
What seems difficult, is that charliecloud does not seem to honor `ENV` layers in docker containers, but only inherits the hosts environment. This might be tricky if a container contains software in locations that are not set in the `$PATH`.

Thanks!

pditommaso:
It should be implemented a *driver* for charliecloud similar to the ones for [Docker](https://github.com/nextflow-io/nextflow/blob/d0cc38de04f9f541e7d5aac12b7294cd62f1616d/modules/nextflow/src/main/groovy/nextflow/container/DockerBuilder.groovy#L26-L212) or [Singularity](https://github.com/nextflow-io/nextflow/blob/d0cc38de04f9f541e7d5aac12b7294cd62f1616d/modules/nextflow/src/main/groovy/nextflow/container/SingularityBuilder.groovy#L30-L160).

tbugfinder:
I have to admit that I got distracted from spending more time on that and I also want to wait for a few changes in charliecloud. Thank you @phue for contributing to it.

phue:
I think I made some progress, feel free to apply the commits listed above to this PR.
Alternatively, I can open a PR against https://github.com/tbugfinder/nextflow/tree/feature/charliecloud

phue:
There is one issue though. Charliecloud does not set Environment variables (see [docs](https://hpc.github.io/charliecloud/command-usage.html#environment-variables)) that are defined in a Docker ENV layer.

My workaround in commits https://github.com/phue/nextflow/commit/9c640a1934c520ef1c16364f3299721559dbf1ee and https://github.com/phue/nextflow/commit/b316d09394363d90f3eba0db6e9c7f5ada5578ee assumes that the container has a key=value file at `/etc/environment` that contains these variables.
However this does not seem to be the case for all base docker images.

So I guess this could be solved by adding missing environment variables in either the `env` scope or via `charliecloud.envWhitelist` and pass them to the container upon execution.
This is not optimal though as it requires additional configuration.

Any other ideas how to handle this? @tbugfinder @pditommaso

pditommaso:
My understanding is that chanrliecloud has a `--set-env` cli option. that should be enough to pass the env variables listed in the `envWhitelist`

phue:
> My understanding is that charliecloud has a `--set-env` cli option. that should be enough to pass the env variables listed in the `envWhitelist`

Absolutely, that works. But it needs to be explicitly passed and won't work out-of-the-box.

Let's take nf-core containers for example. They usually have the dependencies in a conda environment at
`/opt/conda/envs/nf-core-{name}-{version}/bin` which is added as a Docker [ENV layer](https://github.com/nf-core/tools/blob/b67fd2a734a9e2bcf46b37f0f94b5545f7387298/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/Dockerfile#L10)
However Charliecloud ignores such ENV layers, so we would need to manually pass `envWhitelist="PATH=/opt/conda/envs/nf-core-{name}-{version}/bin:$PATH"`

I would like to do this automatically somehow, but I can't find a way to read this environment from of a Docker image.

pditommaso:
Don't think it's correct because the `envWhitelist` is needed to propagate a host variable in the container. For what you are mentioning it should be enough to add the in the nextflow config  `PATH=/opt/conda/envs/nf-core-{name}-{version}/bin:$PATH` (having `{}` resolved ..).

tbugfinder:
Thanks @phue for doing the hard work within #1809.
I'm closing this PR.

phue:
> Thanks @phue for doing the hard work within #1809.
> I'm closing this PR.

Thank you for the initiative!

Issue #1706, opened 2020-08-13
Enable use of Singularity cache dir as repository


pontus:
## New feature

It would be nice to be able to provide singularity container files in a directory as a repository, available for other users.

The current logic for pulling images to the singularity cache requires the use of a lock file in the cache directory 
and makes it not possible to have a directory provided without write permissions (e.g. use for running pipelines as a user without being trusted to manage said repository).

## Usage scenario 

A shared directory can be used on HPC systems to reduce data duplication from container images. It would also simplify providing validated pipelines to users of a system.

## Suggest implementation 

It seems the current implementation of a cache dir could be adopted as a repository with minor logical changes, e.g. 

* redo lock so logic is taken only for downloading
* if requested file exists, return directly
* for downloading (`singularity pull`), use another name in the same directory and rename the file to the correct name afterwards

An alternative implementation would be prepopulating  a hashmap that's currently kept to keep state of requested files during the runtime.

It would also be possible to build a completely different repository feature,but I'm not sure what would be the requirements for that.

pditommaso:
I agree that could be useful. I would consider a pull request implementing this feature.

marcodelapierre:
Just wanting to mention that some users in our HPC centre have the same issue.
A direct solution would be useful, but right now this is our suggested workaround for "read-only" users:

1. define a variable with the image repo dir:
```
params.imageDir = 'file:///path/to/images'
```

2. use it in the container definitions:
```
process DoThis {
    container = "${params}/image.sif"
    """"
    commands here 
    """}
```

If willing to maintain generality and portability of the pipeline, some conditional logic can be used to switch between remote pull and local repo in the config.

pditommaso:
Does singularity require some special command-line option to handle a container as a directory?

pontus:
I'm not sure I'm following the trail of thought here, but:

Nothing special should be required for a sandbox (a container as a directory), but sandboxes have various issues in themselves (e.g. a large number of files make them time consuming to backup).

pditommaso:
Solved. See dd92772b.

PR #1714, opened 2020-08-19
Change check for container in cache to enable read-only use cases ("repository")

pontus:
This is an approximate implementation of support for #1706.

Roughly follows logic outlined in that suggestion:

* check for file in cache early, if found => return
* if no file in cache, take a mutex and try to download, use a temporary file in the same directory as destination
  * in the downloader, check if someone else has already downloaded the current process waited for the mutex
    * if so, remove tempfile and return
  * try to run `singularity pull` to tempfile and return
* clean up mutex

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/container/SingularityCache.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/SingularityCache.groovy
@@ -19,7 +19,9 @@ package nextflow.container
 
 import java.nio.file.FileSystems
 import java.nio.file.Path
+import java.nio.file.Files
 import java.nio.file.Paths
+import java.nio.file.attribute.FileAttribute
 import java.util.concurrent.ConcurrentHashMap
 
 import groovy.transform.CompileStatic
@@ -171,12 +173,23 @@ class SingularityCache {
     Path downloadSingularityImage(String imageUrl) {
         final localPath = localImagePath(imageUrl)
         final file = new File("${localPath.parent}/.${localPath.name}.lock")
+
         final wait = "Another Nextflow instance is pulling the Singularity image $imageUrl -- please wait the download completes"
         final err =  "Unable to acquire exclusive lock after $pullTimeout on file: $file"
 
+        if( localPath.exists() ) {
+            log.debug "Singularity found local store for image=$imageUrl; path=$localPath"
+            return localPath
+        }
+
+        final pullFile = Files.createTempFile(localPath.parent,
+                                              ".${localPath.name}".toString(),
+                                              ".pulling",
+                                              new FileAttribute<?>[0])
+
         final mutex = new FileMutex(target: file, timeout: pullTimeout, waitMessage: wait, errorMessage: err)
         try {
-            mutex .lock { downloadSingularityImage0(imageUrl, localPath) }
+            mutex .lock { downloadSingularityImage0(imageUrl, localPath, pullFile) }
         }
         finally {
             file.delete()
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/container/SingularityCache.groovy
```diff
@@ -202,14 +222,15 @@ class SingularityCache {
 
         final noHttpsOption = (config.noHttps)? '--nohttps' : ''
 
-        String cmd = "singularity pull ${noHttpsOption} --name ${Escape.path(localPath.getFileName())} $imageUrl > /dev/null"
+        String cmd = "singularity pull ${noHttpsOption} -F --name ${Escape.path(pullFile.getFileName())} $imageUrl > /dev/null"
```
pditommaso, line 225: What's `-F` doing?

modules/nextflow/src/main/groovy/nextflow/container/SingularityCache.groovy
```diff
@@ -171,12 +173,23 @@ class SingularityCache {
...
 
+        if( localPath.exists() ) {
+            log.debug "Singularity found local store for image=$imageUrl; path=$localPath"
+            return localPath
+        }
+
+        final pullFile = Files.createTempFile(localPath.parent,
+                                              ".${localPath.name}".toString(),
+                                              ".pulling",
+                                              new FileAttribute<?>[0])
```
pditommaso, line 188: What's the befits of using a temp file name for during the image pull if there's already a lock protecting against race conditions on the file download?

ewels:
Ping @pditommaso 😄

pditommaso:
Oops I was missing this. Seeing again, I'm a bit confused about the stated goal at #1706 (Enable use of Singularity cache dir as repository) and this PR which seems to create the container download with a temporary file name. 

Can you please elaborate more the issue and how you are trying to fix?

pontus:
Sorry if this was unclear, but the container is stored with the "correct" name, it's only pulled with a temporary name to handle the case of others running during the pull time, e.g:

First runner comes and requests container a, which does not exist
  * That runner takes a lock and starts pulling the container to a temporary file

Second runner requests container a, which does still not exist (because of temporary file used for pulling)
  * Waits for lock

First runner finishes pulling and renames temporary file to a and continues.

Second runner gets lock, discovers the file is already there and uses it.

Without the temporary name, the second runner would risk using an incomplete file.

(And we want to avoid the lock taking if the file exists, being the base use case from #1706.)

I hope that makes the intention a bit clearer. (It's possible some of that should go into comment, grateful for opinions on improvements.)

pditommaso:
> Without the temporary name, the second runner would risk using an incomplete file.

I don't think it should happen because race conditions within the same pipeline are prevented by a sync block 

https://github.com/nextflow-io/nextflow/blob/e94e6877c2dc9bed983e873b62e302a08843c891/modules/nextflow/src/main/groovy/nextflow/container/SingularityCache.groovy#L261-L271

and across different pipeline (using the same image) by a (file) lock:

https://github.com/nextflow-io/nextflow/blob/e94e6877c2dc9bed983e873b62e302a08843c891/modules/nextflow/src/main/groovy/nextflow/container/SingularityCache.groovy#L179

pontus:
Yes, but the very point of this PR is getting rid of taking that lock for the base use case (supporting setups where the person running nextflow does not have write permissions in the directory where container images are stored, e.g. a system provided repository of images for various pipelines on common storage on a HPC system).

pditommaso:
I see, more clear now, thanks for detailing the problem. I think it could a simplified a bit avoiding to create a temp file and passing it as the third argument of `downloadSingularityImage0`. 

Instead, `downloadSingularityImage0` can take the name of the image as it's doing now and append a random suffix (current millis?), renaming it when done as you are suggesting.

pontus:
Something such works as well, mostly uses a tempfile to have the java engine figure out that the name does not exist (and even with tempfile, it can easily be adapted to look in other ways).

But the temporary file handling can absolutely be moved into `downloadSingularityImage0`, I'll have a go at that.

For clarity; would you prefer getting rid of the pre-created tempfile (to run `singularity`without `-F`) or do you not have an opinion on that?

pditommaso:
> Something such works as well, mostly uses a tempfile to have the java engine figure out that the name does not exist (and even with tempfile, it can easily be adapted to look in other ways).

yes, using a random suffix should spare a few file system api that with hpc is always a good idea. 

> For clarity; would you prefer getting rid of the pre-created tempfile (to run singularitywithout -F) or do you not have an opinion on that?

Yes, I think it would be preferable to go with the random suffix and thefore w/o the pre-created tempfile and `-F` option

pditommaso:
Ok, thanks. Merged manually with some minor changes. dd92772b

pontus:
Thanks, I was a bit stuck with not understanding why tests worked in my setups but failed in travis.

PR #1734, opened 2020-09-24
Add example for using named stdout in dsl2.

hartzell:
I had to poke around to get this to work, adding it here to save other folks time.

I'm not entirely sure why it doesn't require a comma like the other named output does, but....


pditommaso (approved):
Looks nice. Thanks!

Issue #1733, opened 2020-09-24
"Process <...> has been already used" when creating a process with only optional outputs


hartzell:
## Bug report 

I started [a discussion in the Nextflow gitter](https://gitter.im/nextflow-io/nextflow?at=5f6bb6d9b8a99f4519a1ff42) about this and @pditommaso said: "this looks a kind of bug" so I'm recording it here:

---

I'd like to set up a workflow that starts with a list of things (e.g. directory names, or perhaps sequence ID's or ...) and passes them to a "categorizer" process that examines them (this might be a computationally expensive step, would want to run on e.g. Batch). The "categorizer" would pass the inputs on to different subgraphs of the workflow based on its calculations.

The "categorizer" would run some shell script-ish thing (compiled executable, python script, shell script invoking various applications, ...).

I'd like to use DSLv2 if possible.

Here's a trivial example where things are categorized on color, but in the real world I'll be basing it on more serious bits.

```
                 ┌─────────────────┐
                 │ List of things  │
                 │(paths/names/...)│
                 └─────────────────┘
                          │
                          │
                          ▼
                 ┌─────────────────┐
                 │   Categorizer   │
                 │                 │
                 └─────────────────┘
                        │   │
         ┌──────────────┘   └───────────────┐
         ▼                                  ▼
┌────────────────┐                 ┌────────────────┐
│Process red ones│                 │ Process green  │
│                │                 │      ones      │
└────────────────┘                 └────────────────┘
         │                                  │
         │                                  │
         │                                  │
         ▼                                  ▼
```

### Expected behavior and actual behavior

I believe that the following script should sort the inputs into "valid" and "invalid", but it fails with:

```
$ nextflow run categorizer-2.nf
N E X T F L O W  ~  version 20.07.1
Launching `categorizer-2.nf` [sleepy_sanger] - revision: 7a3e3456b1
Process categorize has been already used -- If you need to reuse the same component include it with a different name or include in a different workflow context

 -- Check script 'categorizer-2.nf' at line: 43 or see '.nextflow.log' file for more details
```

Here's the test case code:

```
nextflow.enable.dsl=2

process categorize {
    input:
        val(name)
    output:
        env valid optional true
        env invalid optional true
    script:
    """
    if [[ $name == /tmp/* ]] ; then
        invalid=$name
    else
        valid=$name
    fi
    """
}

process handle_valid {
    input:
        val(name)
    output:
        val(name)
    """
    /bin/true
    """
}

[...]

PR #1740, opened 2020-09-29
Let output env variables be optional

stekaz:
This PR provides a fix for https://github.com/nextflow-io/nextflow/issues/1733. Basically it just lets variables defined with the 'env' qualifier be optional, for example:

```
nextflow.enable.dsl=2

process test {

    output:
    env(myvalue) optional true

    """
    myvalue=
    """
}

workflow {

    test()
}
```


hartzell:
This looks great and **thank you** for fixing it.  Based on the documentation, I'd think that *all* of the possible output types are allowed to be optional.  Is that the case and the `env` type is the only one that was missing or are there others that should be brought into alignment with the docs (or *vice versa*)?

stekaz:
No worries at all @hartzell. I think the remaining output types are StdOutParam and ValueOutParam. I can't think of a case where making 'val' qualified outputs optional makes much sense, so I think we can skip this one? From my quick investigation with the StdOutParam, it looks like it's always captured as either a String (using 'exec') or UnixPath (using 'script'), so I don't know if there's any need to make it optional? Maybe not returning null sometimes is less than ideal though. I may need to investigate this one a bit more but that will take time.

pditommaso (changes requested):
This is a nice improvement. Thanks for your contribution! any chance to add also a unit test for it in this class 

https://github.com/nextflow-io/nextflow/blob/e94e6877c2dc9bed983e873b62e302a08843c891/modules/nextflow/src/test/groovy/nextflow/script/params/EnvOutParamTest.groovy#L30-L55

You check a similar test here 

https://github.com/nextflow-io/nextflow/blob/e94e6877c2dc9bed983e873b62e302a08843c891/modules/nextflow/src/test/groovy/nextflow/script/params/ParamsOutTest.groovy#L445

hartzell:
@stekaz -- optional values are a powerful way to have portions of a workflow graphs execute in some situations and not in others.

- for `val` output parameters - since they can be defined/populated via groovy code (e.g. the second example [in the Output Values docs](https://www.nextflow.io/docs/latest/process.html#output-values), I can imagine cases where it might not be defined.  In that second example, which defined the output value `x`; if `x` was defined as *optional* and  if the groovy bit of the `script:` block were an `if` that assigned to `x` in one branch but not the other, would that be an error?
- for `stdout`, what is it's value if the exec/script fails to run?  Is it just the empty string?  How is that different than when things run successfully but don't generate any output?

In any case, if only *some* types of outputs are allowed to be optional, it would be great if:
- the docs made it clear when optional was valid and when it wasn't; and
- (this seems like more work) an error was generated when the `optional` flag was used with an unsupported output type.

**Thank you** again for cleaning this up!

stekaz:
Hi @pditommaso, I've added a unit test for this as requested. Please let me know if anything needs updating, thanks!

@hartzell I agree, I think it would be great to make `val` and `stdout` output parameters optional too. I'll try to get back to this when time allows.

pditommaso (approved):
Awesome. thanks!

PR #1751, opened 2020-10-06
feat: add the docs for CLI interface

abhi18av:
This PR adds the `Command Line Interface` section in the docs.

- CLI top-level options

image

- CLI top-level commands

image

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -1,0 +1,1591 @@
+.. _cli-page:
+
+******************
+Command line interface
+******************
+
+`Nextflow` provides a robust command line interface allowing you to manage the execution of a pipeline. The top-level interface consists of two aspects, ``options`` and ``commands``.
+
+Here's what you'll see at the top-level upon invoking ``nextflow`` CLI. ::
+
+
+    $ nextflow
+    Usage: nextflow [options] COMMAND [arg...]
+
+
+
+.. _cli-options:
+CLI options
+============
+
+The top-level ``options`` are meant to be invoked in relation to the core **nextflow runtime**. For child options specific to any ``command``, please refer the `CLI commands` section.
+
+An overview of the `Nextflow` top-level options. ::
+
+
+    $ nextflow
+    Usage: nextflow [options] COMMAND [arg...]
+
+    Options:
+    -C
+        Use the specified configuration file(s) overriding any defaults
+    -D
+        Set JVM properties
+    -bg
+        Execute nextflow in background
+    -c, -config
+        Add the specified file to configuration set
+    -d, -dockerize
+        Launch nextflow via Docker (experimental)
+    -h
+        Print this help
+    -log
+        Set nextflow log file path
+    -q, -quiet
+        Do not print information messages
+    -syslog
+        Send logs to syslog server (eg. localhost:514)
+    -v, -version
+        Print the program version
+
+    Commands...
+
+--------------------
+Hard configuration override
+--------------------
+
+
+**Description**
+
+Use the specified configuration file(s) overriding any defaults
+
+
+**Usage**
+::
+   $ nextflow -C nxf.config COMMAND [arg...]
+
+**Extended description**
+
+The ``C`` option is used to override *all* settings specified in the default config file. For soft override, please refer the ``c`` option.
+
+
+**Examples**
+
+
+- Override **any** previous config with a custom ``config`` file. ::
+    
+  $ nextflow -C nxf.config run nextflow-io/hello
+
+--------------------
+JVM properties
+--------------------
+
+
+**Description**
+
+Set JVM properties
+
+**Usage**
+::
+
+   $ nextflow -Dkey=value COMMAND [arg...]
+
+**Extended description**
+
+The JVM platform, which is utilized by ``nextflow`` as its runtime, allows adding system level properties/values during the startup, which could be accessed during runtime. For specifying other JVM level options, please refer the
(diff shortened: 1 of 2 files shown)


docs/cli.rst
```diff
@@ -0,0 +1,853 @@
...
+| -keep-logs, -k            |   false    | Removes only temporary files but retains execution log entries and metadata.   |                                           
++---------------------------+------------+--------------------------------------------------------------------------------+
+| -quiet, -q                |   false    | Do not print names of files removed.                                           |
++---------------------------+------------+--------------------------------------------------------------------------------+
+
+
+
+Examples
+^^^^^^^^^^^^^^^^^^^^
+
```
abhi18av, line 132: Please let me know if we're open to adding the Asciinema `Shell casts` in `Usage` section. 

These GIFs can improve the beginner friendliness of the docs and this way we can cover the scenario and provide more context for any command.

[asciicast](https://asciinema.org/a/363719)

docs/cli.rst
```diff
@@ -0,0 +1,853 @@
...
+Description
+^^^^^^^^^^^^^^^^^^^^
+Print executions log and runtime info
+
+
+Usage
+^^^^^^^^^^^^^^^^^^^^
+
+The ``log`` command can be invoked like so::
```
abhi18av, line 568: Could you please guide me on the usage of this command - I've tried a few permutations and I only seem to generate the same result.

A couple of examples here would be great 🆘

abhi18av (commented):
Highlighted areas where I'd appreciate some guidance.

docs/cli.rst
evanfloden: See 11.4. Execution provenance at https://seqera.io/training/.

docs/cli.rst
```diff
@@ -0,0 +1,853 @@
...
+
+
+    $ nextflow
+    Usage: nextflow [options] COMMAND [arg...]
+
+    Options:
+    -C
+        Use the specified configuration file(s) overriding any defaults
+    -D
+        Set JVM properties
```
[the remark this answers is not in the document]
evanfloden: We should be clear that these are options for the Nextflow base command and these options are applied when using the child commands (e.g. `nextflow -C nextflow.config run rnaseq.nf`)

docs/cli.rst
evanfloden: Let's leave this out for now.

evanfloden:
For usage, we can remove the line.

`The XXX command can be invoked like so::`

docs/cli.rst
abhi18av: Okey dokey 👍

docs/cli.rst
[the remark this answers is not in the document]
abhi18av: Agreed - let me think about to how put it forward nicely.

docs/cli.rst
abhi18av: Gotcha!

docs/cli.rst
```diff
@@ -0,0 +1,988 @@
...
+
+
+Examples
+^^^^^^^^^^^^^^^^^^^^
+
+
+--------------------
+kuberun       
+--------------------
+
```
abhi18av, line 395: @evanfloden I'd like to confirm, for this iteration, how extensive should be the `EXAMPLES` section for `kuberun` and `run`.

Here are the possible choices:


- Should we document everything extensively right now as the `CLI` doc would probably be referenced quite frequently?

- Should we focus only on the most frequent use-cases and add more examples as per user requests?

docs/cli.rst
```diff
@@ -0,0 +1,988 @@
...
+| -help, -h                 |  false     | Print the command usage.                                                       |
++---------------------------+------------+--------------------------------------------------------------------------------+
+| -o                        |  text      | Output format, either ``text``, ``json`` or ``yaml``.                          |
++---------------------------+------------+--------------------------------------------------------------------------------+
+
+
+
+Examples
+^^^^^^^^^^^^^^^^^^^^
+
```
abhi18av, line 390: While trying out the options for the `info` command, I couldn't make use of the `-o` output feature.
```
(base) Abhinavs-MacBook-Pro:_resources eklavya$ nextflow info -o yaml
  Version: 20.07.1 build 5412
  Created: 24-07-2020 15:18 UTC (20:48 IDT)
  System: Mac OS X 10.15.6
  Runtime: Groovy 2.5.11 on OpenJDK 64-Bit Server VM 1.8.0_192-b01
  Encoding: UTF-8 (UTF-8)

(base) Abhinavs-MacBook-Pro:_resources eklavya$ nextflow info
  Version: 20.07.1 build 5412
  Created: 24-07-2020 15:18 UTC (20:48 IDT)
  System: Mac OS X 10.15.6
  Runtime: Groovy 2.5.11 on OpenJDK 64-Bit Server VM 1.8.0_192-b01
  Encoding: UTF-8 (UTF-8)
```

I've also tried coupling it with `-d` detailed option but the result always seems to be printed out in the `text` format. 

If this is the expected behavior, then perhaps I'm missing some info here 🤔

abhi18av (commented):
Highlighted `info` and `kuberun` related aspects which need some clarifications.

docs/cli.rst
abhi18av: I've opted for the most frequent ones and would add more in case there is user request or I get more ideas.

evanfloden (approved):

abhi18av:
I'm closing this PR since there's another PR which builds upon this work.

https://github.com/nextflow-io/nextflow/pull/1759

Issue #1709, opened 2020-08-17
GCP - Do not require service account key export for google-lifesciences


mozack:
## New feature

Hi, it would be very helpful to use the currently authenticated credentials instead of export service account key for the google-lifesciences executor.  In fact, we cannot use the life-sciences API for most of our use cases without this.

This was done previously for the google-pipelines executor:
https://github.com/nextflow-io/nextflow/issues/1068

## Usage scenario 

There is data available to the public on GCP that requires security access.  Providers of this data do not allow export of security keys because it may circumvent their security measures.  For example, the TCGA and TARGET data stored here: https://isb-cgc.appspot.com/ cannot be accessed from a project with an exported security key.

## Suggest implementation 

It should be possible to just used the currently authenticated user's credentials.  This code snippet from @kdbinder may help:

https://github.com/kdbinder/nextflow/commit/8b6c049821ef859c16b7e6b8c9e3e3129e3ff342

Thanks!

pditommaso:
Not sure to understand how the auth credentials are taken then?

kdbinder:
If nextflow is executed from an existing GCP instance, the [service account is baked into that VM ](https://cloud.google.com/docs/authentication/production#automatically)

snamburi3:
I was able to bypass the service account key export and use the user credentials via `gcloud auth application-default login`.

1. ```gcloud auth application-default login```

2. Add Project Name to the json config file ~/.config/gcloud/application_default_credentials.json
```project_id": "PROJECT NAME",```

3. export values
```
export NXF_VER="20.07.1"
export NXF_MODE=google
export NXF_DEBUG=3
export PROJECT="PROJECT NAME""
export GOOGLE_APPLICATION_CREDENTIALS=~/.config/gcloud/application_default_credentials.json
 ```

mbookman:
Reviewing the description of how Application Default Credentials work, as described here:

https://github.com/googleapis/google-auth-library-java#google-auth-library-oauth2-http

The key bit is:

> The following are searched (in order) to find the Application Default Credentials:
> 
> 1. Credentials file pointed to by the GOOGLE_APPLICATION_CREDENTIALS environment variable
> 1. Credentials provided by the Google Cloud SDK gcloud auth application-default login command
> 1. Google App Engine built-in credentials
> 1. Google Cloud Shell built-in credentials
> 1. Google Compute Engine built-in credentials
>     - Skip this check by setting the environment variable NO_GCE_CHECK=true
[...]

PR #1776, opened 2020-10-28
Google Life Sciences update credential handling.

mbookman:
This is a suggested implementation for updating credentials handling for the Google Life Sciences executor as discussed in #1709.

Automatically pick up Application Default Credentials removing explicit requirements for the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to be set.

I assume there are some tests to update. Feel free to take over this PR or point me to where to updates tests.

### DIFF
--- a/docs/google.rst
+++ b/docs/google.rst
@@ -19,26 +19,25 @@ in your system environment::
 Credentials
 -----------
 
-To allow the deployment in the Google Cloud you need to configure the security credentials using
-a *Security account key* JSON file.
+Credentials for submitting requests to the Google Pipelines API are picked up from your
+environment using `Application Default Credentials <https://github.com/googleapis/google-auth-library-java#google-auth-library-oauth2-http>`_.
+Application Default Credentials are designed to use the credentials most natural to the
+environment in which a tool runs.
 
-Nextflow looks for this file using the ``GOOGLE_APPLICATION_CREDENTIALS`` variable that
-has to be defined in the launching environment.
+The most common case will be to pick up your end-user Google credentials from your
+workstation. You can create these by running the command:
 
-If you don't have it, download the credentials file from the Google Cloud Console following these steps:
+    gcloud auth application-default login 
 
-* Open the `Google Cloud Console <https://console.cloud.google.com>`_
-* Go to APIs & Services → Credentials
-* Click on the *Create credentials* (blue) drop-down and choose *Service account key*, in the following page
-* Select an existing *Service account* or create a new one if needed
-* Select JSON as *Key type*
-* Click the *Create* button and download the JSON file giving a name of your choice e.g. ``creds.json``.
+and running through the authentication flow. This will write a credential file to your gcloud
+configuration directory that will be used for any tool you run on your workstation that
+picks up default credentials.
 
-Finally define the following variable replacing the path in the example with the one of your
-credentials file just downloaded::
+The next most common case would be when running on a Compute Engine VM. In this case,
+Application Default Credentials will pick up the Compute Engine Service Account
+credentials for that VM.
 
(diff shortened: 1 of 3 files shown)


pditommaso (changes requested):
Thanks a lot for submitting this PR. I'm fine with it, however, it would be important to keep backward compatibility with current behavior. 

Therefore my suggestion is that if NF detects the `GOOGLE_APPLICATION_CREDENTIALS` it continues with the current behavior, otherwise it applies the logic you are suggesting.

mbookman:
Thanks Paolo. I have restored the previous behavior when `GOOGLE_APPLICATION_CREDENTIALS` is set.

There is a small drawback to preserving the current behavior if a user wants to use a service account credential file to run workflows in different projects.  In that case, they need to edit the JSON and change the project ID. If it is something that they do frequently, they can have two copies of their credential file, for example:

```
$ ls my_nextflow_creds/

my-creds-file.project_A.json
my-creds-file.project_B.json
```

(where the only difference between the above two files is the `project_id` field) and then set `GOOGLE_APPLICATION_CREDENTIALS` to point to the respective JSON for different runs.

In practice, when people use service accounts, they tend to restrict usage to the associated Cloud project. Further, I think going forward, most use cases won't set `GOOGLE_APPLICATION_CREDENTIALS` anyway and the appropriate Application Default Credentials will get picked up (and the project ID will come from the nextflow config).

pditommaso:
Hi, Sorry for the late reply. 

> There is a small drawback to preserving the current behavior if a user wants to use a service account credential file to run workflows in different projects

In this case, the projectId could be set in the nextflow config. Do you agree? 

Also please, sign-off your commit to make the DCO bot to pass the check https://github.com/nextflow-io/nextflow/pull/1776/checks?check_run_id=1323499154

pditommaso:
One more thing, still the google region or zone to be used should be provided in the nextflow config. Does make sense to you ?

mbookman:
Re: _In this case, the projectId could be set in the nextflow config. Do you agree?_

I think it would be the preferred behavior for the nextflow config to override the service account config, but in the current implementation, that would raise an error:

```
        def credsFile = env.get('GOOGLE_APPLICATION_CREDENTIALS')
        final projectId = GoogleLifeSciencesConfig.getProjectIdFromCreds(credsFile)

        config = GoogleLifeSciencesConfig.fromSession(session)
        if( !config.project )
            config.project = projectId
        else if( config.project != projectId )
            throw new AbortOperationException("Project Id `$config.project` declared in the nextflow config file does not match the one expected by credentials file: $credsFile")
```

pditommaso:
You are right, now I'm also remembering that this was how it was working in the past (the project id in the config had precedence) but users tended to forget to change/update it resulting in the execution hanging. 

I think it would be better to keep it and modify it as following: 

```
        def projectId
        def credsFile = env.get('GOOGLE_APPLICATION_CREDENTIALS')
        if( credsFile && (projectId = GoogleLifeSciencesConfig.getProjectIdFromCreds(credsFile)) ) {
            if( !config.project )
                config.project = projectId
            else if( config.project != projectId )
                throw new AbortOperationException("Project Id `$config.project` declared in the nextflow config file does not match the one expected by credentials file: $credsFile")
        } 
        
        if( !config.project ) {
            throw new AbortOperationException("Missing Google project Id -- Specify it adding the setting `google.project='your-project-id'` in the nextflow.config file")
        }
```

To summarise: 
1. try to fetch the projectId from the creds file 
2. if exists make sure matches the one in the nextflow config
3. use the creds projectId when in the nexflow config is missing 
4. stop if no projectId is given either from creds file or nextflow config 

Happy to merge once the commit is signed-off.

mbookman:
commit is signed off

pditommaso:
Nice. Thanks for contributing to the project. I've merged this PR manually.

PR #1784, opened 2020-11-03
Typo in dsl2 docs

nh13:


pditommaso:
This is actually correct, check the tip green box https://www.nextflow.io/docs/latest/channel.html#channel-factory

Issue #1832, opened 2020-12-14
Add support for cpus throttling when using container execution


pditommaso:
Modern container engines are able to limit CPU usage to a specific amount. This is very useful to prevent CPU hog tasks to use more resources than requested and prevent faulty node condition. 

Docker and Podman support this using the `--cpus` option. It should be investigated if also Singularity and Charliecloud allow the same.

pditommaso:
@abhi18av want to give a try to this?

abhi18av:
Yes @pditommaso, I'd like to have a go at this. Though I'd be bothering you a bit about Nextflow internals 😉

pditommaso:
It should not be that hard. You can follow how the support for `memory` is already implemented 

https://github.com/nextflow-io/nextflow/blob/d0cc38de04f9f541e7d5aac12b7294cd62f1616d/modules/nextflow/src/main/groovy/nextflow/container/ContainerBuilder.groovy#L73-L84

abhi18av:
`cpus` throttling support in 
- `docker` https://docs.docker.com/config/containers/resource_constraints/#cpu ✅ 
- `podman` http://docs.podman.io/en/latest/markdown/podman-run.1.html#cpu-quota-limit ✅ 
- `singularity https://sylabs.io/guides/3.3/user-guide/cli/singularity_exec.html?highlight=cpu ✅ 
- `charliecloud` https://hpc.github.io/charliecloud/search.html?q=cpu&check_keywords=yes&area=default ❌

abhi18av:
@pditommaso , I'll be raising PRs for these containers individually ( I prefer smaller PRs) - the first draft for `Singularity` is here.

https://github.com/nextflow-io/nextflow/pull/1833

Could you please confirm if the direction is okay? The tests are passing locally.

Also, I noticed my auto code reformatting generates multiple changes on IntelliJ - do I need to update any settings?

pditommaso:
Come on, three different PR are needed :)

It looks, in the same vein add also the ones for docker and podman. For the formatting check https://github.com/nextflow-io/nextflow/blob/master/CONTRIBUTING.md#ide-settings

abhi18av:
For `docker`, does this mean we'll drop the `--cpuset-cpus` in favor of `--cpus`?

pditommaso:
Nope, that's a different thing.

pditommaso:
Done via #1833

PR #1833, opened 2020-12-15
Update CPU and RAM for containers

abhi18av:
This PR addresses the discussion done here https://github.com/nextflow-io/nextflow/issues/1832

`cpus` throttling support in 
- `docker` https://docs.docker.com/config/containers/resource_constraints/#cpu 
- `podman` http://docs.podman.io/en/latest/markdown/podman-run.1.html#cpu-quota-limit 
- `singularity` https://sylabs.io/guides/3.3/user-guide/cli/singularity_exec.html?highlight=cpu

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/container/ContainerBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/ContainerBuilder.groovy
@@ -67,8 +67,8 @@ abstract class ContainerBuilder<V extends ContainerBuilder> {
         return (V)this
     }
 
-    V setCpus(String value) {
-        this.cpus = String.format("%.1f", Float.parseFloat(value))
+    V setCpus(Float value) {
+        this.cpus = value
         return (V)this
     }
 
--- a/modules/nextflow/src/main/groovy/nextflow/container/ContainerBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/ContainerBuilder.groovy
@@ -37,7 +37,7 @@ abstract class ContainerBuilder<V extends ContainerBuilder> {
 
     protected List<String> engineOptions = []
 
-    protected String cpus
+    protected Float cpus
 
     protected String cpuset
 
--- a/modules/nextflow/src/main/groovy/nextflow/container/ContainerBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/ContainerBuilder.groovy
@@ -39,6 +39,8 @@ abstract class ContainerBuilder<V extends ContainerBuilder> {
 
     protected String cpus
 
+    protected String cpuset
+
     protected String memory
 
     protected String temp
@@ -65,8 +67,13 @@ abstract class ContainerBuilder<V extends ContainerBuilder> {
         return (V)this
     }
 
-    V setCpus( String value ) {
-        this.cpus = value
+    V setCpus(String value) {
+        this.cpus = String.format("%.1f", Float.parseFloat(value))
+        return (V)this
+    }
+
+    V setCpuset(String value) {
+        this.cpuset = value
         return (V)this
     }
 
--- a/modules/nextflow/src/main/groovy/nextflow/container/DockerBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/DockerBuilder.groovy
@@ -117,7 +117,7 @@ class DockerBuilder extends ContainerBuilder<DockerBuilder> {
         result << 'run -i '
 
         if( cpus )
-            result << "--cpus ${cpus} "
+            result << "--cpus ${String.format("%.1f", cpus)} "
 
         if( cpuset ) {
             if( legacy )
--- a/modules/nextflow/src/main/groovy/nextflow/container/DockerBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/DockerBuilder.groovy
@@ -116,11 +116,14 @@ class DockerBuilder extends ContainerBuilder<DockerBuilder> {
 
         result << 'run -i '
 
-        if( cpus ) {
+        if( cpus )
+            result << "--cpus ${cpus} "
+
+        if( cpuset ) {
             if( legacy )
-                result << "--cpuset ${cpus} "
+                result << "--cpuset ${cpuset} "
             else
-                result << "--cpuset-cpus ${cpus} "
+                result << "--cpuset-cpus ${cpuset} "
         }
 
         if( memory )
--- a/modules/nextflow/src/main/groovy/nextflow/container/PodmanBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/PodmanBuilder.groovy
@@ -109,7 +109,7 @@ class PodmanBuilder extends ContainerBuilder<PodmanBuilder> {
             result << runOptions.join(' ') << ' '
 
         if( cpus ) {
-            result << "--cpus ${cpus} "
+            result << "--cpus ${String.format("%.1f", cpus)} "
         }
 
         if( memory ) {
--- a/modules/nextflow/src/main/groovy/nextflow/container/PodmanBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/PodmanBuilder.groovy
@@ -108,11 +108,20 @@ class PodmanBuilder extends ContainerBuilder<PodmanBuilder> {
         if( runOptions )
             result << runOptions.join(' ') << ' '
 
+        if( cpus ) {
+            result << "--cpus ${cpus} "
+        }
+
+        if( memory ) {
+            result << "--memory ${memory} "
+        }
+
+
         // the name is after the user option so it has precedence over any options provided by the user
-        if( name )
+        if ( name )
             result << '--name ' << name << ' '
 
-        if( registry )
+        if ( registry )
             result << registry
 
         // finally the container name
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -484,6 +484,9 @@ class BashWrapperBuilder {
         if( this.containerMemory )
             builder.setMemory(containerMemory)
 
+        if( this.containerCpus )
+            builder.setCpus(containerCpus)
+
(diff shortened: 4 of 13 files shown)


modules/nextflow/src/main/groovy/nextflow/container/DockerBuilder.groovy
```diff
@@ -117,7 +131,10 @@ class DockerBuilder extends ContainerBuilder<DockerBuilder> {
         result << 'run -i '
 
         if( cpus ) {
-            if( legacy )
+
+            result << "--cpus ${countCpus(cpus)} "
+
+            if (legacy)
```
abhi18av, line 137: @pditommaso , I've tried to accomodate the newer `--cpus` option with the `cpuset`. The `countCpus` private method (defined above) computes it from the `cpus`.

abhi18av (commented):
Highlighted the choice made for `docker`.

Next, I'll test these with the containers locally, to make sure their docs reflect the current status and we can continue with these changes.

modules/nextflow/src/main/groovy/nextflow/container/DockerBuilder.groovy
```diff
@@ -49,21 +49,35 @@ class DockerBuilder extends ContainerBuilder<DockerBuilder> {
...
+        if (params.containsKey('temp'))
             this.temp = params.temp
 
-        if( params.containsKey('engineOptions') )
+        if (params.containsKey('engineOptions'))
             addEngineOptions(params.engineOptions.toString())
 
-        if( params.containsKey('runOptions') )
+        if (params.containsKey('runOptions'))
             addRunOptions(params.runOptions.toString())
```
pditommaso, line 81: Do not commit style changes

modules/nextflow/src/main/groovy/nextflow/container/DockerBuilder.groovy
pditommaso: Hold-on. These are two different things. Much easier to rename the old `cpus` to `cpuset`, and add a new attribute `Float cpus` . The option `--cpus <cpus>` should be formatted to one decimal precision digit.

modules/nextflow/src/main/groovy/nextflow/container/PodmanBuilder.groovy
```diff
@@ -102,17 +102,26 @@ class PodmanBuilder extends ContainerBuilder<PodmanBuilder> {
...
+        if (entryPoint)
             result << '--entrypoint ' << entryPoint << ' '
 
-        if( runOptions )
+        if (runOptions)
             result << runOptions.join(' ') << ' '
 
+        if (cpus) {
+            result << "--cpu-quota ${cpus} "
+        }
```
pditommaso, line 113: Not a podman expert, but I would expect it has the same options of docker, that are 

```
      --cpu-quota int                  Limit CPU CFS (Completely Fair Scheduler) quota
      --cpu-rt-period int              Limit CPU real-time period in microseconds
      --cpu-rt-runtime int             Limit CPU real-time runtime in microseconds
  -c, --cpu-shares int                 CPU shares (relative weight)
      --cpus decimal                   Number of CPUs
```

abhi18av: Yup, you are right. I simply opted for the base option here rather than the alias http://docs.podman.io/en/latest/markdown/podman-run.1.html#cpu-quota-limit

image

modules/nextflow/src/main/groovy/nextflow/container/DockerBuilder.groovy
abhi18av: I'm updating my Groovy code style so that these would not be there in the final code.

Though, an `.editorconfig` might help :)

pditommaso:
Ok, we are almost there. What's it's missing is the setting of the cpus. This could be done the same as for the mem 

https://github.com/nextflow-io/nextflow/blob/59a15eb6baead5e56e258b328031291def562acd/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy#L484-L485

abhi18av:
Paolo, I'm not sure here why the `shold resolve remote config` test from `CmdConfigTest` is failing.

https://travis-ci.org/github/nextflow-io/nextflow/jobs/750009656#L420

pditommaso:
Do the tests pass in your computer?

abhi18av:
Yup, apart  from the `CmdConfigTest` all others pass 

image

pditommaso:
What I mean if it fails in your computer you should be able to debug it :D

abhi18av:
Yes, I limited myself to thinking only within the `container cpu/cpuset/memory` related stuff 😅 

Anyways, on deeper investigation, 

- The `nextflow-io/hello` uses a `quay` different container now - updated on the day as the latest commit on master (perhaps after running the tests).
- 
- I realized that this particular test was sourcing my local copy of `nextflow-io/hello` which was pinned to a previous version.

image


After which, all tests are passing (locally).

image

pditommaso:
It looks fine. It's only missing the setting of the cpus in the container when the task going to be executed. See how it's done for the memory 

https://github.com/nextflow-io/nextflow/blob/59a15eb6baead5e56e258b328031291def562acd/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy#L484-L485

modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
```diff
@@ -484,6 +484,9 @@ class BashWrapperBuilder {
         if( this.containerMemory )
             builder.setMemory(containerMemory)
 
+        if( this.containerCpus )
+            builder.setCpus(containerCpus)
+
```
abhi18av, line 489: @pditommaso , is this not what you meant? I updated the `BashWrapperBuilder` when you mentioned earlier.

pditommaso: You are right! 🎉

modules/nextflow/src/main/groovy/nextflow/container/ContainerBuilder.groovy
```diff
@@ -39,6 +39,8 @@ abstract class ContainerBuilder<V extends ContainerBuilder> {
 
     protected String cpus
```
pditommaso, line 40: This should be `Float`

modules/nextflow/src/main/groovy/nextflow/container/ContainerBuilder.groovy
```diff
@@ -65,8 +67,13 @@ abstract class ContainerBuilder<V extends ContainerBuilder> {
         return (V)this
     }
 
-    V setCpus( String value ) {
-        this.cpus = value
+    V setCpus(String value) {
+        this.cpus = String.format("%.1f", Float.parseFloat(value))
+        return (V)this
+    }
```
pditommaso, line 73: The type should change accordingly to `Float` and formatting moved into the CLI option rendering

modules/nextflow/src/main/groovy/nextflow/container/DockerBuilder.groovy
```diff
@@ -116,11 +116,14 @@ class DockerBuilder extends ContainerBuilder<DockerBuilder> {
 
         result << 'run -i '
 
-        if( cpus ) {
+        if( cpus )
+            result << "--cpus ${cpus} "
```
pditommaso, line 120: Convert float to string here

modules/nextflow/src/main/groovy/nextflow/processor/TaskBean.groovy
```diff
@@ -62,6 +62,8 @@ class TaskBean implements Serializable, Cloneable {
 
     String containerCpuset
 
+    String containerCpus
```
pditommaso, line 65: This should be a `Float` as well

pditommaso (changes requested):
We are almost there, a few changes in the cpus type declaration are needed.

pditommaso:
@phue Any idea if Charliecloud support CPUs limits in the same way as `docker run --cpus` option?

phue:
> @phue Any idea if Charliecloud support CPUs limits in the same way as `docker run --cpus` option?

It seems there is no such feature in Charliecloud, the available run options are listed here:
https://hpc.github.io/charliecloud/command-usage.html#ch-run

pditommaso:
Ok, no pb. thanks for the feedback.

phue:
I tried to read up on this. As Charliecloud makes use of user namespaces, it should also be possible to modify [cgroups](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/resource_management_guide/sec-cpu) to limit the number of cpus (in fact that seems to be what podman and the likes seem to do with the --cpus flag)
But it would only work on linux and I'm not sure if it can be done as unpriviledged user

pditommaso:
I guess Docker and Podman do the same, but it's out of the scope of this issue hack low-level cgroup model.

modules/nextflow/src/main/groovy/nextflow/container/ContainerBuilder.groovy
pditommaso, line 73: I think this should take a Float and just assign to `this.cpus`

abhi18av: Whoops, I reverted this change earlier by mistake along with other style changes.

@pditommaso , shall we work on an `editorconfig` next?

pditommaso:
Nice! think we ca merge this 🎉

modules/nextflow/src/main/groovy/nextflow/container/ContainerBuilder.groovy
pditommaso: No problem. I think I can generate the editor config with Intellij, let me check.

abhi18av: That'd be really helpful :+1:

abhi18av: @pditommaso, here's a plugin which seems to do what we want 
https://plugins.jetbrains.com/plugin/7294-editorconfig

pditommaso: Nice!

PR #1837, opened 2020-12-17
Another typo fix for DSL2 docs

jfy133:
Original

image

You could also rephrase to: "Process output option mode flatten is not available any more. Replace it using flatten on the corresponding output channel."

Regardless `:ref:operator-flatten` doesn't pick up the 'operator' part of the section name in the `.rst` file.


pditommaso:
Make sense. Merged manually f41ec20d.

Issue #1570, opened 2020-04-15
k8s executor must provide cpu resource requests (and limits)


alanbchristie:
### Expected behavior and actual behavior

The k8s executor is not providing Pod resource `request` declarations and compute nodes become over-subscribed with Pods (Jobs) that simply compete with each other for CPU. If a cpu **request** is not provided Kubernetes assumes the corresponding Container needs 0.1 CPU.

For each nextflow **process** I would expect nextflow to set the Pod Container's CPU resources based on the `cpus` declaration in the **process**  block, or `1` if not explicitly defined. i.e.: -

```
spec:
  containers:
    ...
    resources:
      limits:
        cpu: 1
      requests:
        cpu: 1
```

The above will guarantee that the corresponding container is allocated to a compute instance where at least 1 cpu resource is free and can be dedicated to it.

### Steps to reproduce the problem

Use the k8s executor

### Program output 

Workflow runs but compute instances are heavily over-subscribed.

### Environment 

* Nextflow version: `19.10.0`
* Java version: (n/a)
* Operating system: `Kubernetes 1.16.4`
* Bash version: (n/a)

### Additional context

Related to issue #773 which discusses resource limits.

pditommaso:
Related to #773.

pditommaso:
bump

pditommaso:
A bit confused about this, docs say clearly that omitting `request` and specifying `limit`, the scheduler use the same value for both 

https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits

Why are you saying it assume 0.1 CPU?

KevinSayers:
It depends on the k8s implementation you are using I believe, as the defaults can be configured for a namespace. For example see GKE https://cloud.google.com/blog/products/gcp/kubernetes-best-practices-resource-requests-and-limits under the "Container Settings" section.

pditommaso:
I see. Then I think the quick fix is to explicitly specify both `request` and `limit`. Then we should think a bit more about how to extend the `cpus` directive to allow people to specify both of them. I would prefer to a general solution than Pod only one.

KevinSayers:
Ok so for now I will update the pod spec so it also has `requests` based on the `cpus` directive. What do you think about having a default if the `cpus` directive isn't present? Default to 1?

pditommaso:
That's the default for all task 

[...]

PR #1838, opened 2020-12-18
k8s limits addition

KevinSayers:
Adding limits as discussed in #1570


KevinSayers:
Proposed changes @pditommaso after looking at this again I think some further changes would be better. 
* [ ] The `requests` for cpu defaults to 1 no limit by default. This should help with oversubscribing 
* [ ]  Add request and limit to pod directive. `pod limit: cpu, value: 10` `pod request: cpu, value 2`
* [ ] Refactor the accelerator requests to a general requests method.

pditommaso:
Not completely sure. need to understand K8s default, let's discuss on #1570

pditommaso:
Closing this since #1570 has been solved.

Issue #1486, opened 2020-02-04
DAG edges not connected correctly


mes5k:
## Bug report 

The DAG image generated with the `-with-dag` option doesn't produce the correct output. There are erroneous, unlabeled edges added and correctly labeled edges aren't connected to the correct operators and processes.

### Expected behavior and actual behavior

In the example below, there are three edges incoming to the `combine` node, when there should only be two and one of the edges that _should_ be incoming to the `combine` node (`chan_two`) is not.

The edge between the `example` and `view` nodes is labeled with "-" and I'm not sure where that's coming from.

### Steps to reproduce the problem

This program illustrates the problem:

```
nextflow.preview.dsl=2

workflow {
    process example {
        input:
        tuple val(x), val(y)

        output:
        stdout()

        script:
        """
        printf "${x}${y}"
        """
    }

    chan_one = Channel.from(1,2,3)
    chan_two = Channel.from("a","b","c")

    chan_one \
        | combine(chan_two) \
        | example \
        | view

}
```
Run `nextflow run -with-dag bad.png main.nf` to reproduce.

### Program output 

This is the generated dot file:
```
digraph "bad" {
p0 [shape=point,label="",fixedsize=true,width=0.1,xlabel="Channel.from"];
p5 [shape=circle,label="",fixedsize=true,width=0.1,xlabel="combine"];
p0 -> p5 [label="chan_one"];

p1 [shape=point,label="",fixedsize=true,width=0.1,xlabel="Channel.from"];
p2 [shape=point];
p1 -> p2 [label="chan_two"];

p3 [shape=point,label="",fixedsize=true,width=0.1];
p5 [shape=circle,label="",fixedsize=true,width=0.1,xlabel="combine"];
p3 -> p5;

p4 [shape=point,label="",fixedsize=true,width=0.1];
p5 [shape=circle,label="",fixedsize=true,width=0.1,xlabel="combine"];
p4 -> p5;

p5 [shape=circle,label="",fixedsize=true,width=0.1,xlabel="combine"];
p6 [label="example"];
p5 -> p6;

p6 [label="example"];
p7 [shape=circle,label="",fixedsize=true,width=0.1,xlabel="view"];
p6 -> p7 [label="-"];

p7 [shape=circle,label="",fixedsize=true,width=0.1,xlabel="view"];
p8 [shape=point];
p7 -> p8;

}
```

I would expect it to generate something more like this:

```
digraph "expected" {

[...]

PR #1839, opened 2020-12-25
Added fix for duplicate DataflowBroadcast channels in DAG (1486)

mes5k:
Signed-off-by: Mike Smoot <mes@aescon.com>

This PR fixes https://github.com/nextflow-io/nextflow/issues/1486.  The basic idea is that it keeps track of the DataflowBroadcast to DataflowReadChannel conversion that happens in `OpCall.read0` so that when we create the DAG we're able to lookup the expected DataflowBroadcast from the provided DataflowReadChannel to connect the graph correctly.  I've added a unit test and tested on the example that shows the problem, but haven't tested too extensively otherwise.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/extension/OpCall.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/extension/OpCall.groovy
@@ -126,14 +126,19 @@ class OpCall implements Callable {
     Object[] getArgs() { args }
 
     private <T> T read0(source){
+        T readChan = (T)source;
         if( source instanceof DataflowBroadcast )
-            return (T)CH.getReadChannel(source)
+            readChan = (T)CH.getReadChannel(source)
 
         if( source instanceof DataflowQueue )
-            return (T)CH.getReadChannel(source)
+            readChan = (T)CH.getReadChannel(source)
 
-        else
-            return (T)source
+        if( !ignoreDagNode )
+            // Keep track of this relationship so we can retrieve the
+            // DataflowBroadcast or DataflowQueue when rendering the DAG.
+            NodeMarker.addDataflowBroadcastPair(readChan, source)
+
+        return readChan
     }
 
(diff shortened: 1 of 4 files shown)


modules/nextflow/src/main/groovy/nextflow/extension/OpCall.groovy
```diff
@@ -126,14 +126,19 @@ class OpCall implements Callable {
...
         if( source instanceof DataflowQueue )
-            return (T)CH.getReadChannel(source)
+            readChan = (T)CH.getReadChannel(source)
 
-        else
-            return (T)source
+        if( !ignoreDagNode )
+            // Keep track of this relationship so we can retrieve the
+            // DataflowBroadcast or DataflowQueue when rendering the DAG.
+            NodeMarker.addDataflowBroadcastPair(readChan, source)
```
pditommaso, line 139: Wonder if this should also include ` source != readChan`

pditommaso (commented):
Nice. Tests are greens! i've just made a minor comment.

modules/nextflow/src/main/groovy/nextflow/extension/OpCall.groovy
mes5k: I put that check here: https://github.com/mes5k/nextflow/blob/bug_1486_dag_edge_fix/modules/nextflow/src/main/groovy/nextflow/dag/DAG.groovy#L81 in case we need to use that method anywhere else.  Let me know if you'd like me to move it.

pditommaso: Fine than!

pditommaso:
Ok, merged manually 👉 cece1161. Thanks for the patch! this was tricky!

mes5k:
Happy to help, cheers!

Issue #1840, opened 2020-12-29
Jobs that are redirected to another server by PBSPro are considered as "Failed" 


IdoBar:
## Bug report 

This issue seems similar to #412 . 
When running `nf-core/eager` on the QLD HPC system ([QRIS Awoonga](https://www.qriscloud.org.au/support/qriscloud-documentation/92-awoonga-user-guide), running `PBSpro`), jobs that get redirected to another queue (on another server) are being referred to as "failed" with this error message `terminated for an unknown reason -- Likely it has been terminated by the external system`, even tough they complete successfully with an exit status 0 (which nextflow is unable to get).

### Expected behavior and actual behavior

The workflow should keep tracing the jobs even if they're automatically redirected to another queue/server.
I'm not sure how nextflow is tracking job progress, but my guess is that it has to do with the job status being changed to `M` instead of `R` or that nextflow is using the full `job_id.server` to track the job progress rather that using just `job_id` (see output of `qstat` below)

This is the `qstat` summary of a job that had "failed" according to nextflow:
```
Job Id: 510838.awonmgr2
    Job_Name = nf-genotyping_f
    Job_Owner = ibar@awoonga1.local
    job_state = M
    queue = Short@flashmgr2
    server = awonmgr2
    Account_Name = qris-gu
    Checkpoint = u
    ctime = Fri Dec 25 09:28:22 2020
    Error_Path = awoonga1.local:/gpfs1/scratch/30days/ibar/data/Dingo/Dingo_aDNA_NF13_process_24_12_2020/work/83/4cb7e477c5ad7a5236ba0d6d314708/nf-genotyping_f.e510838
    Hold_Types = n
    Join_Path = oe
    Keep_Files = n
    Mail_Points = a
    mtime = Fri Dec 25 09:29:59 2020
    Output_Path = awoonga1.local:/gpfs1/scratch/30days/ibar/data/Dingo/Dingo_aDNA_NF13_process_24_12_2020/work/83/4cb7e477c5ad7a5236ba0d6d314708/.command.log
    Priority = 0
    qtime = Fri Dec 25 09:28:22 2020
    Rerunable = True
    Resource_List.interact1 = 0
    Resource_List.mem = 4096mb
    Resource_List.ncpus = 2
    Resource_List.nodect = 1
    Resource_List.place = free
    Resource_List.select = 1:ncpus=2:mem=4096mb
    Resource_List.walltime = 04:00:00
    substate = 92
[...]

PR #1850, opened 2021-01-11
Check queue status across all available servers

IdoBar:
Check queue status for user-submitted jobs across all queues in all available servers. 
This should fix issue #1840 
Please verify the escaping of the single quotes in line 85.


IdoBar (commented):
This had been tested on Awoonga HPC running PBS Pro v18.1.2

abhi18av:
@IdoBar thanks for this PR!

To make the `Travis CI` build happy regarding the tests, you'd need to update this test here https://github.com/nextflow-io/nextflow/blob/402339f6b054774375ef97abb86d95f87bfa18e6/modules/nextflow/src/test/groovy/nextflow/executor/PbsProExecutorTest.groovy#L150

IdoBar:
@abhi18av, I've updated the test as well.

modules/nextflow/src/main/groovy/nextflow/executor/PbsProExecutor.groovy
```diff
@@ -82,8 +82,8 @@ class PbsProExecutor extends PbsExecutor {
 
     @Override
     protected List<String> queueStatusCommand(Object queue) {
-        String cmd = 'qstat -f'
-        if( queue ) cmd += ' ' + queue
+        String cmd = 'qstat -u $USER -f $( qstat -B | egrep -v \'(^Server|^---)\' | awk -v ORS=" " \'{print "@"$1}\' )'
+       // if( queue ) cmd += ' ' + queue
```
IdoBar, line 86:
```suggestion
        String cmd = 'qstat -u $USER -f ' 
        if( queue ) {
            cmd += queue
        } else {
            cmd += '$( qstat -B | egrep -v \'(^Server|^---)\' | awk -v ORS=" " \'{print "@"$1}\' )'
        }
```

IdoBar:
@pditommaso, @abhi18av, thanks for your feedback, it took me a few iterations, but I managed to pass the tests.

pditommaso:
What exactly `$( qstat -B | egrep -v \'(^Server|^---)\' | awk -v ORS=\' \' \'{print \"@\"\$1}\' )` does ?

IdoBar:
> 
> 
> What exactly `$( qstat -B | egrep -v \'(^Server|^---)\' | awk -v ORS=\' \' \'{print \"@\"\$1}\' )` does ?

It parses all the available servers for the scheduler and makes a list of arguments for `qstat`, so on my HPC, when jobs are automatically redirected to multiple queues at different servers, it will run `qstat -u $USER -f @server1 @server2 @server3` and will capture all the jobs for that user.
I left the original command as is if the user supplies a specific `queue` parameter.

pditommaso:
Is this portable on other PBS cluster? you mentioned you are using a modified version.

IdoBar:
> 
> 
> Is this portable on other PBS clusters? you mentioned you are using a modified version.

Yes, as I've mentioned above it's working as intended on other PBS clusters, whether they support automatic job redirection to other servers or not (it uses "regular" `qstat` commands and I've tested it on another cluster that I have access to).
With this patch I won't need to use my modified version of `qstat` (which is the whole purpose of it).

modules/nextflow/src/main/groovy/nextflow/executor/PbsProExecutor.groovy
```diff
@@ -82,8 +82,12 @@ class PbsProExecutor extends PbsExecutor {
 
     @Override
     protected List<String> queueStatusCommand(Object queue) {
-        String cmd = 'qstat -f'
-        if( queue ) cmd += ' ' + queue
+        String cmd = 'qstat -u $USER -f '
```
IdoBar, line 85:
```suggestion
        String cmd = 'qstat -f ' 
```

modules/nextflow/src/test/groovy/nextflow/executor/PbsProExecutorTest.groovy
```diff
@@ -150,10 +150,11 @@ class PbsProExecutorTest extends Specification {
...
         given:
         def executor = [:] as PbsProExecutor
-
+		def user = System.getenv('USER')
+		
         expect:
-        executor.queueStatusCommand(null) == ['bash','-c', "set -o pipefail; qstat -f | { egrep '(Job Id:|job_state =)' || true; }"]
-        executor.queueStatusCommand('xxx') == ['bash','-c', "set -o pipefail; qstat -f xxx | { egrep '(Job Id:|job_state =)' || true; }"]
+        executor.queueStatusCommand('xxx') == ['bash','-c', "set -o pipefail; qstat -u \$USER -f xxx | { egrep '(Job Id:|job_state =)' || true; }"]
+		executor.queueStatusCommand(null) == ['bash','-c', "set -o pipefail; qstat -u \$USER -f \$( qstat -B | egrep -v '(^Server|^---)' | awk -v ORS=' ' '{print \"@\"\$1}' ) | { egrep '(Job Id:|job_state =)' || true; }"]
```
IdoBar, line 157:
```suggestion
        executor.queueStatusCommand('xxx') == ['bash','-c', "set -o pipefail; qstat -f xxx | { egrep '(Job Id:|job_state =)' || true; }"]
		executor.queueStatusCommand(null) == ['bash','-c', "set -o pipefail; qstat -f \$( qstat -B | egrep -v '(^Server|^---)' | awk -v ORS=' ' '{print \"@\"\$1}' ) | { egrep '(Job Id:|job_state =)' || true; }"]
```

pditommaso:
I noticed the `-u` option has gone. Is this expected to be the final version?

IdoBar:
@pditommaso Yes, this version is tested and working as intended on several HPCs running PBSPro and I'll try to test it on a few others.

pditommaso:
@arnaudceol @cjfields If I'm not wrong you heavily really on the PSB pro executor for Nextflow. Could you validate the proposed change in this PR to make that works in your cluster?   

Essentially it consists of using the following command line to fetch the queue(s) status 

```
set -o pipefail; $( qstat -B | egrep -v '(^Server|^---)' | awk -v ORS=' ' '{print "@"$1}' ) | { egrep '(Job Id:|job_state =)' || true; }
```

instead of 

```
set -o pipefail; qstat -f <queue name> | { egrep '(Job Id:|job_state =)' || true; }
```

Thanks a lot

cjfields:
@pditommaso would love to help but our local HPC switched primarily to SLURM a bit ago.  I can check with our NCSA colleagues, I think they may be running PBS Pro on some systems.

pditommaso:
I see, at your convenience! thanks a lot for your swift reply!

IdoBar:
> 
> 
> @arnaudceol @cjfields If I'm not wrong you heavily really on the PSB pro executor for Nextflow. Could you validate the proposed change in this PR to make that works in your cluster?
> 
> Essentially it consists of using the following command line to fetch the queue(s) status
> 
> ```
> set -o pipefail; $( qstat -B | egrep -v '(^Server|^---)' | awk -v ORS=' ' '{print "@"$1}' ) | { egrep '(Job Id:|job_state =)' || true; }
> ```
> 
> instead of
> 
> ```
> set -o pipefail; qstat -f <queue name> | { egrep '(Job Id:|job_state =)' || true; }
> ```
> 
> Thanks a lot

Just a correction, , the command still starts with `qstat -f`: 
```
set -o pipefail; qstat -f $( qstat -B | egrep -v '(^Server|^---)' | awk -v ORS=' ' '{print "@"$1}' ) | { egrep '(Job Id:|job_state =)' || true; }
```

pditommaso:
My fault, I've misread it. Updated the comment.

arnaudceol:
@pditommaso  I confirm that the command 

```
set -o pipefail; qstat -f $( qstat -B | egrep -v '(^Server|^---)' | awk -v ORS=' ' '{print "@"$1}' ) | { egrep '(Job Id:|job_state =)' || true; }
```

returns the list jobs on all queues (we have only one server, pbs_version = 19.1.2).

IdoBar:
I can also confirm that it works as intended on the HPC at QUT, running PBS version 2020.1.0.20200720152739

pditommaso:
Nice. Thanks a lot both! My only concern is the amount of output this command can produce checking all queues. In a busy cluster, it can be thousands of tasks. What it's preventing limiting `qstat` to the submitting user?

IdoBar:
> 
> 
> Nice. Thanks a lot both! My only concern is the amount of output this command can produce checking all queues. In a busy cluster, it can be thousands of tasks. What it's preventing limiting `qstat` to the submitting user?

`qstat -f` doesn't work with `-u`, these 2 flags are mutually exclusive (except for some reason on our specific HPC cluster).
If a specific queue is provided, however, then only the jobs of that queue will be listed (as it was previously).
I think that since each job ID is collected at submission, it would be best to check only the status of these jobs with `qstat -f $JOB_ID1 $JOB_ID2 ...`.
Alternatively, the running status of the jobs can be parsed from the short version of `qstat -u $USER`, without the `-f` flag (which is must easier to parse) - see an example of the output:
```
> qstat -u $USER
gc-prd-hpcadm:
                                                            Req'd  Req'd   Elap
Job ID          Username Queue    Jobname    SessID NDS TSK Memory Time  S Time
--------------- -------- -------- ---------- ------ --- --- ------ ----- - -----
79184.gc-prd-hp s5000942 workq    KimJob      49510   1  16   36gb 300:0 R 276:4
79974.gc-prd-hp s5000942 workq    KimJob      67429   1   4   24gb 300:0 R 245:3
80145.gc-prd-hp s5000942 workq    KimJob     177233   1   6   24gb 300:0 R 236:0
81011.gc-prd-hp s5000942 workq    KimJob     118637   1   4   24gb 200:0 R 196:3
81407.gc-prd-hp s5000942 workq    KimJob      38131   1   4   24gb 200:0 R 172:5
81408.gc-prd-hp s5000942 workq    KimJob      36075   1   4   24gb 200:0 R 172:5
81680.gc-prd-hp s5000942 workq    KimJob      49214   1  16   24gb 500:0 R 71:33
81683.gc-prd-hp s5000942 workq    KimJob      37961   1   4   24gb 300:0 R 71:25
81695.gc-prd-hp s5000942 workq    KimJob      44326   1   4   24gb 300:0 R 71:21
81699.gc-prd-hp s5000942 workq    KimJob     166262   1   4   24gb 300:0 R 71:02
81700.gc-prd-hp s5000942 workq    KimJob     149223   1   4   24gb 300:0 R 70:52
81703.gc-prd-hp s5000942 workq    KimJob     165458   1   4   24gb 300:0 R 70:38
81877.gc-prd-hp s5000942 workq    KimJob      72323   1   4   24gb 300:0 R 27:59
```
That can be easily parsed to extract the Job ID and Job State with the following
```
> qstat -u $USER | grep "$USER" | awk '{print $1,$10}'
79184.gc-prd-hp R
79974.gc-prd-hp R
80145.gc-prd-hp R
81011.gc-prd-hp R
81407.gc-prd-hp R
81408.gc-prd-hp R
81680.gc-prd-hp R
81683.gc-prd-hp R
81695.gc-prd-hp R
81699.gc-prd-hp R
81700.gc-prd-hp R
81703.gc-prd-hp R
81877.gc-prd-hp R
````

I think that will be much faster and more focused than parsing the entire server(s) or entire queue, but I just need to verify that the parsing command will work on my cluster when jobs are spread across multiple servers (can't see a reason why it wouldn't).

**Comment by pditommaso** [comment]
The -x flag is very useful and I agree that the `qstat -xu` would be a better solution, the problem is that we need to validate across different installations.

**Comment by arnaudceol** [comment]
> The -x flag is very useful and I agree that the `qstat -xu` would be a better solution, the problem is that we need to validate across different installations.

qstat -xu is one I use on our cluster. Be careful, it may provide really long lists. It shows the jobs on all queues (cannot confirm on a multi-server installation).

To get more info about the jobs (if it may help), -ant can be used:

```
# qstat -u MyID

hpcfe01: 
                                                            Req'd  Req'd   Elap
Job ID          Username Queue    Jobname    SessID NDS TSK Memory Time  S Time
--------------- -------- -------- ---------- ------ --- --- ------ ----- - -----
436184.hpcfe01  MyID  workq    pathseq.sh 121545   1   2   40gb   --  R 67:40




# qstat -antu MyID

hpcfe01: 
                                                            Req'd  Req'd   Elap
Job ID          Username Queue    Jobname    SessID NDS TSK Memory Time  S Time
--------------- -------- -------- ---------- ------ --- --- ------ ----- - -----
436184.hpcfe01  MyID  workq    pathseq.sh 121545   1   2   40gb   --  R 67:40
   cn01/3*2
```

pditommaso:
I guess the list gets long when specifying the `-x` because it contains the completed jobs? is so, is there any flag to specify how old complete jobs can be e.g. last hour?

IdoBar:
Is there an actual need to list completed jobs?
At the moment it doesn't list these jobs either with `qstat -f`, so I guess it can stay that way.

pditommaso:
> At the moment it doesn't list these jobs either with qstat -f, so I guess it can stay that way.

That's correct, however, having the ability to check the status of completed jobs would allow the better handing of killed jobs. When the status is not detected from the `qstat` command NF fallback checking on the `.exitcode` file. But this file cannot be created if the job is hard killed by the cluster.

In any case, I'm fine with this patch, since it's only used when the `queue` is not specified.

pditommaso:
Merged manually. 👉  c159890. Thanks for your contribution.

PR #1853, opened 2021-01-12
Charliecloud improvements

phue:
Hi @pditommaso,

I have encountered some rough edges when running nextflow with charliecloud, which occur when running from the user $HOME directory.

I think the problem is that Charliecloud does not support passing mountflags with bind-mounted directories as opposed to Singularity for example (I believe that is what the `readOnlyInputs` flag is there for, please correct me if I'm wrong).

So in order to make sure the work directory is writeable, this removes the `readOnlyInputs` logic for Charliecloud entirely and ensures that mounted directories are always writeable in the container.

Further changes:

* addresses a breaking change that will land in Charliecloud 0.23 (`ch-grow pull` is now `ch-image pull`)
* adds the functionality to specify a directory that is mounted to /tmp within the container

### DIFF
--- a/docs/charliecloud.rst
+++ b/docs/charliecloud.rst
@@ -21,7 +21,7 @@ executed on any platform supporting the Charliecloud engine.
 Prerequisites
 ==============
 
-You will need Charliecloud version ``0.19`` or later installed on your execution environment e.g. your computer or a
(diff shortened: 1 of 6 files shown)


pditommaso:
Awesome!

PR #1862, opened 2021-01-21
Decrease s3 cp sleep

jchorl:
Hi!

tbh, I'm not sure what this will impact. We noticed a process sleeping for an hour+ due to s3 import of 4000+ files. I'm sure aws can handle >1rps. I'm not sure how much. I also see this used in the google life sciences handler.

From `man sleep` on my ubuntu focal machine:
```
 Unlike most implementations that require NUMBER be an integer, here NUMBER may
       be an arbitrary floating point number.
```

Again, I'm not sure what this'll impact. But less wait would be :ok_hand:


pditommaso:
Thanks for pointing this out. i've merged your PR manually setting the sleep to 0.2. It should not harm.

PR #1881, opened 2021-02-03
[Enhancement] Update the default auto-scale formula for Azure Batch Pool resources

abhi18av:
This PR adds support for scaling down for Azure Batch Pool resources to 0 when there are no tasks.

- The pool was created with auto-scale formula (initially 0 nodes)
- Nodes were added after tasks were added to a linked job
- Tasks were executed
- After around 15 minutes nodes will be removed and the count will be set to 0.

### DIFF
--- a/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
+++ b/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
@@ -554,14 +554,14 @@ class AzBatchService implements Closeable {
 
     protected String scaleFormula(AzPoolOpts opts) {
         // https://docs.microsoft.com/en-us/azure/batch/batch-automatic-scaling
-        def DEFAULT_FORMULA = '''\
-                $TargetDedicatedNodes = {{vmCount}};
-                lifespan         = time() - time("{{now}}");
-                span             = TimeInterval_Minute * 60;
-                startup          = TimeInterval_Minute * 10;
-                ratio            = 50;
-                $TargetDedicatedNodes = (lifespan > startup ? (max($RunningTasks.GetSample(span, ratio), $ActiveTasks.GetSample(span, ratio)) == 0 ? 0 : $TargetDedicatedNodes) : {{vmCount}});
-                '''.stripIndent()
+        def DEFAULT_FORMULA = """
+            cappedPoolSize = ${opts.vmCount};
+            \$samples = \$PendingTasks.GetSamplePercent(TimeInterval_Minute * 15);
+            \$tasks = \$samples < 70 ? max(0, \$PendingTasks.GetSample(1)) : max( \$PendingTasks.GetSample(1), avg(\$PendingTasks.GetSample(TimeInterval_Minute * 15)));
+            \$targetVMs = \$tasks > 0? \$tasks:max(0, \$TargetDedicatedNodes/2);
+            \$TargetDedicatedNodes = max(0, min(\$targetVMs, cappedPoolSize));
+            \$NodeDeallocationOption = taskcompletion;
+                """.stripIndent()
 
         final scaleFormula = opts.scaleFormula ?: DEFAULT_FORMULA
         final vars = new HashMap<String,String>()
--- a/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
+++ b/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
@@ -555,17 +555,34 @@ class AzBatchService implements Closeable {
     protected String scaleFormula(AzPoolOpts opts) {
         // https://docs.microsoft.com/en-us/azure/batch/batch-automatic-scaling
         def DEFAULT_FORMULA = '''
+            evaluationInterval = {{interval}};
             cappedPoolSize = {{vmCount}};
-            $samples = $PendingTasks.GetSamplePercent(TimeInterval_Minute * 15);
-            $tasks = $samples < 70 ? max(0, $PendingTasks.GetSample(1)) : max( $PendingTasks.GetSample(1), avg($PendingTasks.GetSample(TimeInterval_Minute * 15)));
-            $targetVMs = $tasks > 0? $tasks:max(0, $TargetDedicatedNodes/2);
-            $TargetDedicatedNodes = max(0, min($targetVMs, cappedPoolSize));
+            timeOfPoolCreation = {{poolCreationTime}};
+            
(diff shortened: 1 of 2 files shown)


abhi18av:
Converted to draft, while investigating the reason behind 
image

abhi18av:
This is functional now, the issue seems to be a bit tangential and related to Docker Client/Compose HTTP timeout settings.

image

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -554,14 +554,14 @@ class AzBatchService implements Closeable {
...
-                $TargetDedicatedNodes = (lifespan > startup ? (max($RunningTasks.GetSample(span, ratio), $ActiveTasks.GetSample(span, ratio)) == 0 ? 0 : $TargetDedicatedNodes) : {{vmCount}});
-                '''.stripIndent()
+        def DEFAULT_FORMULA = """
+            cappedPoolSize = ${opts.vmCount};
+            \$samples = \$PendingTasks.GetSamplePercent(TimeInterval_Minute * 15);
+            \$tasks = \$samples < 70 ? max(0, \$PendingTasks.GetSample(1)) : max( \$PendingTasks.GetSample(1), avg(\$PendingTasks.GetSample(TimeInterval_Minute * 15)));
+            \$targetVMs = \$tasks > 0? \$tasks:max(0, \$TargetDedicatedNodes/2);
+            \$TargetDedicatedNodes = max(0, min(\$targetVMs, cappedPoolSize));
+            \$NodeDeallocationOption = taskcompletion;
+                """.stripIndent()
```
pditommaso, line 565: Instead of using triple `"` use triple `'`, the variables are resolved using `{{vmCount}}` syntax

abhi18av: Paolo, I tried that earlier however it seems that the string resolution isn't working as expected.

The Azure API threw this error (via NF) 

```

Caused by:

  "odata.metadata":"https://nxfazrg.eastus.batch.azure.com/$metadata#Microsoft.Azure.Batch.Protocol.Entities.Container.errors/@Element","code
    "lang":"en-US","value":"The specified auto-scaling formula has evaluation error\nRequestId:fb562edb-6898-4237-9833-6e1be45e3204\nTime:202

}


```

On Groovy console, the following code snippet 

```groovy
def vmCount = 1

'''
cappedPoolSize = {{vmCount}};
$samples = $PendingTasks.GetSamplePercent(TimeInterval_Minute * 15);
$tasks = $samples < 70 ? max(0, $PendingTasks.GetSample(1)) : max( $PendingTasks.GetSample(1), avg($PendingTasks.GetSample(TimeInterval_Minute * 15)));
$targetVMs = $tasks > 0? $tasks:max(0, $TargetDedicatedNodes/2);
$TargetDedicatedNodes = max(0, min($targetVMs, cappedPoolSize));
$NodeDeallocationOption = taskcompletion;
'''.stripIndent()
```

...resolves to 

```
cappedPoolSize = {{vmCount}};
$samples = $PendingTasks.GetSamplePercent(TimeInterval_Minute * 15);
$tasks = $samples < 70 ? max(0, $PendingTasks.GetSample(1)) : max( $PendingTasks.GetSample(1), avg($PendingTasks.GetSample(TimeInterval_Minute * 15)));
$targetVMs = $tasks > 0? $tasks:max(0, $TargetDedicatedNodes/2);
$TargetDedicatedNodes = max(0, min($targetVMs, cappedPoolSize));
$NodeDeallocationOption = taskcompletion;

```

pditommaso:
Cant' be. there's a unit test (that should be update with your version)

https://github.com/nextflow-io/nextflow/blob/101a67c13ec657ebdd5bbd77a3b49f18a6d4fc78/plugins/nf-azure/src/test/nextflow/cloud/azure/batch/AzBatchServiceTest.groovy#L211-L221

pditommaso:
👇 😎 
http://spockframework.org/spock/docs/1.3/spock_primer.html

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
pditommaso: Because it needs to be parsed as it's done by this line 

https://github.com/nextflow-io/nextflow/blob/687043d66d40140265c54f3aef1672931b51a2d7/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L570

abhi18av: Ah, now it makes sense. I've updated the formula since we don't need to use `opts.vmCount` directly in the formula due to the template.

pditommaso: >  I've updated the formula since we don't need to use opts.vmCount

I still see it there `cappedPoolSize = {{vmCount}};`. 

Also is there a way to define the set the num of VMs once the pool is first created, to avoid to await that it scales from 0 to some instances at the beginning?

abhi18av: Yup, we just need to tweak the `auto-scale DSL` a bit - how about initiating a new pool with 1 machine by default ?

pditommaso: I would like to use `vmCount` for the initial VMs and a `maxVmCount` for the capped

abhi18av: > I still see it there cappedPoolSize = {{vmCount}};.

Yeah, this works but `opts.vmCount` wasn't working because of the usage of `vars` later on.

> ...to avoid to await that it scales from 0 to some instances at the beginning?

We could keep the first node as a low priority node maybe OR reduce the formula evaluation time 🤔  

Anyways, there are plenty of ways to optimize the rescaling however they shouldn't be a blocker as of now. The intent of current formula is to follow good heuristic for scaling up and down and faster formula evaluation.

abhi18av:
@pditommaso  I've included time computations to accomodate the following

- Deploy one node by default upon pool creation
- After first interval, scale up/down based on available tasks

pditommaso:
Ok, merged manually with some changes. See 9c791617. Tx!

abhi18av:
Happy to be of help :)

PR #1884, opened 2021-02-11
catch failing commands in nxf_parallel

jchorl:
Greetings!

We hit a brutal edge case - got ratelimited by the metadata endpoint while running thousands of containers through an ec2 instance. Unfortunately `nxf_parallel` is not catching errors, so nextflow failed to stage files and yet the process chugged along. This was no bueno!

Anyway, here's a fix with some tests.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashFunLib.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashFunLib.groovy
@@ -59,7 +59,10 @@ class BashFunLib {
                   ((i+=1))
                 fi
             done
-            ((\${#pid[@]}>0)) && wait \${pid[@]}
+            for p in "\${pid[@]}"
+            do
+                wait \$p
+            done
             )
             unset IFS
         }
--- a/modules/nextflow/src/test/groovy/nextflow/executor/BashFunLibTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/executor/BashFunLibTest.groovy
@@ -1,0 +1,79 @@
+package nextflow.executor
+
+import java.util.concurrent.TimeUnit
+import java.nio.file.Files
+
+import nextflow.util.Duration
+import spock.lang.Specification
+
+/**
+ *
+ * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
+ */
+class BashFunLibTest extends Specification {
+
+    def 'should fail on errors in nxf_parallel' () {
+
+        given:
+        def scriptFile = Files.createTempFile("test", "sh")
+        def script = """
+        #!/bin/bash
+
+        set -e
+        """.stripIndent() +
+
+        BashFunLib.body(5, 5, Duration.of('1 sec')) +
+
+        """
+        cmds=()
+        cmds+=("true")
+        cmds+=("false")
+        nxf_parallel "\${cmds[@]}"
+        """.stripIndent()
+
+        scriptFile.text = script
+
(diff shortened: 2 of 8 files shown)


modules/nextflow/src/main/groovy/nextflow/executor/BashFunLib.groovy
```diff
@@ -59,7 +59,10 @@ class BashFunLib {
                   ((i+=1))
                 fi
             done
-            ((\${#pid[@]}>0)) && wait \${pid[@]}
+            for p in "\${pid[@]}"
+            do
+                wait \$p
+            done
```
pditommaso, line 65: Interesting, how does this version help propagating the error compared to `wait pid1 pid2 etc` ?

jchorl: `wait pid1 pid2 etc` doesn't return non-zero if any of the pids fail. `wait pid1` will return non-zero if the pid fails (and hence bomb due to `set -e`).

From what I can tell, `wait pid1 pid2 etc` isn't a valid use of `wait`: https://man7.org/linux/man-pages/man2/wait.2.html

jchorl: Nevermind, it is a valid use:
```
The exit
       status returned by the wait utility shall be the exit status of
       the process requested by the last pid operand.
```

From: https://www.man7.org/linux/man-pages/man1/wait.1p.html

pditommaso: I see, make sense. But what happens when pid2 fails *before* pid1 terminates correctly? is still the error reported? by which wait?

jchorl: In that case, it'll wait for `pid1`, and then check `pid2`, see it failed, and nuke. From what I read, you actually can wait on procs in different orders than they finish, and it still works.

```
Therefore, the status of background processes was usually lost unless it terminated while wait was waiting for it. This could be  a
       serious	problem  when  a  job  that  was expected to run for a long time actually terminated quickly with a syntax or initialization error
       because the exit status returned was usually zero if the requested process ID was not found. This volume of  IEEE Std 1003.1-2001  requires
       the implementation to keep the status of terminated jobs available until the status is requested, so that scripts like:

	      j1&
	      p1=$!
	      j2&
	      wait $p1
	      echo Job 1 exited with status $?
	      wait $!
	      echo Job 2 exited with status $?

       work without losing status on any of the jobs.
```

- https://www.unix.com/man-page/linux/1/wait/

pditommaso: ok, this is neat.

modules/nextflow/src/test/groovy/nextflow/executor/BashFunLibTest.groovy
```diff
@@ -0,0 +1,79 @@
...
+        def process = "bash ${scriptFile}".execute()
+        process.waitFor(1, TimeUnit.SECONDS)
+
+        expect:
+        process.exitValue() == 0
+
+        cleanup:
+        if( scriptFile ) Files.delete(scriptFile)
+
+    }
```
pditommaso, line 77: Nice job 👏

pditommaso:
Merged manually  👉 dd56f6c1. Thanks for this contribution.

PR #1909, opened 2021-02-19
charliecloud: get container environment from file

phue:
Hi,

With charliecloud v[0.22](https://github.com/hpc/charliecloud/releases/tag/v0.22), ENV layers of docker containers are saved to
a file located in `$IMAGE/ch/environment` when the container is pulled.

Hence it is no longer necessary to pass for example PATH variables explicitly via the env scope, because we can just read it from `/ch/environment`


pditommaso (approved):
Awesome. thanks

PR #1910, opened 2021-02-19
docs: add undocumented features

phue:
I came across two instances of features that were not documented
* `.collectFile(cache:true)`
* `conda.createOptions`


pditommaso:
Thanks for contributing and apology for the late reply. Marged manually ba761d64.

PR #1916, opened 2021-02-22
Update the default AZ vm for vanilla accounts

abhi18av:
Hi @pditommaso ,

I was testing the plugin (and updating the docs) wrt to a new vanilla azure account.

I feel that we should choose the default VM as the one which is available for all default/vanilla/capped Azure accounts. 

For testing, I've used the `nextflow-io/hello` and the `nextflow-io/rnaseq-nf` pipelines with the default settings, but both of these fail due to quota issues. However, after passing 

```nextflow
 azure.batch.pools.auto.vmType = 'STANDARD_D2_V3'
```

... as extra configs in the Launch dialog, both pipeline successfully complete execution.

image

Would it be possible to make an edge release with the new default or shall I stick to the explicit mention of this tweak as part of the docs?

### DIFF
--- a/plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
+++ b/plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
@@ -38,7 +38,7 @@ class AzPoolOpts implements CacheFunnel {
 
     static public final String DEFAULT_PUBLISHER = "microsoft-azure-batch"
     static public final String DEFAULT_OFFER = "centos-container"
-    static public final String DEFAULT_VM_TYPE = "Standard_A3"
+    static public final String DEFAULT_VM_TYPE = "STANDARD_D2_V3"
     static public final OSType DEFAULT_OS_TYPE = OSType.LINUX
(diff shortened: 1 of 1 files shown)


abhi18av:
Here's the quota status for default accounts 

image

Please note that the A series would be phased out soon and by default its not available in the quota - only `DV3` i.e. the suggested `STANDARD_D2_V3` is available.

pditommaso:
I think this heavily depends on user quotas. better to mention in the docs.

jordeu:
At Tower I've proposed to use `Standard_D4_v3` it has 4cpus, 16Gb ram, 100Gb HD at ~$0.192/hour and it is available at all the regions.

The current `Standard_A3` it has 4cpus, 7Gb ram, 285Gb HD at ~$0.24/hour and to use Av2 series is encourage instead of this.

Also, I've created a Batch at three different regions and only `Dv3` series where available using the free tier.

pditommaso:
Fair enough

abhi18av:
Yeah, I was concerned about new Azure account users as well. 

Please let me know if this change might be useful for users or shall I close?

pditommaso:
ok, merged manually using `Standard_D4_v3` as default

abhi18av:
Cool :)

PR #1929, opened 2021-02-24
Clarify how template files can be tested independently from the pipeline

brandoncazander:
This is a super minor change but may prevent some confusion about one of the benefits of using template files—being able to test them independently from the nextflow pipeline.

In this case, using the ablative preposition *from* indicates how the template files may be tested apart from the pipeline.

Signed-off-by: Brandon Cazander <bcazander@canexiahealth.com>


pditommaso (approved):
Thanks a lot

PR #1936, opened 2021-02-27
Sleep with decimal numbers not supported

manuelesimi:
With some non-GNU versions of sleep (like in busybox), sleep does not accept decimal numbers. Here I worked around by sleeping for 1 sec when the first sleep fails. 

These changes have been tested on Azure Batch, locally on MacOS and Ubuntu, and on a SGE cluster.


pditommaso:
Thanks for this. Solved by c4be8d00

PR #1967, opened 2021-03-12
Encode path parameter at GitLab getContentUrl

jordeu:
# Description
Current Gitlab provider implementation fails to retrieve content of files that are not at the root of the repository. This is due to that the `path` parameter at the API URL must be encoded to escape URL characters like a slash.


pditommaso (changes requested):
Nice, worth adding a small test

pditommaso (approved):
Well done

Issue #1860, opened 2021-01-20
[Enhancement] Resolution of dynamic variables like baseDir in params file


abhi18av:
## Bug report 

Hi Paolo, 

Please feel free to close this issue if it's already WIP or there's a decent work-around.

### Expected behavior and actual behavior

If we use the `params` field in `nextflow.config` for `input_fofn = "$baseDir/test_data/manifest-3samples.txt"`  it works as expected. 

However, if we pass this via the `params file` then it fails to be resolved. This could be an issue with pipelines which are adapting for tower.nf 

### Steps to reproduce the problem

<img width="925" alt="image" src="https://user-images.githubusercontent.com/12799326/105221292-abe25d80-5b37-11eb-90aa-b7c693c72288.png">

### Environment 

This bug can be reproduced by simply trying to run any pipeline on Tower. More specifically this has been tested with the following repo and branch https://github.com/seqeralabs/gatk4-germline-snps-indels/pull/22

abhi18av:
Closing this since this feature is now merged.

PR #1972, opened 2021-03-16
Allow variable binding also at params file

jordeu:
Adds the option to alse use variables (following the format `${variable_name}`) when defining parameter values at the JSON or YAML parameters file.

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/config/ConfigBuilderTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/config/ConfigBuilderTest.groovy
@@ -17,6 +17,10 @@
 
 package nextflow.config
 
+import sun.nio.fs.UnixFileSystemProvider
+import sun.nio.fs.UnixPath
+
+import java.nio.file.FileSystems
 import java.nio.file.Files
 import java.nio.file.Paths
 
@@ -455,7 +459,7 @@ class ConfigBuilderTest extends Specification {
         params.text = '''
             alpha: "Hello" 
             beta: "World" 
-            omega: "Last"
+            omega: "${baseDir}/data/omega.txt"            
             '''.stripIndent()
         and:
         def file = Files.createTempFile('test',null)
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/test/groovy/nextflow/config/ConfigBuilderTest.groovy
```diff
@@ -475,12 +479,13 @@ class ConfigBuilderTest extends Specification {
...
-        def result = new ConfigBuilder().setOptions(opt).setCmdRun(run).buildGivenFiles(file)
+        def baseDir = Files.createTempDirectory('assets')
+        def result = new ConfigBuilder().setOptions(opt).setCmdRun(run).setBaseDir(baseDir).buildGivenFiles(file)
 
         then:
         result.params.alpha == 'Hola'   // <-- this comes from the CLI
         result.params.beta == 'Mundo'   // <-- this comes from the CLI as well
-        result.params.omega == 'Last'   // <-- this comes from the params-file
+        result.params.omega == "${baseDir.toUriString()}/data/omega.txt"   // <-- this comes from the params-file and replaces parameters
         result.params.gamma == "I'm gamma"   // <-- from the config
```
pditommaso, line 489: Here the `baseDir` should not be resolved to the actual value?

jordeu: Yes, it is resolved to the actual value. Notice that I'm using double quotes. At this test `baseDir` is a temporary directory that I cannot hardcode beforehand, this is why I'm using the variable itself.

abhi18av:
I guess, this request https://github.com/nextflow-io/nextflow/issues/1860 can now be closed?

jordeu:
> I guess, this request #1860 can now be closed?
Yes, the only thing is that this implementation do not support variables without curly brackets. Only formatted like this `${baseDir}`.

abhi18av:
That should be fine I think since this is explicit. We could include this info in a `Nextflow config (or params) best practices` section.

PR #1973, opened 2021-03-16
Google Lifesciences API slow file staging

aeschriefer:
Improved the speed of google lifescience file staging by allowing two key gsutil parameters to be set using the nextflow config file.  Currently the nextflow defaults are set to the same defaults used by gsutil so no change in performance should be seen.  The only default that I changed was google-sdk:alpine to google-sdk:slim because alpine does not support multiprocessing

More discussion of the issue and benchmarking I did of gsutil on the nextflow forum:
https://groups.google.com/g/nextflow/c/0L7HlD7mmL0

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,7 +48,7 @@ jobs:
     - name: Tests
       run: make test
       env:
-        GRADLE_OPTS: '-Dorg.gradle.daemon=false --stacktrace'
+        GRADLE_OPTS: '-Dorg.gradle.daemon=false'
         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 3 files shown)


pditommaso:
You can run tests locally using `make smoke`

pditommaso:
Stating the snippet below, just using `google-sdk:slim` instead of alpine should be enough to take advantage of multiple processes upload. Does 

```
# On Windows and Alpine Linux, Python multi-processing presents various
# challenges so we retain compatibility with the established parallel mode
# operation, i.e. one process and 24 threads.
should_prohibit_multiprocessing, unused_os = ShouldProhibitMultiprocessing()
if should_prohibit_multiprocessing:
  DEFAULT_PARALLEL_PROCESS_COUNT = 1
  DEFAULT_PARALLEL_THREAD_COUNT = 24
else:
  DEFAULT_PARALLEL_PROCESS_COUNT = min(multiprocessing.cpu_count(), 32)
  DEFAULT_PARALLEL_THREAD_COUNT = 5
```

Therefore `GSUtil:parallel_thread_count` and  `GSUtil:sliced_object_download_max_components` should not be needed. Does make sense?

pditommaso:
Merged manually 2da120b9.

PR #1985, opened 2021-03-23
Parameters set after import are not inherited

### LINKED ISSUE #1984: Module parameter inheritance is only done at import time
The DSL2 documentation is not explicit about the module parameter inheritance when it comes to procedure. Inheritance is only available at the time of inclusion, not later on.

Relevant docs: https://www.nextflow.io/docs/latest/dsl2.html#module-parameters

Note that in the example below we see a [Spanglish](https://en.wikipedia.org/wiki/Spanglish) sentence.

```
$ cat module 
params.foo = 'Hello'
params.bar = 'world!'

def sayHello() {
    println "$params.foo $params.bar"
}

$ cat wf.nf 
nextflow.enable.dsl=2

params.foo = 'Hola'

include {sayHello} from './module'

params.bar = 'Mundo'

workflow {
    sayHello()
}

$ nextflow wf.nf 
Launching `wf.nf` [naughty_lavoisier] - revision: fae2d88139
Hola world!
```

fabianegli:
Closes #1984


jfy133 (approved):
In principle reads fine to me, but should be OK'd by a DSL2 expert. e.g. @drpatelh

fabianegli:
If this behavior is not not expected, or expected to change in the future, it probably should also get a corresponding unit test.

drpatelh:
I didn't even know this was a thing @fabianegli 😅 As a rule of thumb I always set all of the `params` declarations at the top of the module files which is probably why I haven't noticed e.g. https://github.com/nf-core/rnaseq/blob/11aa4bd1e146ede5fa8b334afbd75f389d159ff9/subworkflows/nf-core/fastqc_umitools_trimgalore.nf#L5-L7

fabianegli:
Setting parameters at the beginning is probably a good thing anyway because it puts changing parts at the top. I just thought it would be kind to the user to point out how it works and in this case when the inheritance happens. This might prevent someone less used to programming from having to track down a weird bug.

I am still new to nextflow, so I can't point to any good example of how a parameter would be set dynamically. In my mind it might be possible to set parameters based on the output of some previously included workflow.

If you think it is nonsensical and useless to set parameters past their import, maybe the linting should complain if it happens.

fabianegli:
You can also interpret it as a feature because it allows to import different workflows with the same paremeter name and different value.

fabianegli:
> I didn't even know this was a thing @fabianegli 😅 As a rule of thumb I always set all of the `params` declarations at the top of the module files which is probably why I haven't noticed e.g. https://github.com/nf-core/rnaseq/blob/11aa4bd1e146ede5fa8b334afbd75f389d159ff9/subworkflows/nf-core/fastqc_umitools_trimgalore.nf#L5-L7

Do you propose to include that as a recommendation instead of the changes in this PR?

drpatelh:
To be honest, I don't know if this is expected behaviour or not but to me it seems logical that `params` defined before the `include` statement would be available as opposed to those defined afterwards. So I think these changes are good to describe that behaviour in a little more detail. No harm 👍🏽

fabianegli:
@drpatelh Feel free to approve and/or merge this PR. I am blocked from merging it.

Anything beyond documenting this feature should go to dedicated issues/PRs anyway.

docs/dsl2.rst
```diff
@@ -383,20 +383,21 @@ A module script can define one or more parameters using the same syntax of a Nex
...
+Parameters set before the include statement are inherited from the including context but parameters set later are not. For example::
 
     params.foo = 'Hola'
-    params.bar = 'Mundo'
 
     include {sayHello} from './some/module'
 
+    params.bar = 'Mundo'
+
     workflow {
```
pditommaso, line 394: Even if possible, I think we should not promote this practice. I'd suggest adding a warning mentioning that params set after the import are not inherited.

fabianegli: This is the whole point of this PR - alert users that they need to specify parameters before the import if they want them to have any impact. The marked line has no effect on the workflow and this is clearly stated in the sentence above the code example. Check the diff.

fabianegli: @pditommaso I am not quite sure I understand what you're saying.

My point for adding these lines was "Don't just tell, show." In this case, show that setting the parameter after the import is not "possible" and hence it's also not a "practice". My discussions here in the PR notes, however did go further and might have caused some confusion.

Do you propose to delete the marked lines? If the code example should not be expanded to show the import inheritance explicitly I can remove it.

pditommaso: Apology the late reply. I've added a note following your suggestion 17bbc757. Merging manually. Thanks for contributing.

fabianegli: No worries, we live in a busy world. Happy to help.

Issue #1819, opened 2020-12-02
Failure resolving conda envs in parallel


grst:
### Expected behavior and actual behavior

When nextflow tries to create conda environments in parallel, that share a package, environment creation fails with 
the following error. 
```
InvalidArchiveError("Error with archive /opt/conda/pkgs/openjdk-10.0.2-h14c3975_1015.tar.bz2.  You probably need to delete and re-download or re-create this file.  Message from libarchive was:\n\nFailed to create dir 'legal'")
```

Expected: Create environments one after each other to avoid the problem. 

### Steps to reproduce the problem
Minimal nextflow example (`test.nf`)

```nextflow
#!/usr/bin/env nextflow

nextflow.enable.dsl = 2

process test1 {
    conda "python=3.7 star rsem fastqc gatk"
    
    script:
    """
    exit 0
    """
}

process test2 {
    conda "python=3.8 star rsem fastqc gatk"
    
    script:
    """
    exit 0
    """
}

workflow { 
    test1()
    test2()
}

```

**Note:** If the packages are already cached by conda, there will be no error. To test it in a container: 

```bash
podman pull continuumio/miniconda3
podman run -it -v .:/test continuumio/miniconda3 /bin/bash
> conda install -y -c conda-forge -c bioconda nextflow
> nextflow run /test/test.nf
```

### Program output 

```console
(base) root@dc3f51d38a99:/# ./test/test.nf 
N E X T F L O W  ~  version 20.10.0
Launching `./test/test.nf` [agitated_cori] - revision: 647d8085a5
[-        ] process > test1 -
[-        ] process > test1 -
[-        ] process > test2 -
Error executing process > 'test1'

Caused by:
  Failed to create Conda environment
  command: conda create --mkdir --yes --quiet --prefix /work/conda/env-54b51c21323e895919a03b9a835682b4 python=3.7 star rsem fastqc gatk
  status : 1
  message:
    InvalidArchiveError("Error with archive /opt/conda/pkgs/gettext-0.19.8.1-h0b5b191_1005.tar.bz2.  You probably need to delete and re-download or re-create this file.  Message from libarchive was:\n\nFailed to create dir 'share'")
```

### Environment 

* Nextflow version: 20.10.0
* Java version: openjdk version "11.0.8-internal" 2020-07-14
* Operating system: Linux
* Bash version: 5.0.18

### Additional context

This problem became more relevant with DSL2 pipelines, where every process has its own environment. 
[...]

PR #2030, opened 2021-04-11
Add support for using mamba with conda

abhi18av:
Initiating the work for `mamba` support in conda.

Planned features are 

~A `mamba` directive with `micromamba` option to users to use `micromamba` which doesn't rely on a base environment - see [here](https://github.com/mamba-org/mamba/blob/master/docs/source/micromamba.md#Installation)~

We have decided to implement this as a config setting for `conda` scope.

```nextflow
conda {
   useMamba = true
}
```

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
@@ -305,20 +305,22 @@ class CondaCache {
      */
     @PackageScope
     DataflowVariable<Path> getLazyImagePath(String condaEnv) {
+
+        final binaryName = useMamba ? "mamba" : "conda"
+
         if( condaEnv in condaPrefixPaths ) {
-            log.trace "Conda found local environment `$condaEnv`"
+            log.trace "${binaryName.capitalize()} found local environment `$condaEnv`"
             return condaPrefixPaths[condaEnv]
         }
 
         synchronized (condaPrefixPaths) {
             def result = condaPrefixPaths[condaEnv]
             if( result == null ) {
-                final binaryName = useMamba ? "mamba" : "conda"
                 result = new LazyDataflowVariable<Path>({ createLocalCondaEnv(condaEnv, binaryName) })
                 condaPrefixPaths[condaEnv] = result
             }
             else {
-                log.trace "Conda found local cache for environment `$condaEnv` (2)"
+                log.trace "${binaryName.capitalize()} found local cache for environment `$condaEnv` (2)"
             }
             return result
         }
--- a/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
@@ -72,6 +72,10 @@ class CondaCache {
 
     @PackageScope Path getConfigCacheDir0() { configCacheDir0 }
 
+    @PackageScope String binaryName() {
+        useMamba ? "mamba" : "conda"
+    }
+
     /** Only for debugging purpose - do not use */
     @PackageScope
     CondaCache() {}
@@ -215,10 +219,10 @@ class CondaCache {
      * @return the conda environment prefix {@link Path}
      */
     @PackageScope
-    Path createLocalCondaEnv(String condaEnv, String binaryName = "conda") {
+    Path createLocalCondaEnv(String condaEnv) {
         final prefixPath = condaPrefixPath(condaEnv)
         if( prefixPath.isDirectory() ) {
-            log.debug "The binary '${binaryName}' found local env for environment=$condaEnv; path=$prefixPath"
+            log.debug "The binary '${binaryName()}' found local env for environment=$condaEnv; path=$prefixPath"
             return prefixPath
         }
 
@@ -228,7 +232,7 @@ class CondaCache {
 
         final mutex = new FileMutex(target: file, timeout: createTimeout, waitMessage: wait, errorMessage: err)
         try {
-            mutex .lock { createLocalCondaEnv0(condaEnv, prefixPath, binaryName) }
+            mutex .lock { createLocalCondaEnv0(condaEnv, prefixPath) }
         }
         finally {
             file.delete()
@@ -243,27 +247,27 @@ class CondaCache {
     }
 
     @PackageScope
-    Path createLocalCondaEnv0(String condaEnv, Path prefixPath, String binaryName = "conda") {
+    Path createLocalCondaEnv0(String condaEnv, Path prefixPath) {
 
-        log.info "Creating env using ${binaryName}: $condaEnv [cache $prefixPath]"
+        log.info "Creating env using ${binaryName()}: $condaEnv [cache $prefixPath]"
 
         final opts = createOptions ? "$createOptions " : ''
         def cmd
         if( isYamlFilePath(condaEnv) ) {
-            cmd = "${binaryName} env create --prefix ${Escape.path(prefixPath)} --file ${Escape.path(makeAbsolute(condaEnv))}"
+            cmd = "${binaryName()} env create --prefix ${Escape.path(prefixPath)} --file ${Escape.path(makeAbsolute(condaEnv))}"
         }
         else if( isTextFilePath(condaEnv) ) {
 
-            cmd = "${binaryName} create $opts--mkdir --yes --quiet --prefix ${Escape.path(prefixPath)} --file ${Escape.path(makeAbsolute(condaEnv))}"
+            cmd = "${binaryName()} create $opts--mkdir --yes --quiet --prefix ${Escape.path(prefixPath)} --file ${Escape.path(makeAbsolute(condaEnv))}"
         }
 
         else {
-            cmd = "${binaryName} create $opts--mkdir --yes --quiet --prefix ${Escape.path(prefixPath)} $condaEnv"
+            cmd = "${binaryName()} create $opts--mkdir --yes --quiet --prefix ${Escape.path(prefixPath)} $condaEnv"
         }
 
         try {
             runCommand( cmd )
-            log.debug "${binaryName.capitalize()} create complete env=$condaEnv path=$prefixPath"
+            log.debug "'${binaryName()}' create complete env=$condaEnv path=$prefixPath"
         }
         catch( Exception e ){
             // clean-up to avoid to keep eventually corrupted image file
@@ -310,21 +314,19 @@ class CondaCache {
     @PackageScope
     DataflowVariable<Path> getLazyImagePath(String condaEnv) {
 
-        final binaryName = useMamba ? "mamba" : "conda"
-
         if( condaEnv in condaPrefixPaths ) {
-            log.trace "${binaryName.capitalize()} found local environment `$condaEnv`"
+            log.trace "The binary '${binaryName()}' found local environment `$condaEnv`"
             return condaPrefixPaths[condaEnv]
         }
 
         synchronized (condaPrefixPaths) {
             def result = condaPrefixPaths[condaEnv]
             if( result == null ) {
-                result = new LazyDataflowVariable<Path>({ createLocalCondaEnv(condaEnv, binaryName) })
+                result = new LazyDataflowVariable<Path>({ createLocalCondaEnv(condaEnv) })
                 condaPrefixPaths[condaEnv] = result
             }
             else {
-                log.trace "${binaryName.capitalize()} found local cache for environment `$condaEnv` (2)"
+                log.trace "The binary '${binaryName()}' found local cache for environment `$condaEnv` (2)"
             }
             return result
         }
--- a/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
@@ -72,8 +72,6 @@ class CondaCache {
 
     @PackageScope Path getConfigCacheDir0() { configCacheDir0 }
 
-    Boolean isUsingMamba() { useMamba }
-
     /** Only for debugging purpose - do not use */
     @PackageScope
     CondaCache() {}
@@ -315,7 +313,7 @@ class CondaCache {
         synchronized (condaPrefixPaths) {
             def result = condaPrefixPaths[condaEnv]
             if( result == null ) {
-                final binaryName = isUsingMamba() ? "mamba" : "conda"
+                final binaryName = useMamba ? "mamba" : "conda"
                 result = new LazyDataflowVariable<Path>({ createLocalCondaEnv(condaEnv, binaryName) })
                 condaPrefixPaths[condaEnv] = result
             }
--- a/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
@@ -60,7 +60,7 @@ class CondaCache {
 
     private String createOptions
 
-    private Boolean useMamba = false
+    private boolean useMamba 
 
     private Path configCacheDir0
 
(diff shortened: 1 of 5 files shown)


pditommaso:
Is mamba cli compatible with conda cli? In that case would not make sense just to use it in place of `conda` in the current implementation (with an option to use maba instead of conda?) 

https://github.com/nextflow-io/nextflow/blob/914d7cb40bf58ff977717cd137352e7dfe97dbdb/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy#L246-L254

abhi18av:
Yes Paolo, `mamba` CLI aims to be the superset of `conda` CLI. 

I've put forward a couple of approaches to adding `mamba` support earlier in this comment here https://github.com/nextflow-io/nextflow/issues/1819#issuecomment-783498840 


Please let me know which approach makes more sense to be used for the integration :)

phue:
It seems that `mamba` is CLI-compatible to mamba, but `micromamba` is not. This probably makes the proposed `condaBin` implementation a bit tricky.

Also note that `mamba env create` currently falls back to `conda` to create the environment (See here: https://github.com/mamba-org/mamba/issues/633)
This means that it would speed up the solving part, but would not fix https://github.com/nextflow-io/nextflow/issues/1819 because downloading and caching is handled by `conda`.

However, `micromamba` could be used for creating environments from yaml files using something like this (same syntax as for textfiles. Note that there is no `mamba env` command):
```suggestion
        if( isYamlFilePath(mambaEnv) ) {
            cmd = "micromamba create -p ${Escape.path(prefixPath)} -f ${Escape.path(makeAbsolute(mambaEnv))}"
        }
```

abhi18av:
> It seems that mamba is CLI-compatible to conda, but micromamba is not. This probably makes the proposed condaBin implementation a bit tricky.

Yes, the direction we have decided to take is to implement it as a `directive` which allows enough space to support sibling projects of mamba such as `quetz` and `boa` in future.

Without the `micromamba: true` option, `mamba` would fall back to the conda binary by default.

```nextflow
process foo {
  mamba 'bwa samtools multiqc', micromamba: true

  '''
  your_command --here
  '''
}


```

Thoughts?

phue:
I think that is a great solution :+1:

pditommaso:
I'm not convived by this approach that's introducing a new mamba specific feature and al related options and vars eg `-with-mamba`, `NXF_MAMBA_CACHEDIR`, `mamba` directive etc. 

Since mamba is a drop-in replacement for Conda as a user I would expect to be able to use the usual Conda option and just use Mamba binary in place of conda behind the scene. 

I think we need to introce  a `useMamba` option in the `conda` context  and when set replace conda command lines with the corresponding mamba ones.

drpatelh:
> I think we need to introce a useMamba option in the conda context and when set replace conda command lines with the corresponding mamba ones.

Yep, I agree. If possible, recycling the current Conda parameters and env variables would be much cleaner with the hope that maybe when this stabilises a little we can maybe just set `useMamba: true` by default?

It also means we don't need to add more syntactic clutter to module files [here](https://github.com/nf-core/modules/blob/f7e3b8260ceeb6739093aec6f12d86b99cf7f799/software/fastqc/main.nf#L14) specifically for `mamba` because `useMamba` can be set via a config if required.

pditommaso:
Yes, agree all apart from having `useMamba: true` by default. I think it should be an opt-in feature.

drpatelh:
> I think it should be an opt-in feature.

Given how slow Conda is I think you would have to have way too much time on your hands not to opt in to using mamba by default 😅

abhi18av:
> I think we need to introce a useMamba option in the conda context and when set replace conda command lines with the corresponding mamba ones. 

Cool, I'll refactor accordingly :)

abhi18av:
Okay so I've refactored the PR code, to rely upon the `useMamba` option for the `conda` directive.

One place I need some guidance is how should I accomodate this option within the `BashWrapperBuilder`, I tried to find a way to derive it via `TaskBean / TaskProcessor / TaskRun` classes but it felt like retrofitting.

https://github.com/nextflow-io/nextflow/blob/73899f3b7967f3bd9bc393749a80ce893368a28e/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy#L345

Thoughts?

modules/nextflow/src/main/groovy/nextflow/conda/CondaOptions.groovy
```diff
@@ -0,0 +1,59 @@
...
+ * Model Conda options.
+ *
+ * @author Abhinav Sharma <abhi18av@outlook.com>
+ */
+@CompileStatic
+@ToString(includeNames = true)
+@EqualsAndHashCode(includeFields = true)
+class CondaOptions {
+
+    private Boolean useMamba
```
abhi18av, line 35: Perhaps, I should add the `Path path` option here as well?

pditommaso: Instead of specifying a process level better to add this setting in the conga config (ie nextflow config file). see here for example 

https://github.com/nextflow-io/nextflow/blob/914d7cb40bf58ff977717cd137352e7dfe97dbdb/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy#L82-L93

abhi18av: Okay, so the user would only need to update the `nextflow.config` file like so 

```nextflow
conda {
   useMamba = true
}
```

Hmm, this is interesting and cleaner as it localizes the overall configs.

Cool, I'm implementing this one now.

abhi18av:
**Update:** 

Able to create the environments using `mamba`. 

Please let me know your thoughts on the current solution.

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
```diff
@@ -209,20 +215,20 @@ class CondaCache {
      * @return the conda environment prefix {@link Path}
      */
     @PackageScope
-    Path createLocalCondaEnv(String condaEnv) {
+    Path createLocalCondaEnv(String condaEnv, String binaryName = "conda") {
         final prefixPath = condaPrefixPath(condaEnv)
         if( prefixPath.isDirectory() ) {
-            log.debug "Conda found local env for environment=$condaEnv; path=$prefixPath"
+            log.debug "${binaryName.capitalize()} found local env for environment=$condaEnv; path=$prefixPath"
```
pditommaso, line 221: Since it's a binary name don't think should be capitalised, even better to between quotes. 

```
The binary '$name' was found local env for environment=$condaEnv; path=$prefixPath"
```

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
```diff
@@ -209,20 +215,20 @@ class CondaCache {
      * @return the conda environment prefix {@link Path}
      */
     @PackageScope
-    Path createLocalCondaEnv(String condaEnv) {
+    Path createLocalCondaEnv(String condaEnv, String binaryName = "conda") {
```
pditommaso, line 218: Can we keep `binaryName` as a class attribute instead of passing as argument? or both `conda` and `mamba`can be used in the same class instance?

pditommaso (changes requested):

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
abhi18av: > Can we keep binaryName as a class attribute instead of passing as argument?

Hmm, do you think that it's better to do use the following at class level

```groovy
        private useMamba = false
        private binaryName = useMamba ? "mamba" : "conda"
```

and then simply use the `binaryName` within the methods instead of argument?


> or both conda and mamba can be used in the same class instance?

I am not sure how to interpret that, but from what I understand, it's enough to rely on mamba (i.e. a single binary) CLI since it'll automatically locate `conda` if needed, without accomodating this in the Groovy code.

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
abhi18av: Makes sense, I've updated the message here https://github.com/nextflow-io/nextflow/pull/2030/commits/976f15cd931a51a8fc61bdd8334e5198cde0d632

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
pditommaso: > Hmm, do you think that it's better to do use the following at class level

Let's have a `binaryName` default to `conda` and switch to `mamba` when the flag is enabled

abhi18av: Okay, I have accomodated the change request :)

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
```diff
@@ -70,6 +72,10 @@ class CondaCache {
 
     @PackageScope Path getConfigCacheDir0() { configCacheDir0 }
 
+    @PackageScope String binaryName() {
+        useMamba ? "mamba" : "conda"
+    }
+
```
pditommaso, line 78: Renaming this to `getBinaryName` it becomes a *synthetic* attribute that you can access as `binaryName` instead of `binaryName()`

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
```diff
@@ -60,6 +60,8 @@ class CondaCache {
 
     private String createOptions
 
+    private Boolean useMamba = false
```
pditommaso, line 63: Minor, null is useless in this case better `private boolean useMamba` (false is default)

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
```diff
@@ -90,6 +96,10 @@ class CondaCache {
 
         if( config.cacheDir )
             configCacheDir0 = (config.cacheDir as Path).toAbsolutePath()
+
+        if( config.useMamba )
+            useMamba = config.useMamba as Boolean
```
pditommaso, line 101: `boolean` instead of `Boolean` to match the type declaration

pditommaso (commented):
Ok, we are almost there. Have tried it? also it should be added a note in the docs for conda, explaining this functionality, the benefits, how to use it and that's an experimental feature. 

Please also sign the commit to make the DCO both green. See the contribution readme for details.

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
abhi18av: Copy that 👍

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
abhi18av: Hmm, good to know!

I've refactored accordingly.

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
abhi18av: Interesting - thanks for the tip 😊

abhi18av:
Okay, so I have updated the 

- docs ✅ 
- tested locally with `nextflow-io/rnaseq-nf` ✅ 


**Summary:** 

The `user time` 
- without `mamba` is `2m30.046s` 
- with `mamba` is `1m36.336s`

**Details:**

On my humble local machine, the results without `useMamba` is 

```
(base) Abhinavs-MacBook-Pro:rnaseq-nf eklavya$ time ../../launch.sh main.nf -profile conda
N E X T F L O W  ~  version 21.05.0-edge
Launching `main.nf` [adoring_brahmagupta] - revision: 4ab4121ede
 R N A S E Q - N F   P I P E L I N E
 ===================================
 transcriptome: /Users/eklavya/projects/code/nextflow/_resources/rnaseq-nf/data/ggal/ggal_1_48850000_49020000.Ggal71.500bpflank.fa
 reads        : /Users/eklavya/projects/code/nextflow/_resources/rnaseq-nf/data/ggal/*_{1,2}.fq
 outdir       : results

[-        ] process > RNASEQ:INDEX  -
executor >  local (6)
[c1/99de7b] process > RNASEQ:INDEX (ggal_1_48850000_49020000) [100%] 1 of 1 ✔
[7c/381b31] process > RNASEQ:FASTQC (FASTQC on ggal_liver)    [100%] 2 of 2 ✔
[3b/b06222] process > RNASEQ:QUANT (ggal_gut)                 [100%] 2 of 2 ✔
[94/569293] process > MULTIQC                                 [100%] 1 of 1 ✔

Done! Open the following report in your browser --> results/multiqc_report.html

Completed at: 14-May-2021 12:10:10
Duration    : 2m 52s
CPU hours   : (a few seconds)
Succeeded   : 6



real	3m6.182s
user	2m30.046s
sys	0m22.124s

```

The result with `mamba` is shown below

```

(base) Abhinavs-MacBook-Pro:rnaseq-nf eklavya$ time ../../launch.sh main.nf -profile mamba
N E X T F L O W  ~  version 21.05.0-edge
Launching `main.nf` [compassionate_colden] - revision: 4ab4121ede
 R N A S E Q - N F   P I P E L I N E
 ===================================
 transcriptome: /Users/eklavya/projects/code/nextflow/_resources/rnaseq-nf/data/ggal/ggal_1_48850000_49020000.Ggal71.500bpflank.fa
 reads        : /Users/eklavya/projects/code/nextflow/_resources/rnaseq-nf/data/ggal/*_{1,2}.fq
 outdir       : results

[-        ] process > RNASEQ:INDEX  -
executor >  local (6)
[9b/a2900c] process > RNASEQ:INDEX (ggal_1_48850000_49020000) [100%] 1 of 1 ✔
[1f/b4e9cc] process > RNASEQ:FASTQC (FASTQC on ggal_gut)      [100%] 2 of 2 ✔
[59/d564f1] process > RNASEQ:QUANT (ggal_gut)                 [100%] 2 of 2 ✔
[0e/3bd045] process > MULTIQC                                 [100%] 1 of 1 ✔

Done! Open the following report in your browser --> results/multiqc_report.html

Completed at: 14-May-2021 12:14:11
Duration    : 2m 51s
CPU hours   : (a few seconds)
Succeeded   : 6



real	3m5.407s
user	1m36.336s
sys	0m24.346s

``

File: modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy [hunk 7]
```diff
@@ -212,12 +222,12 @@ class CondaCache {
     Path createLocalCondaEnv(String condaEnv) {
         final prefixPath = condaPrefixPath(condaEnv)
         if( prefixPath.isDirectory() ) {
-            log.debug "Conda found local env for environment=$condaEnv; path=$prefixPath"
+            log.debug "The binary '${binaryName}' found local env for environment=$condaEnv; path=$prefixPath"
             return prefixPath
```
pditommaso, line 226: My fault `The binary ` is useless, let's remote it, sorry.

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
```diff
@@ -301,8 +313,9 @@ class CondaCache {
      */
     @PackageScope
     DataflowVariable<Path> getLazyImagePath(String condaEnv) {
+
         if( condaEnv in condaPrefixPaths ) {
-            log.trace "Conda found local environment `$condaEnv`"
+            log.trace "The binary '${binaryName}' found local environment `$condaEnv`"
```
pditommaso, line 318: Also here and following

abhi18av: Done ✅

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
abhi18av: No worries, I've made the changes ✅

pditommaso (approved):
Awesome! nice contribution @abhi18av!

PR #2031, opened 2021-04-12
Generate XUnit XML reports

jordeu:
This pull requests adds the generation of a XUnit XML test report file at `test-results` folder when a testflow script is run.

It also creates at the folder `test-reports` an HTML report with all the XUnit reports that are at `test-results`. It is using an external tool to generate this report, but it will be better to generate our own html so we can customize it better to give more information useful from a Nextflow point of view (that is not exactly the traditional use case for JUnit like tests). Also, it will be nicer, to create a subcommand (ex: `nextflow test`) that runs all the tests (for example all the scripts with `test_` prefix) and at the end generate the full HTML report with all the tests results.


pditommaso:
> It is using an external tool to generate this report, but it will be better to generate our own html so we can customize it better to give more information useful from a Nextflow point of view

would this spare using the external lib? 

> it will be nicer, to create a subcommand

I agree let's do

jordeu:
> would this spare using the external lib?
Yes, we will be able to remove this dependency

pditommaso (approved):
Good!

pditommaso:
But tests fail: 

```
CmdTestTest > should create reports FAILED
    nextflow.exception.AbortOperationException at CmdTestTest.groovy:68
```

Use `make smoke` or `make test` before pushing

jordeu:
> But tests fail:
> 
> ```
> CmdTestTest > should create reports FAILED
>     nextflow.exception.AbortOperationException at CmdTestTest.groovy:68
> ```
> 
> Use `make smoke` or `make test` before pushing
It worked on my side. 

But I think that there is some error due to concurrency because sometimes the same testflow fails and sometimes it passes. The error is that it seems that the `foo.out[0]` sometimes have the value of another emission, but if I check the `.command.out` file at the run dir it's correct.

pditommaso:
Bump

mmatthews06:
@pditommaso or @jordeu, any guesses why the inconsistent test case return? Is this, and the `testflow` branch still something valuable to contribute to?

pditommaso:
Hi @mmatthews06, not sure what do you mean with "any guesses why the inconsistent test case return" ?

pditommaso:
Closing this because the project nf-test project provides a better alternative https://twitter.com/lukfor/status/1504852647442100227

Issue #911, opened 2018-10-30
PBSPro clusterOptions / additional resources directive


stekaz:
We have a cluster running PBS Pro where we are able to specify custom resources. Some examples include:

```
$ echo "sleep 30" | qsub -N TEST0 -l select=1:ncpus=1:mem=1mb:host=hpcnode001 -l walltime=1:00:00
$ echo "sleep 30" | qsub -N TEST1 -l select=1:ncpus=1:mem=1mb:chip=Intel -l walltime=1:00:00
$ echo "sleep 30" | qsub -N TEST2 -l select=1:ncpus=1:mem=1mb:ngpus=256 -l walltime=1:00:00
```

However, the `select` statement conflicts with additional resources (other than the `walltime` resource/directive, for some reason). This prevents us from using a `clusterOptions = '-l host=hpcnode001'` directive. For example:

```
$ echo "sleep 30" | qsub -N TEST3 -l select=1:ncpus=1:mem=1mb -l walltime=1:00:00 -l host=hpcnode001
qsub: "-lresource=" cannot be used with "select" or "place", resource is: host
```

Are there any provisions for use cases like this? It would be really great if there was some option that could append a list of additional resources to the `select` statement:

https://github.com/nextflow-io/nextflow/blob/master/src/main/groovy/nextflow/executor/PbsProExecutor.groovy#L64

stekaz:
Thanks Paolo, it looks like that will work only with the full resource request:

```
$ echo "sleep 30" | qsub -N TEST4 -l select=1:ncpus=1:mem=1mb -l walltime=1:00:00 -l select=1:host=hpcnode001
qsub: Job has no memory requested!!! Please submit a job requesting memory...
$ echo "sleep 30" | qsub -N TEST5 -l select=1:ncpus=1:mem=1mb -l walltime=1:00:00 -l select=1:host=hpcnode001:ncpus=2:mem=2mb
26374620.hpcpbs
$ qstat -1n | grep "TEST5"
26374620.hpcpbs stekaz test    TEST5       48844   1   2    2mb 01:00 R 00:00 hpcnode001/0*2
```

Interestingly, this also appears to work as expected:

```
$ echo "sleep 30" | qsub -N TEST6 -l ncpus=2 -l mem=1mb -l walltime=01:00:00 -l host=hpcnode002
26381752.hpcpbs
$ qstat -1n | grep "TEST"
[...]

PR #2036, opened 2021-04-13
Append clusterOptions at the beginning of PBS directives

IdoBar:
when multiple competing directives are provided, the first one will take effect, so task.config.clusterOptions should be appended to the beginning of the PBS directives string, not the end of it, see #911

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/PbsProExecutor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/PbsProExecutor.groovy
@@ -44,6 +44,11 @@ class PbsProExecutor extends PbsExecutor {
     @Override
     protected List<String> getDirectives(TaskRun task, List<String> result ) {
         assert result !=null
+        
+        // -- at the beginning append the custom cluster options
+        if( task.config.clusterOptions ) {
+            result << task.config.clusterOptions.toString() << ''
+        }
 
         result << '-N' << getJobNameFor(task)
         result << '-o' << quote(task.workDir.resolve(TaskRun.CMD_LOG))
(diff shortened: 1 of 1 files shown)


pditommaso:
Is there docs from PBS vendor regarding this behaviour?

IdoBar:
According to the [PBSPro 19.2 User Guide](https://www.altair.com/pdfs/pbsworks/PBSUserGuide19.2.3.pdf) (P.69)

> **4.3.8.8  Only One select Statement Per Job**
You can include at most one select statement per job submission.

It doesn't say what happens when you include more than one statement (though I've shown this in #911 and will double-check this on a different HPC cluster running PBS Pro).

IdoBar:
Yes, can confirm this to be the case on two independent HPC clusters running PBS Pro v19.2.3 (20190726105414) and v18.1.2
```
cat .command.test
#!/bin/bash
#PBS -N nf-test
#PBS -l walltime=00:30:00
#PBS -l select=1:ncpus=1:mem=6144mb
#PBS -l select=1:ncpus=2:mem=10kb

sleep 30

$ cat .command.test2
#!/bin/bash
#PBS -N nf-test2
#PBS -l walltime=00:30:00
#PBS -l select=1:ncpus=2:mem=10kb
#PBS -l select=1:ncpus=1:mem=6144mb

sleep 30

$ qsub .command.test
$ qsub .command.test2
$ qstat -u $USER
gc-prd-hpcadm:
                                                            Req'd  Req'd   Elap
Job ID          Username Queue    Jobname    SessID NDS TSK Memory Time  S Time
--------------- -------- -------- ---------- ------ --- --- ------ ----- - -----
100420.gc-prd-h s2978925 workq    nf-test     70894   1   1  6144m 00:30 R 00:00
100421.gc-prd-h s2978925 workq    nf-test2    70983   1   2   10kb 00:30 R 00:00
```

pditommaso (changes requested):
Ok, thank you. In this case please let's add a more explicit comment explaining the reason why the `clusterOption` has to be the first directive . Also a test check-in that would be useful

IdoBar:
Where would you like this comment to be? In this PR on in the actual code?
I will also add a suitable test.

pditommaso:
In the code, thanks!

pditommaso:
Merged manually 75e3f225. Thanks for contributing.

pditommaso:
Included in release `21.06.0-edge`.

PR #2037, opened 2021-04-13
Added support for proxy servers that ask for authentication

stekaz:
Hi, this PR adds support for corporate proxy servers that require authentication using a username and password, which, in my case, requires exporting the following variables in order to wget files over HTTP/HTTPS/FTP for example:

```
export http_proxy='http://username:password@webproxy.domain.ad.lan:8080'
export https_proxy='http://username:password@webproxy.domain.ad.lan:8080'
export ftp_proxy='ftp://username:password@webproxy.domain.ad.lan:8080'
export HTTP_PROXY='http://username:password@webproxy.domain.ad.lan:8080'
export HTTPS_PROXY='http://username:password@webproxy.domain.ad.lan:8080'
export FTP_PROXY='ftp://username:password@webproxy.domain.ad.lan:8080'
```

With the above proxy settings, currently I get:

```
N E X T F L O W  ~  version 21.04.0-edge
Launching `/home/stekaz/devel/test/main.nf` [festering_einstein] - revision: ac766712be
WARN: Unable to stage foreign file: https://hgdownload.cse.ucsc.edu/goldenpath/hg38/chromosomes/chrM.fa.gz (try 1) -- Cause: Unable to access path: /goldenpath/hg38/chromosomes/chrM.fa.gz
WARN: Unable to stage foreign file: https://hgdownload.cse.ucsc.edu/goldenpath/hg38/chromosomes/chrM.fa.gz (try 2) -- Cause: Unable to access path: /goldenpath/hg38/chromosomes/chrM.fa.gz
WARN: Unable to stage foreign file: https://hgdownload.cse.ucsc.edu/goldenpath/hg38/chromosomes/chrM.fa.gz (try 3) -- Cause: Unable to access path: /goldenpath/hg38/chromosomes/chrM.fa.gz
[null] NOTE: Can't stage file https://hgdownload.cse.ucsc.edu/goldenpath/hg38/chromosomes/chrM.fa.gz -- reason: Unable to access path: /goldenpath/hg38/chromosomes/chrM.fa.gz -- Execution is retried (1)
```

Note the weird error message: 'unable to access path' in the root directory. ~~Please see my first commit, which fixes that.~~

In my debugging, I found the return code I was getting was a 407 (proxy authentication required). This PR aims to fix that. It sets the authenticator that will be used when the proxy server asks for authentication. It works, for me at least. Please let me know if there are any issues.


modules/nextflow/src/main/groovy/nextflow/cli/Launcher.groovy
```diff
@@ -593,17 +597,30 @@ class Launcher {
...
+                if( proxy.port ) System.setProperty("${qualifier.toLowerCase()}.proxyPort", proxy.port)
+                if( username && password ) {
+                    log.debug "Setting $qualifier proxy authenticator ..."
+                    Authenticator authenticator = new Authenticator() {
+                        protected PasswordAuthentication getPasswordAuthentication() {
+                            return new PasswordAuthentication( username, password.toCharArray() )
+                        }
+                    }
+                    Authenticator.setDefault(authenticator);
+                }
```
pditommaso, line 623: What's the reason of this?

stekaz: It sets the authenticator (assuming a username and password is provided) that will be used to establish a network connection when the proxy server asks for authentication.

pditommaso: My concern is how is determined to which host apply at these credentials? should not only be for the proxy? 

Also, why the previous username and password handling was not good?

stekaz: My use case is for Nextflow to be able to stage foreign files over HTTP/FTP from behind our institute's proxy server. These credentials are required by our proxy server only. Just so we're on the same page, would you be able to link me the previous username and password handling code? Maybe I missed it. Thanks!

pditommaso: You may be [right](https://stackoverflow.com/a/1626616/395921), however, I fear setting the default authenticator can interfere with any HTTP connection. It May be needed to check the request type as shown in this comment https://stackoverflow.com/a/16340273/395921

pditommaso (changes requested):

stekaz:
Thanks very much for investigating this @pditommaso. I tend to agree, so I've added the request type check. I also added the requesting host and port checks. I think this makes sense, but am happy to be guided on this.

pditommaso:
I've merged this manually with some refactoring. 8ec33b04d. Thanks a lot for this contribution and apology for the late reply.

PR #2038, opened 2021-04-14
Duration `toString()` units

ewels:
We have had problems with Duration units ending up as strings that look like doubles - eg. `2d`.

Here I have changed the `Duration` `toString()` output to use the more verbose unit types, eg `2days` instead of `2d`.

I also updated the documentation to list all of the allowed units instead of just `s`,`m`,`h`,`d`.


ewels:
Hi @pditommaso - I believe this PR followed on from a conversation we had a while back. Are you still on board? Is there anything I need to do for this to progress or was it just a case of conflicting holidays? 😄

Phil

pditommaso:
> was it just a case of conflicting holidays?

a kind of, my last day 😄 

My concern with this PR is that this can break existing tools or scripts relying on this time units since it's extensive used in many components. Is there any chance to manage on your side? Otherwise, we should take in consideration as alternative non-default format

ewels:
> Is there any chance to manage on your side?

We did, but not in a way that I particularly like: https://github.com/nf-core/tools/pull/1012/

```nextflow
// Cast Duration to String
if (p['value'].getClass() == nextflow.util.Duration) {
    new_params.replace(p.key, p['value'].toString().replaceFirst(/d(?!\S)/, "day"))
}
```

I'm not sure how robust the fix is, but it seemed to resolve the most obvious errors for us.

I understand the concern about changing defaults though. If you'd rather not change the behaviour, I can update the PR so that it's just adding to the docs?

Phil

pditommaso:
The docs update is great. Regarding the above snippet, a bit more idiomatic way to write it is: 
```
if (p['value'] instanceof nextflow.util.Duration) {
    new_params.replace(p.key, p['value'].toString().replaceFirst(/d(?!\S)/, "day"))
}
```

But it's essentially the same hack.

ewels:
Reverted the changes to defaults, just the docs updates now 👍🏻

pditommaso (approved):
Awesome thanks!

PR #2044, opened 2021-04-15
Add support custom service account email in Google Cloud LifeSciences executor

yonghaoy:
This is supported by pipeline API, but not Nextflow yet. 
We have use case that we want to use other SA account other than the default compute engine default service account, would be great if we can support this in Nextflow as well.

 https://cloud.google.com/life-sciences/docs/reference/rest/v2beta/projects.locations.pipelines/run#ServiceAccount

Use case: We are building a [FedRamp](https://cloud.google.com/security/compliance/fedramp-guide#section-6) application(https://terra.bio/), and using default compute engine SA is not recommended because of security compliance. We deleted the default compute engine SA when create the project, and create a SA for each user which has very limited IAM role


docs/google.rst
```diff
@@ -130,6 +130,7 @@ google.lifeSciences.debug                      When ``true`` copies the `/google
 google.lifeSciences.preemptible                When ``true`` enables the usage of *preemptible* virtual machines or ``false`` otherwise (default: ``true``)
 google.lifeSciences.usePrivateAddress          When ``true`` the VM will NOT be provided with a public IP address, and only contain an internal IP. If this option is enabled, the associated job can only load docker images from Google Container Registry, and the job executable cannot use external services other than Google APIs (default: ``false``). Requires version ``20.03.0-edge`` or later.
 google.lifeSciences.network                    Set network name to attach the VM's network interface to. The value will be prefixed with global/networks/ unless it contains a /, in which case it is assumed to be a fully specified network resource URL. If unspecified, the global default network is used. Requires version ``21.03.0-edge`` or later.
+google.lifeSciences.serviceAccountEmail        Define the email of the service account to install on the VM. If not specified, the default Compute Engine service account for the project will be used.
```
yonghaoy, line 133: I saw few other fields have `Requires version ``21.03.0-edge`` or later.`
Not sure if I need similar things, and if yes, what would be the value?

plugins/nf-google/src/main/nextflow/cloud/google/lifesciences/GoogleLifeSciencesHelper.groovy
```diff
@@ -176,6 +176,9 @@ class GoogleLifeSciencesHelper {
         disk.setSizeGb(req.diskSizeGb)
 
         def serviceAccount = new ServiceAccount().setScopes( [SCOPE_CLOUD_PLATFORM] )
```
yonghaoy, line 178: not sure if we want/need to specify customized scope. But let me know if we should. 
If yes, what would it be? List of String? 

Thanks

yonghaoy:
Hello @pditommaso not sure what is correct channel to propose PR like this and communicate with you. Please let me know if anything else I should do.

plugins/nf-google/src/main/nextflow/cloud/google/lifesciences/GoogleLifeSciencesHelper.groovy
pditommaso: What's the use of the case of custom scope?

pditommaso (commented):
Thanks for this PR and my apology for the late reply. Could you please describe the need/use case for having to specify the service account email? may be worth mentioning it in the documentation.

plugins/nf-google/src/main/nextflow/cloud/google/lifesciences/GoogleLifeSciencesHelper.groovy
yonghaoy: I can not think of any use case that require custom scope. Seems "cloud-platform API scope" already added by default.  One possible use case might be the batch job need to access other Google APIs besides GCP(e.g. gsuite, drive)? 
That does feel rate, so I did not add support for custom scope in this PR.

yonghaoy:
> Thanks for this PR and my apology for the late reply. Could you please describe the need/use case for having to specify the service account email? may be worth mentioning it in the documentation.

Thanks for getting back to me. I explained the use case in the PR description. Using default compute engine service account is generally not recommended in high security requirement system. I don't feel it's properity&necessary to put in the official document, but let me know if I should.

Thanks!

plugins/nf-google/src/main/nextflow/cloud/google/lifesciences/GoogleLifeSciencesHelper.groovy
kemp-verily: Yeah there isn't really any good reason to modify the scope.  Historically this was useful since there was a specific scope for the Genomics API (predecessor to Life Sciences), and cloud-platform doesn't grant *everything*, but over time most APIs are converging on it and using IAM to control access on a finer grained basis.

docs/google.rst
pditommaso: Ok, we'll merge this into `21.05.0-edge`. You may want to add it. Also, maybe the following reads better 

```
Define the Google service account email to use for the pipeline execution. If not specified, the default Compute Engine service account for the project will be used.
```

plugins/nf-google/src/main/nextflow/cloud/google/lifesciences/GoogleLifeSciencesHelper.groovy
pditommaso: Fair enough, therefore the scope setting can be ignored

docs/google.rst
yonghaoy: Done. Thanks @pditommaso for reviewing again.  I think this PR is ready to go?

pditommaso (approved):
Looks good. Thanks for contributing the project.

pditommaso:
Merging in `develop` branch until new 21.0.4.0 version is released.

PR #2056, opened 2021-04-22
List requirements on pre-existing named pools in Azure.

manuelesimi:
After running some tests on pre-existing named pools in Azure, I found out some requirements not listed in the current documentation but needed by Nextflow to use these pools as queues where to submit tasks.

### DIFF
--- a/docs/azure.rst
+++ b/docs/azure.rst
@@ -198,19 +198,6 @@ requirements:
 2 - the task slots per node must match with the number of cores for the selected VM. Nextflow would return an error like
 "Azure Batch pool 'ID' slots per node does not match the VM num cores (slots: N, cores: Y)".
 
-3 - the ``azcopy`` command must be installed on each node of the pool. A way to do it is to define a Start Task with:
-
-   * a resource file with the following settings:
-        - type of ``Http Url``
-        - value sets to ``https://github.com/nextflow-io/azcopy-tool/raw/linux_amd64_10.8.0/azcopy``
-        - path sets to ``azcopy``
-   * and the following command::
-
-        bash -c "chmod +x azcopy &&
-        mkdir $AZ_BATCH_NODE_SHARED_DIR/bin/ &&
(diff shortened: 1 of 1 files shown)


pditommaso (commented):
Hi Manuele, thanks for the detailed contribution. Regarding the need to adding the `azcopy` tool, NF should be able to download it automatically when the task start (when the pool was not created by NF itself) 

https://github.com/nextflow-io/nextflow/blob/711362ca895cafb3f7f01651050c4dfc0cba9557/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L359-L364

manuelesimi:
Hi @pditommaso, thanks for clarifying.

After I posted my comment I realized that I had the _copyToolInstallMode_ option set to `node` instead of `task`. That indeed does the job. I'd still list `azcopy` among the requirements for pre-created pools and explain to set  _copyToolInstallMode_ to `task` to match the requirement.

pditommaso:
> That indeed does the job. I'd still list azcopy among the requirements for pre-created pools and explain to set copyToolInstallMode to task to match the requirement 

This makes sense tho something still weird. The setting is set automatically so that  when the pool is not created by NF the azcopy is copied for each task 

https://github.com/nextflow-io/nextflow/blob/9e9337d3e847a7a657ac25ca39fcab8bac487938/plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzBatchOpts.groovy#L126-L138

manuelesimi:
Hi @pditommaso, you're right. Another error (with the docker container) led me to manually install the azcopy tool on the nodes (or set copyToolInstallMode to task). Now that I solved it, I see that NF does the installation automatically. 

Bottom line, the code you pasted indeed works :) 
Thanks!

pditommaso:
Good! 😎 then what about point 3 in your PR?

manuelesimi:
> Good! 😎 then what about point 3 in your PR?

Just removed ;)

pditommaso (approved):
Thanks Manuele.

PR #2065, opened 2021-04-27
Fix combination of header and skip parameters when using splitCsv operator

### LINKED ISSUE #2064: Unexpected behavior when combining headers from first line and skip on splitCsv function
I got an expected behavior when I combine headers from first line and skip on splitCsv function.

With this simple csv file
```
a,b,c
1,2,3
4,5,6
7,8,9
```
And this simple workflow
```
workflow {
    Channel.fromPath('csv.csv')
            .splitCsv(header: true, limit: 2)
            .view()

    Channel.fromPath('csv.csv')
        .splitCsv(header: true, skip: 2, limit: 2)
        .view()
}
```
I expect this result
```
[a:1, b:2, c:3]
[a:4, b:5, c:6]
[a:7, :b:8, c:9]
```
Where the two first lines was read by the first `splitCsv` and the last line by the second `splitCsv`.

But I get this result
```
[a:1, b:2, c:3]
[a:4, b:5, c:6]
[4:7, 5:8, 6:9]
```

The documentation is ambiguous on this two parameters
* header | When true the first line is used as columns names.
* skip | Number of lines since the file beginning to ignore when parsing the CSV content.

Whereas `header` indicates that columns names are retrieved from first line, the `skip` indicates the number of lines to ignore.
It is disturbing to see usage of `retrieved records` terms in `limit` documentation while `skip` use `lines`.
And proximity of `skip` and `limit` in documentation let's reader to think they share similar concepts.

alexandreBaronZnk:
Closes #2064

### DIFF
--- a/docs/operator.rst
+++ b/docs/operator.rst
@@ -878,7 +878,7 @@ quote       Values may be quoted by single or double quote characters.
 header      When ``true`` the first line is used as columns names. Alternatively it can be used to provide the list of columns names.
 charset     Parse the content by using the specified charset e.g. ``UTF-8``
 strip       Removes leading and trailing blanks from values (default: ``false``)
-skip        Number of lines since the file beginning to ignore when parsing the CSV content.
+skip        Number of records since the file beginning to ignore when parsing the CSV content.
 limit       Limits the number of retrieved records for each file to the specified value.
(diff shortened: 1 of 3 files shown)


pditommaso:
Thanks for the contribution, but as commented [here](https://github.com/nextflow-io/nextflow/issues/2064#issuecomment-835411217) changing the semantic of options is bad for backward compatibility

PR #2066, opened 2021-04-27
Allow the usage of objects at CLI params and params file

jordeu:
This implementation allows to use parameters that are objects (like `MemoryUnit` and `Duration`) at the CLI params and at the JSON/YAML params file.

To decide to what object has to convert the input string, it checks the type of the default parameter that it is set at the config file and it try to convert into it. If it is not possible the param will be overwriting using the default string representation.


pditommaso:
Can you make an example?

jordeu:
Given this `main.nf`:
```
#!/usr/bin/env nextflow

cheers = Channel.from 'Bonjour', 'Ciao', 'Hello', 'Hola'

process sayHello {
  echo true
  input: 
    val x from cheers
  script:
    """
    echo '$x world! CPUS:${task.cpus}'
    """
}
```
and this `nextflow.config`:
```
process.cpus = 1

params {
   max_time = 5.h
}

if ((params.max_time - 1.h) > 5.h) {
   process.cpus = 2
}

System.out.println("MAX_TIME: ${params.max_time} (${params.max_time.class})")
```

When you run it using production Nextflow:
```
$ nextflow run ./testhello --max_time=6.h
MAX_TIME: 5h (class nextflow.util.Duration)
Launching `./testhello/main.nf` [fervent_williams] - revision: 6968f1918b
MAX_TIME: 6.h (class java.lang.String)
executor >  local (4)
[15/d60dd7] process > sayHello (2) [100%] 4 of 4 ✔
Ciao world! CPUS:2

Hola world! CPUS:2

Hello world! CPUS:2

Bonjour world! CPUS:2
```

But with the new implementation:
```
$ ./launch.sh run ./testhello --max_time=6.h
N E X T F L O W  ~  version 21.04.1-edge
MAX_TIME: 5h (class nextflow.util.Duration)
Launching `./testhello/main.nf` [insane_mandelbrot] - revision: 6968f1918b
MAX_TIME: 6h (class nextflow.util.Duration)
executor >  local (4)
[6e/a9327a] process > sayHello (4) [100%] 4 of 4 ✔
Hello world! CPUS:1

Bonjour world! CPUS:1

Ciao world! CPUS:1

Hola world! CPUS:1
```

Notice that current Nextflow it evaluates wrong this part `params.max_time - 1.h`. It is treating both sides as strings and evaluates to `"6.h" - 1.h = "6.h"` and then it evaluates the `"6.h" > 5.h` both as a `Duration` objects. This can lead to unexpected results very difficult to spot the cause because in fact Groovy is not going to fail anywhere.

pditommaso:
Outdated

Issue #452, opened 2017-09-15
Automatically delete files marked as temp as soon as not needed anymore


andreas-wilm:
To reduce the footprint of larger workflows it's very useful if temporary files (which are marked as such) could be **automatically** deleted once they are not used anymore. Yes, this breaks reruns, but for easily recomputed files or large ones (footprint), this makes sense. Using scratch (see #230) is not always possible/wanted (e.g. for very large files and small scratch). It's also not always possible to delete those files as a user (except at the very end of the workflow), because multiple downstream processes running at different times might require them. This feature is for example [implemented in snakemake](http://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#protected-and-temporary-files), but maybe it's easily done there because the DAG is computed in advance?

Note, this is different from issue #165 where the goal was to remove non-declared files. The issue contains a useful discussion of the topic nevertheless.

Andreas

pditommaso:
I'm adding [this](https://groups.google.com/d/msg/nextflow/-ecC4RrZC_8/9rDts49gBQAJ) for reference. I agree that intermediate files handling needs to be improved but it will require some interval refactoring. Need to investigate. cc @joshua-d-campbell

pditommaso:
I've brainstormed a bit more about this issue and actually it should be possible to remove intermediate output files without compromising the resume feature. 

First problem, runtime generated DAG: tho execution graph is only generated at runtime, it's generally fully resolved immediately after the workflow execution starts. Therefore it would be enough to defer the output delete *after* the full resolution of the execution DAG. That's just after the [run](https://github.com/nextflow-io/nextflow/blob/c84869dcd5c34dc0fddda3e178cacd8d32214f9e/src/main/groovy/nextflow/script/ScriptRunner.groovy#L168) invocation and before the `terminate`. 

[...]

PR #2135, opened 2021-05-27
Document cleanup configuration variable

jfy133:
This documents the currently undocumented `cleanup` option that is avaliable in configuration files.

Based on the discussion on the nf-core slack originally [here](https://nfcore.slack.com/archives/CE56GDKN0/p1583244613045500) and more recently [here](https://nfcore.slack.com/archives/CE6P95170/p1622115069029300)

Note this is a draft as we are unsure if this still an incubating feature and also it is not a traditional 'scope'. However we were unsure where this would fit in the documentation - therefore feel free to move/update as necessary.

It would also potentially close [#452](https://github.com/nextflow-io/nextflow/issues/452)


maxulysse:
You say `There are additional variables`, but you only list one.

jfy133:
That is future proofing in case there are other mysterious things we don't know about yet :grimacing:

mmatthews06:
@pditommaso , is still an officially supported feature, or experimental? And if the latter, is there any plan to make it fully supported?

Obvously #452 would be nice, too.

pditommaso:
Apology to be silent on this. This was never documented because makes the pipeline not resumable. Could not be self-defeating?

drpatelh:
I still think it would be useful. From what I gather the advantage of this option compared to `nextflow clean` is that it will be run automatically at the end of the pipeline without having to execute a separate command. As long as we are explicit in the documentation as to it's limitations it would be a nice, "official" addition. We have quite a few configs using it on [nf-core/configs](https://github.com/nf-core/configs/search?q=cleanup) now too :)

jfy133:
Agreed, as long as it's explicit it's fine.

I can update the documentation accordingly if you wish

Certainly for use at my institute, once a pipeline is complete people rarely use the run again.

pditommaso:
Ok, let's highlight better then. What about the following? 

```
================== ================
Name                Description
================== ================
cleanup             If ``true``, on a successful completion of a run all files in *work* directory are automatically deleted.
================== ================

.. warning:: 
    The use of the above ``cleanup`` option will prevent the use of the *resume* feature on subsequent executions of that pipeline run. 
    Also, be aware that deleting all scratch files can take a lot of time especially when using shared file system or remote cloud storage.

``` 


can you incorporate the change if it's fine?

jfy133:
Fine with me! Updated!

pditommaso (approved):

PR #2157, opened 2021-06-08
Run tasks under user accounts in Azure Batch

manuelesimi:
This PR attaches the user identity to the pool. There are two options:

1) If the configuration specifies `runAs`, the task is run under that user. The assumption here is that the user already exists on the execution node in the pool.

2) `runAs` is omitted. In this case, the usual auto-user identity is explicitly attached to the task. For this identity, it is also possible to set `privileged` to `true` to run the task under elevated access (see https://docs.microsoft.com/en-us/azure/batch/batch-user-accounts#elevated-access-for-tasks).

Technically, Azure Batch also offers the possibility to create a user on the fly in the pool, but I don't feel it is needed in this context. Either the auto-user or the pre-existing user should cover all the use cases I can think of.

### DIFF
--- a/docs/azure.rst
+++ b/docs/azure.rst
@@ -267,18 +267,20 @@ azure.batch.accountName                         The batch service account name.
 azure.batch.accountKey                          The batch service account key.
 azure.batch.endpoint                            The batch service endpoint e.g. ``https://nfbatch1.westeurope.batch.azure.com``.
 azure.batch.location                            The batch service location e.g. ``westeurope``. This is not needed when the endpoint is specified.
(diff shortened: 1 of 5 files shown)


pditommaso (changes requested):
Nice! thanks a lot for this PR! there's a failing test to address. Everything else looks awesome 👍

manuelesimi:
Whoops, I didn't notice the side effect on the hash. I fixed the expected poolId.

pditommaso:
Included in release `21.06.0-edge`.

vinishgeorge:
I have test this and can confirm its working for me . Thank you @pditommaso  @manuelesimi for the quick response !!.

Issue #2220, opened 2021-07-14
publishDir enabled as a dynamic directive


rcannood:
## Bug report 

I want to create a process for which I can use a dynamic directive to determine whether to publish the output or not. 

```groovy
nextflow.enable.dsl=2

process test_process {
  publishDir "output/$id", mode: 'copy', overwrite: true, enabled: { publish } // <--- dynamic
  input:
    tuple val(id), val(publish)
  output:
    tuple val(id), path("output.txt")
  script:
    """
    echo $id > output.txt
    """
}

workflow {
  Channel.fromList([ ["foo", false], ["bar", true] ])
    | test_process
}
```

### Expected behavior and actual behavior
For the example above, I would expect only `output/bar` to be created. The script above creates both, however:

```bash
$ tree output
output
├── bar
│   └── output.txt
└── foo
    └── output.txt
```

### Steps to reproduce the problem

Save the example shown above as `main.nf` and run `NXF_VER=21.04.1 nextflow run main.nf`.

### Program output 

```
$ NXF_VER=21.04.1 nextflow run main.nf 
N E X T F L O W  ~  version 21.04.1
Launching `main.nf` [ecstatic_turing] - revision: 1ad5fd0bcc
executor >  local (2)
[a6/489402] process > test_process (2) [100%] 2 of 2 ✔
```

### Environment 

* Nextflow version: 21.04.1
* Java version: openjdk version "11.0.11" 2021-04-20
* Operating system: Fedora 33
* Bash version: GNU bash, version 5.0.17

pditommaso:
The expected syntax consists of using an interpolated string value, which is implicitly converted to closure under the hood. 

However, it turns out the string value `"true"` and `"false"` are not correctly converted to boolean values. I've uploaded a patch to solve this problem 651835d2d.

rcannood:
Thanks, this solves the unexpected behaviour to a large extent!

PR #2221, opened 2021-07-14
Allow publishDir params to be closures

rcannood:
This is related to #2220. Currently, if I specify something like:
```groovy
publishDir "output/$id", mode: 'copy', overwrite: true, enabled: { publish }
```

Then `result.enabled = params.enabled as boolean` ([`PublishDir.groovy#L156`](https://github.com/nextflow-io/nextflow/blob/190aa4e/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy#L156)) turns this into a `true` value, regardless of the closure type. Similarly, specifying `enabled: "false"` will also result in it being converted to `true`.

This PR adds setters for parameters 'enabled', 'overwrite', and 'pattern' such that they can handle closured and convert strings to booleans properly, if necessary. Alternatively, if the output of a closure is not a boolean, an error could be produced.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
@@ -118,6 +118,42 @@ class PublishDir {
         this.mode = mode
     }
 
+    void setEnabled( obj ) {
+        // first check if obj is a closure
+        if (obj instanceof Closure) {
+            obj = obj.call()
+        }
+        // then check if obj is a string-like
+        if (obj instanceof CharSequence) {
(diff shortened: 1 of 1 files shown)


pditommaso:
Thanks a lot for submitting this PR. However, the problem is solved by 651835d2d, therefore I'm closing this w/o merging it.

rcannood:
Hello @pditommaso,

Thanks for pointing this out. It's good to see that string conversion is being handled in a much better way.

However, for my project, I really need to be able to pass a closure to this parameter; and this is what my PR was mostly about. 

Is it okay for you if I merge your latest changes into my branch, rework my solution, and then create a new PR?

Kind regards,
Robrecht

Issue #2222, opened 2021-07-14
Closures for publishDir params


rcannood:
## New feature

It would be super useful to be able to turn off or on a certain `publishDir` dynamically. I propose allowing closures to be passed to the parameters of `publishDir`: `enabled`, `overwrite` and `pattern`.

## Usage scenario 

I have a pipeline where I might want to publish some files, but not others. By allowing closures, I can define my process as follows:
```groovy
process test_process {
  publishDir "output/$id", mode: 'copy', overwrite: true, enabled: { publish } // <--- dynamic
  input:
    tuple val(id), val(publish)
  output:
    tuple val(id), path("output.txt")
  script:
    """
    echo $id > output.txt
    """
}
```

## Suggest implementation 

I suggested an implementation with PR #2223. I'd be happy to accept suggested changes or discuss the implementation.

pditommaso:
This is already possible using the syntax `enabled: "$publish"` and the patch provided in #2220

rcannood:
Right. It's great that the solution provided in #2220 already allows dynamically configuring this parameters using GStrings!

I think this PR really provides added benefit, because the documentation on [dynamic directives](https://www.nextflow.io/docs/latest/process.html#dynamic-directives) clearly state that you can uses closures to dynamically set your directives. If these particular parameters only allow using GStrings instead of closures to set a *boolean* value, I think this might be a bit confusing for the end user. 

To better understand your point of view, why wouldn't you want to accept this PR?

pditommaso:
This is an interesting point, in principle this should be handled in a uniform way for all directive attributes by this code 

https://github.com/nextflow-io/nextflow/blob/65e2a4d056d8514ae427ac6992c6f06cc13d1271/modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy#L527-L527


This is why I'd avoid handling it separately by for `publishDir`. Still not sure if the missing closure invocation in the above snippet is a bug or was done on purpose.

pditommaso:
Actually, the `param` parameter is meant for that.

pditommaso:
To summarise, this is the expected behvior. Dynamic value for `enabled` can be provided with the equivalent string syntax as commented above. 

I've added a clarification about this in the docs 93ac1040a

PR #2223, opened 2021-07-14
Issue 2222 publishdir closures

rcannood:
This is a suggested implementation for #2222.

It would be super useful to be able to turn off or on a certain `publishDir` dynamically. I propose allowing closures to be passed to the parameters of `publishDir`: `enabled`, `overwrite` and `pattern`.

## Usage scenario 

I have a pipeline where I might want to publish some files, but not others. By allowing closures, I can define my process as follows:
```groovy
process test_process {
  publishDir "output/$id", mode: 'copy', overwrite: true, enabled: { publish } // <--- dynamic
  input:
    tuple val(id), val(publish)
  output:
    tuple val(id), path("output.txt")
  script:
    """
    echo $id > output.txt
    """
}
```

## Suggest implementation 
I added three setters which check whether the object is a closure or not. I also added a few unit tests for verifying whether everything works, and it seems to be ok:

```
$ ./gradlew test
BUILD SUCCESSFUL in 1m 54s
69 actionable tasks: 69 executed
```


pditommaso:
Closing w/o merging as for comment https://github.com/nextflow-io/nextflow/issues/2222#issuecomment-918447423

PR #2262, opened 2021-08-10
Make sure to stage in run script when fixing ownership in aws batch

jchorl:
Weird quirk here -

When fixing container ownership in batch, but not tracing, the run file wasn't staged. This caused the process to erroneously fail.

I thought of adding tests but it doesn't seem useful. There's a fundamental mismatch in dependencies here - the logic to call the script is separate from the staging logic. There's no way (afaict) with the code structured how it is, to validate that _any_ time the trace file is used, it should be staged. Lmk if there's a good way to test for that.

I also improved an error message because I got snagged on that when building from src.

### DIFF
--- a/launch.sh
+++ b/launch.sh
@@ -67,9 +67,10 @@ fi
 JAVA_VER=$(echo "$JAVA_VER" | awk '/version/ {gsub(/"/, "", $3); print $3}')
 major=${BASH_REMATCH[1]}
 minor=${BASH_REMATCH[2]}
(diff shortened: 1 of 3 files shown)


pditommaso (changes requested):
Thanks for this PR. Yes, this makes sense, tho I think it would be preferable to keep the changes to `nextflow` into a  separate PR.

pditommaso:
Ok, the patch for the AWS Batch task handler was merged manually. Thanks for submitting it.

PR #2263, opened 2021-08-12
Look for templates also in the module directory.

manuelesimi:
This PR is related to DSL2. The proposed change adds the module directory to the locations where NF looks for template files.

**Why?**
Templates go along with processes: there is no template without a process. With DSL2, processes can be defined in a module script located in any arbitrary location and be referred with the `include` keyword. 

To me it makes sense to add the possibility to have a dedicate `templates` folder inside the directory where the process is defined. Why processes can be anywhere while their template must be inside the `baseDir\templates` folder? I know that we can use an absolute template path, but this is something I'd avoid in a pipeline versioned on Git. Also, absolute paths would break the pipeline sharing feature.

**How does it work?**
If the template file is not found in `baseDir\templates`, the moduleDir for the task's script is retrieved and checked. This also means that templates in the baseDir override the ones in the moduleDir. This is consistent with the behavior NF keeps with `params`. 

**Use cases**
I see this feature helpful in 2 cases:

1. This first case is actually the one that prompted to propose the change in this PR. Our team is migrating to DSL2 in order to be able to build libraries of processes to share across pipelines. Let's suppose we have the following projects:

A first project with one module script defining 2 processes (P1 and P2) and both use templates:
```
Library L
  |-myModules.nf 
  |-templates
      |-P1-template.sh
      |-P2-template.sh
```
Then, a second project with a workflow that includes P1 and P2
```
Pipeline A
  |-main.nf
```
Finally, a third project with a workflow that includes again P1 and P2
```
Pipeline B
  |-main.nf

```
With the new template location, Library L can be used by A and B without any changes, because it is self-contained. Processes and templates are both in the Library L project. A future Pipeline C would do the same, just cloning L and including its module script. In this regard, I believe this is a step forward toward the complete externalization of the process definition. 

2. As side effect, the PR promotes a more structured organization within the same project. Let's suppose we have tens of module scripts, and all the processes there use templates. A user can now decide to organize the project grouping module scripts and their templates under the same folder as needed. For example:
```
baseDir
  |-main.nf 
  |-Phase0-Modules
    |-mymodules0.nf
    |-mymodules1.nf
    |-templates
        |-P1-template.sh
        |-P2-template.sh
 |-Phase1-Modules
    |-mymodules0.nf
    |-mymodules1.nf
    |-templates
        |-P3-template.sh
        |-P4-template.sh
 |-Phase2-Modules
    |-mymodules0.nf
    |-mymodules1.nf
    |-templates 
        |-P5-template.sh
        |-P6-template.sh
        |-P7-template.sh
```

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskContext.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskContext.groovy
@@ -311,6 +311,14 @@ class TaskContext implements Map<String,Object>, Cloneable {
         if( path.isAbsolute() )
             return path
 
+        // make from the module dir
+        def module = this.script.getBinding()?.getVariable('moduleDir') as Path
+        if( module ) {
+            def target = module.resolve('templates').resolve(path)
+            if (Files.exists(target))
+                return target
+        }
+
         // otherwise make from the base dir
         def base = Global.session.baseDir
         if( base ) {
@@ -319,13 +327,6 @@ class TaskContext implements Map<String,Object>, Cloneable {
                 return target
         }
 
-        //.. or make from the module dir
-        def module = this.script.getBinding()?.getVariable('moduleDir') as Path
-        if( module ) {
-            def target = module.resolve('templates').resolve(path)
-            if (Files.exists(target))
-                return target
-        }
         // if the base dir is not available just use as it is
         return path
     }
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskContext.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskContext.groovy
@@ -17,6 +17,8 @@
 
 package nextflow.processor
 
+import nextflow.script.ScriptMeta
+
 import java.nio.file.Path
 import java.nio.file.Paths
 import java.nio.file.Files
@@ -312,7 +314,7 @@ class TaskContext implements Map<String,Object>, Cloneable {
             return path
 
         // make from the module dir
-        def module = this.script.getBinding()?.getVariable('moduleDir') as Path
+        def module = ScriptMeta.get(this.script)?.getModuleDir()
         if( module ) {
             def target = module.resolve('templates').resolve(path)
             if (Files.exists(target))
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskContext.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskContext.groovy
@@ -19,6 +19,7 @@ package nextflow.processor
 
 import java.nio.file.Path
 import java.nio.file.Paths
+import java.nio.file.Files
 import java.util.concurrent.atomic.AtomicBoolean
 
 import com.esotericsoftware.kryo.io.Input
@@ -310,11 +311,21 @@ class TaskContext implements Map<String,Object>, Cloneable {
         if( path.isAbsolute() )
             return path
 
-        // otherwise make
+        // otherwise make from the base dir
         def base = Global.session.baseDir
-        if( base )
-            return base.resolve('templates').resolve(path)
+        if( base ) {
+            def target = base.resolve('templates').resolve(path)
+            if (Files.exists(target))
+                return target
+        }
 
+        //.. or make from the module dir
+        def module = this.script.getBinding()?.getVariable('moduleDir') as Path
+        if( module ) {
+            def target = module.resolve('templates').resolve(path)
+            if (Files.exists(target))
+                return target
+        }
         // if the base dir is not available just use as it is
         return path
     }
--- a/modules/nextflow/src/main/groovy/nextflow/script/BaseScript.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/BaseScript.groovy
@@ -100,7 +100,7 @@ abstract class BaseScript extends Script implements ExecutionContext {
         binding.setVariable( 'workflow', session.workflowMetadata )
         binding.setVariable( 'nextflow', NextflowMeta.instance )
         binding.setVariable('launchDir', Paths.get('./').toRealPath())
-        binding.setVariable('moduleDir', meta.scriptPath?.parent )
+        binding.setVariable('moduleDir', meta.moduleDir )
     }
 
     protected process( String name, Closure<BodyDef> body ) {
--- a/modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy
@@ -72,8 +72,11 @@ class ScriptMeta {
     /** the script {@link Class} object */
     private Class<? extends BaseScript> clazz
 
-    /** The location path from there the script has been loaded */
+    /** The location path from where the script has been loaded */
     private Path scriptPath
+
+    /** The location path from where the template file is loaded */
+    private Path moduleDir
 
     /** The list of function, procs and workflow defined in this script */
     private Map<String,ComponentDef> definitions = new HashMap<>(10)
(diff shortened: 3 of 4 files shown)


pditommaso:
This is a nice contribution. However, I think it should be reversed logic i.e. before checking in the `moduleDir` and then in the top-level template, otherwise, a module template can be broken by a workflow template with the same added in a later stage.

manuelesimi:
@pditommaso Fair point. I just inverted the checks.

pditommaso:
Think we may improve a bit  with the following: 

1. Add `getModuleDir` to [ScriptMeta](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy#L88-L90) class 
2. Replace [this](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/script/BaseScript.groovy#L103-L103) with the new meta attribute 
3. use the in place of [this line](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/processor/TaskContext.groovy#L315-L315)
4. add a few uni test

pditommaso:
Thanks! 😄

manuelesimi:
> 3. use the in place of [this line](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/processor/TaskContext.groovy#L315-L315)

There is a missing word in this sentence, ("use the ? in place "). Can you please clarify?

Also, thanks for taking the time to explain the changes. It's an internal insight that helps me to better understand the NF codebase (for future PRs :) ).

manuele

pditommaso:
> There is a missing word in this sentence, ("use the ? in place "). Can you please clarify?

Oops, sorry fixed! 

> It's an internal insight that helps me to better understand the NF codebase (for future PRs :) ).

Indeed there's a lot of interesting stuff in the ScriptMeta 😎

manuelesimi:
I pushed some changes partially addressing the points in the comment of @pditommaso .

A bit of explanation:
> 3. use the `getModuleDir` in place of [this line](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/processor/TaskContext.groovy#L315-L315)

I think that this would not have the desired behavior. If we just replace the search in the `baseDir` with a search in the `moduleDir`, we might miss the templates in the `baseDir`. NF should support both, and for this reason I changed the way `moduleDir` is fetched but left the following code about  `baseDir` as it was (see https://github.com/manuelesimi/nextflow/commit/effe7802ccf47ec8d48122aee49b9beb05cbab1b)  

Of course, if you prefer to support templates only under `moduleDir` when this exists, I can proceed as per your comment.

> 4. add a few uni test

Here I have to admit I have some troubles and need help. I don't think that a simple test in `ScriptParserTest` to set and fetch the new `moduleDir` attribute would be meaningful. I mean, we know the parser works.

So I tried to add a test case to `TaskRunTest` that builds the expected structure under a temporary module directory and then  resolve the template there. For some reasons, even following the execution with the debugger, I don't understand why the REGISTRY does not return the ScriptMeta of my mock script.

Here's what I had (but didn't push):

In `TaskRunTest`:
       
     def 'should load a shell template file from the module directory'() {

        given:
        def task = new TaskRun()
        task.processor = [:] as TaskProcessor
        task.processor.grengine = new Grengine()

        // Random generator for unique folder name
        def generator = { String alphabet, int n ->
            new Random().with {
                (1..n).collect { alphabet[nextInt(alphabet.length())] }.join()
            }
        }
        // create a file template under the script dir
        def templateFile = TestHelper.createTemplateFileWithTemplateFolder(
                'template.sh',"myModuleDir-${generator((('A'..'Z')).join(), 9)}", 'just the content')
        def bind = new ScriptBinding([moduleDir: templateFile.parent.parent.toAbsolutePath()?.toString()])
        def script = new MockBindScript(bind)

        task.context = new TaskContext(script, [:],'foo')
        task.config = new TaskConfig().setContext(task.context)

        when:
        def moduleDir = ScriptMeta.get(task.context.getScript())?.getModuleDir()
        task.resolve( new BodyDef({-> template('template.sh')}, 'template...', 'shell'))
        def contextTemplate = task.context.template('template.sh')
        then:
        moduleDir.toAbsolutePath() == templateFile.toAbsolutePath()
        contextTemplate.toAbsolutePath() == templateFile.toAbsolutePath()
        task.source == 'just the content'
    }

With `MockBindScript` defined just as:

    class MockBindScript extends BaseScript {

        MockBindScript(Binding binding) {
            super(binding)
        }
       @Override Object runScript() { return null }
    }

So, `TestHelper.createTemplateFileWithTemplateFolder` was creating the correct module structure, and all was good until the `when` clause. There, out of the three attempts, none of them was able to fetch my mocked script, despite it was registered in the REGISTRY map.

I feel I'm missing something, but I can't figure it out by myself.

Beside, the changes I pushed work :)

pditommaso:
I think you are right [this change](https://github.com/manuelesimi/nextflow/commit/effe7802ccf47ec8d48122aee49b9beb05cbab1b#diff-e9b43ed24a2f9814ebc50c9f94b1d67daeabaeb9eb3cd86d8661227d4b371d56R96) looks good. 

I would avoid having an explicit `setModuleDir`, I think it could be only inferred from the `scriptPath`, or you where thinking to a more general mechanism  


The test, I understand it could be tricky if never used Spock before but likely simpler than the one you are trying. I think it would be possible to create a `Mock` of `TaskContext` and then verify the interaction with the template method. 

https://spockframework.org/spock/docs/1.0/interaction_based_testing.html

But if you reach a dead-end don't worry too much, I will add it.

manuelesimi:
> I think you are right [this change](https://github.com/manuelesimi/nextflow/commit/effe7802ccf47ec8d48122aee49b9beb05cbab1b#diff-e9b43ed24a2f9814ebc50c9f94b1d67daeabaeb9eb3cd86d8661227d4b371d56R96) looks good.
> 
> I would avoid having an explicit `setModuleDir`, I think it could be only inferred from the `scriptPath`, or you were thinking to a more general mechanism?

Nope. no general mechanism. You're absolutely right about setting `moduleDir` implicitly. I pushed the (tested) change.

> 
> The test, I understand it could be tricky if never used Spock before but likely simpler than the one you are trying. I think it would be possible to create a `Mock` of `TaskContext` and then verify the interaction with the template method.
> 
> https://spockframework.org/spock/docs/1.0/interaction_based_testing.html
> 

I tried both the interaction-based and state-based approaches. No luck.
This is my latest (unsuccessful) attempt:

```
   def 'should fetch template from module dir'() {
       setup:
        // Random generator for unique folder name
        def generator = { String alphabet, int n ->
            new Random().with {
                (1..n).collect { alphabet[nextInt(alphabet.length())] }.join()
            }
        }
        // create a file template under the script dir
        def templateFile = TestHelper.createTemplateFileWithTemplateFolder(
                'template.sh',"myModuleDir-${generator((('A'..'Z')).join(), 9)}", 'just the content')
        def bind = new ScriptBinding([moduleDir: templateFile.parent.parent.toAbsolutePath()?.toString()])
        def script = new MockScript(bind)
        //TaskContext context = Mock(TaskContext)
        TaskContext context = new TaskContext(script, [test:'test'], 'ciao')

        when:
        def results = context.template('template.sh').toAbsolutePath()
        then:
        script == context.getScript()
        results instanceof Path
        results.toAbsolutePath() == templateFile.toAbsolutePath()
    }

```

> But if you reach a dead-end don't worry too much, I will add it.

At this point, I'd be really happy to see how you would do it. Never liked to give up on a solution (but this is not "my" project and my time is limited).

pditommaso:
Ok, giving a look to it. Willing to add a mention in the docs to this new feature like in the template section and/or DSL2 docs? 

https://github.com/nextflow-io/nextflow/blob/1e485b32b25ba5a369696c2080432eb06391d641/docs/process.rst#template
https://github.com/nextflow-io/nextflow/blob/1e485b32b25ba5a369696c2080432eb06391d641/docs/dsl2.rst

pditommaso:
Closing this in favour of #2332. I've added a reference to this feature in process template section. If you could contribute a more detailed description in the DSL2 section following the use case described at beginning of this PR it would be great. Thanks!

PR #2285, opened 2021-09-01
Built-in support for Azure File Shares.

manuelesimi:
This PR introduces the capability to mount a selected [Azure File Share](https://azure.microsoft.com/en-us/services/storage/files/) on the pool nodes. 

Files available in the serverless Azure File share can be mounted concurrently on the nodes of a pool executing the pipeline.
These files become immediately available in the file system and can be referred as local files within the processes.  I believe this is especially useful when a task needs to access large amount of data (such as genome indexes) during its execution or to share data across pipelines. At our institute, we have several use cases that would benefit from this feature. 

**Configuration**
The storage section is extended with 3 new optional settings:
```
azure {
 ....
      storage {
        accountName = "<YOUR BLOB ACCOUNT NAME>"
        accountKey = "<YOUR BLOB ACCOUNT KEY>"
        fileName = '<YOUR SOURCE FILE SHARE>'
        relativeMountPath = '<YOUR MOUNT DESTINATION>'
        mountOptions = '<YOUR OPTIONAL MOUNT SETTINGS>'
      }
}
```
Where:

- `fileName` is the source file share (available in the Storage account)
-  `relativeMountPath` is the path where the file share will be mounted. This path is relative to the Batch mounts directory and it is accessible via the `AZ_BATCH_NODE_MOUNTS_DIR` environment variable available on the pool node. For instance, if the mount path is set to `myResources` the mounted files can be from the process' script with `$AZ_BATCH_NODE_MOUNTS_DIR\myResources`.
- `mountOptions` are the same options of the `mount` command in Linux. These are optionals and if not provided, Nextflow sets options common to most use cases.

**Sample usage**
In the `nextflow.config`:
```
storage {
    accountName = 'omitted'
    accountKey = 'omitted'
    fileName = 'exact2resources'
    relativeMountPath = 'resourcesFromAzureFiles'
  }
```
In the process' script:
```
process checkResources {
  publishDir params.resultsDir,  mode: 'copy', overwrite: true

  output:
  path('out.txt') into nothing

  script:
  """
  ls -lrt ${AZ_BATCH_NODE_MOUNTS_DIR}/resourcesFromAzureFiles/* >> out.txt
  """
}
```
After the execution, `out.txt` lists all the files in the File share named `exact2resources` available in the storage account.


manuelesimi:
Since this is still open, I made another change to make the AZ_BATCH_NODE_MOUNTS_DIR variable available inside the docker containers. With the latest commit, the mounted Azure Files can be accessed by both native and docker-based tasks in the same way.

docs/azure.rst
```diff
@@ -48,6 +48,34 @@ the file path with the ``az://`` prefix followed by the container name. For exam
...
+For example::
+
+    azure {
+      storage {
+        accountName = "<YOUR BLOB ACCOUNT NAME>"
+        accountKey = "<YOUR BLOB ACCOUNT KEY>"
+        fileName = '<YOUR SOURCE FILE SHARE>'
+        relativeMountPath = '<YOUR MOUNT DESTINATION>'
+        mountOptions = '<YOUR OPTIONAL MOUNT SETTINGS>'
+      }
```
pditommaso, line 72: Does Azure file share only allow the mount of a single system?

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -335,8 +337,7 @@ class AzBatchService implements Closeable {
         final containerOpts = new TaskContainerSettings()
                 .withImageName(container)
                 // mount host certificates otherwise `azcopy fails
-                .withContainerRunOptions('-v /etc/ssl/certs:/etc/ssl/certs:ro -v /etc/pki:/etc/pki:ro')
-
+                .withContainerRunOptions('-v /etc/ssl/certs:/etc/ssl/certs:ro -v /etc/pki:/etc/pki:ro -v /mnt/resource/batch/tasks/fsmounts:/mnt/resource/batch/tasks/fsmounts:rw -e AZ_BATCH_NODE_MOUNTS_DIR=/mnt/resource/batch/tasks/fsmounts')
```
pditommaso, line 340: This seems hardcoded. What if no Azure share is mounted? Also why the mount path is not configurable? I would expect to be able to mount it in the container into an arbitrary path?

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzStorageOpts.groovy
```diff
@@ -40,6 +43,9 @@ class AzStorageOpts {
         this.sysEnv = env==null ? new HashMap<String,String>(System.getenv()) : env
         this.accountKey = config.accountKey ?: sysEnv.get('AZURE_STORAGE_ACCOUNT_KEY')
         this.accountName = config.accountName ?: sysEnv.get('AZURE_STORAGE_ACCOUNT_NAME')
+        this.fileName = config.fileName ?: sysEnv.get('AZURE_STORAGE_FILE_NAME')
+        this.relativeMountPath = config.relativeMountPath ?: sysEnv.get('AZURE_STORAGE_MOUNT_PATH')
```
pditommaso, line 47: Don't think it's needed to allow configuring it via env vars

pditommaso (commented):
Apology the late revirew. Made some comments, mostly because don't know much Azure file share.

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
manuelesimi: This is an excellent catch because I forgot about this point. At the time I submitted the PR, I thought that was the only mount dir. After some experience and a better read of the documentation, I realized that there are different mount points. Specifically:

* `/mnt/resource/batch/tasks/fsmounts` on CentOS nodes
* `/mnt/batch/tasks/fsmounts` on Ubuntu nodes

So, the mount point should definitely go in the config file. For that, what do you think should be the best.

1. This configuration:
```
        storage {
          accountName = ''
          accountKey = ''
          fileName = '<my File Share Name>'
          relativeMountPath = '<my relative Path>'
          rootMountPath = '/mnt/resource/batch/tasks/fsmounts' // or another path, depending on the selected instance type
        }
```

2. Or this:

```
        storage {
          accountName = ''
          accountKey = ''
          fileName = '<my File Share Name>'
          mountPath = '/mnt/resource/batch/tasks/fsmounts/<my relative Path>' // or another path, depending on the selected instance type
```
Even if the second case, it's easier and more readable, I'm leaning for the first one because it lets to define the AZ_BATCH_NODE_MOUNTS_DIR variable in the docker container without manipulating the mount path. Defining this variable is important for portability, because it is available on the node to scripts that do not run inside a container.
I would like to hear your opinion before coding anything here (of course, any other suggestion is more than welcome!). In both cases, we can let Azure Batch returns the error is misconfigured.

Do note that we cannot mount the File Shares under an arbitrary path simply because it's Azure Batch that mounts it. We can only control the relative path.

Regarding your question what happens if no file share is mounted, the answer is... nothing. the dir referred AZ_BATCH_NODE_MOUNTS_DIR exists even if no File Share is mounted. So, worst case, we mount an empty folder.

pditommaso:
Think I'm missing some basic information how Azure Files works. My understanding that's a shared files system that's required to be mounted in the compute host environment and, once mounted in the host, is then mounted as a container volume. Is that correnct? 

I'm missing how the file system instance is reference in the first instance. 

For example, when using AWS EFS, you need to specify the EFS instance id, for [example](https://github.com/seqeralabs/xpack-amzn/blob/master/docs.md#use-of-aws-efs-file-system-with-nextflow). 

Instead, when using google NFS, the file system is identified using the storage IP/name (see [here](https://github.com/seqeralabs/xpack-google#1-nextflow-configuration)).

How the Azure files storage instance is identified?

manuelesimi:
> Think I'm missing some basic information on how Azure Files works. 

Sorry, this is likely my fault, I'll try to clarify your legitimated doubts.

>My understanding that's a shared files system that's required to be mounted in the compute host environment

This is correct.

> and, once mounted in the host, is then mounted as a container volume. 
>
This is partially correct, in the sense that Batch does not automatically mount the Azure File in the container. Nextflow will do that (if this PR is accepted) when a container is configured for the task. 

> I'm missing how the file system instance is specified.
> 
> For example, when using AWS EFS, you need to specify the EFS instance id, for [example](https://github.com/seqeralabs/xpack-amzn/blob/master/docs.md#use-of-aws-efs-file-system-with-nextflow).
> 
> Instead, when using google NFS, the file system is identified using the storage IP/name (see [here](https://github.com/seqeralabs/xpack-google#1-nextflow-configuration)).
> 
> How the Azure files storage instance is identified?

The Azure File is identified by an URL in the following fixed format:

`"https://{storage-account-name}.file.core.windows.net/{file-share-name}"`

Since we already had the storage account name in the batch configuration section, I thought it would be easier to just add the `fileName` setting to the `storage` section and build the URL inside Nextflow ([here](https://github.com/manuelesimi/nextflow/blob/d72ff75bf8c6b874c5d9a153beafbdc213a2f58c/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L610)).

Hope this helps, but feel free to ask any other question.

pditommaso:
Ok, now it's clear. what about supporting multiple shares using this syntax? 

```
azure.storage.fileShares.'<name>'. mountPath
azure.storage.fileShares.'<name>'. mountOptions
azure.storage.fileShares.'<name>'. rootPath // that should be what you called relativeMountPath if I'm understanding correctly 
```

manuelesimi:
> Ok, now it's clear. what about supporting multiple shares using this syntax?
> 
> ```
> azure.storage.fileShares.'<name>'. mountPath
> azure.storage.fileShares.'<name>'. mountOptions
> azure.storage.fileShares.'<name>'. rootPath // that should be what you called relativeMountPath if I'm understanding correctly 
> ```

Thanks, I'm all in for this change! 
I am going to add it to the branch later today.

manuelesimi:
I  just pushed the support for mounting an arbitrary number of Azure File Shares on each pool.

However, there is a significant difference with the configuration (and thus the implementation) proposed in the latest comment from @pditommaso .

The `mountPath` cannot be associated to the File Share: it belongs to the Pool. In fact, the mount path depends on the image reference selected in the Pool. The same File Share is mounted on different paths if the Image reference is based, for instance, on CentOS or Ubuntu.

Here’s how a minimal configuration looks like:

```
storage {
              accountName = 'OMITTED'
              accountKey = 'OMITTED'
              fileShares {
                  mytestfileshare1 {
		      rootPath = 'resourcesFrom_mytestfileshare1'
                       mountOptions = 'same as mount command' //optionals, NF defaults them
                  }
                  mytestfileshare2 {
                       rootPath = 'resourcesFrom_mytestfileshare2'
                       mountOptions = 'same as mount command' //optionals, NF defaults them
                  }
              }
}
```
By default, NF selects a CentOS-based image, therefore the default path is the one used in such nodes. If the user switches to Ubuntu (or another image reference), the mount path must be specified in the pool: 
```
batch {
              location = 'eastus2'
               // omitted batch settings...
               pools {
                      auto {
                            vmType = 'Standard_D16_v3'
                            vmCount = 4
                            maxVmCount = 4
                            offer = 'UbuntuServer'
                            // omitted ubuntu settings...
                            mountPath = '/mnt/batch/tasks/fsmounts'
                      }
               }
 }
storage {
              accountName = 'OMITTED'
              accountKey = 'OMITTED'
              fileShares {
                  mytestfileshare1 {
		      rootPath = 'resourcesFrom_mytestfileshare1'
                  }
                  mytestfileshare2 {
                       rootPath = 'resourcesFrom_mytestfileshare2'
                  }
              }
}
```
Once again, I remind that the `mountPath` is not under our control (Azure Batch rules here), the `rootPath` is arbitrary.

In both the configuration above, the mounted File Shares are accessed from the tasks in the same way, as follows:
```
echo "Content of File Share 1:"
ls -lrt ${AZ_BATCH_NODE_MOUNTS_DIR}/resourcesFrom_mytestfileshare1
echo "Content of File Share 2:"
ls -lrt ${AZ_BATCH_NODE_MOUNTS_DIR}/resourcesFrom_mytestfileshare2
```
@pditommaso What do you think? Is this implementation reasonable? Any drawbacks?

I'm leaving this PR as draft because:
1. I want to test a bit more the changes (especially with tasks without a container nor File Shares)
2. The doc still needs to be updated according to what just described

modules/nf-commons/src/main/nextflow/Const.groovy
```diff
@@ -58,12 +58,12 @@ class Const {
...
      * The app build time as linux/unix timestamp
      */
-    static public final long APP_TIMESTAMP = 1630341567917
+    static public final long APP_TIMESTAMP = 1632791057194
 
     /**
      * The app build number
      */
-    static public final int APP_BUILDNUM = 5609
+    static public final int APP_BUILDNUM = 5612
```
pditommaso, line 67: Better to not commit changes to Consts since it's modified automatically

pditommaso:
> The mountPath cannot be associated to the File Share: it belongs to the Pool

Ok, but it could be possible to be something smarter, in file share in mounted in the node in the path required by Azure, eg `/mnt/resource/batch/tasks/fsmounts/my-share-1`, `/mnt/resource/batch/tasks/fsmounts/my-share-2`, BUT each of the is mounted in the container using the corresponding `mountPath` provided by the user in the config .

it should be possible?   

Also is there's any variable in the node context to automatically infer the mount root independently the OS used? I mean about those: 

* /mnt/resource/batch/tasks/fsmounts on CentOS nodes
* /mnt/batch/tasks/fsmounts on Ubuntu nodes

pditommaso (changes requested):

modules/nf-commons/src/main/nextflow/Const.groovy
manuelesimi: Sorry, this was a mistake from my side. As you noticed, I never pushed the const.groovy in my other commits, this time I missed to exclude it.

manuelesimi:
> > The mountPath cannot be associated to the File Share: it belongs to the Pool
> 
> Ok, but it could be possible to be something smarter, in file share in mounted in the node in the path required by Azure, eg `/mnt/resource/batch/tasks/fsmounts/my-share-1`, `/mnt/resource/batch/tasks/fsmounts/my-share-2`, BUT each of the is mounted in the container using the corresponding `mountPath` provided by the user in the config .
> 
> it should be possible ~?~

Azure needs 2 information to mount a file shares

1.  the name of the share
2. the relative path (that Azure mounts under the `mountPath` that we associated to the Pool)

So, these must be both in the nextflow configuration.

If I understand correctly, you are now proposing to give to the users the possibility to choose the path where to mount `mountPath/relativePath` in the docker container.

If so, this is another setting to add to the configuration. We would have something like:

batch {
               // omitted batch settings...
               pools {
                      auto {
                            offer = 'UbuntuServer'
                            // omitted ubuntu settings...
                            mountPath = '/mnt/batch/tasks/fsmounts' //fixed, and defaulted if the user does not select a custom offer
                      }
               }
 }
storage {
              accountName = 'OMITTED'
              accountKey = 'OMITTED'
              fileShares {
                  mytestfileshare1 {
                      relativePath='resourcesFrom_mytestfileshare1'
		     customMountPath = '/my/arbitrary/path' //optional
                  }
              }
}

and in the docker run command NF will add the volume as:
` -v mountPath/relativePath:customMountPath`
or, if the custom path is not specified:
` -v mountPath/relativePath:mountPath/relativePath`

Is my understanding correct?

> 
> Also is there's any variable in the node context to automatically infer the mount root independently the OS used? I mean about those:
> 
> * /mnt/resource/batch/tasks/fsmounts on CentOS nodes
> * /mnt/batch/tasks/fsmounts on Ubuntu nodes
> 
> Update: should be the variable `AZ_BATCH_NODE_MOUNTS_DIR`, no?

It is, and yes, it's AZ_BATCH_NODE_MOUNTS_DIR. However, when I tried to mount it with:
`-v $AZ_BATCH_NODE_MOUNTS_DIR:AZ_BATCH_NODE_MOUNTS_DIR -e AZ_BATCH_NODE_MOUNTS_DIR`
I got an error that the var did not exist. 

This was one month ago, when I have less experience with Azure Files. I can try again, eventually this will only spare us the `mountPath` setting in the poo.

manuelesimi:
> I need to disambiguate the meaning of `mountPath` in the above sentence before replying. Is this the path where the shared is going to be mounted in the host node OR the _subdirectory_ within the shared file system that will be mounted? or something else?

With this config:

mountPath = '/mnt/resource/batch/tasks/fsmounts' //fixed, we don't control it and bound to the Pool
relativePath = 'my_folder' // the user chooses it. It's the subdirectory under mountPath where azure batch mounts the File Share

The files in the File Shares will be available in the pool node under:
`/mnt/resource/batch/tasks/fsmounts/my_folder`

and they can be accessed with ${AZ_BATCH_NODE_MOUNTS_DIR}/my_folder

There is no way to mount a subdirectory of the File Share, you can only mount the entire shared file system.

pditommaso:
OK, think I'm starting to understand 😄 

My point is that the `mountPath` and `relativePath` in the above comment are used to mount the share in the node pool. These settings should be hidden to the user and could be inferred (the first) or automatically generated (the second)

The  `mountPath` specified by the user should be used instead for mounting the share in the container. 

Therefore having this nextflow config 

```
storage {
              accountName = 'OMITTED'
              accountKey = 'OMITTED'
              fileShares {
                  mytestfileshare1 {
		     mountPath = '/my/arbitrary/path'
                  }
              }
} 
```

it could be mounted in the host as 


```
/mnt/resource/batch/tasks/fsmounts/mytestfileshare1
```

then in the container as 


```
-v /mnt/resource/batch/tasks/fsmounts/mytestfileshare1:/my/arbitrary/path
```

Makes sense?

manuelesimi:
I'm starting to understand your point too.

Your suggestion does make sense (actually, a lot of sense) and I see the great simplification it will introduce.

I'll work on implementing this approach and push it soon :)

manuelesimi:
Update from the latest commit:

The user can specify this in the config file:
```
	storage {
              accountName = ''
              accountKey = ''
              fileShares {
                    <name of the share> {
		      mountPath = '/mnt/any/place/here1'
                }
                    <name of another share> {
                        mountPath = '/mnt/another/place/subdir/anothersubdir'
                }
              }
            }
```

and in the task's script, the shared files can be accessed with:
```
echo "content of first share:"
ls -lrt /mnt/any/place/here1
echo "fileshare2:"
ls -lrt /mnt/another/place/subdir/anothersubdir
```

Internally, the file shares are mounted in the node as:
```
<OS mount Path>/<name of the share>
<OS mount Path>/<name of another share>
```
but the user has no perception of this.

As side effect of this new implementation, there is no need to expose the AZ_BATCH_NODE_MOUNTS_DIR var inside the container, since the File Shares are mounted under different and arbitrary locations.

docs/azure.rst
```diff
@@ -276,6 +313,7 @@ azure.batch.pools.<name>.vmType                 Specify the virtual machine type
 azure.batch.pools.<name>.vmCount                Specify the number of virtual machines provisioned by the pool identified with ``<name>``.
 azure.batch.pools.<name>.maxVmCount             Specify the max of virtual machine when using auto scale option.
 azure.batch.pools.<name>.autoScale              Enable autoscaling feature for the pool identified with ``<name>``.
+azure.batch.pools.<name>.fileShareRootPath      If mounting File Shares, this is the internal root mounting point. Must be ``/mnt/resource/batch/tasks/fsmounts`` for CentOS nodes or ``/mnt/batch/tasks/fsmounts`` for Ubuntu nodes (default is for CentOS).
```
pditommaso, line 316: think already asked, there's no way to infer this? maybe using  `uname` to get the distribution if there's no better  way?

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -585,10 +590,10 @@ class AzBatchService implements Closeable {
...
                 .withId(spec.poolId)
                 .withVirtualMachineConfiguration(poolVmConfig(spec.opts))
-                // https://docs.microsoft.com/en-us/azure/batch/batch-pool-vm-sizes
+        // https://docs.microsoft.com/en-us/azure/batch/batch-pool-vm-sizes
                 .withVmSize(spec.vmType.name)
-                // same as the num ofd cores
-                // https://docs.microsoft.com/en-us/azure/batch/batch-parallel-node-tasks
+        // same as the num ofd cores
+        // https://docs.microsoft.com/en-us/azure/batch/batch-parallel-node-tasks
                 .withTaskSlotsPerNode(spec.vmType.numberOfCores)
```
pditommaso, line 597: think we still have problem tabs vs space. NF code uses 4 spaces for tab.

docs/azure.rst
manuelesimi: The problem is that we need this info when we submit the tasks, we are not yet on the batch node to get this information on the fly. I tend to think that most users will go with the default and never touch this setting (like the others related to the image reference - they are all connected and should be considered at a whole).

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
manuelesimi: whoops, that's IDEA going its own way! I didn't notice the reformatting, I'll fix it.

manuelesimi: I "should" be OK now.

docs/azure.rst
pditommaso: I see, what if we default this with the `sku` config setting (once merged).

manuelesimi: I'm not sure to follow this. 

Do you want to assign the `fileShareRootPath` automatically based on the `sku` from https://github.com/nextflow-io/nextflow/pull/2360? It's doable, but I, for one, would never be sure that all the ubuntu images mount under the same X path, all the windows images mount under the same Y path, etc. Not to mention custom images created by Azure users.

Because of that, I'd still leave to the user the option to set `fileShareRootPath`. Then if it's not set, we could go with a map like:

/.*ubuntu.*/ -> '/mnt/batch/tasks/fsmounts'
/.*centos.*/ -> '/mnt/resource/batch/tasks/fsmounts'

What do you think?

pditommaso: > Because of that, I'd still leave to the user the option to set fileShareRootPath. Then if it's not set, we could go with a map like:

exactly

pditommaso:
A few conflicts to solve ☝️

manuelesimi:
> A few conflicts to solve ☝️

They should be solved now.

manuelesimi:
> > A few conflicts to solve ☝️
> 
> They should be solved now.

So, I fixed a failing test on the pool ID, but, of course, now it conflicts with the test fixes for https://github.com/nextflow-io/nextflow/pull/2360. Since they both affect the pool ID, the only way I see is to fix it again once merged.

pditommaso:
Nope, you can merge `master` on this branch (solve the conflicts if persists) and push the branch. An now .. buonanotte! 😄

manuelesimi:
> Nope, you can merge `master` on this branch (solve the conflicts if persists) and push the branch. An now .. buonanotte! 😄

:)

manuelesimi:
I did the merge from master, fixed again the pool Id in the test. It should be good.

There is a failing test now, due to AWS credentials:
`
Access Denied (Service: Amazon S3; Status Code: 403; Error Code: AccessDenied; Request ID: 5AHHHXTW8TJF54NP; S3 Extended Request ID: L/1OFafJhODCoEOBW9+hLr1WK/ofciZDphuUQIQq4Jn7X19i/iCZQJDws6O4/iPljjgfsoTI1Is=; Proxy: null)
`

pditommaso:
I'll check why the keys are missing. it should work

pditommaso:
Not sure why this is happening, the docs says 

> Secrets are not passed to workflows that are triggered by a pull request from a fork

manuelesimi:
> Ah, this is a PR from your own fork, therefore it's expected. Need to check how to disable this test in this case.

Yep, this PR was started before I could create branches in the project.

> Merging this commit from master should solve [c405f9e](https://github.com/nextflow-io/nextflow/commit/c405f9ebe778aef6113afdc0261faadf2e2f98cc).

I did it, and all the checks have passed.

pditommaso:
Cool, think last point remain to address is the comment https://github.com/nextflow-io/nextflow/pull/2285#pullrequestreview-771565427

ie. having a sensible default for `fileShareRootPath` depending the `sku`

docs/azure.rst
```diff
@@ -48,6 +48,40 @@ the file path with the ``az://`` prefix followed by the container name. For exam
 container a file named ``foo.txt`` you can access it in your Nextflow script using the following fully qualified
 file path ``az://my-data/foo.txt``.
 
+Azure File Shares
+==================
+
+Nextflow has built-in support also for `Azure Files <https://azure.microsoft.com/en-us/services/storage/files/>`_.
+Files available in the serverless Azure File shares can be mounted concurrently on the nodes of a pool executing the pipeline.
+These files become immediately available in the file system and can be referred as local files within the processes. This
```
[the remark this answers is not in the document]
manuelesimi: Done :)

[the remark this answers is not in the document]
pditommaso: Shouldn't be in the * Azure File Shares* instead of * Azure Blob Storage* .. 😕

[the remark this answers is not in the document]
manuelesimi: oh well... you're a fantastic human debugger and fixer!

Pushed the change.

pditommaso:
That happens sometimes. there should be some crazy race condition that should be investigated

pditommaso:
OK, let's merge it. Thanks for this awesome contribution and the patience with my requests!

manuelesimi:
I'm going to miss this PR, but it's time to let it go :)

Thanks for all the feedback along, I learned a lot more about Nextflow.

PR #2286, opened 2021-09-01
Add sqlite jdbc driver for the nf-sqldb plugin

abhi18av:
I noticed that through the new `nf-sqldb` plugin, we have already added support for 
- h2
- mysql
- mariadb
- postgresql

I thought that it makes sense to add `sqlite` driver since it is quite popular and is often used by CLI tools as a simpler database.

I'd be happy to accomodate any change/enhancement requests.

### DIFF
--- a/plugins/nf-sqldb/build.gradle
+++ b/plugins/nf-sqldb/build.gradle
@@ -41,6 +41,7 @@ dependencies {
     api 'mysql:mysql-connector-java:8.0.22'
     api 'org.mariadb.jdbc:mariadb-java-client:2.7.0'
     api 'org.postgresql:postgresql:42.2.23'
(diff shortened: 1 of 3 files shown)


pditommaso:
This could be useful, tho I think it requires installing the sqlite locally. Have you tried it?

abhi18av:
>  I think it requires installing the sqlite locally.

Yes. Although once, installed there are two ways to use it (i) file based (ii) [memory based](https://www.sqlite.org/inmemorydb.html) 


> Have you tried it?

Yes, I wrote some tests against the locally installed `sqlite3` but since we might not have the binary in CI env, I haven't included those tests. The tests are passing for me, 

image

pditommaso (approved):
Worth merging

Issue #1798, opened 2020-11-18
Local scope of scripts in bin directory


phiweger:
## New feature

When we import a subworkflow into a larger one, and this subwf uses scripts (like some Python parser), then currently this script needs to be copied from the subwf's bin directory to the larger wf's bin directory. However, wouldn't it me more logical for the subwf to look if it can find the script "locally" first?

abhi18av:
I guess that, in the same vein, same update could be made for the script `templates` as well.

grst:
This feature would be super helpful for [nf-core/modules](https://github.com/nf-core/modules)! 

To some degree, this is already possible with `$moduleDir`. It is possible to run a script from the module directory using 

```nextflow
script:
"""
${moduleDir}/bin/my_script.sh
"""
```

However, this breaks the caching functionality, i.e. modifying `my_script.sh` does not invalidate the cache. 

Moving forward, I see two options:
 * support a local `bin` directory that behaves the same way as the global one
 * consider files from `$moduleDir` when calculating the hash. 

I'm slightly in favor of the latter option, because
 * it is explicit, and avoids confusion and namespace clashes with the global bin dir
 * it does not only allow to get binaries, but also other resource files (e.g. report templates) from the module directory while being cache aware. 

---
CC @drpatelh, @ewels

grst:
Another issue with `$moduleDir` is that it doesn't get mounted into docker containers. The scripts will, therefore, not be found when using the docker engine.

drpatelh:
Yup, I agree it would be nice to eventually find a solution to this because it means custom modules that need to be shipped with their own executables can be packaged up and be self-contained in their own right too. @grst has created a proof of principle for us to be able to add such a [custom module](https://github.com/nf-core/modules/issues/294) to nf-core/modules but ran into the Docker mount path issue above. 

As @phiweger mentioned the only viable solution at present is to copy across the executable in the `bin/` directory of the main workflow for it be found correctly on `$PATH`.

grst:
Would be great if at least the mounting-issue could be solved. The caching is less important as at least for nf-core pipelines the end user should never change the script anyway. 

I created a separate issue to track the mounting problem, including a reproducible example: https://github.com/nextflow-io/nextflow/issues/2240

grst:
Before the stale-bot gets at this issue, I would want to throw in that the next nf-core hackathon is approaching and this issue will inevitably come up there again ;)

ewels:
Bump

[...]

PR #2332, opened 2021-09-18
Add ability to define templates library at module level

pditommaso:
This PR replaces #2263 adding some minor refactoring and unit testing.

### DIFF
--- a/docs/dsl2.rst
+++ b/docs/dsl2.rst
@@ -463,8 +463,6 @@ The template files can be made available under the local ``templates`` directory
 			|-P1-template.sh
 			|-P2-template.sh
 
-
-
 Then, we have a second project A with a workflow that includes P1 and P2::
 
 	Pipeline A
--- a/docs/dsl2.rst
+++ b/docs/dsl2.rst
@@ -489,22 +489,22 @@ has several module components, and all them use templates, the project could gro
 		|-Phase0-Modules
 			|-mymodules1.nf
 			|-mymodules2.nf
-				|-templates
-					|-P1-template.sh
-					|-P2-template.sh
+			|-templates
+				|-P1-template.sh
+				|-P2-template.sh
 		|-Phase1-Modules
 			|-mymodules3.nf
 			|-mymodules4.nf
-				|-templates
-					|-P3-template.sh
-					|-P4-template.sh
+			|-templates
+				|-P3-template.sh
+				|-P4-template.sh
 		|-Phase2-Modules
 			|-mymodules5.nf
 			|-mymodules6.nf
-				|-templates
-					|-P5-template.sh
-					|-P6-template.sh
-					|-P7-template.sh
+			|-templates
+				|-P5-template.sh
+				|-P6-template.sh
+				|-P7-template.sh
 
 
 Channel forking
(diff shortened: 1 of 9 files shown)


manuelesimi:
Looks good to me! (Eventually, I still would love to see and learn how to test the `TaskContext` state).

manuelesimi:
I added a `Module Templates` section to the DSL2 doc.

I noticed how you tested the behavior of `TaskContext` with the new template location. Brilliant, congrats! I realize I didn't have enough knowledge of NF to write something like that.... For one, I was forgetting to enable DSL2 in my tests:)

modules/nextflow/src/main/groovy/nextflow/processor/TaskContext.groovy
```diff
@@ -303,17 +307,28 @@ class TaskContext implements Map<String,Object>, Cloneable {
...
+
+        // otherwise make from the base dir
         def base = Global.session.baseDir
-        if( base )
-            return base.resolve('templates').resolve(path)
+        if( base ) {
+            def target = base.resolve('templates').resolve(path)
+            if (Files.exists(target))
+                return target
+        }
```
pditommaso, line 332: I would keep the old version ie. returning the file path even if not existing to avoid unexpected side effects with legacy code.

pditommaso (commented):
Thinks looks great. Just made a comment on file template existence check.

modules/nextflow/src/main/groovy/nextflow/processor/TaskContext.groovy
manuelesimi: Sorry, I didn't mean to break backward compatibility. Just fixed.

abhi18av:
I'm just curious here, the approach used in this PR , could also be applied for `module level bin folder` as well right ?

https://github.com/nextflow-io/nextflow/issues/1798

manuelesimi:
@abhi18av The rational behind #1798  is indeed the same of this PR. They both aim to achieve self-contained module components. However, the solution found in this PR does not apply for the `bin` folder.
The `templates` dir is resolved locally, before the task is submitted to the execution layer. On the other hand, the `bin` dir must be uploaded on the node where the task is executed. Another difference is that there is only one template file per task, while multiple executable scripts can be both in the module's and workflow's `bin` folder.

That said, solving #1798  would nicely pair with the solution we got here.

docs/dsl2.rst
```diff
@@ -447,6 +447,66 @@ The above snippet prints::
...
+
+	baseDir
+		|-main.nf
+		|-Phase0-Modules
+			|-mymodules1.nf
+			|-mymodules2.nf
+				|-templates
+					|-P1-template.sh
+					|-P2-template.sh
+		|-Phase1-Modules
```
pditommaso, line 495: Think this is not correct, the `templates` dir is resolved as a sibling of module file, therefore it should be 

```
		|-Phase0-Modules
			|-mymodules1.nf
			|-mymodules2.nf
			|-templates
				|-P1-template.sh
				|-P2-template.sh
```

manuelesimi: Indeed I indented too much here :)
Good catch! It would have been really confusing for the users.

abhi18av:
> On the other hand, the bin dir must be uploaded on the node where the task is executed.

Ah, I see!

Thanks for the detailed explanation @manuelesimi!

pditommaso:
Ok, think we can merge this. Nice contribution @manuelesimi, thanks!

Issue #1732, opened 2020-09-24
/bin/bash: .command.sh: No such file or directory


Lcornet:
## Bug report 

When I activate the singularity container, nextflow crash and tells me that .command.sh don't exist
This appears after an update to singularity 3.6.0

### Expected behavior and actual behavior

nextflow ORPER.nf --refgroup=Nostocales --outgroup=Gloeobacterales --ribodb=/data1/ORPER/ribo/ --companion=ORPER/ORPER-companion.py --SSU=Nostocales-silva.fasta --cdhit=no

This commands and pipeline work well. but is I activate the container with : 

process.container = '/data1/ORPER/ORPER.sif'
singularity.enabled = true

it crash

### Steps to reproduce the problem

but when  I activate the container with : 

process.container = '/data1/ORPER/ORPER.sif'
singularity.enabled = true

it crash

### Program output 

ommand exit status:
  1

Command output:
  (empty)

Command error:
  /bin/bash: line 0: cd: /data1/ORPER/work/17/5e87858643dcc828837406eecff1b8: No such file or directory
  /bin/bash: .command.sh: No such file or directory

Work dir:
  /data1/ORPER/work/17/5e87858643dcc828837406eecff1b8

### Environment 

* Nextflow version: 20.07
* Java version: openjdk 11.0.8 2020-07-14
* Operating system: Linux 18
* Bash version: GNU bash, version 4.4.20(1)-release (x86_64-pc-linux-gnu)

sckinta:
Is this issue solved? I used most updated nextflow (version 21.10.6.5660), but still have the same error. my singularity version is 3.7.0

Lcornet:
I finally found a solution for our cluster. The config file (nextflow.config) should look like this:

process.container = 'PATH to your container'
singularity.enabled = true
singularity.cacheDir = "$PWD"
singularity.autoMounts = false
singularity.runOptions = '-B <link1> -B <link2>'

the two links can be identified with pwd and readlink -f .

Hope it helps

james-guevara:
When using a singularity container, I have to specify a bind path in the nextflow.config like so:

```
singularity {
    enabled = true
    runOptions = "--bind $PWD"
}
```

where $PWD is the folder with my main.nf file, nextflow.config file, and work folder.

Issue #1598, opened 2020-05-07
Use "/usr/bin/env bash" instead of "/bin/bash"


unode:
## Bug report 

```
% nextflow run hello
N E X T F L O W  ~  version 20.04.1
Launching `nextflow-io/hello` [modest_lamarr] - revision: 1d43afc0ec [master]
WARN: The use of `echo` method is deprecated
executor >  local (4)
[a8/932405] process > sayHello (2) [100%] 1 of 1, failed: 1
WARN: Killing pending tasks (3)
Error executing process > 'sayHello (4)'

Caused by:
  Cannot run program "/bin/bash" (in directory "/tmp/work/7d/a941cf3b6cae277c641136cd0aee2d"): error=2, No such file or directory

Command executed:

  echo 'Hola world!'

Command exit status:
  -

Command output:
  (empty)

Work dir:
  /tmp/work/7d/a941cf3b6cae277c641136cd0aee2d

Tip: you can replicate the issue by changing to the process work dir and entering the command `bash .command.run`
```
The `Tip` actually runs as expected
```
% cd /tmp/work/fa/adc77db68ead7e6681d4e24f3df388
% bash .command.sh 
Bonjour world!
```

### Expected behavior and actual behavior

On some systems `/bin/bash` doesn't exist. This causes failures at many levels since `/bin/bash` is hardcoded in many locations in nextflow's codebase, including the shebang of the `nextflow-X.X.X.all` release files.

The compatible alternative is to use `/usr/bin/env bash` or, if involving shebangs and multiple arguments, `#!/usr/bin/env -S bash -ue`. See also [coreutils - env -S](https://www.gnu.org/software/coreutils/manual/html_node/env-invocation.html#g_t_002dS_002f_002d_002dsplit_002dstring-usage-in-scripts).

### Environment 

* Nextflow version:  20.04.1
* Java version: openjdk 11.0.6-internal 2020-01-14
* Operating system: Linux - NixOS 20.03
* Bash version: (use the command `$SHELL --version`)
```
% $SHELL --version
zsh 5.7.1 (x86_64-pc-linux-gnu)
% bash --version
GNU bash, version 4.4.23(1)-release (x86_64-unknown-linux-gnu)
```

Emiller88:
Same issue on Nixos 20.03 also. 
```sh
#!/usr/bin/env bash
```
Works for me.

unode:
[...]

Issue #73, opened 2015-09-22
cannot in docker 1.8.2a under OSX


huguesfontenelle:
All examples from https://github.com/nextflow-io/examples run fine, but not when using `-with-docker`.

I'm using OSX 10.9.5.

```
$ nextflow run helloworld.nf -with-docker
N E X T F L O W  ~  version 0.15.3
Launching helloworld.nf
[warm up] executor > local
[96/e577c9] Submitted process > printHello (2)
[c9/81db50] Submitted process > printHello (1)
[d5/c9b4a2] Submitted process > printHello (3)
[9e/b6af1f] Submitted process > printHello (4)
Error executing process > 'printHello (3)'

Caused by:
  Process 'printHello (3)' terminated with an error exit status

Command executed:

  echo bonjour

Command exit status:
  1

Command output:
  (empty)

Command error:
  /bin/bash: .command.sh: No such file or directory

Work dir:
  /DEVEL/CLONES/nextflow-io/examples/work/d5/c9b4a21f3b58d4c97e933ae31f6bb5

Tip: you can try to figure out what's wrong by changing to the process work dir and showing the script file named: '.command.sh'

WARN: Killing pending processes (3)
```

I installed Docker through `brew cask`:
```
$ brew cask info dockertoolbox
dockertoolbox: 1.8.2a
Docker Toolbox
https://www.docker.com/toolbox
/opt/homebrew-cask/Caskroom/dockertoolbox/1.8.2a (178M)
https://github.com/caskroom/homebrew-cask/blob/master/Casks/dockertoolbox.rb
==> Contents
  DockerToolbox-1.8.2a.pkg (pkg)
```

Some additional info:
```
$ docker info
Containers: 3
Images: 49
Storage Driver: aufs
 Root Dir: /mnt/sda1/var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 55
 Dirperm1 Supported: true
Execution Driver: native-0.2
Logging Driver: json-file
Kernel Version: 4.0.9-boot2docker
Operating System: Boot2Docker 1.8.2 (TCL 6.4); master : aba6192 - Thu Sep 10 20:58:17 UTC 2015
CPUs: 1
Total Memory: 996.2 MiB
Name: default
ID: HWDD:YHXG:V7J3:CAKH:ZHU2:AQ2N:A6JM:XKEO:F5SI:44KH:S7OW:DDMR
Debug mode (server): true
File Descriptors: 11
Goroutines: 18
[...]

PR #2341, opened 2021-09-22
Replace /bin/bash with /usr/bin/env bash

unode:
This replaces all hardcoded `/bin/bash` paths to `/usr/bin/env bash`.

When additional arguments are given to bash (e.g. `bash -ue`), `/usr/bin/env -S bash -ue` is used instead.
The `env -S` functionality requires coreutils 8.30 or newer. This or a newer version of coreutils is present since Ubuntu 19.04, CentOS 8 and many others.

This PR also addresses issues with containers and Linux OSes where `/bin/bash` doesn't exist or where `bash` is installed in a different path (e.g. NixOS).
The PR should also fix partially or entirely the issues #73 #1598 and #1732 

Additional potential issues remain in the files:
```
modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
modules/nextflow/src/main/groovy/nextflow/k8s/K8sDriverLauncher.groovy
modules/nextflow/src/test/groovy/nextflow/container/DockerBuilderTest.groovy
modules/nextflow/src/test/groovy/nextflow/container/PodmanBuilderTest.groovy
modules/nextflow/src/test/groovy/nextflow/container/ShifterBuilderTest.groovy
modules/nextflow/src/test/groovy/nextflow/container/UdockerBuilderTest.groovy
modules/nextflow/src/test/groovy/nextflow/executor/BashWrapperBuilderTest.groovy
```

Where `/bin/bash` is still hardcoded to work with different images.
In fact, some of that code is likely non-functional as, for instance, the latest `busybox` docker image no longer includes `/bin/bash`.

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -1,2 +1,2 @@
-#!/bin/bash
+#!/usr/bin/env bash
 ./gradlew -q compile exportClasspath
--- a/docker/entry.sh
+++ b/docker/entry.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 #  Copyright 2013-2019, Centre for Genomic Regulation (CRG)
 #
@@ -43,7 +43,7 @@ if [[ "$NXF_USRMAP" ]]; then
 # create a `nextflow` user with the provided ID 
 # then change the docker socker ownership to `nextflow` user 
 addgroup docker
-adduser -u $NXF_USRMAP -G docker -s /bin/bash -D nextflow
+adduser -u $NXF_USRMAP -G docker -s "$(which bash)" -D nextflow
 chown nextflow /var/run/docker.sock  
 # finally run the target command with `nextflow` user
 su nextflow << EOF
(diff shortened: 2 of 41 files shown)


pditommaso:
> The PR should also fix partially or entirely the issues #73 #1598 and #1732

Sorry, can't merge this. This issue was already discussed #1598. The use of `/bin/bash` is a fair assumption. If the host uses a different path for the bash binary, it could be easily fixed by adding a symlink.

unode:
As explained in the description, that's not quite true and there are many cases where this may fail, NixOS being only one. Others include systems where `/bin/bash` is too old or incompatible (e.g. when `LD_LIBRARY_PATH` is tampered with due to complicated software stacks), in which case a newer `bash` may live somewhere in the user's `$PATH`.

As for a symlink, you seem to be assuming that users will have control over `/bin/bash`. This is rarely the case.

The proposed PR addresses all these.

One could make a case for `env -S` but if that's a concern, can we discuss alternatives, such as converting all the hardcoded shebangs and identical bash invocations into a user defined path?

Issue #2361, opened 2021-10-04
Resubmit failed pods


xhejtman:
Hello,

would it be possible to extend kuberun executor so that failed pods get resubmitted?

It can happen that pod is evicted for some reason, e.g. disk pressure, which is not permanent error and not error of pod itself. In such a case, pod should be resubmitted.

The corresponding error status looks like this:

```
"status": {​​​​​​​
          "phase": "Failed",
          "message": "The node was low on resource: ephemeral-storage. ",
          "reason": "Evicted",
          "startTime": "2021-10-04T11:53:32Z"
}​​​​​​​
```

bentsherman:
Related to #2302 I think.

xhejtman:
It does not seem to as the mentioned patch is already included in 21.09.0 version and it is still happening.

pditommaso:
it will be included in the next release

pditommaso:
Solved by f9db9cc77c718b4e5d9700359e0a2b8858fb4d5d

phue:
I can confirm that `21.09.0-edge` it is not resubmitting evicted pods. This is particularly problematic with SPOT instances on AWS as they can be shutdown anytime

xhejtman:
If the patch `f9db9cc` is changed to use NodeTerminationException instead of ProcessFailedException and this additional patch is applied:
```
diff --git a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
index 402cb5d9..3d331028 100644
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -2111,7 +2111,7 @@ class TaskProcessor {
 
             if( task.type == ScriptType.SCRIPTLET ) {
                 if( task.exitStatus == Integer.MAX_VALUE )
-                    throw new ProcessFailedException("Process `${task.name}` terminated for an unknown reason -- Likely it has been terminated by the external system")
+                    throw new NodeTerminationException("Process `${task.name}` terminated for an unknown reason -- Likely it has been terminated by the external system")
 
                 if ( !task.isSuccess() )
                     throw new ProcessFailedException("Process `${task.name}` terminated with an error exit status (${task.exitStatus})")
```

pipeline survives probably all k8s issues including node reboot.

PR #2404, opened 2021-10-22
k8s executor: throw NodeTerminationException on node shutdown

phue:
When using the `k8s` executor, nextflow currently does not handle node shutdown events well

This was supposedly addressed in https://github.com/nextflow-io/nextflow/commit/f9db9cc77c718b4e5d9700359e0a2b8858fb4d5d, however throwing a `ProcessFailedException` does not trigger resubmission of the process as intended.

The commit below applies a patch to throw `NodeTerminationException` instead (this patch was originally proposed by @xhejtman in https://github.com/nextflow-io/nextflow/issues/2361)


bentsherman (approved):

pditommaso (approved):
Yes, makes sense 👍 Thanks for contributing

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -2112,7 +2112,7 @@ class TaskProcessor {
 
             if( task.type == ScriptType.SCRIPTLET ) {
                 if( task.exitStatus == Integer.MAX_VALUE )
-                    throw new ProcessFailedException("Process `${task.name}` terminated for an unknown reason -- Likely it has been terminated by the external system")
+                    throw new NodeTerminationException("Process `${task.name}` terminated for an unknown reason -- Likely it has been terminated by the external system")
```
pditommaso, line 2115: Not sure about this tho. The reason of this error is not well defined. Think it should `ProcessFailedException` and retried is there's a NF error condition

xhejtman: I think this happens if node with the task is rebooted. It reports max exit status, if I remember corectly.

bentsherman: The question is whether this exit code is used for other error cases? As I recall, an exit code of `-1`  (aka `255`) is used for many types of errors. It may also be used differently by different executors. Keep in mind this code segment is not specific to k8s.

xhejtman: perhaps we solve this issue impelementing jobs instead of pods.

bentsherman: Sounds good. @phue Please revert your changes in `TaskProcessor.groovy`. We will merge the k8s specific change now, and we aim to address the other error in a different PR. See #2424 .

pditommaso:
1034 file changes .. something went wrong

phue:
Sorry, made a bit of a mess when rebasing :man_facepalming: 

Here is the clean version of this: https://github.com/nextflow-io/nextflow/pull/2590

PR #2430, opened 2021-11-04
Fixes bug 2428 

mes5k:
To fix this bug I add an additional check to checkInvalidWaitingState that checks if the pod phase is Failed and throw an exception if it is. A test validating this behavior has been included and I have confirmed that the test fails without this fix.

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/k8s/client/K8sClientTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/k8s/client/K8sClientTest.groovy
@@ -845,4 +845,121 @@ class K8sClientTest extends Specification {
         e.message == 'K8s pod configuration failed -- secrets "my-env" not found'
 
     }
+
+    def 'client should throw process exception when initContainer fails' () {
+        // Addresses bug 2428, which is when an initContainer exists and fails
+        // leaving the containerStatuses in "PodInitializing", but the pod
+        // phase is "Failed".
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/test/groovy/nextflow/k8s/client/K8sClientTest.groovy
```diff
@@ -845,4 +845,121 @@ class K8sClientTest extends Specification {
...
+                            "type": "ContainersReady"
+                        },
+                        {
+                            "lastProbeTime": null,
+                            "lastTransitionTime": "2021-11-03T21:37:45Z",
+                            "status": "True",
+                            "type": "PodScheduled"
+                        }
+                    ],
+                    "containerStatuses": [
```
pditommaso, line 905: Maybe the `conditions` can be removed to simplify the test sample

pditommaso (commented):
Thanks for reporting this patch! I've left a comment

modules/nextflow/src/test/groovy/nextflow/k8s/client/K8sClientTest.groovy
mes5k: Sure, just removed them.

pditommaso: Awesome. Thanks Mike!

Issue #1952, opened 2021-03-05
Nextflow 20.10 doesn't recognize AWS China credential?


shanzhong:
## Bug report 
I was trying to set up an analysis pipeline in Nextflow to run with AWS Batch in China (cn-northwest-1 region), starting with a simple test script. Using the same script and config file, the latest stable version (20.10) seems to not recognize the AWS credential and could not access S3, whereas an older version (19.10) runs successfully. Is it a bug, or is there any changes between the two versions that I'm not aware of?

EDIT: tested a few more versions, the script runs successfully till Nextflow 20.04.1, and started to break with 20.07

### Expected behavior and actual behavior

- Expected behavior: the output file (testout.txt) in test.nf should be published to the specified S3 location
- Actual behavior: 
  - Nextflow 20.10 fails to run the test script, giving an error message indicating that the AWS credential is not valid. 
  - Nextflow 19.10 works as expected

### Steps to reproduce the problem
test.nf:
```
#!/usr/bin/env nextflow
outpath = Channel.value(params.outdir)
process test_s3 {
    cpus 1
    memory '1 GB'
    publishDir "${params.outdir}", mode: "copy"
    output:
        path "testout.txt"
    script:
    """
    #!/bin/bash
    echo "hello world" > testout.txt
    """
}
```
nextflow.config:
```
aws {
    accessKey = 'XXXXXX'
    secretKey = 'XXXXXX'
    region = 'cn-northwest-1'
}
```
Running the test script:
```
nextflow run test.nf --outdir s3://<bucket>/<path>/
```

### Program output 
- Nextflow 20.10 output:
```
N E X T F L O W  ~  version 20.10.0
Launching `test.nf` [magical_lichterman] - revision: d7670ccb14
executor >  local (1)
[88/237e0f] process > test_s3 [  0%] 0 of 1
Error executing process > 'test_s3'

Caused by:
  The AWS Access Key Id you provided does not exist in our records. (Service: Amazon S3; Status Code: 403; Error Code: InvalidAccessKeyId; Request ID: XXXXXX; S3 Extended Request ID: XXXXXX)
```
- Nextflow 19.10 output:
```
N E X T F L O W  ~  version 19.10.0
Launching `test.nf` [infallible_magritte] - revision: d7670ccb14
executor >  local (1)
[ae/8ba130] process > test_s3 [100%] 1 of 1 ✔

```
- `.nextflow.log` file from 20.10:
```
[...]

PR #2433, opened 2021-11-05
Add note for aws china endpoint

abhi18av:
This PR adds the configuration for AWS China regions in the docs, a follow up from https://github.com/nextflow-io/nextflow/issues/1952#issuecomment-961276439

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -607,7 +607,7 @@ For example::
         }
     }
 
-.. tip:: To use an AWS China region, please specify the appropriate region code in ``endpoint``. For example ``aws.client.endpoint = "s3.cn-north-1.amazonaws.com.cn"``
+.. tip:: To use an AWS China region, please specify the appropriate region code in ``endpoint``. For example ``aws.client.endpoint = "https://s3.cn-north-1.amazonaws.com.cn"``
 
 .. _config-aws-batch:
 
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -607,6 +607,7 @@ For example::
         }
     }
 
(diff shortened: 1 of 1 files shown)


docs/config.rst
```diff
@@ -607,6 +607,7 @@ For example::
         }
     }
 
+.. tip:: To use an AWS China region, please specify the appropriate region code in ``endpoint``. For example ``aws.client.endpoint = "s3.cn-north-1.amazonaws.com.cn"``
```
pditommaso, line 610: Does not the endpoint start with `https://`?

abhi18av: On the comment which confirmed this setting, there's no `https://` - but yeah, I agree we should be more explicit on our end.

abhi18av:
```suggestion
.. tip:: To use an AWS China region, please specify the appropriate region code in ``endpoint``. For example ``aws.client.endpoint = "https://s3.cn-north-1.amazonaws.com.cn"``
```

pditommaso: Now that you are mentioning it, I'm not sure anymore. Actually, in the docs there are no protocol prefix. 

https://docs.aws.amazon.com/general/latest/gr/s3.html#s3_website_region_endpoints

This we should to a test to verify it

abhi18av: Regarding this, I was able to test access with `https`. Also, this is supported by 


- The [virtual hosting](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html) docs.

- The s3 endpoint docs

pditommaso:
Ok, I've moved to S3 docs page and merged manually 9c8110406. tx!

PR #2446, opened 2021-11-15
Correct description of cross

cjw85:
The property that `a.cross(b)` is not equal to `b.cross(a)` reflects non-commutativity of the operator not reflexivity.


pditommaso (approved):
You are right. Thanks

PR #2457, opened 2021-11-19
add java 16 and 17 to the launch.sh for debugging

mmatthews06:
If this is an error with my configuration, I can fix it, but I'm using Java 17 and this fixed the debug launcher for me


pditommaso:
Nice, thanks a lot

PR #2460, opened 2021-11-22
Allow to use environment variables at parameters

jordeu:
## Description

Current implementation only allows to use `baseDir`, `projectDir`, `launchDir` and `secrets` at the parameters file or from the command line. 

This pull request adds also the option to access the environment variable available at config scope. They can be referenced using the prefix `env.`,  like `${env.MY_ENV_VAR}`.

## Usage

Example on how to use it:
```
$ cat ../params.yml 
outdir: ${env.NXF_WORK}/results_workdir

$ ./launch.sh run nextflow-io/rnatoy -params-file ~/tmp/params.yml -e.NXF_WORK=/home/jordeu/tmp/nextflow -with-docker
```


pditommaso:
Let's close for now. We'll consider this when there's a concrete use case

PR #2462, opened 2021-11-22
Test with java-17 LTS

abhi18av:
Since Java 17 is the new LTS and is already supported by NF, I thought it might be the right to update the automated test template as well.


pditommaso (approved):
Make sense, thanks!

Issue #2282, opened 2021-08-27
Support docker container option --shm-size, per process


maxbates:
## New feature

Some programs which use large file-based databases, like `jackhmmer`, see a significant performance increase when the database can be loaded into memory using a `shm` (e.g. `/dev/shm`) or `tmpfs` volume. 

I would like to specify the size of a ramdisk, dependent on the memory allocated to the container. E.g. If I allocate 64Gb to the container, I would pass `--shm-size=64g`. 

The size needs to be dynamic, so it matches the memory available to the container. Mounting `/dev/shm` is problematic in this case, because I do not want processes to compete for the host machine's shared memory (i.e. when multiple containers are scheduled onto the same instance). 

on AWS, because nextflow creates job definitions and does not support the process directive `containerOptions`, I do not think it is possible to provision a dynamically sized shared memory volume without manually creating a job definition (which I would like to avoid).

By default, docker allocates 64Mb to `/dev/shm`, but can be configured using `--shm-size` (ref)[https://docs.docker.com/engine/reference/run/]. The size cannot be changed from within the container, as it requires remounting the volume.

AWS supports specifying `sharedMemorySize` in the [job definition](https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html#containerProperties), which simply passes through to docker's `--shm-size`

## Usage scenario 

The use of a ramdisk in alphafold's colab notebook for running `jackhmmer` can be see [here](https://github.com/deepmind/alphafold/blob/b1d772d127fcff4cc01d8fa1b4ea6e07da12193d/notebooks/AlphaFold.ipynb) (creating `/tmp/ramdisk`). There is a similar [recommendation on github](https://github.com/soedinglab/hh-suite/issues/281#issuecomment-888689484) for speeding up hhblits.

## Suggest implementation 

Add a process directory `shmSize` and update AWS Batch plugin's `newSubmitRequest(TaskRun task)` [ref](https://github.com/nextflow-io/nextflow/blob/master/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy#L557), and the local container executor.

pditommaso:
Think what could be done here is parsing the `containerOptions` and map selected option to the corresponding Batch API such as `--shm-size`, `--ulimit`, etc

maxbates:
That would absolutely work for me!

pditommaso:
Would this require also the use of `--tmpfs` Docker option

[...]

PR #2471, opened 2021-12-01
Set AWS container properties from container options

manuelesimi:
This contribution addresses the requirements from https://github.com/nextflow-io/nextflow/issues/2282 and https://github.com/nextflow-io/nextflow/pull/2413.

With this change, the container properties of a job definition registered by Nextflow in AWS Batch are set starting from the container options defined in the Nextflow task.

**Example**

Given the following container options:
```
process sayHelloFirst {

  containerOptions '--env MYVAR2=foo2 --env MYVAR3=foo3 --memory-swap 240000 --memory-swappiness 20 --shm-size 16000000 --ulimit nofile=1280:2560 --ulimit nproc=16:32 --privileged'

}
```
the job is registered with the following properties on AWS Batch: 

properties

Do note that:
1. this mapping does not include properties already set by other means in Nextflow (like volumes, mount points, cpus, memory, etc.)
2. not all the docker run options can be mapped into AWS.  See [AWS doc](https://docs.aws.amazon.com/batch/latest/APIReference/API_ContainerProperties.html).


manuelesimi:
I just pushed a fix for tmpfs mounts. They are now correctly set in the container properties:

tmpfs mounts

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AWSContainerOptionsMapper.groovy
```diff
@@ -0,0 +1,170 @@
...
+ * @see <a href="https://docs.docker.com/engine/reference/commandline/run/">Docker run</a>
+ * @see <a href="https://docs.aws.amazon.com/batch/latest/APIReference/API_ContainerProperties.html">API Container Properties</a>
+ * @author Manuele Simi <manuele.simi@gmail.com>
+ */
+class AWSContainerOptionsMapper {
+
+    def options
+
+    protected AWSContainerOptionsMapper(String containerOptions) {
+        options = containerOptions.trim().split("\\s+")
```
pditommaso, line 21: Too weak, the container option values could container blanks and quotes. Also I think the parsing should not be done by this component, which could take instead a Map object

pditommaso (commented):
Hi Manuale, this is welcome. Actually, I had gave a try some time ago but gave up when I realised container options should be managed at Job Definition level (what a design mistake ..) 

My main comment is that this is a good opportunity to manage the `containerOption` as a map. Look at my commit that I've just pushed 

https://github.com/nextflow-io/nextflow/commit/5e9c1cce8c7037be05a77f4c68f91966d7d62521

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AWSContainerOptionsMapper.groovy
manuelesimi: That makes sense. I also agree that the parsing should occur earlier. Let me think about it.

pditommaso: There's an implementation in the commit I've linked above

manuelesimi: I saw it :)

manuelesimi: I saw it, thanks :)

manuelesimi:
Great work (and fast!).
I noticed you pushed your change for the map options to another branch. How should I proceed? Merger yours in this one, wait until you merge yours into master and then merge master back into this, or else?

pditommaso:
I would say to merge my branch in your PR, then you can continue with that

manuelesimi:
@pditommaso There is a problem with your parser. The Map doesn't consider options that can be repeated multiple times (like --env or --tmpfs). Each option should be mapped to a List of values.

manuelesimi:
Also, '=' cannot be used as separator in the `QuoteStringTokenizer`. There are options that uses '=' in their values. E.g. `--env MY_VAR-value`.

pditommaso:
It's getting complicated. Should repeatable options be supported? my original plan was only to support a subset of them.

manuelesimi:
> It's getting complicated. Should repeatable options be supported? my original plan was only to support a subset of them.

Parsing command line options is never easy.

I made a change that seems to work:
From `'--env VAR_FOO -e VAR_FOO2=value2 --env VAR_FOO3=value3 --user nf-user'`, I get:

list

Would that be OK with you?

What's probably more concerning is that we should support only the format "option value" (which is the one documented in the docker run help), not "option=value". So a value with an equal char is not separated (like in the screenshot above).

pditommaso:
Is basically a `Map<String,List>` ?

manuelesimi:
> Is basically a `Map<String,List>` ?

yes, it is.

pditommaso:
Maybe we should model it as a `List` of pairs, or even a custom holder class

manuelesimi:
I'd go for a custom holder where we can add the lookup methods we will need.

pditommaso:
👍

manuelesimi:
Question:

The method `Map getContainerOptionsMap()` in TaskConfig returns a Map. Hence, we have the following options:

1) we change the method signature to return our custom holder class. As far as I see, it is used only [here](https://github.com/nextflow-io/nextflow/blob/6adfce099ecf3c835e03d1a2eb98ea81d0540019/plugins/nf-google/src/main/nextflow/cloud/google/lifesciences/GoogleLifeSciencesTaskHandler.groovy#L349) so it would have a low impact.
2) our custom holder extends a Map (but then we need to cast back to the holder's class to access its methods, which I'd really dislike)
3) we stick to a Map<String,List>

Which one would you favor?

pditommaso:
I would say 1.

manuelesimi:
Also a combination of 1. and 2. would work well.

manuelesimi:
> I would say 1.

Sounds good.

pditommaso:
As you wish 😄

manuelesimi:
@pditommaso I switched to a custom handler for the option map. I also added support for long option with dashes (docker has a few we have to support like `-memory-swap ... --shm-size ...`.
I adapted the AWS container mapper accordingly, which is greatly simplified with respect to the previous version.

See what you think.

PS: I'm marking this PR as draft since I need additional tests on AWS to make sure all is working as expected after these changes.

manuelesimi:
I have tested again the submissions to AWS Batch with the latest changes pushed in this PR. I can confirm that all the container properties are set correctly (see the screenshots from previous comments) according to the options specified in `containerOptions`.

As I mentioned, not all the `docker run` options are supported in the AWS job definition. If this PR is approved, should we also update the documentation?

modules/nf-commons/src/main/nextflow/util/CmdLineOptionMap.groovy
```diff
@@ -0,0 +1,63 @@
+package nextflow.util
+
+/**
+ * Holder for parsed command line options.
+ *
+ * @author Manuele Simi <manuele.simi@gmail.com>
+ */
+class CmdLineOptionMap {
```
pditommaso, line 8: Whenever possible new classes should be annotated with `@CompileStatic`

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsContainerOptionsMapper.groovy
```diff
@@ -0,0 +1,149 @@
...
+import nextflow.util.CmdLineOptionMap
+
+/**
+ * Maps task container options to AWS container properties
+ *
+ * @see <a href="https://docs.docker.com/engine/reference/commandline/run/">Docker run</a>
+ * @see <a href="https://docs.aws.amazon.com/batch/latest/APIReference/API_ContainerProperties.html">API Container Properties</a>
+ * @author Manuele Simi <manuele.simi@gmail.com>
+ */
+class AwsContainerOptionsMapper {
```
pditommaso, line 17: Let's add `@CompileStatic`

modules/nf-commons/src/main/nextflow/util/CmdLineOptionMap.groovy
```diff
@@ -0,0 +1,63 @@
...
+
+    boolean hasMultipleValues(String key) {
+        options.containsKey(key) ? options[key].size() > 1 : false
+    }
+
+    boolean hasOptions() {
+        options.size()
+    }
+
+    def getValues(String key) {
```
pditommaso, line 27: Does this always return a `List`? in this case worth declaring `List getValues(String key)`

modules/nf-commons/src/main/nextflow/util/CmdLineOptionMap.groovy
pditommaso: It's missing `CmdLineOptionMapTest` or I'm wrong?

pditommaso (changes requested):
It looks a nice work! Left a few comments

pditommaso:
> should we also update the documentation?

it would be nice!

manuelesimi:
> > should we also update the documentation?
> 
> it would be nice!

See https://github.com/nextflow-io/nextflow/pull/2471/commits/75c5623c0122c70a6ba431d0cbb140b19cce0102

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsContainerOptionsMapper.groovy
manuelesimi: Done

modules/nf-commons/src/main/nextflow/util/CmdLineOptionMap.groovy
manuelesimi: Indeed, it was missing. Just added.

modules/nf-commons/src/main/nextflow/util/CmdLineOptionMap.groovy
manuelesimi: done

manuelesimi:
> It looks a nice work! Left a few comments

Hi @pditommaso, thanks for your additional comments. Indeed they helped to improve. 

After the changes, I've tested again with AWS Batch and the job properties are diligently matched with the container options :)

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsContainerOptionsMapper.groovy
```diff
@@ -101,25 +104,33 @@ class AwsContainerOptionsMapper {
...
         // tune container memory swappiness
         value = findOptionWithSingleValue('memory-swappiness')
-        if ( value && value.length() > 0 )
+        if ( value && value.length() > 0 ) {
             params.setSwappiness(value as Integer)
+            atLeastOneSet = true
+        }
 
-        containerProperties.setLinuxParameters(params)
+        if ( atLeastOneSet )
```
pditommaso, line 132: Is there a side-effect setting an empty `LinuxParameters` object? 

Also, it may be better to have more functional behaviour having `checkLinuxParameters` returning the `LinuxParameters`. 

Finally `checkLinuxParameters` could be a static method to prevent creating a class for each task.

manuelesimi: > Is there a side-effect setting an empty `LinuxParameters` object?
> 

It's null by default. I wanted to keep it that way if there are no parameters set. This way we have the same (super-tested) behavior previous to this PR. If no param is set the object is null.

> Also, it may be better to have more functional behaviour having `checkLinuxParameters` returning the `LinuxParameters`.

It makes sense.
> 
> Finally `checkLinuxParameters` could be a static method to prevent creating a class for each task.

Makes sense as well, I'll change it (eventually, it'd be an object for each task :) )

manuelesimi: I made the entire AWSContainerMapper class static to avoid creating unnecessary objects for each task.

modules/nf-commons/src/main/nextflow/util/CmdLineHelper.groovy
```diff
@@ -66,12 +75,12 @@ class CmdLineHelper {
         if( val instanceof String ) {
             val = [val]
         }
-
-        for( int i=0; i<val.size(); i++ ) {
-            val[i] = val[i] ?. split(splitter)
+        List<String> vals = val as List<String>
+        for( int i=0; i<vals.size(); i++ ) {
+            vals[i] = vals[i] ?. split(splitter) as String
         }
```
pditommaso, line 82: Think this can be removed.

modules/nf-commons/src/main/nextflow/util/CmdLineOptionMap.groovy
```diff
@@ -0,0 +1,82 @@
...
+        options.each {
+            serialized << "option{${it.key}: ${it.value.each {it}}}"
+        }
+        return "[${serialized.join(', ')}]"
+    }
+
+    @Override
+    Hasher funnel(Hasher hasher, CacheHelper.HashMode mode) {
+        return CacheHelper.hasher(hasher, options, mode)
+    }
```
pditommaso, line 81: I've added the logic to make it hasheable

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsContainerOptionsMapper.groovy
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * @see https://docs.docker.com/engine/reference/commandline/run/
+ * @see https://docs.aws.amazon.com/batch/latest/APIReference/API_ContainerProperties.html
+ *
+ * @author Manuele Simi <manuele.simi@gmail.com>
+ */
+@CompileStatic
+class AwsContainerOptionsMapper {
+
+    static ContainerProperties createContainerOpts(CmdLineOptionMap options) {
```
pditommaso, line 22: Made it return the `ContainerProperties` object

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsContainerOptionsMapper.groovy
```diff
@@ -0,0 +1,159 @@
...
+        // tune container memory swappiness
+        value = findOptionWithSingleValue(options, 'memory-swappiness')
+        if ( value && value.length() > 0 ) {
+            params.setSwappiness(value as Integer)
+            atLeastOneSet = true
+        }
+
+        return atLeastOneSet ? params : null
+    }
+
```
pditommaso, line 132: It should be added a unit test each of the option

plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsBatchTaskHandlerTest.groovy
```diff
@@ -463,7 +463,9 @@ class AwsBatchTaskHandlerTest extends Specification {
         given:
         def IMAGE = 'foo/bar:1.0'
         def JOB_NAME = 'nf-foo-bar-1-0'
-        def handler = Spy(AwsBatchTaskHandler)
+        def handler = Spy(AwsBatchTaskHandler) {
+            getTask() >> Mock(TaskRun) { getConfig() >> Mock(TaskConfig)  }
+        }
         handler.executor = Mock(AwsBatchExecutor)
```
pditommaso, line 469: A few tests were failing

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
```diff
@@ -485,7 +492,10 @@ class AwsBatchTaskHandler extends TaskHandler implements BatchHandler<String,Job
 
         // add to this list all values that has to contribute to the
         // job definition unique name creation
-        hashingTokens.addAll([name, image, awscli, volumes, jobRole])
+        final tokens = [name, image, awscli, volumes, jobRole]
+        if( containerOpts )
+            tokens.add(containerOpts)
+        hashingTokens.addAll(tokens)
```
pditommaso, line 498: Add the `containerOptions` to the hashing key to force a new Job definition when container a value change

pditommaso:
Just realised that `AwsContainerOptionsMapper` was tested via indirectly the `TaskHandler`. Think it would be better to simplify this testing directly the class ie. 

```
 when:
 def map = CmdLineHelper.parseGnuArgs('--ulimit nofile=1280:2560 --ulimit nproc=16:32')
 def properties = AwsContainerOptionsMapper.createContainerOpts(map) 
 then:
 properties.getUlimits().size() == 2
 properties.getUlimits().get(0).toString() == '{HardLimit: 2560,Name: nofile,SoftLimit: 1280}'
 properties.getUlimits().get(1).toString() == '{HardLimit: 32,Name: nproc,SoftLimit: 16}'
```


https://github.com/nextflow-io/nextflow/blob/80dbfa8e64e42b49acf5b89040b8d9e180565c28/plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsContainerOptionsMapperTest.groovy#L12-L156

modules/nf-commons/src/main/nextflow/util/CmdLineOptionMap.groovy
manuelesimi: yep, definitely needed and welcomed.

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsContainerOptionsMapper.groovy
manuelesimi: Looks more readable. Thanks.

plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsBatchTaskHandlerTest.groovy
manuelesimi: thanks!

modules/nf-commons/src/main/nextflow/util/CmdLineHelper.groovy
manuelesimi: > Just realised that `AwsContainerOptionsMapper` was tested via indirectly the `TaskHandler`. Think it would be better to simplify this testing directly the class ie.
> 
> ```
>  when:
>  def map = CmdLineHelper.parseGnuArgs('--ulimit nofile=1280:2560 --ulimit nproc=16:32')
>  def properties = AwsContainerOptionsMapper.createContainerOpts(map) 
>  then:
>  properties.getUlimits().size() == 2
>  properties.getUlimits().get(0).toString() == '{HardLimit: 2560,Name: nofile,SoftLimit: 1280}'
>  properties.getUlimits().get(1).toString() == '{HardLimit: 32,Name: nproc,SoftLimit: 16}'
> ```
> 
> https://github.com/nextflow-io/nextflow/blob/80dbfa8e64e42b49acf5b89040b8d9e180565c28/plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsContainerOptionsMapperTest.groovy#L12-L156

It is also tested there. When you asked for a test, I thought you wanted a dedicated test class. I agree and will move the tests  as you suggested.

manuelesimi:
> Just realised that `AwsContainerOptionsMapper` was tested via indirectly the `TaskHandler`. Think it would be better to simplify this testing directly the class ie.
> 
> ```
>  when:
>  def map = CmdLineHelper.parseGnuArgs('--ulimit nofile=1280:2560 --ulimit nproc=16:32')
>  def properties = AwsContainerOptionsMapper.createContainerOpts(map) 
>  then:
>  properties.getUlimits().size() == 2
>  properties.getUlimits().get(0).toString() == '{HardLimit: 2560,Name: nofile,SoftLimit: 1280}'
>  properties.getUlimits().get(1).toString() == '{HardLimit: 32,Name: nproc,SoftLimit: 16}'
> ```
> 
> https://github.com/nextflow-io/nextflow/blob/80dbfa8e64e42b49acf5b89040b8d9e180565c28/plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsContainerOptionsMapperTest.groovy#L12-L156

It is also tested there. When you asked for a test, I thought you wanted a dedicated test class. I agree and will move the tests as you suggested.

manuelesimi:
> > Just realised that `AwsContainerOptionsMapper` was tested via indirectly the `TaskHandler`. Think it would be better to simplify this testing directly the class ie.
> > ```
> >  when:
> >  def map = CmdLineHelper.parseGnuArgs('--ulimit nofile=1280:2560 --ulimit nproc=16:32')
> >  def properties = AwsContainerOptionsMapper.createContainerOpts(map) 
> >  then:
> >  properties.getUlimits().size() == 2
> >  properties.getUlimits().get(0).toString() == '{HardLimit: 2560,Name: nofile,SoftLimit: 1280}'
> >  properties.getUlimits().get(1).toString() == '{HardLimit: 32,Name: nproc,SoftLimit: 16}'
> > ```
> > 
> 
> It is also tested there. When you asked for a test, I thought you wanted a dedicated test class. I agree and will adjust the tests as you suggested.

done, see https://github.com/nextflow-io/nextflow/pull/2471/commits/d1d9da011fb3e422b08ebc9b3352768f25605984

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsContainerOptionsMapper.groovy
manuelesimi: done, see https://github.com/nextflow-io/nextflow/pull/2471/commits/d1d9da011fb3e422b08ebc9b3352768f25605984

modules/nf-commons/src/main/nextflow/util/CmdLineHelper.groovy
manuelesimi: done, see https://github.com/nextflow-io/nextflow/pull/2471/commits/a724067fe5c8d34ee7685bdb215b212ce7446b83

manuelesimi:
I think I've addressed your latest comments. Pls, let me know if I missed something. Thanks!

manuelesimi:
Hold on, there is a test failing. I know why. I can fix it later.

manuelesimi:
So, I changed two `GoogleLifeSciencesTaskHandlerTest` tests that uses the new config map. I hope I didn't misinterpret them, but basically I'd expect that when the entrypoint is not set in the map, the default entrypoint is returned. See https://github.com/nextflow-io/nextflow/pull/2471/commits/96c3fd4e79ddfd028380fc0b01aef3494450c59e

plugins/nf-google/src/test/nextflow/cloud/google/lifesciences/GoogleLifeSciencesTaskHandlerTest.groovy
```diff
@@ -197,7 +197,7 @@ class GoogleLifeSciencesTaskHandlerTest extends GoogleSpecification {
         then:
         task.getConfig() >> new TaskConfig(containerOptions: [entrypoint:null])
         and:
-        req.entryPoint == null
+        req.entryPoint == GoogleLifeSciencesConfig.DEFAULT_ENTRY_POINT
```
pditommaso, line 201: Interesting, why before it was null? 😕

pditommaso: So, the constant is `/bin/bash`

https://github.com/nextflow-io/nextflow/blob/96c3fd4e79ddfd028380fc0b01aef3494450c59e/plugins/nf-google/src/main/nextflow/cloud/google/lifesciences/GoogleLifeSciencesConfig.groovy#L48-L48

However when is null defaults to `bash` 

https://github.com/nextflow-io/nextflow/blob/96c3fd4e79ddfd028380fc0b01aef3494450c59e/plugins/nf-google/src/main/nextflow/cloud/google/lifesciences/GoogleLifeSciencesHelper.groovy#L245-L246

Think, it would be better to stick to the current behaviour to avoid introducing unexpected side effects. I'll do the chanege

pditommaso: Even more intriguing.. actually when `entrypoint` was not specified it was defaulting to `/bin/bash` 

https://github.com/nextflow-io/nextflow/blob/96c3fd4e79ddfd028380fc0b01aef3494450c59e/plugins/nf-google/src/test/nextflow/cloud/google/lifesciences/GoogleLifeSciencesTaskHandlerTest.groovy#L185-L185

Not clear why there was that test checking for `null`

pditommaso: I've removed that tests 😆

modules/nf-commons/src/main/nextflow/util/CmdLineOptionMap.groovy
```diff
@@ -39,7 +39,7 @@ class CmdLineOptionMap implements CacheFunnel {
     }
 
     def getFirstValue(String key) {
-        getFirstValueOrDefault(key, '')
+        getFirstValueOrDefault(key, null)
```
pditommaso: Usually, it's used `null` for absences of value

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsContainerOptionsMapper.groovy
```diff
@@ -47,15 +47,15 @@ class AwsContainerOptionsMapper {
             final tokens = value.tokenize('=')
             keyValuePairs << new KeyValuePair().withName(tokens[0]).withValue(tokens.size() == 2 ? tokens[1] : null)
         }
-        if ( keyValuePairs.size() > 0 )
```
pditommaso: This is implicit with groovy, have a look at https://groovy-lang.org/semantics.html#Groovy-Truth

manuelesimi: Yep, sorry, that comes from my Java background :)

modules/nf-commons/src/main/nextflow/util/CmdLineOptionMap.groovy
manuelesimi: Got it. Thanks.

pditommaso (approved):
OK, Sunday PR approval! thanks for this, much appreciated!

manuelesimi:
Terrific! This was much needed for our AWS projects.
Thanks for all the help and patience along .

pditommaso:
Released along with version 21.12.1-edge

PR #2489, opened 2021-12-13
Group tuple size

Lehmann-Fabian:
Today I spent some time finding out how to group tuples when the group sizes vary.
Currently, I have to do it this way.
```
Channel
     .from( [1,'A'], [1,'B'], [2,'C'], [3, 'B'], [1,'C'], [2, 'A'], [3, 'D'] )
     .map { key, value -> [ groupKey( key, key == 1 ? 3 : 2 ), value ] }
     .groupTuple()
     .view()
```
Doing it this way, requires an additional mapping, moreover it gets complex if more, and non-consecutive indexes are used.
I propose to do it the following way. I implemented it as an alternative, keeping the other method valid.
```
Channel
     .from( [1,'A'], [1,'B'], [2,'C'], [3, 'B'], [1,'C'], [2, 'A'], [3, 'D'] )
     .groupTuple( size : { it == 1 ? 3 : 2} )
     .view()
```


Lehmann-Fabian:
Hi @pditommaso, can you please review this PR.

bentsherman:
Note: I will probably do something like this for the v2 operators:

```groovy
channel.of( [1,'A'], [1,'B'], [2,'C'], [3, 'B'], [1,'C'], [2, 'A'], [3, 'D'] )
    .groupBy { k, v -> tuple(k, k == 1 ? 3 : 2, v) }
    .view()
```

Basically, you could specify a closure that returns a 3-tuple of (key, size, value) for each incoming item. This should provide maximum flexibility

PR #2495, opened 2021-12-15
Nextflow couldn't parse generated nextflow.config under some circumstances.

Lehmann-Fabian:
Running some nf-core workflows on Kubernetes is currently not possible.
I tested viralrecon and eager. Both fail with errors parsing the generated nextflow.config file on Kubernetes if keys in the config are quoted.

For example:
```
params {
    modules {
       ...
        'sra_fastq_ftp' {
            args            = '-C - --max-time 1200'
            publish_dir     = 'public_data'
            publish_files   = ['fastq.gz':'', 'md5':'md5']
        }
...
```
becomes
```
params {
    ...
    modules {
        ...
        sra_fastq_ftp {
            args = '-C - --max-time 1200'
            publish_dir = 'public_data'
            publish_files {
                'fastq.gz' = ''
                md5 = 'md5'
            }
        }
...
```
when I run:
`nextflow kuberun https://github.com/nf-core/viralrecon -v pvc-host-shared-data:/pvcdata -v project-pvc-host-shared-data:/project -config nextflow.config -profile test,kubernetesConf`
it throws
```
Unable to parse config file: '/pvcdata/fabian/nextflow.config'

  Compile failed for sources FixedSetSources[name='/groovy/script/ScriptBEFEBE235A7D5B7ACE613A834576E6FA/_nf_config_ec39096b']. Cause: org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
  /groovy/script/ScriptBEFEBE235A7D5B7ACE613A834576E6FA/_nf_config_ec39096b: 161: The LHS of an assignment should be a variable or a field accessing expression @ line 161, column 13.
                 'fastq.gz' = ''
                 ^
  
  1 error
```
With my adjustments, the generated file looks like this:
```
params = [
   modules : [
      ...
      sra_fastq_ftp : [
         args : '-C - --max-time 1200',
         publish_dir : 'public_data',
         publish_files : [
            'fastq.gz' : '',
            md5 : 'md5',
         ],
      ],
```
This is successfully parsed.

Moreover, I found the following test case:
https://github.com/Lehmann-Fabian/nextflow/blob/parseFailure/modules/nextflow/src/test/groovy/nextflow/util/ConfigHelperTest.groovy#L171-L191
While the test passes, the String it checks for wouldn't be parsable. Maybe I should delete the test?

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/util/ConfigHelper.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/util/ConfigHelper.groovy
@@ -141,10 +141,13 @@ class ConfigHelper {
 
                     writer.append(TAB*level)
                     writer.append(wrap1(key))
-                    writer.append(' {\n')
+                    writer.append( level == 0 ? ' = ' : ' : ' )
+                    writer.append( '[\n' )
                     canonicalFormat(writer, value, level+1,sort,stack)
                     writer.append(TAB*level)
-                    writer.append('}\n')
+                    writer.append(']')
+                    if ( level > 0 ) writer.append(',')
+                    writer.append( '\n' )
                 }
                 else {
                     // add a new-line to separate simple values from a previous config object
@@ -154,8 +157,9 @@ class ConfigHelper {
 
                     writer.append(TAB*level)
                     writer.append(wrap1(key))
-                    writer.append(' = ')
+                    writer.append( level == 0 ? ' = '  : ' : ')
                     writer.append( render0(value) )
+                    if ( level > 0 ) writer.append(',')
                     writer.append('\n')
                 }
             }
@@ -167,15 +171,7 @@ class ConfigHelper {
 
     static @PackageScope String wrap1(param) {
         final key = param.toString()
-        if( key.startsWith('withLabel:') )  {
-            return 'withLabel:' + wrap0(key.substring('withLabel:'.length()))
-        }
-        else if( key.startsWith('withName:') )  {
-            return 'withName:' + wrap0(key.substring('withName:'.length()))
-        }
-        else {
-            return wrap0(key)
-        }
+        return wrap0(key)
     }
 
     static @PackageScope String wrap0( param ) {
--- a/modules/nextflow/src/test/groovy/nextflow/cli/CmdConfigTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/cli/CmdConfigTest.groovy
@@ -106,23 +106,23 @@ class CmdConfigTest extends Specification {
         cmd.printCanonical0(config, buffer)
         then:
         buffer.toString() == '''
-                    process {
-                       executor = 'slurm'
-                       queue = 'long'
-                    }
+                    process = [
+                       executor : 'slurm',
+                       queue : 'long',
+                    ]
                     
-                    docker {
-                       enabled = true
-                    }
+                    docker = [
+                       enabled : true,
+                    ]
 
-                    mail {
-                       from = 'yo@mail.com'
-                       smtp {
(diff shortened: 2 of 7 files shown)


pditommaso (commented):
Not sure to understand the need for this. It looks declarative arrays are handled as expected already 

https://github.com/nextflow-io/nextflow/pull/2495/files#diff-ab669f4b08b48945b5383c0c48ddf718af307e7c8aaf21a543086697b8948a46L249-L257

Lehmann-Fabian:
My first idea was only to have maps containing keys with special characters as a declarative array. But technically, the syntax is equal, isn't it? So I decided to have it consistently. 
But I could also implement it this way.

pditommaso:
Still not understanding, it looks the formatter only converts `ConfigObject`

https://github.com/nextflow-io/nextflow/blob/a0bff10a2feae8924fbfc347779d5fd47fefc319/modules/nextflow/src/main/groovy/nextflow/util/ConfigHelper.groovy#L136-L136

why your hash map is converted to object notation? I think it should be isolated this behaviour  first

Lehmann-Fabian:
Honestly, I do not fully understand the parsing process in Nextflow.
But I didn't change anything in this regard. The Map 
```
publish_files   = ['fastq.gz':'', 'md5':'md5']
```
is parsed into a ConfigObject why it is then written as:  
```
publish_files {
                'fastq.gz' = ''
                md5 = 'md5'
            }
``` 

You can confirm this, running nf-core/viralcon or eager on K8s.
These are not the only problems I found in the parsing process, but it was the only one I could solve, and they seem not related.
The second problem is that methods defined in a config file get lost at all 
https://github.com/nf-core/viralrecon/blob/master/nextflow.config#L241-L270
and third: For nf-core/eager, the closures are written as:
```
'withLabel:sc_tiny' : [
      cpus : Script5AE9394664207A7B8EC50041BF1DF207$_run_closure1$_closure2$_closure10@638afcaa,
      memory : Script5AE9394664207A7B8EC50041BF1DF207$_run_closure1$_closure2$_closure11@58601e7a,
      time : Script5AE9394664207A7B8EC50041BF1DF207$_run_closure1$_closure2$_closure12@62735b13,
   ],
```

As you authored the code, you may have an idea. Perhaps the problems are related after all.

pditommaso:
It looks the problem that for some reason the map object turns into a `ConfigObject`. However it should not happen, this is captured by this test 

https://github.com/nextflow-io/nextflow/blob/a0bff10a2feae8924fbfc347779d5fd47fefc319/modules/nextflow/src/test/groovy/nextflow/cli/CmdConfigTest.groovy#L248-L257

Lehmann-Fabian:
Maybe the configuration varies if I read from GitHub / use kuberun?

Lehmann-Fabian:
I can confirm:
`nextflow config ~/.nextflow/assets/nf-core/viralrecon/nextflow.config`
and
`nextflow config ~/.nextflow/assets/nf-core/eager/nextflow.config`
resolves both maps right and closures are also printed with body,
So it seems a configuration issue with kuberun.

pditommaso:
I see, `kuberun` is a kind of experiment, it only makes sense to a quick start with K8s deployment. But it's discouraged the use for real use cases

Lehmann-Fabian:
In FONDA, around ten people frequently use `kuberun`, which works fine for our purposes. In case of problems, I'll fix or report them. So maybe we can bring the K8s extension of Nextflow to a new level.
I fixed the problem with your valuable hints, which motivated this PR here: https://github.com/nextflow-io/nextflow/pull/2496. Accordingly, I will close this PR.

PR #2496, opened 2021-12-16
Nextflow couldn't parse generated nextflow.config under some circumstances (root cause fix)

Lehmann-Fabian:
I found the root cause for the problem mentioned yesterday here: https://github.com/nextflow-io/nextflow/pull/2495
Now, K8s generates the config as all other Nextflow modules.


modules/nextflow/src/main/groovy/nextflow/util/ConfigHelper.groovy
```diff
@@ -263,8 +263,9 @@ class ConfigHelper {
         int i=0
         for( Map.Entry entry : map.entrySet() ) {
             if( i++>0 ) result.append(', ')
+            result.append('\'')
             result.append(entry.key)
-            result.append(":")
+            result.append("':")
             result.append(render0(entry.value))
```
pditommaso, line 269: Is this really needed? Quotes should only be needed when using keys that are not valid identifiers e.g. `some-value`, `0zero`, 'with blank`

modules/nextflow/src/main/groovy/nextflow/util/ConfigHelper.groovy
```diff
@@ -277,10 +278,6 @@ class ConfigHelper {
         result.toString()
     }
 
-    static String toCanonicalString(Map map, boolean sort=false) {
-        toCanonicalString(map.toConfigObject(), sort)
-    }
-
     static String toPropertiesString(ConfigObject config, boolean sort=false) {
```
pditommaso, line 281: Think should stay

pditommaso (changes requested):

modules/nextflow/src/main/groovy/nextflow/util/ConfigHelper.groovy
Lehmann-Fabian: This is technically not needed, but it makes the format consistent. I'll use `warp1` instead.

modules/nextflow/src/main/groovy/nextflow/util/ConfigHelper.groovy
Lehmann-Fabian: The method is never used. However, I will undo the deletion but add a warning. Should I mark the method as deprecated?

modules/nextflow/src/main/groovy/nextflow/util/ConfigHelper.groovy
pditommaso: Nice

modules/nextflow/src/main/groovy/nextflow/util/ConfigHelper.groovy
pditommaso: Why not used? I see used in several places https://github.com/nextflow-io/nextflow/pull/2496/files#diff-bbf841c8f39ba480213c8a171b78929cc4e68a38718b4386fa458f8012973705R327

Lehmann-Fabian: Here `config `is a `ConfigObject`, which is processed by this method: https://github.com/nextflow-io/nextflow/blob/67435473b9a5555edc208332bbb28fd8deb3a2ca/modules/nextflow/src/main/groovy/nextflow/util/ConfigHelper.groovy#L274-L278
The method with signature `toCanonicalString(Map map, boolean sort=false)` is never used.

pditommaso: I see. In any case, it may be useful in the future. No need to deprecate it. 👍

modules/nextflow/src/main/groovy/nextflow/k8s/K8sDriverLauncher.groovy
```diff
@@ -117,7 +122,8 @@ class K8sDriverLauncher {
         this.interactive = name == 'login'
         if( background && interactive )
             throw new AbortOperationException("Option -bg conflicts with interactive mode")
-        this.config = makeConfig(pipelineName)
+        this.configObject = makeConfig(pipelineName)
+        this.config = configObject.toMap()
         this.k8sConfig = makeK8sConfig(config)
```
pditommaso, line 127: Bit ugly to keep both `config` and `configObject`. Think this point makes sense to get rid of the first. 

The main difference with the when navigating a `ConfigObject` it automatically creates the request attribute, so for example if you do `configObject.foo`  foo becomes a nested empty object. It should be used `configObject.isSet('foo')` instead

pditommaso (approved):
Nice, thanks for this PR

Issue #785, opened 2018-07-09
Scheduling using K8S 


tdudgeon:
I create this issue to describe some of issues relating to scheduling of NF jobs on K8S. It's really a follow on task to problems 3 and 4 described in #773.

The basic scenario is that a user submits a CPU intensive pipeline to K8S and wants the job to run effectively across all the available nodes. Other jobs might already be running on the nodes. Neither the capacity of the K8S cluster nor the current load will be known to NF when the job is submitted. Ideally NF just submits the job, leaves the scheduling to K8S according to the available CPU and memory and sits back and waits for the job to complete.

I see 3 classes or NF process (considering only the CPU needs):
1. trivial tasks that use minimal amounts of CPU. These can be scheduled anywhere with minimal impact.
2. single threaded tasks that use all of a single core but no more. Ideally you would have as many of these running on each nodes as there are cores on the node. 
3. multi threaded tasks that utilise all the cores on a node. Ideally only one of these jobs would run on each node.

The first class is pretty trivial. They can be submitted without resource limits, or with low resource limits and will only minimally impact other processes.

Considering the second class, K8S has a pretty good way of handling these by specifying resource requests and limits. I've found that if you specify a request of ~0.9 CPU and a limit of ~1.1 CPU the tasks get distributed fairly effectively across the available nodes. With K8S the best approach to this is to use a Job that allows the level of parelellism to be controlled (and adjusted during execution), but the behaviour is much the same when using raw pods. What happens is that pods get started up until there is no CPU capacity left at which points the pods stick in Pending state until CPU capacity becomes available as other pods complete.
Note: this seems to be in contrast to what happens when NF submits pods with CPU limits where submitting the pods causes an error as it couldn't be scheduled, which causes the NF pipeline to throw an error.
In the situation where a previous job is fully utilising the cluster all Pods will presumably be in 'Pending' state until there is capacity. If there is a need to prioritise one type of job over another (e.g. a job from a 'gold' user compared to a 'bronze' user this could probably be handled using labels to distinguish the types. the 'gold' pods will then start in preference to the 'bronze' ones when capacity become available. 

Handling the third class seems more tricky. There does not seem to be a way to specify a CPU resource request of 'all cpus' - only numeric amounts (including fractional amounts) are allowed, and you do not know whether you are running on a machine with 1 core or 72 cores. 
[...]

Issue #773, opened 2018-06-27
Resource limits with K8S


tdudgeon:
The way resource limits are set with the K8S executor has some problems.
It looks like this needs to be set using the cpu directive in process definition. e.g.
```
process my_process {
    cpus 2

    ...
}
```
The problems I encountered are:

Problem 1: specifying the value '1' did not result in anything being created in the pod definition. Maybe 1 is considered a default and handled as a special case, but this is not correct for K8S as a CPU limit of 1 is meaningful and different from there being no limit.

Problem 2: The K8S spec handles this value as a string, allowing values like '100m' (meaning 100 milli-cpus, or 0.1 cpus). NF does not seem to allow specifying this as a text value (presumably this might cause problems with other executors). Specifying a fractional value (e.g. 0.5) seems to be silently ignored.

Problem 3: The 'cpu' value gets incorporated into the pod definition as a resource limit e.g. as
```
resources:
      limits: {cpu: 2}
```
K8S also has a limit for `request`s and probably that is more appropriate that a `limit`, but it might be best if both could be specified independently.

Problem 4: if specifying a CPU limit you can easily get into the situation that NF tries to start a number of pods but not all of them can be scheduled as there are no more CPUs available. When this happens NF throws an error and the pipeline fails. This is a complex topic and should probably be handled as a separate issue, but at the least I think the pod's `restartPolicy` should be set to `OnFailure` rather than `Never`, but how NF handles errors from process could complicate this.

pditommaso:
Many things: 

Problem 1) I think you are right, by default is a process uses 1 cpus, therefore it should be reflect in the resources specification otherwise a pod can consume all available cpus. 

Problem 2) this could be managed in the `cpus` directive but then how the actual command/tool would use that value? I mean the `cpus` is reflected in the `task.cpus` property that is generally used in the command script to set the number of threads/cpus that a multithread program should use. See [for example here](https://github.com/nextflow-io/rnaseq-nf/blob/master/main.nf#L88). Would not make sense to provide there a decimal value. 

Problem 3) This has been done on purpose. When `requests` is omitted it is implicitly equals to the `limits` value. Therefore k8s *guarantees* such resources. I think this makes sense for computational jobs.

Problem 4) Yes, maybe this should be managed as separate issue.

tdudgeon:
Problem 1). So can this be corrected so that a value of '1' does get included?

[...]

PR #2516, opened 2021-12-22
Add support for CPU millis for cpus directive

pditommaso:
This commit adds the ability to CPU millis as unit for the `cpus` directive. 

Target execution platforms should be able to take advantage of this feature, otherwise, it falls back to integer units representing the number of cores.

More in details:
- when providing an integer value for the `cpus` directive nothing change
- when providing a decimal number eg `1.5` platform supporting CPUs millis
  takes the 1500m value; platforms not supporting it, the millis value is
  rounded to the nearest upper value ie `2`
- when providing an integer number ending with `m` the value is intepreted
  as CPUs millis, therefore platform supporting it takes the value `1500m`;
  if the legacy platform takes the nearest upper value ie `2`.


modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
```diff
@@ -281,7 +281,82 @@ class TaskConfig extends LazyMap implements Cloneable {
...
+        // turn millis into an integer unit
+        final str = value.toString()
+        if( str.endsWith('m') ) {
+            try {
+                return Math.ceil((double)str.substring(0,str.length()-1).toInteger() / 1000)
+            }
+            catch (NumberFormatException e) {
+                throw new IllegalArgumentException("Not a valid cpus value: '$value'", e)
+            }
+         // turn float to unit
```
bentsherman, line 308: comment should be on next line?

modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
```diff
@@ -281,7 +281,82 @@ class TaskConfig extends LazyMap implements Cloneable {
...
+        catch (NumberFormatException e) {
+            throw new IllegalArgumentException("Not a valid cpus value: '$value'", e)
+        }
+    }
+
+    static protected int parseCpuMillis(value) {
+        if( value instanceof Integer )
+            return value.toInteger() * 1000
+        if( value instanceof Number )
+            return Math.ceil(value.toDouble() * 1000)
```
bentsherman, line 331: is `Math.ceil` neeeded in this function if we're just trying to get the millis?

modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
```diff
@@ -281,7 +281,82 @@ class TaskConfig extends LazyMap implements Cloneable {
...
+                return str.substring(0,str.length()-1).toInteger()
+            }
+            catch (NumberFormatException e) {
+                throw new IllegalArgumentException("Not a valid cpus value: '$value'", e)
+            }
+        }
+        // turn float to unit
+        if( str.contains('.') ) {
+            try {
+                return Math.ceil(str.toDouble() * 1000)
```
bentsherman, line 347: see previous comment

bentsherman (changes requested):
Do you want me to run the tests myself as part of the review? I assume they work on your end.

Some future items for the record:
- demonstrate this feature in the Nextflow docs
- add a millis unit so that you can do `cpus = 1000.m`

jordeu (approved):

modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
pditommaso: oops yes 👍

modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
pditommaso: Since it's double multipled an integer, the result can be a double, therefore it must be rounded. Then, think it's better to round up to the first integer, to avoid corner cases such as `0.4` => 0

pditommaso:
Since we are moving to a world with fractional CPUs request, I've modelled this adding a `CpuUnit` class as it's done for the memory

pditommaso:
> Do you want me to run the tests myself as part of the review? I assume they work on your end.

Yep, do need to test it

modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
bentsherman: oh right, it's rounding up to the next milli

bentsherman (approved):

bentsherman:
Tests pass for me. Ready to merge.

I will hold off on documentation until we add request / limit for cpus. Side note for docs: you must use `'1000m'` instead of `1000.m` because that already refers to time minutes.

bentsherman:
Going over the related issues #773, @pditommaso is the situation you described here still covered?

> Problem 2) this could be managed in the cpus directive but then how the actual command/tool would use that value? I mean the cpus is reflected in the task.cpus property that is generally used in the command script to set the number of threads/cpus that a multithread program should use. See for example here. Would not make sense to provide there a decimal value.

In other words, what will `task.cpus` be if cpus was provided in millis? If the executor doesn't support millis then I assume `task.cpus` will already be rounded up. But if it does support millis, I'm not sure what it should be.

bentsherman:
From the same thread, this is what I recommend:

> Another thing is how many threads/processes a program should use to perform a computation. Usually it can be set to the fraction of CPU time described above -rounded down when it is larger than 1, rounded up if it is less than 1- num_processes = max(1, floor(cpu_fraction)).

We can explain this in the docs. Basically the `cpus` directive specifies the number of cores to allocate, and `task.cpus` should give the number of threads/processes that are available to use. If you're actually using multi-threading then you probably don't want to specify a fractional CPU anyway.

pditommaso:
Um no, in this PR `task.cpus` always see the (rounder) integer value. Think it's the most safety thing to do. Tools generally need to parallelise by the number of cpus/cores. Don't think it will be useful to give in form of a percentage. If needed, we can expose a `task.cpusMillis`

pditommaso:
Not sure to understand the rationale of the last comment. Why if it was specified `cpus 1.9`, the task should see `1`?

bentsherman:
Honestly... I don't think it matters whether `task.cpus` equals `1` or `2` in that situation. I don't think the user should specify a fractional CPU if the process is doing multi-threading anyway. Perhaps best to leave it as is.

pditommaso:
Disagree, I've seen a lot of use cases in which a task declaring, let's say 4, in reality, it only has a peak using 4 cpus and on average less. I think in these cases declaring a fraction can help to better tune the real usage.

bentsherman:
So in that case we still want to round up the fractional cpu for `task.cpus` right? Fine by me.

bentsherman:
Following my discussion in #785, I recommend we set `task.cpus` to `null` whenever `cpus` is a non-integer value and/or a request/limit range.

pditommaso:
What's the rationale to use `null`? not undertstanding

bentsherman:
Essentially, it is not clear how `task.cpus` should be determined when `cpus` is fractional or request/limit. So we should just set `task.cpus` to `null` to denote that it is undefined behavior. We can add a note to the documentation, so that users know that if they want to request cpus in this way, then they are "on their own".

In the future, if a user thinks `task.cpus` should be set a certain way, then they can make that case.

bentsherman:
Following a private discussion, we will keep the PR as is, `task.cpus` will be rounded up when `cpus` is fractional, for the sake of backwards compatibility.

bentsherman:
@pditommaso This PR is ready to be merged. We agreed that `task.cpus` should be rounded up, which is what it was already doing. The docs are updated now. I think we are good to go. This feature will be useful for testing the resource optimization.

modules/nextflow/src/test/groovy/nextflow/processor/TaskHandlerTest.groovy
```diff
@@ -96,7 +97,7 @@ class TaskHandlerTest extends Specification {
         trace.read_bytes == 20
         trace.write_bytes == 30
         trace.queue == 'longjobs'
-        trace.cpus == 2
+        trace.cpus == CpuUnit.of('2100m').toCores()
```
bentsherman, line 100: Not sure if the trace record should always round up to the next core. Seems like it should depend on the executor, i.e. whether the executor rounds up or not.

bentsherman:
The changes from Andres' branch add fractional cpu support to the local executor, kuberun, google batch, and the task trace record. Comparison support is added to CpuUnit in order to support the local executor.

modules/nextflow/src/test/groovy/nextflow/processor/TaskHandlerTest.groovy
bentsherman: If we do change the trace record, we'll also need to change the `cpus` task metric in Tower to `DECIMAL`

bentsherman:
It looks like AWS Batch also supports fractional CPUs: https://docs.aws.amazon.com/batch/latest/APIReference/API_ResourceRequirement.html

As long as there is at least one vCPU.

bentsherman:
@pditommaso This PR is updated with the latest changes. I also added fractional CPU support to AWS Batch. Ready for your to review.

bentsherman:
Just confirmed that decimal values don't work for AWS Batch:
```
Value 1.1 for type VCPU in resourceRequirement is not valid. Please provide a numeric value. (Service: AWSBatch; Status Code: 400; Error Code: ClientException; Request ID: 84dfe037-5b3e-4e8e-ac58-a504de75a906; Proxy: null)
```

So it looks like decimal values must be less than 1, which is only supported by Fargate. I will revert these changes.

modules/nextflow/src/test/groovy/nextflow/processor/TaskHandlerTest.groovy
bentsherman: For simplicity and backwards compatibility with Tower, I think rounding here is fine. Tower uses `%cpu` anyway for the resource optimization, so it won't be affected.

bentsherman:
@pditommaso I updated and cleaned up this PR. To review, it adds fractional CPU support to the Local, K8s, and Google Batch executors. Otherwise fractional requests are rounded to the next integer. The `task.cpus` and the `cpus` trace field are always rounded up for backwards compatibility.

I think it is good to go. Do you have any remaining concerns?

bentsherman:
@pditommaso I think we can close this PR. I think the original use cases can be covered in other ways:

- setting pod cpus to vary from 0.9 - 1.1 CPUs was more about not setting the CPU limit to allow tasks to burst CPU usage when needed, which Nextflow does now

- submitting many small tasks with e.g. `cpus = 0.1`, I think this is better handled by task batching #3909 

```groovy
// submit 100 tasks with 0.01 cpus, lots of overhead
cpus = '10m'

// submit one task batch with 1 cpu, much more efficient
batch = 100
cpus = 1
```

pditommaso:
Agree

Issue #2519, opened 2021-12-23
Problem with include statement in 21.10+


ATpoint:
The `include` statement is causing some trouble here. 

```
$ NXF_VER=21.04.3 nextflow run main.nf 
N E X T F L O W  ~  version 21.04.3
Launching `main.nf` [elegant_leavitt] - revision: 63425bce5d
executor >  local (1)
[24/1358fa] process > ExampleWorkflow:MODULE [100%] 1 of 1 ✔

$ NXF_VER=21.10.0 nextflow run main.nf // same goes for .1/.2/.3 versions //
N E X T F L O W  ~  version 21.10.0
Launching `main.nf` [stoic_tesla] - revision: 63425bce5d
Missing process or function with name 'include'

 -- Check script 'main.nf' at line: 13 or see '.nextflow.log' file for more details

$ NXF_VER=21.11.0-edge nextflow run main.nf 
N E X T F L O W  ~  version 21.11.0-edge
Launching `main.nf` [disturbed_kalman] - revision: 63425bce5d
Include statement is not allowed within a workflow definition

 -- Check script 'main.nf' at line: 7 or see '.nextflow.log' file for more details
```

Here the MWE:

```
$cat module.nf
process MODULE {
    input:
    val(something)
    output:
    stdout emit: verbiage      
    script: 
    """
    echo $something
    """                
}

$cat main.nf
#! /usr/bin/env nextflow

nextflow.enable.dsl=2

workflow ExampleWorkflow {

    include{ MODULE } from './module'                                                               

    MODULE("HI")

}

workflow { ExampleWorkflow() }
```

pditommaso:
Not a bug, this is not expected to work https://github.com/nextflow-io/nextflow/issues/2162

ATpoint:
I see. I made a PR for the docs to make that clear: https://github.com/nextflow-io/nextflow/pull/2520

Issue #2162, opened 2021-06-09
Include module statement inside a conditional statement throws error when run as a subworkflow


JoseEspinosa:
### Expected behavior and actual behavior

When a module `include` is in a conditional statement it gives an exception if it's run as a subworkflow but not it is run as the `main`.

### Steps to reproduce the problem

I created [this](https://github.com/JoseEspinosa/nxf_issue_conditional_include) repo to reproduce the issue.

If [this](https://github.com/JoseEspinosa/nxf_issue_conditional_include/blob/master/workflows/my_pipeline.nf) script is run as the main it works.

```
git clone https://github.com/JoseEspinosa/nxf_issue_conditional_include.git
cd workflows/
nextflow run my_pipeline.nf
```

However, if it is run from the parent directory with [this](https://github.com/JoseEspinosa/nxf_issue_conditional_include/blob/master/main.nf) script then it throws an exception.

```
cd..
nextflow run main.nf
```
The problem is on [this](https://github.com/JoseEspinosa/nxf_issue_conditional_include/blob/c6b51de6bf7f3bce9e8be8613280442a939e6b31/workflows/my_pipeline.nf#L10) line which has an `include` inside the `if` statement. But the behavior is not consistent when running it as a subworkflow or as the main script.  

### Program output 

```
N E X T F L O W  ~  version 21.04.0
Launching `main.nf` [condescending_mahavira] - revision: 71cf30c291
No such file: Can't find a matching module file for include: ./modules/foo/main.nf

 -- Check script './workflows/my_pipeline.nf' at line: 10 or see '.nextflow.log' file for more details
 ```
 
[nextflow.log](https://github.com/nextflow-io/nextflow/files/6626890/nextflow.log)

### Environment 

* Nextflow version: [21.04.0 build 5552] 
* Java version: [1.8.0_65]
* Operating system: [macOS]
* Bash version: (use the command `$SHELL --version`)[GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin19)]

**Comment by pditommaso** [comment]
Include should not go into `workflow`. This is not expected to be supported in any way.

**Comment by JoseEspinosa** [comment]
The snipped below works, it is ok or it would be better just to have the `include` outside any conditional closure?

```
if (!params.skip_foo) {
    include { FOO } from './modules/foo/main.nf'
}

workflow MY_PIPELINE {    
      
     if (!params.skip_foo) {
         FOO()               
     }
}
```

[...]

PR #2520, opened 2021-12-23
detail on `include` statements

ATpoint:
Make clear that include must be outside of workflow definitions. This worked until 21.04.3 but then was disabled:

- https://github.com/nextflow-io/nextflow/issues/2519
- https://github.com/nextflow-io/nextflow/issues/2162


pditommaso (approved):
Nice, thanks!

PR #2522, opened 2021-12-23
Added affinity option to pod directive

### LINKED ISSUE #2390: Add support for pod / affinity
## New feature

The pod directive currently supports `nodeSelector` for selecting nodes by matching label. There is another property called `affinity` which allows for more complex matching, such as set logic (in, not in) and defining soft conditions:

https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity

I propose we add support for the affinity object in the pod directive. It's easy to do because the PodOptions only needs to save the object as is:

```groovy
    @PackageScope void create(Map<String,String> entry) {
        if( entry.affinity ) {
            this.affinity = entry.affinity as Map
        }
        else if( entry.env && entry.value ) {
```

Here is an example I had to use today to avoid certain GPU nodes on my cluster:
```yaml
spec:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: gpu-type
            operator: NotIn
            values: [T4]
```

And the pod directive I used:
```
pod = [
  affinity: [
    nodeAffinity: [
      requiredDuringSchedulingIgnoredDuringExecution: [
        nodeSelectorTerms: [
          [key: "gpu-type", operator: "NotIn", values: ["T4"]]
        ]
      ]
    ]
  ]
]
```

I patched Nextflow with this code and it works as intended.

bentsherman:
Closes #2390 

This PR adds support for affinity expressions in the pod directive. Examples of affinity can be found [here](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity). Affinity is used to assign pods to nodes, similar to a node selector, but is more expressive and also allows for soft constraints.

Example:
```groovy
pod = [
  affinity: [
    nodeAffinity: [
      requiredDuringSchedulingIgnoredDuringExecution: [
        nodeSelectorTerms: [
          [key: "gpu-type", operator: "NotIn", values: ["T4"]]
        ]
      ]
    ]
  ]
]
```

maps to pod config:
```yaml
spec:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: gpu-type
            operator: NotIn
            values: [T4]
```

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodOptions.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodOptions.groovy
@@ -111,7 +111,7 @@ class PodOptions {
         else if( entry.nodeSelector ) {
             this.nodeSelector = new PodNodeSelector(entry.nodeSelector)
         }
-        else if( entry.affinity ) {
+        else if( entry.affinity instanceof Map ) {
             this.affinity = entry.affinity as Map
         }
         else if( entry.annotation && entry.value ) {
--- a/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodOptions.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodOptions.groovy
@@ -51,6 +51,8 @@ class PodOptions {
 
     private PodNodeSelector nodeSelector
 
+    private Map affinity
+
     private PodSecurityContext securityContext
 
     PodOptions( List<Map> options=null ) {
@@ -109,6 +111,9 @@ class PodOptions {
         else if( entry.nodeSelector ) {
             this.nodeSelector = new PodNodeSelector(entry.nodeSelector)
         }
+        else if( entry.affinity ) {
+            this.affinity = entry.affinity as Map
+        }
         else if( entry.annotation && entry.value ) {
             this.annotations.put(entry.annotation as String, entry.value as String)
         }
@@ -129,14 +134,16 @@ class PodOptions {
 
     Map<String,String> getAnnotations() { annotations }
 
-    PodSecurityContext getSecurityContext() { securityContext }
-
     PodNodeSelector getNodeSelector() { nodeSelector }
 
     PodOptions setNodeSelector( PodNodeSelector sel ) {
         nodeSelector = sel
         return this
     }
+
+    Map getAffinity() { affinity }
+
+    PodSecurityContext getSecurityContext() { securityContext }
 
     PodOptions setSecurityContext( PodSecurityContext ctx ) {
         this.securityContext = ctx
@@ -159,6 +166,7 @@ class PodOptions {
 
     PodOptions plus( PodOptions other ) {
         def result = new PodOptions()
+
         // env vars
         result.envVars.addAll(envVars)
         result.envVars.addAll( other.envVars )
@@ -181,8 +189,11 @@ class PodOptions {
         else
             result.securityContext = securityContext
 
-        // node select
+        // node selector
         result.nodeSelector = other.nodeSelector ?: this.nodeSelector
+
+        // affinity
+        result.affinity = other.affinity ?: this.affinity
 
         // pull policy
         if (other.imagePullPolicy)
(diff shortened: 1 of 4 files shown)


modules/nextflow/src/main/groovy/nextflow/k8s/model/PodOptions.groovy
```diff
@@ -109,6 +111,9 @@ class PodOptions {
         else if( entry.nodeSelector ) {
             this.nodeSelector = new PodNodeSelector(entry.nodeSelector)
         }
+        else if( entry.affinity ) {
```
pditommaso, line 114: It may be better to check `if ( entry.affinity instanceof Map )`, to prevent an invalid type is forced into a map which can be a bit undetermined the result

pditommaso (approved):
Thanks Ben!

bentsherman:
@pdp10 It looks like I never added `matchExpressions` to the pod example. Not sure if I left it out on accident or on purpose, but somehow it worked correctly on the K8s cluster I was using at the time.

It should probably be something like this:
```
pod = [
  affinity: [
    nodeAffinity: [
      requiredDuringSchedulingIgnoredDuringExecution: [
        nodeSelectorTerms: [[
          matchExpressions: [
            [key: "myNode", operator: "In", values: ["true"]]
          ]
        ]]
      ]
    ]
  ]
]
```

Note the extra array brackets because `matchExpressions` is inside an array element of `nodeSelectorTerms`.

pdp10:
@bentsherman 
Just double checking.
With the configuration above, shouldn't I see `myNode=true` in `Node-Selectors` instead of `<none>` when I run `kubectl describe pod <process_name>`?

```
...
QoS Class:                   Guaranteed
Node-Selectors:              <none>
Tolerations:                 node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                             node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:
...
```

Could this be due to the fact that `setAffinity()` is not defined, whereas it is for `nodeSelector`? If I replace nodeAffinity with nodeSelector, I can see `myNode=true` instead of `<none>`.

bentsherman:
I don't think that node affinity rules would be shown under node selectors in that case, because they are a different (although similar) concept. Maybe they are displayed under a separate field, or maybe they aren't shown at all by `kubectl describe`, I'm not sure.

pdp10:
> I don't think that node affinity rules would be shown under node selectors in that case, because they are a different (although similar) concept. Maybe they are displayed under a separate field, or maybe they aren't shown at all by `kubectl describe`, I'm not sure.

It looks like so. This is for a Jenkins pod I have. It contains the declaration:
```
    affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: farmNode
                operator: In
                values:
                - "true"
```

And `kubectl describe pod <pod_name>` returns: 
```
...
QoS Class:       Burstable
Node-Selectors:  farmNode=true
Tolerations:     node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                 node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:
  Type     Reason       Age   From               Message
...
```

bentsherman:
Can you compare the yaml spec between your Jenkins pod and the Nextflow pod? You can do `kubectl get pod -o yaml <pod_name>`.

The reason there is no `setAffinity()` method is because the affinity is set through `PodOptions::create()`.

pdp10:
And this is the Jenkins yaml in the pipeline configuration: 
```
apiVersion: v1
kind: Pod
spec:
  serviceAccount: xxxx
  serviceAccountName: xxxxx
  containers:
  - name: xxxx
    image: xxxxxx
    imagePullPolicy: Always
    command:
    - "sleep"
    args:
    - '1d'
    resources:
      limits:
        cpu: "2"
        memory: 6Gi
      requests:
        cpu: "2"
        memory: 6Gi
    affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: farmNode
                operator: In
                values:
                - "true"
    volumeMounts:
    - mountPath: xxxxxxx
      name: xxxxxxx
    - mountPath: "/home/jenkins/agent"
      name: "workspace-volume"
      readOnly: false
    workingDir: "/home/jenkins/agent"
  dnsPolicy: ClusterFirst
  nodeSelector:
    farmNode: "true"
  restartPolicy: Never
  schedulerName: default-scheduler
  securityContext: {}
  volumes:
  - name: xxxxxxx
    persistentVolumeClaim:
      claimName: xxxxxxx
  - emptyDir:
      medium: ""
    name: "workspace-volume"
```

bentsherman:
Both look correct to me.

pdp10:
@bentsherman 
There was an error in my Jenkinsfile actually. The affinity section was inside containers, whereas it should be outside, but inside specs. I've changed this and now I get the same output as in nextflow pod.

Thanks for your help again. Much appreciated!

PR #2523, opened 2021-12-24
Pipeline reports

jordeu:
## Description

If reports are defined at `nf-<workflow_id>-tower.yml`, the `nf-tower` plugin collects all published files that are reports and writes a `nf-<workflow_id>-reports.tsv` file with all the paths and report keys.

A new event `onFilePublish` has been added at `TraceObserver` to allow `nf-tower` plugin to monitor published files.


plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerReports.groovy
```diff
@@ -0,0 +1,115 @@
...
+                // Initialize report matchers on first event to use the
+                // path matcher of the destination filesystem
+                matchers = patterns.stream()
+                        .map(p -> FileHelper.getPathMatcherFor("glob:**/${p}", destination.fileSystem))
+                        .collect(Collectors.toList())
+            }
+
+            for (int p=0; p < matchers.size(); p++) {
+                if (matchers.get(p).matches(destination)) {
+                    final dst = destination.toAbsolutePath().toString()
```
pditommaso, line 107: Better to use toUriString for this 

https://github.com/nextflow-io/nextflow/blob/f00bfb0d3ad6beb9d4893c0f172b606306df55f3/modules/nf-commons/src/main/nextflow/extension/FilesEx.groovy#L1488-L1488

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerReports.groovy
```diff
@@ -0,0 +1,115 @@
...
+                    final pattern = patterns.get(p)
+                    writer.send { PrintWriter it -> it.println("${pattern}\t${dst}") }
```
pditommaso, line 109: What `pattern` represents here? the glob defined in the config file?

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
```diff
@@ -346,6 +347,8 @@ class PublishDir {
             FileHelper.deletePath(destination)
             processFileImpl(source, destination)
         }
+
+        notifyFilePublish(destination)
```
pditommaso, line 351: Was thinking if it may be better to run this into a separate thread pool

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -1021,6 +1021,10 @@ class Session implements ISession {
         observers.each { trace -> trace.onFlowCreate(this) }
     }
 
+    void notifyFilePublish(Path destination) {
+        observers.each { trace -> trace.onFilePublish(destination) }
```
pditommaso, line 1025: Better to wrap this into a try-catch as done here 

https://github.com/nextflow-io/nextflow/pull/2523/files#diff-bee88a999f8eecd3cec5fcad24b76736162be86626f89d746932f64928f24374R997-R1004

pditommaso (changes requested):

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerReports.groovy
jordeu: Yes. I've added them to the output because is the `key` to get the display name at the reports config.

modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
```diff
@@ -120,4 +122,13 @@ trait TraceObserver {
...
     void onFlowError(TaskHandler handler, TraceRecord trace){}
+
+    /**
+     * Method that is invoke when an output file is published
+     * into a `publishDir` folder.
+     *
+     * @param destination
+     *      The destination path at `publishDir` folder.
+     */
+    void onFilePublish(Path destination){}
```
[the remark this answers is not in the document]
jordeu: ignore

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
jordeu: I think that there is no need to create a new thread for this. This is only triggering the notification to all the observers, but the writing of the file is already done in another thread (using an Agent). Plus, most of the times (if this is not a symbolic link) we invoke this from a thread that is only running the `safeProcessFile` method.

http://github.com/nextflow-io/nextflow/blob/pipeline-reports/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy#L289-L295

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -1021,6 +1023,17 @@ class Session implements ISession {
         observers.each { trace -> trace.onFlowCreate(this) }
     }
 
+    void notifyFilePublish(Path destination) {
+        observers.each { observer -> {
```
pditommaso, line 1027: Why `observer -> { .. } `. Think this creates a nested closure (?). Also better to use just a `for` to keep the same pattern with the other handers

pditommaso:
Can this be moved to *Ready* ?

jordeu:
> Can this be moved to _Ready_ ?

I'm still testing altogether with Tower. When I finish I'll move this to ready.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerReports.groovy
```diff
@@ -145,7 +145,7 @@ class TowerReports {
                     final pattern = patterns.get(p)
                     writer.send { PrintWriter it -> it.println("${pattern}\t${dst}\t${destination.size()}") }
                     final numRep = totalReports.incrementAndGet()
-                    log.debug("Adding report [{}] {} -- {} -- {} -- {} -- {} -- {}", numRep, pattern, dst, destination.getScheme(), destination.toUri(), destination, FilesEx.toUriString(destination))
+                    log.debug("Adding report [{}] {} -- {}", numRep, pattern, dst)
```
pditommaso, line 148: Provided this is OK. Is it not more idiomatic to use groovy string interpolation instead of placeholders ?

jordeu: I think that to use placeholders is better at logging. This avoids to interpolate the string if you are not going to report that log, otherwise the string interpolation is always computed.

pditommaso: AFAIK when using @Slf4j annotation groovy automatically wraps the log statement with a `if` level e.g 

```
if( log.isDebugLevel() ) {
  log.debug "etc"
}
```

jordeu: Yes, you're right. Nice! I'll use groovy string interpolation then.

modules/nextflow/src/main/groovy/nextflow/Session.groovy
pditommaso: there pending comment here

pditommaso:
It looks a test is failing 

```
TowerClientTest > should post create request FAILED
    java.lang.NullPointerException at TowerClientTest.groovy:382
```

pditommaso (approved):
Ok, merging this

abhi18av:
Paolo, would it be possible to configure the Github Action, to upload a `nextflow-pr-xyz-self-contained` build at the PRs which pass the tests?

I have seen this implemented (via some other CI tool) on the bioconda repo https://github.com/bioconda/bioconda-recipes/pull/30297#issuecomment-907612503 

This would speed up the testing of all PRs on Nextflow.

Thoughts?

pditommaso:
@jordeu think there's a problem with this PR, I've got twice the test failing when trying to connecting to tower. 

It seems NF hangs after the execution. Any thread pool not shutdown correclty?

jordeu:
> @jordeu think there's a problem with this PR, I've got twice the test failing when trying to connecting to tower.
> 
> It seems NF hangs after the execution. Any thread pool not shutdown correclty?

I'll check. But what test do you mean?

jordeu:
Ok, may be it was the timer thread that was not always cancelled. I've fixed it and now I'm only creating the timer if I need it.

jordeu:
@pditommaso I'm double checking that the globs are working as expected. Hold a bit the merge please.

pditommaso:
Ok. just for reference this was the [failing test](https://github.com/nextflow-io/nextflow/runs/4969406057?check_suite_focus=true). If you notice the execution is cancelled after 90 mins (timeout)

pditommaso:
> I'm double checking that the globs are working as expected. 

@jordeu any update on this?

jordeu:
> @jordeu any update on this?

I've added some improvements to glob support, fixed one bug, manage malformated yml files and added some tests.

Still I want to test all together at Tower to check if all the reported problems when running `nf-core/rnaseq` pipeline are solved.

jordeu:
After adding new tests, another unrelated test is failing. @pditommaso  do you think it can be related?

pditommaso:
I would say no, this looks some mess with azure image pull 


```
At least one value of specified task container settings is invalid
```

pditommaso (approved):

pditommaso:
@jordeu In the future please sign your commits, otherwise I need to make the bot pass manually. It's enough adding `-s` to the commit

PR #2534, opened 2022-01-04
Fix docker timeout issues.

manuelesimi:
This PR tries to address the issue mentioned in https://github.com/nextflow-io/nextflow/discussions/2510. 

I tried a loop of 10 executions and didn't get the timeout error. Before this change, I got 2 timeouts out of the 10 executions. I can't be 100% sure this fixes it once and for all. But, at least, it seems to limit its frequency.

### DIFF
--- a/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
+++ b/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
@@ -20,6 +20,7 @@ import com.microsoft.azure.batch.protocol.models.AutoUserScope
 import com.microsoft.azure.batch.protocol.models.AutoUserSpecification
 import com.microsoft.azure.batch.protocol.models.AzureFileShareConfiguration
 import com.microsoft.azure.batch.protocol.models.ElevationLevel
+import com.microsoft.azure.batch.protocol.models.EnvironmentSetting
 import com.microsoft.azure.batch.protocol.models.MountConfiguration
 import com.microsoft.azure.batch.protocol.models.UserIdentity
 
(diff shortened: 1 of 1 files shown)


plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -352,7 +353,9 @@ class AzBatchService implements Closeable {
 
         final slots = computeSlots(task, pool)
         log.trace "[AZURE BATCH] Submitting task: $taskId, cpus=${task.config.getCpus()}, mem=${task.config.getMemory()?:'-'}, slots: $slots"
-
+        List<EnvironmentSetting> environmentSettings = new ArrayList<>()
+        environmentSettings << new EnvironmentSetting().withName('DOCKER_CLIENT_TIMEOUT').withValue('120')
+        environmentSettings << new EnvironmentSetting().withName('COMPOSE_HTTP_TIMEOUT').withValue('120')
         final taskToAdd = new TaskAddParameter()
```
pditommaso, line 359: Nice! worth adding a comment with a link to the discussion to highlight the need for those vars

manuelesimi: Done :)

pditommaso: Awesome!

pditommaso:
First merged PR of 2022!

manuelesimi:
Thanks! 🥇 🥇 🥇

pditommaso:
No joy :( 

https://github.com/nextflow-io/nextflow/runs/4708262055?check_suite_focus=true

manuelesimi:
I'm totally puzzled by the fact that it happens **always** on the second process (`FASTQC`) and never on the  first process (`INDEX`). They both use the same image.

It can't be a random network issue.

manuelesimi:
Found this interesting discussion: https://github.com/docker/for-mac/issues/4957

Even if it is mac-related, it suggests that the docker client used by Azure Batch could be the reason. Why we didn't get the error before? An upgrade of the docker client on the VMs from Azure?

manuelesimi:
Shading some light with this issue reported on the AZ Batch repository: https://github.com/Azure/Batch/issues/4 

They mention that if the image is too big, pulling the image can take more than 60 seconds. Hence the error we are also experiencing.

The issue is still open with very recent updates...

pditommaso:
ah, this is a good finding. therefore don't think there's much we can do on our side

manuelesimi:
I agree.

pditommaso:
This PR was reverted d5679ddd0

PR #2182, opened 2021-06-23
Fix extended glob

Lehmann-Fabian:
This PR solves the problems with the extended glob, which can be seen in [https://github.com/nextflow-io/nextflow/pull/2181](https://github.com/nextflow-io/nextflow/pull/2181). Now it passes all tests.
I promised this fix in [https://github.com/nextflow-io/nextflow/pull/2035](https://github.com/nextflow-io/nextflow/pull/2035).

PR #2551, opened 2022-01-11
[Improvement] Extended glob

Lehmann-Fabian:
This PR merges the logic of https://github.com/nextflow-io/nextflow/pull/2182 into https://github.com/nextflow-io/nextflow/pull/2379

With this PR, Nextflow behaves similarly using `-process.scratch true` and not. Moreover, it reduces the number of files copied/moved if the output path contains `**` and the glob parameter is now correctly validated.

Still, all other parameters defined here: https://www.nextflow.io/docs/latest/process.html?highlight=stage#output-path are interpreted as default values. This may still copy/move more files than necessary but could not break functionality. This could be addressed in a following contributions.


Lehmann-Fabian:
The test failure seems not to be related to the changes we made. 
I am looking forward to your feedback @pditommaso.

PR #2559, opened 2022-01-14
Allow to specify memory/cpu resources for the control pod

xhejtman:
If pod is submitted into quota enforcing namespace, cpu/memory limits have to be set. Thus add an ability to set cpu/memory limits.


pditommaso (approved):
Excellent! thanks a lot for the contribution

PR #2562, opened 2022-01-15
Add automountServiceAccountToken option for k8s executor 

xhejtman:
Worker pods do not need to automount service account token usually.
Without token, it starts faster and it is a bit more secure.

Moreover, if for some highly parallel tasks a large number of pods is spawned at once, there are races in mounting. Setting this option to false (not by default), mitigate this issue.


pditommaso (approved):
@bentsherman does look ok to you?

bentsherman (approved):

pditommaso:
Merged. Thanks for this contribution!

Issue #2322, opened 2021-09-16
AWS Batch CLI-mounting potentially disables the container


multimeric:
### Summary
Nextflow mounts the AWS CLI from the AMI into the process container. However the way it does this is dangerous, and can mask the contents of the container itself, causing undefined behaviour.

### Expected behavior and actual behavior
If you get an error like this in AWS Batch:
> CannotStartContainerError: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "/usr/local/env-execute": stat /usr/local/env-execute: no such file or directory: unknown

Then this bug is probably at play. Notably it doesn't *have* to involve `env-execute`. Any missing-file error could result from this bug.

### Steps to reproduce the problem
The steps are documented in this repository here:
https://github.com/multimeric/AwsNextflowBug

The crucial part of this repo are that it:

a) Uses a docker container that depends on the contents of `/usr` being present (all biocontainers do this)
b) Sets the AWS CLI path in the config to `/usr/bin/something`, meaning that the entirety of `/usr` is mounted on top of

### Program output 

```
Error executing process > 'spades'

Caused by:
  Process `spades` terminated for an unknown reason -- Likely it has been terminated by the external system

Command executed:

  spades

Command exit status:
  -

Command output:
  (empty)
```

### Environment 

* Nextflow version: 21.04.0.5552
* Java version: 1.8.0_152-release
* Operating system: Ubuntu 20.04.3 LTS
* Bash version: fish, version 3.2.2

### Discussion of Cause

So I've looked into why this is happening. Whenever you have the AWS config set like this:
```
batch.cliPath = '/usr/bin/foobar'
```
Then nextflow mounts `/usr` from the AMI on top of the container, thereby masking out the entire `/usr/` directory, which is obviously *very bad*. You can see the cause of this behaviour here:
https://github.com/nextflow-io/nextflow/blob/f2f1fdea46631b07fefb48a0d098c66ff53e58fe/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy#L440-L443

One common case where you might have the CLI installed to `/usr/bin/awscli` is if you use the AWS ImageBuilder, which has an AWS CLI component, which automatically installs the CLI into that location.

### Workarounds

[...]

PR #2565, opened 2022-01-17
Docs: Warn about AWS CLI grandparent path in container

manics:
This should help avoid https://github.com/nextflow-io/nextflow/issues/2322

Should we modify the miniconda example to use a path that's definitely unique, e.g. `/opt/aws-ecs-mount/conda` instead of something in `$HOME`, to further reduce the chance of a conflict?


pditommaso (approved):
Nice, thanks for the contribution.

PR #2569, opened 2022-01-18
Scipio extended error report

mguinin:
Hi, thank you very much for developing nexflow.
At Scipio-bioscience we created this fork because we felt the need to have more precise errors report. For now this fork is meant to be more functional than pretty, it is a work in progress.

Our problem was the lines reported in the error messages didn't always point the exact location where the error happened, but rather the general wrapping scope, furthermore the script names were not human readable (numbers) in the extended error stack in .nextflow.log 

As our nextflow code grow in size we had more and more difficulties to find the new errors, so I created this hack were additional information is appended at the end of the cause chain of the error stack, so we are 100% sure the info is always displayed. Obviously it would be cleaner to move those message outside of the error stack. We are open to any suggestion.

Thanks again,
Mathieu Guinin
Scipio-bioscience
mathieu.guinin@scipio.bio


docs/conf.py
```diff
@@ -48,9 +48,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '21.10'
+version = '21.11'
 # The full version, including alpha/beta/rc tags.
-release = '21.10.1'
+release = '21.11.0-SNAPSHOT'
```
pditommaso, line 54: This should not be commited

modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy
```diff
@@ -637,6 +637,7 @@ class TaskPollingMonitor implements TaskMonitor {
         while( runningQueue.size() ) {
 
             TaskHandler handler = runningQueue.poll()
+            println("Killing task: " + handler.task.name)
```
pditommaso, line 640: `println` statement should not be used for logging

modules/nextflow/src/main/groovy/nextflow/script/ExecutionStack.groovy
```diff
@@ -89,3 +131,56 @@ class ExecutionStack {
     }
 
 }
+
+@CompileStatic
+class ERROR_IN_NEXTFLOW_PIPELINE extends Throwable {
```
pditommaso, line 136: The use of all uppercase class names is very unusual for Groovy/Java code convention. Camel case should be used instead

modules/nextflow/src/main/resources/META-INF/plugins-info.txt
```diff
@@ -1,8 +1,8 @@
-nf-amazon@1.3.4
 nf-sqldb@0.2.0
 nf-console@1.0.3
-nf-google@1.1.3
 nf-ignite@1.2.3
 nf-azure@0.11.2
 nf-tower@1.3.0
+nf-amazon@1.3.4
+nf-google@1.1.3
```
pditommaso, line 7: This should not be commited

modules/nf-commons/src/main/nextflow/Const.groovy
```diff
@@ -53,17 +53,17 @@ class Const {
...
+    static public final long APP_TIMESTAMP = 1642433960696
 
     /**
      * The app build number
      */
-    static public final int APP_BUILDNUM = 5647
+    static public final int APP_BUILDNUM = 5684
 
     /**
      * The app build time string relative to UTC timezone
```
pditommaso, line 69: This should not be commited

nextflow
```diff
@@ -16,7 +16,7 @@
 #  limitations under the License.
 
 [[ "$NXF_DEBUG" == 'x' ]] && set -x
-NXF_VER=${NXF_VER:-'21.10.1'}
+NXF_VER=${NXF_VER:-'21.11.0-SNAPSHOT'}
 NXF_ORG=${NXF_ORG:-'nextflow-io'}
```
pditommaso, line 20: This should not be commited

pditommaso (changes requested):
Hello, thanks for considering contributing to this project. 

I've left some comments, most of them related to changes related to files automatically modified by the build script. Those changes should not be committed to preventing conflicts during the merge. 

Also, I need more context on the rationale of this PR. How exactly is modifying the error message? Can you provide a script example showing the PR in action?

pditommaso:
Please also review the contributing document https://github.com/nextflow-io/nextflow/blob/master/CONTRIBUTING.md#contributing-to-nextflow

mguinin:
Thank you for considering my pull request !
I have done the change about the naming convention and remove the unwanted files from commit.


**Here are two examples where the error report is improved:**


1. In the following example the error line in the console point to line 11, the proposed improvement allow to find the root cause of the problem at line 4. Notice that in this example the initCause wrapping is performed in the nextflow code, usually it gets even worse when it happens in groovy code: we frequently got the main workflow opening bracket as the error line. 


```
nextflow.enable.dsl=2

def failTestFunc() {
    throw new Error("this is meant to fail")
}

workflow {
    try {
        failTestFunc()
    } catch (Throwable e) {
        throw new Error("something went wrong").initCause(e)
    }
}
```



2.  In this example nexflow only display the error message, not the full stack. The proposed improvement allow to retrieve the exact error line:

```
nextflow.enable.dsl=2

process failTestProcess {
    input:
        val(x)
    output:
        val(y)

    exec:
        throw new Error("this is meant to fail")
        y=2*x
}

workflow {
    failTestProcess(42)
}
```

pditommaso:
Closing as outdated

Issue #2557, opened 2022-01-12
Error while publishing directories on azure


fbdtemme:
### Expected behavior and actual behavior

On the microsoft azure platform, when a process outputs a directory containing files and subfolders, nextflow fails to copy these to the publishDir. The files are uploaded from the worker node to the process working directory, but not copied to the publishDir.
Copying single files to the publishDir works without problems.

### Steps to reproduce the problem

```#!/usr/bin/env nextflow
nextflow.enable.dsl=2

process sayHello {
    publishDir "${params.outdir}", mode: "copy"

    input: 
    val x

    script:
    """
    echo '$x world!'

    mkdir -p hello-results
    cd hello-results
    echo "$x" > "${x}.txt" 
    """

    output: 
    path "hello-results"         , emit: results
}


workflow {
  sayHello("hello")
}
```

```nextflow run main.nf -c azure.config -w az://myblobstorage/work --outdir az://myblobstorage/results```

### Program output 

```Jan-12 16:10:20.963 [FileTransfer-1] WARN  nextflow.processor.PublishDir - Failed to publish file: az://bloboutput/nextflow/test/15/53b855da5fea98b910d3798b95b9a2/hello-results; to: az://bloboutput/results/test/hello-results [copy] -- See log file for details
com.azure.storage.blob.models.BlobStorageException: Status code 400, "﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>InvalidInput</Code><Message>One of the request inputs is not valid.
RequestId:719fe262-f01e-010b-0cce-074b26000000
Time:2022-01-12T16:10:20.9587019Z</Message></Error>"
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at com.azure.core.http.rest.RestProxy.instantiateUnexpectedException(RestProxy.java:334)
	at com.azure.core.http.rest.RestProxy.lambda$ensureExpectedStatus$5(RestProxy.java:375)
	at reactor.core.publisher.MonoFlatMap$FlatMapMain.onNext(MonoFlatMap.java:118)
[...]

PR #2576, opened 2022-01-19
Use account SAS token instead of service SAS token

### LINKED ISSUE #2563: Nextflow on Azure Complained about "Unable to download path"
Hello community members,

I have a simple pipeline working fine on AWS batch, but not so much on Azure. I am hoping if anyone can shed some light.

Context

1/ I have created the batch account, container registry, and storage account on Azure, and have these values proper defined (I think) on nextflow-azure.config file

2/ nextflow version 21.10.6.5660

3/ cmd
```
nextflow -C nextflow-azure.config run main.nf --run_id ASLDKJGSALKDJG --dev -w az://testdata-output/work
```

4/ image has been pushed to Azure container registry, and I confirm I can pull the image to local after uploading

Potential root causes (IMO)

1/ permission issue to the storage account, or ACR, or both
2/ tenant id?

fbdtemme:
Currently the azure backend is limited to using only the container that contains the workdir.
This is because the SAS token is generated on the container level and does not provide access to different containers in the same storage account.

This fix generates an account level SAS token instead of a service (per container) SAS token.
When performing copy operations between storage accounts we also need to append the SAS token to the source url,

This fixes #2563


abhi18av:
Continuing the conversation from the scenarios set in https://github.com/nextflow-io/nextflow/issues/2563#issuecomment-1016430158 

I tried a build from this PR with scenario-1, input/output/data buckets are all different and came across this issue

```
Error executing process > 'RNASEQ:FASTQC (FASTQC on ggal_gut)'

Caused by:
  At least one value of specified task container settings is invalid

Command executed:

  fastqc.sh "ggal_gut" "ggal_gut_1.fq ggal_gut_2.fq"

Command exit status:
  -

Command output:
  (empty)

Work dir:
  az://nf-workdir/d2/33e17f1389b66268f7e5811efb368e

```

fbdtemme:
This seems like a seperate issue to me: https://github.com/nextflow-io/nextflow/discussions/2510
I can run your example without problems.

abhi18av:
Interesting - could you please tell me which region you're deploying the resources?

abhi18av:
I have tried to use the build from this PR with distinct input/output/work directories using 

1. Azure Container Registry
2. External registry


I must be doing something silly since I'm getting the same results 😅 

```
Error executing process > 'RNASEQ:INDEX (ggal_1_48850000_49020000)'

Caused by:
  One or more container images specified are invalid

Command executed:

  salmon index --threads 1 -t ggal_1_48850000_49020000.Ggal71.500bpflank.fa -i index


```


Could you please share more info about your environment/resource-configuration?

manuelesimi:
Thanks @fbdtemme, this is a really valuable contribution.

I tested these changes in several cases:
1. input files in the same blob storage, same level, or level up/down
2. input files in a different blob storage

It worked as expected in all cases. I can then confirm that this PR fixes what we discussed in https://github.com/nextflow-io/nextflow/issues/2557

TODO:
* the commit must be signed (DCO)

It doesn't look to me that this PR requires new test cases or changes in the existing ones. But this is something I defer to @pditommaso, who should also review and approve (of course!!).

fbdtemme:
@manuelesimi I added the DCO

plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzFileSystem.groovy
```diff
@@ -408,8 +408,15 @@ class AzFileSystem extends FileSystem {
 
     @PackageScope
     void copy(AzPath source, AzPath target) {
+        final sasToken = provider.getSasToken()
+        String sourceUrl = source.blobClient().getBlobUrl()
+
+        if (sasToken != null) {
+            sourceUrl += "?${sasToken}"
+        }
```
pditommaso, line 416: Looking the implementation of `getBlobUrl` the url can contains parameters, therefore appending `?` is a bt tool optimistic  

```
    public String getBlobUrl() {
        String blobUrl = azureBlobStorage.getUrl();
        if (this.isSnapshot()) {
            blobUrl = Utility.appendQueryParameter(blobUrl, "snapshot", getSnapshotId());
        }
        if (this.getVersionId() != null) {
            blobUrl = Utility.appendQueryParameter(blobUrl, "versionid", getVersionId());
        }
        return blobUrl;
    }
```

The `Utility.appendQueryParameter` method should be used instead

fbdtemme: I could not use Utility.appendQueryParameter since that needs a key and the sasToken does not have any but I took possible url parameters into account

pditommaso (approved):

pditommaso:
Thanks a lot for this contribution

Issue #2161, opened 2021-06-09
Finer grain control of K8S security context


tdudgeon:
## New feature

Add ability to define `runAsGroup` and `fsGroup` properties in the `securityContext`.

## Usage scenario 

It is sometimes necessary to have very tight control over the ownership of files that Nextflow creates using the K8S executor.
The k8s scope already provide a `runAsUser` property that gets added to the securityContext of the pod, but it seems that it is not possible to define the `runAsGroup` and `fsGroup` properties. This means we cannot control the file ownership full.

## Suggest implementation 

Add `runAsGroup` and `fsGroup` properties in the same way that `runAsUser` is implemented.

tdudgeon:
Poking this to keep it alive.
What are the chances of the various K8S related issues getting some attention?

bentsherman:
You should be able to do this with the `securityConext` pod option:
```
pod = [
  securityContext: [runAsUser: 1000, runAsGroup: 3000, fsGroup: 2000]
]
```

In looking through the code I saw that this feature is already supported but just isn't documented. I will update the docs. It seems `runAsUser` is directly available as a shortcut, for your case I think simply passing a `securityContext` object will do.

bentsherman:
You can also provide set the `securityContext` in the `k8s` config scope so that it applies to all pods:
```
k8s {
  securityContext = [runAsUser: 1000, runAsGroup: 3000, fsGroup: 2000]
}
```

PR #2584, opened 2022-01-21
Add securityContext pod option to docs

bentsherman:
This PR documents the `securityContext` pod option, which was supported but not documented. Brought up indirectly by #2161 .


pditommaso (approved):
Nice 👌

PR #2591, opened 2022-01-25
Refactor docs process.rst

Kibubu:
Some typos fixed
Examples extend to contain `echo true` where the output would be expected
Example added to input type `path` added to show usual renaming syntax
Emphasis is now done with `*` instead of `` ` ``.


docs/process.rst
```diff
@@ -862,6 +862,9 @@ val(x)          x
 file(x)         (not supported)
 file('name')    'name'
 file(x:'name')  x:'name'
+path(x)         (not supported)
+path('name')    (not supported)
+path(x:'name')  (not supported)
```
pditommaso, line 867: Think, `path` does not support anymore this notation but it should be `path(x, stageAs:'name')`

Kibubu: You are right, just tested and confirmed,

While `path(x)` and `path('name')` work as expected `path(x:'name')` has to be replaced by `path(x, stageAs:'name')`

```suggestion
path(x, stageAs:'name')  (not supported) 
```

docs/process.rst
```diff
@@ -86,10 +89,54 @@ pipeline script.
 .. warning:: Since Nextflow uses the same Bash syntax for variable substitutions in strings, you need to manage them
   carefully depending on if you want to evaluate a variable in the Nextflow context - or - in the Bash environment execution.
 
-When you need to access a system environment variable  in your script you have two options. The first choice is as
-easy as defining your script block by using a single-quote string. For example::
+When you need to access a system's environment variable in your script you have three options. 
+
+The first choise is to use Nexflow and Bash environment variables simultaneously.
```
pditommaso, line 94: think `choise` should be `choice`

docs/process.rst
```diff
@@ -86,10 +89,54 @@ pipeline script.
...
+    process printNextflow {
+       echo true
+
+       """"
+       echo The path is: $PATH
+       """"
+
+    }
+
+While the second example will print Bash's ``PATH`` variable::
```
pditommaso, line 109: More than Bash's variable I think it would be better to say "the ``PATH`` variable defined in the system environment".

docs/process.rst
```diff
@@ -4,13 +4,14 @@
...
-In Nextflow a `process` is the basic processing `primitive` to execute a user script.
+In Nextflow a *process* is the basic processing *primitive* to execute a user script.
 
-The process definition starts with the keyword ``process``, followed by process name and finally the process `body`
+The process definition starts with the keyword ``process``, followed by process name and finally the process *body*
 delimited by curly brackets. The process body must contain a string which represents the command or, more generally,
 a script that is executed by it. A basic process looks like the following example::
 
   process sayHello {
+      echo true
```
pditommaso, line 14: Think here `echo` should not be added since the stdout is redirected to the file.

docs/process.rst
```diff
@@ -86,10 +89,54 @@ pipeline script.
...
+When you need to access a system's environment variable in your script you have three options. 
+
+The first choise is to use Nexflow and Bash environment variables simultaneously. 
+
+Nextflow will use Bash's environment variables unless they are overwritten within the Nextflow context. 
+The first example will print `Nextflow`::
+
+    PATH = "Nextflow"
+    process printNextflow {
+       echo true
```
pditommaso, line 101: It may be useful to add a note box explaining why the use `echo` directive vs the `echo` in the command script

docs/process.rst
```diff
@@ -557,6 +618,12 @@ Will output::
     seq1 seq2 seq3
     ...
 
+.. note::
+  This behaviour requires a limited buffer for the channel. In this examples ``.buffer(size:3)`` was used in the first line.
+
+.. note::
+  Processes with an limiited buffer input channel won't execute unless the buffer is filled. providing
```
pditommaso, line 625: there's a typo in `limiited`

pditommaso (changes requested):
Thanks a lot this is very useful. I left some comments. In general, I think the `echo` directive should not be added to all process. Maybe it can shown how/why to use it and mention the possibility to use the cli option `-process.echo`

pditommaso:
Most of these changes have been incorporated in this #2800 PR. Thanks for this contribution

PR #2594, opened 2022-01-26
Update google cloud docs

bentsherman:
Added some notes about preemptible VMs and resource quotas on Google Cloud.


pditommaso:
Nice, thanks!

PR #2616, opened 2022-02-02
Avoid to read an included config file more than once

jordeu:
## Description

When using Tower (or trace enabled) Nextflow is reading the `includeConfig` file twice. 

First time, it is read as expected when building the config map here:
https://github.com/nextflow-io/nextflow/blob/787a1fd84aead0f285ab94ab8a328093cbe0ec84/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy#L287

and the second time is read when building the resolved config string here:
https://github.com/nextflow-io/nextflow/blob/787a1fd84aead0f285ab94ab8a328093cbe0ec84/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy#L306-L307

It is not possible to reuse the same config object when building the resolve configuration due to the use of the method `setShowClosures` on the second builder.
https://github.com/nextflow-io/nextflow/blob/21792bf727f668c50a6155153b208a0faae88d6c/modules/nextflow/src/main/groovy/nextflow/config/ConfigBuilder.groovy#L830-L835

This is a proposal to memoize the content of the included config file to avoid multiple reads.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/config/ConfigBase.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/config/ConfigBase.groovy
@@ -17,6 +17,8 @@
 
 package nextflow.config
 
+import groovy.transform.Memoized
+
 import java.nio.file.NoSuchFileException
 import java.nio.file.Path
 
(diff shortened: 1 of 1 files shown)


pditommaso:
Can I move this to "Ready for review" ?

jordeu:
Ready @pditommaso

pditommaso:
I cannot think any negative side effects, but I'm not understanding for what reason is invoked twice without the memorisation

pditommaso:
Umm, but then the memoised should work within the config object instance, in the example you are reporting the should different instances. How could it work?

pditommaso:
Got it 👍

pditommaso (approved):

PR #2619, opened 2022-02-03
Extended description of the publishDir Directive

Eric-CH-Chen:
Added a sentence and and example code block to clarify the scope that publishDir will act on. Specifically, outputs to be written by publishDir must be part of the output code block in the process rather than the general outputs of the process itself under the output dir under ``work`` dir.


Eric-CH-Chen:
I didn't know about the DCO (first timer), will close and re-commit and create a new pull request.

PR #2631, opened 2022-02-07
feature: unescape strings

jorgeaguileraseqera:
Search and replace Strings between 3 backticks to avoid escaping them

Signed-off-by: Jorge Aguilera jorge.aguilera@seqera.io


jorgeaguileraseqera:
we need to find another way instead to use a new string syntax

PR #2632, opened 2022-02-08
Accomodate config options for azcopy

abhi18av:
This PR aims to address the discussion https://github.com/nextflow-io/nextflow/discussions/2610 by exposing the azcopy options and allowing users to tweak these settings.

The PR adds relevant options from the [`azcopy copy`](https://docs.microsoft.com/en-us/azure/storage/common/storage-ref-azcopy-copy?toc=/azure/storage/blobs/toc.json#options) page.


abhi18av:
Regarding this PR, 


I noticed that `maxParallelTransfers, maxTransferAttempts and Duration delayBetweenAttempts` are currently being specified at the `azure.batch` level and I'm not sure whether the options for tweaking `azcopy` behavior shoud go within the `azure.batch.azcopy` scope or should they have their own `azure.azcopy` scope?

What are your thoughts @jorgeaguileraseqera ?

plugins/nf-azure/src/main/nextflow/cloud/azure/file/AzBashLib.groovy
```diff
@@ -50,17 +51,17 @@ class AzBashLib extends BashFunLib<AzBashLib> {
         nxf_az_upload() {
             local name=\$1
             local target=\${2%/} ## remove ending slash
-            local base_name="\$(basename "$name")"
-            local dir_name="\$(dirname "$name")"
+            local base_name="\$(basename "\$name")"
+            local dir_name="\$(dirname "\$name")"
```
pditommaso, line 56: This means is all broken what we have now?

abhi18av: Not really, the change is basically to accommodate the interpolation for `--block-blob-tier $blobTier --block-size-mb $blockSize` (and other) options.

pditommaso: ok, what confuse me is that I don't see in the PR the change from a `'''` string block to a `"""` string block

abhi18av: Okay, we can avoid changing the string type and escaping the `$`.


> prepend a snippet that declares that values a Bash variables

Do you mean something like the following? 


```groovy



    protected String setupAzCopyOpts() {
        if( !retryMode )
            return ''
        """
        # custom env variables used for azcopy opts
        export AZCOPY_BLOCK_SIZE_MB=${blockSize}
        export AZCOPY_BLOCK_BLOB_TIER=${blobTier}
        """.stripIndent().rightTrim()
    }


    protected String azLib() {
        '''
            ...
                azcopy cp "$name" "$target?$AZ_SAS" --recursive --block-blob-tier $AZCOPY_BLOCK_BLOB_TIER --block-size-mb $AZCOPY_BLOCK_SIZE_MB
            ...

        '''.stripIndent()
    }


    String render() {
        super.render() + setupAzCopyOpts() + azLib()
    }




                
```

pditommaso: yes!

abhi18av:
@pditommaso , I've tested the code in this PR on Azure Batch system

For the following configuration in the `azure` block

```nextflow 
    azure {
      azcopy {
        blockSize = "10"
        blobTier = "Hot"
      }

      batch {...}

      storage {...}
}

```

The `.../.azcopy_log`, confirms that the settings from `nextflow.config` are picked up and used by the `azcopy` tool while uploading the data from task node to the blob container.

```
2022/02/09 19:33:51 Job-Command cp .command.err https://container.blob.core.windows.net/container/work/b8/7c1686fa3fd8061af41d4ed4c89f4d/.command.err?se=2022-02-09t07%3A33%3A28z&sig=-REDACTED-&sp=rwdlacut&srt=sco&ss=bf&sv=2020-02-10 --block-blob-tier Hot --block-size-mb 10 
2022/02/09 19:33:51 Number of CPUs: 4
2022/02/09 19:33:51 Max file buffer RAM 2.000 GB

```

pditommaso:
Thanks for this PR @abhi18av, I was just thinking if the azcopy config should be top level setting (as it's now) or it should go under `batch?

abhi18av:
Well, to be honest, I think it belongs under the `batch` scope since this is only run in that environment. Let me refactor the code and circle back.

pditommaso (approved):
Thinking more about this, I'd say that's just since azcopy is not a specific component of Batch. Merging it

pditommaso:
Merged manually by [3998a56](https://github.com/nextflow-io/nextflow/commit/3998a56b2b322f986e908b17ad41f8c2b52210a1) (for future contribution open a PR directly in this repo, this allow me to work in the same branch)

PR #2634, opened 2022-02-08
bugfix: Fix build warning + plugins assemble

jorgeaguileraseqera:

### DIFF
--- a/plugins/build.gradle
+++ b/plugins/build.gradle
@@ -63,6 +63,7 @@ subprojects {
      * Creates plugin zip and json meta file in the plugin `build/libs` directory
      */
     task makeZip(type: Jar) {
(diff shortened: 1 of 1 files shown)


jorgeaguileraseqera:
At the end I think is easier to join in a single task both actions (copy and unzip) and avoid the dependencies between them

Also assemble is fixed and running the main assemble task run all plugins assemble tasks

pditommaso:
Like this more! Having this, can `:plugins:assemble` removed from the makefile ? 

https://github.com/nextflow-io/nextflow/blob/a96f23123b22569d428f2d4318979a6ba9e38afe/Makefile#L46-L46

pditommaso (approved):
Nice

Issue #2641, opened 2022-02-11
Configuration tab on run page resolving to n/a when specifying enabled = true in ~/.nextflow/config


jordeu:
## Bug report 

see issue at https://github.com/seqeralabs/nf-tower/issues/338

### Environment 

* Nextflow version: 21.10.6 and 21.04.3

PR #2642, opened 2022-02-11
Resolve config when Tower is enabled using the config file

jordeu:
## Description

Resolve the configuration to show at Tower when it is only enable using configuration file. 

Example config file:
```
tower {
  enabled = true
}
```

See issue #2641 and https://github.com/seqeralabs/nf-tower/issues/338

Signed-off-by: Jordi Deu-Pons <jordi@jordeu.net>


pditommaso (approved):
Well spotted!

PR #2643, opened 2022-02-11
feature: abort session if publish fails

### LINKED ISSUE #2636: Add ability to stop pipeline execution publishDir reports an error
Currently, errors raised during the publishing of output files via `publishDir` are silently ignored and a warning is reported into the `.nextflow.log` file. 

The goal of this feature is to add a new option to the `publishDir` directive named `failOnError: true|false` 

When this option is enabled exception is thrown during the publishing phase (see below) and reported as process error.   

https://github.com/nextflow-io/nextflow/blob/a96f23123b22569d428f2d4318979a6ba9e38afe/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy#L323-L323 

When this happens the usual [errorStrategy](https://www.nextflow.io/docs/latest/process.html#errorstrategy) to decide how to handle the error should take place.

The `failOnError` should be set `false` by default, unless the [strict](https://github.com/nextflow-io/nextflow/blob/a96f23123b22569d428f2d4318979a6ba9e38afe/modules/nextflow/src/main/groovy/nextflow/NF.groovy#L68-L68) is enabled

jorgeaguileraseqera:
resolve #2636
by the moment we'll abort the session in case some publish fails


docs/process.rst
```diff
@@ -1688,6 +1688,16 @@ The number of times a failing process is re-executed is defined by the `maxRetri
...
+As shortcut you can pass a Map, as a second argument, to configure the `maxRetries`_ and `maxErrors`_ directives.
+For example::
+
+    process retryIfFail {
+       errorStrategy 'retry', maxRetries:3, maxErrors:10
+
+       script:
+       <your command string here>
+    }
+
```
pditommaso, line 1701: this is a nice idea, but I would keep outside the scope of this PR

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
```diff
@@ -321,6 +336,11 @@ class PublishDir {
         }
         catch( Throwable e ) {
             log.warn "Failed to publish file: ${source.toUriString()}; to: ${target.toUriString()} [${mode.toString().toLowerCase()}] -- See log file for details", e
+            if( NF.strictMode || failOnError){
+                if( taskProcessor && taskRun ){
+                    taskProcessor.publishFailed(taskRun, e, target)
```
pditommaso, line 341: The session can be accessed directly using this `Global.session` singleton. An example here 

https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy#L106-L106

pditommaso (changes requested):
Ok, left a few comments, mostly let's focus this PR on the core feature without altering the `errorStrategy` directive options

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
jorgeaguileraseqera: yes, I saw we can access to session using Global but I think is better to delegate the logic to who call us (taskProcessor)

pditommaso: fair enough

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
```diff
@@ -102,6 +110,10 @@ class PublishDir {
 
     private String taskName
 
+    TaskProcessor taskProcessor
+
+    TaskRun taskRun
+
```
pditommaso, line 116: Still not convinced about adding those fields just to access the session since it's introducing an unnecessary dependency with task context. 

It's true that *now* `publishDir` can only be executed within a process, however a plan is to make it also available as an operator or from with in a workflow (some discussion [here](https://github.com/nextflow-io/nextflow/discussions/1933)), and in that case there would be not `TaskRun` and `TaskProcessor` instances. 

Think it would be better just using the `Global.session` object, which I agree it's ugly, but it's already used within this class. 

It also would be interesting planning the use of some minimal DI framework to replacing the the use of `Global` object (but this is another story..)

jorgeaguileraseqera: Yes, I agree about unnecessary dependencies ( also I evaluated to define a `PublishDirListener` or similar but doesn't like to much)

Ok, I'll use the `Global`

PR #2645, opened 2022-02-11
bugfix: detect and skip bom mark in files

### LINKED ISSUE #2297: Processing a CSV file with a byte order mark gives incorrect results
## Bug report 

Processing CSV files with a [byte-order mark](https://en.wikipedia.org/wiki/Byte_order_mark) leads to first column being read as `null`.

I expected to be able to read and split a CSV file, accessing the first column of data as described here: https://www.nextflow.io/docs/latest/operator.html#splitcsv

However, the first column was always parsed as `null`. This happened whether I accessed the field by name (row.sampleId) or position (row[0]). 

Removing the byte-order mark (either with `dos2unix` or BBEdit) solved the problem. I initially thought it was a line-ending issue (this file also uses windows line endings) but no, it's the BOM.

Since the BOM is an optional part of UTF-8 (allowed but not required) I believe Nextflow should correctly parse files which include it. Failing that, an informative warning or error message would be very nice.

See attached nextflow script and datafile:
Processing script (had to rename as '.txt' to upload to GitHub): [process_scratch.nf.txt](https://github.com/nextflow-io/nextflow/files/7122571/process_scratch.nf.txt)

Data file it was reading:
[scratch_csv_bom.csv](https://github.com/nextflow-io/nextflow/files/7122564/scratch_csv_bom.csv)

### Program output 

```
Launching `process_scratch.nf` [elegant_lamarr] - revision: a387bc7c40
[﻿sampleId:EXP21001039, read1:L1_read1.fastq, read2:L1_read2.fastq]
[﻿sampleId:EXP21001039, read1:L2_read1.fastq, read2:L2_read2.fastq]
[﻿sampleId:EXP21001039, read1:L3_read1.fastq, read2:L3_read2.fastq]
[﻿sampleId:EXP21001039, read1:L4_read1.fastq, read2:L4_read2.fastq]
[﻿sampleId:EXP21001039, read1:L5_read1.fastq, read2:L5_read2.fastq]
### Incorrect lines below ###
[null, L1_read1.fastq, L1_read2.fastq]
[null, L2_read1.fastq, L2_read2.fastq]
[null, L3_read1.fastq, L3_read2.fastq]
[null, L4_read1.fastq, L4_read2.fastq]
[null, L5_read1.fastq, L5_read2.fastq]
### Incorrect lines above ###
executor >  local (5)
[d5/2449b5] process > printInput (1) [100%] 5 of 5 ✔
```

The incorrect lines are annotated, should have `EXP21001039`  instead of `null`.

### Environment 

* Nextflow version: 20.10.0
* Java version: openjdk version "10.0.2"
* Operating system: macOS
* Bash version: zsh 5.8 (x86_64-apple-darwin20.0)

### Additional context

I don't know if the issue is with `splitCsv` or with `map` or what. It seems like the file is read correctly (hence the first 5 lines in the log, between splitCsv and map, being correct), but the column can't be accessed. Perhaps the BOM is corrupting the ability to access by name, but I also can't access it by position (either 0 or 1).

jorgeaguileraseqera:
resolves #2297 processing a csv file with a byte order mark gives incorrect results


modules/nextflow/src/main/groovy/nextflow/splitter/AbstractTextSplitter.groovy
```diff
@@ -319,4 +327,21 @@ abstract class AbstractTextSplitter extends AbstractSplitter<Reader> {
...
+
+    protected int positionAfterBOM(Reader reader ){
+        if( !reader.markSupported() )
+            return 0
+        char[] beginner = new char[3];
+        reader.mark(beginner.length)
+        int read = reader.read(beginner)
+        reader.reset()
+        if( read ){
+            if( beginner[0] == '\ufeff' || beginner[0] == '\ufffe') {
```
pditommaso, line 339: Out of curiosity, should not this handled by some groovy helper methods? If I'm not wrong I remember something like this implemented by methods in the  `DefaultGroovyMethods` extensions

jorgeaguileraseqera: ummm no idea
I was looking if Groovy has something about it but all I've found redirects to Apache Commons so at the end I choose to implement it

pditommaso: nvm

tests/parse_bom_csv.nf
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env nextflow
+
```
pditommaso, line 2: Unit test should be enough, this could be removed

pditommaso (approved):

PR #2658, opened 2022-02-18
feature: improve error message

### LINKED ISSUE #2627: Incorrect error message "Access to .... is undefined since process doesn't declare any output"

**Expected** An error message matching the problem is printed

**Actual** An error message is printed which seems to state a basic incorrect fact about my pipeline.

I created a simple pipeline like this:

```groovy
nextflow.enable.dsl = 2

process hello {
    echo true

    output:
        path 'test.txt'

    script:
        """
        echo "Hello World" > test.txt
        """
}

 workflow {
     hello
     hello.out.view()
 }
```

Then I run it with:

```
./nextflow run test.groovy
```

### Program output 

```
Launching `...` [awesome_babbage] - revision: fa509e52de
Access to 'hello.out' is undefined since process doesn't declare any output

 -- Check script '....' at line: 20 or see '.nextflow.log' file for more details
```

### Additional context

I'm not quite sure what my real problem is, but whatever the problem, the error message seems misleading since it states my process does not declare an output when it actually does.

jorgeaguileraseqera:
closes #2627

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/ProcessDef.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/ProcessDef.groovy
@@ -149,7 +149,11 @@ class ProcessDef extends BindableDef implements IterableDef, ChainableDef {
     ProcessConfig getProcessConfig() { processConfig }
 
     ChannelOut getOut() {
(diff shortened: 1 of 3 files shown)


pditommaso:
Ok, nice this. But I've realised it can be made a bit more precise by checking if `output==null`. in that case means the process has not been invoked

Issue #2654, opened 2022-02-17
kuberun executor returns 127 but run pod is still running


kfrancisdev:
### Expected behavior and actual behavior

expected: when using kuberun executor if a run pod still has a status of running then the nextflow should not attempt to shutdown. It should wait until the pod status has changed to terminated or some appropriate status. 

actual: shutdown is attempted when run pod is still running and exit code of 127 is return

### Steps to reproduce the problem

(Working on providing a simple test case)

### Program output 

Feb-15 21:20:13.885 [main] DEBUG nextflow.config.ConfigBuilder - User config file: /home/ubuntu/.nextflow/assets/kfrancisdev/ncov2019-artic-nf/nextflow.config
Feb-15 21:20:13.885 [main] DEBUG nextflow.config.ConfigBuilder - Parsing config file: /home/ubuntu/.nextflow/assets/kfrancisdev/ncov2019-artic-nf/nextflow.config
Feb-15 21:20:13.895 [main] DEBUG nextflow.config.ConfigBuilder - Applying config profile: `oke`
Feb-15 21:20:14.397 [main] DEBUG nextflow.config.ConfigBuilder - Available config profiles: [slurm, sanger, sp3, conda, oke, singularity, qib, lsf, phw, docker, gls]
Feb-15 21:20:16.471 [main] DEBUG nextflow.k8s.K8sConfig - Kubernetes workDir=/work/ubuntu/work; projectDir=/work/project; volumeClaims=[default-nextflow-fss-storage-data-pvc, default-nextflow-fss-storage-work-pvc]
Feb-15 21:20:16.905 [main] DEBUG nextflow.k8s.K8sDriverLauncher - Created K8s configMap with name: nf-config-c772c303
Feb-15 21:25:42.592 [main] DEBUG nextflow.k8s.K8sDriverLauncher - Wait for pod termination name=infallible-wilson
Feb-15 21:27:12.759 [main] WARN  nextflow.k8s.K8sDriverLauncher - Unable to fetch pod exit status -- pod=infallible-wilson state=[running:[startedAt:2022-02-15T21:20:18Z]]

### Environment 

* Nextflow version: [20.10.0, 21.10.6] 
* Java version: [11]
* Operating system: [macOS, Linux]
* Bash version: (use the command `$SHELL --version`)

### Additional context

Can the following be considered:

[...]

PR #2667, opened 2022-02-22
Force kuberun to wait for pods before shutdown

kfrancisdev:
- wait for pod to stop running before shutdown.
- Patch for Issue #2654


bentsherman:
Have you tested this fix with your nextflow run? Did the workflow wait until the pods were deleted?

kfrancisdev:
> Have you tested this fix with your nextflow run? Did the workflow wait until the pods were deleted?

I tested this fix manually. It waited until the pod was not running. Specifically, it waited until the pod succeeded. The pod still is present after the status is succeeded. So that behavior does not change. I think the pod should be explicitly deleted in shutdown if it's not running

pditommaso:
That sounds reasonable. Please add sign off the commit to make the DCO bot pass https://github.com/nextflow-io/nextflow/pull/2667/checks?check_run_id=5296377084

kfrancisdev:
Sign-off added.

pditommaso:
Thanks a lot!

Issue #2662, opened 2022-02-21
Allow custom configs `params` to be parsed before nextflow.config


mahesh-panchal:
## New feature

I'm not sure whether to classify this as a feature request or bug.
The issue is that `params` configuration from custom configs (`-c`) appears to be read in after nextflow.config. 
The docs describe that `params` provided in a custom config have a higher priority than those in the nextflow.config. This is true at the time of running the workflow, but not when parsing and evaluating the conditions in the Nextflow config.

For example:
`main.nf`:
```nextflow
#! /usr/bin/env nextflow

nextflow.enable.dsl = 2

workflow {
    FOO()
}

process FOO {

    output:
    path 'myfile.txt'

    script:
    """
    touch myfile.txt
    """
}
``` 
`nextflow.config`:
```nextflow
params.outdir = 'results'
process {
    withName: 'FOO' {
        publishDir = "$params.outdir/foo"
    }
}
```

If one runs:
```console
$ nextflow run main.nf --outdir '/my/new/path'
```
or 
```console
$ cat params.yml
outdir : '/my/new/path'
$ nextflow run main.nf -params-file params.yml
```
Then the `publishDir` is correctly set.

However, if one uses
```console
$ cat custom.config
params.outdir = '/my/new/path'
$ nextflow run main.nf -c custom.config
```
Then the `publishDir` is not correct, instead using the default value.

## Usage scenario 

Lot's of Nextflow users currently provide custom `params` via the `-c` option, not realizing it is not applying to configuration in the nextflow.config.

## Suggest implementation 

(Highlight the main building blocks of a possible implementation and/or related components)

abhi18av:
Good point, this situation seems to also affect launches from the `tw` CLI 


```powershell

tw launch $PIPELINE_URL `
    --profile conda_local `
    --params-file params.yml `              <-- Local params file
    --config slurm.config `                 <-- Local config file
    --compute-env $COMPUTE_ENV_ID `

```

mahesh-panchal:
Is this intended to be addressed in the new config implementation?

abhi18av:
@mahesh-panchal , I think that this would also be addressed as per the overall vision. Though I don't have any more visibility at this point

maxulysse:
I managed to overcome this by supplying params in a yml file

mahesh-panchal:
Not everyone knows to do this. Also configs support things the yml and command line do not, e.g. environment variables (not possible in yml), or closures (config specific).

[...]

PR #2673, opened 2022-02-24
Document params-file and improve self-update 

abhi18av:
Updated the documentation for `--params-file` and `self-update`

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -1291,7 +1291,6 @@ Update the nextflow runtime to the latest available version.
 **Description**
 
 The ``self-update`` command directs the ``nextflow`` cli to update itself to the latest stable release. 
-This command also updates the `Nextflow launcher script <https://github.com/nextflow-io/nextflow/blob/master/nextflow>`_ in addition to the Java artifacts, whereas the ``NXF_VER`` environment variables only updates the Java artifacts.
 
 
 **Examples**
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -1265,7 +1265,7 @@ facilitates rapid iterations, inspections of any pipeline as well as debugging.
     $ nextflow run nextflow-io/hello -with-tower
  
 
-- Invoke the nextflow pipeline execution with a custom parameters ``YML`` file. 
+- Invoke the nextflow pipeline execution with a custom parameters ``YAML/JSON`` file. 
 The parameters which are specified through this mechanism are merged with the resolved configuration (base configuration and profiles) and only the common fields are overwritten by the ``YML`` file.::
 
     $ nextflow run main.nf -params-file pipeline_params.yml
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -1291,7 +1291,7 @@ Update the nextflow runtime to the latest available version.
 **Description**
 
 The ``self-update`` command directs the ``nextflow`` cli to update itself to the latest stable release. 
-The ``self-update`` command also updates the `Nextflow launcher script <https://github.com/nextflow-io/nextflow/blob/master/nextflow>`_ in addition to the Java artifacts, whereas the ``NXF_VER`` environment variables only updates the Java artifacts.
+This command also updates the `Nextflow launcher script <https://github.com/nextflow-io/nextflow/blob/master/nextflow>`_ in addition to the Java artifacts, whereas the ``NXF_VER`` environment variables only updates the Java artifacts.
 
 
 **Examples**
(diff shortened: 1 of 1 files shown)


docs/cli.rst
```diff
@@ -1284,7 +1290,8 @@ Update the nextflow runtime to the latest available version.
 
 **Description**
 
-The ``self-update`` command directs the ``nextflow`` cli to update itself to the latest stable release.
+The ``self-update`` command directs the ``nextflow`` cli to update itself to the latest stable release. 
+The ``self-update`` command also updates the `Nextflow launcher script <https://github.com/nextflow-io/nextflow/blob/master/nextflow>`_ in addition to the Java artifacts, whereas the ``NXF_VER`` environment variables only updates the Java artifacts.
```
abhi18av, line 1294:
```suggestion
This command also updates the `Nextflow launcher script <https://github.com/nextflow-io/nextflow/blob/master/nextflow>`_ in addition to the Java artifacts, whereas the ``NXF_VER`` environment variables only updates the Java artifacts.
```

docs/cli.rst
```diff
@@ -1284,7 +1290,8 @@ Update the nextflow runtime to the latest available version.
 
 **Description**
 
-The ``self-update`` command directs the ``nextflow`` cli to update itself to the latest stable release.
+The ``self-update`` command directs the ``nextflow`` cli to update itself to the latest stable release. 
+This command also updates the `Nextflow launcher script <https://github.com/nextflow-io/nextflow/blob/master/nextflow>`_ in addition to the Java artifacts, whereas the ``NXF_VER`` environment variables only updates the Java artifacts.
```
pditommaso, line 1294: Don't think this is correct. The `self-update` is the same of `-self-update` and they do *not* update the launcher

abhi18av: Whoops!

@pditommaso , then let me cofirm my understanding here

1. `-self-update` updates only java artifacts
2. `NXF_VER` updates java artifacts and the launcher script

Therefore when `NXF_EDGE=1 nextflow -self-update`  is used, only the java artifacts are updated to the latest edge version.

Is this correct?

docs/cli.rst
```diff
@@ -1265,6 +1265,12 @@ facilitates rapid iterations, inspections of any pipeline as well as debugging.
     $ nextflow run nextflow-io/hello -with-tower
  
 
+- Invoke the nextflow pipeline execution with a custom parameters ``YML`` file. 
+The parameters which are specified through this mechanism are merged with the resolved configuration (base configuration and profiles) and only the common fields are overwritten by the ``YML`` file.::
```
bentsherman, line 1269: The params file can be YAML or JSON. I would say YAML/JSON file instead of `YML` file.

bentsherman: Also @pditommaso we should clarify what happens if you provide params through a config file and a params file via CLI arguments. I think the params file takes priority but I'm not certain.

abhi18av: Yes, this question has come up recently https://github.com/nextflow-io/nextflow/issues/2662

docs/cli.rst
```diff
@@ -1265,6 +1265,12 @@ facilitates rapid iterations, inspections of any pipeline as well as debugging.
     $ nextflow run nextflow-io/hello -with-tower
  
 
+- Invoke the nextflow pipeline execution with a custom parameters ``YML`` file.
```
abhi18av, line 1268:
```suggestion
- Invoke the nextflow pipeline execution with a custom parameters ``YAML/JSON`` file. 
```

docs/cli.rst
abhi18av, line 1269:
```suggestion
The parameters which are specified through this mechanism are merged with the resolved configuration (base configuration and profiles) and only the common fields are overwritten by the ``YAML/JSON`` file.::
```

abhi18av (commented):
Accomodate Ben's review about the JSON optoin.

docs/cli.rst
pditommaso: Actually I was wrong here. `self-update` and `-self-update` update both nextflow runtime and the launcher script. 


`NXF_VER` is meant to enforce the use of a specific nf version. Don't think it should be mentioned in this context.

pditommaso (approved):
Ok, updated it removing the refence to the NXF_VER. Eventually it could be added into a *tip* section

docs/cli.rst
pditommaso: Was missing this conversation. The CLI params overrides the ones the json/yaml file.

You can see it here, it reads the params file and put it into the `result` map. then if there are cli `params` copy them over the `result` map. 

https://github.com/nextflow-io/nextflow/blob/0f63f1a7cc7953ed1d014e8d3f821ba2941c6177/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy#L496-L513

PR #2690, opened 2022-03-02
bugfix: allow blank spaces in java path

### LINKED ISSUE #2298: Big Sur installation fails
Hi

I get these errors after issuing

`curl -s https://get.nextflow.io | bash`

on a Mac Book pro running Big Sur

`
...
.../DCBLab/nextflow: line 175: /Library/Internet: No such file or directory
.../DCBLab/nextflow: line 175: exec /Library/Internet: cannot execute: No such file or directory
`

Any ideas about what could be wrong?

Cheers

Marco

PS Please: do not tell me it is the spaces in the pathname.

jorgeaguileraseqera:
resolve #2298

the idea is to extract the first param between double quotes as the java path using a regexp 

### DIFF
--- a/nextflow
+++ b/nextflow
@@ -396,14 +396,14 @@ else
 
     # first string between double quotes is the full path to java, also blank spaces are included
     # remainder string are arguments
-    # we extract first part into cmd and remainder into rest and convert them to array as previous version
-    pat='"([^"]*)"(.*)'
-    [[ "${cli[@]}" =~ $pat ]]
-    cmd=(${BASH_REMATCH[1]})
-    rest=(${BASH_REMATCH[2]})
+    # we extract first part into `cmd_base`` and remainder into `cmd_tail`` and convert them to array as previous version
+    cmd_pattern='"([^"]*)"(.*)'
+    [[ "${cli[@]}" =~ $cmd_pattern ]]
+    cmd_base=(${BASH_REMATCH[1]})
+    cmd_tail=(${BASH_REMATCH[2]})
 
     if [[ "$JAVA_VER" =~ ^(9|10|11|12|13|14|15|16|17) ]]; then
-      launcher="${cmd[@]}"
+      launcher="${cmd_base[@]}"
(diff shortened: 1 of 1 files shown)


nextflow
```diff
@@ -394,8 +394,13 @@ else
     cli=($("$JAVA_CMD" "${JAVA_OPTS[@]}" -jar "$NXF_BIN"))
     [[ $? -ne 0 ]] && echo_red 'Unable to initialize nextflow environment' && exit 1
 
+    pat='"([^"]*)"(.*)'
+    [[ "${cli[@]}" =~ $pat ]]
+    cmd=(${BASH_REMATCH[1]})
+    rest=(${BASH_REMATCH[2]})
+
     if [[ "$JAVA_VER" =~ ^(9|10|11|12|13|14|15|16|17) ]]; then
```
pditommaso, line 402: We have a new Bash guru! 😄

pditommaso: Kidding apart what's the logic of the snippet above?

jorgeaguileraseqera: yes, sorry for the lack of explanations

main problem comes because `cli=($("$JAVA_CMD" "${JAVA_OPTS[@]}" -jar "$NXF_BIN"))` split all the command in an array but bash uses the blank as separator so `launcher=("${cli[@]:0:1}")` assign an incomplete path 

the idea is to apply a regexp and extract the element between double quotes into `cmd` var and the rest into the `rest`var and after we can create the array as doing previously

pditommaso: That's smart, let's add this explanation in the script as a comment. thanks!

jorgeaguileraseqera: done, let me know if it's enough or need more explanation

pditommaso:
Looks nice, just renamed some vars to avoid overlapping with other variables and make it a bit more self-explanatory. Please check it makes sense

jorgeaguileraseqera:
Yes, better these names and avoid overlapping

PR #2708, opened 2022-03-09
bugfix: use first config in stack when is into profile

### LINKED ISSUE #2701: params evaluation in config not working as expected
Finally sat down to figure out why viralrecon is failing to run with the latest edge release of NF. It comes down to an issue with params not being resolved properly in the Nextflow configuration.

I have created a minimal reproducible example.

### `main.nf`

```nextflow
#!/usr/bin/env nextflow

nextflow.enable.dsl = 2
```

### `nextflow.config`

```nextflow
params {
    load_config = null
}

profiles {
    test { includeConfig 'test.config' }
}

if (params.load_config) {
    includeConfig 'process.config'
}
```

### `test.config`

```nextflow
params {
    load_config = true
}
```

### `process.config`

```nextflow
process {
    withName: FOO {
        ext.args = '--quiet'
    }
}
```

### Everything works with `21.10.6`

If we run the latest stable version of Nextflow you will notice that all of the configuration is loaded as expected including the process selector for the `FOO` process:

```console
$ NXF_VER=21.10.6 nextflow config -profile test

params {
   load_config = true
}

process {
   withName:FOO {
      ext {
         args = '--quiet'
      }
   }
}
```

### Doesn't work with `22.02.1-edge`

If we run the latest edge version of Nextflow you will notice that all of the configuration is not loaded as expected and the process selector for the `FOO` process is missing:

```console
$ NXF_VER=22.02.1-edge nextflow config -profile test

params {
   load_config = true
}
```

If I change `nextflow.config` like below then it works again:

```nextflow
params {
    load_config = null
}

profiles {
    test { includeConfig 'test.config' }
}

//if (params.load_config) {     Commenting out the if logic fixes it which indicates that params.load_config isn't being evaluated properly in the configuration
    includeConfig 'process.config'
//}
```

jorgeaguileraseqera:
resolve #2701

the problem comes because we are using always `last` config in the stack but when we are starting to parse the config into a profile we need to use the `first` as it's the main config in the tree

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/config/ConfigParser.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/config/ConfigParser.groovy
@@ -322,6 +322,8 @@ class ConfigParser {
         LinkedList stack = new LinkedList()
         LinkedList profileStack = new LinkedList()
         stack << [config: config, scope: [:]]
+        boolean intoProfile = false
+
         def pushStack = { co ->
             stack << [config: co, scope: stack.last.scope.clone()]
         }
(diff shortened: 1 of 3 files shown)


pditommaso:
The failing test is a new one or a previous existing test? 

```
ConfigBuilderTest > should merge params two profiles FAILED
    nextflow.exception.ConfigParseException at ConfigBuilderTest.groovy:1817

ConfigBuilderTest > should merge profiles with conditions FAILED
    groovy.lang.MissingMethodException at ConfigBuilderTest.groovy:2056
```

jorgeaguileraseqera:
One new and another one previous

my fault I was running the check task in local without error but ... in plugin module!!! 
I'm on it

jorgeaguileraseqera:
Seems a lot of changes but it's because I reorganized a little the `invokeMethod` closure

The idea of this PR is to detect when we are entering into a `profile` section and use the top stack instead of the last as the profile section works starting "from the begin" 

this is done when detecting we are entering into a `conditionalValues`:

```
if( name == 'profiles'){
    intoProfile=true
}
```

so we can decide what stack to use:

`def current = intoProfile ? stack.first : stack.last`

Also, once we are parsed a cloned config we need to merge or include in the stack to no forget values

```
if (current.scope.get(name) instanceof ConfigObject) {
    current.scope.get(name).merge(co)
}else{
    current.scope.put(name,co)
}
```

I've tested using nf-core:

`NXF_VER=22.02.1-edge nextflow run nf-core/viralrecon -profile test,docker --outdir ./results`

With current version process abort with an error but ends in 10mnts with this branch

pditommaso:
> Seems a lot of changes but it's because I reorganized a little the invokeMethod closure


This is brilliant but it would be better to try to keep changes as little as possible, otherwise, it's very difficult to trace back change for future troubleshooting 😄

modules/nextflow/src/main/groovy/nextflow/config/ConfigParser.groovy
```diff
@@ -322,6 +322,8 @@ class ConfigParser {
         LinkedList stack = new LinkedList()
         LinkedList profileStack = new LinkedList()
         stack << [config: config, scope: [:]]
+        boolean intoProfile = false
```
pditommaso, line 325: here `into` stands for *within* a `profile` block right?

jorgeaguileraseqera: yes

pditommaso: cool, i'll change to `withinProfile` to make it a bit more explicit

pditommaso:
Ok, let's merge and cross the fingers 🤞

Issue #2721, opened 2022-03-14
Watchpath will not trigger when the glob is early in the pattern


robsyme:
### Expected behavior and actual behavior

I would expect that we can introduce the glob "*" in a watchPath pattern at a location other than the end of the path.

For example, `Channel.watchPath("watch/*/outputs/*.txt")` where the first glob does not appear at the final directory (the `outputs` directory appears afterwards) will not detect matching events.

### Steps to reproduce the problem

In main.nf:

```groovy
// This will work perfectly
Channel.watchPath("watch/projectA/*/*.txt").view { "Glob at end of pattern: $it" }

// This will not detect any events
Channel.watchPath("watch/*/outputs/*.txt").view { "Glob before end of pattern: $it" }
```

Set up the directories, run the workflow:

```bash
mkdir -p watch/project{A,B,C}/outputs
nextflow run .
```
... and in a separate shell create three files that match the second pattern.

```bash
# ABC permuted to CAB to show it is not an ordering problem
touch watch/project{C,A,B}/outputs/test.txt
```

But only the `projectA/outputs/test.txt` (from the first `watchPath`) is detected:

```
Launching `./main.nf` [gloomy_cray] - revision: 76651a7ada
Glob at end of pattern: <basedir>/watch/projectA/outputs/test.txt
```

The most relevant lines from the `.nextflow.log` are the `DirWatcher` logs:

```
Mar.-14 11:11:30.436 [main] DEBUG nextflow.file.DirWatcher - Watch service for path=watch/projectA/; syntax=glob; pattern=*/*.txt; skipHidden=true; events=create
Mar.-14 11:11:30.518 [main] DEBUG nextflow.file.DirWatcher - Watch service for path=watch/; syntax=glob; pattern=*/outputs/*.txt; skipHidden=true; events=create
```

### Environment 

* Nextflow version: 21.10.6
* Java version: openjdk 11.0.2 2019-01-15
* Operating system: macOS, Linux
* Bash version: (use the command `zsh 5.8 (x86_64-apple-darwin21.0)` and `GNU bash, version 4.2.46`)

### Additional context

I raised this in the Nextflow Slack channel. Paolo offered that it might be a shared filesystem problem, but I've excluded that by running in other filesystems and looking closely at the `DirWatcher` code.

### Specifics

In this section I'm only considering `Channel.watchPath("watch/*/outputs/*.txt")` - the failing case.

[...]

PR #2729, opened 2022-03-15
Dirwatch deeper

robsyme:
A PR to address https://github.com/nextflow-io/nextflow/issues/2721. 

Three main changes to `DirWatcher` class:

**Allow more complex paths**
The first (and motivating) change was to increase the flexibility of the `Channel.watchPath` pattern to allow users to specify known directory names after the glob. That is, paths like `projects/*/outputs/*.done`

**Don't close the channel early**
The second change is to change the behaviour of the watchPath channel closing. By default if any of the watched directories are deleted, the Channel closes. In the example above, if the folder `projects/human` is deleted, then the channel closes, and any files in `projects/mouse/outputs/*.done` will be missed. The changed includes a check to only close the channel if the base folder is removed (`projects/`).

**Don't accidentally miss files**
The last change is to ensure that if a file is quickly created in a newly created directory, these files are added to the Channel, even if the file is created before Nextflow has had a change to attach a listener object to that directory. In essence, if Nextflow is scanning a newly created directory (looking for child dirs to register watchers) and comes across a file that matches the pattern, it is assumed to be a newly created file and emitted in the channel.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/file/DirWatcher.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/file/DirWatcher.groovy
@@ -58,7 +58,7 @@ class DirWatcher implements DirListener {
 
     private WatchEvent.Kind<Path>[] watchEvents
 
-    private PathMatcher dirMatcher
+    private List<PathMatcher> dirMatchers
 
     private PathMatcher fileMatcher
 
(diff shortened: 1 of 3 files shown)


pditommaso:
I see it now, I think it's going be the expected functionality. Thinking also it would not be better to handle multiple pattern at script level and then `mix`ing resulting channel into a single one

robsyme:
I thought about the mixing for my particular use case, but of course not everybody will know ahead of time what those directories might be. The glob should now handle them all neatly.

robsyme:
I've noticed that the ordering of events on the tests is important and I might need to make some changes. I'd recommend holding off review for the moment while I push some changes.

robsyme:
This should be ok now.

In fixing the PR, I noticed another small edge case worth considering, and this PR also includes that fix.

**The problem:** 

Registering watchers is not instantaneous, which can lead to the `DirWatcher` missing newly created files. If a series of directories and a file is created in quick succession:

```
mkdir -p test/a/b
touch test/a/b/one.txt
```

... and we have a watcher looking at `test/*/*/*.txt`, then it is possible that `one.txt` is created before the DirWatcher has time to register a listener on the `b/` directory.

When a new directory is created, DirWatcher traverses down the tree looking for directories to register using a `SimpleFileVisitor`. Previously, the visitor only considered for new directories (matching the pattern) to register via the `preVisitDirectory` method. Even though the directories being traversed are brand new, the `SimpleFileVisitor` did not consider any files there. If the directory is newly created, any files in that directory must also be newly created and should be  emitted by the channel. 

**The fix:**

I added a `visitFile` method to the `SimpleFileVisitor` that checks for new files that match the pattern and emits them iff `watchEvents.contains(ENTRY_CREATE)`.

pditommaso:
Hi Rob, I could not look into this before. However, one important thing is that the DirWatcher logic has been re-implemented using a different strategy to support file system for which notify events were not available.

https://github.com/nextflow-io/nextflow/blob/0af1bcb32d3dc8e01b14b7999b1d9f480a05b62a/modules/nextflow/src/main/groovy/nextflow/file/DirWatcherV2.groovy#L43-L43

Therefore this PR should be updated

robsyme:
No problem. I'm away from my desk this week (in Australia) but will make
the updates next week.

robsyme:
The new polling-based system address the previous problem without any extra changes.

The PR includes the changes to the original DirWatcher class, and also extra methods for DirWatcherV2Test that check the behaviour, but of course these are now totally optional. 

If you'd prefer to avoid changing DirWatcher or if you think that the extra tests are unnecessary, that's ok - let me know and I can delete the PR and branch.

robsyme:
The tradeoff for `DirWatcherV2` is that the apache `FileAlterationMonitor` has to scan the directories before reporting events. In directories with a large number of files, this can mean long delays before events are captured. For this reason, it might be nice to include the changes to `DirWatcher` which is available under the `NXF_DIRWATCHER_LEGACY` env flag.

pditommaso:
Rob, thanks for this, but I believe it's preferable to keep the current basic syntax. It may be simple to support other storage backends in the future.

PR #2731, opened 2022-03-16
Include SCHEMA in toString output

mh11:
The toString is often used to define subpaths / files e.g. file("${myS3dir}/important/file.txt")
When myS3dir is a S3Path (e.g. s3://my/s3/path/to/the/,  the output currently result in a local path and the pipeilne breaks e.g. "/my/s3/path/to/the/important/file.txt" 

Workaround:
The current workaround requires to use the "toUri()" method to be called for a complete definition of the resource e.g. file("${myS3dir.toUri()}/important/file.txt")

Solution:
The proposed solution returns by default a valid string representation of an URI, is more specific (tracking down issues) and allows the simple as well as consistent use of a "Path" variable. 
Also implemented the same way here:
https://github.com/Upplication/Amazon-S3-FileSystem-NIO2/blob/4fe1a64942cd565bb871a4e8c5add19687313f6b/src/main/java/com/upplication/s3fs/S3Path.java#L555

### DIFF
--- a/plugins/nf-amazon/src/main/com/upplication/s3fs/S3Path.java
+++ b/plugins/nf-amazon/src/main/com/upplication/s3fs/S3Path.java
@@ -481,17 +481,7 @@ public class S3Path implements Path, TagAwareFile {
 
 	@Override
 	public String toString() {
-		StringBuilder builder = new StringBuilder();
-
-		if (isAbsolute()) {
-			builder.append(PATH_SEPARATOR);
-			builder.append(bucket);
-			builder.append(PATH_SEPARATOR);
-		}
-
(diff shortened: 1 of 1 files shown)


pditommaso:
I wish I could merge this, but the change you are proposing can break any pipeline relying on the current behaviour. 

Also, it will introduce an inconsistency between S3 paths and other file system providers e.g. Google Storage and Azure Blob containers, that follow the same criteria. 

Unfortunately, the original sin was done in the [Java Path API](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#toString--) that's used under the hood. 

Think we should add a disclaimer in the documentation to clearly highlight this behaviour. 

Closing this.

Issue #2723, opened 2022-03-14
Evolution of Nextflow configuration file


pditommaso:
### Rationale 

The configuration file is a key component of the Nextflow framework since it allows workflow developers to decouple the pipeline logic from the execution parameters and infrastructure deployment settings. 

The current configuration mechanism is extremely flexible and powerful. It uses a customised [GroovySlurper](http://docs.groovy-lang.org/next/html/gapi/groovy/json/JsonSlurper.html) class with extended to handle multiple profiles and config file inclusions. 

A Nextflow configuration file is ultimately a compiled Java class on steroids.  

However, this has serious drawbacks: 

1. The evaluation of config file, requires Java class compilation and interpretation, which can be time-consuming compared to text file parsing. 
2. The evaluation of the config requires ultimately the execution of JVM code, which could be explored by malicious users and pose a non-trivial security burden to be handled in a managed environment. 
3. The current configuration allows the use of arbitrary imperative programming statements within a config file, which is a bad practice, making the config logic poorly readable.
4. The parser implementation is becoming including hard to be maintained, and patches easily introduce unexpected side-effects, resulting in increased stability issues. 

### Goals 

* Explore a new configuration mechanism that can be used in alternative to the current nextflow config 
* Have a well-defined structure and grammar 
* Human-readable 
* Allow grouping correlated setting 
* Allow the definition of named groups of settings (i.e. *profiles*)
* Allow inheritance 
* Allow config files inclusion  

### Non-goals 

* Compatible with current nextlow.config syntax 
* Allowing custom functions
 

### Action

This issue should explore possible alternative implementation and define a migration strategy. 

Currently the best alternatives are:

* [Hashicorp HCL](https://github.com/hashicorp/hcl): this is essentially the configuration language used by Terraform and other Hashicorp tools. It's quite popular, however there isn't a complete parsed for JVM world, other than [this project](https://github.com/bertramdev/hcl4j).
* [Ligthbed HCON](https://github.com/lightbend/config): this is essentially, very curated configuration library. The syntax has a lot of similarities with the current nextflow.config syntax.

jorgeaguileraseqera:
I think the problem comes from the current implementation because we are overwriting the current ConfigSlurper trying to do more things at the same time instead of first parsing the configuration and after applying them

[...]

PR #2747, opened 2022-03-23
Rename pod directive to podOptions

bentsherman:
This PR renames the `pod` directive to `podOptions`, for consistency with `clusterOptions`, `containerOptions`, etc. Nextflow will still accept `pod` but will print a deprecation warning.


pditommaso:
I will the idea, I'd resist making this change only for naming uniformity. I guess there's a lot of config out there that would be impacted and can be annoying to update them. 

I would keep this possibility if at some point there's need to change the pod configuration structure or keep in sync with the config re-engineering #2723

bentsherman:
To be clear, this PR only deprecates the `pod` option, so `pod` is still accepted but users get a deprecation warning. Anyway, this PR is here if you want to merge it. Although I left it as a draft because I wanted to make sure I did the deprecation correctly. Still need to test locally.

pditommaso:
Understand that, but still think we can keep this for when it may be needed to re-organise the K8s config. Also considering the discussion to use jobs instead of pods #2751, does still make sense to use `podOptions`? maybe it may be better turn into `k8sOptions`. 

Closing this for now.

bentsherman:
I think `podOptions` still makes sense because it refers to the pod-specific options, whether the pod is standalone or part of a job. `k8sOptions` would be confusing with the `k8s` config options, we would want to rethink how the k8s options are organized overall.

Issue #1162, opened 2019-05-28
Submit processes as Jobs instead of Pods for k8s / kuberun


bentsherman:
## New feature

One of my main concerns with the k8s executor and `kuberun` right now is that they don't always clean up after themselves. If a process fails, the pod is left hanging around, and `kuberun` also doesn't clean up the ConfigMaps and pods that it creates, regardless of whether they fail or complete successfully.

The problem is that pods are mapped to physical nodes, so if they aren't cleaned up then they take up physical resources even if they just sit there. I hope that Kubernetes improves it's support for batch processing in the future, but for now I think the best solution for nextflow would be to __submit a process as a Job with 1 completion rather than a Pod__. The job will automatically create a pod with the correct spec and delete it when it completes or fails. Similarly, `nextflow kuberun` should submit the controller pod as a Job with 1 completion.

## Usage scenario 

Pods that are submitted by themselves rather than by a controller (aka Job or Deployment) are hard to control in a shared batch processing system. Therefore the sys-admins of my k8s cluster are restricting such standalone pods to 2 CPUs / 8 GB memory / 6 hour wall-time, which is much less than what I need for my nextflow pipelines.

This may seem like a very particular use case but standalone pods are a major issue because they consume physical resources but there is no mechanism to clean them up easily without using a controller. A Job, on the other hand, does not consume resources by itself, and it will automatically manage the creation and deletion of pods.

## Suggest implementation 

Example of an existing pod spec:
```
apiVersion: v1
kind: Pod
spec:
  containers:
  - ...
```

Example of wrapping the pod spec in a job:
```
apiVersion: batch/v1
kind: Job
spec:
  completions: 1
  activeDeadlineSeconds: <value from time directive>
  ttlSecondsAfterFinished: 100
  template:
    spec:
      containers:
      - ...
```

pditommaso:
This is an interesting pod. Unfortunately, we don't have resources in the short term to spend on this. A pull request would be welcomed.

bentsherman:
Added some relevant fields to the job spec. Information about Jobs can be found here:

https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/

bentsherman:
As far as code changes go, the k8s interface will simply create and delete jobs instead of pods. But you must still interact directly with the pod for things like checking status and streaming logs, so you will have to go through the job to do that.

[...]

PR #2751, opened 2022-03-24
Initial version of K8s Jobs

xhejtman:
Preview of initial version of K8s Jobs. Feel free to comment what to change and rework. Tests were not updated yet and not all error cases are managed right now.


pditommaso:
What's the benefit of using jobs in place of pods?

bentsherman:
Resolves #1162

pditommaso:
I agree it's useful to support jobs in place of pods, however, this could impact existing workloads in a non-trivial manner, therefore it should be implemented as an opt-in feature that could be specified in the `k8s` scope. 

Also, this major naming reaction can be distributive for users that may rely on those classes (it's true that's expected to be internal APIs but I guess out there there are forks that use those). 

Keeping part the rename of the classes it looks the required changes are minimal. Likely these

it should be given as an opt-in feature at least initially 

https://github.com/nextflow-io/nextflow/pull/2751/files#diff-780bb5e8895db9ec52a37f703e4c49dd9377c1208d9b36a33c1d0ed638dc0b34L351-R373

https://github.com/nextflow-io/nextflow/pull/2751/files#diff-780bb5e8895db9ec52a37f703e4c49dd9377c1208d9b36a33c1d0ed638dc0b34L351-R373

and a few others

pditommaso:
Yes, I think a first step could be to keep the same interfaces and just make the adaptation in the relevant places to create a job instead of a pod a config setting is provided

bentsherman (changes requested):
I missed the above conversation while writing this review but turned out to have similar comments to Paolo so I'll leave it

xhejtman:
@pditommaso jobs could possible solve an issue if you have large queue like 400 or even 1600, nextflow has some races that pod can disappear sooner than it collect exit status and it ends with:
```
[dd/8506e0] NOTE: Unable to find pod nf-dd8506e0843c3d7fcaf1c418ec32acbd - The pod may be evicted by a node shutdown event. Throwing NodeTerminationException -- Execution is retried
```

but the pod `nf-dd8506e0843c3d7fcaf1c418ec32acbd` actually finished correctly. There are many more that ends like this. I think it should be solved running as jobs as job object will not disappear until deleted.

Btw, it does not support limit of how much cpu/memory can particular workflow consume. It is by design? I think this would be more usefull than limiting queue size. In K8s, this can be set by namespace resource limit, but only after jobs are implemented, pods do not work with quotas. Not sure about other executors.

pditommaso:
> jobs could possible solve an issue if you have large queue like 400 or even 1600, nextflow has some races that pod can disappear sooner than it collect exit status and it ends

I agree, that it would better to sue jobs 

> Btw, it does not support limit of how much cpu/memory can particular workflow consume. It is by design? I think this would be more usefull than limiting queue size. In K8s, this can be set by namespace resource limit, but only after jobs are implemented, pods do not work with quotas. Not sure about other executors

Let's open a separate issue to discuss your proposal

modules/nextflow/src/main/groovy/nextflow/cli/CmdKubeRun.groovy
```diff
@@ -44,13 +44,13 @@ class CmdKubeRun extends CmdRun {
     @Parameter(names = ['-n','-namespace'], description = 'Specify the K8s namespace to use')
     String namespace
 
-    @Parameter(names = '-pod-image', description = 'Specify the container image for the Nextflow pod')
-    String podImage
+    @Parameter(names = '-pod-image', description = 'Specify the container image for the Nextflow job')
+    String jobImage
```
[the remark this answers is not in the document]
xhejtman: should we create PR for `-pod-image` -> `-head-image`? Or perhaps support both for a while as it is breaking change?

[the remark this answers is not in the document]
pditommaso: I don't have a strong preference for this name. My thoughts: it should go into a separate PR, if the name change, the old should be deprecated but kept for backward compatibility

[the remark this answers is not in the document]
bentsherman: I can take care of that. For this PR just keep the variable as `podImage`

xhejtman:
huh, can I reopen it? I pushed new reworked version

xhejtman:
Hello, @romanaduraciova and me created a new version of jobs. It is parmetrized via `nextflow.config`:
```
k8s {
  jobs = true
}
```

default is false, i.e., use pods.

modules/nextflow/src/main/groovy/nextflow/k8s/K8sDriverLauncher.groovy
```diff
@@ -49,6 +50,7 @@ import org.codehaus.groovy.runtime.MethodClosure
 @Slf4j
 class K8sDriverLauncher {
 
+    private String deploymentName = "Pod"
```
bentsherman, line 53: `deploymentName` is confusing because deployment is another type of k8s resource. Perhaps `resourceType` instead? Also throughout the rest of the code

modules/nextflow/src/main/groovy/nextflow/k8s/K8sDriverLauncher.groovy
```diff
@@ -161,22 +163,30 @@ class K8sDriverLauncher {
     protected void waitPodEnd() {
```
bentsherman, line 163: Should be `waitJobEnd()` now?

bentsherman (changes requested):

bentsherman:
Looks good so far. Just need to update the unit tests. I will test on my end when I have some time.

modules/nextflow/src/main/groovy/nextflow/k8s/K8sDriverLauncher.groovy
xhejtman: ok, changed

bentsherman:
Basic test on my end worked. We should try to make sure that it actually solves your scaling issues as well.

modules/nextflow/src/main/groovy/nextflow/k8s/K8sConfig.groovy
```diff
@@ -140,6 +140,8 @@ class K8sConfig implements Map<String,Object> {
 
     String getNamespace() { target.namespace }
 
+    boolean getJob() { target.job }
```
bentsherman, line 143: Can we rename the config option to `jobs` or `useJobs` so that it isn't confused with `pod` options? Also make sure to document the new option in `docs/config.rst`.

xhejtman: job and jobs are too similiar, so `useJobs` then?

bentsherman: sure

xhejtman:
We tested namespace quotas, everything seems to be ok - jobs are pending until free resources. We tested rebooting a node, it is ok except one case. If you have a cluster that has only a single controlplane node that is rebooted, then workflow fails as pods end in unknown state. This is related to K8s itself not nextflow and this situation should be rare, for highavailability, there should always be minimum 3 control plane nodes and in this case, it should be ok. We did not test spawning ~1000 processes to check, whether there is still a race condition or not.

modules/nextflow/src/test/groovy/nextflow/k8s/K8sDriverLauncherTest.groovy
```diff
@@ -472,6 +538,7 @@ class K8sDriverLauncherTest extends Specification {
         config.k8s.storageMountPath == '/this'
         config.k8s.pod == null
         and:
+        and:
```
bentsherman, line 541: looks like an extraneous `and:`

bentsherman (approved):

bentsherman:
@pditommaso ready to merge

xhejtman:
We fixed backoffLimit default value. It was set to `null` instead of 0.

pditommaso:
Was thinking that since the `Job` and `Pod` are almost identical, would not make more sense to use the PodSpecBuilder for that instead of mostly duplicating all the code?

xhejtman:
> Was thinking that since the `Job` and `Pod` are almost identical, would not make more sense to use the PodSpecBuilder for that instead of mostly duplicating all the code?

what do you mean? Code of JobSpecBuilder utilizes PodSpecBuilder for building Pod template, so the actual code of JobSpecBuilder is pretty small. Or do you mean something else?

pditommaso:
Oops, I was missing that. Then it sounds good!

modules/nextflow/src/main/groovy/nextflow/k8s/K8sDriverLauncher.groovy
```diff
@@ -137,24 +139,24 @@ class K8sDriverLauncher {
         this.k8sClient = makeK8sClient(k8sConfig)
         this.k8sConfig.checkStorageAndPaths(k8sClient)
         createK8sConfigMap()
-        createK8sLauncherPod()
-        waitPodStart()
+        createK8sLauncherJob()
+        waitJobStart()
```
pditommaso, line 143: Can we keep the *Pod* naming, since pod is still the default resource?

modules/nextflow/src/main/groovy/nextflow/k8s/model/JobSpecBuilder.groovy
```diff
@@ -0,0 +1,86 @@
...
+
+package nextflow.k8s.model
+
+import groovy.transform.CompileStatic
+import groovy.util.logging.Slf4j
+
+/**
+ * Object build for a K8s job specification
+ *
+ * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
```
pditommaso, line 26: I'm not the author of this :smile:

modules/nextflow/src/main/groovy/nextflow/k8s/model/JobSpecBuilder.groovy
```diff
@@ -0,0 +1,86 @@
...
+@Slf4j
+class JobSpecBuilder {
+
+    String jobName
+
+    String namespace
+
+    String workDir
+
+    int backoffLimit = -1
```
pditommaso, line 38: It would be better to use `Integer` and check for `null` instead of `-1` as absent value.

modules/nextflow/src/main/groovy/nextflow/k8s/model/JobSpecBuilder.groovy
```diff
@@ -0,0 +1,86 @@
...
+ */
+@CompileStatic
+@Slf4j
+class JobSpecBuilder {
+
+    String jobName
+
+    String namespace
+
+    String workDir
```
pditommaso, line 36: We may need to review the need to use a workDir. This has been recently removed for the pod execution. See https://github.com/nextflow-io/nextflow/commit/dc8bfeadc46f82eaadb192833f54efcb6b5bc582

modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
```diff
@@ -307,15 +314,19 @@ class PodSpecBuilder {
...
 
         final metadata = new LinkedHashMap<String,Object>()
-        metadata.name = podName
-        metadata.namespace = namespace ?: 'default'
+        // if podName -> create pod and need podName and namespace, 
+        //  else creating job without podName and namespace
+        if( this.podName ) {
+            metadata.name = podName
+            metadata.namespace = namespace ?: 'default'
+        }
```
pditommaso, line 329: What the logic behind this? I mean why for the Job should be different ?

pditommaso (changes requested):

pditommaso:
I left some comments to address. In general, let's avoid renaming all methods `xxxPod` to `xxxJob` e.g. `createK8sLauncherPod` to `createK8sLauncherJob` since the pod resource is still going to be default one.

pditommaso:
I left some comments to address. In general, let's avoid renaming all methods `xxxPod` to `xxxJob` e.g. `createK8sLauncherPod` to `createK8sLauncherJob` since the pod resource is still going to be default one.

modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
xhejtman: You should set namespace and name only to job part of the object, not pod object. For job case, we do not set podName, thus not filling pod metadata for job, these attributes seem to be ignored at all.

modules/nextflow/src/main/groovy/nextflow/k8s/model/JobSpecBuilder.groovy
xhejtman: Ok, we can remove workDir at all.

modules/nextflow/src/main/groovy/nextflow/k8s/K8sDriverLauncher.groovy
xhejtman: Well, shoud not be using job the target status? We can refactor back to pod, but once jobs will be default, we shall refactor back to job?

modules/nextflow/src/main/groovy/nextflow/k8s/model/JobSpecBuilder.groovy
xhejtman: fair enough ;)

modules/nextflow/src/main/groovy/nextflow/k8s/model/JobSpecBuilder.groovy
xhejtman: Can you give an advice how to distinguish between `null` and `0` for `Integer` object?

modules/nextflow/src/main/groovy/nextflow/k8s/K8sDriverLauncher.groovy
pditommaso: > but once jobs will be default, we shall refactor back to job

Possibly, or even better in the future the pod ones will be deprecated and a new version with `Job` in the name it will be added. 

When doing this kind of change it's always important to not break compatibility with projects that can rely on the nextflow codebase.

modules/nextflow/src/main/groovy/nextflow/k8s/model/JobSpecBuilder.groovy
pditommaso: Having `Integer backoffLimit` you can check for nullability with `backoffLimit == null`

modules/nextflow/src/main/groovy/nextflow/k8s/model/JobSpecBuilder.groovy
xhejtman: fixed

modules/nextflow/src/main/groovy/nextflow/k8s/model/JobSpecBuilder.groovy
xhejtman: fixed

modules/nextflow/src/main/groovy/nextflow/k8s/K8sDriverLauncher.groovy
xhejtman: so ok, refactored to `Pod`

modules/nextflow/src/main/groovy/nextflow/k8s/model/JobSpecBuilder.groovy
xhejtman: at first, we wanted to keep default behaviour of jobs. If backofflimit is not set, it has a default value of 6. However, as restarting pods in place is not desired modus operandi of nextflow, we make it simplier just always setting backoffLimit to 0 unless explicitly specified by withJobBackoffLimit() method. Is this approach ok?

pditommaso: If so, I agree, that it would be better to default to `0` and delegate the retry to nextflow

pditommaso:
I would like to propose some improvements to simplify this implementation, however, I've noticed there is any unit tests for `JobSpecBuilder` class. At least a few should be added to be sure the expected Job spec is created.

modules/nextflow/src/main/groovy/nextflow/k8s/model/JobSpecBuilder.groovy
xhejtman: ok, resolved then.

pditommaso:
Generally it's expected a `XxxTest` unit test class for each class `Xxx` in the code, at least for main method(s)

modules/nextflow/src/test/groovy/nextflow/k8s/K8sDriverLauncherTest.groovy
```diff
@@ -192,6 +192,73 @@ class K8sDriverLauncherTest extends Specification {
...
+            apiVersion: 'batch/v1', 
+            kind: 'Job', 
+            metadata: [name: 'foo-boo', namespace: 'foo'], 
+            spec: [
+               backoffLimit: 0,
+               template: [
+                  apiVersion: 'v1',
+                  kind: 'Pod',
+                  metadata: [labels:[app:'nextflow', runName:'foo-boo']],
+                  spec: [
```
pditommaso, line 232: Looking at the `Job` [spec](https://kubernetes.io/docs/concepts/workloads/controllers/job/) it looks those attributes should not be added. 

I'm going to submit a patch to improve it

pditommaso:
I think this implementation can be significantly simplified using the patch below attached below.

In a nutshell: 
* just use `PodSpecBuilder` adding `buildAsJob` method, which created a pod instead of the Pod resource 
* sanitize the `job.spec.template.spec` removing nested `kind` and `metadata` attributed taken from the pod  
* just default `backoffLimit` to zero. I don't see a concrete need to set it explicitly. Is there? 

I've updated `K8sTaskHandler` and `K8sTaskHandlerTest` accordingly. I didn't do however for `K8sDriverLauncher` and `K8sDriverLauncherTest` classes 


You can apply this patch using `git apply pod-spec-patch.txt` in your working branch

[pod-spec-patch.txt](https://github.com/nextflow-io/nextflow/files/8743470/pod-spec-patch.txt)

Let me know what do you think

xhejtman:
so jobspecbuilder class is not needed anymore? I will test and let you know.

pditommaso:
Essentially no, I've refactored it as `buidAsJob` method into `PodSpecBuilder`. It helps to keep the overall logic more straightforward

xhejtman:
Shall I merge this patch and update PR with commit&push? I found also minor issue with job/pod delete that I fixed.

pditommaso:
That would be good.

xhejtman:
OK, wilco. 

Btw, I might find the reason, why with high number of pods (like 1600) execution fails. It seems that if nodes are like overloaded by pods, K8s starts to remove completed pods. Which is something that nextflow does not expect that a completed  pod may disappear. It was also a problem for the Jobs implementation so I am working on improvement - i.e., missing pod does not always mean that it not started yet, but it also could be removed already. In this case, however, we have info in the Job object as succeeded=1, completed, true. We only need to simulate pod status in this case. I believe, that succeeded=1, completed, true means that exit status was 0. and from job status, we have start/completion time. only container id is missing for proper pod status.

pditommaso:
Let's keep this PR only of Job submit capability and keep other improvements in separate PRs.

xhejtman:
Job status handling is still not correct, will send fix

modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy
```diff
@@ -343,21 +357,13 @@ class K8sTaskHandler extends TaskHandler {
...
-                // finalize the task
-                task.exitStatus = readExitFile()
-                task.stdout = outputFile
-                task.stderr = errorFile
-            }
+            // finalize the task
+            task.exitStatus = readExitFile()
+            task.stdout = outputFile
+            task.stderr = errorFile
             status = TaskStatus.COMPLETED
```
pditommaso, line 367: This should go into a separate PR

xhejtman: this one is not mine, perhaps wrong merge from master? no problem to find and revert

pditommaso:
Made a few more changes. To summarise: 

* Use an `enum` type to model `Pod` and `Job` resource type
* Change the config setting to `computeResourceType`
* Try to not alter `pod` case in the log/print statement 

[k8s-patch.txt](https://github.com/nextflow-io/nextflow/files/8752378/k8s-patch.txt)
 

Please apply this patch, then I think we can merge this

xhejtman:
the patch cannot be applied to this head. Maybe one more patch is missing?

```
patch -sp1 < k8s-patch.txt 
1 out of 2 hunks FAILED -- saving rejects to file modules/nextflow/src/main/groovy/nextflow/k8s/K8sConfig.groovy.rej
6 out of 6 hunks FAILED -- saving rejects to file modules/nextflow/src/main/groovy/nextflow/k8s/K8sDriverLauncher.groovy.rej
11 out of 11 hunks FAILED -- saving rejects to file modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy.rej
The text leading up to this was:
--------------------------
|diff --git a/modules/nextflow/src/main/groovy/nextflow/k8s/model/ResourceType.groovy b/modules/nextflow/src/main/groovy/nextflow/k8s/model/ResourceType.groovy
|index 017a28575..c5f72afdd 100644
|--- a/modules/nextflow/src/main/groovy/nextflow/k8s/model/ResourceType.groovy
|+++ b/modules/nextflow/src/main/groovy/nextflow/k8s/model/ResourceType.groovy
--------------------------
File to patch: 
```

pditommaso:
You are right, please try this 

[k8s-patch-2.txt](https://github.com/nextflow-io/nextflow/files/8752494/k8s-patch-2.txt)

modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy
xhejtman: so what about this? Shall I search how this change got in?

pditommaso: ok, I'll do. thanks!

pditommaso:
Ok, let's merge it!

pditommaso:
Still some glitch to solve, I've merged this PR in the [k8s-jobs](https://github.com/nextflow-io/nextflow/tree/k8s-jobs) branch because I've noticed there was a lot of code duplication in the methods `podState` and `jobState` 

I've simplified it with the commit below, however, when testing it, it seems not working properly. Can you please have a look to it?  

https://github.com/nextflow-io/nextflow/commit/80a08a04c19e69dc458085501d72667ff3b12e06

xhejtman:
The problem is, that 'redundant' podStatus code contained fix for job:
`if( container.name != jobName )`
while original code has:
`if( container.name != podName )`

job name is basically: `nf-431a9203686e45f67aa2f07b09a8188b`, bud pod name is `nf-431a9203686e45f67aa2f07b09a8188b-lstm2`

and container name is derived from the job name.

pditommaso:
I'm not understanding this, why there's a pod even when running a job?

pditommaso:
My understanding is that the job name is assigned here via the `podName` attribute, therefore I would expect it's the same as the container name.  

https://github.com/nextflow-io/nextflow/blob/80a08a04c19e69dc458085501d72667ff3b12e06/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy#L464-L464

But likely I'm missing something?

xhejtman:
this is correct, but if you call podState0() in case of job, you are passing pod's name, not job name. line 375 in K8sClient.groovy. it fails as in case of job, container name is job name, not pod name.

bentsherman:
@pditommaso when you create a job, the job creates a pod, just like a deployment. The pod name is the job name with some additional characters appended, as in the example given above:

> job name is basically: `nf-431a9203686e45f67aa2f07b09a8188b`, but pod name is `nf-431a9203686e45f67aa2f07b09a8188b-lstm2`

pditommaso:
Indeed, I've figured out that. Just fixed 3e064b5a. @xhejtman please have a look and confirm it's ok for you

xhejtman:
Testing. Let you now tomorrow. Btw, name of task is not in annotation any more? I think there was some annotation that contained name of the task such as `MergeBamRecal (ACGT208-ACGT208)`. I do not see it with jobs. Did we miss it?

pditommaso:
> Testing

👌 

> Btw, name of task is not in annotation any more? 

Do you mean the [tag](https://www.nextflow.io/docs/latest/process.html#tag) directive? it should work as expected

xhejtman:
> > Btw, name of task is not in annotation any more?
> 
> Do you mean the [tag](https://www.nextflow.io/docs/latest/process.html#tag) directive? it should work as expected

I think it does not. Because we removed metadata section from pod (if running as job) and we do not set any label/annotation for job itself. And I think the tag is an annotation?

Should I create patch?

pditommaso:
It looks weird. 

> Should I create patch?

Sure, If you think it's needed. Please keep into a separate PR

xhejtman:
Ok, so let's finish this first. I will run a pipeline, if it finishes, lets merge it. I am aware of two things, one is the annotation and second is not optimal handling of some pod error situations (at least, I saw temporary error imagepullbackof in case of jobs, need to investigate a bit more). But I will address these in separate PRs.

pditommaso:
Cool. Thanks for checking it

pditommaso:
Done! c70eb12d

PR #2756, opened 2022-03-28
Deal with k8s api timeouts and similar errors

xhejtman:
Return code 500 from kubernetes api is mostly a temporary error like timeout, so we should try again in these cases.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
@@ -387,7 +387,9 @@ class K8sClient {
 
             try {
                 return makeRequestCall( method, path, body )
-            } catch ( SocketException e ) {
+            } catch ( K8sResponseException | SocketException e ) {
+                if ( ( e instanceof K8sResponseException ) && ( e.response.code != 500 ))
+                    throw e
                 log.error "[K8s] API request threw socket exception: $e.message for $method $path ${body ? '\n'+prettyPrint(body).indent() : ''}"
                 if ( trial < maxTrials ) log.info( "[K8s] Try API request again, remaining trials: ${ maxTrials - trial }" )
                 else throw e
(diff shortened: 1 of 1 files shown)


xhejtman:
E.g., if etcd is not responding, it throws following (with patch applied)
```
[K8s] API request threw socket exception: Request GET /api/v1/namespaces/test-ns/pods/nf-306563d54341960c92bcf2ea650811c5/status returned an error code=500

  {
      "kind": "Status",
      "apiVersion": "v1",
      "metadata": {
          
      },
      "status": "Failure",
      "message": "rpc error: code = Unavailable desc = transport is closing",
      "code": 500
  }
 for GET /api/v1/namespaces/test-ns/pods/nf-306563d54341960c92bcf2ea650811c5/status 


[K8s] Try API request again, remaining trials: 4
```

and workflow resumes once etcd is responding again.

modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
```diff
@@ -387,7 +387,9 @@ class K8sClient {
 
             try {
                 return makeRequestCall( method, path, body )
-            } catch ( SocketException e ) {
+            } catch ( K8sResponseException | SocketException e ) {
+                if ( ( e instanceof K8sResponseException ) && ( e.response.code != 500 ))
```
pditommaso, line 391: Being so, it would not be better to use `e.response.code >= 500` ? 

minor: the expression can be simplified as 

```
if ( e instanceof K8sResponseException && e.response.code != 500  ) 
```

xhejtman: didn't you actually mean `code < 500` `->` throw the exception right now?

pditommaso: My fault, I was misunderstanding the logic here. The rationale is to make `500` error code to behave like the `SocketException` and just rethrow in the other cases 👍

xhejtman: Yes. Should I remove the parantheses and push?

pditommaso: Yes, please

pditommaso (approved):
Nice, thanks for the contribution

PR #2761, opened 2022-03-30
Deprecate pullPolicy in favor of imagePullPolicy

bentsherman:
The `pollPolicy` pod option is an unnecessary shortcut for `imagePullPolicy` and should be removed in order to tighten up the nextflow / k8s interface. This PR simply deprecates `pullPolicy` so that users are warned that it may be removed in a future release.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/k8s/K8sConfig.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/k8s/K8sConfig.groovy
@@ -55,11 +55,14 @@ class K8sConfig implements Map<String,Object> {
             this.podOptions.volumeClaims.add(new PodVolumeClaim(name, mount, subPath))
         }
 
-        // -- shortcut to pod image pull-policy
-        if( target.pullPolicy )
+        // -- shortcut to pod image pull policy
+        if( target.pullPolicy ) {
+            log.warn "The `pullPolicy` option is deprecated -- Use `imagePullPolicy` instead."
             podOptions.imagePullPolicy = target.pullPolicy.toString()
-        else if( target.imagePullPolicy )
+        }
+        else if( target.imagePullPolicy ) {
             podOptions.imagePullPolicy = target.imagePullPolicy.toString()
+        }
 
         // -- shortcut to pod security context
         if( target.runAsUser != null )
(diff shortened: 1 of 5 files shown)


modules/nextflow/src/test/groovy/nextflow/k8s/model/PodOptionsTest.groovy
```diff
@@ -37,20 +37,20 @@ class PodOptionsTest extends Specification {
...
         when:
-        options = new PodOptions([ [pullPolicy:'Always'] ])
+        options = new PodOptions([ [imagePullPolicy: 'Always'] ])
         then:
         options.getImagePullPolicy() == 'Always'
 
         when:
-        options = new PodOptions([ [imagePullPolicy:'latest'] ])
+        options = new PodOptions([ [imagePullPolicy: 'latest'] ])
         then:
```
pditommaso, line 54: It should be kept a test also for `pullPolicy` to be sure it's still working

modules/nextflow/src/main/groovy/nextflow/k8s/K8sConfig.groovy
```diff
@@ -55,11 +55,14 @@ class K8sConfig implements Map<String,Object> {
             this.podOptions.volumeClaims.add(new PodVolumeClaim(name, mount, subPath))
         }
 
-        // -- shortcut to pod image pull-policy
-        if( target.pullPolicy )
+        // -- shortcut to pod image pull policy
+        if( target.pullPolicy ) {
+            log.warn "The `pullPolicy` option is deprecated -- Use `imagePullPolicy` instead."
```
pditommaso, line 60: In principle, only the one in `PodOption` should be enough. It does not?

bentsherman: Apparently it is both a pod option and a `k8s` config option, so I deprecated both.

modules/nextflow/src/test/groovy/nextflow/k8s/model/PodOptionsTest.groovy
bentsherman: Done

PR #2763, opened 2022-03-31
Add tolerations pod option

bentsherman:
This PR adds the `tolerations` pod option, which allows a pod to "tolerate" certain node taints during scheduling. Can be useful for situations where a node is designated for a specific use case (special CPU type, belongs to certain user group, etc), so it will only be scheduled to users who specifically tolerate the taint.

I also reorganized some of the pod-related classes to be more readable. I pushed two separate commits so that you can distinguish the toleration-specific changes from the general changes.


pditommaso (changes requested):
Hi, I think is useful but there are several unrelated changes that make the review difficult. Try to keep the PR as much as possible self-contained. Thanks!

bentsherman:
I wanted to clean up some of the k8s-related files to make them easier to read, but I wasn't sure if they justified a separate PR, so I included them with this one. You can inspect the two commits separately to make the review easier. Or if you prefer I could make a separate "general k8s cleanup" PR.

bentsherman:
Okay I moved the k8s cleanup to a separate draft PR for later. Now you should only see the feature-specific changes.

pditommaso (approved):
Ok, nice 👍

PR #2770, opened 2022-04-05
Document task object

### LINKED ISSUE #2732: Document fields available in task object
The docs should give a table of the fields that are always available in the `task` object. I think it also will include the values of certain directives if they were provided.

https://www.nextflow.io/docs/latest/script.html#process-implicit-variables

bentsherman:
Resolves #2732


pditommaso:
OK, made a few changes

Midnighter:
Awesome 👍🏼

PR #2776, opened 2022-04-07
Rename echo directive to debug

### LINKED ISSUE #1757: Add debug directive as replacement for echo
The [echo](https://www.nextflow.io/docs/latest/process.html#echo) directive is confusing since overlaps with Bash `echo`. 

It should be replaced by `debug` (keeping `echo for backward compatibility).

bentsherman:
Resolves #1757 

There was already a deprecation warning for `echo` so I just left it. If I understand correctly, `echo` should still work but will log the warning and redirect to `debug`.


pditommaso:
Well done, but we need to keep backward compatibility for at least one or two stable releases. Therefore old `echo` should be honoured showing a warning message

modules/nextflow/src/main/groovy/nextflow/script/BaseScript.groovy
```diff
@@ -87,7 +95,7 @@ abstract class BaseScript extends Script implements ExecutionContext {
         if( NF.dsl2Final )
             throw new DeprecationException(msg)
         log.warn(msg)
-        session.getConfig().process.echo = value
+        session.getConfig().process.debug = value
```
bentsherman, line 98: @pditommaso See here -- the original code throws a deprecation exception in DSL2. It looks like the Java 11 CI build fails due to this error but Java 17 doesn't. Aside from this exception, the `echo` directive is still accepted. What do you want to do about this exception?

pditommaso: This `echo` outside the `process` scope. Think we should remove it and without replacing it with `debug`. 

The `echo` in the `process` scope is implemented by [this](https://github.com/nextflow-io/nextflow/blob/c25bc355de078a3e5d3a66b536c5f577e031e47b/modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy#L497-L500) and [this](https://github.com/nextflow-io/nextflow/blob/c25bc355de078a3e5d3a66b536c5f577e031e47b/modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy#L189-L192) 

We should deprecate and keep it for some time, giving `debug` as alternative

pditommaso: Java 17 does not fail because integration tests are only run with Java 11

bentsherman:
Okay I think I fixed everything the way you described. So `BaseScript::echo()` is removed and `ProcessScript` has both `debug()` and `echo()`.

pditommaso (approved):

PR #2783, opened 2022-04-11
2422 process.ext set with withname overrides configuration of all processes

### LINKED ISSUE #2422: process.ext set with withName overrides configuration of all processes

 * Expected: Options set with `withName` only affect the corresponding process(es)
 * Actual: If process.ext gets initalized for all processes, updating it for a single process later overrides that value in all processes

Execute
```bash
nextflow run main.nf 
```

with the following three files

`main.nf`
```nextflow
#!/usr/bin/env nextflow

nextflow.enable.dsl = 2

include { FOO; FOO as BAR } from "./module.nf"

workflow {
    FOO()
    BAR()
}
```

`nextflow.config`
```nextflow
process {
    ext.args = "Hello World!"
    withName: 'BAR' {
        ext.args   = "Hi World!" 
	 cpus = 1
    }
}
```

`module.nf`
```nextflow
process FOO {
    exec:
    println task.ext.args
}
```

### Program output 

```console
N E X T F L O W  ~  version 21.04.0
Launching `main.nf` [hungry_lalande] - revision: 352bf9af95
executor >  local (2)
[6c/88b621] process > FOO [100%] 1 of 1 ✔
[f3/2daf6d] process > BAR [100%] 1 of 1 ✔
Hi World!
Hi World!
```

### Environment 

* Nextflow version: Tested with 21.04.0 and 21.10.0-SNAPSHOT
* Java version: openjdk 11.0.8-internal 2020-07-14
* Operating system: linux
* Bash version: `GNU bash, version 5.1.8(1)-release (x86_64-pc-linux-gnu)`

### Additional context

This bug was discovering while experimenting with the new nf-core dsl2.0 template together with @mahesh-panchal  and @drpatelh

jorgeaguileraseqera:
We maintain in `profileStack` a list of sections visited but we were checking if the property requested was only into the scope of the current stack instead to check if present in the last profile

Another issue was we were checking if the var is present with `if( current.scope[name] )` but this expression creates the variable if not present, so I've changed for `containsKey`

Lastly a little refactor removing the `if-else-if` chain for independents `if` who check if previous didn't find a valid result to return

closes #2422

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/config/ConfigParser.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/config/ConfigParser.groovy
@@ -337,9 +337,6 @@ class ConfigParser {
             def result = null
             if (current.config.get(name)) {
                 result = current.config.get(name)
-            } else if (current.scope.entrySet().find{ it.key==profileStack.last.toString()} ) {
-                result = current.scope.entrySet().find{ it.key==profileStack.last.toString()}.value[name]
-            } else if (current.scope[name]) {
             } else if (current.scope.get(name)) {
                 result = current.scope[name]
             }
--- a/modules/nextflow/src/main/groovy/nextflow/config/ConfigParser.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/config/ConfigParser.groovy
@@ -337,6 +337,9 @@ class ConfigParser {
             def result
             if (current.config.get(name)) {
                 result = current.config.get(name)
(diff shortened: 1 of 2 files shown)


pditommaso:
The guru of ConfigSlurper! 😄

pditommaso:
Can you give please have a look to the conflict?

pditommaso:
Unknown status. Closing it

bentsherman:
This PR needs to be cleaned up but I believe it can be salvaged. Reopening

PR #2788, opened 2022-04-12
Specified deprecation message

Lehmann-Fabian:
As the pipeline fails if the DeprecationException is thrown, I specified the message accordingly.


pditommaso:
Ok, thanks for pointing this out. I going to merge it. however recently the dsl2 *preview* mode has been removed therefore there's any more a different between dsl2 and dsl2Final 

https://github.com/nextflow-io/nextflow/blob/5c59ada56f807f95b4ae74ee62148018df106412/modules/nextflow/src/main/groovy/nextflow/NF.groovy#L50-L58

I'm going to adapt the corresponding error message

PR #2816, opened 2022-04-26
Update docs to be DSL2-first

pditommaso:
This is tentative to update the process section in the docs to DSL2 and remove references to old idioms.


bentsherman:
I will review this today. But is this PR only for updating the process page? Making the docs DSL2-first will probably require a few edits to other pages, and maybe incorporating the DSL2 page into the Nextflow scripting page or something like that. Just wondering if this is meant to be an initial step only.

bentsherman:
There are still some changes missing, so I will add them directly rather than through a review.

bentsherman:
@pditommaso Trying to wrap up my changes this morning. A few questions:
1. Regarding `Channel` vs `channel`, should the docs prefer one over the other? I think we should stick to one or else it gets confusing
2. The [tap](https://www.nextflow.io/docs/latest/operator.html#tap) isn't marked as deprecated, but it depends on `Channel.create()` which is deprecated. I tried the example with `Channel.empty()` but it didn't work. So should we deprecate `tap` ? It seems like we could achieve the same functionality by changing `set` to return itself like `view`.

samuell:
@bentsherman 
> I will review this today. But is this PR only for updating the process page? Making the docs DSL2-first will probably require a few edits to other pages

Indeed, e.g. the Get started page, as was suggested a fix in #2935

bentsherman:
Thanks for the reminder @samuell , I will make sure to review the getting started page as well. This PR should update all pipeline examples in the docs to DSL2.

bentsherman:
Fixed a few inconsistencies with the operators. Some weren't documented, some were deprecated but not documented as such, etc.

`countBy` is deprecated but not documented as such, but actually I think it should be un-deprecated. I'll leave that to a separate issue.

docs/process.rst
```diff
@@ -985,227 +1048,184 @@ Some caveats on glob pattern behavior:
...
-============== =====================
-Name            Description
-============== =====================
-glob            When ``true`` the specified name is interpreted as a glob pattern (default: ``true``)
-hidden          When ``true`` hidden files are included in the matching output files (default: ``false``)
-followLinks     When ``true`` target files are return in place of any matching symlink (default: ``true``)
-type            Type of paths returned, either ``file``, ``dir`` or ``any`` (default: ``any``, or ``file`` if the specified file name pattern contains a `**` - double star - symbol)
-maxDepth        Maximum number of directory levels to visit (default: `no limit`)
-includeInputs   When ``true`` any input files matching an output file glob pattern are included.
-============== =====================
```
pditommaso, line 1077: This has been moved or dropped?

bentsherman: it was moved further up

bentsherman: sorry I probably should not have rearranged things, since then I am trying to minimize changes

pditommaso:
Nice job 👍

PR #2818, opened 2022-04-26
[New Feature] Automatically detect DSL Version if not user-defined

Lehmann-Fabian:
Making DSL 2 the default DSL might confuse because workflows fail with errors that are hard to relate to the DSL version.
I try to detect the DSL if nothing is set automatically. If I am sure it is not DSL 2, it falls back to DSL 1, and a large warning is printed.
Currently, I detect DSL 2 via its `workflow` part:
```
workflow {
a()
b()
c()
}
```

Executing the following script: 
```
process a {

"""
echo hello world
"""

}
```

`nextflow run text.nf` without any config
Prints:

```
N E X T F L O W  ~  version 22.04.0
WARN:
======================================================================
=                               WARNING                              =
=  You have not defined a DSL version: automatically detected DSL1.  =
=        If DSL1 is wrong, define the DSL version explicitly.        =
= Therefore, call the script with -dsl2 or add nextflow.enable.dsl=2 =
= Check documentation: https://www.nextflow.io/docs/latest/dsl2.html =
======================================================================

Launching `test.nf` [maniac_fourier] DSL1 - revision: 0e33a19c82
executor >  local (1)
[6d/53bfc8] process > a [100%] 1 of 1 ✔
```

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/NextflowMeta.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/NextflowMeta.groovy
@@ -17,6 +17,7 @@ import static nextflow.extension.Bolts.DATETIME_FORMAT
 @Singleton(strict = false)
 @ToString(includeNames = true)
 @EqualsAndHashCode
+@Slf4j
 class NextflowMeta {
 
     private static final Pattern DSL_DECLARATION = ~/(?m)^\s*(nextflow\.(preview|enable)\.dsl\s*=\s*(\d))\s*(;?\s*)?(;?\/{2}.*)?$/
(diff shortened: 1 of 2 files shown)


Lehmann-Fabian:
I refactored it a bit. Now there is a separate method to detect the DSL language. This method can be extended later with further tests or for new DSLs.
Moreover, an info is shown if a user calls a script explicitly with `-dsl1` / `-dsl2`, but another version is defined in the script. Until now, it would silently ignore the user's arguments and use the script's definition.

I hope you like the idea of detecting the DSL.

pditommaso:
Thanks for this proposal. This possibility was discussed in the past, but it was decided to not have automatic detection because it could be even more confusing. 

Main reason a requirement is that it should be possible to run code snippets  even without including a `workflow` e.g. 

```
channel.of(1,2,3).view()
```

The system should default to DSL2 without warning the user. 


A warning has been added in the case the runtime default to DSL2 and the script contain one or more `process` and no `workflow`

https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/script/BaseScript.groovy#L187-L189

Maybe it should be made this message for evident using a big banner as in this PR

Lehmann-Fabian:
Thanks for your review, @pditommaso . 
I could add another check to confirm that at least one process is included if I switch to DSL1. DSL2 would fail anyway in that case. So, why not try DSL1 automatically? 
Another opportunity would be to show the warning but not set DSL to DSL1. In contrast to your Exception, I can distinguish between a user who has not defined any DSL and a user who has maybe defined the wrong one. I think the feedback should be different in both cases.

Should I add the other test and/or only show the warning?

In our group, it caused a lot of confusion that workflows that ran before without any problems did not even start with the newest Nextflow version (some switched back to an older Nextflow release to solve), but `-dsl1` would be the trick. For this reason, I think Nextflow should give clear feedback if no DSL is defined and it causes problems.

pditommaso:
I agree it should be reported as a more explicative error message, but I think it would be better to avoid trying to guess the script syntax because it could lead to an ambiguous situation. 

It's better to make the users more aware of the difference between DSL1 and DSL2 and incentivise them to upgrade their scripts to the later. 

I've enhanced the error message taking inspiration from your proposal https://github.com/nextflow-io/nextflow/pull/2871. 

Thanks

pditommaso:
I've reconsidered a bit the idea of dsl detection trying to infer the use of DSL1 by checking for the patterns `input: <type> x from <channel>` or `output: <type> x into <channel>` \*and* the absence of the `workflow` keyword when no explicit config is given. a3b5221a9216f8dd0bbcb4a2fbaef96ba4cdebed

Lehmann-Fabian:
Thank you for your feedback, @pditommaso. I also thought about these bose patterns. I saw a problem if you only have something like this:

```
process a {

echo true

output:
file "a.txt"
file "b.txt" into b

"""
echo "hallo" > a.txt
ln -s a.txt b.txt
"""

}
```
`nextflow example.nf  -dsl1 -process.scratch true -process.stageOutMode move`

file a.txt has to be declared, as it would fail otherwise. I solved this issue here: https://github.com/nextflow-io/nextflow/pull/2478, maybe you can take a look?

So maybe you have to adjust the output pattern a bit to allow repetition.

PR #2820, opened 2022-04-26
feature: new dsl syntax to explicit declaration of plugin extensions

### LINKED ISSUE #2804: DSL2 plugin extension explicit declaration
### Summary 

The goal of this issue is to define a DSL syntax to explicitly declare the extension methods imported by a Nextflow plugin.

This allows developers to precisely control which extensions are going to be used in the workflow script. It can also help code IDEs to check and validate the script syntax. 

### Context 

As of version `21.08.0-edge` nextflow plugins can provide extension methods that add new operations to create and manipulate channels.   

For example, the [nf-sqldb plugin](https://github.com/nextflow-io/nextflow/blob/abf7776b902c9fa37f2b0642bc07f604f0868e6b/plugins/nf-sqldb#L6-L6) adds the method [fromQuery](https://github.com/nextflow-io/nextflow/blob/abf7776b902c9fa37f2b0642bc07f604f0868e6b/plugins/nf-sqldb/src/main/nextflow/sql/ChannelSqlExtension.groovy#L58-L61) that allows the creation of a nextflow channel from a SQL query and the method and the [sqlInsert](https://github.com/nextflow-io/nextflow/blob/abf7776b902c9fa37f2b0642bc07f604f0868e6b/plugins/nf-sqldb/src/main/nextflow/sql/ChannelSqlExtension.groovy#L95-L95) method that allows inserting the channel data into a SQL table. 

The first is called a *custom factory method* because it allows the creation of a channel and it's expected to be invoked as `channel.sql.fromQuery`. 

The second is called a *custom operator* because it's expected to be invoked on a channel instance crated by another factory or process output. 

More [examples here](https://github.com/nextflow-io/nf-sqldb#available-operations).

A plugin that implements DSL extension is expected to implement the interface [ChannelExtensionPoint](https://github.com/nextflow-io/nextflow/blob/abf7776b902c9fa37f2b0642bc07f604f0868e6b/modules/nextflow/src/main/groovy/nextflow/extension/ChannelExtensionPoint.groovy#L16-L16)

The implementing class should also be marked with a [Scoped](https://github.com/nextflow-io/nextflow/blob/abf7776b902c9fa37f2b0642bc07f604f0868e6b/modules/nf-commons/src/main/nextflow/plugin/Scoped.groovy#L23-L23) annotation which defines the context on which the factory method can be invoked. 

For example `channel.sql` for the [sql plugin](https://github.com/nextflow-io/nextflow/blob/abf7776b902c9fa37f2b0642bc07f604f0868e6b/plugins/nf-sqldb/src/main/nextflow/sql/ChannelSqlExtension.groovy#L26-L26). 

Then, the *public* methods in this class can be invoked from the workflow script provided there's a matching signature. 

The channel operators are expected to have as first argument a value of type `DataflowReadChannel`. See for [here](https://github.com/nextflow-io/nextflow/blob/abf7776b902c9fa37f2b0642bc07f604f0868e6b/modules/nextflow/src/main/groovy/nextflow/extension/ChannelExtensionDelegate.groovy#L119-L119) and [here](https://github.com/nextflow-io/nextflow/blob/abf7776b902c9fa37f2b0642bc07f604f0868e6b/plugins/nf-sqldb/src/main/nextflow/sql/ChannelSqlExtension.groovy#L95-L95)

### Goals 

The goal of this issue is to provide an explicit declaration for the extension methods to be used in the workflow script, using the syntax already existing for loading module components. See [here](https://www.nextflow.io/docs/latest/dsl2.html#modules-include).

A possible syntax could be the following 

```
include { fromQuery } from 'plugin/nf-sqldb'
```

Plugin version can be specified using the `@` qualifier, for example  

```
include { fromQuery } from 'plugin/nf-sqldb@0.4.0'
```

Multiple methods can be imported using the `;` separator. For example 

```
include { fromQuery; sqlInsert } from 'plugin/nf-sqldb@0.4.0'
```
 
The `as` keyword can be used to the extension with a new name. For example

```
include { fromQuery as sqlQuery } from 'plugin/nf-sqldb'
```

### Caveats 

A problem with no trivial solution is how to manage the inclusion of different versions of the same plugins across different modules. 

The Nextflow plugin system does not allow the loading of different versions at the same time, therefore if a module A required the plugin `X@1.0.0` and module B the plugin `X@2.0.0`, the latter will cause the stop and the replacement of the former, which can break the pipeline execution. 

Possible solutions: 

* ~~automatically upgrade the plugin to the latest declared version e.g. `X@2.0.0` in the previous example warn the user~~
* do not allow the declaration of the plugin version at inclusion time, and allow the user to explicitly require it only in th nextflow config file

jorgeaguileraseqera:
This PR implements the new `import` plugin functionality by reusing the `IncludeDef` process creating a Map of alias and methods implemented into the plugin

it intercepts the missing static method into the Channel class and reroutes the call to the corresponding ExtensionPoint function

closes #2804


modules/nextflow/src/main/groovy/nextflow/Channel.groovy
```diff
@@ -66,9 +68,11 @@ class Channel  {
     private static CompletableFuture fromPath0Future
 
     static private Session getSession() { Global.session as Session }
-    
 
-    /**
+    static def $static_methodMissing(String name, Object args) {
+        ChannelExtensionDelegate.INSTANCE().invokeFactoryExtensionMethod(name, args as Object[])
+    }
```
pditommaso, line 74: Is `$static_methodMissing` a magic groovy method? if so please add some docs

modules/nextflow/src/main/groovy/nextflow/extension/ChannelExtensionDelegate.groovy
```diff
@@ -136,14 +198,23 @@ class ChannelExtensionDelegate implements DelegatingPlugin {
     }
 
     Object invokeExtensionMethod(Object channel, String method, Object[] args) {
-        final target = operatorExtensions.get(method)
+        final target = aliasOperatorExtensions.get(method)?.v2 ?: operatorExtensions.get(method)
+        method = aliasOperatorExtensions.get(method)?.v1 ?: method
         if( target==null )
```
pditommaso, line 203: What are `v1` and `v2` attributes?

modules/nextflow/src/main/groovy/nextflow/extension/ChannelFactoryInstance.groovy
```diff
@@ -71,7 +71,7 @@ class ChannelFactoryInstance implements ChannelFactory {
     @Override
     Object invokeExtensionMethod(String method, Object[] args) {
         def result = invoke0(method,args)
-        NodeMarker.addSourceNode("channel.${this.scope}.${method}", result)
+        NodeMarker.addSourceNode(scope ? "channel.${this.scope}.${method}" : "channel.${method}", result)
```
pditommaso, line 74: Think we can remove completely the support for `scope`

modules/nextflow/src/main/groovy/nextflow/script/IncludeDef.groovy
```diff
@@ -108,6 +111,10 @@ class IncludeDef {
      */
     void load0(ScriptBinding.ParamsMap ownerParams) {
         checkValidPath(path)
+        if( path.toString().startsWith('plugin/')){
+            loadPlugin0(ownerParams)
+            return
+        }
```
pditommaso, line 117: Let's add a comment about this

modules/nextflow/src/main/groovy/nextflow/script/ScriptBinding.groovy
```diff
@@ -204,6 +205,12 @@ class ScriptBinding extends WorkflowBinding {
         super.getVariables().find { entry -> entry.value?.is(value) }?.getKey()
     }
 
+
+    @CompileDynamic
+    Object invokeDynamicMethod(String name, Object args) {
```
pditommaso, line 210: What's the different `invokeDynamicMethod` vs `invokeMethod` ?

pditommaso (changes requested):
Ok, it looks very good. Mostly think it would be better to comment a bit more especially groovy magic extension methods. 

Also, we can get rid of the *scope* mechanism, and therefore of this 

https://github.com/nextflow-io/nextflow/blob/6b45b50d966589a6678ca8b712d1ae2a8581b06f/modules/nextflow/src/main/java/groovy/runtime/metaclass/NextflowDelegatingMetaClass.java#L69-L76

modules/nextflow/src/main/groovy/nextflow/extension/ChannelExtensionDelegate.groovy
jorgeaguileraseqera: ChannelExtensionDelegate maintains a Map<Tuple> where the key is the alias, and the value is a tuple with the real method and the class how implements it so basically:

if we are calling the invokeExtensionMethod with a method not registered as a plugin we call the method as usual
if we are invoking an alias method, we retrieve the extension and the real method from the map

pditommaso: ok, think you can copy & paste this as a comment in that code 😄

modules/nextflow/src/main/groovy/nextflow/extension/ChannelFactoryInstance.groovy
jorgeaguileraseqera: I think is required to be compatible with current scripts who doesn't use the new syntax

pditommaso: think we can go with a breaking change here

modules/nextflow/src/main/groovy/nextflow/script/ScriptBinding.groovy
jorgeaguileraseqera: upppsss, sorry it was not removed from the previous approach

modules/nextflow/build.gradle
```diff
@@ -34,7 +34,7 @@ dependencies {
     api "com.beust:jcommander:1.35"
     api("com.esotericsoftware.kryo:kryo:2.24.0") { exclude group: 'com.esotericsoftware.minlog', module: 'minlog' }
     api('org.iq80.leveldb:leveldb:0.12')
-    api('org.eclipse.jgit:org.eclipse.jgit:5.2.1.201812262042-r')
+    api('org.eclipse.jgit:org.eclipse.jgit:6.1.0.202203080745-r')
```
pditommaso, line 37: oh no. new jgit landed here 😮‍💨

jorgeaguileraseqera: sorry, I understood this change solves the issue with the pipeline, but I'll rollback

modules/nextflow/src/main/groovy/nextflow/extension/ChannelExtensionDelegate.groovy
```diff
@@ -44,10 +43,12 @@ class ChannelExtensionDelegate implements DelegatingPlugin {
...
     private Session getSession() { Global.getSession() as Session }
 
-    final private Map<String, ChannelFactory> channelFactories = new HashMap<>()
-
     final private Map<String,Object> operatorExtensions = new HashMap<>()
 
+    final private Map<String,Tuple2<String,Object>> aliasOperatorExtensions = new HashMap<>()
+
+    final private Map<String,Tuple2<String,ChannelFactoryInstance>> aliasFactoryExtensions = new HashMap<>()
+
```
pditommaso, line 51: Let's use a custom class instead `Tuple2` that should help a bit to make the underlying logic more accessible. Same for `aliasOperatorExtensions`

modules/nextflow/src/main/groovy/nextflow/extension/ChannelExtensionDelegate.groovy
```diff
@@ -44,10 +43,12 @@ class ChannelExtensionDelegate implements DelegatingPlugin {
 
     private Session getSession() { Global.getSession() as Session }
 
-    final private Map<String, ChannelFactory> channelFactories = new HashMap<>()
-
     final private Map<String,Object> operatorExtensions = new HashMap<>()
 
+    final private Map<String,Tuple2<String,Object>> aliasOperatorExtensions = new HashMap<>()
```
pditommaso, line 48: Wondering if the two could not be unified into a single `operatorExtensions`?

modules/nf-commons/build.gradle
```diff
@@ -36,5 +36,7 @@ dependencies {
 
     /* testImplementation inherited from top gradle build file */
     testImplementation(testFixtures(project(":nextflow")))
+    testImplementation("org.codehaus.groovy:groovy-sql:3.0.10")
+    testImplementation('com.h2database:h2:1.4.200')
```
pditommaso, line 40: Testing a the plugin loading is a good idea, however, I would like to avoid depending the SQL stuff, both in the terms on deps and also because the test logic is quite complicated because requires to run some SQL query. 

Think it would be easier to use the 'nf-hello' plugin https://github.com/nextflow-io/nf-hello

pditommaso (changes requested):
Ok. I've removed the code related to the *scope* handling. Since it was only used by the SQL plugin, we can leave with a breaking chance and simplify the implementation. 

I've left some other comments

modules/nextflow/src/main/groovy/nextflow/extension/ChannelExtensionDelegate.groovy
jorgeaguileraseqera: yes, no problem, was to no complicate and keep separate them but we can use only one Map

modules/nf-commons/build.gradle
jorgeaguileraseqera: The main "problem" is if I'm not wrong, `nf-hello' plugin doesn't have any function to import and test

pditommaso: Let's add it, just commented

pditommaso:
I've made naming refactoring and added some comments. Think we can merge if all tests are ok

PR #2828, opened 2022-04-28
[New Feature] Include the node into the trace report for Kubernetes

Lehmann-Fabian:


jorgeaguileraseqera (approved):

bentsherman:
Isn't the node name just the hostname? If we are going to add a trace record, I think we should call it `hostname` and try to make it more general than just Kubernetes.

We have also considered creating a `trace` directive to define custom trace records in the process definition. In reality I think the hostname / node name would be more appropriate as something for the user to define with such a directive, e.g.
```nextflow
process example {
  // execute bash script like beforeScript, add key/value to .command.trace
  trace "echo node=\${NODE_NAME}"

  """
  your_command ...
  """
}
```

Lehmann-Fabian:
Thanks for your feedback, @bentsherman.
I renamed it from node to hostname.
Yes, this should also apply to other executors, but I have none but Kubernetes here to test and implement. So I started with Kubernetes.

Having manual traces would be nice. The hostname, however, should be a native feature. Pegasus has this as well.

Having manual traces as a directive is maybe not enough. A more powerful implementation would be to always check for a file `.nextflow.user.trace`, which is then structured as the standard trace file. This would allow to, for example, measure the runtime of a part of a script.

```
process example {

"""
whatever
start=`date +%s%N | cut -b1-13`
do something
end=`date +%s%N | cut -b1-13`
echo something_runtime=$((end-start)) > .nextflow.user.trace
"""
}
```

bentsherman:
Fair enough. If we can extend support for `hostname` to the other executors at some point in the future (not necessarily this PR), I'm fine with it.

Lehmann-Fabian:
@bentsherman, so this is ready to be merged. Can you please do that?

pditommaso:
Since this information is going to ignore by most of the users I think it should be an opt-in feature enabled in the config file, under the `k8s` scope e.g. `k8s.fetchNodeName` or something like that.

Lehmann-Fabian:
Thanks for your feedback, @pditommaso. I implemented it accordingly and adjusted the documentation.

docs/config.rst
```diff
@@ -412,6 +412,7 @@ securityContext     Defines the `security context <https://kubernetes.io/docs/ta
 storageClaimName    The name of the persistent volume claim where store workflow result data.
 storageMountPath    The path location used to mount the persistent volume claim (default: ``/workspace``).
 storageSubPath      The path in the persistent volume to be mounted (default: root).
+fetchNodeName       If you trace the hostname, activate this option (default: ``false``).
```
pditommaso, line 415: please change it as 

```
(default: ``false``, requires version ``22.05.0-edge`` or later)
```

docs/tracing.rst
```diff
@@ -234,6 +234,7 @@ workdir                 The directory path where the task was executed.
 script                  The task command script.
 scratch                 The value of the process ``scratch`` directive.
 error_action            The action applied on errof task failure.
+hostname                The host on which the task was executed. Supported only for the Kubernetes executor yet. Activate with ``fetchNodeName = true`` in k8s config.
```
pditommaso, line 237: please change to 


```
Activate with ``k8s.fetchNodeName = true`` in the Nextfliw config file. (requires version ``22.05.0-edge`` or later)
```

modules/nextflow/src/main/groovy/nextflow/k8s/K8sConfig.groovy
```diff
@@ -169,6 +169,11 @@ class K8sConfig implements Map<String,Object> {
         podOptions
     }
 
+    @Memoized
+    boolean fetchNodeName() {
+        target.fetchNodeName as boolean
+    }
```
pditommaso, line 175: Let's add a test in `K8sConfigTest` class

modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
```diff
@@ -223,6 +225,13 @@ class K8sClient {
         return false
     }
 
+    @CompileDynamic
```
pditommaso, line 228: Better to avoid `@CompileDynamic` (not even sure it's required)

pditommaso (changes requested):

pditommaso:
Thanks for the prompt reply. Just made some minor comments

Lehmann-Fabian:
Thanks, @pditommaso . Adjusted everything.

modules/nextflow/src/main/groovy/nextflow/k8s/K8sConfig.groovy
```diff
@@ -169,6 +169,11 @@ class K8sConfig implements Map<String,Object> {
         podOptions
     }
 
+    boolean fetchNodeName() {
+        log.info( target.toString() )
```
pditommaso, line 173: This would be shown in the console. it's really needed I would even decrease to trace

Lehmann-Fabian: Well, this was for testing the behavior of `as boolean` and should not have been committed.

pditommaso (approved):
Nice, thanks

PR #2830, opened 2022-04-29
add -head-prescript option

xhejtman:
Add -head-prescript option to be able to pass init script before running actual nextflow pipeline. The same is supported in tower and can be used to workaround some problems with nextflow.config as kuberun does not copy whole nextflow.config and omits some parts, e.g., function definitions.


xhejtman:
this is not right place to test this option. But I added a test anyway. Also for previous head-cpus and head-memory. Is it ok now?

xhejtman:
btw, it is a question whether `-head-cpus`, `-head-memory`, and `-head-prescript` are intuitive names. They are all used for `kuberun` only. So perhaps `-kube-cpus`, `-kube-memory`, and `-kube-prescript` are better names? `-kuberun-cpus` is a bit long. It is up to you what you prefer.

pditommaso:
Since they are in the context of the `kuberun` command I think the `-head` prefix is fair. Thanks for this

pditommaso (approved):

PR #2831, opened 2022-04-29
[Feature] micromamba support

jpfeuffer:
This enables the support for micromamba, a more lightweight, C-only, env-free alternative of conda/mamba.
It has better parallelism support through lockfiles and therefore solved https://github.com/nextflow-io/nextflow/issues/1819
for me so far.

It does not have a 100% compatible command line, therefore an if-case was needed. See https://github.com/mamba-org/mamba/issues/1654

Let me know what you think.

I was about to just use conda.executable(Name) as a config param but was worried about backwards compatibility.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
@@ -101,12 +101,15 @@ class CondaCache {
         if( config.cacheDir )
             configCacheDir0 = (config.cacheDir as Path).toAbsolutePath()
 
+        if( config.useMamba && config.useMicroMamba)
+            throw new IllegalArgumentException("Both conda.useMamba and conda.useMicromamba were enabled -- Please choose only one")
+        
         if( config.useMamba )
             useMamba = config.useMamba as boolean
 
         if( config.useMicromamba )
             useMicromamba = config.useMicromamba as boolean
-
+        
     }
 
     /**
@@ -182,7 +185,7 @@ class CondaCache {
                 throw new IllegalArgumentException("Conda environment file does not exist: $condaEnv")
             }
             catch( Exception e ) {
-                throw new IllegalArgumentException("Error parsing Conda environment YAML file: $condaEnv -- Chech the log file for details", e)
+                throw new IllegalArgumentException("Error parsing Conda environment YAML file: $condaEnv -- Check the log file for details", e)
             }
         }
         else if( isTextFilePath(condaEnv) )  {
@@ -195,7 +198,7 @@ class CondaCache {
                 throw new IllegalArgumentException("Conda environment file does not exist: $condaEnv")
             }
             catch( Exception e ) {
-                throw new IllegalArgumentException("Error parsing Conda environment text file: $condaEnv -- Chech the log file for details", e)
+                throw new IllegalArgumentException("Error parsing Conda environment text file: $condaEnv -- Check the log file for details", e)
             }
         }
         // it's interpreted as user provided prefix directory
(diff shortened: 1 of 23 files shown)


modules/nf-commons/src/main/nextflow/Const.groovy
```diff
@@ -58,12 +58,12 @@ class Const {
...
-    static public final long APP_TIMESTAMP = 1650736802565
+    static public final long APP_TIMESTAMP = 1651152885778
 
     /**
      * The app build number
      */
-    static public final int APP_BUILDNUM = 5697
+    static public final int APP_BUILDNUM = 5700
 
     /**
```
abhi18av, line 68: This change should not be added to the PR.

modules/nextflow/src/main/resources/META-INF/plugins-info.txt
```diff
@@ -1,7 +1,7 @@
...
+nf-azure@0.13.1
 nf-console@1.0.3
+nf-ga4gh@1.0.3
 nf-google@1.1.4
-nf-azure@0.13.1
-nf-tower@1.4.0
-nf-ga4gh@1.0.3
\ No newline at end of file
+nf-sqldb@0.4.0
+nf-tower@1.4.0
```
abhi18av: This should also not be a part of the PR.

abhi18av (commented):
Thanks for the initiative @jpfeuffer !

I've highlighted some files which shouldn't be a part of the PR.

I was also looking into `micromamba` recently and to me it seems that it doesn't have full compatibility with `conda`.

In case of `mamba`, it _falls back_ to the `conda` layer but `micromamba` complained that "such and such wasn't implemented" in my testing earlier. Perhaps that behavior has changed recently 🤔 

Have you tested locally with a sizeable pipeline?

jpfeuffer:
Yes, it seemed to work for me for our nf-core pipeline. Feel free to try in the copy of my gitpod workspace. Just add micromamba to the path.

https://gitpod.io#snapshot/1fd28fec-e254-4655-b148-d390435206e9

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
```diff
@@ -252,17 +262,20 @@ class CondaCache {
...
         }
         else if( isTextFilePath(condaEnv) ) {
 
-            cmd = "${binaryName} create $opts--mkdir --yes --quiet --prefix ${Escape.path(prefixPath)} --file ${Escape.path(makeAbsolute(condaEnv))}"
+            cmd = "${binaryName} create $opts $mkdir --yes --quiet --prefix ${Escape.path(prefixPath)} --file ${Escape.path(makeAbsolute(condaEnv))}"
         }
 
         else {
-            cmd = "${binaryName} create $opts--mkdir --yes --quiet --prefix ${Escape.path(prefixPath)} $condaEnv"
+            cmd = "${binaryName} create $opts $mkdir --yes --quiet --prefix ${Escape.path(prefixPath)} $condaEnv"
```
pditommaso, line 278: This is not a mistake. the `opts` is expected to have a trailing space.

jpfeuffer: Yes thanks, I just saw it when I re-ran the tests. It has double spaces now.

pditommaso (changes requested):
Thanks for submitting this improvement. In order to merge it, tests should be fixed, also a test should be added to validate the new option. 

Finally, a mention in the docs should be added about the new flag

pditommaso:
This has been merged manually 383e023f. Thanks for the contribution.

jpfeuffer:
Awesome! Thank you very much for jumping in. Last week was very busy.

PR #2832, opened 2022-04-29
Feature groovy4

pditommaso:
Draft PR to evaluate the use of Groovy 4 

Note: discussion about performance issues with Groovy 4 [here](https://lists.apache.org/thread/8vn8h2w8rpxmyz58bjqtkxmcwv2o6trp) and [here](https://issues.apache.org/jira/browse/GROOVY-10307?).


pditommaso:
The most frequent changes look like this 

https://github.com/nextflow-io/nextflow/pull/2832/files#diff-2844a4cf2d2f6d3236ef9fe782c677cf12fba212542db5da5504aba0e8bea4c3L210-R210

Why is required the use of `setProperty` method instead of dot property notation with groovy4 ?

jorgeaguileraseqera:
they are `private` and the compiler marks them as errors ... but only when they are in the `when` and not in the `test` ¯\_(ツ)_/¯

pditommaso:
Not understanding this: 

> but only when they are in the when and not in the test

jorgeaguileraseqera:
For example

I needed to change this

```
when:
        cache.createOptions ----> cache.setProperty('createOptions', '--this --that')
```

but not this:

```
then:
        cache.createOptions == null
```

maybe is something related with spockframework, or new groovy compiler, not sure by the moment why

jorgeaguileraseqera:
I think it's due to use Spy classes with private properties

See 
https://stackoverflow.com/questions/49818211/how-to-create-spock-spy-with-private-property
https://github.com/spockframework/spock/issues/403

Right now is working via setProperty but I agree is not the best way, so probably the best solution is to modify the visibility of these properties and promote them to protected

pditommaso:
Not 100% but `GroovySpy` instead of using `Spy` could be another trick

jorgeaguileraseqera:
I marked also other tests with

`@IgnoreIf({ new Date().after( new SimpleDateFormat("yyyy/MM/dd").parse("2020/05/5"))})`

because they were falling, probably related to the Spy issue, but in this way, I was able to have a first version to validate groovy 4.0.2

I'll take a second round and check if `GroovySpy` helps with them

jorgeaguileraseqera:
I tried with `GroovySpy` and it doesn't work but I founded we can use `constructorArgs` and I think it has all the sense

instead

```
def cache = Spy(CondaCache)
cache.useMamba = true  //useMamba is a private property so GCLib can access to it 
```
we can write

`def cache = Spy(CondaCache, constructorArgs: [new CondaConfig(useMamba:true)])`

pditommaso:
For this make sense. Think it should even be possible to do 

```
def cache = Spy( new CondaCache(new CondaConfig(useMamba:true) )
```


tho, think it remain the problem with private fields are accessed in the `then:` part, right?

jorgeaguileraseqera:
> For this make sense. Think it should even be possible to do
> 
> ```
> def cache = Spy( new CondaCache(new CondaConfig(useMamba:true) )
> ```

yes, for this case it also works, maybe was not the best "example". I've changed for example in K8sTaskHandlerTest as

`def handler = Spy(K8sTaskHandler, constructorArgs: [builder:builder, client:client, task:task])`

removing the `handler.client = client` and it works

about the `them` part I think there is not problem, it's working, only I don't fully understand why it can access to the private property

modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy
```diff
@@ -114,10 +114,10 @@ class CharliecloudCache {
     @PackageScope
     Path getCacheDir() {
 
-        if( config.pullTimeout )
-            pullTimeout = config.pullTimeout as Duration
+        if( config['pullTimeout'] )
+            pullTimeout = config['pullTimeout'] as Duration
```
pditommaso, line 118: As this point is better to add public getters for each of these returning the expected type

jorgeaguileraseqera: As groovy4 is an interesting feature but not urgent, I can mark this PR as a draft and try to maintain it updated (rebasing from master frequently) 

in this way, we can fix this kind of issue not directly related to groovy4 in other branches

pditommaso:
I read this is in the past, look still there are serious performance problems with Groovy 4
 https://lists.apache.org/thread/8vn8h2w8rpxmyz58bjqtkxmcwv2o6trp

pditommaso:
Realised too late that changes i've made to upgrade to Spock 2.1 produced many conflicts in this PR. 

However makes sense to keep it PR only focused on changes required by Groovy 4.

jorgeaguileraseqera:
yes, don't worry, Iwill try to keep this branch updated from master 

I agree it's better to have these improvements in another PRs and restrict this branch only to restricted groovy 4 issues

pditommaso:
it was a nice guy

PR #2843, opened 2022-05-02
Sanitize k8s label and annotation keys, don't sanitize annotation value

### LINKED ISSUE #2834: k8s annotation values should not be sanitized
## Bug report 

As per k8s documentation, annotation values don't have the same restrictions as labels. See https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set. Actually, for labels both the key and the value should be sanitized whereas only the key should be sanitized for annotations.

For example:
```
process {
  pod = [ [annotation: "slurm-job.knoc.io/flags", value: "--job-name=tutorial --time=00:05:00"] ]
}
```
sanitizes the value to:
```
job-name_tutorial_--time_00_05_00
```
which is definitely not the expected behaviour.

IMHO, the example above is self explanatory.

### Program output 

There is no nextflow output as the sanitized annotation value causes the slurm job submission to fail.

### Environment 

* Nextflow version: 22.04.0 build 5697
* Java version: Groovy 3.0.10 on OpenJDK 64-Bit Server VM 17.0.1+12-LTS
* Operating system: Linux

### Additional context

This is using KNoC, implemented using virtual-kubelet. With the appropriate annotation, NodeSelector and toloeration, KNoC impersonates a k8s cluster but actually runs the workload on slurm. I know nextflow has its own slurm executor, but this work in progress potentially allowing better portablilty.

Thank you.

bentsherman:
Closes #2834

### DIFF
--- a/docs/_static/theme.css
+++ b/docs/_static/theme.css
@@ -1,0 +1,16 @@
+
+/* navbar -- logo */
+
+.wy-side-nav-search a:before {
+    display: block;
+    content: url(https://github.com/nextflow-io/trademark/raw/master/nextflow2014_black_white.png);
+    visibility: visible;
+    transform: scale(.43);
+    transform-origin: left;
+    margin: -30px 0 -50px 0;
+}
+
+.wy-side-nav-search a {
+    visibility: hidden;
+}
+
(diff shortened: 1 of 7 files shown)


pditommaso:

Issue #1489, opened 2020-02-08
Optional order for publishDir


srynobio:
## New feature

publishDir is noted as operating in an '_asynchronous manner_', which often is the preferred method of transfer.  However this can cause issues when transferring a file type and its respective index.

Example:
A process creates a VCF file and a tabix indexed file.  The process uses the publishDir to '_copy_' both files to a final directory. However, because the .tbi file copy much quicker (smaller), it receives a older timestamp then the VCF file once it transfers.  Many tools that require the .tbi file to be used with the VCF will error out because index files are not allow to have an older stamp then the original VCF.  

I have also noticed this issue with BAM & BAM indexes.

## Suggest implementation 
Possibly an additional '_parameters_' that allows order to be specified when needed. i.e. synchronous manner.

```
publishDir "my/path/", mode: 'copy', pattern: "*.gz, order: 1
publishDir "my/path/", mode: 'copy', pattern: "*.tbi", order: 2
```

pditommaso:
This is out the scope of `publishDir` directory which is designed to store workflow data result. Ifyou need to synchronise the data output the execution of other tasks/tools a nextflow process has to be used instead.

srynobio:
Okay, given that my example shows the files running through `publishDir` were _store workflow data result[s]_ they appeared to me to fall within this scope, but passing them to another process was second on my list of possible actions.

Thanks!

ghuls:
@pditommaso If publishDir would copy the last modification time from the original files, this would not be a problem at all.
Like `cp` is doing when running it like this:
`cp --preserve=timestamps orgina_file new_file`

jjfarrell:
@srynobio @pditommaso @lindenb 
Several bioinformtics tools output both  the vcf.gz and vcf.gz.tbi files. As described above, the present publishDir copy mode often results with files that have the vcf.gz.tbi  that are dated before the vcf.gz file which is problematic for downsteam processes The copy is not preserving the file times. Simply using the cp -p as mentioned by @ghuls will solve the problem and maintain the  'asynchronous manner'.   This issue has also been mentioned in #3002 for bams.  Often this issue is not a problem when running pipelines on small test files but becomes an issue when testing production runs on wgs crams or large vcf.gz files with larger number of samples. 

[...]

PR #2848, opened 2022-05-04
sort published files so common bioinformatics indexes are processed at the end.

lindenb:
Hi, this small PR is related to the following issue:

https://github.com/nextflow-io/nextflow/issues/1489

When publishing a set of paths, reorder the files so the common indexes (TBI, BAI,CSI...) are copied at the end, so the timestamps of the files are always older than their indexes (otherwise most tools will fail or emit a warning)

I understand that this is specific to bioinformatics, feel free to reject the PR.

I'm not a groovy programmer so I hope it's the right way and the right place to insert that code. I'm not even sure the files are copied in the order of the `List` :sweat_smile: 

P.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
@@ -212,7 +212,7 @@ class PublishDir {
         /*
          * iterate over the file parameter and publish each single file
          */
-        for( Path value : dedupPaths(files) ) {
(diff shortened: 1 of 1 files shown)


lindenb:
@pditommaso thanks,

I'll try to add the option in the next days.

bentsherman:
Solved in the latest edge releases via the `workflow.output.copyAttributes` config option

https://nextflow.io/docs/latest/migrations/25-04.html#miscellaneous

PR #2853, opened 2022-05-06
feature: add support for label/tags

### LINKED ISSUE #2845: Add support for label/tags
### Summary 

The goal of this enhancement is to allow users to propagate label/tag annotations to the corresponding cloud execution platform.

This is a recurrent request from Nextflow users. See [here](https://github.com/nextflow-io/nextflow/issues/2817), [here](https://github.com/nextflow-io/nextflow/issues/1764), [here](https://github.com/nextflow-io/nextflow/issues/948) and [here](https://github.com/nextflow-io/nextflow/pull/2589). 

This support could be implemented by extending the semantic of the label directive, in such a way that when a key-value is provided using the syntax `name=value` is implicitly interpreted as key-value pair that needs to be propagated to the corresponding job execution. 

This is expected to be an opt-in feature therefore the user should enabled either using the env variable: `NXF_ENABLE_LABELS_PROPAGATION=true` or setting the attribute `nextflow.enable.labelsPropagation=true` in the nextflow config file. 

### Supported platforms

* AWS Batch 
  - [Job definition](https://docs.aws.amazon.com/batch/latest/APIReference/API_RegisterJobDefinition.html)
  - [Job submit](https://docs.aws.amazon.com/batch/latest/APIReference/API_SubmitJob.html)
* Google LS
  - [Run pipeline](https://cloud.google.com/life-sciences/docs/reference/rest/v2beta/projects.locations.pipelines/run) allow to specified the labels key-value pairs to be used
* Azure batch (it looks like it does not provide any API for tags/labels annotation)
* Kubernets
  - The task pod should include the user-provided key-value pairs pod [labels](https://github.com/nextflow-io/nextflow/blob/3f247518fbd56074a97379bf31de138cc229f499/modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy#L178-L178).

### Validation 

Each platform has a slightly different set of validation rules for labels name and values. It's to be defined if nextflow should prevent the use of such of invalid pattern or delegate the check to the target platform

jorgeaguileraseqera:
Allow a new syntax in the `label` directive in key-value format

```
  label k: 'value'
  label b: "${params.user}"
```
when this syntax is used we store them in a new Map variable so executors can access them and tag the process if present

closes #2845


pditommaso:
Didn't we agree to go with `label 'name=value'` as first iteration"?

jorgeaguileraseqera:
yes, but once started with the implementation I've found this syntax more powerful and wanted to propose it 

(I'm still working on the issue so I can include also our first idea about `name=value`)

jorgeaguileraseqera:
This PR allows the following syntax:

```
process foo {
      label 'bravo'   //<--- a label as previously
      label 'department=floor 3' 
      label region:'eu-west-1' 
      label region:'eu-west-1' , department:'floor 3'
}
```

pditommaso:
Ok, I'll check soon 👍

modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
```diff
@@ -420,6 +420,10 @@ class TaskConfig extends LazyMap implements Cloneable {
         return CmdLineOptionMap.emptyOption()
     }
 
+    Map<String, String> getStickers() {
+        return (get('sticker') ?: Collections.emptyMap()) as Map<String, String>
+    }
+
```
pditommaso, line 426: Nice idea, but the `sticker` name would introduce a new concept, thats' something we are trying to prevent. Let's rename this to `resourceLabels` / `getResourceLabels`.

modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy
```diff
@@ -232,6 +232,7 @@ class K8sTaskHandler extends TaskHandler {
         result.taskName = task.getName()
         result.processName = task.getProcessor().getName()
         result.sessionId = "uuid-${executor.getSession().uniqueId}" as String
+        result.putAll(task.config?.getStickers() ?: Collections.emptyMap() as Map<String,String>)
         return result
```
pditommaso, line 236: Think the custom labels should not override "core" ones (same for Google LS)

pditommaso (changes requested):
Ok, nice job. The main comment is about naming, let's call these `resourceLabels`

modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
jorgeaguileraseqera: Ok, but this is "transparent" for the DSL, it uses `label` as usual and internally we maintain in a new concept

pditommaso: that's true, but it also a good thing keep the internal domain names consistent with the dsl

dougnukem:
This might be a different feature, but there could be value in allowing users to specify labels at the top-level like on the executor.

Then have those labels propagate down to the `process` to be set. Otherwise we'd have to set those labels/params on every single process.

https://github.com/nextflow-io/nextflow/issues/2845

```
google {
    labels = [ 'user-id': params.user_id, 'project-id': params.project_id ]
    project = 'theprojectid'
    zone = 'europe-west1-b'
}
```

pditommaso:
@dougnukem thanks a good point, however think the current proposal still works, because when setting something like in the config, it would apply to *all* processes 

```
process.labels = [ 'user-id': params.user_id, 'project-id': params.project_id ]
```

docs/process.rst
```diff
@@ -1674,6 +1674,19 @@ in the configuration file to select and configure subset of processes having sim
 
 See the :ref:`config-process-selectors` documentation for details.
 
+The ``label`` directive can be also expressed as a `Map<key-value>` or a `key=value` sentence:
```
pditommaso, line 1677: This now has been implemented with a new `resourceLabels` directive. Please update the docs

modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy
```diff
@@ -241,6 +241,7 @@ class K8sTaskHandler extends TaskHandler {
         if( labels ) {
             result.putAll(labels)
         }
+        result.putAll(task.config?.getResourceLabels() ?: Collections.emptyMap() as Map<String,String>)
```
pditommaso, line 244: Please add a unit test for this

pditommaso (changes requested):
Ok, this looks good a few comments to address. It should also be address the support for Google Batch

pditommaso:
I was looking how to set labels with Google batch but cannot find it, but I'm quite sure it's possible. @bentsherman any clue about that?

bentsherman:
@jorgeaguileraseqera the google batch labels are set through the AllocationPolicy: https://cloud.google.com/java/docs/reference/google-cloud-batch/latest/com.google.cloud.batch.v1.AllocationPolicy.Builder#com_google_cloud_batch_v1_AllocationPolicy_Builder_putAllLabels_java_util_Map_java_lang_String_java_lang_String__

docs/process.rst
jorgeaguileraseqera: @bentsherman can you take a look to (not sure if it's well expressed) https://github.com/nextflow-io/nextflow/pull/2853/files#diff-8b2232009593fccf012f8206c47ff7514a2fb5e4f98d190bbfafebbf36078c8fR1677

docs/process.rst
```diff
@@ -1674,6 +1674,26 @@ in the configuration file to select and configure subset of processes having sim
...
+  process bigTask {
+    resourceLabel organization: 'MyOrganization'
+    resourceLabel department: 'a department', group: 'a group'
+
+    '''
+    <task script>
+    '''
+  }
+
+ResourceLabels are useful, for example, to obtain information about jobs executed in a cloud environment
```
bentsherman, line 1696:
```suggestion
.. _process-resourceLabel:

resourceLabel
-----

The ``resourceLabel`` directive allows you to add metadata to cloud resources associated with a task. Resource labels are particularly useful for tagging compute resources to specific users or teams, e.g. for cost tracking. This feature is currently supported for AWS Batch, Google Batch, and Kubernetes.

Resource labels are specified as key-value pairs::

  process bigTask {
    resourceLabel organization: 'MyOrganization'
    resourceLabel department: 'a department', group: 'a group'

    '''
    <task script>
    '''
  }
```

pditommaso:
Still could not figure out how to use Google Batch API to setup labels. I've open a separate issue for that

beichen1024:
> Still could not figure out how to use Google Batch API to setup labels. I've open a separate issue for that

Do you have an issue for this?

bentsherman:
> Do you have an issue for this?

Fixed in #3170

beichen1024:
> > Do you have an issue for this?
> 
> Fixed in #3170

Actually, yes.
NF version: 22.09.4-edge
executor: google-lifesciences
I added 
```
process.labels = [ 'user-id': params.user_id, 'project-id': params.project_id ]
```
as mentioned in the discussion. And I used `gcloud beta lifesciences operations describe` to check the worker process operation, and in the `resources` section, I am not seeing the labels I added. And nothing showed up in .nextflow.log either.
```
resources:
      regions:
      - us-central1
      virtualMachine:
        bootDiskSizeGb: 500
        bootImage: projects/cos-cloud/global/images/family/cos-stable
        disks:
        - name: nf-pipeline-work
        labels:
          goog-pipelines-worker: 'true'  // only the google default labels
        machineType: custom-1-1024
        nvidiaDriverVersion: 450.51.06
        serviceAccount:
          email: default
          scopes:
          - https://www.googleapis.com/auth/cloud-platform
        volumes:
        - persistentDisk:
            sizeGb: 500
          volume: nf-pipeline-work
    timeout: 604800s
```

I think I have a couple of questions:
1. If it is successfully passed to all process, how do I know? where should I check it? (like what log files)
2. Are those labels suppose to show up in the `resource.labels` section?

Additional information:
If I include a script block in the process itself and use `$task.labels` I can print those values to an output file.

bentsherman:
@beichen1024 it should be `process.resourceLabels`

dougnukem:
@bentsherman The GCP labeling being added appears to be done at the Google Lifesciences Pipeline level, for this to apply to the GCP VMs for cost monitoring this needs to be done at the VirtualMachine.labels level.

https://github.com/nextflow-io/nextflow/pull/2853/files#diff-bd0a55afab4a626d773cac6787aef85e2d2756a3a25d438df70bb98ae0127f1fR479

I believe instead this should be applied instead or in addition at the Resource/VirtualMachine level.

https://cloud.google.com/life-sciences/docs/reference/rpc/google.cloud.lifesciences.v2beta#runpipelinerequest
> RunPipelineRequest
>  labels | map<string, string>User-defined labels to associate with the returned operation. These labels are not propagated to any Google Cloud Platform resources used by the operation, and can be modified at any time.To associate labels with resources created while executing the operation, see the appropriate resource message (for example, VirtualMachine).


https://cloud.google.com/life-sciences/docs/reference/rpc/google.cloud.lifesciences.v2beta#virtualmachine
> VirtualMachine
> labels | map<string, string>Optional set of labels to apply to the VM and any attached disk resources. These labels must adhere to the name and value restrictions on VM labels imposed by Compute Engine.Labels keys with the prefix 'google-' are reserved for use by Google.Labels applied at creation time to the VM. Applied on a best-effort basis to attached disk resources shortly after VM creation.
> 
>
> Optional set of labels to apply to the VM and any attached disk resources. These labels must adhere to the [name and value restrictions](https://cloud.google.com/compute/docs/labeling-resources) on VM labels imposed by Compute Engine.
>
> Labels keys with the prefix 'google-' are reserved for use by Google.
>
>Labels applied at creation time to the VM. Applied on a best-effort basis to attached disk resources shortly after VM creation.

PR #2861, opened 2022-05-11
bugfix: include revision in the Azure Repos provider if specified

jorgeaguileraseqera:
closes #2861 


pditommaso:
it was fast! think we need to add an integration test to verify it works as expected

jorgeaguileraseqera:
> it was fast! think we need to add an integration test to verify it works as expected

ah, ok. I see this test uses a NXF_AZURE_REPOS_TOKEN but not sure if it's injected in the github action

```
   @IgnoreIf({System.getenv('NXF_SMOKE')})
    @Requires({System.getenv('NXF_AZURE_REPOS_TOKEN')})
    def 'should fetch repo tags'() {
```

pditommaso:
Yes, it is

jorgeaguileraseqera:
Added 2 test to verify against a real Azure repo (one to test fake branch and another to check a file is retrieved)

pditommaso:

PR #2876, opened 2022-05-17
gets pipeline metainformation from revision if specified

### LINKED ISSUE #1988: nextflow pull gets pipeline metainformation from the master branch - even if other branch is specified.
## Bug report 

We had an empty repository in a BitBucket distribution. Following good practices the pipeline was developed on a feature branch. To our surprise it was not possible to pull or clone this branch. It seems that NextFlow obtains some metainformation about the pipeline always from the master branch. In our case the master branch was empty.

Here is an example command execution (URLs changed because of security, but they provide reasonable insights into the source of the problems):

```
> nextflow clone pipelines/mypipeline -hub my_bitbucket -r somebranch
WARN: Cannot read project manifest -- Cause: Remote resource not found: https://bitbucket.ourdomain.com/rest/api/1.0/projects/pipelines/repos/mypipeline/raw/nextflow.config
Remote resource not found: https://bitbucket.ourdomain.com/rest/api/1.0/projects/pipelines/repos/mypipeline/raw/main.nf
```

As you can see, the branch name is not used in the API calls. Clearly NextFlow obtain manifest information from the master branch instead of the branch that was selected.

### Environment 

* Nextflow version: 20.10.0

jorgeaguileraseqera:
Tested against a Github repo but must also work with other repositories as revision feature is implemented in all of them

closes #1988

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
@@ -481,7 +481,7 @@ class CmdRun extends CmdBase implements HubOptions {
             if( offline )
                 throw new AbortOperationException("Unknown project `$repo` -- NOTE: automatic download from remote repositories is disabled")
             log.info "Pulling $repo ..."
-            def result = manager.download()
+            def result = manager.download(revision)
             if( result )
                 log.info " $result"
             checkForUpdate = false
--- a/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
@@ -388,8 +388,8 @@ class AssetManager {
     }
 
     AssetManager checkValidRemoteRepo(String revision=null) {
-        def scriptName = getMainScriptName()
         provider.revision = revision
+        def scriptName = getMainScriptName()
         provider.validateFor(scriptName)
         return this
     }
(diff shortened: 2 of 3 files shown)


modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -481,7 +481,7 @@ class CmdRun extends CmdBase implements HubOptions {
             if( offline )
                 throw new AbortOperationException("Unknown project `$repo` -- NOTE: automatic download from remote repositories is disabled")
             log.info "Pulling $repo ..."
-            def result = manager.download()
+            def result = manager.download(revision)
```
pditommaso, line 484: Not sure about this, because there's an explicit checkout immediately after. But I guess the revision is needed here exactly to prevent this issue

jorgeaguileraseqera: exactly
download function tries to validate if it's a valid repo but currently is doing checking the master branch

pditommaso:
There's a similar problem in [getMainScriptName](https://github.com/nextflow-io/nextflow/blob/b23b5e17c1672bb603ecf36a235cc3980265e08a/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy#L391-L391) method. 

Basically, this method tries to read the `manifest.mainScript` attribute in the `nextflow.config`. This should be read from the revision specified for the user, instead, now it's fetching it from the default branch.

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
```diff
@@ -388,8 +388,8 @@ class AssetManager {
     }
 
     AssetManager checkValidRemoteRepo(String revision=null) {
-        def scriptName = getMainScriptName()
         provider.revision = revision
+        def scriptName = getMainScriptName()
         provider.validateFor(scriptName)
```
pditommaso, line 393: oh was that easy! 😄 let's add a comment explaining why this order matter

pditommaso:

PR #2885, opened 2022-05-20
bugfix: parse env booleans

jorgeaguileraseqera:
small issue parsing env variables as boolean

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
@@ -220,7 +220,7 @@ class CmdRun extends CmdBase implements HubOptions {
     String withConda
 
     @Parameter(names=['-offline'], description = 'Do not check for remote project updates')
-    boolean offline = System.getenv('NXF_OFFLINE') as boolean
+    boolean offline = System.getenv().getOrDefault('NXF_OFFLINE', "false").toBoolean()
 
     @Parameter(names=['-entry'], description = 'Entry workflow name to be executed', arity = 1)
     String entryName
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -405,7 +405,7 @@ class CmdRun extends CmdBase implements HubOptions {
     }
 
     private static boolean ignoreHistory() {
-        System.getenv('NXF_IGNORE_RESUME_HISTORY')=='true'
+        System.getenv().getOrDefault('NXF_IGNORE_RESUME_HISTORY',"false").toBoolean()=='true'
```
pditommaso, line 408: I think this is wrong because compare `Boolean` with a string therefore it will be always `false`. 

I think it would be better to only change  `NXF_OFFLINE` using the same "pattern" `System.getenv('NXF_OFFLINE')=='true' which is a bit more safe, because  the `toBoolean` could thrown an exception for a non `true|false` string

jorgeaguileraseqera: toBoolean doesnt throw any exception, only returns `true` if the string is "true", "1"  or "y" 

https://docs.groovy-lang.org/latest/html/groovy-jdk/java/lang/String.html#toBoolean()

pditommaso: I dont know this, and actually is nice, but my concern is to have those kinds of subtle changes across versions and attributes how boolean is interpreted.  

Maybe we should look for variables and change all to use `toBoolean` but I think it' outside the scope of this patch

pditommaso:
Merged manually 40bf2b2a. But checking better is a true mess, boolean are converted using `as Boolean`, `toBoolean`, `Boolean.parseBoolean`, `.toString() == 'true'`

We should make look to all those usage and use a uniform conversion method.

Issue #1694, opened 2020-08-04
Optional inputs


illusional:
## New feature

Pinging @rsuchecki @pditommas (I couldn't find an issue with this, I hope it's okay that I open a new one).

Based on a small conversation on the Gitter ([1 - primary](https://gitter.im/nextflow-io/nextflow?at=5f2019b0e90668200521dfbb) | [2](https://gitter.im/nextflow-io/nextflow?at=5f221341d50f2b289aa3e8d7)), there's interest (a lot from me) to have more direct support for optional inputs - this seems is inline with the goals of DSL2 to produce reusable tool modules / interfaces.

Other workflow specifications have the concept of tool wrappers, which aim to be a "write once, use in all of your workflows". This means the tool wrapper would contain most (if not all) available configuration options, which then the command line is dynamically constructed. This allows the community to build and contribute high quality tool wrappers, for example: [Common Workflow Library](https://github.com/common-workflow-library/bio-cwl-tools) ([CWLibrary#fastqc](https://github.com/common-workflow-library/bio-cwl-tools/blob/release/fastqc/fastqc_2.cwl)), [BioWDL](https://github.com/biowdl/tasks) ([BioWDL#fastqc](https://github.com/biowdl/tasks/blob/develop/fastqc.wdl)) with the tools available for other users to use, or upload to stores like [Dockstore](https://dockstore.org/) or the Galaxy toolshed.

Projects like [aCLImatise](https://github.com/aCLImatise/CliHelpParser) aim to generate tool wrappers, as this process is usually a significant time consuming aspect of building workflows. 

The DSL2 makes good strides towards this, and a stronger concept for optional inputs would take this further.

Relevant discussion:

- [Nextflow patterns#optional-input](https://github.com/nextflow-io/patterns/blob/master/docs/optional-input.adoc)
- [Google Groups: Optional input channels in DSL2](https://groups.google.com/forum/#!topic/nextflow/5qeS4Mh1oMs)

### Command line construction sidenote

I think it would be a bad idea to create a new syntax for building or interpolating command lines, but tool developers could use the groovy environment to build strings for each command option.

## Usage scenario 

Consider fastqc (eg: [nf-core module definition](https://github.com/nf-core/modules/blob/master/software/fastqc/main.nf)), which might have the (simplified) command structure:

```bash
fastqc \
    [-c contaminant file] \
    [ ... other config options ] \
    seqfile1 .. seqfileN
```

[...]

Issue #2257, opened 2021-08-09
Support named channel inputs for processes and workflows


nh13:
## New feature

I would like to have the DSL2 support specifying the channels for processes and workflows via keywords.
Some workflows I've written require many channels (5, 10, even 20!).
It's hard to make sure that the order I specify is the same as the order I require.

## Usage scenario 

```nextflow

workflow foo {
    take:
    ch_in_1
    ch_in_2
    ...
    ch_in_100

    main:
    ...
}

workflow bar {
    ch_in_1   = Channel.from(...)
    ch_in_2   = Channel.from(...)
    ...
    ch_in_100 = Channel.from(...)

    ...
    foo(
        ch_in_1:    ch_in_1,
        ch_in_2:    ch_in_2,
        ....
        ch_in_100:  ch_in_100,
    )
}
```

## Suggest implementation 

?

pditommaso:
This is a good point. Now that DSL2 allows named outputs it makes sense to have named input arguments as well

nh13:
I should circle back to this. Keeping it open.

bentsherman:
Closing as not planned since we ended up supporting similar functionality with records

The best practice with typed processes is to define a single record input and call it with a channel of records. You can use `cross` with named arguments to add singleton values to each record. This approach is more flexible than using named arguments on the process call directly

Similarly, pipeline composition (#7213) proposes importing a pipeline as a named workflow with a single record input rather than using named arguments

We can re-open if we decide to revisit this

PR #2893, opened 2022-05-23
Nullable input/output paths

### LINKED ISSUE #2678: DSL2 - emit tuples with optional values

jorgeaguileraseqera:
new feature to allow nonexisting paths in input and output parameters

closes #2678 

p.d.
it has 2 tests ignored in `ScriptDslTest` because with them `ScriptRunnerTest` fails by Timeout but don't know why


modules/nextflow/src/main/groovy/nextflow/util/NullablePath.groovy
```diff
@@ -0,0 +1,31 @@
+package nextflow.util
+
+
+/**
+ * @author : jorge <jorge.aguilera@seqera.io>
+ *
+ */
+class NullablePath implements CharSequence{
```
pditommaso, line 8: Have you explored the possibility to have `NullablePath implements Path` and delegate the implementation to a fack local path object 

For example https://github.com/nextflow-io/nextflow/blob/c70eb12d1caac1ca3e1acb7c5b2a0adbfae18ebe/modules/nextflow/src/main/groovy/nextflow/scm/ProviderPath.groovy#L45-L52

jorgeaguileraseqera: yes, I've considered it and in this way include NullablePathS to the list of files produced for the task

On the one hand, the code looks great and is simple but I was not sure if we wanted to have this kind of file because the user will need to check if it exists

pditommaso: I think if we go for this approach should be consistent with the expected type, `Path` in this case.

jorgeaguileraseqera: Ok, I'll change to Path

jorgeaguileraseqera:
In the end, I think I've found the way to run all tests although it's a little tricky and probably we need to investigate deeper

Don't know why/where if we run in the same `./gradlew test` command InputNullableTest + OutputNullableTest + ScriptDslTest some of them fail and I think it's due to some static variable related to the execution is not refreshed

So to solve this situation I exclude InputNullable and OutputNullable from the tests and run them in an isolated test

nh13:
Why do we prefer nullable versus optional? The latter could have type Option.  Which is more idiomatic of something that is optional? Also, nullable is less intuitive to folks who don’t know what null means.

pditommaso:
The rationale was described at point 3 [here](https://github.com/nextflow-io/nextflow/pull/2710#issuecomment-1073674150). To summarise we are using already `optional` with a different semantic that's when an optional value is absent no value is emitted. 

This is quite different from emitting a value with a `null` value. Therefore using `optional` for both can be quite confusing.

bentsherman:
Are we calling it `allowNull` or `nullable` ? I think we should call it `nullable` like an adjective.

pditommaso:
I agree, `nullable` sounds better

bentsherman:
We need to improve the e2e pipeline tests, ideally we should have one process emit a nullable path to another process, and it succeeds or fails based on whether or not the nullable paths are specified correctly.

Also, the original motivation for this feature was a tuple that contains a path. I wonder what happens if a process emits a nullable path by itself. I guess it's just a null value?

pditommaso:
@jorgeaguileraseqera This was not supported to be merged. Above all it looks like you force pushed your branch overriding the changes in this PR

jorgeaguileraseqera:
Yes, I have it

I'm reviewing it and trying to reopen the PR. 

Sorry for inconveniences

illusional:
Just to confirm that this would close https://github.com/nextflow-io/nextflow/issues/1694?
If so, could you link the issue from this PR (so merging automatically closes)?

bentsherman:
Not quite. This PR pertains only to nullable paths, whereas that issue pertains to optional inputs in general, and I don't think we have fully worked out the meaning of ?optional", i.e. passing a null or passing nothing. Also, I think #1694 is tied up with #2257 , because you might want to have named inputs so that you can provide only the inputs you want.

bentsherman:
@pditommaso I cleaned up this PR and updated the e2e tests. Also fixed a bug with the nullable input path. I think this PR is ready to go.

Here's the output of the main e2e test, in which process foo emits a nullable path and process bar accepts it:
```console
$ ./launch.sh tests/nullable-path-succeeds.nf 
N E X T F L O W  ~  version 23.02.0-edge
Launching `tests/nullable-path-succeeds.nf` [naughty_fermi] DSL2 - revision: 41393792fb
executor >  local (2)
[de/f4993a] process > foo (1) [100%] 1 of 1 ✔
[51/badde8] process > bar (1) [100%] 1 of 1 ✔
foo
output.txt
output.txt
```

It works but maybe we should figure out how to suppress the file attributes warning?

bentsherman:
In case the unit tests don't run, I confirmed that they pass locally:
```console
$ make smoke
NXF_SMOKE=1 ./gradlew test

BUILD SUCCESSFUL in 3m 14s
91 actionable tasks: 39 executed, 52 up-to-date
```

bentsherman:
While I understand the difference between optional and nullable, I'm wondering if we shouldn't just extend `optional` to be allowed within tuple elements and handle it accordingly, rather than adding a new option. In other words:
- if a tuple or standalone path is declared optional, then emit nothing
- if a nested path is declared optional, then set the tuple element to NullPath

The only thing you lose that way is making a standalone path nullable vs optional, but I don't think that is needed.

pditommaso:
I agree trying to not overload the declaration with new keyword, but what about point 3 [here](https://github.com/nextflow-io/nextflow/pull/2710#issuecomment-1073674150) (I copy it below for your convenience): 

> I think using optional for both the existing mechanism and the new one could bring a serious semantics inconstancy. When using the former the output is NOT emitted when missing. Using the new approach is emitted with null. Imagine the explain the difference between output: val(x) optional and output: val(x, optional:true).

bentsherman:
I would avoid that issue by having the nullable behavior only apply to nested params in a tuple, that way there is no distinction between `val(x), optional: true` and `val(x, optional: true)`.

However, your comment [here](https://github.com/nextflow-io/nextflow/pull/3687#issuecomment-1445785338) makes me think there is a case for standalone paths to be nullable. So now I am leaning back towards having a separate `nullable` option as it is currently implemented.

roskamsh:
Hi there - is there any update on whether this will be merged into the main branch and released with the next version of nextflow? I have a use case in my current pipeline that would really benefit from this (specifically, using the remainder:true option when joining two channels that I pass onto a subsequent process).

bentsherman:
I do need to fix the merge conflict, but otherwise it's just waiting for @pditommaso 's approval.

Issue #2867, opened 2022-05-12
HyperQueue executor 


Nortamo:
## New feature

HyperQueue executor to nextflow (https://it4innovations.github.io/hyperqueue/stable/)

(As a side note we are also looking at an executor for https://github.com/flux-framework/flux-core to enable large workflows with mpi components.)

## Usage scenario 

Running large workflows with very many processes on HPC systems ( ~ 10K or more  ).
The current batch executors (especially the slurm one https://www.nextflow.io/docs/latest/executor.html#slurm ) 
are not suitable to due to the usage of individual jobs, and even if jobsteps were to be used, the amount would be to large,
negatively affecting the HPC scheduler stability. 

So the idea for maximum throughput would be to launch HyperQueue using Slurm (I think they also suport PBS), and then start
nextflow.

## Suggest implementation 

Most of the work is just straightforwardly creating a new Executor which extends the  `AbstractGridExecutor`,
e.g `HqExecutor`.  

Main unsolved issue is that HyperQueue requires the submissions script (i.e the `.command.run` file for each process) to be executable, which is currently
not the case. While prototyping I just added `targetWrapperFile().setExecutable(true,true)`  in the `build` method of the `BashWrapperBuilder` class in `modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy `, but this is perhaps not the way you would like this done, and I'm wondering what would be the best way to achieve this?

So far we managed to run some simple toy workflow on 20 nodes (2560 CPU cores) without apparent issues. Also interested in hearing if you think there will be any issues in nextflow (apart from the file count) if the number of parallel processes is increased more.

pditommaso:
This is interesting. If I'm understanding correctly, you are suggesting submitting the nextflow run as a Slurm job, then the workflow tasks would be run via the HyperQueue executor. Is that correct? 

The support for Flux is also interesting, please report it as a separate ticket.

Nortamo:
Yes. 
I'll create a separate issue for the flux use case.

pditommaso:
> Main unsolved issue is that HyperQueue requires the submissions script (i.e the .command.run file for each process) to be executable

There's a similar need also for the Condor executor. You can see it here 

https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/executor/CondorExecutor.groovy#L158-L159

[...]

PR #2896, opened 2022-05-24
Hyperqueue executor (WIP)

Nortamo:
This PR adds support for the [HyperQueue](https://github.com/It4innovations/hyperqueue) scheduler.
Earlier discussion in #2867 

> Running large workflows with very many processes on HPC systems ( ~ 10K or more ).
The current batch executors (especially the slurm one https://www.nextflow.io/docs/latest/executor.html#slurm )
are not suitable to due to the usage of individual jobs, and even if jobsteps were to be used, the amount would be to large,
negatively affecting the HPC scheduler stability / hitting hard limits.
So the idea for maximum throughput would be to launch HyperQueue using the system batch scheduler and then start
nextflow.

Supported features are:

- clusterOptions
- cpus
- memory
- time
- accelerator ( type of accelerator is ignored)

**Note** requires hyperqueue v0.11 which is not released yet.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/HyperQueueExecutor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/HyperQueueExecutor.groovy
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2020-2022, Seqera Labs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package nextflow.executor
+
+import java.nio.file.Path
+
+import groovy.json.JsonSlurper
+import groovy.transform.CompileStatic
+import nextflow.processor.TaskRun
+import nextflow.util.ServiceName
+import nextflow.util.TupleHelper
+/**
+ * Implements executor for HyperQueue
+ *  https://github.com/It4innovations/hyperqueue/
+ *
+ * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
+ */
+@CompileStatic
+@ServiceName('hq')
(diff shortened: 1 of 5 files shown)


modules/nextflow/src/main/groovy/nextflow/executor/HyperQueueExecutor.groovy
```diff
@@ -0,0 +1,147 @@
...
+import groovy.json.JsonSlurper
+import groovy.transform.CompileStatic
+import nextflow.processor.TaskRun
+import nextflow.util.ServiceName
+import nextflow.util.TupleHelper
+/**
+ * Implements executor for HyperQueue
+ *  https://github.com/It4innovations/hyperqueue/
+ *
+ * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
```
pditommaso, line 31: Please, add yourself as author

pditommaso (changes requested):
Great job. It would be nice to add a few tests for the resources you have added ie. mem and GPUs support

pditommaso:
This could be merged provided a warning message is added to want this is an experimental feature. It could done adding something like this in the HyperQueueExecutor class

```
    @Override
    protected void register() {
        super.register()
        log.warn "The support for HyperQueue is an experimental feature and it may change in a future release"
    }
```

Nortamo:
> Great job. It would be nice to add a few tests for the resources you have added ie. mem and GPUs support

Sure, 

> Great job. It would be nice to add a few tests for the resources you have added ie. mem and GPUs support

I'll throw some together.

modules/nextflow/src/main/groovy/nextflow/executor/HyperQueueExecutor.groovy
```diff
@@ -0,0 +1,153 @@
...
+    @Override
+    protected List<String> getDirectives(TaskRun task, List<String> result) {
+
+        result << '--name' << getJobNameFor(task)
+        result << '--log' << quote(task.workDir.resolve(TaskRun.CMD_LOG))
+        result << '--cwd' << quote(task.workDir)
+
+        // No enforcement, Hq just makes sure that the allocated value is below the limit
+        if( task.config.getMemory() )
+            result << '--resource mem=' + (task.config.getMemory().toBytes()) << ''
```
pditommaso, line 60: It would be a bit cleaner to use instead 

result << '--resource' << "mem=${task.config.getMemory().toBytes()}"

Same for the following ones

pditommaso (approved):

pditommaso:
Ok, let's merge this. Thanks for the awesome contribution!

pditommaso:

PR #2903, opened 2022-05-25
Add min/max/default value for resource directives

### LINKED ISSUE #640: Option to limit process requirements instead of exiting with error
If Nextflow encounters a process that requests resources that are not available on the current system, it fails. For example, if 10 processors are requested and the system only has 8, we exit with an error. This is A Good Thing 👍 

However, when making config files for a pipeline that aim to have generic works-for-most-situations requests, it can be nice to be able to throttle these for small genomes / test datasets / small machines for example.

In most of our pipelines, we now have [this little helper function](https://github.com/nf-core/methylseq/blob/2b2d801bf26bd34bc3ba4c6b9e024b947410dad5/nextflow.config#L105-L136) in `nextflow.config`:

```groovy
// Function to ensure that resource requirements don't go beyond
// a maximum limit
def check_max(obj, type) {
  if(type == 'memory'){
    try {
      if(obj.compareTo(params.max_memory as nextflow.util.MemoryUnit) == 1)
        return params.max_memory as nextflow.util.MemoryUnit
      else
        return obj
    } catch (all) {
      println "   ### ERROR ###   Max memory '${params.max_memory}' is not valid! Using default value: $obj"
      return obj
    }
  } else if(type == 'time'){
    try {
      if(obj.compareTo(params.max_time as nextflow.util.Duration) == 1)
        return params.max_time as nextflow.util.Duration
      else
        return obj
    } catch (all) {
      println "   ### ERROR ###   Max time '${params.max_time}' is not valid! Using default value: $obj"
      return obj
    }
  } else if(type == 'cpus'){
    try {
      return Math.min( obj, params.max_cpus as int )
    } catch (all) {
      println "   ### ERROR ###   Max cpus '${params.max_cpus}' is not valid! Using default value: $obj"
      return obj
    }
  }
}
```

Then, [in each config](https://github.com/nf-core/methylseq/blob/2b2d801bf26bd34bc3ba4c6b9e024b947410dad5/conf/base.config#L20-L22) we wrap the request in the above function, as so:

```groovy
cpus = { check_max( 1 * task.attempt, 'cpus') }
memory = { check_max( 8.GB * task.attempt, 'memory') }
time = { check_max( 2.h * task.attempt, 'time') }
```

This way, the defaults work most of the time. But if you pass `--max_memory "8.GB"` then all processes will use a maximum of `8.GB` and the pipeline will run on your tiny machine. This is really helpful for test datasets and allows us to have a default "base" config file that is throttled by other system-specific configs.

Is this something that could be built into Nextflow as a new feature? Keep the current default behaviour, but have new command line arguments and config scopes to be able to throttle the maximum resources without failing..

Phil

bentsherman:
Closes #640 (also addresses parts of #773, #785)

This PR adds the ability to specify min/max/default values for resource directives (`cpus`, `memory`, `time`, `accelerator`). In general, setting a maximum will allow Nextflow to respect system-wide resource limits, which are currently enforced in nf-core pipelines through the `check_max` function. For Kubernetes in particular, min/max values may correspond to `request` and `limit` in a Pod spec, allowing tasks to have a variable amount of resources over time.

Here are the proposed changes:
- Define classes for `CpuResource`, `MemoryResource`, `TimeResource`, `AcceleratorResource`
- Allow each resource to be specified in the following ways:
  - number (e.g. `cpus = 2`)
  - number with params (e.g. `cpus = 2, max: 4`)
  - map (e.g. `cpus = [value: 2, min: 1, max: 4]`)
- Update pod spec builder to use min and max values for `cpus`, `memory`, `accelerator`
- Update task submitter to apply max thresholds to resource requests before submitting task

The trickiness with this feature is when and how to apply default values. I am still thinking through this part but the logic that I settle on will be reflected in the unit tests.


bentsherman:
There is a conflict between using `limit` to enforce a resource limit vs using `limit` in the context of Kubernetes. In the former case, `limit` means "don't let this resource exceed this limit" whereas in the latter case it means "use this limit in the pod spec".

As a result, I'm not sure that we can use `limit` to implement both meanings. If a user defines `cpus = 2, limit: 64` thinking about resource limits on their HPC cluster, and then runs the pipeline on a k8s cluster, Nextflow will create a pod with 2 cpus that can __burst up to 64 cpus__. Technically it should be fine, but if the task actually tries to burst for whatever reason then it will be unexpected behavior.

Although it's ugly, the only solution may be to use `request` and `limit` only for k8s, and use a separate param such as `max` to enforce system resource limits.

bentsherman:
I am going to try a different approach suggested at https://github.com/nextflow-io/nextflow/issues/640#issuecomment-1139772761 , where by global resource limits are defined in a separate process directive, so that we don't need to extend each process directive individually. This PR is still relevant for implementing `request` and `limit` for K8s (see #785) so I will leave it up for that.

bentsherman:
Regarding k8s request/limit, I am beginning to think it would be better to support this through `pod` directive, similar to `clusterOptions` for executor-specific resource settings. AFAIK K8s is the only executor that supports this kind of upper and lower bound on resource requests.

pditommaso:
> AFAIK K8s is the only executor that supports this kind of upper and lower bound on resource requests

well not really, all HPC batch schedulers have soft and hard resource requirements, that correspond to K8s request/limit model.

pditommaso:
If I'm understanding correctly https://github.com/nextflow-io/nextflow/pull/2911 makes this PR not needed anymore

bentsherman:
Not quite, because this PR would still be needed to implement k8s request/limit for `cpus` and `memory` directives. And if other HPC schedulers support hard/soft resource settings as you say, it could be used for that as well.

bentsherman:
Closing because this thread got confusing. Will open a new PR with the current purpose.

PR #2908, opened 2022-05-26
feature: stripe sensitive data into strings

### LINKED ISSUE #2905: Hide sensitive data from SCM content trace
The content of the SCM file is dumped for troubleshooting purposes. See here 

https://github.com/nextflow-io/nextflow/blob/c3117ada3e4e2b661baf6c3456cd9f2c092be85e/modules/nextflow/src/main/groovy/nextflow/scm/ProviderConfig.groovy#L277-L277

However, it contains sensitive data that should be redacted. An heuristic implementation like done [here](https://github.com/nextflow-io/nextflow/blob/c3117ada3e4e2b661baf6c3456cd9f2c092be85e/modules/nextflow/src/main/groovy/nextflow/scm/ProviderConfig.groovy#L305-L305), should provide a fair solution

jorgeaguileraseqera:
search and replace for password, token, secret and licence keywords as key:value string

closes #2905

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/scm/ProviderConfig.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/ProviderConfig.groovy
@@ -274,7 +274,7 @@ class ProviderConfig {
     static Map getFromFile(Path file) {
         try {
             final content = file.text
-            log.trace "Parsing SCM config path: ${file.toUriString()}\n${content}\n"
+            log.trace "Parsing SCM config path: ${file.toUriString()}\n${StringUtils.stripSecrets(content)}\n"
             final result = parse(content)
             dumpConfig(result)
             return result
--- a/modules/nextflow/src/main/groovy/nextflow/scm/ProviderConfig.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/ProviderConfig.groovy
@@ -273,12 +273,10 @@ class ProviderConfig {
     @PackageScope
     static Map getFromFile(Path file) {
         try {
+            // note: since this can be a remote file via e.g. read via HTTP
+            // it should not read more than once
             final content = file.text
-            try {
-                log.trace "Parsing SCM config path: ${file.toUriString()}\n${StringUtils.stripSecrets(content)}\n"
-            }catch(Exception e){
-                log.error "Error dumping configuration ${file.toUriString()}", e
-            }
+            dumpScmContent(file, content)
             final result = parse(content)
             dumpConfig(result)
             return result
@@ -301,6 +299,14 @@ class ProviderConfig {
         catch( Exception e ) {
             final message = "Failed to parse config file '${file?.toUriString()}' -- Cause: ${e.message?:e.toString()}"
             throw new ConfigParseException(message,e)
+        }
+    }
+
+    static private void dumpScmContent(Path file, String content) {
+        try {
+            log.trace "Parsing SCM config path: ${file.toUriString()}\n${StringUtils.stripSecrets(content)}\n"
+        }catch(Exception e){
+            log.debug "Error dumping configuration ${file.toUriString()}", e
         }
     }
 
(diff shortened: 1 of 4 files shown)


modules/nextflow/src/main/groovy/nextflow/scm/ProviderConfig.groovy
```diff
@@ -274,7 +274,7 @@ class ProviderConfig {
     static Map getFromFile(Path file) {
         try {
             final content = file.text
-            log.trace "Parsing SCM config path: ${file.toUriString()}\n${content}\n"
+            log.trace "Parsing SCM config path: ${file.toUriString()}\n${StringUtils.stripSecrets(content)}\n"
```
pditommaso, line 277: let's wrap into a separate method like this 

https://github.com/nextflow-io/nextflow/blob/c3117ada3e4e2b661baf6c3456cd9f2c092be85e/modules/nextflow/src/main/groovy/nextflow/scm/ProviderConfig.groovy#L303-L310

I've seen StackOverflow exception thrown by regex recently 🤦. better to be sure there's any side effect for this logging

modules/nf-commons/src/main/nextflow/util/StringUtils.groovy
```diff
@@ -47,6 +50,42 @@ class StringUtils {
...
+        addMaskPattern("\"token\"\\s*:\\s*\"(.*?)\"")
+        addMaskPattern("secret\\s*:\\s*\"(.*?)\"")
+        addMaskPattern("\"secret\"\\s*:\\s*\"(.*?)\"")
+        addMaskPattern("licence\\s*:\\s*\"(.*?)\"")
+        addMaskPattern("\"licence\"\\s*:\\s*\"(.*?)\"")
+    }
+
+    static String stripSecrets(String message) {
+        if (multilinePattern == null) {
+            return message;
```
pditommaso, line 73: Not understanding the need for this

jorgeaguileraseqera: do you say the `if`? 
well, it's only a safeguard in the case for whatever reason someday we remove the static block who initializes the list

can be removed of course

pditommaso: yes, but then why it's set [here](https://github.com/nextflow-io/nextflow/pull/2908/files#diff-3e31ee682cd3154c236f35dd401e69d8094a313d9880ef1ae55e94304e57530dR57)? maybe it can be just removed?

pditommaso: (I had the feeling I wrote a similar comment, but can't see it, anyhow) Still understanding the logic of using this static pattern that's the last created by `addMaskPattern` ?

https://github.com/nextflow-io/nextflow/pull/2908/files#diff-3e31ee682cd3154c236f35dd401e69d8094a313d9880ef1ae55e94304e57530dR76

jorgeaguileraseqera: basically, the idea is that the static block is creating a *multiline* regular expression as `"password": value` or `password  : "value"`  with every `addMaskPattern` 

after when we want to stripe a string we iterate over the multilinePattern  groups replacing occurrences with `*` 

Other similar methods we have are oriented to work with Map but the intent of this is to work with raw Strings

(not sure if it's what you are asking)

pditommaso: Looks a bit convoluted, would not much easier doing something like `secret|password|blahblah\s*:\s*"(.*?)"` 

Also not the format is `password = 'something'` or `password = "something"`  (not with `:`, it's a groovy config basically)

jorgeaguileraseqera: true, I missed the equals format

To me is easier to read and maintain a list of expressions than a big expression, but no problem, I'll change it

pditommaso:
🙌

pditommaso (approved):

PR #2911, opened 2022-05-27
Add resourceLimits directive

### LINKED ISSUE #640: Option to limit process requirements instead of exiting with error

bentsherman:
Closes #640 

This PR takes a simpler approach to global resource limits. It adds a `resourceLimits` directive which is simply a map of limits for each resource directive:
```groovy
process {
    resourceLimits = [
        cpus: 4,
        memory: 8.GB,
        time: 16.h
    ]
}
```

The getter methods for these resources in `TaskConfig` have been updated to apply the resource limit if it is defined.


drpatelh:
Ey, when will this be available in an official NF release? The maintainers team on nf-core is considering to rename all the `--max_*` parameters in nf-core pipelines because they are commonly misunderstood. This would solve the problem!

bentsherman:
I just brought it up to date, but it would be good if someone could test it with a real pipeline.

docs/process.md
```diff
@@ -2127,6 +2127,34 @@ Resource labels are currently only supported by the {ref}`awsbatch-executor`, {r
...
+### resourceLimits
+
+:::{versionadded} 23.08.0-edge
+:::
+
+The `resourceLimits` directive allows you to specify environment-specific limits for task resource requests. Resource limits can be specified in a process as follows:
+
+```groovy
+process my_task {
+    resourceLimits [
```
mirpedrol, line 2141:
```suggestion
    resourceLimits = [
```

mirpedrol (commented):
Hello,

I am testing this with the nf-core template. I have the changes in a [branch](https://github.com/nf-core/testpipeline/tree/test-resourcelimits) in case someone wants to have a look.

I set `resolurceLimits` in a [config file](https://github.com/nf-core/testpipeline/blob/test-resourcelimits/conf/base.config#L16-L20) thinking that this would limit the maximum resources for all modules. I set the maximum memory to `6.GB`. But when I run the pipeline, the module [fastqc](https://github.com/nf-core/testpipeline/blob/test-resourcelimits/conf/base.config#L43-L47) uses  `36.GB`, and I get the following error:

```
ERROR ~ Error executing process > 'NFCORE_TESTPIPELINE:TESTPIPELINE:FASTQC (SAMPLE1_PE_T1)'

Caused by:
  Process requirement exceeds available memory -- req: 36 GB; avail: 16 GB
```

Am I missing something? Should I set `resourceLimits` in a different file?

ewels:
Is it because `cpus` etc should also be using `=` instead of `:` @mirpedrol ?

ewels:
```diff
    resourceLimits = [
-        cpus: 1,
-        memory: 6.GB,
-        time: 4.h
+        cpus = 1,
+        memory = 6.GB,
+        time = 4.h
    ]

**Comment by ewels** [comment]
(I'm guessing no, as @bentsherman explicitly said a map, but still - just checking..)

**Comment by mirpedrol** [comment]
I see the same error with `=`

**Comment by bentsherman** [comment]
Julia's code looks right, it should be a map. I will try this pipeline when I'm back next week and see why the limits aren't being applied.

File: docs/process.md (hunk 1 above)
Comment (bentsherman) [reply docs/process.md:2141]: The equals sign is used in the config but not in the process definition. I have added examples for both.

**Comment by bentsherman** [comment]
@mirpedrol how are you running your local build of Nextflow with this PR? You have to run `make compile` and use `./launch.sh` in place of `nextflow`.

However, when I try to do this with your testpipeline, I get this error:
```
ERROR ~ Install is not supported on dev mode - Missing plugin nf-validation
```

So I guess the `launch.sh` doesn't work with custom plugins. I tried to pin the version and download the plugin separately but got the same error.

@pditommaso is there a way to use `launch.sh` with non-core plugins? Or do you have to install the JAR and run it with `nextflow`?

**Comment by bentsherman** [comment]
Putting dev mode aside, I was able to run Julia's test pipeline with the resource limits and it worked.

```bash
# build nextflow locally
git clone https://github.com/nextflow-io/nextflow -b issue-640_resource-limit-directive
cd nextflow
make install

# build number should be 5871
nextflow info

# run pipeline
nextflow run nf-core/testpipeline -r test-resourcelimits -profile test,conda

# check resource limits
nextflow config nf-core/testpipeline -profile test,conda | grep resourceLimits

# check task resources of previous run
nextflow log last -f hash,cpus,memory,time
```

All of the resource limits were correctly applied.

mirpedrol:
I cloned the branch, run `make compine`, but then I used `my/path/netflow` instead of `launch.sh`. If I use `nextflow info`, I get a different build number
```
  Version: 23.07.0-edge build 5870
  Created: 22-07-2023 15:44 UTC (17:44 CEST)
  System: Mac OS X 13.4.1
  Runtime: Groovy 3.0.18 on Java HotSpot(TM) 64-Bit Server VM 11.0.15+8-LTS-149
  Encoding: UTF-8 (UTF-8)
```
Running with `launch.sh` I see the same error 
```

ERROR ~ Install is not supported on dev mode - Missing plugin nf-validation
```
Following your steps with `make install` worked perfect now 😄

docs/process.md
mirpedrol: Oh I see, it's clear now, thanks!

geoffjentry:
This seems like it'd help simply common process resource management techniques

pditommaso:
@bentsherman not sure how to solve conflicts in `TaskConfig`, can you have a look please

modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
```diff
@@ -235,63 +235,74 @@ class TaskConfig extends LazyMap implements Cloneable {
...
+    private Duration getDuration0(Object value, String key) {
         try {
-            new Duration(value.toString().trim())
+            return Duration.of(value)
         }
         catch( Exception e ) {
             throw new AbortOperationException("Not a valid `$key` value in process definition: $value")
         }
-
     }
```
pditommaso, line 296: What's the value of this? only throwing the AbortOperationException on error?

bentsherman: I almost got rid of it, but `getDuration0` is useful to throw an error message with the correct name, as it could be `time`, `resourceLimits.time`, or `maxSubmitAwait`

pditommaso: Ok, thanks. I've reshaped a bit to make it cleaner

pditommaso:
@bentsherman any clue why docs tests fail?

bentsherman:
It's the same troublesome test `process-out-env.nf`:

https://github.com/nextflow-io/nextflow/blob/master/docs/snippets/process-out-env.nf

Thought I fixed it but I guess not. I think it is flaky because it is listing a task directory which maybe gives slightly different output in different environments. Maybe we should just make the snippet simpler.

pditommaso:
Second try it passed. If it cannot be found, the reason should be to consider removing the test. It's a bit annoying having random failure

ewels:
YEAAAAAH!!! Good bye nf-core `check_max()` config function!! You won't be missed!!! 🤣 

Yessss

PR #2914, opened 2022-05-29
Add dry-run like capability

pditommaso:
This PR adds the ability to partially run the script execution, without submitting any job execution. 

To some extent, it may be comparable to a dry-run, even though real dry-run capability simulates the full graph of execution, task by task. This feature instead computes the DAG of process dependencies and terminates the script bypassing the jobs submission, which remains unknown. 

This can be useful to render the pipeline DAG without running any task. It could be also used to evaluate the pipeline inputs and similar tasks. 

Currently, it's activated via the CLI option `-dry`. However, it would be preferable to use a different name for the reason explained above.  

Feedback is welcome.


pditommaso:
I was thinking the same origin, but potentially it could be useful for other purposes. 

@jorgeaguileraseqera is also suggesting `-preview` that sounds good.

pditommaso:
So, *preview* option was merged!

PR #2915, opened 2022-05-29
Fix external pod deletion for jobs

xhejtman:
There is possible race where pod is removed by control plane just after we get pod name, so we suppose there is a pod to get status from, but the pod is just gone. We need to fallback to job state in this case.


modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
```diff
@@ -371,7 +371,15 @@ class K8sClient {
...
         if( podName ) {
-            return podState(podName)
+            try {
+                return podState(podName)
+            } 
+            /* pod might be deleted by control plane just after findPodNameForJob() call
+             * so try fallback to jobState
+             */   
+            catch (NodeTerminationException err) {
+                return jobStateFallback0(jobName)
```
pditommaso, line 381: It may be useful to add a debug line to track this to simplify troubleshooting

xhejtman: Fixed

xhejtman:
I added the same `try` wrapper around podDelete as it can happen here as well:
```
nextflow.k8s.client.K8sResponseException: Request DELETE /api/v1/namespaces/acgt-ns/pods/nf-1b51504b8b211dc9c89ce4f94ad090d6-v522w returned an error code=404

  {
      "kind": "Status",
      "apiVersion": "v1",
      "metadata": {

      },
      "status": "Failure",
      "message": "pods \"nf-1b51504b8b211dc9c89ce4f94ad090d6-v522w\" not found",
      "reason": "NotFound",
      "details": {
          "name": "nf-1b51504b8b211dc9c89ce4f94ad090d6-v522w",
          "kind": "pods"
      },
      "code": 404
  }

        at nextflow.k8s.client.K8sClient.makeRequestCall(K8sClient.groovy:654)
        at nextflow.k8s.client.K8sClient.makeRequest(K8sClient.groovy:610)
        at nextflow.k8s.client.K8sClient.delete(K8sClient.groovy:573)
        at nextflow.k8s.client.K8sClient.delete(K8sClient.groovy)
        at nextflow.k8s.client.K8sClient.podDelete(K8sClient.groovy:176)
        at nextflow.k8s.client.K8sClient.jobDelete(K8sClient.groovy:233)
```

PR #2928, opened 2022-06-04
tag batch jobs with the run name

jchorl:
Similar to the k8s handler, this tags batch jobs with the parent run name. This should make it easier to chase down failures when running tons of concurrent pipelines with tons of concurrent jobs.


jorgeaguileraseqera:
Related https://github.com/nextflow-io/nextflow/pull/2853

pditommaso:
Thanks for submitting this. However support for tags have been provided via `resourceLabels` directive 5d0b7c35608a9c334e548625e91a87a6e982e6d9

PR #2934, opened 2022-06-07
bugfix: fix documentation in channel example [ci fast]

jorgeaguileraseqera:

### DIFF
--- a/docs/basic.rst
+++ b/docs/basic.rst
@@ -66,8 +66,8 @@ A Nextflow script looks like this::
 
 The above example defines two processes. Their execution order is not determined by the fact that the ``blastSearch``
 process comes before ``extractTopHits`` in the script (it could also be written the other way around). Instead, the
-pipe character (``|``) in the workflow between ``blastSearch`` and ``extractTopHits`` forwards the outputs from one
-process to the inputs of the next, automatically.
+pipe operator (``|``) in the workflow between ``blastSearch`` and ``extractTopHits`` forwards the outputs from one
+process to the inputs of the following one.
 
 When the workflow is started, it will create two processes and one channel (``query_ch``)
 and it will link all of them. Both processes will be started at the same time and they will listen to their
--- a/docs/basic.rst
+++ b/docs/basic.rst
@@ -65,18 +65,14 @@ A Nextflow script looks like this::
     }
 
 The above example defines two processes. Their execution order is not determined by the fact that the ``blastSearch``
-process comes before ``extractTopHits`` in the script (it could also be written the other way around).
+process comes before ``extractTopHits`` in the script (it could also be written the other way around). Instead, the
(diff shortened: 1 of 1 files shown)


docs/basic.rst
```diff
@@ -67,13 +67,16 @@ A Nextflow script looks like this::
 The above example defines two processes. Their execution order is not determined by the fact that the ``blastSearch``
 process comes before ``extractTopHits`` in the script (it could also be written the other way around).
 
-Instead, because the first process defines the channel ``top_hits_ch`` in its output declarations, and the
-process ``extractTopHits`` defines the channel in its input declaration, a communication link is established.
+Instead, the pipeline (``|``) after``blastSearch`` and ``extractTopHits`` creates an anonymous channel how will serve
+as output for the fist and input for the second and a communication link is established automatically.
```
ewels, line 71:
```suggestion
Instead, the pipe character (``|``) in the workflow between``blastSearch`` and ``extractTopHits`` creates an anonymous channel which
passes the outputs from one process to the inputs of the next, automatically.
```

docs/basic.rst
```diff
@@ -67,13 +67,16 @@ A Nextflow script looks like this::
...
 
 This linking via the channels means that ``extractTopHits`` is waiting for the output of `blastSearch`, and then
 runs `reactively` when the channel has contents.
 
-.. TODO describe that both processes are launched at the same time
+In this example, when the workflow is started it will create two process and tow channels (``query_ch`` plus the anonymous)
+and it will link all of them. Both process will be started at the same time and they will be listening in their
+respective input channels. Once ``blastSearch`` start to emit values into the anonymous channel ``extractTopHits``
+will start to receiving them (i.e. ``extractTopHits`` is consuming the channel in a `reactive` way)
```
ewels, line 79:
```suggestion
In this example, when the workflow is started it will create two process and two channels (``query_ch`` plus the anonymous pipe ``|``)
and it will link all of them. Both processes will be started at the same time and they will listen to their
respective input channels. Once ``blastSearch`` starts to emit values, the anonymous channel for ``extractTopHits``
will start to receiving them (i.e. ``extractTopHits`` consumes the channel in a `reactive` way).
```

ewels (approved):
Nice! Some suggested tweaks to the wording but nothing major.. 😃

docs/basic.rst
pditommaso: Not sure about the *creates an anonymous channel*. Not even sure that's technically is what it's happening. 

I would say: 

> Instead, the ``|`` (pipe) operator forwards the output of the ``blastSearch`` process to the ``extractTopHits`` processes.

jorgeaguileraseqera: Thanks Paolo, changing it

bentsherman (approved):

pditommaso (approved):

PR #2935, opened 2022-06-07
Remove undefined variable in Docs / Get started code example

samuell:
Running the modified example currently outputs an error:
```
No such variable: letters
```
By removing ` from letters`, the error disappears. As the first code example doesn't include it either, it seems it might be a remnant of some earlier code version(?)

### DIFF
--- a/docs/getstarted.rst
+++ b/docs/getstarted.rst
@@ -170,7 +170,7 @@ process script with the string ``rev $x``, so that the process looks like this::
     process convertToUpper {
 
         input:
-        file x from letters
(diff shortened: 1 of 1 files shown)


bentsherman:
I think this fix will be handled by #2816

bentsherman:
Closing in favor of #2816

samuell:
Sorry for the delay. Just note that this is for the "Get Started" doc (also one of the first things a user sees), so not part of the "Process" section which #2816 covers.

PR #2948, opened 2022-06-09
added activeDeadlineSeconds to specification

romanaduraciova:
Signed-off-by: romanaduraciova <duraciovar@gmail.com>

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy
@@ -216,9 +216,16 @@ class K8sTaskHandler extends TaskHandler {
             builder.withHostMount(mount,mount)
         }
 
-        return useJobResource()
-                ? builder.buildAsJob()
-                : builder.build()
+        if ( useJobResource() ) {
+            if ( task.config.time ) {
+                final duration = task.config.getTime()
+                builder.withActiveDeadline(duration.toMillis() / 1000 as int)
+            }
+            return builder.buildAsJob()
+        }
+        else {
+            return builder.build()
+        }
     }
 
     protected PodOptions getPodOptions() {
--- a/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
@@ -93,7 +93,9 @@ class PodSpecBuilder {
     List<Map> tolerations = []
 
     boolean privileged
-    
+
+    int activeDeadlineSeconds
+
     /**
      * @return A sequential volume unique identifier
      */
@@ -243,6 +245,11 @@ class PodSpecBuilder {
 
     PodSpecBuilder withPrivileged(boolean value) {
         this.privileged = value
+        return this
+    }
+
+    PodSpecBuilder withActiveDeadline(int seconds) {
+        this.activeDeadlineSeconds = seconds
         return this
     }
 
@@ -469,7 +476,9 @@ class PodSpecBuilder {
         spec.backoffLimit = 0
         spec.template = [spec: pod.spec]
 
-        // job spec
+        if ( activeDeadlineSeconds > 0 )
+            spec.activeDeadlineSeconds = activeDeadlineSeconds
+
         final result = [
                 apiVersion: 'batch/v1',
                 kind: 'Job',
--- a/modules/nextflow/src/test/groovy/nextflow/k8s/K8sTaskHandlerTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/k8s/K8sTaskHandlerTest.groovy
@@ -530,7 +530,7 @@ class K8sTaskHandlerTest extends Specification {
         1 * handler.getPodOptions() >> podOptions
         1 * task.getContainer() >> 'debian:latest'
         1 * task.getWorkDir() >> WORK_DIR
-        1 * task.getConfig() >> config
+        2 * task.getConfig() >> config
 
         result == [
             apiVersion: 'batch/v1', 


bentsherman:
I've wanted to see this implemented for a while. Happy to see you're using the `time` directive. Two things:
1. `activeDeadlineSeconds` is supported both for job spec and pod spec. Can you implement it such that it will work regardless of whether the user is using jobs or pods?
2. Please update the k8s docs [here](https://www.nextflow.io/docs/latest/executor.html#kubernetes) by adding `time` to the list of supported directives

romanaduraciova:
Thanks for your comment but if we add activeDeadlineSeconds to pod spec it will always apply time limit (not a case now) and this can change calculation behaviour. Do you really want it this way?

bentsherman:
I assume you mean if someone uses the `time` directive for other executors and assumes it will be ignored by k8s. That is a good point. In that case, implementing it for jobs only is a good way to allow users to opt in to the feature for now.

bentsherman:
It is here: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#podspec-v1-core

pditommaso:
Ah nice, then it's fair to add also for Pod provided it's not been included in a recent K8s API version. Any clue about that?

bentsherman:
Yes but to @romanaduraciova 's point, we might surprise users by doing that. Right now, users may have defined `time` limits for other executors, knowing that k8s will ignore it.

Having reflected on it, I don't think it will be a big deal. It is unlikely that time limits used for other executors will be unreasonable for k8s as well. So I would be fine with implementing it in pod spec. Perhaps you can be the deciding vote Paolo.

pditommaso:
Think if `time` was defined there could be such need or expectation. I think it makes sense to add it also for Pod

romanaduraciova:
I added time limits for pods and also updated k8s docs as you wanted. Feel free to comment :).

bentsherman:
Looks good. However, you accidentally pushed changes to `plugins-info.txt` and `Const.groovy`. Annoyingly Nextflow changes these files whenever it is built. Please revert them and then we will merge.

romanaduraciova:
Fixed :)

modules/nextflow/src/main/resources/META-INF/plugins-info.txt
```diff
@@ -1,7 +1,7 @@
...
+nf-ga4gh@1.0.3
+nf-azure@0.13.2
 nf-console@1.0.3
+nf-amazon@1.8.0
 nf-google@1.2.1
-nf-azure@0.13.2
-nf-tower@1.4.0
-nf-ga4gh@1.0.3
\ No newline at end of file
+nf-sqldb@0.4.1
```
pditommaso: This should not be changed

modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy
```diff
@@ -216,9 +216,16 @@ class K8sTaskHandler extends TaskHandler {
...
+        if ( useJobResource() ) {
+            if ( task.config.time ) {
+                final duration = task.config.getTime()
+                builder.withActiveDeadline(duration.toMillis() / 1000 as int)
+            }
+            return builder.buildAsJob()
+        }
+        else {
+            return builder.build()
+        }
```
pditommaso, line 228: I think we should set irrespective if it's a pod or a job

romanaduraciova:
My apologies. I have already done those things but the commits are wrong. Will fix that with the next push.

romanaduraciova:
Should be all done now.

modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy
```diff
@@ -216,9 +216,15 @@ class K8sTaskHandler extends TaskHandler {
             builder.withHostMount(mount,mount)
         }
 
-        return useJobResource()
-                ? builder.buildAsJob()
-                : builder.build()
+        if ( task.config.time ) {
+            final duration = task.config.getTime()
+            builder.withActiveDeadline(duration.toMillis() / 1000 as int)
```
pditommaso, line 221:
```suggestion
            builder.withActiveDeadline(duration.toSeconds())
```

romanaduraciova:
Based on your comments I made required corrections and fixed problem with DCO.

modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy
```diff
@@ -216,9 +216,15 @@ class K8sTaskHandler extends TaskHandler {
...
 
-        return useJobResource()
-                ? builder.buildAsJob()
-                : builder.build()
+        if ( taskCfg.time ) {
+            final duration = taskCfg.getTime()
+            builder.withActiveDeadline(duration.toSeconds() as int)
+        }
+
+        if ( useJobResource() )
```
pditommaso, line 224:
```suggestion
        return useJobResource()
                ? builder.buildAsJob()
                : builder.build()
```

pditommaso (changes requested):
to proceed it's needed to rollback the changes in the Consts and plugins-info files. Thanks!

romanaduraciova:
I changed the condition as you proposed and reverted the changes in the Consts and plugins-info files. Should be ready to merge :).

xhejtman:
> to proceed it's needed to rollback the changes in the Consts and plugins-info files. Thanks!

should those files go to .gitignore then? It is a bit annoying to always take care not to push them and they get updated in each build.

pditommaso:
> I changed the condition as you proposed and reverted the changes

Thanks appreciated! 

> should those files go to .gitignore then? It is a bit annoying to always take care not to push them and they get updated in each build.

I know it's annoying, but when doing the final release they should be committed. I don't know a nicer way to handle this

pditommaso (commented):
Just realised there's no unit tests for the new feature. Think it should be added both for Pod and Job

romanaduraciova:
Unit tests for the new feature added.

pditommaso (approved):

pditommaso:
Nice, thanks!

PR #2949, opened 2022-06-09
Remove sections from operator docs

bentsherman:
This PR removes the sections from the operator docs. I have found these different categories to hinder as much as they help. Some operators don't clearly fit into one category or another, and you can only have one section expanded at a time, so it's hard to browse all of the operators at once.

With this PR, the operators are simply listed in alphabetical order. I tried to preserve some of the benefit of the sections by adding a blurb at the top, which recommends some commonly used operators for each category.

I did not change the content of any individual operators, I only reordered them.

### DIFF
--- a/docs/operator.rst
+++ b/docs/operator.rst
@@ -4,38 +4,674 @@
 Operators
 *********
 
-Nextflow `operators` are methods that allow you to connect channels to each other or to transform values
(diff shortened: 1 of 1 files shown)


pditommaso:
Nice job!

PR #2950, opened 2022-06-10
feature: improve command line params parse

jorgeaguileraseqera:
allow more complex params from command line as List or Maps parsing as JSON


pditommaso:
Umm, we are already doing too much magic with params. Above all the main concern is backward compatibility. What if somebody is using a parameter to pass a json string ?

jorgeaguileraseqera:
ummmm yes, you're right

PR #2957, opened 2022-06-14
Add MaxErrorRetry to k8s-config (replace hardcoded 5 maxRetry)

naineshmp:


bentsherman:
Thanks for your contribution. I think the option should be called something like `maxRequestRetries` so that it is clearly distinguished from `maxRetries` directive for retrying tasks. This option is about K8s API calls. Also, if you would please document this option in `config.rst`, under the `k8s` config scope.

modules/nextflow/src/main/groovy/nextflow/k8s/client/ClientConfig.groovy
```diff
@@ -37,6 +37,8 @@ class ClientConfig {
 
     String namespace
 
+    Integer maxErrorRetry
```
pditommaso, line 40:
```suggestion
    Integer maxErrorRetry = 5
```

pditommaso (changes requested):
Ok nice. A test should be added to verify that setting is read from the config, otherwise it should default to 5

naineshmp:
> Ok nice. A test should be added to verify that setting is read from the config, otherwise it should default to 5

Thanks for your review. Let me make the change.

naineshmp:
> Ok nice. A test should be added to verify that setting is read from the config, otherwise it should default to 5

I have completed the change, could you please review? Thanks.

modules/nextflow/src/main/groovy/nextflow/k8s/client/ClientConfig.groovy
```diff
@@ -37,6 +37,8 @@ class ClientConfig {
 
     String namespace
 
+    Integer maxErrorRetry = 5
```
pditommaso, line 40: Checking better think this is useless. The client config is used to discover the K8s config in the HOME/.kube/config path. 

Instead this should go in the [K8sConfig](https://github.com/nextflow-io/nextflow/blob/979e39fd0a5e87b1d57918a424382e35d2af1951/modules/nextflow/src/main/groovy/nextflow/k8s/K8sConfig.groovy#L75-L75), tho is not easy then to access it from the client

docs/config.rst
```diff
@@ -423,6 +423,7 @@ storageSubPath      The path in the persistent volume to be mounted (default: ro
 computeResourceType Define whether use Kubernetes ``Pod`` or ``Job`` resource type to carry out Nextflow tasks (default: ``Pod``).
 fetchNodeName       If you trace the hostname, activate this option (default: ``false``, requires version ``22.05.0-edge`` or later).
 volumeClaims        (deprecated)
+maxRequestRetries   Defines the Kubernetes API max request retries (default is set to 5)
```
bentsherman, line 426: Can we name this `maxErrorRetry` as well so that it's consistent with the analogous options for HPC and Azure?

pditommaso:
Merge manually 👉  a6155943. @bentsherman I've accommodated your request adjusting the check considering the different semantics of attempt vs retry

PR #2960, opened 2022-06-15
Support for EC-encrypted client keys (k8s)

pawelpiatkowski1125:
Dear Nextflow maintainers,

at MNM Diagnostics, your tool is widely used for genomic analyses. Quite recently, we found out that our Rancher configuration is incompatible with Nextflow; namely, the default client key encryption (EC) used by Rancher for Kubeconfig files is not supported by Nextflow, which assumes the key to be RSA-encrypted.

This is a small patch that aims to address this limitation by providing a simple fallback mechanism. Don't hesitate to reach out if clarification is needed.

Kind regards,
Paweł


pawelpiatkowski1125:
Please note that DCO reports mismatches in two of my commits – this is due to incorrect encoding of diacritics in my name.

pawelpiatkowski1125:
Thank you, @bentsherman. `k8s.clientKeyStoreEncryption` makes perfect sense! In fact, I started from doing it that way, and only changed approach after I realized (incorrectly, I assume) that `ConfigDiscovery` was an upstream module and didn't get the config info :smile:

pditommaso:
Thanks for this PR, i've merged it manually fd759d09

Issue #2264, opened 2021-08-13
Reconcile cpus/memory with clusterOptions for pbspro executor


bentsherman:
I have a feature request and some discussion around the pbspro executor and how resource settings are determined from `cpus`, `memory`, and `clusterOptions`.

So we have a pipeline with the standard sort of resource settings:
```
# base.config
process {
  cpus = 1
  memory = 6.GB
  time = 24.h
}
```

And in the spirit of nf-core we are developing an institutional config file with settings for our PBS Pro scheduler. Now a wonky thing about our scheduler is that we _must_ specify the interconnect, even if we don't care and just say "any". Our admins have been pretty adamant about this rule. So we put that here:
```
# palmetto.config
process {
    executor = "pbspro"
    clusterOptions = "-l select=1:interconnect=any"
}
```

These settings in combination produce the following kind of pbs headers:
```
#PBS -N nf-kallisto_DRX
#PBS -o [...]
#PBS -j oe
#PBS -l select=1:ncpus=4:mem=6144mb
#PBS -l walltime=24:00:00
#PBS -l select=1:interconnect=any
```

In this situation, PBS seems to ignore the second select line. It gets the right cpus, memory, and walltime, but it never includes the interconnect setting and so the job is rejected. (Actually the job is still accepted if ncpus=1, but I think that is also an artifact of our particular cluster.)

So now I have to null the cpus and memory directives and instead specify them in clusterOptions:
```
process {
    executor = "pbspro"
    cpus = null
    memory = null
    clusterOptions = "-l select=1:ncpus=1:mem=6gb:interconnect=any"
}
```

But at this point things get really hairy because I end up having to do the same kind of thing for any withLabel or withName rules in the pipeline config. The pipeline in question is working towards nf-core compatibility so they aren't going to maintain platform-specific profiles, and I don't think the sys admins are going to budge on this weird interconnect rule.

So I'm wondering if we can solve my problem in Nextflow? It seems to me that if `cpus` and/or `memory` are defined in addition to `clusterOptions`, then the resulting select lines should be merged like so:
```
#PBS -N nf-kallisto_DRX
#PBS -o [...]
#PBS -j oe
#PBS -l select=1:ncpus=4:mem=6144mb:interconnect=any
#PBS -l walltime=24:00:00
```

[...]

PR #2971, opened 2022-06-17
selectOptions directive

romanaduraciova:


pditommaso:
Not sure to understand the advantage of `clusterOptions`

xhejtman:
`clusterOptions` is *replacement* for everything including resource specification. For PBS(Pro), you select resources using: `-l select=1:ncpus=X:memory=Y`. You can use `clusterOptions` to override this, but using `clusterOptions` you just cannot add only. With`clusterOptions` you need to specify whole `select=1` command, so you cannot use single `clusterOptions` for whole pipeline, but you need to specify it for every process of the pipeline. 

with `selectOptions`, we could add options, e.g. ngpus=1 or scratch=10G, this can be global for whole pipeline so it is easier than exact specificatoin of `clusterOptions` per process.

bentsherman:
Hi @xhejtman @romanaduraciova , As a former PBS user myself, I have had such problems with `clusterOptions` before, and I have often wanted to see it's behavior improved. I think we should see if we can improve `clusterOptions` to address your use case before we implement a new but slightly different directive. Let us try to understand what you are trying to do and see if we can make it work with `clusterOptions`.

bentsherman:
Just saw your comment. My understanding is, you can specify multiple lines with `-l` and apply multiple layers of settings. For example:
```
-l select=1:ncpus=X:memory=Y
-l ngpus=1
```

I believe this syntax would apply the `ngpus` setting correctly, although I am rusty in this area.

xhejtman:
> Just saw your comment. My understanding is, you can specify multiple lines with `-l` and apply multiple layers of settings. For example:
> 
> ```
> -l select=1:ncpus=X:memory=Y
> -l ngpus=1
> ```
> 
> I believe this syntax would apply the `ngpus` setting correctly, although I am rusty in this area.

this is not possible. You cannot mix 'old' style `-l ngpus=1` with 'new' tyle `-l select=1`. And you cannot use `-l select` multiple times.

xhejtman:
> This reminds me of an old issue #2264 that I never pursued. I think we might be able to apply my solution from that issue to here.
> 
> Here is my vision. You should be able to specify your base `clusterOptions` at the top-level. Then apply more specific resource settings. Like so:
> 
> ```
> process {
>   clusterOptions = "-l select=1:scratch_local=10GB:cluster=^samson"
> 
>   withName: 'foo' {
>     cpus = 2
>     memory = 6.GB
>   }
> }
> ```
> 
> Then, Nextflow should be able to merge these settings into a single resource list, like so:
> 
> ```
> #PBS -l select=1:ncpus=2:memory=6gb:scratch_local=10GB:cluster=^samson
> ```

this is possible, if interpretation of clusterOptions can be changed. I thought you prefer to keep existing behavior and do not incompatible changes, that's why we implemented it as another option.

bentsherman:
I think we can extend `clusterOptions` in a way that is backwards compatible, practically speaking. We need to update the pbspro executor to always generate only one line of resource settings, and allow the `clusterOptions` setting to be "updated" with values from `cpus` and `memory` if they are provided. It is extremely unlikely that anyone is currently combining these directives in a way that would be affected by these changes, so I think we are safe.

Do you think you can implement these changes or would you like me to try? All the changes should be contained in `PbsProExecutor::getDirectives()`.

xhejtman:
We can give it a try. There are additional things we want to fix:
1. queueStatusCommand() is broken if `queue` is not set, because `server` is abbreviated if it is more than few chars long:
```
qstat -B
Server             Max   Tot   Que   Run   Hld   Wat   Trn   Ext Status
---------------- ----- ----- ----- ----- ----- ----- ----- ----- -----------
cerit-pbs.cerit*     0 37934   718  1823  1622     0     0     3 Scheduling
```
and it does not work that way.
2. the same function, if the `queue` is used and the task is submitted into routing queue (not execution queue), `qstat` does not find the job in that queue and pipeline fails, solution is to do qstat without queue specification
3. `qstat`/`qsub` can fail if e.g., pbs server is just restarting and pipeline fails, we should retry here.

Not related to this, we want to implement support for workdir on S3 for kuberun and at least pbspro executors.

What do you think about all above? Perhaps to discuss it first in some issues?

bentsherman:
Okay. I recommend you use a map to build the cluster options, that way you can override options more easily and construct the PBS command line options at the end.

For everything else you mentioned, let's handle them as two separate issues. One for the queue/qstat issues, one for S3 workdir.

bentsherman:
Yes. When you are loading clusterOptions, cpus, memory, time, etc, load them into a map first and then convert to PBS syntax.

romanaduraciova:
Hello. So I implemented addition of clusterOptions into map. Feel free to comment.

modules/nextflow/src/main/groovy/nextflow/executor/PbsProExecutor.groovy
```diff
@@ -44,12 +46,31 @@ class PbsProExecutor extends PbsExecutor {
     @Override
     protected List<String> getDirectives(TaskRun task, List<String> result ) {
```
bentsherman, line 47: This ended up being more complicated than I expected. I tried to write my own version of `getDirectives` based on your changes:
```groovy
    protected List<String> getDirectives(TaskRun task, List<String> result ) {
        assert result != null

        // extract resource string from cluster options
        def clusterOptionsList = task.config.getClusterOptionsAsList()
        def resIndex = clusterOptionsList.indexOf('-l')
        def res = ''

        if( resIndex != -1 ) {
            res = clusterOptionsList.removeAt(resIndex + 1)
            clusterOptionsList.removeAt(resIndex)
        }

        // parse resource string into a map
        Map<String, String> resMap = [:]
        for ( String entry : res.split(':') ) {
            String[] tokens = entry.split('=')
            resMap.put(tokens[0], tokens[1])
        }

        // append cpus and memory to resource map
        if( task.config.hasCpus() || task.config.memory ) {
            resMap['ncpus'] = task.config.getCpus().toString()
        }
        if( task.config.memory ) {
            // https://www.osc.edu/documentation/knowledge_base/out_of_memory_oom_or_excessive_memory_usage
            resMap['mem'] = "${task.config.getMemory().getMega()}mb".toString()
        }

        // append cluster options
        for( entry : clusterOptionsList ) {
            result << entry
        }

        // append standard options
        result << '-N' << getJobNameFor(task)
        result << '-o' << quote(task.workDir.resolve(TaskRun.CMD_LOG))
        result << '-j' << 'oe'

        // append queue name
        if( task.config.queue ) {
            result << '-q' << task.config.queue.toString()
        }

        // append final resource list
        final resSelect = resMap['select'] ?: 1
        final resList = resMap
            .findAll(key, value -> key != 'select')
            .collect(key, value -> "${key}=${value}")

        if( !resList.isEmpty() ) {
            result << '-l' << "select=${resSelect}:${resList.join(':')}".toString()
        }

        // max task duration
        if( task.config.time ) {
            final duration = task.config.getTime()
            result << "-l" << "walltime=${duration.format('HH:mm:ss')}".toString()
        }

        return result
    }
```

romanaduraciova:
I agree that this is really complex and hard to implement in a way that would be suitable for all users. That's why we suggested selectOptions directive in the first place. What do you think about it now?

bentsherman:
Two other complicating issues are that (1) `clusterOptions` may contain many command line options, not just `-l`, and (2) some resource options can be applied to all chunks by separating with a comma.

Here is the most complex example I can think of:
```
--opt1 arg1 --opt2 arg2 -l "select=4:ncpus=2:mem=8gb+4:ncpus=8:mem=2gb,interconnect=hdr"
```

I think you would have a hard time trying to apply `selectOptions` with this. Unless your idea is to ignore `clusterOptions` in this case and append the select options to the resource list that Nextflow builds from the `cpus` and `memory` directives.

Here is an alternative that should work with Nextflow now. Define your global select options as a param and use them in `clusterOptions` for each process:
```nextflow
params.selectOptions = ':scratch_local=10GB:cluster=^samson'

process proc1 {
  clusterOptions = "-l select=1:ncpus=2:mem=8gb${params.selectOptions}"
}

process proc2 {
  clusterOptions = "-l select=1:ncpus=4:mem=16gb${params.selectOptions}"
}

// ...
```

You still have to define `clusterOptions` for each process instead of using `cpus` and `memory`, but you only have to define your extra select options once.

xhejtman:
I think this is becomming theoretic with such complex requests mainly for mostly simple tasks originating from nextflow. Do you have any example of such complex requests and pipeline? I believe that accepting that `selectOptions` as initially provided 1) should be suitable for most cases, 2) have pretty simple implementation.

I agree that your proposed solution works, but you need to rewrite all processes definition for all used pipelines.. and well, this is not very user friendly for non-IT users.

bentsherman:
I agree that both of our solutions so far would work most of the time, but we have to consider all of these possibilities when we design features like this one. Or we have to add a caveat to the docs that "this feature only works for certain simple use cases X, Y, and Z ..." which will be confusing for users.

On the other hand, I don't think the params workaround would be very burdensome. Nextflow users on PBS are usually already familiar with PBS syntax. As for refactoring each process, the best practice is to create a config profile (e.g. `cerit-sc`) with the `clusterOptions` defined for each process (unlike my example). That way you can edit them all in one place.

There is one more idea that I have for improving `clusterOptions` to meet your needs. 

```nextflow
// nextflow.config
process {
  clusterOptions = { "-l select=1:ncpus=${task.cpus}:mem=${task.memory.toMega()}mb${params.selectOptions}" }
}

// main.nf
process proc1 {
  cpus = 2
  memory = 8.GB
}
```

The nice thing here is you define `clusterOptions` once and you don't have to refactor the processes. The problem is that if you do this right now, Nextflow produces two `-l select=...` lines which is invalid. But I can easily modify Nextflow to not produce the second line if `clusterOptions` contains `-l`. That is much easier than trying to parse and reconstruct the cluster options.

bentsherman:
I tested my latest idea and it worked. Let me know if you like that idea and I will draft a new PR for it.

bentsherman:
Closing in favor of #3015. Please comment there if you have any feedback for the proposed changes.

Issue #1089, opened 2019-03-28
nextflow doesn't recognise k8s current context


olatarkowska:
## Bug report 

Kubernetes nicely manage multiple cluster configs using environment variable `KUBECONFIG`. Context can be switched using `kubectl use-context`, for more details see https://kubernetes.io/docs/reference/kubectl/cheatsheet/#kubectl-context-and-configuration

Nextflow assumes there is one path to the config exported to the KUBECONFIG
```
Mar-28 10:06:10.311 [main] DEBUG nextflow.k8s.client.ConfigDiscovery - K8s config file does not exist: /home/ola/.kube/production-new.config:/home/ola/.kube/config.local
```

### Expected behavior and actual behavior

use currect context `kubectl config current-context	`

### Steps to reproduce the problem

export multiple configs

### Environment 

* Nextflow version: nextflow version 19.01.0.5050
* Java version: JDK 1.8
* Operating system: osx

cc'd: @wikiselev

pditommaso:
It may be related to #1049. Please try version `19.03.0-edge`.

pditommaso:
Sorry I hadn't realised properly the issue you are rasing. Basically, the `KUBECONFIG` can list more than one path separating them with a `:` character that needs to be merged according to the logic described [here](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#merging-kubeconfig-files).

pditommaso:
The merging logic could be tricky to implement. I'm tempted to use the `kubectl config view --raw` to  merge the yaml files and parse the resulting output.

PR #2972, opened 2022-06-17
Support for multiple paths in $KUBECONFIG (#1089)

pawelpiatkowski1125:
Another Kubernetes-related patch. This one aims to address a known bug (#1089).

The code below scans all config files found in $KUBECONFIG, and selects the file that matches current Kubernetes context.
If `context` is not specified, `kubectl` is queried to determine the context.


bentsherman:
My only concern is that it depends on `kubectl` being installed. For `kuberun` it is fine, but running nextflow directly in a k8s cluster with the `k8s` executor does not require `kubectl` to be installed, because nextflow uses the K8s API for everything. However I'm not sure if this method is called in that case. Can you confirm that everything still works if you run nextflow directly with the `k8s` executor?

bentsherman:
I think the unit test failures can be safely ignored because they all pertain to Paolo's CodeCommit feature.

pditommaso:
Just solved that on master, but the use of `kubectl` is a blocker

bentsherman:
There are other places in the code that use `kubectl`, but I think they are all in the context of `kuberun` which makes sense. If Nextflow is running in a pod with `k8s` executor, then it shouldn't need to load the kubeconfig because it's already in the k8s cluster. Also it wouldn't have a kubeconfig in that case anyway. So wouldn't this code only ever be called by `kuberun`?

pawelpiatkowski1125:
@bentsherman @pditommaso Thank you for taking a look into my code. A word of explanation: in fact, `kubectl` is _only_ run if there are multiple paths in $KUBECONFIG _and_ the context is not explicitly specified.

Please compare this to the current situation when multiple paths in $KUBECONFIG would simply crash Nextflow, regardless of whether the context was specified or implied. So basically, this patch depends on `kubectl`, but only when Nextflow would crash anyway.

pditommaso:
My understanding is that `kubectl` is only to determine the current context. Could not be inferred from the `current-context` element in the Kube `config` file?

pawelpiatkowski1125:
That's right, `kubectl` is only to determine the current context – when it's not specified explicitly (by using `k8s.context`) _and_ there's ambiguity to it (more than one kubeconfig file in $KUBECONFIG).

That's why the config files can't be used for that purpose: each has a `current-context` entry, and it's impossible to tell which one is used right now without querying `kubectl`.

Also, regarding your concern: I might be wrong, but isn't `$KUBECONFIG` intrinsically linked to `kubectl`?

pditommaso:
> Also, regarding your concern: I might be wrong, but isn't $KUBECONFIG intrinsically linked to kubectl?

Not necessarily, nextflow uses to find the config file and parse it. 

> That's why the config files can't be used for that purpose: each has a current-context entry, and it's impossible to tell which one is used right now without querying kubectl.

Don't think `kubectl` implements a magic rule. IMO it should be figured out what's governing the finding of the current context in this scenario and provide a native implementation.

pawelpiatkowski1125:
> Not necessarily, nextflow uses to find the config file and parse it.

Ah, I haven't thought about it, sorry.

> Don't think `kubectl` implements a magic rule. IMO it should be figured out what's governing the finding of the current context in this scenario and provide a native implementation.

There's a [Java API](https://javadoc.io/static/io.kubernetes/client-java/12.0.0/io/kubernetes/client/util/KubeConfig.html), but I'm not sure if it doesn't use `kubectl` under the hood...

pditommaso:
Nextflow implements its own K8s client, but the [kube java client](https://github.com/kubernetes-client/java) could be a good start to reverse engineer it and figure out how it works.

pawelpiatkowski1125:
Sadly, the client that you linked to only supports one cluster per client:
https://github.com/kubernetes-client/java/issues/1972#issuecomment-965633246

So does the Java API, as far as I can see.

pditommaso:
Stalled

PR #2975, opened 2022-06-20
feature: clone git repos as bare to allow concurrent pipeline

jorgeaguileraseqera:
This PR tries to clone git repositories using subdirectories for different branches

The idea is to clone the repo in a top directory as bare git and instead to checkout the branch over it, create a copy of it into a subdirectory and do the check out here


jorgeaguileraseqera:
I'll close this PR due I think I'm taking a wrong approach to solving it

PR #2987, opened 2022-06-24
Rename kuberun -pod-image to -head-image

bentsherman:
This PR renames the `-pod-image` option in `kuberun` to be consistent with `-head-cpus`, `-head-memory`, etc


pditommaso:
Nice

PR #2988, opened 2022-06-24
Add CSI ephemeral volume

### LINKED ISSUE #2985: Comprehensive support for K8s ephemeral volumes
Nextflow currently uses PersistentVolumeClaim to use any kind of persistent volume in a generic way, which is good.

However, there are also several types of [ephemeral volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#types-of-ephemeral-volumes):
- emptyDir
- configMap
- downwardAPI
- secret
- CSI ephemeral volume

Of these, Nextflow currently supports configMaps and secrets.
- emptyDir allows you to allocate local scratch storage in a k8s-friendly way, and there is an ancient PR (#1389) to support it in Nextflow
- downwardAPI allows you to pass node/pod information (e.g. hostname) to the pod via files, it is nice but can also be done via `env` / `fieldPath` which Nextflow supports
- CSI is basically a way to provision storage backed by third-party drivers. Some drivers can be persistent while others can be ephemeral. [Secrets Store](https://secrets-store-csi-driver.sigs.k8s.io/getting-started/usage.html) is a CSI ephemeral driver that allows you to inject secrets from something like Hashicorp Vault, Azure Keyvault, etc, which is more secure than using secrets.

These types of volumes don't fit in with the standard PVC model that Nextflow relies on, because you can't define them elsewhere and them reference them generically in the pod -- they're ephemeral! It only makes sense to define them in the pod itself. Therefore, Nextflow is already using the correct model with configMap and secret, and I think we should continue that pattern for the other ephemeral types. The most urgent one to add is `csi`, but I would support `emptyDir` as well. I will draft a PR for `csi` when I have some time.

bentsherman:
Closes #2985 

This PR adds support for CSI ephemeral volumes, following the same pattern used for configMaps and secrets.

The main use case is for [Secrets Store](https://secrets-store-csi-driver.sigs.k8s.io/getting-started/usage.html), which allows you to inject secrets from something like Hashicorp Vault, Azure Keyvault, etc, which is more secure than using regular secrets.


sfehrmann:
Any chance this could get considered sometime?

bentsherman:
Just waiting for @pditommaso to review and approve

modules/nextflow/src/main/groovy/nextflow/k8s/model/PodOptions.groovy
```diff
@@ -66,6 +68,7 @@ class PodOptions {
     PodOptions( List<Map> options=null ) {
         int size = options ? options.size() : 0
         envVars = new HashSet<>(size)
+        mountCsiEphemerals = new HashSet<>(size)
         mountSecrets = new HashSet<>(size)
         mountConfigMaps = new HashSet<>(size)
         mountClaims = new HashSet<>(size)
```
pditommaso, line 74: Setting the map size helps to allocate unneeded memory. Considering that this can be invoked tens of thousand of times is helps a bit to contain the garbage collection activity.

PR #2999, opened 2022-06-28
feature: enable running processes on different regions

jorgeaguileraseqera:
enable running processes on different regions when running on cloud executors
(only AWS by the moment)

The idea of this PR is to separate the preparation (current AwsBatchExecutor) from the Execution (AwsBatchRegionExecutor) creating a "merged" config from the session and the task

basically, AwsBatchExecutor creates an AwsBatchRegionExecutor who prepares an AwsClient with the merged config and runs a TaskHandler with it

p.d.
by the moment it's done only in AWS


pditommaso:
Not planned in the near term

PR #3003, opened 2022-06-29
accelerate s3 downloads

jorgeaguileraseqera:
Currently, the implementation tries to create a `newInputStream` when the source and the target are in different `FileSystem` and copy bytes from one into another

This implementation creates a new interface for file systems who can/want to perform a custom copy in this situation. For example, S3 FileSystem uses `TransferManager` to download from a bucket to a local Path


pditommaso:
Ok, merging this 👍

PR #3004, opened 2022-06-30
Deprecate DSL1 syntax for tap operator

### LINKED ISSUE #2981: Deprecate tap operator, extend set operator to return input channel
The [tap](https://www.nextflow.io/docs/latest/operator.html#tap) isn't marked as deprecated, but it depends on `Channel.create()` which is deprecated. I tried the top example with `Channel.empty()` instead but it didn't work, which makes sense because you shouldn't be able to just write to a channel like that.

So should we deprecate `tap` ? It seems like we could achieve the same functionality by changing `set` to return itself like `view`. Then you could use `set` in the middle of an operator chain to emit into a new channel while proceeding with other channel operators.

bentsherman:
Closes #2981 

I basically just updated the `set` operator to return the input channel directly, for each channel type it supports. I'm not sure if that's the correct way to do it because I'm new to GPars, but the SetOp unit tests are passing.


bentsherman:
@pditommaso This PR is old so I updated it. It can be merged whenever you want.

bentsherman:
You are right, I guess I never tested that in version. In that case we should remove the DSL1 example from the docs and promote the closure syntax. It is confusing for the main example to be something that doesn't work in DSL2. I will update the PR.

bentsherman:
Okay, now the PR only deprecates the tap operator syntax that doesn't work in DSL2. Documentation is clearer now.

@pditommaso ready for review 👍

pditommaso:
Nice

PR #3006, opened 2022-06-30
Override lsf.conf settings with nextflow config settings

### LINKED ISSUE #2862: LSF memory parameters in `nextflow.config` should take precedence over `lsf.conf`
I have encountered some issues with how Nextcloud sets memory limits/reserves memory for LSF jobs. On our cluster, both the memory limit (`-M ...`) and the memory reservation (`-R 'rusage[mem=...]'`) specify the memory per job, not per slot. Nextcloud provides parameters to deal with this, namely `executor.perJobMemLimit` and `executor.perTaskReserve` (see also issue #1071). These parameters configure whether the memory limit and memory reservation, respectively, are on a per job/host basis or on a per slot basis. The problem is that Nextcloud also extracts the parameters from the LSF configuration to be found in `$LSF_ENVDIR/lsf.conf` (see [these lines of code](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/executor/LsfExecutor.groovy#L282-L296)) and when these parameters are found in this file, the user-defined settings in `nextflow.config` are **ignored**. To be precise, the parameter `executor.perJobMemLimit` is ignored, and the parameter `executor.perTaskReserve` is never read from the `nextflow.config` - even if the parameter cannot be found in `lsf.conf`. In my opinion, the settings in `nextflow.config` should take precedence over more global (cluster-wide) settings. Users should always be able to overwrite globally defined parameters if they desire. The LSF configuration can be complex. And as a result of that, the way how Nextflow interprets LSF parameters is not necessarily how LSF truly behaves. Having the ability to overwrite Nextflow's auto-detect behavior enables users to overcome such discrepancies. What do you think?

On a related note: I think there is a bug on [this line](https://github.com/nextflow-io/nextflow/blob/c1e45aa98128e5c83968b2c129b85596b1b2eaec/modules/nextflow/src/main/groovy/nextflow/executor/LsfExecutor.groovy#L295). Shouldn't it say `$perTaskReserve` rather than `$perJobMemLimit`?

bentsherman:
Closes #2862 

Also adds `executor.perTaskReserve` config option to override LSF config file.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -344,6 +344,7 @@ exitReadTimeout       Determines how long the executor waits before to an error 
 killBatchSize         Determines the number of jobs that can be `killed` in a single command execution (default: ``100``).
 submitRateLimit       Determines the max rate of job submission per time unit, for example ``'10sec'`` eg. max 10 jobs per second or ``'50/2min'`` i.e. 50 job submissions every 2 minutes (default: `unlimited`).
 perJobMemLimit        Specifies Platform LSF *per-job* memory limit mode. See :ref:`lsf-executor`.
+perTaskReserve        Specifies Platform LSF *per-task* memory reserve mode. See :ref:`lsf-executor`.
 jobName               Determines the name of jobs submitted to the underlying cluster executor e.g. ``executor.jobName = { "$task.name - $task.hash" }`` Note: when using this option you need to make sure the resulting job name matches the validation constraints of the underlying batch scheduler.
 cpus                  The maximum number of CPUs made available by the underlying system (only used by the ``local`` executor).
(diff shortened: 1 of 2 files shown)


pditommaso (commented):
Well done

bentsherman:
Guess I should have marked this PR as draft. For some reason I only noticed the azure tests failing, which I assume are unrelated. I think the code changes are correct but the unit tests need to be updated. I will submit another PR shortly.

pditommaso:
yeah, weird tests didn't run either. I'll push the changes into a separate branch wait

pditommaso:
This https://github.com/nextflow-io/nextflow/pull/3009

Issue #2912, opened 2022-05-28
Bug with nxf_unstage() module only with storeDir directive in Nextflow v22.04.3 || Urgent 


luisovallex:
### Expected behavior and actual behavior

Expected behavior it's to store the files on S3 using storeDir directive, but the actual behavior it's the pipeline crashing because Nextflow cannot find the output files.

During the debugging process we tested the pipeline with the publishDir directive and everything looks to be fine, so we decide to go trough the work directory and check for the .command.run scripts to see the difference between the execution with publishDir and storeDir, we saw a couple of differences between the codes:

This is the .command.run file generated when running the pipeline with publishDir:
[publishdir.txt](https://github.com/nextflow-io/nextflow/files/8790385/publishdir.txt)

And this is the one generated when running the pipeline with storeDir:
[workdir.txt](https://github.com/nextflow-io/nextflow/files/8790386/workdir.txt)

There are many differences, but the one I don't get it's related with the nxf_s3_upload and nxf_s3_download, because storedir file has this functions and publishdir do not.

After checking and comparing these files we decide to test the storeDir .command.run file by going inside the container, edit the file and then run it manually and we come to the conclusion that the main issue it's related to this piece of code:

```
nxf_unstage() {
    true
    cp .command.out /mnt/efs/ngs_scratch/workdir/210812_NB501331_0892_AH57FVBGXK/updated-path4/14/96e35fb77d096c4ca2d3291e49b398/.command.out || true
    cp .command.err /mnt/efs/ngs_scratch/workdir/210812_NB501331_0892_AH57FVBGXK/updated-path4/14/96e35fb77d096c4ca2d3291e49b398/.command.err || true
    cp .command.trace /mnt/efs/ngs_scratch/workdir/210812_NB501331_0892_AH57FVBGXK/updated-path4/14/96e35fb77d096c4ca2d3291e49b398/.command.trace || true
    [[ ${nxf_main_ret:=0} != 0 ]] && return
    IFS=$'\n'
    for name in $(eval "ls -1d L1.Aligned.sortedByCoord.out.bam L1.Chimeric.out.sam L1.Chimeric.out.junction L1.Log.final.out" | sort | uniq); do
        nxf_s3_upload '$name' s3://k8s-temporal-efs/results/210812_NB501331_0892_AH57FVBGXK/A01-00159459-VL45582-A-d/align/L1 || true
    done
    unset IFS
}
```

[...]

Issue #2904, opened 2022-05-25
storeDir directive does not work with  AWS S3 path


syousry6:
Hello Everyone,

We are running nextflow pipelines on AWS EKS cluster, We use the storeDir directive together with an AWS S3 path to store the results for processes.
We are getting the following error everytime we run a process with storeDir as follows
```
Missing output file(s) `L1.Aligned.sortedByCoord.out.bam` expected by process `alignment`
```

- Here is the error in .nextflow.log
```
Process `alignment (A09-00159428-B-RNA-013:L1)` is unable to find [S3Path]: `/k8s-temporal-efs/results/210811_NB501495_0936_AH555LBGXK/A09-00159428-B-RNA-013/align/L1/L1.Aligned.sortedByCoord.out.bam` (pattern: `L1.Aligned.sortedByCoord.out.bam`)
```

- Here is the process configuration in [modular.nf](http://modular.nf/)
```
params.results = "s3://k8s-temporal-efs/results/"

process alignment {
echo true
tag { sample_name + ":" + lane_id }
container '183449805178.dkr.ecr.us-east-1.amazonaws.com/nextflow-container:latest'

pod = [nodeSelector: "${config.instanceType.alignment}, uuid=${params.uuid}"]
cpus "${config.clustercores.alignment}" 
memory "${config.clustermem.alignment}" 

// limit here to avoid overloading the S3 download pool
maxForks 16
maxRetries 3
errorStrategy { sleep(Math.pow(2, task.attempt) * 15 as long); return 'retry' }

storeDir { "${params.results}/$runid/$sample_name/align/$lane_id" }
//publishDir { "${params.results}/$runid/$sample_name/align/$lane_id" }

input:
file reference
set sample_name, sample_id, lane_id, file(read1), file(read2) from fastqs
//val k from kickoff.first()
val k from kickoff

output:
set sample_name, file ("${lane_id}.Aligned.sortedByCoord.out.bam") into alignments
set sample_name, file ("${lane_id}.Chimeric.out.sam") into fusion_sams
set sample_name, file ("${lane_id}.Chimeric.out.junction") into fusion_jxns

set sample_name, file ("${lane_id}.Log.final.out") into star_logs

set sample_name, sample_id, lane_id into meta

script:
template "alignment.sh"
}
```

- Here is the content of alignment.sh
```
#!/bin/bash

source $baseDir/templates/common.sh
script_begin ${task.process} "$sample_name"

{

	echo "runid = ${runid}"
[...]

PR #3016, opened 2022-07-07
fix: escape unstage outputs with double quotes

### LINKED ISSUE #2790: Using `storeDir` in conjunction with s3 path no longer works

We use the [`storeDir`](https://www.nextflow.io/docs/latest/process.html#storedir) directive together with an AWS s3 path to store the results of a time-consuming step. This has historically worked very well for us. At some point, this stopped working for us, and we don't quite understand why.

When we use an s3 path for our storeDir directive now (pointing to a new bucket that doesn't have the cached output from previous iterations), the process succeeds but nexflow fails with 'Missing output file(s)'

Here's a simple workflow that reproduces the issue for me:

```nextflow
#!/usb/bin/env nextflow
nextflow.enable.dsl=2

filename = "foo.txt"

process foo {
    output:
        path "${filename}"
    storeDir params.storeDir
    """
    echo "Hello World" > "${filename}"
    ls -alh "${filename}"
    """
}

workflow {
    foo()
}
```

If you run this with `nextflow run main.nf --storeDir /tmp/foo` it will succeed and the cached file will be present in that directory and be picked up on subsequent runs.

Running it with `nextflow run main.nf --storeDir s3://my-bucket/foo` fails with:

```
Launching `main.nf` [amazing_archimedes] - revision: 27817df6cb
executor >  local (1)
[e8/2e589d] process > foo [100%] 1 of 1, failed: 1 ✘
Error executing process > 'foo'

Caused by:
  Missing output file(s) `foo.txt` expected by process `foo`

Command executed:

  echo "Hello World" > "foo.txt"
  ls -alh "foo.txt"

Command exit status:
  0

Command output:
  -rw-r--r-- 1 brandon brandon 12 Apr 12 09:18 foo.txt

Work dir:
  /home/brandon/ch/nextflow-file-caching-example/work/e8/2e589d8cfb15e177f3ec3afc3b8d35

```

### Program output 

The relevant part of the .nextflow.log file shows the problem:

```
Apr-12 09:18:33.415 [Task submitter] DEBUG nextflow.executor.LocalTaskHandler - Launch cmd line: /bin/bash -ue .command.run
Apr-12 09:18:33.419 [Task submitter] INFO  nextflow.Session - [e8/2e589d] Submitted process > foo
Apr-12 09:18:33.768 [Task monitor] DEBUG n.processor.TaskPollingMonitor - Task completed > TaskHandler[id: 1; name: foo; status: COMPLETED; exit: 0; error: -; workDir: /home/brandon/ch/nextflow-file-caching-example/work/e8/2e589d8cfb15e177f3ec3afc3b8d35]
Apr-12 09:18:33.865 [Task monitor] DEBUG nextflow.processor.TaskProcessor - Process `foo` is unable to find [S3Path]: `/my-bucket/foo/foo.txt` (pattern: `foo.txt`)
Apr-12 09:18:33.881 [Task monitor] ERROR nextflow.processor.TaskProcessor - Error executing process > 'foo'
```

### Environment 

* Nextflow version: 21.10.6.5660
* Java version: openjdk 11.0.14.1 2022-02-08
* Operating system: Fedora 35

### Additional context

If we switch to use `publishDir` instead of `storeDir` with the same bucket, I have no issues, which shows that it's not a local permissions issue. However, `publishDir` doesn't let us use the cached results.

bentsherman:
Closes #2790 #2904 #2912 

- Multiple users have found that `storeDir` doesn't work with S3
- One user found that the problem doesn't occur with Nextflow 20.10, but it does with multiple versions after that
- One user found an issue with the `nxf_unstage` command generated when using `storeDir`:

```bash
nxf_unstage() {
    # ...
    IFS=$'\n'
    for name in $(eval "ls -1d L1.Aligned.sortedByCoord.out.bam L1.Chimeric.out.sam L1.Chimeric.out.junction L1.Log.final.out" | sort | uniq); do
        nxf_s3_upload '$name' s3://k8s-temporal-efs/results/210812_NB501331_0892_AH57FVBGXK/A01-00159459-VL45582-A-d/align/L1 || true
    done
    unset IFS
}
```

Notice that `'$name'` won't be dereferenced correctly because it is in single quotes.

So I basically found every instance where this command is generated and tried to escape `$name` properly in each case. Users have only reported this issue with S3 so far, but I believe this issue affects all cloud providers.


plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchFileCopyStrategy.groovy
```diff
@@ -122,7 +122,7 @@ class AwsBatchFileCopyStrategy extends SimpleFileCopyStrategy {
             uploads=()
             IFS=\$'\\n'
             for name in \$(eval "ls -1d ${escape.join(' ')}" | sort | uniq); do
-                uploads+=("nxf_s3_upload '\$name' s3:/${Escape.path(targetDir)}")
+                uploads+=("nxf_s3_upload \"\$name\" s3:/${Escape.path(targetDir)}")
```
bentsherman, line 125: I think these cases aren't actually broken, when the script is preparing the `uploads` command. Since the `'$name'` is part of an inner double quoted string `"nxf_s3_upload ..."`, I think it is parsed correctly.

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchFileCopyStrategy.groovy
```diff
@@ -154,7 +154,7 @@ class AwsBatchFileCopyStrategy extends SimpleFileCopyStrategy {
     }
 
     static String uploadCmd( String source, Path target ) {
-        "nxf_s3_upload '$source' s3:/$target"
+        "nxf_s3_upload \"$source\" s3:/$target"
```
bentsherman, line 157: These `uploadCmd` methods must be updated because of how they are called in `SimpleFileCopyStrategy`:
```groovy
        return """\
            IFS=\$'\\n'
            for name in \$(eval "ls -1d ${escape.join(' ')}" | sort | uniq); do
                ${stageOutCommand('$name', targetDir, mode)} || true
            done
            unset IFS""".stripIndent(true)
```

pditommaso, line 157:
```suggestion
        "nxf_s3_upload ${Escape.path(source)} s3:/${Escape.path(target)}"
```

Good finding. Likely the best solution is to follow the implementation of 

https://github.com/nextflow-io/nextflow/blob/287e23b5dfaf2262255d9c41b88b9704558d56f4/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchFileCopyStrategy.groovy#L152-L154

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchFileCopyStrategy.groovy
bentsherman: Reverted this change I as confirmed through testing that it is not needed.

pditommaso: cool

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchFileCopyStrategy.groovy
bentsherman: Done

pditommaso:
When escaping the path the quotes are not needed anymore. tip: test locally before committing using `make test` or `make smoke` (to skip slow tests required inet connection)

bentsherman:
I'm not sure what you mean. This method is called with the literal string `'$name'`:
```groovy
"""
            for name in \$(eval "ls -1d ${escape.join(' ')}" | sort | uniq); do
                ${stageOutCommand('$name', targetDir, mode)} || true
            done
"""
```

So the actual path is generated by Bash during task execution. So I'm not sure it even makes sense to call `Escape.path()` if the path is just `'$name'`.

pditommaso:
Ok, let's merge this

syousry6:
Thanks @bentsherman for your efforts, Any ETA for publishing new release with those changes?

syousry6:
or if you are going to create a pre-release for now so we can validate in our pipeline using **NXF_VER** @bentsherman ?

bentsherman:
I believe the next edge release will be made soon. You can also make a "snapshot" version by building from source as described in the README.

syousry6:
Thanks @bentsherman!

Issue #2150, opened 2021-06-03
Staging S3 folder is painfully slow compared to AWS CLI


birnbera:
### Expected behavior and actual behavior

I'm trying to stage an Illumina run directory from S3 for a pipeline and the staging process takes several hours. The same operation using the AWS CLI takes ~20 minutes. I have tried increasing the amount of memory available to the JVM (`Xmx` and `Xms`) as well as the `aws.client.maxConnections` configuration attribute to no avail.

### Steps to reproduce the problem

Create a channel from the run directory path in S3:

```nextflow
Channel.fromPath("s3://<bucket>/<prefix>", type: 'dir').set { runDirectory }
```

vs.

```console
aws s3 sync s3://<bucket>/<prefix> /tmp/run-dir/
```

### Program output 

Staging via Nextflow using the AWS SDK takes >2 hours. Copying using the AWS CLI takes ~20 minutes.

### Environment 

* Nextflow version: `20.07.1`
* Java version: `openjdk 11.0.8 2020-07-14`
* Operating system: Centos 7
* Bash version: `4.2.46(2)-release (x86_64-redhat-linux-gnu)`

birnbera:
I noticed that in the logs Nextflow is only using a single thread per folder:
```log
Jun-09 15:38:43.257 [FileTransfer-thread-1] DEBUG nextflow.file.FilePorter - Copying foreign file s3://<bucket>/<prefix>/
```
It seems that what's missing is the recursive behavior that is implemented by the AWS CLI.

pditommaso:
Bit more complex than that. It may depend on avail cpu in the running computer

http://github.com/nextflow-io/nextflow/blob/6aa102430972266c4e19663926e68c49d1c41c80/modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy#L55-L55

birnbera:
Interesting, I hadn't noticed that, but I have tried running on a computer with a higher number of threads available and it didn't seem to make a difference.

pditommaso:
As of version  21.12.1-edge, nextflow implements an experimental multi-part download for S3 files. 

If can be enabled adding the following option in the config file 

```
aws.client.downloadParallel = true
```

other config settings: 

```
aws.client.downloadChunkSize = '10 MB' // size of the downloaded chunk 
aws.client.downloadNumWorkers = 10     // number of download worker threads
[...]

PR #3017, opened 2022-07-08
Document aws client parallel download options

bentsherman:
Taken from https://github.com/nextflow-io/nextflow/issues/2150#issuecomment-999836354

I couldn't find these options in the code so I'm not sure what the default values are.


pditommaso:
Likely we are going to deprecate or change these setting due to https://github.com/nextflow-io/nextflow/pull/3003

PR #3026, opened 2022-07-12
Make tests pass if user has set git defaultBranch

robsyme:
Two test (`AssetManagerTest` and `LocalRepositoryProviderTest`) fail if the user has set their own `defaultBranch` in `~/.gitconfig`.

Before this PR, the tests assume that the git `defaultBranch` is `master` and that assumption is hardcoded into the tests. This PR will use the local value if set in the most common way (in .gitconfig without any subsections).

Changing the `defaultBranch` is increasingly common - [Gitlab](https://about.gitlab.com/blog/2021/03/10/new-git-default-branch-name/), [Github](https://github.blog/changelog/2020-10-01-the-default-branch-for-newly-created-repositories-is-now-main/), and [Atlassian](https://bitbucket.org/blog/moving-away-from-master-as-the-default-name-for-branches-in-git) have all moved to using `main` instead of `master`, so many developers now user `main` by default. This does not affect the functionality of Nextflow in any way, but ensures that more devs see all green tests when cloning the repo.

To emphasise - this does not change the default branch name in the nextflow repository, and does not affect the name of the default branch name of the repositories created during tests. The only change is that we try to assume less about the dev environment.

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
@@ -72,6 +72,18 @@ class AssetManagerTest extends Specification {
         AssetManager.root = tempDir.root.toFile()
     }
 
+    // Helper method to grab the default brasnch if set in ~/.gitconfig
+    String getUserDefaultBranch() {
+        def defaultBranch = 'master'
+        def gitconfig = Paths.get(System.getProperty('user.home'),'.gitconfig');
+        if(gitconfig.exists()) {
+            def config = new Config()
+            config.fromText(gitconfig.text)
+            defaultBranch = config.getString('init', null, 'defaultBranch') ?: 'master'
+        }
+        return defaultBranch
+    }
+
     def testList() {
 
         given:
@@ -445,15 +457,6 @@ class AssetManagerTest extends Specification {
         dir.resolve('nextflow.config').text = 'manifest {  }'
         dir.resolve('foo.nf').text = 'this is foo content'
 
-        // Use this user's custom defaultBranch name if set in ~/.gitconfig
-        def defaultBranch = 'master'
-        def gitconfig = Paths.get(System.getProperty('user.home'),'.gitconfig');
-        if(gitconfig.exists()) {
-            def config = new Config()
-            config.fromText(gitconfig.text)
-            defaultBranch = config.getString('init', null, 'defaultBranch') ?: 'master'
-        }
-
         def init = Git.init()
         def repo = init.setDirectory( dir.toFile() ).call()
         repo.add().addFilepattern('.').call()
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
```diff
@@ -442,6 +445,15 @@ class AssetManagerTest extends Specification {
...
 
+        // Use this user's custom defaultBranch name if set in ~/.gitconfig
+        def defaultBranch = 'master'
+        def gitconfig = Paths.get(System.getProperty('user.home'),'.gitconfig');
+        if(gitconfig.exists()) {
+            def config = new Config()
+            config.fromText(gitconfig.text)
+            defaultBranch = config.getString('init', null, 'defaultBranch') ?: 'master'
+        }
+
```
pditommaso, line 456: Wonder if this is this is really needed. Git should not be aware that nextflow uses `master` as default irrespective of the user config?

robsyme: In the test below, we create a new git repository
```groovy
def init = Git.init()
```
... and the newly created repo will use whatever default branch name the user has set on their machine. The choice of Nextflow to use `master` uses does not affect this test repository, but the git configuration on the machine running the test _does_ affect this repo. 

It's important to know what the user has set as default branch name, because we check the branch name later on in the test. Before this PR, we assume that the ephemeral test repository will have the branch named `master`:
```groovy
script.revision == 'master'
```

pditommaso: Fair enough, what about turning that snippet into a helper method `getDefaultBranch()` to keep the test focused on the main logic

modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
```diff
@@ -72,6 +72,18 @@ class AssetManagerTest extends Specification {
         AssetManager.root = tempDir.root.toFile()
     }
 
+    // Helper method to grab the default brasnch if set in ~/.gitconfig
+    String getUserDefaultBranch() {
```
pditommaso, line 76:
```suggestion
    String getDefaultBranch() {
```

pditommaso: The main info is that it's the default branch

pditommaso (approved):
Awesome 🙏

modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
robsyme: Done.
I used the `getUserDefaultBranch` name to avoid confusion with the existing `getDefaultBranch` on the manifest object:
```groovy
holder.manifest.getDefaultBranch()
```
... but the risk of confusion is very minimal.

pditommaso: oh, entering paranoid mode and using `getLocalDefaultBranch`? 😆

pditommaso: great!

PR #3027, opened 2022-07-12
Add `k8s.cpuLimits` config option

### LINKED ISSUE #773: Resource limits with K8S

bentsherman:
Closes #773 and #785

This PR adds the ability to specify request/limit values for resource directives (cpus, memory, time, accelerator). In Kubernetes, request/limit values correspond to request and limit in a Pod spec, allowing tasks to have a variable amount of resources over time. However, this feature could be extended in the future to other executors that support min/max values.

Here are the proposed changes:

- Define classes for CpuResource, MemoryResource, TimeResource, AcceleratorResource
- Allow each resource to be specified in the following ways:
  - number (e.g. cpus = 2)
  - number with params (e.g. cpus = 2, limit: 4)
  - map (e.g. cpus = [request: 2, limit: 4])
- Update pod spec builder to use request/limit values for cpus, memory, accelerator


bentsherman:
Integration test is still failing because closures don't work:
```
Not a valid memory value:  -- Check script '/home/runner/work/nextflow/nextflow/tests/dyn-mem.nf' at line: 27
```

`dyn-mem.nf`
```nextflow
process foo {
  memory { x.size() < 10.B  ? 100.MB : 200.MB }
  
  input: 
  file x from Channel.fromPath(['.small.txt','.big.txt'])
  
  script:
  """
  echo task=$x mem=$task.memory 
  """
}
```

bentsherman:
Looks like everything is ready to go 👍

pditommaso:
Nice

modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
```diff
@@ -453,22 +445,39 @@ class PodSpecBuilder {
             container.env = env
 
         // add resources
+        final res = [
+            requests: new LinkedHashMap(10),
+            limits: new LinkedHashMap(10)
```
bentsherman, line 450: Should the initial size be 4 since there are only 4 possible entries?

jacobsin:
This will help fixing #3898

bentsherman:
@pditommaso I removed the new request/limit syntax for memory and disk. In retrospect, it didn't make sense because memory and disk aren't compressible like CPU time. I think I just added them because I could.

So now this PR only adds request/limit for the `cpus` directive. This is all that was originally requested, and it's more pertinent now that Nextflow doesn't set a CPU limit, so users like @jacobsin need a way to do it themselves in some cases.

The PR is much smaller now, so I feel more confident that it won't break anything 🤞

jacobsin:
@bentsherman thanks a lot for the PR - I do believe we need the same for memory where LimitRange can also be applied by k8s admin

@pditommaso any chance for a review?

bentsherman:
For memory you should be fine already because Nextflow sets the request and limit for memory. But in principle I'm fine with supporting separate request / limit for memory, if there is a use case for it.

jacobsin:
Not the latest nextflow version where the limit was removed :(

bentsherman:
The limit was removed only for cpus, not memory.

pditommaso:
Does not this overlaps with https://github.com/nextflow-io/nextflow/pull/2911?

bentsherman:
No it does not

pditommaso:
But why as a user I should specify twice the limit both in as a `cpus` option and as a `resourceLimits`. it's very confusing

bentsherman:
Some executors like K8s allow a lower and upper bound to be specified at the task level, whereas the `resourceLimits` directive is intended to apply a system-wide limit before the task is submitted.

pditommaso:
But I'd assume the limit is the same; why should I specify twice?

bentsherman:
They are different kinds of limits, and I don't think it's possible to collapse them into one setting.

This PR for CPU request/limit is intended to recover an edge case that we lost when we merged #3338. K8s best practice is to not set a CPU limit, but apparently some K8s clusters require you to set a limit (see #3898), so we need a way for users to apply that limit.

But the `resourceLimits` directive in #2911 is applied to the task resources before the task is submitted.

If I wanted to specify a CPU limit with this PR, I would do the following:
```groovy
process.cpus = 4, limit: 4
```

And that would submit a pod with a CPU request and limit.

But if I used the `resourceLimits` directive to do this:
```groovy
process.resourceLimits = [cpus: 4]
process.cpus = 8
```

That would simply reduce the cpus to 8 and submit a task with request=4 but no limit. The K8s executor doesn't know about the limit from the `resourceLimits` directive.

I don't see a clean way to provide both of these behaviors with a single setting.

pditommaso:
> I don't see a clean way to provide both of these behaviors with a single setting.

Indeed it's a b it tricky, but we need to think how to avoid the proliferation of options, especially having the same name in a similar context, it's very confusing. 

I see like this: 

1. use the `limit` option in the `cpus` and `memory` directive with the semantic of upper-bound (i.e. the same as `resourceLimits`)
2. `requests` is always expected to be <= `limit` 
3. by default `limit` is \*not* applied as K8s resource limits (following the rationale of https://github.com/nextflow-io/nextflow/pull/3338)
4. a setting is added in the `k8s` config scope to enable the use of resource limit when specified, in order to fulfil https://github.com/nextflow-io/nextflow/issues/3898) 
5. the setting `nextflow.defaults.cpus = [limit:x]` is used to define a global limit 


### Example 1 

```
process.cpus = [request: 2, limit: 8]
```

It request 2 cpus, cannot go beyond 8 with automatic retry for all processes. No K8s usage. 


### Example 2 

```
k8s.enableResourceLimits = true
process.cpus = [request: 2, limit: 8]
```

It request 2 cpus, cannot go beyond 8 with automatic retry for all processes. The K8s limit is applied. 


### Example 3 

```
nextflow.defaults.cpus = [limit: 8]
process {
  withName: foo { cpus = 2 }
  withName: bar { cpus = 4 }
}
```

`foo` request 2 cpus, `bar` 4. The can both increased up to 8.


### Example 4


```
k8s.enableResourceLimits = true
nextflow.defaults.cpus = [limit: 8]
process {
  withName: foo { cpus = 2 }
  withName: bar { cpus = 4 }
}
```

`foo` requests 2 cpus and applied 8 cpus k8s limit. `bar` requests 4 cpus and applied 8 cpus k8s limit.

bentsherman:
Let's see how these examples look like with my proposal

### Example 1

System-wide CPU limit, no K8s limit

```groovy
process.resourceLimits = [cpus: 8]
process.cpus = 2
```

### Example 2

System-wide limit with K8s limit

```groovy
process.resourceLimits = [cpus: 8]
process.cpus = [request: 2, limit: 8]
```

### Example 3

```groovy
process.resourceLimits = [cpus: 8]
process {
  withName: foo { cpus = 2 }
  withName: bar { cpus = 4 }
}
```

### Example 4

```groovy
process.resourceLimits = [cpus: 8]
process {
  withName: foo { cpus = [request: 2, limit: 8] }
  withName: bar { cpus = [request: 4, limit: 8] }
}
```

bentsherman:
Both approaches are complex but I think mine is still less confusing:

- There is exactly one way to specify a system-wide limit, the `resourceLimits` directive
- Being a process directive, `resourceLimits` can be different for different processes (i.e. for queues with different resource limits)
- The K8s limit can be specified independently of the system-wide limit, which is an important requirement as they serve different purposes

Whereas in your proposal:
- The system-wide limit can be applied two different ways
- You are using a "defaults" scope to set a "default limit", not quite the same thing
- The system-wide limit and K8s limit are tied together, even though users might want to use them differently (i.e. impose a system-wide limit of 8 cpus but have pods with 1-2 cpus)

modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
```diff
@@ -76,6 +76,8 @@ class PodSpecBuilder {
 
     Integer cpus
 
+    boolean cpuLimits = false
```
pditommaso, line 79:
```suggestion
    boolean cpuLimits
```

`false` is implicit

PR #3028, opened 2022-07-12
Fix 1658

varontron:
The problem here (#1658) arises when a batch queue is defined to use multiple compute environments (each with its own ECS cluster.)  When attempting to obtain the container id using the task and cluster identifiers, the call to `ecsClient.describeTasks` will throw an `InvalidParameterException` if the task is not executing on the specific cluster passed in the argument.  

This exception was not being caught.  This had two effects:  

1. a sea of WARN statements on the console, with a new one generated with each polling loop, rendering it very difficult to read (i.e., noise > signal)
2. a short circuit of the loop and thus failure to proceed to obtain the container id in the next iteration, effectively breaking the cost estimate feature

The fix simply catches the `InvalidParameterException` and returns null, enabling the next iteration of the loop.

The fix was tested using tower community edition, first by building from source and confirming the bug, and then implementing the fix, rebuilding, and confirming the behavior.

The gratuitous commit/revert in the log here is due to forgetting to sign the commit the first time. 😬


plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchHelper.groovy
```diff
@@ -100,23 +102,33 @@ class AwsBatchHelper {
         return instanceId ? getInfoByInstanceId(instanceId) : null
     }
 
-    private String getContainerIdByClusterAndTaskArn(String clusterArn, String taskArn) {
+     private String getContainerIdByClusterAndTaskArn(String clusterArn, String taskArn) {
```
pditommaso, line 105:
```suggestion
    private String getContainerIdByClusterAndTaskArn(String clusterArn, String taskArn) {
```

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchHelper.groovy
```diff
@@ -100,23 +102,33 @@ class AwsBatchHelper {
...
-                *.getContainerInstanceArn()
-        if( containers.size()==1 ) {
-            return containers.get(0)
+        try {
+            describeTasksResult = ecsClient.describeTasks(describeTaskReq)
         }
-        if( containers.size()==0 ) {
-            log.debug "Unable to find container id for clusterArn=$clusterArn and taskArn=$taskArn"
+        catch (InvalidParameterException e) {
+            log.debug "Cannot find container id for clusterArn=$clusterArn and taskArn=$taskArn.  Task is likely running on another cluster."
```
pditommaso, line 114:
```suggestion
            log.debug "Cannot find container id for clusterArn=$clusterArn and taskArn=$taskArn - The task is likely running on another cluster"
```

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchHelper.groovy
```diff
@@ -100,23 +102,33 @@ class AwsBatchHelper {
...
             return null
         }
-        else
-            throw new IllegalStateException("Found more than one container for taskArn=$taskArn")
+        if( describeTasksResult ) {
+            final containers =
```
pditommaso, line 118: Likey this can be moved in the `try` block and spare the use of the `if`

pditommaso (changes requested):
Thanks for submitting this PR. it was tricky one. I just left some minor comments

pditommaso:
Merge manually via 3c4d4d3b

PR #3033, opened 2022-07-14
fix: add ps shared objects to Dockerfile

bentsherman:
The Dockerfile uses a clever trick to install the `ps` binary without all the extra stuff in the `procps-ng` package. However, that leaves ps without several shared library files. This PR copies those shared libraries into the final docker image. Tested on my local machine and ps works now.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,6 +3,8 @@ RUN yum install -y procps-ng
 
 FROM amazoncorretto:17.0.3
 COPY --from=0 /bin/ps /bin/ps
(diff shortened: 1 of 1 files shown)


pditommaso:
maybe a this point makes sense just to run the installer in the target container

bentsherman:
Up to you

pditommaso:
It would be nice if you could do

pditommaso:
Nice, thanks

pditommaso (approved):

PR #3048, opened 2022-07-20
Fix AWS Batch job definition conflict

pditommaso:
Nextflow registers AWS batch jobs implicitly when
a new configuration is detected. A local cache maintained
to prevent unnecesary API request to check if job with
the same features is already created.

This commit fixes conflict in the cache keys creation
that caused two or more processes using the same container
image but different task settins e.g. containerOptions
to use a wrong job definition

### DIFF
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
@@ -396,18 +396,18 @@ class AwsBatchTaskHandler extends TaskHandler implements BatchHandler<String,Job
     }
 
     protected String resolveJobDefinition0(String container) {
-        if( jobDefinitions.containsKey(container) )
-            return jobDefinitions[container]
+        final req = makeJobDefRequest(container)
+        final token = req.getParameters().get('nf-token')
+        final jobKey = "$container:$token".toString()
+        if( jobDefinitions.containsKey(jobKey) )
+            return jobDefinitions[jobKey]
 
-        // this could race a race condition on the creation of a job definition
-        // with another NF instance using the same container name
         synchronized(jobDefinitions) {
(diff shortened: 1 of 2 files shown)


plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
```diff
@@ -396,18 +396,18 @@ class AwsBatchTaskHandler extends TaskHandler implements BatchHandler<String,Job
     }
 
     protected String resolveJobDefinition0(String container) {
-        if( jobDefinitions.containsKey(container) )
-            return jobDefinitions[container]
+        final req = makeJobDefRequest(container)
+        final token = req.getParameters().get('nf-token')
```
bentsherman, line 400: Does `nf-token` correspond to the process name?

pditommaso: Nope, it's a unique hash determined by the request settings https://github.com/nextflow-io/nextflow/blob/master/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy#L437-L441

bentsherman: Perfect

bentsherman (approved):

jchorl:
@pditommaso will this be backported? this is a pretty critical fix.

thanks!

pditommaso:
In reality, it only happens when using task-specific `containerOptions` not sure I'd define critical.

jchorl:
> In reality, it only happens when using task-specific containerOptions not sure I'd define critical.

Fair, probably very few people do this. When you do, it's tough to chase down :P. Especially when using very subtle container options. Anyway, sg.

PR #3052, opened 2022-07-21
3011 explore the possibility to allow the support function overloading

jorgeaguileraseqera:
I've created a FuncionNameDef as a replacement for FunctionName but maybe it's unnecessary


modules/nextflow/src/main/groovy/nextflow/script/ComponentDef.groovy
```diff
@@ -38,6 +38,10 @@ abstract class ComponentDef implements Cloneable {
     abstract Object invoke_a(Object[] args)
 
     Object invoke_o(Object args) {
+        if( this instanceof FunctionNameDef){
+            FunctionNameDef fdef = (this as FunctionNameDef)
+            return fdef.invoke_method(args)
+        }
         invoke_a(InvokerHelper.asArray(args))
```
pditommaso, line 45: Why not overriding `invoke_o` in the subclass instead?

modules/nextflow/src/main/groovy/nextflow/script/FunctionNameDef.groovy
```diff
@@ -0,0 +1,82 @@
...
+import static ChannelOut.spreadToArray
+
+/**
+ * Models a function component that can be included from a module script
+ *
+ * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
+ */
+
+@CompileStatic
+class FunctionNameDef extends ComponentDef implements ChainableDef {
```
pditommaso, line 34: I think it would be better to modify `FunctionDef` instead of adding a new class. it's meant exactly for this kind of stuff

modules/nextflow/src/main/groovy/nextflow/script/ComponentDef.groovy
jorgeaguileraseqera: true

modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
```diff
@@ -1254,7 +1254,7 @@ class NextflowDSLImpl implements ASTTransformation {
                 unit.addError( new SyntaxException("Identifier `$name` is reserved for internal use", node.lineNumber, node.columnNumber+8) )
                 return true
             }
-            if( name in functionNames || name in workflowNames || name in processNames ) {
+            if( /*name in functionNames ||*/ name in workflowNames || name in processNames ) {
```
pditommaso, line 1257: Not sure this is needed to remove it. It should prevent the use of reserved named for function if i'm not wrong

modules/nextflow/src/main/groovy/nextflow/script/FunctionDef.groovy
```diff
@@ -33,37 +35,49 @@ class FunctionDef extends ComponentDef implements ChainableDef {
...
-    Method getMethod() { method }
-
-    String getName() { name }
+    String getName() { alias }
 
     BaseScript getOwner() { owner }
 
     Object invoke_a(Object[] args) {
-        method.invoke(owner, spreadToArray(args))
+        throw new MethodNotSupportedException()
```
pditommaso, line 57: Not sure about this. What's the rationale?

modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy
```diff
@@ -170,14 +170,15 @@ class ScriptMeta {
...
-        final result = new ArrayList(allMethods.length)
+        final result = new ArrayList<FunctionDef>(allMethods.length)
         for( Method method : allMethods ) {
             if( !Modifier.isPublic(method.getModifiers()) ) continue
             if( Modifier.isStatic(method.getModifiers())) continue
             if( method.name.startsWith('super$')) continue
             if( method.name in INVALID_FUNCTION_NAMES ) continue
 
-            result.add(new FunctionDef(script, method))
+            if( result.find{it.name == method.name}) continue
```
pditommaso, line 180: What this does?

modules/nextflow/src/test/groovy/nextflow/script/ScriptIncludesTest.groovy
```diff
@@ -149,7 +149,7 @@ class ScriptIncludesTest extends Dsl2Spec {
         result.val == 'dlrow olleh'
     }
 
-    def 'should not allow duplicate functions' () {
+    def 'should allows duplicate functions' () {
```
pditommaso, line 152: Let's add a bit more tests, e.g. the invocation of `foo()` and at least another test with a more complex signature e.g. a string + a optional map

modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy
jorgeaguileraseqera: simply avoid to have duplicated FunctionDef objects in the list

modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
jorgeaguileraseqera: if we want to have exactly one method (with default args) we can maintain but if we want to have the same method with different signatures is required to remove it from the check

modules/nextflow/src/main/groovy/nextflow/script/FunctionDef.groovy
jorgeaguileraseqera: `invoke_a` is an abstract method from ComponentDef and new FunctionDef not use it as it override `invoke_o`

pditommaso: But it can via [ChainableDef](https://github.com/nextflow-io/nextflow/blob/20cfbfd3595270b0a71a32fe95de52936bbd475b/modules/nextflow/src/main/groovy/nextflow/script/ChainableDef.groovy#L36-L36), [here](https://github.com/nextflow-io/nextflow/blob/20cfbfd3595270b0a71a32fe95de52936bbd475b/modules/nextflow/src/main/groovy/nextflow/extension/ChannelEx.groovy#L181-L181) and [here](https://github.com/nextflow-io/nextflow/blob/20cfbfd3595270b0a71a32fe95de52936bbd475b/modules/nextflow/src/main/groovy/nextflow/extension/ChannelEx.groovy#L203-L203)

jorgeaguileraseqera: If I'm not wrong, and reading the comment of the method, left in both cases is a Process, not a Function

pditommaso: of course, the comment is wrong 😄. look at this example 

https://github.com/nextflow-io/nextflow/blob/20cfbfd3595270b0a71a32fe95de52936bbd475b/modules/nextflow/src/test/groovy/nextflow/script/ScriptPipesTest.groovy#L276-L282

jorgeaguileraseqera: Ok, fixed and added more test

modules/nextflow/src/test/groovy/nextflow/script/ScriptIncludesTest.groovy
jorgeaguileraseqera: done

modules/nextflow/src/main/groovy/nextflow/script/FunctionDef.groovy
```diff
@@ -33,37 +35,49 @@ class FunctionDef extends ComponentDef implements ChainableDef {
...
+                : args instanceof ChannelOut ? (args as List) as Object[]
+                : new Object[]{args};
+        def meta = owner.metaClass.getMetaMethod(name, argsArr)
+        if( meta == null )
+            throw new MissingMethodException(name, owner.getClass(), args)
+        Method callMethod = owner.getClass().getMethod(name, meta.getNativeParameterTypes())
+        if( callMethod == null )
+            throw new MissingMethodException(name, owner.getClass(), args)
+        callMethod.invoke(owner, argsArr)
     }
```
pditommaso, line 72: The only worry I have is about the performance impact of this. But I guess for now is fine.

pditommaso:
Reshaped a bit to make uniform with other "handlers". have a look 930ee432

Issue #3050, opened 2022-07-21
Add support for disk directive when using Google Batch execution


pditommaso:
The Google Batch API allows the control of the size of volume attached to the instance running the job via the [following setting](https://github.com/googleapis/googleapis/blob/master/google/cloud/batch/v1/job.proto#L247). 

The goal of this issue is to allow nextflow users to control this setting by using the [disk](https://www.nextflow.io/docs/latest/process.html#disk) directive.

Issue #3049, opened 2022-07-21
Add support for Google Batch boot disk size


pditommaso:
The Google Batch API allows the control of the boot disk size via the [following setting](https://github.com/googleapis/googleapis/blob/master/google/cloud/batch/v1/task.proto#L42). 

The Goal of this issue is to add support for it in Nextflow via the configuration setting 

```
google.lifeSciences.bootDiskSize = <size using MemUnit type>
```

hnawar:
Quick question: While in Life Sciences API the default behaviour was the API copied the input files to the Data disk, this is not necessary the case with Batch. At the moment the TaskHandler copies the data to the boot disk and simply processes the data on it.
If Disk directive in the process resizes the boot disk then the bootDiskSize is no longer needed, unless the future plans it to still use two or more separate disks (Boot disk + Data siks(s)) then both would be needed

pditommaso:
The use of gcs mount made by the job facilitates things, but still, the task works in the node scratch storage and copies the result back in the bucket. Therefore the support for `disk` can be useful. 

The boot disk instead can be useful when downloading a fat docker image, that the docker driver stores in the boot volume

PR #3055, opened 2022-07-22
Add accelerator, boot disk, cpu platform, disk to google batch

### LINKED ISSUE #3039: Add support for GPU accelerator to Google Batch executor
The Google Batch API provides support for [GPU acceleration](https://cloud.google.com/compute/docs/gpus), however, this feature is currently non-supported by the Nextflow executor. 

The goal of this issues is to extend the Google Batch executor for Nextflow to add the support for the [accelerator](https://www.nextflow.io/docs/latest/process.html#accelerator) directive to allow the use GPU from Nextflow tasks. 

A [similar implementation for Google LS executor](https://github.com/nextflow-io/nextflow/blob/0d779e1f314faaa404587d79b1cfa8380b341997/plugins/nf-google/src/main/nextflow/cloud/google/lifesciences/GoogleLifeSciencesHelper.groovy#L222-L227) can be used as a starting point.

bentsherman:
Closes #3039 #3049 #3050 

This PR adds the following features to google batch:
- GPU support via `accelerator` directive
- Boot disk size (same as life sciences)
- Minimum CPU platform (same as life sciences)
- Disk support via `disk` directive


pditommaso:
Let's keep in separate PRs please

PR #3065, opened 2022-07-25
WIP: add efs support

jchorl:
This adds support for EFS volumes to nextflow on AWS batch.

The syntax is `efs://<efs-id>:<container-mount-point>[:ro|rw]`. I ~copied the `job-definition://` prefix for container.

Here's an example from nextflow.config:
```
aws {
        batch {
                volumes = '/data:/tmp,efs://fs-0222898f792085402:/efsmount:ro'
        }
}
```

The use-case: some files are big to stage in, and it would be great to just use batch efs volumes.

This is WIP. It does work! But please let me know if it is directionally correct and I'd be happy to help add docs/tests.

### DIFF
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
@@ -32,6 +32,7 @@ import com.amazonaws.services.batch.model.DescribeJobDefinitionsRequest
 import com.amazonaws.services.batch.model.DescribeJobDefinitionsResult
 import com.amazonaws.services.batch.model.DescribeJobsRequest
 import com.amazonaws.services.batch.model.DescribeJobsResult
+import com.amazonaws.services.batch.model.EFSVolumeConfiguration
 import com.amazonaws.services.batch.model.EvaluateOnExit
 import com.amazonaws.services.batch.model.Host
 import com.amazonaws.services.batch.model.JobDefinition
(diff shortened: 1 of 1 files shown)


pditommaso:
Thanks for submitting this PR, however this feature overlaps with the support provided by this plugin https://github.com/seqeralabs/xpack-amzn/blob/master/docs.md

jchorl:
Hi @pditommaso ! Thanks for the response. Agreed, it's not useful to merge overlapping features in different places.

In this case, it seems to be a feature implemented by a licensed plugin (even though the plugin is open source).

Can you help us understand what your policy is around fully community-developed features that may overlap with enterprise offerings? Should we be worried that parts of open-source nextflow will be moved behind licensed plugins in the future?

This will help us understand how we should think about future use of nextflow and investment into contributing changes back to the community

pditommaso:
Hi @jchorl! apology for the late reply. 

> Should we be worried that parts of open-source nextflow will be moved behind licensed plugins in the future?

Absolutely not, we are fully committed to keeping Nextflow open source and free with no restrictions. 

Our policy is to license selected plugins that provide support for very specific features that are commonly required by users deploying Nextflow in the industry, like for example the support for AWS EFS or AWS secret manager. 

In other words, we want the open-source Nextflow experience to be as complete as possible. Users who need to "extend" Nextflow in specific ways can pay for commercial extension packs. Of course, anyone is free to develop their own open-source versions of our extension packs. However, our licensed plugins receive dedicated support from Seqera.

Also, extension packs are one of several ways that we fund Nextflow development. While it means that some features won't be made open source, the upside is that we now have several dedicated Nextflow developers who will be able to make the core experience much better. We still welcome contributions from the community, the same as always.

Hope this clarify your doubts.

jchorl:
Got it! All makes sense :)

Thanks for all the hard work.

PR #3066, opened 2022-07-25
Report warning when google batch quota is exceeded

### LINKED ISSUE #3051: Google Batch spot setting results in an request error
When enabling the use of spot instances using the following option 

```
google.batch.spot = true
```

The resulting request returns the following error 

```
Jul-21 08:11:49.759 [Task submitter] TRACE nextflow.processor.TaskProcessor - Handling unexpected condition for
  task: id: 3; name: resolve_align:FASTP (CEG9330132-19-02, L001); type: SCRIPTLET; exit: -; error: null; workDir: /a5/b80963826030fdf9fd1389f4c53966
  error [nextflow.cloud.google.batch.client.BatchResponseException]: Request POST https://batch.googleapis.com/v1/projects/some-nextflow/locations/europe-north1/jobs?job_id=nf-a5b80963-1658391109074 returned an error code=400

  {
      "error": {
          "code": 400,
          "message": "Invalid JSON payload received. Unknown name \"instance\" at 'job.allocation_policy': Cannot find field.",
          "status": "INVALID_ARGUMENT",
          "details": [
              {
                  "@type": "type.googleapis.com/google.rpc.BadRequest",
                  "fieldViolations": [
                      {
                          "field": "job.allocation_policy",
                          "description": "Invalid JSON payload received. Unknown name \"instance\" at 'job.allocation_policy': Cannot find field."
                      }
                  ]
              }
          ]
      }
  }
```

bentsherman:
Closes #3051

### DIFF
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
@@ -232,7 +232,7 @@ class GoogleBatchTaskHandler extends TaskHandler {
                 final eventsCount = newStatus.getStatusEventsCount()
                 final lastEvent = eventsCount > 0 ? newStatus.getStatusEvents(eventsCount - 1) : null
                 if( lastEvent?.getDescription().contains('CODE_GCE_QUOTA_EXCEEDED') )
-                    log.warn "[GOOGLE BATCH] job cannot be run: ${lastEvent.getDescription()}"
+                    log.warn1 "Batch job cannot be run: ${lastEvent.getDescription()}"
             }
         }
         return jobState
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
@@ -221,11 +221,18 @@ class GoogleBatchTaskHandler extends TaskHandler {
         final now = System.currentTimeMillis()
         final delta =  now - timestamp;
         if( !jobState || delta >= 1_000) {
-            def newState = client.getJobState(jobId)
+            final newStatus = client.getJobStatus(jobId)
+            final newState = client.getJobState(newStatus)
             if( newState ) {
                 log.trace "[GOOGLE BATCH] Get job=$jobId state=$newState"
                 jobState = newState
                 timestamp = now
+            }
+            if( newState == 'SCHEDULED' ) {
+                final eventsCount = newStatus.getStatusEventsCount()
+                final lastEvent = eventsCount > 0 ? newStatus.getStatusEvents(eventsCount - 1) : null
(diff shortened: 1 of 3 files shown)


plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -221,12 +221,19 @@ class GoogleBatchTaskHandler extends TaskHandler {
...
             if( newState ) {
                 log.trace "[GOOGLE BATCH] Get job=$jobId state=$newState"
                 jobState = newState
                 timestamp = now
             }
+            if( newState == 'SCHEDULED' ) {
+                final eventsCount = newStatus.getStatusEventsCount()
+                final lastEvent = eventsCount > 0 ? newStatus.getStatusEvents(eventsCount - 1) : null
+                if( lastEvent?.getDescription().contains('CODE_GCE_QUOTA_EXCEEDED') )
+                    log.warn "[GOOGLE BATCH] job cannot be run: ${lastEvent.getDescription()}"
```
pditommaso, line 235:
```suggestion
                    log.warn1 "Batch job cannot be run: ${lastEvent.getDescription()}"
```

Since this is shown to the user better to make it a bit more readable with the `[..]` prefix. Also to prevent printing multiple time the same message better to use `warn1`

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -221,12 +221,19 @@ class GoogleBatchTaskHandler extends TaskHandler {
         final now = System.currentTimeMillis()
         final delta =  now - timestamp;
         if( !jobState || delta >= 1_000) {
-            def newState = client.getJobState(jobId)
+            final newStatus = client.getJobStatus(jobId)
+            final newState = client.getJobState(newStatus)
```
pditommaso, line 225: This calls twice `describeJob` API, which would be better to avoid. It should be enough to use `getJobStatus` and then take the state from the result as done here 

https://github.com/nextflow-io/nextflow/blob/master/plugins/nf-google/src/main/nextflow/cloud/google/batch/client/BatchClient.groovy#L106-L107

bentsherman: The API is only called once, the `getJobState()` is just a helper function to get the state from the status object.

pditommaso: I see it now, but in this case, does not make much sense to make it a client class method. It would be better to method in the GoogleBatchTaskHandler class

bentsherman: Well this method is also used by BatchClient which is why I left it there

bentsherman: how about this, I will move `getJobState(String jobId)` to the task handler as it is the convenience method

PR #3069, opened 2022-07-26
feature: set explicit permissions for test files

jorgeaguileraseqera:
Seems some update of Linux or some vendor Java versions are creating files with unexpected permissions in tests


pditommaso:
Thanks!

PR #3072, opened 2022-07-27
bugfix: emit relative path when relative parameter is set

### LINKED ISSUE #2990: channel.fromPath's `relative` parameter not respected when given a file path

When creating a channel using `channel.fromPath()` and providing a path to a specific file, setting the `relative` parameter to `true` doesn't have any effect, so the resulted path is absolute rather than relative.

Create a `main.nf` file with the following content:
```
data = channel.fromPath("main.nf", relative: true)

process p1 {
    echo true

    input:
    val path from data

    """
    #!/usr/bin/env bash
    echo $path
    """
}
```

### Program output 

```
/absolute/path/to/nextflow/script/main.nf
```

### Environment 

* Nextflow version: 21.10.6
* Java version: 1.8.0_321
* Operating system: macOS
* Bash version: GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin20)

### Additional context

Expected output is:
```
main.nf
```

Adding an asterisk at the end of the path (i.e. `data = channel.fromPath("main.nf*", relative: true)` solves it, but isn't really a good workaround.

jorgeaguileraseqera:
closes #2990

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/ChannelTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/ChannelTest.groovy
@@ -269,6 +269,28 @@ class ChannelTest extends Specification {
         result == ['alpha.txt']
     }
 
+    def testRelative() {
+
+        setup:
+        def file1 = new File('alpha.txt')
+
+        when:
+        def result = Channel
+                .fromPath([relative:false], 'alpha.txt')
+                .toSortedList().getVal()
+        then:
+        result*.toString() == [ file1.absolutePath ]
+
+        when:
+        result = Channel
+                .fromPath([relative:true], 'alpha.txt')
+                .toSortedList().getVal()
+        then:
+        result*.toString() == [ 'alpha.txt' ]
+
+        cleanup:
+        file1.delete()
+    }
 
     def testGlobHiddenFiles() {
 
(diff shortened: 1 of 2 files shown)


pditommaso:
and there's any test? 😄

jorgeaguileraseqera:
> and there's any test? smile

of course but forget to commit it

modules/nf-commons/src/main/nextflow/file/FileHelper.groovy
```diff
@@ -1038,10 +1038,10 @@ class FileHelper {
         final checkIfExists = opts?.checkIfExists as boolean
         final followLinks = opts?.followLinks == false ? [LinkOption.NOFOLLOW_LINKS] : Collections.emptyList()
         if( !checkIfExists || FilesEx.exists(result, followLinks as LinkOption[]) ) {
-            return result
+            return opts?.relative ? path : result
         }
 
-        throw new NoSuchFileException(FilesEx.toUriString(result))
+        throw new NoSuchFileException(FilesEx.toUriString(opts?.relative ? path : result))
```
pditommaso, line 1044:
```suggestion
        throw new NoSuchFileException(opts?.relative ? path : FilesEx.toUriString(result))
```

I think when it's relative makes no sense to use `FilesEx.toUriString`

modules/nextflow/src/test/groovy/nextflow/ChannelTest.groovy
```diff
@@ -269,6 +269,28 @@ class ChannelTest extends Specification {
...
+        then:
+        result*.toString() == [ file1.absolutePath ]
+
+        when:
+        result = Channel
+                .fromPath([relative:true], 'alpha.txt')
+                .toSortedList().getVal()
+        then:
+        result*.toString() == [ 'alpha.txt' ]
+
```
pditommaso, line 290: Let's add a test for the default case (absence of `relative` option)

Issue #2982, opened 2022-06-21
Allow the import for custom function from plugin extensions


pditommaso:
As of version 22.04.x nextflow allows the import of custom channel factories and operators from custom plugins. 

The goal of this issue is to extend this functionality to allow the inclusion of arbitrary custom functions i.e. methods provided by a custom plugin. 

The syntax to allow this is going to remain the same ie. 

```
include { custom_fun } from 'plugin/foo'
```

Allows the users to invoke the `custom_fun` in the including script context. 

The usual aliasing  mechanism should be supported e.g. 

```
include { custom_fun as my_fun } from 'plugin/foo'
```

It will allow to include the function as `my_fun`. 

### Possible implementation 

* Extend the plugin [include mechanism](https://github.com/nextflow-io/nextflow/blob/a06b444236164b88218ef548f7bf834727faa33f/modules/nextflow/src/main/groovy/nextflow/script/IncludeDef.groovy#L214-L214) to allow the inclusion of custom function other than operators 
* To class method(s) declared in the plugin extension class should be annotated with a new annotation `@Function`. 

* Along the same way factory method (eg. method returning a `DataflowWriteChannel`) should be marked as `@Factory` and operates as `@Operator` 

* A new `PluginFunctionDef` should be created as an extension `ComponentDef` to hold the function name, the target object plugin and other relevant metadata .

* The function name when imported should be added to [imports](https://github.com/nextflow-io/nextflow/blob/a06b444236164b88218ef548f7bf834727faa33f/modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy#L90-L90) map of the ScritMeta of the script from when the inclusion is performed. 

### Caveats 

* Function overloading ie. different signatures for the same function name should not be supported. 
* If the user tries to include a function with a name already taken by a process/workflow/function should thrown an error

PR #3079, opened 2022-07-28
2982 allow the import for custom function from plugin extensions

jorgeaguileraseqera:
This is a huge PR witch includes a refactor to test plugins in nf-commons. 

The idea is to remove the nf-hello binary and use a new subproject nf-plugin-template 

About inclusion of custom function the approach of this PR is to include public methods instead annotation approach (which seems more simple)


pditommaso:
I think this diverges too much from what I was proposing in #2982. We should avoid adding a new extension point class (there's too much boilerplate code for plugin writers). 

Instead for the sake of this prototype, it could be used `ChannelExtensionPoint` (then we'll rename it to something different). 

My idea is to mark the methods to be picked as function extensions with a custom annotation `@Function`  (along the same manner, it should be done for `@Operator` and use it [here](http://github.com/nextflow-io/nextflow/blob/daa6edc8e26b08fa8a4cf55332d037bcdd24e2cb/modules/nextflow/src/main/groovy/nextflow/extension/ChannelExtensionProvider.groovy#L153-L153) and `@Factory` to be used [here](http://github.com/nextflow-io/nextflow/blob/daa6edc8e26b08fa8a4cf55332d037bcdd24e2cb/modules/nextflow/src/main/groovy/nextflow/extension/ChannelExtensionProvider.groovy#L169-L169))

jorgeaguileraseqera:
Ok, I see, on it

jorgeaguileraseqera:
Changed and added more tests importing functions in modules and so on

modules/nextflow/src/main/groovy/nextflow/extension/Function.groovy
```diff
@@ -0,0 +1,16 @@
+package nextflow.extension
+
+/**
+ * @author : jorge <jorge.aguilera@seqera.io>
```
pditommaso, line 4: Let's put some description here, please

modules/nextflow/src/main/groovy/nextflow/script/FunctionDef.groovy
```diff
@@ -47,7 +51,7 @@ class FunctionDef extends ComponentDef implements ChainableDef {
 
     String getName() { alias }
 
-    BaseScript getOwner() { owner }
+    Object getOwner() { owner }
```
pditommaso, line 54: I'm wrong, or nothing is using `getOwner()`. In that case I think it's better to remove it

modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy
```diff
@@ -202,6 +202,16 @@ class ScriptMeta {
         return this
     }
 
+    ScriptMeta addFunctionDefinition(FunctionDef component) {
```
pditommaso, line 205: I'm wondering why this check is only needed for function imports. What if a user include process `foo` and then a function `foo` from another module? 

Maybe is needed a general check here? 

https://github.com/nextflow-io/nextflow/blob/2982-allow-the-import-for-custom-function-from-plugin-extensions/modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy#L192-L194

modules/nextflow/src/main/groovy/nextflow/extension/ChannelExtensionProvider.groovy
```diff
@@ -143,8 +148,12 @@ class ChannelExtensionProvider implements ExtensionProvider {
                 ChannelFactoryInstance factoryInstance = new ChannelFactoryInstance(ext)
                 factoryExtensions.put(aliasName, new PluginExtensionMethod(method:realName, target:factoryInstance))
             }
+            else if( definedFunctions.contains(realName) ){
+                FunctionDef functionDef = new FunctionDef(ext, realName, aliasName )
+                meta.addFunctionDefinition(functionDef)
+            }
             else{
-                throw new IllegalStateException("Operator '$realName' it isn't defined by plugin ${existing.getClass().getName()}")
+                throw new IllegalStateException("Operator '$realName' it isn't defined by plugin ${pluginId}")
```
pditommaso, line 156:
```suggestion
                throw new IllegalStateException("Extension '$realName' it isn't defined by plugin ${pluginId}")
```

pditommaso:
Nice, much cleaner now. Wondering if the annotation approach should not be extended also to operators and factories here 

https://github.com/nextflow-io/nextflow/blob/2982-allow-the-import-for-custom-function-from-plugin-extensions/modules/nextflow/src/main/groovy/nextflow/extension/ChannelExtensionProvider.groovy#L162-L192

jorgeaguileraseqera:
> Nice, much cleaner now. Wondering if the annotation approach should not be extended also to operators and factories here
> 
> https://github.com/nextflow-io/nextflow/blob/2982-allow-the-import-for-custom-function-from-plugin-extensions/modules/nextflow/src/main/groovy/nextflow/extension/ChannelExtensionProvider.groovy#L162-L192
> 
> The public code would be much more readable

yes, I was thinking the same. Maybe we can maintain both approaches for a while a deprecated current after a couple of releases

Also we can extend annotations with some attributes, i.e. developer of the Function can specificy and alias (different from the alias of the user) and use it instead of the method

```
@Function(alias = 'myMethod' )
def thisMethodDoThisAndThat(){
}
```

modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy
jorgeaguileraseqera: True, will create a test to cover this situation

pditommaso:
> Maybe we can maintain both approaches for a while a deprecated current after a couple of releases

Make sense 

> Also we can extend annotations with some attributes, i.e. developer of the Function can specificy and alias

let's keep as future extension

modules/nextflow/src/main/groovy/nextflow/extension/ChannelExtensionProvider.groovy
```diff
@@ -154,6 +163,14 @@ class ChannelExtensionProvider implements ExtensionProvider {
         def result = new HashSet<String>(30)
         def methods = clazz.getDeclaredMethods()
         for( def handle : methods ) {
+            // in a future only annotated methodS will be imported
+            if( handle.isAnnotationPresent(Operator)) {
+                def params=handle.getParameterTypes()
+                if( params.length>0 && isReadChannel(params[0]) ) {
```
pditommaso, line 169: It would be better to return an error if `isReadChannel(params[0])` is false. In a second iteration, it would be cool to have an AST rule that check this during the compilation

pditommaso: Let's add a warning instead if the operator is discovered via the deprecated snippet 

```
            if( params.length>0 && isReadChannel(params[0]) )
                result.add(handle.name)
```

modules/nextflow/src/main/groovy/nextflow/extension/ChannelExtensionProvider.groovy
```diff
@@ -170,6 +187,14 @@ class ChannelExtensionProvider implements ExtensionProvider {
         def result = new HashSet<String>(30)
         def methods = clazz.getDeclaredMethods()
         for( def handle : methods ) {
+            // in a future only annotated methodS will be imported
+            if( handle.isAnnotationPresent(Factory)) {
+                def returnType =handle.getReturnType()
+                if( isWriteChannel(returnType) ) {
```
pditommaso, line 193: Also here, let's report an error if the return type is wrong and a warning the deprecated case

modules/nf-commons/build.gradle
```diff
@@ -38,3 +38,4 @@ dependencies {
     testImplementation(testFixtures(project(":nextflow")))
 }
 
+test.dependsOn(":nf-plugin-template:makeZip")
```
pditommaso, line 41: Is this really needed here?

modules/nf-plugin-template/src/main/nextflow/hello/HelloExtension.groovy
```diff
@@ -0,0 +1,155 @@
...
+
+import java.util.concurrent.CompletableFuture
+
+/**
+ * @author : jorge <jorge.aguilera@seqera.io>
+ *
+ */
+@Slf4j
+@Scoped('hello')
+class HelloExtension extends ChannelExtensionPoint{
```
pditommaso, line 28: Is this needed for testing purposes? would not be better to include into `test` source three of another submodule? or a [test fixture](https://docs.gradle.org/current/userguide/java_testing.html#declaring_dependencies_of_test_fixtures)?

modules/nf-commons/build.gradle
jorgeaguileraseqera: this is the way I found to have updated the plugin before running tests automatically. With this hack before to run a test in nf-commons every change in nf-plugin-template is tested

modules/nf-plugin-template/src/main/nextflow/hello/HelloExtension.groovy
jorgeaguileraseqera: The idea of nf-plugin-template is to have a "full" plugin that can be tested as it.

In this way, we build the zip before running the test and the FakeIndex serves it as a typical plugin

pditommaso: Understand that, but this will enter into the distribution I guess. Wondering if moving into test source tree of the nf-commons project would work the same?

jorgeaguileraseqera: I think if we move to nf-commons we can run the typical tests calling functions, validating functions are detected, etc  but I think we'll not be able to test the plugin as integrated in a DSL 

If we decide is not required to test the DSL (at least via spock), probably we can remove the nf-plugin-template and move HelloExtension to nf-commons fixtures

pditommaso: Not sure, I think it should work because when the plugin class in on the classpath it does not have to be loaded and started explicitly. This is indeed how it works during dev (see also [here](https://pf4j.org/doc/system-extension.html))

jorgeaguileraseqera: Not sure how to do it.

I'm able in the test to do `def tmp = Plugins.getExtension(HelloExtension)` and it load the extension (once I've changed the build.gradle)

but the test always fails with `Unknown Nextflow plugin 'nf-plugin-template'` when `includeDef` try to load it so probably I'm missing something

pditommaso: Umm, maybe it's needed to add the `test` path here. https://github.com/nextflow-io/nextflow/blob/2982-allow-the-import-for-custom-function-from-plugin-extensions/modules/nf-commons/src/main/nextflow/plugin/DevPluginClasspath.groovy#L38-L38

jorgeaguileraseqera: Took me a while but I think I found the way to build and load the plugin

https://github.com/nextflow-io/nextflow/blob/c59ea8483b9d79305ff422926a6826b469ee5a2e/modules/nf-commons/src/test/nextflow/plugin/BuildPluginTrait.groovy

pditommaso: cool!

pditommaso:
The build trait is a good idea, but it's not needed. I've managed to make it work adding plugin manager that looks for the plugin in in the `testFixtures` source tree. Look at [TestPluginManager](https://github.com/nextflow-io/nextflow/blob/c240ca98b35d5c331012bc54d1d8f8b3adf5d0ce/modules/nf-commons/src/testFixtures/groovy/nextflow/plugin/TestPluginManager.groovy#L34-L34) and related classes.

This allows also to get rid of the `nf-plugin-template` sub-project. I've made the following changes: 

* `modules/nf-plugin-template/src/main/nextflow/hello/*` was moved to `modules/nf-commons/src/testFixtures/groovy/nextflow/hello/*`
* `modules/nf-plugin-template/src/test/nextflow/hello` was moved to `modules/nf-commons/src/test/nextflow/hello` 
* new plugin manager add `modules/nf-commons/src/testFixtures/groovy/nextflow/plugin/*` 
* some other minor changes to adapt the tests

PR #3086, opened 2022-08-01
Fix #3074

### LINKED ISSUE #3074: Add warning when a NXF_ env variable is set in nextflow.config env scope
Nextflow will ignore attempts to set NXF_ variables in env scope of the nextflow.config file. These environment variables should be set outside Nextflow.

The idea is to:
- [x] mention this explicitly in the documentation (https://github.com/nextflow-io/nextflow/pull/3083)
- [x] Make Nextflow throw a warning when an user attempts to set a NXF_ env variable within the env scope https://github.com/nextflow-io/nextflow/pull/3086

mribeirodantas:
If NXF environment variables other than NXF_DEBUG are set in the
env scope of nextflow.config, a warning should be thrown because
NXF variables are ignored there. They should be set in the
environment where nextflow is launched.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
@@ -310,10 +310,7 @@ class CmdRun extends CmdBase implements HubOptions {
         final nxfEnvs = checkNxfEnv(config)
         if(nxfEnvs) {
             log.warn 'Environment variables in the `env` config scope are not applied to Nextflow -- See https://www.nextflow.io/docs/latest/config.html#scope-env'
-            log.warn 'The following variables should be exported in the environment:'
-            nxfEnvs.each {
-                log.warn it as String
-            }
+            log.warn "The following variables should be exported in the environment: ${nxfEnvs.join(', ')}"
         }
 
         // -- load plugins
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
@@ -306,6 +306,18 @@ class CmdRun extends CmdBase implements HubOptions {
         // check DSL syntax in the config
         launchInfo(config, scriptFile)
 
+        // Warn about setting NXF_ environment variables within nextflow
+        // config file
+        def nxfEnvs = config
+                .env
+                .findAll{
+                    it -> it.toString().startsWith('NXF_') && ! it.toString().startsWith('NXF_DEBUG')
+                }
+        if (nxfEnvs.size() > 0) {
+            log.warn "NXF environment variables are ignored in the env scope of nextflow.config. Check https://www.nextflow.io/docs/latest/config.html#scope-env"
+
+        }
+
         // -- load plugins
         final cfg = plugins ? [plugins: plugins.tokenize(',')] : config
         Plugins.load(cfg)
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
@@ -306,14 +306,8 @@ class CmdRun extends CmdBase implements HubOptions {
         // check DSL syntax in the config
         launchInfo(config, scriptFile)
 
-        // Warn about setting NXF_ environment variables within env config scope
-        final nxfEnvs = config.env.findAll { it ->
-            it.toString().startsWith('NXF_') &&
-            ! it.toString().startsWith('NXF_DEBUG')
-        }
-        if( nxfEnvs.size() > 0 ) {
-            log.warn 'Environment variables in the `env` config scope are not applied to Nextflow -- See https://www.nextflow.io/docs/latest/config.html#scope-env'
-        }
+        // check if NXF_ variables are set in nextflow.config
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -306,6 +306,18 @@ class CmdRun extends CmdBase implements HubOptions {
...
+        // config file
+        def nxfEnvs = config
+                .env
+                .findAll{
+                    it -> it.toString().startsWith('NXF_') &&
+                            ! it.toString().startsWith('NXF_DEBUG')
+                }
+        if (nxfEnvs.size() > 0) {
+            log.warn "NXF environment variables are ignored in the env scope of nextflow.config. Check https://www.nextflow.io/docs/latest/config.html#scope-env"
+        }
```
bentsherman, line 319:
```suggestion
        // Warn about setting NXF_ environment variables within env config scope
        final nxfEnvs = config.env.findAll { it ->
            it.toString().startsWith('NXF_') &&
            ! it.toString().startsWith('NXF_DEBUG')
        }
        if( nxfEnvs.size() > 0 ) {
            log.warn 'Environment variables in the `env` config scope are not applied to Nextflow -- See https://www.nextflow.io/docs/latest/config.html#scope-env'
        }
```

mribeirodantas:
Oops, I didn't know I had to sign off again when accepting co-authorship. Learned one new thing today 😅

bentsherman:
Just showing some conventions we typically follow, but here is some secret groovy starter knowledge:
- prefer `final` over `def` when you don't need to reassign the variable
- prefer single quote strings over double quote when you don't need to inject variables into the string (single quote strings can't be dynamic)

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -306,6 +306,15 @@ class CmdRun extends CmdBase implements HubOptions {
...
 
+        // Warn about setting NXF_ environment variables within env config scope
+        final nxfEnvs = config.env.findAll { it ->
+            it.toString().startsWith('NXF_') &&
+            ! it.toString().startsWith('NXF_DEBUG')
+        }
+        if( nxfEnvs.size() > 0 ) {
+            log.warn 'Environment variables in the `env` config scope are not applied to Nextflow -- See https://www.nextflow.io/docs/latest/config.html#scope-env'
+        }
+
```
pditommaso, line 317: ok, let's move this into separate method that takes the config as an argument, and make a unit test for it

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -306,6 +306,15 @@ class CmdRun extends CmdBase implements HubOptions {
         // check DSL syntax in the config
         launchInfo(config, scriptFile)
 
+        // Warn about setting NXF_ environment variables within env config scope
+        final nxfEnvs = config.env.findAll { it ->
```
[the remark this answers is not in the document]
mribeirodantas: It's an object. It's not entirely clear to me why the value of env, which is a key in the config map, is an object but it is 😅 . That's why I can't call `.keySet()`. I saw in other parts of the code that you use `.navigate()` to check specific values, but it wasn't clear to me how to combine this with `.startsWith()` so I ended up doing it the way I did.

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -348,6 +358,17 @@ class CmdRun extends CmdBase implements HubOptions {
...
     }
 
+    protected Collection checkNxfEnv(ConfigMap config) {
+        // Warn about setting NXF_ environment variables within env config scope
+        final nxfEnvs = config.env.findAll { it ->
+            it.toString().startsWith('NXF_') &&
+                    ! it.toString().startsWith('NXF_DEBUG')
+        }
+        nxfEnvs.collect {
+            it.toString().split('=')[0]
```
jorgeaguileraseqera, line 368: nxfEnvs is a map, probably you can return something as

```
nxfEnvs.collect{ it.key }
```

mribeirodantas: It should, but for some reason it does not 😕.

nxfEnvs is an ArrayList. When I run this
`config.env.findAll { log.warn it.key as String }`
I get `[Static type checking] - No such property: key for class: java.lang.Object`, even though `config.env.getClass()` returns `configMap`.

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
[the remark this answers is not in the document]
mribeirodantas: With help from @jorgeaguileraseqera it all makes sense now 😄

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
mribeirodantas: Now it all makes sense, Jorge! Thanks for the help.

pditommaso:
Ok, the approach used was correct to test it, but after looking more into it I decided to use a trick to capture the warning and keep the check simpler. 

Have a look 👉 009ec256. Merge manually.

PR #3093, opened 2022-08-03
Set aws batch polling monitor capacity from executor queueSize

### LINKED ISSUE #3076: Make `queueSize` work normally on AWS Batch
The `executor.queueSize` config option limits the number of concurrent tasks for a workflow run. That is, it works like this for every executor except for AWS Batch, which has its own option called "queue size" that means something else.

We can easily create a separate config option to map to the AWS-specific setting, but the greater challenge has to do with how the `awsbatch` executor manages jobs. Basically, the executor submits and polls jobs in batches, in order to minimize the rate of API calls. This "batch context" operates like a cache, which makes some operations more complicated. The `queueSize` config option influences whether an individual job will be submitted or delayed, but when jobs are submitted in batches, the logic will look more like "how much of this batch can be submitted at once".

I think to implement `queueSize` correctly, there will be a tradeoff between bandwidth and latency. You can submit larger batches less frequently, or smaller batches more frequently. This would be controlled by an AWS parameter e.g. `aws.batch.submitBatchSize`. For example, if the batch size is 50, Nextflow will wait for 50 jobs to be generated before submitting them. If the workflow generates fewer than 50 jobs, Nextflow needs some way to know that no more tasks are coming and submit a partial batch.

bentsherman:
Closes #3076

### DIFF
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy
@@ -185,12 +185,14 @@ class AwsBatchExecutor extends Executor implements ExtensionPoint {
 
         final pollInterval = session.getPollInterval(name, Duration.of('10 sec'))
         final dumpInterval = session.getMonitorDumpInterval(name)
+        final capacity = session.getQueueSize(name, 1000)
 
         final def params = [
                 name: name,
                 session: session,
                 pollInterval: pollInterval,
-                dumpInterval: dumpInterval
+                dumpInterval: dumpInterval,
+                capacity: capacity
         ]
 
         log.debug "Creating parallel monitor for executor '$name' > pollInterval=$pollInterval; dumpInterval=$dumpInterval"
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy
@@ -219,7 +219,7 @@ class AwsBatchExecutor extends Executor implements ExtensionPoint {
      */
     private ThrottlingExecutor createExecutorService(String name) {
 
-        final qs = session.getQueueSize(name, 5_000)
+        final qs = session.config.navigate("threadPool.${name}.maxQueueSize",5_000) as Integer
         final limit = session.getExecConfigProp(name,'submitRateLimit','50/s') as String
         final size = Runtime.runtime.availableProcessors() * 5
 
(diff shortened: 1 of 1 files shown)


pditommaso:
Nice. It's needed to not use it anymore for the thread pool size, here 

https://github.com/nextflow-io/nextflow/blob/wave-conda/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy#L220-L220

and use instead another setting for (I remember, another config context was used for this thread pool, but cannot find it where is defined)

bentsherman:
I was unsure what to do about that code. Two such thread pools are created here, a submitter and a reaper:

https://github.com/nextflow-io/nextflow/blob/f6c968d6ae99e6b49123b6715807cf6762659432/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy#L182-L184

And the session queue size getter is here:

https://github.com/nextflow-io/nextflow/blob/f6c968d6ae99e6b49123b6715807cf6762659432/modules/nextflow/src/main/groovy/nextflow/Session.groovy#L1279-L1281

So I think this queue size would be `AWSBatch-executor.queueSize` or something like that. So it seemed like the `executor.queueSize` wasn't used in the first place.

pditommaso:
Hold on, I've found it. In the executor setting there's a `submitter` context to configure those pool

https://github.com/nextflow-io/nextflow/blob/wave-conda/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy#L242-L242

and there's already the `queueSize` option. Theforefore it's enough to use 5_000 as default here 

https://github.com/nextflow-io/nextflow/blob/wave-conda/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy#L220-L220

and it can be overridden via the submitter

bentsherman:
Whoops, well I changed it so that now it uses:
- `threadPool.'AWSBatch-executor'.maxQueueSize` for the submitter
- `threadPool.'AWSBatch-reaper'.maxQueueSize` for the reaper

Similar to the FileTransfer thread pool.

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy
```diff
@@ -217,7 +219,7 @@ class AwsBatchExecutor extends Executor implements ExtensionPoint {
      */
     private ThrottlingExecutor createExecutorService(String name) {
 
-        final qs = session.getQueueSize(name, 5_000)
+        final qs = session.config.navigate("threadPool.${name}.maxQueueSize",5_000) as Integer
```
bentsherman, line 222: So you want to just do `final qs = 5_000` and let it be overwritten by the submitter options?

pditommaso (approved):

PR #3103, opened 2022-08-05
Use quiet output mode for hyperqueue executor

bentsherman:
After doing some digging I found that hyperqueue does have a concept of "job state" vs task state, but it doesn't include the job state in the JSON output for `hq job list`. However, the quiet output mode simply returns the id and state of each job. Additionally, submitting a job with quiet output mode simply returns the job id or an error if the submit failed.

Therefore, I changed the executor to use quiet mode for submitting jobs and checking job status. The parsing logic is much simpler and should scale much better because the hq commands now return the minimum required information.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/HyperQueueExecutor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/HyperQueueExecutor.groovy
@@ -19,7 +19,6 @@ package nextflow.executor
 
 import java.nio.file.Path
 
-import groovy.json.JsonSlurper
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
 import nextflow.processor.TaskRun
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/executor/HyperQueueExecutor.groovy
```diff
@@ -77,79 +82,51 @@ class HyperQueueExecutor extends AbstractGridExecutor {
...
-        return TupleHelper.listOf('hq', '--output-mode=json','submit', '--directives=file', scriptFile.getName())
+        return TupleHelper.listOf('hq', '--output-mode=quiet', 'submit', '--directives=file', scriptFile.getName())
     }
 
     @Override
     def parseJobId(String text) {
-        Map result = null
         try {
-            result = parser.parseText(text) as Map
+            text as Integer as String
```
pditommaso, line 91: Creating parsing!

pditommaso (approved):

PR #3116, opened 2022-08-12
Charliecloud support v3

phue:
As of Charliecloud `v0.28`, support in Nextflow stopped working because the expected CLI arguments for the pull command changed in that release. dd69f62c648cf1126225bbdf7e2e08004e8f4f94 addresses that, raising the required Charliecloud version to >=0.28.

Bonus:
Charliecloud has recently gained some nice additions and default settings. This allows for substantial cleanup of the code that
builds the bash wrapper, making some of my earlier workarounds obsolete. See 799e9c33235a1df8cae83fe769dc79b56cac890b for details.

### DIFF
--- a/docs/container.rst
+++ b/docs/container.rst
@@ -21,7 +21,7 @@ suited for use in HPC environments. Its main advantage is that it can be used wi
 making use of user namespaces in the Linux kernel. Charliecloud is able to pull from Docker registries.
 
 .. note::
-    This feature requires Nextflow version ``21.03.0-edge`` or later and Charliecloud ``v0.22`` or later.
+    This feature requires Nextflow version ``21.03.0-edge`` or later.
 
 .. warning::
     This feature is experimental. Using it in a production environment is not recommended.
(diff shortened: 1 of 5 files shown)


pditommaso:
This is nice, but what about potential users that may be using old versions?

phue:
In my opinion, users should be encouraged to upgrade, Charliecloud itself has stabilised quite a bit over the last couple of releases.
It's also straightforward to update/install, doesn't require admin privileges whatsoever.

I think keeping legacy support would create unwarranted overhead, the new implementation is much cleaner and more future proof.

pditommaso:
I tend to agree, but it's always bad to break changes. Though looking at the code, the only way to keep the compatibility with the old version would be have a new class implementing the support for 0.28 and use it as the default implementation. Not sure it's worth it.

phue:
Atleast the documentation states that Charliecloud support is experimental, so that can serve as an excuse for a breaking change :sweat_smile:

phue:
I have added the information about Charliecloud version requirements for different Nextflow releases :+1:

pditommaso (approved):
Nice

PR #3123, opened 2022-08-16
Document `files()` method

### LINKED ISSUE #3109: Add `glob()` method for globs in favor of `file()`
Spawned from the language improvements mega-discussion (#3107). The `file()` method can be used to load a single file or multiple files using a glob string, which means this function can return a single path or a list depending on the input. This is bad. See #1236 and #2425 for similar problems.

Instead, there should be a separate function for globs, e.g. `glob()`, and the glob support via `file()` should be deprecated.

bentsherman:
Closes #3109 

Adds a note about the `files()` method. Just a minimal change for now as this content will probably be overhauled anyway in the new docs.

### DIFF
--- a/docs/script.rst
+++ b/docs/script.rst
@@ -456,6 +456,9 @@ Remove the first number with its trailing whitespace from a string::
 Files and I/O
 ==============
 
+Opening files
+-------------
+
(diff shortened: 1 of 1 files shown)


pditommaso:
Nice

PR #3132, opened 2022-08-22
Allow identity based authentication on Azure Batch

abhi18av:
This PR addresses the credential-less authentication mechanism against Azure Batch and related services.


plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -252,16 +253,32 @@ class AzBatchService implements Closeable {
...
+        } else {
+
+            final batchEndpoint = config.batch().batchEndpoint ?: "https://batch.core.windows.net/";
+            final authenticationEndpoint =  config.batch().authenticationEndpoint ?: "https://login.microsoftonline.com/";
+            final batchServiceEndpoint = config.batch().batchEndpoint
+            final tenantId = config.batch().tenantId
+            final clientId = config.batch().clientId
+            final clientSecret = config.batch().clientSecret
+            cred = new BatchApplicationTokenCredentials(batchServiceEndpoint, clientId, clientSecret, tenantId, batchEndpoint, authenticationEndpoint)
+        }
```
pditommaso, line 278: Ohh, that's it?

abhi18av: On the code side yeah, the changes won't be a lot, but a lot of time was spent on 
1. Getting the batch/pool level configs right for testing
2. Due to Azure quota issues (batch accounts), I can only iterate on a single account 

All this lead to a rabbit hole for terraform templates to bootstrap the testing env.


But not done yet sorry - the compiled code needs testing and improvements :sweat_smile:

pditommaso: Cool, well done

abhi18av:
@pditommaso , I'd appreciate your input on design decisions here.

I have been studying the authentication mechanisms which we can provide as an alternative to the use of storage/batch account keys, and I zeroed in on the service principal based setup, which essentially 

1. Requires the creation of a service principal 
2. Relies on authentication creds (client-id, client-secret, tenant-id) 
3. Can easily be integrated into the current codebase  
4. Minimal changes to an NF-generated pool

However, the downside is still that the authentication creds still would need to be available within the task environment for `azcopy` to login as a service principal via `azcopy --service-principal -application-id $CLIENT_ID --tenant $TENANT_ID` with `AZCOPY_SPA_CLIENT_SECRET` available in the env.

Therefore, the effective change via this mechanism is that we would authenticate the per-node/task `azcopy` CLI for uploads, without storage keys/SAS token but we'd still rely upon the authentication creds (client-id, client-secret, tenant-id) to be made available for every `azcopy` invocation. 

-----------------------

However, I think the central piece which needs to be addressed is basically authenticating `azcopy` running per-node/task in the batch service and [there are basically three modes of authentication](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-authorize-azure-active-directory) which we can explore 
1. Service Principal
2. User-assigned Identity
3. Managed Identity

The user-assigned identity still requires certain metadata to be made available to an Azure Resource such as `client ID, Object ID, or Resource ID `.

I'm currently working on the [**Managed Identity**](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview) and evaluating any corner cases

- Which would only require us to do an `azcopy login --managed` within `.command.run` 
- Requires certain properties to be enabled in the Batch account (such as `Identity = SystemAssigned` and `Storage account authentication = "BatchAccountManagedIdentity"`

Managed Identity can be used with [all major services](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/managed-identities-status) which makes sense in Nextflow (and Tower) context such as 
- Monitoring
- Logging
- Managed Disk
- Azure ARC Kubernetes

---------------


Despite the expectations with the Batch account setup, this direction via Managed Identity seems security friendly - what do you think?

pditommaso:
> However, the downside is still that the authentication creds still would need to be available within the task environment for azcopy to login as a service principal via azcopy --service-principal -application-id $CLIENT_ID --tenant $TENANT_ID with AZCOPY_SPA_CLIENT_SECRET available in the env.

Ouch this is ugly. The problem is also the credentials for the Blob storage on Nextflow side which now requires the account key or SAS token 

https://github.com/nextflow-io/nextflow/blob/6f087fecade0293978af096c465dce9ea1e75f84/plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzFileSystemProvider.groovy#L202-L203

I think the best way to go for this would be to generate a SAS token when the pipeline is launched. This should also solve the problem for the `azcopy` copy tool. have you taken into consideration this possibility?

abhi18av:
> I think the best way to go for this would be to generate a SAS token when the pipeline is launched. This should also solve the problem for the azcopy copy tool. have you taken into consideration this possibility?

Paolo, this is another possibility yes. The tradeoff would be that we get in the business of managing SAS tokens during a pipeline run cycle and the mechanism would not be suitable in case the SAS tokens are not enabled (which is a recommended best practice), therefore we might inevitably run into the problem of using credential-less validation.

I've had good success with the managed identity (credential less) solution which relies on the fact that any VM in the Batch node is an Azure resource and can natively authenticate with the Azure AD (if configured as such) - I feel I'm close  🤞 😊

abhi18av:
Bingo!


```
abhinav@0239xxxxxxx2684bb1xxxxxxxx000:/mnt/batch/tasks/shared/bin$ ./azcopy login --identity
INFO: azcopy: A newer version 10.16.0 is available to download

INFO: Login with identity succeeded.
INFO: azcopy: A newer version 10.16.0 is available to download
```


And this is all we'd need to do in the `.command.run` script to allow any interaction with the Blob storage 🤝

pditommaso:
What about Blob storage client run by nextflow itself?

abhi18av:
> What about Blob storage client run by nextflow itself?

Do you mean the head-job? If so, then we need to accommodate this library https://github.com/AzureAD/microsoft-authentication-library-for-java within the codebase to allow the head job to source the authentication from the VM it's running upon - I'm evaluating this now for our use cases.

pditommaso:
Yes, but it looks to me that it requires too much customization that's hard to test. 

Above all, once Batch is authenticated, nextflow is able to automatically create a SAS token for the Blob storage. 

https://github.com/nextflow-io/nextflow/blob/27636f9d27c56d286e1e6173e66d97c971c5d6d3/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchExecutor.groovy#L97-L98

Therefore keep the blob storage auth on the SAS token and focus on finalise the Batch authentication

pditommaso:
Ouch that was not accurate, the sas token is created by using the storage client 

https://github.com/nextflow-io/nextflow/blob/27636f9d27c56d286e1e6173e66d97c971c5d6d3/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzHelper.groovy#L64-L64

that's create with the storage keys :/

https://github.com/nextflow-io/nextflow/blob/27636f9d27c56d286e1e6173e66d97c971c5d6d3/plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzFileSystemProvider.groovy#L205-L208

abhi18av:
@pditommaso , yes that's why I'm trying find a user-friendly and credential-less auth way to allow users to run Nextflow 

1. Entirely within the Azure infra (by relying upon Identity) - this path is straightforward now.
2. Hybrid mode, with the head job not running on Azure infra ↔️  this is where I'm iterating a bit

abhi18av:
@pditommaso @jorgeaguileraseqera , this PR is ready to be reviewed from my side - I'd appreciate any feedback please 🙏 

- **Service principal** should have the following permissions 
    - `Contributor`
    - `Storage Blob Data Contributor`
    - `Storage Blob Data Owner`

- **Config level changes** 
    - `azure.activeDirectory` config 
    - storage and batch `accountKey` are no longer required
    
```
workDir = "az://rnaseq-nf/rnaseq-nf-workdir"

process {
    executor = 'azurebatch'
    container = 'quay.io/nextflow/rnaseq-nf:latest'
}

params {
    outdir = "az://rnaseq-nf/rnaseq-nf-publishdir"
}

process.executor = 'azurebatch'
process.queue =  $AZURE_BATCH_POOL_NAME"


azure {

    activeDirectory {
        servicePrincipalId = "bXXXbXXX-6XXc-X2xx-8xx4-f02e6xxxxxc" // OR "$AZURE_CLIENT_ID"
        servicePrincipalSecret = "BXXQ~UuxxxO-kXXXXX1yjG5agXXXXXLweXXXX" // OR "$AZURE_CLIENT_SECRET"
        tenantId = "XX51xxxx-XXXX-4XX8-XXXX-1XXXXXXeb" // OR "$AZURE_TENANT_ID"
    }

    storage {
        accountName = "$AZURE_STORAGE_ACCOUNT_NAME"
    }

    batch {
        location = "westeurope"
        accountName = "$AZURE_BATCH_ACOUNT_NAME"

        azure.batch.deleteJobsOnCompletion = false

        pools {
             "$AZURE_BATCH_POOL_NAME" {
                vmType = 'STANDARD_D4_V3'
                privileged = false
            }
        }
    }
}

```


- **Changes to log entries** Logs regarding creation of Blob and Batch clients using Azure Active Directory Identity 

```
...
Sept-14 07:52:53.852 [main] DEBUG n.c.azure.nio.AzFileSystemProvider - Creating Azure Blob storage client using Service Principal credentials

...
Sept-14 07:53:03.274 [Thread-5] INFO  c.a.identity.ClientSecretCredential - Azure Identity => getToken() result for scopes [https://storage.azure.com/.default]: SUCCESS
Sept-14 07:53:03.276 [Thread-5] INFO  c.a.c.i.AccessTokenCache - Acquired a new access token.
...
Sept-14 07:53:13.175 [Task submitter] DEBUG n.cloud.azure.batch.AzBatchService - [AZURE BATCH] Creating Azure Batch client using Service Principal credentials


```

## Tested scenarios

**Case-1**: Same blob container for work/results ✅ 


**Case-2**: Different blob container for work/results ✅ 





## Improvements (TBD)

- Disable logging from the underlying library

```
Sept-14 07:53:14.694 [reactor-http-kqueue-3] WARN  c.a.c.u.s.SerializerEncoding - 'Content-Type' not found. Returning default encoding: JSON

```

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchExecutor.groovy
```diff
@@ -94,8 +94,14 @@ class AzBatchExecutor extends Executor implements ExtensionPoint {
...
-        if( !config.storage().sasToken )
+
+        // Generate an account SAS token using activeDirectory configs
+        if (!config.storage().sasToken && config.activeDirectory().isConfigured()) {
+            config.storage().sasToken = AzHelper.generateContainerSas(workDir, config.storage().tokenDuration)
+        }
+
+        // Generate an account SAS token using storage account keys
+        if (!config.storage().sasToken && !config.activeDirectory().isConfigured())
             config.storage().sasToken = AzHelper.generateAccountSas(workDir, config.storage().tokenDuration)
```
bentsherman, line 105: Consider using a ternary expression here so that you only check each condition once.

bentsherman:
I don't know these APIs very well but I didn't see any issues with the design.

bentsherman (approved):

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchExecutor.groovy
abhi18av: Yup, good point, accommodated the feedback with https://github.com/nextflow-io/nextflow/pull/3132/commits/930cb3b05a01daf493bd390e83071b61f73a2a01

abhi18av:
@pditommaso , I've added the doc entries in the PR now.

abhi18av:
@pditommaso , shall we take this merge forward?

docs/azure.rst
```diff
@@ -345,6 +345,42 @@ Public images from other registries are still pulled (if requested by a Task) wh
   specified via the :ref:`container <process-container>` directive using the format: ``[server]/[your-organization]/[your-image]:[tag]``.
   Read more about image fully qualified image names in the `Docker documentation <https://docs.docker.com/engine/reference/commandline/pull/#pull-from-a-different-registry>`_.
 
+Active Directory Authentication
+===============================
+
+As of version ``22.09.7-edge``, `Service Principal <https://learn.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal>`_ credentials can optionally be used instead of Shared Keys for Azure Batch and Storage accounts.
```
pditommaso, line 351:
```suggestion
As of version ``22.11.0-edge``, `Service Principal <https://learn.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal>`_ credentials can optionally be used instead of Shared Keys for Azure Batch and Storage accounts. 
```

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzActiveDirectoryOpts.groovy
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2021, Microsoft Corp
```
pditommaso, line 2:
```suggestion
```

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzActiveDirectoryOpts.groovy
```diff
@@ -0,0 +1,61 @@
...
+
+    AzActiveDirectoryOpts(Map config, Map<String, String> env = null) {
+        assert config != null
+        this.sysEnv = env == null ? new HashMap<String, String>(System.getenv()) : env
+        this.servicePrincipalId = config.servicePrincipalId ?: sysEnv.get('AZURE_CLIENT_ID')
+        this.servicePrincipalSecret = config.servicePrincipalSecret ?: sysEnv.get('AZURE_CLIENT_SECRET')
+        this.tenantId = config.tenantId ?: sysEnv.get('AZURE_TENANT_ID')
+    }
+
+
```
pditommaso, line 44:
```suggestion
```

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzStorageOpts.groovy
```diff
@@ -29,27 +29,27 @@ import nextflow.util.Duration
 @CompileStatic
 class AzStorageOpts {
 
-    private Map<String,String> sysEnv
+    private Map<String, String> sysEnv
```
pditommaso, line 32:
```suggestion
    private Map<String,String> sysEnv
```

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzStorageOpts.groovy
```diff
@@ -29,27 +29,27 @@ import nextflow.util.Duration
...
     String accountKey
     String accountName
     String sasToken
     Duration tokenDuration
-    Map<String,AzFileShareOpts> fileShares
+    Map<String, AzFileShareOpts> fileShares
```
pditommaso, line 37:
```suggestion
    Map<String,AzFileShareOpts> fileShares
```

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzStorageOpts.groovy
```diff
@@ -29,27 +29,27 @@ import nextflow.util.Duration
...
-        this.fileShares = parseFileShares(config.fileShares instanceof Map ? config.fileShares as Map<String,Map>
-                : Collections.<String,Map>emptyMap())
+        this.tokenDuration = (config.tokenDuration as Duration) ?: Duration.of('48h')
+        this.fileShares = parseFileShares(config.fileShares instanceof Map ? config.fileShares as Map<String, Map>
+                : Collections.<String, Map> emptyMap())
 
     }
 
-    Map<String,Object> getEnv() {
+    Map<String, Object> getEnv() {
```
pditommaso, line 52:
```suggestion
    Map<String,Object> getEnv() {
```

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzStorageOpts.groovy
```diff
@@ -29,27 +29,27 @@ import nextflow.util.Duration
...
+        this.sysEnv = env == null ? new HashMap<String, String>(System.getenv()) : env
         this.accountKey = config.accountKey ?: sysEnv.get('AZURE_STORAGE_ACCOUNT_KEY')
         this.accountName = config.accountName ?: sysEnv.get('AZURE_STORAGE_ACCOUNT_NAME')
         this.sasToken = config.sasToken
-        this.tokenDuration = (config.tokenDuration as Duration) ?: Duration.of('12h')
-        this.fileShares = parseFileShares(config.fileShares instanceof Map ? config.fileShares as Map<String,Map>
-                : Collections.<String,Map>emptyMap())
+        this.tokenDuration = (config.tokenDuration as Duration) ?: Duration.of('48h')
+        this.fileShares = parseFileShares(config.fileShares instanceof Map ? config.fileShares as Map<String, Map>
+                : Collections.<String, Map> emptyMap())
```
pditommaso, line 48:
```suggestion
                : Collections.<String,Map> emptyMap())
```

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzStorageOpts.groovy
```diff
@@ -29,27 +29,27 @@ import nextflow.util.Duration
...
+    AzStorageOpts(Map config, Map<String, String> env = null) {
+        assert config != null
+        this.sysEnv = env == null ? new HashMap<String, String>(System.getenv()) : env
         this.accountKey = config.accountKey ?: sysEnv.get('AZURE_STORAGE_ACCOUNT_KEY')
         this.accountName = config.accountName ?: sysEnv.get('AZURE_STORAGE_ACCOUNT_NAME')
         this.sasToken = config.sasToken
-        this.tokenDuration = (config.tokenDuration as Duration) ?: Duration.of('12h')
-        this.fileShares = parseFileShares(config.fileShares instanceof Map ? config.fileShares as Map<String,Map>
-                : Collections.<String,Map>emptyMap())
+        this.tokenDuration = (config.tokenDuration as Duration) ?: Duration.of('48h')
```
pditommaso, line 46: Why it was increased to 2 days?

plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzFileSystemProvider.groovy
```diff
@@ -58,10 +60,14 @@ class AzFileSystemProvider extends FileSystemProvider {
...
+    public static final String AZURE_CLIENT_ID = 'AZURE_CLIENT_ID'
+    public static final String AZURE_CLIENT_SECRET = 'AZURE_CLIENT_SECRET'
+    public static final String AZURE_TENANT_ID = 'AZURE_TENANT_ID'
+
     public static final String SCHEME = 'az'
 
-    private Map<String,String> env = new HashMap<>(System.getenv())
-    private Map<String,AzFileSystem> fileSystems = [:]
+    private Map<String, String> env = new HashMap<>(System.getenv())
+    private Map<String, AzFileSystem> fileSystems = [:]
```
pditommaso, line 70:
```suggestion
    private Map<String,String> env = new HashMap<>(System.getenv())
    private Map<String,AzFileSystem> fileSystems = [:]
```

pditommaso (changes requested):
Very minor comments

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzStorageOpts.groovy
abhi18av: This was done because in case the workflow goes beyond the previous default limit of `12h`, it is abruptly stopped and it affected some commercial users therefore I've increased it slightly.

pditommaso: Fair enough

docs/azure.rst
```diff
@@ -345,6 +345,42 @@ Public images from other registries are still pulled (if requested by a Task) wh
   specified via the :ref:`container <process-container>` directive using the format: ``[server]/[your-organization]/[your-image]:[tag]``.
   Read more about image fully qualified image names in the `Docker documentation <https://docs.docker.com/engine/reference/commandline/pull/#pull-from-a-different-registry>`_.
 
+Active Directory Authentication
+===============================
+
+As of version ``22.11.0-edge``, `Service Principal <https://learn.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal>`_ credentials can optionally be used instead of Shared Keys for Azure Batch and Storage accounts.
```
pditommaso, line 351: Would not be better to call it "Azure Active Directory service principal" ?

abhi18av: Paolo, the plan is to accommodate other identity within this scope. For example, managed identity and user identity is a type of service principals (with slightly different behavior) therefore it makes sense to use the naming convention to highlight this. 

Therefore if we continue with current design, these future options would be like 

```groovy
azure {
  // common attributes
  tenantId = '<YOUR TENANT ID>'
  subscriptionId = '<YOUR SUBSCRIPTION ID>'

  activeDirectory {
  
      // If vanilla service principal is used 
      servicePrincipalId = '<YOUR SERVICE PRINCIPAL CLIENT ID>'
      servicePrincipalSecret = '<YOUR SERVICE PRINCIPAL CLIENT SECRET>'
  
      // If managed-identity is used 
      useManagedIdentity = true
     
      // If user-identity is used 
      userAssignedIdentityClientID = <YOUR USER ASSIGNED IDENTITY CLIENT IDENTIFIER>
  
      
    }
}

```

However, if we are okay with another level of nesting then we can do something like 

```groovy
azure {

  // common attributes
  tenantId = '<YOUR TENANT ID>'
  subscriptionId = '<YOUR SUBSCRIPTION ID>'

  activeDirectory { 

      // If vanilla service principal is used 
      servicePrincipal {
            id = '<YOUR SERVICE PRINCIPAL CLIENT ID>'
            secret = '<YOUR SERVICE PRINCIPAL CLIENT SECRET>'
           }
  
      // If managed identity is used 
      managedIdentity { 
           enabled = true
          }
   
      // If user identity is used 
      userIdentity { 
           clientId =  <YOUR USER ASSIGNED IDENTITY CLIENT IDENTIFIER>
         }
  }

}
```

Happy to accommodate whatever seems better and more intuitive.

Also, perhaps we should add the note for `<EXPERIMENTAL SUPPORT>` till we stabilize the entire scope and achieve the overall goal of credential-less auth within Azure?

Reference: https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview#managed-identity-types

pditommaso: Ok, I think it's fine 👍

pditommaso (approved):

pditommaso:
Let's merge this 🚀

abhi18av:
<div><br/>via <a href="https://giphy.com/justin/">Justin</a> on <a href="https://giphy.com/gifs/friday-friyay-1APaqOO5JHnWKLc7Bi">GIPHY</a></div>

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchExecutor.groovy
```diff
@@ -94,11 +94,15 @@ class AzBatchExecutor extends Executor implements ExtensionPoint {
...
         batchService = new AzBatchService(this)
-        // generate an account SAS token if missing
-        if( !config.storage().sasToken )
-            config.storage().sasToken = AzHelper.generateAccountSas(workDir, config.storage().tokenDuration)
 
-        Global.onCleanup((it)->batchService.close())
+        // Generate an account SAS token using either activeDirectory configs or storage account keys
+        if (!config.storage().sasToken) {
+            config.storage().sasToken = config.activeDirectory().isConfigured()
+                    ? AzHelper.generateContainerSasWithActiveDirectory(workDir, config.storage().tokenDuration)
```
Skitionek, line 101: This one might reintroducing, this issue:
https://github.com/nextflow-io/nextflow/pull/2576 ( https://github.com/nextflow-io/nextflow/issues/2563 )
just for managed identity

Issue #2668, opened 2022-02-22
Supporting Ignite and using Ceph S3 as working directory


pbelmann:
Dear nextflow developers

As stated in issue (https://github.com/nextflow-io/nextflow/issues/2564) opened by @cnexcale, we are interested in using nextflow in combination with the nextflow feature that allows us to use OpenStack S3 (backed by Ceph) for the working directory. After trying out multiple executors, we figured out that a patched version of the ignite executor is the solution for our use case (using HEAD at commit 2188d51a0f4866ad4249cc5887080390f80cee85): 

* [prevent “missing plugin” exception when creating S3FileSystem during workdir initialization](https://github.com/cnexcale/nextflow/commit/a8010768ee5549a08a58e3a4a6ce92f44b165bac)
* [nf-amazon: prevent sending empty x-amz-tagging headers in putObject requests](https://github.com/cnexcale/nextflow/commit/bc430b20f8c55b3f22ef5260802b6aec9ca76793)
* [nf-ignite: debatable patch: load session config in Ignite BaseTask before deserializing Task payload](https://github.com/cnexcale/nextflow/commit/e3a8e8046be8092ef6ab489d83ba425fc05ecac7)
* [nf-ignite: fixed missing log instance due to shadowing in derived class](https://github.com/cnexcale/nextflow/commit/788da5899b6be0b4d685f5ea2dcd78ea6ef2b1f1)
* [nf-ignite: added .command.env to unstaging scripts after process exec](https://github.com/cnexcale/nextflow/commit/c02e191e59a9baa01c31e4d12b58d41d16f54c0c)

Using ignite might also be a way for us to run Nextflow pipelines on a hybrid cloud setup.
However, we noticed that with commit [d4dc1cfe13fec4bd3db5736fc494fe4fedd5e60c](https://github.com/nextflow-io/nextflow/commit/d4dc1cfe13fec4bd3db5736fc494fe4fedd5e60c) the ignite executor seems to be deprecated and we wonder what are the reasons for not supporting it anymore. Is it a technical reason that makes it hard to continue the support or is it that the nextflow community is not using it? We might be interested in supporting it ourselves (at least we would try it).

We are aware that the fix of storing the Session in config in an IgBaseTask object is way too hacky for a “real” fix. It shouldn’t be a tasks concern to manage configs but in some way or another the AWS credentials and configuration must be present on/published to the worker deamons to correctly initialize an S3FileSytem in order to handle S3 interactions.
[...]

PR #3136, opened 2022-08-22
enabled support for nf-ignite again

cnexcale:
Hi,

as discussed and approved in the issue #2668 we would like be be able to use the nf-ignite plugin with current and future Nextflow builds.

These changes are the bare minimum to be able to start an Ignite daemon node via `nextflow node`

Kind regards
Lukas

### DIFF
--- a/modules/nextflow/src/main/resources/META-INF/plugins-info.txt
+++ b/modules/nextflow/src/main/resources/META-INF/plugins-info.txt
@@ -6,3 +6,4 @@ nf-azure@0.13.4
 nf-tower@1.5.1
 nf-ga4gh@1.0.3
 nf-codecommit@0.1.2
+nf-ignite@1.2.3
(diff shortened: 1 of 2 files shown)


pditommaso:
Those changes are not strictly required to make the nf-ignite plugin work, and it was removed because it's not a core plugin anymore.

The only difference is that the user is required to enable the plugin either via the `-plugins` CLI option or the config file or the `NXF_PLUGINS_DEFAULT`env variable

cnexcale:
`nextflow -plugins nf-ignite@1.2.3 node -bg` or `nextflow node -bg -plugins nf-ignite@1.2.3` (Unknown option: -plugins) or setting the `plugins {}` section in the config did not work .

But `export NXF_PLUGINS_DEFAULT=nf-ignite; ./nextflow node -bg` did the trick, thanks for the direction.

PR #3139, opened 2022-08-22
Add fusion support to local executor

bentsherman:
This PR extends the Fusion support recently added to AWS Batch to the local executor as well. Essentially it allows you to run a pipeline locally while using S3-compatible storage as the work directory. The only requirement is that the tasks must be containerized, so that they can mount the S3 storage as POSIX-like directories.

The current challenge is how to launch the task from Nextflow without a local task directory. I have found three possible approaches:
1. Run Nextflow in a container (e.g. `nextflow run -dockerized`) with the Fusion mounts so that it can access the work directory as normal
2. Use pipes to execute the run script and save the log without writing any local files
3. Move the container logic outside of the run script, so that the launch command itself is executed in the container.

This PR currently uses approach (2), but it isn't quite working. I'm getting this error:
```
Aug-22 10:46:40.668 [Task monitor] DEBUG n.processor.TaskPollingMonitor - Task completed > TaskHandler[id: 1; name: foo; status: COMPLETED; exit: 127; error: -; workDir: s3://nextflow-ci-dev/work/10/c7be191212d409d87ace3e06292acf]
Aug-22 10:46:40.922 [Task monitor] DEBUG nextflow.processor.TaskRun - Unable to dump output of process 'null' -- Cause: java.nio.file.NoSuchFileException: /tmp/temp-s3-7379655129903345232/.command.out
Aug-22 10:46:41.205 [Task monitor] DEBUG nextflow.processor.TaskRun - Unable to dump error of process 'null' -- Cause: java.nio.file.NoSuchFileException: /tmp/temp-s3-68076642700216686/.command.err
Aug-22 10:46:41.410 [Task monitor] DEBUG nextflow.processor.TaskRun - Unable to dump error of process 'null' -- Cause: java.nio.file.NoSuchFileException: /tmp/temp-s3-8956860549820136776/.command.log
Aug-22 10:46:41.415 [Task monitor] ERROR nextflow.processor.TaskProcessor - Error executing process > 'foo'
```

And then the process terminates with exit code 127. Also, this approach relies on `aws s3 cp`, so I don't know if it would work with other cloud providers.

Approach (1) probably requires the fewest code changes but also might require more work from the user.

Approach (3) is probably the most robust but also more complex, because the container builder logic must be extracted from the bash wrapper builder. Also, I don't think this approach will work for HPC because you can't submit jobs inside a container. Maybe we could add another entrypoint to `.command.run` that sets up the container but calls itself with the regular entrypoint?


modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy
```diff
@@ -135,8 +135,8 @@ class TaskPollingMonitor implements TaskMonitor {
         this.dumpInterval = (params.dumpInterval as Duration) ?: Duration.of('5min')
         this.capacity = (params.capacity ?: 0) as int
 
-        this.pendingQueue = new LinkedBlockingQueue()
-        this.runningQueue = capacity ? new ArrayBlockingQueue<TaskHandler>(capacity) : new LinkedBlockingQueue<TaskHandler>()
+        this.pendingQueue = new LinkedBlockingQueue<TaskHandler>()
+        this.runningQueue = new LinkedBlockingQueue<TaskHandler>()
```
pditommaso, line 139: it looks like there are some changes that are not related to this

pditommaso:
Let's try to update this branch with the master

modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy
bentsherman: Weird, not sure what happened, looks clean now

pditommaso:
Giving a try how to make it work

pditommaso:
Ok, there's some cleanup and unit tests to solves, but this works in my computer 😄

modules/nextflow/src/test/groovy/nextflow/config/ConfigBuilderTest.groovy
```diff
@@ -2229,5 +2236,37 @@ class ConfigBuilderTest extends Specification {
...
+        where:
+        ENGINE              | CLAZZ
+        'docker'            | DockerBuilder
+        'podman'            | PodmanBuilder
+        'singularity'       | SingularityBuilder
+        'shifter'           | ShifterBuilder
+        'charliecloud'      | CharliecloudBuilder
+        'udocker'           | UdockerBuilder
+
+    }
```
bentsherman, line 2259: Shouldn't this all be in the `ContainerBuilderTest` ?

pditommaso: Yes, you are right

pditommaso:
Merging this manually since it required some non-trivial changes. Notably: 

* Get rid of the remote bin directory for fusion-based tasks (the container will provide bin scripts) 
* Use of [FusionAwareTask](https://github.com/nextflow-io/nextflow/blob/17160bb0c667a8e498c4fedcbfb7344cae51c5bc/modules/nextflow/src/main/groovy/nextflow/executor/fusion/FusionAwareTask.groovy#L32-L32) trait to re-usume some common logic 
* Re-organise local executor and task handler in its own [package](https://github.com/nextflow-io/nextflow/blob/17160bb0c667a8e498c4fedcbfb7344cae51c5bc/modules/nextflow/src/main/groovy/nextflow/executor/local/LocalTaskHandler.groovy#L174-L174) 

See more here 17160bb0. This should simplify the implementation for the support for the grid executors

bentsherman:
Thanks Paolo, these changes look great. Should make the K8s support much easier to implement.

Thinking aloud about the grid scheduler support because it will be more challenging. The launch command for a task is to submit a job to the HPC scheduler, but then it also needs to be in a container. You could do something like:
```bash
qsub singularity exec /fusion/work/dir/.command.run
```

But then you have to move the submit options (resources, log file redirect, etc) to the submit/launch command. So it will work but the submit options won't be saved anywhere. I guess we could still save them in the wrapper script for posterity. With fusion, the wrapper script can't be run by itself anyway because of the fusion paths.

bentsherman:
Another caveat with the local executor is that native `exec` tasks might not be able to use the task directory. Nextflow should be able to handle S3 paths in a native task using s3fs, but need to test it.

pditommaso:
> Thinking aloud about the grid scheduler support because it will be more challenging. The launch command for a task is to submit a job to the HPC scheduler, but then it also needs to be in a container. You could do something like:
> 
>     qsub singularity exec /fusion/work/dir/.command.run
> 
> But then you have to move the submit options (resources, log file redirect, etc) to the submit/launch command


Yes, the above approach could be problematic. I think we need to continue the batch scheduler needs to continue to launch the wrapper script.  

However, thinking more is not that difficult. The job wrapper structure is quite flexible. By default, it runs the `nxf_main` function. See [here](https://github.com/nextflow-io/nextflow/blob/27636f9d27c56d286e1e6173e66d97c971c5d6d3/modules/nextflow/src/main/resources/nextflow/executor/command-run.txt#L162-L163) and [here](https://github.com/nextflow-io/nextflow/blob/27636f9d27c56d286e1e6173e66d97c971c5d6d3/modules/nextflow/src/main/resources/nextflow/executor/command-run.txt#L162-L163).

It should be simple to notify it to add a new function e.g. `nxf_fusion_launch` as shown below 

```
function nxf_fusion_launch() {
  docker run /some/work/dir/.command.run nxf_main
}
```

and change the entry as `NXF_ENTRY=${1:-nxf_fusion_launch}`, that's it! 


Actually, we should use this approach also for the local executor because it allows allows to re-execute a failing job as usual typing `bash s3://some/path/.command.run` (well bash cannot read directly from s3 but this is a details 😄 )

bentsherman:
I had this thought as well, just add another entrypoint to the run script. But you still need to do `qsub .command.run` and the run script is in S3.

You can pipe a script directly from s3 to bash. That is how I implemented this PR originally:
```bash
aws s3 cp s3://bucket/work/dir/.command.run - | bash
```

The rest of the logic for catching errors and writing the log file can be put in the run script, which as you said could be used by the local executor as well.

pditommaso:
damn, I was missing this. We are doing all this to NOT rely on the `aws` CLI tool. This means it's required to create a local copy of the launcher file to follow this approach

PR #3142, opened 2022-08-23
Add fusion support to K8s executor

bentsherman:
The K8s executor has a similar structure to AWS Batch, so the Fusion support is easier to manage. For each task, the head job creates a pod spec containing the launch command, with or without Fusion.

Of course, the head job still needs to interact with the work directory, so it needs AWS credentials. There are a few ways to do this, e.g. on EKS the cluster might already have access to S3 buckets in the same account. You can provide credentials to the pod as K8s secrets, but I suspect some users will want a more secure solution (see #2988). In any case, having an S3 work directory eliminates the requirement of a PVC, so I still need to work out the logic for that.

One other issue is that the K8sWrapperBuilder overrides `fixOwnership()` with some K8s-specific behavior, so using instead the FusionScriptLauncher breaks this behavior. I think it was a hack anyway, but we'll need to work out a proper solution to it so that the `fixOwnership` config option keeps the same behavior.


pditommaso:
This is a nice progress. I think before moving on, we need to find to share some common logic, e.g. the fusion launch command. 

Also I think we need to get rid of the logic for the remove bin to simplify the implementation. 

https://github.com/nextflow-io/nextflow/blob/6f087fecade0293978af096c465dce9ea1e75f84/modules/nextflow/src/main/groovy/nextflow/executor/fusion/FusionScriptLauncher.groovy#L87-L91

Those binary files should be included in the container by Wave, that's a requirement for Fusion

pditommaso:
> Of course, the head job still needs to interact with the work directory, so it needs AWS credentials

It should not be a problem, EKS pods can access IAM roles via the service account

bentsherman:
Refactored with the new upstream changes. So far so good. I removed the `K8sWrapperBuilder` and moved its logic into `K8sTaskHandler`. The task handler tests are failing, I think because some methods take `TaskRun task` as a parameter while others reference the class field. We should probably clean that up.

pditommaso:
> So far so good. I removed the K8sWrapperBuilder

Umm, was really necessary? I think there should not be any problem to use both of them

bentsherman:
The task handler would use either the K8sWrapperBuilder or the FusionScriptLauncher, so if you use fusion then the behavior of K8sWrapperBuilder is lost.

bentsherman:
@pditommaso Why is there both TaskRun and TaskBean ? They appear to hold the same information. These task handler classes would be easier to manage if we didn't have to think about converting a task run to a task bean at the right moment.

bentsherman:
Okay I found a way to make everything work with minimal extra refactoring. Just need to do an end-to-end test and then it will be ready.

pditommaso:
I see the point of your changes, there's space to improve the `FusionAwareTask` but I think it's better to focus this PR only on K8s. 

Moved `fixOwnership()` in the K8sTaskHandler. This makes it straightforward to fix the tests. 

I've tested this and works beautifully in my local k8s cluster :v:

PR #3146, opened 2022-08-24
Update Google Analytics to use GA4

ewels:


docs/conf.py
```diff
@@ -102,8 +102,7 @@
 # further.  For a list of options available for each theme, see the
 # documentation.
 html_theme_options = {
-  'display_version': False,
-  "analytics_id": "UA-364526-10"
+  'display_version': False
 }
```
pditommaso, line 106: Why just not updating the GA Id ?

ewels: Because the method of including the analytics also changed I think, not just the ID. Or at least, it did on the main website (see https://github.com/nextflow-io/website/commit/5ea9369c1970d00a7681f6f13f2cff84e399cb6b).

Ok, digging into the diffs for this PR.. 

```diff
20a21,22
>         <script async="async" src="https://www.googletagmanager.com/gtag/js?id=G-244N3GEN75"></script>
>         <script src="_static/google_analytics.js"></script>
388,398d389
<     <!-- Theme Analytics -->
<     <script async src="https://www.googletagmanager.com/gtag/js?id=UA-364526-10"></script>
<     <script>
<       window.dataLayer = window.dataLayer || [];
<       function gtag(){dataLayer.push(arguments);}
<       gtag('js', new Date());
< 
<       gtag('config', 'UA-364526-10', {
<           'anonymize_ip': false,
<       });
<     </script> 
```

The new `_static/google_analytics.js`:

```js
if( window.location.hostname == 'localhost' || window.location.hostname == '127.0.0.1' ) { throw new Error('Skip GA on localhost') };
window.dataLayer = window.dataLayer || [];
function gtag(){dataLayer.push(arguments);}
gtag('js', new Date());
gtag('config', 'G-244N3GEN75');
```

ok, yeah.. Maybe not so much. Just ignoring local IPs and removing `'anonymize_ip': false,`.

Fair enough! No need to overcomplicate things! 😅 I'll just change the ID then 👍🏻

ewels:
Closing in favour of https://github.com/nextflow-io/nextflow/pull/3147

I think that the other thing that threw me was the difference between Sphinx and Sphinx-rtd. I guess that the `analytics_id` thing is coming from the theme and not from Sphinx itself, hence the docs I found were a little misleading.

PR #3161, opened 2022-08-30
finalise secrets feature 

### LINKED ISSUE #3156: Finalise secrets feature
The secrets feature should be promoted to a stable feature so that it does not need to be explicitly activated to be used. 

This task requires the following changes: 

- [ ] enable the secrets by default unless the variable NXF_ENABLE_SECRETS is set to false. See [here](https://github.com/nextflow-io/nextflow/blob/27636f9d27c56d286e1e6173e66d97c971c5d6d3/modules/nextflow/src/main/groovy/nextflow/secret/SecretsLoader.groovy#L35-L35).
- [x] #3157
- [ ] Update the docs

jorgeaguileraseqera:
closes #3156


modules/nextflow/src/test/groovy/nextflow/cli/CmdSecretTest.groovy
```diff
@@ -54,10 +55,21 @@ class CmdSecretTest extends Specification {
         field.accessible = true
         def map = (Map<String, String>) field.get(System.getenv())
         map.put('NXF_SECRETS_FILE', secretFile.toString())
+        map.put('NXF_ENABLE_SECRETS', 'true')
```
pditommaso: Too hacking into System env inner implementation. I've added SysEnv wrapper to better handler this see [here](https://github.com/nextflow-io/nextflow/blob/bdfa0f6d3bf31ec8afebb1cd041e867ac27a05e3/modules/nf-commons/src/main/nextflow/SysEnv.groovy#L29-L29) and [here](https://github.com/nextflow-io/nextflow/blob/bdfa0f6d3bf31ec8afebb1cd041e867ac27a05e3/modules/nextflow/src/main/groovy/nextflow/secret/LocalSecretsProvider.groovy#L50-L50)


But not understanding why some tests are failing

jorgeaguileraseqera: Probably because it's also required in `SecretsLoader`

```
 static boolean isEnabled() {
        System.getenv('NXF_ENABLE_SECRETS')=='true'
    }
```

jorgeaguileraseqera: (by other hand, the hack to change `env` values it's "pure gold" , isn't it?)

pditommaso: Kind of :)

jorgeaguileraseqera: I think the error comes because we are using Memoized in `load` method and we need to refresh the cache between tests 

it seems with this commit is working now

https://github.com/nextflow-io/nextflow/commit/d4537712d8e83b2b7c5e75888723b138f0ddb9a7

pditommaso (approved):

PR #3163, opened 2022-08-30
feature: improve dump operator documentation

### LINKED ISSUE #3159: Add pretty print option to `dump`
## New feature

Hi,

Would it be possible to add a "pretty-print" option to the `dump` operator? This would greatly improve readability of large channel dumps

## Usage scenario 

Better readability when using nested maps in a channel

jorgeaguileraseqera:
closes #3159

### DIFF
--- a/docs/operator.rst
+++ b/docs/operator.rst
@@ -674,6 +674,56 @@ Then you will be able to specify the tag ``foo`` or ``bar`` as an argument of th
 either the content of the first or the second channel. Multiple tag names can be specified separating them with a ``,``
 character.
 
+If you want a customization output you can provide a Closure as an optional last argument where you return the string
+desided.
+
+For example, to pretty print a *Map* as a JSON output::
+
+    import static groovy.json.JsonOutput.toJson
+    import static groovy.json.JsonOutput.prettyPrint
+
+
+    workflow {
+        Channel.of(
+            [
+                id: 'test',
+                samples: [
+                    [ id: 'S1', path: '/path/to/s1' ],
+                    [ id: 'S2', path: '/path/to/s2' ],
+                    [ id: 'S3', path: '/path/to/s3' ]
+                ]
+            ]
+        ).dump(tag: 'map', {
+            def json = toJson(it)
+            prettyPrint( json )
+        })
+    }
+
+Will produce::
+
+    Launching `pretty.nf` [maniac_babbage] DSL2 - revision: 9c1c991204
+    [DUMP: map] {
+        "id": "test",
+        "samples": [
+            {
+                "id": "S1",
+                "path": "/path/to/s1"
+            },
+            {
+                "id": "S2",
+                "path": "/path/to/s2"
+            },
+            {
+                "id": "S3",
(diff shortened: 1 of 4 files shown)


bentsherman (approved):

docs/operator.rst
```diff
@@ -673,6 +673,52 @@ Then you will be able to specify the tag ``foo`` or ``bar`` as an argument of th
...
+                    [ id: 'S1', path: '/path/to/s1' ],
+                    [ id: 'S2', path: '/path/to/s2' ],
+                    [ id: 'S3', path: '/path/to/s3' ]
+                ]
+            ]
+        ).dump(tag: 'map', {
+            def json = toJson(it)
+            prettyPrint( json )
+        })
+    }
```
pditommaso, line 696: Even it's a nice trick, not sure we should promote this in the docs. It could make sense to add a pretty print option in the operator

jorgeaguileraseqera: The main "problem" is, if I'm not wrong, that the argument for the closure can be any type

It can be a Map, a String, a List ...

pditommaso: it should as simple as adding an option [here](https://github.com/nextflow-io/nextflow/blob/3c06a598e801611faabdbb61dfa3544b47657a88/modules/nextflow/src/main/groovy/nextflow/extension/DumpOp.groovy#L54-L54)

jorgeaguileraseqera: Yes, I know it, but what logic needs to implement in this method due to the argument can be any object?

Maybe we can implement a default prettyPrint similar to this: https://github.com/nextflow-io/nextflow/issues/3159#issuecomment-1232864270 but not sure if with this we cover all cases or it's very specific to this

pditommaso: What if we pass the target object to `JsonOutput.prettyPrint` ?

modules/nextflow/src/main/groovy/nextflow/extension/DumpOp.groovy
```diff
@@ -36,7 +37,7 @@ import static nextflow.util.CheckHelper.checkParams
 @CompileStatic
 class DumpOp {
 
-    static final private Map PARAMS_DUMP = [tag: String]
+    static final private Map PARAMS_DUMP = [tag: String, prettyPrint: Boolean ]
```
pditommaso, line 40:
```suggestion
    static final private Map PARAMS_DUMP = [tag: String, pretty: Boolean ]
```

Let's be more concise and just use `pretty`

Midnighter:
I think a better approach would be to use a [JsonGenerator](https://groovy-lang.org/json.html#_customizing_output) which for paths could look like:

```groovy
import static groovy.json.JsonOutput.prettyPrint
import groovy.json.JsonGenerator
import java.nio.file.Path

JsonGenerator generator = new JsonGenerator.Options()
        .addConverter(Path) { Path filename -> filename.toString() }
        .build()

prettyPrint(generator.toJson(it))
```

pditommaso:
Merge manually via 4218b299

PR #3167, opened 2022-08-31
Add before-afterScript warning regarding containers

Kibubu:
before and afterScript directives are not executed inside the specified process container by design.

This behavior is not intuitive and will be fixed with `prolog` and `epilog` directives once implemented #540

Added a warning for the time being

Signed-off-by: Kirill Bulert <bulert@infai.org>

### DIFF
--- a/docs/process.rst
+++ b/docs/process.rst
@@ -1283,8 +1283,7 @@ The ``afterScript`` directive allows you to execute a custom (Bash) snippet imme
 This may be useful to clean up your staging area.
 
 .. warning:: In combination with the :ref:`container directive <process-container>`, the ``afterScript`` will be 
-   executed outside the specified container. The ``afterScript`` is always executed in the environment running 
-   ``nextflow``.
+   executed outside the specified container. The ``afterScript`` is always executed in the host environment.
 
 
 .. _process-beforeScript:
@@ -1306,8 +1305,7 @@ For example::
     }
 
 .. warning:: In combination with the :ref:`container directive <process-container>`, the ``beforeScript`` will be 
-   executed outside the specified container. The ``beforeScript`` is always executed in the environment running 
-   ``nextflow``.
+   executed outside the specified container. The ``beforeScript`` is always executed in the host environment.
 
 
 .. _process-cache:
--- a/docs/process.rst
+++ b/docs/process.rst
@@ -1304,7 +1304,7 @@ For example::
       """
     }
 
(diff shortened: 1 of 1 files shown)


bentsherman:
One correction, the `beforeScript` and `afterScript` aren't run in the same environment as nextflow, they are run in the task environment but outside the container. "Host environment" would be more accurate.

docs/process.rst
```diff
@@ -1301,6 +1304,9 @@ For example::
       """
     }
 
+.. warning:: In combination with the :ref:`container directive <process-container>`, the ``beforeScript`` will be 
+   executed outside the specified container. The ``beforeScript`` is always executed in the host environment.
```
bentsherman, line 1308:
```suggestion
.. warning:: When combined with the :ref:`container directive <process-container>`, the ``beforeScript`` will be 
   executed outside the specified container. In other words, the ``beforeScript`` is always executed in the host environment.
```

bentsherman (approved):

docs/process.rst
```diff
@@ -1282,6 +1282,9 @@ afterScript
 The ``afterScript`` directive allows you to execute a custom (Bash) snippet immediately *after* the main process has run.
 This may be useful to clean up your staging area.
 
+.. warning:: When combined with the :ref:`container directive <process-container>`, the ``afterScript`` will be
```
pditommaso, line 1285:
```suggestion
.. note:: When combined with the :ref:`container directive <process-container>`, the ``afterScript`` will be 
```

docs/process.rst
```diff
@@ -1301,6 +1304,9 @@ For example::
       """
     }
 
+.. warning:: When combined with the :ref:`container directive <process-container>`, the ``beforeScript`` will be
```
pditommaso, line 1307:
```suggestion
.. note:: When combined with the :ref:`container directive <process-container>`, the ``beforeScript`` will be 
```

Better to use `note` instead of a `warning`

pditommaso (changes requested):
Thanks for this improvement to the docs. This is actually very useful. Just suggested a minor improvement

Kibubu:
using `note` instead of `warning` is reasonable, I was just salty because my host's `~/.gitconfig` got modified and I did not expect that at all.

pditommaso (approved):

pditommaso:
Thanks

PR #3171, opened 2022-09-01
Set NXF_FUSION_WORK env variable when using fusion

jordeu:
When Fusion enabled, FusionFS needs `NXF_FUSION_WORK` defined to known where to store the FUSE debugging logs file.

This PR is setting this variable to the process work directory, so in this way a `.fusionfs.log` is created next to the process `.nextflow.log` logging file.


modules/nextflow/src/main/groovy/nextflow/executor/fusion/FusionScriptLauncher.groovy
```diff
@@ -102,6 +102,10 @@ class FusionScriptLauncher extends BashWrapperBuilder {
         return buckets
     }
 
+    String fusionWork() {
+        return "/fusion/${scheme}${remoteWorkDir}"
```
pditommaso, line 106:
```suggestion
        return toContainerMount(remoteWorkDir).toString()
```

What about?

jordeu: Yes, better. Done and tested.

pditommaso:
I guess this is going to be needed also for AWS Batch, right?

jordeu:
Yes

PR #3187, opened 2022-09-06
Replace rolling filenames with timestamp-based filenames

### LINKED ISSUE #3186: Get rid of file name rolling for reports, trace, dag and timeline generation
When creating the execution report and other resource files such as timeline, DAG and trace file, nextflow roll the name any existing file with the name. 

https://github.com/nextflow-io/nextflow/blob/3ef915128ee48ba3422cc7b13807c438ed170b4f/modules/nextflow/src/main/groovy/nextflow/trace/TraceFileObserver.groovy#L209-L210

This avoid the user to overwrite accidentally existing files, however, it can be an expensive operation when the file is hosted into a remove bucket repository since renaming it's mapped to a copy & delete operation (up to 10 files!) 

The goal of this issues is to get rid of the file name rolling functionality. 

When nextflow tries to create such files, if a file with such name, a warning should be shown and the corresponding report/trace file should *not* be created. 

If the strict mode is enabled and error should be reported and the execution stopped

bentsherman:
Closes #3186 

The only issue currently is that the log file can have a different timestamp than the other report files, because they are evaluated at different points in the code. We should find some way to compute the timestamp once for all logs and reports.


bentsherman:
Apparently the `Date` class is out of fashion, so I will refactor the changes to use `java.time.*` classes, which should be easier to mock.

modules/nextflow/src/main/groovy/nextflow/cli/Launcher.groovy
```diff
@@ -164,10 +164,12 @@ class Launcher {
...
         if( !options.logFile ) {
+            String timestamp = new Date().format('yyyy-MM-dd_HH-mm-ss')
+
             if( isDaemon() )
-                options.logFile = System.getenv('NXF_LOG_FILE') ?: '.node-nextflow.log'
+                options.logFile = System.getenv('NXF_LOG_FILE') ?: ".node-nextflow_${timestamp}.log"
             else if( command instanceof CmdRun || options.debug || options.trace )
-                options.logFile = System.getenv('NXF_LOG_FILE') ?: ".nextflow.log"
+                options.logFile = System.getenv('NXF_LOG_FILE') ?: ".nextflow_${timestamp}.log"
         }
```
pditommaso, line 173: log files should continue to use the rolling strategy

modules/nextflow/src/main/groovy/nextflow/config/ConfigBuilder.groovy
```diff
@@ -666,7 +669,7 @@ class ConfigBuilder {
             if( cmdRun.withWebLog != '-' )
                 config.weblog.url = cmdRun.withWebLog
             else if( !config.weblog.url )
-                config.weblog.url = WebLogObserver.DEF_URL
+                config.weblog.url = 'http://localhost'
         }
```
pditommaso, line 673: this is unrelated

modules/nextflow/src/main/groovy/nextflow/trace/DefaultObserverFactory.groovy
```diff
@@ -42,12 +42,12 @@ class DefaultObserverFactory implements TraceObserverFactory {
...
+            return
+
         String url = config.navigate('weblog.url') as String
-        if ( isEnabled ) {
-            if ( !url ) url = WebLogObserver.DEF_URL
-            def observer = new WebLogObserver(url)
-            result << observer
-        }
+        def observer = new WebLogObserver(url)
+        result << observer
```
pditommaso, line 50: this is unrelated

modules/nf-commons/src/main/nextflow/extension/FilesEx.groovy
```diff
@@ -1424,48 +1424,6 @@ class FilesEx {
...
-            else{
-                Files.delete(newFile)
-            }
-        }
-
-        FilesEx.renameTo(self, newFile)
-
-    }
-
     static private LINK_PARAMS = [hard:Boolean, overwrite: Boolean]
```
pditommaso, line 1427: This can stay even if not used

pditommaso (changes requested):

modules/nextflow/src/main/groovy/nextflow/config/ConfigBuilder.groovy
```diff
@@ -643,7 +646,7 @@ class ConfigBuilder {
             if( cmdRun.withDag != '-' )
                 config.dag.file = cmdRun.withDag
             else if( !config.dag.file )
-                config.dag.file = GraphObserver.DEF_FILE_NAME
+                config.dag.file = 'dag.dot'
```
pditommaso, line 649: why this name does use the timestamp?

bentsherman: It didn't make sense to me for the DAG to be timestamped, because it doesn't change very often.

bentsherman: On second thought maybe the DAG should be timestamped, because the DAG can look different from run to run based on piipeline params that affect the DAG structure (eg. enabling/disabling steps).

abhi18av: > because the DAG can look different from run to run based on pipeline params that affect the DAG structure (eg. enabling/disabling steps).

Agreed, this would make for a good pipeline dev-time experience.

pditommaso:
Merged manually via https://github.com/nextflow-io/nextflow/commit/a762ed59c1af255183da8a2d26907206964eade0

PR #3193, opened 2022-09-07
bugfix: intermediate multipart upload requires a minimum size

jorgeaguileraseqera:
if after uploading a few bytes the flush is called, a new multipart is added to the queue but the minimum size required from AWS is 1Mb. 
No error is thrown until the end when completeMultiPart is called 

This PR maintains the current buffer as "active" if flush is calling and buffer length is under minimum


pditommaso:
Well spotted!

PR #3196, opened 2022-09-08
Convert docs to Markdown

### LINKED ISSUE #2739: Convert Nextflow docs to markdown
After working on the Nextflow docs multiple times, I've concluded that RST is a real pain to work with. I've encountered many formatting issues that seem to result from confusion about RST syntax, as it is just different enough from markdown to be confusing.

I think we should just convert the Nextflow docs to markdown. It looks like [Sphinx supports markdown](https://www.sphinx-doc.org/en/master/usage/markdown.html) and there are a number of [tools](https://pandoc.org/try/?text=&from=rst&to=markdown_strict&standalone=0) to do it automatically. So the docs won't change outwardly, they will just be easier to edit (in my opinion).

bentsherman:
Closes #2739 

I used the `rst2myst` tool to convert the RST docs to Markdown, which did 99% of the work. There were only two major differences:

1. Same-page links weren't converted correctly, but they were easily fixed with some search-and-replace.
2. In RST, italics can be specified as <span>\`foo\`</span> or <span>\*foo\*</span>. The former syntax was converted to code while the latter syntax was converted correctly to italics (basically they were left the same). So some words were formatted as code whereas before they were italicized. Also fixed with some quick search-and-replace.

A few hours of work and I think the docs are ready. Easiest way to verify the rendered HTML is to built the docs before and after the PR and open them side-by-side in your browser. Then you can scroll through and see that the markdown docs are identical.


ewels:
Amazing!

image

docs/README.md
```diff
@@ -1,50 +1,30 @@
...
-To edit and contribute to the documentation you only need a text editor to change the
-appropriate `.rst` files in this directory.
+## Dependencies
 
-Once you have edited the documentation files verify the your changes are correctly applied
-using the command below to generate the HTML files:
+The build dependencies can be installed with `pip`:
 
+```bash
+pip install sphinx=3.5.4 myst-parser sphinx_rtd_theme
```
abhi18av, line 11: @bentsherman , maybe best to pin the version for `myst-parser` and `sphinx_rtd_theme` the docs are known to work with.

pditommaso:
@bentsherman, please add a mention in the readme how setup the docs build tooling

docs/README.md
bentsherman: Done

bentsherman:
@pditommaso please see the `docs/README.md`, it contains all the instructions you need to build the docs.

docs/README.md
abhi18av: Thanks Ben - really looking forward to this getting merged 🤩

ewels:
@pditommaso - to clarify, basically nothing has changed in the build tooling. Just one additional python dependency in initial setup.

Short version of what's changed:

```diff
# Initial setup
- pip install sphinx==3.5.4 sphinx_rtd_theme
+ pip install sphinx==3.5.4 sphinx_rtd_theme myst-parser==0.17.0 

# Build docs
make clean html
```

bentsherman:
@pditommaso Ready to merge!

please merge 🙏

pditommaso:
I did 

```
pip install sphinx==3.5.4 myst-parser==0.17.0
make html 
```

but 

```
Extension error:
Could not import extension myst_parser (exception: No module named 'myst_parser')
make: *** [html] Error 2
```

bentsherman:
Which version of Python are you using?

bentsherman:
I am not getting that error. I have a Conda environment with Python 3.7. However, I just encountered some other issues that I haven't seen before, so I fixed it by upgrading the dependencies:
```bash
pip install sphinx==5.3.0 myst-parser==0.18.1 sphinx-rtd-theme==1.1.1
```

pditommaso:
Fail also after using the above command. This is my python version 

```
» python3 --version
Python 3.8.9

```

bentsherman:
Okay, I tried a conda environment with 3.8 and it still worked. So I am unable to reproduce your error. I can only offer a few things to try:
- add `import myst_parser` to `conf.py`
- run `less $(which sphinx-build)` and check which python sphinx is using
- use a conda environment: `conda create -n nextflow-docs python=3.7`

pditommaso:
Adding the import line on the `conf.py` I get 

```
Configuration error:
There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "/opt/homebrew/lib/python3.9/site-packages/sphinx/config.py", line 332, in eval_config_file
    exec(code, namespace)
  File "/Users/pditommaso/Projects/nextflow/docs/conf.py", line 16, in <module>
    import myst_parser
ModuleNotFoundError: No module named 'myst_parser'
```

What I'm not understanding is why there's python3.9 in the error trace if I've python 3.8.9 installed

bentsherman:
You should check the `sphinx-build` script and see which python it's trying to use.

bentsherman:
Why don't you just create a conda environment to build the docs?

pditommaso:
That's a good idea, ho It should not be that hard to setup.

bentsherman:
> That's a good idea, ho It should not be that hard to setup.

It's a good practice though, and not much extra effort:
```bash
conda create -n nextflow-docs python=3.7 -y
conda activate nextflow-docs
pip install ...
```

I have different conda environments for each project (nextflow docs, tower docs, groundswell, quilt, etc...) so that they don't clobber each other. And conda uses caching and soft links to minimize the storage footprint.

Hopefully it won't matter once we can build and deploy the docs in a GH action, but it would still be a good practice to follow.

pditommaso:
What about just getting rid off sphinx and move the docs to Markdown + Mkdocs? @ewels made an excellent work with https://training.nextflow.io/

ewels:
What I'd like to do is to integrate the docs into the new Nextflow website. Have it as one build instead of a separate mini-site (maybe also the training too).

This PR along with the training was intended as an intermediate step in that direction. Once the content is in markdown we can start building the new Nextflow website, which can load the docs markdown from this repo.

We can do mkdocs+material here but it feels a bit of a waste / lots of design changes for users if we're then going to change the front again again in a couple of months for the new site.

pditommaso:
> What I'd like to do is to integrate the docs into the new Nextflow website

In principle this can be desirable, however, it should be taken into consideration two strong requirements: 

1. keep the docs source in this repository so that code changes can easily be kept aligned with the corresponding docs 
2. make the docs rendering self-contained so that editing can be easily tested. 

Not sure how this can be done, requiring the docs to be merged in the main website

mribeirodantas:
To be honest, I really really like the first point Paolo mentioned. Having changes/new features closely tied to documentation updates is very nice, in my opinion.

ewels:
Yup, I absolutely agree on both points. I also have both as hard requirements. I have ideas for how to do both, I'm currently prototyping with the new MultiQC website to see how well they work.

pditommaso:
Made some progress here. I've pushed a pre-built container to `nextflow/sphinx:5.3.0`. 

I've noticed that the resulting Markdown uses non-standard syntax e.g. [{ref}](https://github.com/nextflow-io/nextflow/blob/11c860e41f6c141cc03e2af7f657c9ab1ecc0a7f/docs/aws.md#L19-L19), [{note}](https://github.com/nextflow-io/nextflow/blob/11c860e41f6c141cc03e2af7f657c9ab1ecc0a7f/docs/aws.md#L443-L445), and [tables](https://github.com/nextflow-io/nextflow/blob/11c860e41f6c141cc03e2af7f657c9ab1ecc0a7f/docs/config.md#L100-L107). 

I wonder is that compatible with Mkdocs?

bentsherman:
I think that syntax is specific to Sphinx, but it can be converted to mkdocs with some search & replace.

ewels:
Hopefully no need for `[ci skip]` in your commit messages any more @bentsherman! 🙌🏻

bentsherman:
I updated the docs CI because it failed with missing dependencies, but now the PR has changes outside of `docs` so the docs CI doesn't run 😆

bentsherman:
Oh wait it did nevermind

docs/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+RUN micromamba install --yes --name base --channel conda-forge \
+      make=4.3 \
+      python=3.7 \
+      && \
+    micromamba clean --all --yes
+
+RUN eval "$(micromamba shell hook --shell=bash)" && \
+    micromamba activate && \
+    pip install sphinx==5.3.0 myst-parser==0.18.1 sphinx-rtd-theme==1.1.1
```
ewels, line 13: What do you think about keeping these in a `requirements.txt` file under `docs`? Saves repeating them everywhere..

docs/README.md
```diff
@@ -1,50 +1,46 @@
...
-Once you have edited the documentation files verify the your changes are correctly applied
-using the command below to generate the HTML files:
+The most convenient approach is to create a Conda environment with Python 3.7 (other versions may work but haven't been tested).
 
-```
-make clean html
+The build dependencies can be installed with `pip`:
+
+```bash
+pip install sphinx==5.3.0 myst-parser==0.18.1 sphinx-rtd-theme==1.1.1
```
ewels, line 13: See above suggestion - this would then become:

```suggestion
pip install -r docs/requirements.txt
```

docs/Dockerfile
bentsherman: 👍

plugins/nf-tower/build.gradle
```diff
@@ -33,11 +33,11 @@ dependencies {
     compileOnly 'org.slf4j:slf4j-api:1.7.10'
     compileOnly 'org.pf4j:pf4j:3.4.1'
 
-    api("org.codehaus.groovy:groovy-yaml:3.0.14")  { transitive=false }
+    api("org.codehaus.groovy:groovy-yaml:3.0.15")  { transitive=false }
     api "com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.12.5"
     api "com.fasterxml.jackson.core:jackson-databind:2.12.5"
```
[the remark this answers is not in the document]
bentsherman: @sonatype-lift ignore

docs/make-html.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bashg
+docker run -v $(pwd):/tmp nextflow/sphinx:5.3.0 -- make html
```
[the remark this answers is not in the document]
bentsherman: @sonatype-lift ignore

docs/.dockerignore
```diff
@@ -1 +1 @@
-**/*
\ No newline at end of file
+!requirements.txt
```
ewels, line 1: Ahhhh yeah that'd do it 😅

bentsherman:
On it!

bentsherman:
Okay it's ready

pditommaso:
It's m-day!

bentsherman:
I can die happy now

mribeirodantas:
THE DAY HAS COME! HABEMUS MARKDOWN 🤩

PR #3197, opened 2022-09-09
feature: include debug information

jorgeaguileraseqera:
Currently, Nextflow is not compiling scripts with debug information and parsing all scripts as text so the GroovyCompiler is not able to include variable names in the generated binary

This PR sets the debug configuration and compile directly the File when is proportionated

FYI: when the script contains some error, this version fails as the previous version but in some cases, the explanation of the error is different due now the compiler knows the file


pditommaso:
Closing in favour of #4266

PR #3205, opened 2022-09-12
Add fusion support to grid executors

bentsherman:
The strategy here is to create a secondary wrapper script that generates the container launch command, and pipes this script into the executor submit command. Just need to confirm that it works with every grid executor. Most of the executors can read the job script from stdin, but for any that don't, we can just create a temp file.

Here is a table with some useful information about each grid executor (for my convenience):

| Executor   | Supports stdin | Submit command |
| ---------- | -------------- | -------------- |
| Bridge     | ❔ | `ccc_msub .command.run` |
| Condor     | ✔️ | `condor_submit --terse .command.condor` |
| HyperQueue | `--directives=stdin` | `hq --output-mode=quiet submit --directives=file .command.run` |
| LSF        | ✔️ | `bsub < .command.run` |
| Moab       | probably | `msub --xml .command.run` |
| NSQII      | probably | `qsub .command.run` |
| OAR        | ❔ | `oarsub -S ./.command.run` |
| PBS        | ✔️ | `qsub -N <job-name> .command.run` |
| SGE        | ✔️ | `qsub -terse .command.run` |
| SLURM      | ✔️ | `sbatch .command.run` |

Here is an example of the fusion "stdin" script with SLURM:

```bash
#!/bin/bash
#SBATCH -D /nextflow-ci-dev/work/bd/9b1bedf0be5eae30527308f5c109dc
#SBATCH -J nf-sayHello_(2)
#SBATCH -o /nextflow-ci-dev/work/bd/9b1bedf0be5eae30527308f5c109dc/.command.log
#SBATCH --no-requeue
#SBATCH --signal B:USR2@30
sh -c docker run -i -e "NXF_FUSION_WORK=/fusion/s3/nextflow-ci-dev/work/bd/9b1bedf0be5eae30527308f5c109dc" -e "NXF_FUSION_BUCKETS=s3://nextflow-ci-dev" --privileged wave.stage-tower.net/wt/8241ba77e6c7/nextflow/bash:latest bash -o pipefail -c 'trap "{ ret=$?; cp .command.log /fusion/s3/nextflow-ci-dev/work/bd/9b1bedf0be5eae30527308f5c109dc/.command.log||true; exit $ret; }" EXIT; bash /fusion/s3/nextflow-ci-dev/work/bd/9b1bedf0be5eae30527308f5c109dc/.command.run 2>&1 | tee .command.log'
```


modules/nextflow/src/main/groovy/nextflow/executor/GridTaskHandler.groovy
```diff
@@ -99,21 +101,44 @@ class GridTaskHandler extends TaskHandler {
...
     protected ProcessBuilder createProcessBuilder() {
 
-        // -- log the qsub command
+        // -- create the submit command
         def cli = executor.getSubmitCommandLine(task, wrapperFile)
+
+        // -- run script is piped from stdin when fusion is enabled
+        if( fusionEnabled() ) {
+            cli.removeAll { it.contains(wrapperFile.name) }
+        }
```
bentsherman, line 110: @pditommaso The fusion process builder currently removes the wrapper file from the submit command. We could also split `getSubmitCommandLine()` into two methods, one for the command without the wrapper file line and one for the wrapper file, so that this logic isn't needed.

Also, the condor executor copies the wrapper file to `.command.condor`, so we either need to handle that case here or implement my suggestion.

bentsherman:
```
sbatch ... bash ... sh ... bash ... bash ...
```

Damn we really are the masters of wrapping shell commands 😆

bentsherman:
Almost there. The executor-specific headers need to point to a local work directory and log file, so we'll need to take a similar approach as we did with the local executor. We'll need to make a temp directory for each task and save the `.command.log` to that directory.

bentsherman:
SLURM test is a success! @pditommaso this mess is all yours 😄

pditommaso:
Cool! Am wrong or there are some changes coming from some other PR to local and hyper queue executor?

bentsherman:
The hyperqueue submit command had to be changed to accommodate the fusion stdin script, and the local task handler was changed because I removed the `sh -c` wrapper from `FusionHelper.runWithContainer()`.

I think I need to add some tests to verify the fusion stdin script for each grid executor.

bentsherman:
Okay I added tests for the fusion process builder and the slurm executor. @pditommaso I think you should go ahead and review the changes before I add any more tests. There is probably a cleaner way to implement this fusion support but might require more changes to the executor classes.

pditommaso:
Closing this in favour of https://github.com/nextflow-io/nextflow/pull/3516 which uses mostly the same approach of this PR with 2 main differences: 1) only focus on LSF and Slurm and 2) rely on a more granular container configuration implemented by 03126371

PR #3215, opened 2022-09-16
Fix annotations and labels key sanitization

### LINKED ISSUE #3184: Annotation and label keys are sanitized wrongly
## Bug report 

Current main branch code has a regression bug that breaks the usage of annotations and labels that contain `/` on its key

Expected: Annotation and label keys should be allowed to contain `/`. See Kubernetes docs about the syntax and character set of [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set) and [labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set).

Actual: character `/` in annotation and label keys gets replaced with `_`

Having an available Kubernetes cluster, execute a nextflow script with `nextflow.config` as follows:
```
process {
   container = 'quay.io/nextflow/bash'
   executor = 'k8s'
   pod = [[annotation: 'foo.bar/foo-bar', value: 'true'],[label: 'foo.bar/foo-bar', value: 'true']]
}
```

### Program output 
Excerpt from the generated pod manifest:
```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    foo.bar_foo-bar: "true"
  labels:
    app: nextflow
    foo.bar_foo-bar: "true"
```

### Environment 

* Nextflow version: `22.09.1-edge`
* Java version: `11.0.9.1`
* Operating system: Executed from macOS, server side Amazon Linux 2, Kubernetes 1.21
* Bash version: `zsh 5.8.1 (x86_64-apple-darwin21.0)`

### Additional context

This works in the latest stable version. It was introduced in https://github.com/nextflow-io/nextflow/pull/2843/ where the label and annotation keys started to be sanitized.
I should also add that the sanitation of keys is a bit more cumbersome, as the prefix (the part before the `/`) can have up to 253 characters, and not 63, as it is the case for a key that does not contain a `/`

kanor1306:
Closes #3184


bentsherman (approved):

modules/nextflow/src/test/groovy/nextflow/k8s/model/PodSpecBuilderTest.groovy
```diff
@@ -785,6 +785,27 @@ class PodSpecBuilderTest extends Specification {
...
+
+        where:
+        label_key               | str
+        'foo'                   | 'foo'
+        'key 1'                 | 'key_1'
+        'foo.bar/key 2'         | 'foo.bar/key_2'
+        'foo.bar/key 2/key 3'   | 'foo.bar/key_2_key_3'
+        'foo.bar/'              | 'foo.bar'
+        '/foo.bar'              | 'foo.bar'
+        'x2345678901234567890123456789012345678901234567890123456789012345' | 'x23456789012345678901234567890123456789012345678901234567890123'
```
pditommaso, line 804: Since `/` has a special syntax, not sure anymore it should be allowed to provide two or more or them or to allow leading or trailing slashes. Would not be better to report an error?

kanor1306: The reason why I didn't trigger any error is because I think it is consistent behavior with the current one, sort of "I will fix it for you" philosophy, i.e. if you input invalid characters they are just replaced with '_' instead of reporting an error. 

But I can switch it to throw an `IllegalArgumentException` if there are more than on `/` if you think is best. 

What do you think about leading and trailing slashes? Would yo prefer an error or sanitizing as it is now? Currently in this case I remove the slash and consider it a key without prefix.

pditommaso: My understanding is that `/` as special semantic in K8s labels and annotation, therefore if it's wrong formatted i.e. two or more `/` or leading/trailing slashes is very likely the user is messing up thing. Therefore it would be better to throw an error? do you agree? if so I can do quickly

kanor1306: I agree @pditommaso. I think is better to get an error so the user can just adjust the parameters. 

I just pushed the fix and the tests for it.

pditommaso:
Ok, i've merged this manually. Thanks for the contribution! a569afdf

PR #3222, opened 2022-09-20
Change k8s client to use namespace and service account from nextflow config

### LINKED ISSUE #2266: nextflow uses default service account to launch pod in k8s cluster instead of nextflow service account
## Bug report 

We are trying to run a workflow on the EKS cluster. The workflow works well on minikube but on the EKS cluster it doesn't use the context to launch the nextflow pod. It uses `system:serviceaccount:nextflow:default` service account. 

We are running workflow like this:
`
nextflow kuberun https://<gitea-private-repository>
`

nextflow.config file:
```
process.executor = "k8s"

k8s {
   namespace      = "nextflow"
   serviceAccount = "nextflow"
   storageClaimName = "nextflow-pv-claim"
   storageMountPath = "/nextflow/"
   context          = "<current-context>"
}

```

On running the `nextflow kuberun <gitea-private-repository>` it throws access denied:

```
Request GET /api/v1/namespaces/nextflow/persistentvolumeclaims/nextflow-pv-claim returned an error code=403

  {
      "kind": "Status",
      "apiVersion": "v1",
      "metadata": {
          
      },
      "status": "Failure",
      "message": "persistentvolumeclaims \"nextflow-pv-claim\" is forbidden: User \"system:serviceaccount:nextflow:default\" cannot get resource \"persistentvolumeclaims\" in API group \"\" in the namespace \"nextflow\"",
      "reason": "Forbidden",
      "details": {
          "name": "nextflow-pv-claim",
          "kind": "persistentvolumeclaims"
      },
      "code": 403
  }
```

Ideally, it should use the context to launch nextflow pod. 

Moreover, our service account `nextflow` configs looks as follows:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: nextflow-role
rules:
- apiGroups:
  - ""
  resources:
  - pods
  - pods/status
  verbs:
  - get
  - list
  - watch
  - create
  - delete
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: nextflow-rolebinding
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: nextflow-role
subjects:
- kind: ServiceAccount
  name: nextflow
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: nextflow
```

The problem can be resolved by giving `default` service account in the namespace appropriate permissions. 
- Create ns `nextflow`
- Create service account along with role and rolebinding. 
- Create main.nf  with any simple process
- Use nextflow.config shared above
- nextflow kuberun <repo-url>
- It should give the error. 

### Program output 

```
Aug-13 11:28:04.101 [main] ERROR nextflow.cli.Launcher - @unknown
nextflow.k8s.client.K8sResponseException: Request GET /api/v1/namespaces/nextflow/persistentvolumeclaims/efs-claim returned an error code=403

  {
      "kind": "Status",
      "apiVersion": "v1",
      "metadata": {
          
      },
      "status": "Failure",
      "message": "persistentvolumeclaims \"efs-claim\" is forbidden: User \"system:serviceaccount:nextflow:default\" cannot get resource \"persistentvolumeclaims\" in API group \"\" in the namespace \"nextflow\"",
      "reason": "Forbidden",
      "details": {
          "name": "efs-claim",
          "kind": "persistentvolumeclaims"
      },
      "code": 403
  }

	at nextflow.k8s.client.K8sClient.makeRequest(K8sClient.groovy:370)
	at nextflow.k8s.client.K8sClient.makeRequest(K8sClient.groovy)
	at nextflow.k8s.client.K8sClient.get(K8sClient.groovy:379)
	at nextflow.k8s.client.K8sClient.volumeClaimRead(K8sClient.groovy:429)
	at nextflow.k8s.K8sConfig.checkStorageAndPaths(K8sConfig.groovy:216)
	at nextflow.k8s.K8sConfig$checkStorageAndPaths$0.call(Unknown Source)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:47)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:125)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:139)
	at nextflow.k8s.K8sDriverLauncher.run(K8sDriverLauncher.groovy:123)
	at nextflow.cli.CmdKubeRun.run(CmdKubeRun.groovy:75)
	at nextflow.cli.Launcher.run(Launcher.groovy:475)
	at nextflow.cli.Launcher.main(Launcher.groovy:657)
```
### Environment 

* Nextflow version: 21.04.3.5560
* Java version: openjdk 11.0.11
* Operating system: ubuntu 20-04

### Additional context

bentsherman:
Closes #2266


bentsherman:
I used `<class>.metaClass.static.<method>` to mock static methods. Looks like it works for Java classes as well, so it could eliminate the need to create little helper methods like `ConfigDiscovery.path()` just for testing purposes.

modules/nextflow/src/main/groovy/nextflow/k8s/client/ConfigDiscovery.groovy
```diff
@@ -34,27 +34,20 @@ import org.yaml.snakeyaml.Yaml
...
-
-    private String context
-
     /**
      * Discover Kubernetes service from current environment settings
      */
-    ClientConfig discover() {
-
-        config = new ClientConfig()
+    static ClientConfig discover(String context=null) {
```
pditommaso, line 40: not sure this was a good idea, static methods should be avoided as much as possible because it makes hard testing (between the other things)

pditommaso (commented):
Not understanding all these change to only pass the namespaces as argument here ?

https://github.com/nextflow-io/nextflow/blob/783042c41578978c4159cb934645496484b8a384/modules/nextflow/src/main/groovy/nextflow/k8s/client/ConfigDiscovery.groovy#L117-L117

bentsherman:
I thought it was cleaner to separate the config discovery from the auth token discovery, rather than pass the namespace and service account down the call stack to `discoverAuthToken()`. That required making `ConfigDiscovery` a static helper class, which makes sense to me because `ConfigDiscovery` doesn't really have any state like an object. However this testing framework really doesn't play well with static methods. I guess I could make `ConfigDiscovery` a singleton instead, like `OperatorImpl`.

modules/nextflow/src/main/groovy/nextflow/k8s/client/ConfigDiscovery.groovy
bentsherman: Okay, I changed ConfigDiscovery to a singleton to make testing easier

pditommaso:
it's already, what's done for the [context](https://github.com/nextflow-io/nextflow/blob/783042c41578978c4159cb934645496484b8a384/modules/nextflow/src/main/groovy/nextflow/k8s/client/ClientConfig.groovy#L97-L97), that than it's used in the `[formConfig](https://github.com/nextflow-io/nextflow/blob/783042c41578978c4159cb934645496484b8a384/modules/nextflow/src/main/groovy/nextflow/k8s/client/ConfigDiscovery.groovy#L95-L95)` that's the same scope  invoking also [discoverAuthToken](https://github.com/nextflow-io/nextflow/blob/783042c41578978c4159cb934645496484b8a384/modules/nextflow/src/main/groovy/nextflow/k8s/client/ConfigDiscovery.groovy#L117-L117). 

Looks straightforward extending it to support also `namespace` and `serviceAccount`

bentsherman:
`ConfigDiscovery.fromConfig` loads the namespace and user from the current context, but `K8sConfig.getClient` applies the namespace and serviceAccount from the nextflow config if they are provided. Basically, the nextflow `k8s` config settings should be used first, and the current context used as a backup.

I think there is still a logic error here -- if the current context contains a user with a token, it will be used even if a service account is provided in the nextflow config. But I think the service account from the nextflow config should be used first.

pditommaso:
> ConfigDiscovery.fromConfig loads the namespace and user from the current context, but K8sConfig.getClient applies the namespace and serviceAccount from the nextflow config if they are provided. Basically, the nextflow k8s config settings should be used first, and the current context used as a backup.

Indeed, `namespace` and `serviceAccount` in the `k8s` config should take precedence over the one in context. 

Let's focus on this.

bentsherman:
Here is my understanding of things:
1. load user, cluster, and namespace from current context
2. load certs and auth token from current user (if present)
3. use namespace from nextflow config (if present)
4. use service account from nextflow config (if present)

And the priority for credentials is as follows:
1. service account from nextflow config -- discover auth token via kubectl
2. current context user -- use certs from kube config
3. current context user -- use auth token from kube config

If these priorities are correct then I need to adjust the logic slightly. Also need to make sure that `default` namespace and `default` service account are used by default.

pditommaso:
It sounds reasonable

pditommaso:
I think the logic is correct, however the use of `discoverAuthToken` can fail when running it within a pod. 

Is there any Kube API that can be used to find it?

bentsherman:
Okay it's ready. The only caveat now is that if a user+token is defined in the KUBECONFIG, it will override `k8s.serviceAccount` for the purposes of API requests. But I think most users don't do that anyway. I will make another PR later to deal with this minor issue.

So the overall logic is mostly unchanged, except that the namespace and serviceAccount are now used to discover the auth token.

pditommaso:
Merge manually trying to not rely on any intermetied state for the `namespace` and `serviceAccount` attributes. See c3364d0f

shah-newaz:
Sorry, looks like the problem was only resolved for GKE? Still facing the same error for EKS?

Did anyone test it for EKS?

I have the proper ServiceAccount created etc.
Debugging looking at the messages above, I find that

```kubectl -n ${NAMESPACE} get serviceaccount ${SERVICE_ACCOUNT} -o jsonpath='{.secrets[0].name}'```

Actually gives me empty string because EKS serviceaccount does not seem to have a secrets property.

This is what EKS service account I created looks like:

```
apiVersion: v1
kind: ServiceAccount
metadata:
  annotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::SOME_ACCOUNT_ID/nextflow-role
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"ServiceAccount","metadata":{"annotations":{},"name":"nextflow-sa","namespace":"nextflow"}}
  creationTimestamp: "2024-03-18T19:09:19Z"
  name: nextflow-sa
  namespace: nextflow
  resourceVersion: "163905357"
  uid: e9881e36-933e-4e76-820d-ca214d87916e

```

The error I get from NextFlow is:

```
Caused by:
  Request POST /api/v1/namespaces/nextflow/pods returned an error code=403

  {
      "kind": "Status",
      "apiVersion": "v1",
      "metadata": {
          
      },
      "status": "Failure",
      "message": "pods is forbidden: User \"system:anonymous\" cannot create resource \"pods\" in API group \"\" in the namespace \"nextflow\"",
      "reason": "Forbidden",
      "details": {
          "kind": "pods"
      },
      "code": 403
  }
  
  ```
Any help is greatly appreciated!

PR #3239, opened 2022-09-23
[Bugfix] Ignore Kubernetes scheduling problems

Lehmann-Fabian:
It rarely happens in Kubernetes that tasks are faster scheduled than the Kubernetes API refreshes.
It has been a known issue for years and not yet fixed:  https://github.com/kubernetes/kubernetes/issues/34772
If it happens, the pod fails.

Until now, Nextflow does not handle this problem explicitly, leading to random workflow failures.

For example, I ran the eager workflow and the qualimap defines `errorStrategy = { task.exitStatus in [1,143,137,104,134,139, 140] ? 'retry' : task.exitStatus in [255] ? 'ignore' : 'finish' }`. This lead to a complete workflow crash.

```
Sep-22 23:44:36.503 [Task monitor] ERROR nextflow.processor.TaskProcessor - Error executing process > 'qualimap (DA220)'

Caused by:
  K8s pod 'nf-e74e13dc82f81740fb895c536be9e2ab' execution failed - reason: OutOfcpu - message: Pod Node didn't have enough resource: cpu, requested: 2000, used: 30275, capacity: 31900

Command executed:

  qualimap bamqc -bam DA220_rmdup.bam -nt 2 -outdir . -outformat "HTML"  --java-mem-size=4G

Command exit status:
  -

Command output:
  (empty)

Work dir:
  /data/work/e7/4e13dc82f81740fb895c536be9e2ab

```

With this PR I solve the problem.


modules/nextflow/src/main/groovy/nextflow/exception/K8sOutOfCPUException.groovy
```diff
@@ -0,0 +1,7 @@
+package nextflow.exception
+
+import groovy.transform.InheritConstructors
+
+@InheritConstructors
+class K8sOutOfCPUException extends RuntimeException {
```
bentsherman, line 6: Please rename to `K8sOutOfCpuException` to match naming conventions.

modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
```diff
@@ -500,6 +503,12 @@ class K8sClient {
             def msg = "K8s pod '$podName' execution failed"
             if( status.reason ) msg += " - reason: ${status.reason}"
             if( status.message ) msg += " - message: ${status.message}"
+            if ( 'OutOfcpu' == status.reason && (status.message as String).startsWith("Pod Node didn't have enough resource: cpu,")) {
```
bentsherman, line 506: Is the message check needed if the reason already matches?

modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
```diff
@@ -500,6 +503,12 @@ class K8sClient {
...
+                throw new K8sOutOfCPUException(msg)
+            }
+            if ( 'OutOfmemory' == status.reason && (status.message as String).startsWith("Pod Node didn't have enough resource: memory,") ) {
+                throw new K8sOutOfMemoryException(msg)
+            }
             final err = status.reason == 'Shutdown'
                     ? new NodeTerminationException(msg)
                     : new ProcessFailedException(msg)
```
bentsherman, line 514: Why don't we use a match expression to handle all of these cases at once? I guess in Groovy 3 you still need to do a regular switch statement, or use a map.

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -972,8 +975,15 @@ class TaskProcessor {
...
-                if( error.cause instanceof NodeTerminationException )
+            if( task && (error.cause instanceof NodeTerminationException
+                    || error.cause instanceof CloudSpotTerminationException
+                    || error instanceof K8sOutOfCPUException
+                    || error instanceof K8sOutOfMemoryException
+            ) ) {
+                if( error.cause instanceof NodeTerminationException
+                        || error instanceof K8sOutOfCPUException
+                        || error instanceof K8sOutOfMemoryException
+                )
```
bentsherman, line 986: To keep these condition checks from getting out of hand, why don't we make NodeTerminationException and your two new exceptions inherit from a common class. Not sure what the name should be but something that denotes "errors that should trigger a retry instead of workflow termination".

bentsherman (changes requested):

modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
Lehmann-Fabian: I wanted to be 100% sure, but the reason seemed unique, so I adjusted it.

modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
Lehmann-Fabian: Done

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
Lehmann-Fabian: I introduced an interface "TemporaryProblem", good point!

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -971,9 +972,10 @@ class TaskProcessor {
             // -- do not recoverable error, just re-throw it
             if( error instanceof Error ) throw error
 
+            boolean temporaryProblem = (error instanceof TemporaryProblem || error?.cause instanceof TemporaryProblem)
             // -- retry without increasing the error counts
-            if( task && (error.cause instanceof NodeTerminationException || error.cause instanceof CloudSpotTerminationException) ) {
-                if( error.cause instanceof NodeTerminationException )
+            if( task && ( temporaryProblem || error.cause instanceof CloudSpotTerminationException ) ) {
+                if( temporaryProblem )
```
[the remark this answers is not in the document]
Lehmann-Fabian: So not an interface? Some are Runtime Exceptions, some are not.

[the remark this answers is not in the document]
Lehmann-Fabian: Done

Lehmann-Fabian:
@bentsherman, by accident, I somehow signed [1d71fa9](https://github.com/nextflow-io/nextflow/pull/3239/commits/1d71fa97ea427188735a44522b60cbe0a6311980) and cannot undo this. I think I have to reopen this PR in another branch. Otherwise, the DCO check will always fail.

modules/nextflow/src/main/groovy/nextflow/exception/RetriableException.groovy
```diff
@@ -0,0 +1,7 @@
+package nextflow.exception
+
+/**
+ * Exceptions implementing this interface will always lead to a retry
+ */
+interface RetriableException {
+}
```
bentsherman, line 7: Sorry I was rash with the naming, we should really call it `TransientException`. Many errors are "retriable", but these "transient" errors are caused by external factors (e.g. spot reclamation) and therefore shouldn't count towards the max retries limit.

bentsherman:
> @bentsherman, by accident, I somehow signed [1d71fa9](https://github.com/nextflow-io/nextflow/pull/3239/commits/1d71fa97ea427188735a44522b60cbe0a6311980) and cannot undo this. I think I have to reopen this PR in another branch. Otherwise, the DCO check will always fail.

Don't worry about it for now, when the PR is ready we can either rebase or Paolo can commit manually with our co-signatures. The DCO has been acting weird lately anyway.

modules/nextflow/src/main/groovy/nextflow/exception/RetriableException.groovy
Lehmann-Fabian: Okay, done.

bentsherman (approved):

pditommaso:
This seems happening when setting the pod requests.cpu == limits.cpu, see [here](https://unix.stackexchange.com/questions/691933/kubernetes-outofcpu-error-when-requests-cpu-limits-cpu), something that it [was recently changed](https://github.com/nextflow-io/nextflow/pull/3338), so that cpus allocation only uses requests.cpu. 

Wondering if this PR is still relevant?

Lehmann-Fabian:
It is independent of the CPU limit.
The linked issue: https://github.com/kubernetes/kubernetes/issues/106884 defines a procedure to reproduce the error. It only uses requests (https://gist.github.com/shioshiota/84f1f4e162c335fe551c60a118676c3c).

pditommaso:
Ok, but what guarantees that the task does not get resubmitted forever because the cluster lacks of CPUs?

Lehmann-Fabian:
This error is only a temporal issue, and the Kubernetes state is updated relatively quickly. So the second attempt is mostly successful.
If the cluster does not have enough CPUs, the task remains pending but does not throw this error.

pditommaso:
ok, let's go ahead with this. I'd like to suggest the changes in the patch attached to merge 
[patch.txt](https://github.com/nextflow-io/nextflow/files/10086834/patch.txt)

pditommaso:
This was merged manually via d86ddc36

PR #3249, opened 2022-09-27
Print errors and warnings to stderr when ansi logger is unavailable

### LINKED ISSUE #2916: No such file in stdout instead of stderr

Some messages in terminal are printed in stdout instead of stderr. For example, when there is a missing file, the error is printed in stdout.

In `./main.nf` file :
```groovy
#!/usr/bin/env nextflow
nextflow.enable.dsl=2

include { splitData } from './modules/split_data'
```

```bash
$ nextflow run main.nf > out.stdout 2> out.stderr

$ cat out.stdout 
N E X T F L O W  ~  version 22.04.0
Launching `main.nf` [adoring_hamilton] DSL2 - revision: 457b997805
No such file: Can't find a matching module file for include: ./modules/split_data

 -- Check script 'main.nf' at line: 6 or see '.nextflow.log' file for more details

$ cat out.stderr

```
The lines 
```
No such file: Can't find a matching module file for include: ./modules/split_data

 -- Check script 'main.nf' at line: 6 or see '.nextflow.log' file for more details
```
Should be in the stderr for more consistence.

### Environment 

* Nextflow version: 22.04.0 build 5697
* Java version: openjdk version "17.0.3" 2022-04-19
* Operating system: Linux (Fedora 36)
* Bash version: GNU bash, version 5.1.16(1)-release (x86_64-redhat-linux-gnu)

bentsherman:
Closes #2916 

`System.out.println` is used as a backup to log messages when the ansi logger is unavailable. This PR routes errors and warnings to `System.err.println` instead. Also fixed a bug in a related function.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/util/LoggerHelper.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/util/LoggerHelper.groovy
@@ -437,7 +437,7 @@ class LoggerHelper {
     static protected void appendFormattedMessage( StringBuilder buffer, ILoggingEvent event, Throwable fail, Session session) {
         final className = session?.script?.getClass()?.getName()
         final message = event.getFormattedMessage()
-        final quiet = fail instanceof AbortOperationException || fail instanceof ProcessException || ScriptRuntimeException
+        final quiet = fail instanceof AbortOperationException || fail instanceof ProcessException || fail instanceof ScriptRuntimeException
         final normalize = { String str -> str ?. replace("${className}.", '')}
         List error = fail ? findErrorLine(fail) : null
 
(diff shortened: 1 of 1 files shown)


pditommaso:
Is this related to the colouring? stdout/stderr does not affect the colour of the message

bentsherman:
Related but not the same, both issues are affected by the ansi logger being unavailable near the end of execution. This PR only ensures such errors are logged to stdout, does not affect colors.

pditommaso:
I prefer to keep in the current form. This can potentially break some logging relying on the current output over stdout instead of stderr

modules/nextflow/src/main/groovy/nextflow/util/LoggerHelper.groovy
```diff
@@ -437,7 +437,7 @@ class LoggerHelper {
     static protected void appendFormattedMessage( StringBuilder buffer, ILoggingEvent event, Throwable fail, Session session) {
         final className = session?.script?.getClass()?.getName()
         final message = event.getFormattedMessage()
-        final quiet = fail instanceof AbortOperationException || fail instanceof ProcessException || ScriptRuntimeException
+        final quiet = fail instanceof AbortOperationException || fail instanceof ProcessException || fail instanceof ScriptRuntimeException
```
bentsherman, line 440: @pditommaso still should commit this change manually, it's a general bug fix

pditommaso: I see it now, thanks for spotting it. I'm going to apply this patch

Issue #3105, opened 2022-08-08
Behaviour of `.collect()` and `.toList()`


cjw85:
## Bug report 

The semantics of these two operators are poorly documented and counter intuitive. From the documentation its not clear what the different between `.collect()` and `.toList()`, as the examples given are identical. `.toList()` does no in fact return a list.

### Expected behavior and actual behavior

`.collect()`s behaviour is to flatten a **single** level of structure in the input items. This is undocumented and seems arbitrary.

The reasonably expected behaviour would be simply to return a value channel with all items from the input queue channel as a single emission.

`.toList()` returns a channel where it might reasonably be expected to return a... um, well, list.

### Steps to reproduce the problem

```
nextflow.enable.dsl=2

workflow {
    ch1 = Channel.of(
        [1,2],
        [3,4,[5,6]],
        7,
        8
    )
    ch1.view()
    ch1.collect().view()
}
```

### Program output 

```
N E X T F L O W  ~  version 22.04.4
Launching `temp1.nf` [distraught_shockley] DSL2 - revision: 021b10ff49
[1, 2]
[3, 4, [5, 6]]
7
8
[1, 2, 3, 4, [5, 6], 7, 8]
```

Expected output would be:
```
[[1,2], [3, 4, [5,6]], 7, 8]
```

which is what `.toList()` returns.

### Environment 

* Nextflow version: 22.04.4
* Java version: java version "17.0.2" 2022-01-18 LTS
* Operating system: macOS
* Bash version: N/A

### Additional context

(Add any other context about the problem here)

bentsherman:
I discovered the same thing recently. What I found was that `collect` and `toList` are almost identical, except that `collect` has two hidden options `flat` and `sort`, `flat` is true by default, and these options are in the operator docs but commented out. 🙁 I have uncommented them in #2816 as part of updating the docs to DSL2.

However I think the two operators are otherwise the same. They both emit a single list item. Every operator returns one or more channels due to their asynchronous nature.

By the way, thank you for raising these issues about language semantics. I tend to agree with you about the various inconsistencies in Nextflow. I think the language just developed organically and is in need of some cleanup. DSL2 was a huge step in that direction, but operators in particular need to be reviewed. In this case, I think we should keep either `collect` or `toList` and deprecate the other, and maybe deprecate `toSortedList` as well. Having all of these different but mostly similar operators just causes a lot of confusion.

cjw85:
[...]

PR #3276, opened 2022-10-07
Clarify the difference between collect and toList

mribeirodantas:
This PR is a reaction not only to the message on Slack below, but also to other
requests in the past about this difference.

https://nextflow.slack.com/archives/C02T98A23U7/p1665099551564579
https://github.com/nextflow-io/nextflow/issues/3105


mribeirodantas:
I added the difference to `toList` because it is shown after `collect`. Should I repeat the same explanation for the `collect`part?

docs/operator.rst
```diff
@@ -2052,6 +2052,13 @@ and emits the resulting collection as a single item. For example::
     [1,2,3,4]
     Done
 
+
+The difference between ``toList`` and ``collect`` is that while ``collect`` produces
```
bentsherman, line 2056: Let's put this in a note admonition. Also, I believe `collect` automatically flattens the input (check the collect section to confirm) so we should mention that here as well.

mribeirodantas: It doesn't flatten the channel, but the outputs are indeed different.

```
Channel.of([1, [2, 3]]).collect().view { "collect: $it" }
Channel.of([1, [2, 3]]).toList().view { "toList: $it" }
```
The output is:
```
toList: [[1, [2, 3]]]
collect: [1, [2, 3]]
```

bentsherman: This is because of how you are initializing the channel with a single list item, you should either use `fromList` or remove the outer list brackets:
```
> Channel.of([1, [2, 3]]).collect().view { "collect: $it" }  
collect: [1, [2, 3]]

> Channel.fromList([1, [2, 3]]).collect().view { "collect: $it" }  
collect: [1, 2, 3]

> Channel.of(1, [2, 3]).collect().view { "collect: $it" }  
collect: [1, 2, 3]

# collect only flattens one level
> Channel.of(1, [2, [3]]).collect().view { "collect: $it" } 
collect: [1, 2, [3]]
```

docs/operator.rst
```diff
@@ -2052,6 +2052,20 @@ and emits the resulting collection as a single item. For example::
...
+``collect`` also flattens one level.``toList`` is
+equivallent to::
+
+    collect().ifEmpty([])
+
+Regarding the flatenning of one level, it can be seen with the following snippet::
+
+    Channel.fromList([1, [2, 3]]).collect().view { "collect: $it" }
+    Channel.fromList([1, [2, 3]]).toList().view { "toList: $it" }
+
```
bentsherman, line 2068:
```suggestion
.. note::
    There are two differences between ``toList`` and ``collect``:

    * When there is no input, ``toList`` emits an empty list whereas ``collect`` emits nothing.
    * By default, ``collect`` flattens list items by one level.

    In other words, ``toList`` is equivalent to::

        collect(flat: false).ifEmpty([])

```

bentsherman (approved):

mribeirodantas:
LGTM, @pditommaso ? 😄

pditommaso (approved):
Nice, thanks

Issue #2970, opened 2022-06-17
Apptainer support


ghuls:
## New feature

It would be great if Apptainer is natively supported. It should be 99% the same as singularity.

http://apptainer.org/
Singularity has joined the Linux Foundation and is now Apptainer!

## Usage scenario 

Apptainer (as singularity symlink) complains about deprecated usage when it sees SINGULARITYENV env variables.
```
Command error:
WARNING: DEPRECATED USAGE: Forwarding SINGULARITYENV_TMPDIR as environment variable will not be
supported in the future, use APPTAINERENV_TMPDIR instead
WARNING: DEPRECATED USAGE: Forwarding SINGULARITYENV_NXF_DEBUG as environment variable will
not be supported in the future, use APPTAINERENV_NXF_DEBUG instead 
```

## Suggest implementation 

Add APPTAINERENV envs instead of SINGULARITYENV when using apptainer.

matthdsm:
Would it be very involved to add an `apptainer` alias for the `singularity` settings which modifies the env value prefixes?

tbugfinder:
Following the apptainer development it be valuable to add it as its dedicated container engine compared to just "allowing" APPTAINER* ENV variables for singularity binary.

pditommaso:
This is a desirable feature but the community should step in and provide a PR. It should be quite easy to implement the Apptainer support by creating an `AptainerBuilder` as done for singularity 

https://github.com/nextflow-io/nextflow/blob/fdde418f053fd462a54f5ca4b9f3390792ebb9a8/modules/nextflow/src/main/groovy/nextflow/container/SingularityBuilder.groovy#L30-L30

jfy133:
Looks like it was done here: https://github.com/nextflow-io/nextflow/pull/3345

pditommaso:
It is! already in 22.11.0-edge release

matthdsm:
Can this also be added to the docs?
https://www.nextflow.io/docs/latest/container.html#singularity

pditommaso:
Feel free to provide a PR for that

maxulysse:
@matthdsm no PR for the docs?

matthdsm:
Here we go!
https://github.com/nextflow-io/nextflow/pull/3699

PR #3290, opened 2022-10-14
add Apptainer support derived from Singularity implementation.

tbugfinder:
Closes: #2970 

Signed-off-by: tbugfinder <tbugfinder@online.ms>

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -1166,6 +1166,7 @@ class Session implements ISession {
         getContainerConfig0('shifter', engines)
         getContainerConfig0('udocker', engines)
         getContainerConfig0('singularity', engines)
+        getContainerConfig0('apptainer', engines)
         getContainerConfig0('charliecloud', engines)
 
         def enabled = engines.findAll { it.enabled?.toString() == 'true' }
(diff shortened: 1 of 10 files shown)


pditommaso:
Thanks a lot for this valuable PR. I wonder if the Apptainer image format is different from the Singularity one. If they use the same images I guess the `SingularityCache` is not really needed.

tbugfinder:
Apptainer supports and uses the SIF format. It doesn't add its dedicated container format.

pditommaso:
Thanks for clarifying this. Also, I was wondering if specific support in nextflow is really needed. Stating the [Apptainer faq](https://apptainer.org/news/community-announcement-20211130/) the backward should be available, at least at this stage 

>  Apptainer will provide singularity as a command line link and will maintain as much of the CLI and environment functionality as possible. From the user's perspective, very little, if anything, will change.

tbugfinder:
Correct 👍 
Apptainer installation includes singularity alias.
Nevertheless apptainer has deprecated singularity environment variables and might add specific APPTAINER_* vars in upcoming releases. Therefore Apptainer specific support makes sense.

https://apptainer.org/docs/user/main/singularity_compatibility.html#singularity-environment-variable-compatibility

pditommaso:
OK, agree on that. However, I've realised that the current implementation it's just identical to singularility and likely they will not diverge dramatically, at least in the short term. 

I think it would make sense to refactor the current implementation using common base class `BaseSingularityBuilder` with the current logic and parametrising the `singularity|apptainer` naming into an abstract method that can be defined by `SingularityBuilder` and `ApptainerBuilder` sub-classes.

pditommaso:
Closing in favour of #3345

PR #3299, opened 2022-10-19
Add note about default mail ssl protocol [ci skip]

bentsherman:
This PR adds a note to the docs about the incompatibility between some Java versions and 3rd party integrations that require TLSv1.2. Motivated by a number of Tower users who had issues configuring their mail notifications.


pditommaso (approved):
Excellent

PR #3338, opened 2022-11-01
Remove cpu limits from K8s pod spec builder

### LINKED ISSUE #3329: Do not use cpus limit requirement with K8s executor
There are several community advices reporting the use of `limits` to cap CPU usage is a bad idea, that may even result in poor performance and prevent legit CPU usage spikes task.

See [here](https://blog.netdata.cloud/kubernetes-throttling-doesnt-have-to-suck-let-us-help/?utm_source=pocket_mylist) and [here](https://home.robusta.dev/blog/stop-using-cpu-limits) and [here](https://erickhun.com/posts/kubernetes-faster-services-no-cpu-limits/). 

Currently, nextflow set both `requests` and `limits` which according to the above is not a good practice.  

https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy#L418-L422  

The goal of this issue is to only use `requests` for CPUs allocation. 

Other resources such as memory should not be affected by this issue.

bentsherman:
Closes #3329

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
@@ -340,12 +340,6 @@ class PodSpecBuilder {
             env.add(entry.toSpec())
         }
 
-        final res = [:]
-        if( this.cpus )
-            res.cpu = this.cpus
-        if( this.memory )
-            res.memory = this.memory
-
         final container = [ name: this.podName, image: this.imageName ]
         if( this.command )
             container.command = this.command
@@ -417,13 +411,16 @@ class PodSpecBuilder {
             container.env = env
 
         // add resources
-        if( res ) {
-            container.resources = [requests: res, limits: new HashMap<>(res)]
+        if( this.cpus ) {
+            container.resources = addCpuResources(this.cpus, container.resources as Map)
         }
 
-        // add gpu settings
-        if( accelerator ) {
-            container.resources = addAcceleratorResources(accelerator, container.resources as Map)
+        if( this.memory ) {
+            container.resources = addMemoryResources(this.memory, container.resources as Map)
+        }
+
+        if( this.accelerator ) {
+            container.resources = addAcceleratorResources(this.accelerator, container.resources as Map)
         }
 
         // add storage definitions ie. volumes and mounts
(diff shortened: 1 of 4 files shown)


bentsherman:
Matches the behavior in AWS Batch / ECS / Docker (in Linux):

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definitions

> The number of cpu units the Amazon ECS container agent reserves for the container. On Linux, this parameter maps to CpuShares in the [Create a container](https://docs.docker.com/engine/api/v1.38/#operation/ContainerCreate) section of the [Docker Remote API](https://docs.docker.com/engine/api/v1.38/) and the --cpu-shares option to [docker run](https://docs.docker.com/engine/reference/commandline/run/).

> Linux containers share unallocated CPU units with other containers on the container instance with the same ratio as their allocated amount. For example, assume that you run a single-container task on a single-core instance type with 512 CPU units specified for that container, and that task is the only task running on the container instance. In this example, the container can use the full 1,024 CPU unit share at any given time. However, assume then that you launched another copy of the same task on that container instance. Each task is guaranteed a minimum of 512 CPU units when needed, and each container can float to higher CPU usage if the other container was not using it. However, if both tasks were 100% active all of the time, they are limited to 512 CPU units.

modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
```diff
@@ -507,6 +504,36 @@ class PodSpecBuilder {
 
     }
 
+    @PackageScope
+    @CompileDynamic
```
abhi18av, line 508: Unrelated question: 

In which context is the use of these annotation recommended? 

Generally I have only seen the use of `CompileStatic` but curious about these :)

pditommaso: https://groovy-lang.org/metaprogramming.html#xform-CompileStatic
https://groovy-lang.org/metaprogramming.html#xform-PackageScope

😎

pditommaso:
Yes. It looks like K8s `request` correspondß to `--cpu-shares` and `limit` corresponds to `--cpu-quota` or (`--cpus`). 

https://docs.docker.com/config/containers/resource_constraints/

This is interesting because we are using `--cpus` also in docker. 

https://github.com/nextflow-io/nextflow/blob/b725ddc4a6f9ac89dab45b4931d1b3b30f5ec51b/modules/nextflow/src/main/groovy/nextflow/container/DockerBuilder.groovy#L123-L123

Therefore this should change to `--cpu-shares` as well (but in separate PR)

modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
bentsherman: Abhinav, in this specific case, I used `CompileDynamic` because nested maps are difficult to use under static compilation. There might be some best practice that I don't know about, but it would probably require lots of type annotations, so it was easier to just use dynamic compilation. But in general we try to compile static as much as we can for better performance.

abhi18av: Thanks Paolo and Ben - interesting topic

Is there a way we could profile the overall time taken for the various DSL level transformations?

pditommaso: If you enable the logging using `nextflow -log trace run .. etc` you should have some time reported in the log file

pditommaso (approved):
Ok, just removed the use of `@CompileDynamic`

PR #3343, opened 2022-11-02
Add support for s3 glacier storage class

### LINKED ISSUE #3342: Add support for AWS Glacier
AWS S3 Glacier is long-term "cold" storage tier provided by AWS much cheaper compared to plain AWS S3. 

Also, a S3 policy can be created to automatically move data files stored in S3 to be moved into the Glacier tier automatically after a pre-defined rule.  

The drawback of this solution is that to access data stored in the Glacier tier needs an explicit API call to restore them into a "accessible" state. 

The goal of this ticket is to automate the retrieval process so that the overall process is completely transparent from a pipeline developer/deployer point of view.

bentsherman:
Closes #3342 

This PR adds the ability to download files from S3 Glacier using the `ArchiveTransferManager`. My plan is to check the storage class of an object and use the Glacier transfer manager if the storage class is Glacier, but I haven't figured out the best way to do that.

### DIFF
--- a/plugins/nf-amazon/build.gradle
+++ b/plugins/nf-amazon/build.gradle
@@ -40,7 +40,6 @@ dependencies {
 
     api ('javax.xml.bind:jaxb-api:2.3.1')
     api ('com.amazonaws:aws-java-sdk-s3:1.12.129')
-    api ('com.amazonaws:aws-java-sdk-glacier:1.12.129')
     api ('com.amazonaws:aws-java-sdk-ec2:1.12.129')
     api ('com.amazonaws:aws-java-sdk-batch:1.12.129')
     api ('com.amazonaws:aws-java-sdk-iam:1.12.129')
--- a/plugins/nf-amazon/build.gradle
+++ b/plugins/nf-amazon/build.gradle
@@ -40,6 +40,7 @@ dependencies {
 
     api ('javax.xml.bind:jaxb-api:2.3.1')
     api ('com.amazonaws:aws-java-sdk-s3:1.12.129')
+    api ('com.amazonaws:aws-java-sdk-glacier:1.12.129')
     api ('com.amazonaws:aws-java-sdk-ec2:1.12.129')
     api ('com.amazonaws:aws-java-sdk-batch:1.12.129')
     api ('com.amazonaws:aws-java-sdk-iam:1.12.129')
--- a/plugins/nf-amazon/src/main/com/upplication/s3fs/AmazonS3Client.java
+++ b/plugins/nf-amazon/src/main/com/upplication/s3fs/AmazonS3Client.java
@@ -64,6 +64,10 @@ import com.amazonaws.auth.AWSCredentials;
 import com.amazonaws.auth.AWSStaticCredentialsProvider;
 import com.amazonaws.regions.Region;
 import com.amazonaws.regions.RegionUtils;
+import com.amazonaws.services.glacier.AmazonGlacier;
+import com.amazonaws.services.glacier.AmazonGlacierClientBuilder;
+import com.amazonaws.services.glacier.transfer.ArchiveTransferManager;
+import com.amazonaws.services.glacier.transfer.ArchiveTransferManagerBuilder;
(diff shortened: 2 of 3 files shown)


pditommaso:
> My plan is to check the storage class of an object and use the Glacier transfer manager if the storage class is Glacier, but I haven't figured out the best way to do that.

AFAIK, when accessing an S3 file moved into Glacier, the S3 should return an error response with that information. Nextflow should recognise that and start the download from it. 

The best way to tackle this is to start with a test case. Store a file into Glacier and write unit test to retrieve it

https://github.com/nextflow-io/nextflow/blob/dc871c1d698df986cede23260f97008cd66629d7/plugins/nf-amazon/src/test/com/upplication/s3fs/AwsS3NioTest.groovy#L40-L40

bentsherman:
Made some progress today. I uploaded an object to an S3 bucket with a Glacier storage class:
```bash
aws s3 cp --storage-class GLACIER archive.txt s3://nextflow-ci-dev/archive.txt
```

Then I run the following test pipeline:
```groovy
params.infile = 's3://nextflow-ci-dev/archive.txt'

process test {
  input:
  path('archive.txt')

  output:
  stdout

  """
  echo archive.txt
  """
}

workflow {
  test(params.infile)
}
```

With the original code, you get the following error:
```
com.amazonaws.services.s3.model.AmazonS3Exception: The operation is not valid for the object's storage class (Service: Amazon S3; Status Code: 403; Error Code: InvalidObjectState; Request ID: ZR9TGZJJPCWJ2NQK; S3 Extended Request ID: XZ6eY8N1mlOYMQrAJDZx4DxrUWXucQwsajBBXls/53yiATYp+sWQStmwPtu7GxNX7CrlDjP4kfg=; Proxy: null)
```

So I check the error message for "storage class" and retry with the `ArchiveTransferManager`. Now I get this error:
```
com.amazonaws.services.glacier.model.ResourceNotFoundException: The archive ID was not found: archive.txt (Service: AmazonGlacier; Status Code: 404; Error Code: ResourceNotFoundException; Request ID: 8L1am9dXgMsEVgTneuf-2mG-i0JGKPVMlI_jmiVWbj8NJpI; Proxy: null)
```

I think the problem is that the `ArchiveTransferManager` works for Glacier vaults, but not for regular S3 objects that happen to have a Glacier storage class. If you go to the S3 Glacier page in the AWS console, you will see that "Vaults" are a separate concept. AWS suggests you to use the S3 console for a more "integrated" experience, but I think in that you would need to use `AmazonS3Client::restoreObject()` to "restore" the object before downloading it.

So it looks like I will need to scrap the ArchiveTransferManager and use restoreObject() instead.

bentsherman:
Using restoreObject() seems to be on the right track. Now the problem is that it only initiates the restore, we need to manually wait for completion.

bentsherman:
Okay I added some code to check the status of the restore every 5 seconds. As far as I can tell that is the simplest approach that the API provides. I used the following test pipeline:
```groovy
params.infile = 's3://nextflow-ci-dev/archive.txt'

process test {
  input:
  path('archive.txt')
  output:
  stdout
  """
  echo archive.txt
  """
}

workflow { test(params.infile) }
```

And it is currently waiting for the object restore. I'll try to leave it running for the few hours it should take to restore the object.

There are two new parameters, the polling interval and how long to keep the restored object before putting it back into Glacier. They should probably be exposed to the nextflow config.

I also added a unit test, but it will likely need to be tweaked. Probably should use a different bucket or create a dedicated bucket with a glacier file in it.

bentsherman:
The standard `GLACIER` class has a retrieval time of 3-5 hours. There is also Flexible Retrieval (Expedited) which restores in minutes, and Instant Retrieval which restores in milliseconds. So one of those options is probably better for the CI test, but need to make sure it isn't too expensive. Even with a small file and archival time, some storage classes charge minimums:

https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-class-intro.html#sc-compare

pditommaso:
Ok, merged manually b6110766

PR #3344, opened 2022-11-02
Remove dsl1 and deprecated code

bentsherman:
This PR removes the following:
- DSL1 code
- Deprecated channel factories, operators, and other methods available to Nextflow scripts
- Some deprecated internal code

I tried to keep the core pieces of the DSL version checker, in case we need it in the future. So the `-dsl1` and `-dsl2` command line args and the `nextflow.enable.dsl` script variables are still accepted for backwards compatibility, but they are ignored, and trying to enable DSL1 will cause an error.

Lots of tests are failing, so I still need to go through them, but I wanted to go ahead and show what all I plan to remove. In general, it looks like a lot of unit tests depend on DSL1 or otherwise deprecated functionality, so hopefully this PR when it's ready will make our test suite much more robust.


bentsherman:
```
ERROR: You have an error in your code. Tip: fix the error.
```

pditommaso:
Support for DSL1 have been partially removed via 5d290a47, 14257d49, fa400d5f, f664af45 and 23432f38. 

Still remains some DSL1-specific code that can be better kept to provide a more descriptive error message for users that may still be transitioning from DSL1 to DSL2

Issue #3272, opened 2022-10-06
Processes as operator closures


bentsherman:
Based on this discussion: https://github.com/nextflow-io/nextflow/discussions/2521#discussioncomment-2082904

Processes and operators are functionally equivalent, in that they transform input channels into output channels. The different is that operators execute native Groovy code while processes execute bash scripts. Ultimately a process is just syntax sugar for an operator that executes a shell command, and it makes it much easier to work with files.

Some operators take a closure argument, for example, `reduce`. But, what if the closure (i.e. reduction kernel) was a process? That would make it much easier to perform reductions over files, whereas currently you have to use the experimental recursion feature.

If we could simply make processes interchangeable with operator closures, it would make operators a lot more powerful without the need for an entirely separate feature like recursion (although explicit recursion might still have its use).

pditommaso:
this is an interesting point. we can already chain operators and processes, but this would bring the interaction between the two a bit lower level. 

I think the next step is to identify for which operators it makes sense to allow this, or you are thinking instead a general abstraction?

bentsherman:
I'm going to investigate and see whether a general abstraction would make sense, and if it doesn't then I'll pick a few operators to start with.

bentsherman:
Closing this issue. I don't think this idea will solve the problems I originally thought it would. Instead I am working on a much broader solution to nested dataflow. Stay tuned for more

Issue #3243, opened 2022-09-25
DSL2 pipe syntax enhancements


jemunro:
## New feature

Enhancing the pipe operator to allow processes and workflows with multiple inputs to be piped together.

## Usage scenario 
Given the following processes:
```nextflow
process toUpper {
    input: val x
    output: stdout
    script: "echo -n $x | tr a-z A-Z"
}
process strConcat {
    input: val x; val y
    output: stdout
    script: "echo -n $x; echo -n $y"
}
```
If we want to run out data through `toUpper` and into `strConcat`, the best we can do with piping is the following:
```nextflow
workflow {
    append = channel.value('!')
    strConcat(
        channel.of('foo', 'bar') | toUpper,
        append) |
        view()
}
```
which gives output:
```
N E X T F L O W  ~  version 22.04.5
Launching `works.nf` [zen_mclean] DSL2 - revision: 2f1ccf68d9
executor >  local (4)
[68/eb8b8d] process > toUpper (2)   [100%] 2 of 2 ✔
[e9/7fb402] process > strConcat (2) [100%] 2 of 2 ✔
FOO!
BAR!
```
## Suggest implementation 
The following would be more intuitive and readable:
```nextflow
workflow {
    append = channel.value('!')
    channel.of('foo', 'bar') |
        toUpper() | 
        strConcat(append) |
        view()
}
```
Attempting to run the above currently gives the following error:
```
Process `toUpper` declares 1 input channel but 0 were specified
 -- Check script 'main.nf' at line: 17 or see '.nextflow.log' file for more details
```
if we replace `toUpper() |` with  `toUpper |`, the error becomes:
```
Process `strConcat` declares 2 input channels but 1 were specified
``` 

This could be achieved by making the first argument or a process be the piped value (regardless of whether parenthesis are present). This already seems to be the case for operators that accept multiple arguments, e.g.:
```nextflow
channel.of('foo', 'bar') |
        mix(channel.of('baz')) |
        view
```

bentsherman:
Basically you want to be able to curry channel arguments. I think it would work for processes but not for operators, because operators use an object-oriented syntax, e.g. `a | op(b)` is equivalent to `a.op(b)`. It's an interesting idea.

Also, the example with `mix` works because mix can take any number of arguments.

jemunro:
As far as I can tell this already works for most operators, just not for processes.

For example the following works fine:
```nextflow
workflow {
[...]

PR #3356, opened 2022-11-07
exec and eval operators

jemunro:
# New operators `eval` and `exec`
This PR is intended to start a discussion about the potential usefulness of the proposed operators, but also provides an implmentation that could be built upon.

These operators are non-standard in the sense that they don't transform the contents of a  channel but provide a tool for programming with channels, similarly to `set`. These were inspired based on discussion at https://github.com/nextflow-io/nextflow/issues/3243, and are somewhat relevant to https://github.com/nextflow-io/nextflow/issues/3272. 

The main advantage is that they facilitate piping/chaining and result in fewer intermediatate channel variables being assigned.

## `eval`
Evaluate a closure with the Channel or Multi-channel (as output by branch, multimap or processes with multiple outputs) source as delegate and return the result. In the case of named multi-channel sources the name can be used directly. This works similarly to Groovy's `with()` function.

## `exec`
Execute a process using the source channel as first input. Multi-channel inputs are unpacked into subsequent inputs, following by any arguments given to the operator.

## Examples
All examples use the following processes:
```nextflow
process TO_UPPER {
    input: val x
    output: stdout
    script: "echo -n $x | tr a-z A-Z"
}

process CONCATENATE {
    input: val x; val y
    output: stdout; val(x)
    script: "echo -n $x; echo -n $y"
}
```

All examples output the same result_ch:
```
[FOO!, BAR!]
[FOO!, BAZ!]
```

#### Without `eval` and `exec`
```nextflow
workflow {

    append_ch = Channel.value('!')

    upper_ch = TO_UPPER(channel.of('foo', 'bar', 'baz'))
    
    concat_ch = CONCATENATE(upper_ch, append_ch)
    
    branched_ch = concat_ch[0]
        .branch { 
            foo: it == 'FOO!'
            other: true }
    
    result_ch = branched_ch
        .foo
        .combine(branched_ch.other)
    
    result_ch.view()
}
```
#### With `eval` and `exec`
```nextflow
workflow {

    append_ch = Channel.value('!')

    result_ch = channel.of('foo', 'bar', 'baz')
        .exec(TO_UPPER)
        .exec(CONCATENATE, append_ch) 
        .eval { it[0] }
        .branch { 
            foo: it == 'FOO!'
            other: true }
        .eval { foo.combine(other) }

    result_ch.view()
}
```

#### Without `eval` and `exec`
```nextflow
workflow {

    append_ch = Channel.value('!')

    upper_ch = channel.of('foo', 'bar', 'baz') \
        | TO_UPPER
        
    concat_ch = CONCATENATE(upper_ch, append_ch)

    branched_ch = concat_ch[0] \
        | branch { 
            foo: it == 'FOO!'
            other: true }

    result_ch = branched_ch
        .foo \
        | combine(branched_ch.other)
    
    result_ch | view
}
```
#### With `eval` and `exec`
```nextflow
workflow {

    append_ch = Channel.value('!')

    result_ch = channel.of('foo', 'bar', 'baz') \
        | TO_UPPER \
        | exec(CONCATENATE, append_ch) \
        | eval { it[0] } \
        | branch { 
            foo: it == 'FOO!'
            other: true } \
        | eval { foo | combine(other) }

    result_ch | view()
}
```


pditommaso:
This sounds interesting. 

Regarding `eval` if i'm understanding correctly it takes a multi-channel and map it to a channel. Begin so it looks like the reverse of [multiMap](https://www.nextflow.io/docs/latest/operator.html#multimap). is that correct? 

In relation to `exec`, it looks like the main use case it to allow a more fluent composition of processes when there's output/input cardinality mismatch. 

In this case, it would be even nicer to allow of a special input keyword. following your example, something like 

```
channel.of('foo', 'bar', 'baz') \
        | TO_UPPER \
        | CONCATENATE(-, append_ch) \
```

where `-` should represent the channel resulting by the chain operation. This would be more readable than the `exec` proposal. Above all, it would allow more control over *which* process input to be used, for example, it also is possible to use as `CONCATENATE(append_ch, -)`.

Not sure about the feasibility to use `-` in practical terms. It may have some special meaning in the underlying groovy parser.

jemunro:
> This sounds interesting.
> 
> Regarding `eval` if i'm understanding correctly it takes a multi-channel and map it to a channel. Begin so it looks like the reverse of [multiMap](https://www.nextflow.io/docs/latest/operator.html#multimap). is that correct?

The primary use case of of `eval` would be as you say, to take multi-channels and return a single channel after applying some functions/operators/processes, in a way that would work nicely with the pipe operator. For example:

1. Using named multi-channel
    ```nextflow
    Channel.of(['A', 1], ['B', 3], ['C', 4]) \
        | branch { a: it[0] == 'A'; other: true } \
        | eval { a.combine(other) }
        | view
    ```
    ```
    ['A', 1, 'B', 3]
    ['A', 1, 'C', 4]
    ```
    Equivalent to:
    ```nextflow
    ch = Channel.of(['A', 1], ['B', 3], ['C', 4]) \
        | branch { a: it[0] == 'A'; other: true }
    ch.a | combine(ch.other) | view
    ```
    
2. Using unamed multi-channel
    ```nextflow
    process IN_1_OUT_2 {
        input:
        tuple val(x), val(y)

        output: 
        val(x)
        val(y)

        exec: null
    }
    workflow {
        Channel.of(['A', 1]) \
            | IN_1_OUT_2 \
            | eval { first() } \
            | view
    }
    ```
    ```
    'A'
    ```
    equivalent to:
    ```nextflow
    workflow {
        Channel.of(['A', 1]) \
            | IN_1_OUT_2
        IN_1_OUT_2.out.first().view()
    }
    ```
3. Flexible process call
    ```nextflow
    append_ch = Channel.value('!')
    channel.of('foo', 'bar') \
        | eval { CONCATENATE(it, append_ch) } \
        | view
    ```
    ```
    'foo!'
    'bar!'
    ```
    equivalent to:
    ```nextflow
    append_ch = Channel.value('!')
    input_ch = channel.of('foo', 'bar')
    CONCATENATE(input_ch, append_ch) \
        | view
    ```

jemunro:
> In relation to `exec`, it looks like the main use case it to allow a more fluent composition of processes when there's output/input cardinality mismatch.
> 
> In this case, it would be even nicer to allow of a special input keyword. following your example, something like
> 
> ```
> channel.of('foo', 'bar', 'baz') \
>         | TO_UPPER \
>         | CONCATENATE(-, append_ch) \
> ```
> 
> where `-` should represent the channel resulting by the chain operation. This would be more readable than the `exec` proposal. Above all, it would allow more control over _which_ process input to be used, for example, it also is possible to use as `CONCATENATE(append_ch, -)`.
> 
> Not sure about the feasibility to use `-` in practical terms. It may have some special meaning in the underlying groovy parser.

This proposal sounds really nice, but as you say '-' won't work due to the groovy parser. Alternatives could be:
1. Closure
    ```nextflow
        | CONCATENATE { it; append } \
    ```
1. Special reserved variable, e.g. '_'
    ```nextflow
        | CONCATENATE(_, append) \
    ```
But I think both would be harder to implement than an operator such as `exec`, because it would need a rewrite of how Nextflow handles process calls.

jemunro:
> I think `eval` is useful, tho the name is too generic. It should be found something more specific.

I agree it is generic, though I struggle to think of a less generic name. Maybe `chain`? Since it helps to chain more operations together? E.g.:
```nextflow
Channel.of(['A', 1], ['B', 3], ['C', 4]) \
    | branch { a: it[0] == 'A'; other: true } \
    | chain { a.combine(other) } \
    | view
```

> 
> Using `_` instead `-` should be that hard. I would expect something similar to what is currently done for operation chaining in which the left channel is added as a source, see here
> 
> https://github.com/nextflow-io/nextflow/blob/c2f50d74051921f6b033b99e82b363e7f11099db/modules/nextflow/src/main/groovy/nextflow/extension/ChannelEx.groovy#L169-L169
> 
> I think it's worth to give it a try

How does this sound for an implementation approach:
1. Create a class `sourcePlaceholder` which will be assign to `_` in the script binding
2. Add a new boolean field to `processDef` `isPartial`
3. Modify `processDef.run()` at the following section, adding a check for instances of `sourcePlaceholder` in the `params`
https://github.com/nextflow-io/nextflow/blob/76a87c72ed0204b901a613b295fc3202f475cf21/modules/nextflow/src/main/groovy/nextflow/script/ProcessDef.groovy#L167-L180
4. If a `sourcePlaceholder` is present, instead set `isPartial = true` and `return this`. Then the next time `processDef.run()` is called, replace `sourcePlaceholder` with `args`

pditommaso:
it sounds like a plan!

PR #3371, opened 2022-11-10
Improve default value in cli help of `nextflow log -s`

mribeirodantas:
I'm not sure TAB is the best, but we definitely have to think of something
better than \t, that shows up as blank space in the CLI help.

To make it more clear what I mean:

Screenshot 2022-11-10 at 11 27 02

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdLog.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdLog.groovy
@@ -61,7 +61,7 @@ class CmdLog extends CmdBase implements CacheBase {
     static final public NAME = 'log'
 
     @Parameter(names = ['-s'], description='Character used to separate column values')
-    String sep = '\t'
+    String sep = 'TAB'
 
     @Parameter(names=['-f','-fields'], description = 'Comma separated list of fields to include in the printed log -- Use the `-l` option to show the list of available fields')
     String fields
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/cli/CmdLog.groovy
```diff
@@ -61,7 +61,7 @@ class CmdLog extends CmdBase implements CacheBase {
     static final public NAME = 'log'
 
     @Parameter(names = ['-s'], description='Character used to separate column values')
-    String sep = '\t'
+    String sep = 'TAB'
```
pditommaso, line 65: Not understanding this change

mribeirodantas: Have a look at the print screen that I attached to the pull request body text. Even though in the source code it's clearly a tab symbol (`\t`), in the CLI nothing appears as the default value. It could be a number of different things. My change tells the user it's a tab. I don't know if this is the best solution, but we should think on something to put there instead of the current solution.

Type `nextflow log -h` and you'll see it's not clear what's the default value for the `-s` option.

mribeirodantas: Maybe escaping the `\`? `\\t`?

mribeirodantas: Oh, only now I realized. This is probably used somewhere, not only for the `-h` of the command line, right? 😅 My bad. How can we solve this then?

pditommaso: ahh, now understand yes `\\t`!

modules/nextflow/src/main/groovy/nextflow/cli/CmdLog.groovy
```diff
@@ -61,7 +61,7 @@ class CmdLog extends CmdBase implements CacheBase {
...
```
pditommaso, line 64:
```suggestion
    String sep = '\\t'
```

PR #3383, opened 2022-11-13
Container cpu shares

pditommaso:
This PR implements the use of cpu-shares for container CPUs allocation instead of Docker `--cpus` option. 

This allows better utilization of CPUs resources and makes consistent with the semantic of cpus request with other schedulers such as AWS Batch and Kuberneters. 

See also [here](https://www.batey.info/cgroup-cpu-shares-for-kubernetes.html) and #3338

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/container/ContainerBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/ContainerBuilder.groovy
@@ -59,7 +59,7 @@ abstract class ContainerBuilder<V extends ContainerBuilder> {
 
     protected List<String> engineOptions = []
 
-    protected Float cpus
+    protected Integer cpus
 
     protected String cpuset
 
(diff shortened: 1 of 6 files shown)


abhi18av:
Adding the comments here from Azure perspective - the core options from the Azure perspective are implemented via the following snippet

https://github.com/nextflow-io/nextflow/blob/master/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L385-L399 

And the [Azure Batch docs state that](https://learn.microsoft.com/en-us/azure/batch/batch-docker-container-workloads#container-task-command-line) anything [supported by `docker create`](https://docs.docker.com/engine/reference/commandline/create/) can be provided as part of the `containerOpts`, including the `--cpu-shares` option ✅

The relevant options (used within a job) are displayed on the Azure portal as such:

image

bentsherman (approved):

bentsherman:
Looks correct, however I think you should review [this page](https://www.batey.info/cgroup-cpu-shares-for-docker.html) rather than the one you linked, because this one explains how CPU shares work with Docker rather than Kubernetes. The behavior seems to be slightly different.

PR #3384, opened 2022-11-13
Fix support for remote file mail attachment

pditommaso:
This PR fixes the support for remote paths in the mail attachment. 

When a remote path is provided, the file is copied in the local file system to a temp file. 


pditommaso:
Should it be added a warning if the user try to attach a file too big e..g larger than 20MB? @robsyme what do you think?

pditommaso:
Ok, I've added a warning if the attachment is bigger than 20MB. merging!

PR #3386, opened 2022-11-14
include all operators and functions from a plugin

jorgeaguileraseqera:
This PR implements a shortcut to include all functions and operators from a plugin 

I couldn't find the way to use the `*` character so I use `_` 

### DIFF
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -101,3 +101,6 @@ In the same way as operators, functions can be aliased::
 
     include { reverseString as anotherReverseMethod } from 'plugin/my-plugin'
 
+As of version `22.11.0` you can include all operatos and functions a plugin exports using the '_' expression
+
+    include { _ } from 'plugin/my-plugin'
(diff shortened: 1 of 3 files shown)


jorgeaguileraseqera:
this feature can be confusing for users as it can be no clear where functions or operators come from

abhi18av:
I understand that the small benefit might not be worth with engineering and maintenance effort but just for the sake for tracking, I'm outlining the idea I mentioned - [Importing module with prefix](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/import-module?view=powershell-7.3#example-6-import-the-members-of-a-module-and-add-a-prefix) in powershell to avoid name collisions.


Therefore, the idea is that people can import _all_ the functionality (operator / functions) from a plugin using a `prefix`, which dynamically updates the names of all operators/functions. For example


```nextflow
include { _ } from 'plugin/nf-test-plugin-hello' with prefix "test-"

// Instead of sayHello

test-sayHello()
```

This would allow people to iterate faster by including all operators/functions and then cherry-pick the ones they need.

I guess that there might be a benefit with this approach, given that we don't have an IDE or a language server for Nextflow to rely upon.

pditommaso:
Ok, now understand it. It's a kind of namespace mechanism. It's a good idea but I still think it's better to keep a simple and predictable mechanism to import explicitly one by one. At the end rarely are more 4-5 imports per plugin.

PR #3392, opened 2022-11-15
add support for mpijobs via kubeflow operator

xhejtman:
draft implementation of mpijobs. It expects that mpioperator and crd is installed. Example configuration:
```groovy
k8s {
   namespace = 'test-ns'
   runAsUser = 1000
   computeResourceType = 'Job'
   pullPolicy          = 'Always'

}

executor {
  queueSize = 100
}

process {
   time = '24h'
   executor = 'k8s'
   
   pod = [
      [securityContext:[fsGroupChangePolicy:'OnRootMismatch', runAsUser:1000, runAsGroup:1, fsGroup:1]],
      [automountServiceAccountToken:false],
      [mpiJobWorkers: 2],
      [sshAuthMountPath: '/tmp/a'],
      [computeResourceType: 'MPIJob']
   ]
}
```

mpijob is not implemented for k8sdriver as it doubtly runs mpijob directly. it also adds possibility to specify computeresourcetype per task. 

sshAuthMountPath is needed for mpioperator, where to put ssh authorized keys. 

basically, it creates mpijob resource which creates standard job so we can utilize existing job implementation. It is not tested, just hello example passes using launcher jobs.


bentsherman:
Thanks @xhejtman , and yes don't worry about the k8s driver launcher. The implementation looks good so far.

xhejtman:
> Thanks @xhejtman , and yes don't worry about the k8s driver launcher. The implementation looks good so far.

perhaps code around `useJobResource()` should be improved somehow.

bentsherman:
I was going to say, in cases where you are checking all three cases of pod/job/mpijob, just use a switch-case and don't use `useJobResource()`. And if it's not used anywhere else then maybe we can remove it.

modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
```diff
@@ -412,6 +412,10 @@ class TaskConfig extends LazyMap implements Cloneable {
         new PodOptions((List)get('pod'))
     }
 
+    String getComputeResourceType() {
+        get('computeResourceType') as String
+    }
```
pditommaso, line 417: Is this really necessary? I'd avoid to introduce a process directive specific for this

xhejtman: so you want to move it from process to pod options?

pditommaso: think, it makes more sense

xhejtman:
so, reworkerd a bit, removed process option to pod options. if it ok, I will fix tests and add few for mpijob. also I think it would be really useful to provide complete examle e.g., hello-mpi including docker container..

xhejtman:
Example how to use it: https://github.com/CERIT-SC/nf-pi-mpi

Btw, there is one remaining issue. mpioperator creates launcher and workers at once and launcher immediatelly runs `mpirun` command which can fail, if workers are not scheduled, and this is actually happening. kubeflow expects that launcher does several restarts (with error status/crashloopback) before MPI world is established. However, nextflow evaulates this situation as error and terminates execution. Moreover, unavailability of worker is not only due to asynchronicity but also because there are no free resources which can take even several days, not just few restarts.

I believe this could be fixed at mpioperator side, e.g., ensuring all workers are running and then and only then start launcher, which on the other hand breaks the current waiting logic of mpiJobState(). Perhaps instead of wait cycle, should it return just empty status Map? As in the case of pending Pod?

pditommaso:
I agree that would be important to add an example.

> I believe this could be fixed at mpioperator side, e.g., ensuring all workers are running and then and only then start launcher, which on the other hand breaks the current waiting logic of mpiJobState(). Perhaps instead of wait cycle, should it return just empty status Map? As in the case of pending Pod?

I think the best would be to fix at mpioperator side as you are suggesting. Otherwise using an empty map as for the pod should be good. But what if the mpioperator never reach the desired state? would not this cause the execution to stall endlessly?

xhejtman:
Can we proceed here?

xhejtman:
> Looks like some CI tests are failing. Can you update them or do you need help?

They are fixed

modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy
```diff
@@ -253,8 +253,8 @@ class K8sTaskHandler extends TaskHandler implements FusionAwareTask {
         }
 
         return useJobResource()
-            ? builder.buildAsJob()
-            : builder.build()
+             ? builder.buildAsJob()
+             : builder.build()
```
bentsherman, line 257:
```suggestion
            ? builder.buildAsJob()
            : builder.build()
```

vsoch:
hey folks! This looks interesting - if/when it's merged I'd be interested to try it out / and possibly add the FluxOperator. We did some experiments comparing the two if you are interested - will be a talk at Kubecon https://sched.co/1HyaG.

In the meantime I'll watch this issue - I'm super happy to see something like MPI potentially entering the Nextflow space!

bentsherman:
Thank you for patience. We haven't forgotten about this one, just focused on other efforts. I'm going to clean up the routing of pod/job/mpijob since it is now a mess. I think we should also move the options specific to MPIJob into a namespace within the pod options, so you would have e.g. `pod.mpi.workers`, `pod.mpi.sshAuthMountPath`, etc. I will try to finish it next week.

vsoch:
hey @bentsherman that's OK! I can't comment for @xhejtman, but for my comment above, you haven't seen the Flux Operator talk yet (it's next week, woot!) so you should at least watch that and then be convinced if we want to extend the work here to be more generalized for running an operator. The Flux Operator has a more efficient boostrap for the workers and doesn't require an extra launcher node, so it would be of interest for MPI. But you can use the Flux Operator for any kind of job you need to run - it's like spinning up your own on demand HPC cluster inside of Kubernetes (and it doesn't require MPI but works very well for it!(

vsoch:
Yeah that sounds perfect! You might be interested in a csi driver I started too https://github.com/converged-computing/oras-csi you wouldn’t need to build a container every time to add fusion.

xhejtman:
> hey @bentsherman that's OK! I can't comment for @xhejtman, but for my comment above, you haven't seen the Flux Operator talk yet (it's next week, woot!) so you should at least watch that and then be convinced if we want to extend the work here to be more generalized for running an operator. The Flux Operator has a more efficient boostrap for the workers and doesn't require an extra launcher node, so it would be of interest for MPI. But you can use the Flux Operator for any kind of job you need to run - it's like spinning up your own on demand HPC cluster inside of Kubernetes (and it doesn't require MPI but works very well for it!)

Does it create ssh mesh? I.e., for mpirun, you need ssh accessible members of the cluster with pre-set ssh keys. Does it work here?

vsoch:
The ssh bootstrap that the MPI operator uses is actually less efficient than the zeromq bootstrap that the Flux Operator uses. You can watch my talk at Kubecon on Friday (or ask me after and I'll send a video link) to see our experiments that suggest this.

modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy
```diff
@@ -398,6 +435,25 @@ class K8sClient {
...
+	    } 
+            catch (K8sResponseException err) {
+                if( err.response.code != 404 ) {
+	            throw err
+		} else {
+		    final long delay = (Math.pow(3, i - 1) as long) * 250
+                    sleep( delay )
+		}
+            }
+        }
```
bentsherman, line 454: Maybe we can fallback to getting the MPIJob state like we do with Jobs.

modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy
```diff
@@ -462,9 +486,11 @@ class K8sTaskHandler extends TaskHandler implements FusionAwareTask {
         
         if( podName ) {
             log.trace "[K8s] deleting ${resourceType.lower()} name=$podName"
-            if ( useJobResource() )
+            if ( useJobResource() ) {
                 client.jobDelete(podName)
-            else
+                if ( resourceType == ResourceType.MPIJob )
+                    client.mpiJobDelete(podName.minus("-launcher"))
```
bentsherman, line 492: @xhejtman trying to understand this code... does this code mean that an MPIJob creates a separate Job for the launcher, rather than a pod? Also, does deleting the MPIJob not also delete the underlying pods / jobs?

xhejtman: yes, it creates a job for launcher with hardcoded suffix `-launcher`

vsoch: Just for a heads up - the Flux Operator doesn't have/need a launcher node.

bentsherman:
Alright folks, I finished the refactoring. Here are the new changes:

- Add `ResourceProvider` interface, which defines API operations for a K8s resource.

- Move client code from `K8sClient` to implementations of `ResourceProvider` for Pod, Job, and MPIJob. The `K8sClient` now maintains a static registry of API providers, and it routes each request to the appropriate provider based on the given `type` parameter.

- Add `ResourceSpecBuilder` interface, which defines the YAML spec builder for a K8s resource.

- `PodSpecBuilder` now implements `ResourceSpecBuilder`, with separate subclasses for Job and MPIJob.

- `K8sTaskHandler` maintains a static registry of spec builders, and for a given task it uses the correct spec builder for the specified resource type in `pod` directive or `k8s` config.

- Move MPIJob config options to `mpi` option in `pod` directive, e.g.:
  ```groovy
  pod = [mpi: [
    workers: 1,
    sshAuthMountPath: '...',
    waitForWorkers: true
  ]]
  ```

So, to define a new resource like MPIJob, you have to:
1. implement `ResourceProvider` to define the relevant API operations
2. implement `ResourceSpecBuilder` to define how to build the YAML spec

In both cases you can subclass the existing implementations for Pod and Job to reuse code, which I do here for MPIJob. What I would like to do is merge this code without MPIJob and then implement the MPIJob support in a plugin, e.g. `nf-mpijob` or `nf-kubeflow`. So these changes are meant to demonstrate what it would look like. This setup should also support the Flux Operator, although we won't know for certain until we try.

The only lingering issue is the MPIJob config options. Normally we would put them in their own config scope, but these are currently process-specific. I see two possible solutions:
- Put these options in a config scope (e.g. `mpijob`). They will apply to all processes, but they only apply to mpi jobs anyway. Although different mpi processes wouldn't be able to have different number of workers.
- Add a catch-all mechanism to the `pod` directive (like `**kwargs` in Python) and let the MPIJob spec builder inspect those options for what it needs.

@pditommaso let me know what you think about the overall K8s changes.

@xhejtman let me know what you think about how to handle the extra options.

modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy
bentsherman: but does it not delete the launcher automatically when you delete the mpi job?

bentsherman:
@xhejtman The `pod` directive now puts all unmatched options into a map that can be accessed by the MPIJob spec builder. So you would do something like htis:
```groovy
pod = [
    [mpi: [
        workers: 2,
        sshAuthMountPath: '/ssh/path',
        waitForWorkers: true
    ]]
]
```

Please test the PR in your environment and let me know if it works with MPIJobs. If it works, we'll move the MPIJob changes into a plugin.

vsoch:
Another question - why not try a more general approach like using kueue (that supports the MPI operator?) and will support the Flux Operator soon too :)

bentsherman:
My understanding is that Kueue uses K8s Jobs, so it should work out of the box with Nextflow. We're planning to try it soon (also KEDA). If you say it's going to support the MPI Operator and Flux, then in theory a Nextflow user could Kueue for everything. But I don't know how all of that is supported through the Job spec, or how we would pass these MPi options.

vsoch:
Would be happy to! I will likely test out the kueue implementation for the flux operator first, and then can open the issue to comment when I better understand how it works.

bentsherman:
closing due to lack of activity, feel free to re-open if you're still interested

PR #3396, opened 2022-11-16
Store error to `task.error`

scottgigante-immunai:
`task.error` appears to be always null, even on retry. This PR gives users access to e.g. `task.error.message` to inform retry strategy.

Signed-off-by: Scott Gigante <84813314+scottgigante-immunai@users.noreply.github.com>

cc @pditommaso

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -968,6 +968,7 @@ class TaskProcessor {
         ErrorStrategy errorStrategy = TERMINATE
         final message = []
         try {
+            task.error = error
             // -- do not recoverable error, just re-throw it
             if( error instanceof Error ) throw error
 


pditommaso:
Thanks for this PR, but it is actually used in several places. I'm closing this. Feel free to comment if you think this change is necessary 

Screenshot 2022-12-04 at 19 37 41

scottgigante-immunai:
Perhaps I am misunderstanding how this is used, but I can tell you it was always `null` when I printed it in computing resources for a re-attempted task. Example of what I did which printed null every time: https://github.com/openproblems-bio/nf-openproblems/blob/d12cd13df23ac000392c9f0339bf112c6e959f63/nextflow.config#L168

pditommaso:
That field is mainly used to report errors in tasks executing native groovy code or some other specific error conditions. Therefore I think it's better to not alter the current logic without a clear evidence of the problem to be solved.

scottgigante-immunai:
The concrete problem I am trying to solve is [this](https://nextflow.slack.com/archives/C02T98A23U7/p1668617967873289) -- I am unable to access the "Caused by" message when determing re-attempt compute resources. In this case the `task.error.cause` should be `"Essential container in task exited - OutOfMemoryError: Container killed due to memory usage"` while `task.exitStatus == 1`. Since `task.error is null`, I am unable to access this error message at runtime and therefore can't tell if this task timed out, ran out of memory, or failed in some other way (and thus can't use this information to decide whether to retry with more memory, more CPUs, or just give up.)

Issue #3166, opened 2022-08-31
Google Batch should comply with --Error is ignored flag


zyosufzai:
## Bug report 

Google Batch stops running even though nextflow has supplied a --Error is ignored flag

### Expected behavior and actual behavior

Google Batch should continue running even if one of the jobs fails if it has been told to ignore the error. But what actually happens is the workflow terminates with a error message (see below)

### Steps to reproduce the problem

Input the following on the command line:
```
NXF_VER="22.08.1-edge" ./nextflow run nf-core/methylseq -r 1.6.1 -c nextflow.config -profile test,gbatch

```

with the following config file:
```
gbatch{ 
      process.executor = 'google-batch' 
      process.machineType = 'n2-standard-16' 
      process.time = '2h' 
      workDir = 'gs://nextflowdemobucket/zy-test/testrna_gbatch_tmp' 
      google.location = 'us-central1' 
      google.region  = 'us-central1' 
      google.project = ''
      params.outdir = 'gs://nextflowdemobucket/zy-test/testrna_gbatch'
      google.batch.bootDiskSize = 100.GB
      }
```

### Program output 

```ruby
[2e/9436f4] NOTE: Process preseq (SRR389222_sub1) terminated with an error exit status (139) -- Error is ignored
Error executing process > 'preseq (SRR389222_sub2)'
```

```ruby
Caused by:
  Task java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@7426a203[Not completed, task = java.util.concurrent.Executors$RunnableAdapter@6126b8c[Wrapped task = TrustedListenableFutureTask@d542ca6[status=PENDING, info=[task=[running=[NOT STARTED YET], com.google.api.gax.rpc.AttemptCallable@464ab2c1]]]]] rejected from java.util.concurrent.ScheduledThreadPoolExecutor@61fde900[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 0]
```

### Environment 

* Nextflow version: 22.08.1-edge 
* Java version: 17.0.3-internal
* Operating system: Linux
* Bash version: 5.0.3(1)-release
### Additional context

(Add any other context about the problem here)

bentsherman:
Can you provide the full log? There are two different tasks, `SRR389222_sub1` and `SRR389222_sub2`, and the first one is ignored but the second one is triggering workflow termination. Need to see what happened with the second one.

zyosufzai:
[...]

PR #3401, opened 2022-11-17
Ignore Exit status for Runnables in Google Batch

hnawar:
Ignore Exit status for runnables allows Errors to be ignored when specified in the workflow

Signed-off-by: Hatem Nawar <hnawar@google.com>

### DIFF
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
@@ -165,7 +165,8 @@ class GoogleBatchTaskHandler extends TaskHandler {
             .setComputeResource(computeResource)
             .addRunnables(
                 Runnable.newBuilder()
-                    .setContainer(container)
+                        .setContainer(container)
+                        .setIgnoreExitStatus(true)
             )
             .addAllVolumes( launcher.getVolumes() )
 
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
@@ -165,8 +165,8 @@ class GoogleBatchTaskHandler extends TaskHandler {
             .setComputeResource(computeResource)
             .addRunnables(
                 Runnable.newBuilder()
-                        .setContainer(container)
-                        .setIgnoreExitStatus(true)
+                    .setContainer(container)
+                    .setIgnoreExitStatus(true)
             )
             .addAllVolumes( launcher.getVolumes() )
 


hnawar:
#3166

pditommaso:
Hi Hatem, thanks for looking into it. However, I'm not getting how this patch is related to #3166. 

If I'm understanding correctly `setIgnoreExitStatus(true)` applied to child jobs and makes them to not fail when the container task reports a non-zero exit status. 

Instead, the exception in #3166 is reported by the Nextflow runtime? how the ignore flag should affect it?

aaronegolden:
Setting ignoreExitStatus(true) on the Runnable will cause Batch to ignore the exit status of that specific Runnable. If there are multiple Runnables in the Task then it means that Runnable can fail and Batch will just go along and execute the next Runnable. 

In the scenario from #3166, if I understand correctly, there is a Batch Job created for the "preseq (SRR389222_sub1)" NextFlow process, and that Job has a single Task, and that Task has a single Runnable. So if ignoreExitStatus is set on that Runnable then it really means the Job will never fail (unless something goes wrong so early in the process that the Job never runs at all, e.g. the VM can't be provisioned in the first place).

It sounds like what is actually desired for #3166 is for the Batch Job to be allowed to fail, but then for the NextFlow workflow to carry on in spite of that failure.

hnawar:
Would refactoring to have 3 runnables (Staging, Main, Unstaging) with IgnoreExitStatus set to true for the main runnable be a better solution?

The bash script should capture the exit status of the user script/command and the unstaging runnable should still be able to run to copy partial output,logs, and other artifacts

aaronegolden:
I think the Batch ignoreExitStatus field is a red herring. Life Sciences also has that field, and it's not used for the preseq process in #3166. This same process fails when using the Life Sciences backend, but the NextFlow workflow carries on because that part of the workflow uses errorStrategy "IGNORE".

https://github.com/nf-core/methylseq/blob/118077265374f8bd76f60472cc1ab21d116c2a1a/conf/base.config#L54
https://github.com/nf-core/methylseq/blob/master/modules/nf-core/preseq/lcextrap/main.nf#L4

Whether I run the methylseq workflow with executor google-lifesciences or executor google-batch, either way I see a message in my .nextflow.log saying that the preseq process failed, and that the error was ignored, like:

"NOTE: Process `preseq (SRR389222_sub2)` terminated with an error exit status (1) -- Error is ignored"

The difference is, when executor=google-lifesciences the error appears to *actually* be ignored, whereas when executor=google-batch the workflow terminates.

So I think that any changes to the Batch Job submitted by NextFlow for the preseq process, causing the Job to not fail, would just be workarounds. The Job should be allowed to fail, and NextFlow should still complete the workflow due to that process using errorStrategy = 'ignore', as it does when the equivalent Pipeline fails when using the google-lifesciences executor.

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -165,7 +165,8 @@ class GoogleBatchTaskHandler extends TaskHandler {
             .setComputeResource(computeResource)
             .addRunnables(
                 Runnable.newBuilder()
-                    .setContainer(container)
+                        .setContainer(container)
+                        .setIgnoreExitStatus(true)
```
bentsherman, line 169:
```suggestion
                    .setContainer(container)
                    .setIgnoreExitStatus(true)
```

PR #3408, opened 2022-11-18
Clarify limitation of `-with-docker` in the docs [ci skip]

mribeirodantas:
The `-with-docker` CLI option has limitations, when compared to using `docker.enabled` in the `nextflow.config` file.
With `-with-docker`, you either have to provide a container image as an argument in the -with-docker in the CLI, or in
the nextflow.config with the process.container directive, but it has to be the same container image for all processes,
and none can run outside a Docker container. You can’t use selectors here, you really need the process.container.
However, if you do everything in the nextflow.config file (without -with-docker), you can have not only different processes
running in different container images, but you can also have part of the processes running outside containers.

This PR tries to clarify that.


mribeirodantas:
👀 😬

mribeirodantas:
LGTM, @pditommaso ? 😄

docs/container.rst
```diff
@@ -172,9 +172,10 @@ your pipeline execution.
 Multiple containers
 -------------------
 
-It is possible to specify a different Docker image for each process definition in your pipeline script. Let's
-suppose you have two processes named ``foo`` and ``bar``. You can specify two different Docker images for them
-in the Nextflow script as shown below::
+It is possible to specify a different Docker image for each process definition in your pipeline script. However, this
+can't be done with `-with-docker`, since it doesn't support process selectors in the `nextflow.config` file, and doesn't
+check for the `container` process directive in the pipeline script file. Let's suppose you have two processes named ``foo``
+and ``bar``. You can specify two different Docker images for them in the Nextflow script as shown below::
```
pditommaso, line 178:
```suggestion
It is possible to specify a different Docker image for each process definition in your pipeline script. However, this
can't be done with ``-with-docker`` command line option, since it doesn't support process selectors in the ``nextflow.config`` file, and doesn't
check for the ``container`` process directive in the pipeline script file. Let's suppose you have two processes named ``foo``
and ``bar``. You can specify two different Docker images for them in the Nextflow script as shown below::
```

pditommaso (approved):

pditommaso:
yes

Issue #3403, opened 2022-11-17
Flux Framework executor support


vsoch:
## Usage scenario 

I'd like to propose integration with [Flux Framework](https://flux-framework.org/), which is (on a high level) a job manager akin to slurm or similar. However, it also can be launched in slurm or cloud, so it's quite useful! 

My reasons are a bit selfish - I've already added to reframe and Snakemake, and it's because I'm developing the [Flux Operator](https://flux-framework.org/flux-operator/getting_started/custom-resource-definition.html) (a Kubernetes operator) and I'm interesting in being able to show a nextflow workflow running! :partying_face: As we transition to using Flux for the de-facto scheduler (at least the national labs are going this route and I suspect others might follow) I'd like Nextflow users to be able to still share/use their workflows.

## Suggest implementation 

I believe it would be on the level of an executor: https://www.nextflow.io/docs/latest/executor.html. I don't know much about Groovy or developing NextFlow operators, but probably with some guidance I could at least take a first shot! Thank you!

pditommaso:
This is very welcome, Flux was already mentioned in the past by @Nortamo, great to see this proposal moving on. Also, the Flux operator project looks very interesting. 

I think adding the support for Flux to Nextflow should be not too different from the [SlurmExecutor](https://github.com/nextflow-io/nextflow/blob/daf066ce5df6ad90985d26dd3808dbd94545153e/modules/nextflow/src/main/groovy/nextflow/executor/SlurmExecutor.groovy). 

Essentially, the most important thing to know are the Flux equivalent for `sbatch`, `scancel` and `squeue` respectively to submit, cancel and check a job status. 

Happy provide more support if you want to look into this.

vsoch:
I'd love to take a first shot! I've never written in Groovy before so I suspect this will be fun - as long as I can add to a container with Flux and figure out how to build (and thus test) I should be good to go. Will keep you in the loop (and happy Friday!)

pditommaso:
Can Flux be installed on a local computer for testing?

vsoch:
I'm building a container environment to do this. It technically could, but it has it's own compile and set of dependencies that I wouldn't want to add to my development system. I prefer to work in a containerized environment - what I'll do is provide a Dockerfile that builds it, along with a VSCode .devcontainer that the developer can easily copy into the root to immediately have an environment with Flux too. Testing should also be possible using this container approach!

vsoch:
[...]

PR #3412, opened 2022-11-19
add flux executor to nextflow

vsoch:
This work adds the Flux Framework executor, meaning a Nextflow user can submit jobs to flux. 

- This will close https://github.com/nextflow-io/nextflow/issues/3403

## Features Added 

This addition includes:

- the executor itself
- the executor tests 
- a new documentation tutorial flux.rst, which I think is important for interested developers and/or if we want to add more detail for users
- addition to the executor.rst page
- a Docker and/or DevContainers environment for the interested developer / user

## Differences

Notably, Flux is different from a typical executor because it doesn't do any kind of header file, nor does it allow specifying memory. To get around this, basically instead of having the header section populate user preferences, we add them to the command line instead! This makes the testing file uglier, but it's (I believe) the right way to submit the job.

Flux is also interesting because it's technically (originally) an HPC job manager, but we very notably are extending it to more cloud environments (e.g., Kubernetes). I put it under HPC schedulers because conceptually it will probably be more familiar there, and of course was the original intention of its creation.

For submitting a job, I could have used the regular expression OR the identifier that begins with the fancy f, but I chose to keep both. The reason is because I think we first want to look for that pattern, but I'm not aware if this can be tweaked to not be output (in which case we'd want to fall back to just a string job id). This would also allow for easier tweaking if there is some kind of other regular expression honored (in my short experience I haven't seen one yet).

Finally - yesterday afternoon was the first time I saw Groovy, ever, and I'm not a Java developer, so apologies in advance if I royally missed some details! I did my best to add the executor, test it / document it, and test it. Looking forward to discussion about tweaking it further!

### DIFF
--- a/docs/executor.rst
+++ b/docs/executor.rst
@@ -85,6 +85,27 @@ Resource requests and other job characteristics can be controlled via the follow
 * :ref:`process-time`
 
 
+.. _flux-executor:
+
+Flux Framework Executor
+=======================
+
+The ``flux`` executor allows you to run your pipeline script by using `Flux Framework <https://flux-framework.org>`_.
+
+Akin to slurm, Nextflow manages each process as a separate job that is submitted to the cluster by using the ``flux mini submit`` command.
+
+To enable the Flux executor simply set ``process.executor = 'flux'`` in the ``nextflow.config`` file.
+
+Resource requests and other job characteristics can be controlled via the following process directives:
+
+* :ref:`process-clusterOptions`
+* :ref:`process-cpus`
+* :ref:`process-queue`
+* :ref:`process-time`
+
+.. note:: Flux does not support specifying memory. 
+
+
 .. _ga4ghtes-executor:
 
 GA4GH TES
--- a/docs/flux.rst
+++ b/docs/flux.rst
@@ -1,0 +1,213 @@
+.. _flux-page:
+
+**************
+Flux Framework
+**************
+
+`Flux Framework <https://flux-framework.org/>`_ "flux" is modern resource manager that can span the space between
+cloud and HPC. If your center does not provide Flux for you, you can `build Flux on your own <https://flux-framework.readthedocs.io/en/latest/quickstart.html#building-the-code>`_
+and launch it as a job under your resource manager of choice (e.g., SLURM or on the cloud.)
+
+
+Tutorial
+========
+
+In the `docker/flux <https://github.com/nextflow-io/nextflow/tree/master/docker/flux>`_ directory we provide 
+a `Dockerfile for interacting with Flux <https://github.com/nextflow-io/nextflow/tree/master/docker/flux/.devcontainer/Dockerfile>`_
+
+along with a `VSCode Developer Container <https://code.visualstudio.com/docs/devcontainers/containers>`_
+environment that you can put at the root of the project to be provided with a Flux agent and the dependencies
+needed to build Nextflow. There are two ways to use this:
+
+ - Build a container from scratch and bind your code to it (e.g., for development or testing)
+ - Use VSCode and DevContainers to create a more seamless environment.
+
+Both strategies are described below. For this tutorial, you will generally want 
+to prepare a workflow to use the ``flux`` executor, create an environment with Flux, start a Flux instance, and
+interact with it.
+
+Prepare Nextflow Workflow 
+-------------------------
+
+For your workflow, to use Flux you'll want to specify it in your config. Here is an example 
+``nextflow.config``
+
+
+.. code-block:: console
+
+    // This is an example config for running with flux.
+
+    manifest {
+      mainScript = 'demo.nf'
+      homePage = 'https://github.com/nextflow-io/nextflow/tree/master/docker/flux'
+      description = 'Demo using Nextflow with Flux'
+    } 
+
+    process {
+      executor = 'flux'
+    }
+
+Note that the executor is ``flux``. For additional settings for Flux, see :ref:`flux-executor`.
+Here is an a demo workflow ``demo.nf`` of a job we want to run!
+
+.. code-block:: console
+
+    breakfast = Channel.from '🥞️', '🥑️', '🥧️', '🍵️', '🍞️'
+ 
+    workflow {
+      haveMeal(breakfast)
+    }
+    
+    process haveMeal {
+      debug true
+      input:
+      val food
+      """
+      printf '$food for breakfast!'
+      """
+    }
+
+We will be using these files to run our test workflow. Next, assuming you don't have one handy,
+let's set up an envrionment with Flux.
+
+Container Environment
+---------------------
+
+You can choose just to build the Docker image from the root of the repository:
+
+.. code-block:: console
+    
+    $ docker build -f docker/flux/.devcontainer/Dockerfile -t nextflow-flux .
+
+
+And then shell into the container for a development environment. You'd need to bind
+the present working directory to ``/code`` to see your local changes in the container:
+
+.. code-block:: console
+
+    $ docker run -it -v $PWD:/code nextflow-flux 
+
+
+You can also move the .devcontainer directory to the root of your repository, and just
+open it in VSCode.
+
+.. code-block:: console
+
+    $ cp -R docker/flux/.devcontainer .devcontainer
+
+
+Then open in vscode, and select to "Re-open in container"
+
+.. code-block:: console
+
+    $ code .
+
+Then you should be able to open a terminal (Terminal -> New Terminal)
+to interact with the command line. Try running `make`` again!
+Whichever of these two approaches you take, you should be in a container 
+environment with ``flux`` on the path.
+
+Start a Flux Instance 
+---------------------
+
+Once in your container, you can start an interactive flux instance (from which you can submit jobs on
+the command line to test with Nextflow) as follows:
+
+.. code-block:: console
+
+    $ flux start --test-size=4
+
+
+Getting Familiar with Flux
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. note::
+
+    This step is optional!
+
+Here is an example of submitting a job and getting the log for it.
+
+.. code-block:: console
+
+    $ flux mini submit echo "HELLO MOTO"
+    ƒEzWqspb
+
+
+And then getting the log for it:
+
+
+.. code-block:: console
+
+    $ flux job attach ƒEzWqspb
+    HELLO MOTO
+
+
+Try submitting a longer job:
+
+.. code-block:: console
+
+    $ flux mini submit sleep 60
+
+
+And then seeing it in the jobs listing.
+
+.. code-block:: console
+
+    $ flux jobs
+           JOBID USER     NAME       ST NTASKS NNODES     TIME INFO
+       ƒ4tkMUAAT root     sleep       R      1      1   2.546s ab6634a491bb
+
+
+We are going to be issuing similar submit commands via Nextflow!
+
+Submitting with Nextflow
+------------------------
+
+Prepare your ``nextflow.config`` and ``demo.nf`` in the same directory.
+
+.. code-block:: console
+
+    $ ls .
+    demo.nf    nextflow.config 
+
+
+If you've installed Nextflow already with the flux executor, you are good to go! If you are working 
+with development code and need to build nextflow: 
+
+.. code-block:: console
+
+    $ make assemble
+
+Make sure ``nextflow`` is on your PATH (here we are in the root of Nextflow):
+
+.. code-block:: console
+
+    $ export PATH=$PWD:$PATH
+    $ which nextflow
+    /workspaces/nextflow/nextflow
+
+Then cd into the directory with your config and demo file:
+
+.. code-block:: console
+
+    $ cd docker/flux
+
+
+And then run the workflow with flux!
+
+.. code-block:: console
+
+    $ nextflow -c nextflow.config run demo.nf
+
+    # nextflow -c nextflow.config run demo.nf 
+    N E X T F L O W  ~  version 22.10.0
+    Launching `demo.nf` [clever_blackwell] DSL2 - revision: f8cda838cb
+    executor >  flux (5)
+    [4c/f162db] process > haveMeal (3) [100%] 5 of 5 ✔
+    🥞️ for breakfast!
+    🍞️ for breakfast!
+    🍵️ for breakfast!
+    🥑️ for breakfast!
+    🥧️ for breakfast!
+
+And that's it! You've just run a workflow using nextflow and Flux. 
--- a/modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
@@ -1,0 +1,175 @@
+/*
+ * Copyright 2022, Seqera Labs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package nextflow.executor
+import java.nio.file.Path
+import java.util.regex.Pattern
+
+import groovy.util.logging.Slf4j
+import nextflow.processor.TaskRun
+/**
+ * Processor for Flux Framework executor
+ *
+ * See https://flux-framework.org
+ *
+ *
+ * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
+ * @author Vanessa Sochat <sochat1@llnl.gov>
+ */
+@Slf4j
+class FluxExecutor extends AbstractGridExecutor {
+
+    static final private Pattern SUBMIT_REGEX = ~/(ƒ.+)/
+
+    /**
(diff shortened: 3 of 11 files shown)


docs/conf.py
```diff
@@ -52,9 +52,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '22.10'
+version = '22.11'
```
pditommaso, line 55:
```suggestion
version = '22.10'
```

docs/conf.py
```diff
@@ -52,9 +52,9 @@
...
 # The full version, including alpha/beta/rc tags.
-release = '22.10.0'
+release = '22.11.0-SNAPSHOT'
```
pditommaso, line 57:
```suggestion
release = '22.10.0'
```

Changes to this file should not be committed

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
```diff
@@ -0,0 +1,174 @@
...
+ * See https://flux-framework.org
+ *
+ *
+ * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
+ * @author Vanessa Sochat <sochat1@llnl.gov>
+ */
+@Slf4j
+class FluxExecutor extends AbstractGridExecutor {
+
+    static private Pattern SUBMIT_REGEX = ~/(ƒ.+)/
```
pditommaso, line 35:
```suggestion
    static final private Pattern SUBMIT_REGEX = ~/(ƒ.+)/
```

Use `final` to make it constant

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
```diff
@@ -0,0 +1,174 @@
...
+     * @param task A {@link TaskRun} to be submitted
+     * @param result The {@link List} instance to which add the job directives
+     * @return A {@link List} containing all directive tokens and values.
+     */
+    protected List<String> getDirectives(TaskRun task, List<String> result) {
+        return result
+    }
+
+    // Flux does not require a special token or header
+    String getHeaderToken() { '' }
```
pditommaso, line 49:
```suggestion
    String getHeaderToken() { null }
```

Empty strings can be tricky to debug, better to use `null`

modules/nf-commons/src/main/nextflow/Const.groovy
```diff
@@ -53,17 +53,17 @@ class Const {
     /**
      * The application version
      */
-    static public final String APP_VER = "22.10.0"
+    static public final String APP_VER = "22.11.0-SNAPSHOT"
```
pditommaso, line 56:
```suggestion
    static public final String APP_VER = "22.10.0"
```

modules/nf-commons/src/main/nextflow/Const.groovy
```diff
@@ -53,17 +53,17 @@ class Const {
...
 
     /**
      * The app build time as linux/unix timestamp
      */
-    static public final long APP_TIMESTAMP = 1665639892847
+    static public final long APP_TIMESTAMP = 1668828550138
```
pditommaso, line 61:
```suggestion
    static public final long APP_TIMESTAMP = 1665639892847
```

modules/nf-commons/src/main/nextflow/Const.groovy
```diff
@@ -53,17 +53,17 @@ class Const {
...
 
     /**
      * The app build number
      */
-    static public final int APP_BUILDNUM = 5826
+    static public final int APP_BUILDNUM = 5835
```
pditommaso, line 66:
```suggestion
    static public final int APP_BUILDNUM = 5826
```

These changes are automatically generated, should not be committed

tests/nextflow.config
```diff
@@ -22,7 +22,7 @@ manifest {
 } 
 
 process {
-  container = 'quay.io/nextflow/tests'
+  process.executor = 'flux'
```
pditommaso, line 25:
```suggestion
  container = 'quay.io/nextflow/tests'
```

nextflow
```diff
@@ -16,7 +16,7 @@
 #  limitations under the License.
 
 [[ "$NXF_DEBUG" == 'x' ]] && set -x
-NXF_VER=${NXF_VER:-'22.10.0'}
+NXF_VER=${NXF_VER:-'22.11.0-SNAPSHOT'}
```
pditommaso, line 19:
```suggestion
NXF_VER=${NXF_VER:-'22.10.0'}
```

modules/nextflow/src/main/resources/META-INF/plugins-info.txt
```diff
@@ -1,8 +1,8 @@
...
 nf-azure@0.14.2
-nf-tower@1.5.5
+nf-codecommit@0.1.2
 nf-ga4gh@1.0.4
-nf-codecommit@0.1.2
\ No newline at end of file
+nf-wave@0.5.2
+nf-console@1.0.4
+nf-tower@1.5.6
+nf-google@1.4.5
```
pditommaso: This is another generated file. We only commit in the final release otherwise it mess up merges. Better to rollback to the previous version that should be the following 

```
nf-amazon@1.11.0
nf-wave@0.5.2
nf-console@1.0.4
nf-google@1.4.4
nf-azure@0.14.2
nf-tower@1.5.5
nf-ga4gh@1.0.4
nf-codecommit@0.1.2
```

docs/flux.rst
```diff
@@ -0,0 +1,212 @@
...
+
+    process sayHello {
+      debug true
+      input:
+      val food from breakfast
+
+      """
+      printf '$food for breakfast!'
+      """
+    }
```
pditommaso, line 67:
```suggestion

    breakfast = Channel.from '🥞️', '🥑️', '🥧️', '🍵️', '🍞️'
 
    workflow {
      sayHello(breakfast)
    }
    
    process sayHello {
      debug true
      input:
      val food

      """
      printf '$food for breakfast!'
      """
    }
```

Love this example 😄. Note DSL1 is deprecated and it's going to be removed in the upcoming version. I've rewrote the it in DSL2

pditommaso (changes requested):
Thanks for this great PR! very excited about adding support for Flux to Nextflow and the Flux operator project, it looks very interesting. 

The implementation looks just fine! I've only commented about reverting changes to automatically generated files that usually we do not commit. 

It would also be better if you could use a real email address to sign the commit instead of `users.noreply.github.com`.

ps. don't worry about your groovy programming skills, it looks like they are vastly above the average 😄

vsoch:
Will get fixes in soon! Note that the GitHub noreply email is the one I have associated with my account - I don't actually want people sending me emails, but they can find me on here :) If I change that email it likely won't associate with my GitHub account.

modules/nf-commons/src/main/nextflow/Const.groovy
vsoch: haha yeah I was wondering why these files were changed! I think I was testing different parts of the build process and probably this was done with a make command I should not have run :P

docs/flux.rst
vsoch: okay I see why the files changed - when I do `make compile` that seems to change the versions and timestamps. I was doing that to test the new binary, eg., here is testing the updated workflow:

```bash
$ make compile
# files change
$ nextflow -c nextflow.config run demo.nf
N E X T F L O W  ~  version 22.11.0-SNAPSHOT
Launching `demo.nf` [nice_golick] DSL2 - revision: 58592626a9
executor >  flux (5)
[ae/a01395] process > haveMeal (1) [100%] 5 of 5 ✔
🥑️ for breakfast!
🥧️ for breakfast!
🍞️ for breakfast!
🍵️ for breakfast!
🥞️ for breakfast!
```
I'll try doing `make compile` and then git stash to immediately undo the changes.

vsoch:
@pditommaso I'm going to ping some of my Flux colleagues for a final look over to make sure the command line interactions I chose are sane.

docs/executor.rst
```diff
@@ -85,6 +85,30 @@ Resource requests and other job characteristics can be controlled via the follow
...
 
 
+.. _flux-executor:
+
+Flux Framework Executor
+=======================
+
+The ``flux`` executor allows you to run your pipeline script by using `Flux Framework <https://flux-framework.org>`_.
+
+Akin to slurm, Nextflow manages each process as a separate job that is submitted to the cluster by using the ``flux submit`` command.
```
grondo, line 95:
```suggestion
Akin to slurm, Nextflow manages each process as a separate job that is submitted to the cluster by using the ``flux mini submit`` command.
```

docs/executor.rst
```diff
@@ -85,6 +85,30 @@ Resource requests and other job characteristics can be controlled via the follow
...
+
+Being so, the pipeline must be launched from a node where the ``sbatch`` command is available, that is, in a common usage
+scenario, the cluster `head` node.
```
grondo, line 98: `s/sbatch/flux/`. However, this line can probably be removed because the `flux(1)` command should be available everywhere on a cluster or job running flux.

docs/executor.rst
```diff
@@ -85,6 +85,30 @@ Resource requests and other job characteristics can be controlled via the follow
...
+
+To enable the SLURM executor simply set ``process.executor = 'flux'`` in the ``nextflow.config`` file.
```
grondo, line 100:
```suggestion
To enable the Flux executor simply set ``process.executor = 'flux'`` in the ``nextflow.config`` file.
```

docs/flux.rst
```diff
@@ -0,0 +1,213 @@
+.. _flux-page:
+
+**************
+Flux Framework
+**************
+
+`Flux Framework <https://flux-framework.org/>`_ "flux" is modern resource manager that can span the space between
+cloud and HPC. If your center does not provide Flux for you, you can `build Flux on your own <https://flux-framework.readthedocs.io/en/latest/quickstart.html#building-the-code>`_
+and launch jobs under your resource manager of choice (e.g., SLURM or on the cloud.)
```
grondo, line 9:
```suggestion
and launch it as a job under your resource manager of choice (e.g., SLURM or on the cloud.)
```

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
```diff
@@ -0,0 +1,174 @@
...
+    List<String> getSubmitCommandLine(TaskRun task, Path scriptFile ) {
+
+        List<String> result = ['flux', 'mini', 'submit']
+        result << '--setattr=cwd=' + quote(task.workDir)
+        result << '--job-name="' + getJobNameFor(task) + '"'
+        result << '--output=' + quote(task.workDir.resolve(TaskRun.CMD_LOG))  // -o OUTFILE
+
+        if( task.config.cpus > 1 ) {
+            result << '--cores=' + task.config.cpus.toString()
+        }
```
grondo, line 68: I don't know Nextflow, but if `task.config.cpus` here means to allocate > 1 core per task, then I think you might want `--cores-per-task=` here instead of just `--cores`. I _think_ `--cores=N` here will end up running multiple tasks since the default `--tasks-per-core` is 1. (And sorry for the confusion there!)

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
```diff
@@ -0,0 +1,174 @@
...
+     *
+     * @param text The string returned when submitting the job
+     * @return The actual job ID string
+     */
+    @Override
+    def parseJobId(String text) {
+
+        // Parse the special "f" first
+        for( String line : text.readLines() ) {
+            def m = SUBMIT_REGEX.matcher(line)
```
grondo, line 111: The jobid will always be returned in the [F58](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_19.html) encoding, but depending on locale it may start with an `f` or an `ƒ`. It may be simpler here to just grab the first word printed by `flux mini submit` and return it without trying to match or parse the jobid?

If you want to ensure the jobid validity, you could pass the output to `flux job id <jobid>` which will succeed and return an integer for any valid jobid.

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
```diff
@@ -0,0 +1,174 @@
...
+        throw new IllegalStateException("Invalid Flux submit response:\n$text\n\n")
+    }
+
+    @Override
+    protected List<String> getKillCommand() { ['flux', 'job', 'cancel'] }
+
+    @Override
+    protected List<String> queueStatusCommand(Object queue) {
+
+        final result = ['flux', 'jobs', '--suppress-header']
```
grondo, line 131: Suggestion: I worry a little here that if the default `flux jobs` output format changes for some reason, this section of the code could break without warning. Since `flux jobs` allows an output format to be specified, perhaps that could be used to ensure a static output format, e.g.

```java
        final result = [ 'flux', 'jobs', '-a',  '--suppress-header', '--format="{id.f58} {status_abbrev}"' ]
```

would just print a list of jobid/status pairs, e.g.:
```$ flux jobs -a --suppress-header --format="{id.f58} {status_abbrev}"
ƒaNehUumZ1 CD
ƒBgsJ5Yum1 CD
ƒBgkFZWouy CD
ƒBgMtbB71R CD
ƒ48VvoPC7h CD
ƒ28LvWXHXM CD
ƒ28Hp8Cfou CD
```

Note I also added `-a`, without which only "active" jobs are displayed (i.e. those that are pending or running)

grondo (commented):
As requested, I gave this PR a quick review from the Flux perspective and overall looks great! Just some inline comments for typos and a few suggestions. Thanks!

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
pditommaso: Yes, nextflow is cpus per task

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
pditommaso: about `-a` nextflow uses this info also to determine when the task execution is complete. Ideally, it would be nice to include tasks completed in the last x minutes

grondo: `flux jobs` does have a `--since` option:
```
      --since=WHEN           Include jobs that have become inactive since
                             WHEN. (implies -a if no other --filter option is
                             specified)
```

grondo: Oh, and I should mention that `--since` can take an offset like `-30m` or even `15 minutes ago`

pditommaso: awesome

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
vsoch: I sort of do this by way of first checking for the pattern and falling back to the first thing returned. I figured if people have the defaults the first is more targeted / accurate and the second is a catch all.

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
vsoch: Okay I can change to since - what value goes into that? The function looks like it only takes "queue" as an argument?

pditommaso: I think adding `--since -15m` should be enough

vsoch:
okay changes are in! Thanks for the feedback @grondo I knew you'd have wisdom! I gave our breakfast demo one more sanity run and (with the changes) it looks good! :partying_face: 

```console
# nextflow -c nextflow.config run demo.nf 
N E X T F L O W  ~  version 22.11.0-SNAPSHOT
Launching `demo.nf` [spontaneous_hamilton] DSL2 - revision: 58592626a9
executor >  flux (5)
[99/3bff15] process > haveMeal (2) [100%] 5 of 5 ✔
🥞️ for breakfast!
🥧️ for breakfast!
🍞️ for breakfast!
🍵️ for breakfast!
🥑️ for breakfast!
```

pditommaso (approved):

pditommaso:
Merged!

vsoch:
woo!!

@pditommaso you are the king of animated memes I have absolutely loved this thread! :raised_hands:

pditommaso:
Indeed, this PR was quick and fun! thanks a lot!

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
epstein6: Hello,

I appear to be running into this issue.  I get this, because my environment uses `f` to distinguish flux jobids:
```
Caused by:
  Invalid Flux submit response:
  f3t1ZMD36U8w
```

Could the regex at least be expanded to match this?

Thanks!

vsoch: So I understand - you are just looking to add a traditional `f` (not the curly one) to the regex here? https://github.com/nextflow-io/nextflow/blob/938608284c8e592cd22fde9b67dcca6dcf6ee2b6/modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy#L38

epstein6: Yes, I'm currently trying to test, but I think that would help.

vsoch: I'm not sure if there is anything special about regex in Groovy, but I opened a PR where we can at least continue this discussion. https://github.com/nextflow-io/nextflow/pull/5662

Issue #3419, opened 2022-11-22
Flux output warning


vsoch:
## Bug report 

I'm getting a warning when trying to list jobs - the error is an empty string so I'm not sure what it is!

```console
WARN: Training is enabled but pre-trained model(s) are also provided, pre-trained models will be ignored
WARN: [FLUX] queue status cannot be fetched
- cmd executed: flux jobs --suppress-header --format="{id.f58} {status_abbrev}" --since="-15m" --user root
- exit status : 1
- output      :
  flux-jobs: ERROR: ''
```
It just skips over it and keeps going so it's not blocking, but I wanted to open the issue in case anyone has insight. Our output tests work okay, so I am not sure the best way to debug.

* Nextflow version: master branch
* Java version:
```
# java --version
openjdk 11.0.17 2022-10-18
OpenJDK Runtime Environment (build 11.0.17+8-post-Ubuntu-1ubuntu220.04)
OpenJDK 64-Bit Server VM (build 11.0.17+8-post-Ubuntu-1ubuntu220.04, mixed mode, sharing)
```

* Operating system: Linux (ubuntu) in DevContainers environment
* Bash version: GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu

pditommaso:
This error means that the command below failed 

```
flux jobs --suppress-header --format="{id.f58} {status_abbrev}" --since="-15m" --user root
```

maybe something is wrong in the syntax?

vsoch:
It works okay in my terminal (where I'm running everything / debugging) and we know the tests were ok too!

```
# flux jobs --suppress-header --format="{id.f58} {status_abbrev}" --since="-250m" --user root
ƒLgreGCNj CD
ƒLgnApN4P CD
ƒLghWzcmd CD
ƒLcHocVZy CD
ƒLeqVkwrP CD
...
```
Could it be some issue with either the quotes, or how the FluxExecutor is running it? :thinking:

pditommaso:
Not sure. I'd try to rewrite it as wrapping the flux in a `sh` command, see below

```
['sh','-c','flux jobs --suppress-header --format="{id.f58} {status_abbrev}" --since="-15m" --user root']
```

vsoch:
Will try this! Probably tomorrow (err later today?) time to... turn into a pumpkin! :jack_o_lantern:

pditommaso:
[...]

PR #3424, opened 2022-11-22
fix bug with FluxExecutor jobs output

vsoch:
This pull request fixes one bug, and adds a new parameter to the FluxExecutor to support printing of output to stderr/stdout.

- Fixes https://github.com/nextflow-io/nextflow/issues/3419

## Jobs Listing Bug
Currently, for some reason the jobs listing is passing tests, but in testing with a workflow there is an error warning printed to the terminal as reported in https://github.com/nextflow-io/nextflow/issues/3419 The tweak suggested by Paolo there seems to fix the issue - I'm able to run the workflow from the terminal or web UI and I no longer see the warning.

## Output

We started a discussion in https://github.com/nextflow-io/nextflow/discussions/3421#discussioncomment-4204810 about how to enforce the process to print to stderr/stdout. I woke up and realized it was a "this is simple stupid" issue - we needed to add a parameter to the FluxExecutor that would say "please print all output to the terminal." This means the default will still print to the default log, which is expected behavior, but work workflows that use Flux within the RESTFul API I can give instruction for a profile like:

```
profiles {
    flux {
        process.conda = "$baseDir/conda.yml"
        process.executor = 'flux'
        process.terminal_output = true
    }

```

and that seems to fix the issue. And I don't think it's so nuts to require this custom setting for a Flux RESTFul API intended workflow - we typically need to build a custom container base anyway, and we can figure out a more dynamic way down the line (e.g., providing the argument via the running command perhaps) if this isn't desired. For each job that is created for us by nextflow we get the output in the UI if there is output.

image

This was totally empty before! So I'm very happy - I can click a job in the table and see what the heck it was doing.

### DIFF
--- a/docs/executor.rst
+++ b/docs/executor.rst
@@ -103,6 +103,8 @@ Resource requests and other job characteristics can be controlled via the follow
 * :ref:`process-queue`
 * :ref:`process-time`
 
+Additionally, to have Flux print all output to stderr and stdout, add `flux.terminalOutput` to be true.
+
 .. note:: Flux does not support specifying memory. 
 
 
--- a/modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
@@ -61,7 +61,12 @@ class FluxExecutor extends AbstractGridExecutor {
         List<String> result = ['flux', 'mini', 'submit']
         result << '--setattr=cwd=' + quote(task.workDir)
         result << '--job-name="' + getJobNameFor(task) + '"'
-        result << '--output=' + quote(task.workDir.resolve(TaskRun.CMD_LOG))  // -o OUTFILE
+
+        // Only write output to file if user doesn't want written entirely to terminal
+        Boolean terminalOutput = task.config.navigate('flux.terminalOutput') as Boolean
+        if ( !terminalOutput ) {
+            result << '--output=' + quote(task.workDir.resolve(TaskRun.CMD_LOG))  // -o OUTFILE
+        }
 
         if( task.config.cpus > 1 ) {
             result << '--cores-per-task=' + task.config.cpus.toString()
@@ -129,17 +134,18 @@ class FluxExecutor extends AbstractGridExecutor {
     protected List<String> queueStatusCommand(Object queue) {
 
         // Look at jobs from last 15 minutes
-        final result = ['flux', 'jobs', '--suppress-header', '--format="{id.f58} {status_abbrev}"', '--since="-15m"']
+        String command = 'flux jobs --suppress-header --format="{id.f58} {status_abbrev}" --since="-15m"'
 
         if( queue )
-            result << '--queue' << queue.toString()
+            command += ' --queue=' + queue.toString()
 
         final user = System.getProperty('user.name')
         if( user )
-            result << '--user' << user
+            command += ' --user=' + user
         else
             log.debug "Cannot retrieve current user"
 
+        final result = ['sh', '-c', command]
         return result
     }
 
--- a/modules/nextflow/src/test/groovy/nextflow/executor/FluxExecutorTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/executor/FluxExecutorTest.groovy
@@ -89,6 +89,12 @@ class FluxExecutorTest extends Specification {
         task.config.clusterOptions = '--tasks-per-node=4 --cpus-per-node=4'
         then:
         executor.getSubmitCommandLine(task, Paths.get('/some/path/job.sh')) == ['flux', 'mini', 'submit', '--setattr=cwd=/work/path', '--job-name="nf-my_task"', '--output=/work/path/.command.log', '--time-limit=60', '--tasks-per-node=4', '--cpus-per-node=4', '/bin/bash', 'job.sh']
+
+        when:
+        task.config = new TaskConfig()
+        task.config.flux = [terminalOutput: true]
+        then:
+        executor.getSubmitCommandLine(task, Paths.get('/some/path/job.sh')) == ['flux', 'mini', 'submit', '--setattr=cwd=/work/path', '--job-name="nf-my_task"', '/bin/bash', 'job.sh']
     }
 
     def testWorkDirWithBlanks() {
@@ -144,7 +150,7 @@ class FluxExecutorTest extends Specification {
         def executor = [:] as FluxExecutor
         then:
         usr
-        executor.queueStatusCommand(null) == ['flux', 'jobs', '--suppress-header', '--format="{id.f58} {status_abbrev}"', '--since="-15m"', '--user', usr]
-        executor.queueStatusCommand('xxx') == ['flux', 'jobs', '--suppress-header', '--format="{id.f58} {status_abbrev}"', '--since="-15m"', '--queue', 'xxx', '--user', usr]
+        executor.queueStatusCommand(null) == ['sh', '-c', "flux jobs --suppress-header --format=\"{id.f58} {status_abbrev}\" --since=\"-15m\" --user=" + usr]
+        executor.queueStatusCommand('xxx') == ['sh', '-c', "flux jobs --suppress-header --format=\"{id.f58} {status_abbrev}\" --since=\"-15m\" --queue=xxx --user=" + usr]
     }
 }


modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
```diff
@@ -61,7 +61,12 @@ class FluxExecutor extends AbstractGridExecutor {
         List<String> result = ['flux', 'mini', 'submit']
         result << '--setattr=cwd=' + quote(task.workDir)
         result << '--job-name="' + getJobNameFor(task) + '"'
-        result << '--output=' + quote(task.workDir.resolve(TaskRun.CMD_LOG))  // -o OUTFILE
+
+        // Only write output to file if user doesn't want written entirely to terminal
+        Boolean terminal_output = task.config.navigate('terminal_output') as Boolean
```
vsoch, line 66: I'm not sure what the best way to handle a "this could be completely undefined, or defined and a boolean" logic. I first tried checking if the value (as a string) was == 'true' and this worked when it was defined in the config, and threw an error in testing here (not defined). I suspect doing a verbatim `task.config.terminal_output == true` would do the same. So I saw this other pattern to "navigate" but I'm not sure that is safe either - I'm wondering if I need something like https://www.logicbig.com/tutorials/misc/groovy/safe-navigation-operator.html. I'm hoping the above is smart enough to convert whatever it finds to boolean!

pditommaso: I think we have already `process.debug = true` for this

vsoch: But what I’d they want debug level logging but to the output file? Does debug just control where it writes and it was erroneous of me to not use it to direct the output?

vsoch: And to clarify - when I tested `process.debug` here I saw a _ton_ of output in the main process thread (which for me is in the web server logs) and I definitely don't want that. I want the same output in the logs as currently is there (just the log heading and showing the jobs running) but then the robust output in the process stdout/stderr. So I'm not sure that using process.debug is what I want -it seems to do a bit more.

vsoch: So what I'd like to do is:

1. Ensure that there is some flag that tells the FluxExecutor not to specify any output to a file (this is the functionality I need to see the logs in the UI) e.g.,

image

2. Also ensure the main process isn't getting overly verbose (it will clutter up the web server where it's running).

I'm happy to test out any suggestions / make any changes that are desired! In practice I couldn't get anything with `process.debug` to work as I wanted, and if I use it here, I suspect we are going to not satisfy point 2 above.

pditommaso: Nextflow commonly uses these files for task logs: 
1. `.command.out`: nothing more the task `/dev/stdout`
2. `.command.err`: as above, the task `/dev/stdout` 
3. `.command.log`: the task output as returned by the job scheduler, essentially this file contains the task stdout + stderr + some prolog or epilog depending the batch scheduler. 

When you specify `process.debug = true`. Nextflow reads the content of `.command.out` end print it to the main process stdout (if it's missing fallback to .command.err, if also this is missing fallback to .command.log) 

Don't know exactly how Flux works at this regard, but commonly when the task output log is not specified the stdout/err is written to a file (or a couple of files) whose name is chosen by the file system. This should not alter how the job output is rendered by nextlow. 

I'd suggest to focus this PR on fixing the job status warning.

vsoch: Yep, and I don't (can't) read from a file. I need the output and error in the stdout/stderr of the worker, and _not_ the main process.  So using process.debug isn't going to work. I need a way to control the `--output`, as I've suggested in the PR here.

vsoch: Why won't you let me customize the executor? I don't understand. There is a use case that NextFlow does not work for. As I've stated, I cannot read an output file. I also don't want output in the main process - that's going to a server log that a user won't see. The user will see the output for the process, whatever it is (this is not specific to Nextflow, it's for any thing that is being run). With this tiny addition of a config variable, I'm able to have a beautiful table of process runs with logs specific to them in the UI. Without, the logs are all empty and it's useless.

If you don't like this implementation of an extra option, I can do a hack to "throw it" into the catch all extra arguments param, but then I'd need to loop over it, check for one, and set the variable there. Let me know if this is acceptable and I'm happy to test this out.

pditommaso: Hi, PR reviews is all about sharing knowledge and giving advice. I'm not trying to stop customizing the executor, that's good!

Now, I understand you need to control having or not having the flux output log file. 

From Nextflow point is better to have that file, because it can provide useful debugging information to the user if the job fails and for some reason the .commnad.out/err have not been created. So my suggestion is to keep it enabled by default, and add an option to disable it. 

Regarding how to config it. The `process.something` settings are meant to be \*portable* task directives i.e. that applies to all or at least some executors e.g. `cpus`, `memory`, `label`, etc. We tend to add a disk directive only for one executor. 

However, it's perfectly fine to have a `flux { yourSetting = 'here' }` configuration scope in the nextflow config for the option related to Flux (we use camelCase convention for config setting instead of `_` separator). 

I think this should accommodate your needs.

vsoch: Ahh fantastic! So… this is exactly what I had in the first version of the PR with the process.terminal_output flag (false or not present by default leading to printing to output). I thankfully saved that work to a different branch so I’ll see if I can use it to update the PR here, and worst case open a new one.

Thank you! 🙏

vsoch: okay I think that worked! I cloned the repository again (and fetched the different branch with my backup) and then checked out this branch from that branch, and force pushed.

vsoch: oh I missed the camel case though! Let me fix that.

vsoch: okay:

1. rebased
2. changed to camelCase
3. added a test for terminalOutput

I didn't get to run the test locally (my DevContainers session bonked out) but maybe I'll luck out and get it right anyway!

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
```diff
@@ -61,7 +61,12 @@ class FluxExecutor extends AbstractGridExecutor {
         List<String> result = ['flux', 'mini', 'submit']
         result << '--setattr=cwd=' + quote(task.workDir)
         result << '--job-name="' + getJobNameFor(task) + '"'
-        result << '--output=' + quote(task.workDir.resolve(TaskRun.CMD_LOG))  // -o OUTFILE
+
+        // Only write output to file if user doesn't want written entirely to terminal
+        Boolean terminalOutput = task.config.navigate('terminalOutput') as Boolean
```
pditommaso, line 66:
```suggestion
        Boolean terminalOutput = task.config.navigate('flux.terminalOutput') as Boolean
```

docs/executor.rst
```diff
@@ -103,6 +103,8 @@ Resource requests and other job characteristics can be controlled via the follow
 * :ref:`process-queue`
 * :ref:`process-time`
 
+Additionally, to have Flux print all output to stderr and stdout, add `process.terminalOutput` to be true.
```
pditommaso, line 106:
```suggestion
Additionally, to have Flux print all output to stderr and stdout, add `flux.terminalOutput` to be true.
```

modules/nextflow/src/test/groovy/nextflow/executor/FluxExecutorTest.groovy
```diff
@@ -89,6 +89,12 @@ class FluxExecutorTest extends Specification {
         task.config.clusterOptions = '--tasks-per-node=4 --cpus-per-node=4'
         then:
         executor.getSubmitCommandLine(task, Paths.get('/some/path/job.sh')) == ['flux', 'mini', 'submit', '--setattr=cwd=/work/path', '--job-name="nf-my_task"', '--output=/work/path/.command.log', '--time-limit=60', '--tasks-per-node=4', '--cpus-per-node=4', '/bin/bash', 'job.sh']
+
+        when:
+        task.config = new TaskConfig()
+        task.config.terminalOutput = true
```
pditommaso, line 95:
```suggestion
        task.config.flux.terminalOutput = true
```

pditommaso (changes requested):
`flux.` is required to access to flux specific settings

modules/nextflow/src/test/groovy/nextflow/executor/FluxExecutorTest.groovy
vsoch: Oh I understand! So I’m allowed to define arbitrary variables under an “on the fly” flux namespace in the config - and navigate will allow me to do that (without a null error if it’s not defined). Gotcha! Sorry I missed this detail in your description above - it makes perfect sense. I’ll update the PR.

vsoch:
That's a really nice design - we have our own little namespace of options! I couldn't ask for more. I like this a lot now.

image

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
vsoch: I'm wondering - how did this work before with  just terminalOutput when it was under process.terminalOutput? Does navigate do a search of sub-fields? :thinking:

vsoch:
ok hmm those changes seemed to break things - taking a look!

vsoch:
Oh I see I need to set the task.config.flux as something first.

vsoch:
hmm okay @pditommaso I might need some advice here - the "try the thing that looks like a dictionary" didn't seem to work! I'll try setting a TaskConfig there instead next.

pditommaso:
LOL. it broke GitHub 😆

pditommaso:
Well, there's no change in the GH action build so not understanding the reason. Maybe just a GH temporary glitch. I'll relaunch it

vsoch:
I think the bug is valid! I don't think I'm allowed to do this:

```groovy
task.config.flux = { terminalOutput = true }
```
I have to define flux as "something" before that - the "something I'm not sure yet (I'm trying various stuff, so far no luck!)

pditommaso:
Ok, not is a unit test failing, not GitHub. You can test locally running `make smoke` in the project root

vsoch:
Trying:
```
        task.config.flux = [terminalOutput: true]
```

vsoch:
yeah I definitely tried that in my nextflow.config - it doesn't seem to be taking:

```
    flux {
        process.conda = "$baseDir/conda.yml"
        process.executor = 'flux'
        flux {
          terminalOutput = true
        }
    }
```
So we aren't good to merge here until that is figured out (the other PRs I tested all worked)!

bentsherman:
All the tests look good. So it's not working with a manual test? What is the error message?

pditommaso:
it works for me 👍

vsoch:
Log of patterns I'm trying:

```diff
- Does not work
    flux {
        process.conda = "$baseDir/conda.yml"
        process.executor = 'flux'
        flux {
          terminalOutput = true
        }
    }
```

```diff
- Does not work
    flux {
        process.conda = "$baseDir/conda.yml"
        process.executor = 'flux'
        flux: {
          terminalOutput = true
        }
    }
```

```diff
- Does not work
    flux {
        process.conda = "$baseDir/conda.yml"
        process.executor = 'flux'
        flux: [terminalOutput: true]        
    }
```

```diff
- Does not work
    flux {
        process.conda = "$baseDir/conda.yml"
        process.executor = 'flux'
        flux.terminalOutput = true
    }
```

I tried putting this outside of the profile (directly in the config):

```diff
- Does not work
flux {
  terminalOutput = true
}
```
For all of the above, it runs, but there is no output (as shown above) indicating it is not taking. Are we sure that nesting in profiles works? I'm running out of ideas so I will try to look for other examples - something must be missing to map the profile nested config attribute into the code. Once it's there, based on the tests here I think we can see it works.

PR #3427, opened 2022-11-23
allow collect method to be chaining

jorgeaguileraseqera:
as GroovyDefault implements a `collect` method by default for the script the user is not able  to use the collect operator when chaining methods (i.e. channel.of('a') | collect | view )

 importing this method as static via ScriptParser tells the compiler to use it instead default method


pditommaso:
Smart trick, but it's not clear the reason why this problem arises only for collect.

PR #3432, opened 2022-11-24
testing fix for config variables

### LINKED ISSUE #3429: Bug translating nextflow.config nested "flux" variables into workflow
## Bug report 

As reported in https://github.com/nextflow-io/nextflow/pull/3424#issuecomment-1325579816 

The new addition to the code #3424 seems to work when the config attributes are defined in code (e.g., for testing) but I can't figure out how to write this into a `nextflow.config` for a workflow. Here is what I've tried:

Log of patterns I'm trying:

```diff
- Does not work
    flux {
        process.conda = "$baseDir/conda.yml"
        process.executor = 'flux'
        flux {
          terminalOutput = true
        }
    }
```

```diff
- Does not work
    flux {
        process.conda = "$baseDir/conda.yml"
        process.executor = 'flux'
        flux: {
          terminalOutput = true
        }
    }
```

```diff
- Does not work
    flux {
        process.conda = "$baseDir/conda.yml"
        process.executor = 'flux'
        flux: [terminalOutput: true]        
    }
```

```diff
- Does not work
    flux {
        process.conda = "$baseDir/conda.yml"
        process.executor = 'flux'
        flux.terminalOutput = true
    }
```

I tried putting this outside of the profile (directly in the config):

```diff
- Does not work
flux {
  terminalOutput = true
}
```
For all of the above, it runs, but there is no output (as shown above) indicating it is not taking. Are we sure that nesting in profiles works? I'm running out of ideas so I will try to look for other examples - something must be missing to map the profile nested config attribute into the code. Once it's there, based on the tests here I think we can see it works.

The expected behavior is for me to see the output /stdout associated with the process, as I was seeing with an earlier version of that same PR. The current behavior is that I don't see it - and this I think reflects perhaps that we aren't correctly translating the DSL into code.

I have a container build with instructions that I'm following:

https://github.com/rse-ops/flux-hpc/tree/update/nextflow-logs/nextflow/ml-hyperopt

Note that the logs picture was taken after the _first_ fix, with this current approach it looks like:

image

I'm thinking - is there a similar nested-thing in a profile that we know works? Perhaps that would be a starting point for comparison. As always, thank you for the help! I hope we are close and I'm just bad at the DSL. :laughing:

vsoch:
This is a tiny change to ensure that the flux.terminalOutput is correctly passed all the way from the config to the executor! For some reason with task.config, it's not there, likely because it's not known to the task to be parsed?

This was all @pditommaso thank you for the help! Once this is in we are good to go for Nextflow with Flux I think.

 - This will close #3429 

Here is the fixed workflow (before this said the job didn't produce output):

image

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
@@ -63,7 +63,7 @@ class FluxExecutor extends AbstractGridExecutor {
         result << '--job-name="' + getJobNameFor(task) + '"'
 
         // Only write output to file if user doesn't want written entirely to terminal
-        Boolean terminalOutput = task.config.navigate('flux.terminalOutput') as Boolean
+        Boolean terminalOutput = session.config.navigate('flux.terminalOutput') as Boolean
         if ( !terminalOutput ) {
             result << '--output=' + quote(task.workDir.resolve(TaskRun.CMD_LOG))  // -o OUTFILE
         }
--- a/modules/nextflow/src/test/groovy/nextflow/executor/FluxExecutorTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/executor/FluxExecutorTest.groovy
@@ -18,6 +18,7 @@
 package nextflow.executor
 import java.nio.file.Paths
 
+import nextflow.Session
 import nextflow.processor.TaskConfig
 import nextflow.processor.TaskProcessor
 import nextflow.processor.TaskRun
@@ -49,8 +50,12 @@ class FluxExecutorTest extends Specification {
 
     def testGetCommandLine() {
 
-        setup:
-        def executor = [:] as FluxExecutor
+        given:
+        def session = Mock(Session) {
+            getConfig() >> [:]
+        }
+        and:
+        def executor = new FluxExecutor(session: session)
 
         // mock process
         def proc = Mock(TaskProcessor)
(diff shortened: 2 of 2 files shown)


vsoch:
Okay looks like the session is null (and likely needs to be init somehow)
```
Condition failed with Exception:

executor.getSubmitCommandLine(task, Paths.get('/some/path/job.sh')) == ['flux', 'mini', 'submit', '--setattr=cwd=/work/path', '--job-name="nf-my_task"', '--output=/work/path/.command.log', '--queue=delta', '/bin/bash', 'job.sh']
|        |                    |     |     |
|        |                    |     |     /some/path/job.sh
|        |                    |     class java.nio.file.Paths
|        |                    id: null; name: my task; type: null; exit: -; error: null; workDir: /work/path
|        java.lang.NullPointerException: Cannot get property 'config' on null object
|        	at nextflow.executor.FluxExecutor.getSubmitCommandLine(FluxExecutor.groovy:66)
|        	at nextflow.executor.FluxExecutorTest.testGetCommandLine(FluxExecutorTest.groovy:71)
<nextflow.executor.FluxExecutor@5655a930 queueInterval=null lastQueueStatus=null session=null name=null monitor=null memoizedMethodClosure$getStageDir=org.codehaus.groovy.runtime.memoize.Memoize$MemoizeFunction@12234922>

	at nextflow.executor.FluxExecutorTest.testGetCommandLine(FluxExecutorTest.groovy:71)
Caused by: java.lang.NullPointerException: Cannot get property 'config' on null object
	at nextflow.executor.FluxExecutor.getSubmitCommandLine(FluxExecutor.groovy:66)
	... 1 more
```

pditommaso:
The rescue team is coming ..

vsoch:
Oh nice! I didn't know there was a rescue team! We have those in my town when people get lost or injured in the mountains. :laughing:

vsoch:
I'm seeing a few patterns / ideas that could be promising:

- maybe we need to `import nextflow.Session`
- we could try a pattern that checks if it's null first (something like `session ? session ...`) assuming sometimes it's null, other times not.

pditommaso:
Apply this [patch.txt](https://github.com/nextflow-io/nextflow/files/10086949/patch.txt)

pditommaso:
Hold on

pditommaso:
This is the correct one [patch.txt](https://github.com/nextflow-io/nextflow/files/10086957/patch.txt)

vsoch:
Both failed to apply, but I'll look at the changes and see if I can add them to the PR!

vsoch:
oh it seemed to take! not sure why I got an error message. Should we add something to the FluxExecutor when the session isn't defined, or will it always be defined?

pditommaso:
No, the `session` is expected to be defined during normal execution. It's in the test that a fake session needs to be injected

vsoch:
Ah perfect, thank you rescue crew! :raised_hands:

vsoch:
Going to bring up my devcontainer environment and be more responsible to actually try the tests before pushing here!

pditommaso:
Ok, green. Ready to merge it?

vsoch:
I think so - let me do one more sanity check just to be extra sure!

vsoch:
Looks great - ship it! :ship:

vsoch:
I come for the code development, and I stay for the animated gifs! :laughing:

PR #3435, opened 2022-11-25
Add support for custom conda channels

pditommaso:
This PR adds the support for custom conda channels.  

The channels can be specified in the nextflow config e.g. 

```
conda {
  channels = ['bioconda','defaults']
}
```

They are used when creating the Conda environment via punctual packages specified via the `conda` directive e.g. 

```
process foo {
  conda 'samtools'
} 
```
 
Note, custom channels are ignored when the `conda` directive specified a Conda recipe yaml file 

### DIFF
--- a/plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
+++ b/plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
@@ -57,7 +57,7 @@ class WaveClient {
 
     private static Logger log = LoggerFactory.getLogger(WaveClient)
 
-    private static final List<String> DEFAULT_CONDA_CHANNELS = ['defaults', 'conda-forge']
+    private static final List<String> DEFAULT_CONDA_CHANNELS = ['conda-forge','defaults']
 
     final private HttpClient httpClient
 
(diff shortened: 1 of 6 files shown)


plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
```diff
@@ -57,6 +57,8 @@ class WaveClient {
 
     private static Logger log = LoggerFactory.getLogger(WaveClient)
 
+    private static final List<String> DEFAULT_CONDA_CHANNELS = ['defaults', 'conda-forge']
```
jordeu, line 60: The default should be `['conda-forge', 'defaults']`. The channel `conda-forge` should have priority over `defaults`. Is the normal way to use it.

For example if you run:
```
conda create -n test -c defaults -c conda-forge numpy
 ```
  it will install `numpy=1.23.4` from the `defaults` channel. While if you run
```
conda create -n test -c conda-forge -c defaults numpy
```
it will install `numpy=1.23.5` from `conda-forge` channel.

pditommaso: I remember they advice to use `conda-forge` over `defaults`. See here https://conda-forge.org/docs/user/tipsandtricks.html#how-to-fix-it

jordeu: Yes, that's exactly what I mean. To use `conda-forge` over `defaults` the default list of channels should be `['conda-forge', 'defaults']` and not `['defaults', 'conda-forge']`.

pditommaso: Damn, I was convinced I had written the other way around 😄

jordeu (approved):

PR #3438, opened 2022-11-25
At TowerArchiver resolve envar paths relative to baseDir

jordeu:
## Description

When archiving files defined in an environment variable, we should assume that they are relative paths to `baseDir` no to the current working directory.

## Example

Given `NXF_ARCHIVE_DIR="/scratch,s3://fusionfs/archived"` when running a normal pipeline with Tower we have for example this variable `NXF_OUT_FILE=nf-3rvDYTUlUhJ1Ge.txt` that [here](http://github.com/nextflow-io/nextflow/blob/1ed2640a8bba0b43cdb69bb9fdfa5ed3308395bf/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerArchiver.groovy#L133-L133) it ends resolved to this absolute path `/3rvDYTUlUhJ1Ge.txt` because the launcher is running at `/` folder. But then [here](http://github.com/nextflow-io/nextflow/blob/1ed2640a8bba0b43cdb69bb9fdfa5ed3308395bf/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerArchiver.groovy#L125-L125) it is discarded because `baseDir` is `/scratch` (extracted from `NXF_ARCHIVE_DIR`). Notice that both files (`/3rvDYTUlUhJ1Ge.txt` and `/scratch/3rvDYTUlUhJ1Ge.txt`) exists because the [saveCacheFiles](http://github.com/nextflow-io/nextflow/blob/1ed2640a8bba0b43cdb69bb9fdfa5ed3308395bf/plugins/nf-tower/src/main/io/seqera/tower/plugin/CacheManager.groovy#L102-L102) has copied them.

### DIFF
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerArchiver.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerArchiver.groovy
@@ -149,7 +149,7 @@ class TowerArchiver {
 
     protected void archiveFile(String name) {
         if( name )
-            archiveFile(Path.of(name).toAbsolutePath())
+            archiveFile(baseDir.resolve(name))
     }
 
     protected void archiveFile(Path source) {
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerArchiver.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerArchiver.groovy
@@ -17,6 +17,8 @@
 
 package io.seqera.tower.plugin
 
+import nextflow.exception.AbortOperationException
+
 import java.nio.file.FileAlreadyExistsException
 import java.nio.file.Path
 import java.time.temporal.ChronoUnit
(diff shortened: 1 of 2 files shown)


plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerArchiver.groovy
```diff
@@ -149,7 +149,7 @@ class TowerArchiver {
 
     protected void archiveFile(String name) {
         if( name )
-            archiveFile(Path.of(name).toAbsolutePath())
+            archiveFile(baseDir.resolve(name))
```
pditommaso, line 152: It would be better to add a unit test to better keep under control

jordeu:
I've realized that previous solution was wrong because I was assuming that `binDir == workDir`, but that's not necessarily true. Now it is explicitly looking for the Nextflow log files at the working directory. Also I've added a test.

Issue #3353, opened 2022-11-05
Support colon in project names for GCP


danielecook:
## Bug report 

For historical reasons, it is possible for a colon to appear in a GCP project name. When I run Nextflow with Google Batch I observe the following error:

```
Error executing process > 'sayHello (2)'

Caused by:
  io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Unparseable filter: syntax error at line 1, column 58, token ':';
syntax error at line 1, column 59, token '#####/logs/batch_task_logs'
```

Note that this behavior does not occur with the other google executor (life-sciences)

Full traceback:

Note I have redacted the project name and replaced it with `#####`

```

Caused by:
  io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Unparseable filter: syntax error at line 1, column 58, token ':';
syntax error at line 1, column 59, token '#####/logs/batch_task_logs'

com.google.cloud.logging.LoggingException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Unparseable filter: syntax error at line 1, column 58, token ':';
syntax error at line 1, column 59, token '#####/logs/batch_task_logs'
	at com.google.cloud.logging.spi.v2.GrpcLoggingRpc$2.apply(GrpcLoggingRpc.java:208)
	at com.google.cloud.logging.spi.v2.GrpcLoggingRpc$2.apply(GrpcLoggingRpc.java:202)
	at com.google.api.core.ApiFutures$ApiFunctionToGuavaFunction.apply(ApiFutures.java:239)
	at com.google.common.util.concurrent.AbstractCatchingFuture$CatchingFuture.doFallback(AbstractCatchingFuture.java:234)
	at com.google.common.util.concurrent.AbstractCatchingFuture$CatchingFuture.doFallback(AbstractCatchingFuture.java:222)
	at com.google.common.util.concurrent.AbstractCatchingFuture.run(AbstractCatchingFuture.java:133)
	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:31)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1270)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:1038)
	at com.google.common.util.concurrent.AbstractFuture.setException(AbstractFuture.java:808)
	at com.google.api.gax.retrying.BasicRetryingFuture.handleAttempt(BasicRetryingFuture.java:200)
[...]

PR #3464, opened 2022-12-02
Quote the logName in the Cloud Logging filter

aaronegolden:
Quoting will allow special characters, in particular ":", to be parsed. This should prevent the unparseable filter exception in https://github.com/nextflow-io/nextflow/issues/3353.

Signed-off-by: Aaron Golden <aegolden@google.com>


pditommaso:
Awesome thanks

PR #3466, opened 2022-12-02
Use workDir/tmp for collectFile output caching

robsyme:
If we use ephemeral storage in $NXF_TEMP for caching the collectFile outputs, then the caching will always fail if the run is resumed on a new machine, or if the tmp directory is cleaned between runs.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/file/FileCollector.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/file/FileCollector.groovy
@@ -62,6 +62,8 @@ abstract class FileCollector implements Closeable {
 
     boolean resumable
 
+    Path cacheBase
+
     private Path tempDir
 
     private boolean created
@@ -237,20 +239,28 @@ abstract class FileCollector implements Closeable {
 
         // verify if a cached list exists
         final hash = makeHash()?.toString()
-        Path temp = hash ? FileHelper.getLocalTempPath().resolve("${hash}.collect-file") : null
+
+        Path cache = hash ? (cacheBase ?: FileHelper.localTempPath)
+            .resolve(hash.substring(0,2))
+            .resolve(hash.substring(2))
+            .resolve("collectFile.cache") : null
 
         // try to retrieve cached files
         List<Path> items = null
-        if( resumable && temp ) {
-            items = retrieveCachedFiles(temp)
+        if( resumable && cache ) {
+            items = retrieveCachedFiles(cache)
         }
 
         // get the list of files to generated
         if( items == null ) {
             items = saveTo0(target)
             // save the list of collected files
-            if( cacheable && temp ) {
-                cacheCollectedFile(items, temp)
+            if( cacheable && cache ) {
+                if( FilesEx.mkdirs(cache.parent) ) {
+                    cacheCollectedFile(items, cache)
+                } else {
+                    throw new IOException("Unable to create temporary directory: $cache -- Make sure a file with the same name doesn't already exist and you have write permissions")
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


robsyme:
Sorry, this path change should be implemented elsewhere. I'll revisit over the weekend. This PR can be ignored for the moment.

robsyme:
Ok, this looks ok for review. I've kept the changes as modest as possible.

The important change is the addition of `cacheBase` property to the `FileCollector` class. 

If `cacheBase` remains unset, use `FileHelper.localTempPath` to store the collectFile cache. The only change is that before, we used
```groovy
// This is where the collectFile cache used to be stored:
FileHelper.getLocalTempPath()
  .resolve("${hash}.collect-file")
```
and now (if cacheBase is unset), we use the two-directory hash method:
```groovy
FileHelper.getLocalTempPath()
  .resolve(hash.substring(0,2))
  .resolve(hash.substring(2))
  .resolve(".collect-file")
```

I changed this so that it matches the behaviour when `cacheBase` _is_ set. In that case, we use a subdirectory (following the same path pattern) to write the collected results:
```groovy
cacheBase
  .resolve(hash.substring(0,2))
  .resolve(hash.substring(2))
  .resolve(".collect-file")
```

bentsherman (approved):

bentsherman:
@pditommaso Tested and it works on my end. I think this one is ready to go.

modules/nextflow/src/main/groovy/nextflow/file/FileCollector.groovy
```diff
@@ -237,20 +239,28 @@ abstract class FileCollector implements Closeable {
 
         // verify if a cached list exists
         final hash = makeHash()?.toString()
-        Path temp = hash ? FileHelper.getLocalTempPath().resolve("${hash}.collect-file") : null
```
pditommaso, line 240: My concern is that temp path was set on the local disk on purpose because the `collectFile` sorts the collect items using using the local file system. I wonder how this can work using moving it over the pipeline work dir when using a remote path such as `s3://something`

robsyme: Great point. I'll see if we can fix this by writing locally and then moving into workDir. Failing that, perhaps this list of filenames can be stored in the db.

robsyme: Without any changes, this works when using S3 as a workdir. 

I tested this by running Nextflow on an EC2 VM and using the awsbatch executor, running the workflow twice:

image

The collectfile cache is written as expected to `${workDir}/tmp/<hash>/<hash>/collectFile.cache`:

image

pditommaso: I believe you, but HOW? 😄

bentsherman: Rob, I think Paolo is asking about how this change affects the sorting of the files.

My understanding is that this meta file only contains the list of paths but not the contents. On AWS Batch, the chunk files will still be sorted on local scratch storage, whereas the meta file will be generated in memory and written once to the work dir. So sorting performance should not be affected.

robsyme: Yup, you're right Ben. 

The file in question (the one I'm calling `collectFile.cache`) does not store the sorted contents, but is a serialized list of paths.

pditommaso:
@robsyme you are right! I was confusing the sorting path with the index storing path. I made some minor changes 👉 762ba19c7. Thanks for this patch!

PR #3473, opened 2022-12-05
Rename Google location to region

pditommaso:
The use of the `google.location` setting is not consistent with the Google docs where it tend to use the term region instead. 

https://cloud.google.com/batch/docs/locations#regions

I've renamed the setting `location` to `region`.


plugins/nf-google/src/main/nextflow/cloud/google/batch/client/BatchClient.groovy
```diff
@@ -74,25 +74,25 @@ class BatchClient {
     }
 
     Job submitJob(String jobId, Job job) {
-        final parent = LocationName.of(projectId, location)
+        final parent = LocationName.of(projectId, region)
```
pditommaso, line 77: even tho the API use `location` instead 🤦

bentsherman:
Are we sure they won't add support for zones in the future? Considering that the API uses `location` instead of `region` throughout, I figured they decided to only support a few regions at first for simplicity.

mbookman:
My 2 cents is that the configuration should mirror the API in this case, and that the Nextflow documentation should help people understand the differences between `location` and `regions` and `zones` in the context of the Life Sciences API.

The `location` is used for the API endpoint. Many people don't care which API endpoint Nextflow uses, but for those who want to make sure that Nextflow only sends information to the Life Sciences API within a particular geographic region, then they should use the `location` field.

The `region` and `zone` fields are typically more relevant to people in that it impacts the location of the VMs used to execute individual tasks. Someone familiar with the Life Sciences API and any of the other tools that use it would assume that the `region` field for Nextflow configuration is going to impact the VM locations, rather than the API endpoint.

PR #3479, opened 2022-12-08
Update .gitignore to include vscode dot dir

marcodelapierre:
@pditommaso I know this is a bit silly, but it is quite useful to ignore the .vscode directories that get created by MS Visual Studio Code .. for those who use it for development purposes.


pditommaso:
Makes sense!

PR #3492, opened 2022-12-13
Don't set realtime to `complete - start` by default

bentsherman:
The `realtime` trace metric is the runtime of the `.command.sh` script, and it is measured by the command wrapper. However, it is initially set to `complete - start`, presumably as a backup in case the `.command.trace` file can't be recovered.

This is a bad practice from a data science perspective, because then you don't know whether `realtime` is including extra things like staging time (consider the benchmarks we are running right now). Instead, Nextflow should leave `realtime` blank if it can't recover the trace file, and let the user decide how to deal with missing values. The user may choose to impute or ignore the missing values, or just use `complete - start` for consistency, but it should be their choice.


pditommaso:
You are right, but I think removing this it will break features such as timeline when the trace is not enabled. 

I think it's better to keep in this form and plan this change for a trace format v3 that could include other changed like for example #3579

bentsherman:
Reopening until we have a more concrete plan for how to address this issue

PR #3495, opened 2022-12-13
Fix versions to use secrets in the docs  [ci skip]

mribeirodantas:
Clarifies in the docs when secrets was first introduced as a preview and how to use it, and then when it became enabled by default in a later version.

The work done in this PR sums up with the one done in this other PR: https://github.com/nextflow-io/nextflow/pull/3491

The tricky situation here, as explained in the commit messages, is that the use of the `NXF_ENABLE_SECRETS` environment variable requires version `21.09.0-edge` of Nextflow, but its usage changed in version `22.09.2-edge`. I think this should be well explained in the documentation, as two people in the past 24 hours had issues with that (linked below):

https://nextflow.slack.com/archives/C02T98A23U7/p1670956679351909 (by Roger Olivella)
https://nextflow.slack.com/archives/C02T98A23U7/p1670905262984539 (by Nick Youngblut)


pditommaso:
Not sure this is useful. It makes hard even for me to figure out what version to use! 😄

mribeirodantas:
To be honest, if I hadn't run into two people having issues with that in a short period of time, I would agree with you it's too much info haha 😆

But apparently, people are getting confused with the current docs 😓

mribeirodantas:
In the end, what should we do with this PR, @pditommaso ? 😅

pditommaso:
Made my opinionated changes! 👉  edcfba48

PR #3498, opened 2022-12-15
Refactor the AWS configuration 

pditommaso:
The aim of this PR is to refactor the AWS config and move it away from the [Global](https://github.com/nextflow-io/nextflow/blob/d1bb2fe2906303c6c5386e47136b7d5d104b7fec/modules/nf-commons/src/main/nextflow/Global.groovy#L37-L37) object and implement it instead in the nf-amazon module


bentsherman:
Nevermind, I see this is the upstream branch.

PR #3499, opened 2022-12-15
Micronaut CI support

pditommaso:
The goal of this PR is to refactor Nextflow internal component by using the dependency inject pattern via [Micronaut IoC](https://docs.micronaut.io/latest/guide/#ioc) or remove the sue of global singleton classes


pditommaso:
Outdated

PR #3508, opened 2022-12-18
Add warning for exception where channel option must not have comma before

mribeirodantas:
I believe this example with `stdout emit` without any warning gives the impression commas for options are optional (which is something I thought for a while). This warning should fix this misconception.


bentsherman:
Is this true? What error do you get if you use a comma?

pditommaso:
I could not find any example in the tests 

https://github.com/nextflow-io/nextflow/blob/9eb8fe77c70bddea85d2e03a40a669d5713d09a8/modules/nextflow/src/test/groovy/nextflow/script/params/ParamsOutTest.groovy#L1159-L1162

but in principle, the syntax should be 

```
output: 
stdout, emit: something
```

even tho, the above could be reduced to 

```
output: 
stdout(), emit: something
```

which brings that inconsistency

mribeirodantas:
> Is this true? What error do you get if you use a comma?

```
process FOO {
  input:
    val x
  output:
    stdout emit: my_output
  script:
  """
  echo $x
  """
}

workflow {
  FOO(10).my_output.view()
}
```
The code above works fine. If I add a comma after `stdout` (or even `stdout()`), I get the following error:
```
N E X T F L O W  ~  version 22.10.4
Launching `example.nf` [determined_linnaeus] DSL2 - revision: a49e4b3b62
Script compilation error
- file : /Users/mribeirodantas/example.nf
- cause: Unexpected input: '{' @ line 1, column 13.
   process FOO {
               ^

1 error
```

pditommaso:
Ugly, even more the error message!

mribeirodantas:
Yeah 😅 

So... What should we do then? 😬 Change Nextflow to make it work with the comma or add this note to the documentation? 😅

pditommaso:
I think it's worth highlighting but it's not an inconsistency in the syntax because the comma is needed to separate the directive value from the optional attribute. Since in the case of `stdout` there's no value no comma is needed

mribeirodantas:
Oh, that makes sense! I think it's also worth to explain this in the note, to clarify it's not an inconsistency. I will do it! 😉

mribeirodantas:
@pditommaso What do you think of it now? 😄

mribeirodantas:
LGTM? 👀 😬

docs/dsl2.rst
```diff
@@ -180,6 +180,8 @@ The ``emit`` option can be used also to name the stdout::
         sayHello.out.verbiage.view()
     }
 
+.. warning:: Channel options in processes are always added with a comma before to separate the directive value from the optional attribute. The case for ``stdout`` is an
+    exception, as you must not add the comma right after ``stdout`` for adding the ``emit`` channel option. It happens this way because ``stdout`` doesn't have a value, as it is usually the case.
```
bentsherman, line 184:
```suggestion
.. warning::
  Optional params for a process input/output are always prefixed with a comma, except for ``stdout``. Because
  ``stdout`` does not have an associated name or value like other types, the first param should not be prefixed.
```

mribeirodantas:
Thanks, @bentsherman ! It reads better indeed! 🥳

bentsherman (approved):

docs/dsl2.rst
```diff
@@ -180,6 +180,9 @@ The ``emit`` option can be used also to name the stdout::
         sayHello.out.verbiage.view()
     }
 
+.. warning::
```
pditommaso, line 183:
```suggestion
.. note::
```

PR #3511, opened 2022-12-21
Fix FilePorter concurrency issue

pditommaso:
This PR address a concurrency issue in the `FilePorter` logic.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy
@@ -165,6 +165,9 @@ class FilePorter {
         return result
     }
 
+    /**
+     * Model a file stage requirement
+     */
     @Canonical
     static class FileCopy {
         final Path source
--- a/modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy
@@ -212,11 +212,9 @@ class FilePorter {
          */
         Path addToForeign(Path path) {
             // copy the path with a thread pool
-            synchronized (owner) {
-                final copy = getCachePathFor(path, stageDir, owner.stagingTransfers)
-                foreignPaths << copy
-                return copy.target
-            }
+            final copy = owner.getCachePathFor(path, stageDir)
+            foreignPaths << copy
+            return copy.target
         }
 
         /**
@@ -331,7 +329,7 @@ class FilePorter {
         }
     }
 
-    static protected FileCopy getCachePathFor(Path sourcePath, Path stageDir, Map<FileCopy,FileTransfer> stagingTransfers) {
+    synchronized protected FileCopy getCachePathFor(Path sourcePath, Path stageDir) {
         final dirPath = stageDir.toUriString() // <-- use a string to avoid changes in the dir to alter the hashing
         int i=0
         while( true ) {
--- a/modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy
@@ -96,7 +96,7 @@ class FilePorter {
     }
 
     protected FileTransfer getOrSubmit(FileCopy copy) {
-        synchronized (stagingTransfers) {
+        synchronized (this) {
             FileTransfer transfer = stagingTransfers.get(copy)
             if( transfer == null ) {
                 transfer = createFileTransfer(copy.source, copy.target)
@@ -213,9 +213,9 @@ class FilePorter {
         Path addToForeign(Path path) {
             // copy the path with a thread pool
             synchronized (owner) {
-                final target = getCachePathFor(path, stageDir)
-                foreignPaths << new FileCopy(path, target)
-                return target
+                final copy = getCachePathFor(path, stageDir, owner.stagingTransfers)
+                foreignPaths << copy
+                return copy.target
             }
         }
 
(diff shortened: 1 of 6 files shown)


modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy
```diff
@@ -156,16 +165,24 @@ class FilePorter {
...
      */
     static class Batch  {
 
+        final private FilePorter owner
+
         /**
          * Holds the list of foreign files to be transferred
          */
-        private List<Path> foreignPaths = new ArrayList<>()
+        private List<FileCopy> foreignPaths = new ArrayList<>(100)
```
pditommaso, line 185: Instead of the file of source paths use a list of pairs. Each pair represent the source file location and the target location of the staging.

modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy
```diff
@@ -194,8 +212,11 @@ class FilePorter {
          */
         Path addToForeign(Path path) {
             // copy the path with a thread pool
-            foreignPaths << path
-            return getCachePathFor(path, stageDir)
+            synchronized (owner) {
```
pditommaso, line 215: Not 100% this is really needed

modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy
```diff
@@ -76,32 +77,31 @@ class FilePorter {
...
         if( batch.size() ) {
             log.trace "Stage foreign files: $batch"
-            submitStagingActions(batch.foreignPaths, batch.stageDir)
+            submitStagingActions(batch.foreignPaths)
             log.trace "Stage foreign files completed: $batch"
         }
     }
 
-    protected FileTransfer createFileTransfer(Path source, Path stageDir) {
-        final stagePath = getCachePathFor(source,stageDir)
```
pditommaso, line 90: This is the main change. Instead of recomputing it, use the target path that was computed in the previous step

modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy
```diff
@@ -194,8 +212,11 @@ class FilePorter {
          */
         Path addToForeign(Path path) {
             // copy the path with a thread pool
-            foreignPaths << path
-            return getCachePathFor(path, stageDir)
+            synchronized (owner) {
+                final target = getCachePathFor(path, stageDir)
```
jordeu, line 216: This PR will fix the problem but it will download the same files twice, because still it's possible to call this `getCachePathFor` while a transfer is running on the background.

Maybe here we can check if there is already a `FileCopy` at the `FilePorter.stagingTransfers` for the current `path` and reuse it instead of calling a new `getCachePathFor` that it may return a different one.

pditommaso: Not sure, as long as the `FileCopy` identity is made by both source and target paths, if a second copy is created for the same path it will receive the same transfer object here 

https://github.com/nextflow-io/nextflow/blob/4bfbc7e6426016341cdd30e591d372a08f1f34ab/modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy#L99-L110

pditommaso: and this could happen if there's a race condition when checking the existence of the target path and both return false

pditommaso: but it's true that if the check is made while the transfer is on-going it will cause of duplicate download

pditommaso: the problem is that it's hard to know if it's an ongoing download or a broken/partial copy left by another run.

jordeu: But can another run be at `FilePorter.stagingTransfers` ? or it has only the ones from the current one?

pditommaso: Do you mean another nextflow run? it may happen if the work dir is the same 

https://github.com/nextflow-io/nextflow/blob/4bfbc7e6426016341cdd30e591d372a08f1f34ab/modules/nextflow/src/main/groovy/nextflow/executor/Executor.groovy#L120-L120

This can be improved by adding the session id to the stage path. Something like 

```
getWorkDir().resolve("stage-$sessionId")
```

pditommaso: Need to fix the tests, but what about this 👉 [b91629d](https://github.com/nextflow-io/nextflow/pull/3511/commits/b91629d7250e5fe34da21247ec6e24f37ddd5b19)

modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy
```diff
@@ -331,16 +331,19 @@ class FilePorter {
...
-    static protected Path getCachePathFor(Path sourcePath, Path stageDir) {
+    static protected FileCopy getCachePathFor(Path sourcePath, Path stageDir, Map<FileCopy,FileTransfer> stagingTransfers) {
         final dirPath = stageDir.toUriString() // <-- use a string to avoid changes in the dir to alter the hashing
         int i=0
         while( true ) {
             final uniq = List.of(sourcePath, dirPath, i++)
             final hash = CacheHelper.hasher(uniq).hash().toString()
             final targetPath = getCacheDir0(stageDir, hash).resolve(sourcePath.getName())
+            final result = new FileCopy(sourcePath, targetPath)
+            if( stagingTransfers.containsKey(result) )
```
pditommaso, line 342: if found, there's an on-going transfer. Just take it

modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy
```diff
@@ -331,16 +331,19 @@ class FilePorter {
...
+                return result
             final exist = targetPath.exists()
             if( !exist || checkPathIntegrity(sourcePath, targetPath) )
```
pditommaso, line 345: Otherwise use the previous logic

jordeu (approved):

Issue #3514, opened 2022-12-26
Add support for Fusion file system for Sarus containerised task


pditommaso:
### Context 

The [Fusion file system](https://www.nextflow.io/docs/latest/fusion.html) allows using S3-compatible object storage as the task work directory. 

This feature requires mounting the Fusion driver in the container process via a Fuse device. 

When using Docker or Podman this can be achieved using the following options 

```
--device /dev/fuse --cap-add SYS_ADMIN 
```

See #3337 for details. 

### Goal

The goal of this feature is to allow the use of Fusion driver in containers run via the [Sarus container](https://www.nextflow.io/docs/edge/container.html#sarus) engine.

pditommaso:
When trying to run a Fusion enabled container with Sarus, the following error is reported 

```
$ sarus run wave.seqera.io/wt/82f195fe9205/nextflow/rnaseq-nf:v1.1
2022-12-26_20:10:11 Mounting FusionFS v0.6.1-e6da3a5
8:10PM ERR main.go:111 > mount fail error="exec: \"/bin/fusermount\": stat /bin/fusermount: no such file or directory"
```

The documentation reports how to allow [Sarus to access FUSE file system mounted in the host](https://sarus.readthedocs.io/en/stable/user/user_guide.html?highlight=fuse#bind-mounts). 

However, there's no mention of mounting FUSE file system in the container space. Wondering @Madeeks has any suggestions at this regard?

Issue #3513, opened 2022-12-26
Add support for Fusion file system for Singularity containerised task 


pditommaso:
### Context 

The [Fusion file system](https://www.nextflow.io/docs/latest/fusion.html) allows using S3-compatible object storage as the task work directory. 

This feature requires mounting the Fusion driver in the container process via a Fuse device. 

When using docker or podman this can be achieved using the following options 

```
--device /dev/fuse --cap-add SYS_ADMIN 
```

See #3337 for details 

### Goal

The goal of this feature is to allow the use of Fusion driver in containers run via the Singularity and Apptainer engines.

pditommaso:
This issue requires adding in the above snippet the singularity options needed to run the fusion driver with the support for FUSE 

https://github.com/nextflow-io/nextflow/blob/d22337cd9e7391aa9112eb1a4e775103a45659da/modules/nextflow/src/main/groovy/nextflow/container/ContainerConfig.groovy#L87-L96

jordeu:
The current problem is that the way of running fusion with singularity is like:

```
singularity exec <options> docker://<image> /opt/fusion/fusionfs <command>
```
so it's not part of the options, is a pre-command that works as an entrypoint.

pditommaso:
Never a joy

pditommaso:
I think what it could be done is to introduce a `FusionOptions` class in place of using a string here 

https://github.com/nextflow-io/nextflow/blob/827ee97e7132ad64989a68f4a3e30befc71676e4/modules/nextflow/src/main/groovy/nextflow/fusion/FusionHelper.groovy#L53-L53

Then it can be used to customise the exec option by the `SingularityBuilder`. What do you think?

jordeu:
Yes, that seems the less intrusive option. 

An other option can be to undo the [entrypoint](http://github.com/nextflow-io/nextflow/blob/827ee97e7132ad64989a68f4a3e30befc71676e4/modules/nextflow/src/main/groovy/nextflow/container/ContainerBuilder.groovy#L78-L79) deprecation. 

At the end now to use fusion means to use it as entrypoint, in some builders (like docker, podman) the concept exists and we pass it as an option, in the others it means to prepend it before the real command. It should work on all of them in the same way. 

Maybe the only problem of this option is that then the user can override it.

pditommaso:
Yes, all the point was to delegate Wave the handling of the entrypoint

jordeu:
[...]

PR #3516, opened 2022-12-26
Add support for Fusion file system to Slurm and LSF executors

pditommaso:
This PR adds support for [Fusion file system](https://www.nextflow.io/docs/edge/fusion.html) to Slurm and LSF grid executors. 

The PR implements the following changes 
* Mark the [GridTaskHandler](https://github.com/nextflow-io/nextflow/blob/d3a0aa5610da7070f22f7f660d38ca4be027a25d/modules/nextflow/src/main/groovy/nextflow/executor/GridTaskHandler.groovy#L52-L52) as [FusionAwareTask](https://github.com/nextflow-io/nextflow/blob/d3a0aa5610da7070f22f7f660d38ca4be027a25d/modules/nextflow/src/main/groovy/nextflow/fusion/FusionAwareTask.groovy#L32-L32)
* Use the `NXF_CHDIR` variable as the common pattern to specify the job work directory instead of relying on grid-specific directives
* When fusion execution is required, submit the job execution by creating an inline job wrapper piped via `stdin` special file, instead of creating a temporary launcher file 

This feature for production usage requires #3513 or #3514


pditommaso:
Think the Fusion support is ready to go. Main changes: 
* the task work dir is defined via [NXF_CHDIR](https://github.com/nextflow-io/nextflow/pull/3516/files#diff-a1332f56f6111b668185a943c2cda504de84d6afebdcda9a01c9416e0fe7a2ffR86) variable instead of relying on batch scheduler directive 
* Batch scheduler log is sent to `/dev/null` to prevent writing `.command.log` in the local fs
* Improved submit error troubleshooting including [the actual submit command](https://github.com/nextflow-io/nextflow/pull/3516/files#diff-c317d610e9c87f9c425afaedfb79ed6ca92083197d0e0ac9be6b2299aab5860eR239-R243) in the error message  
* Add support for Grid Engine other than Slurm and LSF

jordeu (approved):
I tested and it's really nice, it's magic! :smile: 

The only thing that was not intuitive to me is the need to explicitly set `fusion.exportAwsAccessKeys = true` to make it work.

I'm thinking that with the upcoming support for more data stores this will became a bit of a mess to always explicitly configure and maybe we can have a default way of sharing credentials (using standard env vars) that works out of the box and only when you use a different approach you need to tune your configuration.

pditommaso:
> The only thing that was not intuitive to me is the need to explicitly set fusion.exportAwsAccessKeys = true to make it work.

I agree that's not great but how to handle it better? In principle storage authentication should be delegated to the corresponding cloud infra. 

When using Fusion on-prem I'm expecting the use of S3-compatible storage such as Minio or Ceph. Don't think in this scenario the use of cloud-hosted object storage is a real use case.

jordeu:
MinIO and CEPH also use `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to manage the authentication. So on 99% of the use cases when using Grid platforms + S3 the user will need to pass `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `AWS_S3_ENDPOINT` to the container environment.

pditommaso:
Included in version 23.02.0-edge

JosephLalli:
So great to see this.

Another approach to supporting Fusion in grid-based executors (#3205) supported a wider array of grid-based executors. I am at an institution with a condor-based HTC setup.

Does this PR allow for the use of Fusion in any executors beyond Slurm and LSF? If so, what limits it to those two executors, and what work would need to be done (by me) to allow the use of Fusion with condor? If it matters, I intend to use it with our on-premises Minio-based s3-compatible storage.

pditommaso:
Currently, it is not supported, but in principle, it should be easy to add it. 

To enable this capability what nextflow does it to pipe the launch command to the summit command stdin instead of creating  a wrapper script. 

For example when using Slurm instead of running 

```
sbatch .command.run
```

it does 

```
cat .command.run | sbatch
```

Does Condor allow the same?

bentsherman:
According to [my research](https://github.com/nextflow-io/nextflow/pull/3205#issue-1370457479) it does support stdin.

Also some notes [here](https://github.com/nextflow-io/nextflow/pull/3205#discussion_r968908817) that may be relevant to Condor support.

pditommaso:
I guess so, it's needed somebody that takes care of implementing and validating it

bentsherman:
@JosephLalli if I draft a Fusion / Condor PR, can you test it in your environment? We don't have a Condor environment so makes it hard to develop new features for it. As long as you can run Nextflow, you should be able to build and test it from a branch.

JosephLalli:
Absolutely. I can’t promise any kind of automated test, but I can absolutely run whatever you’d like here. University of Wisconsin CHTC leadership is very eager to get Nextflow working on Condor, and my work has become their test case. I think Fusion paired with our local s3 storage setup should solve a lot of the headaches that the Condor folks have had in the past.



________________________________

bentsherman:
Excellent, I will tag you when we have a PR for it. We'll just need you to run a small pipeline (e.g. `rnaseq-nf`) on your cluster with Fusion.

I also don't know if Fusion works with S3-compatible storage on HPC yet, or at least it hasn't been tested. So we might be waiting for that support to arrive.

pditommaso:
Yes, it does. This an example config 

```
aws {
  client {
    endpoint = '<MINIO ENDPOINT>'
    s3PathStyleAccess = true
  }
  accessKey = '<ACCESS>'
  secretKey = '<SECRET>'
}

wave { enabled = true }

fusion { enabled = true; exportAwsAccessKeys = true }
```

Issue #2839, opened 2022-04-29
Nextflow does not look for AWS_SESSION_TOKEN environment variable


jwarwick-delfi:
It appears that SSO / temporary credentials are only working when specified in an aws config file, not the standard environment variables.

The issue appears to be here, it doesn't look like nextflow is looking for AWS_SESSION_TOKEN.

https://github.com/nextflow-io/nextflow/blob/d30841a58c6a3b7b798b7d6597dcca93ae4f243c/modules/nf-commons/src/main/nextflow/Global.groovy#L116

Issue #1724, opened 2020-09-04
Configuring AWS temporary session token in nextflow.config and environment 


mageshwarm:
As continuity to https://github.com/nextflow-io/nextflow/pull/1265, please enable us to configure sessionToken in nextflow.config and export AWS_SESSION_TOKEN. Currently aws_session_token supported by ~/.aws/credentials only.

pditommaso:
It should be doable adding add the read the `sessionToken` here 

https://github.com/nextflow-io/nextflow/blob/e94e6877c2dc9bed983e873b62e302a08843c891/modules/nf-commons/src/main/nextflow/Global.groovy#L80-L89


and the corresponding env var (not sure about the exact name) 

https://github.com/nextflow-io/nextflow/blob/e94e6877c2dc9bed983e873b62e302a08843c891/modules/nf-commons/src/main/nextflow/Global.groovy#L91-L96

robin-cruk:
the variable is `AWS_SESSION_TOKEN`

kolarji2:
Hi, what is the status of this issue? Is it supported now? How can I configure nextflow with AWS_SESSION_TOKEN?

pditommaso:
This is *partially* supported when the session token is defined in the AWS credentials file. 

However, the refresh of the session token is not supported. Follow #2295 for details

kolarji2:
Hi, thank you for your quick response. But it is not working for me :-(

I put my credentials to: ~/.aws/credentials
I also put them to nextflow.config  and environment variables:
aws {
    accessKey = "${AWS_ACCESS_KEY_ID}"
    secretKey = "${AWS_SECRET_ACCESS_KEY}"
    sessionToken = "${AWS_SESSION_TOKEN}"
    region = 'us-east-1'
}

But I get: Service: Amazon S3; Status Code: 403; Error Code: InvalidAccessKeyId

kolarji2:
I use: nextflow version 22.10.3.5834
And I am sure the credentials are valid. As I can access the files using boto3 directly. That is my current workaround, I pass the file as "val" instead of "path" object and load them from python instead.

pditommaso:
You should be able to check in the `.nextflow.log` file if the token was grabbed. The fact you are passing the file as "val" instead of "path" rings me something wrong

kolarji2:
Thanks, it is working now if I do not set environment variables and I do not set the nextflow.config. Then it is picked from .aws/credentials

[...]

PR #1265, opened 2019-08-07
AWS temporary session token support

nkrumm:
This PR address #934 and adds the ability for nextflow to use temporary AWS credentials from the STS service. There are three credential strings involved AWS_ACCESS_KEY, AWS_SECRET_ACCESS_KEY, and AWS_SESSION_TOKEN, and are read either from the environment or from the `~/.aws/credentials` file.

Note this does depend on a separate PR in the nextflow-s3fs library: https://github.com/nextflow-io/nextflow-s3fs/pull/17. Also, originally this was part of https://github.com/nextflow-io/nextflow/pull/1248, but now is split into two.

PR #3517, opened 2022-12-30
add support for AWS_SESSION_TOKEN

jchorl:
There's a few issues floating around about this:
- https://github.com/nextflow-io/nextflow/issues/1724
- https://github.com/nextflow-io/nextflow/issues/2839
- https://github.com/nextflow-io/nextflow/pull/1265

Figured I'd take a stab at implementing. Wrote a test, but also ran locally and verified that the session token actually works.

I did combine the `AWS_ACCESS_KEY`/`AWS_SECRET_KEY` and `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` creds cases. Sure, technically, it means someone could supply an invalid config and it might work. But also, it unifies the session token logic.

I didn't add this to the config options, because who's putting an ephemeral token in their nextflow config??

Truthfully, I believe we're on the wrong path, and that we should just let the AWS Java SDK resolve credentials. It is opinionated in how it resolves credentials. If we delegated, we wouldn't have to account for every weird possible way of specifying aws creds. Perhaps there's a good reason not to though.

Anyway, let me know what you think. I'd love to have this in.

### DIFF
--- a/modules/nf-commons/src/main/nextflow/Global.groovy
+++ b/modules/nf-commons/src/main/nextflow/Global.groovy
@@ -98,6 +98,7 @@ class Global {
 
         String a
         String b
+        String c
 
         if( config && config.aws instanceof Map ) {
             a = ((Map)config.aws).accessKey
@@ -112,13 +113,14 @@ class Global {
 
         // as define by amazon doc
         // http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html
-        if( env && (a=env.AWS_ACCESS_KEY_ID) && (b=env.AWS_SECRET_ACCESS_KEY) )  {
-            log.debug "Using AWS credentials defined by environment variables AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY"
-            return [a, b]
-        }
+        // supports AWS_ACCESS_KEY, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SECRET_KEY, and AWS_SESSION_TOKEN
+        if( env && (a=(env.AWS_ACCESS_KEY_ID ?: env.AWS_ACCESS_KEY)) && (b=(env.AWS_SECRET_ACCESS_KEY ?: env.AWS_SECRET_KEY)) )  {
+            if ( env && (c=env.AWS_SESSION_TOKEN) ) {
+                log.debug "Using AWS credentials defined by environment variables AWS_ACCESS_KEY(_ID)?, AWS_SECRET(_ACCESS)?_KEY and AWS_SESSION_TOKEN"
+                return [a, b, c]
+            }
 
-        if( env && (a=env.AWS_ACCESS_KEY) && (b=env.AWS_SECRET_KEY) ) {
-            log.debug "Using AWS credentials defined by environment variables AWS_ACCESS_KEY and AWS_SECRET_KEY"
+            log.debug "Using AWS credentials defined by environment variables AWS_ACCESS_KEY(_ID)? and AWS_SECRET(_ACCESS)?_KEY"
             return [a, b]
         }
 
--- a/modules/nf-commons/src/test/nextflow/GlobalTest.groovy
+++ b/modules/nf-commons/src/test/nextflow/GlobalTest.groovy
@@ -38,6 +38,7 @@ class GlobalTest extends Specification {
         Global.getAwsCredentials0([AWS_ACCESS_KEY: 'x', AWS_SECRET_KEY: '222'], null) == ['x','222']
         Global.getAwsCredentials0([AWS_ACCESS_KEY_ID: 'q', AWS_SECRET_ACCESS_KEY: '999'], null) == ['q','999']
         Global.getAwsCredentials0([AWS_ACCESS_KEY: 'x', AWS_SECRET_KEY: '222',  AWS_ACCESS_KEY_ID: 'q', AWS_SECRET_ACCESS_KEY: '999'], null) == ['q','999']
+        Global.getAwsCredentials0([AWS_ACCESS_KEY_ID: 'x', AWS_SECRET_ACCESS_KEY: '222',  AWS_SESSION_TOKEN: 'foobar',], null) == ['x','222', 'foobar']
 
         Global.getAwsCredentials0([AWS_ACCESS_KEY_ID: 'q', AWS_SECRET_ACCESS_KEY: '999'], [aws:[accessKey: 'b', secretKey: '333']]) == ['b','333']
         Global.getAwsCredentials0(null, [aws:[accessKey: 'b', secretKey: '333']]) == ['b','333']


pditommaso:
> Truthfully, I believe we're on the wrong path

I agree. I think as well, using the `AWS_SESSION_TOKEN` was a mistake because it needs to be refreshed over time. 

There's an [open PR](https://github.com/nextflow-io/nextflow/pull/3498) that re-organises the AWS config moving away from the current static object and uses the [STS client](https://github.com/nextflow-io/nextflow/blob/7da48d176af1a072d6270ac3f29c406a73774478/plugins/nf-amazon/src/main/nextflow/cloud/aws/AmazonClientFactory.groovy#L277-L287) to properly handle this.

jchorl:
> > Truthfully, I believe we're on the wrong path
> 
> I agree. I think as well, using the `AWS_SESSION_TOKEN` was a mistake because it needs to be refreshed over time.
> 
> There's an [open PR](https://github.com/nextflow-io/nextflow/pull/3498) that re-organises the AWS config moving away from the current static object and uses the [STS client](https://github.com/nextflow-io/nextflow/blob/7da48d176af1a072d6270ac3f29c406a73774478/plugins/nf-amazon/src/main/nextflow/cloud/aws/AmazonClientFactory.groovy#L277-L287) to properly handle this.

This PR looks like a very welcome change. If I could add a couple cents -

I wouldn't dabble in passing ARNs to nextflow. The Java SDK [supports config files](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials.html).

Config files [support a `role_arn` arg](https://docs.aws.amazon.com/sdkref/latest/guide/feature-assume-role-credentials.html).

Basically, I think this subtle issue is the same reason we're having this conversation - that we're trying to wrap or abstract over aws creds. My personal preference is to just let the sdk be the sdk - it seems to work for everyone.

That said, I'm sure you're more privvy to much more user insight, and if the people must have nextflow-specific config (rather than controlling the host configs), I get it.

Thanks for putting together that PR. It would solve what I was trying to do here - just more involved than the effort I could put in. Thanks again for all the hard work. Feel free to close this one, or merge if you feel yours will take a long time to vet/get merged.

pditommaso:
> I wouldn't dabble in passing ARNs to nextflow

Do you mean this? 

https://github.com/nextflow-io/nextflow/blob/b14e51a630c1d363fee5a822c51b26efb7173771/plugins/nf-amazon/src/main/nextflow/cloud/aws/AmazonClientFactory.groovy#L112-L112

jchorl:
> > I wouldn't dabble in passing ARNs to nextflow
> 
> Do you mean this?
> 
> https://github.com/nextflow-io/nextflow/blob/b14e51a630c1d363fee5a822c51b26efb7173771/plugins/nf-amazon/src/main/nextflow/cloud/aws/AmazonClientFactory.groovy#L112-L112

Yeah exactly - not sure why we wouldn't just instantiate a java client and let it run. Diving into aws creds the last few days (see appendix), theyre quite complex, and configurable, and trying to wrap them will likely lead to:
1. Pain in keeping up with all the configs that they do support (e.g. if you're supporting assume-role, are you also going to support source-profile to figure out which aws profile to use to assume the role?)
2. Potential bugs in munging nf configs into aws configs

Appendix - feel free to stop reading here:
The actual thing I've been battling is 66k `nxf_s3_download` statements in a single `.command.run` absolutely destroying ec2 imds credential fetching

I've implemented a task-side cache for aws cli creds but holy, re-spawning the aws-cli that many times is causing a ton of latency and api call (read: rate limit) overhead.

pditommaso:
I don't remember exactly the reason why the `assumeRoleArn` has been added recently. @bentsherman do you have any clue? 


> The actual thing I've been battling is 66k nxf_s3_download statements in a single .command.run

Not sure how much it can be done on NF side to improve this when using the plain `aws` CLI tool. Have you tried using [Fusion](https://www.nextflow.io/docs/latest/fusion.html), I think it should solve this problem.

jchorl:
Huh, neat, I haven't. I have hand-patched in [goofys](https://github.com/kahing/goofys) for particularly bad processes.

I was looking through the fusion and wave code. I kinda assume you define a custom `scheme://` and hijack that to `mount` the s3 path via NFS (using goofys or something similar) into the container. Looks like you need to run privileged - mostly makes sense. I'm curious why you need to use wave - on cursory glance, I couldn't tell if it puts some kind of daemon or startup script in the container to do the dirty mounting.

If you could point me to the plumbing that actually makes s3 available to the container (if that's open-source), that'd be rad. I'd love to take this thing for a spin - although it looks like it does bypass a lot of the OG nextflow run scripting so will take some time to validate.

pditommaso:
> I kinda assume you define a custom scheme:// and hijack that to mount the s3 path via NFS (using goofys or something similar) into the container

Yes, it's just map S3 paths to `/fusion/s3/something` in the task container

https://github.com/nextflow-io/nextflow/blob/b14e51a630c1d363fee5a822c51b26efb7173771/modules/nextflow/src/main/groovy/nextflow/fusion/FusionHelper.groovy#L78-L91


> Looks like you need to run privileged - mostly makes sense

So far yes. Tho modern Linux kernel and container runtimes do not require it. We are looking how to make it runnable without using privileged permission. Said that, using privileged permission in an ephemeral cluster does not pose security risks.

> I'm curious why you need to use wave

Because the Fusion client runs in the container. Wave allows us to add it on-the-fly without rebuilding existing container images

> If you could point me to the plumbing that actually makes s3 available to the container

Think you are referring the above link 


> it looks like it does bypass a lot of the OG nextflow run scripting so will take some time to validate

it should not! it's just matter to enable it in the nextflow config, nothing else should change

pditommaso:
Regarding the `assumeRoleArn` property, it was added because the STS client requires it 

https://github.com/nextflow-io/nextflow/blob/c856768c0ad8ec8dcc30067f3b3a90b3ec8499a7/plugins/nf-amazon/src/main/nextflow/cloud/aws/AmazonClientFactory.groovy#L281-L283

jchorl:
> Think you are referring the above link

I was curious about the actual process that `wave` injects into the container that mounts s3 as a filesystem. I'd be curious to poke through that code if it's open-source.

> Regarding the assumeRoleArn property, it was added because the STS client requires it

Ah, if you instantiate a raw STS client, sure. But I'd think you don't need to do that. In AWS native config, you can tell it to assume a role: https://docs.aws.amazon.com/sdkref/latest/guide/feature-assume-role-credentials.html

And you can even use multiple profiles, and tell the java client which profile to pick based on an env var or java config property: https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials.html#credentials-profiles

So (correct me if wrong), I think there are two approaches here:
1. Just instantiate a raw java client. It already supports all the credential methods, including assume role. These are configured outside of Nextflow, in the standard AWS config way. This includes config files, env vars, and other. But it's super extensible, and the natural AWS way.
2. Wrap certain parts of credential config in nextflow - e.g. have config fields for access keys, have a config field for an assume-role, etc. And based on nextflow configs, you yourself figure out whether to instantiate a normal client or an STS client.

I'd think 1 is far preferable? Because it supports the totality of what AWS supports, without you needing to play "catch-up" and port aws cred features into nextflow config. I could be missing something here.

pditommaso:
For some legacy reason, Nextflow parses the AWS config file to pick the credentials. 

https://github.com/nextflow-io/nextflow/blob/aws-config-refactor/plugins/nf-amazon/src/main/nextflow/cloud/aws/config/AwsConfig.groovy#L159-L162

But maybe you are right, it would be better to delegate this to the SDK, with exception if the user provided the aws keys or the profile in the nextflow config file

pditommaso:
Likely something like this 

```
        if( accessKey && secretKey ) {
            final creds = new BasicAWSCredentials(accessKey, secretKey)
            return new AWSStaticCredentialsProvider(creds)
        }
        
        if( profile ) {
            return new ProfileCredentialsProvider(profile)
        }
        
        return new DefaultAWSCredentialsProviderChain()
```

pditommaso:
> I was curious about the actual process that wave injects into the container that mounts s3 as a filesystem. I'd be curious to poke through that code if it's open-source.

This is not OSS at this time

pditommaso:
Let's move this discuss to https://github.com/nextflow-io/nextflow/pull/3555

PR #3535, opened 2023-01-07
update of Scripts `à la carte`

nick-youngblut:
Provided explicit examples of using input values & paths in a la carte scripts.

Signed-off-by: Nick Youngblut <nyoungb2@gmail.com>


docs/process.rst
```diff
@@ -134,20 +134,29 @@ To use a language other than Bash, simply start your process script with the cor
...
+        output:
+        path "table.txt"
+    
         """
         #!/usr/bin/python
-
-        x = 'Hello'
-        y = 'world!'
-        print "%s - %s" % (x,y)
+
```
mribeirodantas, line 156:
```suggestion
        // Reads the content of a file, and writes it to `table.txt`
```

mribeirodantas (changes requested):
For beginners, this can be slightly difficult to read so I would suggest to add a comment, such as this one.

mribeirodantas (approved):

docs/process.rst
```diff
@@ -134,20 +134,29 @@ To use a language other than Bash, simply start your process script with the cor
...
+        output:
+        path "table.txt"
+    
         """
         #!/usr/bin/python
-
-        x = 'Hello'
-        y = 'world!'
-        print "%s - %s" % (x,y)
+        // Reads the content of a file, and writes it to `table.txt`
```
mribeirodantas, line 156:
```suggestion
        # Reads the content of a file, and writes it to `table.txt`
```

It's Python now, not Groovy/Nextflow so we should use the appropriate comment symbol.

mribeirodantas (changes requested):

docs/process.rst
```diff
@@ -134,20 +134,29 @@ To use a language other than Bash, simply start your process script with the cor
 `shebang <http://en.wikipedia.org/wiki/Shebang_(Unix)>`_. For example::
 
     process perlStuff {
+        input:
+        val(x)
```
mribeirodantas, line 138:
```suggestion
        val(x)
        output:
        stdout
```

docs/process.rst
```diff
@@ -134,20 +134,29 @@ To use a language other than Bash, simply start your process script with the cor
...
         #!/usr/bin/perl
 
-        print 'Hi there!' . '\n';
+        print 'value: ${x}';
         """
     }
 
     process pythonStuff {
+        input:
+        path(x)
```
mribeirodantas, line 149:
```suggestion
        path(x)
        output:
        stdout
```

mribeirodantas (changes requested):

mribeirodantas (approved):

pditommaso:
Hi, sorry for the late reply. The goal of the example is to show how to invoke other interpreters via the shebang declaration. 

For this reason, I think it's better to add input and keep the example self-contained. Changed  it to add the workflow invocation 6d6b3543

nick-youngblut:
@pditommaso if the goal of this example was to:

>show how to invoke other interpreters via the shebang declaration

...then maybe including a separate example in the docs of how to actually write python/perl code using shell variable interpretation (as in this PR) would be helpful?

pditommaso:
Likely. The key thing it's not going to be a Shell variable, but it should use the syntax of the target scripting language ie. python/perl/R etc. I think this is what confuses people

PR #3537, opened 2023-01-10
Document `userEmulation` option in docker config scope

llewellyn-sl:
Add `userEmulation` to docker scope table.
Alphabetize docker scope table. 

Signed-off-by: Llewellyn vd Berg <113503285+llewellyn-sl@users.noreply.github.com>

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -296,11 +296,6 @@ sudo                Executes Docker run command as ``sudo`` (default: ``false``)
 temp                Mounts a path of your choice as the ``/tmp`` directory in the container. Use the special value ``auto`` to create a temporary directory each time a container is created.
 tty                 Allocates a pseudo-tty (default: ``false``).
 userEmulation       If true, runs all containers as the same user executing the ``nextflow run`` command, and mounts the user home folder in each container (default: ``false``). 
-
-
-
-
-
 ================== ================
 
 The above options can be used by prefixing them with the ``docker`` scope or surrounding them by curly
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -283,18 +283,24 @@ The following settings are available:
 ================== ================
 Name                Description
 ================== ================
-enabled             Turn this flag to ``true`` to enable Docker execution (default: ``false``).
+enabled             If true, enables Docker execution (default: ``false``).
+engineOptions       This attribute can be used to provide any option supported by the Docker engine i.e. ``docker [OPTIONS]``.
 envWhitelist        Comma separated list of environment variable names to be included in the container environment.
+fixOwnership        If true, fixes ownership of files created by the docker container (default: ``false``).
 legacy              Uses command line options removed since version 1.10.x (default: ``false``).
+mountFlags          Add the specified flags to the volume mounts e.g. `mountFlags = 'ro,Z'`
+registry            The registry from where Docker images are pulled. It should be only used to specify a private registry server. It should NOT include the protocol prefix, i.e. ``http://``.
+remove              Cleans up the container after execution (default: ``true``). For details see: https://docs.docker.com/engine/reference/run/#clean-up---rm .
+runOptions          This attribute can be used to provide any extra command line options supported by the ``docker run`` command. For details see: https://docs.docker.com/engine/reference/run/ .
 sudo                Executes Docker run command as ``sudo`` (default: ``false``).
 tty                 Allocates a pseudo-tty (default: ``false``).
 temp                Mounts a path of your choice as the ``/tmp`` directory in the container. Use the special value ``auto`` to create a temporary directory each time a container is created.
-remove              Clean-up the container after the execution (default: ``true``). For details see: https://docs.docker.com/engine/reference/run/#clean-up---rm .
-runOptions          This attribute can be used to provide any extra command line options supported by the ``docker run`` command. For details see: https://docs.docker.com/engine/reference/run/ .
-registry            The registry from where Docker images are pulled. It should be only used to specify a private registry server. It should NOT include the protocol prefix i.e. ``http://``.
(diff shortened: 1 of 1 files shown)


docs/config.rst
```diff
@@ -283,18 +283,24 @@ The following settings are available:
...
+enabled             If true, enables Docker execution (default: ``false``).
+engineOptions       This attribute can be used to provide any option supported by the Docker engine i.e. ``docker [OPTIONS]``.
 envWhitelist        Comma separated list of environment variable names to be included in the container environment.
+fixOwnership        If true, fixes ownership of files created by the docker container (default: ``false``).
 legacy              Uses command line options removed since version 1.10.x (default: ``false``).
+mountFlags          Add the specified flags to the volume mounts e.g. `mountFlags = 'ro,Z'`
+registry            The registry from where Docker images are pulled. It should be only used to specify a private registry server. It should NOT include the protocol prefix, i.e. ``http://``.
+remove              Cleans up the container after execution (default: ``true``). For details see: https://docs.docker.com/engine/reference/run/#clean-up---rm .
+runOptions          This attribute can be used to provide any extra command line options supported by the ``docker run`` command. For details see: https://docs.docker.com/engine/reference/run/ .
```
llewellyn-sl, line 294:
```suggestion
runOptions          Passes extra command line options supported by the ``docker run`` command. For details, see: https://docs.docker.com/engine/reference/run/ .
```

docs/config.rst
```diff
@@ -283,18 +283,24 @@ The following settings are available:
 ================== ================
 Name                Description
 ================== ================
-enabled             Turn this flag to ``true`` to enable Docker execution (default: ``false``).
+enabled             If true, enables Docker execution (default: ``false``).
+engineOptions       This attribute can be used to provide any option supported by the Docker engine i.e. ``docker [OPTIONS]``.
```
llewellyn-sl, line 287:
```suggestion
engineOptions       Passes any options supported by the Docker engine, i.e. ``docker [OPTIONS]``.
```

docs/config.rst
```diff
@@ -283,18 +283,24 @@ The following settings are available:
...
-fixOwnership        Fixes ownership of files created by the docker container.
-engineOptions       This attribute can be used to provide any option supported by the Docker engine i.e. ``docker [OPTIONS]``.
-mountFlags          Add the specified flags to the volume mounts e.g. `mountFlags = 'ro,Z'`
+tty                 Allocates a pseudo-tty (default: ``false``).
+userEmulation       If true, runs all containers as the same user executing the ``nextflow run`` command, and mounts the user home folder in each container (default: ``false``). 
+
+
+
+
+
```
bentsherman, line 303:
```suggestion
```

bentsherman (approved):

pditommaso:
This is undocumented by choice. Likely it should be removed

bentsherman:
I know you said that maybe the `docker.userEmulation` should be removed, but it's used by nf-core/rnaseq now: https://github.com/nf-core/rnaseq/blob/master/nextflow.config#L165

It's quite useful... I was gonna make a PR to add it to rnaseq-nf because without it the pipeline creates everything as `root` in my work directory. I think it would be good to document for local execution.

pditommaso:
I believe we should promote best practices and provide nf-core folks with a better solution to handle this. 

That option was added in the early days, and it does unsafe things e.g. mount the home, mount the password file, etc.  

https://github.com/nextflow-io/nextflow/blob/bb8369603fc6d66cce306e2ef0d1aa40f3a4eef1/modules/nextflow/src/main/groovy/nextflow/container/DockerBuilder.groovy#L39-L39


Likely it can be replaces with `process.containerOptions = 'u $(id -u):$(id -g)'` if needed. 

@ewels @drpatelh what's the need to add `docker.userEmulation` in nf-core config files?

ewels:
It was added years ago, but I think it was likely the user + group file ownership. But maybe more than just that? I have a feeling it was a broad stroke "fix a load of disparate Docker problems in one go" kind of thing.

ewels:
ok, I did a little digging, and it was all 👉🏻  @drpatelh 👈🏻  😆 

https://github.com/nf-core/tools/commit/43e21b349d827a8690066601fc8e8474c70ab230

A bit more digging takes me to an active discussion around the same time, main focus is here: https://github.com/nf-core/sarek/pull/378. I think that's probably where the template update came from, so actually it was all 👉🏻  @maxulysse 👈🏻  🤣 

@maxulysse @pontus @mahesh-panchal @FriederikeHanssen and @heuermh were discussing problems with Spark. I think that's probably where this originally came from.

pditommaso:
Spark aka [Apache Spack](https://spark.apache.org/) ?? 😳

ewels:
@pditommaso - still want to try to get rid of it?? 😆

FriederikeHanssen:
What would be the good practice to compare against?

pontus:
I also think it's unfortunate having something doing many different things to fix issues.

The problem as I see it is that they are all needed at times (sometimes all three of them that I remember) and while all of those can probably be replaced by specific `containerOptions`, I don't see a good way for how a DSL2 module would go about extending those.

* For Linux, for most tools, you don't want to run as root not to get messed up file permissions, also affects things like sending signals to processes or how full a filesystem is allowed to become before write fails.
* Some tools require NSS information (e.g. the running user existing in `/etc/passwd` or `/etc/group`, so requires mapping of passwd file or similar)
* Some tools also require the running user to be able to write in the home directory from NSS (so requires mapping of home directory).

bentsherman:
@Mxrcon the feature is still present, just not documented

Mxrcon:
Thanks @bentsherman, it might be a change on the institution configurations side.

PR #3548, opened 2023-01-13
Fix submit command in error message for grid executors that pipe wrapper script

### LINKED ISSUE #3024: Errors during job submission are not logged anywhere
## Bug report 

Errors that happen during job submission of a cluster job do not appear to be logged anywhere and so they are quite difficult to debug. Moreover, some job submission arguments are not logged in a visible way, making it harder to reproduce.

For example, the [troubleshooting instructions](https://nf-co.re/usage/troubleshooting#manual-debugging-on-clusters-using-schedulers) tell one to rerun the submission command in the work directory. However, the submission command may include other parameters that are not present in the error output shown. In my case at least, for LSF cluster, the arguments to `bsub` (the job-submission command) are documented in the .command.run file in comments that the user cannot be expected to find except by accident. Without including those arguments the error may not happen.

It would be preferable if the error from the failed job submission was reported to the user somehow.

bentsherman:
Closes #3024 

Some grid executors pipe the wrapper script to their submit command (e.g. LSF). In this case, when a task fails and Nextflow prints an error message, the submit command is incorrect because it doesn't include the pipe. For example:
```
Caused by:
  Failed to submit process to grid scheduler for execution

Command executed:

  bsub

Command exit status:
  -

Command output:
  (empty)
```

The submit command in this case should be `bsub < .command.run`.

This PR changes the error message to append `< .command.run` to the command if the wrapper script was piped into the submit command.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/GridTaskHandler.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/GridTaskHandler.groovy
@@ -185,7 +185,11 @@ class GridTaskHandler extends TaskHandler {
             final exitStatus = process.waitFor()
 
             if( exitStatus ) {
-                throw new ProcessNonZeroExitStatusException("Failed to submit process to grid scheduler for execution", result, exitStatus, builder.command())
+                final command = pipeScript
+                    ? builder.command() << '<' << TaskRun.CMD_RUN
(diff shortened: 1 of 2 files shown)


pditommaso (approved):

pditommaso:
Added unit test 👍

PR #3554, opened 2023-01-16
Fix #3553 (breaking environment variable values containing `=`)

### LINKED ISSUE #3553: Issue with outputting string variables containing equal sign (=)

If I use output type `env` and the string variable contains a equal sign `=`, only the part before the `=` will be output, even though the whole string is captured in `.command.env`

I don't think this is the expected behaviour, as standard bash can have string variables containing `=`
```
OUTPUT_VAR='PART1=PART2'
echo $OUTPUT_VAR

PART1=PART2
```

```
process TEST {
    output:
    env OUTPUT_VAR

    shell:
    '''
    OUTPUT_VAR='PART1=PART2'
    '''
}


workflow {
    TEST().view()
}
```

### Program output 

Terminal output:
```
PART1
```

Expected output:
```
PART1=PART2
```

### Environment 

* Nextflow version: 22.10.4
* Java version: OpenJDK 19.0.1
* Operating system: MacOS 12.6.2
* Bash version: zsh 5.8.1

### Additional context
@pditommaso suggests it is likely related to https://github.com/nextflow-io/nextflow/blob/c492ad227d77b729b3c951767395b83a95d0ff6f/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy#L1496-L1504

mribeirodantas:
The use of `tokenize` will separate all substrings separated by
the symbol (equal sign in this case). If an environment variable
contains an equal sign, `tokenize` will split in more than one
place and this will break the string that should be saved. Using
`split` with the `limit` parameter we fix this.

For more details, check https://github.com/nextflow-io/nextflow/issues/3553

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -1497,7 +1497,7 @@ class TaskProcessor {
         final env = workDir.resolve(TaskRun.CMD_ENV).text
         final result = new HashMap(50)
         for(String line : env.readLines() ) {
-            def (k,v) = line.tokenize('=')
+            def (k, v) = line.split('=', limit=2);
             result.put(k,v)
         }
         return result
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -1497,7 +1497,7 @@ class TaskProcessor {
         final env = workDir.resolve(TaskRun.CMD_ENV).text
         final result = new HashMap(50)
         for(String line : env.readLines() ) {
-            def (k, v) = line.split('=', limit=2);
+            def (k, v) = line.split('=', 2);
             result.put(k,v)
         }
         return result


mribeirodantas:
What do you think, @pditommaso ? 😄

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -1497,7 +1497,7 @@ class TaskProcessor {
         final env = workDir.resolve(TaskRun.CMD_ENV).text
         final result = new HashMap(50)
         for(String line : env.readLines() ) {
-            def (k,v) = line.tokenize('=')
+            def (k, v) = line.split('=', limit=2);
```
pditommaso, line 1500:
```suggestion
            def (k, v) = line.split('=', 2);
```

Nextflow is not python!! 😆

mribeirodantas: 😬

mribeirodantas:
LGTM? 😬

pditommaso:
Was still not working for me. I've solved with this 40d56735

PR #3555, opened 2023-01-17
AWS config refactor using AWS credentials chain

pditommaso:
The goal of this PR is to refactor the AWS authentication logic, so that:
* The AWS credentials chain is [used by default](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html)
* The `accessKey` and `secretKey` can be provided alternatively via nextflow.config file in the `aws` scope
* The AWS profile can be provided alternatively via nextflow.config file in the `aws` scope

### DIFF
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/AmazonClientFactory.groovy
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/AmazonClientFactory.groovy
@@ -17,6 +17,9 @@
 
 package nextflow.cloud.aws
 
+import java.nio.file.Files
+import java.nio.file.Paths
+
 import com.amazonaws.AmazonClientException
 import com.amazonaws.ClientConfiguration
 import com.amazonaws.auth.AWSCredentialsProvider
@@ -45,6 +48,7 @@ import com.amazonaws.services.securitytoken.model.GetCallerIdentityRequest
 import groovy.transform.CompileStatic
 import groovy.transform.Memoized
 import groovy.util.logging.Slf4j
+import nextflow.SysEnv
 import nextflow.cloud.aws.config.AwsConfig
 import nextflow.exception.AbortOperationException
 /**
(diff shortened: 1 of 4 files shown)


bentsherman:
I have looked at it, but it is not clear to me what remains to be done. It looks mostly complete. Can you elaborate which parts of the code are incomplete and what manual tests I need to run?

bentsherman:
I tried it with my regular setup and it worked. But I'm not sure which authentication mechanism was used. My creds file looks like this:
```
[default]
aws_access_key_id=...
aws_role_arn=...
aws_secret_access_key=...
aws_session_expiration=2023-01-24T21:09:36.000Z
aws_session_token=...
```

I didn't specify a profile or access/secret key in my config, so I guess that means it used the default credential chain?

pditommaso:
Think so, but it would be better to add some debug line to be able to determine it from the log.

Also it should be tested that works:

* Static creds in the nextflow config
* Profile name in the nextflow config (was not working for me)
* STS role is picked correctly

bentsherman:
Okay. And the `profile` config option, is that a profile in `.aws/config` or an IAM instance profile? I thought it was the former but I am confused by these lines:
```groovy
//        if( !this.accessKey && !this.profile && !fetchIamRole() )
//            throw new AbortOperationException("Missing AWS security credentials -- Provide access/security keys pair or define an IAM instance profile (suggested)")
```

pditommaso:
We may need to review this. Not sure how (on mobile)

bentsherman:
I'm pretty sure it corresponds to a profile in `.aws/config`. But IAM instance profile is also a valid option from an EC2 instance.

I can confirm, Nextflow did use the default credential chain when I tested it. So I will test with a custom AWS profile.

bentsherman:
I made a custom AWS profile that is just a copy of my default profile, and I specified it in the Nextflow config, and it worked.

bentsherman:
But I am still not sure about how it works when running from an EC2 instance. I assume the default credential chain will use the instance profile to get all the credentials it needs. I will try to test that next.

pditommaso:
Yeah, it should but worth to test it

pditommaso:
To be reviewed the use of `anonymous` access with the new client

jchorl:
Just wanted to say thanks for this. Built the latest master, tested locally, it did what I needed it to.

pditommaso:
Happy it works for you.

PR #3568, opened 2023-01-24
Check for incorrect errorStrategies

mribeirodantas:
There are many situations that go silent in Nextflow, including when a non-existent (or typo) error strategy is provided. This Pull Request throws an exception whenever this happens, and includes unit tests for such situations. Because it's not possible to assess closure content at runtime, it's not possible to check for non-existent (or typo) error strategies when a closure is provided as `errorStrategy` so the check will not be performed in such circumstances.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy
@@ -812,6 +812,11 @@ class ProcessConfig implements Map<String,Object>, Cloneable {
      *      The {@link ProcessConfig} instance itself.
      */
     ProcessConfig errorStrategy( strategy ) {
+        listOfStrategies = ['terminate', 'finish', 'ignore', 'retry']
+        if (!listOfStrategies.contains(strategy)) {
+            throw new IllegalArgumentException("Unrecognized error strategy: " + strategy + ". Available strategies are: " + listOfStrategies.join(', '))
+        }
+
         configProperties.put('errorStrategy', strategy)
         return this
     }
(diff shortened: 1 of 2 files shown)


mribeirodantas:
Based on the `test_integration` CI test that failed, @bentsherman just explained to me that sometimes it can be a closure (dynamic expression), so the code in this PR should handle closures too.

modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy
```diff
@@ -813,6 +813,13 @@ class ProcessConfig implements Map<String,Object>, Cloneable {
      *      The {@link ProcessConfig} instance itself.
      */
     ProcessConfig errorStrategy( strategy ) {
+        listOfStrategies = ['terminate', 'finish', 'ignore', 'retry']
```
pditommaso, line 816: Instead of creating a new list for each invocation, it could be used the [ErrorStrategy](https://github.com/nextflow-io/nextflow/blob/34a27733961adbc66635e1a05c3fab8e3e468295/modules/nextflow/src/main/groovy/nextflow/processor/ErrorStrategy.groovy#L25-L25) enum e.g. 

```
strategy.toUpperCase() !in ErrorStrategy.values()
```

maybe even better to declare constant at the top of class, something like this 

```
private static final List<String> VALID_ERROR_STRATEGIES = ErrorStrategy.values()
```

modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy
```diff
@@ -813,6 +813,13 @@ class ProcessConfig implements Map<String,Object>, Cloneable {
...
+        if (strategy !instanceof Closure) {
+            if (!listOfStrategies.contains(strategy)) {
+                throw new IllegalArgumentException("Unrecognized error strategy: " + strategy + ". Available strategies are: " + listOfStrategies.join(', '))
```
pditommaso, line 819: You can use string interpolation instead of `+` concatenation to make it more readable

pditommaso:
Merged manually 954cd921 👍

PR #3579, opened 2023-01-26
Use cgroup for memory trace metrics

### LINKED ISSUE #3496: Option to query trace metrics from cgroup instead of pid
Reminder from our meeting today. The trace metrics are currently taken from `/proc/$pid/...` which captures memory and I/O usage of the task process.

In a container, we should really get these metrics from the cgroup, which gives the container-wide usage, because that is the metric that watchdogs use to determine whether a container is using too much memory and needs to be killed.

Example:
```bash
cat /sys/fs/cgroup/memory/memory.usage_in_bytes 
```

Important for resource optimization in Tower.

cc @jordeu @pditommaso

bentsherman:
Closes #3496 

Changes the `nxf_stat` function in the wrapper script to overwrite the memory-related trace metrics with the corresponding cgroup values if they are available. This way, if a task is running in a container, Nextflow will track the memory usage of the entire container instead of just the task (and its children).

Example:
```console
$ nextflow run hello -with-trace -with-docker ; cat trace.txt
task_id vmem    rss     peak_vmem       peak_rss
2       10.4 MB 10.7 MB 15.1 MB 15.1 MB
3       10 MB   9.7 MB  13.6 MB 13.6 MB
1       9.7 MB  9.9 MB  11.7 MB 11.7 MB
4       9.1 MB  9.3 MB  13.2 MB 13.2 MB
```

We might need to modify the precondition for using these metrics. I noticed that in my WSL environment, the `/sys/fs/cgroup` directory exists even without a container, so that would throw off my metrics for a non-containerized run.

### DIFF
--- a/modules/nextflow/src/main/resources/nextflow/executor/command-trace.txt
+++ b/modules/nextflow/src/main/resources/nextflow/executor/command-trace.txt
@@ -66,6 +66,14 @@ nxf_stat() {
 
     [ $NXF_DEBUG = 1 ] && echo -e "++ stat SUM=${sum[*]}"
 
+    ## use memory metrics from cgroup if available
+    if [ -e /sys/fs/cgroup ]; then
+        sum[2]=$(($(cat /sys/fs/cgroup/memory/memory.memsw.usage_in_bytes) / 1024))
+        sum[3]=$(($(cat /sys/fs/cgroup/memory/memory.usage_in_bytes) / 1024))
+        sum[4]=$(($(cat /sys/fs/cgroup/memory/memory.memsw.max_usage_in_bytes) / 1024))
+        sum[5]=$(($(cat /sys/fs/cgroup/memory/memory.max_usage_in_bytes) / 1024))
+    fi
+
     ## calc peaks
     for i in {1..7}; do
         if [ ${sum[i]} -lt ${cpu_peak[i]} ]; then
--- a/modules/nextflow/src/main/resources/nextflow/executor/command-trace.txt
+++ b/modules/nextflow/src/main/resources/nextflow/executor/command-trace.txt
@@ -67,7 +67,7 @@ nxf_stat() {
     [ $NXF_DEBUG = 1 ] && echo -e "++ stat SUM=${sum[*]}"
 
     ## use memory metrics from cgroup if available
-    if [ -e /sys/fs/cgroup ]; then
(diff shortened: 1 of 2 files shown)


bentsherman:
Also, without this patch, the same run as above reports zero memory usage:
```console
$ nextflow run hello -with-trace trace.1.txt -with-docker ; cat trace.1.txt
task_id vmem    rss     peak_vmem       peak_rss
4       0       0       0       0
1       0       0       0       0
3       0       0       0       0
2       0       0       0       0
```

So that suggests that we are underestimating the memory usage of containerized tasks by ~10 MB at least, and more than that if other things like Fusion are running in the background.

pditommaso:
Interesting. Worth mentioning nextflow does it's best to measure the resources of the actually job task

https://github.com/nextflow-io/nextflow/blob/9f0dac313c7dc5bdc59faa06e8872fcced61404f/modules/nextflow/src/main/resources/nextflow/executor/command-trace.txt#L180-L180

therefore not sure what it's aiming this PR

pditommaso:
Ok, i've go the [rationale](https://github.com/nextflow-io/nextflow/issues/3496). But this approach it will introduce a different approach when using container execution vs non-container. 

I think it should be reported as an new (optional) metric, tho this will make non-trivial to be collected in Tower

modules/nextflow/src/main/resources/nextflow/executor/command-trace.txt
```diff
@@ -66,6 +66,14 @@ nxf_stat() {
 
     [ $NXF_DEBUG = 1 ] && echo -e "++ stat SUM=${sum[*]}"
 
+    ## use memory metrics from cgroup if available
+    if [ -e /sys/fs/cgroup ]; then
```
jordeu, line 70: I think that this may always exists, even if you are not in a container.

bentsherman: I think I will just insert `true` or `false` from the bash wrapper builder based on whether the task is containerized.

bentsherman:
> But this approach it will introduce a different approach when using container execution vs non-container.

Well, it depends on how you define a "task". When you run a task in a container, is the container part of the task? The OOM killer on AWS certainly thinks so.

pditommaso:
We'll add this when it's will be possible to add handle more data points on tower side. closing foe now

PR #3588, opened 2023-01-31
Reinstate exit code for Spack tests

marcodelapierre:

Hi @pditommaso , 

following this commit of yours:
https://github.com/nextflow-io/nextflow/commit/868374d3d591113a6fa3364d4e520d2ef3374aaf#diff-86b11d6f83a372628c1957303e638eccaa9583281cdb8657845b1222b6ff570a

I have re-pulled and retested, as I had done previously, and cannot find issues with these tests, they run successfully.

I believe the bugfix you applied in `SpackCacheTest` probably fixed the issue?

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/spack/SpackCache.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/spack/SpackCache.groovy
@@ -299,7 +299,7 @@ class SpackCache {
     }
 
     @PackageScope
-    void runCommand( String cmd ) {
+    int runCommand( String cmd ) {
         log.trace """spack env create
                      command: $cmd
                      timeout: $createTimeout""".stripIndent()


marcodelapierre:
If you need to test the Spack commands, this is how you get it up and running, with Python 3.7(?) required:
```
spack_root="PICK-YOUR-DIR"

mkdir -p $spack_root
git clone git@github.com:spack/spack $spack_root/spack
cd $spack_root/spack
git checkout v0.19.0  # to make sure you are using a stable release
. share/spack/setup-env.sh
cd -
```

modules/nextflow/src/main/groovy/nextflow/spack/SpackCache.groovy
```diff
@@ -299,7 +299,7 @@ class SpackCache {
     }
 
     @PackageScope
-    void runCommand( String cmd ) {
+    int runCommand( String cmd ) {
```
pditommaso, line 302: Yep, I think the problem was the test. But the return value is not used anywhere. so why adding it?

marcodelapierre: isn't is used in all the unit tests as exitcode, for logical (true/false) checks? 

e.g. see your second edit in https://github.com/nextflow-io/nextflow/commit/868374d3d591113a6fa3364d4e520d2ef3374aaf#diff-86b11d6f83a372628c1957303e638eccaa9583281cdb8657845b1222b6ff570a

pditommaso:
We discussed. It's not needed

PR #3571, opened 2023-01-25
Improve documentation and add warnings re: params-file vs configuration

jfy133:
Improves documentation about this issue: https://github.com/nextflow-io/nextflow/issues/2662

PR #3589, opened 2023-01-31
Improve documentation and add warnings re: params-file vs configuration

jfy133:
Improves documentation about this issue: https://github.com/nextflow-io/nextflow/issues/2662

Originally in #3571 but messed up the DCO and a git reset to try and fix these.

Text originally my myself and updated by @bentsherman.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -1237,10 +1237,26 @@ facilitates rapid iterations, inspections of any pipeline as well as debugging.
     $ nextflow run nextflow-io/hello -with-tower
 
 - Invoke the nextflow pipeline execution with a custom parameters ``YAML/JSON`` file.
-  The parameters which are specified through this mechanism are merged with the resolved configuration (base configuration and profiles) and only the common fields are overwritten by the ``YAML/JSON`` file.::
 
-    $ nextflow run main.nf -params-file pipeline_params.yml
+  The parameters which are specified through this mechanism are merged with the resolved configuration (base configuration and profiles) and only the common fields are overwritten by the ``YAML/JSON`` file. ::
 
+  $ nextflow run main.nf -params-file pipeline_params.yml
+
+  For example, the following params file in YAML format::
+
+    alpha: 1
+    beta: 'foo'
+
+  Or in JSON format::
+
+    {
+      "alpha": 1,
+      "beta": "foo"
+    }
+
+  Is equivalent to the following command line::
+
+    $ nextflow run main.nf --alpha 1 --beta foo
 
 --------------------
 self-update
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -1236,9 +1236,7 @@ facilitates rapid iterations, inspections of any pipeline as well as debugging.
 
     $ nextflow run nextflow-io/hello -with-tower
 
-- Invoke the nextflow pipeline execution with a custom parameters ``YAML/JSON`` file.
-
-  The parameters which are specified through this mechanism are merged with the resolved configuration (base configuration and profiles) and only the common fields are overwritten by the ``YAML/JSON`` file. ::
+- Invoke the nextflow pipeline execution with a custom parameters ``YAML/JSON`` file.::
 
   $ nextflow run main.nf -params-file pipeline_params.yml
 
@@ -1257,6 +1255,8 @@ facilitates rapid iterations, inspections of any pipeline as well as debugging.
   Is equivalent to the following command line::
 
     $ nextflow run main.nf --alpha 1 --beta foo
+
+  The parameters which are specified through this mechanism are merged with the resolved configuration (base configuration and profiles) and only the common fields are overwritten by the ``YAML/JSON`` file. 
 
 --------------------
 self-update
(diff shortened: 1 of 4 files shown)


bentsherman (approved):

docs/cli.rst
```diff
@@ -1237,10 +1237,26 @@ facilitates rapid iterations, inspections of any pipeline as well as debugging.
     $ nextflow run nextflow-io/hello -with-tower
 
 - Invoke the nextflow pipeline execution with a custom parameters ``YAML/JSON`` file.
-  The parameters which are specified through this mechanism are merged with the resolved configuration (base configuration and profiles) and only the common fields are overwritten by the ``YAML/JSON`` file.::
 
-    $ nextflow run main.nf -params-file pipeline_params.yml
+  The parameters which are specified through this mechanism are merged with the resolved configuration (base configuration and profiles) and only the common fields are overwritten by the ``YAML/JSON`` file. ::
```
pditommaso, line 1241:
```suggestion
  The parameters which are specified through this mechanism are merged with the resolved configuration (base configuration and profiles). The values provided via params file overwrites the ones with the same name in the Nextflow configuration::
```

pditommaso: Actually I'd put the sentence "The values provided via params file overwrites the ones with the same name in the Nextflow configuration file" in a tip *after* the example

docs/cli.rst
```diff
@@ -1236,11 +1236,27 @@ facilitates rapid iterations, inspections of any pipeline as well as debugging.
...
+    {
+      "alpha": 1,
+      "beta": "foo"
+    }
+
+  Is equivalent to the following command line::
+
+    $ nextflow run main.nf --alpha 1 --beta foo
+
+  The parameters which are specified through this mechanism are merged with the resolved configuration (base configuration and profiles). The values provided via params file overwrites the ones with the same name in the Nextflow configuration file
```
pditommaso, line 1259:
```suggestion
  The parameters which are specified through this mechanism are merged with the resolved configuration (base configuration and profiles). The values provided via params file overwrites the ones with the same name in the Nextflow configuration file.
```

docs/config.rst
```diff
@@ -26,6 +26,11 @@ first override the same ones that may appear in the second one, and so on.
 .. tip::
   If you want to ignore any default configuration files and use only the custom one, use ``-C <config file>``.
 
+.. note::
+  Pipeline parameters (i.e. the ``params`` scope) should be overwritten through CLI options (``--something value``)
+  or a params file (``-params-file params.yml``). If you supply parameters through a custom config file (``-c my_config``),
+  it will be applied to the pipeline script but not to previous config files, such as the project's ``nextflow.config`` file.
+  See the :ref:`CLI page <cli>`. for more information about how to supply params through CLI options or a params file.
```
pditommaso, line 33: Not understanding this

jfy133: This is in response to the related bug reported here: https://github.com/nextflow-io/nextflow/issues/2662

Where params in a custom config file supplied by `-c`are not actually sent to processes in the pipeline (even though e.g. tower reports they are correctly resolved).

I can try and rephrase to say more bluntly: use only param-files or cli `--` for specifying pipeline parameters. Don't use config files via `-c`. (DSL2 only).

But this seemed a bit extreme/negative

bentsherman: @jfy133 I thought it was the other way around -- that the params are applied to the pipeline script but not the base config?

@pditommaso In any case I think the issue linked by James is a good test case that describes the problem. I think we should either add this note to the docs or fix the inconsistency in Nextflow if it isn't too complicated.

jfy133: Hm, ok yes the current description is flipped I think (I'll make a suggestion to clarify that) - sorry.

I don't know exactly what you mean via pipeline script vs. base config.  Basically what I found is with DSL2: 

if I provide a custom config that has the params scope, if I check `tower` (or whatever is displyed in the top of an nf-core run header), the resolved parameters correctly displays the updated parameters derived from the custom config.

But when you actually look in the `.command.sh` the command itself only has the settings defined in e.g. `base.config`. 

But I found writing this concisely a little difficult 💧

jfy133, line 33:
```suggestion
  Pipeline parameters (i.e. the ``params`` scope) should be overwritten through CLI options (``--something value``)
  or a params file (``-params-file params.yml``). It is not recommend to pass parameters via a custom config file (``-c my_config``) as they may not be passed to processes despite being displayed in the resolved configuration.
  See the :ref:`CLI page <cli>`. for more information about how to supply params through CLI options or a params file.
```

bentsherman: I see. I think we need to find an explanation for this phenomenon in the source code before documenting it. It's a rather confusing edge case, and it might be better to fix it than to try to document it.

jfy133: Fair enough.

I could just set it as `Pipeline parameters (i.e. the ``params`` scope) should be overwritten through CLI options (``--something value``)
  or a params file (``-params-file params.yml``).`
  
  Or should I remove the note from this page entirely? The `params-file` descriptions would still be a useful addition regardless I think

pditommaso: Sorry, still finding confusing the "previous configuration" concept. I've remove this paragraph

pditommaso:
Merged 163eba23

PR #3590, opened 2023-01-31
Add readOnlyInputs option for CharliecloudBuilder

l-modolo:
This merge request is to add the `readOnlyInputs` option as discussed in the  **Charliecloud and shared cacheDir** #3367 issue  and following the  **Add readOnlyInputs option for CharliecloudBuilder** #3477 merge request

With this modification:

- the default `readOnlyInputs` value is `false`
- if `readOnlyInputs` is `true`;
- the `-w` option is removed from the command
- the bound volumes are transformed to only mount the first folder of the path

For example, if I try to bind `/home/user/workdir/` to `/home/` the bound path will be `/home` to `/home`. By doing that, charliecloud doesn't try to `mkdir -p /home/user/workdir/` within the read-only container, the external path is bound at the `/home` directory level so all the subfolders are accessible with the user rights. If one tries to bind `/` I added an expection.

Signed-off-by: Laurent Modolo [laurent.modolo@ens-lyon.fr](mailto:laurent.modolo@ens-lyon.fr)

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/container/CharliecloudBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/CharliecloudBuilder.groovy
@@ -46,6 +46,9 @@ class CharliecloudBuilder extends ContainerBuilder<CharliecloudBuilder> {
         if( params.containsKey('runOptions') )
             addRunOptions(params.runOptions.toString())
 
+        if( params.containsKey('readOnlyInputs') )
+            this.readOnlyInputs = params.readOnlyInputs?.toString() == 'true'
+
         return this
     }
 
@@ -58,7 +61,9 @@ class CharliecloudBuilder extends ContainerBuilder<CharliecloudBuilder> {
     CharliecloudBuilder build(StringBuilder result) {
         assert image
 
-        result << 'ch-run --unset-env="*" -c "$PWD" -w --no-home --set-env '
+        result << 'ch-run --unset-env="*" -c "$PWD" --no-home --set-env '
+        if (!readOnlyInputs)
+            result << '-w '
 
         appendEnv(result)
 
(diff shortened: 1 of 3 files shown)


modules/nextflow/src/main/groovy/nextflow/container/CharliecloudBuilder.groovy
```diff
@@ -78,9 +84,23 @@ class CharliecloudBuilder extends ContainerBuilder<CharliecloudBuilder> {
         return this
     }
 
+    protected String getRoot(String path) {
+        def rootPath = path.split("/")
+
+        if (rootPath.size() >= 1)
+            rootPath = "/${rootPath[1]}"
```
pditommaso, line 91: Not sure about this. What is expected to take? if you want the first path component it should be `rootPath[0]`

l-modolo: Yes, I just want the first path component of the path. However, those are absolute path so `rootPath[0]` is empty

```
println "/test/path".split("/")
[, test, path]
```

pditommaso: use `.tokenize('/')` instead, should avoid that problem

l-modolo: I made the modification in the commit  a5df098ff

pditommaso: ok, wonder tho if there's a unit test for it ?

l-modolo: You are right, I added two test for that in the commit 45de7b8c1

pditommaso: thanks a lot

pditommaso:
Included in version 23.02.0-edge

PR #3598, opened 2023-02-02
Fix #3597 (Remove newline char from tagname)

### LINKED ISSUE #3597: Line break in a process tag
## Bug report 

Hello!
I have a composite (comma-separated) parameter, which should be supplied to a function via a single argument (e.g., `--x "a,b,c,d,e"`). In Nextflow, I created a channel containing all the values, with every single entry emitted separately.

Expected behavior: Nextflow processes each channel value and shows a tag in a screen output.

Actual behavior: If tag is specified, Nextflow fails with error. Without a tag, everithing is OK.

Simple example:

``` bash
cat > test.nf <<'EOT'
#!/usr/bin/env nextflow

params.x = "a,b,c,d,e"

process proc {

    tag "$ind"

    input:
      val  ind

    output:
      path "res.txt"

    script:
    """

    echo ${ind} > "res.txt"

    """
}

workflow {

  ind_ch = Channel.value( params.x )
    .splitText( { it.split(',') } )
    .distinct()
    .flatten()

  proc(ind_ch)

}

EOT

nextflow run test.nf
```

### Program output 

When running a pipeline, Nextflow fails with the following error:
```
.command.run: line 3: syntax error near unexpected token `)'
```

It happened because there is a line break after a tag name in `.command.run`, and line 3 starts with `)`:
```
#!/bin/bash
# NEXTFLOW TASK: proc (e
)
set -e
set -u
...
```

If we comment out the tag and run the script, everything works as expected:
``` bash
sed 's/tag/\/\/tag/' test.nf > test_ok.nf
nextflow run test_ok.nf
```

### Environment 

* Nextflow version: 22.10.6 build 5843
* Java version: openjdk 17.0.3-internal 2022-04-19
* Operating system: Linux
* Shell version: zsh 5.8.1 (x86_64-ubuntu-linux-gnu)

mribeirodantas:
When the process tag name ends with a new line char (`\n`), as it happens if it's based on a string obtained with the `splitText` operator, it will add a new line with a `)` in the `.command.run` and `bash` will fail when trying to interpret the file. This PR fixes this issue by trimming the tag name when writing the `.command.run`.

At first I thought of fixing this at the `processConfig` level, but if the tag name is a variable, at the processConfig file level/time, it hasn't been expanded yet so `this.tag` returns null.

Should I keep working on this, @pditommaso ? I'm wondering if you want to suggest a different approach, for example. Otherwise I will write the unit test based on this.


modules/nextflow/src/main/groovy/nextflow/processor/TaskRun.groovy
```diff
@@ -362,7 +362,7 @@ class TaskRun implements Cloneable {
             try {
                 // -- look-up the 'sampleId' property, and if everything is fine
                 //    cache this value in the 'name' attribute
-                return name = "$baseName (${config.tag})"
+                return name = "$baseName (${config.tag.trim()})"
```
bentsherman, line 365: Might as well remove all newlines:
```suggestion
                return name = "$baseName (${config.tag.replace('\n','')})"
```

PR #3599, opened 2023-02-02
Remove container tag from AWS Batch job definition

pditommaso:
This commit strips the container tag segment from the name associated with the AWS Batch job definition created by nextflow.

The rationale of this change is reduce the proliferation of AWS batch jobs created by nextflow when using containers with the same name and different tags


jordeu (commented):
Given two images with different tags I understand that even with this change you will end with two job definitions with same name but different `nf-token` parameter. What is the benefit of this change?

pditommaso:
It turns out that's quite useless. closing

PR #3615, opened 2023-02-07
wave conda: always installing procps in containers

marcodelapierre:


marcodelapierre:
Would close https://github.com/nextflow-io/nf-wave/issues/6

marcodelapierre:
@pditommaso  the test fail is a rate limit being hit.
if you think this is worthwile we can rerun

pditommaso:
It would be better to rely on the Conda recipe to install this to avoid depending on distro specific command 

https://anaconda.org/conda-forge/procps-ng

marcodelapierre:
Do you know whether `procps-ng` has the same API/CLI than standard `ps`?

marcodelapierre:
Ok, just did some checks on base images for ubuntu, debian, opensuse, alpine, amazonlinux.
I am a bit ignorant on official naming for ubuntu tools, but turns out these folks actually use `procps-ng` as the provider for `ps`. So the `conda` installation should be fine. 
Soon I am going to update the PR accordingly.

marcodelapierre:
Done, please have a look when you get the chance.

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
```diff
@@ -397,6 +397,7 @@ class WaveClient {
         FROM ${config.condaOpts().mambaImage}
         COPY --chown=\$MAMBA_USER:\$MAMBA_USER conda.yml /tmp/conda.yml
         RUN micromamba install -y -n base -f /tmp/conda.yml && \\
+            micromamba install -y -n base conda-forge::procps-ng && \\
```
pditommaso, line 400: I was thinking that instead of automagically adding dependencies in the container build it could be added an option to provide one or packages that should be a added in the resulting container e.g. 

```
wave.conda.basePackages = 'conda-forge::procps-ng'
```

or 

```
wave.conda.addPackages = 'conda-forge::procps-ng'
```

WDT?

marcodelapierre: I like `wave.conda.basePackages`, I think it conveys the message.
Will implement as soon as I am finished with this week's duties.

pditommaso:
Ok, I went head with this. Have a look at 7e827810. Thanks Marco!

marcodelapierre:
cool!

Issue #3110, opened 2022-08-10
Error staging directory from S3 as input


jeremykap:
### Expected behavior and actual behavior

I'd like to be able to stage a directory from S3 as input to task, but am getting a  java.lang.UnsupportedOperationException error when doing so. I'd expect this to work as it does locally, where the local path is symlinked into the working directory, and works like any other file.

### Steps to reproduce the problem

This workflow takes  a directory path as param `input_dir`, and lists it's contents in a file:

```
nextflow.enable.dsl=2


process ListFolder {
    input: 
        path(directory)
    output:
        path("dir_contents.txt")
    """
    ls ${directory}/ > dir_contents.txt
    """
}

workflow {
    input_ch = Channel.fromPath(params.input_dir)
    ListFolder(input_ch)
}
```

### Program output 

When running with local path it succeeds, when running with s3 path, it fails

```
nextflow run test.nf --input_dir {S3_PATH}
N E X T F L O W  ~  version 22.04.5
Launching `test.nf` [jovial_hugle] DSL2 - revision: 5877c29902
[-        ] process > ListFolder -
Error executing process > 'ListFolder'

Caused by:
  java.lang.UnsupportedOperationException
```
from .nextflow.log:
```
Aug-10 21:15:39.290 [Actor Thread 3] DEBUG nextflow.plugin.PluginUpdater - Installing plugin nf-amazon version: 1.7.2
Aug-10 21:15:39.298 [Actor Thread 3] INFO  org.pf4j.AbstractPluginManager - Plugin 'nf-amazon@1.7.2' resolved
Aug-10 21:15:39.298 [Actor Thread 3] INFO  org.pf4j.AbstractPluginManager - Start plugin 'nf-amazon@1.7.2'
Aug-10 21:15:39.316 [Actor Thread 3] DEBUG nextflow.plugin.BasePlugin - Plugin started nf-amazon@1.7.2
Aug-10 21:15:39.332 [Actor Thread 3] DEBUG nextflow.file.FileHelper - > Added 'S3FileSystemProvider' to list of installed providers [s3]
Aug-10 21:15:39.333 [Actor Thread 3] DEBUG nextflow.file.FileHelper - Started plugin 'nf-amazon' required to handle file: {S3_PATH}
Aug-10 21:15:39.343 [Actor Thread 3] DEBUG nextflow.file.FileHelper - Creating a file system instance for provider: S3FileSystemProvider
[...]

PR #3617, opened 2023-02-07
Add java install recommendations to docs

bentsherman:
This PR documents the recommendations provided here: https://github.com/nextflow-io/nextflow/issues/3110#issuecomment-1212014907

I think we should explicitly point users to SDKMAN and at least one recommended distro, since we keep doing it in support tickets.

### DIFF
--- a/docs/getstarted.rst
+++ b/docs/getstarted.rst
@@ -17,6 +17,20 @@ the sharing of tasks input/output files.
 
 Nextflow can also be run on Windows through `WSL <https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux>`_.
 
+.. tip::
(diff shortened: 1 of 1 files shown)


docs/getstarted.rst
```diff
@@ -17,6 +17,20 @@ the sharing of tasks input/output files.
...
+    difficult to debug and are usually resolved by using one of the recommended distros.
+
+    To install Corretto 17::
+
+        sdk install java 17.0.6-amzn
+
+    To install Temurin 17::
+
+        sdk install java 17.0.6-tem
+
```
bentsherman, line 33: This part could be removed, or simplified to not mention a specific version.

pditommaso: I like it!

PR #3630, opened 2023-02-09
Support instance template in google batch

### LINKED ISSUE #3629: Add instance template option to google batch
Google Cloud Batch supports an instanceTemplate option. With this option, batch job will use a specified instance template to create a VM instance. However, current Google Cloud Batch plugin doesn't support this option.

See the details below :
https://cloud.google.com/batch/docs/reference/rest/v1/projects.locations.jobs#instancepolicyortemplate
https://cloud.google.com/batch/docs/create-run-job#create-job-instance-template

daigok:
Closes #3629

Add an instanceTemplate option to create and submit a job using a VM instance template on Google Cloud Batch.

Singed-off-by: Daigo Kobayashi [kobayashidaigo@google.com](mailto:kobayashidaigo@google.com)


docs/google.rst
```diff
@@ -126,6 +126,8 @@ google.enableRequesterPaysBuckets              When ``true`` uses the configured
 google.batch.allowedLocations                  Define the set of allowed locations for VMs to be provisioned. See `Google documentation <https://cloud.google.com/batch/docs/reference/rest/v1/projects.locations.jobs#locationpolicy>`_ for details (default: no restriction. Requires version ``22.12.0-edge`` or later).
 google.batch.bootDiskSize                      Set the size of the virtual machine boot disk, e.g ``50.GB`` (default: none).
 google.batch.cpuPlatform                       Set the minimum CPU Platform, e.g. ``'Intel Skylake'``. See `Specifying a minimum CPU Platform for VM instances <https://cloud.google.com/compute/docs/instances/specify-min-cpu-platform#specifications>`_ (default: none).
+google.batch.instanceTemplate                  Set the Compute Engine instance template to create a job (default: none).
+google.batch.installGpuDrivers                  When ``true`` install the GPU drivers to the virtual machine (default: ``false``).
```
bentsherman, line 130: This option shouldn't be exposed to users, it should be set automatically by Nextflow if the `accelerator` directive was used

bentsherman:
I am hesitant to expose this option through Nextflow, because it would override so many Nextflow directives which are the intended way to configure tasks. What is the benefit of using an instance template that you couldn't do through other the instance policy?

docs/google.rst
daigok: If we only use an `accelerator ` option to add GPUs to an instance, we don't need an `installGpuDrivers` option. However, if we support `instanceTemplate` option, we need the one. Because there is no way to know a created instance template has GPUs or not. It totally depends how a user creates an instance template and we only know the name of the instance template. Hence, if we support an instance template option, an `installGpuDrivers` option is inevitable.

daigok:
Currently, there is no way to specify a custom image to the instance without using an instance template option. 

For example, MPI tasks uses a [HPC-ready instance template](https://cloud.google.com/compute/docs/instances/create-hpc-vm). Please see the [Create a job that uses MPI for tightly coupled tasks](https://cloud.google.com/batch/docs/create-run-job?hl=en#create-job-uses-mpi) for details.

docs/google.rst
bentsherman: Okay, I will just add a note that this option is only needed when using an instance template.

bentsherman, line 130:
```suggestion
google.batch.installGpuDrivers                 When ``true`` install the GPU drivers to the VM (default: ``false``, only needed when specifying an instance template).
google.batch.instanceTemplate                  Specify an instance template with which to create VMs (default: none).
```

docs/google.rst
```diff
@@ -126,13 +126,18 @@ google.enableRequesterPaysBuckets              When ``true`` uses the configured
...
 google.batch.network                           Set network name to attach the VM's network interface to. The value will be prefixed with global/networks/ unless it contains a /, in which case it is assumed to be a fully specified network resource URL. If unspecified, the global default network is used.
 google.batch.serviceAccountEmail               Define the Google service account email to use for the pipeline execution. If not specified, the default Compute Engine service account for the project will be used.
 google.batch.subnetwork                        Define the name of the subnetwork to attach the instance to must be specified here, when the specified network is configured for custom subnet creation. The value is prefixed with `regions/subnetworks/` unless it contains a `/`, in which case it is assumed to be a fully specified subnetwork resource URL.
 ============================================== =================
 
+.. note::
+  Specifying an instance policy will overwrite the following options: ``cpuPlatform``, ``machineType``, `preemptible``, and ``spot``.
```
bentsherman, line 139:
```suggestion
  Specifying an instance policy will overwrite the following options: ``cpuPlatform``, ``machineType``, ``preemptible``, and ``spot``.
```

bentsherman:
@daigok What happens if I specify an instance template but also request task resources like CPU, memory, disk, etc? These resources are added to the task spec, so they won't be overwritten by the instance template. But the instance template defines these resources for the VM. I assume that Google Batch will simply fail with an error if the resource requests are larger than what the instance template provides.

Unlike AWS Batch, which can map multiple tasks to the same VM, Google Batch maps each task to its own VM. So I'm wondering how to best advise the Nextflow user here. If I specify an instance template, should I not specify other directives like `cpus`, `memory`, `disk`, `accelerator` ? Can I assume that the task will just have access to the entire VM?

Also a minor point, from merging the latest upstream changes it looks like the instance template is incompatible with Fusion, because Fusion relies on the instance policy to add the local disk.

daigok:
@bentsherman Google Cloud Batch task spec and machine type are a little bit tricky. 

Google Cloud Batch can also map multiple tasks to the same VM. For example, if you specify an e2-standard-2 (which has 2 vCPUs) machine type and also set cpuMilis: 1000 (1 vCPU)  and parallelism: **2**, Cloud Batch doesn't spin up a new VM (a VM processes two tasks). 
If you specify an e2-standard-2 machine type and set cpuMilis: 1000 (1 vCPU) and parallelism: **4**, then Cloud Batch spins up a new VM (two VMs process tasks and each VM handles 2 tasks).

`cpus`, `memory` and `disk` options are set to the `task`, but if you don't specify a machine type, Cloud Batch decides the machine type based on the specified `cpus` and `memory` (normally it chooses e2-highcpu-x). `accelerator` is set to the instance level, so if you specify an instance template, Cloud Batch ignores the `accelerator` (actually you can only specify either `instance template` or `instance policy`).  

To sum up, `cpus`, `memory` and `disk` are set to a task NOT a VM. Please also see the [task count per node option](https://cloud.google.com/java/docs/reference/google-cloud-batch/latest/com.google.cloud.batch.v1.TaskGroup.Builder#com_google_cloud_batch_v1_TaskGroup_Builder_setTaskCountPerNode_long_).


About Fusion incompatibility, I also think it doesn't work with the instance template unfortunately. I think it is better to mention this limitation in the document.

bentsherman:
Okay, then I think everything should work quite nicely. So Google Batch might actually pack multiple tasks to a VM if the instance template uses a larger machine type.

To use GPUs, the user will have to enable `installGpuDrivers` and specify a GPU-enabled machine type in the instance template. The task container should be able to see the GPUs as with an instance policy.

I added some warnings about Fusion and using GPUs. I think everything is ready to merge, but it would be nice if you could test it with one of your use cases.

bentsherman (approved):

pditommaso:
Didn't we say to specify the template via the machineType? https://github.com/nextflow-io/nextflow/issues/3629#issuecomment-1429281504

bentsherman:
Whoops, forgot about that part. I will add the machineType support using the `template://` prefix.

bentsherman:
Okay, now the instance template is specified via the `machineType` directive, so it can be process-specific. An unfortunate annoyance now is that the `installGpuDrivers` option is global, so if only some of your processes need GPUs, you will still have to install GPU drivers on _all_ of them.

@daigok For this reason, I would like to determine whether the `installGpuDrivers` option is even needed. If the instance template includes a GPU-enabled machine type, then Google Batch might set that option automatically. This is based on past conversations with Google Batch devs, but my memory is fuzzy. Can you perform a test with a GPU-enabled instance template?

@hnawar can you tell us if the `installGpuDrivers` option is required, if the instance template already specified a GPU-based machine type? I remember you saying at some point that it should be set automatically by Google Batch.

docs/google.rst
```diff
@@ -136,17 +137,25 @@ google.batch.subnetwork                        Define the name of the subnetwork
...
-machine type or machine series they should allow to attach local SSD disks, otherwise the job scheduling will fail.
+* A comma-separated list of patterns. The pattern can contain a ``*`` to match any number of characters
+  and ``?`` to match any single character. Examples of valid patterns: ``c2-*``, ``m?-standard*``, ``n*``.
+
+* A predefined Google Compute Platform `machine type <https://cloud.google.com/compute/docs/machine-types>`_ or `custom machine type <https://cloud.google.com/compute/docs/instances/creating-instance-with-custom-machine-type>`_.
+
+* An `instance template <https://cloud.google.com/compute/docs/instance-templates>`_, specified as ``template://<instance-template>``. Using an instance template
+  will overwrite the ``accelerator`` directive, as well as the following Google Batch options: ``cpuPlatform``, ``preemptible``, and ``spot``. To use an instance template
+  with GPUs, you must also set the ``google.batch.installGpuDrivers`` config option to ``true``. To use an instance template with Fusion, the instance template must include
+  a ``local-ssd`` disk with at least 375 GB. Requires version ``23.03.0.edge`` or later.
```
bentsherman, line 154: @jordeu Do you think this will work? Fusion adds a local SSD through the instance policy, which is ignored if an instance template is specified. So I figured, if the instance template includes an SSD disk, then Fusion should still work.

jordeu: And the device name should be "fusion".

See here:
http://github.com/nextflow-io/nextflow/blob/a33a6f47e2160f476de9024b9415d63a57647479/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy#L253-L253

And how then is used to make Google Batch automatically mount it inside the container:
http://github.com/nextflow-io/nextflow/blob/a33a6f47e2160f476de9024b9415d63a57647479/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchFusionAdapter.groovy#L52-L52

bentsherman: Okay, then it will work as long as the device name is `fusion`. The volume mounts are not in the instance policy so they will work the same.

bentsherman:
```suggestion
  with GPUs, you must also set the ``google.batch.installGpuDrivers`` config option to ``true``. To use an instance template with Fusion, the instance template must include
  a ``local-ssd`` disk named ``fusion`` with at least 375 GB. Requires version ``23.03.0.edge`` or later.
```

docs/google.rst
```diff
@@ -136,14 +137,25 @@ google.batch.subnetwork                        Define the name of the subnetwork
...
-any number of characters and ``?`` to match any single character. Examples of valid patterns: ``c2-*``, ```m?-standard*``, ```n*``.
+Processes can be defined as usual. By default, the ``cpus`` and ``memory`` directives are used to find the cheapest machine
+type that is available at the current location and that fits the requested resources. If ``memory`` is not specified, 1GB of memory is allocated per CPU.
 
-Alternatively it can also be used to define a specific predefined Google Compute Platform `machine type <https://cloud.google.com/compute/docs/machine-types>`_
-or a custom machine type.
+The ``machineType`` process directive can be specified in any of the following ways:
+
+* A comma-separated list of patterns. The pattern can contain a ``*`` to match any number of characters
+  and ``?`` to match any single character. Examples of valid patterns: ``c2-*``, ``m?-standard*``, ``n*``.
```
bentsherman, line 147:
```suggestion
* A comma-separated list of patterns. The pattern can contain a ``*`` to match any number of characters
  and ``?`` to match any single character. Examples of valid patterns: ``c2-*``, ``m?-standard*``, ``n*``. Requires  version ``23.02.0-edge`` or later.
```

docs/google.rst
```diff
@@ -144,7 +144,7 @@ type that is available at the current location and that fits the requested resou
 The ``machineType`` process directive can be specified in any of the following ways:
 
 * A comma-separated list of patterns. The pattern can contain a ``*`` to match any number of characters
-  and ``?`` to match any single character. Examples of valid patterns: ``c2-*``, ``m?-standard*``, ``n*``.
+  and ``?`` to match any single character. Examples of valid patterns: ``c2-*``, ``m?-standard*``, ``n*``. Requires  version ``23.02.0-edge`` or later.
```
pditommaso, line 147:
```suggestion
  and ``?`` to match any single character. Examples of valid patterns: ``c2-*``, ``m?-standard*``, ``n*`` (requires  version ``23.02.0-edge`` or later).
```

plugins/nf-google/src/test/nextflow/cloud/google/batch/GoogleBatchTaskHandlerTest.groovy
```diff
@@ -89,6 +90,8 @@ class GoogleBatchTaskHandlerTest extends Specification {
         runnable.getContainer().getOptions() == ''
         runnable.getContainer().getVolumesList() == ['/mnt/disks/foo/scratch:/mnt/disks/foo/scratch:rw']
         and:
+        instancePolicyOrTemplate.getInstanceTemplate() == ''
```
pditommaso, line 93:
```suggestion
        !instancePolicyOrTemplate.getInstanceTemplate()
```

let's make more groovy!

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -224,8 +224,14 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
...
-        if( task.config.getMachineType() )
-            instancePolicy.setMachineType( task.config.getMachineType() )
+        def machineType = task.config.getMachineType()
+        def instanceTemplate = null
+        if( machineType ) {
+            if( machineType.startsWith('template://') )
+                instanceTemplate = machineType.minus('template://')
+            else
+                instancePolicy.setMachineType( machineType )
+        }
```
pditommaso, line 234: This code it's becoming a bit convoluted. Let's add some descriptive comment to help to follow all logic here

plugins/nf-google/src/test/nextflow/cloud/google/batch/GoogleBatchTaskHandlerTest.groovy
bentsherman: Sure, I will do it in another commit

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
bentsherman: I moved the instance policy / instance template pieces into separate branches so that it's clearer

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -199,65 +199,78 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
...
-        if( fusionEnabled() ) {
-            instancePolicy.addDisks(AllocationPolicy.AttachedDisk.newBuilder()
-                    .setNewDisk(AllocationPolicy.Disk.newBuilder()
-                            .setType("local-ssd")
-                            .setSizeGb(375)
-                    )
-                    .setDeviceName("fusion")
-            )
+            machineInfo = findBestMachineType(task.config)
+            if( machineInfo )
```
bentsherman, line 250: @pditommaso @jordeu Should these two conditions be switched? Otherwise the custom machineType will be overwritten by the "best" machine type.

bentsherman: Okay but this comment is about the `findBestMachineType()` method overwriting a custom machine type setting, I think the custom machine type should apply as long as it doesn't contain wildcards.

docs/google.rst
```diff
@@ -136,14 +137,25 @@ google.batch.subnetwork                        Define the name of the subnetwork
...
+
+* A comma-separated list of patterns. The pattern can contain a ``*`` to match any number of characters
+  and ``?`` to match any single character. Examples of valid patterns: ``c2-*``, ``m?-standard*``, ``n*`` (requires  version ``23.02.0-edge`` or later).
+
+* A predefined Google Compute Platform `machine type <https://cloud.google.com/compute/docs/machine-types>`_ or `custom machine type <https://cloud.google.com/compute/docs/instances/creating-instance-with-custom-machine-type>`_.
+
+* An `instance template <https://cloud.google.com/compute/docs/instance-templates>`_, specified as ``template://<instance-template>``. Using an instance template
+  will overwrite the ``accelerator`` directive, as well as the following Google Batch options: ``cpuPlatform``, ``preemptible``, and ``spot``. To use an instance template
+  with GPUs, you must also set the ``google.batch.installGpuDrivers`` config option to ``true``. To use an instance template with Fusion, the instance template must include
+  a ``local-ssd`` disk named ``fusion`` with at least 375 GB. Requires version ``23.03.0.edge`` or later.
```
bentsherman, line 154: @pditommaso I put a warning here in the docs about which options are overwritten by the instance template. I could add warnings for these settings in the code.

Directives like `cpus` and `memory` can be used with the instance template. In this case, if the machine type is bigger than the requested resources then Google Batch will try to pack multiple tasks on the same VM.

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
jordeu: If they only set one machine type `findBestMachineType` will always return that machine type. See here:
https://github.com/nextflow-io/nextflow/blob/1e38274ac573d8016c9704250b701b84e4ffcbbc/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy#L90-L97

I'm not checking for wildcards earlier in the code, because this logic also allows the user to define a list of machine types separated by comma, without any wildcards. On that case the `findBestMachineType` will select the best from the user provided list.

bentsherman: Thanks Jordi, in that case we don't need to set `machineType` directly because `findBestMachineType()` will do that anyway.

daigok:
> @daigok For this reason, I would like to determine whether the `installGpuDrivers` option is even needed. If the instance template includes a GPU-enabled machine type, then Google Batch might set that option automatically. This is based on past conversations with Google Batch devs, but my memory is fuzzy. Can you perform a test with a GPU-enabled instance template?

Unfortunately, we still need the `installGpuDrivers` option. Even if the instance template includes a GPU-enabled machine type, Google Batch does nothing about the gpu drivers. Hence, we explicitly need to specify the option and install the drivers to the VM.

daigok:
@bentsherman 
I've tested the instance template support and it works fine. 

I've accidentally synced and merged 'nextflow-io:master' latest changes to this PR. If it should be reverted, let me know.

docs/google.rst
bentsherman: I ran some tests and wasn't able to get Google Batch to pack tasks onto the same VM. Still, you can use `cpus` and `memory` with `machineType`. I might have been missing something else to make it work.

saberiato:
Any update on finalizing this feature?
I used `machineType = 'template://<instance-name>'` on Nextflow version `23.05.0-edge` and got `null` error.

bentsherman:
Hi @saberiato , it hasn't been merged yet. We need to merge #3861 first and then this one should be ready to go.

saberiato:
Hi @bentsherman. Any update on merging these commits?

bentsherman:
We merged the disk type PR a while ago and @daigok performed a successful test with the latest iteration.

@pditommaso ready for review

saberiato:
I updated to the latest edge version (`23.08.0-edge build 5872`), and tried to use an instance template in the Google batch executor in my pipeline:
```
google {

    project = 'my-project'
    location = 'us-central1'
    batch.serviceAccountEmail = 'my-service-account@my-project.iam.gserviceaccount.com'
    batch.installGpuDrivers = true

}

process {

    executor = 'google-batch'

    withName: MyProcess {
        machineType = 'template://general-gpu'
        container = 'us-central1-docker.pkg.dev/my-project/my-containers/my-image:1.0'
        containerOptions = '--privileged --gpus all'
    }

}
```
And I got this error:
```
ERROR ~ Error executing process > 'MyProcess'

Caused by:
  Process `MyProcess` terminated for an unknown reason -- Likely it has been terminated by the external system
  
  [...]
  
 Command exit status:
  -

 Command output:
  (empty)
```
Is this feature ready to use?

bentsherman:
Hi Ali, this PR hasn't been merged yet. You're welcome to build a custom version of Nextflow from this PR if you'd like to try it out:
```bash
# build nextflow
git clone https://github.com/daigok/nextflow -b google_cloud_batch_instance_template
cd nextflow
make compile

# run nextflow with launch.sh
./launch.sh run ...
```

saberiato:
Hi @bentsherman, thanks!
I'll try the custom version... is there any estimated date to merge this PR?

daigok:
@bentsherman @pditommaso 
I've forgot to set `--signoff` option to my commit [a574385](https://github.com/nextflow-io/nextflow/pull/3630/commits/a574385986c7ce6432f12d0f2fcda37e10cfbbbd). Do I need to amend the commit? 

I guess the PR is squashed and merged into the master branch. So the commit log in this PR is not very important. If it convenient for you, you can delete/amend the commit too.

Also, do we still need to do something before we merge the PR into the master? If something I can help, please let me know.

bentsherman:
Hi @daigok I would just click on the DCO details and follow the instructions to rebase your commit. It will modify the commit history but that should be fine as I think the PR is done. Then I will update the version notes and we should be able to merge

stephenstevethomas:
Hey! Any speculation on when this PR may land?

pditommaso:
@bentsherman a recent conversation on slack, made me think if this should not be managed instance via the queue directive instead of special `machineType`

bentsherman:
I would be interested to see that slack thread. Google Batch doesn't really have queues like AWS Batch or HPC schedulers.

pditommaso:
I know there's no queue concept, but the instance template could be thought as such 

https://nextflow.slack.com/archives/C02TLU7U57F/p1694825242523869

bentsherman:
Not really, it's more like a launch template or job definition. It's a template for a single VM which can be reused. There is no concept of "Max CPUs" or "Allowed instance types", instead you define one machine type for the template.

Also it wouldn't make sense to specify an instance template alongside `machineType`, since the instance template already specifies the machine type.

I spoke to Nathan more in depth directly, turns out he was investigating Google and Azure as alternatives to AWS Batch. He wasn't very familiar with google cloud, so I wouldn't read too much into what he said.

pditommaso:
Does this allows specifying `cpus` and `memory` when using an instance template?

bentsherman:
@pditommaso yes

pditommaso:
Nice. But then, if a machine type is specified in the template, what's the logic taking place? cpus and mem will override the machine type? Tagging @swampie for visibility

bentsherman:
They are independent concepts. The instance template specifies the cpus and memory of the machine, but the `cpus` and `memory` directives specify the task resources. My understanding is that if you use a machine type with more cpus/memory than the task, Google Batch can pack multiple tasks onto the same VM.

pditommaso:
Fixed conflicts and update the docs. Thanks for the contribution and the patience. Merged via df7ed294

PR #3636, opened 2023-02-10
[Beta Test] Wave Containers: add support for the Spack package manager

marcodelapierre:
This PR adds support for Spack as a software provider for Wave containers.
It comes out of some very fruitful conversations myself and Paolo have had recently.

It has the potential to pave the way for on-demand, architecture optimised containers with Nextflow and Wave! (some work yet to be completed). The key missing piece of work in this regard is the adoption of binary caches, to speed up build times (coming later).

Some points:

* Includes support for generalised architectures via `process.arch`
* Tested with a simple pipeline:
    * https://github.com/marcodelapierre/toy-cowsay-nf
* Unit tests written and successful
* Documentation updated
* @pditommaso has added back-end support for Spack YAML files

It is still a DRAFT for now, as 1 last key bit of work is missing:
- [ ] Use of templates for Spack Dockerfiles

@pditommaso @evanfloden have even more fun! :-)

@tgamblin @alalazo I hope this can be another useful contribution from the perspective of the Spack community, too!
@vsochat hopefully this will be merged soon enough .. watch this space ;)


marcodelapierre:
@pditommaso again, not sure why I have some spurious commits from you.

I have just recloned the main nextflow repo and the merge says it is all good ...

pditommaso:
What's wrong in your Git workflow? 😆 

screencapture-github-nextflow-io-nextflow-pull-3636-2023-02-10-16_22_51

marcodelapierre:
> What's wrong in your Git workflow? 😆

Paolo I find it embarrassing 🤣🤣
I will dig further!

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
```diff
@@ -403,11 +434,82 @@ class WaveClient {
...
+         /root/.spack/linux/compilers.yaml) > /opt/spack-env/compilers.yaml \\
+&&  (sed '/^spack:/a  \\ \\ include: [/opt/spack-env/compilers.yaml]\\
+  concretizer:\\
+    unify: true' /tmp/spack.yaml) > /opt/spack-env/spack.yaml \\
+&&  (echo "  packages:" \\
+&&   echo "    all:" \\
+&&   echo "      target: [${config.spackOpts().target}]" \\
+&&   echo "  config:" \\
+&&   echo "    install_tree: /opt/software" \\
+&&   echo "  view: /opt/view") >> /opt/spack-env/spack.yaml
```
pditommaso, line 459: What's the need for the `echo` ?

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
```diff
@@ -425,12 +527,90 @@ class WaveClient {
...
+    ( apk update && apk add --no-cache ${config.spackOpts().osPackages} && rm -rf /var/cache/apk )
+""" //.stripIndent()
+
+        result = addCommands(result)
+
+        result += """\
+
+ENTRYPOINT ["/bin/bash", "--rcfile", "/etc/profile", "-l", "-c", "\$*", "--" ]
+CMD [ "/bin/bash" ]
+"""//.stripIndent()
```
pditommaso, line 597: Would not be better to use a template file instead of a big inline string?

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
marcodelapierre: struggling to reproduce newlines appropriately in the Dockerfile, by using a single `echo` or `cat` ..

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
marcodelapierre: good point, will be back on this one too

pditommaso: Have a look a this class 

https://github.com/nextflow-io/nextflow/blob/762ba19c70006c84070bf1c736825e483c7fdde7/modules/nextflow/src/main/groovy/nextflow/executor/BashTemplateEngine.groovy#L31-L45

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
marcodelapierre: ok, addressed by using a single `echo` . templating (below) later on

pditommaso:
This seems in very good shape. Roadmap to merge: 
1. move the Spack dockerfile to a template file 
2. avoid the use of exotic Bash option e.g. `-l` if possible, for consistency with other containers 
3. identify core `target` platforms to be supported 
4. add the support for Spack mirror on AWS for caching purpose 
5. implemented the corresponding support for Wave backend

marcodelapierre:
> This seems in very good shape. Roadmap to merge:
> 
> 1. move the Spack dockerfile to a template file
> 2. avoid the use of exotic Bash option e.g. `-l` if possible, for consistency with other containers
> 3. identify core `target` platforms to be supported
> 4. add the support for Spack mirror on AWS for caching purpose
> 5. implemented the corresponding support for Wave backend

A memo note here, this is the github issue on arch compliance specifications by @pditommaso : https://github.com/nextflow-io/nextflow/issues/3701

marcodelapierre:
OK, first successful run of a Wave Spack container using a `spack.yaml` file instead of a one-line package specification.

One issue to note. Every time I used Wave to build these Spack containers, the first time I build a new container the pull fails:
```
Command error:
  Unable to find image 'wave.seqera.io/wt/455e45c14efa/wave/build:4a71f62a456a4231cb0f0640b406628e' locally
  docker: Error response from daemon: received unexpected HTTP status: 502 Bad Gateway.
  See 'docker run --help'.
```
Then I just re-run the pipeline and everything goes OK.

Here I have some gaps in expertise. Not sure whether it is an issue arising somehow from Docker locally, or instead from your Wave backend. E.g. I am thinking at some latencies in making newly built layers available for pull....?

@pditommaso what do you think?

pditommaso:
Have you got the build notification email? can't find anything wrong

marcodelapierre:
The build was indeed successful (I have emailed you the notification), as I was suspecting.
So it is something that happens afterwards.

pditommaso:
Umm, `502 Bad Gateway` sounds more like a load balancer glitch. I can pull the container. Have you tried to build it again?

marcodelapierre:
So, Paolo, I was able to run some tests with both `spack` and `conda` Wave builds.

I confirm that once the image is built by the backend, any pull after the first (failing, if applicable) one will work.

Is there a way to force the backend to rebuild the image.
What I have done in my tests is to change a package version to force a new build.

So far: 
* all builds succeeded
* all Spack Wave builds failed on *first pull*
* I have identified at least one large Conda Wave build that fails on *first pull*, too: `process.conda = 'conda-forge::cowpy https://conda.deepmodeling.com::deepmd-kit=*=*cpu https://conda.deepmodeling.com::libdeepmd=*=*cpu'` (3 GB image, DeepMD is big)

Notably, the error I got with Conda was slightly different:
```
Command error:
  Unable to find image 'wave.seqera.io/wt/c7e9f7bb170d/wave/build:593503d65400503e25f885163ecdc304' locally
  docker: Error response from daemon: missing or empty Content-Length header.
  See 'docker run --help'.
```

So it really seems to me about something going badly on the backend side between the build and the client request to pull.

Thoughts?

pditommaso:
Can I use this fork to replicate the error?

marcodelapierre:
Yes, but I imagine (not tested) that you can replicate the `conda` case with the develop branch as well.

pditommaso:
I'd like to focus on the spack problem because it looks more consistent. What is the config/script you are using ?

marcodelapierre:
Use this Paolo:
```
git clone git@github.com:marcodelapierre/toy-cowsay-nf 
cd toy-cowsay-nf 
nextflow run main.nf -profile spackwaveyaml
```
All of the required files are in that github repo, in particular `nextflow.config` and `spack.yaml`.

Right now, you can only do multi-packages (e.g. `samtools blast`) with the YAML file, due to a silly missing bit of implementation for the other case.

The corresponding container image should already be in the Wave cache as I built it yesterday, so I suggest you open the `spack.yaml` and change `nano ^ncurses abi=5` to for instance `nano@6.0 ^ncurses abi=5` (to get always new images you can use a variety of Nano versions down to `4.0`).

(Note nano is not needed by the pipeline, it is really just a way to add some extra minimal complexity to the image build, for testing purposes)

marcodelapierre:
Ah! the moment I read the error  message, I realised what is the issue. Yesterday I had changed the codebase to use `spack.yml` instead of `spack.yaml` .. well Spack does not like it!

I have fixed the code, you can go ahead now with the testing, apologies for the hiccup!

To help us remembering about this, I recommend to also tweak the backend to use `spack.yaml`; let me know when you do it and I will change a couple of corresponding lines in the client code.

NOTICE
```diff
@@ -141,5 +141,10 @@ This software includes source code and libraries developed by:
     Copyright 2008, http://www.snakeyaml.org
     Licensed under Apache License, Version 2.0
 
+    Spack
+    https://github.com/spack/spack
+    Copyright 2013-2023 Lawrence Livermore National Security, LLC and other Spack Project Developers. See the top-level COPYRIGHT file for details.
+    Licensed under Apache License, Version 2.0
```
pditommaso, line 147: Not sure this should be added. What code is taken from Spack?

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
```diff
@@ -411,11 +442,83 @@ class WaveClient {
...
+COPY --from=builder /opt/._view /opt/._view
+COPY --from=builder /etc/profile.d/z10_spack_environment.sh /etc/profile.d/z10_spack_environment.sh
+
+# Near OS-agnostic package addition
+RUN ( apt update -y && apt install -y procps libgomp1 ${config.spackOpts().osPackages} && rm -rf /var/lib/apt/lists/* ) || \\
+    ( yum install -y procps libgomp ${config.spackOpts().osPackages} && yum clean all && rm -rf /var/cache/yum ) || \\
+    ( zypper ref && zypper install -y procps libgomp1 ${config.spackOpts().osPackages} && zypper clean -a ) || \\
+    ( apk update && apk add --no-cache procps libgomp bash ${config.spackOpts().osPackages} && rm -rf /var/cache/apk )
+""" //.stripIndent()
+
```
pditommaso, line 502: Worth to move this into a template file

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
```diff
@@ -441,6 +544,81 @@ class WaveClient {
...
+        result = addCommands(result)
+
+        result += """\
+
+ENTRYPOINT ["/bin/bash", "--rcfile", "/etc/profile", "-l", "-c", "\$*", "--" ]
+CMD [ "/bin/bash" ]
+"""//.stripIndent()
```
pditommaso, line 617: also this it's better to keep into a template

pditommaso:
> To help us remembering about this, I recommend to also tweak the backend to use spack.yaml; let me know when you do it and I will change a couple of corresponding lines in the client code.

Ok, changed this

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
marcodelapierre: agree, it's in my TODO

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
marcodelapierre: agree, it's in my TODO

NOTICE
marcodelapierre: So, I added this because I used their Dockerfile template as a starting point for ours:
https://github.com/spack/spack/blob/develop/share/spack/templates/container/Dockerfile

Not sure whether this is the best way to acknowledge it within Nextflow, what do you reckon?

marcodelapierre:
> > To help us remembering about this, I recommend to also tweak the backend to use spack.yaml; let me know when you do it and I will change a couple of corresponding lines in the client code.
> 
> Ok, I've changed this

Thanks, I have just updated the client.

marcodelapierre:
Hi @pditommaso, plan update on this one:
- will restart on this early next week;
- will merge master with #3786 (generalised `arch` implementation) as soon as approved;
- currently targeting Tue 2 May as worst-case date to finalise this PR.

pditommaso:
I had missed this, it looks a great plan 👍

marcodelapierre:
So @pditommaso , on this one:
> avoid the use of exotic Bash option e.g. `-l` if possible, for consistency with other containers

Right now I have this in the Dockerfile:
```
ENTRYPOINT ["/bin/bash", "--rcfile", "/etc/profile", "-l", "-c", "\$*", "--" ]
CMD [ "/bin/bash" ]
```

The reason for this  is:
```
spack env activate --sh -d . >> /etc/profile.d/z10_spack_environment.sh
```

That is a set of variable configurations to enable the installed Spack packages to execute correctly, things like PATH and so on.

Unfortunately it does not seem trivial nor desirable to generalise it in a hard-codeable way in the Dockerfile. The reason is that some of those variable definitions are dependent on the specific packages being installed. E.g. each Python package installed in the container needs its own specific `PYTHONPATH`.

As such, I would live the ENTRYPOINT as is at this stage.
In the future, a chat with the Spack devs might be useful in figuring out more optimal ways to handle these aspects at the stage of the Spack deployment of packages such as Python ones.

What do you think?

pditommaso:
How that is different from Micromamba ?

https://github.com/mamba-org/micromamba-docker/blob/main/Dockerfile#L60

marcodelapierre:
ah sure, if you are happy with using a service script, I will do that way, thanks!

marcodelapierre:
OK, updated the ENTRYPOINTs in the Dockerfile templates

pditommaso:
Cool! does it mean this PR is ready for review?

marcodelapierre:
Two remaining open points:
* integration with `process.arch`
* refactoring of the Dockerfile templates

At this stage, I am on track to have the PR ready for review by next Monday.

pditommaso:
Update your local copy

marcodelapierre:
> Update your local copy

Done, thanks, I noticed the docs overhaul, going from `rst` to `md`

marcodelapierre:
@pditommaso  for your Docker platform spec (x86, arm, ..), is the backend using the `containerPlatform` string, as sent by the Wave plugin via `SubmitContainerTokenRequest` ?
```
        return new SubmitContainerTokenRequest(
                containerImage: assets.containerImage,
                containerPlatform: assets.containerPlatform,
                containerConfig: containerConfig,
                containerFile: assets.dockerFileEncoded(),
                condaFile: assets.condaFileEncoded(),
                spackFile: assets.spackFileEncoded(),
                buildRepository: config().buildRepository(),
                cacheRepository: config.cacheRepository(),
                timestamp: OffsetDateTime.now().toString(),
                fingerprint: assets.fingerprint()
        )
```

pditommaso:
Nope uses `ContainerPlatform` to handle it 

```
   @Unroll
    def 'should validate platform' () {
        expect:
        ContainerPlatform.of(PLATFORM) == EXPECTED
        ContainerPlatform.of(PLATFORM).toString() == STRING

        where:
        PLATFORM        | EXPECTED                                      | STRING
        null            | new ContainerPlatform('linux','amd64')        | 'linux/amd64'
        'x86_64'        | new ContainerPlatform('linux','amd64')        | 'linux/amd64'
        'amd64'         | new ContainerPlatform('linux','amd64')        | 'linux/amd64'
        'linux/amd64'   | new ContainerPlatform('linux','amd64')        | 'linux/amd64'
        and:
        'arm'           | new ContainerPlatform('linux','arm','v7')     | 'linux/arm/v7'
        'linux/arm'     | new ContainerPlatform('linux','arm','v7')     | 'linux/arm/v7'
        'linux/arm/7'   | new ContainerPlatform('linux','arm','v7')     | 'linux/arm/v7'
        'linux/arm/v7'  | new ContainerPlatform('linux','arm','v7')     | 'linux/arm/v7'
        'linux/arm/5'   | new ContainerPlatform('linux','arm','v5')     | 'linux/arm/v5'
        'linux/arm/v5'  | new ContainerPlatform('linux','arm','v5')     | 'linux/arm/v5'
        and:
        'arm64'         | new ContainerPlatform('linux','arm64')        | 'linux/arm64'
        'linux/arm64'   | new ContainerPlatform('linux','arm64')        | 'linux/arm64'
        'linux/aarch64' | new ContainerPlatform('linux','arm64')        | 'linux/arm64'
        and:
        'linux/arm64/v8'| new ContainerPlatform('linux','arm64')        | 'linux/arm64'
        'linux/arm64/v7'| new ContainerPlatform('linux','arm64','v7')   | 'linux/arm64/v7'

    }
```

marcodelapierre:
yep, but where do you get `PLATFORM` from?  some info expected from the client right?

pditommaso:
It's the value coming from the [request](https://github.com/nextflow-io/nextflow/pull/3636#issuecomment-1527198694) above!

marcodelapierre:
ok, perfect, thanks

marcodelapierre:
Hi @pditommaso , some updates and some questions for you:

1. `process.arch` implemented in Wave+Spack !
2. only remaining minimal bit (under the hood and localised) is using more appropriate templates for the 2 Spack dockerfiles
3. there are two issues that I think require your attention
4. as such, you can start and review the PR, as the only open point on my side is 2. above

Now for the two hot issues for you:
1. the gateway/container pull error that happens every first time that Waves create a new Spack container; we were saying above that this might come from some in-balance in the backend. I imagine it may relate to the big size of some layers in the Spack Wave build. As a partial proof, a few days ago I was testing small changes at the end of the Dockefile (ENTRYPOINT), so new containers were reusing cached layers for the package installation steps, and were NOT failing to pull. To reproduce the error this should be enough `nextflow run marcodelapierre/toy-cowsay-nf -profile spackwave` (define the variable TOWER_ACCESS_TOKEN)
2. issue in `WaveClient.groovy` with the method `defaultFusionUrl()`. This was using the `containerPlatform` property, that was defined at the Wave level. However, the platform variable is now defined at the process level. I have got stuck when trying to refactor in this direction; that method is used by `resolveContainerConfig()`, which among others is used by `sendRequest(String image)`, for which I am struggling to figure out how to pass the process specific platform. As a result, right now some tests within `WaveClientTest.groovy` are inconsistent and/or output errors. Any tip/help here would be much appreciated!

NOTICE
pditommaso: Then, the apache header should go also in the dockerfile

plugins/nf-wave/src/main/io/seqera/wave/plugin/config/WaveConfig.groovy
```diff
@@ -71,7 +73,7 @@ class WaveConfig {
 
     protected List<String> parseStrategy(value) {
         if( !value ) {
-            final defaultStrategy = List.of('container','dockerfile','conda')
+            final defaultStrategy = List.of('container','dockerfile','conda', 'spack')
```
pditommaso, line 76: It'd be better to turn this into a constant and use it also in the check below (line 88)

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
```diff
@@ -230,7 +235,8 @@ class WaveClient {
     }
 
     protected URL defaultFusionUrl() {
-        final isArm = config.containerPlatform()?.tokenize('/')?.contains('arm64')
+        // MARCO MARCO TO BE FIXED - platform is process specific
+        final isArm = false //config.containerPlatform()?.tokenize('/')?.contains('arm64')
```
pditommaso, line 239: I think this can be solved by passing the container platform as parameter to `defaultFusionUrl` and `resolveContainerConfig`. 

When the platform is not available, like in the context of `sendRequest(String image)`, you can yse the default container platform.

pditommaso:
Great work Marco! Regarding the open points: 

1. I'm start to think it can be a problem with the wave backend using a spot instance for the build. It may be not available the very first time.
2. I've commented in the corresponding method 
3. is dropping the support for `ARM32` an option? I don't think it's really used

NOTICE
marcodelapierre: done!

modules/nextflow/src/main/groovy/nextflow/processor/Architecture.groovy
```diff
@@ -73,6 +73,23 @@ class Architecture {
...
+        else if( value == 'arm64/v7' )
+            return 'linux/arm64/v7'
+        else if( value == 'arm' || value == 'arm/v7' || value == 'arm/7' )
+            return 'linux/arm/v7'
+        else if( value == 'arm/v5' || value == 'arm/5' )
+            return 'linux/arm/v5'
+        else
+            throw new IllegalArgumentException("Not a valid `arch` value: ${name}")
+    }
+
```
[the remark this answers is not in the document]
marcodelapierre: minor edit done.

for the unit test, look at this one: 

https://github.com/marcodelapierre/nextflow/blob/add/wave-spack/modules/nextflow/src/test/groovy/nextflow/processor/ArchitectureTest.groovy#L38-L52

Here, column `DOCK` tests the routine `validateArchToDockerArch()`, which you mention, whereas column `SPACK` tests `validateArchToSpackArch()`.

Do you think an additional test would help?

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
```diff
@@ -230,7 +235,8 @@ class WaveClient {
     }
 
     protected URL defaultFusionUrl() {
-        final isArm = config.containerPlatform()?.tokenize('/')?.contains('arm64')
+        // MARCO MARCO TO BE FIXED - platform is process specific
```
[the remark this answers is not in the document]
marcodelapierre: ahah good old coding school at non-CS uni classes LOL

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
```diff
@@ -311,10 +330,10 @@ class WaveClient {
             checkConflicts(attrs, task.lazyName())
 
         //  resolve the wave assets
-        return resolveAssets0(attrs, bundle)
+        return resolveAssets0(attrs, bundle, spackArch, dockerArch)
     }
 
-    protected WaveAssets resolveAssets0(Map<String,String> attrs, ResourcesBundle bundle) {
+    protected WaveAssets resolveAssets0(Map<String,String> attrs, ResourcesBundle bundle, String spackArch, String dockerArch) {
```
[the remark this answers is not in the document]
marcodelapierre: done!

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
```diff
@@ -338,6 +357,25 @@ class WaveClient {
...
         }
 
+        /*
+         * If 'spack' directive is specified use it to create a Dockefile
+         * to assemble the target container
+         */
+        Path spackFile = null
+        if( attrs.spack ) {
+            if( dockerScript )
+                throw new IllegalArgumentException("Unexpected spack and dockerfile conflict")
```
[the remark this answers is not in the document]
marcodelapierre: Good point! However, I added this one by blindly copying the `conda` case above, i.e. `L347-L348` match `L366-L367`. I haven't tested the following,but I actually think these checks should be taken off, because `attrs` has already been validated above at `L327-L330`:
```
        if( config().strategy() )
            attrs = resolveConflicts(attrs, config().strategy())
        else
            checkConflicts(attrs, task.lazyName())
```

I would let you approve whether to delete or not.

marcodelapierre:
> is dropping the support for ARM32 an option? I don't think it's really used. Alternatively, when  ARM32 is specified, it could be thrown an exception in the defaultFusionUrl method saying Fusion does not support ARM32 architecture.

Happy to drop it, unless you know of user cases in commercial cloud.

pditommaso:
> Happy to drop it, unless you know of user cases in commercial cloud

Not really!

modules/nextflow/src/main/groovy/nextflow/processor/Architecture.groovy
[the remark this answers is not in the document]
pditommaso: I see. It's tested indirectly via the constructor. Fair enough, then better to declare it `private`

marcodelapierre:
> > Happy to drop it, unless you know of user cases in commercial cloud
> 
> Not really!

done!

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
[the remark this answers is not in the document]
pditommaso: melius abundare .. :D ok, let's just add `Unexpected spack and dockerfile conflict while resolving wave container`, same for conda

modules/nextflow/src/main/groovy/nextflow/processor/Architecture.groovy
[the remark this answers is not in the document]
marcodelapierre: done!

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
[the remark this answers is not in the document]
marcodelapierre: agree! done!

marcodelapierre:
> I'm start to think it can be a problem with the wave backend using a spot instance for the build. It may be not available the very first time.

Just a memo that It is available when the newly built layers are small

pditommaso:
Could not be a timeout problem?

plugins/nf-wave/src/main/io/seqera/wave/plugin/config/WaveConfig.groovy
marcodelapierre: done!

marcodelapierre:
> Could not be a timeout problem?

the container build succeeds, in fact I get notified via email. It is the subsequent pull that fails.

pditommaso:
Weird. I'll give a try once you have done

pditommaso:
Just checked, and it worked on first try 🤷‍♂️

marcodelapierre:
> Just checked, and it worked on first try 🤷‍♂️

Did you use my pipeline with `-profile spackwave`? I think it worked because the default image is still cached in Wave.

I think you can play with the Dockerfile, `WaveClient.groovy` line 579 and onwards.
First, you add a directive like `RUN echo "hello" >/world` at the end of the Dockerfile, so new layer at the end , should be fine.
But then you add it at the very beginning, invalidating the cache.

pditommaso:
>. Did you use my pipeline with -profile spackwave

yes. let me try again clearing the cache

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
marcodelapierre: so, just to confirm, in `WaveRunCmd.groovy` the `runContainer()` method does not need to know about Fusion?

pditommaso:
Is this the error? 

```
  Unable to find image 'wave.stage-tower.net/wt/2e46a336f730/wave/build/stage:542bc2ddaeecfc5156ab07c0e1f6fcb2' locally
  docker: Error response from daemon: manifest for wave.stage-tower.net/wt/2e46a336f730/wave/build/stage:542bc2ddaeecfc5156ab07c0e1f6fcb2 not found: manifest unknown: repository '195996028523.dkr.ecr.eu-west-1.amazonaws.com/wave/build/stage:542bc2ddaeecfc5156ab07c0e1f6fcb2' not found.
  See 'docker run --help'.


```

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
pditommaso: Not directly. Maybe the best thing is to add a optional `platform` parameter that default to [DEFAULT_DOCKER_PLATFORM](https://github.com/nextflow-io/nextflow/blob/main/plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy#L65-L65)

marcodelapierre:
> Is this the error?
> 
> ```
>   Unable to find image 'wave.stage-tower.net/wt/2e46a336f730/wave/build/stage:542bc2ddaeecfc5156ab07c0e1f6fcb2' locally
>   docker: Error response from daemon: manifest for wave.stage-tower.net/wt/2e46a336f730/wave/build/stage:542bc2ddaeecfc5156ab07c0e1f6fcb2 not found: manifest unknown: repository '195996028523.dkr.ecr.eu-west-1.amazonaws.com/wave/build/stage:542bc2ddaeecfc5156ab07c0e1f6fcb2' not found.
>   See 'docker run --help'.
> ```

No, I typically get one of these two:
```
Command error:
  Unable to find image 'wave.seqera.io/wt/455e45c14efa/wave/build:4a71f62a456a4231cb0f0640b406628e' locally
  docker: Error response from daemon: received unexpected HTTP status: 502 Bad Gateway.
```

```
Command error:
  Unable to find image 'wave.seqera.io/wt/c7e9f7bb170d/wave/build:593503d65400503e25f885163ecdc304' locally
  docker: Error response from daemon: missing or empty Content-Length header.
  See 'docker run --help'.
```

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
marcodelapierre: now updating the unit tests

marcodelapierre: (but having dinner soon)

pditommaso:
Weird, cannot even find it in the logs. When it happen again let me know

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
marcodelapierre: 😂

marcodelapierre: Duff always helps .. done!

marcodelapierre:
Ok, so:
- implemented all your feedback from the past few days
- sorted out issue with `defaultFusionUrl`
- dropped `ARM32`

Tomorrow I will refactor the Dockerfile templates, and then release this PR for your final review and merging.

pditommaso:
Awesome!

marcodelapierre:
@pditommaso I am realising the PR for review and merge!

There are TWO outstanding issues, the first one being new:

1. I have moved to Dockerfile templates, the code builds and the unit tests pass, but .. at runtime I get this error:
```
ERROR ~ Error executing process > 'sayHello (2)'

Caused by:
  Cannot invoke "java.net.URL.openConnection()" because "url" is null
```
e.g. with this trace:
```
java.lang.NullPointerException: Cannot invoke "java.net.URL.openConnection()" because "url" is null
        at org.codehaus.groovy.runtime.ResourceGroovyMethods.configuredInputStream(ResourceGroovyMethods.java:2199)
        at org.codehaus.groovy.runtime.ResourceGroovyMethods.newReader(ResourceGroovyMethods.java:2271)
        at io.seqera.wave.plugin.WaveClient.spackRecipeToDockerFile(WaveClient.groovy:528)
```
it must be something very silly that I am missing about the usage of these methods ... tips?


2. the backend pull errors .. I will run a few other tests once the above is fixed. The dockerfile template makes it much quicker to prototype variations of the dockerfile, that can help troubleshooting this one, e.g. by altering the creation and caching of layers (number and size)

pditommaso:
> have moved to Dockerfile templates, the code builds and the unit tests pass, but .. at runtime I get this error

I believe is happening because BashWrapperBuilder is in another module, try using 

```
        final template = WaveClient.class.getResource('/dockerfile-spack-recipe.txt')
        try(final reader = template.newReader()) {
            final result = new BashTemplateEngine().render(reader, binding)
            return result
        }
```

The key is `WaveClient.class.getResource`, if it works it would also be better to move the files under the path `templates/spack`



2. I'll look into once we merge this

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
marcodelapierre: done

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
marcodelapierre: done

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
marcodelapierre: done

marcodelapierre:
So the last man standing is:

> the backend pull errors .. I will run a few other tests once the above is fixed. The dockerfile template makes it much quicker to prototype variations of the dockerfile, that can help troubleshooting this one, e.g. by altering the creation and caching of layers (number and size)

pditommaso (approved):
Thanks Marco! awesome job!

marcodelapierre:
Thank you Paolo, let's talk at the next PR...!

marcodelapierre:
@tgamblin @alalazo @vsoch it is merged! :-) 

the final bit will be setting up a binary cache of bioinformatics packages to render the service useful...I might ping you on this after my conference frenzy in May!   Todd, Max  see you at ISC!

Issue #3626, opened 2023-02-09
Reports cannot be named with a closure


SamStudio8:
### Expected behavior

Naming a `report`, `timeline` or `trace` file with a closure (in the hope of accessing the `workflow` for introspection) should create a file name that makes sense. 

```
report {
    enabled = true
    file = { "wf-hoot-report-${workflow.sessionId}.html" }
}
trace {
    enabled = true
    file = { "wf-hoot-trace-${workflow.sessionId}.html" }
}
timeline {
    enabled = true
    file = { "wf-hoot-timeline-${workflow.sessionId}.html" }
}
```

### Actual behavior

The files are named with a string description of the parsed closure object.

```
% ls -1 _nf*
_nf_config_186e0cbd$_run_closure1$_closure4@6cd64b3f
_nf_config_186e0cbd$_run_closure2$_closure5@288f173f
_nf_config_186e0cbd$_run_closure3$_closure6@869d87c
```

### Steps to reproduce the problem

* Create a nextflow.config as above
* Run an arbitrary workflow

### Program output 

NA

### Environment 

* Nextflow version: 22.12.0-edge
* Java version: OpenJDK Runtime Environment Homebrew (build 17.0.3+0)
* Operating system: macOS
* Bash version: zsh 5.8

### Additional context

If it is intended that files cannot be named via a closure, Nextflow should abort the workflow upon reading the configuration and warn the user.

bentsherman:
It would be good to document which config options can or can't be closures.

bentsherman:
Following up on this -- Nextflow 25.04 provides strict syntax checking which will report an error if a config option uses a closure where it's not allowed, which should at least improve the error in this case.

I don't think we will add closure support to any more config settings. See #5543 for more discussion about how we might deal with report file names in particular

PR #3651, opened 2023-02-16
Allow closures in file/url for `trace`,`report`,`timeline`,`dag`,`weblog`

fbdtemme:
This adds the option to pass a closure to the following configuration variables:

* `trace.file`
* `report.file`
* `timeline.file`
* `dag.file`
* `weblog.url`

This closure will be resolved in the script execution context but additionally has access to `workflow` and `nextflow`.

Closes: #3626


bentsherman:
@pditommaso what do you think about this idea? I think it would be useful to access variables like the run name in the config. I will need to update this PR but the changes appear to be straightforward.

pditommaso:
Not sure we should this, plain `params` can be used to change those values; along with this the `workflow` was never meant to accessed in the config file

zhilizheng:
@pditommaso  Thanks for the update.  For analysis,  we would run the same pipeline  multiple times with differerent trials, this is very common in analysis (generate log run1, run2, run3..., the users would remember those runName in their note).      We would not like to change the name in the configuration file each time,  especially for the trace or report. A way to automatic those input would be very useful,  because users don't need to remember those, and can put it into the $HOME/.nextflow/config for routines. 
I believe, the pipeline configuration shall be mininal for everyday usage.  So,  a common file would make the user forget those essentials, and keep the project very clean and straightforward for everyday try-and-error works.  Or,  we need to copy those configuration part again and again for each project, it increased the mental efforts.

bentsherman:
Paolo, the main point here is to use the run name in workflow outputs. Can't do that with a param. While I agree that `workflow` shouldn't be accessible in the config file in general, I see nothing wrong with using it in a *closure* in the config file. Many config options already support this, like the `publishDir` path. These report paths should have the same capabilities as `publishDir` path, since they are often published to the same location.

pditommaso:
I'm concerned about the increasing complexity of the configuration file, tho it's true that dynamic values are already supported in the process directives configuration. 

However, this would bring the inconsistency to yet another level. Why `.file` setting could be resolved dynamically and not all settings?

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -380,16 +380,16 @@ class Session implements ISession {
...
 
         // configure script params
         binding.setParams( (Map)config.params )
         binding.setArgs( new ScriptRunner.ArgsList(args) )
 
-        cache = CacheFactory.create(uniqueId,runName).open()
+        this.observers = createObservers()
+        this.statsEnabled = observers.any { it.enableMetrics() }
 
+        cache = CacheFactory.create(uniqueId,runName).open()
```
pditommaso, line 392: Is there a reason why this sequence has been altered?

bentsherman: I think the only necessary change here is to create the workflow metadata before the obsrevers

bentsherman:
> Why .file setting could be resolved dynamically and not all settings?

In theory, yes every config setting could be dynamic. But because each dynamic setting must be configured manually to accept and evaluate a closure, such dynamic settings have been added on an as-needed basis. I don't think that is a reason against making these particular settings dynamic, unless you want to make some generic solution that makes all settings dynamic.

bentsherman:
I added `configDynamic()` to fetch a config option and evaluate it if it's a closure. You just have to give it a binding. Should be a good way to support closures for config options outside of the `process` scope.

bentsherman:
This could also simplify slightly the convention that nf-core uses. Currently they have to fetch their own timestamp:
```groovy
def trace_timestamp = new java.util.Date().format( 'yyyy-MM-dd_HH-mm-ss')
report.file = "${params.outdir}/pipeline_info/execution_report_${trace_timestamp}.html"
```

With this change they could use the timestamp that Nextflow already creates:
```groovy
report.file = { "${params.outdir}/pipeline_info/execution_report_${workflow.start}.html" }
```

Not sure the best way to format an OffsetDateTime, but you get the idea. This way the timestamps are consistent.

pditommaso:
This will introduce yet another exception in the resolution of the config file in which some parts are dynamic and others are not. 

This use case can be addressed by passing a a custom parameter to give the file a different name. Closing without merging.

bentsherman:
No, it cannot be addressed by a param. They want to use the run name which is only available in the workflow metadata. Not wanting to make more settings dynamic is not a good enough reason to not do it IMO. Allowing closures here would enable something that is currently not possible.

risoms:
@fbdtemme @bentsherman Is there any plan on releasing this?

bentsherman:
I don't think so. I suggest using a param as Paolo suggests, even though it requires a bit more logic to set the run name and the param on the command line.

I would like to find a way to better control these report filenames, but I think we can find a better way than just using a closure

PR #3656, opened 2023-02-17
Remove deprecated code (part 1)

bentsherman:
First batch of changes from #3344 . Includes the most straightforward changes that don't break a bunch of tests or require much refactoring.

- DSL1 deprecation errors that are now unreachable since DSL1 will fail at the beginning anyway
- Classes and methods marked as `@Deprecated` that aren't used anywhere else in the code


pditommaso:
Tests are not running

bentsherman:
Yeah I have no idea why... all three PRs are stuck like this

pditommaso:
I suspect the slash in the branch name

PR #3670, opened 2023-02-21
Remove `Channel.from` in favor of `Channel.of`

bentsherman:
Second batch of changes from #3344 . This PR removes the use of `Channel.from` from all tests, since the method is deprecated. I didn't remove the method outright, but I did add a warning to match the documentation.


pditommaso:
I've removed the warnings. `channel.from` is widely used. Let's not annoy users with this. We can continue to have `from` even if marked as deprecated

PR #3671, opened 2023-02-21
Remove deprecated code (part 3)

bentsherman:
Third batch of changes from #3344 . This PR removes some deprecated code related to DSL syntax:
- remove DSL1 process name definitions
- remove old module include syntax
- remove from / into syntax for process inputs/outputs
- remove old process named args syntax
- remove set qualifier for process inputs/outputs

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
@@ -152,63 +152,6 @@ class NextflowDSLImpl implements ASTTransformation {
             super.visitMethod(node)
         }
 
-        protected Statement makeSetProcessNamesStm() {
-            final names = new ListExpression()
-            for( String it: processNames ) {
-                names.addExpression(new ConstantExpression(it.toString()))
-            }
-
-            // the method list argument
-            final args = new ArgumentListExpression()
-            args.addExpression(names)
-
-            // some magic code
-            // this generates the invocation of the method:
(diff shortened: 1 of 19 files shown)


modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
```diff
@@ -498,182 +468,179 @@ class NextflowDSLImpl implements ASTTransformation {
 
             final args = methodCall.arguments as ArgumentListExpression
             final lastArg = args.expressions.size()>0 ? args.getExpression(args.expressions.size()-1) : null
-            final isClosure = lastArg instanceof ClosureExpression
 
-            if( isClosure ) {
-                // the block holding all the statements defined in the process (closure) definition
-                final block = (lastArg as ClosureExpression).code as BlockStatement
+            // the block holding all the statements defined in the process (closure) definition
+            final block = (lastArg as ClosureExpression).code as BlockStatement
```
pditommaso, line 473: How it is inferred that this has to be a ClosureExpression type?

bentsherman: Good catch, we should raise a syntax error here instead of just skipping as before

bentsherman:
@tom-seqera this PR was originally meant to remove a bunch of DSL1 code, but now I think these changes aren't really needed because the whole v1 parser will be deprecated anyway.

So I have pared down this PR to only remove an outdated DSL1 warning, as reported in the linked issue. If the change makes sense to you, please approve it so that I can merge. Let me know if you have any questions

tom-seqera (approved):
LGTM

PR #3685, opened 2023-02-24
Improve transpose and groupTuple examples in the docs

mribeirodantas:
- Add a real example to the groupTuple operator [ci skip]
- Add a real example to the transpose operator [ci skip]


bentsherman:
On second thought, I think it would be better to not use SRA IDs and paths, it feels like too much unnecessary context.

I think you should leave the groupTuple example unchanged, and use the reverse of it for transpose:

```
    Channel
        .of( [1, ['A', 'B', 'C']], [2, ['C', 'A']], [3, ['B', 'D']] )
        .transpose()
        .view()
```

```
    [1, A]
    [1, B]
    [2, C]
    [3, B]
    [1, C]
    [2, A]
    [3, D]
```

mribeirodantas:
Ok! 😉

mribeirodantas:
LGTM?

pditommaso (approved):
Nice

Issue #3507, opened 2022-12-17
Optional input arguments, default value and named arguments.


vruano:
## New feature

In order to increase reusability of processes and (sub)workflows across pipelines we need flexibility in terms of what 
arguments are customized at each invocation. The most convenient syntax would include both:

1. defining default values for arguments as well as
2. the possibility of skipping arguments by naming those that we are giving values to explicitly.

Number (2) is compatible to continue support of positional arguments even with mixtures as is supported by other languages like R.

### Improve dramatically reusability of processes and workflows across pipelines

Very often the same tools are used over and over by different analyses pipelines. These pipeline may need to customize the same tool invocation in different way. Therefore in order to be able to reuse the same code the tool wrapping process/workflow has to be 
a "common multiple" in terms of arguments provide forcing each pipeline to provide sensible defaults for those arguments that are not relevant to them. 

### Shift tool wrapper development to the source

Adding support for optional arguments would give tool providing projects the chance to produce their own Nextflow API/bindings to access their tools for the benefit of Nextflow users. They wouldn't need to guess how their tools are going to be used by third parties. They simply produce flexible wrappers that expose every possible argument where only required one need to be explicitly set.

(What's the main usage case and the deployment scenario addressed by this proposal)

## Suggest implementation 

The syntax for processes definition would be something similar to this:

```nextflow
process tabix {

  input:
        path(input_file)
        val(regions = []) // empty means whole genome
        val(csi = false)
        val(comment = "#")
        val(skip_lines = 0)
        ....
}
```
Then you can invoke it like so:

```nextflow
     tabix(vcf_files, ["chr1", "chr2", "chr3:10000-200000"])
     tabix(input_file = vcf_files, regions =  ["chr1", "chr2", "chr3:10000-200000"])
     tabix(vcf_files, comment = ";")
     tabix(comment = ";", vcf_files) // anonymous parameters are taken in their declaration order.
```

By adding a default value you are implicitly declaring the input 'optional' and so there is no need for the 'optional'  tag/keyword.

bentsherman:
Related to #1694 and #2257 . 100% agree that named/optional params with default arguments would make processes much more re-usable. These features are in our backlog.

pditommaso:
Having a default value when it's not specified it's indeed a good idea. Likely it may be better to rely on the current optional parameters syntax.

notestaff:
[...]

PR #3687, opened 2023-02-24
Default values for process inputs

bentsherman:
Implements default values for process inputs as requested in #3507 

While we could merge this feature as is, I'd like to try to implement optional and named inputs as well, to make sure that they are designed to work together properly.


pditommaso:
This is a good start. If I understand it correctly it works at the level of the process composition ie if the argument is provided matches the ones expected.

But I think there are use cases in which the argument is provided, but the associated channel brings no value (or null). 

A classic example may be [this](https://nextflow-io.github.io/patterns/optional-input/)

bentsherman:
I think the case of a null value could be handled with a separate `nullable` option as discussed [here](https://github.com/nextflow-io/nextflow/pull/2893#issuecomment-1446698892).

If the channel is empty, then the user can use `ifEmpty()` to provide a default value. We could have Nextflow do this automatically if a default value is defined in the process, but I don't know if that is always desirable. It might cause a process to be executed when the user didn't want it. I have to think about this one more.

fabianegli:
In #3714 it is noted that

> The main caveat is that inputs with default values should be declared after inputs with no default, otherwise it's useless.

Is that really a necessary constraint to have? And what does "useless" actually mean on a technical level? It is not quite clear to me from reading through this and the other issue.

bentsherman:
@fabianegli see the docs example I added in this PR for a more detailed explanation. If your process declares an input B with no default value, after an input A with a default value, then you will always have to give a value for A when you call the process, because B comes after A. Thus the default value for A is useless because it can never be used.

I think the caveat effectively goes away if combined with named inputs, because then you can specify the inputs in any order when you call the process.

fabianegli:
I understand the restriction an I am wondering what purpose it might serve. And my point is that I can't think of any situation I'd actually want it. If the defaults are introduced with named inputs, which removes this restriction I see this as something which might surprise some pipeline/module devs rather than a problem.

bentsherman:
> And my point is that I can't think of any situation I'd actually want it.

Not sure what you mean. It's not an artificial restriction that we impose, it's a natural consequence of how arguments are passed in a function call. Most languages that support default values in function calls have this same limitation -- C++, Java, Python, etc.

Also, I guess I should say that named arguments don't "remove" the restriction so much as they provide a workaround by allowing you to order arguments however you like.

fabianegli:
My questioning was aimed to find out wether this behaviour is desirable in this case - irrespective of wether it is introduced intentionally or not. The fact that it is prevalent in other languages seems to hint at it being sensible for some reasons that I am missing. As you say, in some cases it would allow a more organized workflows/processes definition by grouping related inputs independently of wether they have defaults or not. Because the named arguments already make arbitrary input orderings possible, I am not convinced that the effort to remove the order constraint would be worthwhile.

bentsherman:
Gonna fold this one into #4553 doing something like this:
```groovy
process foo {
  input:
    List<Path> inputs
    String args = '--default-option'
  output:
    stdout
  script:
    """
    foo_cmd ${args} ${inputs}
    """
}

workflow {
  foo( file('*.foo') )
}
```

PR #3696, opened 2023-02-28
Update gradle dockerPack

bentsherman:
Brings the gradle dockerPack task up to date. It's useful to build Nextflow locally and run it in a container.

### DIFF
--- a/packing.gradle
+++ b/packing.gradle
@@ -275,7 +275,7 @@ task dockerImage(type: Exec) {
  * Build a docker container image with the current snapshot
  * DO NOT CALL IT DIRECTLY! Use the `make dockerPack` command instead
  */
-task dockerPack(type: Exec, dependsOn: ['pack']) {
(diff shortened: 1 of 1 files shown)


packing.gradle
```diff
@@ -275,7 +275,7 @@ task dockerImage(type: Exec) {
  * Build a docker container image with the current snapshot
  * DO NOT CALL IT DIRECTLY! Use the `make dockerPack` command instead
  */
-task dockerPack(type: Exec, dependsOn: ['pack']) {
+task dockerPack(type: Exec, dependsOn: ['packOne']) {
```
pditommaso, line 278: Not sure about this, what's the need?

bentsherman: This change isn't strictly necessary, just a minor optimization since it doesn't depend on the packAll task.

The other changes are necessary to make the dockerPack work.

pditommaso: Fair enough

PR #3703, opened 2023-03-01
Change access rights of DAG and add a unique ID to Edge and Vertex

Lehmann-Fabian:
To implement the Common Workflow Scheduler as a Nextflow Plugin [(current implementation)](https://github.com/CommonWorkflowScheduler/Nextflow), I need to access the edges and vertices of the DAG.
Moreover, edges and vertices need a unique id.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/dag/DAG.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/dag/DAG.groovy
@@ -40,6 +40,9 @@ import nextflow.script.params.OutParam
 import nextflow.script.params.OutputsList
 import nextflow.script.params.TupleInParam
 import nextflow.script.params.TupleOutParam
+
+import java.util.concurrent.atomic.AtomicLong
+
 /**
  * Model a direct acyclic graph of the pipeline execution.
  *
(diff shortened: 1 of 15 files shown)


bentsherman (approved):

modules/nextflow/src/main/groovy/nextflow/dag/DAG.groovy
```diff
@@ -417,6 +417,10 @@ class DAG {
             indexOf(this)
         }
 
+        int getId() {
+            getOrder()
+        }
+
```
pditommaso, line 423: Maybe `getOrder` can be used. `getId` doesn't look really needed

Lehmann-Fabian: Removed it! Feel free to merge it directly. 
However, introducing an artificial ID independent of the order could avoid potential problems. Should I go for it?

pditommaso: Do you mean using an atomic long in the same it has been done for the `Edge`. It could make sense. Please add a couple of small unit tests in that case

Lehmann-Fabian: Exactly. Should I replace the order or have the id additionally?

pditommaso: I'd to keep `getOrder` in the current form and add `getId` as you are suggesting

Lehmann-Fabian: Okay, done.

pditommaso (approved):
Nice! thanks

Issue #3341, opened 2022-11-02
List of length 1 is converted to path


Mashin6:
## Bug report 
When creating a list with a single element containing a path, this list gets automatically converted to path object.

### Expected behavior and actual behavior
I would expect that list will always be passed to processes as list regardless of number of elements stored in it.

### Steps to reproduce the problem
``` groovy
process C {
    input:
    tuple val(id), path(reads)

    script:
    println reads.size()
}

workflow {
    ch_read_pairs = channel.fromFilePairs( params.reads, checkIfExists: true, size: -1  )
    ch_read_pairs.view()
    C( ch_read_pairs )
}
```

### Program output 
`nextflow run [test.nf](http://test.nf/) --reads *_\{R1,R2\}.fastq` returns number of elements
```
[temp, [/data/temp_R1.fastq, /data/temp_R2.fastq]]
2
```

`nextflow run [test.nf](http://test.nf/) --reads *_\{R1,\}.fastq` returns file size
```
[temp, [/data/temp_R1.fastq]]
62
```
### Environment 

* Nextflow version:  22.04.5 
* Java version: openjdk 17.0.3
* Operating system: linux
* Bash version: GNU bash, 4.4.23

### Additional context

Same happens when defining channel with ` Channel.of( [path('temp_R1.fastq')] )`

bentsherman:
Duplicate of #969, also related to #1236 and #2425. See this comment for the planned solution: https://github.com/nextflow-io/nextflow/issues/2425#issuecomment-1181923896

Issue #2425, opened 2021-11-01
Inconsistent behaviour for glob process outputs


cjw85:
### Expected behavior and actual behavior

According to, https://www.nextflow.io/docs/latest/process.html#multiple-output-files, a glob pattern can be used to emit a list of multiple items from a process into an output channel. What is not clear from the documentation is that if only a single file matches the glob, a length-1 list is not emitted but rather the plain value.

Returning different types from a function (process in this instance) is typically considered bad practice and burdens the caller with having to perform introspection. In the context of of Nextflow this means any operator on a channel first needs to check the returned type, possibly wrap the item thats assumed to be a list, and then safely perform operations such as mapping over the list.

### Steps to reproduce the problem

```
nextflow.enable.dsl=2

process touchOne {
    input:
        val name
    output:
        path "item*"
    script:
        """
        echo "Hello $name"
        touch item1
        """
}

process touchTwo {
    input:
        val name
    output:
        path "item*"
    script:
        """
        echo "Hello $name"
        touch item1
        touch item2
        """
}

workflow {
    touchOne(1).view()
    touchTwo(2).view()
}
```

### Program output 

(Copy and paste here output produced by the failing execution. Please highlight it as a code block. Whenever possible upload the `.nextflow.log` file.)

### Environment 

* Nextflow version: version 20.10.0 build 5430
* Java version: OpenJDK Runtime Environment (build 1.8.0_282-8u282-b08-0ubuntu1~16.04-b08)
* Operating system: Linux
* Bash version: GNU bash, version 4.3.48(1)-release

### Additional context

I can see there being an argument about making a breaking change to Nextflow by enforcing that all globs return a list. Despite Nextflow's eschewing of heavy pattern-matching of file artifacts, I suspect the globbing is obused a lot to emit single files such that a change to emitting lists will break much exisiting code.

However, I do think that at least optionally globbing outputs should be forceable to be lists. I suspect theres some corner cases around what a lenth-0 list might mean and how thats handled. Minimally the documentation needs updating to highlight that the returned type is not always a list.

cjw85:
I'm not aware of any changes in consideration of this issue.

ryankelley-lm:
I agree that this can make workflows brittle, can this issue be re-opened?

cjw85:
That might be a good compromise if you aren't willing to break backwards compatibility.

[...]

Issue #1236, opened 2019-07-22
Output file("*.foo") returns a single file or list


heuermh:
## Bug report 

Capturing wildcard files in an ```output:``` block
```nextflow
process split_foo_gz {
  input:
    set sample, file(foo) from foos
  output:
    set sample, file("*.foo.gz") into splitFoos
...
```
appears to result in either a single file or a list of files, which makes downstream code confusing or tricky, e.g. in this attempt to print the size of the list of files, it prints the size (in bytes presumably?) of a single file

```bash
$ nextflow split_fasta_gz.nf
N E X T F L O W  ~  version 19.04.1
Launching `split_fasta_gz.nf` [amazing_lalande] - revision: 3bb463322d
[warm up] executor > local
executor >  local (2)
[f6/b16ff4] process > split_fasta_gz [  0%] 0 of 2

Split reference to 117968 files:
work/a3/eaa67d4822b8dfe66b5292c4f65352/reference.0.fa.gz

Split transcript to 10 files:
[work/f6/b16ff41dc6aee15e8c45e5f851e6e5/transcript.0.fa.gz, work/f6
/b16ff41dc6aee15e8c45e5f851e6e5/transcript.1.fa.gz, ... >  local (2)
```

### Expected behavior and actual behavior

Always return a list, even if it contains zero or one files.

heuermh:
For example, run `nextflow split_fasta_gz.nf` from https://github.com/heuermh/dishevelled-bio-nextflow

pditommaso:
Yes, this was a bad design choice. Hopefully is going to solved soon in the next syntax review. 

Related #969

bentsherman:
Fixed by https://github.com/nextflow-io/nextflow/commit/42504d3c83145e16179e22df83e9d35ecc995eca

heuermh:
Thank you for following up on this @pditommaso @bentsherman!

The `arity` option appears to be exactly what was needed
> If the arity is `1`, a sole file object will be emitted. Otherwise, a list will always be emitted, even if only one file is produced.

PR #3706, opened 2023-03-01
Add arity option for path inputs and outputs

### LINKED ISSUE #969: Singleton list conversion cause quirk behaviour
## Bug report 

Minimal example:

```Nextflow
    input:
    file input_files from featureCounts_to_merge.collect()

    output:
    file 'merged_gene_counts.txt'

    script:
    //if we only have 1 file, just use cat and pipe output to csvtk. Else join all files first, and then remove unwanted column names.
    def single = input_files instanceof Path ? 1 : input_files.size()
    def merge = (single == 1) ? 'cat' : 'csvtk join -t -f "Geneid,Start,Length,End,Chr,Strand,gene_name"'
    """
    $merge $input_files | csvtk cut -t -f "-Start,-Chr,-End,-Length,-Strand" | sed 's/Aligned.sortedByCoord.out.markDups.bam//g' > merged_gene_counts.txt
    """
```

I expect `input_files.size()` to return the size of the list, e.g. having `n` elements.

Actually, this depends:

- if its just one element, it returns a Path object, so path.size() returns the length of the path as a String
- if its multiple elements, it returns the list size, as expected

See above testcase

### Environment 

* Nextflow version: 18.10.1 
* Java version: 1.8
* Operating system: no-arch

bentsherman:
Closes #969 #1236 #2425 #3341

This PR adds an `arity` option for path inputs and outputs. The arity can be a single number (e.g. `1`) or a range (e.g. `1..*`). The arity is used (1) to validate the expected number of files and (2) to determine, when a path contains only one file, whether to "wrap" that path in a list.

Currently, a path with a single file is always unwrapped. With this PR, the path will be unwrapped only if there is a single file *and* the arity is "single", meaning it can have at most 1 element (in other words, `1` and `0..1` are "single").

The default arity is `1`, or `1..*` if the param has a glob pattern. For optional output params, the default then is `0..1` or `0..*` respectively.

This PR introduces a minor breaking change -- a path input/output with a glob pattern will have a default arity of `1..*`, so it will be wrapped in a list even if there is only one file. Without this PR, this file would be unwrapped. I think this breaking change is worthwhile since it is exactly the inconsistency that users were complaining about.


modules/nextflow/src/main/groovy/nextflow/script/params/ArityParam.groovy
```diff
@@ -0,0 +1,74 @@
...
+
+package nextflow.script.params
+
+
+/**
+ * Implements an arity option for process inputs and outputs.
+ *
+ * @author Ben Sherman <bentshermann@gmail.com>
+ */
+trait ArityParam {
```
pditommaso, line 25: worth adding some unit tests for this trait

pditommaso:
> This PR introduces a minor breaking change -- a path input/output with a glob pattern will have a default arity of 1..*, so it will be wrapped in a list even if there is only one file.

This should be avoided. it could create nasty side effects hard depending the nextflow version. 

Arity should default to `null`, which will implicitly stick to the current behaviour, and use the new rule only when explicitly defined.

bentsherman:
Should be able to make that work. It's too bad though, because the current defaults would probably work 99% of the time without the user having to specify. It would be nice if we could eventually transition to these default settings instead of just `null`.

bentsherman:
ArityParam fails on my laptop but passes in CI 🤷

bentsherman:
I think that is covered by the e2e test I added: [tests/process-arity.nf](https://github.com/nextflow-io/nextflow/pull/3706/files#diff-fd9c080f1d5d455644e7e1da90f3553b1a0508373653974ca6434c707a591ff3)

pditommaso:
Better to add also as unit test

pditommaso:
I found a glitch testing this. If I'm not wrong, when declaring an input `arity: '0..1', it could be a null or a path type, however, I've noticed that when an empty list is given the input type is a list. 

This was my test case 

```
process foo {
  input:
  path x, arity: '0..1'

  script:
  println "inputs [${x.class.simpleName}]=$x"

  """
  echo $x
  """
}


workflow {
  foo([])
}
```

bentsherman:
I think I wasn't sure how to handle that case. You think the list should be coerced into a single null path?

bentsherman:
Now I'm thinking we shouldn't allow the arity to include 0, and instead only support that through the nullable option #2893

pditommaso:
Should be null according this 

> Currently, a path with a single file is always unwrapped. With this PR, the path will be unwrapped only if there is a single file and the arity is "single", meaning it can have at most 1 element (in other words, 1 and 0..1 are "single").

Instead I believe we should drop `nullable` in  favour of this

bentsherman:
> Instead I believe we should drop nullable in favour of this

That could work, actually I think that would be much simpler

> With this PR, the path will be unwrapped only if there is a single file and the arity is "single", meaning it can have at most 1 element (in other words, 1 and 0..1 are "single").

Well, I was only thinking about outputs when I wrote this. If an *input* has `arity = '0..1'`, then it should receive a single file or a "null" file, but not an empty list. An empty list should cause an error.

bentsherman:
Also I think I will add the ability to specify `arity: true` to have the arity inferred from the file pattern. Originally I was doing this by default but this way it can be an opt-in feature.

bentsherman:
I added the `NullPath` from the other PR, and made it fail if a nullable input receives an empty list.

The problem with the null path is that it goes through a bunch of code that it probably shouldn't, and produces this warning:
```
```

bentsherman:
I'm thinking about the [zero one infinity rule](https://en.wikipedia.org/wiki/Zero_one_infinity_rule) and wondering if we should only allow the following:

- `0..1`
- `0..*`
- `1`
- `1..*`

It seems to be that other cases like `1..2` or `2` should be represented with tuples of (possibly nullable) files, for example:

```groovy
// instead of arity: '2'
tuple path('1.txt'), path('2.txt')

// instead of arity: '1..2'
tuple path('1.txt'), path('2.txt', arity: '0..1')
```

bentsherman:
I don't think such a restriction would simplify this PR much, and there might be some weird cases where something like `a..b` is needed, so maybe better to only suggest it as a best practice.

pditommaso:
I've merged this manually, removing the support for `0..1` that opens the problem of how to handle the optional path. It can be added in the following iteration. See 42504d3c.

bentsherman:
Thanks. I will rebase this branch and make a new PR with just the optional features

PR #3712, opened 2023-03-02
Named arguments for processes and workflows

### LINKED ISSUE #2257: Support named channel inputs for processes and workflows

bentsherman:
Closes #2257 #3507 

Adds a `take` option for process inputs that is similar to the `emit` option for outputs. Processes and workflows can now be invoked with named arguments, using the same syntax as native Groovy functions.

Additionally, you don't have to specify the `take` option in the input declaration if the process input has a simple name. For inputs with e.g. GStrings, closures, tuples, the `take` option must be explicit.

It works by interpreting the first process / workflow argument as the named arguments map if it is a map. We have to do this because the invocation happens through `WorkflowBinding.invokeMethod()` rather than an actual function definition. So there will be a problem if you call a process and the first argument is a map, but you meant for that argument to be for the first process input (i.e. as a value channel).

One way that we could avoid this problem is to refactor the NextflowDSL AST transformation to convert process and workflow definitions into actual function definitions, so that the process/workflow inputs are defined as native function arguments.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy
@@ -516,8 +516,10 @@ class ProcessConfig implements Map<String,Object>, Cloneable {
 
     /// input parameters
 
-    InParam _in_val( obj ) {
-        new ValueInParam(this).bind(obj)
+    InParam _in_val( Map opts=null, Object obj ) {
+        new ValueInParam(this)
+                .setOptions(opts)
+                .bind(obj)
     }
 
     InParam _in_file( obj ) {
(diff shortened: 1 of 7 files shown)


bentsherman:
Gonna fold this one into #4553 doing something like this:
```groovy
process foo {
  input:
    List<Path> inputs
    String args = ''
  output:
    stdout
  script:
    """
    foo_cmd ${args} ${inputs}
    """
}

workflow {
  // positional args
  foo( file('*.foo'), '--custom-option' )

  // named args
  foo( inputs: file('*.foo'), args: '--custom-option' )

  // mixing is not allowed
}
```

PR #3719, opened 2023-03-06
Change access rights of submitRateLimit

Lehmann-Fabian:
In my CWS plugin, I overwrite the `submitPendingTasks` method. However, the submitRateLimit is not accessible. Please merge this to give me access.


pditommaso:
Instead of submitting individual PR, would not it be better to create a dedicated PR for there where collecting all required changes?

pditommaso:
In general, I agree, however, considering this effort can take a while several iterations, maybe it's better you prototype into a separate Nextflow branch, so you can prototype it faster and we can discuss how improve it

Lehmann-Fabian:
So, for now, should I join all 4 PRs in a single one, or do we continue as it is?

pditommaso:
Let's move to a uber PR related to CWS effort

PR #3720, opened 2023-03-06
Make GroupKey.target accessible

Lehmann-Fabian:
For my CWS Plugin I need to access the target attribute. I implemented a method similar to the `getGroupSize`.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/extension/GroupKey.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/extension/GroupKey.groovy
@@ -53,6 +53,8 @@ class GroupKey implements CacheFunnel {
 
     int getGroupSize() { size }
 
+    Object getGroupTarget() { target }
+
     /**
      * Delegate any method invocation to the target key object
      *


bentsherman:
Could you elaborate more on why you need this method?

Lehmann-Fabian:
Sure, I transfer information about a task's input to the scheduler. Therefore, I need to extract them. Sometimes the input information is part of the GroupKey.
 Find my code for extraction here: https://github.com/CommonWorkflowScheduler/Nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy#L326

bentsherman:
It looks like the target is functioning like a delegate, even though it isn't marked as `@Delegate`... it would be nice to leverage that somehow, but I don't know how to inspect the type of a delegate without accessing it directly.

pditommaso:
IMO, scheduling should only relay on process input/output interface definition, not internal data structures

Lehmann-Fabian:
Fully agree, but as far as I remember, the target could be an input file in some circumstances. 
Unfortunately, there is not yet a clearly structured Object containing all inputs. At least, I am not aware of it. Is there something I can use?

PR #3721, opened 2023-03-06
Infer commit revision type for azure repos

### LINKED ISSUE #3718: Infer revision type for azure repos
The Azure repository provider should infer the revision type (commit, branch, tag) when fetching a repo. See the [Azure API docs](https://learn.microsoft.com/en-us/rest/api/azure/devops/git/items/get?view=azure-devops-rest-6.0&tabs=HTTP).

For now, should be sufficient to assume it is a commit if it looks like a commit hash. Looks like the API assumes it is a branch by default.

bentsherman:
Closes #3718

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/scm/AzureRepositoryProvider.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/AzureRepositoryProvider.groovy
@@ -30,6 +30,8 @@ import groovy.transform.Memoized
 @CompileStatic
 final class AzureRepositoryProvider extends RepositoryProvider {
 
+    private static final COMMIT_REGEX = ~/[a-zA-Z0-9]{40}/
+
     private String user
     private String repo
     private String continuationToken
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/scm/AzureRepositoryProvider.groovy
```diff
@@ -74,8 +78,12 @@ final class AzureRepositoryProvider extends RepositoryProvider {
                 '$format':'json',
                 'path':path
         ] as Map<String,Object>
-        if( revision )
+        if( revision ) {
             queryParams['versionDescriptor.version']=revision
+
+            if( COMMIT_REGEX.matcher(revision).matches() )
```
pditommaso, line 84: Wonder if it should be supported also partial commit ie. shorter than 40 chars?

bentsherman: I tried, it is not

PR #3723, opened 2023-03-06
Add azure batch pool virtualNetwork option

### LINKED ISSUE #2702: Azure Batch Pool set up within Nextflow can use a private VNet
## New feature

It is useful to run Azure Batch worker pools within a private virtual network.

## Usage scenario 

When using firewalls, private connections, license servers etc it can be useful to run Azure Batch worker pools within a virtual private network. Using the Azure Batch API, it can be configured using an ARM resource identifier string (e.g., `/subscriptions/guid/resourceGroups/resource-group-name/providers/Microsoft.Network/virtualNetworks/network-name/subnet/default`).

Currently we can manually set up a worker pool, then tell Nextflow to execute jobs in this pool, but it would make more sense to add this to the Azure Batch pool specification and allow use of the excellent autopool feature.

## Suggest implementation 

Add `virtualNetwork` to the `azure.batch.pools.<name>` spec, which could be specified in the config in the same manner as sku, vmType etc. 

To achieve this, the `PoolAddParameter` assigned as `poolParams` [here](https://github.com/nextflow-io/nextflow/blob/90ad1dce44ff8caeef09b759d84b7e5db9652a87/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L596) would need to have an additional attribute `NetworkConfiguration` of `NetworkConfiguration` class, where `subnetId` is the string above. See docs [here](https://docs.microsoft.com/en-us/java/api/com.microsoft.azure.batch.protocol.models.networkconfiguration?view=azure-java-stable).

There are additional permissions and quotas required but these are all Azure side. The Microsoft documentation can be found [here](https://docs.microsoft.com/en-us/azure/batch/nodes-and-pools#virtual-network-vnet-and-firewall-configuration).

bentsherman:
Closes #2702 

I followed the suggested implementation from the issue, but I didn't find any unit tests for creating pools. I also haven't tested it end-to-end because I don't have an Azure config set up yet.

### DIFF
--- a/docs/azure.rst
+++ b/docs/azure.rst
@@ -418,6 +418,7 @@ azure.batch.pools.<name>.scaleInterval          Specify the interval at which to
 azure.batch.pools.<name>.schedulePolicy         Specify the scheduling policy for the pool identified with ``<name>``. It can be either ``spread`` or ``pack`` (default: ``spread``).
 azure.batch.pools.<name>.privileged             Enable the task to run with elevated access. Ignored if `runAs` is set (default: ``false``).
 azure.batch.pools.<name>.runAs                  Specify the username under which the task is run. The user must already exist on each node of the pool.
+azure.batch.pools.<name>.virtualNetwork         Specify the subnet ID of a virtual network in which to create the pool (requires version ``22.04.0`` or later).
 azure.registry.server                           Specify the container registry from which to pull the Docker images (default: ``docker.io``, requires ``nf-azure@0.9.8``).
 azure.registry.userName                         Specify the username to connect to a private container registry (requires ``nf-azure@0.9.8``).
 azure.registry.password                         Specify the password to connect to a private container registry (requires ``nf-azure@0.9.8``).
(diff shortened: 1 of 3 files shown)


docs/azure.rst
```diff
@@ -418,6 +418,7 @@ azure.batch.pools.<name>.scaleInterval          Specify the interval at which to
 azure.batch.pools.<name>.schedulePolicy         Specify the scheduling policy for the pool identified with ``<name>``. It can be either ``spread`` or ``pack`` (default: ``spread``).
 azure.batch.pools.<name>.privileged             Enable the task to run with elevated access. Ignored if `runAs` is set (default: ``false``).
 azure.batch.pools.<name>.runAs                  Specify the username under which the task is run. The user must already exist on each node of the pool.
+azure.batch.pools.<name>.virtualNetwork         Specify the subnet ID of a virtual network in which to create the pool (requires version ``22.04.0`` or later).
```
pditommaso, line 421:
```suggestion
azure.batch.pools.<name>.virtualNetwork         Specify the subnet ID of a virtual network in which to create the pool (requires version ``23.03.0-edge`` or later).
```

pditommaso: We are in 2023, are we? 😄

bentsherman: whoopsy

pditommaso:
Ben, we need to understand what's wrong with your PR that does prevent the tests to run

bentsherman:
It's running for the first commit, but the second commit has `[ci skip]`. Maybe the PR status will update when the first commit tests are done.

bentsherman:
They were still running at the time, they are done now. PR status is still yellow though, probably because the latest commit has [ci skip].

pditommaso:
You are right

PR #3724, opened 2023-03-06
Add publish operator

### LINKED ISSUE #1540: Publish operator
## New feature

Publish operator

## Usage scenario 

The current publishDir attribute of processes is not flexible enough. Please consider the following example:

```
process test {
input:
 file('in.txt') from inputChannel
script:
"""
somecmd in.txt -o out1.txt -p out2.txt -r out3.txt
"""
}
```

Now, what if I want to publish out1.txt using move mode, out2.txt using copy mode and output (but not pubish) out3.txt? The current approach is not flexible enough and it becomes complex. I think it would be good t replace the publishDir parameter with a publish operator,like this:

```
output: 
file('out1.txt') into out1
file('out2.txt') into out2
file('out3.txt') into out3
```
And then:
```
out1.publish('.', mode:'move')
out2.publish('output', mode: 'copy')
```

bentsherman:
Closes #1540

### DIFF
--- a/tests/publish-operator.nf
+++ b/tests/publish-operator.nf
@@ -1,0 +1,33 @@
+#!/usr/bin/env nextflow
+/*
+ * Copyright 2023, Seqera Labs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


tests/publish-operator.nf
```diff
@@ -0,0 +1,33 @@
...
+    """
+    touch output1.txt
+    touch output2.txt
+    """
+}
+
+workflow {
+  foo()
+  foo.out[0] | publish('.', mode: 'move')
+  foo.out[1] | publish('outputs', mode: 'copy')
```
bentsherman, line 32: Could be interesting to also support publish on a ChannelOut e.g. `foo | publish`

jmonroynieto:
What behavior would this have with an included process with `publishDir` defined? As a user, my first instinct would be that the workflow call should override it.

bentsherman:
I suspect that using both would cause a race condition. I didn't add any logic to coordinate / deduplicate publishing tasks.

pditommaso:
Closing as not planned

PR #3736, opened 2023-03-09
Add timeout on scheduling wait

pditommaso:
This PR adds a new directive `maxAwait` that allows controlling how long a task can be in submission (pending execution) state, elapsed that time a failure is reported. 

This, along with the use of dynamic directives evaluation and the `task.submitAttempt` attribute make it possible to attempt the execution of the task to a different queue or set of resources requirement.


modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy
```diff
@@ -65,6 +65,7 @@ class ProcessConfig implements Map<String,Object>, Cloneable {
             'machineType',
             'queue',
             'label',
+            'maxAwait',
```
bentsherman, line 68: Think a better name would be `maxSubmitTimeout`

modules/nextflow/src/main/groovy/nextflow/processor/TaskRun.groovy
```diff
@@ -303,6 +303,11 @@ class TaskRun implements Cloneable {
      */
     volatile int failCount
 
+    /**
+     * The number of times the submit of the task has been retried
+     */
+    volatile int submitRetry
```
bentsherman, line 309: Perhaps it should be `submitAttempt` to be consistent with `task.attempt`

modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy
bentsherman: Also maybe should be an executor config option instead of process directive, like all the other timeout / interval settings

modules/nextflow/src/main/groovy/nextflow/processor/TaskRun.groovy
pditommaso: makes sense

modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy
pditommaso: Feel `maxAwait` is more concise. 

> Also maybe should be an executor config option instead of a process directive

Process directives are more powerful because they allow finer control. Usually, it's a defined a executor level, when there isn't a direct process mapping or there's limitation having the setting a process level.

pditommaso:
Think this is ready to go. @bentsherman please give a try to it

bentsherman:
As I understand it, submit timeout errors are limited by `maxRetries`, but the submit retries are counted separately from other retries. So with `maxRetries 3`, a task could fail 3 times due to submit timeout and 3 times due to other retry-able errors, in any order, before the workflow would terminate. Is that right?

pditommaso:
Yes, exactly

bentsherman:
It seems like an odd choice. I would expect either two separate limits for submit retries and regular retries (e.g. `maxSubmitRetries` and `maxRetries`), or for them to be counted together and be subject to the same limit (e.g. `submitRetries + retries <= maxRetries`).

Also, maybe we should call it `maxSubmitAwait` instead of `maxAwait` to denote that it is for the submit time.

pditommaso:
> I would expect either two separate limits for submit retries and regular retries

I see your point, at the same time the submit retry requires setting `errorStrategy=retry` this is why it uses the same setting. 

It could make sense to use a separate setting, but in that case, it should be independent of the `errorStrategy`. 

Let's move forward as "experimental", collect some user feedback and refine it later. 


> maybe we should call it maxSubmitAwait instead of maxAwait to denote that it is for the submit time

Agreed

Issue #3104, opened 2022-08-08
GroupTuple on associative arrays


diego-rt:
Hi,

I'm using an associative array full of metadata as a key for my group tuple. However, samples that differ in one field are being incorrectly merged. Don't know if this is the expected behaviour but I think it is probably not a desirable one. 

```
#!/usr/bin/env nextflow
nextflow.enable.dsl = 1

/*
 * Defines the pipeline inputs parameters
 */

Channel
    .from( [ "sample_name":"sample_A" ], [ "sample_name":"sample_B" ], [ "sample_name":"sample_C" ] )
    .set { samples_ch }

Channel
    .from( ["scaffolds", "file_s"], ["contigs", "file_c"])
    .set { genomes_ch }

samples_ch
    .combine(genomes_ch)
    .view()
    .map { sample_info, genome_name, index -> 
        sample_info."genomeName" = genome_name
        [ sample_info, index ] 
        }
    .view()
    .groupTuple()
    .view()
    
```
### Program output 

Observed:

```
sh-3.2$ ./nextflow run nf-playground.nf 
N E X T F L O W  ~  version 22.04.5
Launching `nf-playground.nf` [happy_einstein] DSL1 - revision: f149707ab9
[[sample_name:sample_A], scaffolds, file_s]
[[sample_name:sample_B], scaffolds, file_s]
[[sample_name:sample_C], scaffolds, file_s]
[[sample_name:sample_A], contigs, file_c]
[[sample_name:sample_B], contigs, file_c]
[[sample_name:sample_C], contigs, file_c]

[[sample_name:sample_A, genomeName:scaffolds], file_s]
[[sample_name:sample_B, genomeName:scaffolds], file_s]
[[sample_name:sample_C, genomeName:scaffolds], file_s]
[[sample_name:sample_A, genomeName:contigs], file_c]
[[sample_name:sample_B, genomeName:contigs], file_c]
[[sample_name:sample_C, genomeName:contigs], file_c]

[[sample_name:sample_A, genomeName:contigs], [file_s, file_c]]
[[sample_name:sample_B, genomeName:contigs], [file_s, file_c]]
[[sample_name:sample_C, genomeName:contigs], [file_s, file_c]]
```

Desired:

```
sh-3.2$ ./nextflow run nf-playground.nf 
N E X T F L O W  ~  version 22.04.5
Launching `nf-playground.nf` [happy_einstein] DSL1 - revision: f149707ab9
[...]

Issue #2660, opened 2022-02-18
[DSL2] modifying a mutable object (like Map) in a duplicated channel affects all instances


Midnighter:
### Expected behavior and actual behavior

I don't know if this is a bug or just a huge gotcha. Either way, I've been affected by it multiple times and so have others. DSL2 is supposed to automatically copy/duplicate channels when they get re-used multiple times. In general, this works perfectly fine. However, when the channel contains a mutable objects like a `Map` and those get modified in place, this affects all channel instances. I don't know the underlying code but I guess for memory reason channels get copied only shallowly, i.e., they reference the same objects. This can lead to more or less subtle bugs in pipelines and surprising behavior.

### Steps to reproduce the problem

Consider the following workflow:

```nextflow
#!/usr/bin/env nextflow

nextflow.enable.dsl = 2

process ECHO {
    input:
    val meta

    output:
    val meta

    script:
    """
    echo '${meta.id}'
    """
}

process ECHO2 {
    input:
    val meta

    output:
    val meta

    script:
    """
    echo '${meta.id}'
    """
}

workflow {
    def ch_input = Channel.of([id: 'test1', idx: 1], [id: 'test2', idx: 2])

    ECHO(ch_input).view()

    ECHO2(
        ch_input.map { meta ->
            meta.id = 'foo'
            meta
        }
    ).view()
}
```

This produces below output where all `id` keys contain the `'foo'` value. If instead you clone the map, the output is more in line with what I expect.

```nextflow
#!/usr/bin/env nextflow

nextflow.enable.dsl = 2

process ECHO {
    input:
    val meta

    output:
    val meta

    script:
    """
    echo '${meta.id}'
    """
}

process ECHO2 {
    input:
    val meta

    output:
    val meta

    script:
    """
    echo '${meta.id}'
    """
}

workflow {
    def ch_input = Channel.of([id: 'test1', idx: 1], [id: 'test2', idx: 2])

    ECHO(ch_input).view()

    ECHO2(
        ch_input.map { meta ->
            def dup = meta.clone()
            dup.id = 'foo'
            dup
        }
    ).view()
}
```

### Program output 

```
N E X T F L O W  ~  version 21.10.6
Launching `mutable.nf` [friendly_koch] - revision: 49b9711e9b
executor >  local (4)
[98/ed1c1a] process > ECHO (1)  [100%] 2 of 2 ✔
[a9/4e05a2] process > ECHO2 (2) [100%] 2 of 2 ✔
[...]

PR #3738, opened 2023-03-09
Make GroupKey cloneable

bentsherman:
See discussions in #3104 and #2660

When modifying maps, it is generally better to copy the map, e.g. `metadata + [id: 'foo']`. However, if the map contains a GroupKey, it won't work. This PR simply makes GroupKey cloneable so that you don't have to remove and re-add the group key to clone a map.


pditommaso:
Fair enough

PR #3739, opened 2023-03-09
Support maps for combine, join, and groupTuple operators

### LINKED ISSUE #3108: Support string indices for operators with `by` option
Spawned from the language improvements mega-discussion (#3107). Operators that combine or group channels based on an index (i.e. `by` option) currently only support integer indices. They should also support string indices so that they can be used with maps.

To my knowledge, these operators include:
- groupTuple
- join
- combine

bentsherman:
Closes #3108 

Certain operators are designed to work with tuples but don't support maps. To encourage the use of maps in Nextflow pipelines, these operators should be extended (or duplicated) to also support maps.

TODO:
- [ ] combine
- [ ] join
- [x] groupTuple

### DIFF
--- a/docs/operator.rst
+++ b/docs/operator.rst
@@ -794,13 +794,44 @@ according to these rules:
 * For any other value, the value itself is used as a key.
 
 
+.. _operator-groupmap:
+
+groupMap
+----------
+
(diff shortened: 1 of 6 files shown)


bentsherman:
Closing since no one is really asking for this feature. Likely will develop something similar for record types one day

PR #3766, opened 2023-03-16
Kuberun: add plugins to config

Lehmann-Fabian:
Using `kuberun` with `-plugins xy` does not yet hand this over to the head-pod. This is implemented in this PR. It would be nice to have this, especially for CWS.


pditommaso:
I feel a test is missing 😄

pditommaso:
You are right.

Lehmann-Fabian:
Thank you, seems to fix it :)

pditommaso (approved):

PR #3784, opened 2023-03-21
Document `-pool-size` run option

### LINKED ISSUE #3014: Missing documentation information about poolSize / -pool-size

I suppose there should have documentation on the `-pool-size` command line parameter and also the `config.poolSize`, that controls the number of threads in the execution pool.

Check the [documentation](https://www.nextflow.io/docs/latest/index.html) and `nextflow help run`.

### Environment 

* Nextflow version: version 22.04.3 build 5703
* Java version: openjdk version "16.0.1" 2021-04-20
* Operating system: macOS Monterey 12.4
* Bash version: GNU bash, version 3.2.57(1)-release (arm64-apple-darwin21)

### Additional context

It's not clear to me how to better describe this on the documentation, so instead of opening a Pull Request I decided to open an issue to make sure this will not be forgotten, as asked by @ewels 😬

bentsherman:
Closes #3014 

Adds an example with the `run` command, showing how to make Nextflow run more concurrent tasks than the number of available CPUs. Useful for a pipeline with many small tasks that run locally.


bentsherman:
Here is an example pipeline to test with the example command:
```groovy
params.sleep = 30

process overwhelm {
   debug true

   input:
      val(idx)

   """
   echo "Be overwhelmed by task ${idx}"
   sleep ${params.sleep}
   """
}

workflow {
   ch_idx = Channel.of(1..1000)
   overwhelm(ch_idx)
}
```

It works with 22.10. Also works with `exec` tasks.

docs/cli.rst
```diff
@@ -1220,10 +1224,17 @@ facilitates rapid iterations, inspections of any pipeline as well as debugging.
...
     $ nextflow run main.nf -with-report
 
-- Invoke the nextflow pipeline execution with a custom queue size. By default, the value of **queue-size** is the same as the number of available CPUs. ::
+- Execute a pipeline with a custom queue size. By default, the default queue size is the number of available CPUs. ::
 
     $ nextflow run nextflow-io/hello -qs 4
 
+- Execute a pipeline locally with more concurrent tasks than the number of available CPUs. ::
+
+    $ nextflow run main.nf -executor.cpus 100 -ps 100 -qs 100
```
pditommaso, line 1233: This can be very inefficient why CPU bound tasks. Why advise it?

bentsherman: It's meant for small local tasks that use less than 1 CPU. It's the best way to handle this case until we support fractional CPUs.

bentsherman: See also this Slack thread: https://nextflow.slack.com/archives/C02T98A23U7/p1656081406989489

bentsherman: I have updated this PR and clarified when this is appropriate to use.

pditommaso:
I think it's preferable to keep hiding the `-ps` and `-executor` options. The first is going to be useless soon with virtual thread, the latter has always been an experiment and likely removed from CLI v2. 

I also believe it should be suggested to use 1k cpus in the CLI docs. Eventually, it's a topic for a blog or a FAQ because it has complex implications to be explained

bentsherman:
Heard about an interesting use case related to this issue -- a Nextflow pipeline that launches 1000s of tasks, where each task makes an API call and waits for a response. In this case you would want to run many more tasks that the number of CPUs. Could be an interesting topic for a blog post on virtual threads.

pditommaso:
Looking forward to reading about that!

PR #3802, opened 2023-03-27
Add initial task graph and metadata json file

bentsherman:
Redo of #3463 . Includes the following changes:

- `ConcreteDAG` class which tracks the concrete task DAG as the pipeline is executed
- `dag.type` config option which allows the user to render the task DAG instead of the regular DAG (currently only supported for Mermaid format)
- `.command.meta.json` file which is written on task completion and contains the task hash, input files, and output files

To test the DAG rendering, launch a pipeline with the following extra config:
```groovy
dag.enabled = true
dag.type = 'concrete'
dag.file = 'pipeline.mmd'
dag.overwrite = true
```

To test the task metadata file, simply launch a pipeline normally and inspect the task directories for `.command.meta.json` files.

Here is an example meta file from `rnaseq-nf`:
```json
{
    "hash": "4fb5c35e186526f64ee5a5cc2720e824",
    "inputs": [
        {
            "name": "ggal_gut",
            "path": "/work/ab/bc682dd08c72ded3c25640d3cb05ef/ggal_gut",
            "predecessor": "abbc682dd08c72ded3c25640d3cb05ef"
        },
        {
            "name": "fastqc_ggal_gut_logs",
            "path": "/work/c0/6030412e5508bc9c2f2d7b053eb882/fastqc_ggal_gut_logs",
            "predecessor": "c06030412e5508bc9c2f2d7b053eb882"
        },
        {
            "name": "multiqc",
            "path": "/.nextflow/assets/nextflow-io/rnaseq-nf/multiqc",
            "predecessor": null
        }
    ],
    "outputs": [
        {
            "name": "multiqc_report.html",
            "path": "/work/4f/b5c35e186526f64ee5a5cc2720e824/multiqc_report.html",
            "size": 1204127,
            "checksum": "a41d49c28135c51b7c92fb70cac70a66"
        }
    ]
}
```

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskHandler.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskHandler.groovy
@@ -21,7 +21,11 @@ import static nextflow.processor.TaskStatus.*
 
 import java.nio.file.NoSuchFileException
 
+import groovy.json.JsonBuilder
 import groovy.util.logging.Slf4j
+import nextflow.dag.ConcreteDAG
+import nextflow.extension.FilesEx
+import nextflow.script.params.FileOutParam
 import nextflow.trace.TraceRecord
 /**
  * Actions to handle the underlying job running the user task.
@@ -212,6 +216,31 @@ abstract class TaskHandler {
         }
 
         return record
+    }
+
+    void writeMetaFile() {
+        final record = [
+            hash: task.hash.toString(),
+            inputs: task.getInputFilesMap().collect { name, path ->
+                [
+                    name: name,
+                    path: path.toString(),
+                    predecessor: ConcreteDAG.getPredecessorHash(path)
+                ]
+            },
+            outputs: task.getOutputsByType(FileOutParam).values().flatten().collect { path ->
+                final file = path.toFile()
+
+                [
+                    name: path.name,
+                    path: path.toString(),
+                    size: file.size(),
+                    checksum: file.isFile() ? FilesEx.getChecksum(path) : null
+                ]
+            }
+        ]
+
+        task.workDir.resolve(TaskRun.CMD_META).text = new JsonBuilder(record).toString()
     }
 
     /**
--- a/modules/nf-commons/src/main/nextflow/extension/FilesEx.groovy
+++ b/modules/nf-commons/src/main/nextflow/extension/FilesEx.groovy
@@ -35,6 +35,7 @@ import java.nio.file.attribute.FileAttribute
 import java.nio.file.attribute.FileTime
 import java.nio.file.attribute.PosixFilePermission
 import java.nio.file.attribute.PosixFilePermissions
+import java.security.MessageDigest
 
 import groovy.transform.CompileStatic
 import groovy.transform.PackageScope
@@ -1599,5 +1600,12 @@ class FilesEx {
 
     static String getScheme(Path path) {
         path.getFileSystem().provider().getScheme()
+    }
+
+    static String getChecksum(Path path) {
+        final data = Files.readAllBytes(path)
+        final hash = MessageDigest.getInstance("MD5").digest(data)
+
+        new BigInteger(1, hash).toString(16)
     }
 }
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -254,6 +254,7 @@ Name                Description
 enabled             When ``true`` turns on the generation of the DAG file (default: ``false``).
 file                Graph file name (default: ``dag-<timestamp>.dot``).
 overwrite           When ``true`` overwrites any existing DAG file with the same name.
+type                Can be ``abstract`` to render the abstract DAG or ``concrete`` to render the concrete (task) DAG (default: ``abstract``).
 ================== ================
 
 The above options can be used by prefixing them with the ``dag`` scope or surrounding them by curly
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -40,6 +40,7 @@ import nextflow.conda.CondaConfig
 import nextflow.config.Manifest
 import nextflow.container.ContainerConfig
 import nextflow.dag.DAG
+import nextflow.dag.ConcreteDAG
 import nextflow.exception.AbortOperationException
 import nextflow.exception.AbortSignalException
 import nextflow.exception.IllegalConfigException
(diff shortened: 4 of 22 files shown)


bentsherman:
I wrote a little Python script to scrape the metadata files and render a task DAG with output files. Here's what the DAG looks like for `rnaseq-nf`:

```mermaid
flowchart TD
    tc8b1fd5c["c8/b1fd5c"]
    tc0603041["c0/603041"]
    tabbc682d["ab/bc682d"]
    tc8b1fd5c -->|index| tabbc682d
    t4fb5c35e["4f/b5c35e"]
    tabbc682d -->|ggal_gut| t4fb5c35e
    tc0603041 -->|fastqc_ggal_gut_logs| t4fb5c35e
    t4fb5c35e -->|multiqc_report.html| o1(( ))
```

I think I will extend the task DAG renderer in this PR to also include the output files like this.

bentsherman:
Currently it only tracks files. I haven't tried to track things like `value` and `env` because they can't be saved as workflow outputs. I guess I'm not clear what the argument is for tracking them -- annotating the task DAG? storing them in the metadata json?

bentsherman:
Cloud executor tests are failing because some of the file operations I'm doing aren't supported, I'll have to think more carefully about how to handle remote files.

modules/nextflow/src/main/groovy/nextflow/processor/TaskHandler.groovy
```diff
@@ -214,6 +218,31 @@ abstract class TaskHandler {
...
+            },
+            outputs: task.getOutputsByType(FileOutParam).values().flatten().collect { path ->
+                final file = path.toFile()
+
+                [
+                    name: path.name,
+                    path: path.toString(),
+                    size: file.size(),
+                    checksum: file.isFile() ? FilesEx.getChecksum(path) : null
+                ]
```
bentsherman, line 239: @pditommaso I think this code doesn't work for object storage because they don't support `path.toFile()`. Is there a better way to get the file size and checksum?

pditommaso: We talked, it'd be better to rely on the checksum provided by the cloud object storage when available and fallback to compute it otherwise. Checksum should be computed using an inputstream to avoid overloading the heap with big files

pditommaso: Even better, it could be used an interface `EtagAware` and implement it in [AzFileAttributes](https://github.com/nextflow-io/nextflow/blob/38c8bd335d286ddc3907254b9d6817581f2a67d0/plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzFileAttributes.groovy#L37-L37) and [S3FielAttributes](https://github.com/nextflow-io/nextflow/blob/38c8bd335d286ddc3907254b9d6817581f2a67d0/plugins/nf-amazon/src/main/com/upplication/s3fs/S3FileAttributes.java#L48-L48) (google is a problem because, there's no a nextflow implementation for it)

bentsherman: You want to implement a filesystem for Google cloud storage? 😄

pditommaso: Not really, but I guess an hack can be found to do the same

modules/nf-commons/src/main/nextflow/extension/FilesEx.groovy
```diff
@@ -1599,4 +1602,20 @@ class FilesEx {
...
+            return null
+
+        if( path instanceof ETagAwareFile )
+            return ((ETagAwareFile)path).getETag()
+
+        final md = MessageDigest.getInstance('MD5')
+        final is = Files.newInputStream(path)
+        final dis = new DigestInputStream(is, md)
+
+        while( dis.read() != -1 ) {}
```
bentsherman, line 1617: With the `DigestInputStream` I have to read through the stream to trigger the md5 computation. This code is reading the stream one byte at a time, which surely is very inefficient. I don't know if the stream is buffering under the hood. Is there a more efficient way to do this? I could also read into a dummy buffer, e.g. 1 MB at a time or something.

pditommaso: Likey buffering should speed it up, but I don't know this interface. What chatgpt says!? 😄

pditommaso: stackoverflow for the win!

bentsherman: Good point, SO recommends ~16KB, so that's what I'll use

https://stackoverflow.com/questions/8748960/how-do-you-decide-what-byte-size-to-use-for-inputstream-read

bentsherman: By the way, I asked chatgpt about the optimal buffer size and it told me to run a benchmark. So I asked it to write the benchmark for me. And it did! And it worked! 😆

modules/nf-commons/src/main/nextflow/extension/FilesEx.groovy
```diff
@@ -1599,4 +1603,20 @@ class FilesEx {
     static String getScheme(Path path) {
         path.getFileSystem().provider().getScheme()
     }
+
+    static String getChecksum(Path path) {
+        if( Files.isDirectory(path) )
+            return null
+
```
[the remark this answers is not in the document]
bentsherman: @sonatype-lift ignore

bentsherman:
Thinking about how the task graph intersects with the automatic cleanup. If we save the task inputs and outputs to the `.nextflow` cache, so that we can resume the task even if the outputs are deleted, maybe then we don't need to save a JSON file for each task. Tower could retrieve this information from the `.nextflow` cache, it could even use Nextflow's `CacheDB` class to do it natively, no need to run a nextflow command.

I'm also thinking this because the automatic cleanup could delete the entire task directory, not just the output files, in that case the JSON file is pointless because it will just be deleted.

bentsherman:
Okay, now it saves the task input/output metadata to the cache db instead of json files, via the TraceRecord. But it doesn't add `inputs` and `outputs` as fields in the TraceRecord, so they won't be included in the events that are pushed to Tower. Instead, Tower can query this data from the cache db.

This work is now folding into the automatic cleanup and S3 cache backend because (1) I also need this input/output metadata to restore deleted tasks and (2) an S3 cache backend would probably be much easier for Tower to work with. So I will focus on those PRs first.

marcodelapierre:
For reference, this discussion: https://github.com/nextflow-io/nextflow/discussions/3447

Issue #3606, opened 2023-02-03
Orphaned processes when using Singularity and trace


krokicki:
### Expected behavior and actual behavior

All descendant processes should be cleaned up when a Nextflow pipeline is interrupted or cancelled. This works in most cases, but when `singularity` and `trace` are both enabled the command.run file produces process trees which are not cleaned up when the pipeline is interrupted (e.g. using CTRL-C). 

### Steps to reproduce the problem

nextflow.config:
```
singularity.enabled = true
trace.enabled = true
```

test.nf
```
#!/usr/bin/env nextflow
nextflow.enable.dsl = 2
process test {
    container 'biocontainers/biocontainers:latest'
    shell:
    '''
    sleep 300
    '''
}
workflow {
    test()
}
```

### Program output 

When trace is not enabled, the process tree like this:

```
31887 ?        Ss     0:00  \_ sshd: rokickik [priv]
31892 ?        S      0:01  |   \_ sshd: rokickik@pts/7
31893 pts/7    Ss     0:07  |       \_ -bash
17285 pts/7    Sl+    0:22  |           \_ /usr/lib/jvm/zulu-17.0.3/bin/java (…) nextflow.cli.Launcher run test.nf
17423 pts/7    S+     0:00  |               \_ /bin/bash -ue .command.run
17439 pts/7    S+     0:00  |                   \_ tee .command.out
17440 pts/7    S+     0:00  |                   \_ tee .command.err
17441 pts/7    S+     0:00  |                   \_ /bin/bash -ue .command.run
17442 pts/7    Sl+    0:00  |                       \_ Singularity runtime parent
17464 pts/7    S+     0:00  |                           \_ /bin/bash -ue /groups/scicompsoft/home/rokickik/dev/test-ctrlc/work/21/7ab499f9054318ba49e7755b76551e/.command.sh
17491 pts/7    S+     0:00  |                               \_ sleep 300
```

When you CTRL-C the pipeline, it cleans up all of these processes via the `on_term` handler (thanks to [#1561](https://github.com/nextflow-io/nextflow/issues/1561)).

However, once you enable trace (or timeline/report/etc), the process tree looks like this:

```
31887 ?        Ss     0:00  \_ sshd: rokickik [priv]
31892 ?        S      0:01  |   \_ sshd: rokickik@pts/7
31893 pts/7    Ss     0:07  |       \_ -bash
[...]

PR #3803, opened 2023-03-27
added termination trap in nxf_trace to call cleanup function

krokicki:
To completed address #3606, the trace version of the script needs to trap TERM/INT signals and invoke on_term.

### DIFF
--- a/modules/nextflow/src/main/resources/nextflow/executor/command-trace.txt
+++ b/modules/nextflow/src/main/resources/nextflow/executor/command-trace.txt
@@ -217,6 +217,7 @@ nxf_trace_linux() {
 }
 
 nxf_trace() {
+    trap on_term TERM INT
     local trace_file=.command.trace
     touch $trace_file
     if [[ $(uname) = Darwin ]]; then


pditommaso:
Was this closed by mistake?

krokicki:
Sorry, I closed it when I realized that I wasn't familiar enough with the contribution process because all of the checks failed. Unfortunately, I don't have time to figure that out right now. And it's a one line change, so I figured the value of a PR was low...

pditommaso:
Well, if it solves the problem, the value is high. I guess some tests very failing. We can look into it

pditommaso:
I've investigated this further, and it turns out that to have the termination signal properly propagated the `--pid` option needs to be provided. 

This commit enables it by default https://github.com/nextflow-io/nextflow/commit/3601b01496c6f7d1d85c6a5cfc798c901db98d88 

Thanks for reporting this problem

krokicki:
Tested with 23.04.0 and it works for cancelling my Spark cluster tasks. Thanks so much, Paolo!

Issue #3308, opened 2022-10-21
Scripts in module bin directory not found


phiweger:
### Expected behavior and actual behavior

Scripts should be found in their module's bin dir as described in #1798 -- However, following the directory structure suggested therein and enabling the feature in `nextflow.config` still errs out with `.command.sh: line 2: xxx.py: command not found`

### Environment 

* Nextflow version: version 22.10.0 (installed via conda)
* Java version:

openjdk version "1.8.0_121"
OpenJDK Runtime Environment (Zulu 8.20.0.5-macosx) (build 1.8.0_121-b15)
OpenJDK 64-Bit Server VM (Zulu 8.20.0.5-macosx) (build 25.121-b15, mixed mode)

* Operating system: macOS
* Bash version: (use the command `$SHELL --version`) GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin20)

mribeirodantas:
Hello, @phiweger !

Could you please share a minimal reproducible example?

phiweger:
I wrote a small example @mribeirodantas -- https://github.com/phiweger/mocks/tree/main/nextflow

The error occurs, when the module imports the process that uses the script, instead of having the process defined in the `main.nf` of the module. If you run the example, it works. However:

If you uncomment 

https://github.com/phiweger/mocks/blob/main/nextflow/modules/bar/main.nf#L1

and comment out the process

https://github.com/phiweger/mocks/blob/main/nextflow/modules/bar/main.nf#L4

it fails.

phiweger:
@pditommaso 

When I add `utils.nf` to the bar module like so:

```
main.nf
utils.nf
resources/
```

it tells me in cannot find the script.

When I add the resources to the other module like so

```
main.nf
processes/
     utils.nf
     resources/
```

It still fails to find the script. What is the folder structure you allude to?

ahstram:
Hi @pditommaso, @phiweger I was running into the same issue and was able to fix it by removing two lines from ScriptMeta.groovy, please see PR #3809

pditommaso:
Solved by 2c5a6878. Credits @ahstram.

PR #3809, opened 2023-03-28
Ensure that we return module bundles even when script name is not mai…

ahstram:
This is intended to address issue #3308


pditommaso:
Thanks for spotting this. I've merged manually with some minor changed 👉 2c5a6878.

PR #3830, opened 2023-04-03
Operator splitJson

lindenb:
Don't merge ! :-)

This PR is preliminary, I just want to know if the NF community would be interested by this new operator that scans a JSON array and emit every item (or scan and emit each key+value of a json object). If yes, i will try fixing things like the doc, spacing etc...

The JsonSplitter class uses the streaming parser from google gson.

So here is an example. The operator split each record of a json array containing a description of 3 movies. For each movie, it download the picture and at the end everything is zipped.

```nf
workflow {
	ch1 = GET_JSON()
	ch2 = GET_POSTER(ch1.output.splitJson([:]))
	ZIP(ch2.output.collect())
}

process GET_JSON {
output:
	path("jeter.json"),emit:output
script:
"""
cat << EOF > jeter.json
[
	{
	"Title":"A new hope",
	"Poster":"https://upload.wikimedia.org/wikipedia/en/thumb/8/87/StarWarsMoviePoster1977.jpg/220px-StarWarsMoviePoster1977.jpg"
	},
	{
	"Title":"Empire Strikes Back",
	"Poster":"https://upload.wikimedia.org/wikipedia/en/thumb/3/3f/The_Empire_Strikes_Back_%281980_film%29.jpg/220px-The_Empire_Strikes_Back_%281980_film%29.jpg"
	},
	{
	"Title":"Return of the Jedi",
	"Poster":"https://upload.wikimedia.org/wikipedia/en/thumb/b/b2/ReturnOfTheJediPoster1983.jpg/220px-ReturnOfTheJediPoster1983.jpg"
	}
]
EOF
"""
}

process GET_POSTER {
tag "${row.Title}"
input:
	val(row)
output:
	path("${row.Title.replaceAll(" ","_")}.jpg"),emit:output
script:
"""
wget -O "${row.Title.replaceAll(" ","_")}.jpg" "${row.Poster}"
"""
}

process ZIP {
input:
	val(L)
script:
"""
zip -0 -j jeter.zip ${L.join(" ")}
"""
}
```

execute:
```
$ ./launch.sh run -resume jeter.nf 
N E X T F L O W  ~  version 23.04.0
Launching `jeter.nf` [tender_knuth] DSL2 - revision: e6c4ac236c
executor >  local (5)
[ae/921395] process > GET_JSON                         [100%] 1 of 1 ✔
[a3/c8f718] process > GET_POSTER (Empire Strikes Back) [100%] 3 of 3 ✔
[1b/76435f] process > ZIP                              [100%] 1 of 1 ✔

lindenb@okazaki:~/src/nextflow$ unzip -l work/1b/76435f00cd3e3cd44e5a37601de89e/jeter.zip 
Archive:  work/1b/76435f00cd3e3cd44e5a37601de89e/jeter.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
    27927  2022-01-23 03:09   Return_of_the_Jedi.jpg
    24538  2022-01-24 00:54   A_new_hope.jpg
    28901  2022-10-25 05:23   Empire_Strikes_Back.jpg
---------                     -------
    81366                     3 files

```


bentsherman:
I like it. I don't think JSON is used as much in life sciences, but it is in ML and data science, plus it would be nice to support the same set of operators for CSV and JSON alike. No new dependencies, so I think your approach should be fine.

My only notes:
- document the new operators in `operator.rst`
- clean up indentation in new files, use 4 spaces instead of tab

lindenb:
@bentsherman thanks , I'll do it tomorrow.

Nevertheless , I'm not sure about the following statements because I don't have a clear view of what they're doing:

https://github.com/nextflow-io/nextflow/blob/0c36159ef1e61ba81f44da5f381ac9250bf7d887/modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy#L58
https://github.com/nextflow-io/nextflow/blob/0c36159ef1e61ba81f44da5f381ac9250bf7d887/modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy#L59


about collectors:

https://github.com/nextflow-io/nextflow/blob/0c36159ef1e61ba81f44da5f381ac9250bf7d887/modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy#L110


https://github.com/nextflow-io/nextflow/blob/0c36159ef1e61ba81f44da5f381ac9250bf7d887/modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy#L87

bentsherman:
The abstract text splitter can accept input from a file or records in memory, and it looks like splitCsv doesn't support file input but other splitters like splitFasta and splitFastq do. Since you're adding the same level of support as splitCsv, that should be fine.

As for the collector, I can see that's an interface from one of the parent classes but I don't know anything about it. Looks like the other splitters don't use it either. When Paolo reviews it we'll see if your code can be simplified at all.

pditommaso:
I haven't looked at the code, but it's a nice idea. How the JSON is going to be split? I mean, the JSON can have an arbitrary data structure, how the operator decides to apply the split?

lindenb:
@pditommaso Thanks. For now I suggest than the root must be an array and all the items of the array will be emitted, or it can be an object and all the key/value will be emitted. 

But one could imagine to add a kind of path as an option to only scan a subpart of the json. Eg: `splitJson("path":"/api/response/body/variants")`

pditommaso:
> For now I suggest than the root must be an array

Fair enough for prototyping, but it would be nice to have a generalisation for final version. 

Jmespath could a nice way to express the path, see https://jmespath.org/

lindenb:
@bentsherman @pditommaso back to you for a first review :-)

The indentation should be fixed now (  I'm working with eclipse. Is there an eclipse configuration for the NF formatting somewhere ? ) and I updated the doc.

As I'm a java programmer, I'm sure you'll find extra semicolons everywhere :-P

Again, I'm not sure for the statements about `validOptions` and  `collector`. 

I implemented a basic parameter `jsonPath` like `key1.key2[2].key3[10]` that specifies which part of the Json document should be scanned.

I also wrote a few tests.

Thanks !

bentsherman:
Thanks, I was able to make your code more Groovy. I also removed the `decompress` and `charset` options because they are only used when consuming files, which splitJson does not.

By the way, I think the `CONTRIBUTING.md` has some guidelines for code style that you can put into your IDE settings.

bentsherman (approved):

modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy
```diff
@@ -0,0 +1,277 @@
...
+            final jsonreader = new JsonReader(reader)
+            result = scanJsonForPath(jsonreader, "", this.jsonPath ?: "")
+
+            jsonreader.close()
+            if( result != null ) {
+                // make sure to process collected entries
+                if( collector && collector.hasChunk() ) {
+                    result = invokeEachClosure(closure, collector.nextChunk())
+                }
+            }
```
pditommaso, line 97: Can be simplified to 

```suggestion
            if( result && collector?.hasChunk() ) {
                result = invokeEachClosure(closure, collector.nextChunk())
            }
```

pditommaso (changes requested):
It looks excellent. Made some proposal

modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy
```diff
@@ -0,0 +1,287 @@
...
+
+    /**
+     * @return A map representing the valid options for the splitter. The map keys define the
+     * accepted parameter names, the values the valid values for each of them.
+     */
+    @Override
+    protected Map<String,?> validOptions() {
+        def baseOptions = super.validOptions()
+        return [
+            jsonPath: String,
```
pditommaso, line 73:
```suggestion
            path: String,
```

Using `path` is a bit more concise and readable e.g. `something.splitJson(path:'x.y.z')`

pditommaso:
Ok, this can be merged but it requires the commit to be signed. @lindenb can you please give it a try? see [here](https://github.com/nextflow-io/nextflow/blob/master/CONTRIBUTING.md#contributing-code-changes) for details

pditommaso:
Looks so, thanks!

PR #3831, opened 2023-04-04
Add an option types to splitCsv() to cast the variable types

### LINKED ISSUE #3763: splitCsv returns all values as strings
Hello,

I am trying to convert the content of a CSV file into a JSON array. When I read the CSV with `splitCSV()` I see that all values are casted as strings.

Is that an expected behaviour? I would expect numbers to be casted as Integers or true/false as booleans, for example. 
And in case it's the expected behaviour, is there a way to automatically cast all these objects to their primitive types?

Thank you!

To reproduce what I see, I have a test.csv:
```
foo,bar
a,1
b,2
```
And this Nextflow code:
```nextflow
import groovy.json.JsonBuilder

def testJSON() {

    fileContent = file("test.csv").splitCsv( header:true, sep:',')

    for (row in fileContent) {
        println("row class: $row.getClass, row: $row")
        for (key in row.keySet()) {
            println("value class: ${row[key].getClass()}, value: ${row[key]}")
        }
    }

    def jsonObj = new JsonBuilder(fileContent)

    println("file content: $fileContent")
    println("json object: $jsonObj")
}

testJSON()
```

mirpedrol:
Close #3763

This adds a new option `types` to splitCsv(). 
splitCsv() returns all values as strings by default. With types=true it will automatically infer the value types, otherwise you can also pass an array of types to define the type of each csv column.


bentsherman:
I love it. Can you document these options in `operator.rst` ? Just need to add the new option to the options table, but you're welcome to add a blurb with an example if you want.

modules/nextflow/src/test/groovy/nextflow/splitter/CsvSplitterTest.groovy
```diff
@@ -150,6 +159,42 @@ class CsvSplitterTest extends Specification {
...
+
+        when:
+        def items = new CsvSplitter().target(typesText).options(header:true,types:true).list()
+        then:
+        items.size() == 4
+
+        items[0].string == 'gamma' && items[0].string instanceof String
+        items[0].integer == 1 && items[0].integer instanceof Integer
+        items[0].boolean == true && items[0].boolean instanceof Boolean
+        items[0].float == 1.1.toDouble() && items[0].float instanceof Double
```
pditommaso, line 172: Not sure, there's a need to implement this. Groovy provides out of the box helper to convert to any type using `toInteger()`, `toLong()`, `toBoolean` etc

bentsherman: Not sure what you mean? The PR is using those methods to do the type casting.

pditommaso: why not using buildit `toString()`, `toInteger()`, etc

pditommaso: I understand that, but I do not understand the above snippet does

mirpedrol: Hello :) I am using those to convert the types [here](https://github.com/nextflow-io/nextflow/pull/3831/files/8e076089ce2d8a01cfe075ae33f39da620fca254#diff-32144a1a47a0ea418938e70e52a8fe650515c61d94b3d31fc421067434430ac5R234). Is that what you are referring to?

bentsherman (approved):

bentsherman:
Your last two commits weren't signed off. It should be `git commit -s`, not `-S`. You can click "Details" on the DCO check and follow the instructions there. Basically undo and redo the last two commits with sign-off and force push.

mirpedrol:
All tests are passing now if you want to merge this PR :) Thanks!

modules/nextflow/src/main/groovy/nextflow/splitter/CsvSplitter.groovy
```diff
@@ -91,6 +98,16 @@ class CsvSplitter extends AbstractTextSplitter {
         if( options.skip )
             skipLines = options.skip as int
 
+        // cast variables to their type
+        if( options.types ) {
+            if( options.types == true )
+                typesAuto = true
+            else if( options.types instanceof List )
+                columnTypes = options.types as List
```
pditommaso, line 106: It would be nice to validate the types list provided

pditommaso (changes requested):

modules/nextflow/src/main/groovy/nextflow/splitter/CsvSplitter.groovy
```diff
@@ -167,13 +185,65 @@ class CsvSplitter extends AbstractTextSplitter {
...
 
+        if( columnTypes && columnTypes.size() != tokens.size() )
+            throw new IllegalArgumentException("The number of column types should match the number of csv columns. Provided types: ${columnTypes}")
+
         def map = [:]
-        for( int i=0; i<columnsHeader.size(); i++ )
-            map[ columnsHeader[i] ] = i<tokens.size() ? tokens[i] : null
+        for( int i = 0; i < tokens.size(); i++ ) {
+            if( typesAuto )
+                map[columnsHeader[i]] = inferValueType(tokens[i])
```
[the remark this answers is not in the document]
mirpedrol: I would like to use this functionality in the nf-validation pluging, where I validate a samplesheet csv file automatically, so the user won't be able to provide the specific types, which makes it difficult to validate the file without the automatic detection of variable types.

If maintaining the automatic type detection is not feasible, implementing it directly in the nf-validation plugin could be the best solution. Do you think this would be the best approach?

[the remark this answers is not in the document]
ewels: But don't we have an explicit schema for the sample sheet? So we should know the types?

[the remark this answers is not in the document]
bentsherman: Paolo, I think we might as well allow the auto type inference and say "use at your own risk". Users can always fall back to explicit types.

It's a common feature in CSV parsers, for example see [pandas.read_csv](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html). So if you're really concerned about it, we could even see what these established implementations are doing and try to imitate it.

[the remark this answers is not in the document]
mirpedrol: > But don't we have an explicit schema for the sample sheet? So we should know the types?

Right, I didn't think about that. We currently only use the schema after parsing the CSV, but we could obtain the types before. The only thing I'm not sure about is the order of properties in the schema, but I can probably solve this with the column names. I'll try it out. Thanks!

modules/nextflow/src/main/groovy/nextflow/splitter/CsvSplitter.groovy
```diff
@@ -107,6 +124,7 @@ class CsvSplitter extends AbstractTextSplitter {
         result.header = [ Boolean, List ]
         result.quote = String
         result.skip = Integer
+        result.types = [ Boolean, List ]
```
bentsherman, line 127: @mirpedrol related to the other thread, I totally forgot about this -- the `types` option should also be allowed to be a map, so that you can assign types by column name. It should be allowed only if a header row is expected.

modules/nextflow/src/main/groovy/nextflow/splitter/CsvSplitter.groovy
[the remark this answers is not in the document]
pditommaso: > Paolo, I think we might as well allow the auto type inference and say "use at your own risk". Users can always fall back to explicit types

the reality people don't care and just use the quickest way, when you will discovered it will be too late

modules/nextflow/src/main/groovy/nextflow/splitter/CsvSplitter.groovy
pditommaso: Not sure about this. IMO it will be more verbose. The explicit list of types sounds better

bentsherman: We should support both. Again, see `pandas.read_csv` for a good example of this

mirpedrol:
@bentsherman @pditommaso I have added the option to provide the types with a map of column headers and removed the option of inferring types automatically. This should work for the sample sheet validation use case and avoid the problems that you were commenting. I hope it's ok :)

mirpedrol:
I have made a final commit that handles errors with type casting. The default behaviour is to return a string.

On another note, I noticed that some validation errors are now being ignored when I used this type casting in the validation plugin with a CSV file. For example, if the CSV file contains an integer, and the expected type in the JSON Schema is a boolean, this integer will be converted to a boolean, and the validation won't fail.

I was wondering if it would be appropriate to override the `castValueType` function in the validation plugin to use automatic type detection or perform more checks before casting the values. Would this be acceptable, or is there a better way to handle this?

mirpedrol:
After some discussion with @bentsherman I will try to implement a `strict` mode to convert the variables only if some check pass (such as a boolean being 'true' or 'false', etc). I will convert this PR to a draft again 🙂  Sorry for the inconvenience!

mirpedrol:
I have implemented the `typesStrict` mode. This option checks if a value is of the correct data type; if it's not, it defaults to a string to prevent loss of information, such as when converting integers or strings to booleans.

bentsherman:
I changed the `typesStrict` option to `strict` as boolean, so you use the `types` option for both cases. Also just made the code more Groovy 😄 

@pditommaso ready for review

pditommaso (changes requested):
Don't want to give bad news, but I've already commented I'm not in favour of merging automatic type detection and map Map-based type definition

bentsherman:
Julia already removed the automatic type detection, but what's your issue with the map?

pditommaso:
To have only a few options. What's the use case of having both List and Map? 

Also, the strict mode could look more useful. If I've read the code correctly in both cases it will result a warning.

mirpedrol:
As I see it, Map is useful when we have column names, then the order of types is not important. List will be used if there isn't a header, but the order of the provided list is important.
Regarding the strict mode, I also find it more useful. The difference is that without the strict mode some values will be casted without a warning. The example that I found is having a number string casted to a boolean which returned "false". So having '100' with strict mode will raire a warning and return a string '100', without the strict mode won't complain and return false.

pditommaso:
> As I see it, Map is useful when we have column names, then the order of types is not important. List will be used if there isn't a header, but the order of the provided list is important.

Fair enough 

> The difference is that without the strict mode some values will be casted without a warning.

Maybe I'm wrong, but from what i'm seeing in the code [castValueType](https://github.com/nextflow-io/nextflow/pull/3831/files#diff-32144a1a47a0ea418938e70e52a8fe650515c61d94b3d31fc421067434430ac5R234), if there a failure trying to convert a string to a type, let's say `long` it with cought in the exception block reporting a warning. 

Instead when using [castValueTypeStrict](https://github.com/nextflow-io/nextflow/pull/3831/files#diff-32144a1a47a0ea418938e70e52a8fe650515c61d94b3d31fc421067434430ac5R265), the extra condition will cause to reach the last `else` , reporting again a warning 🤷‍♂️


Above, if the user declared a specific type falling back to a string and reporting a warning should be avoid, because that will just introduce a bug in your pipeline data that you will only discover when it's too late

mirpedrol:
I see, you mean to throw an error in both cases if we catch the exception and remove the [last else](https://github.com/nextflow-io/nextflow/pull/3831/files#diff-32144a1a47a0ea418938e70e52a8fe650515c61d94b3d31fc421067434430ac5R277-R280) only for the strict conversion, did I understand correctly?

bentsherman:
Julia and I had a separate conversation about the strict mode. Basically the only difference is that strict mode only casts the string to a type if it is "strictly" that type.

For example, without strict mode, you could cast `'20'` to boolean as `true`. With strict mode, it won't do the cast and instead just leave it as `'20'`.

But for Julia it was important to log warnings instead of throwing an error, so that the user receives a list of all the issues that they can fix in one go. If you throw an error, then you can only fix one issue at a time and you have to re-launch Nextflow each time.

pditommaso:
If we go toward this approach what's preventing that a wrong type definition against a 1 mln lines CSV file produces 1 mln warning line trashing time and money? 

I believe it's user role to provide a valid CSV content.

pditommaso:
Alternatively this can be easily implemented using a `slitCsv` chained with `map` and a custom function

mirpedrol:
> If we go toward this approach what's preventing that a wrong type definition against a 1 mln lines CSV file produces 1 mln warning line trashing time and money?

Would removing the warning be an option?

bentsherman:
@mirpedrol , what do you think about implementing the type casting as a custom function, like Paolo suggested? It would take a single CSV record that has already been parsed but not type cast, and you would apply it with a map operator like so:

```groovy
// nf-validation plugin
@nextflow.plugin.extension.Function
Map castToType(Map record, Map types) {
  // ...
}

// main.nf
include { castToType } from 'plugin/nf-validation'

Channel.fromPath('input.csv')
  | splitCsv
  | map { record -> castToType(record, type: [ ... ]) }
```

That way you can evolve it however you want within the nf-validation plugin instead of Nextflow.

I kinda agree with Paolo on this one. The appropriate error-handling strategy is different depending on whether you are in development or production, it would be difficult to force the splitCsv operator to fit both, when we could just have a custom function.

bentsherman:
I see, you are using `splitCsv` on a path. So yes, you can just implement the type casting logic in your plugin code:

```groovy
types = variableTypes(schema_name, baseDir)
fileContent = file_path .splitCsv( /* ... */ ) .collect( record -> castToType(record, types) )
```

You can expose that function if you want, or not if you want to keep your plugin API simple.

mirpedrol:
Ok, this looks more simple than the overriding that I tried at the beginning. Thank you very much for your help! 😄 

Would you like me to change warnings in this PR to errors and still use the type conversion? Or should I close the PR?

bentsherman:
If you implement the casting in a custom function in your plugin then we don't need this PR, you can close it. As for the warnings / errors, you can do whatever you want because it's in your plugin!

PR #3833, opened 2023-04-04
Reorder plugins-info.txt to match build order

bentsherman:
I don't know why, but `make compile` always reorders the list of plugins in `plugins-info.txt` and I always have to revert it. So I just reordered this file to match the way that `make compile` orders it.


pditommaso:
Because it's automatically generated.

bentsherman:
Does it have a deterministic order? It's not alphabetical but it seems to be consistent over time

pditommaso:
i've added a sort to the gradle snippet creating it 👉  d5552ee6

PR #3834, opened 2023-04-04
Convert large tables to definition lists

bentsherman:
I'm experimenting with something in the Markdown docs and wanted to run it by the two of you.

There are all of these huge tables, mostly on the Configuration page. They are hard to use because each option is packed onto a single line and you have to scroll horizontally to read a long option.

I converted some of these tables to definition lists and I think they are much better. Each option description is just a paragraph, so no more scrolling. I also moved version notes to separate lines so that they stand out. I think this change could set us up nicely in the future to expand on options that need it and just be more organized overall.

If you guys like what you see, I'll add these changes to the Markdown PR.

### DIFF
--- a/docs/aws.md
+++ b/docs/aws.md
@@ -14,9 +14,7 @@ You can provide your AWS access keys using the standard AWS variables shown belo
 
 If `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are not defined in the environment, Nextflow will attempt to retrieve credentials from your `~/.aws/credentials` or `~/.aws/config` files. The `default` profile can be overridden via the environmental variable `AWS_PROFILE` (or `AWS_DEFAULT_PROFILE`).
 
-Alternatively AWS credentials can be specified in the Nextflow configuration file.
-
-See {ref}`AWS configuration<config-aws>` for more details.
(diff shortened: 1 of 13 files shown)


bentsherman:
Here's a sample of what it looks like: 
image

ewels:
I agree that the paragraph text is easier to read, however to play devils advocate, I find the config names _slightly_ harder to scan..

For reference, the current view is like this:

image


An alternative way to make these tables more readable could be to remove the maximum page width and allow line-wrapping within tables:


image

Though worth noting that this isn't sufficient on mobile, still tough to read even with the line wrapping. Definition lists would definitely be more readable here.

image

ewels:
The more I look at it, the more I like your approach @bentsherman 👍🏻 

Specifically:

* Scope for consistent rich formatting about default values, minimum version warnings
* Gives room for more extended description of an attribute, if needed
* In your specific example - providing the full config scope for each option (`aws.batch.cliPath`, not just `cliPath`)

I'd be tempted to list the default after the parameter name in each case, to bring it out of the prose description for faster reference.

bentsherman:
Here's another example: the `manifest.nextflowVersion` setting had some extra prose that was included outside of the table. Now we can put it in the setting description: 
image

ewels (approved):

bentsherman:
Here's an example with the groupTuple operator, showing list of options. Possible values for `sort` are displayed inline as a bulleted list: 

image

Issue #2683, opened 2022-02-28
Azure, WARN: Failed to cleanup work dir


des2037:
Hi NextFlow devs/community,

I'm running into an error when I try to clean my working directories (running on Microsoft Azure). Specifically, in my `nextflow.config`, I have (as the last line)

```cleanup = true```

The error I receive is:
```
WARN: Unable to start plugin 'nf-azure' required by az://<my_container_name>/work/2e/8cd...
WARN: Failed to cleanup work dir: /my_container_name/work
```

Here is my log for the run:
[nextflow.log](https://github.com/nextflow-io/nextflow/files/8156729/nextflow.log)

WIth the exception of not cleaning the working directory, the job runs as expected (files are read from and created in my Azure blob containers, so I don't `think` blob permissions the problem).

Any help with this error is greatly appreciated! Please let me know if I can provide any additional information that might be helpful.

Best,
Derek

**Comment by abhi18av** [comment]
Hi @des2037 , 

Thanks for letting us know about this issue. I noticed that the `nf-azure` plugin version is `0.11.2`

Could you please try again with the latest plugin? 

You'd need to update your `nextflow.config` with the following snippet (from [docs](https://www.nextflow.io/docs/latest/plugins.html#configuration))

```nextflow
plugins {
  id 'nf-azure@0.12.0'
}

```


If it fails again, please share the `.nextflow.log` file with us.

des2037:
Hi @abhi18av 

Thanks very much for the fast reply and suggestion. 

I've updated the `nf-azure` plugin to `0.12.0`, but sadly I'm still getting the same error.

Here is my nextflow.log:
[nextflow.log](https://github.com/nextflow-io/nextflow/files/8161920/nextflow.log)


Thanks for the help!

abhi18av:
Thanks @des2037 , I'll look into this soon.

pditommaso:
The problem here is that the cleanup happens after the Azure plugin shutdown :/

```
Feb-28 16:04:09.986 [main] DEBUG nextflow.plugin.BasePlugin - Plugin stopped nf-azure
Feb-28 16:04:10.052 [main] DEBUG nextflow.plugin.PluginsFacade - Using Default plugins manager
Feb-28 16:04:10.054 [main] INFO  org.pf4j.DefaultPluginStatusProvider - Enabled plugins: []
Feb-28 16:04:10.054 [main] INFO  org.pf4j.DefaultPluginStatusProvider - Disabled plugins: []
[...]

PR #3836, opened 2023-04-05
Fix cleanup with remote work dir

### LINKED ISSUE #3645: Clean up option fail when using AWS Batch (and likely other clouds)
## Bug report 

Nextflow fails to perform the pipeline work directory [cleanup](https://www.nextflow.io/docs/latest/config.html?highlight=cleanup#miscellaneous) when using AWS Batch and other cloud providers 

The following error is reported 

```
Feb-14 18:55:04.981 [main] WARN  nextflow.file.FileHelper - Unable to start plugin 'nf-amazon' required by s3://nextflow-ci/work/79/17395af5457010376510e5808a2881
java.lang.NullPointerException: Cannot invoke "nextflow.plugin.CustomPluginManager.getPlugin(String)" because "this.manager" is null
	at nextflow.plugin.PluginsFacade.isStarted(PluginsFacade.groovy:338)
	at nextflow.plugin.PluginsFacade.startIfMissing(PluginsFacade.groovy:432)
	at nextflow.plugin.Plugins.startIfMissing(Plugins.groovy:91)
	at nextflow.file.FileHelper.autoStartMissingPlugin(FileHelper.groovy:288)
	at nextflow.file.FileHelper.asPath(FileHelper.groovy:268)

```

Complete log file is attached 

[nextflow.log](https://github.com/nextflow-io/nextflow/files/10735602/nextflow.log)

bentsherman:
Closes #3645 and #2683 

Fixes an issue with cleanup on a remote work directory, which is caused by the plugin manager being shutdown before the cleanup.

I moved the cleanup code to happen after publishing is done but before the plugin manager and cache db are shutdown. However, I haven't been able to verify that the cleanup works. When I run the `hello` pipeline on AWS Batch with `-trace nextflow`, I don't get the error message anymore, but also the task directories aren't deleted.

Currently I have this code:
```groovy
            log.trace "Cleaning-up workdir"
            cache.eachRecord { HashCode hash, TraceRecord record ->
                log.trace "Cleaning task: ${hash.toString()}"

                def deleted = cache.removeTaskEntry(hash)
                if( deleted ) {
                    // delete folder
                    FileHelper.deletePath(FileHelper.asPath(record.workDir))
                }
            }
            log.trace "Clean workdir complete"
```

The log shows the outer trace messages, but not the inner message for each task. It's like the task records aren't there during the iteration.

The only difference with the old code is that the cleanup now uses the session's cache db reference rather than opening the cache itself. That was necessary because at that point the session's cache had been closed. Also, the old version used openForRead() instead of open(). But that shouldn't matter right?

@pditommaso what do you suggest?

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -655,12 +655,12 @@ class Session implements ISession {
             execService = null
             log.trace "Session > executor shutdown"
 
-            // delete task directories if enabled and run was successful
-            if( config.cleanup && workDir && !(aborted || cancelled || error) )
-                cleanup0()
-
             // -- close db
             cache?.close()
+
+            // clean work directory if enabled and run was successful
+            if( config.cleanup && workDir && !(aborted || cancelled || error) )
+                cleanup0()
 
             // -- shutdown plugins
             Plugins.stop()
@@ -711,13 +711,15 @@ class Session implements ISession {
     }
 
     /**
-     * Delete the workflow work directory from tasks temporary files
+     * Delete the work directories of completed tasks
      */
     final protected void cleanup0() {
+        CacheDB db = null
         try {
-            log.trace "Cleaning-up workdir"
-            cache.eachRecord { HashCode hash, TraceRecord record ->
-                def deleted = cache.removeTaskEntry(hash)
+            log.trace "Cleaning up workdir"
+            db = CacheFactory.create(uniqueId, runName).openForRead()
+            db.eachRecord { HashCode hash, TraceRecord record ->
+                def deleted = db.removeTaskEntry(hash)
                 if( deleted ) {
                     // delete folder
                     FileHelper.deletePath(FileHelper.asPath(record.workDir))
@@ -727,6 +729,9 @@ class Session implements ISession {
         }
         catch( Exception e ) {
             log.warn("Failed to cleanup work dir: ${workDir.toUriString()}")
+        }
+        finally {
+            db.close()
         }
     }
 
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -638,18 +638,26 @@ class Session implements ISession {
     void destroy() {
         try {
             log.trace "Session > destroying"
+
             // shutdown publish dir executor
             publishPoolManager.shutdown(aborted)
+
             // invoke shutdown callbacks
             shutdown0()
-            log.trace "Session > after cleanup"
+            log.trace "Session > after shutdown callbacks"
+
             // shutdown executors
             executorFactory?.shutdown()
             executorFactory = null
+
             // shutdown executor service
             execService?.shutdown()
             execService = null
             log.trace "Session > executor shutdown"
+
+            // delete task directories if enabled and run was successful
+            if( config.cleanup && workDir && !(aborted || cancelled || error) )
+                cleanup0()
 
             // -- close db
             cache?.close()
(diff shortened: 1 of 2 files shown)


bentsherman:
I think the issue is that closing the cache db also waits for it to finish writing anything. So I will try closing and re-opening the cache like before.

bentsherman:
That did the trick. Cleanup works!

So now the Session destroy is as follows:
- shutdown publishing
- shutdown executors
- save, close cache db
- clean work dir if enabled (it will re-open cache db)
- shutdown plugin manager

@pditommaso all yours to review and merge

bentsherman:
@pditommaso bump

bentsherman:
@pditommaso here are the recent commits I found related to the shutdown logic:
1. Fix thread pool race condition on shutdown: 8d2b058
2. Fix issue with empty report file: 9cc4f07

The first commit moved the file transfer shutdown from the start of session destroy to after it. This change caused issues when publishing to S3, because the plugin manager was being shutdown before waiting for all the files to be published.

The second commit fixed this problem by creating separate thread pools for publishing files and the Tower archiver.

Basically, I think the introduction of the Tower archiver created a circular dependency between waiting for file transfers and shutting down the plugin system, and it was resolved by giving the Tower archiver its own file transfer pool.

This PR does not change any of the file transfer shutdown. It only moves the work dir cleanup to before the plugin manager, which is necessary to make it work with cloud storage. This change also puts it before deleting the script classes dir and updating the history file, neither of which are related to the cache DB.

I feel confident that we can merge this change. A good stress test would be to run nf-core rnaseq test on Tower with archiving and cleanup enabled, that should test all of the shutdown logic. Why don't we merge it, run the test, and if we find an error then we'll fix it?

pditommaso:
Not inclined to merge this. The cleanup has been implemented as an experimental feature. when used with object storage with large real-world pipelines can hang the execution for a long time and result in extra charges. Don't think we should promote this behaviour.

bentsherman:
Fair enough. It will be replaced by automatic task cleanup anyway.

Shians:
I would like this to be reconsidered. I am using AWS Batch with Fargate and S3 to process some large VCF files into a smaller size that I pull down for local use. The storage costs for the files in the staging areas far exceed the cost of waiting for them to be cleaned up. I think it should be left to the users whether they want to incur additional execution time costs to reduce storage costs.

CharlesARoy:
@Shians I'm in the same boat and would love to see this functionality implemented (as would many other people from what I can see), but I don't think it's likely in the near future.

If you use one main folder in S3 as the Nextflow work directory, you could create a Lifecycle Rule for the bucket (on the AWS side) that automatically deletes files with a given prefix after some period of time (minimum is 1 day). This would even work if you use a versioned S3 bucket (which probably adds another layer of complexity for a Nextflow-based implementation).

In my case, I create a Nextflow work folder in each project folder, so there isn't a single prefix I can use to limit the application of a lifecycle rule. Consequently, I'd love to be able to refer to Nextflow work files using object tags. It seems that Nextflow already automatically tags the metadata files with `nextflow.io/metadata = true`, e.g.:
Image on 2025-01-17 11 34 50 AM

However, (and this seems like an oversight), Nextflow fails to tag the non-metadata files (i.e. the process input files) it creates in the work folder. If this were fixed/updated, I’d recommend tagging the non-metadata files differently so that they can be handled distinctly from the metadata files by lifecycle rules.

@bentsherman or @pditommaso do you have any insight as to how Nextflow tags the S3 objects that it does and what code would need to be modified so that it would also tag the non-metadata files in the work folder?

pditommaso:
S3 tagging is supported via the [Fusion file system](https://www.nextflow.io/docs/latest/fusion.html). See fusion [tags here](https://www.nextflow.io/docs/latest/reference/config.html#fusion)

PR #3837, opened 2023-04-05
Document return type of operators

bentsherman:
Documents for each operator whether it returns a queue channel or value channel.


pditommaso (approved):
Well done

PR #3839, opened 2023-04-05
Use Google Tag Manager.

ewels:
Documentation: Load Google Analytics via GTM instead of gtag.js

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -9,3 +9,22 @@
         </a>
     </div>
 {% endblock %}
+
+{% block extrahead %}
+    {{ super() }}
+    <!-- Google Tag Manager -->
+    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+    'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+    })(window,document,'script','dataLayer','GTM-TNCXSWG');</script>
+    <!-- End Google Tag Manager -->
+{% endblock %}
+
+{% block footer %}
+    {{ super() }}
+    <!-- Google Tag Manager (noscript) -->
+    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-TNCXSWG"
+    height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+    <!-- End Google Tag Manager (noscript) -->
+{% endblock %}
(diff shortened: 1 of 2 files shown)


pditommaso (approved):

ewels:
Ah, slight problem with https://github.com/nextflow-io/nextflow/pull/3840 - these two CI jobs are _required_ in the repo settings and now they don't run for PRs purely on docs. That means that GitHub just sits and waits for them forever, and merging is blocked:

image

ewels:
I think that there are two ways around this:

* Remove the required status of these tests in the branch protection
* Rework the actions workflow so that it always runs, but that the `job` / `steps` are skipped if it's only the docs that were edited
    * GitHub treats skipped tests as passes for required CI jobs
    * A fair bit more code needed in the workflow YAML, not as clean

pditommaso:
But still I can bypass the checks 😄

pditommaso:
I've seen this also in other PRs e.g. [this](https://github.com/nextflow-io/nextflow/pull/3681), never fully understood why

ewels:
Yeah, you can always admin-force merge.. But it's a bit messy 😉 

I foun another way to do it in these [GitHub docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks) - create a second workflow with the same name, that runs in the opposite scenario. Feels a bit hacky but might be the nicest solution. I'll create a PR..

pditommaso:
you are right!

ewels:
> I've seen this also in other PRs e.g. [this](https://github.com/nextflow-io/nextflow/pull/3681), never fully understood why

This one is something different I think. It looks like the CI build is triggering, but being skipped due to this config in the workflow:

```yaml
if: "github.event == 'push' || github.repository != github.event.pull_request.head.repo.full_name"
```

That's saying that pull-request CI builds should only run if the PR comes from a fork, not from the same repo. The PR you linked to came from a branch on the main repo, hence it was skipped.

Any idea why the workflow is configured this way?

ewels:
> Any idea why the workflow is configured this way?

Ah maybe it's because the CI jobs are running on the `push` already, and so it's assumed that the same CI on the pull-request event is redundant? But the same could be said for forks 🤔 

If that's the motivation, I think it probably makes more sense to only run CI tests for `push` on the `master` branch. Then have `pull_request` CI jobs always run.

ewels:
> Yes, to prevent double triggering the CI for internal PRs

But we're double triggering the CI for external PRs, I don't really see how the two are different..

Also, the two sets of CI tests are slightly different. The `pull_request` one runs the CI on the merged code, `push` only on the base code. So if the PR is coming from a fork that's a long way behind it is more representative of what will happen post-merge.

ewels:
I made https://github.com/nextflow-io/nextflow/pull/3842 to explain what I mean..

PR #3840, opened 2023-04-06
Skip build CI for pull-requests too

ewels:
The build CI tests are already skipped when only editing `docs/**` for the `push` event, but not for the `pull_request` event. This means that they run on every PR even if it's only to the docs.

This adds the `paths-ignore` config to the `pull_request` event, the same as `push`.


pditommaso:
Awesome

ewels:
Follow-up PR: https://github.com/nextflow-io/nextflow/pull/3841

PR #3747, opened 2023-03-13
Add error message for missing container image with Google Batch

bentsherman:
Closes #3746

PR #3842, opened 2023-04-06
Propsal: Different CI trigger logic

ewels:
See discussion in https://github.com/nextflow-io/nextflow/pull/3839#issuecomment-1498736576

This PR changes the CI logic so that jobs only run on `push` to the `master` branch - ie. after PRs are merged, or on direct push to master.

The logic within the build steps is then simplified, so that jobs _always_ run on the `pull_request` event.

This should still mean that CI isn't double-triggered, but makes things a bit simpler and the logic a bit clearer when PRs are coming from forks / internal branches.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,6 +11,7 @@ on:
     types: [opened, reopened, synchronize]
     paths-ignore:
       - 'docs/**'
+  workflow_dispatch:
 
 jobs:
   build:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,10 +4,7 @@ name: Nextflow CI
 on:
   push:
     branches:
-      - '*'
-      - '!refs/tags/.*'
-    tags-ignore:
-      - '*'
+      - 'master'
     paths-ignore:
       - 'docs/**'
   pull_request:
@@ -17,7 +14,6 @@ on:
 
 jobs:
   build:
-    if: "github.event == 'push' || github.repository != github.event.pull_request.head.repo.full_name"
     name: Build
     runs-on: ubuntu-latest
     timeout-minutes: 90


.github/workflows/build.yml
```diff
@@ -4,10 +4,7 @@ name: Nextflow CI
 on:
   push:
     branches:
-      - '*'
-      - '!refs/tags/.*'
-    tags-ignore:
-      - '*'
+      - 'master'
```
pditommaso, line 7: umm. but this will disable tests on branches other than master

bentsherman: I think it disables tests on pushes to other branches, but pull requests will still be tested, see the `pull_request` below 👇

pditommaso: But it's a wrong assumption! it's also useful to push a branch and have tests run even without a PR

bentsherman: Just make a PR

bentsherman: or, what if we add a manual trigger? that way you can test branches only when you want to. also could be a workaround for when PR tests mysteriously don't run

ewels: Added 👍🏻

ewels: If the change doesn't match your workflow (you often merge branches manually / you like seeing CI on branch pushes during development) then we can change or just close. I opened the PR as a draft initially just to propose the suggestion.

pditommaso: Would the CI be trigged when pushing new commit to a PR (both internal or external) ?

ewels: If the PR is already open, then yes in both cases 👍🏻

pditommaso:
```suggestion
      - 'master'
      - 'testing*'
      - 'dev*'
```

Ok, are patterns in the branch names supported? If so it could be a nice compromise

pditommaso: Let's try it!

ewels: @pditommaso you didn't commit the suggestion before merging! 😅

ewels: Added here: https://github.com/nextflow-io/nextflow/pull/3852

pditommaso:
aaaaaand it trigger twice on [this](https://github.com/nextflow-io/nextflow/pull/3498) PR! 

Screenshot 2023-04-12 at 08 49 22

pditommaso:
Oh, the first is the docs build. Should not only start when there are changes in `docs/` ?

ewels:
Hmm, the docs build shouldn't be running. That only runs if there are _only_ changes to the docs - if there are modifications in any other files then the regular CI runs as it always has done. But I agree that I can see it there. Maybe it was triggered when there was a commit pushed with only docs or something? Need to look into this more.. (especially as it's failing 🙈 ). @bentsherman any ideas why this was triggered? 🤔  

Anyway, generally - the CI triggered twice on that PR because you pushed twice. Every time you push a new commit to a PR, the `pull_event` trigger fires and new set of CI runs. If you want to reduce _this_ duplication then that's a different thing. That's job concurrency and we need to set `cancel-in-progress: true` (see [docs](https://docs.github.com/en/actions/using-jobs/using-concurrency)).

The PR you linked to is working correctly, as intended. If you look at the [checks running](https://github.com/nextflow-io/nextflow/pull/3498/checks), you can see that they're all listed as `pull_request`. If you look at an [older PR](https://github.com/nextflow-io/nextflow/pull/3747/checks) then then there are two lots of tests there for `push` and `pull_request` (though in this case, the `push` ones are skipped after being triggered).

Ok, not done yet.. I will start digging 🚧 👷🏻

ewels:
Ah ok, I messed up the logic sorry. From the [docs](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-filters-to-target-specific-paths-for-pull-request-or-push-events):

> If at least one path matches a pattern in the `paths` filter, the workflow runs.

> When all the path names match patterns in `paths-ignore`, the workflow will not run. If any path names do not match patterns in `paths-ignore`, even if some path names match the patterns, the workflow will run.

I was thinking that `paths` wouldn't trigger if there was anything else modified, but that's just the case for `paths-ignore`. Makes sense I guess.

I think that this is actually the desired behaviour though, to run docs CI when we make changes to the docs. To make sure it compiles? In the future (or now, if we want) we can convert this into a docs deployment preview for the PR.

I'll prep a PR to fix some minor niggles now, and to try to get the docs tests to actually pass. I'll see if I can change the name for the UI but in a way that the required tests thing still works. I think I can..

ewels:
Done: https://github.com/nextflow-io/nextflow/pull/3855

PR #3848, opened 2023-04-10
Fix `interval` channel factory

bentsherman:
Not really a high priority item, but I discovered this `Channel.interval` factory and found that it didn't work, it would only emit one item. I checked the [Timer](https://docs.oracle.com/javase/7/docs/api/java/util/Timer.html) API and found that the wrong method was being used, so I fixed it.

I assume no one is using it since it currently doesn't work, but I guess an interval function is a good component of any reactive library. If you want to keep it then I can document it as well.


pditommaso:
Well spotted

mribeirodantas:
We should document it, right?

bentsherman:
Yes

PR #3851, opened 2023-04-11
Use separate script for long stage/unstage commands

### LINKED ISSUE #2852: Slurm Sbatch file too large
## Bug report 

A more detailed description of the issue can be found here already https://github.com/nextflow-io/nextflow/issues/2118 and a preliminary bugfix in Nextflow 22.04.0 wasn't fully effective. The problem arises, when many files are symlinked to a directory by nextflow, thus bloating up the `.command.run` file to be larger than 4MB, which is the default max size allowed by SLURM. The scheduler then  complains about _"too long file names or file size too large"_ and doesn't start the process/job. 

Should be possible to run arbitrarily large file numbers. 

1) Run a job that creates a `.command.run` file > 5MB on any SLURM cluster
2) Watch the above error happen, the job doesn't get scheduled and Nextflow crashes.

Circumvention is possible by using some "hacks", e.g. making the process not produce as many symlinks (e.g. using folders instead of individual files), but that is not always an optimal approach and requires modifying the pipeline itself. One could think about externalizing this step in a separate file, so that the submitted `.command.run` is << 5MB in size and just calls the other externalized script maybe?

Circumvent the issue by running this here: https://github.com/nextflow-io/nextflow/issues/2118#issuecomment-990741937

bentsherman:
Closes #2852 

TODO:
- [x] e2e test


bentsherman:
CI tests were double triggered for push and pull_request, @ewels did we do something wrong?

ewels:
It's because your fork is out of date. I think if you hit "Update branch" it should work as expected... 🤞🏻

ewels:
Sorry, couldn't resist hitting the button to find out 😆

Looking good now though!

bentsherman:
Looking at #2118 , I believe this check was added as a partial solution: https://github.com/nextflow-io/nextflow/blob/6ef14c00c1d72ad5313d3c6ffa1ce2f9ab4f275d/modules/nextflow/src/main/groovy/nextflow/executor/SimpleFileCopyStrategy.groovy#L132-L133

So can we safely remove this check? Now it will be in the `.command.stage` script anyway.

modules/nextflow/src/main/resources/nextflow/executor/command-stage.txt
pditommaso, line 1: This template is not needed. It's enough to write the staging script to a file

modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
```diff
@@ -148,6 +152,10 @@ class BashWrapperBuilder {
         return "NXF_SCRATCH=\"\$(set +u; nxf_mktemp $scratchStr)\""
     }
 
+    protected boolean isStageFileRequired() {
+        inputFiles.size() >= 1000
```
pditommaso, line 156: Worth to parametrise this with an env var, and use it to create a unit test with a few number for files

pditommaso (changes requested):

pditommaso:
Made some improvements: 
- write files using newBufferedWriter to avoid allocation a new `byte[]` buffer that for large file can be expensive 
- use `bash` instead of `source` to read the stage command 
- use `NXF_WRAPPER_STAGE_FILE_THRESHOLD` as bytes threshold instead of num of files to better control it

modules/nextflow/src/main/groovy/nextflow/executor/SimpleFileCopyStrategy.groovy
```diff
@@ -120,19 +120,7 @@ class SimpleFileCopyStrategy implements ScriptFileCopyStrategy {
...
-            // for debugging purpose. However, this can cause the creation
-            // of a very big runner script when a large number of files is
-            // given due to the file name duplication. Therefore the rationale
-            // here is to keep the deletion only when a file input number is
-            // given (which is more likely during pipeline development) and
-            // drop in any case  when they are more than 100
-            if( len<100 )
-                delete << "rm -f ${Escape.path(stageName)}"
-
-            // link them
```
pditommaso, line 135: I'd keep this, to avoid stressing the storage with thousands of unneeded `rm`s

pditommaso: how this has been solved? 😅

tests/checks/stage-file.nf/.checks
pditommaso, line 1: Worth adding a second run with resume and verify it works as expected

pditommaso:
Considering the Slurm limit is 5MB (see [here](https://github.com/nextflow-io/nextflow/issues/2118)), i've creased the default to 1 MB

pditommaso:
Let's merge this. Solve #2852

PR #3857, opened 2023-04-12
Separate script parse from script run

bentsherman:
Closes #3854 

The ScriptParser has a runScript() method which parses the script and then runs it, which is a confusing interface. So I separated the calls to parse and run a script. This fixes the issue where a runtime error is reported as a parse error.

Before:
```console
$ nextflow run gh-3854/
N E X T F L O W  ~  version 23.04.0
Launching `gh-3854/main.nf` [amazing_allen] DSL2 - revision: e672254d14
ERROR ~ Unable to read script: 'gh-3854/./test.nf' -- cause: non-existing-file.txt

 -- Check script 'gh-3854/main.nf' at line: 1 or see '.nextflow.log' file for more details
```

After:
```console
$ ../launch.sh run gh-3854/
N E X T F L O W  ~  version 23.04.0
Launching `gh-3854/main.nf` [magical_meitner] DSL2 - revision: e672254d14
ERROR ~ No such file or directory: non-existing-file.txt

 -- Check script 'gh-3854/./test.nf' at line: 2 or see '.nextflow.log' file for more details
```

Unfortunately a side effect of these changes is that some of the plugin tests are timing out and I can't figure out why 😢

PR #3860, opened 2023-04-13
Separate script runtime errors from parse errors

### LINKED ISSUE #3854: `checkIfExists` gives wrong error when used in an included workflow

When `file("non-existing-file.txt", checkIfExists:true)` is placed inside a script which is included in another script, the following error is thrown (if the file doesn't exist):
```
ERROR ~ Unable to read script: '/home/nvnieuwk/Documents/nextflow/issues/checkexists/./test.nf' -- cause: /home/nvnieuwk/Documents/nextflow/issues/checkexists/non-existing-file.txt
```

while something like this is expected:
```
ERROR ~ No such file or directory: /home/nvnieuwk/Documents/nextflow/issues/checkexists/non-existing-file.txt
```

Following scripts need to be in the same directory:
1. main.nf
```
include { TEST } from './test.nf'

workflow {
    TEST()
}
```

2. test.nf
```
file("non-existing-file.txt", checkIfExists:true)

workflow TEST {
    Channel.fromPath(params.test_file).view()
}
```

Then execute this command to reproduce the problem (make sure `non-existing-file.txt` isn't inside that folder too):
```
nextflow run main.nf
```

### Environment 

* Nextflow version: 22.05.0 and 23.04.0 were tested
* Java version: openjdk 17.0.6 2023-01-17
* Operating system: Linux
* Bash version: 5.0.17

bentsherman:
Closes #3854 

Alternative to #3857 that just solves the reported issue without any extra improvements.


pditommaso:
Smart 👍 Just added `return this` to make more explicit the chaining

PR #3863, opened 2023-04-13
Improve task out redirect

pditommaso:
This PR simplifies the stdout and stderr redirection made in the task launcher script removing the need to use two FIFO pipe redirections.

### DIFF
--- a/modules/nextflow/src/main/resources/nextflow/executor/command-run.txt
+++ b/modules/nextflow/src/main/resources/nextflow/executor/command-run.txt
@@ -102,9 +102,6 @@ on_exit() {
     exit_status=${nxf_main_ret:=$?}
     printf $exit_status {{exit_file}}
     set +u
-    [[ "$tee1" ]] && kill $tee1 2>/dev/null
-    [[ "$tee2" ]] && kill $tee2 2>/dev/null
-    [[ "$ctmp" ]] && rm -rf $ctmp || true
     {{cleanup_cmd}}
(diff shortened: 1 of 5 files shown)


modules/nextflow/src/main/resources/nextflow/executor/command-run.txt
```diff
@@ -153,17 +150,9 @@ nxf_main() {
...
     set +e
-    ctmp=$(set +u; nxf_mktemp /dev/shm 2>/dev/null || nxf_mktemp $TMPDIR)
-    local cout=$ctmp/.command.out; mkfifo $cout
-    local cerr=$ctmp/.command.err; mkfifo $cerr
-    tee {{stdout_file}} < $cout &
-    tee1=$!
-    tee {{stderr_file}} < $cerr >&2 &
-    tee2=$!
-    ( nxf_launch ) >$cout 2>$cerr &
+    (set -o pipefail; (nxf_launch | tee {{stdout_file}}) 3>&1 1>&2 2>&3 | tee {{stderr_file}}) &
```
jordeu, line 153: Amazing pipe switch :1st_place_medal:

jordeu (approved):
I've tested with normal runs and it is working well. I've tested with the use case using Fusion where some processes get block and it seems to be more resilient to the problem but it still can happen (but that is related to a kill order problem I think). 

Overall I think it's a good change because it simplifies the script and it avoids needing to manage the processes kill.

PR #3868, opened 2023-04-14
Fix Azure pool creation timestamp

pditommaso:
This PR fixes the problem with the pool creation timestamp, that's caused by the increased Instant precision in the newer Java version resulting in a timestamp string not parsed by Azure


pditommaso:
Cool

PR #3869, opened 2023-04-14
Support `time` directive in Azure Batch

### LINKED ISSUE #3865: Add MaxWallClockTime to Azure batch pools
## New feature

When using Azure batch and the existing Azure Cloud advanced settings, there is no way to set the maximum wall time for a job to run. This means the default wall time is used, which can be insufficient when running large genomic datasets through various bioinformatics tools.

Would it be possible to extend the Azure Cloud advanced settings options to include a way to set the
`TaskConstraints.MaxWallClockTime Property`  for a pool.

https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.batch.taskconstraints.maxwallclocktime?view=azure-dotnet

## Usage scenario 

When analysing large genomic datasets with pipelines, some steps can take a long time which can exceed the maximum wall time for a task to run. Being able to increase this in the nextflow.config file for a pipeline would allow us to manage this limitation, where currently there isn't a way to do that with nextflow.

## Suggest implementation 

I'm not sure of the details of implementing it into nextflow, but for a user, I could see it available in the config file like this, where `MaxWallClockTime` is an advanced setting in the form `azure.batch.pools.<name>.MaxWallClockTime`

nextflow.config
```groovy
plugins {
  id 'nf-azure'
}

process {
  executor = 'azurebatch'
}

azure {
  registry {
    server = secrets.azure_registry_server
    userName = secrets.azure_registry_username
    password = secrets.azure_registry_password
  }
  batch {
    location = secrets.azure_batch_location
    accountName = secrets.azure_batch_accountName
    accountKey = secrets.azure_batch_accountKey
    allowPoolCreation = true
    pools {
      'default' {
        autoScale = true
        vmCount = 1
        maxVmCount = 4
        vmType = 'Standard_D8_v3'
        sku = 'batch.node.ubuntu 20.04'
        offer = 'ubuntu-server-container'
        publisher = 'microsoft-azure-batch'
        MaxWallClockTime = '48h'
        privileged = true
      }
      'big' {
        autoScale = true
        vmCount = 1
        maxVmCount = 4
        vmType = 'Standard_D32_v3'
        sku = 'batch.node.ubuntu 20.04'
        offer = 'ubuntu-server-container'
        publisher = 'microsoft-azure-batch'
        MaxWallClockTime = '120h'
        privileged = true
      }
    }
  }
```

Thanks
James

bentsherman:
Closes #3865 

The issue OP technically asked for a time constraint on pools, but AFAIK it is only possible for tasks, also that is the model that Nextflow already uses via the `time` directive.

### DIFF
--- a/docs/executor.rst
+++ b/docs/executor.rst
@@ -33,6 +33,8 @@ The pipeline can be launched either in a local computer, or an EC2 instance. EC2
 Resource requests and other job characteristics can be controlled via the following process directives:
 
 * :ref:`process-accelerator`
+* :ref:`process-container`
+* :ref:`process-containerOptions`
 * :ref:`process-cpus`
 * :ref:`process-memory`
 * :ref:`process-queue`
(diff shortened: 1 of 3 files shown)


pditommaso:
Nice! let's mention about it in the [Azure Batch](https://www.nextflow.io/docs/latest/executor.html#azure-batch) executor (see AWS batch for an example)

bentsherman:
Yes it is there

pditommaso:
All green!

PR #3875, opened 2023-04-17
Fix string comparison in S3 client

### LINKED ISSUE #3873: Invalid string comparing using equals operator
The following snipper uses a `!=` operator two compare two strings instead of using `.equals` (notice it's a Java code, not groovy) 

https://github.com/nextflow-io/nextflow/blob/a2f3bb24d7b027e617e6c2b28c4061cf39191bcf/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java#L586-L587

Also it's not clear the intent of the comparing because for example the expiration would be set of even for non glacier storage class such as `STANDARD`

bentsherman:
Closes #3873

### DIFF
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
@@ -583,7 +583,7 @@ public class S3Client {
 			RestoreObjectRequest request = new RestoreObjectRequest(bucketName, key);
 
 			String storageClass = client.getObjectMetadata(bucketName, key).getStorageClass();
-			if( storageClass != "INTELLIGENT_TIERING" )
+			if( !storageClass.equals("INTELLIGENT_TIERING") )
 				request.setExpirationInDays(glacierExpirationDays);
 
 			if( glacierRetrievalTier != null )
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
@@ -583,7 +583,7 @@ public class S3Client {
 			RestoreObjectRequest request = new RestoreObjectRequest(bucketName, key);
 
 			String storageClass = client.getObjectMetadata(bucketName, key).getStorageClass();
-			if( !storageClass.equals("INTELLIGENT_TIERING") )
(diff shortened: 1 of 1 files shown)


plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
```diff
@@ -583,7 +583,7 @@ protected void restoreFromGlacier(String bucketName, String key) {
 			RestoreObjectRequest request = new RestoreObjectRequest(bucketName, key);
 
 			String storageClass = client.getObjectMetadata(bucketName, key).getStorageClass();
-			if( storageClass != "INTELLIGENT_TIERING" )
+			if( !storageClass.equals("INTELLIGENT_TIERING") )
```
pditommaso, line 586: Cannot `storageClass` be null?

bentsherman: I don't think that's possible because it comes from the object metadata. I don't think an object can have an undefined storage class.

pditommaso: Better being defensive. a trick could be to reverse the comparison i.e `!"INTELLIGENT_TIERING".equals(storageClass)` tho not sure the logic makes sense

bentsherman:
```suggestion
			if( storageClass && !storageClass.equals("INTELLIGENT_TIERING") )
```

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
```diff
@@ -583,7 +583,7 @@ protected void restoreFromGlacier(String bucketName, String key) {
 			RestoreObjectRequest request = new RestoreObjectRequest(bucketName, key);
 
 			String storageClass = client.getObjectMetadata(bucketName, key).getStorageClass();
-			if( storageClass != "INTELLIGENT_TIERING" )
+			if( storageClass && !storageClass.equals("INTELLIGENT_TIERING") )
```
pditommaso, line 586:
```suggestion
			if( storageClass!=null && !storageClass.equals("INTELLIGENT_TIERING") )
```

pditommaso (approved):

PR #3877, opened 2023-04-17
Add reference in Google section in the docs to the nextflow's blogpost

mribeirodantas:
In other sections of the official documentation, such as the Amazon section, a blog post is mentioned as an additional source of content on that topic. With that in mind, it's also worth referring to the Google Cloud Batch blog post in the Nextflow blog in the Google section of the official documentation.

### DIFF
--- a/docs/google.rst
+++ b/docs/google.rst
@@ -193,7 +193,6 @@ the ``-work-dir`` command line options. For example::
 .. warning::
   The Google Storage path needs to contain at least sub-directory. Don't use only the bucket name e.g. ``gs://my-bucket``.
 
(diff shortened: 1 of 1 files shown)


bentsherman:
In general we prefer not to link to Nextflow blog posts from the Nextflow docs. I believe the example you referenced was for an Amazon blog post, which is fine. But if a Nextflow blog post has some pertinent information for the docs, then the blog post content should be incorporated directly into the docs, like what we did for the [caching and resuming](https://nextflow.io/docs/latest/cache-and-resume.html) page.

Issue #3878, opened 2023-04-18
add logic for normalizing snake_case pipeline params


ahvigil:
## New feature

Several naming conventions are floating around for how to pass params into a pipeline. Some of the nextflow built-in params use [`camelCase`](https://github.com/nextflow-io/nf-validation/issues/9), nf-core [suggests using `snake_case`](https://nf-co.re/docs/contributing/modules#naming-conventions), and more generally GNU [suggests `kebab-case` for long options](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html), a de facto standard for many POSIX tools.

It can be annoying as a developer working with many tools using many conventions to remember if something takes `--long-option` or `--longOption` or `--long_option`. The distinction between these is rarely ever functionally relevant except to create headaches for the user, and the mix of underscores and hyphens `-_` can be difficult to distinguish and prone to typos.

Nextflow already accepts `--longOption` and `--long-option` (as documented [here](https://www.nextflow.io/docs/latest/cli.html#pipeline-parameters)) but nf-core represents a fairly large community of nextflow content and its a shame that `--long_option` pipelines have to form a separate usage group. The proposal here is to make the existing behavior more consistent and inclusive by including `snake_case` arguments during param normalization.

## Usage scenario 

Running
```
nextflow run main.nf --param_name
```
should be equivalent to
```
nextflow run main.nf --param-name
```
which is already equivalent to
```
nextflow run main.nf --paramName
```

so that authors and users can use whatever convention they prefer based on context and personal preference, and developers can stop arguing over which one is right ;)

## Suggest implementation 

Nextflow already [normalizes camel and kebab cases](https://github.com/nextflow-io/nextflow/blob/17fcc67814b6e8a5566469fc2873408fa476d262/modules/nextflow/src/main/groovy/nextflow/script/ScriptBinding.groovy#L286-L287). Extend this to support snake case by checking for hyphens as well.

https://github.com/nextflow-io/nextflow/pull/3884

ahvigil:
one more thought- the existing implementation creates duplicate `longName` and `long-name` entries in the underlying map as part of the call to `.put`- it might be more graceful to normalize these at the point where we access the map ( in `.get`) to avoid the duplicate entries

PR #3884, opened 2023-04-19
normalize CLI pipeline params

ahvigil:
fixes https://github.com/nextflow-io/nextflow/issues/3878. There are two changes here:

- `snake_case` pipeline param names are now handled in a way that is consistent with what nextflow does for `kebab-case` and `camelCase` params, which are already interchangeable on the command line. That means `--long-name`, `--longName`, and `--long_name` are parsed as the same param.
- change the way param names get normalized so that instead of storing duplicate entries in the ParamsMap for each parameter representation, we normalize before accessing the underlying map. With the previous implementation, `ParamsMap('foo-bar':1)` produces a ParamsMap of size 2, which seems kind of confusing.

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/script/ScriptBindingTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/script/ScriptBindingTest.groovy
@@ -75,29 +75,30 @@ class ScriptBindingTest extends Specification {
 
     }
 
-    def 'should convert hyphen separated string to camel case' () {
+    def 'should normalize strings to camel case' () {
 
         expect:
-        ScriptBinding.ParamsMap.hyphenToCamelCase('a') == 'a'
-        ScriptBinding.ParamsMap.hyphenToCamelCase('A') == 'A'
-        ScriptBinding.ParamsMap.hyphenToCamelCase('a-b-c-') == 'aBC'
-        ScriptBinding.ParamsMap.hyphenToCamelCase('aa-bb-cc') == 'aaBbCc'
-        ScriptBinding.ParamsMap.hyphenToCamelCase('alpha-beta-delta') == 'alphaBetaDelta'
-        ScriptBinding.ParamsMap.hyphenToCamelCase('Alpha-Beta-delta') == 'AlphaBetaDelta'
-
-    }
-
-    def 'should convert camel case string to hyphen separated' () {
-
-        expect:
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/test/groovy/nextflow/script/ScriptBindingTest.groovy
```diff
@@ -133,7 +134,7 @@ class ScriptBindingTest extends Specification {
 
         then:
         map['field1'] == 1
-        map['field-1'] == null
```
ahvigil, line 136: this test case changes because camel case is a bit ambiguous with numbers- `field-1` normalizes to `field1` but field1 wasn't converting via `camelCaseToHyphen` into `field-1`.

modules/nextflow/src/test/groovy/nextflow/script/ScriptBindingTest.groovy
```diff
@@ -184,7 +185,7 @@ class ScriptBindingTest extends Specification {
         when:
         def params = new ScriptBinding.ParamsMap('foo-bar':1)
         then:
-        params.size() == 2
+        params.size() == 1
```
ahvigil, line 188: `ParamsMap.size` now more accurately reflects the number of params it contains.

ahvigil:
CLI is being rewritten in https://github.com/nextflow-io/nextflow/pull/3600 so this code will soon be deprecated, closing

bentsherman:
Since the params normalization code is separate from the CLI, it should not be affected by the new CLI, so let's keep this open.

However I would prefer to merge this fix #4188 first.

ahvigil:
updated with merge from `master` to incorporate changes from https://github.com/nextflow-io/nextflow/pull/4188 with necessary updates, and made a small update to mention in the docs

Nish9:
Looks like this is not merged. Any idea when this feature will be available?

bentsherman:
Unfortunately I don't think this will be merged soon because the params are being overhauled (#4669) and this change would likely need to be incorporated into that. Feel free to comment there to share your thoughts.

I have been thinking about snake case though, and I think, if we have a params schema then we can intelligently convert kebab-case CLI params to either snake-case or camel-case depending on whether that param is defined in the schema.

Also there is no need to keep the kebab-case params in the params map for the script since you wouldn't access those anyway (see #4702)

pditommaso:
Outdated

bentsherman:
Note: I would like to make CLI v2 (#3600) normalize CLI params to snake_case instead of camelCase, since snake_case is basically always used for params.

PR #3902, opened 2023-04-27
Cleanup version notes in docs

bentsherman:
This PR attempts to standardize the wording and placement of version notes in the documentation, so that they are easy to spot without distracting from the text.

- New features are marked as *New in version ...`, changes are marked as *Changed in version ...*.
- Use italics instead of admonitions so that there aren't so many admonitions.
- Add tons of missing version notes for new features from the last few release cycles.

Basically just more prep work for the website overhaul.

### DIFF
--- a/docs/_static/theme.css
+++ b/docs/_static/theme.css
@@ -19,3 +19,43 @@
     margin-top: 0.5rem;
     width: 60%;
 }
+
+/* Version change admonitions */
+div.deprecated, div.versionadded, div.versionchanged {
+    color: #555555;
+    border-radius: 0.25rem;
+    box-shadow: 0 0.2rem 0.5rem var(--pst-color-shadow),0 0 0.0625rem var(--pst-color-shadow)!important;
+    margin: 1.5625em auto;
+    overflow: hidden;
+    padding: 0.2rem 0.6rem;
+    page-break-inside: avoid;
+    position: relative;
+    transition: color .25s,background-color .25s,border-color .25s;
+    vertical-align: middle;
+}
+div.deprecated p, div.versionadded p, div.versionchanged p {
+    margin-bottom: 0.6rem;
+}
+div.deprecated p:last-child, div.versionadded p:last-child, div.versionchanged p:last-child {
+    margin-bottom: 0;
+}
+div.deprecated p:first-child::before, div.versionadded p:first-child::before, div.versionchanged p:first-child::before {
+    font-family: FontAwesome;
+    display: inline-block;
+    font-style: normal;
+    font-weight: 400;
+    line-height: 1;
+    text-decoration: inherit;
+    margin-right: 0.5rem;
+    content: "";
+    -webkit-font-smoothing: antialiased;
+    color: #ffffff;
+    padding: 0.4rem 0.6rem;
+    margin: -0.2rem 0.6rem -0.2rem -0.6rem;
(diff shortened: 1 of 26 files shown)


ewels:
> standardize the wording and placement of version notes in the documentation
> New features are marked as *New in version ...`, changes are marked as Changed in version ....

Yes, this is awesome. This kind of standardisation of wording and placement is a great improvement 👍🏻 

> Use italics instead of admonitions so that there aren't so many admonitions.

I actually dislike this, sorry. Admonitions can easily be customised to make their styling more subtle with CSS. Italics is just regular typography though, so we lose the ability to target these specific elements.

In the future we can have also have collapsable blocks as done with [material mkdocs here](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#collapsible-blocks).
Could look into Sphinx admonitions / MyST class name annotations in the short term. It should be fairly easy to add a bit of custom CSS to a specific class to make them less annoying, maybe even collabsible.

> Add tons of missing version notes for new features from the last few release cycles.

Generally speaking, I wonder if we should have a policy for how far back the _"New in version xxx"_ notices go? I see some old ones moved in here which go back to 2019! This becomes especially relevant if we manage to get versioned docs in the future, as they serve much less purpose (could maybe then be removed entirely?).

bentsherman:
Here are some example docs that I drew inspiration from:
- [Python (asyncio)](https://docs.python.org/3/library/asyncio-task.html)
- [pandas (read_csv)](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html)
- [Kubernetes (Volumes)](https://kubernetes.io/docs/concepts/storage/volumes/)

Overall, I'm in favor of doing italics as they are, or a smaller admonition like you see in the pandas docs: 
image

I agree that we should have some kind of "version note" component so that we can control the styling from one place. But I haven't found an easy way to do it with Sphinx/MyST, so for me the italics was the simplest solution because at least we can control them all with a search & replace, e.g. with a version note component in Astro. Do you just not like that it's not a directive, or do you also not like the italics? If it's only the former, would you be fine with just having the italics for now until we make a proper component in Astro?

As for the additional version notes, most of them were recent, but I may have added a few from the `19.x.x` series just because I noticed them. I removed many of the really old notes (i.e. from the `0.x.x` series) a while ago. I don't know what the cutoff should be, but if we do decide to maintain versioned docs, it would be better to have them in there so that we know which sections to keep or remove.

ewels:
I had a quick look and found an [admonition type](https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html#admonition-types) dedicated to this use case. These admonition types are used in the examples you gave, just with varying CSS styling.

I've had a stab at implementing this in https://github.com/nextflow-io/nextflow/pull/3907 (based on this branch).

Let me know what you think 😄

bentsherman:
Yo! I can't believe I missed this, that's exactly what pandas uses 😆

ewels:
And Python 😉 If it makes you feel any better, I only noticed it half way through building my own custom admonition type 😆

bentsherman:
And we're back. Phil, I adjusted the custom CSS to make the admonitions not pop as much. I also standardized the "experimental" warnings and added versions to the deprecation notes.

Here are some examples:

image

image

image

image

@pditommaso This PR is ready for you.

PR #3904, opened 2023-04-28
Fix problem args were not correctly represented in debug yaml

Lehmann-Fabian:
This problem occurred when a workflow was run with `k8s.debug.yaml = true`
The created `.command.yaml` file contained an args like this:
```
...
    args:
    - /bin/bash
    - -ue
    - !!org.codehaus.groovy.runtime.GStringImpl {}
...
```
Converting the GString to String looks like it should:
```
...
args: [/bin/bash, -ue, /input/data/work/aa/abf3408753f4c094567abc16d1e543/.command.run]
...
```


bentsherman (approved):

pditommaso:
Well spotted 👍

PR #3892, opened 2023-04-21
Job arrays

bentsherman:
Closes #1477 (and possibly #1427)

Summary of changes:

- Adds `array` directive to submit tasks as array jobs of a given size

- `TaskArrayCollector` collects tasks into arrays and submits each array job when it is ready to the underlying executor. The executor must implement the `TaskArrayAware` interface. Each process has its own array job collector.

  When all input channels to a process have received the "poison pill", the process is "closed" and the array job collector is notified so that it can submit any remaining tasks. All subsequent tasks (e.g. retries) will be submitted as individual tasks.

- `TaskArray` is a special type of `TaskRun` for an array job that holds the list of child task handlers. For an executor that supports array jobs, the task handler can check whether its task is a `TaskArray` to apply perform job specific behavior.

- `TaskHandler` has a few more methods, which the array job collector uses to create the array job script. This script simply defines the list of child work directories, selects a work dir based on the index, and launches the child task using an executor-specific  launch command.

- `TaskPollingMonitor` has been modified to handle both array jobs and regular tasks. The array job is handled like any other task, but then discarded once it has been submitted. The task stats reported by Nextflow are the same with or without array jobs -- array jobs are not included in the task stats.

Here's the pipeline I'm using as the e2e test:
```groovy
params.n_tasks = 50

process foo {
    array 10

    input: val index
    output: path 'output.txt'

    """
    echo "Hello from task ${index}!" > output.txt
    """
}

process bar {
    debug true
    array 10

    input: path 'input.txt'

    """
    cat input.txt
    """
}

workflow {
    Channel.of(1 .. params.n_tasks) | foo | bar
}
```

TODO:
- [x] documentation
- [x] unit tests
- extra
  - [x] AWS Batch: kill array job instead of child jobs
  - [x] Google Batch: kill array job instead of child jobs
  - [x] Grid executors: add array index environment var to container
  - [x] handle retried tasks with dynamic resources
- manual e2e tests
  - [x] SLURM
  - [x] SLURM + Fusion
  - [x] AWS Batch
  - [x] AWS Batch + Fusion
  - [x] Google Batch
  - [x] Google Batch + Fusion

PR #3905, opened 2023-05-01
Implement job array using TaskRun as carrier

pditommaso:
This PR is a refactoring of #3892 that removes a lot of code duplication introduced by the submitter pattern. 

The main idea is to avoid specialised task submitter classes, e.g. `GridArrayTaskSubmitter`  and  `ArrayTaskSubmitter`, that introduce a lot of code duplication. Instead, it creates a new `TaskRun` instance for each array of tasks that is used as the carrier for the execution of job array. 

Using this approach, the job-array launch is managed like any other task, with the only difference that the task script is the Bash script to launch the children's tasks. See here 

https://github.com/nextflow-io/nextflow/blob/535f483c3f839cd88612ea1ba3956058b0c1d3c2/modules/nextflow/src/main/groovy/nextflow/processor/ArrayTaskCollector.groovy#L110-L110

The array-specific options are exected to be implemented as any other job directive, see for example here

https://github.com/nextflow-io/nextflow/blob/535f483c3f839cd88612ea1ba3956058b0c1d3c2/modules/nextflow/src/main/groovy/nextflow/executor/SlurmExecutor.groovy#L58-L60

The proposed implemented works as expected, tho some things remain to be solved: 

- [ ] There's inconsistency with num of tasks reported.  
- [ ] The job index variable, e.g. `SLURM_ARRAY_TASK_ID`, should be added to the list of exported variables when the child job is executed via a container 
- [ ] When the array job is made by 1 task, it should be submitted as a plain task.
- [ ] It should take into account the impact of job array with Fusion, which is expecting the task to be confined in its own work dir. Instead job array launch the execution of the child job into a diffetent task work directory
- [ ] Unit tests need to be fixed. 

This PR was meant to be against the branch `1477-job-array-refactor`, at commit 3e38ec05. However, the latest commits caused too many conflicts. I've created an independent commit for now.


modules/nextflow/src/main/groovy/nextflow/processor/ArrayTaskCollector.groovy
```diff
@@ -0,0 +1,135 @@
...
+            closed = true
+        }
+        finally {
+            sync.unlock()
+        }
+    }
+
+    protected void submit0(List<TaskRun> array) {
+        // final submit the task array
+        final task = processor.createTaskArray(array)
```
pditommaso, line 109: Creates a `TaskRun` instance referencing the array of tasks.

modules/nextflow/src/main/groovy/nextflow/processor/ArrayTaskCollector.groovy
```diff
@@ -0,0 +1,135 @@
...
+        task.script = createTaskArrayScript(array)
```
pditommaso, line 110: Creates the Bash script that executes the job array

modules/nextflow/src/main/groovy/nextflow/processor/ArrayTaskCollector.groovy
```diff
@@ -0,0 +1,135 @@
...
+        executor.submit(task)
```
pditommaso, line 111: Submit as any other task

modules/nextflow/src/main/groovy/nextflow/processor/ArrayTaskCollector.groovy
```diff
@@ -0,0 +1,135 @@
...
+        // the launcher entry point
+        final index = executor.getArrayIndexName()
+        if( index ) {
+            script << '# execute the i-th task\n'
+            script << '/bin/bash ${tasks[' << index << ']}\n'
+        }
+        else {
+            script << '# execute all tasks\n'
+            script << 'for t in ${tasks[@]}; do /bin/bash $t; done\n'
+        }
```
pditommaso, line 132: This ideally, will handle the execution of tasks in batches. The implementation could be very similar

bentsherman:
Also, while I said something about not allowing task arrays and task grouping to be used together, I think I had it backwards, actually it might be interesting to use them together for processes that execute ~100,000 tasks. For that reason we'll want to keep both features modular rather than try to manage them with the same code.

pditommaso:
Not sure to agree, the idea of this approach is to reduce a job array as any other task executed by nextflow and therefore there should be a TaskRun to model it.  

This has the advantage it's not needed to handle exception or custom logic for handling the task array because it mostly fit in the current design. 

Also, the job array directories are placed where all other cluster directives are currently coded. 

> Consider that a TaskRun could be retried multiple times, and each time it could belong to a different task array.

This is fair point, but it can be easily managed by turning off the retry for `TaskRun` marked as a task array. 


Also, I'd like to not have a separate implementation for task grouping, don't see a big value mixing job array and job group submissions. 

Instead, what could be interesting is to be able to create tasks array and tasks group that span across multiple processes (provided they have the same resource requirements)

bentsherman:
But an array job is not a regular job, it is a special way to submit many jobs. Once it is submitted, the child jobs become regular jobs completely independent of each other. Each job could start and stop at any time regardless of the others. If you model the array job as a task run, each child will have to wait for the entire array job to finish before it can be finalized and trigger downstream tasks.

For task groups this makes perfect sense, because the child tasks are run sequentially on the same node, so they do have to wait for the entire task group to finish.

bentsherman:
Regarding the retry, I'm not sure what you mean by turning if off. I'm not talking about the array job itself, but the child tasks. If a child task fails and is retried, it will be retried in a different array job. So it's array index will change on each attempt.

I think array jobs and task groups would be useful together at large scale. For example, consider a process that executes ~100,000 short-running tasks. You decide to execute them in groups of 100 each because that is a reasonable runtime, but then you still have to submit 1,000 jobs. Now combine that with array jobs, you can submit a single array job of 1,000 tasks. These features should complement each other quite well at large scales.

Mixing processes in the same array job or task group is possible but I think very unlikely. They would need to have the exact same resource requirements and container, and I think the container alone makes it very unlikely. Consider how diverse the nf-core pipelines are in terms of how many different containers they use.

bentsherman:
Putting a note here so that I don't forget. One thing I could try is to have a generic task array submitter that creates a task run with the array script (like in this PR), but simply call the `submit()` method on it instead of submitting it to the executor. So the child tasks will still go through the monitor and defer to the array submitter as in my main PR, but the array submitter can defer to a regular task handler, no need to specialize for each executor.

A minor annoyance still is that the array job will create an extra task directory and the array script will be needlessly wrapped in a wrapper script, but maybe we can address those things with some conditional logic in the right place.

bentsherman:
Okay, I found a way to incorporate your ideas into my array job PR with only a few differences:
- create a TaskArray subclass of TaskRun instead of adding a field to TaskRun
- instead of submitting the array TaskRun to the executor, create the task handler directly and only call `prepareLauncher()` and `submit()` on it
- submit the child tasks to the monitor as in my PR, so they can be managed independently

So now the array job PR is much simpler overall, while avoiding the issues I mentioned previously in this thread. I will test when I get back next week. The array job wrapper script likely still needs some tweaking.

pditommaso:
> Each job could start and stop at any time regardless of the others. If you model the array job as a task run, each child will have to wait for the entire array job to finish before it can be finalized and trigger downstream tasks.


That's true for the child job, but also the array job has it's own status, look for example Slurm 

```
squeue
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
        4098_[1-2]      long   sbatch    paolo PD       0:00      1 (Resources)
            4095_1      long   sbatch    paolo  R       0:49      1 ip-172-31-37-61
            4095_2      long   sbatch    paolo  R       0:49      1 ip-172-31-37-61
            4095_3      long   sbatch    paolo  R       0:49      1 ip-172-31-37-61
            4098_0      long   sbatch    paolo  R       0:01      1 ip-172-31-37-61
```

and, for example, [AWS Batch](https://docs.aws.amazon.com/batch/latest/userguide/array_jobs.html) 

> The parent array job can enter a SUBMITTED, PENDING, FAILED, or SUCCEEDED status. An array parent job is updated to PENDING when any child job is updated to RUNNABLE

This is why this PR creates a TaskRun both for the parent job and the child jobs. The status of each of them is monitored independently 

> Regarding the retry, I'm not sure what you mean by turning if off. I'm not talking about the array job itself, but the child tasks. If a child task fails and is retried, it will be retried in a different array job. So it's array index will change on each attempt

I was referring the re-try of the parent job that should be avoid, only the child jobs should be retried. 


> Mixing processes in the same array job or task group is possible but I think very unlikely

it's not so unlikely, I see several pipeline using the same container across many process. But I agree that's a secondary goal.

bentsherman:
Paolo, I took a closer look at your changes to the task polling monitor, and I see now that our PRs are doing a very similar thing. I'm fine with this approach since it removes the need for the array submitter class, so I will try to incorporate these changes into my PR.

However, there is no point to tracking the status of the array job, because it is just an aggregate of the child job statuses. For some HPC schedulers I'm still not sure if they even support array job status. So we can discard the array job handler once the child jobs are submitted.

pditommaso:
> However, there is no point to tracking the status of the array job, because it is just an aggregate of the child job statuses

The good thing is that it comes for free, but I agree that's in some cases it cannot be possible or it may need adaptation. In this cases it should be fairly easy to specialize the `checkIfCompleted` and `checkIfRunning` methods in the corresponding handler classes

bentsherman:
Well my main point is that there is no need to track the array job status.

I also see what you mean by the inconsistency in the number of tasks reported. The log observer knows the correct number of submitted tasks but not created tasks, so it reports something like `[1000%] 10 of 1`. I'm not sure why that's happening but I will dig in and try to fix it.

A related issue, the array job allows you to bypass the queue limit. You could have a queue limit of 100, submit 100 array jobs of size 100 each, then you'll have 10,000 actual jobs in the queue. Should we account for that when considering whether an array job can be submitted? I think some schedulers also allow you to define how many child tasks can be run concurrently...

bentsherman:
The task reporting is off because the task polling monitor only notifies the array job as pending and not the child jobs. I guess we can fix it using similar logic as with the submit.

pditommaso:
> Well my main point is that there is no need to track the array job status.

Understand, think you are right. it may simplify some stuff. To achieve it should be enough to avoid to added the "parent" handler to the `runningQueue` 

https://github.com/nextflow-io/nextflow/blob/d24d97645f31b6156aec4499cff194ee014661ec/modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy#L226-L228

this should also solve the problem with the queue limit you are reporting. 


In the same manner, it should not trigger the task submit event here

https://github.com/nextflow-io/nextflow/blob/d24d97645f31b6156aec4499cff194ee014661ec/modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy#L228-L228

and along the same lines the task pending event here 

https://github.com/nextflow-io/nextflow/blob/d24d97645f31b6156aec4499cff194ee014661ec/modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy#L256-L256

bentsherman:
That is exactly what I did in my PR 😄 

The task reporting should be correct now

pditommaso:
In the regard of tracking the parent array task, it should be taken into consideration that a critical use case is to allow the deletion of tasks array jobs via the parent job, instead of the individual tasks. 

This is essential to improve the capability to terminate workflow launching a large number of tasks. For example when using AWS Batch deleting one by one can take up to hours!

bentsherman:
Generally speaking, HPC schedulers allow you to kill an array job, it simply expands to killing the list of child jobs in a single command. Since you already have a mechanism to delete jobs in batches via `BatchCleanup`, it's already as good as it's gonna get. Adding a separate mechanism for killing array jobs won't bring any improvement.

This is mostly downstream of the fact that child jobs are independent jobs, so the array job only improves the scheduling efficiency.

bentsherman:
As for AWS Batch, it looks like you can only delete one job at a time through the API. So there might be some value to deleting an array job. But we would want to make sure that it's actually faster. Surely it must be, because it wouldn't be acceptable for a single API call to take hours. But a simple test would be to create an array job with 1,000 tasks and delete it, then create 1,000 regular tasks and delete them, and observe the time difference.

If it is actually faster, we could implement it just for AWS Batch by having the reaper detect and de-duplicate API calls for child jobs.

bentsherman:
This PR has been incorporated into #3892, closing in favor of that one

PR #3906, opened 2023-05-01
Enable static compilation for task processing code

bentsherman:
I noticed Paolo added static compilation to `TaskProcessor` in a since closed PR, so I decided to take that idea to completion. This PR enables static compilation for most classes in `nextflow.executor` and `nextflow.processor`. These packages are responsible for processing tasks, so they should be the hottest parts of the codebase, and static compilation seems to be the best way to improve performance in Groovy. Also should help us find bugs at compile time instead of run time.


abhi18av:
Hi @bentsherman 😊 

Curious about this PR - I noticed that the major changes are related to the refactoring of method calls from Groovy style to (basically) Java style? 

If the entire nextflow-core codebase is made compatible for static compilation - would it by any chance help the compilation of pipelines or the Nextflow binary itself via GraalVM?

bentsherman:
Yes, some classes like `TaskConfig` and `TraceRecoard` use "metaclass" methods like `getProperty()` and `propertyMissing()` to route property assignments to method calls. Under static compilation, these metaclass methods are ignored, so you have to make explicit method calls.

For the Nextflow codebase I think the static compilation is very helpful, both in terms of improving performance and error checking. But we would need to statically compile the entire Nextflow codebase in order to run it as a GraalVM native image, and certain parts of the codebase would be hard to make static. On top of that, since GraalVM native image doesn't support Groovy directly, you have to compile the Groovy code into Java bytecode and provide metadata about method calls that use reflection. Jordi does this for Tower CLI, which is written entirely in Java, and he says it's still a huge hassle to maintain.

For Nextflow scripts, static compilation probably wouldn't improve performance very much because the script execution is a relatively small part of overall runtime. But I imagine many users would appreciate having better compile-time type checking. I think the blocker there is the custom DSL for processes and workflows, which relies on a lot of dynamic Groovy magic. We might have to come up with a DSL syntax that better allows for static typing first.

abhi18av:
Hmm, interesting points there Ben. The point regarding the custom DSL might be the hard blocker here since other dynamic languages on the JVM, such as Clojure, have been used with GraalVM to create a native app https://github.com/babashka/babashka.

bentsherman:
Well they are really two separate efforts. The native image is a far off dream, but static typing in the Nextflow DSL is something we can pursue right now. And we don't have to statically compile the entire Nextflow codebase to make that work.

pditommaso:
Nice cleanup

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -1781,7 +1786,7 @@ class TaskProcessor {
         if( obj instanceof Path )
             return obj
 
-        if( obj == null )
+        if( !obj == null )
```
bentsherman: Is this not a bug?

bentsherman:
Thanks. Do you still want to try to improve the static compilation of the task processor and task run? I see you undid most of their changes, but they are probably the most performance-critical classes in the PR.

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
pditommaso: Indeed. I wanted to make a separate PR with a test but then I forgot about that

pditommaso:
> Do you still want to try to improve the static compilation of the task processor and task run? 

One reason I've reverted it. It's quite sure it was going to conflict with the work with job array. Also it made a bit less readable. 

Don't think there's a major performance benefit using @CompileStatic in this code. It's mostly useful because it makes easier to spot mistakes at compile time

bentsherman:
Regarding the array jobs, I found that if I create a TaskRun in a compile-static context, I have to provide the script in the constructor. If I try to assign the script after the fact, I got an error about the `script` field being read-only. Which is weird because it's a public field. But it only happens if the caller is compile-static. I don't think that will cause any conflict though.

Anyway, I haven't done any performance profiling so no way to really know. Based on my reading, static compilation usually improves performance by 2-3x, but also the tasks are probably a greater bottleneck than the workflow engine. It would be nice for catching errors where it doesn't require lots of type-casting.

PR #3907, opened 2023-05-01
Phil docs version notes

ewels:
Converted the new italic version notices added by @bentsherman in #3902 to admonitions, using dedicated version-change admonition types:

https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html#admonition-types

```markdown
:::{versionadded} 19.07.0
A new feature was added.
:::

:::{versionchanged} 22.08.0-edge
Something changed.
:::

:::{deprecated} 21.04.0
:::
```

Added a bit of custom CSS to style them:

CleanShot 2023-05-01 at 23 36 12@2x

This syntax should be no problem in a new Astro site. Worst case scenario, they will be easier to change using global find + replace.

Sanity checks / making sure I didn't break anything / suggestions on styling welcome 🙏🏻 

Note that we can [make admonitions collapsible](https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html#collapsible-admonitions) if we want as well.

### DIFF
--- a/docs/_static/theme.css
+++ b/docs/_static/theme.css
@@ -19,3 +19,43 @@
     margin-top: 0.5rem;
     width: 60%;
 }
+
+/* Version change admonitions */
+div.deprecated, div.versionadded, div.versionchanged {
+    color: #555555;
+    border-radius: 0.25rem;
+    box-shadow: 0 0.2rem 0.5rem var(--pst-color-shadow),0 0 0.0625rem var(--pst-color-shadow)!important;
+    margin: 1.5625em auto;
+    overflow: hidden;
+    padding: 0.2rem 0.6rem;
+    page-break-inside: avoid;
+    position: relative;
+    transition: color .25s,background-color .25s,border-color .25s;
+    vertical-align: middle;
+}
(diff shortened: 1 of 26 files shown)


bentsherman:
Baller. Thanks for looking into this.

I made a few changes to the CSS because some of the longer admonitions looked awkward. I used the MyST docs you linked as guidance.

Here are some before/after examples. The after examples have smaller width only because I had the dev tools open 😅 

Before:

image

image

After:

image

image

If this looks good to you, I'll merge into my PR and we can continue from there. I still need to adjust some of the "EXPERIMENTAL" admonitions...

ewels:
Nice! Yeah I started with this design and then was messing around with the other. Not sure why I changed. Agree it looks crap when broken onto two lines. Yours is better.

My only styling concern is that these are too "loud", especially when there are lots of them (for both styles). Don't want them to be too visually heavy when there are lots. Could maybe drop the background colour opacity a little? 🤔

ewels:
LGTM to merge anyway 👍🏻

bentsherman:
Okay, let's merge and see what Paolo thinks

PR #3927, opened 2023-05-10
Azure jobs correctly deleted after completion

### LINKED ISSUE #3926: Azure Batch jobs not cleaned up after task failure
## Bug report 

(Please follow this template replacing the text between parentheses with the requested information)

When an Azure Batch task fails, Nextflow leaves the job in active state where it can fill the active job quota. 

Instead, it should move the job to the completed or deleted state upon workflow completion.  It could use automatic termination after submitting all tasks by setting the [onAllTasksComplete](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.batch.cloudjob.onalltaskscomplete?view=azure-dotnet#microsoft-azure-batch-cloudjob-onalltaskscomplete) parameter.

Azure documentation: https://learn.microsoft.com/en-us/azure/batch/jobs-and-tasks#job-manager-tasks-and-automatic-termination 

main.nf:

```nextflow
#!/usr/bin/env nextflow
nextflow.enable.dsl=2 

process FAIL {
    container 'ubuntu:20.04'

    script:
    """
    exit 1
    """
}

workflow {

    main:
       FAIL()

}
```

nextflow.config:

```
process {
    executor = 'azurebatch'
}

azure {
    storage {
        accountName = "$AZURE_BATCH_STORAGE_NAME"
        accountKey  = "$AZURE_BATCH_STORAGE_KEY"
    }
    batch {
        location                = "$AZURE_REGION"
        accountName             = "$AZURE_BATCH_ACCOUNT_NAME"
        accountKey              = "$AZURE_BATCH_ACCOUNT_KEY"
        autoPoolMode            = true
        allowPoolCreation       = true
        deletePoolsOnCompletion = true
        deleteJobsOnCompletion  = true
    }
}
```

Command:

```
nextflow run main.nf -w az://work/
```

In my case, a job called `job-5f163d784f2037f1ec3f-FAIL` is left in active state with no tasks in it's queue. Since it contains no tasks or logs it is not useful and just uses up the quota. This should be removed upon failure. 

### Program output 

[Uploaded log.](https://github.com/nextflow-io/nextflow/files/11433887/nextflow.log.log)

### Environment 

* Nextflow version: 
```
       N E X T F L O W
      version 23.04.0 build 5857
      created 01-04-2023 21:09 UTC (22:09 BST)
      cite doi:10.1038/nbt.3820
      http://nextflow.io
```
* Java version:
```
openjdk version "17.0.6" 2023-01-17
OpenJDK Runtime Environment Temurin-17.0.6+10 (build 17.0.6+10)
OpenJDK 64-Bit Server VM Temurin-17.0.6+10 (build 17.0.6+10, mixed mode, sharing)
```
* Operating system:
Ubuntu 20.04 via WSL2 Windows 11
```
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.1 LTS
Release:        22.04
Codename:       jammy
```
```
GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
```

### Additional context

This also affects Tower head node submissions.

adamrtalbot:
Nextflow was leaving Azure Batch jobs around in active state after completion. This does the following:

 - Nextflow now deletes jobs even if a task has failed
 - If config `azure.batch.TerminateJobsOnCompletion` is true (default), it will set the job to automatically terminate after all tasks have finished. This no longer counts towards Azure Batch quota.

Fixes #3926

### DIFF
--- a/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
+++ b/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
@@ -38,8 +38,10 @@ import com.microsoft.azure.batch.protocol.models.ContainerConfiguration
 import com.microsoft.azure.batch.protocol.models.ContainerRegistry
 import com.microsoft.azure.batch.protocol.models.ElevationLevel
 import com.microsoft.azure.batch.protocol.models.ImageInformation
+import com.microsoft.azure.batch.protocol.models.JobUpdateParameter
 import com.microsoft.azure.batch.protocol.models.MountConfiguration
 import com.microsoft.azure.batch.protocol.models.NetworkConfiguration
+import com.microsoft.azure.batch.protocol.models.OnAllTasksComplete
 import com.microsoft.azure.batch.protocol.models.OutputFile
 import com.microsoft.azure.batch.protocol.models.OutputFileBlobContainerDestination
 import com.microsoft.azure.batch.protocol.models.OutputFileDestination
@@ -783,14 +785,37 @@ class AzBatchService implements Closeable {
         apply(() -> client.taskOperations().deleteTask(key.jobId, key.taskId))
     }
 
+    protected void terminateJobs() {
+        /* 
+        We set the job to terminate when all tasks are complete rather than directly terminating, this allows Azure Batch to handle the termination for us.
+        */
+
+        for( Map.Entry<TaskProcessor,String> entry : allJobIds ) {
+            final proc = entry.key
+            final jobId = entry.value
+
+            try {
+                log.trace "Terminating Azure job ${jobId}"
+
+                CloudJob job = apply(() -> client.jobOperations().getJob(jobId))
+                final poolInfo = job.poolInfo()
+
+                JobUpdateParameter jobParameter = new JobUpdateParameter()
+                        .withOnAllTasksComplete(OnAllTasksComplete.TERMINATE_JOB)
+                        .withPoolInfo(poolInfo)
+
+                apply(() -> client.jobOperations().updateJob(jobId, jobParameter))
+            }
+            catch (Exception e) {
+                log.warn "Unable to terminate Azure Batch job ${jobId} - Reason: ${e.message ?: e}"
+            }
+        }
+    }
+
     protected void cleanupJobs() {
         for( Map.Entry<TaskProcessor,String> entry : allJobIds ) {
             final proc = entry.key
             final jobId = entry.value
-            if( proc.hasErrors() ) {
-                log.debug "Preserving Azure job with error: ${jobId}"
-                continue
-            }
 
             try {
                 log.trace "Deleting Azure job ${jobId}"
@@ -826,6 +851,11 @@ class AzBatchService implements Closeable {
     }
     @Override
     void close() {
+        // Terminate existing jobs to prevent them occupying quota
+        if( config.batch().terminateJobsOnCompletion!=Boolean.FALSE ) {
+            terminateJobs()
+        }
+
         // cleanup app successful jobs
         if( config.batch().deleteJobsOnCompletion!=Boolean.FALSE ) {
             cleanupJobs()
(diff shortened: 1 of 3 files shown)


plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -783,14 +785,37 @@ class AzBatchService implements Closeable {
...
+                log.trace "Terminating Azure job ${jobId}"
+
+                CloudJob job = apply(() -> client.jobOperations().getJob(jobId))
+                final poolInfo = job.poolInfo()
+
+                JobUpdateParameter jobParameter = new JobUpdateParameter()
+                        .withOnAllTasksComplete(OnAllTasksComplete.TERMINATE_JOB)
+                        .withPoolInfo(poolInfo)
+
+                apply(() -> client.jobOperations().updateJob(jobId, jobParameter))
```
bentsherman, line 807: Instead of updating the job we should set this property when the job is created in `getOrCreateJob()`

Also, do we even need the config option? Can we just use this setting all the time?

adamrtalbot: Probably, but I thought we might as well add an option.

adamrtalbot: You must set this after adding the tasks, otherwise the job will terminate immediately.

adamrtalbot: How about setting it after runTask()?

https://github.com/nextflow-io/nextflow/blob/2959f0915ce5b77520680f4b11dcdb822b1ef425/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L320

bentsherman: You would have to wait until all tasks for a job have been created, so it might be correct as is

bentsherman: I think an Azure job corresponds to a Nextflow process, so you could update the job once the process is "closed" (all tasks have been created) and that would be the most eager approach. But might be overkill. Let's see what Paolo thinks.

adamrtalbot: Yep, that's what I was thinking. Just not sure where to look in the code for that.

I can update this to just terminateTask now and convert this to a setJobAutoTermination method that's ready to use after all tasks have been submitted?

The other question is, does Nextflow re-use an Azure Batch job when it resumes? I don't think so but worth checking.

bentsherman: It doesn't look like jobs are re-used across runs.

We just need an "onProcessClose" event to notify when all tasks are created. I've already added it in a number of PRs, like this one: #3818

So if Paolo likes the idea then we can just rip the code from that PR.

bentsherman: https://github.com/nextflow-io/nextflow/pull/3818/files#diff-3250bcd11fe7bb7aeeaf9839c44ac5d737dc27ddc9a734954789d09eb9dab399

adamrtalbot: For now, I'm going to leave this as it is, keep this PR simple and just fix one bug.

pditommaso: Make sense. It can be improved in a following iteration

PR #3929, opened 2023-05-10
Document use of aws scope for non-Amazon/AWS S3. 

jfy133:
I was today trying to set up a nextflow run to store output on an s3 bucket hosted on an Openstack infrastructure.

I eventually realised that I had to use the `aws` scope even if I was not using a AWS/Amazon cloud.

This PR adds a little note clarifying this (and hopefully also makes functionality compatible with Openstack (and presumably non AWS/GCP/Azue infrastructure more findable within the docs).

Note: I am only just getting into cloud infrastructure so my terminologies may be wrong - please feel to make direct modifications to the PR.

Derived from [nf-core slack discussion](https://nfcore.slack.com/archives/CE6SDBX2A/p1683711013957429)

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -141,6 +141,10 @@ aws {
 }
 ```
 
+:::{tip}
+This scope can also be used for non-Amazon S3 methods such as Ceph (often used with Openstack clouds)
+:::
+
 Read the {ref}`aws-page` and {ref}`amazons3-page` pages for more information.
 
 The following settings are available:
--- a/docs/config.md
+++ b/docs/config.md
@@ -142,7 +142,7 @@ aws {
 ```
 
 :::{tip}
-This scope can also be used to configure access to S3-compatible storage outside of AWS.
+This scope can also be used to configure access to S3-compatible storage outside of AWS, such as [Ceph](https://ceph.com/en/) and [MinIO](https://min.io/).
 :::
 
 Read the {ref}`aws-page` and {ref}`amazons3-page` pages for more information.
--- a/docs/config.md
+++ b/docs/config.md
@@ -142,7 +142,7 @@ aws {
 ```
 
 :::{tip}
-This scope can also be used for non-Amazon S3 methods such as Ceph (often used with Openstack clouds)
+This scope can also be used to configure access to S3-compatible storage outside of AWS.
 :::
 
 Read the {ref}`aws-page` and {ref}`amazons3-page` pages for more information.
--- a/docs/amazons3.md
+++ b/docs/amazons3.md
@@ -2,7 +2,7 @@
 
 # S3 storage
 
-Nextflow includes support for S3 storage. Files stored in an S3 bucket can be accessed transparently in your pipeline script like any other file in the local file system. The following sections focus on Amazon S3 storage, however the same concepts apply to any S3 compatible storage such as Ceph or MinIO storage.
+Nextflow includes support for S3 storage. Files stored in an S3 bucket can be accessed transparently in your pipeline script like any other file in the local file system. The following sections focus on Amazon S3 storage, however the same concepts apply to any S3 compatible platforms such as Ceph or MinIO storage.
 
 ## S3 path
 
--- a/docs/amazons3.md
+++ b/docs/amazons3.md
@@ -1,8 +1,8 @@
 (amazons3-page)=
 
-# Amazon S3 storage
+# S3 storage
 
-Nextflow includes support for Amazon S3 storage. Files stored in an S3 bucket can be accessed transparently in your pipeline script like any other file in the local file system.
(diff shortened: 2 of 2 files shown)


docs/config.md
```diff
@@ -141,6 +141,10 @@ aws {
 }
 ```
 
+:::{tip}
+This scope can also be used for non-Amazon S3 methods such as Ceph (often used with Openstack clouds)
```
bentsherman, line 145:
```suggestion
This scope can also be used to configure access to S3-compatible storage outside of AWS.
```

bentsherman:
@ewels I think something is wrong with the GH action, it is triggering the integration tests even though it's a docs-only PR.

ewels:
@bentsherman it's working as expected - it is triggered (necessary to be able to pass branch protection) but it skips basically all steps and completes in a few seconds instead of several minutes:

CleanShot 2023-05-10 at 14 07 30@2x

docs/config.md
ewels: @bentsherman I actually suggested to @jfy133 that he specifically include the term "openstack" so that if people search the docs for it, it'll come up..

jfy133: Yes... I originally tried searching for 'openstack' and 'ceph' in this case, but didn't find anything. It's very misleading the whole scope is called `aws` as I thought that was restricted to just Amazon hosted infrastructure.

bentsherman:
I see, it looks like you're mostly right, but for some reason the `test_integration` still ran for both Java 11 and 19?

docs/config.md
bentsherman: I removed the openstack mention because I don't want to give the impression that you can use the `aws` scope for other clouds, only S3-compatible storage. So you could access the S3-compatible storage in your openstack instance but not the compute resources, right?

I think it would be fine to mention examples of S3-compatible storage like Ceph and MinIO.

jfy133: Fair enough! Mentioning Ceph and MinIO would be sufficient for me if you're OK with it :) I did try searching `ceph`, and `minIO` is another storage/api option in the OpenStack instance I'm on ([de.NBI cloud](https://cloud.denbi.de/wiki/quickstart/))  so even having that would have worked.

It would require a much larger change but I wonder if the `aws` scope could be split/partially duplcated with an `s3` scope instead (But like I said, I'm naïve with all this stuff so I don't know if that would be valid either).

bentsherman: We definitely should try to communicate better how to use Nextflow with S3-compatible storage, maybe add a section to the page on Amazon S3. For this PR I will just add the examples to the tip...

docs/config.md
```diff
@@ -141,6 +141,10 @@ aws {
 }
 ```
 
+:::{tip}
+This scope can also be used to configure access to S3-compatible storage outside of AWS.
```
bentsherman, line 145:
```suggestion
This scope can also be used to configure access to S3-compatible storage outside of AWS, such as [Ceph](https://ceph.com/en/) and [MinIO](https://min.io/).
```

bentsherman (approved):

docs/config.md
jfy133: Sounds good! Would you like me to make an issue for more generalised docs?

bentsherman: Actually, we might as well do it in this PR... can you add a brief section to the S3 page, showing how you configure S3-compatible storage? You could use your OpenStack / Ceph setup as an example. Basically just show the necessary configuration, I can help with the prose.

jfy133:
Sorry for the many commits @bentsherman, I kept reflecting after each commit and made minor tweaks, and then the inevitable sign-off issues.

To be honest, for my personal case with the Ceph storage I was interacting with, there was very minimal differences to what is described on the existing S3 Page. If you dropped all 'AWS' and 'Amazon' words/titles, it would pretty much still fit with what I had to configure, except for the region and IAM roles. I've tried to do something along those lines in the last couple of commits.

For example, the configuration I have to set is (based on this [info on configuring `s3cmd`](https://cloud.denbi.de/wiki/Tutorials/ObjectStorage/)):

```
aws.client.accessKey = '<key>'
aws.client.secretKey = '<key>'
aws.client.endpoint = 'openstack.cebitec.uni-bielefeld.de:8080' // For Bielefeld
aws.client.protocol = 'https'
aws.client.s3PathStyleAccess = true
```

The PathStyleAccess might not be being necessary.

So you can see how similar it is to what is on the current S3 page.

ewels:
> I see, it looks like you're mostly right, but for some reason the `test_integration` still ran for both Java 11 and 19?

Yeah, you're right. It's only the `Build (n)` tests that have to run. The `test_` ones can be skipped I think.

docs/config.md
jfy133: Done!

pditommaso:
Thanks, this is useful. Ive merged it manually making some changes 95a59734

PR #3933, opened 2023-05-10
Walk file tree on directory publish

### LINKED ISSUE #3372: Validation of published outputs
## Published Output Validation

At the moment, Nextflow is conservative in when it writes published outputs. If a file exists in the publishDir location, then Nextflow does not re-publish the file. If the destination file is truncated, or has been changed by some external process, Nextflow doesn't check to see if the source (in the task work directory) is different to the destination (in the publishDir location).

It would be good if Nextflow made stronger checks above simple filename matching. In cases where the source and destination are both on S3, we can provide strong guarantees by checking the file contents hash in the etags provided by S3.

## Usage scenario 

1. Nextflow runs to completion
2. A user or external process makes changes to an output file (but does not change the file name)
3. Nextflow runs again with `-resume`

In this case, we would expect that Nextflow should notice that the published file is "stale", and that it needs to be re-published.

## Suggest implementation 

In cases where both files are on S3, we can do cheap file integrity checking by the hash value in the S3 etags.

### LINKED ISSUE #3754: Safe guard when using publishDir directive
## New feature

Hi! Thanks for using Nextflow and submitting the proposal 
for a new feature or the enhancement of an existing functionality. 

Please replace this text providing a short description of your 
proposal.

## Usage scenario 

I made a mistake of doing something like this 

```
process foo {
    publishDir "${params.db}", mode: 'copy'

    input:
    path db

    output:
    tuple path(db), path(some/other/file)

    script:
    """
    some command
    """
}
```

Essentially I accidentally published the directory "db" to itself. If I understand it correctly, then when there's a change in the output then the output will be copied to the published directory? In my case, as the "db" got copied to itself, then it was changed and the process continued, creating something like db/db/db/db/... - until it reached either time limit or memory limit. So I essentially created a virus by doing so...

## Suggest implementation 

It would be great to have a safe guard to prevent users from doing something like this again. Sorry if this is meant to be an issue in Bug report

### LINKED ISSUE #4684: PublishDir step fails on push to S3 even though pipeline succeeds
## Bug report 

Reported by a user.

I recently observed a run on Nextflow Tower that was reported to be completed successfully on platform, but the publish directory on S3 was empty. 
The corresponding directory in the working directory was fully populated with all expected outputs.

Log file showed the following:
```
Nov-16 07:14:35.678 [PublishDir-1] WARN  nextflow.processor.PublishDir - Failed to publish file: s3://s3-path/scratch/2NQLCVMB9up84P/dd/be6eea71f1d896340271b30c751e39/results; to: s3://s3-path/users/kkosheleva/tower/EXP23002963/MAX078/results [copy] -- See log file for details
java.nio.file.NoSuchFileException: the path: /s3-path/users/kkosheleva/tower/EXP23002963/MAX078/results/variants not exists
    at com.upplication.s3fs.S3FileSystemProvider.delete(S3FileSystemProvider.java:561)
    at java.base/java.nio.file.Files.delete(Files.java:1152)
    at nextflow.file.FileHelper$3.postVisitDirectory(FileHelper.groovy:1007)
    at nextflow.file.FileHelper$3.postVisitDirectory(FileHelper.groovy)
    at java.base/java.nio.file.Files.walkFileTree(Files.java:2828)
    at java.base/java.nio.file.Files.walkFileTree(Files.java:2882)
    at nextflow.file.FileHelper.deleteDir0(FileHelper.groovy:999)
    at nextflow.file.FileHelper.deletePath(FileHelper.groovy:991)
    at nextflow.processor.PublishDir.processFile(PublishDir.groovy:390)
    at nextflow.processor.PublishDir.safeProcessFile(PublishDir.groovy:359)
    at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
    at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.base/java.lang.reflect.Method.invoke(Method.java:568)
    at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:107)
    at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:323)
    at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1258)
    at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1035)
    at org.codehaus.groovy.runtime.InvokerHelper.invokePogoMethod(InvokerHelper.java:1036)
    at org.codehaus.groovy.runtime.InvokerHelper.invokeMethod(InvokerHelper.java:1019)
    at org.codehaus.groovy.runtime.InvokerHelper.invokeMethodSafe(InvokerHelper.java:97)
    at nextflow.processor.PublishDir$_apply1_closure1.doCall(PublishDir.groovy:332)
    at nextflow.processor.PublishDir$_apply1_closure1.call(PublishDir.groovy)
    at groovy.lang.Closure.run(Closure.java:493)
    at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
    at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
    at java.base/java.lang.Thread.run(Thread.java:833)
Nov-16 07:14:35.679 [main] DEBUG nextflow.util.ThreadPoolManager - Thread pool 'PublishDir' shutdown completed (hard=false)
```

bentsherman:
Close #3372 

I just changed the publish dir to explicitly publish each file in a directory. I think it means that if the publish mode is symlink, each file in the directory will be symlinked instead of just the directory.

@robsyme do we just need to add a checksum comparison to S3 copies and uploads? I think I can just rip the code from this PR #3802


bentsherman:
This PR is ready for review. The publish dir now publishes each file in a directory instead of just the directory, which should improve detection of Reports in Seqera Platform and fix some issues with Fusion symlinks (#4725).

It does not yet solve #3372 , for that we also need to verify the checksum of published files when deciding whether a file needs to be re-published. I can implement that here or in a separate PR, up to you @pditommaso .

robsyme:
Looks good, thanks Ben!

In the case where publish method is symlinks and we are emitting/publishing a directory, what does the published output look like? 

Is it a symlinked directory full of symlinks, or a regular directory full of symlinks, or something else entirely?

bentsherman:
It will be a regular directory of symlinks

pditommaso (commented):
Not sure we should do this. On some storage traversing a directory path is extremely expensive e.g. Lustre, NFS, etc.

bentsherman:
I share your concern, although copying a directory also involves walking the directory tree at some level, so may it's not an issue. Either way, there are other reasons we should do this.

In general, I think it is a bad practice to define a directory output. When I see a process with a directory output, I know nothing about the file structure within that directory.

So I think we should either not support directory outputs or support them correctly. Currently we just support it incorrectly:
- Tower reports in a directory output are not detected
- Fusion symlinks in a directory output are not resolved
- If we implemented #4729 without this PR, the validation of published outputs would not work for directory outputs

The performance implications you mention only happen if the user decides to use directory outputs.. I think we could simply document this point and encourage users to use flat files as much as possible.

Consider also that we already walk the work directory to collect the output files:

https://github.com/nextflow-io/nextflow/blob/279c410b8de69627dbb7992e37e98ad36883c180/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy#L1654

which has similar performance issues when there are nested directories

bentsherman:
Indeed, when we publish a directory we are already traversing it:

https://github.com/nextflow-io/nextflow/blob/279c410b8de69627dbb7992e37e98ad36883c180/modules/nf-commons/src/main/nextflow/file/FileHelper.groovy#L948

This PR is just doing the traversal earlier

bentsherman:
@robsyme what do you think about what I said about publishing a symlink directory? It is certainly a change in behavior, but I'm wondering if it's more desirable to symlink all of the files rather than only symlink the directory.

If not, I can modify the logic to only traverse the directory when the publish mode isn't symlink, but I feel like that would make the behavior inconsistent based on the publish mode.

robsyme:
Thanks for moving this forward Ben, and apologies for my slow response times! My feeling is that we're stuck between a rock and a hard place, but inconsistencies between publishDir modes is the worse outcome.

As for
> bad practice to define a directory output

I think it's inevitable that people will want to publish results that contain directories (many tools have complex/nested outputs).

bentsherman:
Indeed, even though a directory output is harder to reason about, it seems to be convenient for the first iteration of a Nextflow process for certain tools

pditommaso:
I think this is solved by #5005

bentsherman:
The fusion symlinks are no longer an issue, but there are other reasons to merge this:

- when the workflow output definition uses the `deep` overwrite mode, which compares checksums to decide whether to overwrite a file, comparing each individual file instead of the entire directory would allow for finer-grained caching

- walking the directory emits a publish event for each individual file, which makes them easier to capture for reports

R3myG:
@bentsherman What's the status of this fix? Encountering is systematically with the https://github.com/nf-core/spatialvi pipeline. Alternatively, any known workaround?

bentsherman:
The PR works but I'm not sure if we still want to merge it or not. We might take a different approach.

I'm not sure why your issue is happening because you don't need to create directories beforehand in S3. It could be a different bug in Nextflow. Feel free to submit an issue here if you aren't able to solve it.

A workaround to publishing a directory is to have a tuple of paths for each individual file in the directory. But this gets very impractical if there are many files

PR #3938, opened 2023-05-11
Fixing bug 2428 for errorStrategy ignore

mes5k:
This fix simply handles PodUnschedulableException in the same way that NodeTerminationException is already handled. The synthetic state object updates the task correctly and the failing pod (for any PodUnschedulableException reason) will be ignored, but the task will still be cleaned up.


pditommaso (approved):

pditommaso:
Thanks Mike!

Issue #3567, opened 2023-01-22
docs on converting to/from MemoryUnit (and similar)


nick-youngblut:
## New feature

As far as I can tell, the [Nextflow docs](https://www.nextflow.io/docs/latest/index.html) do not contain any information about converting to/from the `nextflow.util.MemoryUnit` (or similar Nextflow classes). It would be helpful if some official docs were included. 

## Usage scenario 

Is this a good way of converting a `task.memory` value to an integer:

```
task.memory.toString().replaceFirst(/ .+/, '').toInteger()
```

For another example: What is the best way of converting a `file.size()` value (bytes; integer) to a `MemoryUnit`? This is useful to setting the memory for the process job (e.g., set memory required to the size of the reference database file).

***

I would submit a PR, but I do not know the best-practices for such scenarios.

bentsherman:
We will make sure to document the MemoryUnit API as part of the ongoing docs overhaul.

In the meantime, you can extract the numeric value of a MemoryUnit using methods like `toGiga()`, `toMega()`, and `toBytes()`, and you can create a MemoryUnit from bytes value using `MemoryUnit.of(file.size())`

Source: https://github.com/nextflow-io/nextflow/blob/master/modules/nf-commons/src/main/nextflow/util/MemoryUnit.groovy

PR #3953, opened 2023-05-16
Improve documentation on Nextflow scripts

bentsherman:
This PR makes a few key improvements to the "Nextflow scripting" docs page, in preparation for the website / docs overhaul.

- Add sections on built-in functions and classes
- Move sections on regexes and closures into the Groovy section
- Consolidate several subsections under Files & I/O into a section on "filesystem operations"
- Convert tables to definition lists (as we have done throughout the docs) to be more concise and readable

### DIFF
--- a/docs/script.md
+++ b/docs/script.md
@@ -554,21 +554,7 @@ The following functions are available in Nextflow scripts:
 
 ## Implicit classes
 
-The following classes are available in Nextflow scripts:
-
-`Channel`
-: Provides channel factory methods. See {ref}`channel-factory`.
-
-`Duration`
-: Represents a duration. See the {ref}`process-time` process directive.
-
-`MemoryUnit`
-: Represents a quantity of bytes. See the {ref}`process-memory` process directive.
-
-`ValueObject`
-: An AST transformation for classes and enums, which simply combines [AutoClone](http://docs.groovy-lang.org/latest/html/gapi/groovy/transform/AutoClone.html) and [Immutable](https://docs.groovy-lang.org/latest/html/gapi/groovy/transform/Immutable.html).
-
-Additionally, the following classes are imported from standard libraries:
+The following classes are imported by default in Nextflow scripts:
 
 - `java.lang.*`
 - `java.util.*`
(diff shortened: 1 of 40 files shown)


docs/script.md
```diff
@@ -320,281 +502,258 @@ In the above snippet the `task.cpus` holds the value for the {ref}`cpus directiv
...
 
-```groovy
-myMap = ["China": 1 , "India" : 2, "USA" : 3]
+`Duration`
+: Represents a duration. See the {ref}`process-time` process directive.
 
-result = 0
-myMap.keySet().each( { result+= myMap[it] } )
+`MemoryUnit`
+: Represents a quantity of bytes. See the {ref}`process-memory` process directive.
```
bentsherman, line 569: Eventually, would be good to document these classes more thoroughly, and include the unit suffix tables with them, rather than with the time / memory directives. But for now this is fine.

docs/script.md
```diff
@@ -320,281 +502,258 @@ In the above snippet the `task.cpus` holds the value for the {ref}`cpus directiv
...
-By default, closures take a single parameter called `it`, but you can also create closures with multiple, custom-named parameters. For example, the method `Map.each()` can take a closure with two arguments, to which it binds the `key` and the associated `value` for each key-value pair in the `Map`. Here, we use the obvious variable names `key` and `value` in our closure:
+`sendMail( params )`
+: Send an email. See {ref}`mail-page`.
 
-```groovy
-printMapClosure = { key, value ->
-    println "$key = $value"
-}
+`stop()`
+: Interrupts the iteration when using split operators.
```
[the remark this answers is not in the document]
pditommaso: I not even remember. I'd remove it

docs/script.md
pditommaso: Not sure, these were meant to be internal class only

bentsherman: Indeed, users should never need to reference `MemoryUnit` directly, but some of its methods, like `toGiga()` and `toMega()`, are useful. Same thing with `Duration`. But I'm not sure yet what is the best way to document that.

bentsherman: This issue is relevant: #3567

bentsherman: Okay I expanded these sections in a way that makes sense to me. Let me know what you think.

bentsherman:
@pditommaso can this be merged?

docs/script.md
```diff
@@ -320,290 +502,348 @@ In the above snippet the `task.cpus` holds the value for the {ref}`cpus directiv
...
+The following functions are available in Nextflow scripts:
 
-```groovy
-square = { it * it }
-```
+`branchCriteria( closure )`
+: Create a branch criteria to use with the {ref}`operator-branch` operator.
 
-The curly brackets around the expression `it * it` tells the script interpreter to treat this expression as code. The `it` identifier is an implicit variable that represents the value that is passed to the function when it is invoked.
+`cacheableDir( key )`
```
pditommaso, line 514: This is a legacy API do be removed

docs/script.md
```diff
@@ -320,290 +502,348 @@ In the above snippet the `task.cpus` holds the value for the {ref}`cpus directiv
...
+: Create a unique folder for the given key. The same folder name is always returned for the same key.
 
-Once compiled the function object is assigned to the variable `square` as any other variable assignments shown previously. Now we can do something like this:
+`cacheableFile( key, name = null )`
```
pditommaso, line 517: This is a legacy API do be removed

docs/script.md
```diff
@@ -320,290 +502,348 @@ In the above snippet the `task.cpus` holds the value for the {ref}`cpus directiv
...
+: Create a unique file name for the given key. The same file name is always returned for the same key. If `name` is not specified, it will be generated from the key.
 
-```groovy
-println square(9)
-```
+`error( message = null )`
+: Throw a script runtime error with an optional error message.
 
-and get the value 81.
+`exit( exitCode = 0, message = null )`
```
pditommaso, line 523: Should be deprecated because it uses `System.exit`. Better to not document it

docs/script.md
```diff
@@ -320,290 +502,348 @@ In the above snippet the `task.cpus` holds the value for the {ref}`cpus directiv
...
+: Stop the pipeline execution and return an exit code and optional error message.
 
-This is not very interesting until we find that we can pass the function `square` as an argument to other functions or methods. Some built-in functions take a function like this as an argument. One example is the `collect` method on lists:
+`fasta( obj )`
```
pditommaso, line 526: This has always been an experiment but I think it should be removed and only keep at channel level because it can create a blocking point in the script execution

docs/script.md
```diff
@@ -320,290 +502,348 @@ In the above snippet the `task.cpus` holds the value for the {ref}`cpus directiv
...
+: Create a FASTA splitter for an object.
 
-```groovy
-[ 1, 2, 3, 4 ].collect(square)
-```
+`fastq( obj )`
```
pditommaso, line 529: Same as `fasta`

pditommaso (changes requested):

bentsherman:
Thanks @pditommaso , I removed the legacy methods from the docs and marked them as deprecated in the code. Since I think some people still use `exit()`, I will leave it in the docs with a deprecation warning so that we get them to move away from it.

ewels:
It'd be good to have some explicit docs about how to properly exit. It comes up all the time and I can never remember the details. Would be really useful to have a docs heading to link to.

pditommaso:
Think there's [error](https://github.com/nextflow-io/nextflow/blob/d3788f9f41ef783a06573d11cc6870c6d20fbb19/modules/nextflow/src/main/groovy/nextflow/Nextflow.groovy#L223-L225) for it

bentsherman:
Once the changes from this PR make it into the docs, you can link people to the implicit functions: https://nextflow.io/docs/latest/script.html#implicit-functions

ewels:
Yup 👍🏻  Do you think it's worth throwing in an extra warning admonition there, explaining why `exit()` and `System.exit()` are bad? Or would that be overkill?

PR #3956, opened 2023-05-16
Remove `--no-home` default option from charliecloud builder

### LINKED ISSUE #3951: Failure to execute ch-run with current charliecloud version (0.32).

Failure to execute ch-run with current charliecloud version (0.32).

Given a regular call: 
```nextflow run -without-docker -with-charliecloud ontresearch/wf-pore-c:sha51e5a1f25db6981a781851783e93efd05f8ef266 -ansi-log ...```

### Program output 

Yields:
```
Command exit status:
  64

Command output:
  (empty)

Command error:
  ch-run: unrecognized option '--no-home'
  Try `ch-run --help' or `ch-run --usage' for more information.
```

### Temporary kludge

A ch-run wrapper in the PATH that removes the '--no-home' from the argument list. The workflow completes like charm :)

### Environment 

* Nextflow version: 23.04.1
* Java version: openjdk-17.0.2.8
* Operating system: Linux
* Bash version: 5.1.4

### Additional context

The change was introduced by https://github.com/hpc/charliecloud/pull/1536

Without having tried, it looks to me that `--no-home` should be removed in `CharliecloudBuilder.groovy`, when running current and future versions of charliecloud.

Best,
  Thomas

bentsherman:
Close #3951 

In Charliecloud v0.32, the `--no-home` option was removed. It has been the default since version v0.31. Since Nextflow always adds this option, Nextflow won't work with Charliecloud v0.32 or later.

This PR removes the `--no-home` option from the builder. Users can preserve the old behavior by adding `--no-home` to `charliecloud.runOptions` or `process.containerOptions`.

I can add a note to the container docs as well if you want.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/container/CharliecloudBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/CharliecloudBuilder.groovy
@@ -61,7 +61,7 @@ class CharliecloudBuilder extends ContainerBuilder<CharliecloudBuilder> {
     CharliecloudBuilder build(StringBuilder result) {
         assert image
 
-        result << 'ch-run --unset-env="*" -c "$PWD" --no-home --set-env '
+        result << 'ch-run --unset-env="*" -c "$PWD" --set-env '
         if (!readOnlyInputs)
             result << '-w '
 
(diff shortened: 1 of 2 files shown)


pditommaso:
Adding @l-modolo and @phue for review

l-modolo (commented):
The `--no-home` option is now the default in the recent version of charliecloud, v0.32 or later. Thus keep it will throw an error for those versions.

The corresponding code and test are correctly updated.

l-modolo (approved):

PR #3958, opened 2023-05-18
Use full map of valid options for JsonSplitter creation

robsyme:
At the moment, the JsonSplitter abandons the list of valid options from `super`, leading to errors when splitting any Json string as noted by @mribeirodantas.

This fix ensures that the inherited valid methods are preserved.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy
@@ -70,10 +70,8 @@ class JsonSplitter extends AbstractTextSplitter {
     @Override
     protected Map<String,?> validOptions() {
         def baseOptions = super.validOptions()
-        return [
-            path: String,
-            limit: baseOptions.limit
-        ]
+        baseOptions.path = String
+        return baseOptions
     }
 
     @Override
--- a/modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy
@@ -69,7 +69,7 @@ class JsonSplitter extends AbstractTextSplitter {
      */
     @Override
     protected Map<String,?> validOptions() {
-        def baseOptions = super.validOptions()
+        final baseOptions = super.validOptions()
         return [
             autoClose: baseOptions.autoClose,
             into: baseOptions.into,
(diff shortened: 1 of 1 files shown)


modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy
```diff
@@ -70,10 +70,8 @@ class JsonSplitter extends AbstractTextSplitter {
     @Override
     protected Map<String,?> validOptions() {
         def baseOptions = super.validOptions()
-        return [
-            path: String,
-            limit: baseOptions.limit
-        ]
```
pditommaso, line 76: I think this has been done on purpose because `splitJson` does only support `path`. What cause the error?

robsyme: Try running `main.nf`:
```groovy
Channel.of('[1,null,["A",{}],true]')
| splitJson()
| view
```
with
```shell
NXF_VER=23.05.0-edge nextflow run .
```

Returns

```
N E X T F L O W  ~  version 23.05.0-edge
Launching `./main.nf` [lethal_kay] DSL2 - revision: 2bdd9780c3
ERROR ~ Unknown argument 'into' for operator 'Json' -- Possible arguments: path, limit

 -- Check script 'main.nf' at line: 2 or see '.nextflow.log' file for more details
```

Perhaps we just need to add `into` and `autoClose`:

```groovy
 return [
    into: baseOptions.into,
    autoClose: baseOptions.autoClose,
    path: String,
    limit: baseOptions.limit
]
```

mribeirodantas: I tried multiple examples added to the docs in the PR that created the channel operator. None of the ones I tried works 😢

modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy
```diff
@@ -71,8 +71,10 @@ class JsonSplitter extends AbstractTextSplitter {
     protected Map<String,?> validOptions() {
         def baseOptions = super.validOptions()
         return [
-            path: String,
+            autoClose: baseOptions.autoClose,
+            into: baseOptions.into,
             limit: baseOptions.limit
+            path: String,
```
bentsherman, line 77:
```suggestion
            autoClose: baseOptions.autoClose,
            into: baseOptions.into,
            limit: baseOptions.limit,
            path: String,
```

bentsherman (approved):

modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy
```diff
@@ -71,8 +71,10 @@ class JsonSplitter extends AbstractTextSplitter {
     protected Map<String,?> validOptions() {
         def baseOptions = super.validOptions()
```
pditommaso, line 72:
```suggestion
        final baseOptions = super.validOptions()
```

PR #3959, opened 2023-05-18
Add doc tests, move some snippets to separate files

bentsherman:
Summary of changes:

- Delete unused doctests
- Move testable snippets from docs to separate files
- Add CI test for doc snippets
- Fix another bug in JSON splitter
- Tighten up language in operator docs
- Add section about syntax sugar, remove random tips about syntax sugar elsewhere

For now it's mostly just the operator examples since they are complete snippets and don't require external files. Snippets that have an expected output are also verified against that output. Since items can be emitted in arbitrary order, we sort the outputs before comparing.

I also added some notes to clarify the differences between various clusters of similar operators:
- `collect`, `toList`, `toSortedList`
- `combine`, `cross`, `join`
- `distinct` and `unique`
- `groupTuple`, `transpose`

Where appropriate, I try to use the same example across similar operators so that the user can easily see the difference in their output.


modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy
```diff
@@ -217,9 +217,9 @@ class JsonSplitter extends AbstractTextSplitter {
                 break
             }
 
-            final map = Map.of(
-                 OBJECT_KEY, reader.nextName(),
-                 OBJECT_VALUE, fromJson(reader))
+            def map = new LinkedHashMap<>(1)
+            map[OBJECT_KEY] = reader.nextName()
+            map[OBJECT_VALUE] = fromJson(reader)
```
bentsherman, line 222: This change fixed a weird bug where the JSON splitter would fail when parsing an object with a null value in it. It also makes the key order consistent which is good for the CI test 😄

bentsherman:
This PR got quite out of hand... I started out with a specific goal and ended up making a ton of changes to the operator docs that I've been wanting to do.

To ease the reviewing process, I have moved all unrelated changes to a separate branch. Now this PR only contains:
- Move operator snippets to separate files
- Add integration test for doc snippets
- Fix bug in splitJson operator (which is revealed only by the doc tests)

I would like to move this PR forward.

modules/nextflow/src/main/groovy/nextflow/splitter/JsonSplitter.groovy
pditommaso: Ok, better to keep in its own commit for history purposes 👉 8ec14dd2

PR #3960, opened 2023-05-19
Patch Google batch instance types

pditommaso:
This PR adds default machine types when local SSD is not required

### DIFF
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy
@@ -71,6 +71,8 @@ class GoogleBatchMachineTypeSelector {
      */
     private static final List<String> DEFAULT_FAMILIES_WITH_SSD = ['n1-*', 'n2-*', 'n2d-*', 'c2-*', 'c2d-*', 'm3-*']
 
+    private static final List<String> DEFAULT_FAMILIES_WITH_HDD = ['n1-*', 'n2-*', 'n2d-*', 'c2-*', 'c2d-*', 'm1-*', 'e2-*']
+
     @Immutable
     static class MachineType {
         String type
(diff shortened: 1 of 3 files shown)


plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy
```diff
@@ -71,6 +71,8 @@ class GoogleBatchMachineTypeSelector {
      */
     private static final List<String> DEFAULT_FAMILIES_WITH_SSD = ['n1-*', 'n2-*', 'n2d-*', 'c2-*', 'c2d-*', 'm3-*']
 
+    private static final List<String> DEFAULT_FAMILIES_WITH_HDD = ['n1-*', 'n2-*', 'n2d-*', 'c2-*', 'c2d-*', 'm1-*', 'e2-*']
```
jordeu, line 74: My only doubt here if we also add `m2` and `m3` that are types with very big memories
https://cloud.google.com/compute/docs/memory-optimized-machines#m2_machine_types

jordeu: Also these types are not "HDD" types. The difference is that there are some types that allow to mount a local SSD disk and others that only allow network disks. So maybe better naming can be here `DEFAULT_FAMILIES` and for Fusion `DEFAULT_FAMILIES_WITH_LOCAL_SSD`

jordeu (commented):
I was also doing this PR and I included in the same another fix, just to remove these two lines:
https://github.com/nextflow-io/nextflow/blob/f48a473c070f297bce6f97f6b076e4e92d25e00a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy#L231-L232

Do you want me to add this also here or in a different PR?

pditommaso:
> Do you want me to add this also here or in a different PR?

Yes, please

PR #3961, opened 2023-05-19
Fix invalid machine type setting when no valid machine type is found

jordeu:
## Description

When the method `findBestMachineType` fails to find a best candidate the `machineType` provided by the user should only be used if it is a correct machine type, and not if it's a list of families that can contain a wildcard. In that case, `machineType` should be empty so Google Batch will decide a custom type depending on the CPU and Memory allocated.


pditommaso:
All failed!

pditommaso (approved):

PR #3969, opened 2023-05-25
Add shared storage support for AWS Batch executor

shahim:
## Description

AWS Batch integration always assumes that the workDir is an S3 bucket so by default uses AwsBatchFileCopyStrategy. This PR add a check of schema of workDir to use SimpleFileCopyStrategy if the schema is not s3. Therefor shared storage can be used with AWS Batch integration.

### DIFF
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy
@@ -110,16 +110,6 @@ class AwsBatchExecutor extends Executor implements ExtensionPoint {
         session.bucketDir ?: session.workDir
     }
 
-    protected void validateWorkDir() {
-        /*
-         * make sure the work dir is a S3 bucket
-         */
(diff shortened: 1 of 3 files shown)


bentsherman:
I think local work directories are not supported for the AWS Batch executor because that is provided by the [xpack-amzn](https://github.com/seqeralabs/xpack-amzn) plugin.

shahim:
@pditommaso  

Hi Paolo, Regarding local project bin/ directory, I managed to get it work by setting NXF_ASSETS to the shared storage so containers can access binaries in the project bin path. 

I appreciate that this feature is currently available through the enterprise extension, and I would like to suggest the possibility of incorporating it into the open-source project as well. Including this feature in the open-source version would be beneficial to the community, considering that the primary value of the enterprise extension lies in the enterprise support it offers.

Thanks,
Kambiz

pditommaso:
We consider shared file systems a legacy technology and our plan is to privileged Fusion file system over them. 

Consider using [Fusion](https://www.nextflow.io/docs/latest/fusion.html) instead, it's as fast as FSx with no extra cost other than S3.

PR #3990, opened 2023-05-31
feat: add remote bin support for TES in a workdir

jherr-dev:
Adds support for using remote bin with TES in a shared file system workdir. The shared file system mount on the server running the pipeline needs to be the same the mount in the image, so a mount off root is easiest. Code is based on the AWS Batch plugin's remote bin support. 

Also sets the TES name to the task name.

@pditommaso I see you have done all the updates in the plugin repository for the official plugins, do you want/need to handle that or should I create a pr there to go along with this one?

### DIFF
--- a/plugins/nf-ga4gh/src/main/nextflow/ga4gh/tes/executor/TesExecutor.groovy
+++ b/plugins/nf-ga4gh/src/main/nextflow/ga4gh/tes/executor/TesExecutor.groovy
@@ -55,14 +55,8 @@ class TesExecutor extends Executor implements ExtensionPoint {
 
     @Override
     protected void register() {
-        if( session.binDir && !session.binDir.empty() && !session.workDir ) {
-            session.abort()
-            throw new AbortOperationException("ERROR: TES executor does not allow the use of custom scripts in the `bin` folder without a workdir")
-        } else {
-            uploadBinDir()
-        }
-
         super.register()
+        uploadBinDir()
 
         client = new TaskServiceApi( new ApiClient(basePath: getEndPoint()) )
     }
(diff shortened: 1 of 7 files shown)


plugins/nf-ga4gh/src/main/nextflow/ga4gh/tes/executor/TesExecutor.groovy
```diff
@@ -44,11 +48,18 @@ class TesExecutor extends Executor implements ExtensionPoint {
...
+     */
+    private Path remoteBinDir = null
+
     @Override
     protected void register() {
-        if( session.binDir && !session.binDir.empty() ) {
+        if( session.binDir && !session.binDir.empty() && !session.workDir ) {
             session.abort()
-            throw new AbortOperationException("ERROR: TES executor does not allow the use of custom scripts in the `bin` folder")
+            throw new AbortOperationException("ERROR: TES executor does not allow the use of custom scripts in the `bin` folder without a workdir")
```
bentsherman, line 60: You can remove this check entirely, the work directory is always defined.

bentsherman: Or if you're saying that it only works with local/shared filesystem, you can check that with `session.workDir.fileSystem == FileSystems.default`. That condition will be false when the work directory is object storage like S3.

bentsherman:
Thanks for working on this. Have you tested it with your own TES backend? Also, could you add a unit test to `TesTaskHandlerTest` ?

jherr-dev:
> Thanks for working on this. Have you tested it with your own TES backend? Also, could you add a unit test to `TesTaskHandlerTest` ?

Yeah I've tested and it works with my TES backend. I've added a test for the submit function.

pditommaso (approved):

PR #3992, opened 2023-06-01
Fallback to anonymous credentials in S3 client

### LINKED ISSUE #3989: nf-amazon 2.0.0: Unable to load AWS credentials when running on Github Actions
## Bug report 

When running on Github actions (and probably some other systems that lack AWS credentials) Nextflow reports the error:
```
Downloading plugin nf-amazon@2.0.0
ERROR ~ Unable to load AWS credentials from any provider in the chain: [EnvironmentVariableCredentialsProvider: Unable to load AWS credentials from environment variables (AWS_ACCESS_KEY_ID (or AWS_ACCESS_KEY) and AWS_SECRET_KEY (or AWS_SECRET_ACCESS_KEY)), SystemPropertiesCredentialsProvider: Unable to load AWS credentials from Java system properties (aws.accessKeyId and aws.secretKey), WebIdentityTokenCredentialsProvider: You must specify a value for roleArn and roleSessionName, com.amazonaws.auth.profile.ProfileCredentialsProvider@c2dab10: profile file cannot be null, com.amazonaws.auth.EC2ContainerCredentialsProviderWrapper@32029cd: The requested metadata is not found at http://169.254.169.254/latest/meta-data/iam/security-credentials/]
```

Full run details:
https://github.com/nf-core/smrnaseq/actions/runs/5134355097/jobs/9238248639?pr=254 

The data is a public source so it should not require credentials.

Because this was a Github action, we can go back in history and see the same code working fine a few weeks ago. Here is a working version:
https://github.com/nf-core/smrnaseq/actions/runs/4958065849/jobs/8870496729

The main delta I can see is the working version was:
Nextflow v22.10.1 and nf-amazon v1.11.0
Nextflow v23.04.1 and nf-amazon v1.16.2

vs 

Nextflow v23.05.0-edge and nf-amazon v2.0.0

I'm trying some combinations [here](https://github.com/nf-core/smrnaseq/pull/256). You can see some combinations aren't compatible which makes this fairly tricky to get right by just selecting versions: https://github.com/nf-core/smrnaseq/actions/runs/5135450691/jobs/9240888639

Tests should pull data from AWS without credentials and not throw an error.

https://github.com/nf-core/smrnaseq/pull/256

### Program output 

To follow.

### Environment 

* Nextflow version: Various (22.10.1, 23.04.1, 23.05.0-edge, multiple nf-amazon versions).
* Java version: [?]

### Additional context

bentsherman:
Close #3989 

Based on the [default S3 credentials provider](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/S3CredentialsProviderChain.java#L29), which uses the default provider chain with a fallback to anonymous credentials. Since Nextflow doesn't use the default provider chain anymore, I think we have to add this fallback ourselves.


pditommaso:
Nice point. However, I'm not understanding why it should happen. when the anonymous is specified this branch should be taken that uses the `AnonymousAWSCredentials` in explicit manner 

https://github.com/nextflow-io/nextflow/blob/1daebeef025f45b8e9e5cab62e783ddb3327118f/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3FileSystemProvider.java#L838-L839


Also worth adding an integration tests in the `tests/` folder

bentsherman:
I guess it's because the `anonymous` property isn't set:

https://github.com/nextflow-io/nextflow/blob/1daebeef025f45b8e9e5cab62e783ddb3327118f/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3FileSystemProvider.java#L836-L840

As for the e2e test, how would we negate the CI runner's AWS creds before running the test?

bentsherman:
I see, you're talking about `aws.client.anonymous` in the Nextflow config. But if you enable that option, it looks like you would only be able to access public buckets. What if you want to access public and private buckets in the same run?

bentsherman:
Since I'll be out, if you can get it to work, a simple e2e test would be:
```groovy
println file('s3://ngi-igenomes/igenomes/Homo_sapiens/Ensembl/GRCh37/Annotation/README.txt').text
```

pditommaso:
Ok, added the integration tests and refactored the anonymous class to `S3CredentialsProvider` class

bentsherman:
Where is the integration test? I don't see it in the PR

pditommaso:
You are right. It went lost 👉 680f0a90

bentsherman:
@pditommaso should we deprecate the `aws.client.anonymous` config option? It seems useless if Nextflow can fallback to anonymous creds anyway.

pditommaso:
I still can be useful to avoid the normal auth flow when using non-aws cloud, I guess

PR #3998, opened 2023-06-05
Enhanced support for Spack + wave

pditommaso:
Enhance the support for Spack mapping Spack packages to a Spack yaml file


marcodelapierre:
now going to work on the mirror/buildcache edits

pditommaso:
This is ready. Main changes: 

1. Add support for Spack binary cache on Wave 
2. Simplifies Spack options
   - `basePackages`: One or more packages to be added to all Spack based images (in the same manner as it's done for Conda_
   - `baseImage`: The base image used for building the Spack container (for consistency with the same option for Conda)
   - `commands`: One or more Dockerfile `RUN`commands to customise the resulting container 
3. The following variables are resolved by the Wave backend 
   - `builder_image`
   - `spack_cache_path`
   - `spack_key_file`

pditommaso:
@marcodelapierre is this ok on your side? I'd like to merge this and include in upcoming 23.06.0-edge

marcodelapierre:
Still on leave (travelling) today, I recommend holding until tomorrow COB, so that I can check a couple of things.

Then, in my to-do list there would still be adding the code lines to populate the cache, and test them with the updated backend. However, this bit could wait as well until the next release.

pditommaso:
Ok, let's talk tomorrow. Just to reply to your point, the cache population has been added (see it below) and tested with the Wave backend 

https://github.com/nextflow-io/nextflow/blob/b7c6abbcf3e2498ff8337fd2467f1bc857965c79/plugins/nf-wave/src/resources/templates/spack/dockerfile-spack-file.txt#L24-L30

marcodelapierre:
Reminder to myself to also review this one:

https://github.com/seqeralabs/wave/pull/249

pditommaso:
Considering the problem updating the local cache, I've removed the spack `push` command and added the public cache mirror 👉 f96d64337ed1a596a56517fe79ab292feeb33

Also, I think this part should be moved to the backend to better control it.

marcodelapierre:
TO DO: 
* a couple of pipeline test
* review Paolo's latest commits

@pditommaso had an unexpected, will finalise review of your work tomorrow (Wednesday)

marcodelapierre:
```

Command error:
  Unable to find image 'wave.seqera.io/wt/dd174cbd2609/wave/build:8469ca1f8a0fe8a68f1173753bce999c' locally
  docker: Error response from daemon: unknown: INFO[0000] Resolved base name {{builder_image}} to builder 
  ERRO[0000] Error while retrieving image from cache: {{builder_image}} could not parse reference: {{builder_image}} 
  INFO[0000] Retrieving image manifest {{builder_image}}  
  error building image: could not parse reference: {{builder_image}}.
  See 'docker run --help'.
```

pditommaso:
You need to use the testing version. Put in the config 

```
wave.endpoint = 'https://wave.stage-tower.net'
tower.endpoint = 'https://api.stage-tower.net'
```

pditommaso:
I've moved forward moving the Spakc builder part of the Dockerfile to the backend. See [fcb4a92](https://github.com/nextflow-io/nextflow/pull/3998/commits/fcb4a9210380aaf3ee8b1775e310aa44c5dafa46)

marcodelapierre:
Ok, I am quite happy with the testing and review:
- trialled spack wave with both yaml file and list of specs: see comment below
- reviewed my changes: OK
- reviewed most of your changes: see inline comment in code

I have noticed a funny behaviour, depending on whether I use 1. or 2. in the config snippet below:
```
  process.spack = 'cowsay nano ^ncurses abi=5'   //    1

  process.spack = 'spack.yaml'    //   2
```

With 1, a run of 8 nextflow processes generates 8 distinct container tags (all pointing to the same image of course);
with 2, the processes all use the same container tag.

What may be making the hashes different in the 1st case? It would be nice to fix it.

Note: TO BE UPDATED

plugins/nf-wave/src/main/io/seqera/wave/config/SpackOpts.java
```diff
@@ -27,32 +27,33 @@
...
-    public final String builderImage;
-    public final String runnerImage;
-    public final String osPackages;
-    public final String cFlags;
-    public final String cxxFlags;
-    public final String fFlags;
+    /**
+     * The default base image from which the resulting container is created
+     */
+    final static public String DEFAULT_SPACK_BASE_IMAGE = "ubuntu:22.04";
```
marcodelapierre, line 33: I have some thoughts here on the original builder vs runner image:
```
    final static public String DEFAULT_SPACK_BUILDER_IMAGE = "spack/ubuntu-jammy:v0.20.0";
    final static public String DEFAULT_SPACK_RUNNER_IMAGE = "ubuntu:22.04";
```

1. Note how the OS (type+version) should be consistent between the two. Hence, if the builder images cannot be chosen, the runner one should not be customisable neither (or viceversa, they should be flexible).

2. Depending on the model which is adopted for growing the binary cache, it might be good to have an option to choose the Spack version, i.e. the tag of the builder image. The scenario which I am thinking about is one where we advertise the ability for endusers to add new Spack recipes and use them promptly with Wave. In this case, the Spack version that needs to be used would not be the release one, but the next immediate weekly develop snapshot version, which will contain the newly added recipes. Or in alternative: we clearly state that only recipes available in the current release version of Spack are supported. 

I think this is an important point for the overall design of the service.

What do you think?

pditommaso:
> With 1, a run of 8 nextflow processes generates 8 distinct container tags (all pointing to the same image of course);

nice spot, yes, should be caused by the hashing of the temporary file

plugins/nf-wave/src/main/io/seqera/wave/config/SpackOpts.java
pditommaso: Are you suggesting the runner image should be defined on backend side?

pditommaso: Ok, i've moved it to the backend

marcodelapierre:
@pditommaso the only outstanding comment from the ones I made:

Depending on the model which is adopted for growing the binary cache, it might be good to have an option to choose the Spack version, i.e. the tag of the builder image. The scenario which I am thinking about is one where we advertise the ability for endusers to add new Spack recipes and use them promptly with Wave. In this case, the Spack version that needs to be used would not be the release one, but the next immediate weekly develop snapshot version, which will contain the newly added recipes. Or in alternative: we clearly state that only recipes available in the current release version of Spack are supported.

What do you think?

pditommaso:
We can review this at a later stage. If a concrete need arises we can add an option to choose it. Also, for this reason, I preferred to remove the ability to choose the running container image and keep the selection at the backend for now.

marcodelapierre:
Makes sense to me!

marcodelapierre:
Last test OK:

image tags are now unique (1 tag for any number of processes) for the new `master` branch, as in:
- conda yaml
- conda spec
- spack yaml
- spack spec

PR #4010, opened 2023-06-09
Update `workflow.revision` max length

tcrespog:
Since Tower 22.4.0, the max length for the `workflow.revision` field is 100: update schema on Nextflow side so the field does not get truncated to 40 (the previous limit) before sending it to Tower.


pditommaso (approved):
Cool

Issue #4024, opened 2023-06-14
Allow closure type as input in PublishDir.setMode()


dwishsan:
Hi everyone !

First of all, a big THANK YOU for nextflow, it is very handy :)

## Some context on my feature request
I am using nextflow with configuration files to customize processes. 
This mean, among other things, that I **DO NOT** define the `publishDir` directive for each process, like what is done in the nextflow documentation, I rather define the `publishDir {}` one and one time only for all processes (in a `process {}` directive) using a `closure` for the `publisDir.path`.
This is very handy and it works very well!

But recently, I wanted to add the possibility to copy the published files instead of doing symlinks (the default). In theory, this is very easy to do, I just have to set `publishDir.mode` with the wanted `closure`.

But I get this error:
`Invalid method invocation 'setMode' with arguments:  (Script989C00012DEDEFE63A87A761299AE910$_run_closure6$_closure56) on PublishDir type`

Indeed, [in the code](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy#L132), `PublishDir.setMode()` can take either a `nextflow.processor.PublishDir.Mode` or a `String`, but not a `closure`.

## New feature

Add a `setMode(def value)` method for the `PublishDir class`.

## Usage scenario 

I do not know if other individuals are using closure to set their `publishDir.path`, but it would be very helpfull for me if this feature is added in Nextflow :)

## Suggest implementation 

I THINK this should work:
```groovy
void setMode( Closure value ) {
  setMode(value.call())
}
``` 

Best,
Audric

dwishsan:
I just realised that `publishDir.enabled` does not work with `closures` either. A workarround, suggested here: #3692, could be to set `publishDir.path` to `null`, but this is not allowed by Nextflow (`Path string cannot be empty`). This is due to the following piece of code:
```groovy
void apply( Set<Path> files, TaskRun task ) {

        if( !files || !enabled )
            return

        if( !path )
            throw new IllegalStateException("Target path for directive publishDir cannot be null")

        if( nullPathWarn )
            log.warn "Process `$task.processor.name` publishDir path contains a variable with a null value"

        this.sourceDir = task.targetDir
        this.sourceFileSystem = sourceDir.fileSystem
        this.stageInMode = task.config.stageInMode
        this.taskName = task.name

        apply0(files)
    }
``` 

[...]

PR #4027, opened 2023-06-15
Allow publish mode to be a closure

dwishsan:
Add `Closure` as possible `PublishDir.setMode()` input
This is related to #4024

This have been tested and work.

Exemple:

```groovy
process {
  ext {
    Output_copy = false
  }
  publishDir {
    mode = { task.ext.Output_copy ? 'copy' : 'symlink' }
  }

  withName: 'FOO' {
    ext.Output_copy = true
  }

  withName: 'BAR' {
    ext.Output_copy = false
  }
}
``` 

Only results of the FOO process will be copied.

**WARNING**:
I expect the closure to return a String, so it won't work if it returns null for exemple. Maybe it should handle it?


pditommaso:
Closing in favour of https://github.com/nextflow-io/nextflow/issues/4186.

Issue #4037, opened 2023-06-20
Wave Containers using Podman: permission denied error


marcodelapierre:
I am testing Conda wave containers with Podman as container runtime, and getting a pemission denied error when the container attempts to write the output on the filesystem:
```
nextflow run marcodelapierre/toy-cowsay-nf -profile condawavepodman

[..]

  .command.sh: line 2: out_hello: Permission denied

Work dir:
  /home/ubuntu/toy-cowsay-nf/work/2c/68ce823fdef2dabe43298b1a2abf4b

Tip: you can try to figure out what's wrong by changing to the process work dir and showing the script file named `.command.sh`
```

I strongly suspect the issue comes from the Mamba base image switching to a non-root user: https://github.com/mamba-org/micromamba-docker/blob/main/Dockerfile#L53

Hence, a simple fix might be (to be tested) to revert the user to `root` in the Wave conda dockerfile.

@pditommaso

marcodelapierre:
An alternative fix might be to launch podman and docker with `docker/podman run -u $USER ..`, to force the host user inside the running container

jordeu:
I've been reviewing this issue, and I think that there are two possible solutions to match the owner of the mounted volume with the user that is running the container.

1. Given that by default rootless podman maps host UID to root inside the container one option is to force the container to always run as root, independently of what `USER` is defined at the image level. This can be done adding `podman.runOptions = '-u root'` to the config.

2. Another solution is to map the host UID to same image UID inside the container, this can be achieve using the option [userns](https://docs.podman.io/en/latest/markdown/options/userns.container.html) on Podman if you know a priori which is the user used inside the image. Example usage `podman.runOptions = '--userns=keep-id:uid=1000,gid=1000'`. The main problem is that this option is only available starting from Podman v4.4 (released on Feb 2023).

marcodelapierre:
Hi @pditommaso  @jordeu, sorry for the wait, I am back from leave.

I think option 1. is preferable, as it does not require knowledge of the userid within the container image, nor of a specific podman version, nor of specific UID remapping features in the OS.
Do you see any disadvantage?

Then I think we should implement `podman.runOptions = '-u root'` in the Nextflow code, to make it transparent to users.

I am happy to update my PR #4038 with this, if you give me green lights.

jordeu:
I agree that solution one is more simple, but not sure if we need to make this the default setup on Nextflow.

[...]

PR #4038, opened 2023-06-20
wave, conda dockerfile templates: always switch to USER root

marcodelapierre:


marcodelapierre:
Fixes https://github.com/nextflow-io/nextflow/issues/4037

pditommaso:
I am wondering if this is the best thing to do. The problem will remain for manually provided docker file using a different user. @jordeu any suggestion?

marcodelapierre:
Thanks Jordi.
Basically, this amounts to having the config option `podman.runOptions = '-u root'` always set in the case of Podman Wave containers. 
Do you folks have a tip on the best place to put it?

PR #4043, opened 2023-06-21
Ignore accelerator type for AWS Batch

bentsherman:
Nextflow is applying the accelerator type to the AWS Batch job definition. However, according to the [AWS documentation](https://docs.aws.amazon.com/batch/latest/APIReference/API_ResourceRequirement.html), the type must be "GPU", and does not support specific GPU types. It looks like the only way to control the GPU type is indirectly through the allowed instance types in a compute environment.

This PR ignores the accelerator type for AWS Batch and adds a warning if the type is specified, so that it's clear that it isn't supported.

### DIFF
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
@@ -59,7 +59,6 @@ import nextflow.container.ContainerNameValidator
 import nextflow.exception.ProcessSubmitException
 import nextflow.exception.ProcessUnrecoverableException
 import nextflow.executor.BashWrapperBuilder
-import nextflow.executor.res.AcceleratorResource
 import nextflow.fusion.FusionAwareTask
 import nextflow.processor.BatchContext
 import nextflow.processor.BatchHandler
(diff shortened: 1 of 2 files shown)


plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
```diff
@@ -715,8 +715,12 @@ class AwsBatchTaskHandler extends TaskHandler implements BatchHandler<String,Job
         if( task.config.getCpus() > 1 )
             resources << new ResourceRequirement().withType(ResourceType.VCPU).withValue(task.config.getCpus().toString())
 
-        if( task.config.getAccelerator() )
-            resources << createGpuResource(task.config.getAccelerator())
+        final accelerator = task.config.getAccelerator()
+        if( accelerator ) {
+            if( accelerator.type )
```
pditommaso, line 720: This could be simplified with 
```suggestion
        if( accelerator?.type ) {
```

pditommaso (commented):
Good catch, it may be useful to mention in the nextflow docs (?)

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
bentsherman: Not quite, if you look closer you will see that both checks are required

pditommaso (approved):

Issue #3711, opened 2023-03-02
Validation of task outputs


robsyme:
## Fail tasks where stageOut fails

This is similar to existing issue https://github.com/nextflow-io/nextflow/issues/3372, but instead of being concerned with the `publishDir` step (run work directory to publish output location), this issue is concerned with the `output` step (task work directory to run work directory)

## Usage scenario 

When task outputs are copied from the task working directory to the run work directory using the aws cli, the `aws s3 cp` process can fail. Nextflow ignores this failure with a `|| true`, which means that in the event of failure, the downstream tasks expecting this input will fail. This makes debugging harder than it should be, as users will be looking for the error in the downstream task rather than in the task that actually failed.

## Suggest implementation 

I would recommend that we remove the `|| true` on L187 here:

https://github.com/nextflow-io/nextflow/blob/2a224258cdd75321f8b3a239c93fdd688c7dc370/modules/nextflow/src/main/groovy/nextflow/executor/SimpleFileCopyStrategy.groovy#L184-L190

In addition, it may be helpful to provide a more descriptive warning in the event of an upload failure, but I'm not 100% sure how best to communicate that to the Nextflow process.

pditommaso:
This is somehow related to [this comment](https://github.com/nextflow-io/nextflow/issues/3639#issuecomment-1427004081). @mribeirodantas tried removing `|| true` but reported no difference. 

Some Bash hacker may want to give it another try

jchorl:
I've hit this as well, when a `cp` fails for any reason and it silently fails to stage out. Even worse if the output is optional, because nextflow can't differentiate between not-produced and failed-to-stage-out. Seeing the linked issues here would indicate it's somewhat important.

> This is somehow related to [this comment](https://github.com/nextflow-io/nextflow/issues/3639#issuecomment-1427004081). @mribeirodantas tried removing `|| true` but reported no difference.
> 
> Some Bash hacker may want to give it another try

Do you recall why? in theory if `set -e` is set and a `cp` operation fails, the exit code should be non-zero and it should fail.

```
$ cp doesntexist foo
cp: cannot stat 'doesntexist': No such file or directory
$ echo $?
1
```

pditommaso:
[Here](https://github.com/nextflow-io/nextflow/pull/3858) there's a tentative solution but it was aborted

jorgee:
[...]

PR #4050, opened 2023-06-24
Wait for _all_ child processes in nxf_parallel

hartzell:
`nxf_parallel` runs a number of processes in parallel, distributing them
across a limited number of workers.

Prior to PR #1884, errors were not being checked.  PR #1884 was
intended to remedy this but effectively only checked the _last_
command (or some subset of the commands) to exit (which was,
unfortunately the case that was tested for).

Processes that exit while there is still work going on were not being
"waited" on, so their exit status was not being checked, and failures
were escaping unnoticed.

This commit:

- updates the loop that's tracking the work so that it waits on
  processes that have exited; and
- updates the test cases so that the failing command is _not_ the
  last in the list and ensures that the templated text is as expected.

Note that an alternative change would be:

    [[ -e /proc/\$x ]] && copy+=(\$x) || wait \$x

While this wins at vim golf, it's not really clear about what's happening.

We also experimented with maintaining a separate array containing the
ids of all the processes that have been started and then waiting on
them before exiting, this made what was going on clearer but didn't
short circuit work if/when there was an error.

I based this change off of the STABLE-23.04.x branch because I was
unable to get master to test cleanly and am not groovy enough to figure
out how to just run _these_ tests.  It might be nice to include some basic
developer documentation (or highlight it, I swear I looked...).

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashFunLib.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashFunLib.groovy
@@ -95,7 +95,11 @@ class BashFunLib<V extends BashFunLib> {
             while ((i<\${#cmd[@]})); do
                 local copy=()
                 for x in "\${pid[@]}"; do
-                  [[ -e /proc/\$x ]] && copy+=(\$x)
+                  if [[ -e /proc/\$x ]]; then
+                    copy+=(\$x)   # process still exists, remember it
+                  else
+                    wait \$x      # process exited, wait on it
+                  fi
                 done
                 pid=("\${copy[@]}")
                 
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashFunLib.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashFunLib.groovy
@@ -95,11 +95,9 @@ class BashFunLib<V extends BashFunLib> {
             while ((i<\${#cmd[@]})); do
                 local copy=()
                 for x in "\${pid[@]}"; do
-                  if [[ -e /proc/\$x ]]; then
-                    copy+=(\$x)   # process still exists, remember it
-                  else
-                    wait \$x      # process exited, wait on it
-                  fi
+                  # if the process exist, keep in the 'copy' array, otherwise wait on it to capture the exit code
+                  # see https://github.com/nextflow-io/nextflow/pull/4050 
+                  [[ -e /proc/\$x ]] && copy+=(\$x) || wait \$x 
                 done
                 pid=("\${copy[@]}")
                 
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashFunLib.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashFunLib.groovy
@@ -96,8 +96,8 @@ class BashFunLib<V extends BashFunLib> {
                 local copy=()
                 for x in "\${pid[@]}"; do
                   # if the process exist, keep in the 'copy' array, otherwise wait on it to capture the exit code
-                  # see https://github.com/nextflow-io/nextflow/pull/4050 
-                  [[ -e /proc/\$x ]] && copy+=(\$x) || wait \$x 
+                  # see https://github.com/nextflow-io/nextflow/pull/4050
+                  [[ -e /proc/\$x ]] && copy+=(\$x) || wait \$x
                 done
                 pid=("\${copy[@]}")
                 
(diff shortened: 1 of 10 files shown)


hartzell:
My first push had a rebase foible in the template test for this change.  I've fixed it, and tests run on my system except for something related to AWS region.
I'm not sure what's causing the CI failures here.  I don't _think_ it's me....

bentsherman:
Hi @hartzell can you redo your PR against the `master` branch instead of `23.04`? The stable branches are only for backports.

hartzell:
@bentsherman -- I can rebase them, but the tests don't run cleanly (mine do, but they're lost in the noise).  Is that ok?

bentsherman:
That's fine, let's put it against master first and then we can deal with any CI failures

hartzell:
Whoops, that didn't do what I wanted.  Hang on.

hartzell:
There we go, needed to change the base branch in the MR.

hartzell:
@bentsherman -- Ok, looks like it's at the point where things are failing that are unrelated to the `nxf_parallel` change.  Can you take a look?

bentsherman:
Actually the current CI failures do appear to be related to your changes. Some tests in the nf-amazon plugin were likely affected by your changes and need to be updated. Please check the tests reported by the CI and update them accordingly.

bentsherman:
Almost there! Now some azure tests are failing, just replace `nf-amazon` with `nf-azure`

hartzell:
And then the google tests.  Easier to just leave the module test off.....

bentsherman:
Unfortunately I think even `make` will skip any downstream modules if an upstream module fails 🙁

hartzell:
Woo hoo!

modules/nextflow/src/main/groovy/nextflow/executor/BashFunLib.groovy
```diff
@@ -95,7 +95,11 @@ class BashFunLib<V extends BashFunLib> {
             while ((i<\${#cmd[@]})); do
                 local copy=()
                 for x in "\${pid[@]}"; do
-                  [[ -e /proc/\$x ]] && copy+=(\$x)
+                  if [[ -e /proc/\$x ]]; then
+                    copy+=(\$x)   # process still exists, remember it
+                  else
+                    wait \$x      # process exited, wait on it
+                  fi
```
pditommaso, line 102:
```suggestion
                  # if the process exist, keep in the 'copy' array, otherwise wait on it to capture the exit code
                  # see https://github.com/nextflow-io/nextflow/pull/4050 
                  [[ -e /proc/\$x ]] && copy+=(\$x) || wait \$x 
```

pditommaso: I'd suggest a more compact notation

modules/nextflow/src/test/groovy/nextflow/executor/BashFunLibTest.groovy
```diff
@@ -88,8 +92,8 @@ class BashFunLibTest extends Specification {
 
         """
         cmds=()
-        cmds+=("true")
         cmds+=("false")
+        cmds+=("true")
```
pditommaso, line 96: Is there a real reason for this change? I'd say no?

hartzell: Yes, there is.  With the previous/original order, the test case does not catch failing tasks unless they're at the end of the list.  This test exercises the more general case.
This change, in a separate test script, is what allowed me to track down the errant behavior.

hartzell:
Rebased to current tip of master.

pditommaso (approved):

bentsherman:
@pditommaso will this PR related at all to fix #3711 where failed stage out commands are ignored? I'm wondering if it would allow us to remove the `|| true` here:

https://github.com/nextflow-io/nextflow/blob/0f4d886776831a187c2a78cafff10a0c1b0c51c8/modules/nextflow/src/main/groovy/nextflow/executor/SimpleFileCopyStrategy.groovy#L186

pditommaso:
This a slightly different problem. I've tried to resolve it here, but I think it's fine in the current form

hartzell:
**_Thank you_** for working with me on this change and getting it merged.

Can you point me at an explanation of the release cycle that would help me figure out when it will be in a release that the team can use?

bentsherman:
We make an edge release every month, so this change should be included in the upcoming `23.07.0-edge` release, sometime later this month.

PR #4057, opened 2023-06-27
Apply job metadata to pod

### LINKED ISSUE #4053: Pod missing metadata e.g. labels when using computeResourceType = Job

(Give an brief description of the expected behavior 
and actual behavior)

Expected pods created by nf jobs via k8s executor still have labels and annotations metadata as given in pod directive

Actual pods missing metadata provided, likely due to PodSpecBuilder.buildAsJob() only setting pod.spec to job.template.spec but not pod.metadata
https://github.com/nextflow-io/nextflow/blob/263ecca8fedf5df6ba7edb4a3dd7b98d5c40d03e/modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy#L575

1. add any workflow a process with `pod label: 'FOO', value: 'bar'`
2. add nextflow.config with `process.executor = 'k8s'` and  `k8s.computeResourceType = 'Job'`
3. run the workflow on k8s
4. `kubectl get job -oyaml -lapp=nextflow`

### Program output 

Job yaml only has the label 'FOO' under its own metadata, but not the pod template level

### Environment 

* Nextflow version: 22.10.08
* Java version:  openjdk 17.0.3
* Operating system: [macOS, Linux, etc] linux
* Bash version: (use the command `$SHELL --version`) 5.1.4

### Additional context

bentsherman:
Close #4053

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/k8s/model/PodSpecBuilderTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/k8s/model/PodSpecBuilderTest.groovy
@@ -1001,23 +1001,24 @@ class PodSpecBuilderTest extends Specification {
                 .buildAsJob()
 
         then:
-        spec ==  [ apiVersion: 'batch/v1',
-                    kind: 'Job',
-                    metadata: [name:'foo', namespace:'default'],
+        spec == [
+            apiVersion: 'batch/v1',
+            kind: 'Job',
+            metadata: [name: 'foo', namespace: 'default'],
+            spec: [
+                backoffLimit: 0,
+                template: [
+                    metadata: [name: 'foo', namespace: 'default'],
                     spec: [
-                            backoffLimit: 0,
-                            template: [
-                                    spec: [
-                                            restartPolicy:'Never',
-                                            containers:[
-                                                    [name:'foo',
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/test/groovy/nextflow/k8s/model/PodSpecBuilderTest.groovy
```diff
@@ -1099,36 +1103,4 @@ class PodSpecBuilderTest extends Specification {
...
-                                                    image:'busybox',
-                                                    command:['echo', 'hello'],
-                                                   ]
-                                           ]
-                                   ]
-                           ]
-                   ]
-        ]
-    }
-
```
pditommaso, line 1133: What was wrong with this test?

bentsherman: it is extraneous, the previous test already tests activeDeadlineSeconds with pod spec, no need to test it with job spec as well

pditommaso: Fair enough

PR #4063, opened 2023-06-29
Add output file patterns to task script for Fusion

### LINKED ISSUE #4031: Tag different process output files when using Fusion
## Description

When using Fusion files stored at S3 are always tagged following this pattern:
https://github.com/nextflow-io/nextflow/blob/cb9592063783576a10ec78800e0ce99ff459bfb8/modules/nextflow/src/main/groovy/nextflow/fusion/FusionConfig.groovy#L31

This allow us to differentiate only two groups of files:
- Nextflow generated files
- Process generated files.

Given that when using Fusion we run with `scratch=false` it will be useful to differentiate which of the process generated files are output of the process and which are just intermediate or temporal files that the process is not cleaning at the end.

It will be interesting to tag they different. 

## Usage scenario 

This will allow to define a S3 cleaning policy to remove all intermediate files that are not needed even to resume the pipeline. 

Another future use case, could be that Fusion directly do not upload them to reduce storage usage.

bentsherman:
Close #4031 

Files that match a process output are tagged by Fusion as `nextflow.io/output`. For other files tagged as `nextflow.io/temporary`, Fusion can try to cache them locally instead of uploading them to object storage.


jordeu (approved):

modules/nextflow/src/main/groovy/nextflow/fusion/FusionEnvProvider.groovy
```diff
@@ -39,9 +39,10 @@ class FusionEnvProvider {
         }
         // tags setting
         if( config.tagsEnabled() ) {
-            result.FUSION_TAGS = config
-                    .tagsPattern()
-                    .replace('[*]', "[${bean.outputFiles.join('|')}](nextflow.io/output=true),[*]")
+            def tagsPattern = config.tagsPattern()
+            if( bean.outputFiles )
+                tagsPattern = tagsPattern.replace('[*]', "[${bean.outputFiles.join('|')}](nextflow.io/output=true),[*]")
```
jordeu, line 44: I think that it may be more safe to use a place holder instead of `[*]` to decide where to add these dynamic tags. We control the default value but is also possible that the user defines his own tag patterns. I will do something like `[.command.*|.exitcode|.fusion.*](nextflow.io/metadata=true),[##PROCESS_OUTPUT##](nextflow.io/output=true),[*](nextflow.io/temporary=true)` and then you replace only this `##PROCESS_OUTPUT##` if exists.

jordeu: Yes, no file is going to match `[](nextflow.io/output=true)` tag.

jordeu (approved):
Do we have somewhere an example or definition of possible output patterns? (I'll add more tests on Fusion to be sure that we are compatible)

bentsherman:
I added only a few examples to the unit test. If you want more, you could grep the process outputs in [nf-core/modules](https://github.com/nf-core/modules) and see what you find...

jordeu:
@bentsherman would it be possible to do the same but with the `publishDir` patterns? 

@pditommaso if we add the publish pattern here we could make convert into a normal file any symbolic link that is a file to be published.

bentsherman:
Yes we could add the publish paths to the Fusion tags, but what exactly are you trying to do with them?

jordeu:
> Yes we could add the publish paths to the Fusion tags, but what exactly are you trying to do with them?

Fusion stores the symbolic link of the staged inputs as a normal object with the target file as content. Nextflow is not aware that this file is a symbolic link and if the process publish the input file then the "symbolic link" is copied, instead of the target file that is what the user would expect.

The idea is that Fusion at shutdown does this symbolic link de-reference if it matches the publish pattern.

pditommaso:
I think we should decouple this from object tagging, and move the output file pattern as special metadata in the .command.run script e.g.:
 
```
## nf-out: "*.txt"
## nf-out: ..
## nf-out:end
```

jordeu:
> I think we should decouple this from object tagging, and move the output file pattern as special metadata in the .command.run script e.g.:

Not really a fan of adding this in the `.command.run` and with a custom format. I'd prefer to add a `.command.yaml` metadata file. But I know that you want to avoid adding more files. 


What about using a standard format, like YAML, but inside the `.command.run`?  

Something like:
```
#!/bin/bash

## container: quay.io/nextflow:bash 
## publishDir:
##    - *.bam
##    - *.pdf
## outputs:
##    - *.bam
...
```

This will be easy to parse at Fusion side, because we use a standard format, and also it will allow to easily add more things if needed.

pditommaso:
Live this, think it should include proper yaml head and termination eg


```
##---
##publishDir
##etc
##...
```

jordeu:
Do you mean to use "## --" line as begin/end YAML delimiters?

pditommaso:
Yes `## ---` beginning and `## ...` end. 

Actually think we should `###`because nextflow strips away `##` sequence

jordeu:
> Yes `## ---` beginning and `## ...` end.
Ok, make sense
 
> Actually think we should `###`because nextflow strips away `##` sequence
When writing the file? or when reading? Is Nextflow going to use this for anything?

pditommaso:
The [template engine](https://github.com/nextflow-io/nextflow/blob/89ede15c71e79858560f93df71064b74d2da2cf4/modules/nextflow/src/main/groovy/nextflow/util/MustacheTemplateEngine.groovy#L31-L31) used by NF interprets `##` asa comment and removes it from the rendered script

bentsherman:
Folded into #3892

PR #4069, opened 2023-06-30
List process container images in preview mode

### LINKED ISSUE #3340: New command/option to download containers for offline usage
We don't currently have the possibility to get the `container` definitions that are used within any given Nextflow pipeline via `nextflow config` or any other existing nextflow command. This is particularly useful especially when you need to pre-download Singularity containers locally so they can be transferred to offline environments before running the pipeline. We have implemented the [`nf-core download`](https://nf-co.re/tools/#downloading-pipelines-for-offline-use) command to do exactly this but it's a hack because the only way we can currently get all of the container definitions is by writing Python code to physically parse the module main scripts which is fragile (see [here](https://github.com/nf-core/tools/blob/dec66abe1c36a8975a952e1f80f045cab65bbf72/nf_core/download.py#L467)).

It would be great to have a dedicated `nextflow download` command that fully resolves all `container` definitions provided in the pipeline. A couple of obvious options:

1. `nextflow download -list nf-core/rnaseq -r 3.9`: List all container definitions in the pipeline
2. `nextflow download -fetch nf-core/rnaseq -r 3.9`: Download Singularity images directly if `container` definitions are prefixed by `https` otherwise convert from Docker to Singularity as done natively by Nextflow

For further inspiration, these are the options we currently use in `nf-core download` and these cover pretty much all bases we have encountered so far:

```
$ nf-core download --help

                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 2.6 - https://nf-co.re



 Usage: nf-core download [OPTIONS] <pipeline name>

 Download a pipeline, nf-core/configs and pipeline singularity images.
 Collects all files in a single archive and configures the downloaded workflow to use relative
 paths to the configs and singularity images.

╭─ Options ────────────────────────────────────────────────────────────────────────────────────────╮
│ --revision                       -r  TEXT                       Pipeline release                 │
│ --outdir                         -o  TEXT                       Output directory                 │
│ --compress                       -x  [tar.gz|tar.bz2|zip|none]  Archive compression type         │
│ --force                          -f                             Overwrite existing files         │
│ --container                      -c  [none|singularity]         Download software container      │
│                                                                 images                           │
│ --singularity-cache-only/--sin…                                 Don't / do copy images to the    │
│                                                                 output directory and set         │
│                                                                 'singularity.cacheDir' in        │
│                                                                 workflow                         │
│ --parallel-downloads             -p  INTEGER                    Number of parallel image         │
│                                                                 downloads                        │
│ --help                           -h                             Show this message and exit.      │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```

bentsherman:
Close #3340 

Currently, this PR just lists the container image for each process during a preview run. It creates a basic task config for each process and tries to resolve the container directive. Even if the directive is dynamic, as long as it's defined in terms of variables that are defined at the pipeline level (e.g. `workflow`, `ext` directive from config), it will work.

If for some reason the container is defined in terms of some task specific property, it resolve will fail and just print null. But in practice I think this is extremely rare.

Successful tests:
- rnaseq
- sarek

Remaining questions:

1. Do we want to create a separate `download` command for this?
2. What should the output format be? CSV? JSON?

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -468,10 +468,24 @@ class Session implements ISession {
         CH.broadcast()
 
         if( preview ) {
+            dumpContainerImages()
             terminated = true
         }
         else {
             callIgniters()
+        }
+    }
+
+    private void dumpContainerImages() {
+        for( def vertex : dag.vertices ) {
+            // skip nodes that are not processes
+            if( !vertex.process )
+                continue
+
+            // print process name and container image
+            def process = vertex.process
+            log.debug "Querying container image for process `${process.name}`..."
+            log.info "${process.name} ${process.getContainer()}"
         }
     }
 
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -367,6 +367,23 @@ class TaskProcessor {
 
     boolean hasErrors() { errorCount>0 }
 
+    /**
+     * Get the container image of the process, or return
+     * null if it can't be resolved.
+     */
+    String getContainer() {
+        try {
+            def config = config.createTaskConfig()
+            config.context = new TaskContext(this)
+
+            return config.getContainer()
+        }
+        catch( Exception e ) {
+            log.warn1 "Unable to resolve container image for process `${name}`", e
+            return null
+        }
+    }
+
     protected void checkWarn(String msg, Map opts=null) {
         if( NF.isStrictMode() )
             throw new ProcessUnrecoverableException(msg)


ewels:
> 2\. What should the output format be? CSV? JSON?

Personally, I quite like the idea of JSON. It's likely that containers could be shared between multiple processes in a lot of pipelines. If we use JSON then we could structure the output so that we have a deduplicated list of container URLs, each with an array of process names that correspond to it.

Having said that, it's not too difficult to deduplicate a CSV for a user. So... maybe both? :trollface:

bentsherman:
Instead of making a `download` command, I think we should make it be just another report for the `run` command, like the workflow diagram. That way, you can generate the container manifest with the exact command line you would use to run the pipeline. Then the downloading could be handled by a Bash script, or a separate Nextflow command that takes the container manifest file as input.

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -468,13 +468,27 @@ class Session implements ISession {
         CH.broadcast()
 
         if( preview ) {
+            dumpContainerImages()
```
pditommaso, line 471: It would be better to enable this using a dedicated option that implicitly enable *preview*

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -468,13 +468,27 @@ class Session implements ISession {
...
 
+    private void dumpContainerImages() {
+        for( def vertex : dag.vertices ) {
+            // skip nodes that are not processes
+            if( !vertex.process )
+                continue
+
+            // print process name and container image
+            def process = vertex.process
+            log.debug "Querying container image for process `${process.name}`..."
```
pditommaso, line 487: This is not needed

pditommaso (commented):
That's nice! I think at this point it could make sense to make it generic and allow the user to provide the list of directives they want to preview

bentsherman:
Okay, I added some config options for preview. Now it will write a JSON file a report, and you can specify which process directives to preview. Here is a sample of the default preview with `nf-core/rnaseq`:

```json
{
    "NFCORE_RNASEQ:RNASEQ:PREPARE_GENOME:GUNZIP_GTF": {
        "container": "nf-core/ubuntu:20.04",
        "cpus": 1,
        "memory": "6 GB",
        "time": "4h"
    },
    "NFCORE_RNASEQ:RNASEQ:PREPARE_GENOME:GUNZIP_ADDITIONAL_FASTA": {
        "container": "nf-core/ubuntu:20.04",
        "cpus": 1,
        "memory": "6 GB",
        "time": "4h"
    },
    "NFCORE_RNASEQ:RNASEQ:PREPARE_GENOME:CAT_ADDITIONAL_FASTA": {
        "container": "biocontainers/python:3.9--1",
        "cpus": 1,
        "memory": "6 GB",
        "time": "4h"
    },
    // ...
}
```

@pditommaso if you want to enable it separately, we could add a `preview.enabled` config option. But that might be confusing with the `-preview` CLI option. A separate option like `-preview-directives` could be fine, but also I could see the preview report expanding to include other information. I think it would be fine to just enable it with `-preview`, as this is the kind of thing users are expecting from the preview mode anyway.

modules/nextflow/src/main/groovy/nextflow/trace/PreviewReportWriter.groovy
```diff
@@ -0,0 +1,119 @@
...
+    private boolean overwrite
+
+    static PreviewReportWriter create(Session session) {
+        final directives = session.config.navigate('preview.directives', DEF_DIRECTIVES) as List
+        def file = session.config.navigate('preview.file', DEF_FILE_NAME)
+        file = (file as Path).complete()
+        final overwrite = session.config.navigate('preview.overwrite', false) as boolean
+
+        return new PreviewReportWriter(session.dag, directives, file, overwrite)
+    }
```
pditommaso, line 59: would not be better to keep the creation login the same as for the other observers? 

https://github.com/nextflow-io/nextflow/blob/main/modules/nextflow/src/main/groovy/nextflow/trace/DefaultObserverFactory.groovy#L12-L12

pditommaso:
> But that might be confusing with the -preview

What if we call this *inspect* and turns on the preview flag implicitly?

bentsherman:
I think that would be more confusing 😆 In that case I would rather have a `preview.enabled` config option that implicitly enables `-preview`. That's the best way I can think of to enable this report separately from the existing preview behavior.

cc @ewels @drpatelh for their thoughts

ewels:
I asked @bentsherman to summarise this for me:

> 1. `-preview` CLI option automatically enables the preview report
> 2. `preview.enabled` config option enables the preview report and implicitly enables `-preview` (but not the other way around)
> 3. same as (2), except we called it `inspect.enabled` (or `-inspect` CLI option)

For the use cases we had been thinking about, I think a CLI flag makes most sense. This is information that the end user wants in a one-off way (usually for deployment). It's not a report that would normally need to be generated for every run. It's also not really specific to pipeline, but rather the user - so wouldn't make sense to put in a pipeline config. Sure, can use a user config, but one-off user-level spells CLI flag to me.

So my +1 is for either `-preview` or `-inspect`, don't really mind which.

bentsherman:
I agree that it should be a CLI option. If it's going to be separate from `-preview`, then I think it should be of the form `-preview-*`, like `-preview-config` or `-preview-report`, so that it's clear that it's an extension of the preview feature rather than a completely separate thing.

Calling it `-inspect` would draw the same confusion that we get over similar operators like `distinct()` vs `unique()`, `combine()` vs `cross()` vs `join()`, `collect()` vs `toList()`, etc.

pditommaso:
To move on this, I think we should focus on the primary use case, that's the previous of pipeline container configuration. In this extend I'd suggest the following 

1. specialise the implementation for the container setting, instead of arbitrary directives 
2. enable it with the CLI option `-preview-containers` 
3. make it wave aware, this may be not so trivial because it will require a `TaskRun` instance to invoke this [code](https://github.com/nextflow-io/nextflow/blob/19a72c402495e72df0140c6af05d299330fa5ba9/modules/nextflow/src/main/groovy/nextflow/processor/TaskRun.groovy#L618-L631)

pditommaso (commented):
it looks nice. A couple of points:

1. it would be nicer to only print the containers preview to stdout by default and hide any other output. it should be possible disabling the ansi output and running the setting the log in quiet mode 
2. what about structuring a bit more the json output 

instead of 

```json
{
    "RNASEQ:INDEX": "wave.seqera.io/wt/d76be4922513/nextflow/rnaseq-nf:v1.1",
    "RNASEQ:QUANT": "wave.seqera.io/wt/d76be4922513/nextflow/rnaseq-nf:v1.1",
    "MULTIQC": "wave.seqera.io/wt/d76be4922513/nextflow/rnaseq-nf:v1.1",
    "RNASEQ:FASTQC": "wave.seqera.io/wt/d76be4922513/nextflow/rnaseq-nf:v1.1"
}
```

maybe something like 

```
{
  processes: [
   { name: "foo", container: "bar" }, 
   .. 
  ]
}
```

pditommaso:
IMO it would be more handy having it to the stdout if not file name is specified. Also, because what's printed instead is completely useless

bentsherman:
@ewels @drpatelh what do you guys think? If no file is given, just `-preview-containers`, would you rather it disable the log and print to stdout, or save to a JSON/config file?

drpatelh:
> @ewels @drpatelh what do you guys think? If no file is given, just -preview-containers, would you rather it disable the log and print to stdout, or save to a JSON/config file?

Not particularly fussed. Personally, I find it a bit confusing when you have optional args like this for parameters.

ewels:
I'm with @drpatelh - I'd prefer either stdout or a file but find both a bit unnecessary. File makes most sense to me here if the filename is being used to choose between JSON and Nextflow config syntax.

Not super fussed about the stdout if writing to a file. Could be nothing.

bentsherman:
I also don't think the stdout is necessary.

Besides, you can always make it work with a little Bash magic 😉 
```bash
nextflow -q run rnaseq-nf -preview-containers preview.json ; cat preview.json
```

pditommaso:
Then I think we should go with stdout version, it's the most intuitive behaviour and it could be easily redirected to a file

ewels:
Fine by me, but then a second CLI flag is needed to define the output format (JSON or NF config).

pditommaso:
It can be the argument

bentsherman:
I still prefer the file output. Also, suppressing the other console output seems not so trivial. Even with the "quiet" flag, the pipeline script can still print warnings to the console.

pditommaso:
`print`s are overridden in the [BaseScript](https://github.com/nextflow-io/nextflow/blob/f369cef37557b7a9dd646cc695971e8f195f16e6/modules/nextflow/src/main/groovy/nextflow/script/BaseScript.groovy#L207-L226).  it should be easy to suppress them

bentsherman:
Not worth the added complexity IMO

bentsherman:
I will try the stdout approach... just for you Paolo 😉

pditommaso:
You will figure out that's simpler than expected

pditommaso:
I've pushed further the implementation of this PR, moving this feature to its own command `inspect`. This allows a simpler implementation and allows the use of specific option to control the ~~preview~~ inspect output. 

Moreover, it now fully support Wave. Some example 

```
» ./launch.sh inspect rnaseq-nf 
[
    {
        "name": "RNASEQ:INDEX",
        "container": "quay.io/nextflow/rnaseq-nf:v1.1"
    },
    {
        "name": "RNASEQ:QUANT",
        "container": "quay.io/nextflow/rnaseq-nf:v1.1"
    },
    {
        "name": "MULTIQC",
        "container": "quay.io/nextflow/rnaseq-nf:v1.1"
    },
    {
        "name": "RNASEQ:FASTQC",
        "container": "quay.io/nextflow/rnaseq-nf:v1.1"
    }
]
```

Enabling Wave 

```
» ./launch.sh inspect rnaseq-nf 
[
    {
        "name": "RNASEQ:INDEX",
        "container": "wave.seqera.io/wt/dda7a502619d/nextflow/rnaseq-nf:v1.1"
    },
    {
        "name": "RNASEQ:QUANT",
        "container": "wave.seqera.io/wt/dda7a502619d/nextflow/rnaseq-nf:v1.1"
    },
    {
        "name": "MULTIQC",
        "container": "wave.seqera.io/wt/dda7a502619d/nextflow/rnaseq-nf:v1.1"
    },
    {
        "name": "RNASEQ:FASTQC",
        "container": "wave.seqera.io/wt/dda7a502619d/nextflow/rnaseq-nf:v1.1"
    }
]
```

Finally enabling the Wave freeze mode with 

```
wave {
  enabled = true
  freeze = true
  build.repository = 'docker.io/pditommaso/wave-tests'
}

```

The output is 

```
» ./launch.sh inspect rnaseq-nf 
[
    {
        "name": "RNASEQ:INDEX",
        "container": "docker.io/pditommaso/wave-tests:b270c282d96c2a2ea41115acaa211e26"
    },
    {
        "name": "RNASEQ:QUANT",
        "container": "docker.io/pditommaso/wave-tests:b270c282d96c2a2ea41115acaa211e26"
    },
    {
        "name": "MULTIQC",
        "container": "docker.io/pditommaso/wave-tests:b270c282d96c2a2ea41115acaa211e26"
    },
    {
        "name": "RNASEQ:FASTQC",
        "container": "docker.io/pditommaso/wave-tests:b270c282d96c2a2ea41115acaa211e26"
    }
]
```

This almost matches the needs of nf-core folks. Thoughts?

bentsherman:
Let's just hope people aren't confused by all the similar vocabulary like stub run, preview, inspect, etc. But at least the separate command will make it easier to add capabilities to pipeline "inspection" in the future

docs/cli.md
```diff
@@ -1070,7 +1070,12 @@ The `run` command is used to execute a local pipeline script or remote pipeline
 `-preview`
 : :::{versionadded} 22.06.0-edge
   :::
-: Run the workflow script skipping the execution of all processes
+: Run the workflow script skipping the execution of all processes.
+
+`-preview-containers` (`json`)
+: :::{versionadded} 23.08.0-edge
+  :::
+: Print a preview of the containers used for each process to stdout (implicitly enables `-preview`). An optional argument specifies the output format, can be `config` or `json`.
```
bentsherman, line 1078: TODO: add docs for `inspect` command

modules/nextflow/src/main/groovy/nextflow/container/inspect/ContainersInspector.groovy
```diff
@@ -0,0 +1,114 @@
...
+    ContainersInspector withFormat(String format) {
+        if( format !in ['config', 'json'] )
+            throw new IllegalArgumentException("Invalid format for container preview: '${format}' -- should be 'config' or 'json'")
+        this.format = format
+        return this
+    }
+
+    ContainersInspector withIgnoreErrors(boolean ignore) {
+        if( format !in ['config', 'json'] )
+            throw new IllegalArgumentException("Invalid format for container preview: '${format}' -- should be 'config' or 'json'")
```
bentsherman, line 54: can be removed

plugins/nf-wave/src/main/io/seqera/wave/plugin/config/WaveConfig.groovy
```diff
@@ -48,11 +48,13 @@ class WaveConfig {
     final private RetryOpts retryOpts
     final private HttpOpts httpClientOpts
     final private Boolean freezeMode
+    final private Boolean awaitMode
 
     WaveConfig(Map opts, Map<String,String> env=System.getenv()) {
         this.enabled = opts.enabled
         this.endpoint = (opts.endpoint?.toString() ?: env.get('WAVE_API_ENDPOINT') ?: DEF_ENDPOINT)?.stripEnd('/')
         this.freezeMode = opts.freeze as Boolean
+        this.awaitMode = opts.navigate('awaitMode', true)
```
bentsherman, line 57: TODO: document `wave.awaitMode`

bentsherman: on second thought, doesn't seem to have any purpose outside of container preview

bentsherman:
> 1. Does it work with singularity containers too?

Yes, it is independent of the container runtime so it works with all of them.

> 2. Can you pass parameters / configs to it? (eg. as we use [here](https://github.com/nf-core/rnaseq/blob/3bec2331cac2b5ff88a1dc71a21fab6529b57a0f/modules/nf-core/fastqc/main.nf#L6-L8))

You can specify config files and profiles, but not CLI params or params file. Currently params can only be included through config files.

> 3. Any intentions to add other things like `cpus`, `memory`, `time` etc as done further up?

I think that's a great idea...

modules/nextflow/src/main/groovy/nextflow/cli/CliOptions.groovy
```diff
@@ -30,6 +31,8 @@ import org.fusesource.jansi.Ansi
 @Slf4j
 class CliOptions {
 
+    static @PackageScope boolean quiteDefault = false
```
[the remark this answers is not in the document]
pditommaso: 😄

pditommaso:
> Does it work with singularity containers too?

Yes. Above all we are going to have Singulatiy native builds via Wave. Local conversion is not going to be needed any more!
 

> Can you pass parameters / configs to it? (eg. as we use [here](https://github.com/nf-core/rnaseq/blob/3bec2331cac2b5ff88a1dc71a21fab6529b57a0f/modules/nf-core/fastqc/main.nf#L6-L8))

Currently, only the profile, but adding the support for params is straightforward. I'm going to add it 

> Any intentions to add other things like cpus, memory, time etc as done further up?

Not in the very short term, but the use of a dedicated command would make it possible to easily extend this functionality.

pditommaso:
Added support for `-revion`, `-params-file` and pipeline custom params. Fixed `quiet` mode. Almost ready. This is the output for rnaseq 👇. 

I'm curios about the feedback from @drpatelh. I remember he was working on a python script doing something similar. Wonder if this address his problem.  

```
[
    {
        "name": "NFCORE_RNASEQ:RNASEQ:FASTQ_FASTQC_UMITOOLS_TRIMGALORE:FASTQC",
        "container": "quay.io/biocontainers/fastqc:0.11.9--0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BEDGRAPH_BEDCLIP_BEDGRAPHTOBIGWIG_FORWARD:UCSC_BEDCLIP",
        "container": "quay.io/biocontainers/ucsc-bedclip:377--h0b8a92a_2"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BAM_MARKDUPLICATES_PICARD:BAM_STATS_SAMTOOLS:SAMTOOLS_IDXSTATS",
        "container": "quay.io/biocontainers/samtools:1.17--h00cdaf9_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:DESEQ2_QC_STAR_SALMON",
        "container": "quay.io/biocontainers/mulled-v2-8849acf39a43cdd6c839a369a74c0adc823e2f91:ab110436faf952a33575c64dd74615a84011450b-0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:ALIGN_STAR:BAM_SORT_STATS_SAMTOOLS:BAM_STATS_SAMTOOLS:SAMTOOLS_STATS",
        "container": "quay.io/biocontainers/samtools:1.17--h00cdaf9_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:MULTIQC_CUSTOM_BIOTYPE",
        "container": "quay.io/biocontainers/python:3.9--1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:PREPARE_GENOME:UNTAR_SALMON_INDEX",
        "container": "quay.io/nf-core/ubuntu:20.04"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:ALIGN_STAR:STAR_ALIGN",
        "container": "quay.io/biocontainers/mulled-v2-1fa26d1ce03c295fe2fdcf85831a92fbcbd7e8c2:1df389393721fc66f3fd8778ad938ac711951107-0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:CUSTOM_DUMPSOFTWAREVERSIONS",
        "container": "quay.io/biocontainers/multiqc:1.14--pyhdfd78af_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BAM_MARKDUPLICATES_PICARD:BAM_STATS_SAMTOOLS:SAMTOOLS_FLAGSTAT",
        "container": "quay.io/biocontainers/samtools:1.17--h00cdaf9_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:FASTQ_FASTQC_UMITOOLS_TRIMGALORE:TRIMGALORE",
        "container": "quay.io/biocontainers/trim-galore:0.6.7--hdfd78af_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BEDGRAPH_BEDCLIP_BEDGRAPHTOBIGWIG_FORWARD:UCSC_BEDGRAPHTOBIGWIG",
        "container": "quay.io/biocontainers/ucsc-bedgraphtobigwig:377--h446ed27_1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:CAT_FASTQ",
        "container": "quay.io/nf-core/ubuntu:20.04"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BAM_MARKDUPLICATES_PICARD:SAMTOOLS_INDEX",
        "container": "quay.io/biocontainers/samtools:1.17--h00cdaf9_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BAM_RSEQC:RSEQC_JUNCTIONANNOTATION",
        "container": "quay.io/biocontainers/rseqc:3.0.1--py37h516909a_1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:PREPARE_GENOME:CUSTOM_GETCHROMSIZES",
        "container": "quay.io/biocontainers/samtools:1.16.1--h6899075_1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUANTIFY_SALMON:SALMON_TX2GENE",
        "container": "quay.io/biocontainers/python:3.9--1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUANTIFY_STAR_SALMON:SALMON_QUANT",
        "container": "quay.io/biocontainers/salmon:1.10.1--h7e5ed60_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUANTIFY_STAR_SALMON:SALMON_SE_GENE_SCALED",
        "container": "quay.io/biocontainers/bioconductor-summarizedexperiment:1.24.0--r41hdfd78af_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BEDGRAPH_BEDCLIP_BEDGRAPHTOBIGWIG_REVERSE:UCSC_BEDCLIP",
        "container": "quay.io/biocontainers/ucsc-bedclip:377--h0b8a92a_2"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BAM_RSEQC:RSEQC_JUNCTIONSATURATION",
        "container": "quay.io/biocontainers/rseqc:3.0.1--py37h516909a_1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:DESEQ2_QC_SALMON",
        "container": "quay.io/biocontainers/mulled-v2-8849acf39a43cdd6c839a369a74c0adc823e2f91:ab110436faf952a33575c64dd74615a84011450b-0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:PREPARE_GENOME:GUNZIP_GTF",
        "container": "quay.io/nf-core/ubuntu:20.04"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:DUPRADAR",
        "container": "quay.io/biocontainers/bioconductor-dupradar:1.28.0--r42hdfd78af_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:FASTQ_SUBSAMPLE_FQ_SALMON:FQ_SUBSAMPLE",
        "container": "quay.io/biocontainers/fq:0.9.1--h9ee0642_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:STRINGTIE_STRINGTIE",
        "container": "quay.io/biocontainers/stringtie:2.2.1--hecb563c_2"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BAM_MARKDUPLICATES_PICARD:PICARD_MARKDUPLICATES",
        "container": "quay.io/biocontainers/picard:3.0.0--hdfd78af_1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:SUBREAD_FEATURECOUNTS",
        "container": "quay.io/biocontainers/subread:2.0.1--hed695b0_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUANTIFY_SALMON:SALMON_TXIMPORT",
        "container": "quay.io/biocontainers/bioconductor-tximeta:1.12.0--r41hdfd78af_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BEDTOOLS_GENOMECOV",
        "container": "quay.io/biocontainers/bedtools:2.30.0--hc088bd4_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BAM_RSEQC:RSEQC_INNERDISTANCE",
        "container": "quay.io/biocontainers/rseqc:3.0.1--py37h516909a_1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:PREPARE_GENOME:GUNZIP_ADDITIONAL_FASTA",
        "container": "quay.io/nf-core/ubuntu:20.04"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUANTIFY_SALMON:SALMON_QUANT",
        "container": "quay.io/biocontainers/salmon:1.10.1--h7e5ed60_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUANTIFY_STAR_SALMON:SALMON_TXIMPORT",
        "container": "quay.io/biocontainers/bioconductor-tximeta:1.12.0--r41hdfd78af_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUANTIFY_STAR_SALMON:SALMON_SE_GENE",
        "container": "quay.io/biocontainers/bioconductor-summarizedexperiment:1.24.0--r41hdfd78af_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:ALIGN_STAR:BAM_SORT_STATS_SAMTOOLS:SAMTOOLS_INDEX",
        "container": "quay.io/biocontainers/samtools:1.17--h00cdaf9_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUANTIFY_STAR_SALMON:SALMON_TX2GENE",
        "container": "quay.io/biocontainers/python:3.9--1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BAM_RSEQC:RSEQC_BAMSTAT",
        "container": "quay.io/biocontainers/rseqc:3.0.1--py37h516909a_1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BAM_RSEQC:RSEQC_READDUPLICATION",
        "container": "quay.io/biocontainers/rseqc:3.0.1--py37h516909a_1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BAM_MARKDUPLICATES_PICARD:BAM_STATS_SAMTOOLS:SAMTOOLS_STATS",
        "container": "quay.io/biocontainers/samtools:1.17--h00cdaf9_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BBMAP_BBSPLIT",
        "container": "quay.io/biocontainers/bbmap:39.01--h5c4e2a8_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:INPUT_CHECK:SAMPLESHEET_CHECK",
        "container": "quay.io/biocontainers/python:3.9--1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUANTIFY_SALMON:SALMON_SE_GENE",
        "container": "quay.io/biocontainers/bioconductor-summarizedexperiment:1.24.0--r41hdfd78af_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:PREPARE_GENOME:BBMAP_BBSPLIT",
        "container": "quay.io/biocontainers/bbmap:39.01--h5c4e2a8_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:ALIGN_STAR:BAM_SORT_STATS_SAMTOOLS:BAM_STATS_SAMTOOLS:SAMTOOLS_IDXSTATS",
        "container": "quay.io/biocontainers/samtools:1.17--h00cdaf9_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUANTIFY_STAR_SALMON:SALMON_SE_GENE_LENGTH_SCALED",
        "container": "quay.io/biocontainers/bioconductor-summarizedexperiment:1.24.0--r41hdfd78af_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUALIMAP_RNASEQ",
        "container": "quay.io/biocontainers/qualimap:2.2.2d--1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BAM_RSEQC:RSEQC_READDISTRIBUTION",
        "container": "quay.io/biocontainers/rseqc:3.0.1--py37h516909a_1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUANTIFY_SALMON:SALMON_SE_GENE_SCALED",
        "container": "quay.io/biocontainers/bioconductor-summarizedexperiment:1.24.0--r41hdfd78af_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:FASTQ_SUBSAMPLE_FQ_SALMON:SALMON_QUANT",
        "container": "quay.io/biocontainers/salmon:1.10.1--h7e5ed60_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:PREPARE_GENOME:STAR_GENOMEGENERATE",
        "container": "quay.io/biocontainers/mulled-v2-1fa26d1ce03c295fe2fdcf85831a92fbcbd7e8c2:1df389393721fc66f3fd8778ad938ac711951107-0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:PREPARE_GENOME:GTF2BED",
        "container": "quay.io/biocontainers/perl:5.26.2"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:ALIGN_STAR:BAM_SORT_STATS_SAMTOOLS:SAMTOOLS_SORT",
        "container": "quay.io/biocontainers/samtools:1.17--h00cdaf9_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BAM_RSEQC:RSEQC_INFEREXPERIMENT",
        "container": "quay.io/biocontainers/rseqc:3.0.1--py37h516909a_1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:MULTIQC",
        "container": "quay.io/biocontainers/multiqc:1.14--pyhdfd78af_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUANTIFY_STAR_SALMON:SALMON_SE_TRANSCRIPT",
        "container": "quay.io/biocontainers/bioconductor-summarizedexperiment:1.24.0--r41hdfd78af_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:PREPARE_GENOME:CAT_ADDITIONAL_FASTA",
        "container": "quay.io/biocontainers/python:3.9--1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:ALIGN_STAR:BAM_SORT_STATS_SAMTOOLS:BAM_STATS_SAMTOOLS:SAMTOOLS_FLAGSTAT",
        "container": "quay.io/biocontainers/samtools:1.17--h00cdaf9_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUANTIFY_SALMON:SALMON_SE_GENE_LENGTH_SCALED",
        "container": "quay.io/biocontainers/bioconductor-summarizedexperiment:1.24.0--r41hdfd78af_0"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:BEDGRAPH_BEDCLIP_BEDGRAPHTOBIGWIG_REVERSE:UCSC_BEDGRAPHTOBIGWIG",
        "container": "quay.io/biocontainers/ucsc-bedgraphtobigwig:377--h446ed27_1"
    },
    {
        "name": "NFCORE_RNASEQ:RNASEQ:QUANTIFY_SALMON:SALMON_SE_TRANSCRIPT",
        "container": "quay.io/biocontainers/bioconductor-summarizedexperiment:1.24.0--r41hdfd78af_0"
    }
]
```

ewels:
>  I remember he was working on a python script doing something similar.

Yup - there are several. nf-core/tools is a python script that does it, and I'm aware of a few others. The problem is that none of them really _understand_ the code and mostly work by just trying to parse the text strings. So whenever we change anything (such as the recent adoption of `docker.registry`, but there are others), everything breaks 💥  It's also very fragile to whitespace / other insignificant syntactic differences (eg. types of quote marks).

> Wonder if this address his problem.

Hopefully! That was the motivation for the issue and this resulting PR. It's certainly looking that way, I'm happy! 😅

pditommaso:
Made a few changes: 

1. Added a prompt confirmation when Wave + freeze mode is enabled, to prevent massive build operation if unwanted 
2. Added a few tests more
3. Restructure the JSON using as below: 

    ```json
            {
                "processes": [
                    {
                        "name": "proc2",
                        "container": "container2"
                    },
                    {
                        "name": "proc1",
                        "container": "container1"
                    }
                ]
            }
     ```

Think it's ready

pditommaso:
Think more on this, i'm not super convinced of adding a prompt confirmation when using Wave for building the container images.

A better solution could be to perform a dry-run request to wave (currently not existing) by default, and submit a real build request when a specific option is provided.

pditommaso:
Thanks Ben for updating the docs. Let's merge this

drpatelh:
Sorry, just catching up with this. Thank you 🙏

It is indeed the perfect replacement for the custom Python scripts we have cobbled together to scrape container information.

pditommaso:
Excellent. I need to show you guys the native build for singularity containers. It should help you a lot.

PR #4077, opened 2023-07-04
Prefer to use CacheFunnel implementation when available for caching non-trivial objects

robsyme:
This change will allow us to define custom caching strategies for objects that are also Maps (custom Meta objects, for example).

Currently, if a Map-like object implements CacheFunnel, it is never used because the CacheHelper's `value instanceof Map` will return true earlier than the test `value instanceof CacheFunnel`.

### DIFF
--- a/modules/nf-commons/src/main/nextflow/util/CacheHelper.java
+++ b/modules/nf-commons/src/main/nextflow/util/CacheHelper.java
@@ -162,6 +162,10 @@ public class CacheHelper {
         if( value instanceof byte[] )
             return hasher.putBytes( (byte[])value );
 
+        if( value instanceof CacheFunnel ) {
+            return ((CacheFunnel) value).funnel(hasher,mode);
+        }
+
         if( value instanceof Object[]) {
             for( Object item: ((Object[])value) )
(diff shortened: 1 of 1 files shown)


bentsherman:
Hi Rob :smile: 

I reviewed the linked issues and PRs. As I understand it, the original motivation was to create an immutable map that only supports the `+` operator instead of mutation

I think we will be able to achieve this natively with record types. The `record()` function can be used in lieu of a map literal to create a record, which is just an immutable map with the interface you describe [here](https://github.com/nextflow-io/nf-schema/issues/23#issuecomment-2071598995)

I suspect the custom hashing strategy won't be needed once we implement some related hash improvements: #5351

I will probably just fold both of these PRs into the record types PR once it is ready

robsyme:
The original intent was not just immutable Maps, but any custom class that inherits from a Map. For example - you may want a custom Map class that is able to compute new fields from existing fields, or to pull properties/values from a LIMS system, perhaps.

That said - this all predates our discussions and your thoughts on record types, so I'm happy to see how many of the fundamentals problems dissolve into simpler solutions when we have proper record types.

bentsherman:
Closing in favor of #6679

PR #3664, opened 2023-02-18
Remove redundant DSL version checks

bentsherman:
First batch of changes from #3344 . Includes the most straightforward changes that don't break a bunch of tests or require much refactoring.

* DSL1 deprecation errors that are now unreachable since DSL1 will fail at the beginning anyway
* ~~Classes and methods marked as `@Deprecated` that aren't used anywhere else in the code~~

PR #4085, opened 2023-07-10
Fix incorrect error message on missing comma

### LINKED ISSUE #3185: wrong error when forgetting comma

When forgetting a comma in the input section:

```nextflow
process PLINK_ASSOC_GLM {

    input: 
    tuple path(bedfile), path(bimfile), path(famfile), path(covar) path(EP_list_hg19)
                                                                  ^
}
```

The script throws the wrong error:
```
Process 'PLINK_ASSOC_GLM' has been already used -- 
If you need to reuse the same component include it with a different name
or include in a different workflow context
```

Instead of the error: `you forgot a comma`

### Program output 
Log attached
[log.txt](https://github.com/nextflow-io/nextflow/files/9497043/log.txt)

### Environment 
in log file
```
  Version: 22.04.5 build 5708
  Created: 15-07-2022 16:09 UTC (18:09 CEST)
  System: Linux 4.19.0-21-amd64
  Runtime: Groovy 3.0.10 on OpenJDK 64-Bit Server VM 11.0.16+8-post-Debian-1deb10u1
  Encoding: UTF-8 (ISO-8859-1)
  Process: 1926@login3.lisa.surfsara.nl [145.101.32.96]
  CPUs: 16 - Mem: 16 GB (15.2 GB) - Swap: 16 GB (14.7 GB)
```

bentsherman:
Close #3185 

This PR fixes a particularly nasty error message that occurs with missing commas in process inputs/outputs.

Consider the following example:
```groovy
process foo {
  debug true

  input:
  tuple val(x) val(y) // missing comma

  script:
  """
  echo "foo ${task.index} ${task.workDir}"
  """
}

workflow {
  foo(['x', 'y'])
}
```

Currently, the script compiles despite the missing comma but then Nextflow reports that the process has been used twice:
```console
 nextflow run gh-3185/bad-error-message-4.nf
N E X T F L O W  ~  version 23.06.0-edge
Launching `gh-3185/bad-error-message-4.nf` [loving_keller] DSL2 - revision: 63eb6918e1
Process 'foo' has been already used -- If you need to reuse the same component, include it with a different name or include it in a different workflow context

 -- Check script 'gh-3185/bad-error-message-4.nf' at line: 14 or see '.nextflow.log' file for more details
```

The transpiled Groovy code looks like this:
```groovy
this._in_tuple(
    new nextflow.script.TokenValCall(new nextflow.script.TokenVar('x'))
).val(new nextflow.script.TokenVar('y'))
```

So I just add a `methodMissing()` implementation to `BaseParam` to catch the incorrect `.val()` call. With that change, the error now looks like this:
```console
$ ../launch.sh run gh-3185/bad-error-message-4.nf
N E X T F L O W  ~  version 23.06.0-edge
Launching `gh-3185/bad-error-message-4.nf` [nasty_bohr] DSL2 - revision: 63eb6918e1
Invalid method call `val([nextflow.script.TokenVar(y)])` -- possible syntax error

 -- Check script 'gh-3185/bad-error-message-4.nf' at line: 5 or see '.nextflow.log' file for more details
```

The error message is correctly called a syntax error and it gives the exact line where it happened.

Note however that it doesn't catch all missing comma errors. For example if you forget a comma before `emit: ...`, the script fails to compile so we can't catch the error.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/params/BaseParam.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/params/BaseParam.groovy
@@ -18,7 +18,7 @@ package nextflow.script.params
 
 import groovy.util.logging.Slf4j
 import nextflow.exception.ScriptRuntimeException
-
+import nextflow.script.TokenVar
 /**
  * Base class for input/output parameters
  *
@@ -146,7 +146,25 @@ abstract class BaseParam implements Cloneable {
      * Report missing method calls as possible syntax errors.
      */
     def methodMissing( String name, def args ) {
-        throw new ScriptRuntimeException("Invalid method call `${name}(${args})` -- possible syntax error")
+        throw new ScriptRuntimeException("Invalid function call `${name}(${argsToString0(args)})` -- possible syntax error")
     }
 
+    private String argsToString0(args) {
+        if( args instanceof Object[] )
+            args = Arrays.asList(args)
+        if( args instanceof List ) {
+            final result = new ArrayList()
+            for( def it : args )
+                result.add(argsToString1(it))
+            return result.join(',')
+        }
+        return argsToString1(args)
+    }
+
+    private String argsToString1(arg) {
+        if( arg instanceof TokenVar )
+            return arg.name
+        else
+            return String.valueOf((Object)arg)
+    }
 }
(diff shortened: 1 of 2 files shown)


bentsherman:
If we remove old DSL1 syntax (#3664 and #3671), this change should also catch things like `from` and `into`.

pditommaso:
What about adding a unit test for capture this behavior

pditommaso (approved):

PR #4097, opened 2023-07-14
Add S3 cache store

bentsherman:
Close https://github.com/seqeralabs/nf-tower-cloud/issues/4822

Adds a new cache store backed by S3. The cache entry for each task is saved to the following S3 path:
```
s3://<bucket>/<path>/<session-id>/<task-hash>
```

It can be enabled by setting `NXF_CACHE_STORE` to `s3` and `NXF_CACHE_PATH` to the desired S3 path.

Notes:
- ~~Currently included in the nf-amazon plugin, but could be generalized to any object storage since it uses the Path API. Maybe better to make it part of the nextflow module and call it something like `ObjectCacheStore`.~~
- Consider exposing config options in addition to environment variables.
- Consider using buckets for the task hash (like the work directory) instead of storing all task entries under the same S3 prefix. I don't know if this will actually help though. Usually Nextflow just needs to try a particular key, not iterate through all keys.
- Consider setting the default cache path based on the work directory (e.g. `${workDir}/../cache`


modules/nextflow/src/main/groovy/nextflow/cache/PathCacheStore.groovy
```diff
@@ -0,0 +1,170 @@
...
+
+    @Override
+    void close() {
+        lock.delete()
+    }
+
+    @Override
+    void writeIndex(HashCode key, boolean cached) {
+        getCachePath(key).bytes = [ cached as byte ] as byte[]
+    }
```
pditommaso, line 116: Not understanding this logic, the index file is essentially an append log, that keep the order of the execution of the tasks 

https://github.com/nextflow-io/nextflow/blob/s3-cache-store/modules/nextflow/src/main/groovy/nextflow/cache/DefaultCacheStore.groovy#L134-L137

pditommaso (changes requested):
Using the underlying virtual file system is a good idea. I'm concerned to on  503 slow down error that can be reported when resulting large pipeline because the object storage API will be under pressure. 

Also, it would make sense to move this into its own plugin to be activated to enable this feature

bentsherman:
> I'm concerned to on 503 slow down error that can be reported when resulting large pipeline because the object storage API will be under pressure.

Me too. The AWS SDK should be able to handle it with it's built-in adaptive retry, but I don't remember if it's enabled by default.

> Also, it would make sense to move this into its own plugin to be activated to enable this feature.

I made it enabled with an environment variable `NXF_CACHE_STORE='path'` so that it doesn't have to reside in a separate plugin. It still uses the default cache store by default.

modules/nextflow/src/main/groovy/nextflow/cache/PathCacheStore.groovy
bentsherman: I think I did this because I wasn't sure if the index file and leveldb were independent, i.e. writing a cache entry to leveldb doesn't affect the index file. If they are independent, I will just save the index file as an S3 object and use it exactly as the default store.

bentsherman: However, since I can't append to an S3 object, I will have to re-upload the entire index file every time a task is written. That should be okay for now, but ideally we would want to avoid that.

The index file iteration is used by the `clean` and `log` commands, but they don't need the order of execution. This is why I think we could discard the index file and instead simply list the directory of cache entries when we need to iterate.

pditommaso: Yes, you can create an output stream with `Files.newOutputStream(Path)`

bentsherman: So that does some kind of streaming upload that continues for the entire pipeline run? Seems like magic but should make the index file easy to manage. I will do that.

bentsherman: I know... I'm just curious how the S3 output stream manages the upload under the hood.

bentsherman:
I moved the patch cache to a plugin and added the index file to behave like the default.

One problem with the plugin is that there is no way to enable it with the `clean` and `log` commands that use the cache. How do you recommend we instrument those commands?

pditommaso:
Likely the only way to enable it in those commands is via the NXF_PLUGINS_DEFAULT, which is not very handy. However, the primary need for this plugin is to use it with Tower, therefore I would not care too much

bentsherman:
Okay, then this PR is ready for review. I tested with rnaseq-nf by running and then resuming the pipeline. Looks like there is still a problem with `packing.gradle`, but I'm not sure how to fix it.

packing.gradle
```diff
@@ -21,13 +21,14 @@ dependencies {
...
-    amazon  project(':plugins:nf-amazon')
-    azure   project(':plugins:nf-azure')
-    tower   project(':plugins:nf-tower')
-    wave    project(':plugins:nf-wave')
+    amazon    project(':plugins:nf-amazon')
+    azure     project(':plugins:nf-azure')
+    console   project(':plugins:nf-console')
+    ga4gh     project(':plugins:nf-ga4gh')
+    google    project(':plugins:nf-google')
+    pathcache project(':plugins:nf-path-cache')
```
pditommaso, line 29: No really needed adding it here, for example nf-codecommit plugin is not include in this list

plugins/nf-path-cache/src/main/nextflow/PathCachePlugin.groovy
```diff
@@ -0,0 +1,31 @@
...
+import nextflow.plugin.BasePlugin
+import org.pf4j.PluginWrapper
+/**
+ * Nextflow path cache plugin
+ *
+ * @author Ben Sherman <bentshermann@gmail.com>
+ */
+@Slf4j
+@CompileStatic
+class PathCachePlugin extends BasePlugin {
```
pditommaso, line 25: What about calling it CloudCacheXxx

docs/plugins.md
```diff
@@ -13,6 +13,7 @@ Currently the following functionalities are implemented as plugin components and
 - `nf-console`: Implement Nextflow [REPL console](https://www.nextflow.io/blog/2015/introducing-nextflow-console.html).
 - `nf-ga4gh`: Support [GA4GH APIs](https://www.ga4gh.org/).
 - `nf-google`: Support for Google cloud.
+- `nf-path-cache`: Support for path-based Nextflow cache.
```
pditommaso, line 16: I'd proposed to call this `nf-cloudcache` or `nf-cloudcachedb`, at the end the primary use case is to implement a cache DB for cloud-based runs

pditommaso (changes requested):

validation/google.sh
```diff
@@ -68,4 +64,18 @@ $NXF_CMD -C ./gls.config run ./test-overwrite.nf
 ## re-executing should overwrite the published file
 [ `$NXF_CMD -C ./gls.config run ./test-overwrite.nf -resume | { grep 'Failed to publish file' -c || true; }` == 0 ] && echo OK || { echo 'Failed to publish file' && false; }
 
+NXF_CLOUDCACHE_PATH=gs://rnaseq-nf/cache \
+$NXF_CMD -C ./google.config \
+    run nextflow-io/rnaseq-nf \
+    -with-report \
+    -with-trace $OPTS \
```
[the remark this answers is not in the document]
pditommaso: @sonatype-lift ignore

pditommaso:
Ok, [those](https://github.com/nextflow-io/nextflow/pull/4097/commits/24f52ac141cc95b88d0a52e3034a545abb394179) tests works, but it should be added some checks that's effectively resuming the tasks and using the cloud cache. Logging adding some grep in the log file. @bentsherman can you please give it a try? then it can be merged

validation/awsbatch.sh
```diff
@@ -36,7 +36,18 @@ $NXF_CMD run test-complexpaths.nf -resume -c awsbatch.config
...
+    -plugins nf-cloudcache
+
+NXF_CLOUDCACHE_PATH=s3://nextflow-ci/cache \
+$NXF_CMD run nextflow-io/rnaseq-nf \
+    -profile batch \
+    -with-report \
+    -with-trace \
+    -plugins nf-cloudcache \
+    -resume
+[[ `grep -c 'Cached process > ' .nextflow.log` == 4 ]] || false
```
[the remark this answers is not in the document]
bentsherman: @sonatype-lift ignore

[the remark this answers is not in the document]
pditommaso: this is ok, tho it does not guarantee that's using the S3 based cache, no?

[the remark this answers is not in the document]
pditommaso: Cool, let's add the same also to Google batch and Azure

validation/awsbatch.sh
```diff
@@ -36,7 +36,20 @@ $NXF_CMD run test-complexpaths.nf -resume -c awsbatch.config
...
 $NXF_CMD run test-subdirs.nf -c awsbatch.config
 
+NXF_CLOUDCACHE_PATH=s3://nextflow-ci/cache \
 $NXF_CMD run nextflow-io/rnaseq-nf \
     -profile batch \
     -with-report \
-    -with-trace
+    -with-trace \
+    -plugins nf-cloudcache
+[[ `grep -c 'Using Nextflow cache factory: nextflow.cache.CloudCacheFactory' .nextflow.log` == 1 ]] || false
```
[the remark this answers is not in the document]
bentsherman: @sonatype-lift ignoreall

pditommaso:
I've removed sonatype-lif, just annoying

pditommaso (approved):

abhi18av:
This is b-e-a-utiful, thanks guys! 🎉

bentsherman:
@pditommaso is there a way to trigger the cloudcache plugin when the `NXF_CLOUDCACHE_PATH` variable is set? Currently you have to enable the plugin explicitly, seems like you shouldn't have to do that since it's a core plugin. I'm not sure how to do this through the priority extensions loader.

PR #4098, opened 2023-07-14
Remove logging of report JSON data

### LINKED ISSUE #4093: Add option to turn off rendering of single-line JSON of trace report in log files
## New feature

It would be to have the option to turn off rendering of the entire trace/execution report as a single-line JSON at the end of `.nextflow.log` runs.

## Usage scenario 

In the cases of particularly large or complex pipelines, this results in a very large  log files and makes them difficult to parse (e.g. grepping for single process name when trying to debug.).

Having the ability to turn off the printing of the JSON to the `.nextflow.log` file will be beneficial for reading logs, also particularly given the information can be saved as an explicit file (e.g. HTML).

For example, trying to grep for the process 'BOWTIE2_PHIX_REMOVAL_BUILD' in the attached log file, still requireslots of scrolling (or grep expression specifications) to get to actually useful information about whether the process was executed correctly. etc.

[nextflow.txt](https://github.com/nextflow-io/nextflow/files/12037009/nextflow.txt)

## Suggest implementation 

I believe the 'offending' section is here:

https://github.com/nextflow-io/nextflow/blob/84f7a61a1f5c763c86cd09fa8f9c4d1a3eb0db99/modules/nextflow/src/main/groovy/nextflow/trace/ReportObserver.groovy#L239-L243

It would be nice to have an option (e.g. in the corresponding configuration scope block) so that the `log.debug` only is used (or just the ${result} included) when requested.

For example 

```groovy
    protected String renderSummaryJson(printToLog) {
        final result = aggregator.renderSummaryJson()

       if ( printToLog ) {
          log.debug "Execution report summary data:\n  ${result}"
       }
       
        return result
    }
```

I would try this myself as it seems like a very small change, but when I looked into it I found I didn't have enough experience to parse through all the layers of functions that end up calling that function, nor how to specify a configuration variable to control for this.

bentsherman:
Close #4093

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/trace/ReportObserver.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/trace/ReportObserver.groovy
@@ -237,9 +237,7 @@ class ReportObserver implements TraceObserver {
     }
 
     protected String renderSummaryJson() {
-        final result = aggregator.renderSummaryJson()
-        log.debug "Execution report summary data:\n  ${result}"
-        return result
+        aggregator.renderSummaryJson()
     }
 
     protected String renderPayloadJson() {


modules/nextflow/src/main/groovy/nextflow/trace/ReportObserver.groovy
```diff
@@ -237,9 +237,7 @@ class ReportObserver implements TraceObserver {
     }
 
     protected String renderSummaryJson() {
-        final result = aggregator.renderSummaryJson()
-        log.debug "Execution report summary data:\n  ${result}"
-        return result
+        aggregator.renderSummaryJson()
```
pditommaso, line 240:
```suggestion
        return aggregator.renderSummaryJson()
```

pditommaso (approved):

PR #4109, opened 2023-07-18
Fix stage script in Fusion script launcher

bentsherman:
The stage script, which is written to a separate file if it is too large, is not written correctly with Fusion because Nextflow was trying to write to the Fusion path instead of the S3 path.


pditommaso (approved):
Thanks!

PR #4111, opened 2023-07-19
Fix container status check on aws batch

jchorl:
From the comment, it seems the intention is to use the container status if it exists. However, because 0 is falsy, it was still falling back to reading `.exitcode` when the batch job finished successfully.

Tbh, the old behaviour was a good sanity check to make sure the process really finished and all files got written. I'm not sure if other such checks exist.

But anyway, I believe the behaviour now matches the commented behaviour.


pditommaso:
Not sure, this should be merged. I'd prefer to amend the comment, unless there's a clear reason  to change the code

pditommaso:
Closing untile further evidence is found

PR #4114, opened 2023-07-19
Add deleteTasksOnCompletion to Azure Batch configuration

adamrtalbot:
Deleting Azure Tasks was checking the configuration object `deleteJobsOnCompletion` which was incorrect since a task belongs to a job. This adds the equivalent configuration for tasks which is checked before deleting the tasks.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -340,7 +340,10 @@ The following settings are available:
 : Delete all compute node pools when the workflow completes (default: `false`).
 
 `azure.batch.deleteTasksOnCompletion`
-: Delete each task when it completes (default: `false`).
+: :::{versionadded} 23.08.0-edge
+  :::
+: Delete each task when it completes (default: `true`).
+: Although this setting is enabled by default, failed tasks will not be deleted unless it is explicitly enabled. This way, the default behavior is that successful tasks are deleted while failed tasks are preserved for debugging purposes.
 
 `azure.batch.endpoint`
 : The batch service endpoint e.g. `https://nfbatch1.westeurope.batch.azure.com`.
--- a/docs/config.md
+++ b/docs/config.md
@@ -346,7 +346,7 @@ The following settings are available:
 : The name of the batch service region, e.g. `westeurope` or `eastus2`. This is not needed when the endpoint is specified.
 
 `azure.batch.terminateJobsOnCompletion`
-: When the workflow completes, set all jobs to complete on task completion. (default: `true`).
+: When the workflow completes, set all jobs to terminate on task completion. (default: `true`).
 
 `azure.batch.pools.<name>.autoScale`
 : Enable autoscaling feature for the pool identified with `<name>`.
--- a/docs/config.md
+++ b/docs/config.md
@@ -332,12 +332,15 @@ The following settings are available:
 
 `azure.batch.deleteJobsOnCompletion`
 : Delete all jobs when the workflow completes (default: `false`).
+: :::{versionchanged} 23.08.0-edge
+  Default value was changed from `true` to `false`.
+  :::
 
 `azure.batch.deletePoolsOnCompletion`
 : Delete all compute node pools when the workflow completes (default: `false`).
 
 `azure.batch.deleteTasksOnCompletion`
-: Delete each task when it completes (default: `true`).
+: Delete each task when it completes (default: `false`).
 
 `azure.batch.endpoint`
 : The batch service endpoint e.g. `https://nfbatch1.westeurope.batch.azure.com`.
@@ -346,6 +349,8 @@ The following settings are available:
 : The name of the batch service region, e.g. `westeurope` or `eastus2`. This is not needed when the endpoint is specified.
 
 `azure.batch.terminateJobsOnCompletion`
+: :::{versionadded} 23.05.0-edge
+  :::
 : When the workflow completes, set all jobs to terminate on task completion. (default: `true`).
 
 `azure.batch.pools.<name>.autoScale`
--- a/docs/config.md
+++ b/docs/config.md
@@ -331,7 +331,7 @@ The following settings are available:
 : Specify where the `azcopy` tool used by Nextflow. When `node` is specified it's copied once during the pool creation. When `task` is provider, it's installed for each task execution (default: `node`).
 
 `azure.batch.deleteJobsOnCompletion`
-: Delete all jobs when the workflow completes (default: `true`).
+: Delete all jobs when the workflow completes (default: `false`).
 
 `azure.batch.deletePoolsOnCompletion`
 : Delete all compute node pools when the workflow completes (default: `false`).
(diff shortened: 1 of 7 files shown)


modules/nf-commons/src/main/nextflow/Const.groovy
pditommaso, line 1: Do not include this file in the PR, please

pditommaso (changes requested):
Ok, tho cleanup logic is becoming a bit convoluted.

modules/nf-commons/src/main/nextflow/Const.groovy
adamrtalbot: woops, sorry about that.

adamrtalbot: Done now.

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchTaskHandler.groovy
```diff
@@ -130,15 +130,15 @@ class AzBatchTaskHandler extends TaskHandler implements FusionAwareTask {
     }
 
     private Boolean shouldDelete() {
-        executor.config.batch().deleteJobsOnCompletion
```
pditommaso, line 133: Why `deleteJobsOnCompletion` is gone?

adamrtalbot: > Why `deleteJobsOnCompletion` is gone?

It was checking `deleteJobsOnCompletion`, then deleting the _tasks_. I thought that doesn't really make sense, so a better option would be to add an equivalent `deleteTasksOnCompletion` configuration item.

pditommaso:
Tagging @bentsherman for reviewing

docs/config.md
```diff
@@ -330,24 +330,24 @@ The following settings are available:
...
-: Delete tasks after successful completion. This deletes them on the Azure Batch service but files and resources may persist on the compute nodes (default: `true`).
-
 `azure.batch.deleteJobsOnCompletion`
 : Enable the automatic deletion of jobs created by the pipeline execution (default: `true`).
 
 `azure.batch.deletePoolsOnCompletion`
 : Enable the automatic deletion of compute node pools upon pipeline completion (default: `false`).
 
+`azure.batch.deleteTasksOnCompletion`
+: Delete tasks after successful completion. This deletes them on the Azure Batch service but files and resources may persist on the compute nodes (default: `true`).
```
adamrtalbot, line 340: Had a quick thought - does this default to true? I think it actually defaults to false.

bentsherman (commented):
Adam, do you think it would be better to remove the task deletion logic? If they are already deleted when the job is deleted (for which there is already a config option), then it seems like that would be better by reducing the number of API calls to Azure Batch. On the other hand, I suppose the tasks would be deleted only when the entire job queue is done, rather than immediately. I'm not sure whether that matters.

docs/config.md
bentsherman: I think you are right, all of these delete flags are false by default.

bentsherman:
From the [documentation](https://learn.microsoft.com/en-us/rest/api/batchservice/task/delete?tabs=HTTP):

> When a Task is deleted, all of the files in its directory on the Compute Node where it ran are also deleted (regardless of the retention time).

That makes me think there is an advantage to deleting tasks as soon as possible. Unless someone wants to be able to control task deletion and job deletion, I am fine with the current behavior, which controls both via `deleteJobsOnCompletion`.

bentsherman:
On the other hand, [Dave on StackOverflow](https://stackoverflow.com/questions/34618115/is-deleting-tasks-and-pools-necessary-in-azure-batch-jobs) says:

> You don't need to delete tasks or their jobs at all and there's no real downside of leaving them in the system. You do however need to terminate jobs on completion as there is a quota on the number of active jobs within a Batch account.

I think I will go with the Microsoft docs 😆

docs/config.md
bentsherman: Actually it defaults to false there too:

https://github.com/nextflow-io/nextflow/blob/9fc1d3bd951b1d9d75ba8e64a5aa497cc087f62c/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchTaskHandler.groovy#L137-L138

bentsherman: Probably this logic should happen in `AzBatchOpts`

bentsherman:
But my point is, if the tasks aren't deleted until the entire job is done, will that cause the task files on the VM to accumulate and run out of storage? Or are the task files already deleted some other way?

bentsherman:
Okay, then let's add your config option making it true by default, and if a user has rate limit issues then they have that option as a lever. Let me push a few minor changes first.

bentsherman:
The way it works currently is that `deleteTasksOnCompletion` can be true, false, or null. It is true by default, but failed tasks are deleted only if it is *explicitly* true. It's like a hidden third state. So by default, successful tasks will be deleted and failed tasks will be preserved in case you want to inspect it. I think that comment is backwards, so I changed it to "preserve failed tasks for debugging purposes".

If you think that isn't useful, I would be fine with removing it, thereby deleting all tasks by default.

docs/config.md
```diff
@@ -330,21 +330,24 @@ The following settings are available:
...
+: Delete each task when it completes (default: `true`).
 
 `azure.batch.endpoint`
 : The batch service endpoint e.g. `https://nfbatch1.westeurope.batch.azure.com`.
 
 `azure.batch.location`
 : The name of the batch service region, e.g. `westeurope` or `eastus2`. This is not needed when the endpoint is specified.
 
+`azure.batch.terminateJobsOnCompletion`
+: When the workflow completes, set all jobs to complete on task completion. (default: `true`).
```
bentsherman, line 349: Did you mean "terminate" or are they interchangeable?

```suggestion
: When the workflow completes, set all jobs to terminate on task completion. (default: `true`).
```

docs/config.md
```diff
@@ -330,21 +330,24 @@ The following settings are available:
 `azure.batch.copyToolInstallMode`
 : Specify where the `azcopy` tool used by Nextflow. When `node` is specified it's copied once during the pool creation. When `task` is provider, it's installed for each task execution (default: `node`).
 
-`azure.batch.terminateJobsOnCompletion`
-: Enables the Batch Job to automatically terminate a job once all tasks have completed (default: `true`).
-
 `azure.batch.deleteJobsOnCompletion`
-: Enable the automatic deletion of jobs created by the pipeline execution (default: `true`).
+: Delete all jobs when the workflow completes (default: `true`).
```
bentsherman, line 334:
```suggestion
: Delete all jobs when the workflow completes (default: `false`).
```

bentsherman:
Okay I updated the tests and docs to reflect that `deleteJobsOnCompletion` is now false by default. But we should still make `deleteTasksOnCompletion` true by default right? So that we are still deleting the task files from the VMs as soon as possible?

pditommaso (approved):

PR #4129, opened 2023-07-25
Document behavior of withName selector with included aliases

### LINKED ISSUE #2490: Process selector of original process name applies to modules included as original name and as aliased name.
## Bug report 

When including a process via `include`, and then configuring it using `withName`, the original process name configuration applies to both the originally named process and the aliased process.

The expected behaviour is that the `withName: FOO` configuration, should only apply to `FOO` and not `BAR`.
The actual behaviour is that `withName: FOO` applies to both `FOO` and `BAR`.

`main.nf`:
```groovy
nextflow.enable.dsl = 2

include { FOO; FOO as BAR } from './foo'

workflow {
    FOO ( 'This is foo' )
    BAR ( 'This is bar' )
}
```
`foo.nf`:
```groovy
process FOO {

    input:
    val message

    script:
    """
    echo $message
    """
}
```
`nextflow.config`:
```groovy
process {
    withName: 'FOO' {
        echo = true
    }
}
```

### Program output 

```console
$ NXF_VER=21.10.5 nextflow run main.nf 
CAPSULE: Downloading dependency io.nextflow:nextflow:jar:21.10.5
CAPSULE: Downloading dependency io.nextflow:nf-commons:jar:21.10.5
CAPSULE: Downloading dependency io.nextflow:nf-httpfs:jar:21.10.5
N E X T F L O W  ~  version 21.10.5 
Launching `main.nf` [confident_babbage] - revision: 06a11bc636
executor >  local (2)
[da/3129d7] process > FOO [100%] 1 of 1 ✔
[05/2f2c5b] process > BAR [100%] 1 of 1 ✔
This is foo

This is bar

```

The issue also occurs if `FOO` is omitted from the workflow ( and only `BAR` is used ).

### Environment 

* Nextflow version: 21.10.0, 21.10.5
* Java version: openjdk 11.0.9.1
* Operating system: macOS
* Bash version: GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin20)

bentsherman:
Close #2490

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1162,6 +1162,10 @@ process {
 }
 ```
 
+:::{note}
+The `withName` selector applies to a process even when it is included from a module under an alias. For example, `withName: hello` will apply to any process originally defined as `hello`, regardless of whether it is included under an alias. Similarly, it will not apply to any process not originally defined as `hello`, even if it is included under the alias `hello`.
+:::
+
 :::{tip}
 Label and process names do not need to be enclosed with quotes, provided the name does not include special characters (`-`, `!`, etc) and is not a keyword or a built-in type identifier. When in doubt, you can enclose the label name or process name with single or double quotes.
 :::


pditommaso (approved):

saulpierotti:
Hi, the documentation for this seems inaccurate to me. While the first part is true that the withName selector applies also to the original name of a module, not only the import name, in my usage experience it does also apply to the import name. For example if I have a process named `FOO` and I import it with `import FOO as BAR`, a `withName:BAR` selector will work, and also a `withName:FOO` selector will. The documentation suggests that only `withName:FOO` works, which I believe to be incorrect.

PR #4130, opened 2023-07-25
Hide processes that haven't been executed yet

bentsherman:
Original discussion: https://github.com/nextflow-io/nextflow/discussions/3107?sort=new#discussioncomment-3388088

Hides the process status from the log until it has created some tasks.

I suggest we add an environment variable to toggle it.


pditommaso:
Not sure this should be done. it's useful to see the list of pending processes even if no tasks have been running

bentsherman:
Some people have requested it, that's why I think it should be opt-in with an environment var.

bentsherman:
Closing in favor of #4573

PR #4132, opened 2023-07-27
Add tower logs checkpoint

pditommaso:
This PR refactor the Tower plugin cache manager turning it into a checkpoint mechanism for log and other tower-created files on a periodic basis. 
  
The overall idea is to rely on the nf-cloudcache plugin to store nextflow cache metadata in the cloud object storage, and use the logs checkpoint to backup logs files periodically in the object storage. This should prevent the lost of those files when the execution is interrupted abrutely


plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFactory.groovy
```diff
@@ -39,13 +39,25 @@ class TowerFactory implements TraceObserverFactory {
...
         if( !isEnabled )
             return Collections.emptyList()
 
+        final result = new ArrayList(1)
+        // create the tower client
+        final tower = createTowerClient(session, config)
+        result.add(tower)
+        // create the logs checkpoint
+        if( env.containsKey('NXF_CLOUDCACHE_PATH') )
+            result.add( new LogsCheckpoint() )
```
bentsherman, line 52: Why not enable the logs checkpointing all the time?

pditommaso: This is mostly needed when logs are created in the container and need to be copied in the remote work dir (to be accessed by Tower). it's not required in a normal usage scenario

PR #4142, opened 2023-08-01
Add `-value` option to `config` command

### LINKED ISSUE #4138: Fetch a supplied property from a config file via the Nextflow CLI
## New feature

Fetch a named property from a resolved Nextflow config file via the Nextflow CLI. The feature would make it easier for users to utilise config files outside of a nextflow script. 

Note that I originally [suggested this feature on Slack](https://nextflow.slack.com/archives/C02T98A23U7/p1690796177956779), and @ewels recommended I raise a feature request here. 

## Usage scenario 

Retrieving config file properties outside of nextflow scripts, particularly in an automated setting (eg in pre/post-processing routines and testing). For example, our codebase has a number of dev/testing tools that share common parameters with our pipeline, and we would like to centralise these configurations with the nextflow config. 

It would also make it easy for users to quickly lookup values as they are resolved in the pipeline. This is especially useful in complex configs. 

## Suggest implementation 

Modify `CmdConfig.groovy` to accept a new `-property <name>` optional. The `<name>` can reference nested properties using dot syntax:
```console
$ nextflow config -property aws.region
eu-west-1
```

If the property does not exist, an error message with non-zero exit code should be returned. For example:
```console
$ nextflow config -property does.not.exist
Error: does.not.exist does not exist in the config file. 
$ echo $?
2 
```

Evaluation of the property should adhere to the `-profile` optional if defined. For example:
```console
$ nextflow config -property process.executor -profile standard
local
$ nextflow config -property process.executor -profile cluster
sge
```

afishman:
Close #4138 

The config CLI `-property` optional has been implemented here. I have also updated the docs accordingly. I tested the feature with the following config and test script:

## `test.config`
```groovy
some {
    nullval = null
    nested = [
        key: "value"
    ]
}

profiles {
    myprofile {
        some.nested.key = "alt-value"
    }
}
```

## `test.bash`
```bash
echo "Running tests..."

# Run tests
echo "Should print 'value'"
./launch.sh -C test.config config -property some.nested.key
echo ""

echo "Should print 'alt-value'"
./launch.sh -C test.config config -property some.nested.key -profile myprofile
echo ""

echo "Should print 'null'"
./launch.sh -C test.config config -property some.nullval -profile myprofile
echo ""

echo "Should return a non-zero exit code"
./launch.sh -C test.config config -property does.not.exist
echo "Exit Code: $?"
```

## Output
```
Running tests...
Should print 'value'
value
Should print 'alt-value'
alt-value
Should print 'null'
null
Should return non-zero exit code error
Property 'does.not.exist' not found
Exit Code: 1
```

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -406,7 +406,9 @@ The `config` command is used for printing the project's configuration i.e. the `
 : Print config using Java properties notation.
 
 `-property`
-: Prints a config property using Java properties notation
+: :::{versionadded} 23.08.0-edge
+  :::
+: Print the value of a config property, or fail if the property is not defined.
 
 `-a, -show-profiles`
 : Show all configuration profiles.
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -405,6 +405,9 @@ The `config` command is used for printing the project's configuration i.e. the `
 `-properties`
 : Print config using Java properties notation.
 
+`-property`
+: Prints a config property using Java properties notation
+
 `-a, -show-profiles`
 : Show all configuration profiles.
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -453,7 +453,7 @@ process.executor = local
 Print out the value of a specific configuration property.
 
 ```console
-$ nextflow config -property process.executor
+$ nextflow config -value process.executor
 local
 ```
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -405,16 +405,16 @@ The `config` command is used for printing the project's configuration i.e. the `
 `-properties`
 : Print config using Java properties notation.
 
-`-property`
-: :::{versionadded} 23.08.0-edge
-  :::
-: Print the value of a config property, or fail if the property is not defined.
-
 `-a, -show-profiles`
 : Show all configuration profiles.
 
 `-sort`
 : Sort config attributes.
+
+`-value`
+: :::{versionadded} 23.08.0-edge
+  :::
+: Print the value of a config option, or fail if the option is not defined.
 
 **Examples**
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -450,6 +450,13 @@ docker.enabled = true
 process.executor = local
 ```
 
+Print out a single config property.
+
+```console
+$ nextflow config -property process.executor
+local
+```
+
 Print out all profiles from the project's configuration.
 
 ```console
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -450,7 +450,7 @@ docker.enabled = true
 process.executor = local
 ```
 
-Print out a single config property.
+Print out the value of a specific configuration property.
 
 ```console
 $ nextflow config -property process.executor
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdConfig.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdConfig.groovy
@@ -59,7 +59,7 @@ class CmdConfig extends CmdBase {
     @Parameter(names = '-sort', description = 'Sort config attributes')
     boolean sort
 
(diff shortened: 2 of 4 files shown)


bentsherman:
Can you follow the instructions [here](https://github.com/nextflow-io/nextflow/pull/4142/checks?check_run_id=15523555214) to add the DCO correctly? I would just undo your commits, add a single signed commit, and force push it to github.

Also, please revert the edits to `Const.groovy`. They are automatically added by the build process but should not be committed.

afishman:
Sure thanks, just made those changes.

bentsherman (approved):

modules/nextflow/src/main/groovy/nextflow/cli/CmdConfig.groovy
```diff
@@ -123,6 +135,21 @@ class CmdConfig extends CmdBase {
...
+     * Prints a property of a {@link ConfigObject}.
+     *
+     * @param config The {@link ConfigObject} representing the parsed workflow configuration
+     * @param name The {@link String} representing the property name using dot notation
+     * @param output The stream where output the formatted configuration notation
+     */
+    @PackageScope void printProperty(ConfigObject config, String name, OutputStream output) {
+        final map = config.flatten()
+        if( !map.containsKey(name) )
+            throw new IllegalArgumentException("Property '$name' not found")
```
pditommaso, line 148:
```suggestion
            throw new AbortOperationException("Configuration property '$name' not found")
```

Using `AbortOperationException` should provide a slightly better error message

docs/cli.md
```diff
@@ -445,6 +450,13 @@ docker.enabled = true
 process.executor = local
 ```
 
+Print out a single config property.
```
pditommaso, line 453:
```suggestion
Print out the value of a specific configuration property.
```

modules/nextflow/src/main/groovy/nextflow/cli/CmdConfig.groovy
```diff
@@ -50,6 +50,9 @@ class CmdConfig extends CmdBase {
     @Parameter(names=['-profile'], description = 'Choose a configuration profile')
     String profile
 
+    @Parameter(names = '-property', description = 'Print the value of a config property, or fail if the property is not defined.')
```
pditommaso, line 53: I was thinking if it should not be used something more idiomatic, e.g `-value` or `-eval`? thoughts ?

afishman:
Thanks for the feedback, I have updated and added an example to the docs. Looks like a few of the `CmdConfig` unit tests are failing.

modules/nextflow/src/main/groovy/nextflow/cli/CmdConfig.groovy
afishman: Also makes sense. My thinking was that `-property` is most appropriate because there is already a `-properties` option.

modules/nextflow/src/main/groovy/nextflow/cli/CmdConfig.groovy
bentsherman: My bad, @afishman you had it right the first time 😄

modules/nextflow/src/main/groovy/nextflow/cli/CmdConfig.groovy
```diff
@@ -62,6 +59,8 @@ class CmdConfig extends CmdBase {
     @Parameter(names = '-sort', description = 'Sort config attributes')
     boolean sort
 
+    @Parameter(names = '-value', description = 'Print the value of a config option, or fail if the option is not defined.')
```
pditommaso, line 62:
```suggestion
    @Parameter(names = '-value', description = 'Print the value of a config option, or fail if the option is not defined')
```

PR #4150, opened 2023-08-03
Fix strict mode docs

### LINKED ISSUE #4148: failOnMismatch without remainder cannot be set to false in strict mode
The documentation in edge for Nextflow's strict mode states:

> When using the join operator, the failOnMismatch option is true by default (unless remainder is also true)

This would imply to me that `failOnMismatch` could still be set to `false` by a developer regardless of strict mode, but that is not the case [as the `failOnMismatch` setting is overridden to `true` when in strict mode (if remainder is false)](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/extension/JoinOp.groovy#L67).

```
this.failOnMismatch = params?.failOnMismatch || (!this.remainder && NF.isStrictMode())
```

It is unclear to me whether this is a bug or a design decision? On the one hand, if I have explicitly allowed mismatches with `failOnMismatch: false` I would not expect strict mode to override me and raise an error anyway; but perhaps this was a decision to make a developer explicitly handle mismatches (eg. by keeping a remainder and discarding entries with `null`)?

If the former, in the case of strict mode, Nextflow should not set `failOnMismatch` to `true` if a value of `false` has been explicitly provided by the developer. If the latter, perhaps Nextflow should raise an error when trying to set `failOnMismatch: false` when using strict mode ([similar to what is done when trying to set remainder && failOnMismatch](https://github.com/nextflow-io/nextflow/blob/19a72c402495e72df0140c6af05d299330fa5ba9/modules/nextflow/src/main/groovy/nextflow/extension/JoinOp.groovy#L70)); and the documentation for strict mode should perhaps read:

> When using the join operator, a mismatch is not allowed unless remainder is also true

bentsherman:
Close #4148 

When I documented strict mode, I said that certain options are true "by default", when in fact they are true regardless of any user setting.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1713,11 +1713,11 @@ Some features can be enabled using the `nextflow.enable` and `nextflow.preview` 
 
   - When merging params from a config file with params from the command line, Nextflow will fail if a param is specified from both sources but with different types
 
-  - When using the `join` operator, the `failOnDuplicate` option is `true` by default
+  - When using the `join` operator, the `failOnDuplicate` option is `true` regardless of any user setting
 
-  - When using the `join` operator, the `failOnMismatch` option is `true` by default (unless `remainder` is also `true`)
+  - When using the `join` operator, the `failOnMismatch` option is `true` (unless `remainder` is also `true`) regardless of any user setting
 
-  - When using the `publishDir` process directive, the `failOnError` option is `true` by default
+  - When using the `publishDir` process directive, the `failOnError` option is `true` regardless of any user setting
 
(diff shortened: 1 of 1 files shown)


jchorl:
Does this mean that it is impossible to override `failOnMismatch` in strict mode?

As a user, that is surprising to me. I would expect strict-mode to enforce better defaults and validate configuration more strictly, but I would not expect it to silently ignore certain settings.

bentsherman:
I agree it is strange 🤷‍♂️ Maybe it was not intentional

PR #4153, opened 2023-08-03
Update tip about modifying maps

### LINKED ISSUE #4146: What is an "update" map mentioned in the Language basics:Maps section of the docs?
The Language basics:Maps section of the docs has the following Tip:

`Tip!: Appending an “update” map is a safer way to modify maps in Nextflow, specifically when passing maps through channels. This way, any references to the original map elsewhere in the pipeline won’t be modified.`
https://www.nextflow.io/docs/latest/script.html#maps

I understand this to mean making a deep copy of a map rather than adding a new reference (which I agree can be important).  Is this what was intended?  Unfortunately, I can not figure out what the document means by an "update" map.  I am not a native Groovy developer, but I searched the Nextflow documentation, google, ChatGPT, and the official Groovy Maps documentation (https://groovy-lang.org/groovy-dev-kit.html).  

This tip should be clarified, perhaps with a link to the documentation for "update" maps?

bentsherman:
Close #4146

### DIFF
--- a/docs/script.md
+++ b/docs/script.md
@@ -114,7 +114,7 @@ new_scores = scores + ["Pete": 3, "Cedric": 120]
 When adding two maps, the first map is copied and then appended with the keys from the second map. Any conflicting keys are overwritten by the second map.
 
 :::{tip}
-Copying a map with the `+` operator is a safer way to modify maps in Nextflow, specifically when passing maps through channels. This way, any references to the original map elsewhere in the pipeline won't be modified.
+Copying a map with the `+` operator is a safer way to modify maps in Nextflow, specifically when passing maps through channels. This way, a new instance of the map will be created, and any references to the original map won't be affected.
 :::
 
 Learn more about maps:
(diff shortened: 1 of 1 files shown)


docs/script.md
```diff
@@ -114,7 +114,7 @@ new_scores = scores + ["Pete": 3, "Cedric": 120]
 When adding two maps, the first map is copied and then appended with the keys from the second map. Any conflicting keys are overwritten by the second map.
 
 :::{tip}
-Appending an "update" map is a safer way to modify maps in Nextflow, specifically when passing maps through channels. This way, any references to the original map elsewhere in the pipeline won't be modified.
+Copying a map with the `+` operator is a safer way to modify maps in Nextflow, specifically when passing maps through channels. This way, any references to the original map elsewhere in the pipeline won't be modified.
```
pditommaso, line 117:
```suggestion
Copying a map with the `+` operator is a safer way to modify maps in Nextflow, specifically when passing maps through channels. This way, a new instance of the map object will be created and therefore any references to the original map elsewhere in the pipeline won't be affected.
```

pditommaso (approved):

PR #4155, opened 2023-08-04
Save aws batch instance id to trace record

bentsherman:
Nextflow already fetches the instance id in order to fetch the machine info, it just wasn't saving it. This PR saves the instance id to the `hostname` field of the trace record. Set `aws.batch.fetchInstanceType = true` to enable it.

If you want to avoid changing the CloudMachineInfo, an alternative approach would be to separate the calls to fetch the instance id vs the machine info. That would require more changes but happy to do it if you prefer.

Quick test:
```console
$ ../launch.sh run hello -profile awsbatch -with-trace
N E X T F L O W  ~  version 23.07.0-edge
Launching `https://github.com/nextflow-io/hello` [nostalgic_goodall] DSL2 - revision: 1d71f857bb [master]
[ce/bff2a9] process > sayHello (2) [100%] 4 of 4 ✔
Hello world!

Hola world!

Ciao world!

Bonjour world!

$ cat trace-20230804-48381761.txt 
task_id hash    native_id       %cpu    peak_rss        realtime        hostname
3       47/e2dabf       504ded65-e21b-49dd-b193-85810f1fdf6a    -       -       454ms   i-01a2837115229b513
4       c8/c36c49       4d5557a2-84bc-4df4-93e7-bc136319e6af    -       -       461ms   i-00ff2821fe2499302
1       ad/ac4632       9ac30279-42ee-4307-94c1-bbe0afafe940    -       -       2.9s    i-01a2837115229b513
2       ce/bff2a9       3053e752-9577-432f-8b68-69e03be09136    -       -       705ms   i-00ff2821fe2499302
```

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -169,6 +169,10 @@ The following settings are available:
 `aws.batch.delayBetweenAttempts`
 : Delay between download attempts from S3 (default: `10 sec`).
 
+`aws.batch.fetchInstanceType`
+: When `true` fetches the instance type and instance ID of each task (default: `false`).
+: The instance ID can be accessed from the trace file under the `hostname` field.
+
 `aws.batch.jobRole`
 : The AWS Job Role ARN that needs to be used to execute the Batch Job.
 
--- a/docs/tracing.md
+++ b/docs/tracing.md
@@ -304,8 +304,12 @@ The following table shows the fields that can be included in the execution repor
 
 `hostname`
 : :::{versionadded} 22.05.0-edge
+  Supported by Kubernetes with `k8s.fetchNodeName = true`.
   :::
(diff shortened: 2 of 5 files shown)


bentsherman:
Thinking about Tower, to be honest it would be better to save the instance id to Tower alongside the other cloud info, and let Tower compute the total cost. The issue with saving to the trace file is that the other cloud machine info cannot be saved to the trace file currently, and even if it did, the user would still have to write some code to compute the cost from the raw data, whereas Tower is already doing it.

bentsherman:
Noting some internal discussions here. This feature is blocked until the platform's data layer is upgraded to better support additional task metadata. We could merge it now just for Nextflow, but Paolo doesn't want to do this until we can take full advantage of it in the platform (i.e. improve cost estimate, traceability)

pditommaso:
Closing until we don't have a concrete use case for it

PR #4159, opened 2023-08-07
more clear explanation about join() operators

dxu104:
I propose adding a line to clarify this as follows:

In the output, items from the 'left' channel always precede the corresponding items from the 'right' channel."

And @mribeirodantas mentioned "the concat operator has an equivalent comment, so I  guess join operator deserves the same." Thank you Marcel

This would make it clearer for new users and provide a more complete understanding of the function's behavior. Please let me know if this change would be acceptable.

### DIFF
--- a/docs/operator.md
+++ b/docs/operator.md
@@ -952,6 +952,10 @@ left.join(right).view()
 [X, 1, 4]
 ```
 
+:::{note}
+In the output, items from the 'left' channel always precede the corresponding items from the 'right' channel."
+:::
+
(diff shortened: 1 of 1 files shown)


pditommaso:
It's more clear this note to me. In the `concat` the sematic is that if you have the left channel with `[a,b,c]` and the right with `[x,y,z]`, you will get [a,b,c,x,y,z]`. 

In this case, I guess you mean the resulting tuple will be composed with one item from the left channel and the second from the right channel. I'd argue that the expected semantic for the join operation, not sure it's needed to explicitate.

PR #4160, opened 2023-08-07
Enable cloud cache based on environment variable

bentsherman:
By the way, tested with nf-core/rnaseq and it works 👍


pditommaso (changes requested):
Worth adding a unit test. There are already some examples

pditommaso (approved):

PR #4163, opened 2023-08-09
more clear explanation about splitCsv() operators

dxu104:
From the provided official documentation, I noticed that the description of how the `.splitCsv()` operation handles CSV data and the nature of its returned data structures is somewhat ambiguous. The documentation showcases examples with and without the `header` parameter, but it doesn't explicitly mention the exact type of returned data (`List` vs`LinkedHashMap`).

I believe that such clarity would be invaluable to readers aiming to understand the `.splitCsv()` operation fully.  For the sake of clarity and comprehensiveness in the documentation, I propose the following clarifications:

**- If the header option is not used, .splitCsv() returns each row as a List.
- If header: true is used, or if a list of strings is provided for the header, then .splitCsv() returns each row as a LinkedHashMap**

Incorporating this information into the official documentation would offer readers a more lucid and actionable understanding.


bentsherman:
Fwiw I think I have done this in [#3959](https://github.com/nextflow-io/nextflow/pull/3959/files#diff-6af56712b9b41089d7acb8cb87a03b1ba65ada8f41bef38302034a7938155dfaR1101-R1103)

dxu104:
Thank you for your mention. But I did not find  your pull request related with mine. Do you think my pr can be merge

docs/operator.md
```diff
@@ -1384,6 +1384,10 @@ Channel
     .splitCsv(header: ['col1', 'col2', 'col3'], skip: 1 )
     .view { row -> "${row.col1} - ${row.col2} - ${row.col3}" }
 ```
+:::{note}
+- If the header option is not used, .splitCsv() returns each row as a List.
+- If header: true is used, or if a list of strings is provided for the header, then .splitCsv() returns each row as a LinkedHashMap
```
pditommaso, line 1389:
```suggestion
- By default, the `splitCsv` operator returns each row as a *list* object. Items are accessed by using the 0-based column index.
- When the `header` is specified each row is returned as a *map* object (also known as dictionary). Items are accessed via the corresponding column name. 
```

PR #4165, opened 2023-08-09
more clear explanation about map() operators

dxu104:
In Groovy, the default Map implementation is LinkedHashMap. As a result, when you declare a Map without specifying a type, Groovy will use LinkedHashMap by default. This is the reason why the input and output order is consistent here. While Java's Map is unordered, Nextflow, due to its pipeline nature, defaults to LinkedHashMap to preserve order.

I propose the following clarifications:

In Groovy, the default Map is a LinkedHashMap, ensuring consistent input and output order. This is especially important in Nextflow's pipeline context, where order preservation is often needed.

### DIFF
--- a/docs/operator.md
+++ b/docs/operator.md
@@ -1026,6 +1026,10 @@ Channel
 Done
 ```
 
+:::{note}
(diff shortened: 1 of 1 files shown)


pditommaso:
Indeed

dxu104:
Thank you for your reminder. They are not related with each other. I will close this PR

PR #4171, opened 2023-08-09
Allow secrets to be used in pipeline script

### LINKED ISSUE #3262: Provide a mechanism for using secrets in `onComplete` and `onError`
## Usage scenario 

Secrets can currently either be used in the configuration (but not assigned to `params`) or accessed in processes; it would be stellar to also be able to access them in the `onComplete` (and `onError`) handlers. That's the typical place to send pipeline reports etc. which will need access to authentication information.

Alternatively, if it was possible to define custom configuration scopes, then I could assign secrets just like to the built-in scopes and use that in my handlers. Maybe this is already possible via plugins?

## Suggest implementation 

I don't know about the nextflow code itself but the usage could look like either allowing direct usage:

```nextflow
workflow.onComplete {
    sendChat('Pipeline status', secrets.API_TOKEN)
}
```

or via a custom config scope and handler (possibly defined in a plugin):

`nextflow.config`:
```nextflow
chat {
    authentication = secrets.API_TOKEN
}

workflow.onComplete = {
    sendChat('Pipeline status')
}
```

bentsherman:
Close #3262


pditommaso:
@marcodelapierre can you please review and test?

marcodelapierre:
Sure!

docs/secrets.md
```diff
@@ -71,3 +71,20 @@ The secrets are made available in the process context running the command script
...
+Secrets can be accessed in the pipeline script using the `secrets` variable. For example:
+
+```groovy
+workflow.onComplete {
+    println("The secret is: ${secrets.MY_SECRET}")
+}
+```
+
+:::{note}
+This feature is only available when using the local or grid executors (Slurm, Grid Engine, etc). The AWS Batch executor allows the use of secrets when deploying the pipeline execution via [Nextflow Tower](https://seqera.io/blog/pipeline-secrets-secure-handling-of-sensitive-information-in-tower/).
```
pditommaso, line 89: Should not this work also with Batch when using xpack?

bentsherman: I was going to say, it should be "Seqera Platform" instead of Nextflow Tower

pditommaso: Right 👉  0d1dad3f

PR #4173, opened 2023-08-10
Fix typos in source code comments

mribeirodantas:
Replace typos in source code comments and output messages. I avoided changing outside comments/output messages as changing those could break code.


pditommaso:
What a mess! 😆 i've reverted th `isChunckComplete` to address it separately. thanks!

mribeirodantas:
🥳

PR #4177, opened 2023-08-10
Allow secrets to be used with includeConfig

### LINKED ISSUE #3502: Unable to use secrets in the includeConfig directive
We would like to be able to use config from a separate private repository and use it within the includeConfig context. This is possible if we pass the pat/token to the raw URL link but that means exposing the token on the raw text file. We tried to run this using secrets but it doesn't look like the includeConfig directive is able to parse/access the content of the secret.

```
Unable to parse config file: '/Users/mlpanganiban/nf/nextflow.config'

  Illegal character in query at index 85: https://raw.githubusercontent.com/markpanganiban/nf-test/master/ubuntu.config?token=${token}


------


Unable to parse config file: '/Users/mlpanganiban/nf/nextflow.config'

  Compile failed for sources FixedSetSources[name='/groovy/script/Script3B7B1D673C896BC4338945B5D6A2FB78/_nf_config_c782925a']. Cause: org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
  /groovy/script/Script3B7B1D673C896BC4338945B5D6A2FB78/_nf_config_c782925a: 51: Unexpected character: '\' @ line 51, column 100.
     aster/ubuntu.config?token=' + \${TOKEN}
                                   ^
```

bentsherman:
Close #3502 

The easiest way to make it work is to evaluate the secrets eagerly instead of through the `SecretsContext` which evaluates later when the secret provider has been bound.

To be honest, I'm not sure what value the `SecretsContext` is providing. The secrets provider can be initialized at any point, and I don't see any logic to hide secrets when they are printed, so I don't see any disadvantage to just evaluating the secrets eagerly. But I may be missing something.


bentsherman:
Btw it does work:

```groovy
// nextflow.config
includeConfig "https://raw.githubusercontent.com/markpanganiban/nf-test/master/ubuntu.config?token=${secrets.MY_TOKEN}"
```

```console
$ nextflow secrets set MY_TOKEN foo
$ nextflow run main.nf

  Config file does not exist: https://raw.githubusercontent.com/markpanganiban/nf-test/master/ubuntu.config?token=foo
```

pditommaso:
The rationale of `SecretContext` was to prevent the rendering of the secret values when using  `nextflow config` for security purposes. 

It turns out it can be simplified using a last transformation.

bentsherman:
Closing in favor of #4840

PR #4178, opened 2023-08-11
Support resource labels for Azure Batch

### LINKED ISSUE #3248: Azure metadata tags (resource labels) propagation for child jobs
## New feature

With the recent improvements to resource labelling https://github.com/nextflow-io/nextflow/commit/d64eeffcb0b4c7d2d2773f0492fff15e05399e21  and https://github.com/nextflow-io/nextflow/pull/2853 , it'd be great if the child jobs can be tagged on Azure Batch executor as well.

## Usage scenario 

When submitting jobs to Azure Batch with Nextflow, the tags provided as the process level should be propagated to the generated job definitions.

## Suggest implementation 

We would need to implement the current labelling mechanism within the AzBatchService, by (probably) enhancing the [createPool](https://github.com/nextflow-io/nextflow/blob/master/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L616) implementation. 

The metadata tags would show up at the job metadata menu on the portal

image

bentsherman:
Close #3248 

Based on discussions in the issue and with Adam, it seems cost analysis in Azure is tied to pools, and job metadata is not propagated to VMs. As a result, this PR applies `resourceLabels` to the pool when it is created, and the docs make clear that resource labels in Azure should be defined once for all processes, or else they will be non-deterministic.

Happy to be proven wrong if anyone knows better. But this feature is still useful for pipeline-level labels, like pipeline name, user, project, etc.

@adamrtalbot can you test it?

### DIFF
--- a/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
+++ b/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
@@ -564,9 +564,11 @@ class AzBatchService implements Closeable {
             throw new IllegalArgumentException(msg)
         }
 
-        final key = CacheHelper.hasher([vmType.name, opts]).hash().toString()
+        def metadata = task.config.getResourceLabels()
+
+        final key = CacheHelper.hasher([vmType.name, opts, metadata]).hash().toString()
         final poolId = "nf-pool-$key-$vmType.name"
-        return new AzVmPoolSpec(poolId: poolId, vmType: vmType, opts: opts)
+        return new AzVmPoolSpec(poolId: poolId, vmType: vmType, opts: opts, metadata: metadata)
     }
 
     protected void checkPool(CloudPool pool, AzVmPoolSpec spec) {
@@ -619,7 +621,7 @@ class AzBatchService implements Closeable {
         def pool = getPool(spec.poolId)
         if( !pool ) {
             if( config.batch().canCreatePool() ) {
-                createPool(spec, task)
+                createPool(spec)
             }
             else {
                 throw new IllegalArgumentException("Can't find Azure Batch pool '$spec.poolId' - Make sure it exists or set `allowPoolCreation=true` in the nextflow config file")
@@ -677,7 +679,7 @@ class AzBatchService implements Closeable {
                 .withContainerConfiguration(containerConfig)
     }
 
-    protected void createPool(AzVmPoolSpec spec, TaskRun task) {
+    protected void createPool(AzVmPoolSpec spec) {
 
         def resourceFiles = new ArrayList(10)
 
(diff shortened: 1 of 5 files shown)


plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -618,7 +619,7 @@ class AzBatchService implements Closeable {
         def pool = getPool(spec.poolId)
         if( !pool ) {
             if( config.batch().canCreatePool() ) {
-                createPool(spec)
+                createPool(spec, task)
```
pditommaso, line 622: It would make sense to keep the labels in the AzVmPoolSpec object, likely when it's created here

https://github.com/nextflow-io/nextflow/blob/3051cd13da0298eaca142509ac05c7080a1c4328/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L589-L607

bentsherman: Makes sense. For the auto pool spec, we could add the metadata to the unique hash so that different resource labels trigger different auto pools. That could lead to many different pools being created, and I don't know whether that would be problematic.

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -563,9 +564,11 @@ class AzBatchService implements Closeable {
             throw new IllegalArgumentException(msg)
         }
 
-        final key = CacheHelper.hasher([vmType.name, opts]).hash().toString()
+        def metadata = task.config.getResourceLabels()
+
```
pditommaso, line 568:
```suggestion
        final metadata = task.config.getResourceLabels()
```

pditommaso:
This [test](https://github.com/nextflow-io/nextflow/blob/3051cd13da0298eaca142509ac05c7080a1c4328/plugins/nf-azure/src/test/nextflow/cloud/azure/batch/AzBatchServiceTest.groovy#L324-L351) could be updated

bentsherman:
@adamrtalbot what do you think about creating different auto pools for different resource labels? This is how it currently works, if two processes have different resource labels, Nextflow will create two different pools for them.

This would allow the resource labels to be process-specific, but could potentially create lots of otherwise identical pools. I'm wondering if you think it's worth having.

bentsherman:
My concern is that it could lead to lower resource utilization, since each process would have its own pool rather than many processes sharing one or a few pools, which are more likely to be fully utilized the whole time.

bentsherman:
Wonderful, then this PR is ready to go

pditommaso (approved):

PR #4190, opened 2023-08-14
Fix `workflow.container` map resolution

### LINKED ISSUE #3394: Documentation vs reality: workflow.container

The metadata documentation indicates that `workflow.container` returns a map of **all** containers:

> Docker image used to run workflow tasks. When more than one image is used it returns a map object containing *[process name, image name]* pair entries.

A map is not returned when multiple containers defined, only the top level value when defined (`process.container`).

If top level `process.container` has not been defined we get `[:]`.

nextflow.config
```groovy
params {
    // output folder options
    outdir          = 'results'
    tracedir        = "${params.outdir}/pipeline_info"
}

docker.enabled = true

process {
    container     = 'ubuntu:18.04' // gives this value as result, comment to get `[:]`

    withName: bash_step {
        executor = 'local'
        container = null
    }

    withName: job_a {
        container = 'quay.io/wtsicgp/expansion_hunter:5.0.0'
    }

    withName: job_b {
        container = 'quay.io/wtsicgp/pcap-core:5.7.0'
    }
}
```

main.nf
```groovy
#!/usr/bin/env nextflow
nextflow.enable.dsl=2

process bash_step {
    input:
        val(example)
    output:
        path '*.out', emit: wfl
    script:
        """
        echo ${workflow.container} > ${example}.out
        """
}

process job_a {
    input:
        path(wfl)
    output:
        path '*.out.gz', emit: gzip
    script:
        """
        gzip -c ${wfl} > ${wfl}.gz
        """
}

process job_b {
    input:
        path(wfl)
        path(gzip)
    
    output:
        path 'compress.comp'

    publishDir "${params.tracedir}", mode: "copy"

    shell:
        '''
        echo native: $(wc -c !{wfl}) > compress.comp
        echo compressed: $(wc -c !{gzip}) >> compress.comp
        echo "## content follows ##" >> compress.comp
        cat !{wfl} >> compress.comp
        echo "## EOC ##" >> compress.comp
        '''
}

workflow {
    main:
        bash_step('bob')
        job_a(bash_step.out.wfl)
        job_b(bash_step.out.wfl, job_a.out.gzip)
}
```

Run with:
```
nextflow run -stub-run main.nf
# cat results/pipeline_info/compress.comp
```

Gives:
```
native: 13 bob.out
compressed: 41 bob.out.gz
## content follows ##
ubuntu:18.04
## EOC ##
```

Would expect:

```
[bash_step : null, job_a : 'quay.io/wtsicgp/expansion_hunter:5.0.0', job_b: 'quay.io/wtsicgp/pcap-core:5.7.0']
```

(possibly no entry for `bash_step`)

As above

### Program output 

As above

### Environment 

* Nextflow version: 22.10.2.5832
* Java version: openjdk version "17.0.5" 2022-10-18
* Operating system: Linux (GitHub CodeSpaces)
* Bash version: GNU bash, version 5.1.4(1)-release (x86_64-pc-linux-gnu)

### Additional context

Appears related to #1758, if this is not a suitable way to introspect this value please update documentation.

bentsherman:
Close #3394 

Keeps the old `$<name>` syntax for backwards compatibility. See also: `Session::validateConfig()`

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -1268,7 +1268,14 @@ class Session implements ISession {
              * look for `container` definition at process level
              */
             config.process.each { String name, value ->
-                if( name.startsWith('$') && value instanceof Map && value.container ) {
+                if( name.startsWith('$') ) {
+                    name = name.substring(1)
+                }
+                else if( name.startsWith('withName:') ) {
+                    name = name.substring('withName:'.length())
+                }
+
+                if( value instanceof Map && value.container ) {
                     result[name] = resolveClosure(value.container)
                 }
             }
--- a/modules/nextflow/src/test/groovy/nextflow/SessionTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/SessionTest.groovy
@@ -525,25 +525,25 @@ class SessionTest extends Specification {
         when:
         text = '''
                 process {
-                    $proc1 { container = 'alpha' }
-                    $proc2 { container ='beta' }
+                    withName:'proc1' { container = 'alpha' }
+                    withName:'proc2' { container = 'beta' }
                 }
                 '''
         then:
-        new Session(cfg(text)).fetchContainers() == ['$proc1': 'alpha', '$proc2': 'beta']
+        new Session(cfg(text)).fetchContainers() == ['proc1': 'alpha', 'proc2': 'beta']
 
 
         when:
         text = '''
                 process {
-                    $proc1 { container = 'alpha' }
-                    $proc2 { container ='beta' }
+                    withName:'proc1' { container = 'alpha' }
+                    withName:'proc2' { container = 'beta' }
                 }
 
                 process.container = 'gamma'
                 '''
         then:
-        new Session(cfg(text)).fetchContainers() == ['$proc1': 'alpha', '$proc2': 'beta', default: 'gamma']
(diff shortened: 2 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -1268,7 +1268,14 @@ class Session implements ISession {
              * look for `container` definition at process level
              */
             config.process.each { String name, value ->
-                if( name.startsWith('$') && value instanceof Map && value.container ) {
+                if( name.startsWith('$') ) {
+                    name = name.substring(1)
+                }
```
pditommaso, line 1273: Let's comment this is legacy code that likely will be removed

modules/nextflow/src/main/groovy/nextflow/config/ConfigBuilder.groovy
```diff
@@ -787,7 +787,7 @@ class ConfigBuilder {
                 return true
 
             def result = process
-                            .findAll { String name, value -> name.startsWith('$') && value instanceof Map }
+                            .findAll { String name, value -> name.startsWith('withName:') && value instanceof Map }
```
pditommaso, line 790: Does this not break `$`-prefixed config?

bentsherman: It will cause containerized execution to fail unless a global container is specified. That is currently what happens with the `withName:` syntax. We can keep both if you want

bentsherman: Done

pditommaso (approved):

PR #4192, opened 2023-08-14
FIx nested InvocationTargetException

### LINKED ISSUE #3418: `Unexpected error [InvocationTargetException]` when calling included function in local function and included function throws error
### Expected behaviour:
Print error message of the original error.

### Actual behaviour:
Generic error message `Unexpected error [InvocationTargetException]` is printed.

Two files, `main.nf` and `funcs.nf`:
```
// main.nf
include { imported_func } from "./funcs.nf"

def local_func() {
    return imported_func()
}


workflow {
    local_func()
}
```
```
// funcs.nf
def imported_func() {
    error "error in imported_func"
}
```

### Program output 
```
N E X T F L O W  ~  version 22.10.0
Launching `main.nf` [goofy_pauling] DSL2 - revision: f561796fa1
Unexpected error [InvocationTargetException]

 -- Check script 'main.nf' at line: 5 or see '.nextflow.log' file for more details
```
### Environment 

* Nextflow version: 22.10.0
* Java version: `openjdk version "11.0.13" 2021-10-19`
* Operating system: Linux
* Bash version: `GNU bash, version 4.3.48(1)-release (x86_64-pc-linux-gnu)`

bentsherman:
Close #3418 

When a method invoked via reflection throws an exception, the exception is wrapped in an `InvocationTargetException`. `BaseScript` already has a catch block to unwrap this exception, but for some reason there are some cases where the exception is wrapped twice, such as the linked issue.

I tried burying the user exception in more function calls and workflow calls, but it was only ever wrapped once or twice. So it does not seem to depend on the call depth, but I can't explain why it is wrapped sometimes once and sometimes twice. So I just check both cases.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -1275,7 +1275,7 @@ class TaskProcessor {
 
 
     static String err0(Throwable e) {
-        final fail = e instanceof InvocationTargetException ? e.targetException : e
+        final fail = e instanceof InvocationTargetException ? e.cause : e
 
         if( fail instanceof NoSuchFileException ) {
             return "No such file or directory: $fail.message"
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -1275,7 +1275,7 @@ class TaskProcessor {
 
 
     static String err0(Throwable e) {
-        final fail = e instanceof InvocationTargetException ? e.cause : e
+        final fail = e instanceof InvocationTargetException ? e.targetException : e
 
         if( fail instanceof NoSuchFileException ) {
             return "No such file or directory: $fail.message"
--- a/modules/nextflow/src/main/groovy/nextflow/script/BaseScript.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/BaseScript.groovy
@@ -193,7 +193,10 @@ abstract class BaseScript extends Script implements ExecutionContext {
         }
         catch( InvocationTargetException e ) {
             // provide the exception cause which is more informative than InvocationTargetException
-            throw e.cause?.cause ?: e.cause ?: e
+            Throwable target = e
+            while( target.cause )
+                target = target.cause
+            throw target
         }
         finally {
             ExecutionStack.pop()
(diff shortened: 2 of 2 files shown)


bentsherman:
A more general solution...

```groovy
def err = e
while( err instanceof InvocationTargetException )
    err = err.cause
```

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -1275,7 +1275,7 @@ class TaskProcessor {
 
 
     static String err0(Throwable e) {
-        final fail = e instanceof InvocationTargetException ? e.targetException : e
+        final fail = e instanceof InvocationTargetException ? e.cause : e
```
pditommaso, line 1278: This code doesn't look related with the issue

pditommaso:
yeah, but it looks that it would make more sense 

```
def err = e
while( err instanceof InvocationTargetException )
    err = err.target
```

pditommaso:
Not according to the reality .. oops, screenshot was broken.

modules/nextflow/src/main/groovy/nextflow/script/BaseScript.groovy
```diff
@@ -191,9 +191,12 @@ abstract class BaseScript extends Script implements ExecutionContext {
         try {
             run0()
         }
-        catch(InvocationTargetException e) {
+        catch( InvocationTargetException e ) {
             // provide the exception cause which is more informative than InvocationTargetException
-            throw(e.cause ?: e)
+            Throwable target = e
+            while( target.cause )
```
bentsherman, line 197: I think we should still check `target instanceof InvocationTargetException`, because the target exception could have it's own cause

pditommaso: Agree

modules/nextflow/src/main/groovy/nextflow/script/BaseScript.groovy
```diff
@@ -191,9 +191,12 @@ abstract class BaseScript extends Script implements ExecutionContext {
         try {
             run0()
         }
-        catch(InvocationTargetException e) {
+        catch( InvocationTargetException e ) {
             // provide the exception cause which is more informative than InvocationTargetException
-            throw(e.cause ?: e)
+            Throwable target = e
+            while( target.cause instanceof InvocationTargetException )
```
bentsherman, line 197:
```suggestion
            while( target instanceof InvocationTargetException )
```

pditommaso: Then `do .. while`! 😎

bentsherman: It's the Lord's loop!

PR #4194, opened 2023-08-15
Add check for latest version

pditommaso:
This PR adds the check for latest version of Nextflow on startup, printing a notification message is available. 

The check can be disabled using the env variable `NXF_DISABLE_CHECK_LATEST`. 

However, with some basic test it looks like it takes ~4-500ms for a simple http request, which is too slow.


pditommaso:
Yep, an env variable.

bentsherman:
I just realized you already had the env variable... should have read the entire PR 🤦

bentsherman (approved):

bentsherman:
If you don't like how long the HTTP request takes, then I would just switch the env variable to enable it rather than disable it.

pditommaso:
Think it's mostly related to my slow connection. Can you please give it a try?

bentsherman:
```diff
# check for update
- check_latest
+ time check_latest
```

```console
$ for i in `seq 10`; do nextflow info 2>&1 | grep real ; done
real    0m0.112s
real    0m0.075s
real    0m0.072s
real    0m0.081s
real    0m0.070s
real    0m0.070s
real    0m0.084s
real    0m0.073s
real    0m0.075s
real    0m0.074s
```

PR #4202, opened 2023-08-18
the usage of .out syntax

dxu104:
Hi maintainers of Nextflow,

While working extensively with Nextflow, I realized that there was a gap in the documentation concerning the emit feature in processes. I found that new users might face challenges in understanding when and how to use the direct referencing versus the .out syntax, especially if they're not introduced explicitly. Therefore, I add a clear explanation about this point.

credit for @jfy133 Thank you for your help.


docs/dsl2.md
```diff
@@ -153,6 +153,12 @@ workflow {
     foo.out.samples_bam.view()
 }
 ```
+If you're referencing the named output right at the point where the process is being executed, you don't need to use the `.out` syntax. This is the more concise method. For instance:
+```groovy
```
ewels, line 157:
````suggestion

If referencing a named output as the process is invoked, you can use a more concise syntax:

```groovy
````

ewels (approved):
Nice! Minor rewording suggestion.

dxu104:
> Nice! Minor rewording suggestion.

@ewels  Highly appriciate your help!

PR #4214, opened 2023-08-24
Dry run

### LINKED ISSUE #844: "dry run" option?
When using the -resume option, it would be helpful to see which processes are going to be run and which will be reused from the cache. 
Additionally it would be useful to indicate which hashes have changed requiring a cached process to be run. I know this can be done manually with the -dump-hashes option but this is not very user friendly.

### LINKED ISSUE #4894: [feature] Resume: print cache invalidation reason to console
Idea for making broken `-resume` caches less frustrating: get Nextflow to identify the change that caused an invalidated cache, then print to the console. Give immediate feedback, rather than forcing the end user to dig into hashes etc to try to figure this out.

Related to https://github.com/nextflow-io/nextflow/pull/4214

bentsherman:
Close #844

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -1072,6 +1072,11 @@ The `run` command is used to execute a local pipeline script or remote pipeline 
 `-disable-jobs-cancellation`
 : Prevent the cancellation of child jobs on execution termination
 
+`-dry`
+: :::{versionadded} 25.10.0
+  :::
+: Print the tasks that will be executed, then exit without executing any tasks. Can be combined with `-resume` to test how much a previous run can be resumed.
+
 `-dsl1`
 : :::{deprecated} 23.09.0-edge
   :::
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -129,6 +129,11 @@ class Session implements ISession {
     boolean resumeMode
 
     /**
+     * whenever it has been launched as a dry run
+     */
+    boolean dryRun
+
+    /**
      * The folder where workflow outputs are stored
      */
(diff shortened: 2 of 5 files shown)


bentsherman:
Added basic cache invalidation reporting to this PR. It requires the cache to also load the index file of the previous run, so that when a cache entry can't be found, it can search the previous run for a matching task (i.e. matching process + tag) and report any differences between the two tasks.

Currently it only reports the different hashes and the scripts (if they are different), but could be expanded to include things like:
- container / conda / spack / modules config
- stdin
- input environment vars
- input files (compare metadata or checksum based on cache mode)

The last point requires the input file metadata to be saved to the cache as discussed for #3802 and #3849 . I would be fine to address that in a future iteration, since for now, even just reporting the changed hashes is a great improvement.

Other notes:
- we can use [java-diff-utils](https://github.com/java-diff-utils/java-diff-utils) to print a nice diff of the script text
- comparing the previous run adds a lot of read pressure to the cache store, I'm concerned it might slow down large runs a lot. I would consider enabling this "deep" cache analysis only when the dry run is enabled

docs/reference/cli.md
```diff
@@ -1072,6 +1072,11 @@ The `run` command is used to execute a local pipeline script or remote pipeline
 `-disable-jobs-cancellation`
 : Prevent the cancellation of child jobs on execution termination
 
+`-dry`
+: :::{versionadded} 25.10.0
+  :::
+: Print the tasks that will be executed, then exit without executing any tasks. Can be combined with `-resume` to test how much a previous run can be resumed.
```
christopher-hakkaart, line 1078:
```suggestion
: Print the tasks that will be executed, then exit without executing any tasks. Can be combined with `-resume` to test how much of a previous run can be resumed.
```

PR #4218, opened 2023-08-25
Add timeout operator

### LINKED ISSUE #2248: Adding ending condition for watchPath
## New feature

normally, a pipeline ends with all jobs done. It is usable adding condition to make watchPath end in a proper time. 

## Usage scenario 

by default, watchPath waits endlessly. 

waits until reaching triggering limits.
`watchPath("*.fastq", "times: 8")`

waits until deadline.
`watchPath("*.fastq", "duration: 3h")`

bentsherman:
Close #2248 

It's very close, but the only problem is that the workflow doesn't terminate. The log reaches the message `Execution complete -- Goodbye` at the very end, but then hangs forever.

@pditommaso can you tell why it hangs?

### DIFF
--- a/docs/channel.md
+++ b/docs/channel.md
@@ -378,6 +378,27 @@ Available options:
 `protocol`
 : Allow choosing the protocol for the resulting remote URLs. Available choices: `ftp`, `http`, `https` (default: `ftp`).
 
+(channel-interval)=
+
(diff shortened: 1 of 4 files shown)


bentsherman:
Meh, this isn't really needed

PR #4224, opened 2023-08-25
Fail if glob does not exist with `checkIfExists: true`

### LINKED ISSUE #4223: Incorrect checkIfExists results for file() method with glob pattern

When a glob path pattern is used for the file method, e.g. `file("*glob-non-existing-file.txt", checkIfExists:true)` , you would expect an exception to be thrown when not a single path matches the pattern. 

Instead the file method returns an empty list, as described in the [documentation](https://www.nextflow.io/docs/latest/script.html?highlight=checkifexists#files-and-i-o), but without an error.

I would expect something like this when there's no match:
`ERROR ~ No such file(s) or directory(-ies) : "*glob-non-existing-file.txt"`

1. main.nf
`workflow {
    Channel.value(file("*glob-non-existing-file.txt", checkIfExists:true)).view()
}`
2. run
`nextflow run main.nf`

### Program output 
`[]`

### Environment 

* Nextflow version: 23.04.1]
* Java version: openjdk 17.0.3-internal 2022-04-19
* Operating system: Linux
* Bash version: 5.0.17

### Additional context
Somewhat similar to #3323

bentsherman:
Close #4223


modules/nextflow/src/main/groovy/nextflow/Nextflow.groovy
```diff
@@ -135,7 +135,11 @@ class Nextflow {
         }
 
         // revolve the glob pattern returning all matches
-        return fileNamePattern(splitter, options)
+        final result = fileNamePattern(splitter, options)
+        if( glob && options?.checkIfExists && result.isEmpty() )
+            throw new NoSuchFileException(path.toString())
```
jorgee, line 140: Maybe, we could add a new case in the [NexflowTest.'should check if file exists'](https://github.com/nextflow-io/nextflow/blob/aacb69f8b6ff55d1cb09cb53c6ac5e41781ab6a6/modules/nextflow/src/test/groovy/nextflow/NextflowTest.groovy#L350) to cover this case. Something like this
```
when:
        Nextflow.file("$folder/*.txt", checkIfExists: true)
 then:
        e = thrown(NoSuchFileException)
        e.message == "$folder/*.txt".toString()
```

bentsherman: Good catch

jorgee (approved):

PR #4228, opened 2023-08-25
Fix `executor.queueSize = 0` ignored

bentsherman:
Setting `executor.queueSize` to 0 is equivalent to having no limit. However, due to an incorrect null check, the queue size is ignored and falls back to the default value if it is set to 0 in the config. This PR fixes that, as well as a null reference error that happens when the queue size is 0.


pditommaso (changes requested):
Some tests are needed to valide this

pditommaso (approved):

PR #4229, opened 2023-08-26
Document aws.batch.retryMode = 'built-in'.

rorynolan:
This undocumented retryMode is handy. This is a retryMode that will catch any exception type. The AWS modes only catch specific types. I needed something to catch "No space left on device." This 'built-in' mode does that, nothing else does.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -189,7 +189,7 @@ The following settings are available:
 : Max number of downloads attempts from S3 (default: `1`).
 
 `aws.batch.retryMode`
-: The retry mode configuration setting, to accommodate rate-limiting on [AWS services](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html) (default: `standard`)
+: The retry mode configuration setting, to accommodate rate-limiting on [AWS services](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html) (default: `standard`, other options: `legacy`, `adaptive`); this handling is delegated to AWS. To have Nextflow handle retries instead, use `built-in`.
 
 `aws.batch.schedulingPriority`
 : :::{versionadded} 23.01.0-edge


bentsherman:
That setting is used to retry API requests to AWS Batch, but it sounds like you are talking about a task failure, i.e. a task ran out of disk storage. Task failures are handled by a different mechanism, these things are not related.

rorynolan:
I am using amazon genomics CLI. It sets up the machine to attach new storage drives as current storage is filling up. However, sometimes it's a bit slow acquiring the new storage and a download can fail due to "no space left" and then succeed upon retry because by then the extra storage is attached. retryMode = 'build-in' will make the download retry where it otherwise would not. This saves the task. The task would have failed due to the download error but being able to retry the download saves it. 

```
    AwsBatchConfig(Map opts) {
        cliPath = parseCliPath(opts.cliPath as String)
        maxParallelTransfers = opts.maxParallelTransfers as Integer ?: MAX_TRANSFER
        maxTransferAttempts = opts.maxTransferAttempts as Integer ?: defaultMaxTransferAttempts()
        delayBetweenAttempts = opts.delayBetweenAttempts as Duration ?: DEFAULT_DELAY_BETWEEN_ATTEMPTS
        maxSpotAttempts = opts.maxSpotAttempts as Integer ?: DEFAULT_MAX_SPOT_ATTEMPTS
        volumes = makeVols(opts.volumes)
        jobRole = opts.jobRole
        logsGroup = opts.logsGroup
        retryMode = opts.retryMode ?: 'standard'
        shareIdentifier = opts.shareIdentifier
        schedulingPriority = opts.schedulingPriority as Integer ?: 0
        if( retryMode == 'built-in' )
            retryMode = null // this force falling back on NF built-in retry mode instead of delegating to AWS CLI tool
        if( retryMode && retryMode !in AwsOptions.VALID_RETRY_MODES )
            log.warn "Unexpected value for 'aws.batch.retryMode' config setting - offending value: $retryMode - valid values: ${AwsOptions.VALID_RETRY_MODES.join(',')}"

    }
```

```
String stageInputFile( Path path, String targetName ) {
        // third param should not be escaped, because it's used in the grep match rule
        def stage_cmd = opts.maxTransferAttempts > 1 && !opts.retryMode
                ? "downloads+=(\"nxf_cp_retry nxf_s3_download s3:/${Escape.path(path)} ${Escape.path(targetName)}\")"
                : "downloads+=(\"nxf_s3_download s3:/${Escape.path(path)} ${Escape.path(targetName)}\")"
        return stage_cmd
    }
```

You can see here that 'built-in' gets me a `nxf_cp_retry nxf_s3_download ` which saves my job.

pditommaso (approved):

pditommaso:
Thanks!

PR #4241, opened 2023-08-30
More helpful warnings when publish fails to create parent directory.

robsyme:
At the moment, if publishing a file fails because of permissions, Nextflow will return an error indicating the Access Denied exception, but will not indicate which file was the problem.

This commit adds the file name to the error message, so that the user can determine from the logs alone which file path was the problem. Catching this exception also triggers the safeProcessFile catch block which also gives the helpful message saying "Failed to publish file". It's not entirely clear to me why this catch block does not catch the Access Denied exception being thrown inside the `makeDirs` method.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
@@ -483,6 +483,9 @@ class PublishDir {
         catch( FileAlreadyExistsException e ) {
             // ignore
         }
+        catch( Exception e ) {
+            log.warn "Failed to create directory for publishing file: ${dir.toUriString()}"
+        }
         finally {
             makeCache.put(dir,true)
         }


robsyme:
I'd like to note that I'm very unclear why this change is necessary. From my reading of the code, `safeProcessFile` calls `processFile` which calls `makeDirs`, which throws the exception:

```
safeProcessFile
  | processFile
     | makeDirs -> throws the Exception
```

But I'm very unclear why this exception isn't caught by the try block
```groovy
protected void safeProcessFile(Path source, Path target) {
    try {
        processFile(source, target)
    }
    catch( Throwable e ) {
        log.warn "Failed to publish file: ${source.toUriString()}; to: ${target.toUriString()} [${mode.toString().toLowerCase()}] -- See log file for details", e
        if( NF.strictMode || failOnError){
            final session = Global.session as Session
            session?.abort(e)
        }
    }
}
```

Ben/Paolo, can you explain why the try block in the `safeProcessFile` method isn't catching the Exception?

bentsherman:
Sorry for letting this one fall through. Did you confirm that `makeDirs` was called via `processFile` ? Because it is also called via `apply0 > createPublishDirs`.

If that's not it, then I don't know why the exception wouldn't be caught. But in any case it would probably be helpful to have this warning directly in `makeDirs` so that it is properly logged

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
```diff
@@ -483,6 +483,9 @@ class PublishDir {
         catch( FileAlreadyExistsException e ) {
             // ignore
         }
+        catch( Exception e ) {
+            log.warn "Failed to create directory for publishing file: ${dir.toUriString()}"
```
pditommaso, line 487: But this changes the existing behavior which was causing the execution to stop, instead not only a warning is returned. Is that made on purpose?

bentsherman:
Closing without merging, since we weren't able to validate the change. Both `makeDirs` calls have `catch` blocks that should catch any errors and report the offending file

Feel free to re-open if you see the issue again and can provide a log

PR #4247, opened 2023-08-31
Checkout remote tag if checkout remote branch fails

### LINKED ISSUE #4246: nextflow pull -r needs to be run twice to pull remote tag
### Expected behavior

`nextflow pull -r <tag>` checks out a new tag from the remote repository.

### Actual behavior

* `nextflow pull -r <tag>` yields the following error: `Cannot find revision <tag> -- Make sure that it exists in the remote repository`
* A second `nextflow pull -r <tag>` is successful

```
# Check out a known past tag
nextflow pull -r v1.0.0 epi2me-labs/wf-basecalling
LOCAL_PATH=$(nextflow info epi2me-labs/wf-basecalling | grep 'local path' | sed -nE 's, local path  : (.*),\1,p')
cd ${LOCAL_PATH}
# Nuke a known "future" tag
git tag -d v1.0.1
# Try and check it out with nextflow pull
nextflow pull -r v1.0.1 epi2me-labs/wf-basecalling
# check out again, successfully
nextflow pull -r v1.0.1 epi2me-labs/wf-basecalling
```

### Environment 

* Nextflow version: 23.04.2
* Java version: OpenJDK 64-Bit Server VM Homebrew (build 17.0.3+0, mixed mode, sharing)
* Operating system: macOS
* Bash version: zsh 5.8 (x86_64-apple-darwin21.0)

### Additional context

Running Nextflow pull on an existing asset repo, triggers the following events:
* [AssetManager.download](https://github.com/nextflow-io/nextflow/blob/v23.04.2/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy#L575)
* [`git.checkout() .setName(revision) .call()`](https://github.com/nextflow-io/nextflow/blob/v23.04.2/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy#L622), which will fail because the repository has not been fetched yet
* [`checkoutRemoteBranch(revision)`](https://github.com/nextflow-io/nextflow/blob/v23.04.2/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy#L629C27-L629C47), triggered by the failure of the first `git.checkout()`
* [`fetch.call()`](https://github.com/nextflow-io/nextflow/blob/v23.04.2/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy#L941), where the remote repository is now fetched
* [A second and fatal `git.checkout()`](https://github.com/nextflow-io/nextflow/blob/v23.04.2/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy#L942), which fails because it tries to specifically track an upstream `origin/<revision>` branch (but revision is a tag)

When running the command a second time, the repository is successfully updated by the first `git.checkout().setName(revision)`.

Some initial thoughts on how to work around this:
* the repository should always be fetched before attempting the first git.checkout().setName(revision), or
* after failing to track a branch in `checkoutRemoteBranch`, `download` could attempt to checkout the revision as a tag

bentsherman:
Close #4246 

`checkoutRemoteBranch()` now falls back to checking for remote tag if the branch checkout fails (which it will if the revision is a tag).

For some reason though, the returned ref is null, even though a tag is considered a ref, so the message will say "checkout at null: revision [abcdef]". Not sure how to deal with that.

To test:
```bash
./launch.sh drop epi2me-labs/wf-basecalling
./launch.sh pull -r v1.0.0 epi2me-labs/wf-basecalling

cd $HOME/.nextflow/assets/epi2me-labs/wf-basecalling
git tag -d v1.0.1
cd -

./launch.sh pull -r v1.0.1 epi2me-labs/wf-basecalling
```


pditommaso (changes requested):
Any chance to add a unit test?

modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
```diff
@@ -580,4 +580,25 @@ class AssetManagerTest extends Specification {
...
+        def folder = tempDir.getRoot()
+        def token = System.getenv('NXF_GITHUB_ACCESS_TOKEN')
+        def manager = new AssetManager().build('nextflow-io/nf-test-branch', [providers: [github: [auth: token]]])
+
+        when:
+        manager.download("v0.1")
+        then:
+        folder.resolve('nextflow-io/nf-test-branch/.git').isDirectory()
+        and:
+        folder.resolve('nextflow-io/nf-test-branch/workflow.nf').text == "println 'Hello'\n"
```
pditommaso, line 596: it would be much better to use a local Git repo. There's an example here

https://github.com/nextflow-io/nextflow/blob/91c1ab15e6572b2492eb85d3d1439a69eb7f6e9e/modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy#L459-L463

bentsherman: Don't think that is possible, because you need a remote git repo to replicate the issue, i.e. checkout a remote tag that hasn't been pulled locally yet

pditommaso: Fair enough

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
```diff
@@ -629,7 +629,7 @@ class AssetManager {
              */
             catch ( RefNotFoundException e ) {
                 def ref = checkoutRemoteBranch(revision)
-                return "checkout-out at ${ref.getObjectId().name()}"
+                return "checked out at ${ref?.getObjectId()?.name()}"
```
pditommaso, line 632: Can `ref` be null? in this case the message reported would not be nice. Otherwise it can be kept the `ref.getObjectId().name()` version

bentsherman: I think I added the null checks because at some point during testing it was null, which crashed the entire run with an even worse error message

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
```diff
@@ -628,8 +628,11 @@ class AssetManager {
              * Try to checkout it from a remote branch and return
              */
             catch ( RefNotFoundException e ) {
-                def ref = checkoutRemoteBranch(revision)
-                return "checkout-out at ${ref.getObjectId().name()}"
+                final ref = checkoutRemoteBranch(revision)
+                final commitId = ref?.getObjectId()
+                return commitId
+                    ? "checked out at ${commitId.name()}"
+                    : "checked out revision $revision}"
```
bentsherman, line 635:
```suggestion
                    : "checked out revision ${revision}"
```

pditommaso: Oops, it late friday!

PR #4249, opened 2023-08-31
Document operators that maintain order when collecting

nh13:
This includes collect, collectFIle, multiMap, and toList

### DIFF
--- a/docs/operator.md
+++ b/docs/operator.md
@@ -236,7 +236,7 @@ See also: [buffer](#buffer) operator.
 
 *Returns: value channel*
 
-The `collect` operator collects all the items emitted by a channel to a `List` and return the resulting object as a sole emission. For example:
+The `collect` operator collects all the items emitted by a channel to a `List` and return the resulting object as a sole emission, maintaining the order of the items. For example:
 
 ```groovy
 Channel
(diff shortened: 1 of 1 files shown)


nh13:
And if order of the items is not preserved (except when using sort keywords), I think that's not a good thing, and should be changed.

bentsherman:
Had some internal Slack discussions about this PR. See also #2682 about order preservation of operators.

This question usually comes up because someone has a process with multiple inputs that really should be a single tuple input, so they use `multiMap` to split an incoming tuple into multiple channels and `merge` or `join` to recombine the downstream outputs if neccessary.

While this happens to work, it is not a "guaranteed semantic" for any operator to preserve the order of inputs. It only happens to work because operators are single-threaded... if one day they become multi-threaded, it will break your pipeline.

Instead, you should fix the root problem, which is that your process has multiple inputs that should be a single tuple input. See [this note](https://nextflow.io/docs/latest/process.html#multiple-input-channels) in the docs about using multiple inputs.

nh13:
@bentsherman is there a place we could document the alternative solution, and also make it explicit in the docs that order is not maintained (I really think this is important not to be implicit)?

bentsherman:
I think the docs note about multiple process inputs covers this point, also there is a note for [multiMap](https://nextflow.io/docs/latest/operator.html#multimap) since it comes up a lot

PR #4264, opened 2023-09-02
make store public

jchorl:
Went to give `nf-weblog` a go.

It blows up hard here: https://github.com/nextflow-io/nf-weblog/blob/4a6b16832252426851b3b015a2e46f6ba2fafe69/plugins/nf-weblog/src/main/nextflow/trace/WebLogObserver.groovy#L221

Hard enough that I didn't see any stacktrace in the logs - not sure why.

Anyway, I think I chased it down.

I'm not sure if the member should be public, or if there should be a getter, or other. First I tried to implement a getter, but it broke some other methods that rely on `store`.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
@@ -238,7 +238,12 @@ class TraceRecord implements Serializable {
     }
 
 
-    public Map<String,Object> store
+    @PackageScope
+    Map<String,Object> store
+
+    Map<String,Object> getStore() {
+        new LinkedHashMap(store)
+    }
 
     @Memoized
     Set<String> keySet() {
--- a/modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
@@ -238,8 +238,7 @@ class TraceRecord implements Serializable {
     }
 
 
-    @PackageScope
-    Map<String,Object> store
+    public Map<String,Object> store
 
     @Memoized
     Set<String> keySet() {
--- a/modules/nextflow/src/test/groovy/nextflow/trace/TraceRecordTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/trace/TraceRecordTest.groovy
@@ -288,14 +288,14 @@ class TraceRecordTest extends Specification {
         when:
         rec.env = 'aws_key=1234'
         then:
-        rec.store.env == 'aws_key=[secure]'
+        rec.@store.env == 'aws_key=[secure]'
     }
 
     def 'should retrieve safe env' () {
         given:
         def rec = new TraceRecord()
         when:
-        rec.store.env = 'aws_key=1234'
+        rec.@store.env = 'aws_key=1234'
         then:
         rec.env == 'aws_key=[secure]'
     }


modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
```diff
@@ -238,8 +238,7 @@ class TraceRecord implements Serializable {
     }
 
 
-    @PackageScope
-    Map<String,Object> store
+    public Map<String,Object> store
```
pditommaso, line 241: It would be preferable to add a getter method instead of making this public

jchorl:
> It would be preferable to add a getter method instead of making this public

Agreed.

I tried this first (and just tried again). I can't really get the plugin build and nextflow build to play nice together, so it's really hard for me to test/debug locally. Seeing:

```
Cannot invoke "org.pf4j.Plugin.start()" because the return value of "org.pf4j.PluginWrapper.getPlugin()" is null
```

Anyway, I just pushed a change. I suspect this is close, but it's **not** working. I thought the plugin accessing `.store` would auto-call the getter due to groovy shenanigans, but alas this still seems broken (with my latest push). I'm not sure if it's as easy as updating the nf-weblog code to `.getStore()`, as I'm having trouble building/testing.

Here's what I did try to test, pretty much just following the readme from https://github.com/nextflow-io/nf-weblog:
```
docker run -it --rm \
    -v /home/j/third-party/nf-weblog:/home/j/third-party/nf-weblog \
    -w /home/j/third-party/nf-weblog \
    -v /home/j/third-party/nextflow:/home/j/third-party/nextflow \
    -e NXF_TRACE=nextflow.trace \
    -e NXF_ANSI_LOG=false \
    --net=host \
    amazoncorretto:17 \
    bash
bash-4.2# yum install -y make
bash-4.2# pwd
/home/j/third-party/nf-weblog
bash-4.2# echo "includeBuild('../nextflow')" >> settings.gradle
bash-4.2# make compile
bash-4.2# ./launch.sh run nextflow-io/hello -plugins nf-weblog
N E X T F L O W  ~  version 23.08.1-edge
Pulling nextflow-io/hello ...
 downloaded from https://github.com/nextflow-io/hello.git
Launching `https://github.com/nextflow-io/hello` [naughty_moriondo] DSL2 - revision: 1d71f857bb [master]
ERROR ~ Cannot invoke "org.pf4j.Plugin.start()" because the return value of "org.pf4j.PluginWrapper.getPlugin()" is null

```

pditommaso:
Much simpler, clone [nf-weblog](https://github.com/nextflow-io/nf-weblog) as sibling project. Change to the nf-weblog project root and add `includeBuild('../nextflow')` in the `settings.gradle`, then compile and run nextflow using `./launch.sh` from the nf-weblog root.

jchorl:
> Much simpler, clone [nf-weblog](https://github.com/nextflow-io/nf-weblog) as sibling project. Change to the nf-weblog project root and add `includeBuild('../nextflow')` in the `settings.gradle`, then compile and run nextflow using `./launch.sh` from the nf-weblog root.

Thanks - I think this is what I did though? You can see in my log:
```
$ pwd
/home/j/third-party/nf-weblog
```

bentsherman:
I was not able to replicate your original issue. I tried `nextflow run hello -with-weblog` and it ran just fine.

I don't think `store` is the issue because it is marked `@PackageScope`, which means it is accessible to other classes in the same package, which includes the weblog observer.

Can you share your original error message?

jchorl:
> I was not able to replicate your original issue. I tried `nextflow run hello -with-weblog` and it ran just fine.
> 
> I don't think `store` is the issue because it is marked `@PackageScope`, which means it is accessible to other classes in the same package, which includes the weblog observer.
> 
> Can you share your original error message?

When you say it ran just fine, do you mean it succeeded, or it _worked correctly_? For me there were no error messages, but I did not see any e.g. `process_started` messages go through the weblog. Only the nextflow run start and complete messages came through. Again, no error messages.

jchorl:
Another way to check is to enable debug logging and look for this line in the logs: https://github.com/nextflow-io/nf-weblog/blob/4a6b16832252426851b3b015a2e46f6ba2fafe69/plugins/nf-weblog/src/main/nextflow/trace/WebLogObserver.groovy#L256

When I ran it, I only hit that line for the workflow start/end, not for each task start/complete, because it was bombing silently before.

bentsherman:
Sorry I missed your original comment that there was no error message. You were right, it did not log any of the task events, but the exception wasn't being caught because the request happens in a separate thread. With your proposed change, it works again.

Can you redo your last commit and sign it so the DCO will pass? Also need to fix the failing test

I will also push some changes to nf-weblog to try and capture any other silent errors...

bentsherman:
Change this line in the last test in `TraceRecordTest` to fix the failing test:

```diff
-        rec.store.env = 'aws_key=1234'
+        rec.@store.env = 'aws_key=1234'
```

jchorl:
Done - thank you largely for the help. TIL about `@` in groovy.

bentsherman (approved):

pditommaso:
It's not really needed to create a copy of the `store` field. Merged manually via 07582f0b

PR #4266, opened 2023-09-03
Add support for remote debug

pditommaso:
This PR adds the support for remote debugging aka IDE interactive debug to Nextflow script, following the method proposed by Jorge Aguilera in PR #3197. 

The remote debug can be enabled by using the top level option `-remote-debug` both for `nextflow` and `launch.sh` script (for development purposes). 

By default it uses the port `5005`. it can be overridden with by using the env variable `NXF_REMOTE_DEBUG_PORT`. 

Credits https://fediverse.tv/w/3dsuE1tX25orRLmNYtBJrD


pditommaso:
The problem of this approach is that as along as it parses the script file, the [computeClassName](https://github.com/nextflow-io/nextflow/blob/909de8516d8df944d1d857e3124782fb360bae26/modules/nextflow/src/main/groovy/nextflow/script/ScriptParser.groovy#L141-L141) is ignored. 

As a side effect, script containing an identifier as the file name (that's compiled into the class name) will throw an error. For example having a script `hello.nf` with the content 

```
hello = 1 
println hello
```

pditommaso:
Ok, merging this as an experimental feature. It will be needed #4267 for a more soluting support.

PR #4278, opened 2023-09-07
Update the endpoint to be v1

vsmalladi:


bentsherman:
I remember being confused about this point. When I generated the new Java client with the v1.1 openapi spec, it removed the v1 prefix from these requests, even though the v1 is still in the [spec](https://github.com/ga4gh/task-execution-schemas/blob/develop/openapi/task_execution_service.openapi.yaml#L68)

oh well, I will just merge for now

PR #4284, opened 2023-09-08
Raise syntax error if script variable has the same name as script

### LINKED ISSUE #4267: Replace script aliasing with conflicting identifier names
When parsing a script file, Nextflow replaces the script file name with a [unique alias](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/script/ScriptParser.groovy#L138-L152) and parses the script content instead of using the script file itself. 

This is required to prevent the use of a variable having the same name as the script name itself, throws an expected error condition. For example, having a script named `hello.nf` with the following content would result in an execution error. 

```
hello = 1
println hello
```

This mechanism has, however, two downsides: 
1. when a runtime error is reported, the error message may report the script alias name, instead of the real script name, making it difficult to identify the cause of the error.
2. unable to generate compiler debugging information, useful for interactive debugging, that requires the real file path information. 

To avoid those problems, the goal of this ticket is to replace the file name aliasing with the detection of the use of a variable identifier having the same name as the script file, during the script AST parsing, and reporting an actionable error message when the conflict is detected.

bentsherman:
Close #4267

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
@@ -106,23 +106,23 @@ class NextflowDSLImpl implements ASTTransformation {
 
     @Override
     void visit(ASTNode[] astNodes, SourceUnit unit) {
-        createVisitor(unit).visitClass((ClassNode)astNodes[1])
+        final node = (ClassNode)astNodes[1]
+        createVisitor(unit, node.getName()).visitClass(node)
     }
 
     /*
      * create the code visitor
      */
-    protected ClassCodeVisitorSupport createVisitor( SourceUnit unit ) {
-        new DslCodeVisitor(unit)
+    protected ClassCodeVisitorSupport createVisitor( SourceUnit unit, String className ) {
+        new DslCodeVisitor(unit, className)
     }
 
     @CompileStatic
     static class DslCodeVisitor extends ClassCodeVisitorSupport {
 
-
         final private SourceUnit unit
 
-        private String className
+        final private String className
 
         private String currentTaskName
 
@@ -141,15 +141,9 @@ class NextflowDSLImpl implements ASTTransformation {
         protected SourceUnit getSourceUnit() { unit }
 
 
-        DslCodeVisitor(SourceUnit unit) {
+        DslCodeVisitor(SourceUnit unit, String className) {
             this.unit = unit
-        }
-
-        @Override
-        void visitClass(ClassNode node) {
-            if( !className )
-                className = node.name
-            super.visitClass(node)
+            this.className = className
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


pditommaso:
Excellent. Made minor improvements

modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
```diff
@@ -263,7 +257,7 @@ class NextflowDSLImpl implements ASTTransformation {
                 stm.setExpression(loadCall)
             }
             if( isInvalidVariableAssignment(stm) ) {
-                syntaxError(stm, 'Cannot declare a script variable with the same name as the script')
+                syntaxError(stm, 'Cannot declare a variable identifier with the same name as the script file')
```
bentsherman, line 260: "Script variable" would be more precise because local variables are still allowed. Note from the tests that you need only add a `def` to make it work.

Actually we should probably add that suggestion to the error message...

bentsherman:
Damn... you add one little thing and the whole world explodes

bentsherman:
The problem is that `MockScriptRunner` saves the script to a temporary file, but the underlying file system `JimFs` doesn't support the `toFile()` method that is called when parsing the file instead of the text.

Would it be fine to just change the mock script runner to not create the temp file?

bentsherman:
Ah, but the script runner itself expects a script file...

modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
pditommaso: yeah, that's true. just because don't like the repetition "script variable" and "script file"

pditommaso:
yep, it should be used `.toUri`, I'm on that

pditommaso:
Not sure anymore we should do this. The plan was to get rid of the hashed script name. However I've realised that using the real file name the stack trace would be something like 

```
groovy.lang.MissingPropertyException: No such property: puts for class: nextflow.script.WorkflowBinding
	at groovy.lang.Binding.getVariable(Binding.java:61)
	at nextflow.script.WorkflowBinding.getVariable(WorkflowBinding.groovy:140)
	at groovy.lang.Binding.getProperty(Binding.java:116)
	at nextflow.script.WorkflowBinding.getProperty(WorkflowBinding.groovy:129)
	at org.codehaus.groovy.runtime.InvokerHelper.getProperty(InvokerHelper.java:190)
	at groovy.lang.Closure.getPropertyTryThese(Closure.java:320)
	at groovy.lang.Closure.getPropertyDelegateFirst(Closure.java:310)
	at groovy.lang.Closure.getProperty(Closure.java:296)
	at org.codehaus.groovy.runtime.callsite.PogoGetPropertySite.getProperty(PogoGetPropertySite.java:49)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callGroovyObjectGetProperty(AbstractCallSite.java:341)
	at t$_runScript_closure1$_closure3.doCall(main.nf:5)
```

Considering that almost all modules are named `main.nf`, it will result in a less precise error message instead of a more detailed one.

bentsherman:
Is it not possible to make the error trace use the script path?

pditommaso:
No idea

pditommaso:
Closing for now

pditommaso:
Replaced by this https://github.com/nextflow-io/nextflow/commit/c056a74ec08542f5d232de74f00ec844ae29e4a6

PR #4285, opened 2023-09-08
Add task entry prefix to cloud cache

bentsherman:
Before we start pushing the cloud cache more... I am concerned that it will not perform optimally against object storage at scale. This PR is meant to address that concern.

My understanding is that when you request many S3 objects, each prefix is processed in parallel. So if you have many files then it is better to spread them across many prefixes, as we do for the work directory.

But a workflow run with 30,000 tasks will have an S3 directory with 30,000 objects... I suspect that will be hard to manage and hurt performance a little bit.

It may not be a huge issue for resuming, since usually you would resume a few tasks at a time rather than all at once. But if you ran e.g. 1,000 samples in parallel, then the resume would be fetching 1,000 objects at a time, from the same prefix.

### DIFF
--- a/plugins/nf-cloudcache/src/main/nextflow/cache/CloudCacheStore.groovy
+++ b/plugins/nf-cloudcache/src/main/nextflow/cache/CloudCacheStore.groovy
@@ -162,6 +162,7 @@ class CloudCacheStore implements CacheStore {
     }
 
     private Path getCachePath(HashCode key) {
-        dataPath.resolve(key.toString())
+        final keyStr = key.toString()
+        dataPath.resolve(keyStr.substring(0, 2)).resolve(keyStr.substring(2))
     }
 }


bentsherman:
We should try to decide on this before the 23.10 release, since it is a breaking change to the cloud cache.

pditommaso:
Not sure, object storage systems are big key-value databases. The directory structure is simulated, interpreting the `/` in a special manner by the client, but it is stored in the same key obtained concatenating all path components.

bentsherman:
It looks like Amazon addressed this issue back in 2018 according to [this article](https://aws.amazon.com/about-aws/whats-new/2018/07/amazon-s3-announces-increased-request-rate-performance/), so we shouldn't need to worry about maintaining separate prefixes.

PR #4292, opened 2023-09-11
Use shell directive with trace command in wrapper script

### LINKED ISSUE #4209: Using tracing invalidates use of `shell` directive
## Bug report 

When using the `-with-trace` argument, Nextflow does not adhere to the shell directive when generating the .command.run file.

I would expect that setting the `-with-trace` configuration should not affect the behaviour of the `shell` directive.

Given the simple `main.nf` that uses the `shell` directive for a non-standard bash location:

```groovy
process Test {
    shell '/usr/local/bin/bash'
    script: "echo hi"
}

workflow {
    Test()
}
```

If I run
```shell
nextflow run . -with-docker 'docker.io/bash:5.2'
```

Nextflow generates a `.command.run` that respects the shell directive in `.command.run` (newlines added for emphasis):

```shell
nxf_launch() {
    docker run -i --cpu-shares 1024 -v $TASKDIR:$TASKDIR -w "$PWD" --name $NXF_BOXID docker.io/bash:5.2 \
        /usr/local/bin/bash $TASKDIR/.command.sh
}
```

If I enable tracing enabled,
```shell
nextflow run . -with-docker 'docker.io/bash:5.2' -with-trace trace.txt
```

Nextflow ignores my request to change the interpreter. Again, in `.command.run`:

```shell
nxf_launch() {
    docker run -i --cpu-shares 1024 -e "NXF_DEBUG=${NXF_DEBUG:=0}" -v $TASKDIR:$TASKDIR -w "$PWD" --name $NXF_BOXID docker.io/bash:5.2 \
        /bin/bash $TASKDIR/.command.run nxf_trace
}
```

### Environment 

* Nextflow version: `23.04.2`
* Java version: openjdk version "17.0.5" 2022-10-18
* Operating system: macOS
* Bash version: zsh 5.9

bentsherman:
Close #4209


rollf:
@bentsherman Could this PR be updated and merged? (Asking for a friend.)

pditommaso:
This is a -1 for me because it's changing the flag used to run the trace script adding `-eu`. While this seems innocuous and even better, it can introduce subtle different behaviour hard to predict and above all to tests.

pditommaso:
This [392073b](https://github.com/nextflow-io/nextflow/pull/4292/commits/392073bc4415590e314718cf1cf01822ad44c255) should achieve the same without changing the default behaviour

bentsherman:
Makes sense to me

sstrong99:
what is a good way to track when this change is released?

bentsherman:
Edge release should be coming out in the next week or two. You can probably subscribe to release notifications? You can also check the merge commit and it will tell you which releases have the commit.

pditommaso:
Done! https://github.com/nextflow-io/nextflow/releases/tag/v24.12.0-edge

PR #4295, opened 2023-09-11
Retry TimeoutException in azure file system

bentsherman:
Requested from Freshdesk ticket

- Should we also check the `cause` like we do for `IOException`?
- Any simple way to test this?


pditommaso:
> Should we also check the cause like we do for IOException

it would be better 

> Any simple way to test this?

No idea how to replicate

PR #4299, opened 2023-09-12
Fix issues with workflow complete handler

### LINKED ISSUE #3129: Unpredictable behaviour for workflow.onComplete handler
## Bug report 

In certain situations, the workflow.onComplete handler will not run. When using `error`. For example, in the following script the line "OnComplete: Ending Workflow" should appear in the logs according to the documentation: https://www.nextflow.io/docs/latest/metadata.html#completion-handler. You can see the `onError` context has been entered correctly.

The workflow.onComplete context should be entered and scripts within executed.

```nextflow
#! /usr/bin/env nextflow
nextflow.enable.dsl=2

workflow {
    CREATE_ERROR ()
}

workflow CREATE_ERROR {
    error "Error: Oops!"
}

workflow.onComplete {
    log.info("""
    OnComplete: Ending Workflow
    """)
}

workflow.onError {
    log.info("""
    OnError: An error occured.
    """)
}

```

### Program output 

```
Launching `./main.nf` [nauseous_ardinghelli] - revision: 8ba6ace425

    OnError: An error occured.
    
Error: Oops!

 -- Check script 'main.nf' at line: 11 or see '.nextflow.log' file for more details
(nextflow) 
```
[nextflow.log](https://github.com/nextflow-io/nextflow/files/9374002/nextflow.log)

### Environment 

* Nextflow version: 21.10.6
* Java version:
```
openjdk version "11.0.9.1" 2020-11-04 LTS
OpenJDK Runtime Environment Zulu11.43+55-CA (build 11.0.9.1+1-LTS)
OpenJDK 64-Bit Server VM Zulu11.43+55-CA (build 11.0.9.1+1-LTS, mixed mode)
```
* Operating system: [macOS, Ubuntu 20.04]
* Bash version: `zsh 5.8 (x86_64-apple-darwin20.0)`, `GNU bash, version 4.4.20(1)-release (x86_64-pc-linux-gnu)`

### Additional context

This seems to impact NF-Core template pipelines which exit prematurely before sending the email in the `onComplete` handler.

bentsherman:
Close #3129 

- Call workflow complete handler on all workflow errors, including session `abort()` (as well as other shutdown callbacks and observers)
- Add `log` to complete/error handler context so that they can use the logger when defined in a config file

For testing (1), see the reproducible example in the linked issue.

For testing (2), add the following to Nextflow config:

```groovy
workflow.onComplete = {
    log.info "~ config handler was also here ~"
}

workflow.onError = {
    log.error "~ config handler was also here ~"
}
```

I'm not 100% on either of these changes. Not sure if adding the shutdown callbacks to session `abort()` has other unintended side effects. And the delegate for the workflow handlers should probably be done more carefully to make sure that the handler can still access global variables in the pipeline script.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -764,6 +764,7 @@ class Session implements ISession {
                 log.debug(status)
             // force termination
             notifyError(null)
+            shutdown0()
             ansiLogObserver?.forceTermination()
             executorFactory?.signalExecutors()
             processesBarrier.forceTermination()
--- a/modules/nextflow/src/main/groovy/nextflow/script/WorkflowMetadata.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/WorkflowMetadata.groovy
@@ -385,6 +385,7 @@ class WorkflowMetadata {
 
         onCompleteActions.each { Closure action ->
             try {
+                action.delegate = [log: log]
                 action.call()
             }
             catch (Exception e) {
@@ -402,6 +403,7 @@ class WorkflowMetadata {
         setErrorAttributes()
         onErrorActions.each { Closure action ->
             try {
+                action.delegate = [log: log]
                 action.call(trace)
             }
             catch (Exception e) {


bentsherman:
But if we don't make these changes, we should document them as limitations instead.

pditommaso:
No chance to add some unit tests?

bentsherman:
When I'm not sure about a PR, I like to get your overall opinion before I write unit tests

bentsherman:
Also, I'm not sure that unit tests would help here. Probably better to add an integration test based on the reproducible examples.

bentsherman:
@pditommaso I have trimmed this PR to focus on calling the onComplete on all error conditions, including the `error` function. Also added an e2e test

Notes:

- This PR calls all shutdown callbacks on abort, which includes the onComplete and the [task monitor cleanup](https://github.com/nextflow-io/nextflow/blob/bfd5174ce80895030835c815449886ee7aad6c80/modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy#L317-L318). This seems like a good thing? In fact I thought it was already doing this but I don't see any logic for it

- The onComplete is also allowed to call the `error` function, which could lead to the onComplete being called multiple times. It shouldn't cause an infinite recursion thanks to the existing error-handling logic in Session::abort() and other places. I think it's still fine to call `error()` in the onComplete, and as long as you do it in a sensible way, you will get sensible error messages

jorgee (approved):

PR #4300, opened 2023-09-13
Raise error if cpus value is negative

bentsherman:
Related to customer ticket: https://support.seqera.io/helpdesk/tickets/3774

An alternative would be to simply log a warning and return 1 cpu if the cpus value is negative.


bentsherman:
Related to #4917 -- I think I might refactor this PR into a more general solution that adds stricter checks to `ProcessConfig` for all process directives -- types, min/max values where appropriate. The main challenge is to make sure that the process DSL and process configuration are routed through the same methods

pditommaso:
Closing in favour of af2e4ef7

PR #4301, opened 2023-09-13
Include FileHolder stage name in task cache

### LINKED ISSUE #4106: Cache is reused with -resume even if 'stageAs' is changed/added
## Bug report 

When changing the value of stageAs in a module, it did not register as a change relative to the cache, so I had to manually delete the task work directory. For the module I used, changing stageAs was needed to make it work. 

If stageAs changes, the cache hash will change and the process will be rerun. 

Run any pipeline with `-resume`. Add stageAs to any process to change the file names for staged files. Rerun pipeline and note how the cache is still used for the changed process and the new code is not run.

### Environment 

* Nextflow version: version 23.04.1 build 5866
* Java version: openjdk 11.0.15-internal 2022-04-19
* Operating system: Pop!_OS 22.04 LTS

bentsherman:
Close #4106

### DIFF
--- a/modules/nf-commons/src/main/nextflow/util/CacheHelper.java
+++ b/modules/nf-commons/src/main/nextflow/util/CacheHelper.java
@@ -191,8 +191,11 @@ public class CacheHelper {
             return hasher;
         }
 
-        if( value instanceof FileHolder )
-            return CacheHelper.hasher(hasher, ((FileHolder) value).getSourceObj(), mode );
+        if( value instanceof FileHolder ) {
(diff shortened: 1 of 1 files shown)


pditommaso (commented):
It could make sense. My concern is that this will break the cache for existing runs. We may want to park this until there's a more concrete need to push a series of breaking changes all together

bentsherman:
Sounds good. You can check the `cache-breaking` label linked here to see those. Currently there are four.

If we introduce the CID store as an alternative cache, that would also be a good opportunity to re-design the task hash

bentsherman:
Closing because I don't think this is needed

PR #4313, opened 2023-09-15
Map type for process inputs/outputs

### LINKED ISSUE #2127: Allow map output from a process
## New feature

I would like to be able to produce a single channel with multiple named values (ie a `Map`) from my processes, ie `my_proccess.out.view()` should return:
```
[a:abc, b:123, c:false]
[a:abc, b:123, c:false]
[a:def, b:456, c:true]
```

Currently we have a `tuple` qualifier, which creates an output channel which is a tuple of other types. However this tuple is unlabelled, so users have to extract values from this channel by position, which results in confusing code. 

It is also possible to produce multiple output channels, each of which has its own name. However, these channels can't easily be combined into a single channel containing maps or tuples, because the `merge` operator has been deprecated, and in general joining channels by position is discouraged.

## Usage scenario 

This would be useful when a user is working with mostly map data in their channels, likely because they want each field to be labelled instead of unlabelled as in a tuple.

## Suggest implementation 

I would envisage a new `map` qualifier, which is used like this:
```groovy
process hmmer_search {
    container "quay.io/biocontainers/hmmer:3.3.2--h1b792b2_1"
    input:
      path profile
      path database
    output:
        map [table: path('table.txt'), human_readable: path('match.txt')]
    script:
      """
      hmmsearch -o match.txt --tblout table.txt ${profile} ${database} 
      """
}
```

bentsherman:
Close #2127

### DIFF
--- a/tests/process-map-input-output.nf
+++ b/tests/process-map-input-output.nf
@@ -1,0 +1,34 @@
+#!/usr/bin/env nextflow
+
+
+process foo {
+  input:
+    val id
+  output:
+    map( [id: val(id), first: path('first.txt'), second: path('second.txt')], emit: 'records' )
+  script:
+    """
+    echo 'first' > first.txt
+    echo 'second' > second.txt 
+    """
+}
+
+process bar {
+  input:
+    map( [id: val(id), first: path('first.txt'), second: path('second.txt')] )
+  output:
+    stdout
+  script:
+    """
+    echo ${id}
+    cat first.txt
+    cat second.txt
+    """
+}
+
+workflow {
+  Channel.of(1, 2, 3)
+    | foo
+    | bar
+    | view
+}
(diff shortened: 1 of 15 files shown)


tests/process-map-input-output.nf
```diff
@@ -0,0 +1,34 @@
...
+
+process bar {
+  input:
+    map( [id: val(id), first: path('first.txt'), second: path('second.txt')] )
+  output:
+    stdout
+  script:
+    """
+    echo ${id}
+    cat first.txt
```
mribeirodantas, line 24: Would it be possible to do `cat ${first}`?

mribeirodantas: Oh, ok, it's clear now. I wasn't sure about what `id` was `${id}` referring to, but it's not the key 😬

pditommaso:
This is an interesting capability but requires more in-depth research on the implication of this change on the use of map objects with existing operators. The side effects are not trivial. Above this is out of the scope of DSL3, which has higher priority. Closing for now.

bentsherman:
records are better anyway

I think we will be able to support maps automatically with the annotation API I'm working on

PR #4336, opened 2023-09-22
Document cloud cache

llewellyn-sl:


bentsherman:
Putting it with the `cache` directive was clever, but there is a separation between how the cache is defined (i.e. the task hash components) versus how the cache is stored (i.e. local file vs remote storage, etc), and the process directive is really only concerned with the former.

For now I think the best option is to document the cloud cache on the S3 page. But the cache directive is probably a good place to elaborate on how the cache is defined -- what are the components of a task hash? what information is stored in the cache entry?

But in the future I think it would be better to dedicate an entire page to the cache and resume mechanism, as it is a major feature of Nextflow and currently the various aspects of it are scattered throughout the docs. That way we could cover both the definition and available backends in one place. And maybe also some troubleshooting tips that so far have only been covered in blog posts.

pditommaso:
The `cache` directive is misleading because it's related to the task compute cache, instead `NXF_CLOUDCACHE_PATH` is about the execution metadata. 

Likey we'll need a new section to better describe how the nextflow cache and resume mechanism. Until then I think it's enough the entry in the config section documenting the  `NXF_CLOUDCACHE_PATH` variable

bentsherman:
Closing in favor of #4371 

@llewellyn-sl don't worry about giving a deep review if you're busy, but let me know if the cloud cache section is sufficient

PR #4337, opened 2023-09-22
Use Mermaid diagram in HTML DAG

### LINKED ISSUE #4316: Use Mermaid output in HTML DAG
Currently, the DAG output implementation for HTML is stand-alone, and needs to be maintained separately. The Mermaid DAG has been improved quite a bit recently (https://github.com/nextflow-io/nextflow/pull/4070). Rather than trying to maintain the two codebases separately, we could just reuse the Mermaid output for the HTML file.

Basically, if HTML is requested - create the Mermaid DAG output file and embed it within a static HTML template. The template should include the relevant Mermaid javascript dependencies and script blocks, as described in the [Mermaid docs](https://mermaid.js.org/intro/n00b-gettingStarted.html#_3-calling-the-javascript-api).

This should give us nicer functionality and less maintenance burden.

Note that the HTML DAG config scopes should probably still stay in place, but would then just forward on to the Mermaid DAG generation code.

👉🏻  Related to https://github.com/nextflow-io/nextflow/issues/4315

bentsherman:
Close #4316 

- Make HTML the default DAG format
- Change HTML format to embed Mermaid diagram instead of Cytoscape

The stretch goals I mentioned about hyperlinks and interactivity... will be done in a separate PR😄 

Testing is easy:

```bash
./launch.sh run <pipeline> -preview -with-dag dag.html
```

Then open `dag.html` in your browser!


bentsherman:
For large pipelines like nf-core/rnaseq, it is better for the diagram to be left-to-right instead of the default top-to-bottom. It is easy to edit the HTML file to do this, just change `flowchart TD` to `flowchart LR`, but it might be good to provide a config option to control this.

pditommaso:
Agree, let's add `direction` option to control it. 

Also there are a bunch static type errors reported by Intellij (tho the compiler seems ok), [here](https://github.com/nextflow-io/nextflow/blob/371b70972a94b43a1e6ac6234530132bcc9f6854/modules/nextflow/src/main/groovy/nextflow/dag/MermaidRenderer.groovy#L69-L69), [here](https://github.com/nextflow-io/nextflow/blob/371b70972a94b43a1e6ac6234530132bcc9f6854/modules/nextflow/src/main/groovy/nextflow/dag/MermaidRenderer.groovy#L76-L76) and here https://github.com/nextflow-io/nextflow/blob/371b70972a94b43a1e6ac6234530132bcc9f6854/modules/nextflow/src/main/groovy/nextflow/dag/MermaidRenderer.groovy#L291-L292 and [here](https://github.com/nextflow-io/nextflow/blob/371b70972a94b43a1e6ac6234530132bcc9f6854/modules/nextflow/src/main/groovy/nextflow/dag/MermaidRenderer.groovy#L346-L346)

bentsherman:
Okay I see these errors in IntelliJ, I guess the IDE is stricter than Groovy on some things. I will fix them.

pditommaso (approved):

mribeirodantas:
🥳

Issue #4330, opened 2023-09-21
Plugins offline: `Nextflow self-contained distribution allows only core plugins`


ewels:
Talking to @Alexander-Stuckey about running offline with plugins, I pointed to [these docs](https://github.com/nextflow-io/nextflow/blob/master/docs/plugins.md#offline-usage) for fetching plugins for offline use, but it seems that it doesn't work if you use the `-all` distribution of Nextflow:

```
N E X T F L O W  ~  version 23.09.1-edge
Launching `/pgen_int_work/BRS/rif/helix_replacement_project/sarek/resources/nf-core-sarek_3.3.1/3_3_1/main.nf` [distracted_meninsky] DSL2 - revision: 19395081df
WARN: Nextflow self-contained distribution allows only core plugins -- User config plugins will be ignored: nf-validation@0.3.2
```

This is currently the [recommended method](https://nf-co.re/docs/usage/offline#nextflow) from nf-core, so it's causing some confusion.

We're starting to see these kinds of problems a lot now, since @nf-core pipelines are adopting the [nf-validation](https://github.com/nextflow-io/nf-validation) plugin. I think that this is probably some of the first very widespread adoption of a non-core Nextflow plugin.

ewels:
Related to https://github.com/nextflow-io/nextflow/issues/4329

bentsherman:
I think #4340 will make it possible to get around this problem by using the regular Nextflow distribution and downloading the desired set of core plugins. Otherwise we will need to change the self-contained distribution to support non-core plugins, and I don't know how difficult that would be.

ewels:
Ok, yeah sounds like we probably need to drop the `-all` distribution at some point. I know it's been spoken about before hand.

I guess tied to this is that we could do with a user friendly way to distribute Nextflow in offline systems to replace that dist. Running Nextflow on another system and then transferring a bunch of directories is quite indirect and unintuitive.

Could we have some kind of "downloader" web pages, that generates an `-all` dist on the fly with the selected plugins? Or some other way that we can make it easy for people to easily get a Nextflow install that works offline?

Related to https://github.com/nextflow-io/nextflow/issues/2028 and https://github.com/nextflow-io/nextflow/issues/2951

bentsherman:
Could we extend `nf-core download` for this purpose, or is it only used for container images?

You could write a bash script to install everything you need for offline usage:
1. install nextflow like normal
[...]

Issue #4036, opened 2023-06-13
Pinning plugin major version


ewels:
## Problem

When adding a plugin to a pipeline we have to pin a specific version. This is good for reproducibility, but bad for bug fixing. We're about to add `nf-validation` into the nf-core template, but we know very well that we will hit lots of bugs and need to do patch releases. This will require a lot of manual work updating the template and many pipelines on a regular basis.

### Language
A classic way is to give syntax to say "this version or above" (eg. Nextlow itself handles this with `>=` and `!>=`). This would presumably require a language change in Nextflow to allow inclusion of pipelines like this.

### Major version pins

Possibly a simpler way is to use major version pinning. This is done with GitHub Actions (see official [GitHub docs](https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-tags-for-release-management)): when a new release is made, a new specific git tag is added (`v1.4.2`) but also the tag for the major version is bumped to point to the new release (`v1`). That way, users can pin either the specific release, or they can pin `v1` and get updated patch releases as they are available, automatically.

A similar approach should be viable for Nextflow plugins I think - add a `v1` major version to the JSON file and update the download link and sha every time a new release is made. Then include the plugin in the pipeline with the major version tag.

Any thoughts @pditommaso @bentsherman? Tagging @mirpedrol @mashehu for visibility.

pditommaso:
It sounds reasonable, tho I'm not 100% sure about the implementation. Nextflow relies on a library to handle plugins. 

It may be possible to implement this logic rewriting this logic 

https://github.com/nextflow-io/nextflow/blob/master/modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy#L402-L403

pditommaso:
I don't believe so. Have you tried it?

ewels:
I got excited for a moment there that `>=` operator might already work (so we could do `plugins { id 'nf-validation@>=1.0.0' }`). But @mirpedrol just tried it and no luck:

* with 0.2.1 installs 0.2.1
* without a tag installs 0.2.1
* with @>=0 there’s an error, it’s trying to download version >=0
* and the same happens with >0 or >0.0.0

mirpedrol:
I have tried the suggested approach https://github.com/nextflow-io/nextflow/issues/4036#issuecomment-1598271009 and it seems it would need some implementation if we want this to work.
If we define the major version with only one number, it fails because the library expects three digits.
```json
{
  "version": "0",
[...]

PR #4340, opened 2023-09-25
Pin only major/minor plugin version with `+` suffix

### LINKED ISSUE #4329: Plugins: Use any available cached version if `NXF_OFFLINE` is true
Currently, running offline with Nextflow plugins [is possible](https://github.com/nextflow-io/nextflow/blob/22638d460ae61d308103bf7291f22c9586d10f87/docs/plugins.md#offline-usage), but it is a little difficult. I think that with some fairly minor changes we could streamline the process a little.

'cc @remiolsen and @MatthiasZepper to correct / add more detail if needed.

## Usage scenario

Currently, working offline with plugins involves the following steps:

* Fetch required plugins on an online system
* Copy the `$NXF_HOME` dir / Nextflow plugins cache to the offline system
* Use a custom config file to overwrite the plugin include in the pipeline to specifically pin the version of the plugin that was downloaded
    * This step is difficult, especially with more complex deployments such as running on Tower / with the Tower CLI

Pinning the plugin version within the workflow itself is undesirable, as that means that any bugfixes in the plugin do not make it out to pipelines for a long time. This is one the main benefits of moving code into a plugin.

My suggestion is that if `NXF_OFFLINE` is set for the workflow run (and potentially if the request to the plugins JSON times out / fails), and the plugin version in the workflow config is _unpinned_, then Nextflow should check the local plugins cache and use the latest available version.

If the plugin version is pinned and _not_ available in the cache, then fine - the workflow must fail. No question there.

bentsherman:
Close #4329 

This PR changes the plugin system to not download plugins in offline mode (`NXF_OFFLINE`). Instead, Nextflow will search the plugins directory for the latest plugin version and use that one.

This change makes it possible to run a pipeline in offline mode without pinning the version of each plugin, which allows the pipeline to automatically receive plugin updates. The assumption is that users will download all the plugins they need beforehand and transfer them to the offline system, using these plugins as a local cache.

I tested this PR by installing nf-validation 0.3.0 (`nextflow plugin install nf-validation@0.3.0`) and running a script with the following config:
```groovy
plugins {
  id 'nf-validation'
}
```

Normally, Nextflow would automatically download the newest plugin version (currently 0.3.2), but now if you run Nextflow with `NXF_OFFLINE=true`, it will use the already downloaded version 0.3.0 and not download anything new. I even disabled my internet connection just to be sure 😉 

One limitation however is that the offline version checker doesn't verify that a plugin version is compatible with the current Nextflow version. We can get the version range from the plugin manifest file, but maybe there is some code somewhere that already does this?

### DIFF
--- a/modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
+++ b/modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
@@ -205,26 +205,23 @@ class PluginUpdater extends UpdateManager {
             return pluginPath
         }
 
-        // 1. determine the version
-        if( !version )
-            version = getLastPluginRelease(id)
         log.info "Downloading plugin ${id}@${version}"
 
-        // 2. Download to temporary location
-        Path downloaded = safeDownloadPlugin(id, version);
+        // 1. download to temporary location
+        Path downloaded = safeDownloadPlugin(id, version)
 
-        // 3. unzip the content and delete downloaded file
+        // 2. unzip the content and delete downloaded file
         Path dir = FileUtils.expandIfZip(downloaded)
         FileHelper.deletePath(downloaded)
 
-        // 4. move the final destination the plugin directory
+        // 3. move the final destination the plugin directory
         assert pluginPath.getFileName() == dir.getFileName()
 
         try {
             safeMove(dir, pluginPath)
         }
         catch (IOException e) {
-            throw new PluginRuntimeException(e, "Failed to write file '$pluginPath' to plugins folder");
+            throw new PluginRuntimeException(e, "Failed to write file '$pluginPath' to plugins folder")
         }
 
         return pluginPath
@@ -316,14 +313,25 @@ class PluginUpdater extends UpdateManager {
     private boolean load0(String id, String version) {
         assert id, "Missing plugin Id"
 
-        if( version == null )
-            version = getLastPluginRelease(id)?.version
-        if( !version )
-            throw new IllegalStateException("Cannot find latest version of $id plugin")
+        def pluginPath = null
+        if( SysEnv.get('NXF_OFFLINE')=='true' ) {
+            if( version == null )
+                version = getLastPluginReleaseOffline(id)
+            if( !version )
+                throw new IllegalStateException("Cannot find local version of $id plugin (offline mode)")
 
-        def pluginPath = pluginsStore.resolve("$id-$version")
-        if( !FilesEx.exists(pluginPath) ) {
-            pluginPath = safeDownload(id, version)
+            log.debug "Update disabled in offline mode -- using local version $version of $id plugin"
+            pluginPath = pluginsStore.resolve("$id-$version")
+        }
+        else {
+            if( version == null )
+                version = getLastPluginRelease(id)?.version
+            if( !version )
+                throw new IllegalStateException("Cannot find latest version of $id plugin")
+
+            pluginPath = pluginsStore.resolve("$id-$version")
+            if( !FilesEx.exists(pluginPath) )
+                pluginPath = safeDownload(id, version)
         }
 
         // verify the plugin install path contains the expected manifest path
@@ -363,8 +371,19 @@ class PluginUpdater extends UpdateManager {
         // resolve the plugins
         pluginManager.resolvePlugins()
         // finally start it
-        PluginState state = pluginManager.startPlugin(id);
+        PluginState state = pluginManager.startPlugin(id)
         return PluginState.STARTED == state
+    }
+
+    private String getLastPluginReleaseOffline(String id) {
+        final specs = FilesEx.list(pluginsStore)
+            .collect( dir -> PluginSpec.parseDirectory(dir) )
+            .findAll( spec -> spec.id == id )
+
+        if( !specs )
+            return null
+
+        specs.max( (a,b) -> Version.valueOf(b.version) <=> Version.valueOf(a.version) )?.version
     }
 
     /**
@@ -381,13 +400,13 @@ class PluginUpdater extends UpdateManager {
             throw new PluginRuntimeException("Plugin $id cannot be updated since it is not installed")
         }
 
-        PluginInfo pluginInfo = getPluginsMap().get(id);
+        PluginInfo pluginInfo = getPluginsMap().get(id)
         if (pluginInfo == null) {
             throw new PluginRuntimeException("Plugin $id does not exist in any repository")
         }
 
         if (!pluginManager.deletePlugin(id)) {
-            return false;
+            return false
         }
 
         load0(id, version)
--- a/modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
+++ b/modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
@@ -378,7 +378,7 @@ class PluginUpdater extends UpdateManager {
 
     private String getLastPluginReleaseOffline(String id) {
         final specs = FilesEx.list(pluginsStore)
-            .collect( dir -> PluginSpec.parseDirectory(dir) )
+            .collect( dir -> PluginSpec.parseDirName(dir) )
             .findAll( spec -> spec.id == id )
 
         if( !specs )
--- a/modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
+++ b/modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
@@ -313,25 +313,26 @@ class PluginUpdater extends UpdateManager {
     private boolean load0(String id, String version) {
         assert id, "Missing plugin Id"
 
-        def pluginPath = null
-        if( SysEnv.get('NXF_OFFLINE')=='true' ) {
-            if( version == null )
-                version = getLastPluginReleaseOffline(id)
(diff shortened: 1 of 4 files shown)


ewels (approved):
Super nice, thanks @bentsherman!

This problem is currently causing a _lot_ of pain within nf-core at the moment, since we recently added the first non-core plugin to all pipelines (nf-validation). So suddenly, everyone running offline is hitting a wall.

Getting this PR merged + released ASAP would be a great help.  So I'd mark this as one of my highest priority Nextflow PRs currently @pditommaso 🙏🏻

pditommaso:
Need to review it. AFAIK there's already in place a logic to pick the current installed plugin in place

bentsherman:
@ewels in the meantime, can they not work around the issue by specifying the list of plugins with versions in their config or CLI? That should work without needing to fork the pipeline code.

jfy133:
How does `NXF_OFFLINE` get determined exactly? Does this need to be physically set or will Nextflow assume that if it can't contact a server it should set that env for you?

Overly specific context: I ask as one of my bugbears since the nf-validation plugin being added to nf-core pipelines, is when developing on German trains where the internet connection is _awful_ is I will start testing a pipeline, it works a couple of times, then the internet goes and then the next test run fails because it can't download the pipeline because the internet connection has dropped. That keeps happening until the connection comes back.

It would be nice for the default (IMO) is to always look for a plugin in cache _unless_ a new version is specified, not just when NXF_OFFLINE is specified (in my context above, I don't want to keep having to sporadically having to set/deactivate the parameter...)

pditommaso:
> It would be nice for the default (IMO) is to always look for a plugin in cache unless a new version is specified

this is what exactly happens for core plugins. For non-core plugins, when no version is specified, nextflow checks for the latest version remotely

jfy133:
> > It would be nice for the default (IMO) is to always look for a plugin in cache unless a new version is specified
> 
> this is what exactly happens for core plugins. For non-core plugins, when no version is specified, nextflow checks for the latest version remotely

Is there a particular reason why the same behaviour could not also apply for non-core (assuming the plugin has already been cached)?

pditommaso:
Actually I want to test this scenario

pditommaso:
I confirm a plugin has a pinned version e.g. `nf-validation@0.3.2` and it's configure locally, nextflow does not try to download it

ewels:
> @ewels in the meantime, can they not work around the issue by specifying the list of plugins with versions in their config or CLI? That should work without needing to fork the pipeline code.

Yes that's what we're doing (actually I forgot that you could do it on the CLI, that's nicer). But I've walked people through this about 5 times in the past week and it's getting tiring 😅  We're also starting to get [XY questions](https://xyproblem.info/) related to stripping the nf-validation plugin out of pipeline code, which is bad.

> I confirm a plugin has a pinned version e.g. nf-validation@0.3.2 and it's configure locally, nextflow does not try to download it

Yup, we know - https://github.com/nextflow-io/nextflow/issues/4329 is specifically about making this work _without_ pinning.

* We don't want to pin the plugin version in nf-core pipelines as then we can't ship bugfixes to the plugin.
* We also don't want to introduce the overhead of making every offline user set up their own config to pin the plugin, just so that the pipeline runs.

The folks at @SciLifeLab @NationalGenomicsInfrastructure are trying to launch from Tower in an offline environment, and they were struggling to get the custom config with the pinned plugin versions to resolve correctly at all, so this is a production-blocking problem for them. Can go into more detail on that in a separate issue if you're interested, I don't want to derail the thread here about Tower config resolution.

I'm hoping that this PR means that non-core plugins will "just work" for people running offline.

modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
```diff
@@ -316,14 +313,25 @@ class PluginUpdater extends UpdateManager {
...
+        }
+        else {
+            if( version == null )
+                version = getLastPluginRelease(id)?.version
+            if( !version )
+                throw new IllegalStateException("Cannot find latest version of $id plugin")
+
+            pluginPath = pluginsStore.resolve("$id-$version")
+            if( !FilesEx.exists(pluginPath) )
+                pluginPath = safeDownload(id, version)
```
pditommaso, line 334: This change is too convoluted. Ideally it should something like 

```
        if( !version )
            version = offline ? getLastPluginReleaseOffline(id) : getLastPluginRelease(id)
```

bentsherman: I would like to do that too, but the error messages are different and the online path also calls `safeDownload()`. You think it would be better to just check `offline` multiple times?

pditommaso: but once the version is resolved (not null) the `safeDownload` just use the local installation if it can find it

bentsherman: true 👍

ewels:
This PR is getting very urgent. There's a long thread over in the @nf-core slack about pinning exact plugin versions in all pipelines, because so many people are hitting difficulties fetching plugins in offline systems / systems with patchy internet.

My current thinking is that we should extend the logic for plugins to work in the same manner as Nextflow itself: if a plugin is not pinned, use the latest _local_ version available if there is one.

Currently anyone developing a pipeline on a train can suddenly not test it when they go into a tunnel, even if they've run it 20 times in the past ten minutes. Nextflow _requires_ reaching out to the internet on every run if a plugin is not pinned. We don't want to pin plugins in pipeline code, because pipeline release cycles can be very slow.

What ideally what we need ASAP is:

* This PR to be merged, preferably without needing `NXF_OFFLINE`, instead just working like this for all runs
* To be able to pin minimum versions in pipelines, not only exact versions: https://github.com/nextflow-io/nextflow/issues/4036

pditommaso:
I'm not sure we should encourage the use of unversioned plugins. Even minor version changes can introduce unexpected behaviour. 

Likely this has been a mistake done, it would be a practice to make the plugin version mandatory. This a common best practice for many package managers. 

To solve the problem of offline execution, I think the plugins should be downloaded along with the pipeline assets, and make nextflow able to recognise the plugins in local project assets path. 

How are the pipelines made available in such environment, if there's no network connection?

bentsherman:
> How are the pipelines made available in such environment, if there's no network connection?

You start in an environment that is connected to the internet but also has e.g. SSH access to the offline environment. You download everything you need (pipeline, plugins, containers) and then transfer it to the offline environment.

bentsherman:
I sketched out a process for preparing an offline environment here: https://github.com/nextflow-io/nextflow/issues/4330#issuecomment-1741025196

Now that we have `nextflow inspect`, we could have another command like `nextflow download` or extend `nf-core download` to perform this whole process. I originally imagined having to specify the plugins that you need, but I guess Nextflow could also infer it from the configuration.

@pditommaso it seems a lot of nf-core folks would agree with you to simply pin the plugin versions from now on. Still, it makes good sense to make NXF_OFFLINE also apply to plugins. The expectation of offline mode is that it allows you to work *completely offline* -- no downloading pipeline code, plugins, or other software deps.

pditommaso:
> Still, it makes good sense to make NXF_OFFLINE also apply to plugins

I agree! as such I think this PR should: 

1) when NXF_OFFLINE=false, keep the current behaviour 
2) when NXF_OFFLINE=true
  - throw an error if the plugin version is missing (do not check remote latest)
  - throw an error if a specific version of the plugin is not available locally 

A separate PR to enforce the plugin version for non-core plugins. 

Does it make sense ?

bentsherman:
I agree it would be useful to specify `>=` to automatically fetch patch releases. In offline mode, Nextflow could simply grab the minimum version from the local cache.

However, suppose I specify `>= 1.4.0`, then Nextflow downloads `1.4.1`, then I go into offline mode. Then it is not clear how Nextflow should recover `1.4.1` from the local cache when the pipeline specified 1.4.0 as the minimum version. It needs the plugin index file from GitHub, or we need to have some kind of lock file for plugins.

For now, I have created a separate PR #4487 to improve the error message with the expectation that pipeline devs will have to pin their plugin versions. That will give a solution for now while we debate over longer-term options.

ewels:
> Then it is not clear how Nextflow should recover `1.4.1` from the local cache when the pipeline specified `1.4.0` as the minimum version.

This is just standard semver version matching, no? We have a list of versions that we have locally, we have a range of accepted versions that we can use. It's just a case of getting a subset of acceptable versions and then picking the one with the most recent version number..

Nextflow is already using the [jsemver](https://github.com/zafarkhaja/jsemver) library to work with version numbers, and that has functionality to compare versions.

bentsherman:
Okay, based on our discussion from the last weekly meeting, we will adapt this PR to provide a `>=` mechanism to fetch the latest patch release (e.g. `1.0.x`) or use the local one when offline. The semver library can be used for this.

bentsherman:
Given the discussion here -- https://github.com/nextflow-io/nextflow/issues/5016#issuecomment-2126800485 -- regarding the meaning of `+` when checking the Nextflow version, we should use that instead of `>=` to implement this feature, so that they are consistent.

I always thought `+` just meant `>=`, but actually it means "use this exact version or a greater patch version"... which is exactly what we want here

PR #4348, opened 2023-09-26
Resolve Fusion symlinks when publishing files

### LINKED ISSUE #4309: Incorrect content when a process's published file is also an input and Fusion is enabled

When using Fusion if a process published a file that is also an input of that process the content of the published file is just a Fusion path to the input file. This happens because this is the way of Fusion to manage symbolic links.

The expected behavior is that the published file is a copy of the input file, with same content. 

```
$ nextflow run jordeu/nf-tests -r publish_input -w s3://fusion-develop/scratch
N E X T F L O W  ~  version 23.09.1-edge
Launching `https://github.com/jordeu/nf-tests` [stoic_rutherford] DSL2 - revision: b020907dd8 [publish_input]
executor >  local (fusion enabled) (2)
[b3/d224be] process > CREATE  [100%] 1 of 1 ✔
[b5/6c474f] process > PUBLISH [100%] 1 of 1 ✔

$ cat results/data.txt 
/fusion/s3/fusion-develop/scratch/b3/d224be3bcb4eefe89547058baba223/data.txt
```

### Environment 

```
$ nextflow info
  Version: 23.09.1-edge build 5881
  Created: 11-09-2023 10:08 UTC (12:08 CEST)
  System: Linux 5.15.0-83-generic
  Runtime: Groovy 3.0.19 on OpenJDK 64-Bit Server VM 17.0.4+8-LTS
  Encoding: UTF-8 (UTF-8)
```

bentsherman:
Close #4309


modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
```diff
@@ -283,6 +284,12 @@ class PublishDir {
         this.stageInMode = task.config.stageInMode
         this.taskName = task.name
 
+        // resolve Fusion symlinks
+        if( FusionHelper.isFusionEnabled(Global.session as Session) ) {
+            final inputFiles = task.getInputFilesMap()
+            files = files.collect { inputFiles.getOrDefault(it.name, it) } as Set<Path>
+        }
+
```
pditommaso, line 292: Not sure to understand how this works?

bentsherman: Probably was easier to read before I compacted it into three lines 😄 

When Fusion is enabled, if an output file is found to be an input, then we know that it is a Fusion symlink, so we should instead publish the link source, which we can obtain by reading the link file from S3 or from the input files map. Obviously the second way is more efficient.

The input files maps stage name to source path, so we check each output file to see if it's name is in the map, and if it is then replace the output file with the source path.

bentsherman:
I updated Jordi's example script as follows:
```groovy
nextflow.enable.dsl=2

process CREATE {

    output:
    path "data.txt"

    script:
    """
    echo HELLO > data.txt
    """
}

process FORWARD {

    input:
    path "data.txt"

    output:
    path "data.txt"

    script:
    """
    echo AND >> data.txt
    """
}

process PUBLISH {
    publishDir "${params.outdir}"

    input:
    path "data.txt"

    output:
    path "data.txt"

    script:
    '''
    echo BYE >> data.txt
    '''
}

workflow {
    CREATE | FORWARD | PUBLISH
}
```

And I found that the output file in `PUBLISH` links to the output from `FORWARD` rather than `CREATE`. I will try to fix that in Nextflow.

bentsherman:
Actually, it might not be so simple. When an input is forwarded as an output multiple times, there are conflicting needs:

- linking to the original file regardless of the degrees of separation is nice to avoid multiple levels of links
- however, these links are needed to determine task provenance (and the automatic work directory cleanup that is WIP)

If we can find a way to resolve the actual symlinks but preserve Nextflow's metadata, then maybe we can have it both ways. For example, we could add a field to `FileHolder` e.g. `originalStorePath` that points to the original file regardless of intermediate links.

bentsherman:
I worked around the previous problem by using the `.fusion.symlinks` files. However there is one more issue:

https://github.com/nextflow-io/nextflow/blob/945c01fb7a756a661d9ed9d3dee5ed5f17b30643/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy#L329

This line does not work with the resolved Fusion symlink, because now the resolved file is not in the current task directory. So instead of publishing to `results/data.txt` it goes to something like `results/dc/14a14e8ef9bc91cbf92e14e33715da/data.txt`.

I think this line is meant to deal with files that are in a subdirectory of the task directory, I should be able to fix it by resolving the Fusion symlink after this line.

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
```diff
@@ -400,6 +369,11 @@ class PublishDir {
     @CompileStatic
     protected void processFile( Path source, Path destination ) {
 
+        // resolve Fusion symlink if applicable
+        if( FusionHelper.isFusionEnabled(Global.session as Session) )
+            if( source.name in taskInputs )
```
jordeu, line 374: Given that we are relaying on Fusion file `.fusion.symlinks` to determine which are the links I think that here is better to use same mechanism. In this way if it's the user that creates a symbolic link it will also be supported.
```suggestion
            if( source.name in getFusionLinks(source.parent) )
```

bentsherman:
`PublishOpTest` is failing, but since `PublishOp` is not used and we decided not to make a `publish` operator, maybe it's better to delete

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
bentsherman: My concern is that this change will greatly increase the number of S3 requests when Fusion is enabled, even if they are small requests.

Currently this code happens only when an output is a symlink via `includeInputs`, which is rare. This change will make Nextflow check the `.fusion.symlinks` file for every task that publishes outputs, whether or not it uses `includeInputs`.

A user-provided symlink with Fusion is probably an anti-pattern anyway, so I would rather not impose the performance cost on the majority of use cases to support it.

On the other hand, it's a small file and probably has little impact on performance, so maybe it doesn't matter.

jordeu: The only case that I imagine that this can be a problem is if they decide to rename an input. But it's a bit a corner case, so make sense to keep as it is to avoid extra S3 requests.

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
```diff
@@ -387,6 +397,31 @@ class PublishDir {
...
+     * Resolve a Fusion symlink by following the .fusion.symlinks
+     * file in the task directory until the original file is reached.
+     *
+     * @param file
+     */
+    @CompileStatic
+    protected Path resolveFusionLink(Path file) {
+        while( file.name in getFusionLinks(file.parent) )
+            file = file.text.replace('/fusion/s3/', 's3://') as Path
+        return file
```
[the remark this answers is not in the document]
robsyme: To bake in support for Azure and GCP.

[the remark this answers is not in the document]
jordeu: Good catch. Yes we should be schema agnostic.

jordeu (approved):

jordeu:
We have some open tickets related to this problem. I'd be good to make it progress to master.

bentsherman:
Is there a simple way to mock the S3 interactions here? Or do you want to use real S3 objects?

pditommaso:
I'd try to moch the `resolveFusionLinks` method with and test the methods relying on it with a `Spy` .

Alternatively [localstack](https://github.com/localstack/localstack) could here

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
```diff
@@ -364,6 +369,11 @@ class PublishDir {
     @CompileStatic
     protected void processFile( Path source, Path destination ) {
 
+        // resolve Fusion symlink if applicable
+        if( FusionHelper.isFusionEnabled(Global.session as Session) )
+            if( source.name in taskInputs )
```
pditommaso, line 374: What's the semantics for `in` operator against a map? does it check it via `.containsKey`?

pditommaso:
Regex "compilation" can be slow. Moved into a separate const, so it's parsed only the very first time 👉 [0d853ea](https://github.com/nextflow-io/nextflow/pull/4348/commits/0d853ea5cda3b9c5538cccbb6cf2f687082b25a8)

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
bentsherman: Apparently the `in` operator casts the corresponding value to a boolean, which means e.g. `'foo' in ['foo': null]` would return false:

https://www.baeldung.com/groovy-collections-find-elements#3-map

Aside from that it is the same as `containsKey`

robsyme: Under the hood, the [membership operator](https://groovy-lang.org/operators.html#_membership_operator) is semantic sugar for [.isCase()](https://docs.groovy-lang.org/latest/html/groovy-jdk/java/util/Map.html#isCase(java.lang.Object)) for most objects (including Maps), and [.contains()](https://docs.groovy-lang.org/latest/html/groovy-jdk/java/lang/Iterable.html#contains(java.lang.Object)) for Lists.

The semantics are exactly as you describe Ben.

pditommaso:
Still failing

bentsherman:
I think the integration tests are not set up to use Fusion, seems like AWS credentials are not provided

pditommaso:
Umm, this it should be added `exportStorageCredentials` (good opportunity to update here as well)  

https://github.com/nextflow-io/nextflow/blob/816216614dd90b83a0261192729fccf85fc55e76/validation/wave-tests/example6/nextflow.config#L1-L21

Issue #3965, opened 2023-05-23
Bug: Wave does not support 'null' container engine


adamrtalbot:
## Bug report 

When running nf-core/sarek on AWS via Tower (with Wave and Fusion enabled), the error `Wave does not support 'null' container engine` occurred while executing process. 

### Expected behavior and actual behavior

Command launched by Tower:

```
nextflow run 'https://github.com/nf-core/sarek'
		 -name lethal_shannon
		 -params-file 'https://api.tower.nf/ephemeral/uivdKNFGgFNqOBSBZroySA.json'
		 -with-tower
		 -r 3.1.2
		 -profile test
```

### Steps to reproduce the problem

After retrying the pipeline, it worked fine so looks like an intermittent fault.

### Program output 

[nf-4WLnF5thJHWzde.txt](https://github.com/nextflow-io/nextflow/files/11542577/nf-4WLnF5thJHWzde.txt)
[nf-4WLnF5thJHWzde.log](https://github.com/nextflow-io/nextflow/files/11542578/nf-4WLnF5thJHWzde.log)

Relevant lines:
```
May-23 02:40:29.285 [main] DEBUG nextflow.processor.TaskProcessor - Starting process > NFCORE_SAREK:SAREK:MULTIQC
May-23 02:40:29.285 [main] DEBUG nextflow.script.ScriptRunner - > Awaiting termination 
May-23 02:40:29.285 [main] DEBUG nextflow.Session - Session await
May-23 02:40:29.477 [Actor Thread 17] DEBUG nextflow.sort.BigSort - Sort completed -- entries: 1; slices: 1; internal sort time: 0.002 s; external sort time: 0.014 s; total time: 0.016 s
May-23 02:40:29.477 [Actor Thread 30] DEBUG nextflow.sort.BigSort - Sort completed -- entries: 1; slices: 1; internal sort time: 0.002 s; external sort time: 0.014 s; total time: 0.016 s
May-23 02:40:29.691 [Actor Thread 17] DEBUG nextflow.file.FileCollector - Saved collect-files list to: /seqeralabs-showcase/work/sarek/work-20230523/collect-file/4061ae76e9a565e648322056d3a4b0f1
May-23 02:40:29.704 [Actor Thread 17] DEBUG nextflow.file.FileCollector - Deleting file collector temp dir: /tmp/nxf-3322981256689798888
[...]

PR #4349, opened 2023-09-27
Add debug messages for wave null container engine error

bentsherman:
Related to #3965

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -1186,6 +1186,8 @@ class Session implements ISession {
     @Memoized
     ContainerConfig getContainerConfig(String engine) {
 
+        log.debug "Fetching container config for '$engine' engine"
+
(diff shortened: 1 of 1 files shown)


pditommaso:
This looks suspicious 

https://github.com/nextflow-io/nextflow/blob/e4a878b6ce6d62bfd809fbbb88a6d4f2639e214f/modules/nextflow/src/main/groovy/nextflow/script/WorkflowMetadata.groovy#L242-L242

I fear the `.with` has some side effects on the ContainerConfig object. WDYT?

pditommaso:
I've followed a different approach, following back to `docker` when the engine is null and logging the wrong config object

bentsherman:
Okay, your changes seem more likely to help.

I don't understand though how the `.with()` would affect anything, since the workflow metadata container engine isn't referenced anywhere else in the code

pditommaso:
> I don't understand though how the .with() would affect anything

yeah, I don't believe is related. Just removed it because it's an unusual pattern

PR #4359, opened 2023-10-02
Add bioconda to wave conda-channel defaults.

ewels:
Based on Slack conversation with @pditommaso, who identified the line of code that needed updating.

Not sure if anything else needs changing as well? Or if this is sufficient.

Tagging @jason-seqera for docs visibility.

### DIFF
--- a/plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
+++ b/plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
@@ -79,7 +79,7 @@ class WaveClient {
                         'Accept','application/vnd.docker.distribution.manifest.v2+json',
                         'Accept','application/vnd.docker.distribution.manifest.list.v2+json' }
 
-    private static final List<String> DEFAULT_CONDA_CHANNELS = ['conda-forge','defaults']
+    private static final List<String> DEFAULT_CONDA_CHANNELS = ['conda-forge','bioconda','defaults']
 
     private static final String DEFAULT_SPACK_ARCH = 'x86_64'
 


plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
```diff
@@ -79,7 +79,7 @@ class WaveClient {
                         'Accept','application/vnd.docker.distribution.manifest.v2+json',
                         'Accept','application/vnd.docker.distribution.manifest.list.v2+json' }
 
-    private static final List<String> DEFAULT_CONDA_CHANNELS = ['conda-forge','defaults']
+    private static final List<String> DEFAULT_CONDA_CHANNELS = ['conda-forge','bioconda','defaults']
```
pditommaso, line 82: Let's use [this](https://github.com/nextflow-io/rnaseq-nf/blob/master/nextflow.config#L41) (order matter)

ewels: Order does indeed matter, and `conda-forge` must have higher priority than `bioconda`. https://bioconda.github.io/#explain-commands

```suggestion
    private static final List<String> DEFAULT_CONDA_CHANNELS = ['seqera','conda-forge','bioconda','defaults']
```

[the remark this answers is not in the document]
ewels:
```suggestion
    private static final List<String> DEFAULT_CONDA_CHANNELS = ['seqera','conda-forge','bioconda','defaults']
```

pditommaso: Then pulling @jordeu about conda channels, I remember this was discussed and, if I'm not wrong, it was said to use `seqera,bioconda,conda-forge,defaults`

jordeu: The logic one would be `seqera,bioconda,conda-forge,defaults`. And I think this was the recommended some time ago by bioconda. I see that now they say "note that the recommended configuration has changed over the years" and they recommend to give `conda-forge` more priority (even I do not understand why... unless there is some versioning mess on bioconda channel) 

So, @ewels suggestion seems the correct one.

pditommaso: but then it should be `conda-forge,bioconda,seqera,defaults

jordeu: Ok, I see that bioconda now also recommends to set `channel_priority: strict` that is not the default behavior (it's `flexible` the default). This may be the reason to set conda-forge before bioconda.
https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html

Not sure that I understand how `flexible` and `strict` work, specially when setting a different architecture

pditommaso: what a mess :D

ewels: I remember this going around a few years ago. I think that there are some deps on bioconda that are outdated but kept for reproducibility purposes, but conda-forge should be the preferred option for maintaining non-bioinformatics dependencies. So that's why it gets the higher priority. Or something like that, anyway.

It was quite a while ago now. I had lots of people complaining that running `conda install -c bioconda multiqc` didn't work any more...

pditommaso: ok, when you feel ready mark the PR as "Ready for review"

ewels: My questions:

1. Does top priority go first in the string (left) or last (right)?
2. Should the `seqera` channel have top priority? Or before / after bioconda?
3. Do we need to set it to strict mode somewhere, if so how. And will it break stuff @jordeu ? 😬

jordeu: Good questions :smile: 

1. left = top priority

The other two I've no idea. :cowboy_hat_face:

ewels: Poking around for where to add conda settings to look where to set `strict`, I get back to this line:

https://github.com/nextflow-io/nextflow/blob/0bce2be098bfd61bd6d78eb5e5379fd1002fea15/plugins/nf-wave/src/main/io/seqera/wave/plugin/config/WaveConfig.groovy#L23

So are the default conda options set in the Wave codebase?

pditommaso: There's no control over `strict` in nextflow and wave

ewels: Can support for this be added somehow? From the [conda docs](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html#strict-channel-priority):

> Strict channel priority can dramatically speed up conda operations and also reduce package incompatibility problems. We recommend setting channel priority to "strict" when possible.

pditommaso: Any clue how this is set in the Conda yaml spec?

ewels: After poking around a bit, I don't think that it is possible to set this in conda env YAML files 😞 

However, it can be set in a `.condarc` config file:

```yaml
channel_priority: strict
```

These files can also contain default channel priorities too, see [here](https://conda-forge.org/docs/user/tipsandtricks.html#how-to-fix-it). So if we use a `.condarc` file for global settings then we wouldn't have to specify channels in every env file.

According to [this SO comment](https://stackoverflow.com/questions/58555389/set-the-channel-priority-in-conda-environment-yaml#comment117120468_65983247), `strict` mode will become the default in Conda 5? But conda has since changed to calendar versioning, and I haven't found any other mention of it 😓

pditommaso: Umm, then should it be added to the CLI? This what is run to build the image 

https://github.com/seqeralabs/libseqera/blob/84f6a27de06901da7d1101c63ded647a5d291dc3/wave-utils/src/main/resources/templates/conda/dockerfile-conda-file.txt#L1-L7

ewels: Possibly, I'm getting a bit lost with all the different places that parts of this are happening. Wherever the conda environment is created (both Nextflow when using conda, and Wave when building from conda?), the `.condarc` file would go in the root directory of that environment. I think.

[docs](https://conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#channel-locations-channels):
> To select channels for a single environment, put a `.condarc` file in the root directory of that environment (or use the --env option when using `conda config`).

ewels: Alternatively, I wonder if this is getting too complicated. Maybe we just set the channels as in this PR and leave it to the Nextflow user to configure conda appropriately? That's already an improvement over the current status, and it's possible that setting this config won't be a requirement for very much longer anyway.

As I read it, having the default `channel_priority` will mostly make the environment resolution slower, not sure that it will necessarily break (unless you disagree @jordeu?).

Could then just drop a `.condarc` in the home directory of the wave base image for conda that says `channel_priority: strict` and we should be good to go...

pditommaso: Reading the docs I believe if we use this, the seqera channel is not going to work anymore 

> Setting strict channel priority makes it so that if a package exists on a channel, conda will ignore all packages with the same name on lower priority channels.

pditommaso: I don't know

jordeu: For me from the documentation is not clear what happens with the architecture restriction, if they apply before or after building the list of candidates. 

I'd say the `seqera,conda-forge,bioconda,defaults` and `flexible` (the default) channel priority should work well in "all" cases.

pditommaso: ok, let's merge this

pditommaso (changes requested):
Test broken :/

pditommaso:
I've tested this both as `seqera,bioconda,conda-forge,defaults` and ``seqera,conda-forge,bioconda,defaults`, and both works. 

Considering however, this is going to impact the builds of the associated containers, would bot be better to keep the minimal and leave this to the pipeline developer?

ewels:
I would prefer to have it as default. Otherwise, almost every wave user is going to need to specify this, which deviates from the goal of having minimal config.

We've effectively already been leaving it to the end user for a long time with conda support in Nextflow already. The channel setup is _such_ a common problem for Nextflow users that we have an auto-responder in the nf-core Slack with the solution:



People hit this problem (which never comes with a straightforward error message) very regularly.

pditommaso:
Ok, agree. However I believe the order they are suggesting is linked to the use of strict which is not supported by nextflow. 

I find more rational the order `'seqera','bioconda','conda-forge','defaults'` because it should check in a channel and fallback to the following one.

ewels:
> because it should check in a channel and fallback to the following one (from left to the right).

That's the point of having conda-forge first though - it should check in conda-forge before bioconda. Then it preferentially fetches dependencies (like Java) from conda-forge, where possible.

We don't want to prefer bioconda, only fall back to that if the package is not available on conda-forge.

Seqera I don't think it matters much as the channel is much smaller and we presumably don't have old core libs floating around there in the same way.

pditommaso:
i've tested all rnaseq pipeline, worked smoothly

rpetit3:
In most cases you will probably be OK, but if you are going to set a default, go with `seqera,conda-forge,bioconda,defaults` At least this will save future channel order issues. If you need extra incentive, you can just say its based on Bioconda recommendation:

> The conda-forge channel (see [docs](https://conda-forge.org/docs/index.html)) enables installation of general-purpose packages. Since bioconda heavily depends on the conda-forge channel, it is set to highest priority.

You might even consider removing "defaults" as it falls under under Anaconda's ToS when it comes to organizations with 200+ employees requiring a commercial license. This ToS does not apply to non-Anaconda channels (e.g. seqera, conda-forge, bioconda). But also I'm not a lawyer.

pditommaso:
I need a rationale explanation why `seqera` should go before `conda-forge` and `bioconda` after

rpetit3:
I am assuming `seqera` has custom built packages (e.g. arm) you prefer over the available `conda-forge` or `bioconda`packages, so it would be first. 

Then conda-forge, followed by bioconda. The main reasoning for conda-forge before Bioconda, is to avoid older packages being installed from bioconda that were moved to conda-forge (e.g perl packages not originally supported by conda-forge)

pditommaso:
More clear, thanks for the explanation. Does this apply only to the packages listed in the channel or also to any transitive dependencies?

rpetit3:
I'm unsure on transitive dependencies. Let me see if I can find an example of two packages that were moved off bioconda that depended on one or the other.

rpetit3:
Looks like it is also transitive dependencies.

Here is one `seqlike`

#### bioconda first

```
mamba create -n test-order -d -c bioconda -c conda-forge seqlike | grep -E "biopython|seqlike|weblogo"
  + biopython                    1.81  py311h2582759_0      conda-forge/linux-64       3 MB
  + seqlike                     1.1.6  pyh5e36f6f_0         bioconda/noarch           55 KB
  + weblogo                     3.7.9  pyhdfd78af_0         bioconda/noarch          406 KB
```

`weblogo` was pulled from `bioconda` even though a newer version is available from `conda-forge`. I'm guessing the solver checked for "newest" version without looking at the patch (e.g. `bioconda::weblogo=3.7.x == conda-forge:weblogo=3.7.x`, use bioconda one)

#### conda-forge first

```
mamba create -n test-order -d -c conda-forge -c bioconda seqlike | grep -E "biopython|seqlike|weblogo"
  + biopython                    1.81  py311h2582759_0      conda-forge/linux-64       3 MB
  + seqlike                     1.5.1  pyhd8ed1ab_0         conda-forge/noarch       321 KB
  + weblogo                    3.7.12  pyhd8ed1ab_0         conda-forge/noarch       524 KB
```

`weblogo` was pulled from `conda-forge`

pditommaso:
Ok, this clarifies my doubts. Solved 6672c6d7

ewels:
Thanks all! ❤️

Issue #4186, opened 2023-08-14
Add ability to provide custom defaults for publishDir directive via config file


pditommaso:
This ticket aims to provide the ability to custom default settings for the `publishDir` directive via the `nextflow.config` file. 

The default settings should be provided using the `nextflow.defaults.publishDir` configuration scope. For example: 

```
nextflow {
  defaults {
    publishDir = [mode: 'copy', enabled: false]
  }
}
```

When provided, they should override the built-in default settings and applied when creating a [PublishDir](https://github.com/nextflow-io/nextflow/blob/3051cd13da0298eaca142509ac05c7080a1c4328/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy#L169-L200) instance.

bentsherman:
@pditommaso you closed a bunch of issues and PRs about supporting closures for these fields, so do you intend for these default settings to support closures in some way? Otherwise it isn't really addressed those other issues. In many cases they want to configure the PublishDir with task-specific inputs which requires closures.

bentsherman:
If the publishDir directive can be specified in the pipeline code (#4375), then you could instead use a function to provide defaults:

```groovy
def publishDir(path, opts=[:]) {
  final defaults = [
    path: params.outdir,
    mode: params.publish_mode
  ]
  defaults + [path: path] + opts
}
```

bentsherman:
This convenience is no longer needed once the workflow output DSL #4784 is implemented

bentsherman:
Closing in favor of the output i.e. "publish" DSL and #4839 which provides some config options for publish retry strategy

pditommaso:
Not 100% this should be thrown away

PR #3793, opened 2023-03-23
Incorporate DSL2 features into docs

bentsherman:
This PR moves the DSL2 features (functions, workflows, modules) into the main docs, since they are well-established language features now. The DSL2 page has been refactored into a "DSL1 migration guide" so that there are still some DSL1 examples, since the rest of the docs use DSL2 now and we don't maintain docs for older versions.

The sections on functions, workflows, and modules were mostly copy/paste into their own sections, with some minor edits for clarity.

These changes should be consistent with what we were planning anyway for the website overhaul, but I decided to move forward with it now instead of wait, since the overhaul has been stalled for so long. These changes should make the docs truly DSL2-first.

PR #4375, opened 2023-10-04
Add process programmatic configuration

### LINKED ISSUE #4322: [proposal] Add process programmatic configuration context
## Context 

The use of DSL2 introduces the use of module components that allow the inclusion and re-use a process more than one time with different names.

A common requirement in this scenario is that a process configuration needs to be customised, for example, to provide a specific `publishDir` path or command arguments. 

Currently, this need is addressed by creating a monolithic configuration file in which the process configuration is provided by using the target process name. For example here 

https://github.com/nf-core/rnaseq/blob/master/conf/modules.config#L174-L192

In some case, there's a conditional logic that needs to be duplicated in the workflow script. For example here: 

https://github.com/nf-core/rnaseq/blob/master/conf/modules.config#L209-L224

This approach has several disadvantages: 

* Monolithic configuration file
* The process configuration is out of the context from the workflow code 
* Very fragile; it's enough a typo in the config file to break the process  configuration 
* The use of monolithic breaks the workflow modularization 
* Very hard to maintain. 

## Solution 

A possible solution is to allow the process configuration in the context of the workflow definition instead of the configuration file.

A possible syntax could be the following: 

```
include { FOO as BAR } from 'some/module'

workflow myAnalyis {

  // define process custom config ahead of invocation 
  BAR.config.publishDir.path = "${params.outdir}/sample_fastq/fastq" 
  BAR.config.publishDir.mode = params.publish_dir_mode

  // invoke the process execution 
  BAR()

  // access the output 
  BAR.out.view() 
 
} 
```

The `.config` syntax will allow overriding the process default configuration in the context of the workflow execution. 

This has several advantages: 

* no need to use closures for lazy evaluation, see [here](https://github.com/nf-core/rnaseq/blob/master/conf/modules.config#L178)
* the config is keep in the same context where the logic is defined 
* prevent bug monolitic config file 
* enable modulariation 
* the `.config` property is consistent with the already existing syntax of process acceessors e.g. `.out`
* relatively easy to implement

bentsherman:
Close #4322 

Example test case:

```groovy
// main.nf

include { FOO ; FOO as BAR } from './module.nf'

params.outdir = 'results'
params.publish_dir_mode = 'copy'

workflow {

    FOO.config.ext.extension = 'foo'
    FOO( Channel.of('1', '2', '3') )
    FOO.out.view { "FOO: $it" }

    BAR.config.ext.extension = 'bar'
    BAR.config.publishDir params.outdir, mode: params.publish_dir_mode
    BAR( Channel.of('1', '2', '3') )
    BAR.out.view { "BAR: $it" }

}
```

```groovy
// module.nf

process FOO {
    input:
    val(id)

    output:
    tuple val(id), path("sample${id}.${task.ext.extension}")

    script:
    """
    echo '${id} was here' > sample${id}.${task.ext.extension}
    """
}
```

```console
$ ./launch.sh run main.nf
N E X T F L O W  ~  version 23.09.2-edge
Launching `main.nf` [astonishing_goldwasser] DSL2 - revision: 9b1341eb15
executor >  local (6)
[cd/f6fe43] process > FOO (3) [100%] 3 of 3 ✔
[16/d722fe] process > BAR (1) [100%] 3 of 3 ✔
BAR: [2, work/ad/d60943813d18751a8ded972ee9d798/sample2.bar]
BAR: [3, work/89/9295138db148949f054a22708895a4/sample3.bar]
FOO: [2, work/48/659fef3cb5e492e2f181c975ea232c/sample2.foo]
FOO: [1, work/db/43db80ff13d8d3c7301467e14184db/sample1.foo]
BAR: [1, work/16/d722fe4b621da24e9bd1e7f30cfd1e/sample1.bar]
FOO: [3, work/cd/f6fe432184777c0f08c9e5c1edee30/sample3.foo]

$ ls -al results/
total 1224
drwxr-xr-x  3 bent bent    4096 Oct  3 12:44 ./
drwxr-xr-x 88 bent bent   12288 Oct  3 12:38 ../
-rw-r--r--  1 bent bent      11 Oct  3 12:38 sample1.bar
-rw-r--r--  1 bent bent      11 Oct  3 12:38 sample2.bar
-rw-r--r--  1 bent bent      11 Oct  3 12:38 sample3.bar

$ cat results/*.bar
1 was here
2 was here
3 was here
```
```

### DIFF
--- a/docs/dsl2.md
+++ b/docs/dsl2.md
@@ -183,6 +183,26 @@ workflow {
 Optional params for a process input/output are always prefixed with a comma, except for `stdout`. Because `stdout` does not have an associated name or value like other types, the first param should not be prefixed.
 :::
 
+### Process config
+
+:::{versionadded} 23.10.0
+:::
+
+A process can be configured with additional directives from a workflow by using the `config` attribute on a process. For example:
+
+```groovy
+include { foo ; foo as bar } from './module.nf'
+
+workflow {
+    foo()
+
+    bar.config.publishDir 'results', mode: 'copy'
+    bar()
+}
+```
+
+Process directives can be defined using the same syntax as in a process definition, and they are applied separately to different module aliases. In the above example, the `publishDir` directive is applied to process `bar` but not to process `foo`.
+
 ## Workflow
 
 ### Workflow definition
--- a/docs/dsl2.md
+++ b/docs/dsl2.md
@@ -183,6 +183,8 @@ workflow {
 Optional params for a process input/output are always prefixed with a comma, except for `stdout`. Because `stdout` does not have an associated name or value like other types, the first param should not be prefixed.
 :::
 
+(dsl2-process-config)=
+
 ### Process config
 
 :::{versionadded} 23.10.0
--- a/modules/nextflow/src/main/groovy/nextflow/script/ProcessDef.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/ProcessDef.groovy
@@ -145,7 +145,10 @@ class ProcessDef extends BindableDef implements IterableDef, ChainableDef {
 
     String getBaseName() { baseName }
 
-    ProcessConfig getProcessConfig() { processConfig }
+    ProcessConfig getConfig() {
+        if( processConfig == null ) initialize()
+        processConfig
+    }
 
     ChannelOut getOut() {
         if( output==null )
@@ -165,7 +168,8 @@ class ProcessDef extends BindableDef implements IterableDef, ChainableDef {
     @Override
     Object run(Object[] args) {
         // initialise process config
-        initialize()
+        if( processConfig == null )
+            initialize()
 
         // get params 
         final params = ChannelOut.spread(args)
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: docs/dsl2.md [hunk 1]
```diff
@@ -183,6 +183,26 @@ workflow {
 Optional params for a process input/output are always prefixed with a comma, except for `stdout`. Because `stdout` does not have an associated name or value like other types, the first param should not be prefixed.
 :::
 
+### Process config
```
bentsherman, line 186: Ideally would like to merge #3793 first, then this would become part of the "Workflows" page which is more prominent.

modules/nextflow/src/main/groovy/nextflow/script/ProcessDef.groovy
bentsherman, line 1: These changes were just the minimal hacking required to make the test case work. Ideally the `initialize()` should be called in a more deterministic way, while still supporting module aliases.

bentsherman:
I added the config setting for publishDir defaults as defined in #4186 so that I can further improve the nf-core/rnaseq POC. Doesn't have to be merged with the final PR but it's a simple change.

drpatelh:
This is awesome! Thank you! Will take a closer look when I get a moment. Also, thanks for testing with nf-core/rnaseq!

bentsherman:
Closing in favor of #4510

PR #4376, opened 2023-10-04
Throw error if a process defines multiple inputs with the same name

### LINKED ISSUE #4366: Throw error if a process has multiple inputs with the same name
Consider the following example:
```groovy
workflow {
    ch_foo = Channel.of( ['1', '1.foo'], ['2', '2.foo'] )
    ch_bar = Channel.of( ['2', '2.bar'], ['1', '1.bar'] )
    gather(ch_foo, ch_bar)
}

process gather {
    input:
    tuple val(id), file(foo)
    tuple val(id), file(bar)
    """
    merge_command $foo $bar
    """
}
```

Many users follow this pattern assuming that Nextflow will automagically match incoming items based on `id`.

But it does not.

Instead it simply grabs input pairs as they come in, like the `merge` operator. If you're lucky, and the two channels just happen to have the same order, nothing bad will happen. If you're not lucky, bad things will happen and you might not even notice.

To fix this, Nextflow should throw an error if a process defines multiple inputs with the same name. And probably add a little blurb that "Nextflow is not doing what you think it's doing" or link to the docs. I'm writing a new page about caching behavior which will have this example as an error case.

bentsherman:
Close #4366

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -417,6 +417,22 @@ class TaskProcessor {
         if ( !taskBody )
             throw new IllegalStateException("Missing task body for process `$name`")
 
+        // -- check that multiple inputs with the same name are not defined
+        def inputNames = []
+        config.getInputs()
+            .collect { InParam param ->
+                param instanceof TupleInParam
+                    ? param.inner
+                    : param
+            }
+            .flatten()
+            .each { InParam param ->
+                if( param.name in inputNames )
+                    throw new IllegalArgumentException("Input '${param.name}' defined multiple times in process `$name`")
+                else
+                    inputNames << param.name
+            }
+
         // -- check that input set defines at least two elements
         def invalidInputSet = config.getInputs().find { it instanceof TupleInParam && it.inner.size()<2 }
         if( invalidInputSet )


mribeirodantas:
Stuff like this, right? 😓

CleanShot 2023-10-09 at 11 19 33@2x

pditommaso:
Yep. ideally it should break only if the values do not match. Not sure about the impact of implement this tho

bentsherman:
@mribeirodantas those are outputs, this error is for inputs. Inputs should *never* have duplicate names, and if any nf-core pipelines are doing it, it is a bug and should cause an error message. Let me know if you find any such examples in nf-core pipelines

mribeirodantas:
I referred to the highlighted part of the print screen (meta). That print screen is of the output block, but it is not hard to find an example of multiple metas in the input block. Check [here](https://github.com/nf-core/modules/blob/988193a6dfbe2c8571c1e57b4900c32213bebc41/modules/nf-core/peka/main.nf#L10).

bentsherman:
I found some more examples:
- [hic/bowtie2/align](https://github.com/nf-core/hic/blob/fe4ac656317d24c37e81e7940a526ed9ea812f8e/modules/nf-core/bowtie2/align/main.nf#L11)
- [nascent/bwa/mem](https://github.com/nf-core/nascent/blob/9ff33c78968717b2cee89d7b5be2b5110facf928/modules/nf-core/bwa/mem/main.nf#L11)
- [raredisease/tiddit/cov](https://github.com/nf-core/raredisease/blob/c371602095ff83064f51747fe90dfc405ca7a864/modules/nf-core/tiddit/cov/main.nf#L11)
- [raredisease/gatk4/mergevcfs](https://github.com/nf-core/raredisease/blob/c371602095ff83064f51747fe90dfc405ca7a864/modules/nf-core/gatk4/mergevcfs/main.nf#L11)

What's interesting is that in all cases they are named `meta` and `meta2`:
```groovy
    input:
    tuple val(meta) , path(reads)
    tuple val(meta2), path(index)
```

So actually they would get around this error check 😅 

@ewels is there some nf-core convention when a process has multiple inputs with the same meta map? Technically they should be joined into a single tuple, but it looks like in all the cases linked above, the second input is some value channel like an index, so there is no risk of sample mismatch. But the fact that they are named `meta` and `meta2` makes me think that people are aware of this problem...

ewels:
Tagging @mribeirodantas to reply to this one please ☝🏻

mribeirodantas:
I replied already 😅. This print screen you shared was attached above too, and even though it's an output block, in my last message above I linked to an nf-core pipeline that contains two input variables with the name meta 😃

bentsherman:
We are going in circles... maybe we should just discuss this at the hackathon

ewels:
Ahh sorry @mribeirodantas - multitasking too much and didn't read properly 🤦🏻‍♂️ 



> But the fact that they are named `meta` and `meta2` makes me think that people are aware of this problem...

Yes I think you're probably right 👍🏻  Shouldn't be too hard to check? Can look at the hackathon. But yeah if it's just inputs then I think it'll probably fine.

Any conventions etc we have should be documented here: https://nf-co.re/docs/contributing/modules - I can't find anything from a quick search though.

bentsherman:
After more investigation I found that for these examples with `meta` and `meta2`, they are actually different meta maps, usually for samples and an index.

So this PR should not break any nf-core pipelines. And if a pipeline is doing this, the error will force a critical bug fix. But if it would make you happier @pditommaso , we could have it be a warning instead.

bentsherman:
Closing in favor of the new script parser which will report an error for this

PR #4382, opened 2023-10-06
Add some details about what a workflow can contain

maxulysse:
A workflow can contains no process or no operator, I'd say even none of them, but that'll be useless. But at least we can specify in the docs that it can be one or the other.


pditommaso:
Thanks Max. Find it clear in the current form

PR #4387, opened 2023-10-09
Support for Fusion unprivileged execution

pditommaso:
This PR the config option 

```
fusion.privileged = true|false // defualt true
```
 
Enable or disable the privileged execution required by Fusion. The implementation depends on the target executor


pditommaso:
Ready

jordeu (approved):
It's working! :tada: 

I've double checked that the pod do not uses privileged mode.

pditommaso:
Awesome

PR #4416, opened 2023-10-16
Conceptual error in process selector documentation

saulpierotti:
Hi, the documentation for this seems inaccurate to me. While the first part is true that the withName selector applies also to the original name of a module, not only the import name, in my usage experience it does also apply to the import name. For example if I have a process named FOO and I import it with import FOO as BAR, a withName:BAR selector will work, and also a withName:FOO selector will. The documentation suggests that only withName:FOO works, which I believe to be incorrect.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1195,7 +1195,7 @@ process {
 ```
 
 :::{note}
-The `withName` selector applies to a process even when it is included from a module under an alias. For example, `withName: hello` will apply to any process originally defined as `hello`, regardless of whether it is included under an alias. Similarly, it will not apply to any process not originally defined as `hello`, even if it is included under the alias `hello`.
+The `withName` selector applies to a process even when it is included from a module under an alias. For example, `withName: hello` will apply to any process originally defined as `hello`, regardless of whether it is included under an alias. However, a selector referring to the alias of a process will also be applied.
 :::
 
 :::{tip}


saulpierotti:
I actually find it useful to have it this way - I can have the same module and import it with two different aliases, and then have it run with different parameters/modes according to the alias. An example is https://nf-co.re/modules/stitch, which I import once for generating the input only with certain ext.args, and once for actually running the imputtaion with other ext.args

mribeirodantas:
It seems to be the case, I can reproduce on my side, @bentsherman.

```
.
├── main.nf
├── modules
│   └── foo
│       └── main.nf
└── nextflow.config

3 directories, 3 files
```

./main.nf
```
include { FOO        } from './modules/foo/main.nf'
include { FOO as BAR } from './modules/foo/main.nf'

workflow {
  FOO()
  BAR()
}
```

nextflow.config
```
process {
  withName: FOO {
    cpus = 1
  }
  withName: BAR {
    cpus = 2
  }
}
```

./modules/foo/main.nf
```
process FOO {
  debug true

  output:
  stdout

  script:
  """
  echo ${task.cpus}
  """
}
```

Output:
```
Launching `main.nf` [trusting_colden] DSL2 - revision: 466eb880a9
executor >  local (2)
[9f/cf382e] process > FOO [100%] 1 of 1 ✔
[0e/748233] process > BAR [100%] 1 of 1 ✔
1

2
```

bentsherman (approved):
Fair enough, I guess if you don't want the base config to be propagated to aliases then you can simply not use it and alias every process inclusion.

bentsherman:
cc @marcodelapierre 

consider folding this PR into the module config #4510

bentsherman:
Closing in favor of #4510

PR #4421, opened 2023-10-17
Reflect current macOS naming convention

HarryHung:
OS X has been officially known as macOS since 2016.
I think it would be a good idea to use the official naming convention to avoid user confusion. 
Might need to consider updating the [Nextflow home page](https://www.nextflow.io/) as well


pditommaso (commented):
Right!

Issue #4386, opened 2023-10-09
'versions' directive in process


adamrtalbot:
## New feature

Many pipelines attempt to catch the software versions running within the container. The most common method is to catch a file with the software version written to it, typically using some fun string parsing:

https://github.com/nf-core/modules/blob/72f633b541815ae9565599e5b3cf66d863736ae1/modules/nf-core/fastqc/main.nf#L38-L41

https://github.com/epi2me-labs/wf-template/blob/3edea44cdf9449093a2adc2f08bb16ec143373bf/main.nf#L20-L30

This populates the output directive with an additional channel and prevents people using the pipe operator amongst other things.

Why not get this into a separate, native directive? It could automatically populate some property of the `process` which is available to use as a string (or map of strings).

## Usage scenario 

Anyone who wants to record their software at runtime. 

## Suggest implementation 

```nextflow
process FASTQC {
    // process goes here
    versions:
    fastqc = 'fastqc --version | sed -e "s/FastQC v//g"'
    python = 'python --version'
}
```

Somewhere within the .command.run the .task.versions file is created:
```
fastqc: 0.12.1
python: 3.11.2
```

Back in the pipeline code:
```nextflow
FASTQC.versions == [ name: ${task.process}, versions: [ fastqc: "0.12.1", python: "3.11.2" ]  ]

// It's syntactic sugar around the FASTQC.out.versions channel, so can be used like so:
ch_versions = ch_versions.mix(FASTQC.versions)
```

ewels:
I think that this is basically a duplicate of an issue that I wrote with @pditommaso back in 2018 😬 https://github.com/nextflow-io/nextflow/issues/879

But same motivation, similar syntax even (this is slightly better though, agree that the ability to have more than 1 version per process is necessary 👍🏻 )

adamrtalbot:
Ah bother. I did search for `versions` but I see that feature is slightly more generic. You could probably use my this feature to implement generic info as well as versions. 

The biggest flaw I can see right now is stealing a secret. From the example above:

```
versions:
sneaky-secret-steal = "echo $SECRET"
```

adamrtalbot:
In my personal experience, conda only contains about 25% of tools used in bioinformatics (although they are the big and important ones).

ewels:
[...]

PR #4425, opened 2023-10-18
Allow to add custom traces and use them as metadata

jordeu:
# Description 

This PR allows the user to add custom traces on each process. Also exposes them as workflow metadata, so the user can use them without any extra channel.

This is a more generic approach to solve the version tracking problem described at #4386 

The idea is to allow the user to collect custom traces before running the command. The custom traces is a map where the key is the name of the custom trace and the value is an string with the bash script to run to collect that trace. The output is always parsed as an string.

Then all the traces of the completed and resumed tasks are available as workflow metadata at `workflow.traces`. If `workflow.traces` is used as part of a process script it will contain all the traces of completed task just before submitting that process.

You can use `workflow.traces` at `workflow.onComplete` if you want to store custom traces into a file or do something else.

# Example pipeline

See the whole pipeline [here](https://github.com/jordeu/nf-tests/tree/versions) 

**main.nf**
```nextflow
def parseVersions(traces) {
  result = ''
  for( t in traces ) {
    v = t.getStore().keySet().findAll{ it.startsWith('custom_version_') }.collect{ "$it: ${t.getStore().get(it)}" }
    if( v ) { 
      result += t.processName + ":\n    " + v.join("\n    ") + "\n"
    }
  } 
  return result
}


process FASTQC {
  conda "bioconda::fastqc=0.11.9"
  customTraces version_fastqc: "fastqc --version | sed -e 's/FastQC v//g'"

  output:
    path 'output.txt'

  """
  echo "FASTQC" > output.txt
  """
}


process BAMTOOLS {
  conda "bioconda::bamtools=2.5.2"
  customTraces version_bamtools: "bamtools --version | grep -e 'bamtools' | sed 's/^.*bamtools //'"

  output:
    path 'output.txt'

  """
  echo "BAMTOOLS" > output.txt
  """
}


process MULTIQC {

  input:
    path 'fastqc.txt'
    path 'bamtools.txt'

  script:
    versions = parseVersions(workflow.traces)
    """
    cat <<-END_VERSIONS > versions.yml
${versions}
END_VERSIONS

    cat fastqc.txt bamtools.txt > result.txt
    """
}

workflow {

  FASTQC()
  BAMTOOLS()

  MULTIQC(FASTQC.out, BAMTOOLS.out)
}

workflow.onComplete {
  println "EXAMPLE VERSIONS YAML:\n${parseVersions(workflow.traces)}"   
}
```

# Notes

- This is still a PoC, open to discuss this approach and alternatives.
- The `workflow.traces` is an array of `TraceRecord` of the completed tasks
- Maybe instead of using directly `TraceRecord` to expose traces to the user, would be better to expose them using a custom class more user friendly.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -323,6 +323,9 @@ class BashWrapperBuilder {
         // patch root ownership problem on files created with docker
         binding.fix_ownership = fixOwnership() ? "[ \${NXF_OWNER:=''} ] && (shopt -s extglob; GLOBIGNORE='..'; chown -fR --from root \$NXF_OWNER ${workDir}/{*,.*}) || true" : null
 
+        binding.custom_trace_collect = getCustomTraceCollect()
+        binding.custom_trace_write = getCustomTraceWrite()
+
         binding.trace_script = isTraceRequired() ? getTraceScript(binding) : null
         
         return binding
@@ -424,6 +427,26 @@ class BashWrapperBuilder {
         String result=''
         for( String it : moduleNames) {
             result += moduleLoad(it) + ENDL
+        }
+        return result
+    }
+
+    private String getCustomTraceCollect() {
+        if( !customTraces )
+            return null
+        String result=''
+        for( String key : customTraces.keySet() ) {
+            result += "local custom_${key}=\$(${customTraces.get(key)})"
+        }
+        return result
+    }
+
+    private String getCustomTraceWrite() {
+        if( !customTraces )
+            return null
+        String result=''
+        for( String key : customTraces.keySet() ) {
+            result += "echo \"custom_${key}=\$custom_${key}\" >> \$trace_file"
         }
         return result
     }
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskBean.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskBean.groovy
@@ -51,6 +51,8 @@ class TaskBean implements Serializable, Cloneable {
 
     List<String> moduleNames
 
+    Map<String,String> customTraces
+
     Path workDir
 
     Path targetDir
@@ -156,6 +158,7 @@ class TaskBean implements Serializable, Cloneable {
         this.stageOutMode = task.config.getStageOutMode()
 
         this.resourceLabels = task.config.getResourceLabels()
+        this.customTraces = task.config.getCustomTraces()
     }
 
     @Override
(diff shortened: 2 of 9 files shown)


muffato:
I very much like the idea of "custom traces". That can be used to collect input sizes (for instance size in bytes of some files, or number of lines, etc).

Question about the interface `workflow.traces.getStore().keySet().findAll(...)`. In the case of a pipeline with processes running concurrently, a process accessing that store will see different values, depending of how early it is run in comparison to the other processes. For reproducibility, I would expect a process to only have access to the traces of its parents. To reliably have access to _all_ traces, the reader process would have to be marked as depending on every stream of processes, to make sure it's run last. Or if a Nextflow process is not required, `workflow.onComplete` should have access to it too ?

pditommaso:
What would be a use-case for this PR?

jordeu:
I see two use cases:

 - Collect info of all the tools used in the pipeline and report them on the last process with a tool like MultiQC
 - Collect custom metrics of your processes (ex: free disk space, input file sizes...) and on `workflow.onComplete` store them in a file.

jordeu:
> Question about the interface `workflow.traces.getStore().keySet().findAll(...)`. In the case of a pipeline with processes running concurrently, a process accessing that store will see different values, depending of how early it is run in comparison to the other processes. For reproducibility, I would expect a process to only have access to the traces of its parents. To reliably have access to _all_ traces, the reader process would have to be marked as depending on every stream of processes, to make sure it's run last. Or if a Nextflow process is not required, `workflow.onComplete` should have access to it too ?

It's true that when you use `workflow.traces` you get all the completed tasks and this is not deterministic if you are using this in a process that runs concurrently with others. To only give access to upstream traces would be a bit nicer solution, but don't seems easy to implement. I think that traces are just metadata and should not be used for anything that affects the reproducibility of a pipeline.

bentsherman:
TODO:
- [ ] rename `customTraces` to `trace`
- [ ] allow `trace` to be defined multiple times like `label`, `publishDir`, etc
- [ ] remove `custom_` prefix, instead make sure custom names don't conflict with standard trace fields
- [ ] document process directive
- [ ] document `workflow.trace` as a metadata field that is only available in `onComplete` and `onError` handlers

pditommaso:
I'm not convinced we should proceed down this path. It would lead to even more do-it-yourself metadata, which the nextflow runtime would be totally unaware of. Think we should have a more controlled approach

bentsherman:
Let's take this discussion back to the original issue #4386

pditommaso:
I've looked more into this, and while the idea is interesting and we should pursue it to some extent, what I'm concerned about between the other things is that it's very fragile. it's enough an extra new-line introduced by a custom command to break the trace file. Also, an invalid command can break the command wrapper script, and alter the time metrics collection. 

If we want to go ahead with the idea of collecting custom metadata running third-party tools, it should be independent by the trace mechanism

jordeu:
> If we want to go ahead with the idea of collecting custom metadata running third-party tools, it should be independent by the trace mechanism

When you say independent... do you mean to isolate them in a new function at the `.command.run` or something else?

pditommaso:
Yes, possibly. Likely it should be executed just before or after the task command. https://github.com/nextflow-io/nextflow/issues/540

pditommaso:
Closing in favor of https://github.com/nextflow-io/nextflow/pull/4459

PR #4434, opened 2023-10-19
Add k8s job ttlSecondsAfterFinished option

bentsherman:
See [Kubernetes docs](https://kubernetes.io/docs/concepts/workloads/controllers/job/#ttl-mechanism-for-finished-jobs) for details


modules/nextflow/src/main/groovy/nextflow/k8s/model/PodOptions.groovy
```diff
@@ -67,6 +67,8 @@ class PodOptions {
     private List<Map> tolerations
 
     private Boolean privileged
+
+    private Integer ttlSecondsAfterFinished
```
pditommaso, line 71: it could make sense to set a sensible default for this e.g. 1 hour

bentsherman: I didn't want to do that initially because I thought it could override a cluster-level setting, but after checking it seems there is no such cluster-level setting, and in fact the K8s docs recommends setting this option explicitly:

> It is recommended to set ttlSecondsAfterFinished field because unmanaged jobs (Jobs that you created directly, and not indirectly through other workload APIs such as CronJob) have a default deletion policy of orphanDependents causing Pods created by an unmanaged Job to be left around after that Job is fully deleted. Even though the [control plane](https://kubernetes.io/docs/reference/glossary/?all=true#term-control-plane) eventually [garbage collects](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-garbage-collection) the Pods from a deleted Job after they either fail or complete, sometimes those lingering pods may cause cluster performance degradation or in worst case cause the cluster to go offline due to this degradation.

I will set the default to 1 hour

pditommaso:
OK, just removed the default to 1h. We can it in a later stage if really needed

PR #4440, opened 2023-10-25
Add support for Singularity OCI mode

pditommaso:
This PR adds support for the OCI-mode provided by Singularity 4. More details at [this link](https://docs.sylabs.io/guides/4.0/user-guide/oci_runtime.html#oci-mode).

### Guidelines for testing 

1. Install Singularity 4 and `runc` (see [here](https://docs.sylabs.io/guides/4.0/user-guide/oci_runtime.html#oci-mode) for details)

2. enable the singularity OCI mode in the nextflow config 

     ```
     singularity.enabled = true
     singularity.oci = true
     ``` 

3. run a test pipeline 

     ```
     nextflow run hello
     ```


pditommaso:
@emi80 you may want to give a try to this branch

emi80:
Hey @pditommaso,

tryed to run as from your guideline and got the following error on an Ubuntu 22.04 machine and `singularity-ce version 4.0.1-jammy`:

```
Command error:
  FATAL:   While initializing: while applying crun cgroup workaround: couldn't create cgroup manager: rootless cgroups require a D-Bus session - check that XDG_RUNTIME_DIR and DBUS_SESSION_BUS_ADDRESS are set
```

It seems like exporting those two env variables in the command makes it work.

pditommaso:
oohh, should not those vars be provided by the host environment?

emi80:
Yes, but the Singularity command in `.command.run` is prepended with `env - ` and I think that removes the vars from the environment...

emi80:
If I run the command:
```
set +u; env - XDG_RUNTIME_DIR="$XDG_RUNTIME_DIR" DBUS_SESSION_BUS_ADDRESS="$DBUS_SESSION_BUS_ADDRESS" PATH="$PATH" ${TMP:+SINGULARITYENV_TMP="$TMP"} ${TMPDIR:+SINGULARITYENV_TMPDIR="$TMPDIR"} ${NXF_TASK_WORKDIR:+SINGULARITYENV_NXF_TASK_WORKDIR="$NXF_TASK_WORKDIR"} singularity exec --no-home --pid --oci -B /data/epalumbo/test-nf/oci/work/45/55fcd590ebc3c1dc5d7c9d891bf748 docker://quay.io/nextflow/bash /bin/bash -ue /data/epalumbo/test-nf/oci/work/45/55fcd590ebc3c1dc5d7c9d891bf748/.command.sh
```

it pulls the image, but then gives me the following error:
```
❯ bash .command.run
INFO:    Using cached OCI-SIF image
INFO:    --oci runtime uses a PID namespace by default, pid flag is redundant.
chdir: No such file or directory
```

emi80:
Removing `--no-home` or adding `-B $HOME` works

pditommaso:
> I think that removes the vars from the environment

ok, then adding those to the `singularity.envWhitelist` should solve it 

https://www.nextflow.io/docs/latest/config.html#scope-singularity

> Removing --no-home or adding -B $HOME works

Independently of the above variables?

emi80:
> ok, then adding those to the singularity.envWhitelist should solve it

This adds `${XDG_RUNTIME_DIR:+SINGULARITYENV_XDG_RUNTIME_DIR="$XDG_RUNTIME_DIR"} ${DBUS_SESSION_BUS_ADDRESS:+SINGULARITYENV_DBUS_SESSION_BUS_ADDRESS="$DBUS_SESSION_BUS_ADDRESS"}` but it does not work. Still getting:
```
Command error:
  FATAL:   While initializing: while applying crun cgroup workaround: couldn't create cgroup manager: rootless cgroups require a D-Bus session - check that XDG_RUNTIME_DIR and DBUS_SESSION_BUS_ADDRESS are set
```

That is because it prepends `SINGULARITYENV_` to the variable names.

> Independently of the above variables?

No,  `XDG_RUNTIME_DIR` and `DBUS_SESSION_BUS_ADDRESS` must be set

pditommaso:
Ok, those variables can added explicitly, but are those names portable? 


Also, I'm still confused by the `-B $HOME` binding. Why is that necessary?

emi80:
As far as I understand those variables are standard and are needed to use cgroups v2.

> Also, I'm still confused by the -B $HOME binding. Why is that necessary?

Maybe this is something related to the pipeline I am running: `nextflow-io/hello`.

emi80:
I can confirm that with `nextflow-io/rnaseq-nf` The `-B $HOME` binding is not needed.

pditommaso:
Ok, i'll just add the support for `XDG_RUNTIME_DIR` and `DBUS_SESSION_BUS_ADDRESS` env vars then

emi80:
Thanks!

It seems that `rnaseq-nf` works because the bind `-B /home/epalumbo/.nextflow/assets/nextflow-io/rnaseq-nf` is added to the Singularity command in `.command.run` while this is not done with `hello`.

Indeed, it seems to be an issue with Singularity and OCI:
```
❯ singularity exec --no-home docker://ubuntu echo Hello
INFO:    Using cached SIF image
Hello

❯ singularity exec --no-home --oci docker://ubuntu echo Hello
INFO:    Using cached OCI-SIF image
2023-10-26T14:12:46.000537361Z: chdir: No such file or directory
```

pditommaso:
You can use `NXF_SINGULARITY_HOME_MOUNT=true` to enable it

emi80:
I opened https://github.com/sylabs/singularity/issues/2289. Let's see what they tell me...

pditommaso:
Where did you added XDG_RUNTIME_DIR and DBUS_SESSION_BUS_ADDRESS vars to test it?

emi80:
It's a problem with the low-level runtime that Singularity uses in `oci` mode.

See https://github.com/sylabs/singularity/issues/2289#issuecomment-1781273779.

emi80:
It looks like setting `--cwd` in the SIngularity command would fix the `--no-home` issue and would be more portable as it would work with different OCI runtimes.

See https://github.com/sylabs/singularity/issues/2289#issuecomment-1781290057

pditommaso:
If it's a singularity problem, I'm inclined to ignore it, and only add the variables

emi80:
Indeed, it's actually an OS problem so you can safely ignore it 👍

pditommaso:
Done

emi80:
BTW, the `--pid` should be removed in OCI mode as it's redundant

pditommaso:
Uff, why?

emi80:
```
INFO:    --oci runtime uses a PID namespace by default, pid flag is redundant.
```

pditommaso:
I see

pditommaso:
Done. what else? 😆

emi80:
Mostly good news...

It works with the `nextflow-io/hello` pipeline both using the default container and with Wave.

With the `nextflow-io/rnaseq-nf` pipeline it works using biocontainers but it's not working with Wave (using the `conda` strategy), e.g.:

```
Command error:
  INFO:    Using cached OCI-SIF image
  .command.sh: line 2: salmon: command not found
```

pditommaso:
Damn, I guess it's ignoring the entry point

pditommaso:
But then, how biocontainers can work?

emi80:
The tools are in the PATH!

```
❯ singularity exec --no-home --oci docker://quay.io/biocontainers/salmon:1.10.2--hecfa306_0 which salmon
INFO:    Using cached OCI-SIF image
/usr/local/bin/salmon
```

dtrudg:
> Damn, I guess it's ignoring the entry point

Hiya - if there's a bug in entrypoint handling, or anything else, please let us know via an issue. We have e2e-tests around that, but perhaps we are missing something.

Much appreciate any feedback / bug reporting on the `--oci` mode. It's a big chunk of new stuff, has a lot of dependency on kernel and runc/crun (with their own fixes etc.) and we didn't have a ton of feedback in the RC period, so I wouldn't expect it to be free of issues.

pditommaso:
Hi Dave, thanks for jumping here. It looks like the entrypoint is not honored. It should be this one. 

https://github.com/mamba-org/micromamba-docker/blob/main/Dockerfile#L62

@emi80 can you please confirm and report an issue in the Singularity repo if confirmed ?

emi80:
Hey @pditommaso @dtrudg,

I just confirmed the problem and will open an issue in the Singularity repo

emi80:
Hey @pditommaso , it looks like this is the expected behavior for `exec`. If we need the entrypoint we should use `run`.

https://github.com/sylabs/singularity/issues/2290#issuecomment-1782553422

pditommaso:
Then we have two options. Doing the same export in the docker build 

https://github.com/seqeralabs/libseqera/blob/2eb6b7cb6925e3077a6805e2383cadb2bca3fa65/wave-utils/src/main/resources/templates/conda/singularityfile-conda-file.txt#L9-L10

or use `singularity run` (already implemented [here](https://github.com/nextflow-io/nextflow/blob/ef6688cdecb1545689f1c5cdc03f5e68ff91d75f/modules/nextflow/src/main/groovy/nextflow/container/SingularityBuilder.groovy#L65-L65))

emi80:
I confirm that `rnaseq-nf` with `singualrity run` with Wave and `conda` strategy works as expected.

What's your opinion @pditommaso? I actually don't really like that we need to use non-default Nextflow behavior to make this work...

pditommaso:
Likely, exporting the Conda root directory is the best solution.

pditommaso:
@emi80 do you think that can have any undesired side effects?

emi80:
Hey @pditommaso, I think that should be fine as it's one of the things the entrypoint is actually doing. I don't see any side effect given that those images will be mainly used to run tools installed in the micromamba enviroment

pditommaso:
Ok, let's go ahead with this. Can you please make this test by adding in your nextflow config this line to make sure it's solving the problem? 

```
wave.build.conda.commands = ['ENV PATH="$MAMBA_ROOT_PREFIX/bin:$PATH"']
```

emi80:
Hey @pditommaso, 
I can confirm that setting PATH with `wave.build.conda.commands` fixes the issue.

@marcodelapierre thanks for taking a look at this!

In case anybody need it, the final config file I used to run `nextflow-io/rnaseq-nf` with wave and Singularity OCI mode is the following:
```
singularity {
   enabled = true
   oci = true
}

wave {
   enabled = true
   strategy = ['conda', 'container']
   build {
      conda {
         commands = ['ENV PATH="$MAMBA_ROOT_PREFIX/bin:$PATH"']
      }
   }
}

process {
   withName:MULTIQC {
      conda = 'multiqc=1.17'
   }
}
```

pditommaso:
Thanks both! 

> To avoid annoying issues with funny point C. above, I would suggest to enforce either A or B, i.e. in the Nextflow config

A. Done, already 
B. You suggest the `oci` option should be considered a 3-state flag: to enable and disable it when specified or to be ignored when omitted. Makes sense

pditommaso:
done

marcodelapierre:
I have just run a local test on a Ubuntu 22.04, Singularity 4.0 box and this branch, using the `nextflow/blast-example` pipeline with a minimally modified config:

```
manifest {
  nextflowVersion = '>= 22.04.0'
}

process {
  container = 'nextflow/examples'
}

singularity.enabled = true
singularity.autoMounts = true

singularity.oci = true
```

The pipeline fails. 

A few notes:
1. the usual warning `WARN: Singularity cache directory has not been defined -- Remote image will be stored in the path: /home/ubuntu/blast-example/work/singularity -- Use the environment variable NXF_SINGULARITY_CACHEDIR to specify a different location` is not printed out
2. singularity correctly outputs `INFO:    Using cached OCI-SIF image`, but then no OCI-SIF file is saved in the Nextflow path above, `/home/ubuntu/blast-example/work/singularity/`
3. The failure is due to a file access issue, `.command.sh: line 2: blast_result: Permission denied`. I can fix it if I add the flag `--no-compat`, which I believe can be simplified, as this is just a meta-flag for a collection of flags (to be checked).

marcodelapierre:
@pditommaso  please have a look at points 1. and 2. in my last comment.

As regards point 3., "compatibility behaviour", here is the doc section on it: https://docs.sylabs.io/guides/latest/user-guide/singularity_and_docker.html#docker-like-compat-flag

My impression is this mostly a behaviour of look-and-feel for the Singularity CLI user. In other words, by default with `--oci` you get all of those Docker-like **isolation from host** behaviours. On the other hand, with `--no-oci` you have the usual Singularity **host integration** behaviours. 

SUGGESTION: 

I am tempted to say .. for our purposes, we can probably implement OCI mode as `--oci --no-compat`, to avoid battling all of those behavioural changes that do not alter the pipeline but overcomplicate the Nextflow internal interface to Singularity. 
Exception to this line of thought: any Singularity defaults that cause troubles to users, and that do not occur with Docker.

Thoughts?

pditommaso:
Regarding this 

> WARN: Singularity cache directory has not been defined 

Tha't reported by the [SingularityCache](https://github.com/nextflow-io/nextflow/blob/f50c89de0dc71a57408eb133c5ff13a4e703b1d8/modules/nextflow/src/main/groovy/nextflow/container/SingularityCache.groovy#L263-L263), but it should not be used because when OCI mode is enabled is skipped here

https://github.com/nextflow-io/nextflow/blob/f50c89de0dc71a57408eb133c5ff13a4e703b1d8/modules/nextflow/src/main/groovy/nextflow/container/ContainerHandler.groovy#L70-L71


Make sure to update your branch and try to debug it still happens

pditommaso:
@marcodelapierre any finding about the warning? otherwise we should assume the test made by @emi80 is fine

marcodelapierre:
> @marcodelapierre any finding about the warning? otherwise we should assume the test made by @emi80 is fine

I will be on this one next Paolo, feedback by tomorrow, apologies for the wait.

marcodelapierre:
I am working on this one, had some blockers on the way, including

https://github.com/nextflow-io/rnaseq-nf/pull/20

marcodelapierre:
Found the root cause of my failing run, in the Nextflow code . 
Will knock it tomorrow morning.

emi80:
Hi @marcodelapierre, thanks a lot for digging into this.

I tried to replicate you run of `nextflow-io/blast-example` with Singularity in OCI mode. As far as I understand, for this specific example, the problem is that the task is not run in the task directory but in user home, since there is not specification of the current working directory neither a `cd`. Adding `--cwd $NXF_TASK_WORKDIR`makes it work properly.

At this point I am not sure whether we want to use the `--no-compat` flag or treat OCI mode as a Docker run and specify the working directory. Any thoughts @pditommaso?

marcodelapierre:
> At this point I am not sure whether we want to use the --no-compat flag or treat OCI mode as a Docker run and specify the working directory. Any thoughts @pditommaso?

Exactly, this is what I was thinking. 

The key advantage of `--no-compat` would be to minimise changes/differences in the codebase between handling OCI and non-OCI Singularity modalities. 

The key limitation I can think of is any practical user cases where the Docker-like behaviour might be preferred .. are we aware of anything like this? Otherwise I would ignore this until we get community feedback on this.

pditommaso:
As a user, I'd expect the container to behave as Singularity, irrespective of the image type. Therefore I think it should default to `--no-compact` mode. 

In any case, I've added also an option `singularity.compact = true|false` to control it

marcodelapierre:
Let's community contribute:

- https://github.com/ContinuumIO/docker-images/pull/444
- https://github.com/mamba-org/micromamba-docker/pull/381
- https://github.com/sylabs/singularity/issues/2305
- https://github.com/sylabs/singularity/pull/2306

marcodelapierre:
Thanks @pditommaso .  Some comments:

- I have renamed all occurrencies of `compact` to `compat` (as in "compatibility")

- in this case, the Singularity vs Docker behaviour is not really exposed directly to users; the vast majority of the impact of using one or the other behaviour is on the Nextflow interface (the key impact that may go to human end users is in the case of errors, thought this might be small in scope); in other words, I would argue the "user" in this context is the Nextflow internal interface. 

- additional argument to the point above: having to maintain the additional combo `--oci --compat` can prove little advantage (why should the user explore this?) and on the other hand require complicating our script templates to maintain it. This PR in its current form had to implement changes, and even then some types of containers wouldn't execute (see the blast_example case).

- small by-product of the current implementation: with `--no-compat` being default, any non OCI container image being run will produce the warning `WARNING: --no-compat applies to --oci mode only, ignoring`. 

My proposition to address the points above:  we don't need a user facing `compat` toggle, and we should think instead of implementing `oci = <true/false>` as adding or not the flag combo `--oci --no-compat`. Simpler implementation in the templates/codebase of Nextflow, easier to maintain, little (no?) impact on the human end user experience.

Thoughts?

marcodelapierre:
@pditommaso  Still actively working on this one - will report on the further progress

marcodelapierre:
In my latest commit:
- simplified other edits by assuming `--no-compat` always on with `--oci` (to be discussed, see above)
- fixed image file not being saved in `work/singularity/`
- correctly saving image file in OCI SIF format

see my comments along the code

marcodelapierre:
My remaining to-do: 
- investigate root cause of mismatching line in `.command.run` for blast-example
- update unit tests
- test with Wave

pditommaso:
-1 for latest changes (apart of course for `compact` -> `compat` renamin). 

I agree to default `--no-compact`, but considering this is a new feature can be useful to keep control over it. 

SingularityCache should be disabled when OCI mode is enabled

marcodelapierre:
> I agree to default --no-compat, but considering this is a new feature can be useful to keep control over it.

then we should put a clear warning on the docs, `--oci --compat` may result in unexpected behaviours and pipeline failure.

marcodelapierre:
Otherwise OK with what you say - see also Sylabs Dave's comments around OCI  mode in Singularity - I did not have a clear enough picture of the scope for OCI mode.

We should aim to have `--oci` to work without `--no-compat` . 

I will work on it, and I think we're on track to finalise this on Monday.

pditommaso:
Ok, now I've realised that --oci implies --compat by default (i had understood the opposite) 

What's the side effect of using the same default when OCI is enabled on our side?

pditommaso:
@marcodelapierre moved your changes to https://github.com/nextflow-io/nextflow/pull/4483.

pditommaso:
I'm testing this branch with classic nf-core/rnaseq + Singularity OCI mode enabled (using defaults) + Wave and it looks working fine. 

Instead, some problems are reported when enabling the Fusion support. I've opened an issue specific for it separately

marcodelapierre:
I am about to do a commit that should fix **some** issues, not sure about yours, hold on

marcodelapierre:
So @pditommaso  @emi80  I confirm that Dave helped me build a clearer and correct picture of OCI mode in Singularity 4: the key aspect is that by definition, OCI mode aims to have Singularity behave as close as possible to Docker.

I only have one outstanding little fix to propose, and will then perform some extra tests.

pditommaso:
Ok, please make a separate PR.

marcodelapierre:
Here we go @pditommaso : https://github.com/nextflow-io/nextflow/pull/4484

marcodelapierre:
> Ok, let's go ahead with this. Can you please make this test by adding in your nextflow config this line to make sure it's solving the problem?

@pditommaso  just saw this bit of thread right now. It is the usual issue with conda/mamba base images not being designed for Singularity. We have gone through this before, and now I actually have 2 PRs opened with Conda and Mamba devs to try and fix these issues (for Mamba: https://github.com/mamba-org/micromamba-docker/pull/381).

Ultimately, it is the very same issue with the rnaseq-nf container: https://github.com/nextflow-io/rnaseq-nf/blob/master/docker/Dockerfile#L13

The same workaround can be applied inside Wave when building the container for the `conda` strategy .

pditommaso:
Indeed, the same patch has been applied to the Wave build for Docker https://github.com/seqeralabs/libseqera/commit/348cf29b9330b3525816b0886e195f9a7a12e62b

marcodelapierre:
.... and https://github.com/nextflow-io/nextflow/pull/4484 also ready for review

modules/nextflow/src/main/groovy/nextflow/container/SingularityBuilder.groovy
```diff
@@ -117,9 +127,15 @@ class SingularityBuilder extends ContainerBuilder<SingularityBuilder> {
...
             result << '--no-home '
 
-        if( newPidNamespace )
+        if( newPidNamespace && !oci )
             result << '--pid '
 
+        if( oci != null )
+            result << (oci ? '--oci ' : '--no-oci ')
+
+        if( oci && compat != null )
```
marcodelapierre, line 136: After conversation with Sylabs folks, suggestion:
```
        if( ! oci && compat )
            result << '--compat '
// unlikely to be useful, possibly just for internal debugging
//        if( oci && ! compat )
//            result << '--no-compat '
```

marcodelapierre (changes requested):
- One change on `compat` option, file modules/nextflow/src/main/groovy/nextflow/container/SingularityBuilder.groovy

- Test files may need updating as a consequence: modules/nextflow/src/test/groovy/nextflow/container/SingularityBuilderTest.groovy , modules/nextflow/src/test/groovy/nextflow/executor/BashWrapperBuilderTest.groovy

Note I would NOT document the `compat` option, should only be for internal debug use

marcodelapierre:
after chat: Let's remove the `compat` option completely

pditommaso:
Done

marcodelapierre (approved):

pditommaso:
For sake of consistency of this PR, I've added also the fix for the singularity `cd` using `$PWD` here, see https://github.com/nextflow-io/nextflow/pull/4440/commits/dbbc8e636dcc8437cbc21d93996b45c66cb75f06. 

Then, we'll merge https://github.com/nextflow-io/nextflow/pull/4484 over master to propagate the use of TASK_WORK_DIR

pditommaso:
ok, let's merge this

PR #4441, opened 2023-10-25
Fix and improve warning about attribute setting in config profiles [ci-skip]

mribeirodantas:


mribeirodantas:
LGTM? 😃

bentsherman:
Thanks Marcel for your contribution. Sorry that it was neglected for so long.

I was always unsure about which cases were actually affected by this bug, so I tested a bunch of combinations and the one added by this PR is the only failing case:

```groovy
profiles {
    foo {
        process.memory = '2 GB'
        process {
            cpus = 2
        }
    }
}
```

```console
$ nextflow config -profile foo
process {
   cpus = 2
}
```

Since we are planning to merge #4744 soon, I will incorporate this example into that PR so that I can recommend the strict config syntax as a solution. Otherwise it just looks like we are documenting a bug.

PR #4443, opened 2023-10-25
Upgrade to Groovy 4

pditommaso:
This PR upgrade the nextflow runtime to Groovy 4


pditommaso:
Some tests fails due this issue https://issues.apache.org/jira/browse/GROOVY-11211

marcodelapierre:
Blocks these PRs:

https://github.com/seqeralabs/libseqera/pull/7

https://github.com/seqeralabs/wave/pull/318#issuecomment-1815749581

bentsherman:
Notes from Groovy 4 [release notes](https://groovy-lang.org/releasenotes/groovy-4.0.html) regarding indy bytecode:

> For many versions, Groovy could generate classic call-site based bytecode or bytecode targeting the JDK7+ invoke dynamic ("indy") bytecode instructions. You could switch between them with a compiler switch and we had two sets of jars ("normal" and "-indy") built with and without the switch enabled. In Groovy 4.0, only bytecode using the latter approach can be generated. There is now one set of jars and they happen to be indy flavored.

> Currently, the Groovy runtime still contains any necessary support for classes compiled using older versions of Groovy. Please use Groovy versions up to 3.x if you need to generate the older style bytecode.

> This work was originally planned for Groovy 3.0, but there were numerous places where "indy" code was noticeably slower than "classic" bytecode. We have made numerous speed improvements (starting with [GROOVY-8298](https://issues.apache.org/jira/browse/GROOVY-8298)) and have some ability to tune internal thresholds (search the code base for groovy.indy.optimize.threshold and groovy.indy.fallback.threshold). That work gave us useful speed improvements, but we welcome further feedback to help improve overall performance of the indy bytecode.

This change is the main reason I was concerned about performance, but it sounds like they have largely addressed this issue. Just looking at the Nextflow CI tests they are about 4:30-5:00 for both this PR and master. So far so good, but I will still try to test rnaseq.

Other notable additions in Groovy 4 are switch expressions and record types. Java's record syntax could be a good pattern to follow in the DSL, especially an internal DSL.

modules/nextflow/src/main/groovy/nextflow/cli/Launcher.groovy
```diff
@@ -209,7 +209,7 @@ class Launcher {
 
             // when the first argument is a file, it's supposed to be a script to be executed
             if( i==1 && !allCommands.find { it.name == current } && new File(current).isFile()  ) {
-                normalized.add(0,CmdRun.NAME)
+                normalized.add(0,CmdRun.NAME.toString())
```
bentsherman, line 212: Why is this needed when CmdRun.NAME is already declared as a String?

pditommaso: Don't remember

bentsherman:
Just ran nf-core/rnaseq small test with groovy 3 and groovy 4 from the CI instance. Both runs took about 30 min. Gonna to a full test next

bentsherman:
nf-core/rnaseq full test completed successfully with comparable runtime. Is there anything else you want to test?

pditommaso:
Let's try also nf-core/fetchngs and Sarek

bentsherman:
sarek full test was successful ✅ 

Note that for all three pipelines I had to use older versions that didn't use nf-validation, This plugin (and likely others) will need to be updated, with the current nf-validation I am getting this error:
```
groovy.lang.MissingMethodException: No signature of method: nextflow.processor.TaskHandler.getTraceRecord() is applicable for argument types: () values: []
Possible solutions: getTraceRecord(), safeTraceRecord()
        at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.unwrap(ScriptBytecodeAdapter.java:72)
        at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.invokeMethodOnSuperN(ScriptBytecodeAdapter.java:148)
        at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.invokeMethodOnSuper0(ScriptBytecodeAdapter.java:166)
        at nextflow.cloud.aws.batch.AwsBatchTaskHandler.getTraceRecord(AwsBatchTaskHandler.groovy:836)
        at nextflow.Session.notifyTaskPending(Session.groovy:980)
        at nextflow.processor.TaskPollingMonitor.schedule(TaskPollingMonitor.groovy:229)
        at nextflow.executor.Executor.submit(Executor.groovy:98)
        at nextflow.processor.TaskProcessor.submitTask(TaskProcessor.groovy:2250)
        at nextflow.processor.TaskProcessor.checkCachedOrLaunchTask(TaskProcessor.groovy:823)
        at nextflow.processor.TaskProcessor.invokeTask(TaskProcessor.groovy:639)
        at nextflow.processor.InvokeTaskAdapter.call(InvokeTaskAdapter.groovy:52)
        at groovyx.gpars.dataflow.operator.DataflowOperatorActor.startTask(DataflowOperatorActor.java:120)
        at groovyx.gpars.dataflow.operator.ForkingDataflowOperatorActor.access$001(ForkingDataflowOperatorActor.java:35)
        at groovyx.gpars.dataflow.operator.ForkingDataflowOperatorActor$1.run(ForkingDataflowOperatorActor.java:58)
        at java.base/java.util.concurrent.ThreadPerTaskExecutor$TaskRunner.run(ThreadPerTaskExecutor.java:314)
        at java.base/java.lang.Thread.run(Thread.java:1583)
```

fetchngs could not be tested because the test profiles for older versions no longer seem to work

bentsherman:
Investigating further, it looks like this exception happens whenever I install the groovy4 build using `make install`, regardless of the pipeline. When I run with `launch.sh` this error doesn't happen. So it must be an error with the PR itself. The stack trace is coming from the awsbatch executor, but I'm not sure why this error would happen, since the nf-amazon plugin is built alongside nextflow, so it should be using the Nextflow API built with groovy4.

bentsherman (approved):

pditommaso:
We are ready for this!

bentsherman:
Here we go 😅

PR #4458, opened 2023-10-30
Allow error() to be used in onComplete handler

### LINKED ISSUE #4397: Exceptions and `error()` are unusable in onComplete handler
Slack thread: https://nextflow.slack.com/archives/C02T98A23U7/p1696364290193959

Exceptions can't be thrown from the workflow onComplete handler because they are immediately caught by Nextflow and logged as an error:

https://github.com/nextflow-io/nextflow/blob/dfb5bf092066ea5ac81f3c17ed14e1a42a126366/modules/nextflow/src/main/groovy/nextflow/script/WorkflowMetadata.groovy#L391-L398

This also affects the `error()` function, which throws a `WorkflowScriptErrorException` under the hood. Throwing an exception is a useful way to fail the entire run if some final validation does not succeed.

On the other hand, if some unexpected exception is thrown, the current logic prevents the entire run from failing, which can be nice.

A good solution might be to re-throw `WorkflowScriptErrorException` but catch and log all others.

bentsherman:
Close #4397

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/WorkflowMetadata.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/WorkflowMetadata.groovy
@@ -29,6 +29,7 @@ import nextflow.NextflowMeta
 import nextflow.Session
 import nextflow.config.ConfigBuilder
 import nextflow.config.Manifest
+import nextflow.exception.WorkflowScriptErrorException
 import nextflow.trace.WorkflowStats
 import nextflow.util.Duration
 import org.codehaus.groovy.runtime.InvokerHelper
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/script/WorkflowMetadata.groovy
```diff
@@ -392,6 +393,9 @@ class WorkflowMetadata {
             try {
                 action.call()
             }
+            catch (WorkflowScriptErrorException e) {
+                throw e
```
pditommaso, line 397:
```suggestion
                // re-throw it to allow `error` function to be invoked by compleation handler
                throw e
```

pditommaso (approved):

modules/nextflow/src/main/groovy/nextflow/script/WorkflowMetadata.groovy
```diff
@@ -392,6 +393,10 @@ class WorkflowMetadata {
             try {
                 action.call()
             }
+            catch (WorkflowScriptErrorException e) {
+                // re-throw it to allow `error` function to be invoked by compleation handler
```
pditommaso, line 397:
```suggestion
                // re-throw it to allow `error` function to be invoked by completion handler
```

PR #4459, opened 2023-10-30
Add "topic" channel 

pditommaso:
This PR implements a draft implementation for topic channel described [here](https://github.com/nextflow-io/nextflow/discussions/2842)


marcodelapierre:
Also see this PR by Jordi: https://github.com/nextflow-io/nextflow/pull/4425

bentsherman:
I tested with nf-core/rnaseq, PR is here: https://github.com/nf-core/rnaseq/pull/1109

The topic channel worked perfectly, but there is a downstream error with multiqc that needs to be resolved, I think it is pipeline-specific rather than a bug with the topic channel.

I will add an integration test and some docs

modules/nextflow/src/main/groovy/nextflow/extension/CH.groovy
```diff
@@ -30,7 +30,15 @@ class CH {
...
     }
 
-    static private Map<DataflowQueue, DataflowBroadcast> bridges = new HashMap<>(10)
+    static class Topic {
+        String name
+        DataflowBroadcast broadcaster = new DataflowBroadcast()
+        List<DataflowWriteChannel> writers = new ArrayList<>(10)
+    }
+
+    static final private List<Topic> allTopics = new ArrayList<>(10)
```
bentsherman, line 39: Would it not be safer/clearer to model this as a `Map<String,Topic>` ?

pditommaso: yeah

docs/channel.md
```diff
@@ -424,6 +424,47 @@ Y
...
+### topic
+
+:::{versionadded} 23.11.0-edge
+:::
+
+:::{note}
+This feature requires the `nextflow.preview.topic` feature flag to be enabled.
+:::
+
+The `topic` method is used to create a "topic" channel, which is a queue channel that can receive items from multiple sources.
```
bentsherman, line 438: Maybe you want to describe "topic" channels as a new channel type alongside queue and value channels. Since a topic channel seems to behave like a queue channel, I described it in this way

marcodelapierre:
Question on the naming, @pditommaso  @bentsherman : should we pick a different name, something like the `metadata` channel?

To me, `topic` sounds a bit generic jargon, not too specific to the workflows/pipelines/data/computing world.

docs/channel.md
pditommaso: Good point. Surely it deserves to be expanded. Another possibility could be introduce the "broadcast" channel type, along "queue" and "value". A broadcast channel can even many writers and many readers (opposed to a queue channel than can have exactly one write and reader) and it's identified by a "topic" name.

marcodelapierre: I like "broadcast" more than "topic", sounds to me like more appropriate jargon in the scope of Nextflow

bentsherman: Broadcast is not specific enough IMO because a value channel is also a broadcast (it can have many readers). Even the underlying GPars class is `DataflowBroadcast`. The topic channel is distinct because it can have many *writers*, but I don't know of any special term for that. I will see if I can find something from stream processing or digital circuits terminology...

bentsherman: I didn't find anything beyond the [event bus](https://dzone.com/articles/design-patterns-event-bus) pattern. On further reflection, I don't think we need to distinguish topic channels as a special channel type. A topic channel is just shorthand for a mix operation:

```groovy
ch_foo = mix(foo1, foo2, foo3, foo4, foo5)
```

It's just a bunch of queue channels, and the "topic" is what brings them together.

I'm open to other words than topic if we can find a better one. Some alternatives include "category", "label", "tag"... but label and tag are already concepts in Nextflow and category is too broad IMO. "metadata" implies that the topic channel can only be used to collect metadata, but that need not be the case. I like the idea of describing the topic channel as an event bus, but any channel could be called a "bus".

How about... [`Channel.mixer()`](https://en.wikipedia.org/wiki/DJ_mixer)

pditommaso: I think you were right it should be introduced a new channel definition because in several places it's said that a "queue" channel only allows one writer and one reader, therefore this is an important change in the nextflow dataflow model. 

I'd keep aligned with gpars definition: 
* DataflowQueue -> queue channel 
* DataflowValue -> value channel 
* DataflowBroadcast -> broadcast channel 

And a "broadcast" channel can be identified by "topic" name. The topic is common in event based systems, and [Kafka made](https://dzone.com/refcardz/apache-kafka-patterns-and-anti-patterns) it popular

bentsherman: I think we are misunderstanding each other. Broadcast is *not* a good term to use because it only means multiple readers, not necessarily multiple writers. I'm talking about the meaning of the word, not the semantics of `DataflowBroadcast`.

Also, Nextflow uses `DataflowBroadcast` under the hood to implicitly fork a queue channel when it is read multiple times in a pipeline, so there isn't really a one-to-one mapping between GPars and Nextflow channel types. In reality, both queue and value channels can have multiple readers in Nextflow, so they can all be "broadcast".

There isn't really a term for a channel that can have multiple sources, I guess because there are different terms for different kinds of aggregation (mix, collect, gather, concatenate, sum, etc). Since we are simply "mixing" the source channels (as in the `mix` operator), the best term I can think of is a "mixer" channel. So I recommend we declare a new "mixer" type or just don't call it a new type at all.

pditommaso: I see better your point, however, I still claim "broadcast" fits better here. It's true that's not perfectly aligned with the real semantics but at least it's a term more common in the message-oriented systems. 

Also if you notice gpars docs, they use DataflowQueue both for one-to-many and many-to-many 

http://gpars.org/1.2.1/javadoc/index.html?groovyx/gpars/dataflow/DataflowBroadcast.html

pditommaso: > Broadcast tells you nothing about the number of writers

I know, but a simple word cannot capture the semantics. The important thing is that it's evocative enough to give an idea of the different behaviour. The docs will give a formal definition of the semantics and how it can be used

bentsherman: Paolo, that does not make any sense 😆 

Besides, we can do better than merely "evocative" -- we can actually use a correct word! That word is "mixer"

pditommaso: To me "mixer" make no sense 🤷‍♂️, and the consistency with gpars terminology it's an important +1 for me.

marcodelapierre:
Just adding some references:

docs/channel.md
pditommaso: Under the hood is a composition of `mix` operator, however the logic model is distinct

bentsherman: I think it's inconsistent of you to treat the alignment with DataflowBroadcast as such a benefit while brushing aside the mix operator as merely an implementation detail. But I will put that aside because I don't like "mixer" for other reasons. My actual preference is to not declare a new channel type at all.

I think I would rather just call it a "topic" channel. The multiple writers and "broadcast" are not the important aspect -- any channel can be broadcast in Nextflow, and operators can have "multiple writers" -- but the implicit aggregation of inputs based on a _topic_. And as you have said, "topic" is a well-established term from the domain of stream processing.

Look over the docs I have written. I have described what a topic channel is and how it is used in a pipeline. Do you really think these docs will be improved by adding another term aside from "topic", especially one that is _not correct_ ?

modules/nextflow/src/main/groovy/nextflow/extension/CH.groovy
```diff
@@ -30,7 +30,14 @@ class CH {
...
 
-    static private Map<DataflowQueue, DataflowBroadcast> bridges = new HashMap<>(10)
+    static class Topic {
+        List<DataflowWriteChannel> sources = new ArrayList<>(10)
+        DataflowBroadcast target = new DataflowBroadcast()
+    }
+
+    static final private Map<String, Topic> allTopics = new HashMap<>(10)
+
+    static final private Map<DataflowQueue, DataflowBroadcast> bridges = new HashMap<>(10)
```
bentsherman, line 40: I've been meaning to ask, what is the purpose of this bridges map? My guess is that it is used to enable queue channels to be re-used without the old `into` operator, is that right?

docs/channel.md
marcodelapierre: Just read the thread. 

I am not super convinced of "topic", but I cannot think of a better one neither. At first I was liking "broadcast", but Ben's comments around inconsistencies that can turn confusing to users/devs is fair. 

However, I do think it should be advertised as a new channel type, at the same level of value and queue. The distinctive trait is indeed the ability to collect input from multiple processes. (this is opposite to considering it a specialisation of the queue type)

So, I would go for the `topic` channel, which is a new third type of channel.

docs/channel.md
```diff
@@ -424,6 +424,47 @@ Y
...
+### topic
+
+:::{versionadded} 23.11.0-edge
+:::
+
+:::{note}
+This feature requires the `nextflow.preview.topic` feature flag to be enabled.
+:::
+
+The `topic` method is used to create a "topic" channel, which is a queue channel that can receive items from many sources implicitly based on a "topic" identifier.
```
marcodelapierre, line 438: As mentioned in the main comments thread, I think that `topic` should be advertised as a new channel type, at the same level of `value` and `queue`. 

The distinctive trait is indeed the ability to collect input from multiple processes. (this is opposite to considering it a specialisation of the `queue` type)

docs/channel.md
```diff
@@ -424,6 +424,47 @@ Y
...
+
+Any process can send items to a topic by using the `topic` option on an output:
+
+```groovy
+process foo {
+  output:
+  val('foo'), topic: 'my-topic'
```
marcodelapierre, line 445: Let me be provocative here, keen on your views @bentsherman @pditommaso and probably also @ewels .

This currently proposed syntax does not convince me vs DSL2:
```
output:
val('foo'), topic: 'my-topic'
```
It looks to me like a silent partial going back to DSL1 style.

Instead DSL2 specifies channel connections between processes within the `workflow` block, not the `process` block, and hence I think the main legal syntax should be along the lines:
```
output:
val('foo'), emit: 'foo_val'

[..]

workflow {
    foo.foo_val.topic('my-topic')
}
```

And then, eventually (but not necessarily) also allow the following, more DSL1 styled variant (where the `emit` attribute is not needed if no other process requires this as input):
```
output:
val('foo'), emit: 'foo_val', topic: 'my-topic'
```

marcodelapierre (changes requested):

docs/channel.md
marcodelapierre: This sort of DSL1-esque inconsistency became apparent to me while looking at Ben's test implementation in rnaseq:  https://github.com/nf-core/rnaseq/pull/1109#issuecomment-1808450986

Why that outlier syntax line?

marcodelapierre (changes requested):

docs/channel.md
bentsherman: That is an interesting approach, specifying the topic with the process invocation instead of the definition. It would keep the process def modular and be more flexible -- you could send any process output to any topics you want without patching the module, and you could use emits and topics together.

The only downside is that it's more verbose, although still simpler than the current approach used by nf-core

I think I am keen to try it, but curious to see what others think..

pditommaso: > It looks to me like a silent partial going back to DSL1 style.

Possibly, but this is the key point because it decouples the topic from the channel compositions. 

Having a convention that all processes in, for example, nf-core, use `versions` to track the tool version, it's enough to add a new module and the receiver will get the module tool version without doing anything else

modules/nextflow/src/main/groovy/nextflow/extension/CH.groovy
bentsherman: If my guess is correct, it would illustrate my point that the topic channel is no more a "broadcast" than queue channels, in the sense that they both use `DataflowBroadcast` primarily to support multiple readers. Meanwhile, the topic channel uses `MixOp` to receive from multiple sources

docs/channel.md
bentsherman: To me, the key distinguishing feature of the topic channel is not the multiple sources, but the ability to collect those sources *implicitly* via a matching name which is the "topic".

I don't see it as a new channel type, but an operation that produces a queue channel. It is essentially an implicit "mix" operation, but we don't create a new channel type for every operator like mix, collect, concat, etc, we just call them channel operators. Or consider the `Channel.watchPath()` factory, which emits file events for a path or glob -- it is doing a special kind of operation, but ultimately produces a queue channel. So I think the `topic` factory should be understood as an operation on queue channels.

Maybe we should call it `Channel.fromTopic` to be consistent with `fromPath`, `fromList`, etc

marcodelapierre: Thanks for this perspective Ben, I get your point, it makes sense to me now: in the end the multiple writers end up pouring their output in waht is a single queue - the description you made of implicit multiple mixes got me.

@pditommaso ?

docs/channel.md
marcodelapierre: See discussion above

docs/channel.md
marcodelapierre: I love `fromTopic`, will comment further down

docs/channel.md
marcodelapierre: Thanks for the feedbacks. I get your main point about conciseness. Indeed I see how my proposition results in a verbosity that scales with the number of writers - vs your approach, that just adds one module.

1. To bring things together, I am citing @bentsherman 's suggestion above of having `Channel.fromTopic()`. I think this has the potential to ultimately define the new functionality not as a new channel type, but indeed as a new way to define a queue channel.  You **label** outputs by topic, and then you use the new `fromTopic` to create a channel out of all those outputs with that given label. This way I think it essentially clarifies that, as Ben's says, this is a way of implicitly composing multiple `mix`es.

2. As a second consequence of this picture, I would then see the topic specifier in the process definition as a way to apply the *topic* label (as defined above) to the output:
```
output:
val('foo'), topic: 'my-topic'
```
hence, I would still suggest to consider such `topic` as an additional optional attribute to the output, in addition/alternative to the `emit` specifier, rather than in strict replacement for it. This would allow cases where the output has to be consumed also through non-topic channels. 

(as a final note, both the definition of `fromTopic`, and the definition of `topic` as an additional attribute aside with `emit` seem to me more consistent with DSL2 compared to the starting point - from my point of view)

pditommaso: I'm fine to stay with "topic" channel definition, but they should be documented as a new channel type because they have different semantics (many writers). 

This should it should be `channel.topic`, not `fromTopic`

bentsherman: > Having a convention that all processes in, for example, nf-core, use versions to track the tool version, it's enough to add a new module and the receiver will get the module tool version without doing anything else

I agree with Paolo on this point. I think it is better to use the current approach which is the least verbose and has a narrow use case, i.e. for clear conventions like nf-core tool versions. There are other use cases like collecting logs for multiqc which would require more flexibility to be a channel topic, but at that point I think you might as well just use explicit channel logic.

docs/channel.md
bentsherman: Quoting Paolo's comment which got mixed up:

> I'm fine to stay with "topic" channel definition, but they should be documented as a new channel type because they have different semantics (many writers).
>
> Thus it should be `channel.topic`, not `fromTopic`

Thinking more on this, I still think we should call it `Channel.fromTopic` and describe it as an operation that creates a queue channel rather than a new channel type. Marco, Phil, and I seem to be in agreement on this, but since this thread has meandered and the meeting didn't reach a solid conclusion, here is my argument put concisely:

A channel topic is literally an operation on queue channels. There is no new channel type under the hood, just queue channels coming in and a queue channel going out. It uses the DataflowBroadcast only to support multiple readers (in the same way as queue channels) and it uses the *mix operator* to support the multiple writers. In fact many operators support multiple writers, so that alone is not enough to warrant a new channel type. The implicit linking via topic is more unique, but when I tried to document it as a new channel type, I just found it unnecessary and more confusing.

docs/channel.md
marcodelapierre: yep, agree on maintaining the current approach.

I would still suggest to consider such `topic` as an additional optional attribute to the output, in **addition/alternative** to the `emit` specifier, rather than in strict replacement for it. This would allow cases where the output has to be consumed also through non-topic channels.

docs/channel.md
marcodelapierre: > it uses the mix operator to support the multiple writers. In fact many operators support multiple writers, so that alone is not enough to warrant a new channel type.

👍

docs/channel.md
bentsherman: Actually `emit` and `topic` already can be used together, I must have misread something. I will update the docs and integration test

marcodelapierre (commented):
My only outstanding points at this stage are in line with Ben: 
- considering the topic a method to create a queue channel as opposed to a new type of channel
- `fromTopic` as opposed to `topic

marcodelapierre (commented):
My only outstanding points at this stage are in line with Ben (rationale discussed between us in comments above):

- considering the topic a method to create a queue channel as opposed to a new type of channel

- `fromTopic` as opposed to `topic`

docs/channel.md
```diff
@@ -424,6 +424,47 @@ Y
...
+Then, the `topic` method can be used to consume all items in the topic:
+
+```groovy
+Channel.topic('my-topic').view()
+```
+
+This approach is a convenient way to collect related items from many different sources without explicitly defining the channel logic (the `topic` method is essentially an implicit {ref}`operator-mix` operation). You can name topics however you want, and you can use different names to collect items for different "topics", as long as your process outputs and channel logic are consistent with each other.
+
+:::{warning}
+Any process that consumes a channel topic should not send any outputs to that topic, or else the pipeline will hang forever.
```
pditommaso, line 463: What does it mean? A circular dependency?

bentsherman: yes

pditommaso: it should be possible to detect the condition and report an error

bentsherman: I guess, if a process has a topic output, you could walk back up the DAG and check for topic factory calls. Because there could be operators between the topic factory and process. Can you think of anything simpler?

pditommaso: Not sure sure, but while composing the dag or in the "normalization" phase it should be possible to identify it 

https://github.com/nextflow-io/nextflow/blob/topic-channel/modules/nextflow/src/main/groovy/nextflow/dag/DAG.groovy#L339-L340

marcodelapierre (approved):

docs/channel.md
pditommaso: I believe you need to see things from the proper perspective. The most important thing is that the topic channel introduces a different way to compose and think Nextflow channels. 

Instead of having one-to-one, producer-to-consumer messaging, the topic allows many producers to send messages over the same topic to many consumers. 

It doesn't matter it could have been implemented using a composition of `mix` operators or how it's implemented under the door. The topic type is important to highlight the different paradigm that is introduced by this feature. 

I've made a few changes in the docs to reflect this view. In any case, this is marked as experimental, we can always review and changes along the way in future releases.

ewels:
Awesome, very happy to see this merged. Thanks all! 🎉

marcodelapierre:
image

docs/channel.md
marcodelapierre: > the topic channel introduces a different way to compose and think Nextflow channels.
> The topic type is important to highlight the different paradigm that is introduced by this feature.

I had this chat with Paolo as well. I now better get the point around proposing a new paradigm, that unlocks new ways for devs to describe their pipelines. 

Overall it was a good discussion, and most importantly it is good to have it as `experimental`, to leave room for upgrades in case we identify the need for them.

marcodelapierre:
Post-merge, post-weekend thought specifically around the naming.

I keep thinking that the `topic` misses something compared to typical software devs jargon. 

Hence, I was wondering whether `forum` might be more appropriate.  
The **forum** channel enables multiple writers to collect information around a topic, that is reflected in the channel name itself.


@pditommaso @bentsherman ?  
Paolo, this is one for you to think about during one of those yummy Boston breakfasts!!

docs/channel.md
bentsherman: I do see the other perspective Paolo, after all I started this whole thread, but my thoughts evolved and I no longer think it's necessary to describe channel topics as a new type. Although it doesn't always happen this way, in this case I think the implementation details are quite instructive in how to describe it. If someone figures out to use a channel topic in some way other than an implicit mix, I might be convinced otherwise.

But I'm glad you went ahead and merged it. Better to get the feature out there for users to play with it. We can refine the docs as needed.

ewels:
Hmm, don't love `forum` sorry. That feels less intuitive to me, I quite liked `topic` personally.

marcodelapierre:
Fair enough, thanks for the feedback Phil 😊

PR #4485, opened 2023-11-06
Add support for K8s pod schedulerName

rgo3:
For my master thesis I needed a way to specify which scheduler is used to schedule nextflow tasks. Maybe it could be useful for others as well. This adds a `schedulerName` field to `PodOptions` and a shortcut to it via `K8sConfig`.

K8s docs reference: https://kubernetes.io/docs/tasks/extend-kubernetes/configure-multiple-schedulers/#specify-schedulers-for-pods

Given I already had the code, I opted to create a PR directly instead of creating an issue first. If this is not needed or already worked on, feel free to close :)

### DIFF
--- a/docs/process.md
+++ b/docs/process.md
@@ -2075,6 +2075,9 @@ The following options are available:
 `runAsUser: '<uid>'`
 : Specifies the user ID with which to run the container. Shortcut for the `securityContext` option.
 
+`schedulerName: '<name>'`
+: Specifies which [scheduler](https://kubernetes.io/docs/tasks/extend-kubernetes/configure-multiple-schedulers/#specify-schedulers-for-pods) is used to schedule the container. 
+
 `secret: '<secret>/<key>', mountPath: '</absolute/path>'`
 : *Can be specified multiple times*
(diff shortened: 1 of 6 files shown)


bentsherman (approved):

docs/process.md
```diff
@@ -2075,6 +2075,9 @@ The following options are available:
 `runAsUser: '<uid>'`
 : Specifies the user ID with which to run the container. Shortcut for the `securityContext` option.
 
+`schedulerName: '<name>'`
+: Specifies which [scheduler](https://kubernetes.io/docs/tasks/extend-kubernetes/configure-multiple-schedulers/#specify-schedulers-for-pods) is used to schedule the container.
```
pditommaso, line 2079: Which K8s version was this feature introduced? (worth to mention)

bentsherman: It is quite old, at least as old as many other options for which we don't mention the minimum version

pditommaso: Ok, was not aware of that. then we can assume it's ok

pditommaso (approved):

PR #4488, opened 2023-11-07
Update logging filter for Google Batch provider

zetxqx:
Batch used to use generic_task as monitored resource type. However, the type is not only for Batch. Google Batch is planning to switch it's own monitored resource type which is "batch.googleapis.com/Job". To have a smooth changing, add a OR filter to have the new monitored resource type included. After we finished transition, the generic_task filter could be removed.

### DIFF
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/logging/BatchLogging.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/logging/BatchLogging.groovy
@@ -70,7 +70,7 @@ class BatchLogging implements Closeable {
         final stdout = new StringBuilder()
         final stderr = new StringBuilder()
         // use logging here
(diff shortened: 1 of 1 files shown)


bentsherman:
Thanks @zetxqx , is this change documented anywhere? Or if you are from the Google Batch team I'll just take your word for it 😉

zetxqx:
Hi @bentsherman, yes I'm from the Google Batch team. Confirmed, we are planning to use the new monitored resource. Sorry I don't have a public document right now. But our [reference](https://cloud.google.com/batch/docs/reference/rest/v1alpha/projects.locations.jobs#logspolicy) page will be soon published.

In the PR, I added the OR logging filter. After we finish the migration, I will remove the old "generic_task" filter as well in another PR. Thanks!

bentsherman (approved):

PR #4493, opened 2023-11-08
Add cmd output type

pditommaso:
Implements the support for `cmd` output type to collect task tool version and metadata as described here https://github.com/nextflow-io/nextflow/issues/4386#issuecomment-1788614325

POC usage:

```
process foo {
  output:
  cmd 'git --version', emit: gitVersion
  
  '''
  echo Ciao
  '''
}

workflow {
  foo()
  foo.out.gitVersion | view
}
```


bentsherman:
I kinda like the idea of overloading `stdout` for this purpose, since it can only be the stdout of the command. Otherwise we should use the full word `command`

pditommaso:
I wanted to keep simple, at least as POC

pditommaso:
Next steps:
- [ ] Handle command failure (how?) 
- [ ] Handle multiline command output (likely currently breaks)
- [ ] Add + fix tests

bentsherman:
Is this ready to test with rnaseq? Even if it only supports single-line bash commands, should be enough.

pditommaso:
it should

bentsherman:
It doesn't seem to work with tuples, gonna need that for something like:
```groovy
tuple val('gunzip'), cmd("gunzip --version")
```

bentsherman:
I will add it

marcodelapierre:
> I kinda like the idea of overloading `stdout` for this purpose, since it can only be the stdout of the command. Otherwise we should use the full word `command`

@pditommaso  sounds tempting to me, too, on this same ground by Ben.

marcodelapierre:
I think this comment is interesting:  https://github.com/nextflow-io/nextflow/issues/4386#issuecomment-1802790996
Also, it relates to your point above Paolo: `Only works for Bash based tasks`

If multi-line is allowed, then we might leverage the *shebang* to allow non-bash executions, including other shells, Groovy, popular scripting langs such as Python .. a bit like the current scoping for the `script` block , really.

Thoughts on this point?

marcodelapierre:
> Handle command failure (how?)

I would say, if the command fail, the task should fail. I.e. consistent with the behaviour as for the main `script` block.

pditommaso:
> If multi-line is allowed, then we might leverage the shebang to allow non-bash executions

not sure it's possible

bentsherman:
> I would say, if the command fail, the task should fail. I.e. consistent with the behaviour as for the main script block.

On this point, I think I suggested that the user should be able to ignore the failure by specifying `optional: true`, although it would be helpful to still log a warning in that case

> If multi-line is allowed, then we might leverage the shebang to allow non-bash executions

Due to the compact nature of process inputs and outputs, it would be unwieldy to specify a multi-line script here. Instead, it should already be possible to reference a script from the `bin` directory in the cmd output. That would be useful anyway because, in rnaseq for example, many processes use the same tools and so the same version commands are duplicated.

Regarding the shebang, we could add a `shell` option to the command output and still support one-liners:
```groovy
output:
cmd("print('Hello!')", shell: 'python')
```

modules/nextflow/src/main/groovy/nextflow/script/params/TupleInParam.groovy
```diff
@@ -79,6 +80,9 @@ class TupleInParam extends BaseInParam {
             else if( item instanceof TokenStdinCall ) {
                 newItem(StdInParam)
             }
+            else if( item instanceof TokenCmdCall ) {
+                throw new IllegalArgumentException('Command input declaration is not supported')
+            }
```
bentsherman, line 85: I keep wondering what it would mean to have a "command input"... here are some ideas

If we ever decide to do more fine-grained resource optimization based on task inputs, I would like to be able to specify custom metrics based on the input files. We can already get the file size, but file size isn't always the most salient metric for resource usage. See [here](https://github.com/SystemsGenetics/KINC-nf/blob/77545302244c50278b322c8150ca68c6f4a14ee9/main.nf#L179-L181) how I specified custom metrics for my PhD research.

Perhaps a "command input" could serve this purpose, but the challenge is that it needs to run before the task is launched, if it is to be used to influence process directives like cpus, memory, etc

Here is a sketch based on the example I linked:
```groovy
def predict_memory(n_rows, n_cols) {
  // query prediction model
}

process IMPORT_EMX {
  memory { predict_memory(n_rows, n_cols) }

  input:
  path 'input.txt'
  cmd 'tail -n +1 input.txt | wc -l', name: 'n_rows'
  cmd 'head -n +1 input.txt | wc -w', name: 'n_cols'

  // ...
}
```

This syntax is confusing because the command inputs aren't really process inputs, i.e. the user doesn't need to supply input channels for them. Seems like the command input should really just be a function that has access to the task inputs:

```groovy
process IMPORT_EMX {
  memory {
    def n_rows = cmd('tail -n +1 input.txt | wc -l')
    def n_cols = cmd('head -n +1 input.txt | wc -w')
    predict_memory(n_rows, n_cols)
  }

  input:
  path 'input.txt'

  // ...
}
```

This is still dubious because directives are evaluated before the task environment is created and input files staged. We might just have to limit ourselves to task metadata:

```groovy
  memory {
    predict_memory( task.inputs['input.txt'].size() )
  }
```

This is all outside the scope of this PR, just leaving my notes here for future reference.

bentsherman:
> > I kinda like the idea of overloading `stdout` for this purpose, since it can only be the stdout of the command. Otherwise we should use the full word `command`
> 
> @pditommaso sounds tempting to me, too, on this same ground by Ben.

I'm starting to lean back towards a separate `command` or `cmd` type... , this command output feels different enough from stdout that it should have a different name, and it wouldn't make sense to extend `stdin` in the same way

modules/nextflow/src/main/groovy/nextflow/script/params/TupleInParam.groovy
marcodelapierre: Shall you fork out a new issue out of this interesting note?! :-)

marcodelapierre:
> > > I kinda like the idea of overloading `stdout` for this purpose, since it can only be the stdout of the command. Otherwise we should use the full word `command`
> > 
> > 
> > @pditommaso sounds tempting to me, too, on this same ground by Ben.
> 
> I'm starting to lean back towards a separate `command` or `cmd` type... , this command output feels different enough from stdout that it should have a different name, and it wouldn't make sense to extend `stdin` in the same way


I agree on the leaning back, based on the `optional` and `script` attributes you mentioned above, https://github.com/nextflow-io/nextflow/pull/4493#issuecomment-1816692368,  both of which make sense to me.

modules/nextflow/src/main/groovy/nextflow/script/params/TupleInParam.groovy
bentsherman: See #4545

bentsherman:
> Handle multiline command output (likely currently breaks)

So the multi-line output does not fail, but the newlines are converted to spaces:
```console
$ echo nxf_out_cmd_1=$(bash --version) > .command.env
$ cat .command.env
nxf_out_cmd_1=GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu) Copyright (C) 2020 Free Software Foundation, Inc. License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html> This is free software; you are free to change and redistribute it. There is NO WARRANTY, to the extent permitted by law.
```

The command itself is easily fixed with some quotes:
```console
$ echo nxf_out_cmd_1="$(bash --version)" > .command.env
$ cat .command.env 
nxf_out_cmd_1=GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

But that would break the `.command.env` file which expects a key-value pair on each line.

The same limitation already exists for `env` outputs, e.g. if I do `FOO=$(bash --version)` in the process script then the process output will be squeezed onto a single line. For this reason, and the fact that `cmd` outputs are intended for small things like metadata, while larger outputs can always be written to a file, I am fine with leaving the current behavior as it is.

pditommaso:
A single line is a too weak assumption. If you notice, nf-core uses a multi-lines YAML snippet! 

https://github.com/nf-core/rnaseq/blob/a10f41afa204538d5dcc89a5910c299d68f94f41/modules/nf-core/salmon/index/main.nf#L42-L45

bentsherman:
That is only because they are eagerly formatting it to YAML. Instead in my proof-of-concept PR I use tuples and separate `cmd` outputs:

https://github.com/nf-core/rnaseq/pull/1115/files#diff-d81c6cc6d9c866f8162450ce077b2413b28ecd47881a70a5c24a1ec1f1b70205

pditommaso:
and where the formatting is done?

bentsherman:
It can be done when all the tool versions are collected: https://github.com/nf-core/rnaseq/pull/1115/files#diff-8925ff36f26422427d3219852d89d7d83bd417fcb5aa37a48f6421bc08b94edd

modules/nextflow/src/test/groovy/nextflow/script/params/CmdOutParamTest.groovy
```diff
@@ -0,0 +1,59 @@
...
+        def outs = process.config.getOutputs() as List<CmdOutParam>
+
+        then:
+        outs.size() == 2
+        and:
+        outs[0].name == 'nxf_out_cmd_1'
+        outs[0].target == 'foo --version'
+        and:
+        outs[1].name == 'nxf_out_cmd_2'
+        outs[1].target == 'bar --help'
```
bentsherman, line 55: This test fails now because I added a test to `TaskRunTest` which affected the global counter in `CmdOutParam` 🤦

bentsherman:
This `cmd` output is basically a small syntax sugar over an `env` output, and I'm starting to wonder whether it's worth adding.

I will sketch out a process with (1) the current nf-core convention, (2) an env output, (3) a cmd output, to show the differences in syntax.

```groovy
// current nf-core convention
process FOOBAR {
    output:
    path 'versions.yml', topic: versions

    """
    # ...

    cat <<-END_VERSIONS > versions.yml
    "${task.process}":
        foo: \$(foo --version)
        bar: \$(bar --version)
    END_VERSIONS
    """
}

// env output
process FOOBAR {
    output:
    tuple val("${task.process}"), val('foo'), env(FOO_VERSION), topic: versions
    tuple val("${task.process}"), val('bar'), env(BAR_VERSION), topic: versions

    """
    # ...

    FOO_VERSION=\$(foo --version)
    BAR_VERSION=\$(bar --version)
    """
}

// cmd output
process FOOBAR {
    output:
    tuple val("${task.process}"), val('foo'), cmd('foo --version'), topic: versions
    tuple val("${task.process}"), val('bar'), cmd('bar --version'), topic: versions

    """
    # ...
    """
}
```

The `cmd` approach is slightly less verbose and has better locality of behavior, but the `env` approach is clearer about how the version commands are executed, and it's more flexible because it's just Bash. In other words, `cmd` provides a restricted wrapper around Bash whereas `env` just exposes the Bash interface directly. So maybe the `cmd` output is not worth the added complexity

marcodelapierre:
> The command itself is easily fixed with some quotes:
> 
> ```
> $ echo nxf_out_cmd_1="$(bash --version)" > .command.env
> $ cat .command.env 
> nxf_out_cmd_1=GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)
> Copyright (C) 2020 Free Software Foundation, Inc.
> License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
> 
> This is free software; you are free to change and redistribute it.
> There is NO WARRANTY, to the extent permitted by law.
> ```
> 
> But that would break the `.command.env` file which expects a key-value pair on each line.
> 
> The same limitation already exists for `env` outputs, e.g. if I do `FOO=$(bash --version)` in the process script then the process output will be squeezed onto a single line. For this reason, and the fact that `cmd` outputs are intended for small things like metadata, while larger outputs can always be written to a file, I am fine with leaving the current behavior as it is.


> But that would break the `.command.env` file which expects a key-value pair on each line.

Why do we need this constraint for the command output type (and env type as well, for what is worth in this scope)?  I don't get it.

For the output type specifically, a command output by design **can** be multi-line, so we should look for an implementation that supports it.   Say we store the multi-line output in a `.command.cmd` file, then this gets passed to the process that needs it as input, which will operate on it the way it needs to. What is wrong in my thinking here?

marcodelapierre:
> This `cmd` output is basically a small syntax sugar over an `env` output, and I'm starting to wonder whether it's worth adding.

At this stage I am inclined to think we should have the `cmd` type. If I think from a user perspective:
- the saved verbosity is good, not only the process def is more concise, but the extra bash boilerplate is also avoided
- whereas I can accept the content of an environment variable to be constrained to one line, I really want the ability to handle a multi-line command output

modules/nextflow/src/main/groovy/nextflow/script/params/TupleInParam.groovy
marcodelapierre: thank you!

pditommaso:
> It can be done when all the tool versions are collected

It looks nice, well done. I agree with Marco to stay with a `cmd` to make it more concise and being able to support multiline output. 

To make it fails on the very first error it may be enough pretending the command with `set -e;` (or pipeline?)

docs/process.md
```diff
@@ -1057,43 +1057,38 @@ To sum up, the use of output files with static names over dynamic ones is prefer
...
+
+The `cmd` qualifier allows you to capture the standard output of an arbitrary shell command:
+
+```{literalinclude} snippets/process-out-cmd.nf
+:language: groovy
 ```
 
+Only one-line Bash commands are supported. You can use a semi-colon `;` to specify multiple Bash commands on a single line, and many interpreters can execute arbitrary code on the command line, e.g. `python -c 'print("Hello world!")'`.
+
+If the command fails, the task will also fail. In Bash, you can append `|| true` to a command to suppress any command failure.
```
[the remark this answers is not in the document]
bentsherman: > To make it fails on the very first error it may be enough pretending the command with set -e; (or pipeline?)

Unfortunately no, even with `bash -ueo pipefail` it still doesn't fail. I can't find a way to make it fail without assigning it to an environment var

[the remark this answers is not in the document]
bentsherman: As a workaround we could do:
```bash
FOO=$(bash --version) # fails if inner command fails
echo FOO="$FOO"
```

bentsherman:
> I agree with Marco to stay with a cmd to make it more concise and being able to support multiline output.

It's worth mentioning that nf-core does not need multi-line output to collect the tool versions. But in any case, if we keep the `cmd` output then we need to figure out how to support the multi-line output.

A simple solution would be to escape newlines:
```bash
$ FOO=$(bash --version) ; echo FOO="${FOO//$'\n'/\\n}"
FOO=GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)\nCopyright (C) 2020 Free Software Foundation, Inc.\nLicense GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>\n\nThis is free software; you are free to change and redistribute it.\nThere is NO WARRANTY, to the extent permitted by law.
```

Then unescape them when reading from `.command.env`. We could also extend this to `env` outputs at some point.

pditommaso:
For multiline output, it could be enough to make the .env parsing a bit smarter and use the pattern key=.. as the starting line, somehow similar to a Fasta file format

pditommaso:
All the point of this feature is to remove the boiler plate in the script definition. In this extent `cmd` is better then `env`.

bentsherman:
It does not remove much boilerplate compared to `env`, and there are trade-offs in usability and implementation complexity that make me think it is not really worth adding. But I would like to hear from some nf-core devs about which way they prefer. If we add this feature and then they just use the `env` approach then it will have been a waste of time

bentsherman:
There is another advantage of the `cmd` output that I think has swayed me. If a process defines multiple possible task scripts (e.g. based on some condition) then with an `env` you would have to define the env variables each time. But with a `cmd` output you would only specify it once. I guess you could define the environment vars logic once and reuse them, but still, the `cmd` output is nicer in this case.

bentsherman:
nice!

bentsherman:
It looks like you force pushed and deleted a bunch of stuff

bentsherman:
Initial nf-core sentiment seems to prefer the `env` approach: https://nfcore.slack.com/archives/C043FMKUNLB/p1701792739273019

marcodelapierre:
*** please ignore until working day ***    
(working myself this weekend to knock out my list 😂)

I say, let's consider the phase of community feedback gathering is completed, and make the move to finalise this, there isn't much work left. Once the feature is available, users will be able to assess `env` vs `cmd` and make their choice.

Please let's review the tick-box list above: https://github.com/nextflow-io/nextflow/pull/4493#issuecomment-1842409641   
I have addded names to items.

marcodelapierre:
If I try and run the following with this branch:
```
process foo {
  output:
  stdout emit: Out
//  env(HELLO), emit: Version
  cmd('git --version'), emit: Version

  '''
  echo Ciao
  export HELLO=world
  '''
}

workflow {
  foo()
  foo.out.Out | view
  foo.out.Version | view
}
```

I get the error:
```
ERROR ~ No such variable: Version
```

Paolo, I am wondering whether this relates to what Ben noticed above, about force pushed/deleted stuff.

marcodelapierre:
OK, I have done some quick meaningful testing with:
```
params.cmd='git'
params.env='HELLO'

process foo {
  output:
  stdout emit: Out
//  env("${params.env}"), emit: Version
//  cmd('"${params.cmd}" --version'), emit: Version
  cmd('git --version'), emit: Version

  '''
  echo Ciao
  export HELLO=world
  export BYE=moon
  '''
}

workflow {
  foo()
  foo.out.Out | view
  foo.out.Version | view
}
```

- Multi-line output seems OK
- **Error handling NOT OK**: if I write a failing command, the workflow won't fail
- One more thing: `cmd` does not seem to support use of `params` inside the comand string, I think this should work, for consistency with other output types - adding to tick-box list above

pditommaso:
Ok, fixed the handling of errors and the use of variables in the `cmd` statement. 

Proper error handling was not simply because it's tricky in Bash to capture the stdout, stderr and command exit status as separate variables. 

However this state-of-art of Bash snippet does the magic (courtesy of Stackoverflow) 

https://github.com/nextflow-io/nextflow/blob/624d864483ef8c5fb98b18f4cd129f2f225cccee/modules/nextflow/src/main/resources/nextflow/executor/command-env.txt#L26-L32

To summarise both `env` and `cmd` are stored in the .command.end using this syntax 

```
VAR_NAME=output line 1
output line 2
output line n
/VAR_NAME/=exit:NUMBER
```

Where `VAR_NAME` is an arbitrary env variable name or command temporary variable name. `/VAR_NAME/` is used as termination placeholder. 

The optional `=exit:NUMBER` holds the exit status of the output command execution, that's expected to be zero. 

When non-zero is interpreted as error condition and the variable contains the failed command stderr.

marcodelapierre:
I am seeing an issue with `tuples`, when trying to pass the output of the `cmd` as a simple input for another process, ie via an input `val`.

Without `tuple` it works:
```
process one {
  output:
  cmd("git --version")

  script:
  """
  echo Ciao >file
  """
}

process two {
  input:
  val(due)
  output:
  stdout

  script:
  """
  due="$due"
  echo "Two is \$due"
  """
}

workflow {
  one | two | view
}
```

But with a `tuple` it errors:
```
process one {
  output:
//  tuple path('file'), val(trial)
  tuple path('file'), cmd("git --version")

  script:
  trial = 'trial'
  """
  echo Ciao >file
  """
}

process two {
  input:
  tuple path('file'), val('due')
  output:
  stdout

  script:
  """
  uno=\$( cat file )
  due="$due"
  echo "One is \$uno ; Two is \$due"
  """
}

workflow {
  one | two | view
}
```

```
ERROR ~ Unknown process directive: `cmd`

 -- Check script 'tuple.nf' at line: 4 or see '.nextflow.log' file for more details
```

Am I doing something wrong? or is it a bug in the code?

marcodelapierre:
Latest summary of outstanding points:

- [ ] Bug in `tuple` composition  (see above)
- [ ] Add + fix tests
- [ ] Add documentation
- [ ] Check most recent commits for force push/deletion of a bunch of stuff  @pditommaso

pditommaso:
yeah, think I've messed up with Ben commit fixing the support for tuple

pditommaso:
Support for tuples fixed

bentsherman:
My changes also added docs for the cmd type, you should be able to recover it

TODO: add note to `env` output type that multi-line output support was added with this PR

pditommaso:
Unless you are a local copy, don't think it's possible to recover via GH

marcodelapierre:
last man standing is the documentation , then ready to merge

pditommaso:
While testing multiline `cmd` it turns out that only the first one works, the problem is the that having command line like this `cmd('bash --version\ngit --version')` it is turned into this snippet 

```
nxf_catch STDOUT STDERR bash --version
git --version
```

therefore the lines after the first are ignored

marcodelapierre:
> Support for tuples fixed

I can confirm that all my tests now work.

bentsherman:
> Unless you are a local copy, don't think it's possible to recover via GH

You can, just look through the event log on this thread for where you force pushed changes, then click "Compare" to see what was lost

docs/process.md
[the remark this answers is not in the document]
marcodelapierre: resolved

marcodelapierre (approved):
Good for merge for me!

marcodelapierre:
@pditommaso AKA the nextflow merge-nator for you! 😎

marcodelapierre:
@pditommaso  any blocker for this to be merged? Conversation around the review seems pretty much settled

pditommaso:
Ok, I've cleaned up and refactored a but this. The main change is `output: cmd` become `output: eval`. The main reason is that *command* is too generic term and it was overlapping with the task command. I think *eval* express better the overall idea of a small computation that should be evaluated to capture the output value.

ewels:
Nice! I like `eval`, good call 👍🏻

docs/process.md
```diff
@@ -1049,43 +1049,49 @@ To sum up, the use of output files with static names over dynamic ones is prefer
...
+
+(process-set)=
+
+### Output type `set`
+
+:::{deprecated} 19.08.1-edge
+Use `tuple` instead.
+:::
+
```
bentsherman, line 1094: Looks like this was added back in from a merge commit, it should be removed

pditommaso: Well spotted 👍  cce07c92

PR #4501, opened 2023-11-09
Add docs section on container requirements

### LINKED ISSUE #4413: Update official documentation to list recommended container base images
As per the Slack discussion [here](https://nextflow.slack.com/archives/C02T97HAV5M/p1696986142936759) I discovered that Nextflow containers meant for usage on AWS Batch should not be built on an Alpine Linux base image;

https://github.com/nextflow-io/nextflow/issues/1865

some more details here; 

https://github.com/aws/aws-cli/issues/4971

Technically this is not a Nextflow issue but instead an AWS Batch issue it seems with the relation between AWS and Alpine Linux, however, considering that we already have some official docs for Nextflow about Container usage;

https://nextflow.io/docs/latest/container.html

maybe we can update these docs to include a couple **recommended** base images to use for packaging tools and scripts inside Docker containers to run in a Nextflow pipeline. 

In my case, I finally settled on Ubuntu 22.04 ( https://hub.docker.com/_/ubuntu/tags ) as a base image for my custom tools since it seems to work with everything out of the box (has all required [reporting tools](https://nextflow.io/docs/latest/tracing.html#execution-report-tasks) along with `ps` and `bash`). 

Having a quick note recommending this or some other container that "just works" would be really helpful!

bentsherman:
Close #4413 

- Add "Requirements" section to Containers page
- Minor improvements to AWS page


docs/container.md
```diff
@@ -4,9 +4,11 @@
...
 
-:::{note}
-When creating your container image to use with Nextflow, make sure that Bash (3.x or later) and `ps` are installed in your image, along with other tools required for collecting metrics (See  {ref}`this section <execution-report-tasks>`). Also, Bash should be available on the path `/bin/bash` and it should be the container entrypoint.
-:::
+## Requirements
+
+When creating a container image to use with Nextflow, make sure that Bash (3.x or later) and `ps` are installed in the image, along with other tools required for collecting metrics (See {ref}`this section <execution-report-tasks>`). Bash should be available on the path `/bin/bash` and it should be the container entrypoint.
+
+When using AWS Batch, containers must be able to use the {ref}`AWS CLI <aws-batch-cli>` (unless you are using {ref}`wave-page` and {ref}`fusion-page`).
```
pditommaso, line 11: Don't think this should be mentioned here. I'd replace with a tip about Wave e.g. "Consider using Wave containers to streamline the provisioning containers on-demand for your data pipeline"

docs/aws.md
```diff
@@ -138,24 +138,22 @@ Nextflow provides built-in support for AWS Batch, allowing the seamless deployme
...
+(aws-batch-cli)=
 
 ### AWS CLI
 
 Nextflow needs the [AWS command line tool](https://aws.amazon.com/cli/) (`aws`) to be available in the container in which tasks are executed, in order to stage input files and output files to and from S3 storage.
 
-:::{tip}
-When using {ref}`wave-page` and {ref}`fusion-page`, the AWS command line tool is not needed for task containers or the underlying EC2 instances when running Nextflow on AWS Batch. See the {ref}`fusion-page` documentation for more details.
-:::
+The `aws` command can be made available in any of the following ways:
```
pditommaso, line 147: Think Wave tip should stay on top and it should be made clear that installing the `aws` cli  is considered a legacy approach

PR #4188, opened 2023-08-14
Fix ParamsMap copyWith param aliases

bentsherman:
Close #3476 

When updating a params map, the aliases weren't being removed from the read-only set, which prevented the params map from being updated correctly.

Issue #3476, opened 2022-12-06
The value taken by a parameter is affected by uppercase letters in the parameter name


fortune9:
### Expected behavior and actual behavior

The uppercase and lowercase parameters behave differently in value override. For example, the parameters params.msg and params.msG in the following workflow takes the values from different scopes, just because their names containing uppercase letters or not. 

Expectation: both parameters should behave the same, no matter whether there are uppercase letters in the parameter names.

### Steps to reproduce the problem

file: wf_addparams.nf

```groovy
nextflow.enable.dsl=2

params.msg="workflow msg lower case"
params.msG="workflow msg upper case"

// include the process
include { print_msg } from \
    './msg.nf' addParams (
        msg: "include msg lower case",
        msG: "include msg upper case"
        )


workflow {
    print_msg()
}
```

The included process file is as follows:
file: msg.nf

```groovy
process print_msg {

    echo true

    script:
    """
    echo msg: ${params.msg}
    echo msG: ${params.msG}
    """
}
```

### Program output 

- Run 1

    * Command
       nextflow run ./wf_addparams.nf
    * Output
    N E X T F L O W ~ version 22.04.5 Launching `./wf_addparams.nf` [soggy_swirles] DSL2 - revision: 61b19cc590 WARN: The `echo` directive has been deprecated - use to `debug` instead executor > local (1) [ef/9126b4] process > print_msg [100%] 1 of 1 ✔ msg: include msg lower case msG: workflow msg upper case
    * Problem: as you can see, params.msg takes the value set at addParams, while params.msG takes the global value set at the beginning of the workflow file.

- Run 2

    * Command
       nextflow run ./wf_addparams.nf--msg "cmd msg lower case" --msG "cmd msg upper case"
    * Output
    N E X T F L O W ~ version 22.04.5 Launching `./wf_addparams.nf` [sad_turing] DSL2 - revision: 61b19cc590 WARN: The `echo` directive has been deprecated - use to `debug` instead executor > local (1) [50/bebd58] process > print_msg [100%] 1 of 1 ✔ msg: include msg lower case msG: cmd msg upper case
    * Problem: again, params.msg takes the value set at addParams, while params.msG takes the value set at the command line.

### Environment 

* Nextflow version: 22.04.5
* Java version: openjdk version "17.0.5"
* Operating system: 18.04.1-Ubuntu
* Bash version: GNU bash, version 4.4.20(1)-release (x86_64-pc-linux-gnu)

PR #4510, opened 2023-11-13
Module config

### LINKED ISSUE #4422: [proposal] Add local module named configuration template
## Context 

The use of DSL2 introduces the use of module components that allow the inclusion and re-use a process more than one time with different names.

A common requirement in this scenario is that a process configuration needs to be customised, for example, to provide a specific `publishDir` path or command arguments. 

Currently, this need is addressed by creating a monolithic configuration file in which the process configuration is provided by using the target process name. For example here 

https://github.com/nf-core/rnaseq/blob/master/conf/modules.config#L174-L192

In some case, there's a conditional logic that needs to be duplicated in the workflow script. For example here: 

https://github.com/nf-core/rnaseq/blob/master/conf/modules.config#L209-L224

This approach has several disadvantages: 

* Monolithic configuration file
* The process configuration is out of the context from the workflow code 
* Very fragile; it's enough a typo in the config file to break the process  configuration 
* The use of monolithic breaks the workflow modularization 
* Very hard to maintain. 

## Solution 

A possible solution could be to introduce a module-level configuration file that allows the definition of named configuration template that can applied to one or more processes in the same module. 

For example the following configuration file could be added in a workflow module: 

```nextflow 
config.FOO.ext.args = "--x one --y two" 
config.FOO.publishDir = [path: "/some/path"]

config.BAR.ext.args = "--x three --y four" 
config.BAR.publishDir = [path: "/other/path/${params.suffix}"]
```

Note: in the example above, `FOO` and `BAR` are arbitrary identifiers the user chooses. They have no relationship with the process name where they are meant to be used. 

In the workflow module, they could be used using the following syntax: 

```
include { PROC1 } from './module/path1'
include { PROC2 } from './module/path2'

workflow NAME {
  PROC1(config.FOO)
  PROC2(config.BAR) 
}
```

*Note: explorative syntax, it may change*

In the above example, the process `PROC1` is executed having the config setting defined by the template with name `FOO` is applied. Same with with process `PROC2` with config `BAR`. 

Conditional configuration could be defined using usual `if` and expression statement e.g. 

```
workflow NAME {
  def cfg = params.something ? config.FOO : config.BAR
  PROC1(cfg)
}
```

This approach has several advantages: 
* keep the config settings in the nextflow config file, aside the module where it's expected to be used
* promote a declarative approach for configuration files 
* decouple the configuration from the target process name and therefore, increase reusability 
* Avoid the verbosity of `withName`selector mechanism   

## Caveats
* Named config templates should only accessible in the module where they are defined 
* A conventional config file should used e.g. `module.config` or `main.config` 
* The `withName` selector should be able to override settings defined by a config template (to not break existing config e.g. groundswell)

bentsherman:
Close #4422


docs/module.md
```diff
@@ -271,6 +271,52 @@ Those scripts will be made accessible like any other command in the task environ
...
+// module.config
+withName:'FOO:BAR' {
+    ext.args = '--n-iters 1000'
+    publishDir = "${params.outdir}/foo_bar"
+}
+```
+
+In the above example, the module config defines process config settings using the same syntax as a Nextflow config file (including {ref}`process selectors <config-process-selectors>`) but implicitly within the process config scope. The selector `FOO:BAR` matches the process `BAR` invoked by workflow `FOO`.
+
+Futhermore, the selector `BAR` would have also worked in this case, even if `BAR` is used elsewhere in the pipeline, because the module config is only applied to this module. This assurance is the advantage of defining process config in module config files instead of the global pipeline config.
```
bentsherman, line 309:
```suggestion
Furthermore, the selector `BAR` would have also worked in this case, even if `BAR` is used elsewhere in the pipeline, because the module config is only applied to this module. This assurance is the advantage of defining process config in module config files instead of the global pipeline config.
```

marcodelapierre (commented):
Local build and test successful.

Also tested the examples provided in the bug report in the root issue https://github.com/nextflow-io/nextflow/issues/3476 .

I only have one question Ben. In this case:
```
nextflow run ./wf_addparams.nf --msg "cmd msg lower case" --msG "cmd msg upper case"
```

It is correct that the winning set of parameters is the one coming from the `addParams` at the `include`, NOT the cmd line ones, right?

```
include { print_msg } from \
    './msg.nf' addParams (
        msg: "include msg lower case",
        msG: "include msg upper case"
        )
```

marcodelapierre (approved):

docs/config.md
```diff
@@ -1247,12 +1247,12 @@ The above configuration snippet sets 2 cpus for the processes annotated with the
 
 #### Selector priority
 
-When mixing generic process configuration and selectors the following priority rules are applied (from lower to higher):
+Process configuration settings are resolved as follows (from lowest to highest priority):
```
marcodelapierre, line 1250: Should we include also settings defined when including processes from modules?

See e.g. https://github.com/nextflow-io/nextflow/pull/4188

docs/module.md
```diff
@@ -271,6 +271,52 @@ Those scripts will be made accessible like any other command in the task environ
...
+    ext.args = '--n-iters 1000'
+    publishDir = "${params.outdir}/foo_bar"
+}
+```
+
+In the above example, the module config defines process config settings using the same syntax as a Nextflow config file (including {ref}`process selectors <config-process-selectors>`) but implicitly within the process config scope. The selector `FOO:BAR` matches the process `BAR` invoked by workflow `FOO`.
+
+Furthermore, the selector `BAR` would have also worked in this case, even if `BAR` is used elsewhere in the pipeline, because the module config is only applied to this module. This assurance is the advantage of defining process config in module config files instead of the global pipeline config.
+
+Process configuration is resolved as follows (from highest to lowest priority):
```
marcodelapierre, line 311: should this list and the analogous one in `config.md` be made consistent ?

docs/module.md
```diff
@@ -271,6 +271,52 @@ Those scripts will be made accessible like any other command in the task environ
...
+}
+
+workflow FOO {
+    BAR()
+}
+```
+
+```groovy
+// module.config
+withName:'FOO:BAR' {
```
marcodelapierre, line 301: Questions:
- what happens in this PR when using `withName: BAR` in the module config, without the `FOO:`?
- what happens in this PR when defining module config lines with no Name/Label selectors?

(asking from a user experience perspective - haven't looked at the code nor tested yet, on purpose)

modules/nextflow/src/test/groovy/nextflow/script/ProcessDefTest.groovy
marcodelapierre, line 1: shall we have unit tests about the composition of configs from multiple sources (ideally, process, module and workflow) ?

modules/nextflow/src/main/groovy/nextflow/script/ExecutionStack.groovy
marcodelapierre, line 1: unit tests for edits in ExecutionStack ?

modules/nextflow/src/main/groovy/nextflow/script/IncludeDef.groovy
marcodelapierre, line 1: new unit tests required for IncludeDef?

modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy
marcodelapierre, line 1: new unit tests required for ScriptMeta?

marcodelapierre (changes requested):

marcodelapierre:
Removed from edge November milestone, to allow time for team discussion

docs/module.md
bentsherman: The standard priorities apply, from lowest to highest:
- generic process config
- process definition
- withLabel
- withName (base name)
- withName (imported name)
- withName (fully qualified name)

marcodelapierre:
@pditommaso @bentsherman  Ran out of capacity to finalise some testing and articulate my thoughts on this one. The general lines is that I am not entirely bought on the proposed syntax.  

Keen to resume on this in the new year, however if you two are confident on the way forward, I do not want to be the blocker.

marcodelapierre:
OK, had a good review of the PR today, from the point of view of the design, also having a look at the Rnaseq POC.

I like a lot how a simple design change, i.e. having the module/subwf specific config files, can make such a big impact, i.e. enabling to modularise these configuration entries for improved readability, maintainability, reusability. 

One thing I reflected on was Paolo's point on the fragility coming from process selector typos.  I had a conversation with @bentsherman  around this, and he mentioned the existence of a parameter, I believe `nextflow.enable.configProcessNamesValidation`, that enables spitting Warnings for no matches of process selectors.  
Nf-core developers used to turn it off, to avoid false positives with the old monolithic process configuration.  
Now, with modular configurations, Ben points out that false positives can be eliminated, and hence the parameter can be turned on. This will inform pipeline devs/users of no match / typo  situations with process selectors.

In summary, I thing the design of this PR, beside matching most original requirements in https://github.com/nextflow-io/nextflow/issues/4422, provides at least 2 additional advantages:
1. simplicity
2. use of existing syntax (process selector): devs do not have to learn new syntax

@pditommaso , further comments on the design of this functionality?

I will go ahead tomorrow and review documentation and implementation.

docs/config.md
```diff
@@ -4,20 +4,20 @@
...
+When a pipeline script is launched, Nextflow looks for configuration files in multiple locations. Since each configuration file may contain conflicting settings, they are resolved as follows (from highest to lowest priority):
 
 1. Parameters specified on the command line (`--something value`)
 2. Parameters provided using the `-params-file` option
-3. Config file specified using the `-c my_config` option
-4. The config file named `nextflow.config` in the current directory
-5. The config file named `nextflow.config` in the workflow project directory
+3. Config file specified using the `-c <config-file>` option
+4. The config file `nextflow.config` in the launch directory
+5. The config file `nextflow.config` in the project directory
```
marcodelapierre, line 13: After 4. and 5., I would add an item in this list as well:
```
6. The module config file `module.config` in the project module directory (`process` parameters only, see Module config(REF) section in the Modules(REF) page)
```

I reckon the module.config takes higher priority then the HOME config file, `$HOME/.nextflow/config`, right?

docs/config.md
```diff
@@ -1343,12 +1343,12 @@ The above configuration snippet sets 2 cpus for the processes annotated with the
 
 #### Selector priority
 
-When mixing generic process configuration and selectors the following priority rules are applied (from lower to higher):
+Process configuration settings are resolved as follows (from lowest to highest priority):
```
marcodelapierre, line 1346: can you flip the list from highest to lowest priority, for consistency with the other lists? (and hence improved overall readability)

marcodelapierre (changes requested):
Thanks Ben, 

I have finished reviewing, see my outstanding comments from my previous review (docs and tests), plus the new ones on the docs

marcodelapierre:
@pditommaso  

https://github.com/nextflow-io/nextflow/pull/4416 to be also merged as part of this unit of work

docs/config.md
bentsherman: It would be lower than the HOME config file, because the config files from 3-6 are resolved into a single "pipeline config" which takes precedence over module configs. I will add it as 7

docs/config.md
bentsherman: I think I actually prefer going from lowest to highest, I find it easier to reason about. What if we switched the other lists to be low to high?

docs/module.md
```diff
@@ -271,6 +271,52 @@ Those scripts will be made accessible like any other command in the task environ
 This feature requires the use of a local or shared file system for the pipeline work directory, or {ref}`wave-page` when using cloud-based executors.
 :::
 
+## Module config
```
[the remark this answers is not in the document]
bentsherman: Agreed

docs/module.md
```diff
@@ -271,6 +271,52 @@ Those scripts will be made accessible like any other command in the task environ
...
+// main.nf
+process BAR {
+}
+
+workflow FOO {
+    BAR()
+}
+```
+
+```groovy
```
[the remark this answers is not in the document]
bentsherman: Agreed

docs/module.md
```diff
@@ -271,6 +271,52 @@ Those scripts will be made accessible like any other command in the task environ
 This feature requires the use of a local or shared file system for the pipeline work directory, or {ref}`wave-page` when using cloud-based executors.
 :::
 
+## Module config
+
+:::{versionadded} 23.11.0-edge
+:::
+
+Modules can define a config file called `module.config` in the module directory. This config file can be used to apply process configuration to processes that are invoked in the module. Unlike a regular Nextflow configuration file, the module config assumes the {ref}`process config scope <config-process>` and only supports process config settings.
```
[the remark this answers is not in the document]
bentsherman: Sure

docs/config.md
bentsherman: Switching the other lists, let me know what you think

marcodelapierre: Agree

modules/nextflow/src/test/groovy/nextflow/script/ProcessDefTest.groovy
marcodelapierre: @bentsherman for your consideration

modules/nextflow/src/main/groovy/nextflow/script/ExecutionStack.groovy
marcodelapierre: @bentsherman for your consideration

modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy
marcodelapierre: @bentsherman for your consideration

bentsherman:
Now that we're converging on a solution for the workflow publish definition, I'm having second thoughts about this module config idea.

The module config further complicates the rules for config resolution, and at this point it would only be used to define `ext` config for processes. In other words, the module config only serves as a convenient shortcut for what should really be pipeline params.

I think we should discourage the use of `ext` to provide inputs to the process (i.e. values that are used in the script body), and instead encourage those settings to be made explicit inputs. If they need to be configurable by the user, they can be made so with pipeline params. This can be done today, but it isn't done because process selectors + `ext` is more convenient.

Moving `ext` config to explicit inputs will make the params list much longer, at least for nf-core/rnaseq, but it's more transparent and makes the params more comprehensive. Right now, the "inputs" of a process could come from the actual inputs, or pipeline params, or `ext` config, and that makes it harder to read. Instead, everything should be a process input, and the calling workflow can decide how to provide those inputs.

The big picture that I'm beginning to see is, there are many language features which were fine in DSL1 when everything was in a single script, but become troublesome in DSL2 with modules and subworkflows:
- publishDir (already discussed)
- using `params` anywhere (they should only be used in the anonymous workflow)
- arbitrary Groovy code anywhere (it should only be allowed in the process / workflow / function body)
- process selectors (overcomplicated by all the rules for defined vs imported name, simple vs fully qualified name, etc)

Another way to look at it is, if you consider the list of process directives, they all affect *how* tasks are executed (executor, hardware resources, software deps, error handling, etc), as opposed to *what* is executed (inputs, outputs, script). With two notable exceptions:
1. `publishDir`, which we are moving away from
2. `ext`, if it is used to provide inputs into the script (which I have argued is an anti-pattern)
   a. there is a valid use case for `ext` to define the *how*, e.g. YellowDog uses it for custom executor settings

So if process directives are used to control *how* a process is executed, and the *how* is usually specific to a pipeline-level config profile, then there is no need for module config because modules typically do not describe how they are executed, or when they do, such as `conda` / `container`, it is usually in the process definition.

pditommaso:
Completely agree. We should consider deprecating `ext` config option

bentsherman:
Okay, then I will close this PR. There are some general docs improvements here about the config resolution, I will make a separate PR for those

PR #4517, opened 2023-11-17
Makefile clean to also remove buildSrc/build

marcodelapierre:
@pditommaso  this should fix weird behaviours observed when testing Nextflow builds with source modifications.
A while ago the fix you handed me was:
```
find . -name MANIFEST.MF | grep build/tmp | xargs rm
```

Upon inspection, the command above reduces to the `Makefile` change in this PR.


pditommaso:
well done

PR #4518, opened 2023-11-17
Add "fusion.cacheSize" config option

jordeu:
## Description

Add `cacheSize` Fusion configuration option that allows to limit the size of the local cache that creates Fusion.

### DIFF
--- a/docs/fusion.md
+++ b/docs/fusion.md
@@ -212,3 +212,6 @@ The following configuration options are available:
 
 `fusion.tagsPattern`
 : The pattern that determines how tags are applied to files created via the Fusion client (default: `[.command.*|.exitcode|.fusion.*](nextflow.io/metadata=true),[*](nextflow.io/temporary=true)`)
+
+`fusion.cacheSize`
+: The maximum size of the local cache used by the Fusion client.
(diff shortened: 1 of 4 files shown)


pditommaso (approved):

jordeu:
I didn't added the "new in version" at the docs. Do you think is better to add always?

pditommaso:
You are right better to add it

jordeu:
> You are right better to add it

23.11.0-edge ???

pditommaso:
yes

docs/fusion.md
```diff
@@ -212,3 +212,8 @@ The following configuration options are available:
 
 `fusion.tagsPattern`
 : The pattern that determines how tags are applied to files created via the Fusion client (default: `[.command.*|.exitcode|.fusion.*](nextflow.io/metadata=true),[*](nextflow.io/temporary=true)`)
+
+`fusion.cacheSize`
+: :::{versionadded} 23.11.0-edge
+  :::
+: The maximum size of the local cache used by the Fusion client.
```
bentsherman, line 219: it would be nice to keep these options in alphabetical order, even though the Fusion options aren't perfectly ordered at the moment

bentsherman: I would say, keep `enabled` first but then alphabetize the rest

jordeu: done

PR #4521, opened 2023-11-17
Bypass google batch price query if task cpus and memory are defined

### LINKED ISSUE #4520: Add option to disable google batch vm price query
Google Batch queries `cloudinfo.seqera.io` to fetch VM prices in order to try and find an optimal machine type for the task requirements. While it can always fall back to the default behavior in the event of an error, it can cause delays if Nextflow does not have access to the public internet (e.g. a locked-down enterprise environment).

https://github.com/nextflow-io/nextflow/blob/82e56799654eab43f37fb41ada579b34637f56f6/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy#L144-L164

We should add a config option to disable this logic, and maybe also disable it automatically if `NXF_OFFLINE` is enabled.

bentsherman:
Close #4520 

Maybe we still want to add an explicit option to disable the price query though

### DIFF
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy
@@ -90,7 +90,6 @@ class GoogleBatchMachineTypeSelector {
     }
 
     MachineType bestMachineType(int cpus, int memoryMB, String region, boolean spot, boolean fusionEnabled, List<String> families) {
-        final machineTypes = getAvailableMachineTypes(region, spot)
         if (families == null)
             families = Collections.<String>emptyList()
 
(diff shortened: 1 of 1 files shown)


bentsherman:
We should have done this anyway because currently every Nextflow run on google batch is pinging `cloudinfo.seqera.io` for every single task execution...

bentsherman:
Well, I guess it's not so bad since the price query is memoized.

On that note, one more thing we could add is a try-catch around this line: https://github.com/nextflow-io/nextflow/blob/dd32454eb0f03707452b464c3e0e0eda30d49140/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy#L146

To return an empty list on timeout errors, that way the result is cached and the timeout only happens once per pipeline run in the worst case

jordeu (approved):

jordeu:
> Well, I guess it's not so bad since the price query is memoized.
> 
> On that note, one more thing we could add is a try-catch around this line:
> 
> https://github.com/nextflow-io/nextflow/blob/dd32454eb0f03707452b464c3e0e0eda30d49140/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy#L146
> 
> To return an empty list on timeout errors, that way the result is cached and the timeout only happens once per pipeline run in the worst case

Is the timeout very long? It is possible to just reduce the timeout?

Be aware, that if you do this empty list thing, then you need to handle better this case on `bestMachineType` to throw a more meaningful error message.

pditommaso:
Merging this in the current form. The retry policy can be added in a separate PR

PR #4535, opened 2023-11-22
Support disk image for Google Batch via `disk` directive

### LINKED ISSUE #4534: Support specifying disk image for additional disk with Google Batch (e.g. Image with reference files)
## New feature

Support specifying a disk image for additional disk when using Google Batch. This would be useful in many cases but particularly with large reference files which can be both time consuming and expensive.
Using DIsk image is faster and cheaper as the Disk is created from the image before the provisioning the VM

## Usage scenario 

User creates a Disk image containing all reference files 
User specifies the image as part of the disk directive in the process
## Suggest implementation 
Update Google Batch Task handler to use disk image if both type and image are specified. This will take advantage of the current code which applies to disk type

This could look like this

I've created a sample implementation in a PR

hnawar:
## New feature
Support specifying a disk image for additional disk when using Google Batch. This would be useful in many cases but particularly with large reference files which can be both time consuming and expensive.
Using DIsk image is faster and cheaper as the Disk is created from the image before the provisioning the VM

## Usage scenario
User creates a Disk image containing all reference files
User specifies the image as part of the disk directive in the process

#4534


plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -277,21 +277,43 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
...
-                    )
-                    .setDeviceName('scratch')
-            )
+            if ( disk?.image ) {
+                instancePolicy.addDisks(
+                    AllocationPolicy.AttachedDisk.newBuilder()
+                        .setNewDisk(
+                            AllocationPolicy.Disk.newBuilder()
+                                .setType(disk.type)
+                                .setImage(disk.image)
```
bentsherman, line 286: If I understand correctly, when the image is specified the size is irrelevant?

hnawar: Yes, When image or snapshot are specified as a source, then the disk size is ignored.
I kept it in the example documentation as I cannot leave it empty or I would get an error "Not a valid disk value: null"

bentsherman: And is there a reason why the type is also required when using an image?

hnawar: Disk type is not implied from the Image and would still be needed.
We can also add a default disk type if not specified by the user if you think this would be a better experience

bentsherman: Just making sure I understand the requirements from google batch. Actually Nextflow uses the disk type to decide whether to apply the disk directive to the boot disk size or a new disk, so this should be fine as it is.

bentsherman:
This change by itself is fine, but the disk directive now has three purposes (resize boot disk, attach empty disk, attach disk with image) but can be specified only once. That means that e.g. you couldn't attach a disk with reference inputs and also attack a local SSD to use as scratch storage.

To solve this issue we just need to allow the disk directive to be specified multiple times. But I am fine with merging this PR as it is and implementing multiple disk directives in a separate PR.

bentsherman (approved):

hnawar:
Thanks Ben :)
Note, that the disk can still be used as scratch space if it is not completely full and is mounted in the same path with read/write permissions

I did think about the multiple disk directive, but did not have an answer for that

PR #4538, opened 2023-11-23
Fix the missing labels field in Job request for Google Batch

zetxqx:
This change will add the missing labels to Batch Job request.
Without some of the labels, some Batch side optimization is not recognized.


pditommaso (approved):

bentsherman:
@zetxqx do you think we should still add these labels to the allocation policy?

https://github.com/nextflow-io/nextflow/blob/5a3a0fa2cc6236ff94e427cf02dfee5609bcf3bc/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy#L337

I'm guessing they are both useful as one is for the job metadata and the other is for the physical resources.

zetxqx:
Yep, there are still some labels valuable for allocationPolicy, we might need to expose more fields on the NextFlow side as well. Currently, this is a comprimised solution not breaking things. Will make more change after I understand the nextFlow code base better.

bentsherman:
Sure, feel free to ping me if you have questions about the codebase

PR #4545, opened 2023-11-27
Add example of dynamic directive with input file

bentsherman:
See https://github.com/nextflow-io/nextflow/pull/4493#discussion_r1397589519

I don't really like this particular example because it is probably an anti-pattern, i.e. many tools are designed to stream the input data so that they don't have to load entire datasets in memory. But I want to explicitly show that input files can also be used to configure directives.


drpatelh:
There are some other real-world examples in this talk:
https://youtu.be/oAdSpdUrJag?t=16928

Would be awesome to have something like this to tune resources.

bentsherman:
@christopher-hakkaart for your consideration. It would be good to have an example of setting the memory based on the task inputs, but ideally would be something that is proven to work in the wild. Instead of using `data.size()`, which I just made up, you might find a more concrete example in the talk that Harshil linked.

Even something like this could be better:
```groovy
// allocate 1 GB for every million reads, up to 32 GB
memory { meta.num_reads / 1_000_000 * 1.GB }
```

bentsherman:
@christopher-hakkaart can you give this one a review when you have a chance

docs/process.md
```diff
@@ -1311,6 +1311,19 @@ disk { [request: 375.GB * task.attempt, type: 'local-ssd'] }
 ```
 :::
 
+Task resources can also be defined in terms of task inputs, for example:
+
+```nextflow
+process foo {
+    memory { 8.GB + 1.GB * Math.ceil(input_file.size() / 1e9) }
```
pditommaso, line 1318: Do we want really promote the use of `1e9` notation in the docs?

docs/process.md
```diff
@@ -1311,6 +1311,19 @@ disk { [request: 375.GB * task.attempt, type: 'local-ssd'] }
 ```
 :::
 
+Task resources can also be defined in terms of task inputs, for example:
```
christopher-hakkaart, line 1314:
```suggestion
Task resources can also be defined in terms of task inputs. For example:
```

docs/process.md
```diff
@@ -1311,6 +1311,19 @@ disk { [request: 375.GB * task.attempt, type: 'local-ssd'] }
...
+```nextflow
+process foo {
+    memory { 8.GB + 1.GB * Math.ceil(input_file.size() / 1e9) }
+
+    input:
+    path input_file
+}
+```
+
+In this example, each task requests 8 GB, plus the size of the input file rounded up to the next GB. This way, each task requests only as much memory as it needs based on the size of the inputs. The specific function that you use should be tuned for each process.
```
christopher-hakkaart, line 1325:
```suggestion
In this example, each task requests 8 GB of memory, plus the size of the input file rounded up to the next GB. This way, each task requests only as much memory as it needs based on the size of the inputs. The specific function that you use should be tuned for each process.
```

christopher-hakkaart (approved):
I'm happy with the language. The 1e9 notation could be removed or a note could be added to explain. It's not blocking for me.

PR #4549, opened 2023-12-01
welcome latch

kennyworkman:


kennyworkman:
Thank you @bentsherman, we are unable to use the plugin because our platform require a full description of the workflow DAG and typed bindings between nodes before it can be executed. This helps with code correctness, provides reproducibility guarantees and yields nice visualizations of worklow graphs for scientific teams. From reading a handful of the plugins, they rely on some executor that receives tasks one at a time. Would love to find a way to write a new type of plugin that supports these constraints as our integration matures. 

For now, we will just maintain a fork because we are using this codebase a bit differently (the PRd changes manipulate the script to invoke single process / operators within nodes of our static graph). But we will produce technical documentation throughout this effort that we can share here to brainstorm. Very fun problem

bentsherman:
You should be able to use a customer executor for this, because you have complete control over how to execute tasks. Even though they are submitted one at a time, Nextflow just watches for status changes which you control. So you could for example receive many tasks and submit them all at once. Also, you can access the complete DAG in your executor or a trace observer and post-process / send it to your backend before executing any tasks. The nf-cws plugin does something similar to this.

But there is a fundamental limitation in task scheduling, which is that Nextflow does not know the entire task graph from the beginning. It knows the "abstract" graph of processes and operators, but since a task could produce an arbitrary number of output files that trigger an arbitrary number of downstream tasks, the task graph is "discovered" as the pipeline runs. This is in the general case, in theory you could know the full task graph if you don't have any glob outputs.

Anyway, please feel free to share what you're working on. We are working on all sorts of improvements around static typing, error checking, graph optimizations, etc... always happy to brainstorm

kennyworkman:
The problem is that requiring some program (nextflow) that runs for the entire lifecycle of an execution, submitting and monitoring the status of tasks that it submits, is just not consistent with how we execute workflows on our platform. Workflow structure is defined at registration time and that structure is sufficient for the workflow to complete without an additional scheduler.

If you treat each process as a blocking, mappable task, where channels become lists of values, and dataflow operators are dispatched to their own tasks, there is a path forward to constructing a best effort static graph.

OK, sounds good - will keep you in the loop

bentsherman:
I see. That is a much more hybrid approach. I'm curious to understand how you do this with Flyte and Snakemake, like do you even use their respective runtimes to schedule tasks? Or you are just invoking them to generate a task and nothing else?

Nextflow has the ability to invoke a subworkflow in isolation, but subworkflows are not represented in the DAG anyway. The ability to invoke a single process with specific inputs would be useful for testing if nothing else, currently you have to write wrapper workflow like [this one](https://github.com/nf-wrap/main) just to test individual tools.

Operators are another story... 😅 Not all operators are as simple as a map or filter. Operators like `collect`, `combine`, `join`, `reduce`, etc might consume multiple items or even the entire channel before emitting any outputs. So you will have to either implement them directly in your backend or wait until a channel is "complete" before sending it to a reduction operator (limiting parallelism).

Issue #3976, opened 2023-05-29
Suggestion: Prettier terminal outputs


ewels:
A few ideas for making the terminal output from Nextflow a bit prettier / easier to use (taken from a discussion on Slack with @bentsherman):

1. More use of colour
    * For example: Cached = grey, failed but ignored = orange, etc etc. Even small things like the `[` characters in dim / grey. Nextflow version number in grey, etc.
1.  Use of spinners to make it clear that Nextflow hasn't crashed, so that there's always movement even when waiting for stuff.
    * Bonus points: combine with a progress bar when we know the number of expected tasks for a process.
1. _Ambitious:_ ability to show stdout from running commands, but in a truncated rolling tail of 3-4 lines.
    * Example: `docker compose`, which shows the build log output as it goes along, but without taking up the whole terminal.

Just to get the conversation moving, I threw a Nextflow log into a Python script and coloured it using [Rich](https://github.com/Textualize/rich). Gives an idea of the kind of thing I'm thinking of for the colours (I stole the header style from Astro). 

<details>
<summary><code>demo.py</code></summary>

```python
from time import sleep
from random import randint
from rich.console import Console
from rich.progress import Progress, TextColumn, SpinnerColumn, BarColumn, ProgressColumn

console = Console()

log_msg = """
 🚀  [black on green] N E X T F L O W [/]  [dim]~  [/][green]version 23.04.1[/]

 ┃ Launching [magenta]`../main.nf`[/] [dim][[/][bold cyan]boring_montalcini[/][dim]][/] [dim]DSL2 - [/][dim magenta]revision: [/][magenta]d420d96c87[/]

[dim]executor >  local (3)[/]
[dim][[/][blue]89/615a07[/][dim]][/] [dim]process >[/] [grey]NFCORE_METHYLSEQ:METHYLSEQ:PREPARE_GENOME:BISMARK_GENOMEPREPARATION[/] [dim]([/][yellow]BismarkIndex/genome.fa[/][dim])[/] [dim][[/][green]100%[/][dim]][/] 1 of 1 [green]✔[/]
[...]

PR #4573, opened 2023-12-11
Coloured logs

ewels:
To do:

- [x] Don't use ANSI for the header when `-ansi false` is set
- [ ] Respect env var `NO_COLOR`
- [ ] Try harder to auto-detect tty / ansi support and switch off automatically
- [x] Look into adding colour to the task label (bit in brackets at the end)
- [x] More testing, refine colour use
- [x] Try to get exact Nextflow green RGB, at least for the header

---

See: https://github.com/nextflow-io/nextflow/issues/3976

> [pditommaso](https://github.com/pditommaso) commented [on May 29](https://github.com/nextflow-io/nextflow/issues/3976#issuecomment-1567219430)
> That's very beautiful, but not sure exist something so cool as Rich for JVM world

> [pditommaso](https://github.com/pditommaso) commented [on May 30](https://github.com/nextflow-io/nextflow/issues/3976#issuecomment-1568071479)
> Phil. I think this a great opportunity to improve your Groovy skills 😆

I was never going to be able to back down from such a challenge 😆 

https://github.com/nextflow-io/nextflow/assets/465550/ba009ef6-88a5-4a20-90f2-ab4b8f63444b

Does a few things:

* Truncates the start / middle of the process name (instead of the end), so we can see the end. This is usually the interesting bit.
* Hides processes with zero tasks if there are more rows to print than available in the terminal (with a message at the end saying how many are hidden)
* Drops the (`process > `) at < 180 cols width to save horizontal space
* Drop the % age completion at < 120 cols to save horizontal space
* Has oodles of colour 🌈 🌟 

Known limitations:

* Code is probably terrible style
* Some of it isn't switched off when ANSI is set to false
* Likely some nasty ansi codes will go to log files when there isn't a tty attached, should try to auto-detect and disable
* Probably doesn't respect `NO_COLOR` env variables etc.


ewels:
Wondering if we can use the `isAnsiSupported()` function from `jline` to automatically disable ANSI log output when running in a non-interactive shell 🤔 

https://jline.github.io/jline2/apidocs/reference/jline/Terminal.html#isAnsiSupported()

ewels:
Updates since the video above:

* Percentage only green when it gets to 100%
* Processes with zero tasks hidden if we run out of terminal
* Switched `...` with `…` to save a couple of characters width

CleanShot 2023-12-12 at 09 25 56@2x

bentsherman:
Phil, I had an idea about how we could improve the percentage tracking. Until a process receives all the tasks it will execute, the percentage is always overestimating the true percentage. What if we show the percentage as `???` or `-` at first, and when a process receives the poison pill (which usually happens long before it finishes all of its tasks), the ANSI log can receive that signal and start showing the percentage

bentsherman:
I already implemented this event hook for array jobs and automatic cleanup, so just need to copy some code from there

ewels:
@bentsherman I go back and forth on this a little. I'm not sure which is best, the current output:

```
[-        ] NFCORE_RNASEQ:RNASEQ:FASTQ_FASTQC_UMITOOLS_TRIMGALORE:TRIMGALORE    [ 75%] 3 of 4 ✔
[-        ] NFCORE_RNASEQ:RNASEQ:ALIGN_STAR:STAR_ALIGN                          [ 33%] 1 of 3 ✔
```

or what you're proposing:

```
[-        ] NFCORE_RNASEQ:RNASEQ:FASTQ_FASTQC_UMITOOLS_TRIMGALORE:TRIMGALORE    [ 75%] 3 of 4 ✔
[-        ] NFCORE_RNASEQ:RNASEQ:ALIGN_STAR:STAR_ALIGN                          [ ???] 1 of ? ✔
```

Even if the `of 3` is a lie, at least you can see how many tasks are currently running there. With `1 of ?` you have no idea.

I guess you could go halfway and remove the percentages and keep the `of n`:

```
[-        ] NFCORE_RNASEQ:RNASEQ:FASTQ_FASTQC_UMITOOLS_TRIMGALORE:TRIMGALORE    [ 75%] 3 of 4 ✔
[-        ] NFCORE_RNASEQ:RNASEQ:ALIGN_STAR:STAR_ALIGN                          [ ???] 1 of 3 ✔
```

..but I sort of feel that you're still "lying" about the proportion of tasks done. So not sure how much of an improvement it is over the current behaviour.

Am I missing something / wrong about anything here?

bentsherman:
The third option is what I'm proposing, but without the check mark:
```
[-        ] NFCORE_RNASEQ:RNASEQ:FASTQ_FASTQC_UMITOOLS_TRIMGALORE:TRIMGALORE    [ 75%] 3 of 4 ✔
[-        ] NFCORE_RNASEQ:RNASEQ:ALIGN_STAR:STAR_ALIGN                          [ ???] 1 of 3
```

The question marks in the percentage and the lack of a check mark tell you that this process hasn't received all of its tasks yet, so the true progress is unknown. I think that's reasonable

ewels:
Ah sorry, the check mark was me manually copying and pasting stuff for the example. That's not there in the current behaviour (see above screenshots).

bentsherman:
I see, well my point is basically the same

ewels:
Yup, agreed. I like it! Feel free to push whatever to this branch to add it 👍🏻

bentsherman:
Okay that is done. Another idea I had related to this is to not include failed tasks in the percentage and total count. When a task fails and is retried, the percentage increases slightly (e.g. it goes from 3/4 = 75% to 4/5=80%), giving the illusion that the pipeline made progress when it didn't.

Instead, the number of failed tasks should not influence either the total or completed count, they should just be noted to the side as they already are.

bentsherman:
Here's an example process that runs 4 tasks where 1 task fails and retries:
```
[70/0de758] process > bar (3) [ 75%] 3 of 4, retries: 1
```

Then the process shows 100% when the retried task succeeds:
```
[70/0de758] process > bar (3) [100%] 4 of 4, retries: 1 ✔
```

If instead the failed task is ignored, it simply reduces the number of total tasks, allowing the process to reach 100% while still noting the number of ignored tasks:
```
[f2/decd6c] process > bar (4) [ 75%] 2 of 3, ignored: 1
```
```
[98/47ff99] process > bar (4) [100%] 3 of 3, ignored: 1 ✔
```

pditommaso:
Can we keep change in the logic separated from the colouring?

bentsherman:
You mean like a separate PR? Sure

pditommaso:
Thanks

ewels:
@bentsherman are you ok to yank your commits out of this PR again? I'll leave it to you to handle that bit. Shout if you want me to do anything.

bentsherman:
Already done 👍

ewels:
oh nice, thanks!

ewels:
Ok, styling is I think about as far as I want to take it for now. Most of the rest of the TODO is to do with automatically switching ANSI stuff on and off, that might be better in a separate PR.

Ready for you all to tear this PR down and rewrite it as you see fit now 😆  Or any other critique..

CleanShot 2023-12-14 at 22 43 12@2x
CleanShot 2023-12-14 at 22 43 30@2x

bentsherman:
well, step 1 is to fix the tests in `AnsiLogObserverTest` 😄

ewels:
Hoping that `NO_COLOR` and `FORCE_COLOR` support can be handled upstream, rather than doing our own implementation: https://github.com/fusesource/jansi/issues/279

ewels:
Tweaked the colours slightly:

CleanShot 2023-12-15 at 15 03 00@2x

CleanShot 2023-12-15 at 15 03 51@2x


Also tested on a terminal with a light colour scheme:

CleanShot 2023-12-15 at 15 02 47@2x


CleanShot 2023-12-15 at 15 02 23@2x

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -310,7 +314,19 @@ class CmdRun extends CmdBase implements HubOptions {
 
         checkRunName()
 
-        log.info "N E X T F L O W  ~  version ${Const.APP_VER}"
+        if( launcher.options.ansiLog ){
+            log.debug "N E X T F L O W  ~  version ${Const.APP_VER}"
+
+            def fmt = ansi()
+            fmt = fmt.a("\n")
+            fmt = fmt.a(" 🚀  ")
```
pditommaso, line 322: Not I'm ready for this 😆

ewels: You have a favourite emoji you'd like to use instead? 👾 , ✨ or maybe 🙈 ? 😆

But yeah we can of course remove it. Looks pretty awesome with the proper Nextflow green background colour now, so not sure that the emoji really adds much!

ewels:
```suggestion
```

bentsherman:
I figured out how to make the tests pass. I will commit the changes shortly but then @ewels you will need to review the tests to make sure that the expected values match your intent.

ewels:
Awesome!

bentsherman:
Tests look good. However, I would adjust the test case so that it includes the percentages as before.

ewels:
Ok, extended the tests to also test different terminal widths. Extended them all to be full width, and added some test cases across the different break points.

Padding is being stripped in a slightly weird way that I don't really understand, but hopefully this is ok.

ewels:
Just noticed this, which needs cleaning up:

> ` more processes waiting for tasks`

When the pipeline is complete (and there are no more processes waiting)

mribeirodantas:
I'm looking forward to thisssss 😻

pditommaso:
There are some build errors in this PR. Also, the line "more processes waiting for tasks" looks inconsistent. It's shown even when all processes are rendered. 

```
[20/14f848] RNASEQ:INDEX (ggal_1_48850000_49020000) [100%] 1 of 1 ✔
[13/bc29ee] RNASEQ:FASTQC (FASTQC on ggal_gut)      [100%] 1 of 1 ✔
[09/0f40e6] RNASEQ:QUANT (ggal_gut)                 [100%] 1 of 1 ✔
[b1/b2c8bc] MULTIQC                                 [100%] 1 of 1 ✔
 more processes waiting for tasks…

```

  
@ewels You may want to create a new PR against the main repo instead of your own to make simpler to contribute back in the branch.

ewels:
> There are some build errors in this PR.

Looks like that was caused when you pulled upstream changes, which switched out some instances of `Const.APP_VER` to `BuildInfo.version` but not all. Just pushed a fix.

> Also, the line "more processes waiting for tasks" looks inconsistent. It's shown even when all processes are rendered.

Agreed, I spotted the same thing two comments up. I hadn't had a chance to look into it, but I'll try to sort it now.

> You may want to create a new PR against the main repo instead of your own to make simpler to contribute back in the branch.

I can do, but then we'll lose the PR history. You should be able to contribute back to the branch without that - I have the following tick box selected, which is the default when the PR is coming from a fork on a user's account (not from an organisation).

<br>


This should mean that you can check out this PR and push to it directly, even though it's coming from a fork. I personally do this with the [GitHub CLI](https://cli.github.com/), it's super easy as it sets up all of the remotes etc for you automatically:

```bash
gh pr checkout 4573
# make some changes
git commit ...
git push
```

Anyway, happy to push the commits to a branch and open a new PR anyway if you prefer.

ewels:
> Also, the line "more processes waiting for tasks" looks inconsistent. It's shown even when all processes are rendered.

Fixed - was missing squiggly brackets for a multi-line `if`. Too used to Python 😆

pditommaso:
> This should mean that you can check out this PR and push to it directly, even though it's coming from a fork

pditommaso:
I've realised that on older version of `org.fusesource.jansi.Ansi` embedded with [jline](https://github.com/nextflow-io/nextflow/blob/main/modules/nextflow/build.gradle#L44-L44). 

Keeping both on the classpath sounds a very bad thing, and version 2.4 was only used for the `bgRgb` and `fgRbg` methods. Think [this](https://github.com/nextflow-io/nextflow/pull/4573/commits/a39fe9b47ad3686afcd16b753d36769296519e25) should do the same.

modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
```diff
@@ -371,34 +386,75 @@ class AnsiLogObserver implements TraceObserver {
...
+            term.fg(Color.YELLOW).a(Attribute.INTENSITY_FAINT).a(' (').reset()
+            term.fg(Color.YELLOW).a(labelTag)
+            term.a(Attribute.INTENSITY_FAINT).a(')').reset().a(labelSpaces)
+        }
+
+        // No tasks
+        if( tot == 0 ) {
+            term.a(' -')
+            return term
+        }
```
pditommaso, line 429: Well, this code was simple, but now it's totally incomprehensible :cry:

ewels: Yeah - I spent a while trying to add a helper function into the `term` object, but tied myself into Groovy knots and gave up. So it's very verbose as it stands.

The Jansi library doesn't really give any alternative option though. So a custom helper function was the only thing I could think of.

pditommaso: A few comments e.g. what regexs do and similar can help!

modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
```diff
@@ -371,34 +388,78 @@ class AnsiLogObserver implements TraceObserver {
...
-    protected String line(ProgressRecord stats) {
+    private final static Pattern TAG_REGEX = ~/ \((.+)\)( *)$/
+    private final static Pattern LBL_REPLACE = ~/ \(.+\) *$/
+
+    protected Ansi line(ProgressRecord stats) {
+        final term = ansi()
         final float tot = stats.getTotalCount()
         final float com = stats.getCompletedCount()
         final label = fmtWidth(stats.taskName, labelWidth, Math.max(cols-50, 5))
+        final tagMatch = TAG_REGEX.matcher(label)
```
pditommaso, line 402: Inline regex e.g. `~/something/` requires to "compile" the pattern every time. A static variable should be a bit more efficient

modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
```diff
@@ -371,34 +388,78 @@ class AnsiLogObserver implements TraceObserver {
...
+        final labelTag = tagMatch ? tagMatch.group(1) : ''
+        final labelSpaces = tagMatch ? tagMatch.group(2) : ''
+        final labelNoTag = LBL_REPLACE.matcher(label).replaceFirst("")
+        final labelFinalProcess = labelNoTag.tokenize(':')[-1]
+        final labelNoFinalProcess = labelFinalProcess.length() > 0 ? labelNoTag - ~/$labelFinalProcess$/ : labelNoTag
```
pditommaso, line 407: What `labelNoTag - ~/$labelFinalProcess$/` does? just strips `labelFinalProcess` from `labelNoTag`?

modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
ewels: That I can do!

pditommaso: 🙏

modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
ewels: Yup - I'm basically breaking that string into components so that I can colour them differently. I wanted to highlight the final process name (after the last `:`) as that's usually the most interesting bit.

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -372,6 +376,21 @@ class CmdRun extends CmdBase implements HubOptions {
         runner.execute(scriptArgs, this.entryName)
     }
 
+    protected void printBanner() {
+        if( launcher.options.ansiLog ){
+            log.debug "N E X T F L O W  ~  version ${BuildInfo.version}"
+
+            def fmt = ansi()
+            fmt.a("\n")
+            fmt.bg(Color.CYAN).fg(Color.BLACK).bold().a(" N E X T F L O W ").reset()
```
ewels, line 385: @pditommaso this isn't quite the same as the `bgRgb` that was here before. The previous RGB value was the _exact_ Nextflow colour and would appear the same on every terminal - irrespective of what theme is being used. The same for the `fgRbg` with (0,0,0).

The problem with terminal colours is that "cyan" and "black" are interpreted differently by everyone's setup, depending on what theme they use. For example, in a dark mode termainal, "black" is actually... white 👀  By using the exact rgb values it always looks the same and it ensures that it'll be readable.

This is minor, but I wanted to point out why I did this particular bit the way I did (and why it was different to everything else).

ewels: To illustrate this, the output I get now looks like this: (not green and poor text contrast)

CleanShot 2024-02-10 at 19 46 53@2x

Whereas before it looked like this:

CleanShot 2024-02-10 at 19 49 24@2x

pditommaso: Umm, this stuff is little more of a wrapper around an escape code. Considering this is the only place where it's used, is there any chance to find out the code and use it directly?

modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
ewels: Ok, just pushed a load of comments walking through the code. It was a good exercise for myself as well 😅  Hopefully it's a bit clearer now..

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
ewels: Ok - Yup, can do 👍🏻

ewels: Ok, got it to work after a bit of trial and error in 32e3a166885285d77dcb7cd3a71058cc345c201b

Now looks exactly as in the second screenshot above.

modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
ewels: I added a chunk of comments a bit further up with an example, and what each variable will end up with.

modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
pditommaso: Much better now! thanks!

modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
pditommaso: I think the regex is not necessary, have a look at feb0cc54

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
pditommaso: Not working for me. Either we give up with the background color or we may need to use a different library for Ansi colors. [This](https://github.com/dialex/JColor) can be a better choice

ewels:
Ok, I see some difference in the Nextflow header RGB support based depending on which terminal program I use. Mac OSX basically doesn't support the background:

CleanShot 2024-02-12 at 15 37 49@2x

I guess that this is due to Apple Terminal lacking support. I'll try to take a look.

pditommaso:
I suspect they have different escape codes ..

ewels:
I think it's because the Apple Terminal program only supports 256 colours, rather than `truecolor` (16.7 million). See [this blog post](https://medium.com/@skeough117/the-mac-default-terminal-lacks-true-color-capabilities-7ee42eb27aa1), but I've also come across it before in the Python world I remember now.

Basically, Apple Terminal is kind of crap 😅

pditommaso:
Let's do it!

PR #4574, opened 2023-12-12
Fix import of environment variables into container

### LINKED ISSUE #4572: Docker enabled is failing if the directory contains special characters
## Bug report 

Customer reported issue. 
Nextflow docker-enabled run encounters an error when the directory used contains special characters.

To replicate:
1. Clone the project https://github.com/cjw85/wf-fun-times
2. `cd wf-fun-times/fun\(\)times/`
3. `nextflow run main.nf -with-docker ubuntu`

### Program output 

```
ubuntu@ip-x:~/nf/wf-fun-times/fun()times$ ~/nf/nextflow run main.nf
Launching `main.nf` [admiring_pare] DSL2 - revision: c64afdbebf
executor >  local (1)
[74/739307] process > echoStuff [100%] 1 of 1 ✔
/home/ubuntu/nf/wf-fun-times/fun()times/work/74/739307ec8921c28e1d55b21ce8fe4c/stuff.txt

ubuntu@ip-x:~/nf/wf-fun-times/fun()times$ ~/nf/nextflow run main.nf -with-docker ubuntu
Launching `main.nf` [fervent_franklin] DSL2 - revision: c64afdbebf
executor >  local (1)
[5f/09d425] process > echoStuff [100%] 1 of 1, failed: 1 ✘
ERROR ~ Error executing process > 'echoStuff'

Caused by:
  Process `echoStuff` terminated with an error exit status (127)

Command executed:

  echo.sh > stuff.txt

Command exit status:
  127

Command output:
  (empty)

Command error:
  .command.sh: line 2: echo.sh: command not found

Work dir:
  /home/ubuntu/nf/wf-fun-times/fun()times/work/5f/09d4257c29b2256c3071f47749d930

```

### Environment 

* Nextflow version: 23.10.0
* Java version: 17.0.6-tem
* Operating system: Ubuntu

### Additional context

bentsherman:
Close #4572 

Based on the solution suggested by the customer (see Freshdesk ticket)

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -1988,26 +1988,27 @@ class TaskProcessor {
         if( !environment )
             return null
 
-        final List script = []
+        final List exports = []
         for( String name : environment.keySet() ) {
             String value = environment.get(name)
             if( !ENV_VAR_NAME.matcher(name).matches() )
                 log.trace "Illegal environment variable name: '${name}' -- This variable definition is ignored"
             else if( !value ) {
                 log.warn "Environment variable `$name` evaluates to an empty value"
-                script << "export $name=''"
+                exports << "$name=''"
             }
             else if( !escape ) {
-                script << /export $name="$value"/
+                exports << /$name="$value"/
             }
             else {
                 // escape both wrapping double quotes and the dollar var placeholder
-                script << /export $name="${Escape.variable(value)}"/
+                exports << /$name="${Escape.variable(value)}"/
             }
         }
-        script << ''
 
-        return script.join('\n')
+        return exports.size() > 0
+            ? "export ${exports.join(' ')}\n"
+            : ''
     }
 
     final protected int makeTaskContextStage1( TaskRun task, Map secondPass, List values ) {
--- a/modules/nextflow/src/test/groovy/nextflow/executor/BashWrapperBuilderTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/executor/BashWrapperBuilderTest.groovy
@@ -506,10 +506,7 @@ class BashWrapperBuilderTest extends Specification {
         then:
         binding.containsKey('task_env')
         binding.containsKey('container_env')
-        binding.task_env == /\
-                export FOO="aa"
-                export BAR="bb"
-                /.stripIndent()
+        binding.task_env == 'export FOO="aa" BAR="bb"\n'
         binding.container_env == null
 
         when:
@@ -521,8 +518,7 @@ class BashWrapperBuilderTest extends Specification {
         binding.container_env == /\
                 nxf_container_env() {
                 cat << EOF
-                export FOO="aa"
-                export BAR="bb"
+                export FOO="aa" BAR="bb"
                 EOF
                 }
                 /.stripIndent()
@@ -533,7 +529,7 @@ class BashWrapperBuilderTest extends Specification {
         then:
         binding.containsKey('task_env')
         binding.containsKey('container_env')
-        binding.task_env == 'export FOO="aa"\nexport BAR="bb"\n'
+        binding.task_env == 'export FOO="aa" BAR="bb"\n'
         binding.container_env == null
         !binding.launch_cmd.contains('nxf_container_env')
         binding.kill_cmd == '[[ "$pid" ]] && nxf_kill $pid'
@@ -1052,11 +1048,7 @@ class BashWrapperBuilderTest extends Specification {
         binding = newBashWrapperBuilder(environment: ENV).makeBinding()
         then:
         binding.container_env == null
-        binding.task_env ==  '''
-                export FOO="hello"
-                export BAR="hello world"
-                export PATH="/some/path:$PATH"
-                '''
+        binding.task_env == 'export FOO="hello" BAR="hello world" PATH="/some/path:$PATH"\n'
                 .stripIndent().leftTrim()
 
         when:
@@ -1069,9 +1061,7 @@ class BashWrapperBuilderTest extends Specification {
         binding.container_env ==  '''
                 nxf_container_env() {
                 cat << EOF
-                export FOO="hello"
-                export BAR="hello world"
-                export PATH="/some/path:\\$PATH"
+                export FOO="hello" BAR="hello world" PATH="/some/path:\\$PATH"
                 EOF
                 }
                 '''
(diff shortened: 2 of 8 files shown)


bentsherman:
The integration test `env-container.nf` is failing because `$NXF_CMD run foo.nf -with-docker` produces the wrong output. The task wrapper script generates this code for the container environment:
```bash
nxf_container_env() {
cat << EOF
export FOO="HOLA" PATH="\$PATH:/nextflow/tests/checks/env-container.nf/bin" BAR="Hello World!"
EOF
}
```

Which fails for some reason:
```console
$ bash .command.run 
/bin/bash: line 1: export: `World!': not a valid identifier
BAR=Hello
FOO=HOLA
HOME=/root
HOSTNAME=0ee183a3f1a8
NXF_TASK_WORKDIR=/nextflow/tests/checks/env-container.nf
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/nextflow/tests/checks/env-container.nf/bin
PWD=/nextflow/tests/checks/env-container.nf
SHLVL=0
_=/usr/bin/env
Hola mundo
```

And outputs `BAR=Hello` instead of `BAR=Hello World!`. I don't really understand why this error is happening. I have tried quoting things many different ways without success. Here is the context:
```bash
docker run -i --cpu-shares 1024 -e "NXF_TASK_WORKDIR" -v /nextflow/tests/checks/env-container.nf:/nextflow/tests/checks/env-container.nf -w "$NXF_TASK_WORKDIR" --name $NXF_BOXID quay.io/nextflow/bash /bin/bash -c "eval $(nxf_container_env); /bin/bash -ue /nextflow/tests/checks/env-container.nf/work/dc/81c4efb034b065a8ef48448a1af2f0/.command.sh"
```

pditommaso:
Not sure to understand your approach to solve this. 

it looks like the problem arises because the `()` characters are escaped in the container env ie.. 

```
nxf_container_env() {
cat << EOF
export SOMERANDOMCONTENTFORPRERUNSCRIPT="1"
export PATH="\$PATH:/Users/pditommaso/Projects/wf-fun-times/fun\(\)times/bin"
EOF
}

```

Instead it should be 

```
nxf_container_env() {
cat << EOF
export SOMERANDOMCONTENTFORPRERUNSCRIPT="1"
export PATH="\$PATH:/Users/pditommaso/Projects/wf-fun-times/fun()times/bin"
EOF
}

```

pditommaso:
The weird thing, is that removing the `\` escape makes it works only for `(` and `)`. But still does not work for blank and other special char. not sure what's the best way to handle this

bentsherman:
> Not sure to understand your approach to solve this.

To be honest, I don't really understand it either 🤷‍♂️ I don't know why packing the exports onto a single line affects how they are evaluated, it was a solution suggested by the customer that worked, but clearly it is not robust enough

pditommaso:
I didnt' see any difference putting on the same line. What have you observed?

bentsherman:
Try the test case in the linked issue. It fails normally and succeeds when the exports are on a single line

marcodelapierre:
Not sure how much this can help (love these CLI riddles):

How do we pass env variables to docker?  The `-e`/`--env` option seems to be doing a neat job:

```
$ docker run -e CIAO="hello world" quay.io/nextflow/bash env 
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=caf3cb7c82d7
CIAO=hello world
HOME=/root

$ docker run -e CIAO="hello()world" quay.io/nextflow/bash env 
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=ef59ab755b09
CIAO=hello()world
HOME=/root
```

marcodelapierre:
So basically we could adapt `nxf_container_env` to be a one-liner list of `-e` entries, to be added as runtime options to docker/podman. singularity also accepts that if I remember well, probably (?) also apptainer. Not sure about the other engines.

True that the current approach, when fixed, is more robust to new container runtimes.

bentsherman:
I was wondering the same thing, why we don't just mount the environment vars directly, but I figured it was done this way so that you could augment `$PATH` inside the container. Otherwise you would be forcing the `PATH` of the launch environment into the container

marcodelapierre:
Running out of brain juice .. leaving this to you folks until tomorrow.

Please feel free to tingle me on this, happy to spare some extra time tomorrow to sort this out.

marcodelapierre:
Also suggesting to add the round bracket case to the integration tests: https://github.com/nextflow-io/nextflow/pull/4574/commits/b2ade6cd69d468d04e0c0ed0143ca641d46f1ad6

marcodelapierre:
@pditommaso  is there any room for altering the behaviour of the `Escape.variable()` method?  In this context maybe it is escaping too much stuff?

See https://github.com/nextflow-io/nextflow/blob/master/modules/nf-commons/src/main/nextflow/util/Escape.groovy

marcodelapierre:
> > So essentially we should only escape the $ for the env variables as far as I am concerned, and no other special characters. If we then have the double quotes wrapping the variables values we should be good to go.
> 
> Have you tried this? I thought the same but it doesn't look so

I have tested it outside Nextflow but with Docker, by executing this bash script, let's call it `run.sh`:
```
#!/bin/bash

nxf_container_env() {
cat << EOF
export FOO="HOLA"
export PATH="\$PATH:/Users/marco.delapierre/test/special-chars/bin"
export ZIO="Ciao()Mondo!"
export ZIA="Ciao Mondo!"
EOF
}

nxf_launch() {
    docker run --rm quay.io/nextflow/bash /bin/bash -c "eval $(nxf_container_env); env"
}


nxf_launch
```

```
bash run.sh
```
```
HOSTNAME=19cab8a26926
PWD=/
FOO=HOLA
HOME=/root
ZIA=Ciao Mondo!
ZIO=Ciao()Mondo!
SHLVL=0
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/Users/marco.delapierre/test/special-chars/bin
_=/usr/bin/env
```

The idea I wanted to test in Nextflow (though I didn't get the time to action today) was that of a new method for the `Escape` class, similar to [`variable()`](https://github.com/nextflow-io/nextflow/blob/master/modules/nf-commons/src/main/nextflow/util/Escape.groovy#L109-L111), but restricted only to escaping the essential; this would be the dollar sign, and eventually the outer quotes if this is not already taken care of (did not check myself).

pditommaso:
Marco, you are right. Special chars other than `$` do not need to be escaped. Closing in favour of https://github.com/nextflow-io/nextflow/pull/4594

PR #4579, opened 2023-12-13
Build optimizations

bentsherman:
This PR includes two improvements:
- Remove `buildInfo` as a dependency of `compile` and add it as a dependency of `make assemble`, since the build info is only needed when building a release
- Add a make rule `make compileCore` which compiles only Nextflow without the plugins. Useful when testing core features that don't depend on the plugins

The first optimization alone brings my cached build time from ~45s to ~3s

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,12 +38,15 @@ clean:
 	rm -rf plugins/*/build
 	./gradlew clean
 
+compileCore:
+	BUILD_PLUGINS=0 ./gradlew compile exportClasspath
+
 compile:
 	./gradlew compile exportClasspath
 	@echo "DONE `date`"
 
 assemble:
-	./gradlew compile assemble
+	./gradlew buildInfo compile assemble
 
 check:
 	./gradlew check
--- a/Makefile
+++ b/Makefile
@@ -38,9 +38,6 @@ clean:
 	rm -rf plugins/*/build
 	./gradlew clean
 
-compileCore:
-	BUILD_PLUGINS=0 ./gradlew compile exportClasspath
-
 compile:
 	./gradlew compile exportClasspath
 	@echo "DONE `date`"
--- a/settings.gradle
+++ b/settings.gradle
@@ -31,15 +31,13 @@ rootProject.children.each { prj ->
     prj.projectDir = new File("$rootDir/modules/$prj.name")
 }
 
-if( System.getenv('BUILD_PLUGINS') !in ['0', 'false'] ) {
-    include 'plugins'
-    include 'plugins:nf-amazon'
-    include 'plugins:nf-google'
-    include 'plugins:nf-ga4gh'
-    include 'plugins:nf-tower'
-    include 'plugins:nf-console'
-    include 'plugins:nf-azure'
-    include 'plugins:nf-codecommit'
-    include 'plugins:nf-wave'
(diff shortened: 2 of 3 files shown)


pditommaso:
Well spotted. I've move the build num and other info into the BuildInfo class. 👉  ec8083d4

bentsherman:
That's nice but doesn't solve the problem. The key change is to remove `buildInfo` from the `compile` task and invoke it separately. Otherwise it invalidates the build cache, doesn't really matter whether you modify `Const.groovy` or a resource file.

Also this allows you to build the project without changing a file in Git, solving the age-old problem where people submit PRs with extraneous changes from the build.

bentsherman:
Disabling the plugins doesn't actually save much compared to making a change and building. It's about 30s for me with or without the plugins. So we can forget about that part, I just want to separate `buildInfo` from `compile`.

Makefile
```diff
@@ -38,12 +38,15 @@ clean:
 	rm -rf plugins/*/build
 	./gradlew clean
 
+compileCore:
+	BUILD_PLUGINS=0 ./gradlew compile exportClasspath
```
pditommaso, line 42:
```suggestion
	BUILD_PLUGINS=false ./gradlew compile exportClasspath
```

settings.gradle
```diff
@@ -31,13 +31,15 @@ rootProject.children.each { prj ->
...
-include 'plugins:nf-amazon'
-include 'plugins:nf-google'
-include 'plugins:nf-ga4gh'
-include 'plugins:nf-tower'
-include 'plugins:nf-console'
-include 'plugins:nf-azure'
-include 'plugins:nf-codecommit'
-include 'plugins:nf-wave'
-include 'plugins:nf-cloudcache'
+if( System.getenv('BUILD_PLUGINS') !in ['0', 'false'] ) {
```
pditommaso, line 34:
```suggestion
if( System.getenv('BUILD_PLUGINS') != 'false' ) {
```

pditommaso:
Fair enough. I was expecting the resource files to no invalidate the build cache

Makefile
```diff
@@ -38,12 +38,15 @@ clean:
...
+
```
bentsherman, line 43:
```suggestion
```

settings.gradle
```diff
@@ -31,13 +31,15 @@ rootProject.children.each { prj ->
...
+    include 'plugins:nf-amazon'
+    include 'plugins:nf-google'
+    include 'plugins:nf-ga4gh'
+    include 'plugins:nf-tower'
+    include 'plugins:nf-console'
+    include 'plugins:nf-azure'
+    include 'plugins:nf-codecommit'
+    include 'plugins:nf-wave'
+    include 'plugins:nf-cloudcache'
+}
```
bentsherman, line 45:
```suggestion
include 'plugins'
include 'plugins:nf-amazon'
include 'plugins:nf-google'
include 'plugins:nf-ga4gh'
include 'plugins:nf-tower'
include 'plugins:nf-console'
include 'plugins:nf-azure'
include 'plugins:nf-codecommit'
include 'plugins:nf-wave'
include 'plugins:nf-cloudcache'
```

bentsherman:
Okay I removed the plugins part. This is ready to merge

PR #4581, opened 2023-12-13
Add AWS_SESSION_TOKEN when exporting credentials for Fusion

jordeu:
## Description

In some scenarios the `AWS_SESSION_TOKEN` variable is also needed by Fusion to authenticate AWS. 

This PR exports this variable if exists and `fusion.exportStorageCredentials` is set to `true`.


pditommaso:
I think we should into a different approach, exposing creds an env vars is not safe

pditommaso:
Ok, we discussed it. Agreed a temporary solution since it does not alter the existing problem. We need to look for better long-term alternatives

PR #4593, opened 2023-12-15
Fix bug with Fusion symlink resolution

bentsherman:
Follow up to #4348 , based on multiple customer reports that Fusion symlinks are still not being published correctly.

It looks like some last minute changes on the original PR introduced a bug, but the integration test was not robust enough to catch it.

While I'm confident that this PR will fix the bug, I haven't been able to make the integration test fail. Whenever I run `fusion-symlink.nf` locally, the `data.txt` output in each task directory is a copy instead of a symlink, so it is not actually testing the symlink resolution.


pditommaso:
Related #4309

danielecook:
This issue appears to impact GCP Batch workloads / files on GCS.

bentsherman:
Can you share the process definition for a process that is still publishing Fusion symlinks? We have confirmed that directory outputs still aren't working

Issue #4611, opened 2023-12-21
Allow specifying plugins to supply the FUSE device for the Fusion file system


ehfd:
## New feature

Support the usage of various Kubernetes resources (not limited to `nextflow.io/fuse: 1`), including `smarter-devices/fuse: 1` and `github.com/fuse: 1`, `squat.ai/fuse: 1`, and others to mount the `/dev/fuse` FUSE device for using `fusion.privileged = false`.

## Usage scenario 

Currently, the Fusion file system allows non-privileged containers using `fusion.privileged = false`. This currently requires the use of Nextflow's [k8s-fuse-plugin](https://github.com/nextflow-io/k8s-fuse-plugin) for use with Kubernetes. However, FUSE can be mounted using many kinds of Kubernetes device plugins, including the original [fuse-device-plugin](https://github.com/kuberenetes-learning-group/fuse-device-plugin), [smarter-device-manager](https://gitlab.com/arm-research/smarter/smarter-device-manager), and [generic-device-plugin](https://github.com/squat/generic-device-plugin).

This feature is very important for unprivileged multitenant Kubernetes environments including the NSF-funded Expanse Supercomputer, National Research Platform, and the NIH Biowulf cluster.

## Suggest implementation 

We should either allow specifying arbitrary resources to mount the FUSE device or use a mechanic to provide/set up a custom resource for provisioning the FUSE device. I believe that this is mostly a configuration format enhancement for the resulting yaml file in Kubernetes.

ehfd:
CC @dimm0 @davidparks21 and others.

PR #4612, opened 2023-12-21
Add support for custom fuse device plugin

pditommaso:
When Nextflow uses Fusion over K8s with unprivileged mode requires the use of the [k8s-fuse-plugin](https://github.com/nextflow-io/k8s-fuse-plugin) to mount the FUSE device. 

Other plugins are available which requires the use of a custom fuse device name to be specified in the K8s spec. 

This PR adds the ability to specify a custom fuse device name via the nextflow config file. For example: 

```
k8s {
   fuseDevicePlugin = ['smarter-devices/fuse':1] 
}
```

Solved #4611

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1006,6 +1006,11 @@ The following settings are available:
   :::
 : If you trace the hostname, activate this option (default: `false`).
 
+`k8s.fuseDevicePlugin`
+: :::{versionadded} 24.01.0-edge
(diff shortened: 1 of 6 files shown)


ehfd:
Wow, this is less than half an hour, wonderful!

bentsherman (approved):
Looks good, just need to document the new option

PR #4613, opened 2023-12-21
Nextflow parser

bentsherman:
This PR adds a custom parser for Nextflow scripts and uses it instead of the Groovy parser. The Nextflow parser is generated from an ANTLR grammar, which currently contains a subset of Groovy syntax with some additional rules for processes, workflows, and include statements.

To bypass the Groovy parser, we invoke the GroovyShell with a placeholder script that simply wraps the actual script in a string expression. Then in an AST transform, we extract the string value, parse it with the Nextflow parser, and insert the resulting Groovy AST into the placeholder script.

This approach allows us to control the parsing process -- including the syntax and detecting syntax errors -- while still leveraging the Groovy compiler for execution. In other words, we can define whatever grammar we want, as long as we can "compile" it into a Groovy AST. If you look at `AstBuilder`, you'll see that it converts processes / workflows / includes into the same Groovy AST structures produced by `NextflowDSLImpl`.

The hack I'm doing to make this work seems fine but a more robust solution might be to use internal Groovy classes in such a way that allows us to pass our AST directly to the Groovy compiler, instead of going through the GroovyShell and AST transforms. That will take time to understand which components we'll need to rip out. But the advantage is that we don't have to implement our own compiler backend.

I developed this code in a separate project and only just now incorporated it into Nextflow. I haven't tested extensively so there are likely some issues around the edges. Just wanted to finish a basic prototype before the holidays.

TODO:
- [ ] Figure out how to use the generated lexer/parser without manual copy
- [ ] Bring `AstBuilder` to parity with `NextflowDSLImpl`
- [ ] Restore backwards compatibility
- [ ] Pass unit tests and integration tests


bentsherman:
Some sweets-infused holiday thoughts... right now I am just producing the same AST expected by the runtime to keep this PR as simple as possible. But, like I said, we can produce whatever Groovy AST we want, so we could produce Groovy code that more effectively enables new features like static types, default arguments, etc.

The main example I'm thinking of is the annotation API (see https://github.com/nextflow-io/rnaseq-nf/pull/24). I originally designed it as user-facing code, but it could also be an intermediate representation that is produced by the parser. If we "compile" the process and workflow definitions to actual function definitions, then we can more easily leverage the Groovy type checking.

This is just an example. We may not need the annotation API exactly, but it would be good to explore alternative AST representations, perhaps in a second iteration.

bentsherman:
Another aside... GraalVM implements an AST model for every language that it supports. Here is the [Graal Python AST](https://github.com/oracle/graalpython/tree/master/graalpython/com.oracle.graal.python/src/com/oracle/graal/python/nodes) source code.

So we could also have the parser produce a Graal/Python AST and thereby allow the pipeline code to use Python semantics instead of Groovy semantics.

We would need to design a DSL syntax for processes and workflows that would make sense with Python. Likely it would look more like Snakemake. Using native Python syntax (i.e. functions with decorators) is also an option but would likely be more verbose. We would still need to implement our own IDE tooling, but centered around Python syntax instead of Groovy syntax.

The point is, if we rely on the semantics (and compiler backend) of an existing language, it doesn't have to be Groovy. It could easily be any language supported by GraalVM.

pditommaso (commented):
ok, most the comments overlaps with the config parser PR

bentsherman:
Success! Tests are passing 😄 

As a bonus, you can use nextflow inspect to check for errors without running:
```bash
NXF_ENABLE_STRICT_SYNTAX=true nextflow inspect <pipeline>
```

Anyway, let's merge the config parser before this one.

pditommaso (changes requested):
This looks very clean. My only suggestion is to keep `tests` as they were to be sure there's any regression in the parser v1 and create a copy for `tests-v2` to tests the parser v2

bentsherman:
I *really* do not want to maintain two whole copies of the e2e tests

pditommaso:
we don't need to maintain both, we'll add new ones only in `tests-v2`, but I'd prefer to not risk regressions on v1. As a rule of thumb existing stuff should not change.

bentsherman:
Well, if we're going to have a clean break then I would want to review the `tests` folder and see if it can be pruned for the v2 parser

bentsherman:
I have restored a subset of tests that use old syntax patterns. These tests are named `<test>-v1.nf` and `<test>-v2.nf`. The CI tests for both parsers run through the `tests` folder, but the v1 parser ignores the v2 tests and the v2 parser ignores the v1 tests.

I still would prefer to just create a test suite that covers the old syntax patterns more explicitly, rather than copying over tests that happen to use that syntax. If you're concerned about regressions then let's just test for them explicitly in the v1 parser. We can see exactly which syntax patterns were changed.

pditommaso:
Sorry Ben if I'm insisting but I think it is better a no brainer solution having exact the same tests for v1 and new copy for v2 tests. This will make comparisons much simpler just using a a diff across the two directories without caring picking individual files and distinct file names. I don't think this will have any impact from maintenance point of view because we'll update only the version v2 when needed (and those tests are actually changed/updated rarely). It would also help tracking changes across the two different version.

bentsherman:
I do not see it as a no-brainer solution. I don't believe that we will actually get by without having to change the v1 tests at all. If we want to add some new language feature to both v1 and v2 parsers (which we have discussed for e.g. workflow inputs and outputs), or fix some bug in the v1 parser, we must change the v1 tests.

But the bigger problem with your approach is that the v1 parser does not run the updated tests that use the strict syntax. This is an important requirement, that the strict syntax (for now) is purely a subset of the lenient syntax. Once we add new syntax that breaks compatibility, we can add them as `v2-*.nf` so that the v1 parser doesn't run them. But all of the other tests that aren't prefixed with `v1` or `v2` should be tested by both parsers, and there should be only one copy of them.

bentsherman:
Basically, I need to be able to manage three categories of e2e tests:
1. tests that only work with the v1 parser (old syntax)
2. tests that work for both parsers
3. tests that only work with the v2 parser (new syntax in the future)

Most of the tests will fall under (2). These tests are important to verify that there are no unexpected breaking changes in the strict syntax. They should be tested by both parsers, and the should not be duplicated.

The tests for (1) should be clearly marked as "v1 only". In the future, when we add new language features to the v2 parser, those tests will be marked as "v2 only", so that it is clear what the breaking changes are.

We can use a different naming convention if you want, but we can't just duplicate the entire e2e test suite and allow them to diverge. That will just make it harder to verify the differences between the v1 and v2 parsers.

pditommaso:
>  I don't believe that we will actually get by without having to change the v1 tests at all

Why not? as long as there's no change in parser v1 they would run as they are. 

> But the bigger problem with your approach is that the v1 parser would not run the updated tests that use the strict syntax

This is why I want to keep them separated. My point is that as long as the syntax of all tests is updated (even with minor changes e.g. `script:` or variable definition) you are already broke the backward compatibility from a user point of view. 

In my view tests-v1 should be the current ones, tests-v2 should be the version you made (excluding v1-* thing). 

Let me give a try to it in a separate pr

bentsherman:
There is no loss in backwards compatibility. That is ensured by the `v1-*` tests that I added. And it's much better this way because I only need to look at the "v1 only" and "v2 only" tests to see what the breaking changes were, rather than scanning the entire test suite.

pditommaso:
I find much more effective comparing dirs/version with diff. You have an immediate view of the changes

Screenshot 2025-03-06 at 15 12 35

bentsherman:
You shouldn't need to diff anything. Just look at the `v1-*` tests to see what only works with v1, and (in the future) the `v2-*` tests to see what only works with v2.

bentsherman:
Let me know if you like the new test structure. In the meantime, I need to refactor some other parts of the code to enable the full error checking

pditommaso:
I'd like to go with https://github.com/nextflow-io/nextflow/pull/5858 as we discussed. It duplicates some tests but there's no need to optimise in this context.

bentsherman:
Yes that's what I have done

bentsherman:
That was a doozy. Time to summarize what I just refactored:

The v1 script parser loads included modules during script execution. Each new include triggers a separate compilation via ScriptLoaderV1. However, that approach would prevent the new parser from resolving the includes at compile-time.

So I refactored the v2 script parser to recursively compile all included modules in a single compilation. You can see this in ScriptCompiler. It's a bit trickier to make this work with Groovy's CompilationUnit but it gets the job done. Each module is still executed in the same lazy manner as before.

I also renamed ScriptParser -> ScriptLoader to prevent confusion with the various classes that comprise the "script parser":
- `ScriptParser`: parse the source file into an AST (auto-generated by ANTLR)
- `ScriptCompiler`: compile a set of source files into Groovy classes
- `ScriptLoader`: run a script in order to build the workflow DAG (or just load the process/workflow definitions in inspect mode)
- `ScriptRunner`: run a workflow DAG (probably should be called "WorkflowRunner", oh well)

Hope that all makes sense. Let's see how the tests go...

pditommaso:
Should not CompilationUnit correspond to a single compilation file?

pditommaso:
There are a lot of errors reported by the IDE. Consider using Intellij 

Screenshot 2025-03-10 at 12 08 55

pditommaso:
> Let me know if you like the new test structure

Maybe https://github.com/nextflow-io/nextflow/pull/5858 was slightly simpler, but I'm ok with this. However can be avoided to rename test (e.g. removing `-dsl2` suffix). I understand the intent, but it would be preferable reduce the deltas across version (we can rename once we drop the v1 ones)

bentsherman:
> it would be preferable reduce the deltas across version

I renamed the tests in both folders, so they should be the same? What does the directory comparison look like?

pditommaso:
> I renamed the tests in both folders, so they should be the same?

Yes, it would be better 

>  What does the directory comparison look like?

Screenshot 2025-03-10 at 16 24 58

bentsherman:
Are you comparing different branches? My directory compare looks like this (I have a few local changes too):

image

bentsherman:
Regarding the IntelliJ errors, I have IntelliJ open but I am not seeing them. Are they preventing you from building?

IntelliJ's Groovy analyzer does not seem to like these lambdas, though I don't know why, because they are correct

If the IntelliJ errors aren't blocking, what I'd rather do is merge this one and in a separate PR, port ScriptCompiler back to Java, because I know the Java compiler won't have those errors. I just might need to adjust some underlying Groovy classes like BaseScript to make it work

modules/nextflow/src/main/groovy/nextflow/script/parser/v2/ScriptCompiler.groovy
```diff
@@ -0,0 +1,295 @@
...
+/**
+ * Compile a Nextflow script into a Groovy class.
+ *
+ * @see groovy.lang.GroovyShell::parse()
+ * @see groovy.lang.GroovyClassLoader::doParseClass()
+ *
+ * @author Ben Sherman <bentshermann@gmail.com>
+ */
+@CompileStatic
+public class ScriptCompiler {
```
pditommaso, line 55:
```suggestion
class ScriptCompiler {
```

modules/nextflow/src/main/groovy/nextflow/script/parser/v2/ScriptCompiler.groovy
```diff
@@ -0,0 +1,295 @@
...
+
+    public static final String DEFAULT_CODE_BASE = "/groovy/shell"
+    public static final String MAIN_CLASS_NAME = "Main"
+
+    private final CompilerConfiguration config
+    private final GroovyClassLoader loader
+
+    private Compiler compiler
+
+    public ScriptCompiler(Session session) {
```
pditommaso, line 65:
```suggestion
    ScriptCompiler(Session session) {
```

modules/nextflow/src/main/groovy/nextflow/script/parser/v2/ScriptCompiler.groovy
```diff
@@ -0,0 +1,295 @@
...
+
+    public List<SyntaxErrorMessage> getErrors() {
+        if( !compiler )
+            return null
+        return compiler.compilationUnit()
+            .getErrorCollector()
+            .getErrors()
+            .stream()
+            .filter(e -> e instanceof SyntaxErrorMessage)
+            .map(e -> (SyntaxErrorMessage) e)
```
pditommaso, line 119: This looks just to cast the type, find a bit useless

modules/nextflow/src/main/groovy/nextflow/script/parser/v2/ScriptCompiler.groovy
```diff
@@ -0,0 +1,295 @@
...
+        compiler.addSource(su)
+
+        unit.addSource(su)
+        unit.setClassgenCallback(collector)
+        unit.compile(Phases.CLASS_GENERATION)
+
+        // collect script classes
+        final classes = collector.getLoadedClasses().stream()
+            .map((o) -> (Class) o)
+            .filter((c) -> c.getSuperclass() == BaseScript.class)
```
pditommaso, line 141:
```suggestion
            .filter((Class c) -> c.getSuperclass() == BaseScript.class)
```

modules/nextflow/src/main/groovy/nextflow/script/parser/v2/ScriptCompiler.groovy
```diff
@@ -0,0 +1,295 @@
...
+        }
+
+        @Override
+        public void addPhaseOperation(final ISourceUnitOperation op, final int phase) {
+            super.addPhaseOperation((source) -> {
+                // skip main script on second conversion pass
+                if( phase == Phases.CONVERSION && source == entry )
+                    return
+                op.call(source)
+            }, phase)
```
pditommaso, line 210: I think the IDE is correct to report an ambiguity. Even if it compiles it's not clear what method is invoking

pditommaso (changes requested):
It's not even compiling for me 

```
./gradlew compile exportClasspath
Starting a Gradle Daemon, 1 incompatible Daemon could not be reused, use --status for details

> Task :nextflow:compileGroovy
/Users/pditommaso/Projects/nextflow/modules/nextflow/src/main/groovy/nextflow/config/parser/v2/ConfigToGroovyVisitor.java:21: error: cannot find symbol
import nextflow.config.ast.ConfigAppendNode;
                          ^
  symbol:   class ConfigAppendNode
  location: package nextflow.config.ast
/Users/pditommaso/Projects/nextflow/modules/nextflow/src/main/groovy/nextflow/config/parser/v2/ConfigToGroovyVisitor.java:71: error: cannot find symbol
        if( node instanceof ConfigAppendNode ) {
                            ^
  symbol:   class ConfigAppendNode
  location: class ConfigToGroovyVisitor
Note: /Users/pditommaso/Projects/nextflow/modules/nextflow/src/main/groovy/nextflow/sort/BigSort.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
2 errors
startup failed:
Compilation failed; see the compiler output below.
/Users/pditommaso/Projects/nextflow/modules/nextflow/src/main/groovy/nextflow/config/parser/v2/ConfigToGroovyVisitor.java:21: error: cannot find symbol
import nextflow.config.ast.ConfigAppendNode;
                          ^
  symbol:   class ConfigAppendNode
  location: package nextflow.config.ast
/Users/pditommaso/Projects/nextflow/modules/nextflow/src/main/groovy/nextflow/config/parser/v2/ConfigToGroovyVisitor.java:71: error: cannot find symbol
        if( node instanceof ConfigAppendNode ) {
                            ^
  symbol:   class ConfigAppendNode
  location: class ConfigToGroovyVisitor
Note: /Users/pditommaso/Projects/nextflow/modules/nextflow/src/main/groovy/nextflow/sort/BigSort.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
2 errors

1 error
```

Apart this I don't really feel comfortable merging having IDE syntax errors, especially in such central component (compiler)

pditommaso:
After refreshing my copy, I can compile.

modules/nextflow/src/main/groovy/nextflow/script/parser/v2/ScriptCompiler.groovy
bentsherman: It doesn't compile without the type cast 😄

modules/nextflow/src/main/groovy/nextflow/script/parser/v2/ScriptCompiler.groovy
bentsherman: Same here, this does not compile

modules/nextflow/src/main/groovy/nextflow/script/parser/v2/ScriptCompiler.groovy
pditommaso:
```
        // collect script classes
        final classes = (collector.getLoadedClasses() as Collection<Class>)
            .findAll((Class c) -> c.getSuperclass() == BaseScript.class)
            .toList()
```

this works

bentsherman: I would rather not use hard casts or Groovy `findAll`, which copies the list and converts the lambda into a closure

I'm going to leave it as it is. Stream operators are cheap, they don't make an extra copy like Groovy `collect` or `findAll` does

modules/nextflow/src/main/groovy/nextflow/script/parser/v2/ScriptCompiler.groovy
bentsherman: I figured out how to keep the script compiler in Java, so these IntelliJ errors should go away.

I had to re-do some commits to fix the DCO, so you might have to reset and re-pull
```bash
git reset --hard 6cd60d1
git pull
```

bentsherman:
Tests are passing! Approve when you're ready and I'll merge it

pditommaso:
I'll check by today

pditommaso:
i've very confused, I don't see the updates on this branch 

```
» git log 
e6b4048b - (HEAD -> ben-config-parser, origin/ben-config-parser) Minor change [ci fast] (2 weeks ago) <Paolo Di Tommaso>
53565bcf - Rename NXF_ENABLE_STRICT_SYNTAX -> NXF_SYNTAX_PARSER(=v2) (2 weeks ago) <Ben Sherman>
df99b11f - Separate ClosureToStringVisitor from xform (2 weeks ago) <Ben Sherman>
5511f655 - Refactor legacy/impl -> v1/v2 (2 weeks ago) <Ben Sherman>
ca26e26e - Merge branch 'master' into ben-config-parser (2 weeks ago) <Paolo Di Tommaso>

» git log origin/ben-config-parser
- Minor change [ci fast] [e6b4048b]
- Rename NXF_ENABLE_STRICT_SYNTAX -> NXF_SYNTAX_PARSER(=v2) [53565bcf]
- Separate ClosureToStringVisitor from xform [df99b11f]
- Refactor legacy/impl -> v1/v2 [5511f655]
- Merge branch 'master' into ben-config-parser [ca26e26e]
- cleanup [09f1498b]
- Fix Prevent S3 global option when using custom endpoints (#5779) [ed9da469]
- Fix dynamic publish path (#5809) [6a81c015]
```

bentsherman:
It should be `origin/ben-custom-parser`

pditommaso:
Ok, just reverted old tests nomenclature.

PR #4645, opened 2024-01-09
Ignore stored process message when ansi log is enabled

### LINKED ISSUE #4582: storeDir "[skipping] Stored process" messages prevent display of progress bar in console
## New feature

Hi there,
I have been actively developing my Nextflow pipeline for a long time now, and make regular changes to the latter steps. Instead of having to re-run things when a pipeline stage is modified, I rely on storeDir to skip a step if the files for that step are present.

The only downside of using storeDir, is that I get literally hundreds of thousands of unnecessary messages like the following in my console, when all I really want to see is the progress bars.

```
[skipping] Stored process > prep:asvoPrep (1184618352)
```

according to `.nextflow.log`, these messages are at level `INFO`

These messages come in so quickly that the lovely progress bar is pushed far outside my shell's scrollback limit, and I can never see it when doing large jobs.

## Usage scenario 

Nextflow pipelines which rely on storeDir instead of resume for caching of results. e.g.

```nextflow
// nextflow-23.10.0-all run test.nf
nextflow.enable.dsl=2

process CACHEMEIFYOUCAN {
  input: val(i)
  output: path("test_${i}")
  storeDir "."
  script: "touch test_${i}"
}

workflow {
  Channel.from(0..2000) | CACHEMEIFYOUCAN // whatever your terminal scrollback limit is
}
```
gives
```
Launching `test.nf` [deadly_linnaeus] DSL2 - revision: 2e21c16da7                                                                                                                                                                                                                              
[skipped  ] process > CACHEMEIFYOUCAN (53) [100%] 101 of 101, stored: 101 ✔                                                                                                                                                                                                                    
[skipping] Stored process > CACHEMEIFYOUCAN (69)                                                                                                                                                                                                                                               
[skipping] Stored process > CACHEMEIFYOUCAN (38)                                                                                                                                                                                                                                               
[skipping] Stored process > CACHEMEIFYOUCAN (52)                                                                                                                                                                                                                                               
[skipping] Stored process > CACHEMEIFYOUCAN (4)  
...
```

## Suggest implementation 

perhaps a param like

```nextflow
process {
    ignore_stored_process = true

    withName: 'CACHEMEIFYOUCAN' {
        ignore_stored_process = true
    }
}
```

bentsherman:
Close #4582

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
@@ -121,7 +121,9 @@ class AnsiLogObserver implements TraceObserver {
         boolean warn
         if( isHashLogPrefix(message) && !(warn=message.indexOf('NOTE:')>0) )
             return
-        
+        if( message.contains('[skipping]') )
+            return
+
         if( !started || !statsObserver.hasProgressRecords() ) {
             println message
         }
--- a/modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
@@ -121,7 +121,7 @@ class AnsiLogObserver implements TraceObserver {
         boolean warn
         if( isHashLogPrefix(message) && !(warn=message.indexOf('NOTE:')>0) )
             return
-        if( message.contains('[skipping]') )
+        if( message.startsWith('[skipping]') )
             return
 
         if( !started || !statsObserver.hasProgressRecords() ) {


modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
```diff
@@ -121,7 +121,9 @@ class AnsiLogObserver implements TraceObserver {
         boolean warn
         if( isHashLogPrefix(message) && !(warn=message.indexOf('NOTE:')>0) )
             return
-        
+        if( message.contains('[skipping]') )
```
pditommaso, line 124:
```suggestion
        if( message.startsWith('[skipping]') )
```

It would be a bit less "ambiguous"

pditommaso (approved):

PR #4655, opened 2024-01-11
Add NXF_CACHE_DIR environment var

### LINKED ISSUE #4283: Request: Allow setting cache/history/plugin path
## New feature

Please enable setting different paths than `./.nextflow` for the cache/history/plugin folder.

## Usage scenario 

In my experience:

When invoking nextflow from multiple different directories, it leaves multiple '.nextflow' files.
This makes it harder to track them and keep the filesystem clean.

It also means that you have to change to a specific directory to use `nextflow log`.
I guess that the caching then misses too.

## Suggest implementation 

Any one or more configuration options: environment-variable, nextflow-config or commandline would help.

I would prefer having it as an environment variable, along the lines of `NXF_DB_DIR`, because it could be set on an os level, so that nextflow runs can be collected together, and it also eases moving the cache/history from dir to dir or env to env.

i am unsure, but i guess this line is responsible:

[HistoryFile.groovy line40](https://github.com/nextflow-io/nextflow/blob/23c4ec1d551cc55d5c35a70015aa137cfa7d5a56/modules/nextflow/src/main/groovy/nextflow/util/HistoryFile.groovy#L40C51-L40C51)

bentsherman:
Close #4283 

This PR adds a new environment variable `NXF_CACHE_DIR` which controls the location of the default cache store (default `.nextflow` in the launch directory).

Some users want to be able to launch Nextflow pipelines in many different directories but maintain a single `.nextflow` cache that they can query from anywhere with `nextflow .log`.


modules/nextflow/src/main/groovy/nextflow/util/HistoryFile.groovy
```diff
@@ -34,7 +34,7 @@ import nextflow.exception.AbortOperationException
 @Slf4j
 class HistoryFile extends File {
 
-    public static final String FILE_NAME = '.nextflow/history'
+    public static final String FILE_NAME = "${System.getenv('NXF_CACHE_DIR') ?: '.nextflow'}/history"
```
pditommaso, line 37: Let's use `SysEnv.get` and add some tests

Issue #4637, opened 2024-01-07
add docs about methods of GroupKey objects


bguo068:
## Bug report 

This is not a bug report per se. It is mainly a request to add more docs about the GroupKey object.

### Expected behavior and actual behavior

Expected behavior: the key created by calling `groupKey (value, size)` would be better to be converted back to the `value` itself after being used in the `groupTuple` operator. If not, it would be helpful to remind the user that the key after calling the `groupTuple` operator is an object of a special class `GroupKey` and document how to get the original `value` back from the `GroupKey` object.

Actual behavior: the key created by calling `groupKey (value, size)` is not converted back to the `value` itself after being used in the `groupTuple` operator. There is no documentation about the type of the key after calling the `groupTuple` operator. It takes me a while to realize the key is a special class   `GroupKey` (when the grouped channel is `combine`d with a channel with a `by` parameter it won't match with `value`).  I have to look at the source code https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/extension/GroupKey.groovy to figure out how to get the original `value` from the `GroupKey` object.

### Steps to reproduce the problem

```nextflow
a = Channel.fromList([
    ["k1", 1],
    ["k1", 2],
    ["k1", 3],
])

// Not calling getGroupTarget() on groupkey
a_grp1 = a.map{key, value -> [groupKey(key, 3), value]}
        .groupTuple(sort: (a, b) -> a <=> b)
        .map{key, value -> [key, value, "additional info with a"]}

// Calling getGroupTarget() on groupkey
a_grp2 = a.map{key, value -> [groupKey(key, 3), value]}
        .groupTuple(sort: (a, b) -> a <=> b)
        .map{key, value -> [key.getGroupTarget(), value, "additional info with a"]}

b = Channel.fromList([
    ["k1","K1","A"],
    ["k1","K1","B"],
    ["k1","K1","C"],
    ["k1","K2","D"],
    ["k1","K2","E"],
    ["k1","K2","F"],
])

b_grp = b.map{key, KEY, value -> [groupKey([key,KEY], 3), value]}
        .groupTuple(sort: (a, b) -> a <=> b)
        .map{key, value -> [key[0], key[1], value, "additional info with b"]}

// does not work, get 0 item(s)
a_grp1.combine(b_grp, by:0).count().view{"Not calling getGroupTarget() on groupkey: get ${it} item(s)."}

// works, get 2 items
[...]

PR #4656, opened 2024-01-12
Add unwrap option to groupTuple

### LINKED ISSUE #4104: Channel emits differently depending on order of dependent task completion
## Bug report 

A channel may emit all, some, or none of what is expected, depending on the order of completion of upstream tasks. This can be caused when using the `groupTuple` operator in combination with `groupKey`.

It seems that the underlying issue may be related to the output from groupKey not being a string in some contexts (seemingly inconsistent, it works in some places but not in others). This can be remedied by calling `.toString()` on the value output by `groupKey` but it would be great if this emitted a string so channels would behave predictably and consistently across contexts.

Expected behavior for a channel would be that the values emitted would be the same (perhaps in a different order) so long as the input channels emit the same values (again, order should not matter).

There exists an example where this is not the case. Depending on the order of upstream task completions, the channel may emit fewer tuples (without error) than expected, or even none as demonstrated in the example below.

In the below example, this is setup to ensure that the `by_sample` tasks will finish first (`sleep 1`) and `by_batch` tasks will finish after all samples have completed (`sleep 30`). This will create a situation where *nothing* is emitted from the `batch_done` channel (even though we'd expect 2 tuples to emitted, one for each batch).

```groovy
process by_sample {
    input:
    val sample_id

    output:
    val sample_id, emit: sample_ids

    script:
    """
    sleep 1
    echo $sample_id
    """
}

process by_batch {
    input:
    val batch_id

    output:
    val batch_id, emit: batch_ids

    script:
    """
    sleep 30
    echo $batch_id
    """
}

workflow {
    // Input samples, each belonging to one of 2 batches.
    samp_ch  = Channel.from(
        ['SAMP1', 'BATCH1'],
        ['SAMP2', 'BATCH1'],    
        ['SAMP3', 'BATCH2'],
        ['SAMP4', 'BATCH2'],
        ['SAMP5', 'BATCH2'],
    )

    // A process which scatters on sample
    by_sample(samp_ch.map { sample, batch -> sample })
    by_sample.out.sample_ids.dump(tag: 'by_sample')

    by_batch(samp_ch.map { sample, batch -> batch }.unique())
    by_batch.out.batch_ids.dump(tag: 'by_batch')

    // This will create a channel resulting in:
    // tuple(batch_id, num_samples_in_batch)
    sample_counts = samp_ch
        .map { sample_id, batch -> tuple(batch, sample_id) }
        .groupTuple()
        .map { batch, sample_ids -> tuple(batch, sample_ids.size()) }

    sample_counts.dump(tag: 'sample_counts')

    // Using the above counts of samples per batch, we can create a channel
    // which will emit a batch ID when the following conditions are met:
    //   1. All samples in a single batch have completed process "by_sample"
    //   2. The batch has completed the "by_batch process"
    batch_done = by_sample.out.sample_ids
        .join(samp_ch)
        .map { sample, batch -> tuple(batch, sample) }
        .combine(sample_counts, by:0)
        .map { batch, sample, batch_size  -> tuple(groupKey(batch, batch_size), sample) }
        .groupTuple()
        .join(by_batch.out.batch_ids)
        .map { batch, samples -> tuple(batch, "batch_done") }.dump(tag: 'batch_done')
} 
```

### Program output 
If samples finish before batches (as enforced in the example above), we do not see any emitted from the `batch_done` channel:
```
N E X T F L O W  ~  version 23.04.2
Launching `src/bioinformatics/workflows/nextflow/hello_world/main.nf` [goofy_albattani] DSL2 - revision: 6badd408f3
executor >  local (7)
[e0/e080c9] process > by_sample (3) [100%] 5 of 5 ✔
[1c/dafd4a] process > by_batch (1)  [100%] 2 of 2 ✔
[DUMP: sample_counts] ['BATCH1', 2]
[DUMP: sample_counts] ['BATCH2', 3]
[DUMP: by_sample] 'SAMP1'
[DUMP: by_sample] 'SAMP5'
[DUMP: by_sample] 'SAMP2'
[DUMP: by_sample] 'SAMP4'
[DUMP: by_sample] 'SAMP3'
[DUMP: by_batch] 'BATCH2'
[DUMP: by_batch] 'BATCH1'
```

If we switch it up and enforce the batches to finish before samples (i.e. `by_batch`: `sleep 1` and `by_sample`: `sleep 30`), the output looks similar to (notice that now `batch_done` emits a tuple for each batch):
```
N E X T F L O W  ~  version 23.04.2
Launching `src/bioinformatics/workflows/nextflow/hello_world/main.nf` [confident_mercator] DSL2 - revision: 717a98d981
executor >  local (7)
[11/334d03] process > by_sample (4) [100%] 5 of 5 ✔
[d4/c85a77] process > by_batch (1)  [100%] 2 of 2 ✔
[DUMP: sample_counts] ['BATCH1', 2]
[DUMP: sample_counts] ['BATCH2', 3]
[DUMP: by_batch] 'BATCH2'
[DUMP: by_batch] 'BATCH1'
[DUMP: by_sample] 'SAMP2'
[DUMP: by_sample] 'SAMP5'
[DUMP: by_sample] 'SAMP3'
[DUMP: by_sample] 'SAMP1'
[DUMP: batch_done] [BATCH1, 'batch_done']
[DUMP: by_sample] 'SAMP4'
[DUMP: batch_done] [BATCH2, 'batch_done']
```

### Environment 

* Nextflow version: 23.04.2
* Java version: `OpenJDK Runtime Environment AdoptOpenJDK-11.0.11+9 (build 11.0.11+9)`
* Operating system: macOS (but also tested on Linux with same behavior)
* Bash version: zsh 5.9 (x86_64-apple-darwin22.0)

### Additional context

This can also be fixed in a couple other ways which may demonstrate the underlying problem. If we change the channel creation of `batch_done` to the following, we always see the expected output, no matter the order upstream tasks complete in (notice the call to `batch.toString()` in line 7):
```
    batch_done = by_sample.out.sample_ids
        .join(samp_ch)
        .map { sample, batch -> tuple(batch, sample) }
        .combine(sample_counts, by:0)
        .map { batch, sample, batch_size  -> tuple(groupKey(batch, batch_size), sample) }
        .groupTuple()
        .map { batch, samples -> tuple(batch.toString(), samples) }
        .join(by_batch.out.batch_ids)
        .map { batch, samples -> tuple(batch, "batch_done") }.dump(tag: 'batch_done')
```

bentsherman:
Close #4104 and #4637 

I would prefer to be aggressive here and unwrap the group key by default because it will almost always be the desired behavior. But I will leave that decision up to you. We could add the option now and change the default later.

### DIFF
--- a/docs/operator.md
+++ b/docs/operator.md
@@ -692,6 +692,11 @@ Available options:
   - `'deep'`: Similar to the previous, but the hash number is created on actual entries content e.g. when the item is a file, the hash is created on the actual file content.
   - A custom sorting criteria used to order the nested list elements of each tuple. It can be a {ref}`Closure <script-closure>` or a [Comparator](http://docs.oracle.com/javase/7/docs/api/java/util/Comparator.html) object.
 
+`unwrap`
+: :::{versionadded} 24.01.0-edge
+  :::
+: If the key is wrapped with `groupKey()`, unwrap the group key when the group is emitted (default: `false`).
+
 (operator-ifempty)=
 
 ## ifEmpty
--- a/modules/nextflow/src/main/groovy/nextflow/extension/GroupTupleOp.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/extension/GroupTupleOp.groovy
@@ -31,7 +31,13 @@ import nextflow.util.CheckHelper
 @Slf4j
 class GroupTupleOp {
 
-    static private Map GROUP_TUPLE_PARAMS = [ by: [Integer, List], sort: [Boolean, 'true','natural','deep','hash',Closure,Comparator], size: Integer, remainder: Boolean ]
+    static private Map GROUP_TUPLE_PARAMS = [
+            by: [Integer, List],
+            remainder: Boolean,
+            size: Integer,
+            sort: [Boolean, 'true', 'natural', 'deep', 'hash', Closure, Comparator],
+            unwrap: Boolean
+    ]
 
     static private List<Integer> GROUP_DEFAULT_INDEX = [0]
 
@@ -55,6 +61,8 @@ class GroupTupleOp {
 
     private sort
 
+    private boolean unwrap
+
     GroupTupleOp(Map params, DataflowReadChannel source) {
 
         CheckHelper.checkParams('groupTuple', params, GROUP_TUPLE_PARAMS)
@@ -64,6 +72,7 @@ class GroupTupleOp {
         size = params?.size ?: 0
         remainder = params?.remainder ?: false
         sort = params?.sort
+        unwrap = params?.unwrap ?: false
 
         defineComparator()
     }
@@ -150,6 +159,10 @@ class GroupTupleOp {
             sortInnerLists(tuple, comparator)
         }
 
+        if( unwrap ) {
+            unwrapGroupKeys(tuple)
+        }
+
         target.bind( tuple )
     }
 
@@ -234,14 +247,20 @@ class GroupTupleOp {
         return target
     }
 
-    private static sortInnerLists( List tuple, Comparator c ) {
-
+    static private void sortInnerLists( List tuple, Comparator c ) {
         for( int i=0; i<tuple.size(); i++ ) {
-            def entry = tuple[i]
-            if( !(entry instanceof List) ) continue
+            final entry = tuple[i]
+            if( entry !instanceof List ) continue
             Collections.sort(entry as List, c)
         }
+    }
 
+    static private void unwrapGroupKeys(List tuple) {
+        for( int i=0; i<tuple.size(); i++ ) {
+            final entry = tuple[i]
+            if( entry !instanceof GroupKey ) continue
+            tuple[i] = entry.target
+        }
     }
 
     static protected int sizeBy(List target)  {
--- a/modules/nextflow/src/test/groovy/nextflow/extension/GroupTupleOpTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/extension/GroupTupleOpTest.groovy
@@ -17,10 +17,8 @@
 package nextflow.extension
 
 import nextflow.Channel
-
-import spock.lang.Specification
-
 import nextflow.Session
+import spock.lang.Specification
 
 /**
  *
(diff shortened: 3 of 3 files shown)


bentsherman:
Closing in favor of v2 operator library

kgalens:
> Closing in favor of v2 operator library

What does this mean? Is this something I can use to solve the issues I've been seeing with grouped keys?

bentsherman:
The solution for now is to explicitly unwrap the groupKey in your pipeline code. I have a broader solution in mind for permanently addressing issues like this one (and many others) in the operator library.

kgalens:
Is there an issue or anything I might be able to subscribe to in order to get updates on the task? It would be great to be able to track resolutions to issues/topics that we're interested in.

bentsherman:
Unfortunately I don't have a trackable issue for the long-term solution at the moment. But the short-term solution is to unwrap the group key explicitly in your code:

```groovy
        .map { batch, sample, batch_size -> tuple(groupKey(batch, batch_size), sample) }
        .groupTuple()
        .map { batch, samples -> tuple(batch.getGroupTarget(), samples) }
```

kgalens:
Thanks for the suggestion @bentsherman . This is similar to our current solution:
`.map { batch, samples -> tuple(batch.toString(), samples) }`. 

Looking forward to not accidentally missing this in new code that gets committed. I guess I'll just keep checking back in future releases to see when this is no longer required.

robsyme:
Heya @bentsherman. Can I (on behalf on Kevin and other) ask that we reconsider this feature? Even if we leave `unwrap:false` as the default to avoid introducing breaking changes, this argument would both:
1) Improve visibility to this potential to introduce workflow bugs in workflows with two groupTuple operators
2) Simplify the vast majority of workflows that currently have to include a supplementary `map` operation.

bentsherman:
In my view there is little difference between adding a new option and adding a map operation to unwrap the groupKey. Both require the user to find it in the docs, so we might as well just document the solution that already works for all versions of Nextflow

There are a hundred little quirks like this in the operator library that cannot be solved without breaking forwards/backwards compatibility or bloating the operators with more options. Meanwhile, operators are still complicated and annoying to use on a basic level. This is the stage at which it makes more sense to just introduce a new thing (operators v2) and encourage users to migrate to it

Issue #2870, opened 2022-05-13
Allow the concurrent run of multiple pipeline revisions


pditommaso:
### Summary 

Nextflow relies on built-in integration with Git to pull and run a workflow. 

When the user specifies the Git repository URL on then run command line, Nextflow carry out a Git clone command, stores the pipeline code into the `$HOME/.nextflow/assets` directory and launch the execution from there. 

When the user specifies the `-r` (revision) CLI option, the repository is checked out at the specified revision ie. branch, tag or even commit id. 

This however poses a problem when if two or more users run different versions at the same time, because the last performing the operation would override the previous repository code, which could be a disruptive operation. 

This is not such an unlikely event considering a pipeline execution can last for hours or even days. 

To mitigate this problem nextflow refuses to perform a run if the project is currently checkout to a non-default version and the `run` does not specify the revision to be executed in an explicit manner. However, this is the cause of other unexpected side effects. See [here](https://github.com/nextflow-io/nextflow/issues/1670).

#### Goal 

The goal of this enhancement is to allow the concurrent use of multiple pipeline revision in the same computer and deprecated the need for the *stick* revision check. 

This could be achieved by downloading the Git repository with [bare clone](https://www.saintsjd.com/2011/01/what-is-a-bare-git-repository/) instead of a normal clone, and checkout the work tree into a separate subdirectory named as the commit id associated with the specified revision. 

For example, if the user runs 

```
nextflow run https://github.com/nextflow-io/hello
```

nextflow should clone the repo above with the bare option and store in the path `$HOME/.nextflow/assets/nextflow-io/hello.git`

Then implicitly the default branch is checkout, therefore the associate commit should be retrieved e.g. `4eab81bd42eed592f4371cd91b755ec78df25fe9`, therefore the following path should be created containing the work tree accessible for the execution 

```
$HOME/.nextflow/assets/nextflow-io/hello.git/.nextflow/revs/4eab81bd42eed592f4371cd91b755ec78df25fe9
```

When the user-specified a different revision e.g. 

```
nextflow run https://github.com/nextflow-io/hello -r dev
```

A new subdirectory with the corresponding commit id should be created. 

The commit id should be resolved against the local git clone, unless the `-latest` option is specified.

pditommaso:
Good point. If already exists think should report a warning message maybe?

jorgeaguileraseqera:
so, report with a warning message (maybe with some instructions to remove the current repo) and stop the command, right?

[...]

Issue #2655, opened 2022-02-17
git checkout of pipeline can be changed mid-execution


notestaff:
## Bug report 

If I start a nextflow run on a particular git revision of a pipeline, and then, while this run is executing, do a nextflow pull  or start another nextflow run on a different revision, the pipeline checkout used by the still-executing original run (under ~/.nextflow/assets/scm) will be overwritten, possibly leading to wrong results.    Instead, nextflow should create a separate checkout for each git revision (perhaps using git worktree) for which a pipeline run is started, and keep that checkout immutable throughout the pipeline's run.

pditommaso:
This is a good point. Ideally, this should require having Nextlow to clone the repo as a [bare repository](https://www.saintsjd.com/2011/01/what-is-a-bare-git-repository/). 

Then, when the execution is launched it should check it out into a sub-directory named as the requested versions

leonorpalmeira:
I would have also expected `main.nf` to be read once and kept into RAM (or copied on disk into the execution directory) while nextflow executes the pipeline. Instead, I experience the same issue as above in the following use case:

```
cd MyGitRepo/
git checkout master
nextflow main.nf
git checkout newFeatureBranch
nextflow main.nf
git checkout otherFeatureBranch
```

The execution of the sub-processes will execute the state in which the code is at the time where those processes are run, not the state in which they were when the pipeline was launched. I may be wrong but it seems that the only solution is to have one working repository per execution?

bentsherman:
Seems to be a duplicate of #2870

PR #4659, opened 2024-01-15
Multiple revisions 1st: AssetManager, Cmd Clone Pull Run

marcodelapierre:


bentsherman:
You could save disk space by cloning with depth = 0 when a revision is specified

marcodelapierre:
> You could save disk space by cloning with depth = 0 when a revision is specified

Thanks @bentsherman , great idea, will do, for `nextflow pull` but not for `clone`. Why not for any case, including no revision specified?

bentsherman:
yes I suppose we could just clone with depth = 0 in all cases, since no revision just defaults to `master`. the `clone`, `pull`, and `run` commands all provide a `-deep` option for this, so users can change it if they want

marcodelapierre:
Ah! unfortunately we cannot implement the `deep` functionality, as with a depth of 1 we cannot checkout to other branches/tags/commits.  
This would be supported : `git clone -b <branch/tag> --deep <deep>  <repo>` , but only works for branches and tags, not commits. Git still does not support a way to clone to a specific commit straight away.

bentsherman (approved):
Works on my end. These changes are pretty straightforward.

Initially I thought that no revision should default to `master`, but with the current approach it should use the default branch of the git repository without needing to query the repository provider, so I think this behavior is worth keeping

marcodelapierre:
Thanks for the review @bentsherman ! 

@pditommaso   ready  for you

pditommaso:
This implementation is slick and effectively handles each revision as an independent project. however it introduces some minor inconsistencies.

For example I could do 

```
nextflow run nextflow-io/hello -r testing 
```

Later on I could use the command `info` to look all revisions of this pipeline, but it would report an error 

```
» nextflow info nextflow-io/hello 
Unknown project `nextflow-io/hello`
```

Along the same line with `drop`, which was meant to delete all revisions. 


I wonder if, at this point, we should handle the revision as part of the project name, e.g. and use this rule consistently across all commands. 


```
nextflow run nextflow-io/hello@testing 
```

This would make more explicit that the command is referring to a very specific version. Also it may be a bit more intuitive when list the project e.g. 

```
nextflow list 
```

would report 

```
nextflow-io/hello@master
nextflow-io/hello@testing
nextflow-io/hello@v1.2
```


Thoughts?

pditommaso:
Another option could to make the command `info` *smarter*, so that it tries to pick the first  revision that's checkout out and pull a fresh copy of the master automatically. 

Also, `list` may arrange available projects differently, e.g.  

```
nextflow/rnaseq-nf
nextflow-io/hello
  @testing 
  @v1.2
```

marcodelapierre:
Thank you Paolo, very good points!

I want to comment on two in particular:
- inconsistencies: one root consistency I noticed when working on the PR is the different status of the pipeline without revision input vs all the others
  - one way I think we should try and address it is to have Nextflow make the match between the revision-less pipeline and its actual revision name
- default layout: grouping vs not grouping projects by pipeline name

More details coming later..

marcodelapierre:
Following NF meeting:

1) `nextflow list` to only show repos without revisions ; add mark for projects with multiple downloaded revisions
2) `nextflow info` to show all revisions of a repo ; downloaded revisions show the corresponding path

3) need to get rid of duplication of default branch; want to do it dynamically, without assuming that `master` is default (which is often not the case)

For 3),  I was pointing out that `git` has no way of querying info on the master branch from the remote repo; the Github API has a way, need to check for Gitlab, Bitbucket...

marcodelapierre:
On the default branch point above, I may have found the Jgit way to probe it: https://stackoverflow.com/questions/50481123/get-the-default-branch-of-a-remote-repository-with-jgit#50482486

pditommaso:
In principle it should be defined in the config file 

https://github.com/nextflow-io/nextflow/blob/9d6dd6a06707034e7b761de5a7aed85177e9ed02/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy#L428-L430

marcodelapierre:
Yes, good point. 

However the current fallback for an undefined config master is the hard-coded `master`.  I would like to see if this can instead be sourced dynamically.

bentsherman:
Here's a hacky solution I thought of:
1. Clone the repo with no revision specific
2. Get the current branch from the local repo
3. Rename the repo directory accordingly

marcodelapierre:
> Here's a hacky solution I thought of:
> 
> 1. Clone the repo with no revision specific
> 2. Get the current branch from the local repo
> 3. Rename the repo directory accordingly

Yes, good point.  And, in addition, for the `defaultBranch`, also add some sort of `:DEFAULT` suffix to the repo path, so that NF can tell when the default branch is already present locally.

pditommaso:
Note the method [getManifest](https://github.com/nextflow-io/nextflow/blob/a7f233053cc18b17192ede4c869ea252f23bc8d8/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy#L437-L464) is already able to fetch get the manifest from the remote repo if it has not yet been download.

bentsherman:
That has always seemed circular to me because you have to know the default branch to get the manifest which defines the default branch...

marcodelapierre:
Indeed Ben   ... I was banging my head on the circularity right now  😂

will find a way

pditommaso:
I agree it can be done better, tho doesn't look directly related to this PR (time to close it)

marcodelapierre:
Summary as of today:

- [x] `nextflow list` to only show repos without revisions ; add mark for projects with multiple downloaded revisions ; add `-a/-all-revisions` option to also show local revisions
- [x] `nextflow drop` to add `-a/-all-revisions` option to delete all revisions of given project

- [ ] `nextflow info` to show all revisions of a repo ; downloaded revisions show the corresponding path (quick one)

- [ ]  get rid of duplication of default branch (small bit yet to figure out)

- [ ] documentation for 2 new options above ; unit test for `listRevisions` method in `AssetManager`

marcodelapierre:
Note on this addition: `Get rid of duplication of default branch`

This only works properly in one of these 2 cases:
- the pipeline project has a manifest which specifies the  `defaultBranch`
- the pipeline project default branch coincides with the `DEFAULT_BRANCH` (currently `master`)

On the other hand, if the undeclared default branch differs from `master`:
- pulled default branch may be duplicated (i.e. both pull without revision, and pull with default revision)
- project information on default branch via `info` will be incorrect

marcodelapierre:
@pditommaso apart from the above caveat on undeclared non-standard default branch,  this is now ready for your review.

marcodelapierre:
Some examples to facilitate review.

Nextflow List
```
$ nextflow  list
 nextflow-io/hello
 marcodelapierre/toy-acacia-nf
 marcodelapierre/toy-cowsay-nf

$ nextflow  list -a
 nextflow-io/hello
   toy
   v1.1
   (default)
   v1.2
 marcodelapierre/toy-acacia-nf
   main
 marcodelapierre/toy-cowsay-nf
   (default)
```

Nextflow Drop
* no flags: deletes default revision of specified project
* `-r <rev>`: deletes <rev> revision of specified project
   * if `<rev>` is the default revision, redirects to it
* `-a`: deletes all revisions of specified project

Nextflow Info   
Notes: 
* revision selection allowed (local path and main script are/may be specific to revision)
    * if no revision specified and default revision not pulled locally, print info on first found revision
    * if non locally pulled revision specified, stops with error (as any other subcommand related to proects)
* `*` indicates selected revision in CLI (has to be local)
* `P` indicates all locally pulled revs
* `(default)` indicates default revision
```
$ nextflow info nextflow-io/hello
 project name: nextflow-io/hello
 revision    : master
 repository  : https://github.com/nextflow-io/hello
 local path  : /Users/marco.delapierre/test/add-multirev/HOME_PR/assets/nextflow-io/hello
 main script : main.nf
 revisions   : 
 * P master (default)
     mybranch
     testing
   P toy
   P v1.1 [t]
   P v1.2 [t]

# non default branch selection
$ nextflow info -r toy nextflow-io/hello
 project name: nextflow-io/hello
 revision    : toy
 repository  : https://github.com/nextflow-io/hello
 local path  : /Users/marco.delapierre/test/add-multirev/HOME_PR/assets/nextflow-io/hello:toy
 main script : main.nf
 revisions   : 
   P master (default)
 * P toy
     mybranch
     testing
   P v1.1 [t]
   P v1.2 [t]

# default branch selection -> redirects to it
$ nextflow info -r master nextflow-io/hello
 project name: nextflow-io/hello
 revision    : master
 repository  : https://github.com/nextflow-io/hello
 local path  : /Users/marco.delapierre/test/add-multirev/HOME_PR/assets/nextflow-io/hello
 main script : main.nf
 revisions   : 
 * P master (default)
     mybranch
     testing
   P toy
   P v1.1 [t]
   P v1.2 [t]

# no branch selected, only non default branches locally available -> defaults to first found (main in this case)
$ nextflow info marcodelapierre/toy-acacia-nf
 project name: marcodelapierre/toy-acacia-nf
 revision    : main
 repository  : https://github.com/marcodelapierre/toy-acacia-nf
 local path  : /Users/marco.delapierre/test/add-multirev/HOME_PR/assets/marcodelapierre/toy-acacia-nf:main
 main script : main.nf
 revision    : * P main
```

All commands (including info, run, ..):
* default revision redirection requires either default (`master`) naming of specification of `defaultBranch` in Manifest, otherwise local pull duplication (fair enough)
* **CHECK THIS** when default branch non locally pulled, every interaction with other local branches of same project will require internet connection to retrieve default project manifest on Git repo
* **CHECK THIS** inconsistencies may arise for a project where `defaultBranch` is updated often - unlikely??

marcodelapierre:
(note : last CI failed due to ... Quay.io not responding to usual container requests - sigh)

modules/nextflow/src/main/groovy/nextflow/cli/CmdClone.groovy
```diff
@@ -15,6 +15,9 @@
  */
 
 package nextflow.cli
+
+import static nextflow.scm.AssetManager.revisionDelim
```
pditommaso, line 19: By convention constanta are upper case and separated by `_`

modules/nextflow/src/main/groovy/nextflow/cli/CmdClone.groovy
```diff
@@ -68,9 +72,9 @@ class CmdClone extends CmdBase implements HubOptions {
         }
 
         manager.checkValidRemoteRepo()
-        print "Cloning ${manager.project}${revision ? ':'+revision:''} ..."
-        manager.clone(target, revision, deep)
+        print "Cloning ${manager.project}${revisionSuffix} ..."
```
pditommaso, line 75: Considering the `AssetManager` knows the `revision` it's worth replacing `${manager.project}${revisionSuffix}` with a small helper method like `manager.getFullQualifiedProjectName()` which reports the project plus the revision

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
```diff
@@ -542,16 +565,39 @@ class AssetManager {
...
         list().each {
             def items = it.split('/')
-            if( items[1] == name )
-                exact << it
-            else if( items[1].startsWith(name ) )
-                partial << it
+            def itemsRev = items[1].tokenize(':')
+            if( (!revision && !itemsRev[1]) || (revision && itemsRev[1] == revision) ) {
+                if( itemsRev[0] == name )
+                    exact << it.tokenize(':')[0]
```
pditommaso, line 597: This `:` should not be `revisionDelim` ?

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
```diff
@@ -542,16 +565,39 @@ class AssetManager {
...
+                else if( itemsRev[0].startsWith(name ) )
+                    partial << it.tokenize(':')[0]
```
pditommaso, line 599: Some comments in this block would be useful to understand the overall logic

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
```diff
@@ -737,30 +793,32 @@ class AssetManager {
...
+     *     * P master (default)
+     *         patch-x
+     *         v1.0 (t)
+     *         v1.1 (t)
      * </pre>
      *
-     * The star character on the left highlight the current revision, the string {@code (default)}
-     *  ticks that it is the default working branch (usually the master branch), while the string {@code (t)}
-     *  shows that the revision is a git tag (instead of a branch)
+     * The star character on the left highlights the selected revision,
```
pditommaso, line 802: I think the star does not have any more sense because in this model there isn't anymore an "active" revision

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
```diff
@@ -94,39 +104,52 @@ class AssetManager {
...
-    AssetManager build( String pipelineName, Map config = null, HubOptions cliOpts = null ) {
+    AssetManager build( String pipelineName, String revision = null, Map config = null, HubOptions cliOpts = null ) {
+
+        // if requested revision corresponds to the default branch, then unset it
+        // this avoids duplication of the default branch
+        if ( revision ) {
+            def referenceManager = new AssetManager(pipelineName, null, cliOpts)
+            if ( revision == referenceManager.getDefaultBranch() ) {
+                revision = null
+            }
```
pditommaso, line 142: A nested `AssetManager` creation smell very bad. Why not calling directly `this.getDefaultBranch()` ?

pditommaso:
Looking more into this I think this approach is flawed. Consider this scenario in which two users share the same pipeline asset directory: 

1. User A pull a pipeline from the master and launch a run 
2. User B upload a pipeline change in the Git repository on master and wants to run it. 
3. User B run the pipeline using the flag `-latest` 
4. As result the version for user A is altered

bentsherman:
Seems more like a pitfall of using branches instead of a fixed revision like a tag or commit. Especially if multiple users are sharing the same NXF_ASSETS directory, which is not very common.

pditommaso:
It's not common but [happens](https://github.com/nextflow-io/nextflow/issues/2655). We have this same problem in the Platform as well. 

This is why the [original ticket](https://github.com/nextflow-io/nextflow/issues/2870) was planning the checkout the repo as a bare repository, and then checkout the branch into a path named as the commit it associated the branch

bentsherman:
The problem is really with using a branch which can map to any number of revisions. It can be avoided by using a commit or tag.

pditommaso:
This does not fit the user expectation

bentsherman:
Maybe we should change their expectation. It is well known that you shouldn't use `latest` tag of a container if you care about reproducibility because it can change at any time. Why not promote the same best practice for the git repos -- "if you run pipelines in a shared environment, avoid using branches in favor of commits / tags because branches can be changed by other users"

Or we could append the commit hash to the directory name if a pipeline is checked out to a particular revision, i.e. `.nextflow/assets/<org>-<project>-<branch>-<commit>`

modules/nextflow/src/main/groovy/nextflow/cli/CmdClone.groovy
marcodelapierre: thanks, fixed

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
marcodelapierre: thanks, fixed

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
marcodelapierre: comments added

modules/nextflow/src/main/groovy/nextflow/cli/CmdClone.groovy
marcodelapierre: added and used method `getProjectWithRevision()`

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
marcodelapierre: thanks, let's review this after the design points below are sorted out

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
marcodelapierre: To use `this.getDefaultBranch` we still need to mock a number of things, so that we can then get the manifest from the provider (or at least, this is the best solution I could find):
```
        this.providerConfigs = ProviderConfig.createFromMap(config)

        // these four are temporary, just to get the default branch
        this.project = resolveName(pipelineName, null)
        this.localPath = checkProjectDir(this.project, null)
        this.hub = checkHubProvider(cliOpts)
        this.provider = createHubProvider(hub)

        // if requested revision corresponds to the default branch, then unset it
        // this avoids duplication of the default branch
        if ( revision && revision == this.getDefaultBranch() ) {
            revision = null
        }

        // now the proper initialisation
        this.revision = revision
        this.project = resolveName(pipelineName, this.revision)
        this.localPath = checkProjectDir(this.project, this.revision)
        this.hub = checkHubProvider(cliOpts)
        this.provider = createHubProvider(hub)
        this.provider.setRevision(this.revision)

        [...]
```

I went for the nested creation in the PR for conciseness basically. Happy to move to the alternate solution above. What are your thoughts @pditommaso ?

marcodelapierre:
I agree with Ben, we need to make expectations explicit, and remind that sharing the same NF repo collection between multiple people may result in not running the expected branch commit.  

Even Paolo's solution with explicit commit id in the path does not solve the situation Here's the case that breaks it IMO.
- Person A pulls a pipeline branch, commit 1, which is shared also by person B. 
- Days and commits later
- Person A updates that branch, new implicit path created with commit 5
- Days and commits later
- Person B updates that branch, new implicit path created with commit 10

And now in this new situation, how do people get to use the version "they think they are using"? All of the implicitly requested commits are available, but how do you reliably pick one or the other for usage? I don't see an implicit way to achieve this.  

The simplest way (for the implementation and also, crucially, for the users) is to advise them to use tags or commits if they really care about reproducibility.

So my suggestion is: let's keep this implementation, and add a paragraph to the docs to highlight these reproducibility caveats (doing a commit shortly).

marcodelapierre:
These are actually relatively important caveats to clarify, because I have heard multiple instances of site (cluster/service) administrators willing to curate a collection of pipelines for their users, including multiple revisions.  
In this case, the admin is the person that might change revisions under the hood to all users; in this case, though, it seems quite clear to me that it is up to the admin to communicate with their users about updates to the curated list of pipelines/revisions.

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
marcodelapierre: ok, on this one now

marcodelapierre: OK, done:
```
$ nextflow info hello
 project name: nextflow-io/hello
 repository  : https://github.com/nextflow-io/hello
 local path  : /home/ubuntu/add-worktree/HOME_PR/assets/nextflow-io/hello
 main script : main.nf
 revisions   : 
 P master (default)
   mybranch
   testing
 P toy
   v1.1 [t]
 P v1.2 [t]
```

`nextflow info <project` now:
- does not take a revision parameter
- always show `local path` for the default revision (also documented in the docs)
- does not show the sticky `*` for current revision, instead shows `P` for locally pulled revisions

@pditommaso

marcodelapierre:
I have been studying and making experiments using a local bare repository, as a local intermediate between the remote repo and the various checked out revisions. 

In my first round of attempts, I was trying to use the local bare repo as the source of truth for any local checkouts, in substitution for the remote repo. Essentially, the idea was:
- clone the bare from the remote, only update when needed
- otherwise, for most other operations use the local bare as the source of truth

This does not work at present, as the general assumption in AssetManager methods is that the "source of truth" has everything available, i.e. both repo information and file contents. However a bare repo lacks the latter.

Hence, I started an attempt of targeted usage of either bare or remote, depending on the required information. This not only is confusing, but currently fails in my tests, again because of assumptions that cross each other around the existence and contents of the various types of repos.

At the end of today:

To bring the effort to a more manageable and productive ground, I have decided to move forward with a radically simplified approach:
- keep using the remote repo as much as possible
- only use the local bare repository to map requested revisions into commit IDs

This approach, if workable, tackles the main goal of this round of PR revision: to locally use commits to store and run pipelines, to avoid clashes when updating repositories.

marcodelapierre:
I have started to add localBarePath. Scope is:
- to dereference branches/tags into commit IDs
- to source the local git config file

Ok, I have fixed the chicken and egg at AssetManager instantiation, between defining localPath and devining the repository provider.  Using localBarePath has been instrumental.

Next:
- actual dereferentiation to commits
- effective update of bare repo, on demand and transparent (i.e. only when download or run with latest)

marcodelapierre:
@pditommaso Any tips on how to get a commitID for a branch or tag, I am a taker.   
Otherwise, I will go into it next week.

pditommaso:
Not sure how much I can help this week. @bentsherman you want to have look at this

marcodelapierre:
thanks Paolo. did good progress myself.

marcodelapierre:
Good progress today.

Notes on revision dereferencing (working)
            - key attributes are localPath and commitId
            - do it at instantiation
            - if cmdpull/cmdrun, use updateLocalBareRepo() to update the attributes
            - also use updateLocalBareRepo() if revision cannot be found locally

            Let's simplify and minimise usage of the local bare
            0. get local gitconfig
            1. dereference revisions (branches!/tags?) to commits
            2. use commits in localpath
            3. use commits in CLI Cmds
            4. use commits for download()
            5. use local bare path just to dereference locally without need for remote repo, and for local gitconfig
            6. also use local bare for download() of revisions/commits

marcodelapierre:
On point 2. above, wondering whether we should leverage user-provided information more, when it comes to printing out information on repository in use.  I.e.:
- commit ID as de-referenced from user request
- (if not null) revision as requested by the user
- (eventually, also infer branch)

marcodelapierre:
Have addressed point 3. above with a fix

marcodelapierre:
As dicussed in meeting, on point 1., let me re-implement with the original class signature for `AssetManager` - no revision attribute.  Commit ID  and hence local path will be set when the revision is set.


Note on available code so far:
1. current branch for this PR is `add/mult_revisions` - impl with local bare repo, and commit id in local path
2. branch `add/mult_revisions_bkp15apr` - first impl, uses revision in local path
3. branch `add/mult_revisions_bkp29may` - backup of 1.

marcodelapierre:
Going forward, let's manage these through 2 separate new PRs:
1. https://github.com/nextflow-io/nextflow/pull/5087 : Jan-April implementation, using revision name in repo localPath
2. TBD : June implementation, using commit ID in repo localPath

The first one is for the record and for reference, unlikely to go forward.

The second one should be considered for revision and merging.

bentsherman:
Closing in favor of #6620

PR #4663, opened 2024-01-16
LocalSecretsProvider: invalid permissions tests made more robust

marcodelapierre:
This PR fixes a unit test failure that happens if the machine you run on has tight default file permissions (e.g. as obtained via `umask`, for instance `umask 067`).   

The error reads:
```
LocalSecretsProviderTest > should fail with invalid permissions FAILED
    org.spockframework.runtime.WrongExceptionThrownError at LocalSecretsProviderTest.groovy:114
```

To succeed, this tests is expected to raise an exception as a result of the permissions of a created file being different from `rw-------`.  If this configuration is the default in the machine (more secure),  the test will fail. 

The PR explicitly sets the file permissions for the test to something different, i.e. `rw-r-----` , instead of relying on the machine default.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -4,7 +4,7 @@ NEXTFLOW CHANGE-LOG
 - Fix bug with Fusion symlink resolution (#4593) [f28c9e48]
 - Fix Fusion symlinks when publishing files (#4348) [1fa5878a]
 - Fix Inspect command fails with Singularity [25883df3]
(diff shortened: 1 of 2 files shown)


pditommaso:
What this fix? (please fill out the description)

marcodelapierre:
sure, doing soon - thought you had seen the thread on Slack

pditommaso:
Nice

PR #4675, opened 2024-01-18
Set scratch default to false when using Fusion

bentsherman:
I spoke to Jordi and found that we originally defaulted scratch to true with Fusion at a time when Fusion was only staging input data. But now that Fusion handles both inputs and outputs staging, it makes no sense to ever use scratch=true with Fusion because it eliminates the performance benefit of Fusion. I think we can safely set the default to false now.


pditommaso:
Agreed

PR #4678, opened 2024-01-19
Update Wave Client tests for Spack: add OS packages 

marcodelapierre:
This is a matching PR for the Libseqera PR    https://github.com/seqeralabs/libseqera/pull/10.   
See there for rationale.


marcodelapierre:
I reckon that PR fails because tests on the Dockerfile/Singularityfile compare the updated provided text string with that generated by the released `libseqera`, which does not implement the changes yet (see https://github.com/seqeralabs/libseqera/pull/10).

pditommaso:
Dependencies should be added via Spack packages

Issue #4585, opened 2023-12-14
Text alignment of multi-line error messages is not aligned.


mahesh-panchal:
## Bug report 

This is a small thing, but the error messages from multiline error messages are not aligned correctly when printed.

### Expected behavior and actual behavior

The expectation is that the multiline messages are aligned correctly, for example, so `assert` points to the correct variable.
Also the source block script alignment is also off.
Presumably these could be fixed by using a `.stripIndent()` on the script text, and a `.replace('\n', '\n<indent>')` on the multi-line error message.

### Steps to reproduce the problem

```nextflow
workflow {
    TASK(Channel.of('A'..'C','a')).view()
}

process TASK {
    input:
    val letter

    script:
    assert letter.isUpperCase()
    """
    echo $letter
    """

    output:
    stdout
}
```

### Program output 

```console
$ nextflow run main.nf 
N E X T F L O W  ~  version 23.10.0
Launching `main.nf` [nostalgic_curie] DSL2 - revision: 458e0253ae
[-        ] process > TASK [  0%] 0 of 3
ERROR ~ Error executing process > 'TASK (4)'

Caused by:
  assert letter.isUpperCase()
       |      |
       'a'    false -- Check script 'main.nf' at line: 10

Source block:
  assert letter.isUpperCase()
  """
      echo $letter
      """

Tip: you can try to figure out what's wrong by changing to the process work dir and showing the script file named `.command.sh`

 -- Check '.nextflow.log' file for details


```

### Environment 

* Nextflow version: 23.10.0
* Java version: OpenJDK 64-Bit Server VM (build 17.0.9-internal+0-adhoc..src, mixed mode, sharing)
* Operating system: Linux (Gitpod)
* Bash version: GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)

### Additional context

(Add any other context about the problem here)

PR #4681, opened 2024-01-23
Fix error text alignment

mahesh-panchal:
Fixes https://github.com/nextflow-io/nextflow/issues/4585

## Input output

Using this toy:
```nextflow
workflow {
    TASK(Channel.of('a')).view()
}

process TASK {
    input:
    val letter

    script:
    assert letter.isUpperCase()
    def puzzle = null
    if ( puzzle ) {
        println "Puzzled"
    } else {
        println "LGTM"
    }
    """
    echo $letter
    """

    output:
    stdout
}
```
I now get a properly formatted error:
```
N E X T F L O W  ~  version 23.12.0-edge
Launching `../main.test.nf` [spontaneous_galileo] DSL2 - revision: 81c47780ef
[-        ] process > TASK -
ERROR ~ Error executing process > 'TASK (1)'

Caused by:
  assert letter.isUpperCase()
         |      |
         'a'    false -- Check script '/workspace/main.test.nf' at line: 10


Source block:
  assert letter.isUpperCase()
  def puzzle = null
  if ( puzzle ) {
      println "Puzzled"
  } else {
      println "LGTM"
  }
  """
  echo $letter
  """

```

## Additional notes

`task.source` seems to be formatted strangely such `stripIndent` doesn't work. As soon as it encounters indented code, all indentation is preserved. E.g.
```nextflow
    script:
    if (meta.id) {
        args = "-f $meta.id"
    } else {
        args = "-f $bam.baseName"
    } 
```
becomes:
```nextflow
if (meta.id) {
        args = "-f $meta.id"
    } else {
        args = "-f $bam.baseName"
    } 
```

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
@@ -471,9 +471,9 @@ class NextflowDSLImpl implements ASTTransformation {
                         }
                         body.add(stm)
                 }
-
-                readSource(stm, source, unit)
             }
+            // read the closure source
+            readSource(closure, source, unit)
 
             final bodyClosure = closureX(null, block(scope, body))
             final invokeBody = makeScriptWrapper(bodyClosure, source.toString(), 'workflow', unit)
@@ -759,21 +759,35 @@ class NextflowDSLImpl implements ASTTransformation {
                 final line = unit.source.getLine(i, null)
 
                 // prepend first-line indent
-                def indent = ''
                 if( i == first ) {
                     int k = 0
                     while( k < line.size() && line[k] == ' ' )
                         k++
-                    indent = line.substring(0, k)
+                    buffer.append( line.substring(0, k) )
                 }
-
-                // prepend statement label
-                if( node.statementLabel )
-                    buffer.append(indent).append(node.statementLabel).append(':\n')
 
                 final begin = (i == first) ? colx - 1 : 0
                 final end = (i == last) ? colz - 1 : line.size()
-                buffer.append(indent).append( line.substring(begin, end) ).append('\n')
+                buffer.append( line.substring(begin, end) ).append('\n')
+            }
+        }
+
+        private void readSource( ClosureExpression node, StringBuilder buffer, SourceUnit unit ) {
+            final colx = node.getColumnNumber()
+            final colz = node.getLastColumnNumber()
+            final first = node.getLineNumber()
+            final last = node.getLastLineNumber()
+            for( int i=first; i<=last; i++ ) {
+                def line = unit.source.getLine(i, null)
+                if( i==last ) {
+                    line = line.substring(0,colz-1).replaceFirst(/}.*$/,'')
+                    if( !line.trim() ) continue
+                }
+                if( i==first ) {
+                    line = line.substring(colx-1).replaceFirst(/^.*\{/,'').trim()
+                    if( !line ) continue
+                }
+                buffer.append(line) .append('\n')
             }
         }
 
--- a/modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
@@ -471,9 +471,9 @@ class NextflowDSLImpl implements ASTTransformation {
                         }
                         body.add(stm)
                 }
+
+                readSource(stm, source, unit)
             }
-            // read the closure source
-            readSource(closure, source, unit, true)
 
             final bodyClosure = closureX(null, block(scope, body))
             final invokeBody = makeScriptWrapper(bodyClosure, source.toString(), 'workflow', unit)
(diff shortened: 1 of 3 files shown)


marcodelapierre:
Hi @mahesh-panchal thank you for taking the time to open this PR.

The fix for the Error Cause looks spot on to me.

On the other hand, I would like to inspect a bit more the other fix, on the Task Source.  In fact, the current approach in the PR does not seem general  (see example below).  
I see how `task.source` is formatted weirdly, perhaps investigating the root cause there will help find a general fix.

@bentsherman  what do you think?


A breaking example (with irregular indentations at various lines):
```nextflow
workflow {
    TASK(Channel.of('a')).view()
}

process TASK {
    input:
    val letter

    script:
    assert letter.isUpperCase()
    def puzzle = null
      if ( puzzle ) {
          println "Puzzled"
    } else {
      println "LGTM"
    }
    """
    echo $letter
    """

    output:
    stdout
}
```

```
ERROR ~ Error executing process > 'TASK (1)'

Caused by:
  assert letter.isUpperCase()
         |      |
         'a'    false -- Check script 'test.nf' at line: 10


Source block:
  assert letter.isUpperCase()
  def puzzle = null
  if ( puzzle ) {
        println "Puzzled"
  } else {
    println "LGTM"
  }
  """
  echo $letter
  """
```

marcodelapierre:
Hi @mahesh-panchal thank you for taking the time to open this PR.

The fix for the Error Cause looks spot on to me.

On the other hand, I would like to inspect a bit more the other fix, on the Task Source.  In fact, the current approach in the PR does not seem general  (see example below).  
I see how `task.source` is formatted weirdly, perhaps investigating the root cause there will help find a general fix.

@bentsherman  what do you think?


A breaking example : 
- added some irregular indentations at various lines
- in the error output, the line `if ( puzzle ) {` has the wrong indentation

```nextflow
workflow {
    TASK(Channel.of('a')).view()
}

process TASK {
    input:
    val letter

    script:
    assert letter.isUpperCase()
    def puzzle = null
      if ( puzzle ) {
          println "Puzzled"
        println "Double puzzled"
    } else {
      println "LGTM"
    }
    """
    echo $letter
    """

    output:
    stdout
}```

```
ERROR ~ Error executing process > 'TASK (1)'

Caused by:
  assert letter.isUpperCase()
         |      |
         'a'    false -- Check script 'test.nf' at line: 10


Source block:
  assert letter.isUpperCase()
  def puzzle = null
  if ( puzzle ) {
        println "Puzzled"
      println "Double puzzled"
  } else {
    println "LGTM"
  }
  """
  echo $letter
  """
```

**Comment by mahesh-panchal** [comment]
My problem was I couldn't find where that was coming from quickly so if someone can point me in the right direction, I can amend the PR.

**Comment by bentsherman** [comment]
The weird indentation comes from how the source code is extracted from the AST:

https://github.com/nextflow-io/nextflow/blob/5c37fcc248ba47efc36681758015a458f19bf6c2/modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy#L753-L776

The main AST transform wraps the script body in a BodyDef constructor and appends the source text as an argument. Each AST node includes the line and column range for the node, so it uses that to get the source text for each statement in the script body.

The problem is that the text range doesn't include the indent before the statement, so if the statement spans multiple lines (e.g. an if-else statement), the first line is not indented.

I think the ideal solution would be to adjust `readSource()` to grab that first-line indent, basically set `colx` to 0 instead of the starting column number. But we need consider any edge cases. A line can have multiple statements separated by semi-colons... it is not "idiomatic" Nextflow, but it is valid syntax.

I will play around with this idea, and Mahesh you are more than welcome to do the same if you feel like delving into the AST 😄

**Comment by bentsherman** [comment]
ooh I can plug my developer docs: https://nextflow.io/docs/latest/developer/nextflow.ast.html

**Comment by bentsherman** [comment]
I think what we can do is check if the AST node is preceded only by whitespace on the first line and if so, include it as the indent. I will try to make this work. Also need to handle this annoying edge case with `stripBrackets`

**Comment by bentsherman** [comment]
Done! Works with both of your examples and also handles multi-statement lines, so the following code:
```groovy
    script:
    assert letter.isUpperCase() ; def puzzle = null
```

Will be printed as:
```groovy
    script:
    assert letter.isUpperCase()
    def puzzle = null
```

@marcodelapierre if you're satisfied with this solution, let's merge it

marcodelapierre:
@bentsherman  not quite there yet. 

2 unit tests are failing in a quite significant way:

```
Condition not satisfied:

meta.getWorkflow('bravo') .source.stripIndent(true) == '''              take: foo take: bar main: print foo print bar emit: foo+bar '''.stripIndent(true)
|    |                     |      |                 |                                                                                   |
|    |                     |      |                 |                                                                                   take: foo
|    |                     |      |                 |                                                                                   take: bar
|    |                     |      |                 |                                                                                   main:
|    |                     |      |                 |                                                                                     print foo
|    |                     |      |                 |                                                                                     print bar
|    |                     |      |                 |                                                                                   emit: 
|    |                     |      |                 |                                                                                     foo+bar
|    |                     |      |                 false
|    |                     |      |                 25 differences (62% similarity)
|    |                     |      |                 (------)foo\n(------)bar\n(------~)  print foo\n  print bar\n(-----) (-~) (-)foo+bar\n
|    |                     |      |                 (take: )foo\n(take: )bar\n(main:\n)  print foo\n  print bar\n(emit:) (\n) ( )foo+bar\n
|    |                     |      foo
|    |                     |      bar
|    |                     |        print foo
|    |                     |        print bar
|    |                     |        foo+bar
|    |                                   foo
|    |                                   bar
|    |                                     print foo
|    |                                     print bar
|    |                                     foo+bar
|    WorkflowDef[workflow bravo]
<nextflow.script.ScriptMeta@40107927 clazz=class Script1 scriptPath=null definitions=[bravo:WorkflowDef[workflow bravo], alpha:WorkflowDef[workflow alpha], delta:WorkflowDef[workflow delta], empty:WorkflowDef[workflow empty]] imports=[:] module=false functionsCount=[:] memoizedMethodClosure$getModuleBundle=org.codehaus.groovy.runtime.memoize.Memoize$MemoizeFunction@51728c08>

	at nextflow.script.WorkflowDefTest.should parse workflow(WorkflowDefTest.groovy:101)
```


```
Condition not satisfied:

workflow.getSource().stripIndent(true) == '''                            take: foo main: print x emit: foo '''.stripIndent(true)
|        |           |                 |                                                                       |
|        |           |                 |                                                                       take:
|        |           |                 |                                                                         foo
|        |           |                 |                                                                       main:
|        |           |                 |                                                                         print x 
|        |           |                 |                                                                       emit: 
|        |           |                 |                                                                         foo  
|        |           |                 false
|        |           |                 28 differences (36% similarity)
|        |           |                 (------~--)foo\n(------~--)print x(-)\n(-------~--)foo(--)\n
|        |           |                 (take:\n  )foo\n(main:\n  )print x( )\n(emit: \n  )foo(  )\n
|        |           foo
|        |           print x
|        |           foo
|                        foo
|                        print x
|                        foo
WorkflowDef[workflow alpha]

	at nextflow.script.WorkflowDefTest.should capture workflow code(WorkflowDefTest.groovy:347)
```

bentsherman:
The failing tests are due to changing the workflow xform to extract the source of each statement instead of the entire body, which leaves out labels and whitespace. I just adjusted the tests but I don't like warping the workflow body source this much. I think I will go back to reading the entire workflow body and strip out the brackets some other way

bentsherman (approved):

mahesh-panchal:
Thanks for pushing this over the finish line.

pditommaso:
Thanks for opening this PR!

PR #4686, opened 2024-01-25
Add ignoreThenFail error strategy

### LINKED ISSUE #4446: Defer error exit status to end of pipeline run
> really go as far as possible in running the pipeline, but then exit 1 at the very end if any errors were ignored.

This is a pretty useful feature. Imagine the scenario where you have 96 independent samples, if 1 fails and it brings the whole pipeline crashing down that's annoying, but if you use errorStrategy = 'ignore' then you have to write your own solution. We could have `process.errorStrategy = 'catch'`. This might be a separate ticket though.

Having some form of introspection on tasks in the workflow.onComplete block would also help for similar reasons and be more generally useful.

_Originally posted by @adamrtalbot in https://github.com/nextflow-io/nextflow/issues/4365#issuecomment-1759191011_

bentsherman:
Close #4446 

Test example:
```groovy
process foo {
  errorStrategy 'ignoreThenFail'

  input:
  val id

  script:
  """
  test $id -ne 4
  """
}

workflow {
  foo( Channel.of(1..10) )
}
```

This pipeline will ignore the failed task but then return a non-zero exit code.


pditommaso (changes requested):
Tests are needed

bentsherman:
`ignoreFail` ? 😆 

unfortunately the shorter one is not as clear

pditommaso:
Was thinking a couple of other implementation alternatives: 
1. use `ignoreThenFail` but map it to `IGNORE` enum value and tracking the *fail* flag independently 
2. instead of introducing new `errorStrategy` value, define it as an `executor` or `nextflow` scope setting

bentsherman:
I will review the docs and then it should be good to go

PR #4702, opened 2024-01-30
Don't save kebab-case aliases in the params map

### LINKED ISSUE #4033: Params case translation between camelCase and kebab-case doesn't always work as expected

Parameters are sometimes incorrectly translated between camelCase and kebab-case. Particularly, when default values of `params` are defined in nextflow.config, the translation to kebab-case doesn't work as expected, but when they are defined in the main script, they seem to work fine. 

Here's a simple script:
```nextflow
#!/usr/bin/env nextflow

params.firstParam = false
params.secondParam = "default value"

workflow {
  println(params)
}
```

If we run it without any arguments, we can see the default values, with all parameters successfully translated into kebab-case:
```bash
$ ./test.nf
N E X T F L O W  ~  version 23.04.2
Launching `./test.nf` [grave_ptolemy] DSL2 - revision: 5375fb2bec
[firstParam:false, first-param:false, secondParam:default value, second-param:default value]
```

If we run it with options in camelCase or kebab-case, everything seems to work fine:
```bash
$ ./test.nf --firstParam --secondParam 'new value'
N E X T F L O W  ~  version 23.04.2
Launching `./test.nf` [pedantic_easley] DSL2 - revision: 5375fb2bec
[firstParam:true, first-param:true, secondParam:new value, second-param:new value]

$ ./test.nf --first-param --second-param 'new value'
N E X T F L O W  ~  version 23.04.2
Launching `./test.nf` [sharp_noyce] DSL2 - revision: 5375fb2bec
[first-param:true, firstParam:true, second-param:new value, secondParam:new value]
```

However, when the default values of the parameters are specified in `nextflow.config`, it behaves differently. In this example, we have the same script minus the definition of the params:
```nextflow
#!/usr/bin/env nextflow
nextflow.enable.dsl=2

workflow {
  println(params)
}
```

In addition, we have a `nextflow.config` file where the default parameter values are defined:
```nextflow
params {
  firstParam = false
  secondParam = "default value"
}
```

When we run it with no options, we see the default values:
```bash
$ ./test.nf
N E X T F L O W  ~  version 23.04.2
Launching `./test.nf` [sick_faggin] DSL2 - revision: 5d765fd59e
[firstParam:false, first-param:false, secondParam:default value, second-param:default value]
```

If we run it with camelCase options, it behaves as expected:
```bash
$ ./test.nf --firstParam --secondParam 'new value'
N E X T F L O W  ~  version 23.04.2
Launching `./test.nf` [lonely_mirzakhani] DSL2 - revision: 5d765fd59e
[firstParam:true, first-param:true, secondParam:new value, second-param:new value]
```

However, when we run it with kebab-case options, the values don't get picked up:
```bash
$ ./test.nf --first-param --second-param 'new value'
N E X T F L O W  ~  version 23.04.2
Launching `./test.nf` [fabulous_wescoff] DSL2 - revision: 5d765fd59e
[firstParam:false, first-param:false, secondParam:default value, second-param:default value]
```

There seems to be a disconnect somewhere in the logic that handles the command-line option case translation. If params are defined in the main script it works fine, but if they're defined in the config file, it doesn't.

### Environment 

* Nextflow version: 23.04.2 build 5870
* Java version: openjdk version "11.0.19"
* Operating system: Ubuntu 20.04.6 LTS
* Bash version: 5.0.17(1)-release (x86_64-pc-linux-gnu)

### Additional context

nextflow.log here:
[nextflow.log](https://github.com/nextflow-io/nextflow/files/11774303/nextflow.log)

Also, as an aside, the `--option=value` mode of parameter passing doesn't seem to be supported. This is may be as designed, but it's something I noticed while testing this:
```bash
# the following *don't* work, although that may be as designed:
$ ./test.nf --first-param --second-param='new value'
N E X T F L O W  ~  version 23.04.2
Launching `./test.nf` [infallible_yalow] DSL2 - revision: d14c7a6085
[first-param:--second-param=new value, firstParam:--second-param=new value, secondParam:default value, second-param:default value]

mykle at kinilau (screen *flowtest*) in ~/tmp
$ ./test.nf --firstParam --secondParam='new value'
N E X T F L O W  ~  version 23.04.2
Launching `./test.nf` [nauseous_agnesi] DSL2 - revision: d14c7a6085
[firstParam:--secondParam=new value, first-param:--secondParam=new value, secondParam:default value, second-param:default value]
```

bentsherman:
Close #4033 

This PR attempts to resolve some quirks with the params map by separating the alias behavior from the read-only behavior. It adds an `AliasMap` which can be used to create a map that aliases between kebab-case and camelCase. The params map re-uses this class internally.

Still need to add some tests.


bentsherman:
Confirmed that it works with the test case in the linked issue

bentsherman:
Simplified it further:
- When CLI params and params file are parsed, kebab-case params are automatically converted to camelCase. Only the camelCase key is stored.
- The params map does not do any aliasing, since kebab-case params have already been converted, and the kebab-case variant does not need to be read

modules/nextflow/src/test/groovy/nextflow/script/ScriptBindingTest.groovy
```diff
@@ -160,68 +124,17 @@ class ScriptBindingTest extends Specification {
...
-        def Y='hello'
-        def binding = new ScriptBinding([:])
-        binding.setVariable('foo', X)
-        binding.setVariable('bar', Y)
-
-        expect:
-        binding.getVariableName(X) == 'foo'
-        binding.getVariableName(Y) == 'bar'
-        binding.getVariableName(new String(Y)) == null
-        binding.getVariableName(null) == null
```
pditommaso, line 224: Any reason this was deleted?

pditommaso:
Intellij reports a problem with `ParamsMap` 

Screenshot 2024-09-27 at 17 40 16

modules/nextflow/src/test/groovy/nextflow/script/ScriptBindingTest.groovy
bentsherman: weird, I'm not sure. maybe some mistake in resolving a merge conflict

bentsherman:
This PR can be simplified further, I will fix those issues and make a few more edits

pditommaso (approved):
Much better now, thanks Ben

PR #4713, opened 2024-02-05
Automatic task cleanup (without resumability)

bentsherman:
Spun off from #3849 without the WIP resumability piece.


modules/nextflow/src/main/groovy/nextflow/trace/TaskCleanupObserver.groovy
```diff
@@ -0,0 +1,452 @@
...
+     *
+     * @param task
+     */
+    private void deleteTask(TaskRun task) {
+        log.trace "[${task.name}] Deleting task directory: ${task.workDir.toUriString()}"
+
+        // delete task
+        threadPool.submit({
+            try {
+                FileHelper.deletePath(task.workDir)
```
robsyme, line 378: This looks like we remove the whole task working directory, which would also destroy the .command files. These files are used by the Seqera Platform and are generally useful for task provenance. Would it be sensible to consider whitelisting the Nextflow (and maybe Fusion) dotfiles from deletion?

bentsherman: My plan was to save this provenance information in the `.nextflow` metadata cache, but right now that code is sitting in #3849. This information can also be preserved by using nf-prov

How else are they used by Seqera Platform?

bentsherman:
Closing in favor of [nf-boost](https://github.com/bentsherman/nf-boost) plugin

modules/nextflow/src/main/groovy/nextflow/trace/TaskCleanupObserver.groovy
bentsherman: @robsyme I managed to move this PR to a plugin, so feel free to make an issue there if you want to continue this discussion about deleting the task directory.

I could add a config option to control whether to delete the task directory or only the outputs. Also, the upcoming workflow output DSL might make it very difficult to know when a task directory could be deleted, so it might become a moot point

Issue #4725, opened 2024-02-08
Fusion symlink resolution doesn't work with directories


bentsherman:
We recently added some logic to handle Fusion symlinks, which happen when an input file is included in a published output. We use the `.fusion.symlinks` file to detect and resolve any Fusion symlinks when they are published.

However, when publishing a directory, Nextflow does not walk the directory tree and publish each file individually, it just publishes the directory. If Fusion symlinks were staged into this directory, they will not be detected and resolved. 

Given a list of input files:
```console
$ ls -1 freshdesk/fd-4463/files/ 
1.txt
2.txt
3.txt
4.txt
5.txt
```

The following pipeline script will demonstrate this issue:

```groovy
process AGGREGATE {
  container "quay.io/nextflow/bash"
  publishDir "results", mode: "copy"

  input:
  path(samples), stageAs: 'AnalysisFiles/'

  output:
  path("*")

  script:
  """
  for name in AnalysisFiles/*.txt; do
    touch AnalysisFiles/Analysis_on_\$(basename \${name} .txt)
  done
  """
}

workflow {
    AGGREGATE( files("files/*") )
}
```

One workaround that I tried is to make the outputs more explicit:

```groovy
  output:
  path("AnalysisFiles/*.txt", includeInputs: true)
  path("AnalysisFiles/Analysis_on_*")
```

But this doesn't quite work because of a small bug in the symlink resolution. I will submit a patch so that at least this workaround works.

A more permanent solution would be to walk the directory tree and publish each file explicitly instead of only publishing the directory (#3933). There are other benefits to making this change, see also #3372.

bentsherman:
Here is another workaround that works with the current version of Nextflow:
```groovy
process AGGREGATE {
  container "quay.io/nextflow/bash"
  publishDir "results/AnalysisFiles", mode: "copy"

  input:
  path("*.txt")

  output:
  path("*.txt", includeInputs: true)
  path("Analysis_on_*")

  script:
  """
  for name in *.txt; do
    touch Analysis_on_\$(basename \${name} .txt)
  done
  """
}
```

So there is something to be said for trying to keep the task directory as flat as possible and using the publish mechanism to create the desired directory structure. Still, this may not be possible or easy to do with certain tools, so I would still like to fix it in Nextflow.

IreneRobles:
Hi, 

I leave this message to encourage finding a solution to the issue.

[...]

PR #4726, opened 2024-02-08
Fix Fusion symlink resolution with nested files

bentsherman:
Partial solution to #4725 (full solution also requires #3933)

Test case:
```groovy
process AGGREGATE {
  container "quay.io/nextflow/bash"
  publishDir "results", mode: "copy"

  input:
  path(samples), stageAs: 'AnalysisFiles/'

  output:
  path("AnalysisFiles/*.txt", includeInputs: true)
  path("AnalysisFiles/Analysis_on_*")

  script:
  """
  for name in AnalysisFiles/*.txt; do
    touch AnalysisFiles/Analysis_on_\$(basename \${name} .txt)
  done
  """
}

workflow {
    AGGREGATE( files("files/*.txt") )
}
```

Where `files` should be populated with some text files. Run with Fusion enabled.


bentsherman:
@marcodelapierre this one would be a good inaugural PR for the "merge to master" effort. It is a simple bugfix and does not affect the core runtime, only some auxiliary logic for Fusion symlinks. If you can review it and approve, I will merge it 👀

marcodelapierre:
I was attempting the test case described above.  However I am getting an unrelated error, both with stable Nextflow and this build:
```
Unexpected path for Fusion script launcher: /Users/marco.delapierre/test/fusion-fix/work/45/681d22ba7f16b2567dc00f7a356900
```

```
$ cat nextflow.config 
docker {
    enabled = true
}

fusion {
    enabled = true
}

wave {
    enabled = true
}
```

What am I missing?

marcodelapierre:
Just confirming I have the above issue when testing Fusion, with any of stable, edge, build from this PR. Both on x86 and ARM, also tried with a different container.   I must be missing something trivial in the config...?

bentsherman:
I just uploaded the test case to github so that you can run it out-of-the-box:
```
nextflow run bentsherman/nf-tests -r fusion-symlinks
```

I couldn't reproduce your issue, but maybe you can try this example. Maybe you weren't mounting the AWS credentials into the container

Even stranger though, I just ran the test case with 23.10.1 and didn't reproduce the original issue of the symlinks not being resolved. When I list the work directory in S3, there is no `.fusion.symlinks` file and the input files are not symlinked.

@jordeu have you changed the behavior of Fusion symlinks? the above test case used to create Fusion symlinks but now it doesn't

marcodelapierre:
Thanks for the extra details Ben, by defining workDir and amazon credentials properly I could run the tests. 

I confirm your finding with Nextflow 23.10.1 stable and any of the example workflows in #4725: all inputs and output files are published, and no `.fusion.symlinks` files are found in the S3 work directory.

I see this functional behaviour is also consistent with what a user reports in #4725. 

I also tested the PR and confirm that the patch enables to correctly identify Fusion symlinks when running the workaround workflow in #4725 .

I am therefore approving the PR, as it effectively patches targeted case.

Of course, the key question remains of what change is now making the sample workflows to work.

marcodelapierre (approved):
Targeted corner case is fixed by the patch - effectiveness of `resolveFusionLink` .

I would still merge as this improves consistency of Fusion link resolution.

pditommaso (approved):

bentsherman:
False alarm, I set the publish mode to "copy" in the test which is another workaround

Fixing that, I now see the `.fusion.symlinks` in the folder and the symlink issue which is fixed by this PR. I will merge once the CI passes

bentsherman:
@pditommaso google batch logging test failed again

pditommaso:
This looks different 

```
Method timed out after 600.00 seconds
	at java.base@19.0.2/java.lang.Thread.sleep(Thread.java:465)
	at nextflow.cloud.google.batch.logging.BatchLoggingTest.should fetch logs(BatchLoggingTest.groovy:135)

```

bentsherman:
Now the google test failed on 21 but not 11 🤔

PR #4727, opened 2024-02-08
Rename Tower -> Seqera Platform in docs and log messages

bentsherman:


pditommaso (commented):
Worth adding in some key places `(formerly Tower Cloud)`

PR #4729, opened 2024-02-09
Overwrite published outputs only if they are stale

### LINKED ISSUE #3372: Validation of published outputs

bentsherman:
Close #3372 

This PR builds on #3933 , which ensures that files in a directory are explicitly published. Given this behavior, we can then compare the source and target checksums to decide whether to publish a file.

- If the checksums match then the copy can be skipped
- If the checksums don't match, then the target should *always* be overwritten, even if the task was cached, because if the previous run was terminated and publishing cancelled, the published file might be incomplete.

The checksum code is ripped from #3802 , as it is also needed for task provenance and the resumability for automatic cleanup (#3849). It tries to use a pre-computed checksum (such as ETag for cloud storage paths) where possible to avoid scanning the file contents.

The only qualm I have right now is that the checksum code doesn't use the ETag for Google cloud storage paths. This is because we use the built-in file system provider from the Google SDK rather than our own. Maybe we can just extend this FS provider and register it over the existing one, but I'm not sure yet how to do it.


plugins/nf-google/src/main/nextflow/cloud/google/nio/GsPath.groovy
```diff
@@ -0,0 +1,168 @@
...
+        if( metadata == null ) {
+            final googleOpts = GoogleOpts.create(Global.session as Session)
+            final client = StorageOptions.newBuilder()
+                    .setProjectId(googleOpts.projectId)
+                    .build()
+                    .getService()
+            metadata = client.get(target.bucket(), target.toString())
+        }
+        return metadata?.getMd5ToHexString()
+    }
```
bentsherman, line 158: I added a custom NIO filesystem for Google cloud storage which just delegates to the existing one, but allows us to add custom behavior like this ETag support.

It turns out that the ETag returned by the Java SDK is a CRC32 instead of MD5. The MD5 is also available through this storage API, but I haven't gotten it to work yet. The metadata is always null, even though I provided the projectId and I have the credentials file. Need to investigate further when I have more time

bentsherman:
Some notes from discussion today (related to workflow output DSL):

- checksum verification should be optional as it can be very expensive for non-object storage
- should also have a lightweight alternative that simply checks the last modified timestamp and file size (similar to `cache` directive)

Issue #1792, opened 2020-11-13
Params override is not honoured 


pditommaso:
Use the folowing `nextflow.config` 

```
profiles {
    local {
        params.genome = '/local/genome'
    }

    cloud {
        params.genome = 'gs://genome'
    }
}
```

then run 

```
nextflow config -profile cloud,local
```

If prints 

```
params {
   genome = 'gs://genome'
}
```

Instead of 

```
params {
   genome = '/local/genome'
}
```

PR #4744, opened 2024-02-14
Config parser (and loader)

### LINKED ISSUE #2723: Evolution of Nextflow configuration file

bentsherman:
Close #2723 

Implements the [strict config syntax](https://nextflow.io/docs/latest/updating-syntax.html#configuration-syntax) in Nextflow, using the shared "compiler" module from the language server.

To use the strict config syntax, simply set `NXF_SYNTAX_PARSER=v2` in your environment when running `nextflow`.

This config loader is much simpler and easier to read. There is no more dependence on metaclasses, instead there is a simple builder DSL for executing a config file and producing a map.

NOTE: While this PR uses Jitpack to load the shared module, this dependency should be inverted before the release of 25.04. That is, eventually the compiler module should reside in Nextflow and the language server should consume it.


bentsherman:
Research updates:
- Groovy compiler provides an interface to use a custom parser (in order to ease the transition from Antlr2 to Antlr4), this can be used to inject our custom parser without much hassle
- Config schema effort can be pursued independently of the parser, see #4201 
- Might be able to use the [Java security manager](https://docs.oracle.com/javase/8/docs/api/java/lang/SecurityManager.html) to prevent execution of unsafe code in the config file such as file I/O, subprocesses

bentsherman:
Here's an example to illustrate how the config parser "executes" the config file. Given this config:
```groovy
foo.bar = 'hello'

foo {
  bar = 'hello'
}

includeConfig 'foobar.config'
```

The parser transforms this code into the following:
```groovy
assign(['foo', 'bar'], 'hello')

block('foo', {
  assign(['bar'], 'hello')
}

includeConfig('foobar.config')
```

Basically each statement is translated to a DSL method (see `ConfigDsl`). This "hidden DSL" allows us to control how each statement is executed. In other words, the config parser + hidden DSL implement a simple interpreter for the Nextflow config language, separating the config language from the config execution.

I think this is a good model for how we might separate the Nextflow language from the runtime. Instead of hooking directly into Groovy methods and relying on Groovy MOP, metaclass, etc, we can introduce a hidden DSL which allows us to better control how a given statement (e.g. invoking a process in a workflow) is executed by the runtime.

bentsherman:
Java security manager is being [removed](https://snyk.io/blog/securitymanager-removed-java/). I will keep an eye out for other sandboxing techniques, but in the meantime, a simple solution is to provide a "safe" execution mode where assignment expressions are wrapped in string literals.

ewels:
Really like the way that this is going, great work 👏🏻

pditommaso:
@bentsherman can you please solve the conflicts?

bentsherman:
The PR I wanted you to review is this one -- #4201 

However I just looked at it and it's not really in a review-able state. I'll ping you again when it's more complete.

bentsherman:
Trying to get jitpack to work. So far I can build with mavenLocal but not with jitpack.

- Language server PR: https://github.com/nextflow-io/language-server/pull/95
- Following the jitpack docs [here](https://docs.jitpack.io/) and [here](https://docs.jitpack.io/building/).

I had to move all of the repositories config to settings.gradle and add this setting for jitpack: https://github.com/nextflow-io/nextflow/blob/a4cbea481a89abb0655925d0b3f7a5de6949a90c/settings.gradle#L35-L38

Then I would specify the dependency like this:
```groovy
    implementation 'com.github.nextflow-io:language-server:compiler:jitpack-SNAPSHOT'
```

But it doesn't even seem to use the jitpack repository. It just searches the other ones and fails.

I can pull from mavenLocal like this:
```groovy
    implementation 'com.github.nextflow-io:compiler:jitpack-SNAPSHOT'
```

So I have updated the CI to do this for now since it is slightly better. But I might need some help to get the jitpack to work.

build.gradle
```diff
@@ -85,14 +85,6 @@ allprojects {
...
     }
 
-    repositories {
-        mavenCentral()
-        maven { url 'https://repo.eclipse.org/content/groups/releases' }
-        maven { url 'https://oss.sonatype.org/content/repositories/snapshots' }
-        maven { url = "https://s3-eu-west-1.amazonaws.com/maven.seqera.io/releases" }
-        maven { url = "https://s3-eu-west-1.amazonaws.com/maven.seqera.io/snapshots" }
-    }
-
```
pditommaso, line 95: This should not be removed instead `	maven { url 'https://jitpack.io' }` should be added

bentsherman: I had to move all of the repositories config to `settings.gradle` because the `dependencyResolutionManagement` recommended by jitpack was not recognized here

pditommaso: It should be enough adding `maven { url 'https://jitpack.io' }` here and nothing else

bentsherman: Turns out the includeGroup for jitpack was too strict. Local build is now working with jitpack. I will now see if I can move the config back into the build.gradle

bentsherman:
@pditommaso config parser is ready to review

Also, this google test is failing again. Do the credentials need to be refreshed?

```
BatchLoggingTest > should fetch logs FAILED
    org.spockframework.runtime.ConditionNotSatisfiedError at BatchLoggingTest.groovy:149
```

modules/nextflow/src/test/groovy/nextflow/config/parser/legacy/ConfigParserLegacyTest.groovy
```diff
@@ -739,6 +683,26 @@ class ConfigParserTest extends Specification {
 
     }
 
+    def 'should apply profiles in the order they were defined' () {
```
bentsherman, line 686: @pditommaso this test is related to #1792 . We determined that config profiles are applied this way with the legacy parser. We could patch it, but I figure there's no point now, so just added a test to verify it's behavior.

bentsherman:
@pditommaso tests are passing, let's review and merge

docs/updating-syntax.md
```diff
@@ -487,8 +487,14 @@ The process `when` section is deprecated. Use conditional logic, such as an `if`
 
 The process `shell` section is deprecated. Use the `script` block instead. The VS Code extension provides syntax highlighting and error checking to help distinguish between Nextflow variables and Bash variables.
 
+(updating-config-syntax)=
+
 ### Configuration syntax
 
+:::{versionadded} 25.02.0-edge
+The strict config syntax can be enabled in Nextflow by setting `NXF_ENABLE_STRICT_SYNTAX=true`.
```
christopher-hakkaart, line 495:
```suggestion
Enable the strict config syntax by setting `NXF_ENABLE_STRICT_SYNTAX=true`.
```
I wonder if this should be worded actively as the rest of the section relies on this being set. Don't consider this a blocker if you disagree.

.github/workflows/build.yml
```diff
@@ -134,6 +134,7 @@ jobs:
       matrix:
         java_version: [17, 23]
         test_mode: ["test_integration", "test_docs", "test_aws", "test_azure", "test_google", "test_wave"]
+        strict_syntax: [true, false]
```
pditommaso, line 137: This results in too many builds. I was suggesting to add it as a `test_mode` then hook in the test.sh script and run `test_integration` using the new parser

docs/reference/env-vars.md
```diff
@@ -92,6 +92,11 @@ The following environment variables control the configuration of the Nextflow ru
   :::
 : Enable Nextflow *strict* execution mode (default: `false`)
 
+`NXF_ENABLE_STRICT_SYNTAX`
```
pditommaso, line 95:
```suggestion
`NXF_ENABLE_SYNTAX_PARSER_V2` (?)
```

modules/nextflow/src/main/groovy/nextflow/config/ConfigParserFactory.groovy
```diff
@@ -0,0 +1,42 @@
...
+package nextflow.config
+
+import groovy.transform.CompileStatic
+import groovy.util.logging.Slf4j
+import nextflow.SysEnv
+import nextflow.config.parser.legacy.ConfigParserLegacy
+import nextflow.config.parser.ConfigParserImpl
+
+/**
+ * @author Ben Sherman <bentshermann@gmail.com>
```
pditommaso, line 27: Add a description please

modules/nextflow/src/main/groovy/nextflow/config/parser/ConfigToGroovyVisitor.java
```diff
@@ -0,0 +1,110 @@
...
+import org.codehaus.groovy.ast.VariableScope;
+import org.codehaus.groovy.ast.expr.Expression;
+import org.codehaus.groovy.ast.stmt.ReturnStatement;
+import org.codehaus.groovy.ast.stmt.Statement;
+import org.codehaus.groovy.control.SourceUnit;
+
+import static org.codehaus.groovy.ast.tools.GeneralUtils.*;
+
+/**
+ *
```
pditommaso, line 36: Add a description please

modules/nextflow/src/main/groovy/nextflow/config/parser/legacy/ConfigBase.groovy
```diff
@@ -14,14 +14,15 @@
  * limitations under the License.
  */
 
-package nextflow.config
+package nextflow.config.parser.legacy
```
pditommaso, line 17: it would be preferable to not shuffle stuff around and use `Legacy`.

modules/nextflow/src/main/groovy/nextflow/config/parser/legacy/ConfigParserLegacy.groovy
pditommaso, line 1: Consider keep current naming and use `ConfigParserV2` for the new one

pditommaso (changes requested):

docs/updating-syntax.md
bentsherman: Yeah, I prefer this wording because the user might follow these migration notes with the language server or with Nextflow 25.04, so it's not essential that they enable this variable

.github/workflows/build.yml
bentsherman: That makes more sense. I didn't want to double the number of builds either

docs/reference/env-vars.md
bentsherman: Well, I think it will be better to call it `NXF_ENABLE_STRICT_SYNTAX` to keep consistent with our terminology around [strict syntax](https://nextflow.io/docs/latest/updating-syntax.html#preparing-for-strict-syntax). It's not just a new and improved parser, it's a formal language spec.

modules/nextflow/src/main/groovy/nextflow/config/parser/legacy/ConfigParserLegacy.groovy
bentsherman: `ConfigParser` now refers to the shared interface which both parsers implement. That's why I had to rename them. I could call them `ConfigParserV1` and `ConfigParserV2`.

modules/nextflow/src/main/groovy/nextflow/config/parser/legacy/ConfigBase.groovy
bentsherman: I think this is a reasonable change to keep related classes organized. You have done a similar thing many times, e.g. `nextflow.executor.local` or your proposed operator cleanup for the provenance tracking

docs/updating-syntax.md
christopher-hakkaart: Valid - all good 👍 Thanks!

docs/config.md
```diff
@@ -276,41 +275,47 @@ profiles {
         process.container = 'cbcrg/imagex'
         docker.enabled = true
     }
-
 }
 ```
 
-This configuration defines three different profiles: `standard`, `cluster`, and `cloud`, that each set different process
-configuration strategies depending on the target runtime platform. The `standard` profile is used by default when no profile is specified.
+This configuration defines three profiles: `standard`, `cluster`, and `cloud`. Each profile provides a different configuration for a given execution environment. The `standard` profile is used by default when no profile is specified.
```
christopher-hakkaart, line 281:
```suggestion
The above configuration defines three profiles: `standard`, `cluster`, and `cloud`. Each profile provides a different configuration for a given execution environment. The `standard` profile is used by default when no profile is specified.
```

christopher-hakkaart (approved):
One minor suggestion but otherwise the docs look great.

bentsherman:
Another flaky CI test:

```
~ Test 'publish-s3-kms.nf' run failed
   + /home/runner/work/nextflow/nextflow/nextflow -q run ../../publish-s3-kms.nf -c .config
   ++ check_kms_key
   ++ aws s3api head-object --bucket nf-kms-xyz --key work/ci-test/publish-s3/HELLO.tsv
   ++ grep -c e5109f93-b42d-4c26-89ee-8b251029a41d
   
   An error occurred (404) when calling the HeadObject operation: Not Found
```

docs/reference/env-vars.md
bentsherman: The strict feature flag won't be needed as we continue with the strict language spec, so I would rather just phase it out than try to work around it.

modules/nextflow/src/main/groovy/nextflow/config/parser/legacy/ConfigBase.groovy
pditommaso: That's fair, what I think it should be avoid the use of the term "legacy" in class or package name, as it's a bad practice to use the term "new" when introducing new class. It would be preferable just to use `v1` for current ones and v2 for new ones.

PR #4760, opened 2024-02-20
Save built-in reports as Tower reports

### LINKED ISSUE #3460: Allow to publish timeline, trace and DAG files as a Tower report
## New feature

Current Tower plugin is only able to use as reports files that are notified with the `onFilePublish` event. This excludes some files generated by Nextflow like the timeline, trace and DAG files.

Allow the Tower plugin use these files too. 

## Usage scenario 

- A user wants to add the pipeline timeline as a final report and visualize it directly in Tower
- A user wants to download the trace file from Tower
- A user want to download or visualize the DAG file from Tower.

## Suggest implementation 

Trigger a `onFilePublish` event for each of these files.

bentsherman:
Close #3460 

This PR saves the built-in reports as Tower reports by inspecting the Nextflow config. If a report is enabled, it will be eagerly sent to the reports file on workflow completion.

Note that if a report fails to create for some reason, it will be displayed in the platform run page but will not load since the file doesn't exist. In this case, the failure will have been logged as a warning to the console output.

### DIFF
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerReports.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerReports.groovy
@@ -204,35 +204,6 @@ class TowerReports {
         return false
     }
 
-    /**
-     * Send any built-in reports that are enabled to the reports file.
-     */
-    void sendRuntimeReports() {
-        final config = session.config
-        final Map<String,Map<String,String>> entries = [:]
-        if( config.navigate('report.enabled') ) {
-            final reportFile = config.navigate('report.file', ReportObserver.DEF_FILE_NAME) as String
-            entries[reportFile] = Map.of('display', 'Nextflow execution report')
-        }
-        if( config.navigate('timeline.enabled') ) {
-            final reportFile = config.navigate('timeline.file', TimelineObserver.DEF_FILE_NAME) as String
-            entries[reportFile] = Map.of('display', 'Nextflow timeline report')
-        }
-        if( config.navigate('trace.enabled') ) {
-            final reportFile = config.navigate('trace.file', TraceFileObserver.DEF_FILE_NAME) as String
-            entries[reportFile] = Map.of('display', 'Nextflow trace report')
-        }
-        if( config.navigate('dag.enabled') ) {
-            final reportFile = config.navigate('dag.file', GraphObserver.DEF_FILE_NAME) as String
-            entries[reportFile] = Map.of('display', 'Nextflow workflow diagram')
-        }
-
-        for( def entry : entries ) {
-            final destination = (entry.key as Path).complete()
-            writer.send((PrintWriter it) -> writeRecord(it, entry, destination))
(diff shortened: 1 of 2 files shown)


mbosio85:
Hi! how are the reports saved? I mean the names.
If there are user-defined reports, do the nextflow reports fall in the middle of the alphabetically sorted list or are they at the top/bottom, separated from the rest?

bentsherman:
They will be lumped in with the user-defined reports. For this reason I made the display name "Nextflow execution report", "Nextflow timeline report", etc so that the Nextflow reports are at least grouped together

As a second iteration, we could have the platform reports tab display "runtime" reports separately from "user" reports, and Nextflow could denote this type when it sends the report information to the platform

jordeu (commented):
The implementation seems correct. 

My only concern is that it might be better to make this configurable somewhere. With this approach, if the report/timeline/trace/dag are enabled, they will always be added as Tower reports, and there is no way to keep generating them without adding them as Tower reports.

This should be optional. Many users will still want to generate the report/timeline/trace/dag files without showing them on the Tower reports tab.

jordeu (approved):

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerReports.groovy
```diff
@@ -220,4 +224,27 @@ class TowerReports {
...
+            files << config.navigate('timeline.file', TimelineObserver.DEF_FILE_NAME)
+
+        if( config.navigate('trace.enabled') )
+            files << config.navigate('trace.file', TraceFileObserver.DEF_FILE_NAME)
+
+        if( config.navigate('dag.enabled') )
+            files << config.navigate('dag.file', GraphObserver.DEF_FILE_NAME)
+
+        for( def file : files )
+            filePublish( (file as Path).complete() )
```
pditommaso, line 248: This seems overlapping with the logic in the LogsHandler

https://github.com/nextflow-io/nextflow/blob/3460-publish-nextflow-reports/plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsHandler.groovy#L61-L70

bentsherman: Why is the logs handler saving the timeline report but not the others?

pditommaso: Because that was the requirement

bentsherman: What requirement? 😆 

In any case, the logs handler is saving the timeline file whereas this PR is adding the built-in reports to the Tower reports list. So there is no overlap

pditommaso (approved):

ewels:
Should we document this somewhere?

bentsherman:
Probably in the platform docs: https://docs.seqera.io/platform/23.4.0/reports/overview

Issue #4661, opened 2024-01-16
[proposal] Move some `publishDir` attributes to `process`-level directives


ewels:
_Based on a comment originally posted in https://github.com/nextflow-io/nextflow/issues/4205#issuecomment-1704966418_

`publishDir` supports multiple attributes that control file publishing behaviour. The problem with having them structured this way is that if _any_ need to be overwritten, then _entire_ `publishDir` statement has to be repeated. This leads to some attributes being repeated verbatim many times in pipeline configs.

This issue is to suggest moving attributes out to new directives, so that they can be set to a single value for the entire pipeline. This reduces duplicated configuration code. Could be all current attributes, but main priorities (most duplicated) are:

* `path` (ideally split into two, with "base" and current `path` attribute)
* `mode`
* `saveAs`

If possible, this would be under a `publishDir` scope, though I don't know if this is possible technically.

So ideally, it'd be something like taking the following example:

```nextflow
publishDir = [
    path: { "${params.outdir}/genome" },
    mode: params.publish_dir_mode,
    saveAs: { filename -> filename.equals('versions.yml') ? null : filename },
    enabled: params.save_reference
]
```

And it becoming:

```nextflow
// Defaults for all processes
process {
    publishDir { 
        base = params.outdir
        mode = params.publish_dir_mode
        saveAs = { filename -> filename.equals('versions.yml') ? null : filename }
    }
}
// Specific process overrides
process withName: "trimData" {
    publishDir.path: "trimmed"
}
process withName: "prepGenome" {
    publishDir.path: "genome",
    publishDir.enabled: params.save_reference
}
```

bentsherman:
`publishDir` defaults are tricky because it is a list of publish "commands" that can target specific outputs. I would suggest something like this:
```groovy
// process directive
process FASTQC {
    publish 'results', mode: 'copy', saveAs: { ... }

    // ...
}

// config option
process {
    publish.path = 'results'
    publish.mode = 'copy'
    publish.saveAs = { ... }
}
```

It is essentially giving up the ability to target specific outputs differently, instead the setting would apply to the entire process. It seems to be what people do almost every time, but I would like to investigate it more.

But I do suggest making a new process directive, even something as simple as `publishDir` -> `publish`, so that users have to make an explicit transition, because this new directive would be structured differently in the runtime.

ewels:
> It is essentially giving up the ability to target specific outputs differently, instead the setting would apply to the entire process.

[...]

Issue #4042, opened 2023-06-20
Flag files for downstream reporting via `publishDir`


ewels:
## New feature

An emerging standard for Nextflow pipelines is a root `tower.yml` file, used for [providing reports to Tower](https://help.tower.nf/23.1/reports/overview/#providing-reports).

A potential alternative is to instead define this metadata as part of `publishDir`, within the Nextflow config. This has a few advantages:
* Removes the need for yet-another-config-file in the repository root
* Keeps configuration of published files in a single location, not spread across multiple files
* Less Tower-specific, more community friendly

In this location, Nextflow will know about the _report_ status of files during the publish step and could potentially match patterns against actual files created, allowing some kind of metadata with precise file paths + report status to be generated in memory / in some kind of report.

## Suggest implementation

My suggestion is to add a new directive: `report` (`int`). Non-zero values (or >0) could include that files should be shown within downstream reporting functionality. The integer value itself could then be used as a weighting factor when sorting that list.

The directive should be paired together with the ability to filter the published files for a given process based on filename / a closure.

## Usage scenario

Based on the [`publishDir` config for a process in the nf-core/rnaseq pipeline](https://github.com/nf-core/rnaseq/blob/3bec2331cac2b5ff88a1dc71a21fab6529b57a0f/conf/modules.config#L1072-L1078), syntax / usage could potentially look something like this:

```diff
  withName: '.*:BAM_RSEQC:RSEQC_READDISTRIBUTION' {
      publishDir = [
          path: { "${params.outdir}/${params.aligner}/rseqc/read_distribution" },
          mode: params.publish_dir_mode,
          saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
+         report: { filename -> filename ==~ '.*\.pdf' ? 10 : null }
      ]
  }
```

Here, any PDF files published by this process would be given a report priority of `10`. The integer > 0 indicates that they should be shown in a report interface, value `10` gives weighting score for sorting the list of files there.

The results of of this directive then need to be handled somehow. I expect this to be the most contentious part of this suggestion! My suggestion would be a new optional output file, similar to reports and trace files. This could potentially tie into future efforts for provenance tracking of published files.

evanfloden:
The file filtering is a nice touch.

1. Do you envision either the need or the ability to have multiple `report: ` lines within the directive? For example, if a user wanted to have two different reports with different weightings.

[...]

PR #4784, opened 2024-02-28
Workflow output definition

### LINKED ISSUE #4670: Workflow output definitions and schema
Spun off from https://github.com/nextflow-io/nextflow/issues/4042#issuecomment-1875924234

Currently, the `publishDir` directive is used to define "workflow outputs". It is a "push" model in which individual processes are responsible for deciding which of their outputs should be workflow outputs, and how to publish them. It was created in the days of DSL1 when there was no concept of workflows / subworkflows.

With DSL2 and the introduction of workflows, publishDir is an unwieldy way to define workflow outputs. For one thing, it is extremely difficult to get a full picture of a workflow's outputs because they are scattered across the process definitions. Additionally, a process could be invoked many different times by different workflows, and each workflow might have a different idea of its own outputs. All in all, it no longer makes sense to define workflow outputs in the process definition.

Instead, it should be possible to specify an output schema for a workflow, which defines the full set of published outputs, as well as metadata, in terms of the processes (and subworkflows) that the workflow invokes.

The two options discussed so far are:
1. define the workflow outputs as part of the workflow definition (e.g. an `output` block) which can reference the outputs of subcomponents
2. define the workflow outputs in a separate file similar to what is done for [nf-core modules](https://github.com/nf-core/modules/blob/9e3a40a18f118df5b2a3df2a9a6568bfa85265c6/modules/nf-core/parabricks/fq2bam/meta.yml#L48-L77)

See also #4661 for discussion of where to put additional publish options such as mode and saveAs. The publish mode could become a config option or process directive (e.g. `publishOptions.mode`) to allow for process-specific configuration, or they could become part of the output schema.

bentsherman:
Close #4670 

This PR adds a DSL for defining workflow outputs. It is meant to completely replace the publishDir directive, and centralize the definition of published outputs into one location.

This first iteration is essentially a builder DSL for a directory structure, which also collects the essential elements of the existing publish mechanism: process selectors to select process outputs, and publish options like `pattern` and `enabled` (other options like `mode` can also be supported this way).

Under the hood, there is now a `WorkflowPublisher` which contains all of the publish "selectors" and uses `PublishDir` just like the directive. So the publishing itself is also handled in one place rather than during the task finalization.

See https://github.com/nf-core/fetchngs/pull/275 for a prototype with nf-core/fetchngs


bentsherman:
Some thoughts after working on https://github.com/nf-core/rnaseq/pull/1227

- The `path` / `select` model is very flexible and seems to scale well to arbitrarily complex directory structures. Selectors can be ordered arbitrarily, e.g. by process instead of directory name, which seems useful at the scale of rnaseq. There are often many ways to express the same thing, which I usually don't like, but in this case it feels natural.

- The tokens `path` and `select` are necessary boilerplate with a builder DSL, but could likely be eliminated with a custom Nextflow parser, for example:

    ```groovy
    "${params.outdir}" mode params.publish_dir_mode {
        'fastq' {
            'ASPERA_CLI|SRA_FASTQ_FTP|SRATOOLS_FASTERQDUMP' pattern '*.fastq.gz'
        }
    }
    ```

- The process selectors in rnaseq are quite ugly, because some defensive boilerplate is needed since the selectors are applied globally. If subworkflows outputs can be defined in a modular way (like #4510), it should simplify the selectors. Of course some rnaseq workflow names are just very long 🤷‍♂️ 

- The way I see it, we can either use process selectors (pull) or workflow emits (push) to publish outputs. The latter approach would require every "published" channel to be forwarded all the way up to the top-level workflow emits, which I suspect would require more boilerplate code compared to process selectors.

  We would also need to think about how to publish channel values that aren't files. Process selectors can only select process output files, which forces the user to explicitly serialize in-memory data structures to files in a process. I think I'm fine with this restriction, it is still flexible enough to support any use case. For example, it will discourage the use of `collectFile()` in favor of a custom process, which I think is a good thing because `collectFile()` is a rather annoying exception to the general rule of "only processes should create and modify files".

pditommaso:
Thanks Ben, however [this approach](https://github.com/nf-core/rnaseq/pull/1227/files#diff-37d1fd9d970fac487954bd14de07dcbf27a4ed9d9b2a4f54392d282a34d479d5R351-R375) is substantially different to what was suggest in the [issue](https://github.com/nextflow-io/nextflow/issues/4670#issuecomment-1956040723)

bentsherman:
Not really, though I guess it ended up more like your [original proposal](https://github.com/nextflow-io/nextflow/issues/4042#issuecomment-1875946614)

pditommaso:
The proposal was this 

```
output {
  directory = params.outdir 

   samples {
      topic 'someName'
      path 'subdir/where/store/the/files'
      pattern '*.assembly.fa.gz' 
      index {
         type 'csv'
         path 'where/to/store/samples.csv'
      }
   }

   moreHere {
     ..
   }
}
```

and if I'm understanding correctly we have this 

```
output {
    if( !params.save_reference )
        return

    path(params.outdir, mode: params.publish_dir_mode) {
        path('genome') {
            select 'GUNZIP_.*'
            select 'MAKE_TRANSCRIPTS_FASTA'
            select 'GFFREAD'
            select 'GTF2BED'
            select 'CAT_ADDITIONAL_FASTA'
            select 'PREPROCESS_TRANSCRIPTS_FASTA_GENCODE'
            select 'GTF_FILTER'
            select 'CUSTOM_GETCHROMSIZES'
        }

        path('genome/index') {
            select 'UNTAR_.*'
            select 'STAR_GENOMEGENERATE'
            select 'STAR_GENOMEGENERATE_IGENOMES'
            select 'HISAT2_BUILD'
            select 'HISAT2_EXTRACTSPLICESITES'
            select 'SALMON_INDEX'
            select 'KALLISTO_INDEX'
            select 'RSEM_PREPAREREFERENCE_GENOME'
            select 'PREPARE_GENOME:BBMAP_BBSPLIT'
        }
    }
}
```

bentsherman:
Not sure what your point is. I have explained my choices and the relevant trade-offs here and in the fetchngs PR.

pditommaso:
My point is does not match not only the proposed syntax, but also the semantic (index files creation)

bentsherman:
Again, please see me comments in the fetchngs PR on this point

bentsherman:
Let's just discuss at the retreat if my comments were not clear

marcodelapierre:
adding to conversation, I would like to review together the current proposal vs Ben's hypothesis from last week: 

https://github.com/nextflow-io/nextflow/issues/4670#issuecomment-1965377052

bentsherman:
See also https://github.com/nf-core/fetchngs/pull/302, an alternative approach which uses workflow emits instead of process selectors.

I think it would be better to define the outputs under a `publish:` block in the top-level workflow, in order to distinguish it from output channels. Basically there are two kinds of workflow "outputs", the publishing and the emits.

pinin4fjords:
IMO the distinction between publishing and outputing is a bit odd for outsiders, and I'm not sure it's worth retaining or emphasising if we have a choice. 

I think there are just:

-  'temporary outputs'
-  'outputs we want to keep'

So in my head at least it would be better to just have one outputs concept (emit?), with markers of some sort for things to retain (=publish).

modules/nextflow/src/main/groovy/nextflow/extension/OperatorImpl.groovy
```diff
@@ -1239,4 +1240,12 @@ class OperatorImpl {
                 .getOutput()
     }
 
+    DataflowWriteChannel topic(DataflowReadChannel source, String name) {
+        if( !NF.topicChannelEnabled ) throw new MissingMethodException('topic', OperatorImpl.class, InvokerHelper.EMPTY_ARGS)
+        final op = new IntoTopicOp(source, name)
+        final target = op.apply()
+        OpCall.current.get().outputs.addAll( op.outputs )
+        return target
+    }
```
bentsherman, line 1249: I added a `topic` operator using `tap` as inspiration, but the problem is that it hangs. You can reproduce it by updating the e2e example as follows:
```groovy
nextflow.preview.topic = true

// ...

workflow {
  // ...
  foo | topic('foo')
}

output {
  path('data') {
    // ...
    topic 'foo', mode: 'link'
  }
}
```

@pditommaso maybe you can spot if I'm doing something wrong with the topic channel logic here

pditommaso: Have a look at this 88d7eadd

bentsherman: Thanks for adding the test. However the problem turned out to be that the source `foo.out` is a value channel. I guess it never sends a poison pill and the pipeline hangs.

- converting `foo.out` to a queue channel e.g. `foo.out | flatMap | topic('foo')` fixed it
- weirdly it does not happen when the topic is specified in the process definition
- it turns out `top` has the same problem, it hangs when the source is a value channel, will address separately
- `into` op seems to address it by sending a poison pill when necessary

So I guess we need to either (1) ensure the source channel sends a poison pill or (2) change the topic internal behavior to handle this case

Tbh I'm not sure why the topic behaves this way since it's just a mix op under the hood. I will copy the `into` op approach for now

bentsherman:
Ready for review

Notes:
- the task processor normally sets the publish `overwrite` option based on whether the task was cached. now that the publishing is decoupled from the task lifecycle, we should merge #4729 to recover this behavior via checksum verification

- the resolution of output directory setting between CLI / config / DSL is tricky, see my comment in the code

- likely still need to fix some failing tests

modules/nextflow/src/main/groovy/nextflow/script/WorkflowDef.groovy
```diff
@@ -254,3 +274,116 @@ class WorkflowParamsResolver {
...
+        'overwrite',
+        'pattern',
+        'saveAs',
+        'storageClass',
+        'tags'
+    )
+
+    private Binding binding
+
+    private Path directory = (Global.session as Session).outputDir
```
bentsherman, line 300: @pditommaso I assume we want the CLI option / config option to override the output directory defined in the output DSL. But this is tricky because the output directory has already been resolved and sent to the workflow metadata (`workflow.outputDir`). Not sure what the best way would be to handle this correctly.

bentsherman:
@marcodelapierre you asked me the other day if the file pattern filter would still be supported and I said it wouldn't, but I forgot that `publishDir` also has a `pattern` option to filter by file pattern. So it will be supported that way 👍

modules/nextflow/src/main/groovy/nextflow/script/WorkflowDef.groovy
bentsherman: It is unusual to have a runtime setting like this be configurable in the pipeline script. For example the work directory can only be set on the command line or in the config, then in the script it is assumed to be constant

If we want to treat the output directory as a core runtime setting, maybe it would be better to think of the output DSL as being relative to the output directory without needing to specify it explicitly in the DSL.

It will be overridden 99% of the time anyway, and if the user wants to provide a default value for the pipeline, they can set `outputDir` in the config file

pditommaso: I don't think it's expected to control the overwrite setting via the CLI, only via config file and script

bentsherman: It's pretty common for people to set `params.outdir` from the command line:
```bash
nextflow run nf-core/rnaseq --outdir results ...
```

bentsherman: Whereas they would never set it in the script because it makes more sense to put the default in the config file

bentsherman: @ewels could use your opinion here

tests/publish-dsl.nf
```diff
@@ -0,0 +1,85 @@
...
+  my_combine.out.view{ it.text }
+
+  foo | topic('foo')
+}
+
+output {
+  directory 'results'
+
+  'data' {
+    select align.out[0], mode: 'copy'
```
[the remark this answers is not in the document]
pditommaso: Also not sure we should use `select` here, we could use `from` to avoid introducing another keyword

modules/nextflow/src/main/groovy/nextflow/script/WorkflowDef.groovy
pditommaso: it's enough keep using using 

```
output {
  directory params.outdir 
  ...
}
```

tests/publish-dsl.nf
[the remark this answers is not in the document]
bentsherman: All publish options are supported at any level. The example you proposed can be written using the `defaults` directive:
```groovy
output {
  directory 'result'
  defaults mode: "symlink"

  'data' {
    defaults mode: 'copy'
    select align.out[0]
  }
}
```

This works for every publish option, not just mode, so that users can transition from publishDir to the output DSL without losing functionality.

> Also not sure we should use select here, we could use from to avoid introducing another keyword

I'm not married to a particular keyword, `from` seems fine to me. Would you then use `fromTopic` for the topic selector?

tests/publish-dsl.nf
```diff
@@ -0,0 +1,85 @@
...
+workflow {
+  def input = Channel.of('alpha','beta','delta')
+  align(input)
+
+  def bam = align.out[0].toSortedList { it.name }
+  def bai = align.out[1].toSortedList { it.name }
+  my_combine( bam, bai )
+  my_combine.out.view{ it.text }
+
+  foo | topic('foo')
```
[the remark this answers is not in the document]
bentsherman: I would support a `topic:` block as it's a nice way to give some structure to how topics are used.

What if the user wants to send multiple channels to the same topic? Simplest thing would be for them to use `mix`:

```groovy
topic:
foobar = foo.out | mix( bar.out )
```

modules/nextflow/src/main/groovy/nextflow/script/WorkflowDef.groovy
bentsherman: I thought the point of this was to remove `params.outdir`. If you have it as a CLI option and config option then you don't need a separate param

pditommaso: My point is that it's not blocking right now. It could be added in a later iteration

tests/publish-dsl.nf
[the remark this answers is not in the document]
pditommaso: > The example you proposed can be written using the defaults directive

Not aware of that. From where does it come from? 


> from seems fine to me. Would you then use fromTopic for the topic selector?

Ideally, always `from` could be used

tests/publish-dsl.nf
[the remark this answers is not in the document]
pditommaso: It think also repeting it, should be fine e.g .

```
topic:
foobar = foo.out 
foobar = bar.out
```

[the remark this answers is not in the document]
bentsherman: I was thinking the same, though it would be best to avoid a proliferation of different ways to write the same thing. Maybe we can emphasize only one way in the docs

tests/publish-dsl.nf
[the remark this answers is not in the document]
bentsherman: Sounds good, string -> from topic and channel -> from channel

Check out the docs I wrote in this PR (or the docs deploy preview), everything is explained there

modules/nextflow/src/main/groovy/nextflow/script/WorkflowDef.groovy
bentsherman: Okay then let's not bother with a CLI or config option yet, just the DSL piece and see if users form an opinion on it

I have a feeling that this `directory` method won't be used long-term if there is a CLI + config option, but it doesn't hurt anything if unused

tests/publish-dsl.nf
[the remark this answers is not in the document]
pditommaso: I'm more scared about the fact that it's commonly assumed the latter assignment, override the former one.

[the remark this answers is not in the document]
bentsherman: Yes that would be confusing for many people

I've always thought of topics as a push-based approach, whereas assignment is pull-based. Assignment doesn't fit very well when you can send multiple things to the same topic

modules/nextflow/src/main/groovy/nextflow/script/WorkflowDef.groovy
```diff
@@ -234,23 +253,174 @@ class WorkflowParamsResolver {
...
+    @Override
+    Object invokeMethod(String name, Object args) {
+        if( name == '_into_topic' ) {
+            final args0 = args as Object[]
+            if( args0[0] instanceof DataflowWriteChannel )
+                _into_topic(args0[0] as DataflowWriteChannel, args0[1] as String)
+            else if( args0[0] instanceof ChannelOut )
+                _into_topic(args0[0] as ChannelOut, args0[1] as String)
+            else
+                throw new IllegalArgumentException("Workflow topic source should be a channel")
```
bentsherman, line 292: This is ugly but it works. Maybe there is a cleaner way to do it. For me it's a moot point because this stuff will likely be refactored as we separate the language from the runtime

modules/nextflow/src/main/groovy/nextflow/script/WorkflowDef.groovy
```diff
@@ -234,23 +253,179 @@ class WorkflowParamsResolver {
...
+            if( opts.ignoreErrors )
+                opts.failOnError = !opts.remove('ignoreErrors')
+            new PublishOp(CH.getReadChannel(source), defaults + opts + [path: path]).apply()
+        }
+
+        void from(Map opts=[:], ChannelOut out) {
+            if( out.size() != 1 )
+                throw new IllegalArgumentException("Cannot publish a multi-channel output")
+            from(opts, out[0])
+        }
```
bentsherman, line 416: @pditommaso what do you think about allowing a multi-channel output to be published

e.g. if I specify `from foo.out` and `foo` has multiple outputs, then publish all of them under the hood

Currently it is only allowed if the process has only one output

bentsherman:
rnaseq POC is updated: https://github.com/nf-core/rnaseq/pull/1227

It doesn't work yet because the main subworkflows where all of the publishing happens, RNASEQ and PREPARE_GENOME, are nested within two workflows. Unlike fetchngs, it's hard to access the relevant channels from the top-level workflow

Of course we could propagate these channels through emits and topics, but I'm wondering if a better solution would be for the top-level output definition to be able to import definitions from subworkflows, e.g.

```groovy
output {
  include PREPARE_GENOME
  include RNASEQ
}
```

Then we don't have to go through all the ceremony of propagating the channels. I don't think the nf-core folks will want to do all of that, and I don't think they will remove these wrapper workflows either.

It might be good to get another round of feedback from scidev at this point.

pinin4fjords:
Liking https://github.com/nf-core/rnaseq/pull/1227 now. If we really need the output block (rather than doing something with emit), this is a nice readable way of doing it. 

I actually don't have a problem with being 'ceremonial' and propagating the channels, but the include makes sense as a shortcut.

tests/publish-dsl.nf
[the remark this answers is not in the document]
marcodelapierre: We touched on this briefly while meeting yesterday, but cannot recall where we landed: for the sake of more concise code, it would be good if the keyword `defaults` could be avoided

modules/nextflow/src/main/groovy/nextflow/script/WorkflowDef.groovy
marcodelapierre: or, prompt to be explicit on the channel of interest:

`Cannot publish a multi-channel output` -> `This is a multi-channel output: please specify the channel you want to publish.`

I would lean towards one or the other option also based on typical nf-core usage: do multiple channels from a process get typically published in the exact way (same output path, same options..)?  If so, I would prefer Ben's.   
Otherwise, if output statements typically need to be distinct, probably better to prompt the pipeline dev to be specific.

bentsherman: I did encounter some cases in rnaseq where an entire process was published with no extra options. In this case it would be nice to just say `from foo.out` instead of enumerating all of the outputs of `foo`

pditommaso: Yes, it should be specified the punctual one, either with array notation or emit name

pditommaso:
I think the output `include` hides too much information. One goal is also to make transparent the composition of the output structure

modules/nextflow/src/main/groovy/nextflow/script/WorkflowDef.groovy
bentsherman: Fine, It will just be more verbose. But probably better in this case to be explicit and robust to e.g. adding a process output that you don't intend to publish

tests/publish-dsl.nf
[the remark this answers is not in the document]
bentsherman: I have stuck with `defaults` instead of explicit methods for each publish option so that the syntax for publish options is the same for directory defaults and individual `from` selectors. Otherwise I think it would be harder to understand:
```groovy
'results' {
  mode 'copy'
  'foo' {
    from foo.out, pattern: '*.txt'
  }
}
```

It's not as clear here that `mode` and `pattern` are both publish options being specified with different syntax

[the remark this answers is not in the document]
marcodelapierre: Agree on the non clarity of yout snippet above. What prevents us from adopting the following? 
```groovy
'results' {
  mode 'copy'
  'foo' {
    pattern: '*.txt'
    from foo.out
  }
}
```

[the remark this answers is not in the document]
bentsherman: Because then the pattern applies to the entire directory instead of the individual rule. These options often need to be applied to individual rules, so if they can only be specified for a directory and there are multiple rules in the same directory, you would have to repeat the directory which is burdensome:
```groovy
'results' {
  mode 'copy'
  'foo' {
    pattern '*.txt'
    from foo.out
  }
  'foo' {
    from foobar.out
  }

  // instead of this
  'foo' {
    from foo.out, pattern: '*.txt'
    from foobar.out
  }
}
```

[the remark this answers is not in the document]
pditommaso: This dsl should be an opportunity for simplification. I'm with Marco that options should be at group level not at from level. It's very likely that the same pattern needs to be applied to all `from` statements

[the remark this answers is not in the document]
bentsherman: > It's very likely that the same pattern needs to be applied to all from statements

If you look at the rnaseq prototype you will see that this is simply not true

I am all for simplification but we cannot simplify to the point of not supporting the behavior that is needed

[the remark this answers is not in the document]
bentsherman: The output DSL has already simplified things in three major ways:
1. the publish mode is specified once for the entire directory instead of once for every rule
2. putting the directory structure first, instead of specifying the full path for every rule, removes a lot of duplication of base paths
3. the pattern option is often not needed because you can select individual channels instead of entire processes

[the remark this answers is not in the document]
bentsherman: Then you need to demonstrate the bad pattern and offer something better. Right now you are hand waving and not really engaging with the actual pipeline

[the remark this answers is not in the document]
marcodelapierre: Thanks for clarifying and for all the feedbacks.   

I now better appreciate the merits of enabling both path-level and rule-specific options, considering some big pipelines. 

I think I am a bit distracted by `defaults` both adding verbosity and at the same time not being super expressive about the meaning.

In other words, I would not find this confusing (open to other views, incl scidev):

```groovy
'results' {
  mode 'copy'
  'foo' {
    mode 'symlink'
    from foo.out, pattern: '*.txt'
  }
}
```

As long as we document that all publish options can be applied at both path level (standalone line) and rule level (comma separated within a `from` specification). 

There might be some degree of error proneness (spare or missing comma may trigger unwanted results), but we could promote the best practice of always having the path level options stated **before** the list of `from` specifications.

docs/channel.md
```diff
@@ -491,21 +483,38 @@ process bar {
...
+
+    emit:
+    bar.out
+}
+```
+
+Finally, the `Channel.topic()` factory can be used to consume the resulting channel for a given topic name, which can be used like any other channel:
 
 ```groovy
 channel.topic('my-topic').view()
```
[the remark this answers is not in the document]
bentsherman: Reducing the topic to a string is too much syntax sugar in my opinion, there needs to be some marker to clarify that it is a channel topic

bentsherman:
Updates from latest iteration:
- rename `topic:` section to `publish:`
- instead of channel topics, use a simple mapping of channel -> name
- change the output DSL to be a simple flat list of rules which map an arbitrary name to a set of publish options
- replace `defaults` with dedicated methods for each option i.e. `mode`, `enabled`, etc
- if a rule is not defined for a name, use the name as the path

TODO:
- [ ] implement `publish:` section for processes
- [ ] allow publish rules for a process/subworkflow to be overwritten when invoked by a workflow

Notably, the current prototype does not use topics at all. It simply maintains a global map of channels to names (i.e. "topics"), which is more amenable to overriding in different contexts. This way, the publishing logic is completely independent of our original use case for topics of collecting the tool versions. Of course we could still use topics under the hood if we think it would be useful.

When it comes time to publish, the output DSL simply publishes each channel from this global mapping with any applicable rules, using the "topic" name as the path unless otherwise specified. fetchngs becomes extremely simple as a result, will update the PR shortly. rnaseq will be updated later.

bentsherman:
Sorry just realized I didn't push the changes for the process `publish:` section, pushing now

pditommaso:
Would be useful to have a version https://github.com/nextflow-io/rnaseq-nf using the output dsl, too complex to follow nf-core/rnaseq

bentsherman:
Sure I'll make a PR for that

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
```diff
@@ -100,24 +90,22 @@ class PublishOp {
...
+        if( path == null )
             return null
-        def result = getTaskDir0(path, session.workDir)
-        if( result == null )
-            result = getTaskDir0(path, session.bucketDir)
-        return result
+        return getTaskDir0(path, session.workDir.resolve('tmp'))
+            ?: getTaskDir0(path, session.workDir)
+            ?: getTaskDir0(path, session.bucketDir)
     }
```
bentsherman, line 100: Normally, the `overwrite` publish option is set to `!task.cached`. Since we are inferring the task work directory here, we should be able to recover also whether the task was cached by querying the cache db.

However that could put a lot of pressure on the cache db. Might be able to alleviate with a second cache in the publish op of task dir -> cached true/false.

Or we can implement #4729 and overwrite based on stale-ness. We would want to have a "lenient" mode (compare file save) and "deep" mode (compare file checksums) as there is a trade-off between performance and accuracy.

pditommaso: I'd ignore the `task.cached` attribute and just check if the file metadata changed has it's done for regular resume mode. Cfr https://github.com/nextflow-io/nextflow/blob/509f271e70bd99f0dd867ea07ecf9c7831342091/modules/nf-commons/src/main/nextflow/util/CacheHelper.java#L382-L382

tests/publish-dsl.nf
```diff
@@ -0,0 +1,87 @@
...
+
+  publish:
+  align.out[0]    >> 'data'
+  align.out[1]    >> 'data'
+  my_combine.out  >> 'more/data'
+  foo.out         >> 'data'
+}
+
+output {
+  directory 'results'
```
pditommaso, line 78: I think `path` can be used instead of `directory` both for internal and `process output` consistency

pditommaso:
Think this feature is taking a nice shape. Ideally, we could introduce as an experimental feature in 23.04 and final in 23.10. 

Possible improvements: 
* remove the need for quotes in the `publish` and `output` sections, and only keep when publish/output scope contains a special char e.g. `/` or `.` 
* allow the use of multichannel output in the `publish` section as a shortcut for many single-channel declarations 
* add automatic creation of index files

tests/publish-dsl.nf
bentsherman: I considered the same, but there is an important difference with `directory` -- whereas with other publish options the rule overrides the generic option, the `path` in a rule is appended to the top-level `path` i.e. `directory`. So I think the different name is actually helpful to denote the different semantics

bentsherman:
> remove the need for quotes in the publish and output sections, and only keep when publish/output scope contains a special char e.g. / or .

This should already work in the output block, though I was thinking to disallow it in the formal grammar. It is yet another example of "multiple ways to do the same thing" which we have been lamenting lately and I thought we would try to minimize going forward.

I think it would be especially confusing to allow the target name in the `publish:` section to be an identifier or string. As a novice user I would wonder if there is some significance to why some `publish:` rules point to an identifier while others point to string. For the identifier I would wonder if that's a variable name I can use somewhere, if I could define a variable with the same name in the workflow, if that would somehow be connected to the publish rule, etc.

bentsherman:
Index file generation and shortcut for multi-channel would both be worth exploring in a second iteration in my opinion

marcodelapierre:
> Index file generation and shortcut for multi-channel would both be worth exploring in a second iteration in my opinion

I would consider prioritising multi-channel shortcut sooner rather than later, as it will make the pipeline more concise straight from the feature first release, helping showcase its impact

pditommaso:
> Index file generation and shortcut for multi-channel would both be worth exploring in a second iteration in my opinion

I would consider prioritising multi-channel shortcut sooner rather than later, as it will make the pipeline more concise straight from the feature first release, helping showcase its impact

I'd consider equally important and include in this iteration

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
bentsherman: I had to do a bit of hacking on `CacheHelper` to make it work. It normally uses the absolute file name, but here we want to use relative names. It's a bit ugly, might be cleaner to make CacheHelper into a builder i.e. `new HashBuilder(opts).with(...).with(...).build()`

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
```diff
@@ -199,7 +203,7 @@ class PublishDir {
             result.pattern = params.pattern
 
         if( params.overwrite != null )
-            result.overwrite = Boolean.parseBoolean(params.overwrite.toString())
+            result.overwrite = params.overwrite
```
pditommaso, line 206: what guarantees that's a `Boolean` ?

bentsherman:
I'm starting to come around to the idea of redirecting to `null` to disable publishing. Marco had suggested redirecting to an explicit target which uses the `enabled` option, but this is essentially the same thing with extra steps:
```groovy
workflow {
  ch_foo = Channel.of( 1, 2, 3 )

  publish:
  ch_foo >> 'foo'
}

publish {
  directory 'results'

  'foo' {
    enabled params.save_foo
  }
}
```

Instead you could do:
```groovy
workflow {
  ch_foo = Channel.of( 1, 2, 3 )

  publish:
  ch_foo >> params.save_foo ? 'foo' : null
}

publish {
  directory 'results'
}
```

These seems more concise to me. Even though I don't like null, I find this very compelling because it eliminates the need to customize a bunch of rules in the `publish` block just to use `enabled`. It is also easy to implement. I think I will try this.

bentsherman:
Another improvement I'm considering from looking at rnaseq is the ability to redirect entire publish targets in addition to individual channels.

In rnaseq there are many targets being redirected at the top-level like so:
```groovy
publish {
  // ...

  'dupradar' {
    path "${params.aligner}/dupradar"
  }

  // ...
}
```

What if you could do this redirect in the workflow `publish:` section:
```groovy
workflow {
  // ...
  publish:
  'dupradar' >> "${params.aligner}/dupradar"
}
```

It is more concise, but my concern is that it is more error prone. With channels you are restricted to the channels in the workflow body, but with this you can redirect *any* target *anywhere* in the pipeline. It's much easier to screw things up. Maybe it could work if you could allow only targets mapped to channels in the current workflow body, but I fear that is getting too complicated.

marcodelapierre:
On redirecting publish targets from within the `publish:` section:

Would this only be allowed at the workflow level, not process level? If so, maybe it wouldn't be as error prone.   
As in, typically channels are forwarded to publish targets at the process and sub-workflow levels; but then finalisation of the publish target properties is only allowed at the workflow level, i.e. essentially with same scope as the `publish { }` block.  So, redirection can be used for conciseness with respect to the full block, which would only be required if options other than the path need to be specified.   

Though I am also wondering whether this would confuse developers/users as to when to use one form or the other.

bentsherman:
That's a good point, it would be safe if only allowed in the top-level workflow. But I agree that it would be confusing if you can re-route paths in the top workflow or publish block but not in a named workflow. Unless we can eliminate the publish block entirely in favor of using the top workflow, I'd rather leave it as is.

modules/nextflow/src/main/groovy/nextflow/extension/PublishIndexOp.groovy
bentsherman, line 1: Do you think the index file should be formally published so that a publish event is emitted for it?

bentsherman:
Here is one way we could eliminate the publish block entirely:

- add `-publish-dir` CLI option and `publish.directory` config option like we discussed, by default it is unset and nothing is published

- allow publish options (`contentType`, `ignoreErrors`, `mode`, `overwrite`, `storageClass`, `tags`) to be set globally in the config file via `publish` scope. target-specific options are not supported
  - publish retry options are currently under `nextflow.publish`, I would recommend moving to `publish` scope
  - need to look into process-specific mode in sarek and see if there's an alternative solution

- allow the top-level workflow to redirect entire targets as suggested by Marco

- no more index DSL 😞 instead, provide the `mergeCsv()` function for writing CSV files and let users create index files in the pipeline the normal way
  - note: fetchngs likely wouldn't use the current index DSL because (1) you can only create one index file per target but fetchngs creates two and (2) the index file must go in the same directory as the published files but fetchngs puts the samplesheets in a different directory

This way, publish behavior would be defined entirely in the workflow `publish:` sections and the `publish` config scope. Not advocating for it yet, just some food for thought.

bentsherman:
This is what @ewels said about process-specific publish mode https://github.com/nextflow-io/nextflow/issues/4661#issuecomment-1896175464:

> Definitely, I know of at least one use case for having mode process specific: some tools hate symlinks so you might want to set mode to copy for a single process (not the whole pipeline as that wastes disk space).

Seems like more of a practical issue than a fundamental one. I think we could go ahead with only global publish options and address the above comment in other ways. Off the top of my head, you can always fall back to `copy` and use automatic cleanup 😄

bentsherman:
Alternative way to preserve the index DSL if everything else is moved to config / workflow `publish:`

```groovy
index {
  "foo/" {
    path "index.csv"
    header true
    // index options ...
  }

  "bar/" {
    // ...
  }
}
```

This feels like the beginning of an output schema definition, maybe better to explore later as part of that effort.

marcodelapierre:
Thanks for your thoughts and reflections on the hypotethical dropping of the `publish` block Ben, quite interesting.

How restrictive would be to developers not being able to define target-specific options? Probably a good question for SciDev?  Judging from the rnaseq and fetchngs POCs, doesn't seem to be a major issue.   
Again on this, in the rnaseq POC all the target-specific statements in the top-level [`main.nf`](https://github.com/nf-core/rnaseq/pull/1227/files#diff-6401496ba455b9488ffa902a6e4d7732b2c60ff2d77c5c3ef96b28a7ac7d3b28) are `path` redefinitions;  these would then be handled in a `publish:` section thanks to whole-target `>>` redirection, right?

@pditommaso how do you feel about the `index` DSL suggestion in Ben's previous comment? I don't have a strong opinion on this at the moment.

pditommaso:
I think we should keep the publishing block because because it's the goal of this feature! 

What I'm more concerned is the overlapping in between `publish:` and `publish { }` definitions which is entirely confusing.  

It would be better to use `output {}` instead of `publish { }` and keep `publish` for the target definition

bentsherman:
@pditommaso I'm essentially proposing that the `publish` block should become config:
```groovy
// nextflow.config
publish {
  directory = params.outdir // this must be defined to enable publishing
  mode = 'copy'
  overwrite = 'standard'
  // ...
}
```

This seems to be aligned with separating the *what* from the *how* of publishing.

That would actually free up the `output` block in the DSL to focus on defining the output schema, which we haven't really thought through. As a first iteration, it could just be a collection of index files:
```groovy
workflow {
  // ...
}

output {
  index {
    target "foo/"
    path "index.csv"
    header true
    // index options ...
  }

  index {
    target "bar/"
    // ...
  }
}
```

pditommaso:
I still see more valuable a single output block in the script context (for conciseness) but giving also the ability to override/define partially some settings (or even all?) via the config. 

This could open the ability to reuse the same `output` across different pipelines

bentsherman:
But what is the purpose of this output block? You can't publish channels, that can only be done in the process and workflow `publish:`. The only thing left is configuration and index files. The purpose seems unclear and fragmented

bentsherman:
nf-core/rnaseq POC is working on aws and google ✅

pditommaso:
No way, it hangs on Google Batch with Fusion consistently

bentsherman:
Are you talking about job arrays?

pditommaso:
Yes, sorry

marcodelapierre:
A note while testing a simple pipeline, [rnaseq-nf](https://github.com/nextflow-io/rnaseq-nf/pull/27), with the output DSL.  
Here "simple" implies, among other aspects, not having named workflows or sub-workflows; hence `publish` declarations of channel happen at the module level.  

A channel must be named in order to be publishable.  
I was wondering whether there should be some syntax for treating unnamed output channels (e.g. akin to `<channel>.out` in a workflow), but I would say it is probably an overshooting. 

Just flagging, probably this doesn't need extra thinking.

bentsherman:
Not sure what you mean, you can publish channels from the `.out` of a process/workflow

marcodelapierre:
Let me try and clarify, in the case of [rnaseq-nf](https://github.com/nextflow-io/rnaseq-nf/pull/27/files#diff-fc43b51ca38d78479af71264a7ea0188ceecb825e30fe2d8315bec7c24532113) we need to give an emit name to a module channel to be able to publish it, e.g. see this diff:

```
      output:
-    path "fastqc_${sample_id}_logs" 
+    path "fastqc_${sample_id}_logs", emit: logs
+
+    publish:
+    logs >> '.'
```

So my thinking point is whether we should have a syntax for the `publish` section that does not require giving the emit name `logs` to the channel to be published.  As I was saying above, my personal answer is that we don't need it, but still wanted to leave this for your consideration.

bentsherman:
With static types we are moving towards a syntax where every process output will have a name by definition (`<type> <name> = <value>`), so in time this will not be an issue

marcodelapierre:
thanks for the reminder Ben

marcodelapierre:
I am happy with the testing of the feature with the Fetchngs POC (linux vm); there I disabled nf-validation.

For the rnaseq POC, how do I get around the following error?
```
ERROR ~ Install is not supported on dev mode - Missing plugin nf-validation 1.1.3
```

(nf-validation here seems needed to use `fromSamplesheet`)

bentsherman:
You can install the local build with `make install`. It will replace the installation of your current Nextflow version and allow you to run pipelines with third-party plugins. It won't use the local build of core plugins like nf-amazon, but that's okay because this PR doesn't modify any core plugins.

marcodelapierre:
Testing Rnaseq was also successful. As Ben expected, output directory layout is not identical, but output files are there.

modules/nextflow/src/main/groovy/nextflow/script/PublishDef.groovy
```diff
@@ -0,0 +1,276 @@
...
+}
+
+/**
+ * Implements the DSL for publishing workflow outputs
+ *
+ * @author Ben Sherman <bentshermann@gmail.com>
+ */
+@Slf4j
+@CompileStatic
+class PublishDsl {
```
pditommaso, line 64: Maybe this now should be `OutputDsl`?

pditommaso: Also would not make sense to move it into its own source file?

pditommaso: There are no tests for this class

modules/nextflow/src/main/groovy/nextflow/extension/PublishIndexOp.groovy
pditommaso: Good point, likely is should like any other file

modules/nextflow/src/main/groovy/nextflow/script/OutputDsl.groovy
```diff
@@ -0,0 +1,250 @@
...
+            opts.overwrite = 'standard'
+
+        final path = opts.path as String ?: name
+        if( path.startsWith('/') )
+            throw new ScriptRuntimeException("Invalid publish target path '${path}' -- it should be a relative path")
+        opts.path = directory.resolve(path)
+        return opts
+    }
+
+    static class TargetDsl {
```
pditommaso, line 153: Not 100% sure, but likely inner class should be annotated `@CompileStatic`

bentsherman: Since I've been digging though the groovy codebase... I can confirm it is not needed 😄 

https://github.com/apache/groovy/blob/55d43696d2b13fa03067ef3c2b200eb7c84148ca/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java#L450-L451

modules/nextflow/src/main/groovy/nextflow/script/PublishDef.groovy
pditommaso: still not test of `OutputDsl` 😭

docs/workflow.md
```diff
@@ -403,3 +392,288 @@ workflow {
 ```
 
 In the above snippet, the initial channel is piped to the {ref}`operator-map` operator, which reverses the string value. Then, the result is passed to the processes `foo` and `bar`, which are executed in parallel. Each process outputs a channel, and the two channels are combined using the {ref}`operator-mix` operator. Finally, the result is printed using the {ref}`operator-view` operator.
+
+(workflow-output-def)=
+
+## Publishing outputs
```
pditommaso, line 398: I think the core concept should be "Workflow output definition" which essentially allows structuring the output of a workflow execution and collect publishing targets

pditommaso:
While testing the output index I've got this 

```
ERROR ~ Index file record must be a list or map: /Users/pditommaso/Projects/rnaseq-nf/work/10/f200de889f31fcba450bab2d2964a3/fastqc_ggal_gut_logs [UnixPath]

```

What about instead creating a simple index as "file name", "file path" ?

pditommaso:
Made a few tests and works like a charm. I was expecting a much bigger impact in the codebase, great work Ben.

I've commented a few things to address above

docs/workflow.md
bentsherman: I hesitate to call it "workflow output definition" here because there is already a section called "workflow outputs (emit)". Even with the "emit" qualification, it still feels confusing to have two different sections about workflow "outputs". That is why I focus on "publishing" here to better distinguish from emits.

pditommaso:
Better, but it's not an index file 

```
"/Users/pditommaso/Projects/rnaseq-nf/results/fastqc/fastqc_ggal_gut_logs"
```

I was expecting as 

```
"fastqc_ggal_gut_logs", "/Users/pditommaso/Projects/rnaseq-nf/results/fastqc/fastqc_ggal_gut_logs"

**Comment by bentsherman** [comment]
That can be done with the index mapper:
```groovy
index {
  mapper { file -> [file.name, file] }
}
```

pditommaso:
I know, but it would be nice to have it as default behaviour

docs/workflow.md
pditommaso: Ok, for now. We may want to reshape it a bit when promoting it stable

pditommaso:
and it's ready!

PR #4808, opened 2024-03-11
Add defaults configuration settings 

pditommaso:
This PR aims to make default settings configurable via the nextflow configuration file. 

The primary goal is to allow the control the defaults of `publishDir` directive. It can be extend to cover the settings of almost any other configuration attribute.

Solves https://github.com/nextflow-io/nextflow/issues/4186


bentsherman:
I don't see the point in doing this. We are planning to replace `publishDir` with the workflow output DSL, which I have shown can provide a simple way to provide publish defaults without so much repetition (https://github.com/nf-core/fetchngs/pull/275). And when you put aside `publishDir`, there is no other process directive that would benefit from this sort of defaults scope.

pditommaso:
Closing for now

PR #4817, opened 2024-03-14
added labels in ContainerConfig

munishchouhan:
This PR will provide config option to add labels to the container image build by wave


marcodelapierre:
minor note: we will need to document the new option in the Wave doc page.

PR #4842, opened 2024-03-22
Docs sidebar: Show version

ewels:
Show the version of the docs in the sidebar:

| Stable release | Edge release | No release (show current branch) |
|--|--|--|
| CleanShot 2024-03-22 at 14 32 52@2x | CleanShot 2024-03-22 at 14 34 08@2x | CleanShot 2024-03-22 at 14 31 57@2x |

Also auto-generate the version string from git commands, so that this no longer needs to be bumped manually in `conf.py`.

I was going to put in a version switcher dropdown, but because this will only affect docs builds going forwards I aborted because it'll be mostly non-functional.


bentsherman:
Since you are fetching the version automatically in `conf.py`, there are a few related changes which you can see by doing a project search for "conf.py":
- remove update of `conf.py` from `build.gradle`
- remove the note about not changing `conf.py` from the PR template
- I think `sync-doc.sh` can be deleted because it was used by the old website build system?

ewels:
Great spots @bentsherman, addressed 👍🏻

docs/conf.py
```diff
@@ -283,7 +296,7 @@
 epub_title = u'Nextflow'
 epub_author = u'Paolo Di Tommaso'
 epub_publisher = u'Paolo Di Tommaso'
-epub_copyright = u'2013-2018, Centre for Genomic Regulation'
+epub_copyright = u'2013-2024, Seqera Labs, Centre for Genomic Regulation'
```
bentsherman, line 299: I think the copyright can just be Seqera, @pditommaso to confirm

docs/conf.py
```diff
@@ -66,9 +68,20 @@
 # built documents.
 #
 # The short X.Y version.
-version = '24.02'
+# version = '24.02' # AUTO-GENERATED BELOW
 # The full version, including alpha/beta/rc tags.
-release = '24.02.0-edge'
+# release = '24.02.0-edge' # AUTO-GENERATED BELOW
```
bentsherman, line 73: To be deleted before merge

docs/conf.py
ewels: True - this is what's used elsewhere. I'll just copy this:

https://github.com/nextflow-io/nextflow/blob/744fff2a7753b75b0125caf073ad4b30a89433ef/build.gradle#L2

ewels:
Added a little dropdown to switch between versions:

CleanShot 2024-03-22 at 21 51 57

Goes to an error page because I don't have multiple versions running locally, but hopefully you get the idea.

bentsherman (approved):

ewels:
Thanks both!

bentsherman:
@ewels I just noticed from the commit log that you didn't squash merge this PR, please remember to do that in the future.

Github likes to play around with the default button option, that might be what threw you off.

@pditommaso I believe it is possible to select which merge options are available for a repo in the repo settings, you should be able to allow only the "squash merge" option

ewels:
Oops, sorry. Yes I just clicked Merge - whatever was shown by default.

PR #4860, opened 2024-03-27
Add `CODEOWNERS` for docs

ewels:
Add a `.github/CODEOWNERS` file that tags the `/docs/` and `.github/workflows/` folders with the @nextflow-io/docs team. This should auto-assign them as reviewers on any PR that changes this content.

The intention is not necessarily that a review from them is required to merge, but rather to ensure that they are aware of updates to the Nextflow docs.


.github/CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+# DOCUMENTATION -------------------------------------------------------
+# Docs folder ownership
+/docs/ @nextflow-io/docs
+/.github/workflows/ @nextflow-io/docs
```
pditommaso, line 4: is `github/workflows` really needed?

ewels: Not really - I was using the Platform repo as reference and they had it there. Both workflows in this repo are to do with publishing the docs (`Test docs build` and `Check for spelling errors`) so I figured why not. But happy to remove if you'd prefer.

pditommaso: I'd not expect docs team to work directly on GH actions. I'd keep just `/docs/`

ewels:
```suggestion
```

ewels: Done ✅

pditommaso (approved):

PR #4841, opened 2024-03-22
Azure Batch configurable startTask

adamrtalbot:
Azure Batch: Configurable startTask per pool

Additions:
 - Make Azure Batch pool startTask configurable.
 - Can be used to modifiy the existing worker node. Use cases:
    - Install more recent version of AzCopy
    - Install monitoring software like Azure Insights
    - Use NVME SSDs for storage
    - Flash Dragen firmware to FPGA chips

Pool specific so you can differentiate between different hardware requirements.

Closes #4259

~Also spotted a bug with the dev version and the container configuration so fixed it (first commit, can reorganise if we want it in a second PR).~ Closed with https://github.com/nextflow-io/nextflow/issues/4854

To use it, you can try this config:

```
process {
    executor = 'azurebatch'
}

azure {
    storage {
        accountName = "$AZURE_STORAGE_ACCOUNT_NAME"
        accountKey = "$AZURE_STORAGE_ACCOUNT_KEY"
    }
    batch {
        location = "$AZURE_LOCATION"
        accountName = "$AZURE_BATCH_ACCOUNT_NAME"
        accountKey = "$AZURE_BATCH_ACCOUNT_KEY"
        autoPoolMode = true
        deletePoolsOnCompletion = true
        pools {
            auto {
                startTask = "echo 'HELLO WORLD'"
            }
        }
    }
}
```

PR #4883, opened 2024-04-04
Improve control on azcopy install

pditommaso:
This PR improves the control over the installation of the `azcopy` tool used by Azure batch execution by adding the ability to disable its installation using the `none` option or when Fusion is enabled. 

Moveover it prevents the installation in the compute node when the `task` mode is specified


adamrtalbot (approved):
Simple enough but will conflict with #4841. We can merge this then fix it there.

PR #4884, opened 2024-04-04
Hyperqueue breaking change (memory request)

njspix:
Hyperqueue version ≥ 0.17 needs memory requested in megabytes, not bytes. 
(https://github.com/It4innovations/hyperqueue/blob/main/CHANGELOG.md#breaking-change-1)
Worth adding a hyperqueue version check somehow??

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/HyperQueueExecutor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/HyperQueueExecutor.groovy
@@ -64,7 +64,7 @@ class HyperQueueExecutor extends AbstractGridExecutor {
 
         // No enforcement, Hq just makes sure that the allocated value is below the limit
         if( task.config.getMemory() )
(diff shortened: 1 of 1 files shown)


pditommaso:
It's a bit ugly. Is there no way to keep backward compatibility ?

njspix:
I agree. Hyperqueue (and its support in nextflow) is still in beta, I think, so breaking changes are not unexpected, but agree would like to avoid if all possible. 
Currently it raises a job submission error. Do you think it would be feasible to check the content of the error (I think it's pretty specific, something like "couldn't coerce requested memory to <some type>") and then submit all remaining jobs with the memory in mb?

Or maybe it's simpler to just run a shell `hq --version` and check the output?

pditommaso:
I'd avoid adding extra logic to check the error or the version. Worth adding a note in documentation on the minimal version of HyperQueue expected. 

Also note there are some failing tests

njspix:
Closing for now, not urgent. Thanks!

pditommaso:
Merged manually via b45f7c48

Issue #4870, opened 2024-04-01
Nextflow process output type 'eval' does not work with bash pipes


stevekm:
Testing the new `eval` output type

https://www.nextflow.io/docs/master/process.html#output-type-eval

along with the `topic` channels https://www.nextflow.io/docs/master/channel.html#topic

- `main.nf`

```groovy

nextflow.enable.dsl=2
nextflow.preview.topic = true

process FOO {
    tag "${id}"
    container "quay.io/biocontainers/multiqc:1.14--pyhdfd78af_0"

    input:
    val(id)

    output:
    tuple val(task.process), eval('multiqc --version'), topic: versions

    script:
    """
    """
}

process BAR {
    tag "${id}"
    container "quay.io/biocontainers/multiqc:1.14--pyhdfd78af_0"

    input:
    val(id)

    output:
    tuple val(task.process), eval('multiqc --version | sed -e "s/multiqc, version //g"'), topic: versions

    script:
    """
    """
}

workflow {
    input_ch = Channel.from("Sample1") 

    FOO(input_ch)
    BAR(input_ch)

    channel.topic('versions').view()
}

```

run like this

```
$ NXF_VER=24.02.0-edge nextflow run main.nf

 N E X T F L O W   ~  version 24.02.0-edge

 ┃ Launching `main.nf` [happy_tesla] DSL2 - revision: 432d989a79

executor >  local (2)
[66/012ea3] process > FOO (Sample1) [100%] 1 of 1 ✔
[e2/7aff93] process > BAR (Sample1) [100%] 1 of 1 ✔
[FOO, multiqc, version 1.14]
[BAR, ]

```

I have tried a bunch of different one-liners in the `eval` statement and none of them seem to return any output when there is a bash `|` pipe in the command.

This is surprising since, as far as I can tell, the vast majority of one-liners used for e.g. return process "versions" output require editing with `sed` or similar to clean up the version string. 

Not sure if this is a bug?

stevekm:
after inspecting the `.command.sh` for these tasks, I see this

```bash
nxf_eval_cmd() {
    {
        IFS=$'\n' read -r -d '' "${1}";
        IFS=$'\n' read -r -d '' "${2}";
        (IFS=$'\n' read -r -d '' _ERRNO_; return ${_ERRNO_});
[...]

PR #4887, opened 2024-04-05
Fix eval output type via bash -c wrapping

marcodelapierre:
This PR stems from issue #4870 , and fixes the shell implementation of the `eval` output type by wrapping the command within a `bash -c "<cmd>"` expression.  In this way, for instance, piping of commands becomes supported (see issue). 

Unit tests have been updated accordingly.

Note at this stage there is no special treatment of quote characters, may require double escaping in the input with this implementation.  @bentsherman have you any seen specific needs/use cases in this space?

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -235,7 +235,7 @@ class BashWrapperBuilder {
         // out eval
         for( Map.Entry<String,String> eval : outEvals ) {
             result << "#\n"
-            result <<"nxf_eval_cmd STDOUT STDERR ${eval.value}\n"
+            result <<"nxf_eval_cmd STDOUT STDERR bash -c \"${eval.value}\"\n"
             result << 'status=$?\n'
             result << 'if [ $status -eq 0 ]; then\n'
             result << "  echo $eval.key=\"\$STDOUT\" >> ${TaskRun.CMD_ENV}\n"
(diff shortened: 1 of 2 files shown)


bentsherman:
This is why I was concerned about wrapping a command when you could just use an `env`, because now we have to think about all the different edge cases involved. So we might need to put a boundary on how much we want to support for `eval` or else we'll be chasing edge cases like this forever.

But for now it should be enough to wrap in `bash -c "..."` as you have done and also escape double quotes in the command

marcodelapierre:
ok, handling double quotes properly now

modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
```diff
@@ -235,7 +235,7 @@ class BashWrapperBuilder {
         // out eval
         for( Map.Entry<String,String> eval : outEvals ) {
             result << "#\n"
-            result <<"nxf_eval_cmd STDOUT STDERR bash -c \"${eval.value}\"\n"
+            result <<"nxf_eval_cmd STDOUT STDERR bash -c \"${eval.value.replace('"','\\\"')}\"\n"
```
pditommaso, line 238: is it really needed? since the outmost quote are escape, the inner ones are ok.

marcodelapierre: yes, needed, the resulting `.command.sh` snippet needs to be:
```
bash -c "  command with \"double\" quotes   "
```

pditommaso (approved):

PR #4888, opened 2024-04-05
Fix for Azure certificates compatibility

### LINKED ISSUE #4828: [Azure Batch] Cannot use a different distro as the host because of the forced mount of /etc/ssl/certs and /etc/pki

When using a centos/redhat container in a ubuntu host, the centos container should run normally

Start an Ubuntu azure batch instance and run

```
sudo docker run -ti -v /etc/ssl/certs:/etc/ssl/certs:ro -v /etc/pki:/etc/pki:ro
```

### Program output 

```
docker: Error response from daemon: failed to create shim: OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: rootfs_linux.go:75: mounting "/etc/ssl/certs" to rootfs at "/etc/ssl/certs" caused: mkdir /mnt/resource/docker/overlay2/1639d1e2a072f2b3679abf18c69b1789b4339cd7deedd187786191ed0a52a6a4/merged/etc/pki/tls: read-only file system: unknown.
ERRO[0000] error waiting for container: context canceled
```

### Environment 

* Nextflow version: 23.10.1
* Operating system: Ubuntu Azure default VM

### Additional context

This seems to be caused by something that is not configurable here
https://github.com/nextflow-io/nextflow/blob/master/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L398

Would it be possible to make it configurable?

adamrtalbot:
When running a CentOS container on a Debian based virtual machine on Azure (default), an error was thrown while mounting the certificates. This was because we tried to add them by default. This PR changes the behaviour so this no longer happens. A user can still renable these options using the `container.runOptions` as per any other Docker runner.

Fixes #4828 

Signed-off-by: Adam Talbot <12817534+adamrtalbot@users.noreply.github.com>


validation/azure.config
```diff
@@ -9,6 +9,8 @@ process {
   queue = 'nextflow-ci'
 }
 
+docker.runOptions = '-v /etc/ssl/certs:/etc/ssl/certs:ro -v /etc/pki:/etc/pki:ro'
+
```
pditommaso, line 13: This doesn't look nice

adamrtalbot: Removed.

pditommaso:
It looks like this was never solved. Inclined to close it

PR #4890, opened 2024-04-05
Add finalizer thread to task polling monitor

bentsherman:
Possible fix for a customer issue:

- pipeline with many parallel tasks, each task generates several files
- maxForks set to 500, but only a few jobs are running at a time
- initial throughput is high but degrades after the first wave completes
- Nextflow reports a "full" queue (i.e. submitted+running = 500) but this could be misleading because "running" includes completed tasks waiting to be finalized
- task finalization takes ~2s per task, likely due to querying S3 for task output files

This PR creates an additional queue and processing thread for finalizing tasks, with the idea that the task monitor need not wait for a task to finalize before checking the next task.

Another angle would be to optimize the querying of output files, as it was designed around local filesystems rather than remote object storage. This fix simply removes the finalization as a blocker for now.


starnsd:
Hey all, anything I can do to help move this along? Gabriel from Profound is still awaiting an update on this.

bentsherman:
@starnsd see my last message to @robsyme on the ticket. We need to try to replicate the bottleneck with a minimal example and confirm that this change actually removes the bottleneck. If Rob or someone from support could do this it would be a great help. My message in the ticket gives some guidance for designing such a test case

starnsd:
@robsyme Is this something you can test? I'd like to get some updates to the client if possible.

Also, potentially dumb question, but @bentsherman assuming this fix gets approved and merged how will this be released? Do we releases patches or will this wait for another major version?

bentsherman:
Next edge release 24.03.0-edge is going out next week, if we can move quickly we might be able to make that release. Otherwise we can shoot for 24.04.0 which will likely go out right before Boston.

pditommaso:
Not keen to include this without more evidence and more tests.

bentsherman:
@pditommaso I believe I have reproduced the bottleneck with this test case:
```bash
nextflow run bentsherman/nf-tests -r task-finalize-bottleneck
```

It launches 20,000 `exec` tasks that each take 5 s, up to 500 at a time. If you add a 2 second delay to the task finalize in Nextflow (i.e. `Thread.sleep(2000)`), you will see that the first wave of 500 tasks finishes quickly and then it slows down to 1 task per 2 seconds.

Now using this PR (also with the 2 second delay), the slowdown does not happen. However, because the tasks are still finalized one at a time, the pipeline will spend about ~11 hours finalizing tasks even after all jobs have finished.

We have two options:
1. try to make `TaskProcessor::collectOutputs()` more efficient against S3
2. use a thread pool to finalize tasks

I'm gonna go ahead with (2) because it is much simpler. It will basically become like the publishing issue, you can either allocate more threads to make it go faster or use virtual threads.

bentsherman:
If the thread pool feels like overkill, we could use it only when `session.workDir.fileSystem != FileSystems.default`, i.e. when the work directory is remote

bentsherman:
When I resume the run with the same 2 second delay in `collectOutputs()`, I do not see such a large delay, only a few seconds. I think this is because the task processor can process the tasks in parallel, so in this case it can finalize 500 tasks in parallel whereas the task monitor could only finalize one at a time.

Interesting considering that we also have reports of resume taking too long. We'll have to keep investigating on that one.

pditommaso:
Note there are some tests failing

pditommaso:
Wonder if it would not be better using virtual threads instead of introducing yet another thread pool

pditommaso:
Wonder if you considered to make async only [collectOutFiles](https://github.com/nextflow-io/nextflow/blob/ed96e3892ead22a4cec92a86cb76bb96ebf225d6/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy#L1530-L1530)

bentsherman:
I think it's simpler to keep the entire task finalization as one synchronous operation and let the task monitor handle the rest, since it is already doing all of the async stuff

pditommaso:
Agree. Considering this was tested little i've added a feature toggle variable `NXF_ENABLE_ASYNC_FINALIZER` (default false). Once we are sure there are no side effects, it will promoted as default

bentsherman:
Sounds good

This XFileSystemProviderTest just started breaking on the job arrays PR too. Wonder if there is a networking issue with github

PR #4892, opened 2024-04-06
Do not print a new line when stdout is empty

jordeu:
## Description

When using `process.debug = true`, even if any process generates `stdout`, a new line is printed to `stdout` for each executed task.

I was expecting that if the process does not generate any `stdout`, then Nextflow does not print anything to `stdout`.

## Example
Screenshot from 2024-04-06 06-48-51

## With the fix in this pull request
Screenshot from 2024-04-06 07-06-27

**NOTE:** If ANSI is disabled, this will not happen.


pditommaso:
Nice, thanks

PR #4901, opened 2024-04-10
Fix local paths for TES Azure

vsmalladi:


plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzPath.groovy
```diff
@@ -319,14 +320,16 @@ class AzPath implements Path {
 
 
     String toUriString() {
+        final accountName = AzConfig.getConfig().storage().accountName
         if( path.isAbsolute() ) {
-            return "${AzFileSystemProvider.SCHEME}:/${path.toString()}"
+            return "/${accountName}${path.toString()}"
```
bentsherman, line 325: Should there not still be the `az://` prefix here? Or you're saying that Azure will mount the files at `/<accountName>/...` ?

vsmalladi: Azure-Tes will mount these files

plugins/nf-ga4gh/src/main/nextflow/ga4gh/tes/executor/TesExecutor.groovy
```diff
@@ -158,6 +160,14 @@ class TesExecutor extends Executor implements ExtensionPoint {
...
         new TesTaskHandler(task, this)
     }
+
+    protected BashWrapperBuilder createBashWrapperBuilder(TaskRun task) {
+        // creates the wrapper script
+        final builder = new BashWrapperBuilder(task)
+        // job directives headers
+        builder.headerScript = "NXF_CHDIR=${Escape.path(task.workDir)}"
+        return builder
+    }
```
bentsherman, line 170: This should be done here instead: https://github.com/nextflow-io/nextflow/blob/048c3955236cb7ff1233323c55554c0b4e966bb0/plugins/nf-ga4gh/src/main/nextflow/ga4gh/tes/executor/TesTaskHandler.groovy#L170-L172

TesBashBuilder is a subclass of BashWrapperBuilder which is specialized for TES, so something like this should do:
```groovy
    protected TesBashBuilder newTesBashBuilder(TaskRun task, String remoteBinDir) {
        final builder = new TesBashBuilder(task, remoteBinDir)
        builder.headerScript = "NXF_CHDIR=${Escape.path(task.workDir)}"
        return builder
    }
```

plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzPath.groovy
pditommaso: This will break everything

plugins/nf-ga4gh/src/main/nextflow/ga4gh/tes/executor/TesExecutor.groovy
vsmalladi: Good call will fix

plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzPath.groovy
vsmalladi: What do you mean @pditommaso ?

vsmalladi: Do you mean the Azure plugin will break is that it?

pditommaso: `toUriString` should return path with the scheme prefix, using just `toString` if you don't need that

vsmalladi: okay let me see if I can refactor accordingly

bentsherman: @pditommaso this is a temporary fix for testing TES Azure, the ultimate goal is to include the account name in the AZ path only when appropriate

however now that I understand it's really the local path, likely we should implement a custom file copy strategy for it:
- `AzFileCopyStrategy`: current strategy which uses `azcopy` to transfer files
- `TesAzFileCopyStrategy`: custom strategy for TES Azure which simply the files mounted at `/<account-name>`

bentsherman: The main problem is that TES Azure expects AZ paths in a different format from the Azure Batch executor

vsmalladi: so whats the recomendation?

vsmalladi: Agreed and we have to find a switch if its TES vs not

plugins/nf-ga4gh/src/main/nextflow/ga4gh/tes/client/ApiClient.java
```diff
@@ -68,7 +68,7 @@
 
 public class ApiClient {
 
-    private String basePath = "http://localhost";
+    private String basePath = "/ga4gh/tes/v1";
```
MattMcL4475, line 71: is this correct?  typically TES has a url like: `https://tes.azurewebsites.net/v1/tasks`

vsmalladi: Correct. This does get over written by the config. I changes from default of localhost to `/ga4gh/tes/v1` to share what the spec has an example url. I can change it back it has no implication on the code.

vsmalladi:
@bentsherman mostly working. Just have to work on the last part of output file checks

bentsherman:
Regarding the output file type, when you are declaring the output files:
```groovy
        task.getOutputFilesNames().each { fileName ->
            body.addOutputsItem(outItem(fileName))
        }
```

Instead you can do something like this:
```groovy
        task.getOutputsByType(FileOutParam).keySet().each { param ->
            final type = param.type
            final patterns = param.getFilePatterns(task.context, task.workDir)
            patterns.each { pattern -> body.addOutputsItem(outItem(pattern, type)) }
        }
```

No big deal if you can't get it to work, I can just add it in the main PR

plugins/nf-ga4gh/src/main/nextflow/ga4gh/tes/executor/TesTaskHandler.groovy
```diff
@@ -168,17 +176,31 @@ class TesTaskHandler extends TaskHandler {
...
+        String tes_end_point = executor.getEndpoint()
+        if ( tes_end_point.contains('azure.com') ) {
+            String azure_work = WORK_DIR.replace("/", "")
+            final result = new ArrayList(BashWrapperBuilder.BASH)
+            String bashAsString = result.join(' ')
+            exec.command = List.of('/bin/bash', '-c', "${bashAsString} ${azure_work}/${wrapperFile.getName()} &> ${azure_work}/${TaskRun.CMD_LOG}".toString())
+
+        }
+        else 
+            exec.command = List.of('/bin/bash', '-c', "${BashWrapperBuilder.BASH} ${wrapperFile.getName()} &> ${TaskRun.CMD_LOG}".toString())
```
bentsherman, line 198: @vsmalladi why is this change needed?

vsmalladi: for our TES it was having issues with the nested list

`/bin/bash -c [/bin/bash -ue] .command.run &> .command.log`

So this change was to bring it back to old way of doing it without changing our Executor too much.

`/bin/bash -c /bin/bash -ue .command.run &> .command.log`

bentsherman: I'm talking about specifically this line:
```diff
+            String azure_work = WORK_DIR.replace("/", "")
```

bentsherman: Does the container not start in `/work` ?

vsmalladi: ya on azure we don't start in the work directory

bentsherman: where do you start?

vsmalladi: Our tesRunner, runs at root.

So then i do `NXF_CHDIR=/work` in `.command.run`

bentsherman: but the TES task submitted by Nextflow speciffies the work dir as `/work`:
```groovy
        exec.workdir = WORK_DIR
```

so why aren't you respecting that?

vsmalladi: It has to do how our runner runs in the docker container. I do use it as 

`builder.headerScript = "NXF_CHDIR=${Escape.path(WORK_DIR)}"`

I tried using /work/.command.sh but seems to be an issue what we have access to

bentsherman: So you can do `bash work/.command.run` but not `bash /work/.command.run` ?

vsmalladi: Let me look at this again.

vsmalladi: > So you can do `bash work/.command.run` but not `bash /work/.command.run` ?

Yep its somehow the runner outside the docker container can access the outfiles. Its super odd. But works.

bentsherman: okay I did some cleanup, let's merge this into the main PR and discuss as a whole

PR #4923, opened 2024-04-16
Use full (String) form of S3 uris in error message

robsyme:
Before this fix, error messages that referred to S3 paths would print potentially-confusing messages that omit the `s3:/` prefix:

> `reason: the path: /my-bucket/path/to/file does not exist `

With this fix, the message prints the full URI:

> `reason: the path: s3://my-bucket/path/to/file does not exist `


bentsherman (approved):

pditommaso (approved):

pditommaso:
Nice, thanks

PR #4941, opened 2024-04-22
[ci skip] docs: azure batch autopool feature more comprehensive documentation

### LINKED ISSUE #4304: Extend Azure autopool feature to create multiple pools based on process requirements
## New feature

Currently, Azure runs the entire process within a pool of a single VM type. [Recent changes ](https://github.com/nextflow-io/nextflow/pull/4178)mean Nextflow will create multiple pools of the same size based on resource labels, we should extend this feature to make a new pool based on process requirements [similar to GCP Batch](https://www.nextflow.io/docs/latest/google.html#process-definition). 

This will help:
 - Resource management and using wastefully large machines (i.e. the size of the largest Azure job).
 - We can avoid some [strangeness in the Azure Autoscale formula ](https://github.com/nextflow-io/nextflow/discussions/3731) by just manually controlling pool size with Azure Batch API calls (pool size can be adjusted after creation).
 - A common complaint with Azure is disks run out of space. There isn't any way of controlling this, but by supporting the [disk process directive](https://www.nextflow.io/docs/latest/process.html#disk) we can select an appropriate sized machine. 

It would require a way of finding appropriate VM sizes from Azure. This might be possible via API with a default of fallbacks. 

It would also require the user to have appropriate quota, which is largely on them but it would be good to make it user friendly within Nextflow.

## Usage scenario 

Bringing feature parity with AWS and GCP Batch.

## Suggest implementation 

- Add support for the `machineType` process directive. I imagine typically a user would select from a single family (e.g. `Standard_E*d_v5`)
- Nextflow would select from a list of available machines based on the machine requirements.
- The config could look something like this:
```
process.queue = 'auto'
process.machineType = 'Standard_E*d_v5'
```
As far as I can see, the Azure config wouldn't change much? It could default to process.queue = 'auto' unless otherwise stated.

adamrtalbot:
Adds details about the Azure Batch autopool feature including how to use variable sized pools and nodes. Should make it more obvious to end users that Nextflow supports splitting work across variable sized machines.

Closes #4304 (kinda, it documents a feature that already existed).

### DIFF
--- a/docs/azure.md
+++ b/docs/azure.md
@@ -157,11 +157,36 @@ See the [Batch documentation](https://docs.microsoft.com/en-us/azure/batch/quick
 
 ### Pools configuration
 
-When using the `autoPoolMode` option, Nextflow automatically creates a `pool` of compute nodes to execute the jobs in your pipeline. By default, it only uses one compute node of the type `Standard_D4_v3`.
+When using the `autoPoolMode` option, Nextflow automatically creates a `pool` of compute nodes appropriate for your pipeline.
+
+By default, the `cpus` and `memory` directives are used to find the smallest machine type that fits the requested resources in the Azure VM type, specified by `machineType`. If `memory` is not specified, 1 GB of memory is allocated per CPU. By default, it only uses one compute node of the type `Standard_D4_v3`.
+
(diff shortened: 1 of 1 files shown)


pditommaso (approved):

vsmalladi:
@adamrtalbot how would these all play out in the nextflow code:

`Standard_e*d_v5`
`Standard_e*_v5`
`Standard_e*_v*`
`Standard_e*`

What is the priority of choosing a VM. Does it limit to the Family and then match the compute requirements?

PR #4959, opened 2024-04-30
Guarantees K8s pod name is unique on resume

pditommaso:
This PR guarantees that a pod name is unique so that when resuming there is going to be no conflict


bentsherman (approved):

modules/nextflow/src/main/groovy/nextflow/k8s/K8sTaskHandler.groovy
```diff
@@ -159,7 +159,8 @@ class K8sTaskHandler extends TaskHandler implements FusionAwareTask {
     }
 
     protected String getSyntheticPodName(TaskRun task) {
-        "nf-${task.hash}"
+        final suffix = System.currentTimeMillis().toString().md5()[-5..-1]
+        return "nf-${task.hash}-$suffix"
```
pditommaso, line 163:
```suggestion
        return "nf-${task.hash}-${suffix}"
```

PR #4964, opened 2024-05-01
Allow multiple cluster options to be specified as a string list

### LINKED ISSUE #4935: clusterOptions does not work with CLI flags (LSF executor)
## Bug report 

when specigying a clusterOptions which includes a flag the Grid executor header is constructed wrongly by nextflow

e.g profile.config :
```
clusterOptions = { "-g /${_queue_group} -app ${params.fast_track ? params.lsf_fast_track_profile : task.process.tokenize(':')[-1].toLowerCase()} -sp ${params.hpc_priority} -P ${params.lsf_project_name} -rn" }
```

results in the following BSUB command (LSF executor)

```
#!/bin/bash
#BSUB -o /home/tim/nextflow/13/e0aeb3d2121ed180dc3272e28de8a0/.command.log
#BSUB -q high
#BSUB -W 24:00
#BSUB -J nf-DELIVERY_INTEGRITY_MOVE_DELIVERY_(HX23697398_LP5221738-DNA_H07)
#BSUB -rn -g
#BSUB /high -app
# NEXTFLOW TASK: DELIVERY_INTEGRITY:MOVE_DELIVERY (HX23697398/LP5221738-DNA_H07)
set -e
set -u
NXF_DEBUG=${NXF_DEBUG:=0}; [[ $NXF_DEBUG > 1 ]] && set -x
NXF_ENTRY=${1:-nxf_main}

```

I would expect flags (keys with no values) to be correctly rendered in the BSUB option lists
```
#!/bin/bash
#BSUB -o /home/tim/nextflow/13/e0aeb3d2121ed180dc3272e28de8a0/.command.log
#BSUB -q high
#BSUB -W 24:00
#BSUB -J nf-DELIVERY_INTEGRITY_MOVE_DELIVERY_(HX23697398_LP5221738-DNA_H07)
#BSUB -rn 
#BSUB -g /high
#BSUB  -app something
# NEXTFLOW TASK: DELIVERY_INTEGRITY:MOVE_DELIVERY (HX23697398/LP5221738-DNA_H07)
set -e
set -u
NXF_DEBUG=${NXF_DEBUG:=0}; [[ $NXF_DEBUG > 1 ]] && set -x
NXF_ENTRY=${1:-nxf_main}
```

(Provide a test case that reproduce the problem either with a self-contained script or GitHub repository)

### Program output 

(Copy and paste here output produced by the failing execution. Please highlight it as a code block. Whenever possible upload the `.nextflow.log` file.)

### Environment 

* Nextflow version: [?] 
* Java version: [?]

### Additional context

bentsherman:
Close #4935 

- Allow `clusterOptions` to be a string or string list
- Each list element is written to a separate line when required by the executor, otherwise the list is combined to a single string
- Log warning when a grid executor which requires one option per line (i.e. LSF, OAR) receives multiple options in a string, to discourage users from this approach


modules/nextflow/src/main/groovy/nextflow/executor/OarExecutor.groovy
```diff
@@ -81,10 +81,17 @@ class OarExecutor extends AbstractGridExecutor {
...
-          }
+        final opts = task.config.getClusterOptions()
+        if( opts instanceof Collection ) {
+            for( String line : opts )
+                result << line << ''
+        }
+        else if( opts != null ) {
+            final lines = opts.toString().tokenize(';')
+            if( lines.size() > 1 )
+                log.warn1 "[OAR] Specifing multiple cluster options in a string is deprecated, use a string list instead: clusterOptions = [ ${lines.collect(v -> "'$v'").join(', ')} ]"
```
pditommaso, line 92:
```suggestion
                log.warn1 "[OAR] Specifyng multiple cluster options in a string is deprecated, use a string list instead: clusterOptions = [ ${lines.collect(v -> "'$v'").join(', ')} ]"
```

docs/executor.md
```diff
@@ -407,12 +407,19 @@ Resource requests and other job characteristics can be controlled via the follow
 - {ref}`process-queue`
 - {ref}`process-time`
 
-### Known Limitations
+:::{versionchanged} 24.04.0
+Prior to this version, if `clusterOptions` was specified as a string, multiple options were separated by semicolons to ensure that the job script was formatted correctly:
+```groovy
+clusterOptions = '-t besteffort;--project myproject'
+```
```
pditommaso, line 414: It would be preferable to keep the current behavior and extend it adding the support for list of options

bentsherman: It is still supported, wasn't sure if we should try to phase out this syntax or support both in perpetuity

pditommaso: I think we can keep. We have many settings can be both string and list

bentsherman: Okay docs updated

modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
```diff
@@ -429,24 +436,6 @@ class TaskConfig extends LazyMap implements Cloneable {
...
-        }
-
-        if( opts instanceof Collection ) {
-            return new ArrayList<String>(opts)
-        }
-        else {
-            return CmdLineHelper.splitter( opts.toString() )
-        }
-    }
-
```
pditommaso, line 449: It turns out list was already supported

bentsherman: yes, but not in the way we wanted. you had to specify each option/value separately and you'd have to include a blank string for options with no value, which nobody knew about and is what led to this issue. considering it wasn't documented, I think it's fine to change.

I ended up removing this method and adding `getClusterOptionsAsString()` because most executors just add the cluster options as a single line, whereas only a few executors use this list form and they all do it differently. This method was used only by LSF for instance

pditommaso:
Closing in favour of https://github.com/nextflow-io/nextflow/pull/4993. Main changes: 

* keep common logic in the `AbstractGridExecutor` 
* minimise as less as possible the changes in respect current implementation to prevent unexpected side effects

PR #4968, opened 2024-05-03
Update stale documentation in overview.md

greenberga:
#4812 updated the syntax of the workflow in the example script to no longer use the `|` operator, but the accompanying description didn't change. This commit edits the description to bring it in line with the change in the workflow syntax.


docs/overview.md
```diff
@@ -59,7 +59,7 @@ workflow {
 }
 ```
 
-The above example defines two processes. Their execution order is not determined by the fact that the `blastSearch` process comes before `extractTopHits` in the script (it could also be written the other way around). Instead, the pipe operator (`|`) in the workflow between `blastSearch` and `extractTopHits` forwards the outputs from one process to the inputs of the following one.
+The above example defines two processes. Their execution order is not determined by the fact that the `blastSearch` process comes before `extractTopHits` in the script (it could also be written the other way around). Instead, execution order is determined by their sequence in the _workflow_. In that block, `blastSearch` comes first, so it will execute first. `extractTopHits` executes after, and uses the output from `blastSearch`.
```
bentsherman, line 62:
```suggestion
The above example defines two processes. Their execution order is not determined by the fact that the `blastSearch` process comes before `extractTopHits` in the script (it could also be written the other way around). Instead, execution order is determined by their _dependencies_ -- `extractTopHits` depends on the output of `blastSearch`, so `blastSearch` will be executed first, and then `extractTopHits`.
```

bentsherman (changes requested):
This whole thing is gonna be rewritten but the correction makes sense. Only clarify to refer to dependencies instead of order within the workflow block

greenberga:
Thank you for clarifying!

pditommaso (approved):
Nice. Thanks

PR #4972, opened 2024-05-04
Fix Invalid DER: object is not integer when using EC private key

jagedn:
Trying to run pipelines using the k8s operator against a k3s cluster it seems the EC private key is not well supported and following exception is raised:

`Caused by: java.io.IOException: Invalid DER: object is not integer`

Fabric8io suffered same error and they fixed it using bouncy-castle library:

https://github.com/fabric8io/kubernetes-client/pull/1349/files#diff-6c5691ee4cb5995ee4820377a4703695a3cdbc4ea4711022199ea0388ad8370aR124

This PR use same approach and it's tested against a k3d cluster


modules/nextflow/src/test/groovy/nextflow/k8s/client/ConfigDiscoveryTest.groovy
```diff
@@ -415,8 +415,8 @@ class ConfigDiscoveryTest extends Specification {
...
+        final KEY = 'LS0tLS1CRUdJTiBFQyBQUklWQVRFIEtFWS0tLS0tCk1IY0NBUUVFSUNvQTNvRHkzN3NXdmszM3JGRGtRdlZ1Wkh1cCt1Uk40V3RqbUlPR1c4cHBvQW9HQ0NxR1NNNDkKQXdFSG9VUURRZ0FFL3RQME82aEZoNy9SNUV0RndaWjJGdks4S21lcktVaEhhSXZXMkxTeGRTMGxoL241VEVmMQpvRGJ1eHhwRnFhT2Z1QXpXVDdveUpqMnViTHlzV0NGbWFnPT0KLS0tLS1FTkQgRUMgUFJJVkFURSBLRVktLS0tLQo='.decodeBase64()
```
pditommaso, line 419: In principle existing tests should not be affected or it mean the current support it's broken?

bentsherman: @jagedn would be better to keep this test the same and add a new test for the EC algorithm

pditommaso: @bentsherman the test is named `should create key managers from an EC-encrypted client key` therefore my question above

jagedn: Sorry for the late response

I changed the cert/keys because I thought was a mistake and they were not EC so I changed with someone generated by my local k3s

digging a little more I founded the Parser can return 2 kinds of object so this PR try to cover these 2 cases

pditommaso: Excellent

pditommaso:
Merged manually via 4dbb5193. Thanks Jorge

PR #4973, opened 2024-05-05
Job array refactor 

pditommaso:
This PR refactor the support for job array provided by #3892. Main changes 

* Remove the need for `getWorkDir` and `getLaunchCommand` in `TaskHandler` 
* Add similar methods to TaskArrayExecutor providing default implementation 
* Fix support for Fusion 
* Add tests

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/TaskArrayExecutor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/TaskArrayExecutor.groovy
@@ -61,6 +61,12 @@ interface TaskArrayExecutor {
         getWorkDir().fileSystem== FileSystems.default
     }
 
+    /**
+     * Get a {@link TaskHandler} work directory for the task array resolution
+     *
+     * @param handler
+     * @return
+     */
     default String getArrayWorkDir(TaskHandler handler) {
         return isFusionEnabled()
             ? FusionHelper.toContainerMount(handler.task.workDir).toString()
--- a/modules/nextflow/src/main/groovy/nextflow/executor/TaskArrayExecutor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/TaskArrayExecutor.groovy
@@ -16,16 +16,19 @@
 
 package nextflow.executor
 
+import java.nio.file.FileSystems
 import java.nio.file.Path
 
+import groovy.transform.CompileStatic
+import nextflow.fusion.FusionHelper
 import nextflow.processor.TaskHandler
 import nextflow.processor.TaskRun
-
 /**
  * Interface for executors that support job arrays.
  *
  * @author Ben Sherman <bentshermann@gmail.com>
  */
+@CompileStatic
 interface TaskArrayExecutor {
 
     String getName()
@@ -54,4 +57,25 @@ interface TaskArrayExecutor {
      */
     String getArrayTaskId(String jobId, int index)
 
+    default boolean isWorkDirDefaultFS() {
+        getWorkDir().fileSystem== FileSystems.default
+    }
+
+    default String getArrayWorkDir(TaskHandler handler) {
+        return isFusionEnabled()
+            ? FusionHelper.toContainerMount(handler.task.workDir).toString()
+            : handler.task.workDir.toUriString()
+    }
+
+    default String getArrayLaunchCommand(String taskDir) {
+        if( isFusionEnabled() ) {
+            return "bash ${taskDir}/${TaskRun.CMD_RUN}"
+        }
+        else if( isWorkDirDefaultFS() ) {
+            return "bash ${taskDir}/${TaskRun.CMD_RUN} 2>&1 > ${taskDir}/${TaskRun.CMD_LOG}"
+        }
+        else {
+            throw new IllegalStateException("Executor ${getName()} does not support array jobs")
+        }
+    }
 }
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskArrayCollector.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskArrayCollector.groovy
@@ -25,11 +25,8 @@ import groovy.util.logging.Slf4j
 import nextflow.executor.Executor
 import nextflow.executor.TaskArrayExecutor
 import nextflow.file.FileHelper
-import nextflow.fusion.FusionAwareTask
-import nextflow.fusion.FusionHelper
 import nextflow.util.CacheHelper
 import nextflow.util.Escape
-
 /**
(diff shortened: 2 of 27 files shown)


marcodelapierre:
Successfully tested this refactored PR with Slurm, as well.   
Tested with `hello`, `rnaseq-nf` (tweaked for multiple samples), and `rnaseq` (test profile).

pditommaso:
Thanks Marco!

modules/nextflow/src/main/groovy/nextflow/processor/TaskArrayCollector.groovy
```diff
@@ -148,8 +147,9 @@ class TaskArrayCollector {
         Files.createDirectories(workDir)
 
         // create wrapper script
-        final script = createTaskArrayScript(handlers)
-
+        final script = createArrayTaskScript(handlers)
+        log.debug "Creating task array run >> $workDir\n$script"
```
bentsherman, line 151:
```suggestion
        log.trace "Creating task array run >> $workDir\n$script"
```

modules/nextflow/src/main/groovy/nextflow/executor/TaskArrayExecutor.groovy
```diff
@@ -54,4 +57,31 @@ interface TaskArrayExecutor {
...
+            ? FusionHelper.toContainerMount(handler.task.workDir).toString()
+            : handler.task.workDir.toUriString()
+    }
+
+    default String getArrayLaunchCommand(String taskDir) {
+        if( isFusionEnabled() ) {
+            return "bash ${taskDir}/${TaskRun.CMD_RUN}"
+        }
+        else if( isWorkDirDefaultFS() ) {
+            return "bash ${taskDir}/${TaskRun.CMD_RUN} 2>&1 > ${taskDir}/${TaskRun.CMD_LOG}"
```
bentsherman, line 81:
```suggestion
            return "bash ${taskDir}/${TaskRun.CMD_RUN} 2>&1 | tee ${taskDir}/${TaskRun.CMD_LOG}"
```

I believe this should use `tee` so that the log output is also forwarded to stdout like regular tasks

pditommaso: But then the parent job will collect the output of child tasks 

https://github.com/nextflow-io/nextflow/blob/037a67d9e5cedadd5302107cb76c1f55f0f940d7/modules/nextflow/src/main/groovy/nextflow/executor/LsfExecutor.groovy#L72-L72

modules/nextflow/src/main/groovy/nextflow/processor/TaskArrayCollector.groovy
bentsherman: Fair enough, eventually we should demote it to trace or truncate it

modules/nextflow/src/main/groovy/nextflow/executor/TaskArrayExecutor.groovy
bentsherman: Actually I guess we should disable this option for the job array, it makes no sense to send anything to the log file in the parent directory

bentsherman: Then we'll still want to use `tee` so that the task log is also the job stdout. Some schedulers like PBS Pro can display the stdout of a job while it's running, but you need to either print to stdout or set the `-o` option. We can't do the latter because it's a job array but we can do the former, each child job will print its log to its own stdout

pditommaso: I'm worried funneling the stdout of all tasks to the same log. Read this 

> Note: if you only use '%A' in the log all array tasks will try to write to a single file. The performance of the run will approach zero asymptotically. Make sure to use both %A and %a in the log file name specification.

https://help.rc.ufl.edu/doc/Submitting_Array_Jobs

modules/nextflow/src/main/groovy/nextflow/executor/TaskArrayExecutor.groovy
```diff
@@ -54,4 +57,31 @@ interface TaskArrayExecutor {
...
+     * @param handler
+     * @return
+     */
+    default String getArrayWorkDir(TaskHandler handler) {
+        return isFusionEnabled()
+            ? FusionHelper.toContainerMount(handler.task.workDir).toString()
+            : handler.task.workDir.toUriString()
+    }
+
+    default String getArrayLaunchCommand(String taskDir) {
```
bentsherman, line 76: While I understand the motivation to move this logic into the job array interface, I fear we will have to revert it once we implement task batching: https://github.com/nextflow-io/nextflow/pull/3909/files#diff-c317d610e9c87f9c425afaedfb79ed6ca92083197d0e0ac9be6b2299aab5860eR110-R123

I implemented `getWorkDir()` and `getLaunchCommand()` in the task handler specifically so that it can be used both for job arrays and task batching

modules/nextflow/src/main/groovy/nextflow/executor/TaskArrayExecutor.groovy
bentsherman: But the `taskDir` in this method refers to the child task dir, so each child job will send its stdout to its own log and also to its own stdout

Also, the note you mentioned won't happen anyway if we disable the `-o` option when submitting a job array

bentsherman: Well, I'm not 100% certain that all grid executors handle the job array stdout in this way. Might be better to play it safe for now, only sending the child job stdout to its log. But either way we should remove the `-o` option for job arrays, it can only cause trouble

pditommaso: If we disabled the `-o` option in the parent job then `tee` should not be used otherwise it won't be consumed by anything.

modules/nextflow/src/main/groovy/nextflow/executor/TaskArrayExecutor.groovy
pditommaso: let's see this later

modules/nextflow/src/main/groovy/nextflow/executor/TaskArrayExecutor.groovy
pditommaso: > But either way we should remove the -o option for job arrays, it can only cause trouble

Make sense to remove it. Let's do in the parent PR

PR #4974, opened 2024-05-06
Improve error message for image pull time-out for Singularity/Apptainer/Charliecloud

### LINKED ISSUE #4823: singularity.pullTimeout lack of error message
Hi Team,

My team has recently discovered (after a long time digging around) that Nextflow will SIGTERM `singularity/apptainer pull`s after 20mins. We've since fixed it by increasing the `singularity.pullTimeout` config. In our case, the long time taken to convert Docker images to Singularity is due to the high-latency shared filesystem we use on our cluster.

Firstly, could I get some clarification on why this is necessary? I notice that Docker doesn't have a similar config.

Secondly, is it possible to add a meaningful error message when the timeout is reached? The current error message looks like:

```
Caused by:
 Failed to pull singularity image
 command: singularity pull  --name ghcr.io-wehi-researchcomputing-haemosphere-latest.img.pulling.1710721200833 docker://ghcr.io/wehi-researchcomputing/haemosphere:latest > /dev/null
 status : 143
 message:
   INFO:    Converting OCI blobs to SIF format
   INFO:    Starting build...
   Getting image source signatures
   Copying blob sha256:9d19ee268e0d7bcf6716e6658ee1b0384a71d6f2f9aa1ae2085610cf7c7b316f

 ... skipped output ...

   2024/03/18 11:31:02  info unpack layer: sha256:e5a50e1173436b4242b02cff1322507cbd5cce18c675ea94dc11ccb6bb8e3ef0
```

Which, besides saying the pull has failed with error code 143, doesn't provide much more meaningful information. Adding more meaningful information could save others lots of time debugging. E.g. saying something to the tune of `pullTimeout of "20 mins" reached. singularity/apptainer pull terminated. Increase timeout with singularity/apptainer.pullTimeout.`.

marcodelapierre:
Closes #4823  by adding an explicit suggestion in the error message, to increase the `pullTimeout` config for the relevant container runtime.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy
@@ -240,7 +240,7 @@ class CharliecloudCache {
         def status = proc.exitValue()
         if( status != 0 ) {
             consumer.join()
-            def msg = "Charliecloud failed to pull image\n  command: $cmd\n  status : $status\n  hint   : Try and increase charliecloud.pullTimeout in the config (default is $DEFAULT_PULLTIMEOUT_DURATION)\n  message:\n"
+            def msg = "Charliecloud failed to pull image\n  command: $cmd\n  status : $status\n  hint   : Try and increase charliecloud.pullTimeout in the config (default is "${DEFAULT_PULLTIMEOUT_DURATION}")\n  message:\n"
             msg += err.toString().trim().indent('    ')
             throw new IllegalStateException(msg)
         }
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy
```diff
@@ -47,7 +49,7 @@ class CharliecloudCache {
 
     private boolean missingCacheDir
 
-    private Duration pullTimeout = Duration.of('20min')
+    private Duration pullTimeout
```
pditommaso, line 52:
```suggestion
    private Duration pullTimeout = DEFAULT_PULLTIMEOUT_DURATION
```

pditommaso: Awesome

Reziya: > See last commit, I have actually restored the original code here without defining the extra `DEFAULT_PULLTIMEOUT_DURATION`, which is not really needed. In the very first pass I was using the default in the error message, but then I realised the current value would be more informative for the user.

I have same issue ,how did you solved this problem?

Issue #4921, opened 2024-04-16
Slurm: Add config scope for `-A` (account / project)


ewels:
It would be great to have a config attribute specifically for the slurm account (aka project) to submit jobs to.

See the [Slurm docs](https://slurm.schedmd.com/sbatch.html):
> `-A`, `--account=<account>`
> Charge resources used by this job to specified account. The _account_ is an arbitrary string. The account name may be changed after job submission using the`scontrol` command.

We have [44 different institutional configs](https://github.com/search?q=repo%3Anf-core%2Fconfigs%20-A&type=code) in nf-core doing this by using `clusterOptions` plus a parameter currently. For [example](https://github.com/nf-core/configs/blob/59ac7ecea321f8f4cc6d4cea66985811e29b6259/conf/hasta.config#L23-L26):

```nextflow
process {
    executor = 'slurm'
    clusterOptions = { "-A $params.priority ${params.clusterOptions ?: ''}" }
}
```

It'd be great to remove the need for this workaround as it is so common.

pditommaso:
Something like ?

```
slurm {
  account = 'something'
}
```

PR #4975, opened 2024-05-06
Add account config option for grid executors

pditommaso:
This PR adds the ability to define a Slurm account in the nextflow config. For example: 

```
executor {
  account = 'some-project-xyz'
}
```

Solve https://github.com/nextflow-io/nextflow/issues/4921

Updated to use `executor` instead of `slurm` on 7th May

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -646,6 +646,12 @@ The `executor` scope controls various executor behaviors.
 
 The following settings are available:
 
+`executor.account`
+: :::{versionadded} 24.04.0
+  :::
+: *Used only by the {ref}`slurm-executor`, {ref}`lsf-executor`, {ref}`pbs-executor` and {ref}`pbspro-executor` executor.*
+: Allows specifying the project or organisation account that should be charged for running the pipeline jobs.
+
 `executor.cpus`
 : The maximum number of CPUs made available by the underlying system. Used only by the `local` executor.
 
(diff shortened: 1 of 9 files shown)


bentsherman:
In this case I recommend also moving `executor.perCpuMemAllocation` to `slurm.perCpuMemAllocation`.

marcodelapierre:
I would suggest exposing this option as a generic executor one, `executor.account`, rather than creating the new `slurm` config scope.   
Indeed, other executors implement it, e.g. PBS has `-P <account/project>`.

muffato:
Hi. I was going to suggest the same thing as @marcodelapierre . LSF has the `-G user_group`: https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=o-g-1

pditommaso:
Fair enough

pditommaso:
@bentsherman @marcodelapierre please have a look and let's merge it

bentsherman (approved):

marcodelapierre (approved):
Good to go for me!
(committed minor edits)

pditommaso:
Thanks both. I've reverted [9486d68](https://github.com/nextflow-io/nextflow/pull/4975/commits/9486d68d45edcbea944161cad484bea3a3e433f7) because it's better to keep explicit from the name is coming from

Issue #4953, opened 2024-04-26
ifEmpty operator doesn't return expected value (value channel)


onuryukselen:
### Expected behavior and actual behavior

I anticipate receiving the file() block when the channel is empty. However, it looks like it returns queue channel and it gets consumed after one job submission. Are there any other way to check whether the channel is empty?

### Steps to reproduce the problem

Here is the example code to reproduce the problem:
```
ch = channel.from( 1, 3, 5, 7 )
ch_empty_file = file("$baseDir/NO_FILE")
secondChannel = channel.empty()
testChn= secondChannel.ifEmpty(ch_empty_file)

process test {

input:
   val(a)
   file(b)

   script:
   """
   """
}

process test2 {

input:
  val(a)
  file(b)

  script:
  """
  """
}

workflow {
test (ch,ch_empty_file)
test2(ch,testChn)
}
```

### Program output 

N E X T F L O W  ~  version 23.10.1
Launching `/opt/run1372/test.nf` [exotic_swartz] DSL2 - revision: 663225ff37
[1b/66845b] Submitted process > test (1)
[8c/7fb0e1] Submitted process > test (2)
[41/eda731] Submitted process > test2 (1)
[f5/470f5e] Submitted process > test (3)
[ac/f0ad1a] Submitted process > test (4)

It executes test2 only once. I was expecting 4 test2 jobs similar to test process.

### Environment 

* Nextflow version: [23.10.1] 
* Java version: [openjdk 18.0.2-ea 2022-07-19]
* Operating system: [Linux]
* Bash version: (5.1.16)

onuryukselen:
Hello, have you been able to replicate the issue? Do you think my expectation about the ifEmpty operator is correct?

bentsherman:
It turns out the docs are incorrect. They say that `ifEmpty` returns a value channel but actually it always returns a queue channel. I will review the operator return types more carefully and update the docs

onuryukselen:
I see. Could you please suggest another way to check whether the channel is empty or not? I want to return Value Channel if channel is empty.

onuryukselen:
My goal is creating optional input for a process. For example if secondChannel is empty, I want to use file("$baseDir/NO_FILE") so that process will work even secondChannel is empty.

Similar to https://nextflow-io.github.io/patterns/optional-input/ but I need to check channel. So is there any way to write condition like this:
``` 
testChn= secondChannel.ifEmpty("empty")
if (testChn.value == "empty"){
[...]

PR #4976, opened 2024-05-07
Update operator return types

bentsherman:
Update operator docs to distinguish between operators that always return a queue channel and operators that may return a queue or value channel matching the source channel type.


bentsherman:
See also #4953

pditommaso (approved):

PR #5003, opened 2024-05-16
Remove `seqera` and `defaults` from Conda default channels

pditommaso:
This PR removes `seqera` from the Conda default channels as it's not going to be maintained anymore


pditommaso:
This also removes `defaults` from Conda default channels

Issue #4967, opened 2024-05-02
Publishing a symlink when using fusion copies link (mock symlink) rather than link target


robsyme:
## Bug report 

Given a simple workflow that creates a file 

```groovy
workflow {
    Channel.value("Seqera")
    | MakeJson
    | PublishIdCard
}

process MakeJson {
    input: val(name)
    output: tuple val(name), path('*.json')
    script: 
    """
    echo '{"name":"$name"}' > id.${name}.json
    """
}

process PublishIdCard {
    publishDir 'results'

    input: tuple val(name), path("jsons/*")
    output: path("symlink.json")
    script: "ln -s jsons/id.${name}.json symlink.json"
}
```
... and configuration that sets fusion:

```groovy
workDir = 's3://my-bucket-name/work'
wave.enabled = true
fusion.enabled = true
docker.enabled = true
process.scratch = false
process.container = 'ubuntu:latest'
```

### Expected behavior and actual behavior

I would expect the contents of the file published at `./results/symlink.json` to include the json document

```
{"name":"Seqera"}
```
...but instead it is a text file containing:

```
jsons/id.Seqera.json
```

### Notes

The .fusion.symlinks document in the task working directory of the PublishIdCard task contains the text `symlink.json`, but Nextflow doesn't seem to follow the symlink when publishing.

bentsherman:
Not sure this is worth supporting. Normally we detect fusion symlinks by checking if the output file is also an input, which is cheap because the list of inputs is in memory. But detecting this would require checking for `.fusion.symlnks` for *every* task, including subdirectories. This is because, though this example is simple, there is no way to know in general if a task will link inputs to outputs. So now you're doing a bunch of extra S3 calls for _every_ task, even if you never use this pattern.

Instead, I would recommend that you either copy the file or declare the input file as an output directly (e.g. `output: path("jsons/id.${name}.json")`). I understand this might be more verbose compared to e.g. a single output directory containing a bunch of different files. It might be more readable when we have static types #4553 and can output a record type representing a directory of diverse output files.

pditommaso:
it's not even running for me 

```
ERROR ~ Error executing process > 'MakeJson'

Caused by:
  No signature of method: nextflow.script.ScriptBinding.echo '{() is applicable for argument types: () values: [] -- Check script 't.nf' at line: 10

Source block:
  "echo '{"name":"$name"}' > id.${name}.json"

[...]

PR #5004, opened 2024-05-16
Remove Fusion symlink resolution

### LINKED ISSUE #4725: Fusion symlink resolution doesn't work with directories

bentsherman:
Close #4725 #4967 

Fusion 2.3 will resolve output symlinks going forward, so this code can be removed.


pditommaso (approved):

pditommaso:
Requires #5005

tests/checks/fusion-symlink.nf/.checks
pditommaso, line 1: Think we should keep the test to be sure the support is implemented correctly by the Fusion runtime

pditommaso: Reverted

PR #5008, opened 2024-05-19
Add enabled property to output DSL

pditommaso:
This PR adds the `enabled` option to output DSL that allows to turn it on/off specific publish targets.


bentsherman:
Needs to be documented:
```md
`enabled`
: Enable or disable publishing (default: `true`).
```

pditommaso:
you are right, please add it 🙏

PR #5013, opened 2024-05-21
Fix the default for async finalizer

pditommaso:
The PR fixes the setting for the `NXF_ENABLE_ASYNC_FINALIZER` env variable. 

The expression `'some string' as boolean` is always true unless the string is null or empty. 

Therefore the setting is always enabled unless `NXF_ENABLE_ASYNC_FINALIZER` is set to an empty string.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy
@@ -118,7 +118,7 @@ class TaskPollingMonitor implements TaskMonitor {
     @Lazy
     private ExecutorService finalizerPool = { session.taskFinalizerExecutorService() }()
 
-    private boolean enableAsyncFinalizer = SysEnv.get('NXF_ENABLE_ASYNC_FINALIZER','false') as boolean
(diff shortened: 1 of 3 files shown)


pditommaso:
Wonder if it should be set to `true` by default at this point

bentsherman:
Might be better to make it true by default and allow people to disable it if it breaks something for them. We'll get much better testing feedback that way

modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy
```diff
@@ -118,7 +118,7 @@ class TaskPollingMonitor implements TaskMonitor {
     @Lazy
     private ExecutorService finalizerPool = { session.taskFinalizerExecutorService() }()
 
-    private boolean enableAsyncFinalizer = SysEnv.get('NXF_ENABLE_ASYNC_FINALIZER','false') as boolean
+    private boolean enableAsyncFinalizer = SysEnv.getBool('NXF_ENABLE_ASYNC_FINALIZER',false)
```
pditommaso, line 121:
```suggestion
    private boolean enableAsyncFinalizer = SysEnv.getBool('NXF_ENABLE_ASYNC_FINALIZER',true)
```

pditommaso:
ok

PR #5018, opened 2024-05-22
Charliecloud: registry, docs

nschan:
Hi,

I noticed that my previous attempt (https://github.com/nextflow-io/nextflow/pull/4879) at making charliecloud handle the registry does not work. This PR fixes this by following the general approach of other container engines: letting normalizeDockerImageName handle this.
I also realized that my previous PR added three new options for charliecloud, but these were not documented, so this also updates the docs accordingly. This is identical to #5017, except that here I signed off the commits.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy
@@ -86,7 +86,7 @@ class CharliecloudCache {
             if( !registry.endsWith('/') ) {
                 registry += '/'
             }
-            name += registry
+            name = registry + name
         }
 
         name = name.replace(':','+').replace('/','%')
--- a/modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy
@@ -86,7 +86,7 @@ class CharliecloudCache {
             if( !registry.endsWith('/') ) {
                 registry += '/'
             }
-            imageUrl = registry + imageUrl
+            name = registry + name
         }
         name = name.replace(':','+').replace('/','%')
         return name 
--- a/modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy
@@ -86,8 +86,9 @@ class CharliecloudCache {
             if( !registry.endsWith('/') ) {
                 registry += '/'
             }
-            name = registry + name
+            name += registry
         }
+
         name = name.replace(':','+').replace('/','%')
         return name 
     }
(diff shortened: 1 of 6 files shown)


modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy
```diff
@@ -79,11 +80,14 @@ class CharliecloudCache {
...
-        // add registry
-        if( registry )
-            name = registry + name
         
+        // add registry
+        if( registry ) {
+            if( !registry.endsWith('/') ) {
+                registry += '/'
+            }
+            imageUrl = registry + imageUrl
```
pditommaso, line 89: Is this correct? why updating `imageUrl` if it's not used

modules/nextflow/src/main/groovy/nextflow/container/ContainerHandler.groovy
```diff
@@ -271,4 +274,32 @@ class ContainerHandler {
...
-}
+
+    /**
+     * Normalize charliecloud image name resolving the absolute path
+     *
+     * @param imageName The container image name
+     * @return Image name in canonical format
+     */
+     @PackageScope
+     String normalizeCharliecloudImageName(String img) {
```
pditommaso, line 285: Please add a unit test

modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy
nschan: Good catch, it is incorrect, should be `name`. I will update.

modules/nextflow/src/main/groovy/nextflow/container/ContainerHandler.groovy
nschan: Added a unit test in 75152b8

modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy
```diff
@@ -79,11 +80,14 @@ class CharliecloudCache {
...
+        
         // add registry
-        if( registry )
+        if( registry ) {
+            if( !registry.endsWith('/') ) {
+                registry += '/'
+            }
             name = registry + name
-        
+        }
```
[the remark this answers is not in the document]
nschan: OK, see 8e8c84a

modules/nextflow/src/test/groovy/nextflow/container/ContainerHandlerTest.groovy
```diff
@@ -200,6 +200,34 @@ class ContainerHandlerTest extends Specification {
         result == 'shifter://image'
             
     }
+    @Unroll
```
pditommaso, line 203:
```suggestion

    @Unroll
```

pditommaso (approved):

PR #4840, opened 2024-03-22
Improve secrets rendering

pditommaso:
This PR simplifies the rendering of secrets in the the `nextflow config` using a AST transformation instead of using the `SecretContext` class.

See also https://github.com/nextflow-io/nextflow/pull/4177

PR #5020, opened 2024-05-22
Fix the invocation of function with gstring

pditommaso:
This fixes the invocation of a module function having a String type in the signature and  GString value as argument.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/FunctionDef.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/FunctionDef.groovy
@@ -16,7 +16,6 @@
 
 package nextflow.script
 
-import java.lang.reflect.Method
 
 import groovy.transform.CompileStatic
 /**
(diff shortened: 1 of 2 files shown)


bentsherman:
I think we can solve this in a much simpler way: https://github.com/nextflow-io/nextflow/blob/e2e608140cdde1da39df4c911f56286015538228/modules/nextflow/src/main/groovy/nextflow/config/ConfigBuilder.groovy#L835

This logic which was removed by #4840:
```groovy
        else if( config instanceof GString ) {
            final holdSecrets = config.values.any { it instanceof SecretHolder }
            return holdSecrets ? config : config.toString()
        }
```

So let's just add it back without the secrets part:
```groovy
        else if( config instanceof GString ) {
            return config.toString()
        }
```

bentsherman:
Never mind, I just confirmed that it happens regardless of the config setting. Let's go with the more general solution

bentsherman (approved):

PR #5026, opened 2024-05-26
AWS Batch: use custom log group name when requesting task logs, if it is defined

nathanthorpe:
Nextflow allows the `aws.batch.logsGroup` parameter to be defined to override the log group name. The `getTaskLogStream` method is only able to use the default log group `/aws/batch/job`. 

Update this method to allow support for retrieving logs when the custom log group name is defined.

Also update the value of the default log group name.


plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchHelper.groovy
```diff
@@ -199,22 +200,25 @@ class AwsBatchHelper {
...
             return null
         }
+        // Fallback to the default log group name if log group name is not set
+        final logGroupName = containerDetail.logConfiguration?.options?.get('awslogs-group')
+            ?: '/aws/batch/job'
 
         final logRequest = new GetLogEventsRequest()
-                .withLogGroupName("/aws/batch/job")
-                .withLogStreamName(streamId)
+                .withLogGroupName(logGroupName)
```
pditommaso, line 213: If I'm understanding correctly, the log group should match the one define in the config 

https://github.com/nextflow-io/nextflow/blob/master/plugins/nf-amazon/src/main/nextflow/cloud/aws/config/AwsBatchConfig.groovy#L62-L65

nathanthorpe: It should be yes, but I thought it would be easier to pull directly from the container response since we're already grabbing that information.

I can change it to pass in the log group from the config file in the `getTaskLogStream` method?

pditommaso:
Solved via 30195838

PR #5044, opened 2024-06-02
Allow dynamic `cache` directives using closures

### LINKED ISSUE #5022: Request to enable dynamic `cache` directive
## New feature

It would be helpful to be able to set the `cache` directive via a closure.

## Usage scenario 

It would be sometimes helpful to force a re-run of a specific task (in cases where the outputs are corrupted, for example).  For users that don't have access to the run workdir, it would be helpful to set the following configuration:

```groovy
process {
  withName: MyProcess {
    cache = { task.tag != "mytag" }
  }
}
```

At the moment, this closure is not evaluated, and is simply compared directly do the available options, and we get the warning:
```
WARN: Unknown cache mode: Script_a35edbdc3426f6a0$_runScript_closure2$_closure5@30922f8d
```

## Suggest implementation 

Something in `ProcessConfig`, I suppose :D

greenberga:
Closes #5022

### DIFF
--- a/modules/nextflow/src/test/groovy/FunctionalTests.groovy
+++ b/modules/nextflow/src/test/groovy/FunctionalTests.groovy
@@ -750,4 +750,109 @@ class FunctionalTests extends Dsl2Spec {
         and:
         runner.session.fault.report ==~ /(?s).*-- Check script '(.*?)' at line: 10.*/
     }
+
+    def 'should handle dynamic cache directives' () {
+        given:
+        String config = '''
+            process {
+                withName: MyProcess {
+                    cache = { task != null }
+                }
+            }
+            '''
+        String script = '''
+                process MyProcess {
+                    tag 'mytag'
+
+                    script:
+                    'echo hello'
+                }
+
+                workflow { MyProcess() }
+                '''
+
+        when:
+        new MockScriptRunner(new ConfigParser().parse(config))
+                .setScript(script)
+                .execute()
+        def processor = TaskProcessor.currentProcessor()
+        then:
+        processor instanceof TaskProcessor
+        processor.config.cacheable == false
+
+
+        when:
+        config = ''
+
+        and:
+        script = '''
+                process foo {
+                    cache = { false }
+                    script:
+                    'echo hello'
+                }
+
+                workflow { foo() }
+                '''
+        and:
+        new MockScriptRunner(new ConfigParser().parse(config))
+                .setScript(script)
+                .execute()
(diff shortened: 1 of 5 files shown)


greenberga:
@bentsherman I spent some time looking into this issue. There's a good chance I'm missing something, but as far as I can tell, the `TaskConfig` isn't aware of the hash mode. I can pull the `cache` property from `task.config`, but it's still a closure (i.e., it hasn't been evaluated).

Further, isn't the hash mode slightly different than whether or not to cache? The mode itself can be `standard`, `lenient`, `deep`, or `sha256`&mdash;but you can also set a boolean, in which case the mode will be `null`:

https://github.com/nextflow-io/nextflow/blob/4c54db6ac634cc80a820752d8f4edb959fc832e4/modules/nf-commons/src/main/nextflow/util/CacheHelper.java#L45-L50

Assuming there is some step where a script's process directives are merged with the directives from a config's `process` scope directives, it seems like you'd want to evaluate the closure at that point, in the context of the task. Let me know if that doesn't make sense. Thanks!

modules/nextflow/src/test/groovy/FunctionalTests.groovy
```diff
@@ -750,4 +750,109 @@ class FunctionalTests extends Dsl2Spec {
...
+
+    def 'should handle dynamic cache directives' () {
+        given:
+        String config = '''
+            process {
+                withName: MyProcess {
+                    cache = { task.tag != 'mytag' }
+                }
+            }
+            '''
```
greenberga, line 762: This currently doesn't work because `task` is `null` when this closure gets evaluated. If you define a closure like this in a `nextflow.config` file, will it ever work? Isn't it attempting to capture a local variable that never exists in the context of the config file? In other words, `task`s come from scripts, not config files. If closures work the way I understand them to, referencing the `task` variable in `nextflow.config` won't work because tasks don't get defined until the processor executes a script.

bentsherman:
I think you just need to move the `getHashMode()` method from the process config to the task config, something like this:
```groovy
    HashMode getHashMode() {
        HashMode.of(get('cache')) ?: HashMode.DEFAULT()
    }
```

The `get('cache')` will handle the closure evaluation.

greenberga:
Thanks for the point in the right direction. `get('cache')` does indeed seem to evaluate the closure properly. The only thing I can't quite figure out is why it's enough to move `HashMode` retrieval from the `ProcessConfig` to the `TaskConfig`. Whether `get('cache')` returns `true` or `false`, `HashMode` will be `STANDARD`.

PR #5045, opened 2024-06-02
Target java17

pditommaso:
This set Java 17 as target bytecode generation and lang compatibility. 

TLDR; so far mock created by Spock used cglib, which is not maintained anymore and does not support Java 17. Bytebuddy should replace it, but some tests fail. 

 https://spockframework.org/spock/docs/2.0-M4/faq.html


pditommaso:
@bentsherman when you have some chance, it would be nice to fix the failing tests so we can move to Java 17 as target bytecode

bentsherman:
So this will drop support for Java 11?

pditommaso:
Not needed in the short term. Closing for now

modules/nextflow/build.gradle
```diff
@@ -49,10 +49,10 @@ dependencies {
 
     // test configuration
     testFixturesApi ("org.apache.groovy:groovy-test:4.0.21") { exclude group: 'org.apache.groovy' }
-    testFixturesApi ("cglib:cglib-nodep:3.3.0")
-    testFixturesApi ("org.objenesis:objenesis:3.2")
-    testFixturesApi ("org.spockframework:spock-core:2.3-groovy-4.0") { exclude group: 'org.apache.groovy'; exclude group: 'net.bytebuddy' }
-    testFixturesApi ('org.spockframework:spock-junit4:2.3-groovy-4.0') { exclude group: 'org.apache.groovy'; exclude group: 'net.bytebuddy' }
+    testFixturesApi ("org.objenesis:objenesis:3.4")
+    testFixturesApi ("net.bytebuddy:byte-buddy:1.14.17")
```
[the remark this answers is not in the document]
pditommaso: `testFixtures` allow the use of [these classes](https://github.com/nextflow-io/nextflow/tree/2a71206d099f7f0640b39b7b7dee33c832982aa7/modules/nextflow/src/testFixtures) in the other projects for testing

bentsherman:
The factor connecting all of the failing tests is that they are mocking or spying `RepositoryProvider` or a subclass of it. The GithubRepositoryProviderTest is the only test that mocks a subclass, all of the other provider subclasses do not.

- Adding a default constructor didn't fix it (I think these are not needed anyway thanks to objenesis).
- Making RepositoryProvider concrete instead of abstract didn't fix it
- Moving TagInfo and BranchInfo out of RepositoryProvider didn't fix it (read somewhere that bytebuddy has issues mocking inner classes)

modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy
```diff
@@ -242,7 +242,7 @@ abstract class RepositoryProvider {
     }
 
     @Memoized
-    protected <T> List<T> invokeAndResponseWithPaging(String request, Closure<T> parse) {
+    protected List invokeAndResponseWithPaging(String request, Closure parse) {
```
bentsherman, line 245: This method seems to be the issue. You can either remove `@Memoized` or remove the generics, and the relevant tests stop failing. 🤷

bentsherman (approved):

modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy
pditommaso: Good finding, but it's a pity we need to remove the generic type

pditommaso:
@bentsherman do you have a concrete need for Java 17? I guess there are still people running on Java 11

bentsherman:
The main feature I'm using is this pattern matching syntax:
```java
if( node instanceof Expression exp ) {
  // ...
}
```

Whereas with Java 11 I would have to do:
```java
if( node instanceof Expression ) {
  var exp = (Expression) node;
  // ...
}
```

I use this a lot in the parsing and AST code, so it does make the code more concise. But I can go back to the old way if lots of people still rely on Java 11. I figured it should be pretty easy to upgrade to Java 17 these days, I don't know how many people would actually be stuck with 11

pditommaso:
Understand. Are you sure that this requires Java 17 bytecode generation and not Java 17 lang compatibility?

bentsherman:
I did try that:
```groovy
java {
  sourceCompatibility = JavaVersion.VERSION_17
  targetCompatibility = JavaVersion.VERSION_11
}
```

But I was surprised to find it wasn't allowed:
```
Execution failed for task ':language-server:compileJava'.
> warning: source release 17 requires target release 17
```

bentsherman:
Java 11 seems to be in a similar EOL status as Java 8, at least for Oracle Java: https://endoflife.date/oracle-jdk

So I think it would be safe to require Java >=17. Users always have the previous versions of Nextflow

Maybe it would be helpful to look up the compatibility matrices of other Java projects of similar size to see what they're doing

pditommaso:
I agree this is a nice to have. I'd like just postpone merging this to post 24.10.x to not fragment too much Java requirement for this year stable releases.

bentsherman:
Fine by me. I was actually going to suggest waiting until 25.04 . We can merge the new config parser then too. That will give me more time to stabilize some things

bentsherman:
* or an edge release after 24.10

docs/install.md
```diff
@@ -6,12 +6,16 @@
...
 ## Requirements
 
-Nextflow can be used on any POSIX-compatible system (Linux, macOS, etc), and on Windows through [WSL](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux). It requires Bash 3.2 (or later) and [Java 11 (or later, up to 23)](http://www.oracle.com/technetwork/java/javase/downloads/index.html). You can see which version you have using the following command:
+Nextflow can be used on any POSIX-compatible system (Linux, macOS, etc), and on Windows through [WSL](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux). It requires Bash 3.2 (or later) and [Java 17 (or later, up to 23)](http://www.oracle.com/technetwork/java/javase/downloads/index.html) to be installed. You can see which version you have using the following command:
 
 ```bash
 java -version
 ```
 
+:::{versionchanged} 24.08.0-edge
```
pditommaso, line 15:
```suggestion
:::{versionchanged} 24.11.0-edge
```

pditommaso:
Moving forward this

PR #5051, opened 2024-06-06
Add note that beforeScript runs inside the container for some execution environments

ewels:
Wording isn't great, so feel free to rewrite as needed @nextflow-io/docs 🙏🏻 😊

### DIFF
--- a/docs/process.md
+++ b/docs/process.md
@@ -1410,6 +1410,7 @@ process foo {
 
 :::{note}
 When combined with the {ref}`container directive <process-container>`, the `beforeScript` will be executed outside the specified container. In other words, the `beforeScript` is always executed in the host environment.
+The exception to this is container-native execution environments such as cloud batch and kubernetes, where `beforeScript` must run inside the container.
 :::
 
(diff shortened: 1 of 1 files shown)


docs/process.md
```diff
@@ -1410,6 +1410,7 @@ process foo {
 
 :::{note}
 When combined with the {ref}`container directive <process-container>`, the `beforeScript` will be executed outside the specified container. In other words, the `beforeScript` is always executed in the host environment.
+The exception to this is container-native execution environments such as cloud batch and kubernetes, where `beforeScript` must run inside the container.
 :::
```
bentsherman, line 1414:
```suggestion
When the process is containerized (using the {ref}`process-container` directive), the `beforeScript` will be executed in the container only if the executor is *container-native* (e.g. cloud batch executors, Kubernetes). Otherwise, the `beforeScript` will be executed outside the container.
```

bentsherman (changes requested):

pditommaso (approved):

PR #5055, opened 2024-06-07
timelime -> timeline

lindenb:
fix a typo for a message giving a wrong advice : 
```
enable the 'timelime.overwrite' option in your config 
```

to

```
enable the 'timeline.overwrite' option in your config 
```


pditommaso (approved):
Well spotted!

PR #5056, opened 2024-06-10
Remove GA4GH plugin

pditommaso:
This PR removes the GA4GH TES plugin from the Nextflow core codebase. The plugin source code will be maintained in the [nf-ga4gh](https://github.com/nextflow-io/nf-ga4gh) repository


bentsherman:
Why is the latest version 1.1.0-patch1 when the previous version was 1.3.0? Isn't that going to mess up the plugin version picker?

pditommaso:
It should not. `1.3.0` is greater than `1.1.0-patch1`

bentsherman:
I'm confused. You made the standalone plugin 1.1.0 , so Nextflow will always select the "old" version 1.3.0 by default?

pditommaso:
it should not, because it requires 23.10.2 

https://github.com/nextflow-io/plugins/blob/main/plugins.json#L1214-L1218

bentsherman:
so if I release a new version of the plugin, should I start from `1.1.0` or `1.4.0` ?

pditommaso:
1.4.0

PR #5061, opened 2024-06-13
Fix the cleanup of local secret file

pditommaso:
This PR fixes the cleanup of secret files created by the local provider

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/secret/LocalSecretsProvider.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/secret/LocalSecretsProvider.groovy
@@ -215,6 +215,9 @@ class LocalSecretsProvider implements SecretsProvider, Closeable {
         // make sure the file can only be accessed by the owner user
         path.setPermissions(ONLY_OWNER_PERMS)
         path.text = result
+        // remove it on completion
+        path.toFile().deleteOnExit()
+        // return the temp path
         return path
     }
 }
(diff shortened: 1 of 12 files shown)


marcodelapierre (approved):
I have added `deleteOnExit()` also for when the path is preexisting.  
This makes the behaviour retroactive, for tmp files with matching hash.

What do you think?

pditommaso:
Umm, not sure anymore it should be deleted 😄

pditommaso:
The point that file could be deleted during a by a run while a concurrent execution is running. 

I think the temp file is removed, then the file name should be unique i.e. using a random uuid

marcodelapierre:
See now - added a random UUID to the filename.  
I did not use the session uniqueID, as it seemed an overshooting to expose the session object all the way to this class.

modules/nextflow/src/main/groovy/nextflow/secret/LocalSecretsProvider.groovy
```diff
@@ -198,11 +198,13 @@ class LocalSecretsProvider implements SecretsProvider, Closeable {
         if( !secretsMap )
             return null
 
-        final name = ".nf-${CacheHelper.hasher(secretsMap.values()).hash()}.secrets"
+        final name = ".nf-${CacheHelper.hasher(secretsMap.values()).hash()}-${UUID.randomUUID().toString()}.secrets"
```
pditommaso, line 201: Let's just use the random uuid

marcodelapierre: done now

marcodelapierre:
unrelated commits should be fine now, see changelog

marcodelapierre:
(still showing extra commits in commit list, as I rebased to sign off one commit)

PR #5065, opened 2024-06-14
Do not require tentantId when using Azure system-assigned managed identity

adamrtalbot:
Azure system-assigned managed identity no longer needs the tenant ID after we updated the Java SDK, but the checks prevented it being allowed. This PR removes the check for system + tenant ID so system assigned identities can be used.

I've tested this to work on an Azure VM with a managed identity.

### DIFF
--- a/plugins/nf-azure/src/test/nextflow/cloud/azure/batch/AzBatchServiceTest.groovy
+++ b/plugins/nf-azure/src/test/nextflow/cloud/azure/batch/AzBatchServiceTest.groovy
@@ -710,7 +710,7 @@ class AzBatchServiceTest extends Specification {
     }
 
     @Unroll
-    def 'should create manage credentials token' () {
+    def 'should create user-assigned managed identity credentials token' () {
         given:
         def config = Mock(AzConfig)
         def exec = Mock(AzBatchExecutor) {getConfig() >> new AzConfig(CONFIG) }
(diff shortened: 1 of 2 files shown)


plugins/nf-azure/src/test/nextflow/cloud/azure/batch/AzBatchServiceTest.groovy
```diff
@@ -729,4 +729,24 @@ class AzBatchServiceTest extends Specification {
...
+        def config = Mock(AzConfig)
+        def exec = Mock(AzBatchExecutor) {getConfig() >> new AzConfig(CONFIG) }
+        AzBatchService service = Spy(new AzBatchService(exec))
+
+        when:
+        def token = service.createBatchCredentialsWithManagedIdentity()
+        then:
+        config.managedIdentity() >> { Mock(AzManagedIdentityOpts) }
+        then:
+        token instanceof ManagedIdentityCredential
```
adamrtalbot, line 745: @pditommaso it's not clear to me how to test this. There should not be a `clientId` property, but should be a valid token.

pditommaso: how would you check for the valid token?

adamrtalbot: 🤷

pditommaso: I meant what's the condition you want to check

pditommaso:
Merged manually a639a17d

PR #5071, opened 2024-06-17
4446 ignore then fail strategy v3

pditommaso:
This PR implements the fail on ignore by using by setting the option `nextflow.enable.failOnIgnore` in the nextflow config file.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -223,7 +223,7 @@ class Session implements ISession {
 
     private Barrier monitorsBarrier = new Barrier()
 
-    private volatile boolean failOnComplete
+    private volatile boolean failOnIgnore
 
     private volatile boolean cancelled
 
@@ -826,7 +826,7 @@ class Session implements ISession {
 
     boolean isCancelled() { cancelled }
 
-    boolean isSuccess() { !aborted && !cancelled && !failOnComplete }
+    boolean isSuccess() { !aborted && !cancelled && !failOnIgnore }
 
     void processRegister(TaskProcessor process) {
         log.trace ">>> barrier register (process: ${process.name})"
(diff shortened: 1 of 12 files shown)


bentsherman:
I would suggest exposing it as `nextflow.failOnIgnore` so as to not confuse it with feature flags. Long term I don't think we should allow feature flags in the config because they are inherent to the pipeline code.

bentsherman:
In other words:
- `nextflow` config scope is used for runtime options and is only relevant in the config file
- feature flags are used for language options and are only relevant in the script

pditommaso:
That was the first option, but then I realised it's not a feature flag but more a workflow (error handling) behaviour, and as such reported in the workflow metadata

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -866,6 +866,10 @@ class Session implements ISession {
         config.navigate('nextflow.enable.moduleBinaries', false) as boolean
     }
 
+    boolean failOnIgnore() {
+        config.navigate('workflow.failOnIgnore', false) as boolean
```
bentsherman, line 870:
```suggestion
        config.navigate('nextflow.failOnIgnore', false) as boolean
```

pditommaso: Check my comment https://github.com/nextflow-io/nextflow/pull/5071#issuecomment-2174019219

bentsherman: It should go in this config scope: https://nextflow.io/docs/latest/config.html#scope-nextflow

pditommaso: Disagree

bentsherman: I'm not talking about feature flags, I'm talking about the `nextflow` config scope which contains the publish retry settings. It makes sense to group them together

pditommaso: yeah, that was somehow temporary. We are going to deprecated `publishDir`, therefore those settings will likely go under `workflow.output.retryPolicy`.

PR #5072, opened 2024-06-17
Move Wave docs to Wave documentation

jason-seqera:
This PR shifts the Wave-specific guides over to the Wave repo and the Wave documentation site at: https://docs.seqera.io/wave


ewels (approved):
LGTM 👍🏻 

@jason-seqera - this repo needs signed commits, which is why the CI check is failing. Easy to fix:

```bash
git commit --amend --sign
# Accept / save commit message as before
git push --force
```

pditommaso:
I think we should keep on hold until we don't have the new wave docs ready

ewels:
Ah I assumed that there was a paired PR to add these back into the Seqera docs. Yes agree, they should be added there before they're removed here.

jason-seqera:
@pditommaso @ewels - We could move over these tutorials in the meantime, though, and then that portion of the project would be complete. None of these specifically depend on the overall content refresh being complete.

ewels:
Aha - it's here! https://github.com/seqeralabs/wave/pull/516

Sorry, didn't see it before. Yes if we can merge that first ☝🏻 then the docs will appear at docs.seqera.io. Then I think it's fine to merge this PR as the content will still be available and we can point there.

So just a question of sequencing the merges correctly from my point of view.

@pditommaso is it not better to do this before before the big Wave docs rewrite? I feel like this content should probably be included in the docs review for the new wave docs, rather than being tacked on afterwards...

pditommaso:
I'm not okay with deleting the Wave docs from Nextflow until there's an update and comprehensive documentation for Wave. 

My rationale: 1) moving pieces around just creates confusion, 2) the Wave docs are not blocked by this, and 3) there are many places, including Platform docs, that link the Wave docs in Nextflow.

jason-seqera:
This content now also lives here: https://docs.seqera.io/wave/nextflow

PR #5087, opened 2024-06-24
Enable multiple git revisions (Jan-Apr 2024) : using revision name in repo localPath

marcodelapierre:
This is spawned out of #4659 .

Original attempt, using revisions not commit IDs, in localPath (as per original thread).

Pros:
- simple implementation

Cons:
- if project is updated, running pipelines with same project will see underlying pipeline files change during the run. Troublesome when using `run -latest` and in multi-user settings
- AssetManager instantiates itself circularly to gather info on default branch (to avoid local duplication of default branch directories)


bentsherman:
Closing in favor of #6620

Issue #4920, opened 2024-04-16
Azure Batch: Add disk size to slots calculation


adamrtalbot:
## New feature

When using Azure Batch, Nextflow will reject a process if it has too many CPUs for the worker machine.

```
Caused by:
  Process requirement exceeds available CPUs -- req: 32; avail: 10
```

However, Azure Batch VMs come with a fixed disk and it's common that the Nextflow process runs out of storage. There are many, many issues about this on the Nextflow Slack! The typical workaround is to increase the number of CPUs an individual process requires, however it would be better to support the disk directive so we can directly enforce the VMs have the right sized disk.

Although we can't enforce it properly (i.e. make sure tasks are only assigned to a VM with enough space), being able to prevent users trying to run a task on a machine which is too small would catch some of the issues.

## Usage scenario 

When running on Azure Batch, raise an error if a task is assigned to a queue which does not contain sufficient storage.

## Suggest implementation 

```nextflow
process HELLO {
    disk 12.TB

    """
    echo Hello
    """
}

workflow {
    HELLLO()
}
```

```
Caused by:
  Process requirement exceeds available storage -- req: 10TB; avail: 1TB
```

bentsherman:
how would you determine the available disk storage for a given VM / queue?

adamrtalbot:
I'm thinking way simpler than that. 

if `process.disk = '1024GB'` and VM disk size is 512GB ([we have the disk size](https://github.com/nextflow-io/nextflow/blob/master/plugins/nf-azure/src/resources/nextflow/cloud/azure/vm-list-size-westus.json)) then prevent a job being submitted to that queue. It's not foolproof, but it would catch errors early.

bentsherman:
I see, so if the user specifies a machine type or a queue with an implied machine type then we can use `disk` to validate the requirement, but if the user specifies `cpus` and `memory` with auto-pools enabled then we could actually use `disk` to narrow the query of valid machine types from this list. Does that sound right?

adamrtalbot:
> I see, so if the user specifies a machine type or a queue with an implied machine type then we can use disk to validate the requirement, but if the user specifies cpus and memory with auto-pools enabled then we could actually use disk to narrow the query of valid machine types from this list. Does that sound right?

Correct. It's not perfect, but it might help catch a few mistakes.

We would need to add the disk directive here:
[...]

PR #5120, opened 2024-07-08
Support disk directive in Azure Batch

adamrtalbot:
This PR adds the `process.disk` directive support to Azure Batch. After adding this directive, Azure Batch will:

- Factor disk and VM disk size into the number of slots a single task occupies
- Create autopools that use VMs with sufficiently large enough storage disks for a single Azure Batch task
- When using autopools, select the node with the fewest features (likely to be the cheapest)

This should help with the recurring issue where Azure Batch nodes run out of disk space because too many tasks are packed onto a single machine. Since we don't support using a network attached disk as the working directory this is the only solution to handling larger files.

Similarly, when using Fusion we can ringfence a larger fraction of the local storage drive to ensure the local cache is not overloaded.


plugins/nf-azure/src/test/nextflow/cloud/azure/batch/AzBatchServiceTest.groovy
```diff
@@ -211,19 +211,21 @@ class AzBatchServiceTest extends Specification {
...
-        2     | 64    | 4         | 64        |   4
+        1     | 1     | 1    | 4         | 64        | 4       |   1
+        1     | 8     | 1    | 4         | 64        | 4       |   1
+        1     | 16    | 1    | 4         | 64        | 4       |   1
+        1     | 32    | 1    | 4         | 64        | 4       |   2
+        1     | 1     | 4    | 4        | 64        | 4       |   4
+        1     | 1     | 4    | 4        | 64        | 8       |   2
+        2     | 1     | 1    | 4         | 64        | 4       |   2
+        4     | 1     | 1    | 4         | 64        | 4       |   4
+        2     | 64    | 1    | 4         | 64        | 4       |   4
```
adamrtalbot, line 228:
```suggestion
        1     | 1     | 1    | 4         | 64        | 4       |   1
        1     | 8     | 1    | 4         | 64        | 4       |   1
        1     | 16    | 1    | 4         | 64        | 4       |   1
        1     | 32    | 1    | 4         | 64        | 4       |   2
        1     | 1     | 4    | 4         | 64        | 4       |   4
        1     | 1     | 4    | 4         | 64        | 8       |   2
        2     | 1     | 1    | 4         | 64        | 4       |   2
        4     | 1     | 1    | 4         | 64        | 4       |   4
        2     | 64    | 1    | 4         | 64        | 4       |   4
```

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -188,20 +188,24 @@ class AzBatchService implements Closeable {
...
+                def score = computeScore(cpus, mem, disk, entry)
                 if( score != null )
                     scores.put(score, entry.name as String)
             }
         }
 
+        if ( scores ) {
+            log.debug "[AZURE BATCH] Found VMs: ${scores}"
+        }
+
```
adamrtalbot, line 208: This is purely to help me debug.

```suggestion
```

bentsherman:
close #4920

docs/azure.md
```diff
@@ -266,6 +266,18 @@ azure {
 ```
 :::
 
+### Task packing on nodes
+
+Each node is given a number of task slots, which is the number of tasks that can be run concurrently on the node. The number of task slots is determined by the number of cores for the selected VM. Nextflow will assign each process a number of task slots equal to a percentage of the total resources avaiable on the node, based on the `cpus`, `memory` and `disk` directives.
```
adamrtalbot, line 271:
```suggestion
Each node is given a number of task slots, which is the number of tasks that can be run concurrently on the node. The number of task slots is determined by the number of cores for the selected VM. Nextflow will assign each process a number of task slots equal to a percentage of the total resources available on the node, based on the `cpus`, `memory` and `disk` directives.
```

PR #5122, opened 2024-07-08
Link to other docs on the failOnIgnore option

jfy133:
As requested by @ewels .

Note there is an earlier reference to the option in the docs, but I didn't feel it fit in there in the same way, as there is not a particular markdown entry for the specific option on `metadata.md`


jfy133:
@ewels the page cross-links aren't rendering  on the preview, I'm not sure why, I copied and pasted from other examples of such links (although admittedly it doesn't make sense to me how they work)

docs/process.md
```diff
@@ -1691,6 +1691,8 @@ To illustrate the differences between error strategies, it can be helpful to ima
 - `finish`: Nextflow will allow any existing jobs to conclude but not submit any more and report an error
 - `ignore`: Nextflow will proceed to continue submitting processes for the remaining 95 samples and proceed to completion. Nextflow will ignore the error and report a successful pipeline completion
 - `ignore` and `workflow.failOnIgnore`: Nextflow will proceed to continue submitting processes for the remaining 95 samples and proceed to completion, then exit with an error status and report the error
+
+See {ref}`metadata` for more information on `workflow.failOnIgnore`.
```
adamrtalbot, line 1695:
```suggestion
See {ref}`metadata-workflow` for more information on `workflow.failOnIgnore`.
```

docs/metadata.md
```diff
@@ -63,6 +63,8 @@ The following table lists the properties that can be accessed on the `workflow`
   :::
 : Whether the `workflow.failOnIgnore` config option was enabled.
 
+: See {ref}`errorStrategy` for use-cases of this option.
```
adamrtalbot, line 66:
```suggestion
: See {ref}`process-error-strategy` for use-cases of this option.
```

adamrtalbot (commented):
I think...

adamrtalbot (approved):
Yay it works. You can customise the text using the snippet I shared in a comment.

docs/process.md
```diff
@@ -1691,6 +1691,8 @@ To illustrate the differences between error strategies, it can be helpful to ima
 - `finish`: Nextflow will allow any existing jobs to conclude but not submit any more and report an error
 - `ignore`: Nextflow will proceed to continue submitting processes for the remaining 95 samples and proceed to completion. Nextflow will ignore the error and report a successful pipeline completion
 - `ignore` and `workflow.failOnIgnore`: Nextflow will proceed to continue submitting processes for the remaining 95 samples and proceed to completion, then exit with an error status and report the error
+
```
pinin4fjords, line 1694: Suggestions for above lines (since I was too slow in the last PR, and I don't have permission to push to this one):

```
To illustrate the differences between error strategies, consider the following example. Let's say you are analyzing 96 patient samples and the data from one is corrupted, causing the associated task to fail. The different ways to handle this failure are as follows:

- **errorStrategy `terminate`**: Nextflow will cancel any other ongoing tasks at the time of the failure, exit the pipeline, and report an error.
- **errorStrategy `finish`**: Nextflow will allow any other existing tasks to conclude (but not submit any more) and report an error.
- **errorStrategy `ignore`**: Nextflow will continue submitting tasks for the remaining 95 samples, complete the workflow, and report a successful pipeline completion.
- **errorStrategy `ignore` and `workflow.failOnIgnore` set to `true` in configuration**: The same behavior as setting the errorStrategy alone, except the pipeline will return an exit status of -1 and report an error.
```

jfy133: Works for me if @adamrtalbot is happy?

ewels: I like it 👍🏻 

"same as setting the `ignore` errorStrategy"

on the last line maybe. But otherwise great 👍🏻

pinin4fjords: resolved in https://github.com/nextflow-io/nextflow/pull/5123

pditommaso (approved):

PR #5124, opened 2024-07-08
chore: Upgrade docs version. 

gavinelder:
# Summary

The following upgrades our version of Python from 3.7 to 3.12 along with relevant dependency upgrades to support to version upgrade, the most notable of which is the major version upgrade of Sphinx from 5 -> 7 

## Testing

Testing was carried out as per the relevant documentation by building the docker container locally , generating the relevant documentation then serving the html locally, checking for build & output errors along with doing a visual comparison across all pages to ensure correct rendering.  

image

## Changes to Azure.md

The `azure.md` file contained invalid [Pygments lexer's](https://pygments.org/docs/lexers/) such as `config` & `nextflow` these have been updated to `groovy` to match existing doc-styles.

There was also some drive-by mark-down formatting. 

## Docker changes.

- Replaced deprecated Maintainer tag with a label.
- Moved the installed versions out to arguments to make changes & version pinning easier to parse.

## Requirements changes.

- Pip froze all installed dependancies in requirements.txt via `pip freeze > requirements.txt`
- defined top level dependancies in `requirements-top-level.txt` to help identify direct vs indirect-dependancies and allow a regeneration of requirements and to prune old unneeded dependancies if they're ever removed in the future.


pditommaso:
I had to revert this because it was insanely slow when running in my laptop 👉 770b3dde

PR #5141, opened 2024-07-12
Improve Google Batch 5000x error class handling

pditommaso:
This PR improve the handling of Google Batch 5000x error class in such a way that 

1. All 50001 .. 50006 error codes managed, see [here](https://cloud.google.com/batch/docs/troubleshooting#vm_preemption_50001)
2. When this error condition arises is converted into `ProcessException` including the source error description 
3. The usual Nextflow `errorRetry` strategies can be applied to handle the error condition


bentsherman:
See also: https://github.com/nextflow-io/nextflow/blob/cb63ec9a905d81510f8bdc4d2c6af3827fde1462/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy#L262-L277

VasLem:
Just a user opinion, I would prefer these errors not to change the task attempt number, as in a dynamic configurations this would lead to unnecessary increase of resources, or unnecessary termination/ignoration of the process.

pditommaso:
This change makes it possible to recover errors from 50002 to 50006, similarly to 50001. Currently, they are causing the termination of the pipeline execution. 


@bentsherman Regarding the Batch automatic recovery, I didn't change it on purpose because the semantic settings are specific for spot interruption since it's named `maxSpotAttempts`. 

The this failures can be recovered with the usual error strategy.

bentsherman:
From meeting discussion, we suggested using a new config setting e.g. `google.batch.autoRetryExitCodes` to provide a list of exit codes to retry natively (without increasing the task attempt)

pditommaso:
Merging this. I'll open a separate PR for the retry logic at [this comment](https://github.com/nextflow-io/nextflow/pull/5141#issuecomment-2228587075)

PR #5148, opened 2024-07-15
Make auto retry codes configurable

pditommaso:
This PR introduces a new config option that allows setting one more exit codes when running a task with Google Batch that should trigger an automatic retry when a task returns an error code and the opinion `google.batch.maxSpotAttempts` is set to a value greater than zero.

By default this option is set to `[50001]`. See the [Google docs](https://cloud.google.com/batch/docs/troubleshooting#vm_preemption_50001) for more details.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -848,8 +848,7 @@ The following settings are available for Google Cloud Batch:
 `google.batch.autoRetryExitCodes`
 : :::{versionadded} 24.06.0-edge
   :::
-: Defines the task exit codes that determine an automatic task execution retry when the setting `maxSpotAttempts` is set
-to a value greater than 0. (default `[50001]`). For more details check out the [Google Batch documentation](https://cloud.google.com/batch/docs/troubleshooting#vm_preemption_50001).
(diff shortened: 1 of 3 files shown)


docs/config.md
```diff
@@ -865,6 +871,7 @@ The following settings are available for Google Cloud Batch:
 : :::{versionadded} 23.11.0-edge
   :::
 : Max number of execution attempts of a job interrupted by a Compute Engine spot reclaim event (default: `5`).
+See also `autoRetryExitCodes`.
```
pditommaso, line 874: @bentsherman How to link each other these settings?

bentsherman: We don't typically do this, but you should be able to put a link target over the config option like `(google-batch-auto-retry-exit-codes)=` and reference it like any other link:
```
: See also: {ref}`google-batch-auto-retry-exit-codes`
```

pditommaso: yeah, it would bye much better having automatic links for each option

docs/config.md
```diff
@@ -845,6 +845,11 @@ Read the {ref}`google-page` page for more information.
 
 The following settings are available for Google Cloud Batch:
 
+`google.batch.autoRetryExitCodes`
+: :::{versionadded} 24.06.0-edge
```
pditommaso, line 849:
```suggestion
: :::{versionadded} 24.07.0-edge
```

docs/config.md
bentsherman: I think Phil and the docs team want to do this at some point. Maybe we can tackle this in the future as we migrate docs

PR #5157, opened 2024-07-20
[Docs] Plugin test repository note

llewellyn-sl:
Add `NXF_PLUGINS_TEST_REPOSITORY` JSON filename admonition, per  https://seqera.slack.com/archives/C040MFU06AE/p1721422335135079

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -365,7 +365,7 @@ You can also use a different plugin registry with the `NXF_PLUGINS_TEST_REPOSITO
 For example:
 
 ```bash
-# custom registry at https://github.com/my-org/plugins
+# custom plugin repository at https://github.com/my-org/plugins
 export NXF_PLUGINS_TEST_REPOSITORY="https://raw.githubusercontent.com/my-org/plugins/main/plugins.json"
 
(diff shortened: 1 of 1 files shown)


docs/plugins.md
```diff
@@ -374,6 +374,10 @@ export NXF_PLUGINS_TEST_REPOSITORY="https://github.com/nextflow-io/nf-hello/rele
 nextflow run <pipeline> -plugins nf-hello
 ```
 
+:::{tip}
+To differentiate an `NXF_PLUGINS_TEST_REPOSITORY` value that points to a JSON file containing a single plugin (second example above) instead of a full custom repository (first example), the single-plugin JSON filename must match the regular expression `/(.+)-(\d+\.\d+\.\d+\S*)-meta\.json/`.
```
pditommaso, line 378: Let's improve something like: 

> This variable can be either the URI of a plugins repository JSON file or the URI of a specific repository JSON meta file. In the latter case it should match the pattern `<plugin id>-X.Y.Z-meta.json`.

pditommaso: Merged manually 7703f0d0

PR #5166, opened 2024-07-25
Improve fusion docs

pditommaso:
This PR improves the documentation for Fusion adding specific section for Google and Azure


docs/fusion.md
```diff
@@ -13,15 +13,103 @@ Support for Google Cloud Storage.
...
+tower.accessToken = '<your platform access token>'
+```
+
+Then run your pipeline using the usual command:
+
+```
+nextflow run <your pipeline> -work-dir az://<your blob container>/scratch
+```
+
+[TODO add some notes on Azure Blob permission and the need for SSD disk (?)]
```
adamrtalbot, line 48:
```suggestion
[TODO add some notes on Azure Blob permission]

Azure machines come with fast SSDs attached, therefore no additional storage configuration is required however it is recommended to use the machine types with larger data disks attached, denoted by the suffix 'd' after the core number (e.g. Standard_E32*d*_v5). These will increase the throughput of Fusion and reduce the chance of overloading the machine.
```

docs/fusion.md
```diff
@@ -13,15 +13,103 @@ Support for Google Cloud Storage.
...
+
+process {
+    executor = 'awsbatch'
+    queue = '<YOUR BATCH QUEUE>'
+}
+
+aws {
+    region = '<YOUR AWS REGION>'
+}
+```
```
adamrtalbot, line 75: This should be consistent with the Azure and GCP syntax.

```groovy
process.executor = 'awsbatch'
process.queue = '<YOUR BATCH QUEUE>'
wave.enabled = true
fusion.enabled = true
tower.accessToken = '<your platform access token>'
aws.region = '<YOUR AWS REGION>'
```

(or go the other way which looks nicer!)

docs/fusion.md
```diff
@@ -13,15 +13,103 @@ Support for Google Cloud Storage.
...
+
+#### NVMe storage
+
+The Fusion file system implements a lazy download and upload algorithm that runs in the background to transfer files in
+parallel to and from object storage into a container-local temporary folder. This means that the performance of the temporary
+folder inside the container (`/tmp` in a default setup) is key to achieving maximum performance.
+
+The temporary folder is used only as a temporary cache, so the size of the volume can be much lower than the actual needs of your
+pipeline processes. Fusion has a built-in garbage collector that constantly monitors remaining disk space on the temporary folder
+and immediately evicts old cached entries when necessary.
```
adamrtalbot, line 95: Perhaps this should go at the top to indicate it's a general feature of Fusion.

docs/fusion.md
pditommaso: Any instance type? really?

jordeu: Yes, all the instance types in Azure come with a local SSD disk, but some can be too small. So I agree with Adam suggestion.

pditommaso: Please push the changes

docs/fusion.md
```diff
@@ -54,75 +142,30 @@ The AWS S3 bucket should be configured with the following IAM permissions:
...
-}
-
-aws {
-    region = '<YOUR AWS REGION>'
-}
+nextflow run <your pipeline> -work-dir gs://<your google bucket>/scratch
 ```
 
-Then you can run your pipeline using the following command:
+[TODO add some notes on Google Storage permission and the need for SSD disk and instance type(?)]
```
jordeu, line 166:
```suggestion
When using Fusion, if the `process.disk` is not set, Nextflow will attach a single local SSD disk to the machine. The size of this disk can be much lower than the actual needs of your pipeline processes because Fusion uses it only as a temporal cache. Fusion is also compatible with other types of `process.disk`, but better performance is achieved when using local SSD disks.
```

docs/fusion.md
```diff
@@ -13,15 +13,105 @@ Support for Google Cloud Storage.
...
+tower.accessToken = '<your platform access token>'
+```
+
+Then run your pipeline using the usual command:
+
+```
+nextflow run <your pipeline> -work-dir az://<your blob container>/scratch
+```
+
+[TODO add some notes on Azure Blob permission]
```
jordeu, line 48: I'm not aware that, to use Fusion, we need any special permission on Azure Blob

jordeu (approved):

docs/fusion.md
```diff
@@ -13,15 +13,104 @@ Support for Google Cloud Storage.
...
+fusion.enabled = true
+tower.accessToken = '<your platform access token>'
+```
+
+Then run your pipeline using the usual command:
+
+```
+nextflow run <your pipeline> -work-dir az://<your blob container>/scratch
+```
+
```
adamrtalbot, line 47:
```suggestion
```

docs/fusion.md
pditommaso: make sense

docs/fusion.md
pditommaso: agree

PR #5169, opened 2024-07-26
Update script.md for mklink

bjlang:
The documentation for mklink indicates options as 2nd parameter ( mklink( linkName, options = [:] ) ), however they need to be provided as 1st parameter, following the function signature ( [mklink(java.util.Map opts, java.nio.file.Path link)](https://javadoc.io/static/io.nextflow/nextflow/20.04.1-edge/nextflow/processor/TaskPath.html#mklink(java.util.Map,%20java.nio.file.Path)) )

### DIFF
--- a/docs/script.md
+++ b/docs/script.md
@@ -1186,12 +1186,13 @@ The following methods are available for manipulating files and directories in a 
   file('any/path').mkdirs()
   ```
 
-`mklink( linkName, options = [:] )`
+`mklink( options = [:], linkName )`
 : Creates a *filesystem link* to a given path:
 
   ```groovy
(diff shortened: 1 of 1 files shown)


bentsherman:
Named args can be specified in any order. Groovy collects them into a Map and passes it as the first argument.

Though I see your point that if you provided an explicit map rather than named args, you would need to provide it first. I think we can clarify the docs by saying `[options]` instead of `options = [:]`, but we'll need to do that across the board. I will make a PR for this. Thanks for your suggestion

bentsherman:
Closing in favor of #5171

PR #4766, opened 2024-02-21
Separate reference docs from manual pages

bentsherman:
This PR reorganizes the docs in a number of ways:

- reorganize manual pages into "Running pipelines" and "Developing pipelines" to be more aligned with the user journey
- separate reference content (config scopes, process directives, etc) from the main docs
- improve section on managing dependencies

The original exploration was around refactoring the docs into a user guide vs reference manual. After a few iterations, the main focus now is the first two items, separating the reference docs and reorganizing the sections to be more user friendly.

PR #5171, opened 2024-07-26
Improve documentation of named args

bentsherman:
Merge after #4766 

Based on #5169 . Refactors the docs for functions with named args by removing the `[:]` since that is not how named args are typically passed and it is technically incorrect. Also added a named arg to the `mklink` example.


christopher-hakkaart (approved):
Nice change!

Issue #5176, opened 2024-07-26
containerOptions directive does not support shm-size where the memory unit is specified (when running on awsbatch)


robsyme:
## Bug report 

Docker's `--shm-size` parameter supports memory unit sizes (docs). When this argument is specified in a `containerOptions` directive, the parameter is passed directly to the `docker run` command run using docker, but when using the `awsbatch` executor, we parse those args so that they can be turned into jobDescription API arguments.

When we pass those arguments, the value is assumed to be a simple integer which is cast using `as Integer`: 

https://github.com/nextflow-io/nextflow/blob/2fb5bc07f2ad1309c9743b8675bb8003892e3eb7/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsContainerOptionsMapper.groovy#L107-L111

If the user specifies a unit:

```groovy
process.containerOptions = "--shm-size 256m"
```

This string `256m` will be cast into an Integer, which will fail.

### Expected behavior and actual behavior

Nextflow should use the (very handy) MemoryUnit class to turn this string into an integer number of MiB. Note that this needs to be in MiB and not bytes (as per the [AWS ContainerProperties documentation](https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html))

### Steps to reproduce the problem

Try to submit a task that uses `process.containerOptions = "--shm-size 256m"` to AWS Batch.

### Program output 

Fails with 

```
ERROR ~ Error executing process > 'WithSharedMemory'

Caused by:
  For input string: "250m"

 -- Check '.nextflow.log' file for details
```

[Nextflow log](https://github.com/user-attachments/files/16397048/nextflow.log) attached.

PR #5177, opened 2024-07-26
More robust parsing of shm-size containerOptions

robsyme:
This commit ensures that the `shm-size` option is correctly parsed from the containerOptions. This is necessary to ensure that the correct value in MiB is passed to the AWS Batch API when creating a job definition.

Fixes https://github.com/nextflow-io/nextflow/issues/5176

### DIFF
--- a/plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsContainerOptionsMapperTest.groovy
+++ b/plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsContainerOptionsMapperTest.groovy
@@ -94,7 +94,25 @@ class AwsContainerOptionsMapperTest extends Specification {
         def map = CmdLineHelper.parseGnuArgs('--shm-size 12048024')
         def properties = AwsContainerOptionsMapper.createContainerProperties(map)
         then:
-        properties.getLinuxParameters().getSharedMemorySize() == 12048024
+        properties.getLinuxParameters().getSharedMemorySize() == 11
+    }
+
+    def 'should set shared memory size with unit in MiB'() {
+
+        when:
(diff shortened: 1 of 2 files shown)


robsyme:
This does adjust an existing test, but I'd argue that the original test was incorrect, given that AWS measure shared memory in MiB and not in bytes (see [sharedMemorySize docs](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html)).

plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsContainerOptionsMapperTest.groovy
```diff
@@ -94,7 +94,25 @@ class AwsContainerOptionsMapperTest extends Specification {
         def map = CmdLineHelper.parseGnuArgs('--shm-size 12048024')
         def properties = AwsContainerOptionsMapper.createContainerProperties(map)
         then:
-        properties.getLinuxParameters().getSharedMemorySize() == 12048024
+        properties.getLinuxParameters().getSharedMemorySize() == 11
```
robsyme, line 97: Because AWS measures shared memory size in MiB rather than bytes (rather than the docker standard of bytes), the `AwsContainerOptionsMapper` should do the conversion automatically.

pditommaso: Well spotted Rob. Just made a minor change to make it more explicit 548d40a7

pditommaso (approved):
Excellent! Thanks Rob.

ewels:
Concerning forum post that could indicate a problem with this PR: https://community.seqera.io/t/shared-memory-bug/1090

PR #5185, opened 2024-07-30
Finalize workflow output definition

### LINKED ISSUE #5103: Finalize workflow output definition
Continuation of #4670 

Enumerating the proposed changes that we've collected so far:

- **Support additional index file formats** (json, yaml)

- ~~**Generate output schema.** Essentially a list of index file schemas. Should be generated on each run or via some separate command. Should eventually be used with parameter schema for chaining pipelines. See https://github.com/nf-core/fetchngs/pull/312 for a concrete example (`schema_outputs.yml`).~~ -> #5213 

- **Dynamic path mapping.** Allow the `path` option in target definition to be a closure:
  ```groovy
  workflow {
    main:
    ch_foo = Channel.of( [ [id: 1], file('foo.fastq') ] )

    publish:
    ch_foo >> 'foo'
  }

  output {
    'foo' {
      path { meta, fastq -> "foo/${meta.id}" }
    }
  }
  ```

  Note that the dynamic path need only define the directory, not the full filename. Since a channel value may contain multiple files, an alternative syntax could be to provide the file and the channel value to the closure, so that it's clear which file is being published:
  ```groovy
      path { file, value -> "foo/${value[0].id}" }
  ```

- **Move publish options to config.** Publish options like `directory` and `mode` typically need to be configurable by the user, which currently would require you to define a special param for each option. Therefore it makes more sense for them to be config settings rather than pipeline code:
  ```groovy
  // nextflow.config
  workflow {
    output {
      directory = 'results'
      mode = 'copy'

      withTarget:'foo' {
        mode = 'link'
      }
    }
  }

  // main.nf
  output {
    'foo' {
      index {
        path 'index.csv'
        // ...
      }
    }
  }
  ```

  The output block should be used only to define index files (i.e. the output schema). In other words, the pipeline code should define *what* is published and the config should define *how* it is published.

  For the output directory, it has also been proposed to provide a CLI option for it e.g. `-output-dir` and shorter config option e.g. `outputDir`. The output directory would be available in the pipeline code as part of workflow metadata i.e. `workflow.outputDir`.

- **Remove publish section from process definition**. Still under discussion. The rationale for the process publish section was to provide some sensible defaults which can be overridden, however I've come to think that it only makes it harder to determine what is being published. Instead of enumerating the publish targets in one place, they are scattered throughout the pipeline code. Also, process definitions are abstracted from any particular pipeline, so it doesn't make much sense to put pipeline-specific details like params and publishing in the process definition.

  A better way to give some sensible defaults for a process would be to write an entry workflow in the process definition that gives an example usage:
  ```groovy
  process foo {
    // ...
  }

  workflow {
    main:
    params.input = '...'
    foo( params.input )

    publish:
    foo.out >> 'foo'
  }
  ```

  This workflow will be ignored when importing the process as a module, but it provides a concrete example and can even be used to run the process directly. In fact it could even replace the custom nf-test DSL eventually.

- **Allow publish section only in entry workflow**. I am less certain about this one but wanted to lay out the case for it. Building on the previous item, having publish sections potentially spread across several workflows in a pipeline makes it hard to see what all is being published. Instead, maybe named workflows should only be allowed to emit outputs, and only the entry workflow be able to publish outputs. As with the previous point, you could write an entry workflow for each named workflow which gives some example publishing (and allow you to run the workflow as a standalone pipeline).

  This idea is inspired by a principle in software engineering that side effects (a.k.a. I/O, publishing) should be pushed to the "boundaries" of the code, to make it more readable and testable, and to make it easier to swap out different I/O strategies (file copy, database insert, API call, etc).

  At the same time, I appreciate that publishing from a named workflow is a convenient shorthand, especially when you considering having to propagate outputs back up through potentially several nested workflows. But I wonder if being more strict here would be better in the long run. The example entry workflow is something that will be written anyway, both for testing and to run workflows as standalone pipelines.

- **Runtime enhancements**
  - report target names in the output block that weren't defined in a process or workflow
  - prevent leading or trailing slashes
  - detect published file collisions at runtime
  - detect outputs that aren't published or used by downstream processes

- **Include output targets in inspect command**. Similar to how inspect lists all processes with some resolved directives, etc, it could also show the resolved list of output targets. Not as essential if we implement some of the above points, but still useful for things like resolving params.

bentsherman:
Close #5103 

TODO:

language improvements:
- [x] support dynamic path
- [x] move publish options to config
- [x] remove publish section from process definition
- [ ] preserve preview version ?

runtime improvements:
- [x] prevent leading/trailing slashes
- [ ] report undefined target names
- [ ] detect file collisions at runtime ?
- [ ] detect outputs that aren't published or used downstream ?

new features (might be handled separately):
- [ ] support JSON index file
- [ ] generate output schema
- [ ] include publish targets in inspect command ?

### DIFF
--- a/docs/workflow.md
+++ b/docs/workflow.md
@@ -41,12 +41,12 @@ The `main:` label can be omitted if there are no `take:` or `emit:` blocks.
 Workflows were introduced in DSL2. If you are still using DSL1, see the {ref}`dsl1-page` page to learn how to migrate your Nextflow pipelines to DSL2.
 :::
 
-## Implicit workflow
+## Entry workflow
 
-A script can define a single workflow without a name (also known as the *implicit workflow*), which is the default entrypoint of the script. The `-entry` command line option can be used to execute a different workflow as the entrypoint at runtime.
+A script can define a single workflow without a name (also known as the *entry workflow*), which is the default entrypoint of the script. The `-entry` command line option can be used to execute a different workflow as the entrypoint at runtime.
 
 :::{note}
-Implicit workflow definitions are ignored when a script is included as a module. This way, a script can be written such that it can be either imported as a module or executed as a pipeline.
+Entry workflow definitions are ignored when a script is included as a module. This way, a script can be written such that it can be either imported as a module or executed as a pipeline.
 :::
 
 ## Named workflows
@@ -82,7 +82,7 @@ workflow {
 ```
 
 :::{tip}
-The use of global variables and params in named workflows is discouraged because it breaks the modularity of the workflow. As a best practice, every workflow input should be explicitly defined as such in the `take:` block, and params should only be used in the implicit workflow.
+The use of global variables and params in named workflows is discouraged because it breaks the modularity of the workflow. As a best practice, every workflow input should be explicitly defined as such in the `take:` block, and params should only be used in the entry workflow.
 :::
 
 ## Workflow inputs (`take`)
@@ -397,30 +397,29 @@ In the above snippet, the initial channel is piped to the {ref}`operator-map` op
 
 ## Publishing outputs
 
-:::{versionadded} 24.04.0
+:::{versionadded} 24.10.0
 :::
 
 :::{note}
-This feature requires the `nextflow.preview.output` feature flag to be enabled.
+This feature was introduced in 24.04.0 as a preview. It can be enabled in that version using the `nextflow.preview.output` feature flag.
 :::
 
-A script may define the set of outputs that should be published by the implicit workflow, known as the workflow output definition:
+A workflow may define the set of outputs that should be published:
 
 ```groovy
 workflow {
+    main:
     foo(bar())
-}
 
-output {
-    directory 'results'
+    publish:
+    foo.out >> 'foo'
+    bar.out >> 'bar'
 }
 ```
 
-The output definition must be defined after the implicit workflow.
-
 ### Publishing channels
 
-Processes and workflows can each define a `publish` section which maps channels to publish targets. For example:
+Workflows can define a `publish` section which maps channels to publish targets. Any channel defined in the workflow, including process and subworkflow outputs, can be published. For example:
 
 ```groovy
 process foo {
@@ -429,39 +428,31 @@ process foo {
     output:
     path 'result.txt', emit: results
 
-    publish:
-    results >> 'foo'
-
     // ...
 }
 
-workflow foobar {
+workflow {
     main:
     foo(data)
     bar(foo.out)
 
     publish:
-    foo.out >> 'foobar/foo'
-
-    emit:
-    bar.out
+    foo.out.results >> 'foo'
+    bar.out >> 'bar'
 }
 ```
 
-In the above example, the output `results` of process `foo` is published to the target `foo/` by default. However, when the workflow `foobar` invokes process `foo`, it publishes `foo.out` (i.e. `foo.out.results`) to the target `foobar/foo/`, overriding the default target defined by `foo`.
+In the above example, the output `results` of process `foo` is published to the target `foo`, and all outputs of process `bar` are published to the target `bar`.
 
-In a process, any output with an `emit` name can be published. In a workflow, any channel defined in the workflow, including process and subworkflow outputs, can be published.
-
-:::{note}
-If the publish source is a process/workflow output (e.g. `foo.out`) with multiple channels, each channel will be published. Individual output channels can also be published by index or name (e.g. `foo.out[0]` or `foo.out.results`).
+:::{tip}
+A workflow can override the publish targets of a subworkflow by "re-publishing" the same channels to a different target. However, the best practice is to define all publish targets in the entry workflow, so that all publish targets are defined in one place at the top-level.
 :::
-
-As shown in the example, workflows can override the publish targets of process and subworkflow outputs. This way, each process and workflow can define some sensible defaults for publishing, which can be overridden by calling workflows as needed.
 
 By default, all files emitted by the channel will be published into the specified directory. If a channel emits list values, any files in the list (including nested lists) will also be published. For example:
 
 ```groovy
 workflow {
+    main:
     ch_samples = Channel.of(
         [ [id: 'sample1'], file('sample1.txt') ]
     )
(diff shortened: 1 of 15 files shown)


bentsherman:
Second preview is ready for review, POCs have also been updated:
- https://github.com/nf-core/fetchngs/pull/312
- https://github.com/nf-core/rnaseq/pull/1227

docs/workflow.md
```diff
@@ -516,167 +486,208 @@ results/
...
-  : Overwrite existing files when the file size or last modified timestamp is different.
+The closure can even define a different path for each individual file by returning an inner closure, similar to the `saveAs` option of the {ref}`publishDir <process-publishdir>` directive:
 
-`path`
-: Specify the publish path relative to the output directory (default: the target name). Can only be specified within a target definition.
+```groovy
+output {
+    'fastq' {
+        path { meta, fastq_1, fastq_2 ->
+            { file -> "fastq/${meta.id}/${file.baseName}" }
```
maxulysse, line 582: <3

nvnieuwk:
I just tried the dynamic publishing with some dummy code and it worked perfectly! I'm looking forward to implementing this in my pipelines :partying_face:

modules/nextflow/src/main/groovy/nextflow/script/OutputDsl.groovy
```diff
@@ -35,76 +37,28 @@ import nextflow.file.FileHelper
...
-        setDefault('storageClass', value)
-    }
-
-    void tags(Map value) {
-        setDefault('tags', value)
-    }
-
-    void enabled( boolean value ) {
-        setDefault('enabled', value)
-    }
```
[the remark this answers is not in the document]
bentsherman: yes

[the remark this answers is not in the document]
pditommaso: nice, and output top level? only in the config or config and script definition?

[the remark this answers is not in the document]
bentsherman: only the config. default output directory is `results`

docs/reference/config.md
```diff
@@ -1612,3 +1620,76 @@ The `workflow` scope provides workflow execution options.
 
 `workflow.onError`
 : Specify a closure that will be invoked if a workflow run is terminated. See {ref}`workflow-handlers` for more information.
+
+`workflow.output.contentType`
+: *Currently only supported for S3.*
+: Specify the media type a.k.a. [MIME type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_Types) of published files (default: `false`). Can be a string (e.g. `'text/html'`), or `true` to infer the content type from the file extension.
```
christopher-hakkaart, line 1626:
```suggestion
: Specify the media type, also known as [MIME type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_Types), of published files (default: `false`). Can be a string (e.g. `'text/html'`), or `true` to infer the content type from the file extension.
```

christopher-hakkaart (commented):
Added small language suggestion.

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
```diff
@@ -222,7 +222,7 @@ class PublishDir {
     protected void apply0(Set<Path> files) {
         assert path
 
-        final retryOpts = session.config.navigate('nextflow.publish.retryPolicy') as Map ?: Collections.emptyMap()
+        final retryOpts = session.config.navigate('workflow.output.retryPolicy') as Map ?: Collections.emptyMap()
```
pditommaso, line 225: What's if there's an usage of this setting using `nextflow.publish.retryPolicy`?

modules/nextflow/src/main/groovy/nextflow/script/OutputDsl.groovy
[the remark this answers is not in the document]
pditommaso: ok

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
bentsherman: We can keep the old syntax if you want, but I don't think we need to. I don't think anyone has been using these settings

pditommaso: the problem is that `nextflow.publish.retryPolicy` than it would be better to keep it as it as even if I agree may not be used.

bentsherman: something like this?

```groovy
        final retryOpts = session.config.navigate('workflow.output.retryPolicy') as Map ?: session.config.navigate('workflow.publish.retryPolicy') as Map ?: Collections.emptyMap()
```

pditommaso: I guess the problem is that the component is not aware if it's related to `output` or `publish` scope. What about passing the type (it could be just a string) when creating that instance ?

pditommaso: ok, agreed. let's use `workflow.output`. let's mention in the docs `workflow.publish.retryPolicy` has been replaced by `workflow.output.retryPolicy`

pditommaso: and reporting a warning if `workflow.publish.retryPolicy` is used in place of `workflow.output.retryPolicy`

bentsherman: done

docs/workflow.md
```diff
@@ -516,167 +486,208 @@ results/
...
+
+See [Reference](#reference) for the list of available index directives.
+
+### Migrating from first preview
+
+The first preview of workflow publishing was introduced in 24.04. The second preview, introduced in 24.10, made the following breaking changes:
 
-`header`
-: When `true`, the keys of the first record are used as the column names (default: `false`). Can also be a list of column names.
+- The process `publish:` section has been removed. Channels should be published only in workflows, ideally the entry workflow.
```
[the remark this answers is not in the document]
bentsherman: It's too confusing. If a process publishes some outputs by default and you don't want to publish them, then you have to explicitly disable them. And one of the goals of workflow outputs was to have all published outputs in one place, instead of being scattered across the pipeline

My initial feeling was to only allow publishing in the entry workflow, but that would also create a lot of boilerplate to pass all the channels up from the subworkflows. So a happy middle ground for now is to allow publishing from any workflow, but not processes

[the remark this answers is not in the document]
bentsherman: Also, the original motivation for process publishing is to provide some sensible publishing defaults for a module. I think this is better handled by providing an entry workflow with the module:
```groovy
process ASPERA_CLI {
    input:
    // ...

    output:
    tuple val(meta), path("*fastq.gz"), emit: fastq
    tuple val(meta), path("*md5")     , emit: md5

    script:
    // ...
}

workflow {
    main:
    ASPERA_CLI ( params.input, params.args )

    publish:
    ASPERA_CLI.out.fastq >> 'fastq'
    ASPERA_CLI.out.md5 >> 'md5'
}
```

This gives an example of how to use the module, both params and publishing, but they are opt-in. The user can incorporate these publishing rules into their pipeline explicitly, if they want to.

pditommaso:
@bentsherman can you please have a look at the conflicts

Issue #5187, opened 2024-07-31
Join behaviour non-deterministic


robsyme:
## Bug report 

When the `join` operator is joining channels where the join key is a Map, the comparison can fail in a nondeterministic manner.

Inside the join operator, we `get` the key from the `buffer` object here:

https://github.com/nextflow-io/nextflow/blob/cc6ec3142de0131b7dbeb8fed38b0e6506e86bdc/modules/nextflow/src/main/groovy/nextflow/extension/JoinOp.groovy#L180

This `get` operation uses `.equals()` to compare the argument to the keys in the `buffer` Map. The Nextflow script below produces keys where:
- `self.equals(other) = false` _and_
- `self == other      = true`

Because the Map `get` method uses `.equals()` to compare objects, the join operator fails to join the elements correctly in some cases.

### Expected behavior and actual behavior

The example workflow below correctly outputs two elements in the `JOINED:` channel only about half the time. You may need to run the workflow multiple times to see the nondeterminism.

```nextflow
#!/usr/bin/env nextflow

workflow {
    Channel.of(
        [[id:"id1"], "value1", "item1"],
        [[id:"id1"], "value2", "item2"],
        [[id:"id2"], "value3", "item3"]
    )
    | groupTuple
    | map { meta, channel_value, list ->
            newMeta = meta + [list: list]
            [newMeta, channel_value.flatten()]
        }
    | view { it -> 
        """
        CH1: ${it} 
            key = ${it[0]}
            hashCode == ${it[0].hashCode()}) 
        """.stripIndent()
    }
    | set { ch1 }

    Channel.of(
        [[id:"id1"], "another1", "item1"],
        [[id:"id1"], "another2", "item2"],
        [[id:"id2"], "another3", "item3"]
    )
    | groupTuple
    | map { meta, channel_value, list ->
            newMeta = meta + [list: list]
            [newMeta, channel_value.flatten()]
        }
    | view { it -> 
        """
        CH2: ${it} 
            key = ${it[0]}
            hashCode == ${it[0].hashCode()}) 
        """.stripIndent()
    }
    | set { ch2 }

    ch1
    | join(ch2)
    | view { result -> "JOINED: ${result}" }
}
```

I would expect that every time I run the workflow, the final "joined" channel would always return two items:

```
JOINED: [[id:id1, list:[item1, item2]], [value1, value2], [another1, another2]]
JOINED: [[id:id2, list:[item3]], [value3], [another3]]
```

[...]

PR #5189, opened 2024-07-31
Explicit delegation to target class for ArrayBag

robsyme:
Solves #5187


modules/nextflow/src/main/groovy/nextflow/util/ArrayBag.groovy
```diff
@@ -63,6 +63,16 @@ class ArrayBag<E> implements Bag<E>, List<E>, KryoSerializable {
...
 
+   @Override
+   int hashCode() {
+       target.hashCode()
+   }
+
+   @Override
+   boolean equals(Object o) {
+       target.equals(o)
+   }
```
bentsherman, line 74: There's a reason why these methods were commented out, related to some Groovy bug, but now I can't recall what it was

pditommaso:
@robsyme is this PR solving the issue for you?

modules/nextflow/src/main/groovy/nextflow/util/ArrayBag.groovy
robsyme: I suspected that might have been the case (reasonable looking code that is commented out is always a bit ominous).
The changes don't break any tests, if that gives you any confidence.

robsyme:
Yup, the changes solve both the very minimal example in the new test included in this PR and also the slightly more complicated example in #5187

modules/nextflow/src/main/groovy/nextflow/util/ArrayBag.groovy
pditommaso: Still not sure to understand \*how* it solves the problem

robsyme: Yeah, me neither. I'll carve out some time from customer things this week to try and dig into the *why*.

bentsherman: This is the Groovy issue I was thinking of: https://github.com/nextflow-io/nextflow/issues/5187#issuecomment-2262906488

@pditommaso maybe it would be good to convert ArrayBag to Java, so that we don't have to deal with this weird Groovy behavior

pditommaso: it's not that weird, the semantic of equals and hashCode in groovy is based on the collection content, not the object id, as it's in Java. 

It should be enough to implement the same logic as it's has been done [here](https://github.com/nextflow-io/nextflow/blob/71f6ed0bcda3462c415a23fb92c14f020ac55bce/modules/nf-commons/src/main/nextflow/util/HashBuilder.java#L159-L159), to make it work. 


@robsyme can you please provide a unit test to validate your solution?

pditommaso: I mean a unit test for JoinOp, not the ArrayBag class 

https://github.com/nextflow-io/nextflow/blob/5c9e305d8a506c980d2fb086b6c90f482ecd12fb/modules/nextflow/src/test/groovy/nextflow/extension/JoinOpTest.groovy#L29-L29

robsyme: Added the JoinOpTest unit test, thanks Paolo.

robsyme:
If the point of the ArrayBag class is to provide a container for objects that order-invariant, it makes sense to ensure that hashCode and equivalence methods also reflect this property. The commented-out equals method only tests for length equivalence and a containsAll test, which would fail when testing ArrayBags with the same items with differing frequencies, e.g. `[1,1,2]` and `[1,2,2]`. I added a slighty more computationally costly equals method, but it will not fail for these cases. I also added a JoinOp unit test for this case.

pditommaso:
This opens an interesting point: should `[1,1,2]` be the same of `[1,2,2]` ?

pditommaso:
Also it results into an inconsistency that `hashCode` is honoured, but it does not `equals` and `==` operator. If you check this tests will fail 

```
     given:
        def alpha = new ArrayBag(['abc',123,'x', 9])
        def delta = new ArrayBag([123,9,'abc','x'])

        expect:
        alpha == delta
        alpha.equals(delta)
```

bentsherman:
ArrayBag itself seems over-complicated and contradictory. It's basically trying to be a `List` in some places and a `Set` in others. Instead I would argue that you should just use a `List` or a `Set` depending on what you want.

Here are all the places where ArrayBag is used:
- `collect` operator: produce a List if sorting is enabled, otherwise produce a Set
- `groupTuple` operator: same as above
- process glob input: it is an unordered collection, use a `Set`

The two operators should have return type `Collection`, so that in general order is not guaranteed in general, but can be depending on runtime conditions. Or even better, separate the sort functionality from `collect` and `groupTuple` so that they always return `Set`, then the sorting can be done is a separate operation.

Glob patterns in general are unordered, so it doesn't make sense to give the illusion of ordering in this case. What does `files('*.fastq')[0]` mean? It has no meaning

If we do all of that, we can remove ArrayBag and avoid all of this mess

bentsherman:
Example of sorting groups explicitly after a groupTuple ([source](https://github.com/bentsherman/nf-boost/blob/4702ca3f695718c167918023ceb8eb24a31a020b/examples/merge-text.nf#L75-L81)): 

```groovy
// ...
    | groupTuple
    | map { group, items ->
      def sorted = items
        .sort { item -> item[0].id }
        .collect { meta, csv -> csv }
      return [group, sorted]
    }
```

robsyme:
Thanks for doing that analysis Ben, I agree that replacing the ArrayBag seems very sensible.

pditommaso:
The main reason for ArrayBag is to not invalidate the cache when files are emitted in different order 

https://github.com/nextflow-io/nextflow/blob/ensure-arraybag-equivalence/modules/nf-commons/src/main/nextflow/util/HashBuilder.java#L158-L159

What is tricky is that `hashCode` and `equals` is honoured when invoked by Java SDK (and therefore fixes the `JoinOp`), and it is not when invoked by Groovy (and pipeline code). 

Given that adding `hashCode` and `equals` methods at least fixes the join implementation, I think it's fine to add them with a big notice. 

I would avoid however the check for differing frequencies tho. 


Do you agree?

bentsherman:
> The main reason for ArrayBag is to not invalidate the cache when files are emitted in different order

my point is that this is already supported for Set, so ArrayBag should just be replaced with Set. if the order of a collection doesn't matter for purposes of caching, it shouldn't matter for the user either.

bentsherman:
If the order doesn't matter... then they shouldn't be relying on it. Doing something like `foo[0]` on an ArrayBag is non-deterministic. Tbh it might explain why people have mysterious caching issues from time to time

That being said, I would be willing to defer this change to the introduction of static type checking. We'll have to clean up APIs like this anyway and then we can have a feature flag like `nextflow.enable.types` to make it opt-in

pditommaso:
Sound like a reasonable plan

bentsherman:
I forgot that `Set` requires uniqueness, which makes it the wrong abstraction for `collect` and `groupTuple` (should be fine for file globs though).

The `Bag` interface is actually the right abstraction for collecting values from a channel, as it is a generic collection, but serves as a marker for the hash builder to hash it like a set.

The problem is that ArrayBag implements List, which gives a false sense that it is ordered. If a user starts doing indexed access on an ArrayBag, they will be relying on non-deterministic behavior that could cause cryptic resume issues -- even though the ArrayBag itself is deterministic, their *usage* of the ArrayBag could lead to non-deterministic hashes downstream.

You mentioned frequencies as a way to test equality and that it would be expensive, but... what if we just implement Bag as a frequency map instead of an array list? Call it a `HashBag`. The elements already need to be hashable so it shouldn't break anything. Then the equality test is the same as for a map.

It looks like Guava even has this, it's called a [Multiset](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/Multiset.html):

> A collection that supports order-independent equality, like Set, but may have duplicate elements. A multiset is also sometimes called a bag.

bentsherman:
In any case, I think the migration plan is the same, since replacing ArrayBag with Multiset would drop indexing, better to change it as part of the static type API.

PR #5195, opened 2024-08-01
Allow local cloudcache path

bentsherman:
Allow cloudcache to use a local path so that it works with Fusion V1


bentsherman:
I remember we were planning to merge cloudcache into the main codebase. That might affect how we handle this change.

The cloud cache store is configured to always override the default store, so the only way to control which cache store is used is to enable/disable the cloudcache plugin.

So if we get rid of the plugin, Nextflow needs to decide which cache store to use. Could be an explicit environment var like `NXF_ENABLE_CLOUDCACHE`, or check whether `NXF_CACHE_PATH` is a local path vs remote path.

bentsherman:
Closing, we will address this issue on the platform side

PR #5198, opened 2024-08-02
Update nextflow install path

christopher-hakkaart:
- Update nextflow install path
- Update NXF_VER example


bentsherman (approved):

docs/install.md
```diff
@@ -60,12 +60,16 @@ Nextflow is distributed as a self-installing package, in order to make the insta
     chmod +x nextflow
     ```
 
-3. Move Nextflow into an executable path:
+3. Move Nextflow into an executable path. For example:
 
     ```bash
-    sudo mv nextflow /usr/local/bin
+    mv nextflow $HOME/.local/bin/
```
pditommaso, line 66: Not sure about this, because `$HOME/.local/bin` is not a standard path. Mac does not have it (not sure across different linux distribution).

christopher-hakkaart: Good point - it would be nice to have a suggested path rather than saying "Install on your path". Do you have any suggestions?

pditommaso: maybe it can be fixed adding `mkdir -p $HOME/.local/bin/` before `mv` 

and changing the tip something like:

> make sure the directory `$HOME/.local/bin/` is in your `PATH` variable, otherwise .. (how to do it). Alternatively you can use any other directory already in your path. Be aware nextflow will try to override this file  when upgrading it, etc ..

bentsherman (approved):

docs/install.md
```diff
@@ -60,12 +60,21 @@ Nextflow is distributed as a self-installing package, in order to make the insta
...
+    mkdir -p $HOME/.local/bin/
+    mv nextflow $HOME/.local/bin/
     ```
 
+    :::{tip}
+    Ensure the directory `$HOME/.local/bin/` is included in your `PATH` variable. Temporarily add this directory to `PATH` by setting `export PATH="$PATH:$HOME/.local/bin"`. Add the directory to `PATH` permanently by adding the export command to your shell configuration file, such as `~/.bashrc` or `~/.zshrc`. Alternatively, move the `nextflow` executable to a directory already in your `PATH`.
+    :::
+
+    :::{warning}
+    Nextflow will update its executable during the self update process and it should not be placed in a directory with restricted permissions.
```
pditommaso, line 75:
```suggestion
    Nextflow will update its executable during the self update process, therefore the update can fail if the executable is placed in a directory with restricted permissions.
```

People may want to keep in path with restricted permission anyway

PR #5201, opened 2024-08-02
Convert Duration and MemoryUnit to Java

bentsherman:
Exploring the possibility of converting some lower-level classes to Java, to move towards native compilation. Duration and MemoryUnit will be used by the language server to do type checking.


modules/nf-commons/src/main/nextflow/util/Duration.java
```diff
@@ -0,0 +1,367 @@
...
+import org.codehaus.groovy.runtime.StringGroovyMethods;
+import org.apache.commons.lang.time.DurationFormatUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+/**
+ * A simple time duration representation
+ *
+ * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
+ */
+@EqualsAndHashCode(includes = "durationInMillis")
```
pditommaso, line 37: Does groovy annotation works with java classes ?

bentsherman: Nope, I will implement them directly

bentsherman:
There is one minor sticking point, which is the Duration `toString()`. Durations between 1s and 60s are rendered with one decimal place, e.g `1.5s`. However, Groovy has some extra magic -- it will print `3.0` as `3` whereas Java will print `3.0`. I don't know if this was an actual requirement so I just updated the failing tests to include the `.0`.

It's a minor change but also this PR is pretty low-stakes, so you can take it or leave it

bentsherman:
Closing in favor of implementing Duration / MemoryUnit interfaces from a shared library

Issue #5188, opened 2024-07-31
`nextflow config` unexpectedly writes logging to stdout when not using tty


SamStudio8:
### Expected behavior and actual behavior

**Expected**
`nextflow config` should yield a valid nextflow configuration and not care about my tty.

**Actual**
`nextflow config` can write debug information to stdout, preventing the resulting configuration from being read by nextflow.

### Steps to reproduce the problem

```
$ echo "params.hoot = 0" > nextflow.config
$ export COLUMNS="bad"
$ NXF_VER=24.04.3 nextflow config > my.config
$ cat my.config
17:18:14.153 [main] DEBUG nextflow.cli.Launcher -- Unexpected terminal $COLUMNS value: bad
params {
   hoot = 0
}
```

### Environment 

* Nextflow version: 24.04.3
* Java version: 17.0.3
* Operating system: macOS
* Bash version: 3.2.57

### Additional context

Generally I'd expect information like this to be written to stderr. The offending log line is here https://github.com/nextflow-io/nextflow/blob/cc6ec3142de0131b7dbeb8fed38b0e6506e86bdc/modules/nextflow/src/main/groovy/nextflow/cli/Launcher.groovy#L186.

bentsherman:
Likely because the error happens before the logger has been set up. We can either replace with `System.err.println` or find a way to make the logger fallback to stderr instead of stdout.

Eventually I'd like to refactor the CLI with picocli which would remove this code entirely

pditommaso:
In principle it should be possible to add a logback config file to provide the default log setting and turn off debug level in the console

pditommaso:
The PR above solves the problem, tho I don't like: 
* the use of xml config likely adds a little overhead in the nextflow launch
* any debug logging during the bootstrap is lost

bentsherman:
Is there some way to redirect `log.debug` to stderr instead of stdout during bootstrap?

pditommaso:
Maybe, dont' remember

PR #5206, opened 2024-08-06
Fix bootstrap log level

pditommaso:
This PR prevent the output of debug line to the console during nextflwo bootstrap i.e. before the CLI args are parsed and the custom logging config has been applied. 

Solve #5188


pditommaso:
Ok, this should solve https://github.com/nextflow-io/nextflow/issues/5188

bentsherman (approved):
Looks good as long as you confirmed it fixed the reported issue

PR #5227, opened 2024-08-14
Recommend using Nextflow secrets for sensitive credentials in configs

ewels:
Put together after pointing someone in the community towards Nextflow Secrets.

Initially was just going to do a couple of tip boxes, but figured that if secrets are the recommended method then maybe the code examples should be using them as well.

Happy to scale this back if the feeling is that it's overkill. On the other hand, also happy to be told if I missed a spot.

@netlify /docs/latest/amazons3.html#aws-access-and-secret-keys


ewels:
@christopher-hakkaart this might be a nice small one for you to review / wrap into other docs changes 😊

christopher-hakkaart:
PR to wrap into other docs changes here -> #5325

pditommaso:
I believe we should keep config examples simpler, even it can make sense provide to not hard code keys in the config file. Secrets can be a solution, but even better to rely on instance profile or env variables.

ewels:
@christopher-hakkaart can this PR be closed now?

PR #5229, opened 2024-08-15
Draft proposal for supporting Flux (different to flux mini)

ewels:
Porting a set of changes from an email thread for discussion.

Context is that a user was trying to use Nextflow with Flux. The executor is written for _flux mini_ and didn't work for him.

After some back and forth and investigation, he found the following set of changes got the executor to work on their system. This led to the creation of the following docs page for the system users: https://doku.lrz.de/nextflow-on-hpc-systems-test-operation-788693597.html

This was all only in an email thread, so moving here so that it doesn't get lost. I'm curious whether the flux executor support could be tweaked to work with both flux core and flux mini? (maybe with some configuration?). The changes seem pretty minor.

'cc @pditommaso and I think the original implementation was by @vsoch

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
@@ -61,9 +61,10 @@ class FluxExecutor extends AbstractGridExecutor {
     @Override
     List<String> getSubmitCommandLine(TaskRun task, Path scriptFile ) {
 
-        List<String> result = ['flux', 'mini', 'submit']
+        List<String> result = ['flux', 'submit']
         result << '--setattr=cwd=' + quote(task.workDir)
         result << '--job-name="' + getJobNameFor(task) + '"'
+        result << '--flags=waitable'
 
         // Only write output to file if user doesn't want written entirely to terminal
         Boolean terminalOutput = session.config.navigate('flux.terminalOutput') as Boolean
@@ -77,7 +78,8 @@ class FluxExecutor extends AbstractGridExecutor {
 
         // Time limit in minutes when no units provided
         if( task.config.getTime() ) {
-            result << '--time-limit=' + task.config.getTime().format('mm')
+            log.debug "Custom task wallclock time request is not currently supported here."
+//            result << '--time-limit=' + task.config.getTime().format('mm')
         }
 
         // Flux does not support asking for specific memory
@@ -87,7 +89,8 @@ class FluxExecutor extends AbstractGridExecutor {
 
         // the requested partition (a.k.a queue) name
         if( task.config.queue ) {
-            result << '--queue=' + (task.config.queue.toString())
+            log.debug "Queue parameter not supported by standalone flux."
+//            result << '--queue=' + (task.config.queue.toString())
         }
 
         // Any extra cluster options the user wants!


pditommaso:
Who is going to maintain this?

vsoch:
This looks OK - the only change with flux is we deprecated mini, so it's just flux submit.

ewels:
Thanks @vsoch! Are you happy / do you understand the logic behind the other changes to the flags? (I haven't read up on them yet so don't).

ewels:
Also do you mean that flux / flux mini are not two different systems? So we can just make this change and it's valid for all people using flux?

We might need some docs updates here as well.

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
```diff
@@ -61,9 +61,10 @@ class FluxExecutor extends AbstractGridExecutor {
     @Override
     List<String> getSubmitCommandLine(TaskRun task, Path scriptFile ) {
 
-        List<String> result = ['flux', 'mini', 'submit']
+        List<String> result = ['flux', 'submit']
         result << '--setattr=cwd=' + quote(task.workDir)
         result << '--job-name="' + getJobNameFor(task) + '"'
+        result << '--flags=waitable'
```
vsoch, line 67: This would require you to have `flux wait` somewhere, so I need more detail about why you are adding it here. The `flux submit` is the same as `flux mini submit` (I removed the mini early on, because it wasn't needed).

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
```diff
@@ -77,7 +78,8 @@ class FluxExecutor extends AbstractGridExecutor {
 
         // Time limit in minutes when no units provided
         if( task.config.getTime() ) {
-            result << '--time-limit=' + task.config.getTime().format('mm')
```
vsoch, line 80: Why was this removed? This should still work.

```
  -t, --time-limit=MIN|FSD    Time limit in minutes when no units provided, otherwise in Flux standard duration, e.g. 30s, 2d, 1.5h   
```

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
```diff
@@ -87,7 +89,8 @@ class FluxExecutor extends AbstractGridExecutor {
 
         // the requested partition (a.k.a queue) name
         if( task.config.queue ) {
-            result << '--queue=' + (task.config.queue.toString())
+            log.debug "Queue parameter not supported by standalone flux."
```
vsoch, line 92: This is also still available:

```
 -q, --queue=NAME            Submit a job to a specific named queue
```

vsoch:
And this title is not correct:

> Draft proposal for supporting Flux (different to flux mini)

They are exactly the same.

pditommaso:
@ewels not sure what to do with this PR

vsoch:
@ewels could you please address my questions/comments so we can move the PR forward?

ewels:
@vsoch because I just pasted the code diff from an email, I don't know the answers sorry. I'll follow up with him again and try to find out.

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
ewels: Quoting original author:

---

`--flags=waitable` was not necessary (I thought that I had explained that already somewhere ... if not, I'm sorry). Background was that I didn't know whether nextflow itself is doing some blocking here anyhow ... Otherwise, with "flux sbatch" flux returns usually immediately to the shell prompt, and letting flux do its job in the background.

Initially, I feared that if I omit that waitable flag, nextflow would submit the task dependency tree of tasks to flux, and return. In that case, the slurm job had just finished, and the flux instance been killed by the slurm daemon. I was happy to learn that this fear was unsubstantiated.

---

So I think that we can remove this 👍🏻 

```suggestion
```

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
ewels: Quoting the original author:

---

`--time-limit` doesn't apply here either. In theory it could be infinite. But our nextflow users have mostly anyway no clue about how long the single flux tasks (nextflow tasks) run. And the hard limit is the Slurm job's wall clock time ... So, I removed that in order to reduce the complexity for the users.

---

I think that this is a misunderstanding by him on how this works, given that the end user should never need to set this manually - it's taken from the pipeline config. So let's leave this in place 👍🏻

modules/nextflow/src/main/groovy/nextflow/executor/FluxExecutor.groovy
ewels: Reply from the original author:

---

Uhm ... `--queue` doesn't fit here anyway because there is no queue (names) when you run "flux start" manually. So, I didn't know what to do with it.

(To explain: As I tried to outline previously once ... I start a flux instance inside a Slurm job ... as user. For jobfarming. So there is no centrally managed flux instance with different queues etc. Just flux-in-slurm ... )

---

I'm not quite sure what's going on here. But either way, I think that they can just omit `queue` in the Nextflow config and this won't be used. So no need to delete from the executor source 👍🏻

ewels:
ok after feedback from the original author, this PR gets much simpler - now just a single line PR. I've also removed the `mini` from the docs.

Is this good to merge now @vsoch? Thanks for the review, and apologies for the slightly odd nature of this PR.

vsoch:
This now LGTM! :+1:

docs/executor.md
```diff
@@ -95,7 +95,7 @@ Resource requests and other job characteristics can be controlled via the follow
 
 The `flux` executor allows you to run your pipeline script using the [Flux Framework](https://flux-framework.org).
 
-Nextflow manages each process as a separate job that is submitted to the cluster by using the `flux mini submit` command.
+Nextflow manages each process as a separate job that is submitted to the cluster by using the `flux submit` command.
```
christopher-hakkaart, line 98:
```suggestion
Nextflow submits each process to the cluster as a separate job using the `flux submit` command.
```

pditommaso:
Tests green, there are some unresolved comments

ewels:
Comments all resolved 👌🏻

Issue #5125, opened 2024-07-09
Support scratchless execution with Google Cloud Batch


siddharthab:
### Expected behavior and actual behavior

Because Nextflow uses Cloud Storage Volumes by default in Cloud Batch, one could assume that scratch is not needed anymore because Cloud Storage Volumes will take care of staging things in scratch space and then moving to Cloud Storage. However, when I try to set `process.scratch = false`, all my processes fail with messages like: `/bin/bash: /mnt/disks/[workdir-bucket]/[workdir-prefix]/[task-id]/.command.sh: Too many levels of symbolic links`

### Steps to reproduce the problem

nextflow.config:
```
process.executor = 'google-batch'
process.scratch = false
process.container = 'ubuntu'
docker.enabled = true

google {
  ... project name and location ...
}
```

main.nf (same as in [tutorial](https://www.nextflow.io/docs/latest/your-first-script.html)):
```
params.str = 'Hello world!'

process splitLetters {
    output:
    path 'chunk_*'

    """
    printf '${params.str}' | split -b 6 - chunk_
    """
}

process convertToUpper {
    input:
    path x

    output:
    stdout

    """
    cat $x | tr '[a-z]' '[A-Z]'
    """
}

workflow {
    splitLetters | flatten | convertToUpper | view { it.trim() }
}
```

Run with:
```
nextflow run main.nf -w gs://[workdir-bucket]/[workdir-prefix]
```

### Program output 

```
executor >  google-batch (1)
[30/7bdc02] process > splitLetters   [100%] 1 of 1, failed: 1 ✘
[-        ] process > convertToUpper -
ERROR ~ Error executing process > 'splitLetters'

Caused by:
  Process `splitLetters` terminated with an error exit status (1)


Command executed:

  printf 'Hello world!' | split -b 6 - chunk_

Command exit status:
  1

Command output:
  /bin/bash: /mnt/disks/[workdir-bucket]/[workdir-prefix]/[task-id]/.command.sh: Too many levels of symbolic links

Command error:
  /bin/bash: /mnt/disks/[workdir-bucket]/[workdir-prefix]/[task-id]/.command.sh: Too many levels of symbolic links

Work dir:
  gs://[workdir-bucket]/[workdir-prefix]/[task-id]

Tip: when you have fixed the problem you can continue the execution adding the option `-resume` to the run command line

 -- Check '.nextflow.log' file for details
```

### Environment 

* Nextflow version: 24.04.2 build 5914
[...]

PR #5256, opened 2024-08-23
Support scratchless execution by Google Batch

siddharthab:
Addresses the observation in #5125.

This change comes with its own caveats. So please feel free to reject. But if we are rejecting, we should at least give a proper error message to users who set `process.scratch = false`.


pditommaso:
Closing this as not planned. Not sure what you mean with "we should at least give a proper error message to users who set"

PR #5260, opened 2024-08-27
fix: missing .command.env when eval is used and task runs on a cloud env

thanhleviet:
The file `.command.env` is missing in `nxf_unstage` when `eval` is used in the process on a cloud env (tested on AWS Batch)


pditommaso:
Well spotted! merged manually adding some tests https://github.com/nextflow-io/nextflow/commit/4a6b54aa12a9f5e9bdb83afe26e233440c4bc383

Thanks!

PR #5266, opened 2024-08-29
Change GitHub clone URL implementation

nathanthorpe:
We've come across an issue where the 60 requests per hour rate limit causes the workflow to error out. It stems from having to lookup the Clone URL via the API. 

I propose that the Clone URL be hardcoded so that we don't need to call the GitHub API. 

Alternatively, I could put it under try / catch for the RateLimitExceeded exception

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/scm/GithubRepositoryProvider.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/GithubRepositoryProvider.groovy
@@ -76,13 +76,7 @@ class GithubRepositoryProvider extends RepositoryProvider {
     /** {@inheritDoc} */
     @Override
     String getCloneUrl() {
-        Map response = invokeAndParseResponse( getEndpointUrl() )
-
-        def result = response.get('clone_url')
-        if( !result )
-            throw new IllegalStateException("Missing clone URL for: $project")
-
-        return result
+        return "${config.server}/${project}.git"
     }
 
     @Override


modules/nextflow/src/main/groovy/nextflow/scm/GithubRepositoryProvider.groovy
```diff
@@ -76,13 +76,7 @@ class GithubRepositoryProvider extends RepositoryProvider {
...
     @Override
     String getCloneUrl() {
-        Map response = invokeAndParseResponse( getEndpointUrl() )
-
-        def result = response.get('clone_url')
-        if( !result )
-            throw new IllegalStateException("Missing clone URL for: $project")
-
-        return result
+        return "${config.server}/${project}.git"
```
pditommaso, line 79: Considering the method `invokeAndParseResponse` is annotated with `@memoized` it should be invoked only once. 

https://github.com/nextflow-io/nextflow/blob/4a6b54aa12a9f5e9bdb83afe26e233440c4bc383/modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy#L279-L280

Not sure why this is happening.

nathanthorpe: It only happens when there are many concurrent nextflow runs. 

It is being performed in AWS Batch, so it doesn't have the opportunity to use the cache.

pditommaso: Understand. However I believe this would solve much because in case other API calls will be made to access the config file, cloning the repo, etc. 

Is this not happening because the GitHub API token is not specified?

nathanthorpe: Ah right ok, didn't realize the config file was read that way.

Cloning the repo luckily doesn't use the rate quota.

I would rather not have users configure GitHub creds, but maybe I can find another solution for this.

PR #5268, opened 2024-08-30
In Google Batch, support user specified boot images

siddharthab:
The default boot image is batch-cos, but it may not be desired in all
situations. In particular, there is an ongoing issue in which the
batch-cos image does not retry pulling a docker image if there was a
network issue. The user may also have some some other configuration
pre-configured in their custom boot image.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -886,6 +886,11 @@ The following settings are available for Google Cloud Batch:
   :::
 : Define the set of allowed locations for VMs to be provisioned. See [Google documentation](https://cloud.google.com/batch/docs/reference/rest/v1/projects.locations.jobs#locationpolicy) for details (default: no restriction).
 
+`google.batch.bootDiskImage`
+: :::{versionadded} 24.08.0-edge
+  :::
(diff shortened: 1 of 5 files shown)


pditommaso (changes requested):
Thanks for this PR. It would be useful to add some unit tests to validate the changes

siddharthab:
> Thanks for this PR. It would be useful to add some unit tests to validate the changes

Added a very basic unit test; tested with `./gradlew ':plugins:nf-google:test'`. Please let me know if you would like more.

pditommaso (approved):
all fine. thanks for this contribution

bentsherman:
@siddharthab do you know if the boot image can be specified in an instance template? if so, we should add a warning that the boot image option you added will be overridden by the instance template if one is used:

https://github.com/nextflow-io/nextflow/blob/bf0cd326572eea34cb32dbd26f94ed7fc9faed6a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy#L299-L318

siddharthab:
> if so, we should add a warning

You are right. I missed the warnings section. Will send another PR.

Issue #4995, opened 2024-05-13
Improve stability of virtual threads at scale


bentsherman:
See https://nextflow.slack.com/archives/C02T98A23U7/p1715351870737659

When a workflow publishes many files using S3-to-S3 copy, virtual threads are needed to maximize request throughput, since each publish task is simply waiting on an HTTP request and doesn't need to occupy an OS thread the whole time.

However, enabling virtual threads currently creates two problems:

- The AWS Java SDK imposes a separate limit on the number of max HTTP connections (`aws.client.maxConnections`), and if this limit isn't also increased, Nextflow will likely fail with a message like `Timeout waiting for connection from pool`

- Virtual threads have no queue size limit by default, but in reality we might still want to limit the number of in-flight HTTP requests so as to not overload the network

To this end, I propose the following improvements:

- Allow virtual threads to be used with a queue size (similar to `executor.queueSize`) to control the number of concurrent publish tasks

- Automatically increase the AWS connection limit to match the virtual threads queue size so that the user doesn't have to remember it

fortune9:
Hi @bentsherman , I met the error "Timeout waiting for connection from pool" by enabling virtual threads, and I wonder any update you have on this issue. Thank you.

bentsherman:
The recommended way to limit concurrent access to a resource with virtual threads is to use a semaphore. I think we would need to add a semaphore for each shared resource (e.g. the S3 client) and make each API call acquire a semaphore slot. Then it should be safe to use virtual threads with no limit.

This blog post summarizes the caveats with virtual threads: https://blog.ycrash.io/pitfalls-to-avoid-when-switching-to-virtual-threads/

bentsherman:
@jorgee Notes from our discussions today

- Handle concurrency below the filesystem API (S3) and use virtual threads above it

- Keep the ability to disable virtual threads until we are confident that it works well across aws / azure / google, but don't worry about e.g. adding a semaphore to PublishDir

- Simplify aws configuration for s3 client: user should specify either a numeric limit (`maxConnections`) which controls all sub-clients in the s3 client.

  Might be useful to allow setting the target throughput based on a VM's allocated bandwidth, but still need a default value for `maxConnections` since target throughput only affects the transfer manager

- Keep the multipart upload options since they are used by the transfer manager, not just S3OutputStream

  Should be able to remove `multipartCopyObject()` from S3Client as it is not used anymore (uses transfer manager instead)

PR #5270, opened 2024-09-01
Disable automatic detection of virtual threads

pditommaso:
The use of Java virtual threads is supported as experimental feature. This PR disables the automatic use of virtual threads with Java 21 or later.


bentsherman (approved):
Good idea. See #4995 to a proposed solution to the problems reported around virtual threads.

I think the problem is not the virtual threads themselves but the lack of a pool size. They say "you don't need a pool anymore, just make a virtual thread!" but if every thread is doing heavy I/O (e.g. publishing) then you will still overwhelm the system.

So we need some kind of queue that allows for unlimited threads but limits the level of concurrent processing and I/O. Maybe that's just the existing platform thread pool... need to think on it more.

Issue #5273, opened 2024-09-02
Clarify that you must use `publish:` to produce output files with new syntax


adamrtalbot:
The new workflow outputs syntax just suggests you can use an output block to create outputs:
```
workflow {
    foo(bar())
}

output {
    directory 'results'
}
```

![image](https://github.com/user-attachments/assets/74472b74-e6c4-49bd-a174-76f5c6df7644)

However, this isn't the case and you must use the `publish:` directive in the process or workflow syntax to make it work:

```
workflow {
    foo(bar())

    publish:
    foo.out >> 'foo'
}

output {
    directory 'results'
}
```

We should clarify the documentation to say this.

ewels:
Will be fixed by #5185

PR #5274, opened 2024-09-02
[Docs] Workflow output: Include `publish` in example.

ewels:
Closes https://github.com/nextflow-io/nextflow/issues/5273

[ci-skip]

### DIFF
--- a/docs/workflow.md
+++ b/docs/workflow.md
@@ -409,6 +409,9 @@ A script may define the set of outputs that should be published by the implicit 
 ```groovy
 workflow {
     foo(bar())
+
+    publish:
+    foo.out >> 'foo'
 }
 
 output {


docs/workflow.md
```diff
@@ -409,6 +409,9 @@ A script may define the set of outputs that should be published by the implicit
 ```groovy
 workflow {
     foo(bar())
+
+    publish:
+    foo.out >> 'foo'
 }
```
adamrtalbot, line 415: Would it be better to remove fake details? This way draws your attention to the output block which is the focus of the statement.

```suggestion
workflow {
    // ...
}
```

ewels: Maybe, but having it there does implicitly tell anyone skim-reading just this one section that nothing is going to happen without `publish`, which  I quite like..

Issue #5214, opened 2024-08-09
Foreign file staging can allow pipelines to proceed with corrupt files


scwatts:
### Expected behavior and actual behavior

Description and background

- foreign file staging via HTTP/S can report successful transfer following connection interruption
- under certain conditions this allows pipelines to continue with incomplete and corrupt inputs
- this scenario has been observed intermittently for [nf-core/oncoanalyser](https://github.com/nf-core/oncoanalyser) when retrieving reference data from Cloudflare R2
  - most recently [reported by a user](https://nfcore.slack.com/archives/C04CQ1LK7U1/p1722527647438539) in the nf-core Slack Workspace
  - this issue has also been reported directly to me by collaborators
  - I've also personally encountered this error with regular but low frequency over the past 8 months

Desired behaviour

- minimally to emit an error message and halt pipeline execution, or
- ideally retry / resume download with required byte range to complete staging

### Steps to reproduce the problem

Given the difficultly involved in reproducing this intermittent problem I've put together toy example that can replicate behaviour
locally with small data. The simple example is not a perfect replication but fundamentally demonstrates the problem
described above.

<details><summary><b>File: <i>run_simple_http_server.py</i> (click to show)</b></summary>

```python
#!/usr/bin/env python3
import argparse
import http.server
import time


TAR_DATA = b'\x1f\x8b\x08\x00\xb7\xa3\xb5f\x00\x03\xed\xd1M\n\xc20\x10\x05\xe0\xac=ENP\'\xe9\xfc\x9c\xa7BB\x17\xc5H\x1b\xc1\xe3\x1bm)"\xb4\xbb(\xe2|\x9b\xb7\x19\x92I^L\xe9h*\x03\x00!\xb2\xcf\xe49\xc1\xe3\x9c\x0b\xeb\x90\x89Z\xef\x98\xc5\x82\xf3(\xceX\xaa\xbd\xd8\xc3u\xca\xddXV\x99r\xb8\xf4\xe1\xbc9W\xc6b\xdc9gy\xc7\x9a?"\x96\xfeO\xdd\xd8\xe4[\xaevG\xf9\x0fF\xdc\xeb\x9f\xdf\xfaoE\xbc\xb1Pm\xa3\x17\x7f\xde\x7f\x1f\x86!\x1d\xbe\xbd\x85RJ\xa9O\xbb\x03\xe7+\x86\x11\x00\n\x00\x00'


[...]

PR #5275, opened 2024-09-02
Fix http corrupted download

pditommaso:
This PR tries to address the issue with corrupted download reported by https://github.com/nextflow-io/nextflow/issues/5214. 

The overall approach consist on using an input stream filter that checks that the number of bytes read matches the expected one. 

However, I think this kind of check is already implemented by the http client, and therefore still persists.

### DIFF
--- a/modules/nf-httpfs/src/main/nextflow/file/http/FixedInputStream.groovy
+++ b/modules/nf-httpfs/src/main/nextflow/file/http/FixedInputStream.groovy
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2013-2024, Seqera Labs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


pditommaso:
Actually using the python server included in https://github.com/nextflow-io/nextflow/issues/5214 it looks working properly. 

@scwatts would you be able to test this branch will real data?

scwatts:
@pditommaso I've also confirmed this branch catches corrupt HTTP downloads as served by the toy example I gave. Testing this is in a real-world situation is tough since it relies on the intermittent failure of the Cloudflare R2 server.

I have oncoanalyser running in a loop configured only to stage and decompress the large VIRUSBreakend database with both Nextflow compiled from this branch and the v24.04.4 release. So far I have only captured one instance of such a server failure, which was in a run with this branch.

Your changes look to correct the behaviour: [oncoanalyser.nextflow_branch.stdstrm.failure_captured.txt](https://github.com/user-attachments/files/16845373/oncoanalyser.nextflow_branch.stdstrm.failure_captured.txt).

I'll post here to confirm the negative case under the same conditions where I am able observe that as well.

pditommaso:
Excellent!

scwatts:
After leaving the oncoanalyser loops to run overnight I've now replicated six fatal failures with v24.04.4 and an additional three successfully handled failures using this branch.

Thank you for this fix @pditommaso!

Issue #5081, opened 2024-06-20
Azure Fusion integration should not fail when managed identity auth is present


swampie:
## Usage scenario 

Now that [Azure Managed Identity](https://github.com/nextflow-io/nextflow/pull/4897) has been merged the support for fusion should also be enabled. The goal of this feature is to prevent failures when a managed identity is correctly configured and pass the configuration to fusion.

## Suggest implementation 

- one change should be the removal/extension of [this](https://github.com/nextflow-io/nextflow/blob/dbe6238f1a6c45fdcf69a7cb54cc0486a8870b98/plugins/nf-azure/src/main/nextflow/cloud/azure/fusion/AzFusionEnv.groovy#L43-L44) logic

swampie:
one possibility to prevent changes on fusion (if it relies on sas tokens) is to change the `getOrCreateSasToken()` to try to generate a token with the managed identity (unsure if this is possible)

alberto-miranda:
I'm currently writing a PR solving this in order to unblock https://github.com/seqeralabs/fusion/issues/526, but there's one thing I don't understand: why are we always generating a SAS token with `getOrCreateSasToken()` and returning it as a `result.AZURE_STORAGE_SAS_TOKEN` even if the `accountKey` authentication method has been configured? 

I was initially thinking about setting `result.AZURE_STORAGE_SAS_TOKEN` only for if the SAS token auth was requested, but I don't want to break any existing logic relying on this.

adamrtalbot:
> returning it as a result.AZURE_STORAGE_SAS_TOKEN even if the accountKey authentication method has been configured?

I _believe_ the SAS key is created by Nextflow, which is sent to the worker nodes as a temporary token for Fusion to use. This is because the accountKey isn't available to worker nodes (?). We can simply bypass this if we believe the worker nodes have an MI associated. 

Relates to https://github.com/nextflow-io/nextflow/issues/5232 which is the non-Fusion equivalent.

alberto-miranda:
> > returning it as a result.AZURE_STORAGE_SAS_TOKEN even if the accountKey authentication method has been configured?
> 
> 
> 
> I _believe_ the SAS key is created by Nextflow, which is sent to the worker nodes as a temporary token for Fusion to use. This is because the accountKey isn't available to worker nodes (?). We can simply bypass this if we believe the worker nodes have an MI associated. 
> 
> 
> 
[...]

PR #5278, opened 2024-09-03
Support Azure Managed Identities in Fusion configuration logic

alberto-miranda:
## Description
This PR extends the `nextflow.cloud.azure.fusion.AzFusionEnv` class so that it now takes into account a potential Azure Managed Identity in Nextflow's [configuration file](https://www.nextflow.io/docs/latest/azure.html#managed-identities). This also fixes a configuration error (#5081) by which Nextflow complained of a missing SAS token despite having a Managed Identity properly configured.


pditommaso (approved):
Great work!

PR #5293, opened 2024-09-10
Fix failOnIgnore causes task monitor to stop submitting tasks

### LINKED ISSUE #5291: Workflow option `failOnIgnore` causes workflow to hang

When using the `ignore` errorStrategy with the workflow option `failOnIgnore`, the pipeline hangs when there's a task failure.

workflow.nf
```
process process1 {
    input:
    val sample_id

    output:
    val sample_id, emit: sample_ids

    script:
    """
    if [[ $sample_id == "SAMP1" ]]; then
        exit 2
    fi
    ls -lah .*
    """
}

process process2 {
    input:
    val ready

    output:
    stdout

    script:
    """
    ls -lah .*
    """

}

workflow {
    input_channel = channel.of("SAMP1", "SAMP2", "SAMP3")
    process1(input_channel)
    process2(process1.out.sample_ids.collect())
}
```

Nextflow Config
```
workflow {
    failOnIgnore = true
}
process {
  errorStrategy = 'ignore'
}
```

I would expect that the workflow would complete with a non-zero exit status.

### Program output 

```Nextflow 24.08.0-edge is available - Please consider updating your version to it

 N E X T F L O W   ~  version 24.05.0-edge

Launching `/path/to/workflows/nextflow/hello_world/main.nf` [infallible_cajal] DSL2 - revision: fe2c285334

executor >  local (3)
executor >  local (3)
[09/916c3a] process1 (2) [100%] 3 of 3, failed: 1 ✔
[-        ] process2     [  0%] 0 of 1
[f2/61fcf1] NOTE: Process `process1 (1)` terminated with an error exit status (2) -- Error is ignored
```

And it hangs and doesn't finish.

### Environment 

* Nextflow version: [24.05.0-edge] 
* Java version: [openjdk version "21.0.2" 2024-01-16]
* Operating system: [macOS]
* Bash version: (use the command `$SHELL --version`) [zsh 5.9 (arm-apple-darwin22.1.0)]

bentsherman:
Close #5291 

I would prefer something that doesn't require referencing failOnIgnore directly in the task polling monitor, as they seem like separate concerns, but maybe I am just nit-picking.


pditommaso:
> I  would prefer something that doesn't require referencing failOnIgnore directly in the task polling monitor

Indeed. I've added a dedicated method with a more explicit semantic to keep logic more consistent.

PR #5294, opened 2024-09-11
Add Fusion version pinning info to Fusion config scope

llewellyn-sl:


docs/config.md
```diff
@@ -812,7 +812,7 @@ The following settings are available:
 : The maximum size of the local cache used by the Fusion client.
 
 `fusion.containerConfigUrl`
-: The URL from where the container layer provisioning the Fusion client is downloaded.
+: The URL from where the container layer provisioning the Fusion client is downloaded. Use this option to pin a specific Fusion version to use with Nextflow. For example, set `https://fusionfs.seqera.io/releases/v2.4.2-amd64.json` to use Fusion v2.4.2 for AMD64 CPU architectures.
```
pditommaso, line 815:
```suggestion
: The URL from where the container layer provisioning the Fusion client is downloaded. This option is useful to specify a development Fusion version for debugging purposes. 
```

The use of a non default Fusion version can lead to incompatibilities problem, as such it should not be advised as a practice to follow.

PR #5300, opened 2024-09-12
Update container handling with charliecloud

nschan:
Here are changes to the charliecloud driver.

- `charliecloud.writeFake` is true by default (!)

- if `charliecloud.writeFake` is true and `$CH_IMAGE_STORAGE` is set, images will go into `$CH_IMAGE_STORAGE` and be run from there with --write-fake (will be run by name in that case).

- if charliecloud.writeFake is true, but `$CH_IMAGE_STORAGE` is unset, images will go into `charliecloud.cacheDir`, `$NXF_CHARLIECLOUD_CACHE` or `work/charliecloud` (order of priority) and will be run by path (since run by name only works if `$CH_IMAGE_STORAGE `is set)

- if `charliecloud.writeFake` is set to `false`, images will need to be run with `-w` to enable mounting of the workDir. In this case `$CH_IMAGE_STORAGE` will not be used to store images, since `$CH_IMAGE_STORAGE` is incompatible with `-w`. Images will instead go into `charliecloud.cacheDir`, `$NXF_CHARLIECLOUD_CACHE` or `work/charliecloud` (order of priority). If the folder used to store images is the same as `$CH_IMAGE_STORAGE`, throw an exception. Images will be run by path.

In summary:

- Give preference to `$CH_IMAGE_STORAGE` in the default case of `writeFake`

- Make some effort to avoid using `$CH_IMAGE_STORAGE` when `writeFake` is `false`.

- No longer create temporary containers for each task. 

This is the same as #5212, but without the many merge conflicts created by a failed sign-off.

@phue I have updated the documentation. I have not added a unit test for switching between `image` and `imageName` in the builder, since I do not know of a way to set an env var for the tests. I have tested this locally (i.e. running tests with the var set or not) and it produced the expected results. Is it really worth adding a way to inject fake environments into the BuilderTest to test this specific behaviour?

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -491,6 +491,9 @@ The `charliecloud` scope controls how [Charliecloud](https://hpc.github.io/charl
 
 The following settings are available:
 
+`charliecloud.writeFake`
+: Enable `writeFake` with charliecloud (default) This allows to run containers from storage in writeable mode, using overlayfs. `writeFake` requires `overlayfs` (Linux kernel >= 5.11). For full suppornt tempfs with xattrs in the user namespace (Linux kernel >= 6.6) is required , see [charliecloud documentation](https://hpc.github.io/charliecloud/ch-run.html#ch-run-overlay) for details.
+
 `charliecloud.cacheDir`
 : The directory where remote Charliecloud images are stored. When using a computing cluster it must be a shared folder accessible to all compute nodes.
 
@@ -511,12 +514,6 @@ The following settings are available:
 
 `charliecloud.registry`
 : The registry from where images are pulled. It should be only used to specify a private registry server. It should NOT include the protocol prefix i.e. `http://`.
-
-`charliecloud.writeFake`
-: Enable `writeFake` with charliecloud (default) This allows to run containers from storage in writeable mode, using overlayfs. `writeFake` requires `overlayfs` (Linux kernel >= 5.11). For full suppornt tempfs with xattrs in the user namespace (Linux kernel >= 6.6) is required , see [charliecloud documentation](https://hpc.github.io/charliecloud/ch-run.html#ch-run-overlay) for details.
-
-`charliecloud.useSquash`
-: Create a temporary squashFS container image in the process work directory instead of a folder.
 
 Read the {ref}`container-charliecloud` page to learn more about how to use Charliecloud containers with Nextflow.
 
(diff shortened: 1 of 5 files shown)


docs/config.md
```diff
@@ -488,16 +488,18 @@ The following settings are available:
...
 The following settings are available:
 
-`charliecloud.cacheDir`
-: The directory where remote Charliecloud images are stored. When using a computing cluster it must be a shared folder accessible to all compute nodes.
-
 `charliecloud.enabled`
 : Enable Charliecloud execution (default: `false`).
 
+`charliecloud.writeFake`
+: Enable `writeFake` with charliecloud (default: `true`) This allows to run containers from storage in writeable mode, using overlayfs. `writeFake` requires `overlayfs` (Linux kernel >= 5.11). For full suppornt tempfs with xattrs in the user namespace (Linux kernel >= 6.6) is required , see [charliecloud documentation](https://hpc.github.io/charliecloud/ch-run.html#ch-run-overlay) for details.
```
phue, line 498:
```suggestion
: Enable `writeFake` with charliecloud (default: `true`) This allows to run containers from storage in writeable mode, using overlayfs. `writeFake` requires unprivileged `overlayfs` (Linux kernel >= 5.11). For full support, tempfs with xattrs in the user namespace (Linux kernel >= 6.6) is required, see [charliecloud documentation](https://hpc.github.io/charliecloud/ch-run.html#ch-run-overlay) for details.
```

nschan: I have pushed this in a new commit, I cant sign-off correctly via github.

nschan:
> > #3566 has been fixed a while ago
> 
> How was it fixed?

Trying to catch it and retry https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy#L186

nschan:
Thanks for the helpful discussion and review @phue ! @pditommaso do you have any additional comments on this?

modules/nextflow/src/main/groovy/nextflow/container/CharliecloudCache.groovy
```diff
@@ -130,19 +131,38 @@ class CharliecloudCache {
         if( config.pullTimeout )
             pullTimeout = config.pullTimeout as Duration
 
+        def writeFake = true as Boolean
```
pditommaso, line 134:
```suggestion
        def writeFake = true
```

pditommaso (commented):
Just lest a minor comment

nschan:
Thanks @pditommaso, I have put in the suggested change.

pditommaso:
Nice, let's wait for the tests

Issue #4034, opened 2023-06-19
Additional `manifest` config fields


ewels:
It would be great to have a few new `manifest` config variables to help describe pipelines:

```nextflow
manifest {
    docsURL = "https://mypipeline.com/docs"
    iconURL = "https://mypipeline.com/icon.png"
    icon = "assets/icon.png"
    organisation = "University of Phil"
    license = "MIT"
}
```

abhi18av:
Mmm, interesting suggestion Phil!

Perhaps, it should be kept open-ended like with a `manifest.meta` field so that other tools/plugins could possibly populate or read from that field?

P.S. `"University of Phil"` sounds nice! 😄

ewels:
I don't really understand how anything could (or should) populate it, other than the pipeline developers?

Having a field for an arbitrary map of keys could be nice to have a flexible use case 👍🏻  However, standardisation is nice. We could add that in addition to (some of?) the suggested keys above.

abhi18av:
I was just wondering that since there are some Git related fields within `manifest` like `recurseSubmodules`, `defaultBranch` then it might make sense to logically combine further options together through an open-ended map like 

```nextflow 

manifest {
    homePage = 'http://foo.com'
    description = 'Pipeline does this and that'

    .. existing fields...

    meta {
         documentationUrl = "https://doc.pipeline.com"
         license = "MIT" 
         researchGroup = "some specific research group"

         ...arbitrary fields ...
    }
}

```

Also the idea regarding open-ended metadata stems from https://github.com/nextflow-io/nextflow/issues/2723, which might result in a config which is interchangeable with the JSON format.

ewels:
Ah ok. Following that thread, it might make sense to move _all_ descriptive fields (including current) to a `meta` sub-map then?

```nextflow
manifest {
    defaultBranch
    recurseSubmodules
    mainScript
    name
    nextflowVersion
    version
    meta {
        author
        description
        doi
        homePage
        docsURL
        iconURL
        icon
        organisation
        license
    }
}
```

eg. depreciate `manifest.description` and `manifest.doi` etc, instead replacing with `manifest.meta.description`

ewels:
I think deprecating the old ones and leaving them there for a version or two is probably ok. Like you say, we hopefully have a new config coming one day anyway.. 🤷🏻‍♂️

bentsherman:
I recommend adding the following fields based on my experience with the prov formats:
- license
[...]

PR #5314, opened 2024-09-19
Include additional fields to manifest

jorgee:
Include new fields in the manifest and update the documentation with the new fields

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1246,7 +1246,7 @@ The following settings are available:
 `manifest.homePage`
 : Project home page URL.
 
-`manifest.iconUrl`
+`manifest.icon`
 : Project related icon location (Relative path or URL).
 
 `manifest.license`
--- a/docs/config.md
+++ b/docs/config.md
@@ -1250,7 +1250,7 @@ The following settings are available:
 : Project related icon location (Relative path or URL).
 
 `manifest.license`
-: Workflow license.
+: Project license.
 
 `manifest.mainScript`
 : Project main script (default: `main.nf`).
--- a/docs/config.md
+++ b/docs/config.md
@@ -1237,14 +1237,26 @@ The following settings are available:
 `manifest.description`
 : Free text describing the workflow project.
 
+`manifest.docsUrl`
+: Project documentation URL.
+
 `manifest.doi`
 : Project related publication DOI identifier.
 
 `manifest.homePage`
 : Project home page URL.
 
+`manifest.iconUrl`
+: Project related icon location (Relative path or URL).
+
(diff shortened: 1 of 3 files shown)


docs/config.md
```diff
@@ -1237,15 +1237,27 @@ The following settings are available:
...
 : Project related publication DOI identifier.
 
 `manifest.homePage`
 : Project home page URL.
 
+`manifest.iconUrl`
+: Project related icon location (Relative path or URL).
+
+`manifest.license`
+: Workflow license.
```
bentsherman, line 1253:
```suggestion
: Project license.
```

jorgee:
> Looks neat. Thin @ewels was suggesting to user `icon` instead of `iconUrl` in a recent comment

Yes, changed.

pditommaso:
Well done! 🎉

ewels:
Ooh awesome! Thanks @jorgee!

Looks like author / maintainer fields are just strings here. Could we optionally have these as maps as well with additional subfields such as ORCiD, affiliatiob and email? See https://github.com/nextflow-io/nextflow/issues/4034#issuecomment-1766364814

'cc @pditommaso @bentsherman

pditommaso:
Think we said it would be better to keep it flat

bentsherman:
The BCO supports a bunch of contributor metadata as a list of maps, so you can do something like name, email, ORCID

https://wiki.biocomputeobject.org/Provenance-domain#Contributors_%E2%80%9Ccontributors%E2%80%9D

I think we could support this structure directly in the config. The `manifest.author` could be either a string for comma-separated list, or a list of maps for the complete structure. I think name, affiliation, email, and ORCID should be enough, but always easy to add more fields later.

See also: https://github.com/nextflow-io/nf-prov/issues/10

bentsherman:
So the list of maps is mainly to provide a more complete BCO manifest

jorgee:
In fact, there is no check in the author, so you can put whatever valid expression (String, list, list of maps). 
```
manifest {
     author = [ [ name: 'First Author', email: 'author1@mail.com'], [ name: 'Second Author', email: 'author@mail.com']]
}
```
The above configuration should work. In the `nextflow info`, it will print the string serialization of the list of maps. 
I think we just need to validate the map fields.

ewels:
Ok so if it already works, shall we say that a short term solution is to just modify the documentation to say that it can be a string or a map? Sounds like that would be sufficient for now. We can standardise on nf-core and come back to this if needed.

ewels:
👉🏻  https://github.com/nextflow-io/nextflow/pull/5316

PR #5316, opened 2024-09-20
config.md - mention string or map

ewels:
See https://github.com/nextflow-io/nextflow/pull/5314#issuecomment-2363535388

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1229,7 +1229,7 @@ The `manifest` scope allows you to define some meta-data information needed when
 The following settings are available:
 
 `manifest.author`
-: Project author name (use a comma to separate multiple names).
+: Project author name (String or map).
 
 `manifest.defaultBranch`
 : Git repository default branch (default: `master`).
@@ -1256,7 +1256,7 @@ The following settings are available:
 : Project main script (default: `main.nf`).
 
 `manifest.maintainer`
-: Project maintainer name (use a comma to separate multiple names).
+: Project maintainer(s) (String or map).
 
 `manifest.name`
 : Project short name.
--- a/docs/config.md
+++ b/docs/config.md
@@ -1256,7 +1256,7 @@ The following settings are available:
 : Project main script (default: `main.nf`).
 
 `manifest.maintainer`
-: Project maintainer(s) (String or map).
+: Project maintainer(s). Can be a string of comma-separated values or a list of maps.
 
 `manifest.name`
 : Project short name.


docs/config.md
```diff
@@ -1229,7 +1229,7 @@ The `manifest` scope allows you to define some meta-data information needed when
 The following settings are available:
 
 `manifest.author`
-: Project author name (use a comma to separate multiple names).
+: Project author name (String or map).
```
bentsherman, line 1232:
```suggestion
: Project author name. Can be a string of comma-separated values or a list of maps.
```

docs/config.md
```diff
@@ -1256,7 +1256,7 @@ The following settings are available:
 : Project main script (default: `main.nf`).
 
 `manifest.maintainer`
-: Project maintainer name (use a comma to separate multiple names).
+: Project maintainer(s) (String or map).
```
bentsherman, line 1259:
```suggestion
: Project maintainer(s). Can be a string of comma-separated values or a list of maps.
```

docs/config.md
```diff
@@ -1256,7 +1256,7 @@ The following settings are available:
 : Project main script (default: `main.nf`).
 
 `manifest.maintainer`
-: Project maintainer(s) (String or map).
+: Project maintainer(s). Can be a string of comma-separated values or a list of maps.
```
pditommaso, line 1259: Nope, this is declared as a String 

https://github.com/nextflow-io/nextflow/blob/62e260437fa65c29de5a60d2491ad9aaf6f1886f/modules/nextflow/src/main/groovy/nextflow/config/Manifest.groovy#L62-L64

We could use comma separate string for more than one author as for `maintainer` (map becomes a mess to handle on backend side)

ewels: Yup, this is following this comment: https://github.com/nextflow-io/nextflow/pull/5314#issuecomment-2362956680 and originally https://github.com/nextflow-io/nextflow/issues/4034#issuecomment-1766364814

Essentially for both author and maintainer fields, I'd like to be able to optionally specify a number of fields:
* Name
* Email
* Affiliation
* ORCiD
* GitHub

..for multiple authors / maintainers.

Having this info would allow us to roll it into other provenance outputs, like BCO and RO Crate. Which would in turn give rich authorship information in downstream platforms like WorkflowHub, Seqera Pipelines, the nf-core website and more. We could also build automation to fill it from GitHub contribution statistics in @nf-core and other nice stuff.

The string is simple, but not really useful. People put all kinds of things in there in different formats and so it's impossible to use for anything downstream.

pditommaso: This is an important deviation from what was implemented. Not sure any more all those should the manifest

bentsherman: It is true the author is declared as a string, so we can't re-use it. We'll have to add a new field like `authors` and `maintainers`, maybe deprecate the old ones

But we really do need more detailed fields for the author list so that it can be easily mapped to provenance manifests like BCO

ewels:
Closing in favour of https://github.com/nextflow-io/nextflow/pull/5322

PR #5318, opened 2024-09-20
Include tower prefix in Batch job names

jorgee:
Include tower prefix in the AWS Batch job name when the TOWER_WORKFLOW_ID env var is defined


plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
```diff
@@ -843,6 +849,8 @@ class AwsBatchTaskHandler extends TaskHandler implements BatchHandler<String,Job
     }
 
 
+
+
```
pditommaso, line 853:
```suggestion
```

plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsBatchTaskHandlerTest.groovy
```diff
@@ -16,6 +16,8 @@
 
 package nextflow.cloud.aws.batch
 
+import org.bouncycastle.crypto.engines.EthereumIESEngine
```
pditommaso, line 19: Not sure why this is needed

plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsBatchTaskHandlerTest.groovy
```diff
@@ -1100,4 +1102,18 @@ class AwsBatchTaskHandlerTest extends Specification {
...
+        given:
+        def handlerWithEnv = new AwsBatchTaskHandler(environment: [TOWER_WORKFLOW_ID: '1234'])
+        def handlerWithoutEnv = new AwsBatchTaskHandler()
+        def name = 'job_1'
+        when:
+        def convertedName = handlerWithEnv.includeTowerPrefix(name)
+        def nonConvertedName = handlerWithoutEnv.includeTowerPrefix(name)
+        then:
+        convertedName == "tw-1234-job_1"
+        nonConvertedName == "job_1"
```
pditommaso, line 1116: This is OK, but it can be written using a more idiotic approach with `where:` something like: 

```
where: 
ENV                          | EXPECTED
[TOWER_WORKFLOW_ID: '1234']  |  "tw-1234-job_1"
[:]                          | "job_1"
```

there are plenty of examples look around

pditommaso (changes requested):
Left a few comment. Considering that's quite simple, it could make sense doing the same for Google Batch and Azure cloud

jorgee:
> Left a few comment. Considering that's quite simple, it could make sense doing the same for Google Batch and Azure cloud

Can I use the same PR or shoud I open a new issue?

pditommaso:
> Can I use the same PR or shoud I open a new issue?

Let's keep everything here

pditommaso:
That's the place for [Google Batch](https://github.com/nextflow-io/nextflow/blob/62e260437fa65c29de5a60d2491ad9aaf6f1886f/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy#L105-L105) and Azure [job](https://github.com/nextflow-io/nextflow/blob/62e260437fa65c29de5a60d2491ad9aaf6f1886f/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L363-L363)

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -375,12 +375,18 @@ class AzBatchService implements Closeable {
                 .trim()
                 .replaceAll(/[^a-zA-Z0-9-_]+/,'_')
 
-        final String key = "job-${Rnd.hex()}-${name}"
+        final String key = includeTowerPrefix("job-${Rnd.hex()}-${name}", System.getenv())
```
pditommaso, line 378: Better using [SysEnv.get](https://github.com/nextflow-io/nextflow/blob/6900c1b46822fa9e20d3ab135b29a52cd4b3b1e4/modules/nf-commons/src/main/nextflow/SysEnv.groovy#L40-L40) that `System.getenv`. In this case it's essentially the same, but it can simplify testing in some cases

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -375,12 +375,18 @@ class AzBatchService implements Closeable {
...
         // Azure batch job max len is 64 characters, however we keep it a bit shorter
         // because the jobId + taskId composition must be less then 100
         final MAX_LEN = 62i
         return key.size()>MAX_LEN ? key.substring(0,MAX_LEN) : key
     }
+    protected static String includeTowerPrefix(String name, Map<String,String> environment) {
```
pditommaso, line 384: Maybe, this can be turned into a common utility method used by all implementations. Something like 

```
class TaskUtils {
  
   static String includeTowerPrefix(String name, Map<String,String> env) { .. } 
 
}
```

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
jorgee: Removes in new commits

plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsBatchTaskHandlerTest.groovy
jorgee: Removed

plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsBatchTaskHandlerTest.groovy
jorgee: Changed

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
jorgee: Done

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
jorgee: Class TaskUtils created

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -103,6 +105,7 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
         this.client = executor.getClient()
         this.executor = executor
         this.jobId = customJobName(task) ?: "nf-${task.hashLog.replace('/','')}-${System.currentTimeMillis()}"
+        this.jobId = TaskUtils.includeTowerPrefix(this.jobId, SysEnv.get())
```
pditommaso, line 108: This open an interesting point, should we alter a custom name?!

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
```diff
@@ -712,7 +713,7 @@ class AwsBatchTaskHandler extends TaskHandler implements BatchHandler<String,Job
         final opts = getAwsOptions()
         final labels = task.config.getResourceLabels()
         final result = new SubmitJobRequest()
-        result.setJobName(normalizeJobName(task.name))
+        result.setJobName(TaskUtils.includeTowerPrefix(normalizeJobName(task.name), this.environment))
```
pditommaso, line 716: I've move `includeTowerPrefix` in `normalizeJobName`, so it will be easier to test

pditommaso: I'll give a try to this

pditommaso:
This is a bit tricker than planned: 

* AWS batch does not allow `-` in the job, therefore i've changed to `_` as separator 
* It maybe not possible to have a common prefix for all schedulers. For example K8s does not allow `_` in the job names. We may be to have different strategy of use labels 
* Platform/Tower prefix should always be added, even when a custom name is used 
* Job name normalisation should always have priority to truncate names too long

pditommaso:
I'm merging this but we need to review the overall strategy based on previous comment

jorgee:
> This is a bit tricker than planned:
> 
> * AWS batch does not allow `-` in the job, therefore i've changed to `_` as separator

I think AWS accepts hyphens as job name. I didn't got any error when testing it in the three clouds. Jobs appeared in the console with the '-' in the prefix.

pditommaso:
You are right, the docs says 

> The name of the job. It can be up to 128 letters long. The first character must be alphanumeric, can contain uppercase and lowercase letters, numbers, hyphens (-), and underscores (_).

Then we need to update the normalisation function in the aws handler

pditommaso:
I've partially reverted and kept only for AWS Batch, other cloud better to handle via tags/labels.  See 42dd4ba8

Issue #5320, opened 2024-09-23
E2E are not triggered in a pull request


pditommaso:
## Bug report 

When pushing a commit with the comment `[e2e stage]` end-2-end tests should be run. See here 

https://github.com/nextflow-io/nextflow/blob/master/.github/workflows/build.yml#L180

however for some reason they are skipped, see for example [here](https://github.com/nextflow-io/nextflow/actions/runs/10992221379?pr=5287)

jorgee:
The mentioned line should work for push events but not for pull_request events. The 'github.event.head_commit' does not exists in pull_request. It seems this property was part of the pull_request event in the past but not in current versions. I will check if there is a workaround to make it work in both events.

pditommaso:
ugly!

pditommaso:
I've made some research with chatgpt. It is suggesting this approach 

```
jobs:
  get-commit-message:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout code
        uses: actions/checkout@v3

      - name: Get the commit message via GitHub API
        id: get_commit_message
        run: |
          PR_NUMBER=${{ github.event.pull_request.number }}
          COMMIT_SHA=${{ github.event.pull_request.head.sha }}
          COMMIT_MESSAGE=$(curl -s \
            -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
            https://api.github.com/repos/${{ github.repository }}/commits/$COMMIT_SHA | jq -r '.commit.message')
          echo "Commit message: $COMMIT_MESSAGE"
          echo "::set-output name=commit_message::$COMMIT_MESSAGE"

      # Now you can use the commit message in an if statement
      - name: Conditional step based on commit message
        if: contains(steps.get_commit_message.outputs.commit_message, 'fix')
        run: echo "The commit message contains the word 'fix'."

```

However it looks a bit convoluted especially it we want to have it working both for push event and PR events. 

At this point better to move this logic in the run script. It already takes the commit message via using `git`. it should be moving the logic on top and skip it when the expected "tag" is missing

pditommaso:
Solved by  #5326

PR #5326, opened 2024-09-24
Fix e2e trigger event on PR

pditommaso:
This PR fixes the triggering of e2e tests on PR events by moving the conditional logic in the `test-platform/run.sh` script. 

Solves #5320


pditommaso:
Ok, it should be ok now. Learned at lot about GHA with this PR. Thanks @jorgee for pointing in the right direction.

PR #5328, opened 2024-09-24
Fix Azure Fusion env misses credentials when no key or SAS provided

pditommaso:
This PR fixes two issues: 
* Azure Fusion missing authentication if accountKey is provided
* Catch when service principal exists but no keys or SAS


pditommaso:
Finally all tests are green

PR #5336, opened 2024-09-25
Document formal script syntax

bentsherman:
Now that the docs are in better shape, we can begin work towards a formal description of the Nextflow language, as well as prepare the docs and tests for the strict syntax that will be introduced by #4613 and #4744 .

Refer to the changes to snippets and tests for examples of strict syntax:
- replace implicit `it` closure parameter with explicit parameter
- use quotes for process `env` input/output name
- move top-level statements into process / function / workflow definitions

Currently the syntax description is split across several pages:
- Scripts
- Processes
- Workflows
- Modules

It would be nice to have a single page that describes the entire language. At the same time, I don't want to move these pages around much more, and they are too hefty individually to combine them.

Since these pages cover a lot of other topics like APIs and runtime behavior, maybe it would be better to add a new page e.g. "Reference > Syntax" which gives a top-down syntax description.

Looking to the [Groovy](https://www.groovy-lang.org/documentation.html#languagespecification) docs as an example, I feel like the Scripts page could be expanded to also describe process/workflow/include syntax. It just feels weird to mention them here when there are already dedicated pages.


christopher-hakkaart:
I agree having everything on the same page would be too much, and I'm also not sure if that would make the most sense for someone accessing the docs.

So I guess the question is where the bulk of the syntax should be described in detail (a single page or separate pages).

I wonder if there can be a compromise where the bulk or common syntax is described on a Syntax page but also have Syntax sections (e.g., [Groovy closures](https://www.groovy-lang.org/closures.html)). The "main" syntax page could have small sections about the other syntax sections and then link out, rather than build everything into one page.

bentsherman:
I was thinking something similar, have a top-level page that links out to the existing pages. Two questions around that:
1. Do you think this page should be under "Developing pipelines" or a reference page?
2. Do you think the page should absorb the current "Scripts" content or also link to it?

I think we could just repurpose the "Scripts" page with an outline like this:

- (overview)
- workflows (briefly describe and link out)
- processes (briefly describe and link out)
- functions
- "basic syntax"
  - numbers, strings, lists, variables, closures, etc
  - link out to "Syntax" reference page

Then add a "Syntax" reference page which is actually a full description of the script grammar. This way the "Scripts" page doesn't have to be comprehensive on basic syntax, it can just provide the most important examples. The reference page can get very specific on things like punctuation, what kind of code is allowed where, etc

In any case, we need to have a broader conversation on how to evolve the docs from here, but I think for now I can do a first draft and let us iterate on that

christopher-hakkaart:
I was thinking something similar:

- Scripts (living under developing pipelines) have a high-level overview with minimal examples and lots of links to reference pages. The purpose is to describe the language and how it fits together but not get into details about every little thing.
- New reference page that describes syntax in detail and gives examples.
    - A focus would be to make everything very findable, so some clear grouping would be essential.

I think this would align nicely with splitting user and reference content too.

pditommaso:
Please keep PR small

docs/module.md
```diff
@@ -210,15 +210,15 @@ Pipeline B
 └── main.nf
 ```
 
-With the possibility to keep the template files inside the project L, A and B can use the modules defined in L without any changes. A future project C would do the same, just cloning L (if not available on the system) and including its module script.
+With the possibility to keep the template files inside the project L, A and B can use the modules defined in L without any changes. A future project C would do the same, just cloning L (if not available on the system) and including its module.
```
christopher-hakkaart, line 213: This is hard to follow. The relationship between Project L, and script A and B isn't clear.

christopher-hakkaart (commented):
Adding a few comments for visibility now.

bentsherman:
Okay I think I found a good scope for this PR:
- Add a new [syntax](https://deploy-preview-5336--nextflow-docs-staging.netlify.app/reference/syntax) reference page which describes the entire Nextflow syntax
- Cleanup existing pages now that they don't have to cover syntax so much, just link to syntax page instead
- Add section to scripts page on how to transition from code snippet to full pipeline script with processes, workflows, etc
- Update code snippets to comply with formal syntax
- Remove unnecessary references to Groovy in favor of Nextflow

Some items for future PRs:
- Update all e2e tests to comply with formal syntax
- Expand standard library docs to cover things like List, Map, Set APIs
- Continue to massage the pages under Developing Pipelines to be more focused

Thanks @christopher-hakkaart for the comments, I will address them shortly

docs/module.md
```diff
@@ -2,15 +2,15 @@
 
 # Modules
 
-In Nextflow, a **module** is a script that may contain functions, processes, and workflows (collectively referred to as *components*). A module can be included in other modules or pipeline scripts and even shared across workflows.
+Nextflow scripts can include **definitions** (workflows, processes, and functions) from other scripts. When a script is included in this way, it is referred to as a **module**. Modules can be included by other modules or pipeline scripts and can even be shared across workflows.
```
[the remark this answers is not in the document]
bentsherman: The original text is correct, a module is a script included by other scripts, not the other way around

docs/module.md
bentsherman: We can address this in a separate PR if you'd like. I mainly wanted to clean up some vocabulary here, don't really want to re-work this example in this PR

docs/reference/syntax.md
```diff
@@ -0,0 +1,935 @@
...
+- `&&`: logical AND
+- `||`: logical OR
+- `?:` elvis (i.e. short ternary)
+
+### Ternary expression
+
+A ternary expression consists of a *test expression*, a *true expression*, and a *false expression*, separated by a question mark and a colon:
+
+```groovy
+x % 2 == 0 ? 'x is even!' : 'x is odd!'
```
ewels, line 885: I've thought with a few of these that it'd be nice to show the example being assigned to a variable.

When it's so "naked" it feels a little too abstract. Personally I'd find this easier to understand:

```nextflow
def is_even = x % 2 == 0 ? 'x is even!' : 'x is odd!'
```

Could also have more verbose explanation of this one:

```nextflow
expression ? value_if_true : value_if_false
```

docs/reference/syntax.md
```diff
@@ -0,0 +1,935 @@
...
+- `^`
+- `|`
+- `&&`
+- `||`
+- `?:` (ternary)
+- `?:` (elvis)
+
+## Deprecations
+
+The following legacy features were excluded from this page because they are deprecated:
```
ewels, line 931: Nice 👍🏻

ewels (commented):
Had a quick run through of just the new `syntax.md` file and dropped a few first impressions.

This is great!

docs/process.md
```diff
@@ -2,39 +2,23 @@
 
 # Processes
 
-In Nextflow, a **process** is the basic processing primitive to execute a user script.
+In Nextflow, a **process** is a function that is specialized for executing scripts in a scalable and portable manner.
 
-The process definition starts with the keyword `process`, followed by process name and finally the process body delimited by curly braces. The process body must contain a string which represents the command or, more generally, a script that is executed by it. A basic process looks like the following example:
+Here is an example process definition:
```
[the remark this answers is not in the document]
bentsherman: Because the process page is still a bit reference-ish, I feel like a small example is needed at the top to give the user a taste of what is coming. I would be more amenable to removing this example if this page were more focused as a guide

Right now, this page goes through all the variations of the script/shell/exec section, then the stub, before it finally gets to inputs/outputs, and then it simply goes through each input/output type exhaustively. This initial example is basically a band-aid until the overall page is improved

docs/process.md
```diff
@@ -171,7 +155,7 @@ process align {
 }
 ```
 
-In the above example, the process will execute one of the script fragments depending on the value of the `mode` parameter. By default it will execute the `tcoffee` command, but changing the `mode` variable will cause a different branch to be executed.
+In the above example, the process will execute one of several scripts depending on the value of the `mode` parameter. By default it will execute the `tcoffee` command.
```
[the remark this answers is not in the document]
bentsherman: Not really, the `script` label is only required if other sections like `input` and `output` are present. This is covered in the syntax page

docs/process.md
```diff
@@ -492,7 +478,7 @@ In this case, `x.name` returns the file name with the parent directory (e.g. `my
 
 ### Multiple input files
 
-A `path` input can also accept a collection of files instead of a single value. In this case, the input variable will be a Groovy list, and you can use it as such.
+A `path` input can also accept a collection of files instead of a single value. In this case, the input variable will be a list, and you can use it as such.
```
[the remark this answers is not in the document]
bentsherman: "collection" is intentional here as I want to make it very clear that it's a single list item coming in from the input channel, not multiple individual items

docs/reference/syntax.md
```diff
@@ -0,0 +1,935 @@
...
+Only one output block may be defined in a script. Refer to {ref}`workflow-output-def` for the set of available target directives.
+
+## Statements
+
+Statements should be separated by a newline or semi-colon:
+
+```groovy
+// newline
+println 'Hello!'
+println 'Hello again!'
```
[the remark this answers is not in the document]
bentsherman: I think we should only show them in the docs when there are multiple statements on one line

docs/reference/syntax.md
```diff
@@ -0,0 +1,935 @@
...
+
+// semi-colon
+println 'Hello!' ; println 'Hello again!'
+```
+
+### Variable declaration
+
+Variables can be declared with the `def` keyword:
```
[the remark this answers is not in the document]
bentsherman: I am stilling thinking about the best way to communicate this, because for a time there will be a mis-match between the formal syntax and what Nextflow allows in practice

From a "formal syntax" point of view, I think we should simply say that `def` is always required. But in practice there is still a good reason to omit `def` in a process (so that you can re-use the variable in the outputs). But it seems weird to talk about processes in this section about variable declarations

So for now there is a note in the Scripts page to the effect of "sometimes variables can be declared with `def` but this is usually dangerous so you shouldn't do it"

docs/process.md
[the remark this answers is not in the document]
christopher-hakkaart: Fair - looking at it on the deploy preview it's not as noticeable. I'll resolve this one.

docs/reference/syntax.md
```diff
@@ -0,0 +1,935 @@
...
+
+If the condition is true, the if branch will be executed, otherwise the else branch will be executed.
+
+If / else statements can be chained any number of times by making the else branch another if / else statement:
+
+```groovy
+def grade = 89
+if( grade >= 90 )
+    println 'You get an A!'
+else if( grade >= 80 )
```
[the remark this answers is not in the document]
bentsherman: It is explained in the first sentence albeit a bit jargon-y

docs/process.md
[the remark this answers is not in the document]
christopher-hakkaart: Updated my suggestion.

docs/reference/syntax.md
```diff
@@ -0,0 +1,935 @@
...
+            if( grade >= 60 ) {
+                println 'You get a D!'
+            }
+            else {
+                println 'You failed.'
+            }
+        }
+    }
+}
+```
```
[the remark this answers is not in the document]
bentsherman: Maybe not. I just thought it was cool

docs/reference/syntax.md
```diff
@@ -0,0 +1,935 @@
...
+```groovy
+throw new Exception('something failed!')
+```
+
+:::{note}
+In general, the appropriate way to raise an error is to use the {ref}`error <stdlib-functions>` function:
+```groovy
+error 'something failed!'
+```
+:::
```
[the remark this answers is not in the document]
bentsherman: That is already covered by the stdlib entry for `exit`

docs/script.md
```diff
@@ -421,73 +345,68 @@ Mark = Williams
...
-// shorthand
-[1, 2, 3].each { println it }
+```groovy
+workflow {
+    println 'Hello!'
+}
 ```
 
-If the last argument is a closure, the closure can be written outside of the parentheses:
+This block is called the *entry workflow*. A script can only have one entry workflow, and it serves as the entrypoint when the script is executed. In fact, whenever a script contains only simple statements like `println 'Hello!'`, Nextflow simply treats it as an entry workflow!
```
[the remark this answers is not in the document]
christopher-hakkaart: Workflows page, the first few handful of lines.

docs/workflow.md
```diff
@@ -2,73 +2,42 @@
 
 # Workflows
 
-In Nextflow, a **workflow** is a composition of processes and dataflow logic (i.e. channels and operators).
+In Nextflow, a **workflow** is a function that is specialized for composing processes and dataflow logic (i.e. channels and operators).
 
-The workflow definition starts with the keyword `workflow`, followed by an optional name, and finally the workflow body delimited by curly braces. A basic workflow looks like the following example:
+A script can define a workflow without a name, known as the *entry workflow*, which is the entrypoint of the script:
```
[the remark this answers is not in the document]
bentsherman: Tried to make this section read better, let me know what you think

docs/reference/syntax.md
```diff
@@ -0,0 +1,935 @@
...
+```groovy
+nextflow.preview.topic = true
+```
+
+### Include
+
+An include declaration consists of an *include source* and one or more *include clauses*:
+
+```groovy
+include { foo ; bar as baz } from './some/module'
```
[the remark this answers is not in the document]
bentsherman: As this is a reference page and not a guide, I'm not as concerned about guiding the user here

docs/reference/syntax.md
```diff
@@ -0,0 +1,935 @@
...
+    THURSDAY,
+    FRIDAY,
+    SATURDAY,
+    SUNDAY
+}
+```
+
+Enum values can be accessed as `Day.MONDAY`, `Day.TUESDAY`, and so on.
+
+:::{note}
```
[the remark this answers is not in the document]
bentsherman: I don't think this rises to the level of a warning

docs/reference/syntax.md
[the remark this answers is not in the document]
bentsherman: Well, I think it is useful to clarify exactly how else statements are evaluated, as users sometimes get tripped up (see [dangling else](https://en.wikipedia.org/wiki/Dangling_else))

docs/reference/syntax.md
bentsherman: I can add a `println` so that it will at least print something when copied into a script

bentsherman: Though in general I don't want to add too much extra stuff especially around these expressions. It's becoming increasingly clear to me that this page is more of a "language specification" rather than a guide, so it can be a little abstract

docs/reference/syntax.md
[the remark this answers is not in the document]
bentsherman: I think the Scripts page is the appropriate place to address this for now, as it deals more with semantics (i.e. meaning) while this page is focused on syntax (i.e. structure)

Both of these pages can still be expanded significantly, but I'd rather save that for a future iteration

docs/reference/syntax.md
```diff
@@ -0,0 +1,957 @@
...
+- Parameter definitions
+- Workflow definitions
+- Process definitions
+- Function definitions
+- Enum types
+- Output block
+
+Script declarations are in turn composed of statements and expressions.
+
+A script may contain one or more [statements](#statements), if there are no top-level declarations. In this case, the entire script will be treated as an entry workflow. For example:
```
christopher-hakkaart, line 40:
```suggestion
If there are no top-level declarations a script may contain one or more [statements](#statements). In this case, the entire script will be treated as an entry workflow. For example:
```

docs/reference/syntax.md
```diff
@@ -0,0 +1,957 @@
...
+
+### Include
+
+An include declaration consists of an *include source* and one or more *include clauses*:
+
+```groovy
+include { foo as bar } from './some/module'
+```
+
+The include source should be a string literal and should refer to either a local path (e.g. `./module.nf`) or a plugin (e.g. `plugin/nf-hello`). Each include clause should specify a name, and may also specify an *alias*. In the example above, `foo` is included under the alias `bar`.
```
christopher-hakkaart, line 82:
```suggestion
The include source should be a string literal and should refer to either a local path (e.g. `./module.nf`) or a plugin (e.g. `plugin/nf-hello`). Each include clause should specify a name, and may also specify an *alias*. In the above example, `foo` is included under the alias `bar`.
```

docs/reference/syntax.md
```diff
@@ -0,0 +1,957 @@
...
+}
+```
+
+- Only one entry workflow may be defined in a script.
+
+- The `main:` section label can be omitted if the publish section is not specified.
+
+- The publish section consists of one or more *publish statements*. A publish statement is a [right-shift expression](#binary-expressions), where the left-hand side is an expression that refers to a value in the workflow body, and the right-hand side is an expression that returns a string.
+
+In order for a script to be executable, it must either define an entry workflow or use the implicit workflow syntax described [above](#top-level-declarations).
```
christopher-hakkaart, line 172:
```suggestion
A script must define an entry workflow or use the implicit workflow syntax described [above](#top-level-declarations) to be executable.
```

docs/reference/syntax.md
```diff
@@ -0,0 +1,957 @@
...
+
+```groovy
+(x, y) = [ 1, 2 ]
+```
+
+### Expression statement
+
+Any [expression](#expressions) can be a statement.
+
+In general, the only expressions that can have any effect as expression statements are function calls that have side effects (e.g. `println`) or an implicit return statement (e.g. in a function or closure).
```
christopher-hakkaart, line 407:
```suggestion
In general, the only expressions that can have any effect as expression statements are function calls that have side effects (e.g. `println`) or an implicit return statement (e.g. in a [function](#function) or [closure](#closure)).
```

docs/reference/syntax.md
```diff
@@ -0,0 +1,957 @@
...
+- `?:` (ternary)
+- `?:` (elvis)
+
+## Deprecations
+
+The following legacy features were excluded from this page because they are deprecated:
+
+- The `addParams` and `params` clauses of include declarations (see {ref}`module-params`)
+- The `when:` section of a process definition (see {ref}`process-when`)
+- The implicit `it` closure parameter
```
christopher-hakkaart, line 957:
```suggestion
- The `addParams` and `params` clauses of include declarations. See {ref}`module-params` for more information.
- The `when:` section of a process definition. See {ref}`process-when` for more information.
- The implicit `it` closure parameter
```

docs/reference/syntax.md
```diff
@@ -0,0 +1,957 @@
...
+
+The first line of a script can be a [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)):
+
+```sh
+#!/usr/bin/env nextflow
+```
+
+### Feature flag
+
+A feature flag declaration is an assignment, where the target should be a valid {ref}`feature flag <config-feature-flags>` and the source should be a literal (i.e. number, string, boolean):
```
christopher-hakkaart, line 68:
```suggestion
A feature flag declaration is an assignment. The target should be a valid {ref}`feature flag <config-feature-flags>` and the source should be a literal (i.e. number, string, boolean):
```

docs/reference/syntax.md
```diff
@@ -0,0 +1,957 @@
...
+
+The following definitions can be included:
+
+- Functions
+- Processes
+- Named workflows
+
+### Parameter
+
+A parameter declaration is an assignment, where the target should be a pipeline parameter and the source should be an expression:
```
christopher-hakkaart, line 112:
```suggestion
A parameter declaration is an assignment. The target should be a pipeline parameter and the source should be an expression:
```

docs/reference/syntax.md
```diff
@@ -0,0 +1,957 @@
...
+
+A list literal consists of a comma-separated list of zero or more expressions, enclosed in square brackets:
+
+```groovy
+[1, 2, 3]
+```
+
+### Map
+
+A map literal consists of a comma-separated list of one or more *map entries*, where each map entry consists of a *key expression* and *value expression* separated by a colon, enclosed in square brackets:
```
christopher-hakkaart, line 696:
```suggestion
A map literal consists of a comma-separated list of one or more *map entries*. Each map entry consists of a *key expression* and *value expression* separated by a colon and enclosed in square brackets:
```

docs/script.md
```diff
@@ -2,13 +2,9 @@
 
 # Scripts
 
-Nextflow is a domain-specific language (DSL) based on Groovy, a general-purpose programming language for the Java virtual machine. Nextflow extends the Groovy syntax with features that ease the writing of computational pipelines in a declarative manner.
+Nextflow is a workflow language that runs on the Java virtual machine (JVM). Nextflow's syntax is very similar to [Groovy](https://groovy-lang.org/), a scripting language for the JVM, but Nextflow is specialized for writing computational pipelines in a declarative manner. See {ref}`syntax-page` for a full description of the Nextflow language.
```
christopher-hakkaart, line 5:
```suggestion
Nextflow is a workflow language that runs on the Java virtual machine (JVM). Nextflow's syntax is very similar to [Groovy](https://groovy-lang.org/), a scripting language for the JVM. However, Nextflow is specialized for writing computational pipelines in a declarative manner. See {ref}`syntax-page` for a full description of the Nextflow language.
```

docs/script.md
```diff
@@ -20,41 +16,44 @@ Nextflow scripts have a maximum size of 64 KiB. To avoid this limit for large pi
...
+println flag
 
-x = "Hi"
-println x
+def str = "Hi"
+println str
 ```
 
+:::{warning}
+Variables can also be declared without `def` in many cases, but this practice is discouraged outside of simple code snippets because it can lead to a {ref}`race condition <cache-global-var-race-condition>`.
```
christopher-hakkaart, line 48:
```suggestion
Variables can also be declared without `def` in many cases. This practice is discouraged outside of simple code snippets as it can lead to a {ref}`race condition <cache-global-var-race-condition>`.
```

docs/script.md
```diff
@@ -307,110 +289,52 @@ println patch   // 3
...
 
 ```groovy
 println square(9)
 ```
 
-and get the value 81.
+The above example prints `81`.
 
-This is not very interesting until we find that we can pass the function `square` as an argument to other functions or methods. Some built-in functions take a function like this as an argument. One example is the `collect` method on lists:
+The main use case for a closure, however, is as an argument to a higher-order function:
```
christopher-hakkaart, line 314:
```suggestion
The main use case for a closure is as an argument to a higher-order function:
```

docs/script.md
```diff
@@ -421,73 +345,68 @@ Mark = Williams
...
-printf('Hello %s!\n', 'World')
+## Script definitions
 
-// shorthand
-printf 'Hello %s!\n', 'World'
-```
+So far, we have been focusing on the basic building blocks of Nextflow code, like variables, lists, strings, and closures.
 
-It is especially useful when calling a function with a closure parameter:
+In practice, however, Nextflow scripts are composed of *workflows*, *processes*, and *functions* (collectively known as *definitions*), and they can *include*  definitions from other scripts.
```
christopher-hakkaart, line 383:
```suggestion
In practice, however, Nextflow scripts are composed of *workflows*, *processes*, and *functions* (collectively known as *definitions*), and can *include*  definitions from other scripts.
```

docs/script.md
christopher-hakkaart, line 393:
```suggestion
This block is called the *entry workflow*. A script can only have one entry workflow. It serves as the entrypoint when the script is executed. Whenever a script contains only simple statements like `println 'Hello!'`, Nextflow simply treats it as an entry workflow!
```

docs/config.md
```diff
@@ -22,37 +22,65 @@ You can use the `-C <config-file>` option to use a single configuration file and
...
-### Variables
+```groovy
+params.helper_file = "${projectDir}/assets/helper.txt"
+```
 
-Configuration properties can be used as variables in the configuration file by using the usual `$propertyName` or `${expression}` syntax.
+### Blocks
 
-For example:
+A config scope can also be specified as a block, allowing multiple configuration options to be set within that block. For example:
```
christopher-hakkaart, line 51:
```suggestion
A config scope can also be specified as a block and allows multiple configuration options to be set within that block. For example:
```

docs/config.md
```diff
@@ -62,7 +90,11 @@ process.memory = '10G'
 includeConfig 'path/foo.config'
 ```
 
-When a relative path is used, it is resolved against the actual location of the including file.
+When a relative path is used, it is resolved against the location of the including file.
```
christopher-hakkaart, line 93:
```suggestion
A relative path is resolved against the location of the including file.
```

docs/process.md
```diff
@@ -276,6 +260,8 @@ Hello Mr. a
 Hello Mr. c
 ```
 
+A native process is very similar to a {ref}`function <syntax-function>`, but provides additional capabilities such as parallelism, caching, and progress logging.
```
christopher-hakkaart, line 263:
```suggestion
A native process is very similar to a {ref}`function <syntax-function>`. However, it provides additional capabilities such as parallelism, caching, and progress logging.
```

christopher-hakkaart (changes requested):
This is looking awesome.

I've added some suggestions to turn passive into active voice and split some sentences. I don't see any of these as blockers but opportunities to improve clarity. If you're happy with these I think it's good to go.

christopher-hakkaart (approved):
Thanks for adding my suggestions. The docs look good to me.

docs/config.md
```diff
@@ -22,37 +22,65 @@ You can use the `-C <config-file>` option to use a single configuration file and
...
+    retry {
+        maxAttempt = 5
+    }
+}
+```
 
 ### Includes
 
-A configuration file can include one or more configuration files using the keyword `includeConfig`. For example:
+A config file can include any number of other config files using the `includeConfig` keyword:
```
pditommaso, line 83: Is not `configuration` better here?

docs/dsl1.md
```diff
@@ -88,7 +88,7 @@ In DSL1, the entire Nextflow pipeline must be defined in a single file (e.g. `ma
 DSL2 introduces the concept of "module scripts" (or "modules" for short), which are Nextflow scripts that can be "included" by other scripts. While modules are not essential to migrating to DSL2, nor are they mandatory in DSL2 by any means, modules can help you organize a large pipeline into multiple smaller files, and take advantage of modules created by others. Check out the {ref}`module-page` to get started.
 
 :::{note}
-With DSL2, the Groovy shell used by Nextflow also imposes a 64KB size limit on pipeline scripts, so if your DSL1 script is very large, you may need to split your script into modules anyway to avoid this limit.
+DSL2 scripts cannot exceed 64 KB in size. Large DSL1 scripts may need to be split into modules to avoid this limit.
```
pditommaso, line 91: DSL1 does not exist any more since ages, it can be ignored and just say: 

> Nextflow scripts cannot exceed 64 KB in size.

docs/snippets/process-out-env.nf
```diff
@@ -1,6 +1,6 @@
 process myTask {
     output:
-    env FOO
+    env 'FOO'
```
pditommaso, line 3: This is expected to work without quotes

docs/config.md
bentsherman: I guess either one is fine, but it's verbose to write configuration all the time, so we tend to write configuration the first time and config there after

docs/snippets/process-out-env.nf
bentsherman: The formal syntax requires quotes, otherwise the parser will treat it as a variable when it's really a name. Also the quotes are more readable

I will add a note about this restriction to the other PR

docs/dsl1.md
bentsherman:
```suggestion
Nextflow scripts cannot exceed 64 KB in size.
```

bentsherman: This is the DSL1 migration page so it's fair to mention DSL1. The only purpose of this change was to remove unneeded references to Groovy

docs/snippets/process-out-env.nf
pditommaso: Actually, this was designed to work primarily without the need for quotes 

https://github.com/nextflow-io/nextflow/blob/500f49c0911bb33e3fa0f765df4a20da07a8f761/modules/nextflow/src/test/groovy/nextflow/script/params/ParamsInTest.groovy#L335-L337

docs/config.md
pditommaso: True that verbose, but this is a formal docs, "configuration" sounds better imo

docs/snippets/process-out-env.nf
bentsherman: both are supported, but the one without quotes is problematic because it looks like a variable declaration when in reality it is not

pditommaso: at nextflow level you mean?

docs/config.md
bentsherman: okay

bentsherman:
```suggestion
A configuration file can include any number of other configuration files using the `includeConfig` keyword:
```

docs/snippets/process-out-env.nf
bentsherman: yes

pditommaso: Make sense. Agree

docs/snippets/first.nf
```diff
@@ -15,5 +15,5 @@ Channel.of( 1, 2, 'a', 'b', 3 )
 
 // emits the first item for which the predicate evaluates to true: 4
 Channel.of( 1, 2, 3, 4, 5 )
-    .first { it > 3 }
+    .first { v -> v > 3 }
```
pditommaso, line 18: While I agree it's generally better to use the extended syntax. it would be nice to mention about `it` as implicit argument and the possibility to have one more argument in a closure, when describing what a closure is: 


```
A closure is a function that can be used like a regular value. Typically, closures are passed as arguments to higher-order functions to express computations in a declarative manner.

For example:

def square = { v -> v * v }
The above example defines a closure, which takes one parameter named v and returns the “square” of v (v * v). The closure is assigned to the variable square.
```

bentsherman: It is mentioned at the bottom of the syntax page:

```md
## Deprecations

The following legacy features were excluded from this page because they are deprecated:

...
- The implicit `it` closure parameter
```

The migration notes for implicit `it` are covered in the private docs PR

pditommaso: Not sure, `it` should be considered deprecated. worth noting there are also slight different semantic with it. I think it should be mentioned as a note or tip also because it's widely adopted, and therefore it would be informative to know about it

pditommaso: Disagree

bentsherman: Thought about this more. For the sake of new users reading existing code, I have added a note about the implicit closure parameter to the Scripts page. There is also an example of a closure with multiple parameters there.

pditommaso:
I've left out the mention to deprecation because deserver further discussion 👉 https://github.com/nextflow-io/nextflow/pull/5336/commits/3f6f37796065fd644813a1b75ed91fea623080c7

Issue #5208, opened 2024-08-06
Class cannot be created (missing no-arg constructor): nextflow.file.http.XPath


nvnieuwk:
### Expected behavior and actual behavior

We encountered this error when debugging a PR in [nf-core/demultiplex](https://github.com/nf-core/demultiplex/pull/224). This issue only occurs with the last version of Nextflow edge (24.07.0-edge)

<details>
<summary>Error traceback</summary>
<br>
<pre>
com.esotericsoftware.kryo.KryoException: Class cannot be created (missing no-arg constructor): nextflow.file.http.XPath
	at com.esotericsoftware.kryo.Kryo$DefaultInstantiatorStrategy.newInstantiatorOf(Kryo.java:1228)
	at nextflow.util.InstantiationStrategy.newInstantiatorOf(SerializationHelper.groovy:247)
	at com.esotericsoftware.kryo.Kryo.newInstantiator(Kryo.java:1049)
	at com.esotericsoftware.kryo.Kryo.newInstance(Kryo.java:1058)
	at com.esotericsoftware.kryo.serializers.FieldSerializer.create(FieldSerializer.java:547)
	at com.esotericsoftware.kryo.serializers.FieldSerializer.read(FieldSerializer.java:523)
	at com.esotericsoftware.kryo.Kryo.readClassAndObject(Kryo.java:761)
	at org.codehaus.groovy.vmplugin.v8.IndyInterface.fromCache(IndyInterface.java:321)
	at nextflow.util.KryoHelper.deserialize(SerializationHelper.groovy:181)
	at nextflow.util.KryoHelper.deserialize(SerializationHelper.groovy)
	at nextflow.file.SortFileCollector.saveFile(SortFileCollector.groovy:315)
	at nextflow.file.FileCollector.saveTo0(FileCollector.groovy:228)
	at nextflow.file.FileCollector.saveTo(FileCollector.groovy:251)
	at org.codehaus.groovy.vmplugin.v8.IndyInterface.fromCache(IndyInterface.java:321)
	at nextflow.extension.CollectFileOp.emitItems(CollectFileOp.groovy:187)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
[...]

PR #5339, opened 2024-09-26
Add XPath default serializer

jorgee:
Add XPath default serializer

### DIFF
--- a/modules/nf-httpfs/src/main/nextflow/file/http/XPath.groovy
+++ b/modules/nf-httpfs/src/main/nextflow/file/http/XPath.groovy
@@ -45,6 +45,8 @@ class XPath implements Path {
 
     private String query
 
+    XPath(){}
+
     XPath(XFileSystem fs, String path) {
         this(fs, path, EMPTY)
     }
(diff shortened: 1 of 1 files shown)


pditommaso:
Any test unit tests that can replicate the issue and validate the patch?

jorgee:
> Any test unit tests that can replicate the issue and validate the patch?

I have tried to reproduce the error. In fact, there is a unit test checking the XPath serialized and deserialized from a byte array that is the one used in the error in SortFileCollector.saveFile function. This test is not failing even without the no-arg constructor. 
I think, the issue is more related to how a particular XPath instance has been serialized that forces Kyro to use the no-args constructor when deserializing in this particular case and not in the others. Finding the particular case could take some time, and I think just adding the default constructor should solve the issue. I can continuing looking for this particular case, but I don't know if it is good to invest more time on it.

pditommaso:
Curious to see the error trace. Is there a link to the issue?

jorgee:
> Curious to see the error trace. Is there a link to the issue?

https://github.com/nextflow-io/nextflow/issues/5208

pditommaso (approved):

PR #5345, opened 2024-09-30
Detecting errors in data unstaging

### LINKED ISSUE #3711: Validation of task outputs

jorgee:
close #3711 

- Enables fast failure (set -e) at unstage part and removes "|| true" statement in the unstage.
- Splits unstage in control and task outputs.
- Unstage of control outputs after task outputs to see errors in .command.err/out
- Command to skip task outputs unstaging is 'true' instead of null (Google Life sciences part)


bentsherman:
@pditommaso are we still planning to include this in 24.11 edge ? Asking on behalf of @FriederikeHanssen and some customers

modules/nextflow/src/main/resources/nextflow/executor/command-run.txt
```diff
@@ -99,7 +99,13 @@ nxf_fs_fcp() {
 }
 
 on_exit() {
-    exit_status=${nxf_main_ret:=$?}
+    ## Capture possible errors.
+    ## Can be caused either by the task script, unstage script or after script if defined
+    local last_err=$?
+    ## capture the task error first or fallback to unstage error
+    exit_status=${nxf_main_ret:=0}
+    [[ ${exit_status} -eq 0 && ${nxf_unstage_ret:0} -ne 0 ]] && exit_status=${nxf_unstage_ret}
```
pditommaso, line 107: Why there's `${nxf_unstage_ret:0}` in the condition and `exit_status=${nxf_unstage_ret}` in the assignment? it may be better to use a local intermediate variable as for `local last_err=$?`

modules/nextflow/src/main/resources/nextflow/executor/command-run.txt
```diff
@@ -150,11 +159,17 @@ nxf_main() {
...
+    ## Deactivate fast failure to allow uploading stdout and stderr files later
     set +e
     (set -o pipefail; (nxf_launch | tee {{stdout_file}}) 3>&1 1>&2 2>&3 | tee {{stderr_file}}) &
     pid=$!
     wait $pid || nxf_main_ret=$?
-    {{unstage_cmd}}
+    if [[ ${nxf_main_ret:=0} == 0 ]]; then
+        ## Data unstaging redirecting stdout and stderr with append mode
+        (set -e -o pipefail; ({{unstage_cmd}} | tee -a {{stdout_file}}) 3>&1 1>&2 2>&3 | tee -a {{stderr_file}})
+        nxf_unstage_ret=$?
```
pditommaso, line 170: Was this tested which S3 uploads that are made in a parallel logic? see [here](https://github.com/nextflow-io/nextflow/blob/test_3711/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchFileCopyStrategy.groovy#L121-L127)

modules/nextflow/src/main/resources/nextflow/executor/command-run.txt
jorgee: This is to decide which exit status to print in the exit_file. First, I check the task ('nxf_main_ret'), then the unstage if exists ('nxf_unstage_ret') and finally the after script ('last_err'). Regarding the local, do you mean make 'exit_status' local?

modules/nextflow/src/main/resources/nextflow/executor/command-run.txt
jorgee: I tested with azure, that is also doing the copies in parallel.

jorgee: I have just seen, I created a validation test using a fake aws command to produce errors in stage_out

modules/nextflow/src/main/resources/nextflow/executor/command-run.txt
pditommaso: > do you mean make 'exit_status' local?

yes, as long as is not used anywhere else


Let's add extensive comment to describe all this logic that's quite complex

modules/nextflow/src/main/resources/nextflow/executor/command-run.txt
```diff
@@ -99,7 +99,13 @@ nxf_fs_fcp() {
...
 
 on_exit() {
-    exit_status=${nxf_main_ret:=$?}
+    ## Capture possible errors.
+    ## Can be caused either by the task script, unstage script or after script if defined
+    local last_err=$?
+    ## capture the task error first or fallback to unstage error
+    local exit_status=${nxf_main_ret:=0}
+    [[ ${exit_status} -eq 0 && ${nxf_unstage_ret:=0} -ne 0 ]] && exit_status=${nxf_unstage_ret:=0}
+    [[ ${exit_status} -eq 0 && ${last_err} -ne 0 ]] && exit_status=${last_err}
```
pditommaso, line 108: This is still very hard to decode, it would be nice to be more verbose about the different of `last_err` vs `nxf_main_ret` vs `nxf_unstage_ret` and the general logic.

PR #5351, opened 2024-10-02
Hash improvements

### LINKED ISSUE #4916: Changes in object input structure use cached values rather than starting new runs for exec

Changes to objects passed as input should trigger new runs when using `-resume`. When changing an object's structure, e.g. putting it in an array, or map, the process uses the cached values from before, rather than executing new runs.

First run:
```nextflow
workflow {
    Channel.of(
        [ id:'foo', taxid: '632'],
        [ id:'bar', taxid: '632']
    )
    | TASK
    | view
}

process TASK {
    input:
    val meta

    exec:
    file("$task.workDir/node_id.txt").text = meta.taxid
    
    output:
    tuple val(meta), path('node_id.txt'), emit: node_id
}
```
Changed code (run 2):
```nextflow
workflow {
    Channel.of(
        [map:[ id:'foo', taxid: '632']],
        [map:[ id:'bar', taxid: '632']]
    )
    | TASK
    | view
}

process TASK {
    input:
    val meta

    exec:
    file("$task.workDir/node_id.txt").text = meta.taxid
    
    output:
    tuple val(meta), path('node_id.txt'), emit: node_id
}
```

### Program output 

Run 1:
```console
$ nextflow run main.nf 
N E X T F L O W  ~  version 23.10.1
Launching `main.nf` [zen_leibniz] DSL2 - revision: eaf2b5bb3d
executor >  local (2)
[31/f024e0] process > TASK (2) [100%] 2 of 2 ✔
[[id:foo, taxid:632], /workspace/Nextflow_sandbox/work/8e/853bdb1d954eef65d82eb81ba481e1/node_id.txt]
[[id:bar, taxid:632], /workspace/Nextflow_sandbox/work/31/f024e0f621ef2b27d3567887a7c5d3/node_id.txt]
```
Run 2: Expect error due to change in input object structure, but uses cached values instead.
```console
$ nextflow run main.nf -resume
N E X T F L O W  ~  version 23.10.1
Launching `main.nf` [cheesy_allen] DSL2 - revision: 8074d4d3e0
[6b/fb2322] process > TASK (1) [100%] 2 of 2, cached: 2 ✔
[[id:bar, taxid:632], /workspace/Nextflow_sandbox/work/56/2935d71088d17207dab38381590386/node_id.txt]
[[id:foo, taxid:632], /workspace/Nextflow_sandbox/work/6b/fb2322b25197873c2aaf6f0c75084e/node_id.txt]
```

### Environment 

* Nextflow version: 23.10.1
* Java version: openjdk 17.0.10-internal 2024-01-16
* Operating system: Linux (Gitpod)

bentsherman:
Close #4916 

- Map should be hashed as an unordered collection of map entries
- all collections other than List should be hashed as unordered collection (in practice what we're already doing but make it more clear)
- while we're changing the hash, might as well make the enum hash more efficient

Tests need to be updated. This PR will break everyone's cache.

### DIFF
--- a/modules/nf-commons/src/main/nextflow/util/HashBuilder.java
+++ b/modules/nf-commons/src/main/nextflow/util/HashBuilder.java
@@ -144,10 +144,13 @@ public class HashBuilder {
             for( Object item : ((Object[])value) )
                 with(item);
 
-        // note: should map be order invariant as Set ?
-        else if( value instanceof Map )
-            for( Object item : ((Map)value).values() )
+        else if( value instanceof List ) {
+            for( Object item : ((List)value) )
                 with(item);
+        }
+
+        else if( value instanceof Map )
+            hashUnorderedCollection(hasher, ((Map) value).entrySet(), mode);
 
         else if( value instanceof Map.Entry ) {
             Map.Entry entry = (Map.Entry)value;
@@ -155,12 +158,8 @@ public class HashBuilder {
             with(entry.getValue());
         }
 
-        else if( value instanceof Bag || value instanceof Set )
+        else if( value instanceof Collection )
             hashUnorderedCollection(hasher, (Collection) value, mode);
-
-        else if( value instanceof Collection)
-            for( Object item : ((Collection)value) )
-                with(item);
 
         else if( value instanceof FileHolder )
             with(((FileHolder) value).getSourceObj());
@@ -185,8 +184,10 @@ public class HashBuilder {
         else if( value instanceof CacheFunnel )
             ((CacheFunnel)value).funnel(hasher, mode);
 
-        else if( value instanceof Enum )
-            hasher.putUnencodedChars( value.getClass().getName() + "." + value );
+        else if( value instanceof Enum ) {
+            hasher.putUnencodedChars(value.getClass().getName());
+            hasher.putUnencodedChars(value);
+        }
 
         else {
             Bolts.debug1(log, FIRST_ONLY, "[WARN] Unknown hashing type: " + value.getClass());


yfarjoun:
This will be a great fix! thanks for working on it.

Can I suggest that the keys also be part of the hash? And the pairings? 

Perhaps this should be hashed as a set of length-2 lists [key,value] ? that way if someone were to change the mapping, or the keys, the cache would not match.

yfarjoun:
Also, the current implementation has hash([item]) == hash(item). This is also wrong... perhaps one could "salt" the hash function with the type of object being hashed:

in (terrible) shorthand:

if "object":
   with(["object", item])
if "list':
   with(["list",item]) <- here item will be the list
if "map":
   with(["map",item]) <- here item will be the map

bentsherman:
@yfarjoun this line does exactly what you describe for maps:

```java
        else if( value instanceof Map )
            hashUnorderedCollection(hasher, ((Map) value).entrySet(), mode);
```

The "entry set" is the set of key-value pairs.

yfarjoun:
thanks @bentsherman does that mean that a entry [key:value] will has to the same as the string "keyvalue"? or the set {"keyvalue"}?

bentsherman:
I guess each entry would have the same hash as the list `[key, value]`

yfarjoun:
I still think that every object should be prefixed with it's type (or some sentinel thereof) in order to avoid easy-to-prevent hash collisions.... but even without it, your solution will be better than the current state. 

🌟

bentsherman:
Closing in favor of #6679

PR #5356, opened 2024-10-02
Add retry policy to google batch describe task

bentsherman:
Raised by Adam during some testing. The Google Batch describeTask didn't have a retry policy like the other methods and sure enough a 5xx error struck


pditommaso (approved):
Nice catch

PR #5359, opened 2024-10-03
Fixing issue when retrying with stub

jorgee:
When retrying with stubs, the TaskRun copy in the retry is set with the body instead of the stub. 
The PR add the same conditional as in invoke task to put the correct closure according to the stub flag.


pditommaso:
Worth adding a test. Some examples here 

https://github.com/nextflow-io/nextflow/blob/fix-boostrap-log-level/modules/nextflow/src/test/groovy/nextflow/script/ScriptRunnerTest.groovy#L659-L772

jorgee:
> Worth adding a test. Some examples here
> 
> https://github.com/nextflow-io/nextflow/blob/fix-boostrap-log-level/modules/nextflow/src/test/groovy/nextflow/script/ScriptRunnerTest.groovy#L659-L772

I have tried to add a unit test but to test the feature it requires that the task fails to retry and check it is using the stub. The MockTaskHandler used by the MockScriptRunner do not support it. I have also tried with the normal ScriptRunner but it stop the execution at the end with an AbortException.

At the end, I added the test as a integration test in the tests folder.

bentsherman:
Marking your PR as a draft. In general, please submit PRs as drafts and then mark them "ready for review" when it's ready

pditommaso (approved):

PR #5365, opened 2024-10-04
Fix incorrect total count at the end of the execution

jorgee:
The number of total tasks during the execution did not match with the total tasks at the end of the execution. This was due to  aborted taks where not included in the ProgressRecord total count. This PR fixes it, including aborted tasks to the total count.


bentsherman (approved):

pditommaso:
Tagging @swampie for visibility in the case you will notice any chance in platform totals (it should not)

pditommaso (approved):

PR #5366, opened 2024-10-04
Fix workflow onComplete and onError in the entry workflow

### LINKED ISSUE #5261: Workflow onComplete/onError handlers don't work when defined in the entry workflow
The following example:
```groovy
workflow {
    params.outdir = 'results'

    workflow.onComplete {
        if( workflow.success )
            log.info "Success! View results: $params.outdir"
        else
            log.info "Failure!"
    }
}
```

Fails with an error that `workflow` is null in the onComplete handler, even though it should be in scope. The same error will happen for `params` if it is accessed first. These handlers should work regardless of whether they are defined at the script level or in the entry workflow.

There are two workarounds:
1. Assign the onComplete handler: `workflow.onComplete = { ...`
2. Declare local variables that reference `workflow` and `params`:
   ```groovy
   def myParams = params
   def myWorkflow = workflow
   workflow.onComplete {
     // ...
   }
   ```

bentsherman:
Close #5261 

Tested with this script:
```groovy
workflow {
    params.outdir = 'results'

    def local = 'local'

    workflow.onComplete {
        if( workflow.success )
            log.info "Success! View results: $params.outdir"
        else
            log.info "Failure!"

        log.info "local variable: ${local}"
    }
}
```

The problem is that this change breaks the WorkflowMetadata unit tests. But these unit tests do not accurately represent how the onComplete/onError is defined

I could rewrite the unit tests to print to stdout and try to capture the stdout. But these tests probably just need to be e2e tests in order to test the actual script context

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/WorkflowMetadata.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/WorkflowMetadata.groovy
@@ -303,12 +303,7 @@ class WorkflowMetadata {
      * @param action The action handler
      */
     void onComplete( Closure action ) {
-
-        final clone = (Closure)action.clone()
-        clone.delegate = NF.binding.variables
-        clone.resolveStrategy = Closure.DELEGATE_FIRST
-
-        onCompleteActions.add(clone)
+        onCompleteActions.add(action)
     }
 
     /**
@@ -335,12 +330,7 @@ class WorkflowMetadata {
      * @param action
      */
     void onError( Closure action ) {
-
-        final clone = (Closure)action.clone()
-        clone.delegate = NF.binding.variables
-        clone.resolveStrategy = Closure.DELEGATE_FIRST
-
-        onErrorActions.add(clone)
+        onErrorActions.add(action)
     }
 
     /**


bentsherman:
@jorgee can you help me move this PR forward when you have some time? I'm hoping to get it into a 24.10.x patch release

We need to fix these failing tests:
```
WorkflowMetadataTest > should populate workflow object FAILED
    org.spockframework.runtime.SpockComparisonFailure at WorkflowMetadataTest.groovy:112

WorkflowMetadataTest > should access workflow script variables onComplete FAILED
    org.spockframework.runtime.SpockComparisonFailure at WorkflowMetadataTest.groovy:174

WorkflowMetadataTest > should access workflow script variables onError FAILED
    org.spockframework.runtime.SpockComparisonFailure at WorkflowMetadataTest.groovy:244
```

Currently these tests are trying to capture the stdout, and they are setting up the onComplete handler in a different way than it is actually used. I think they just need to be rewritten as e2e tests that check the console output for the printed lines, as other e2e tests do.

jorgee:
The changes were also breaking the integration_test `workflow_oncomplete.nf`. In the case, that the closure comes from the configuration, we need to change the closure delegate. 
I have added new functions to manage when onComplete and onError are called from the configuration.

bentsherman:
Closing since this doesn't really matter anymore with the new `onComplete:` and `onError:` sections

Issue #3593, opened 2023-01-31
Main branch not recognised when executing an online workflow


marcodelapierre:
I have a github repo whose primary branch is called `main` instead of `master`, which is becoming increasingly common.
Nextflow asks me to specify the branch with `-r `:
```
$ nextflow run marcodelapierre/toy-cowsay-nf -profile condawave

N E X T F L O W  ~  version 22.12.0-edge
Pulling marcodelapierre/toy-cowsay-nf ...
 downloaded from https://github.com/marcodelapierre/toy-cowsay-nf.git
Project `marcodelapierre/toy-cowsay-nf` is currently stickied on revision: main -- you need to explicitly specify a revision with the option `-r` in order to use it
```

All good with `-r`:
```
$ nextflow run -r main marcodelapierre/toy-cowsay-nf -profile condawave

N E X T F L O W  ~  version 22.12.0-edge
Launching `https://github.com/marcodelapierre/toy-cowsay-nf` [cranky_hypatia] DSL2 - revision: cd804ac170 [main]
executor >  local (8)
[b7/d27846] process > sayHello (1) [100%] 4 of 4 ✔
[35/cabd24] process > cowSay (4)   [100%] 4 of 4 ✔
```

I suggest to enable Nextflow to look for both `master` and `main` branches.

marcodelapierre:
pinging this one back as I think it should be discussed - should be a quick fix, if approved for action

@pditommaso @bentsherman @jordeu

bentsherman:
Basically the AssetManager should be modified to accept both `master` and `main` as "default branch" names: https://github.com/nextflow-io/nextflow/blob/fe5bea992e2c0dd99a2f72388df386975c32c83a/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy#L911-L915

marcodelapierre:
is there a way to know what is the default branch of a repo?

jordeu:
> is there a way to know what is the default branch of a repo?

The default branch concept do not exists on Git, it is a provider concept. I know that using Github API you can check which is the default branch.

pditommaso:
Indeed https://stackoverflow.com/a/16501903/395921. 

Quite surely it's available also on BitBucket and GitLab

ewels:
This just came up again, @Biowilko was asking if it was the case. I had totally forgotten about this issue and assured him that `master` couldn't be hardcoded in the codebase 👀 (then I tried it myself)

[...]

PR #5375, opened 2024-10-07
Identify default git branch when downloading pipelines (#3593)

tom-seqera:
Uses the symbolic ref approach described here https://stackoverflow.com/a/44750379 to identify the default branch when downloading a pipeline repository.

jgit doesn't automatically create the symbolic ref, but does checkout the remote's default branch. This change creates the symbolic ref after cloning, and uses it when determining the default branch. This change does not attempt to retroactively create the symbolic ref on already downloaded local repos.

If the manifest specifices a `defaultBranch`, then that will override the symbolic ref. If neither `defaultBranch` nor symbolic ref is present, `master` will be used as a fallback.


modules/nextflow/src/test/groovy/nextflow/cli/CmdInfoTest.groovy
```diff
@@ -89,7 +89,7 @@ class CmdInfoTest extends Specification {
         json.repository == "https://github.com/nextflow-io/hello"
         json.localPath == "$tempDir/nextflow-io/hello"
         json.manifest.mainScript == 'main.nf'
-        json.manifest.defaultBranch == 'master'
+        json.manifest.defaultBranch == null
```
pditommaso, line 92: Not getting this

tom-seqera: The `Manifest` object now used to return `master` if `defaultBranch` isn't specified in the manifest file. 

Now it returns `null` so the `AssetManager` can tell the difference between not-set-in-manifest vs explicitly-set-to-master.

pditommaso (changes requested):
Let's add a test checking an ad-hoc test repo using a non-default branch

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
```diff
@@ -730,6 +755,10 @@ class AssetManager {
     }
 
 
+    private static boolean isRemoteBranch(Ref ref) {
```
pditommaso, line 758: Likely a unit test can be added for this

tests/checks/main-branch.nf/.checks
```diff
@@ -0,0 +1,6 @@
+#
+# run remote pipeline, which has default branch 'main'
+#
+$NXF_CMD -q run nextflow-io/socks $WITH_DOCKER --out result.txt | tee stdout
+[[ `grep 'Launching \`https://github.com/nextflow-io/socks\`' .nextflow.log | grep -c '[main]'` == 1 ]] || false
+[[ $? == 0 ]] || exit 1
```
pditommaso, line 6: Let's keep this as a unit test. Something like this 

https://github.com/nextflow-io/nextflow/blob/e794e868fd3d62346e09e5deeba612976dd5143e/modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy#L160-L178

tom-seqera: My initial attempt was along these lines, but downloading isn't enough to trigger the "stuck on revision" error - it only happens when trying to run the pipeline.

tom-seqera: It can be triggered by calling two methods:
```
manager.download()
manager.checkout(null)
```

but is that perhaps making too many assumptions about how the code will executed when actually running?

pditommaso: > but downloading isn't enough to trigger the "stuck on revision" error

is that really needed? at the end the test should verify that the repo is cloned with the expected  (non-default) branch (i guess)

tom-seqera: The repo is successfully cloned - even before these changes - and jgit already automatically checks out the remote's default branch (in the socks case, `main`).

The issue happens later: when running the pipeline without specifying a revision, nextflow checks that the current branch matches what it expects the default branch to be (ie `master`).

pditommaso: I see, then let's go with [this](https://github.com/nextflow-io/nextflow/pull/5375#discussion_r1791872286) approach, no?

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
tom-seqera: Added a direct test in `AssetManagerTest.groovy`, and modified `CmdInfoTest.groovy` to test indirectly.

tests/checks/main-branch.nf/.checks
tom-seqera: Updated. 

It's a simpler test this way, but it does carry a small risk of failing to catch a regression if the real codepath is modified to work differently.

pditommaso: Excellent

pditommaso:
Could not resist to make this change 6795ca99. @tom-seqera would make sense to add a unit test for `getRemoteBranch` or it's too paranoic ?

PR #5391, opened 2024-10-10
AWS: switch from containerProperties to ecsProperties

### LINKED ISSUE #5096: AWS: Switch from `containerProperties ` to `ecsProperties`
According to @delagoya [we need to change this](https://x.com/delagoya/status/1806036909623329209) ASAP:

> I don't know who needs to hear this, but:
> 
> If you use AWS Batch, you should default to using the new `ecsProperties` parameter for all your job definitions. Migrate any past job definition to a new version that moves from `containerProperties` 
>
> Otherwise you are going to Experience Bij later when you have to submit jobs with overrides and forget where in the API those overrides go, depending on the job definition

See also: https://github.com/nextflow-io/nextflow/issues/4979 where this was a requirement for supporting `ipcMode`.

NB also the proposed move to Java SDKv2 (https://github.com/nextflow-io/nextflow/issues/4741), though this shouldn't be required for this switch according to https://github.com/nextflow-io/nextflow/issues/4979#issuecomment-2101439206

tom-seqera:
Closes #5096 

Use [EcsProperties](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/batch/model/EcsProperties.html) instead of [ContainerProperties](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/batch/model/ContainerProperties.html) to configure AWS Batch jobs.

See [AWS docs](https://docs.aws.amazon.com/batch/latest/userguide/multi-container-jobs.html)


plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchHelper.groovy
```diff
@@ -181,7 +181,8 @@ class AwsBatchHelper {
         final response = batchClient.describeJobs(request)
         if( response.jobs ) {
             final detail = response.jobs[0]
-            return detail.container.logStreamName
+            return detail.ecsProperties?.taskProperties[0].containers[0].logStreamName
```
pditommaso, line 184:
```suggestion
            return detail.ecsProperties?.taskProperties[0]?.containers[0]?.logStreamName
```


if `detail.ecsProperties` can be null all the other requires `?.`

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchHelper.groovy
```diff
@@ -181,7 +181,9 @@ class AwsBatchHelper {
         final response = batchClient.describeJobs(request)
         if( response.jobs ) {
             final detail = response.jobs[0]
-            return detail.container.logStreamName
+            return detail.ecsProperties?.taskProperties[0]?.containers[0]?.logStreamName
```
pditommaso, line 184: Not sure if with an empty list `[0]` returns `IndexOutOfBoundException`, may be it's better to wrap this (and other similar) in helper method and use if try/catch(Exception) returning a `null` if something is going wrong

tom-seqera: Just checked and it seems `emptyList[0]` returns null so I think that's fine. Unfortunately `obj.nullList[0]` throws a NPE.

This is a bit awkward, but would work:
```groovy
detail.ecsProperties?.taskProperties?[0]?.containers?[0]?.logStreamName
```

pditommaso: Too ugly, make sense to have little helper method

pditommaso:
Closing because this changes impact also the xpack plugin and there's a priority need for it. We'll reconsidering this on the upcoming migration to AWS SDK v2

PR #5402, opened 2024-10-15
Exposing trace record meta info in the task context

jorgee:
When retry trace record is generated and included in the task context with variable `trace`

In the first attempt, trace is null. I tried to set a partial trace record for the first attempt generating the trace record just after the creation of the TaskHandler. But the evaluation of the directives is executed before the generation of the first TaskHandler at executor.submit. So, the value was null. One alternative, could be to move the generation of the trace record to the TaskRun but it implies to move some data such as the status, start and stop time in the TaskRun.

Included a suggestion for documentation

### DIFF
--- a/docs/process.md
+++ b/docs/process.md
@@ -1290,7 +1290,7 @@ Similarly to the previous scenario, task resources can be also updated according
 
 ```groovy
 process foo {
-    memory { (trace != null) ? trace.memory * 2 : (1.GB) }
+    memory { task.attempt > 1 ? trace.memory * 2 : (1.GB) }
     errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
     maxRetries 3
 
--- a/docs/process.md
+++ b/docs/process.md
@@ -1284,13 +1284,13 @@ If the task execution fail reporting an exit status in the range between 137 and
 The directive {ref}`process-maxretries` set the maximum number of time the same task can be re-executed.
 
 ### Dynamic task resources with previous execution trace
-:::{versionadded} 24.09.1-edge
+:::{versionadded} 24.09.2-edge
 :::
-Similarly to the previous scenario, task resources can be also updated according to the metrics included in the {ref}`trace record <trace-report>` of the previous task attempt. The metrics can be accessed through the `trace` variable. For example:
+Similarly to the previous scenario, task resources can be also updated according to the metrics included in the {ref}`trace record <trace-report>` of the previous task attempt. The metrics can be accessed through the `task.previousTrace` variable. For example:
 
 ```groovy
 process foo {
-    memory { task.attempt > 1 ? trace.memory * 2 : (1.GB) }
+    memory { task.attempt > 1 ? task.previousTrace.memory * 2 : (1.GB) }
     errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
     maxRetries 3
 
--- a/docs/process.md
+++ b/docs/process.md
@@ -1286,6 +1286,7 @@ The directive {ref}`process-maxretries` set the maximum number of time the same 
 ### Dynamic task resources with previous execution trace
 :::{versionadded} 24.09.2-edge
 :::
+
 Similarly to the previous scenario, task resources can be also updated according to the metrics included in the {ref}`trace record <trace-report>` of the previous task attempt. The metrics can be accessed through the `task.previousTrace` variable. For example:
 
 ```groovy
--- a/docs/process.md
+++ b/docs/process.md
@@ -1290,7 +1290,7 @@ Similarly to the previous scenario, task resources can be also updated according
 
 ```groovy
 process foo {
-    memory { task.attempt > 1 ? task.previousTrace.memory * 2 : (1.GB) }
+    memory { (task.attempt > 1) ? task.previousTrace.memory * 2 : (1.GB) }
     errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
     maxRetries 3
 
--- a/docs/process.md
+++ b/docs/process.md
@@ -1291,7 +1291,7 @@ Similarly to the previous scenario, task resources can be also updated according
 
 ```groovy
 process foo {
-    memory { (task.attempt > 1) ? task.previousTrace.memory * 2 : (1.GB) }
+    memory { task.attempt > 1 ? task.previousTrace.memory * 2 : (1.GB) }
     errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
     maxRetries 3
 
--- a/docs/process.md
+++ b/docs/process.md
@@ -1299,7 +1299,7 @@ process foo {
     <your job here>
 }
 ```
-In the above example the {ref}`process-memory` is set according to previous trace record metrics. In the first attempt, when no trace metrics are available, it is set to one GB, while in the other attempts it is doubling the previously allocated memory. The available metrics in the trace record can be found in the {ref}`Trace Report section <trace-report>`.
+In the above example, the {ref}`process-memory` is set according to previous trace record metrics. In the first attempt, when no trace metrics are available, it is set to one GB. In the subsequent attempts, it doubles the previously allocated memory. See {ref}`trace-report` for more information about trace records.
 
 
 ### Dynamic retry with backoff
--- a/docs/process.md
+++ b/docs/process.md
@@ -1283,6 +1283,24 @@ If the task execution fail reporting an exit status in the range between 137 and
 
 The directive {ref}`process-maxretries` set the maximum number of time the same task can be re-executed.
 
+### Dynamic task resources with previous execution trace
+:::{versionadded} 24.09.1-edge
+:::
+Similarly to the previous scenario, task resources can be also updated according to the metrics included in the {ref}`trace record <trace-report>` of the previous task attempt. The metrics can be accessed through the `trace` variable. For example:
+
+```groovy
+process foo {
+    memory { (trace != null) ? trace.memory * 2 : (1.GB) }
+    errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
+    maxRetries 3
+
+    script:
+    <your job here>
+}
+```
+In the above example the {ref}`process-memory` is set according to previous trace record metrics. In the first attempt, when no trace metrics are available, it is set to one GB, while in the other attempts it is doubling the previously allocated memory. The available metrics in the trace record can be found in the {ref}`Trace Report section <trace-report>`.
+
+
 ### Dynamic retry with backoff
 
 There are cases in which the required execution resources may be temporary unavailable e.g. network congestion. In these cases immediately re-executing the task will likely result in the identical error. A retry with an exponential backoff delay can better recover these error conditions:
--- a/docs/process.md
+++ b/docs/process.md
@@ -1287,7 +1287,7 @@ The directive {ref}`process-maxretries` set the maximum number of time the same 
 :::{versionadded} 24.10.0
 :::
 
-Similarly to the previous scenario, task resources can be also updated according to the metrics included in the {ref}`trace record <trace-report>` of the previous task attempt. The metrics can be accessed through the `task.previousTrace` variable. For example:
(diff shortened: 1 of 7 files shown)


pditommaso:
Embarrassingly simple! tagging @luisas for visibility

pditommaso:
what if accessing `trace` in the `script` context?

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -2384,6 +2388,8 @@ class TaskProcessor {
             collectOutputs(task)
         }
         catch ( Throwable error ) {
+            def trace = handler.getTraceRecord()
+            task.context.put('trace', trace)
```
pditommaso, line 2392: Not getting this. Why adding only in task context on there's an exception

jorgee: If I am not wrong, this is the case where the task is retried. In other cases it is going to finalize the task and I think there is no need to put the record in the context.

pditommaso: I see your point. My point is that if this is going to be available only on error condition, we may want to expose using a different context e.g. `error` that could bring more information about the error condition + the `trace`

jorgee:
> what if accessing `trace` in the `script` context?

I though it was working in a similarly as in the directives, but it fails with No such variable. I need to add a default trace value, as null or empty TraceRecord. In the first case, users case easily check if there is no previous trace as in the example. @pditommaso what do you think?

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
jorgee: Do you mean adding the another implicit variable called 'error' in the TaskContext and put as condition something like Throwable object apart form the 'trace'? Or include the 'trace' inside this 'error' variable?

docs/process.md
```diff
@@ -1283,6 +1283,24 @@ If the task execution fail reporting an exit status in the range between 137 and
...
 The directive {ref}`process-maxretries` set the maximum number of time the same task can be re-executed.
 
+### Dynamic task resources with previous execution trace
+:::{versionadded} 24.09.1-edge
+:::
+Similarly to the previous scenario, task resources can be also updated according to the metrics included in the {ref}`trace record <trace-report>` of the previous task attempt. The metrics can be accessed through the `trace` variable. For example:
+
+```groovy
+process foo {
+    memory { (trace != null) ? trace.memory * 2 : (1.GB) }
```
bentsherman, line 1293:
```suggestion
    memory { task.attempt > 1 ? trace.memory * 2 : (1.GB) }
```

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
jorgee: The only problem that I see of putting the property trace inside error is that it can be confused with the stack trace.  Maybe we could call the property 'traceRecord'

bentsherman: In reality I think I would prefer something like `task.previous` to make it clear that it is information about the previous attempt that only exists when attempt > 1:

```groovy
task.previous.exitStatus
task.previous.errorMessage
task.previous.trace
```

This also keeps everything contained within the implicit variable `task`

jorgee: I like it. @pditommaso, what do you think?

pditommaso: `previous` implies an ordering which is something it cannot be guaranteed. Maybe `task.last` or `task.failure` ?

jorgee: I prefer `task.last`.  As I mentioned before, associating `trace` with `task.failure` or `error` could be interpreted as the stack trace of the error

pditommaso: thinking more, if this is going to be exposed on a task retry, I believe it would be better to use `task.failure` or a similar concept. `task.last` or `task.previous` makes me think it's available in any case

pditommaso: Maybe we are over engineering here. Considering those variables are already in the `task` scope, does not make much sense adding a new `last` context including again the `exitStatus`. 

What about just adding `lastTrace` as it's done for `exitStatus`, `retryCount`, etc?

bentsherman: That's fine too, but I would prefer `previousTrace` because `last` can mean either "previous" or "final", it often confuses people

jorgee:
I have done an update setting the trace record in `task.previousTrace` if you finally think it is better to have ` lastTrace` I can easily do. I have also included the access to the trace in the script context and updated the documentation.

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
bentsherman: Can we also include the error as e.g. `task.previousError` ? Or at least the error message, to not expose runtime details. Since the original use case is to determine the type of error that occurred, it seems like the error message would be just as important as the trace

jorgee: Sure!

modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy
```diff
@@ -617,6 +619,9 @@ class TaskPollingMonitor implements TaskMonitor {
             if (evict(handler)) { 
                 handler.decProcessForks()
             }
+            //Add trace of the previous execution in the task context for next execution
+            handler.task.config.put(TRACE_PROPERTY_NAME, handler.getTraceRecord())
+            handler.task.config.put(ERROR_PROPERTY_NAME, error.getMessage())
```
pditommaso, line 624: it would be better to keep this in the `resumeOrDie` method both for consistency and avoid code duplication   

https://github.com/nextflow-io/nextflow/blob/500f49c0911bb33e3fa0f765df4a20da07a8f761/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy#L1060-L1063

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -1036,6 +1042,7 @@ class TaskProcessor {
                 session.getExecService().submit {
                     try {
                         taskCopy.runType = RunType.RETRY
+
```
pditommaso, line 1045:
```suggestion
```

modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy
jorgee: I think no problem for the case of the error. I need to check for the trace record as we need the TaskHandler

jorgee: Done! I have included the traceRecord as optional parameter of the resumeOrDie and inside the function I modify task.config inside the function

docs/process.md
```diff
@@ -1283,6 +1283,25 @@ If the task execution fail reporting an exit status in the range between 137 and
 
 The directive {ref}`process-maxretries` set the maximum number of time the same task can be re-executed.
 
+### Dynamic task resources with previous execution trace
+:::{versionadded} 24.09.2-edge
```
pditommaso, line 1287:
```suggestion
:::{versionadded} 24.10.0
```

docs/reference/process.md
```diff
@@ -26,6 +26,18 @@ The following task properties are defined in the process body:
...
 : The current task name.
 
+`task.previousError`
+: :::{versionadded} 24.10.0
+  :::
+: The error message of the task execution.
+: Since the error message is only available after the task has been executed, it can only be used when `task.attempt` is bigger than 1.
+
+`task.previousTrace`
+: :::{versionadded} 24.09.2-edge
```
pditommaso, line 36:
```suggestion
: :::{versionadded} 24.10.0
```

docs/process.md
```diff
@@ -1283,6 +1283,25 @@ If the task execution fail reporting an exit status in the range between 137 and
...
+:::
+
+Similarly to the previous scenario, task resources can be also updated according to the metrics included in the {ref}`trace record <trace-report>` of the previous task attempt. The metrics can be accessed through the `task.previousTrace` variable. For example:
```
christopher-hakkaart, line 1290:
```suggestion
Task resource requests can be updated relative to the {ref}`trace record <trace-report>` metrics of the previous task attempt. The metrics can be accessed through the `task.previousTrace` variable. For example:
```

docs/process.md
```diff
@@ -1283,6 +1283,25 @@ If the task execution fail reporting an exit status in the range between 137 and
...
+process foo {
+    memory { task.attempt > 1 ? task.previousTrace.memory * 2 : (1.GB) }
+    errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
+    maxRetries 3
+
+    script:
+    <your job here>
+}
+```
+In the above example the {ref}`process-memory` is set according to previous trace record metrics. In the first attempt, when no trace metrics are available, it is set to one GB, while in the other attempts it is doubling the previously allocated memory. The available metrics in the trace record can be found in the {ref}`Trace Report section <trace-report>`.
```
christopher-hakkaart, line 1302:
```suggestion
In the above example, the {ref}`process-memory` is set according to previous trace record metrics. In the first attempt, when no trace metrics are available, it is set to one GB. In the subsequent attempts, it doubles the previously allocated memory. See {ref}`trace-report` for more information about trace records.
```

christopher-hakkaart (changes requested):
Suggestions to shorten sentences.

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -1061,6 +1063,10 @@ class TaskProcessor {
                 task.config.exitStatus = task.exitStatus
                 task.config.errorCount = procErrCount
                 task.config.retryCount = taskErrCount
+                //Add trace of the previous execution in the task context for next execution
+                if ( traceRecord )
+                    task.config.previousTrace = traceRecord
+                task.config.previousError = error.getMessage()
```
pditommaso, line 1069:
```suggestion
                task.config.previousException = error
```

would not make more sense expose the exception object at this point ?

jorgee: I think @bentsherman commented it https://github.com/nextflow-io/nextflow/pull/5402#discussion_r1805002690 . No sure if exposing the full object could imply a security vulnerability. I think the Exception type could be also nice to have

pditommaso: The exception object can be useful for task running native java/groovy code

bentsherman: Wouldn't you handle native task errors with a try / catch?

I agree security is not really an issue. I was thinking more about exposing "internal" errors to the user that are not useful or could change across releases

pditommaso: Definitively not handling exception, much more for collecting error info. Exception classes are useful to understand the type of error, and specific exception brings their own info, think for example http status code, with http client exceptions

bentsherman: Agreed, in nf-core you will see they typically use try-catch to handle things like http errors in helper functions

jorgee: Updated with changes in docs and test according to the modification

Issue #5360, opened 2024-10-03
FIxedInputStream raises false error when reading only part of a file


bentsherman:
`FixedInputStream` was added to detect truncated files:

https://github.com/nextflow-io/nextflow/blob/324de3d162e476ed3bf58a9c77ae8e95e27cb3ee/modules/nf-httpfs/src/main/nextflow/file/http/FixedInputStream.groovy#L67-L72

However this does not handle the case where only part of a file is read. Consider this example from sarek:

```groovy
// Get first line of a FASTQ file
def readFirstLineOfFastq(path) {
    def line = null
    try {
        path.withInputStream {
            InputStream gzipStream = new java.util.zip.GZIPInputStream(it)
            Reader decoder = new InputStreamReader(gzipStream, 'ASCII')
            BufferedReader buffered = new BufferedReader(decoder)
            line = buffered.readLine()
            assert line.startsWith('@')
        }
    } catch (Exception e) {
        log.warn "FASTQ file(${path}): Error streaming"
        log.warn "${e.message}"
    }
    return line
}
```

It will log this error every time:
```
Read data length does not match expected size - bytes read: 2756; expected: 16013759
```

cc @maxulysse

pditommaso:
Related to https://github.com/nextflow-io/nextflow/pull/5275

pditommaso:
Solved by https://github.com/nextflow-io/nextflow/pull/5405

PR #5405, opened 2024-10-15
Allow http file to be closed early

tom-seqera:
Proof of concept for #5360 

This uses Java's [OpenOption](https://docs.oracle.com/javase/8/docs/api/java/nio/file/OpenOption.html) mechanism to pass a configuration param to the `XFileSystemProvider`, allowing a script to declare that it doesn't care if the input stream is closed before being fully read:

```groovy
import nextflow.file.http.HttpOpenOption

def firstLine(path) {
    path.withInputStream(HttpOpenOption.ALLOW_EARLY_CLOSE) {
        def r = new BufferedReader(new InputStreamReader(it))
        return r.readLine()
    }
}
```

This approach is quite flexible and generic, since it would also allow other `OpenOptions` to be passed to other `FileSystemProviders`, and as a result might be a little in conflict with the formalisation of a nextflow script syntax. An alternative could be to add a single-purpose method to `XPath`, eg:

```groovy
InputStream newLenientInputStream() {
   ...
}
```

### DIFF
--- a/modules/nf-commons/src/main/nextflow/file/CopyMoveHelper.java
+++ b/modules/nf-commons/src/main/nextflow/file/CopyMoveHelper.java
@@ -41,6 +41,10 @@ import org.slf4j.LoggerFactory;
  * Helper class to handle copy/move files and directories
  */
 public class CopyMoveHelper {
+    /**
+     * True if currently performing a copy of a foreign file.
+     */
+    public static final ThreadLocal<Boolean> IN_FOREIGN_COPY = new ThreadLocal<>();
 
     private static Logger log = LoggerFactory.getLogger(CopyMoveHelper.class);
 
@@ -81,14 +85,16 @@ public class CopyMoveHelper {
     private static void copyFile(Path source, Path target, boolean foreign, CopyOption... options)
             throws IOException
     {
-
         if( !foreign ) {
             source.getFileSystem().provider().copy(source, target, options);
             return;
         }
 
+        IN_FOREIGN_COPY.set(true);
         try (InputStream in = Files.newInputStream(source)) {
             Files.copy(in, target);
+        } finally {
+            IN_FOREIGN_COPY.set(false);
         }
     }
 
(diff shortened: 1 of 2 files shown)


pditommaso:
The problem I see in this approach is that only works for `withInputStream`, there are plenty of other ways to open a stream. 

What about if we something much more crazy, for example when opening (or closing the connection) inspect the current stack trace and determine the request is coming from a file path operation.

pditommaso:
Ok, forget about the stack trace thing, because still does not solve the problem (you may still want to read partially a stream via a Path). 

I think instead, we can apply something very similar to what you are suggesting using a reverse logic, i.e. enforcing the size check only when file transfer operation are made (not for generic stream operations). 

All file copies are made by the `FilePorter` [here](https://github.com/nextflow-io/nextflow/blob/8d0e28270dd625a62e00ba4b04f9b3dd2408cb4b/modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy#L357-L364) . 

The [fileCopy](https://github.com/nextflow-io/nextflow/blob/8d0e28270dd625a62e00ba4b04f9b3dd2408cb4b/modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy#L364-L364) take some options that could be extend as you have implemented. 

What do you think?

tom-seqera:
I tried reversing the logic, so it only applies the check on file copy.

It generally seems to work, but the Jimfs handler doesn't like it very much, and throws a `java.lang.UnsupportedOperationException` [here](https://github.com/google/jimfs/blob/52d1285a324440ef46f1dce8abb3763ed7d426ff/jimfs/src/main/java/com/google/common/jimfs/Options.java#L103) - so we'd potentially have to add logic so the option is only added for http/ftp inputstreams

pditommaso:
If jimfs is blocking we could also consider something more dirty as a thread local flag

pditommaso:
@tom-seqera any eta or blocking issue for this ?

tom-seqera:
I managed to make it work by implementing `copy()` on the `XFileSystemProvider`. It's not the cleanest because it relies on provider exceptions to determine their supported operations. I'm not sure if it's better or worse than a thread local.

pditommaso:
Even if the OpenOption is a nice solution, I think it would be better to not change the copy implementation to make it work with a testing library. A thread local could be a better hack here

pditommaso:
Neat! likely "Ready for review" ?

PR #5406, opened 2024-10-17
In Google Batch, install GPU drivers for GPU VMs

### LINKED ISSUE #5372: Support accelerator optimized VMs in Google Cloud
## Usage scenario 

Google Cloud Batch has [accelerator optimized VMs](https://cloud.google.com/compute/docs/accelerator-optimized-machines), that need to be configured a little differently. They don't need to specify the accelerator type and count, but would still need the `installGpuDrivers` set to true. See [Google documentation](https://cloud.google.com/batch/docs/create-run-job-gpus#use-gpu-vms).

## Suggest implementation 

For [this line](https://github.com/nextflow-io/nextflow/blob/64bb5a9235f635650334496c1fad375c46658cdd/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy#L335C42-L335C62), we also need to check if the task's machine type is `g2-*`, `a2-*` or `a3-*` even if accelerator type and count were not set. I think this should be enough. Happy to send a PR.

siddharthab:
Also clean up some old logic for container options when using GPUs.
These are now automatically handled by Google Cloud.

Fixes #5372.


plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -324,17 +313,6 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
...
-                final accelerator = AllocationPolicy.Accelerator.newBuilder()
-                    .setCount( task.config.getAccelerator().getRequest() )
-
-                if( task.config.getAccelerator().getType() )
-                    accelerator.setType( task.config.getAccelerator().getType() )
-
-                instancePolicy.addAccelerators(accelerator)
-                instancePolicyOrTemplate.setInstallGpuDrivers(true)
-            }
-
```
pditommaso, line 337: Not sure this should be removed. Is this API still available on Google Batch ?

siddharthab: It's not removed, I just moved it to be after machineType logic. That way, setting accelerators can override the installGpuDrivers field set through machineType.

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy
```diff
@@ -249,4 +257,19 @@ class GoogleBatchMachineTypeSelector {
...
+     * @param machineType Machine type
+     * @return Boolean value indicating if GPU drivers should be installed.
+     */
+    protected boolean installGpuDrivers(MachineType machineType) {
+        if ( machineType.gpusPerVm > 0 ) {
+            return true
+        }
+        // Cloud Info service currently does not currently return gpusPerVm values (or the user
+        // could have disabled use of the service) so also check against a known set of families.
+        return ACCELERATOR_OPTIMIZED_FAMILIES.find { matchType(it, machineType.type) }
```
bentsherman, line 272:
```suggestion
        return ACCELERATOR_OPTIMIZED_FAMILIES.any { matchType(it, machineType.type) }
```

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -224,19 +224,8 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
...
-            container
-                .addVolumes('/var/lib/nvidia/lib64:/usr/local/nvidia/lib64')
-                .addVolumes('/var/lib/nvidia/bin:/usr/local/nvidia/bin')
-        }
-
         def containerOptions = task.config.getContainerOptions() ?: ''
-        // accelerator requires privileged option
-        // https://cloud.google.com/batch/docs/create-run-job#create-job-gpu
-        if( task.config.getAccelerator() || fusionEnabled() ) {
+        if( fusionEnabled() ) {
```
bentsherman, line 228: Why is this logic removed?

siddharthab: It does not seem to be needed anymore since Google does this for you automatically if `installGpuDrivers` bit is set. It is not mentioned in the official Google documentation as well.

The only thing is that people might still need to add `/usr/local/nvidia/bin` to their PATH and `/usr/local/nvidia/lib64` to their LD_LIBRARY_PATH depending on their app (some apps automatically look there anyway), but it is not happening right now as well.

bentsherman: Indeed, the library path should be handled by the container

bentsherman:
@siddharthab overall the changes make sense. I think we'll need to do some testing on our end to make sure the machine type selection still works with Fusion + SSD

Can you confirm that it works for you with some GPU-enabled tasks?

siddharthab:
Yes, I confirm that I tested the following configurations:

| GPU Directive | bootDiskImage | Fusion | Result |
| -------------- | --------------- | ------- | ------ | 
| machineType | batch-cos | No | PASS |
| accelerator | batch-cos | No | PASS |
| machineType | batch-cos | Yes | PASS† |
| accelerator | batch-cos | Yes | PASS |
| machineType | batch-debian | No | FAIL‡ |
| accelerator | batch-debian | Yes | FAIL‡ |

† Needed new logic in `findValidLocalSSDSize` for the larger machine types, now pushed as an update to this PR.

‡ It looks like `installGpuDrivers` works properly only when `bootDiskImage` is `batch-cos` (the default). With `batch-debian`, the installer aborted with the following in journalctl. This is independent of any changes made in this PR (mentioning here because it came up in my testing).
```
google_metadata_script_runner[846]: startup-script-url: [2024-10-09 19:40:44] Executing: lspci -n
google_metadata_script_runner[846]: startup-script-url:
google_metadata_script_runner[846]: startup-script-url: 00:00.0 0600: 8086:1237 (rev 02)
google_metadata_script_runner[846]: startup-script-url: 00:01.0 0601: 8086:7110 (rev 03)
google_metadata_script_runner[846]: startup-script-url: 00:01.3 0680: 8086:7113 (rev 03)
google_metadata_script_runner[846]: startup-script-url: 00:03.0 0000: 1af4:1004
google_metadata_script_runner[846]: startup-script-url: 00:04.0 0200: 1af4:1000
google_metadata_script_runner[846]: startup-script-url: 00:05.0 00ff: 1af4:1005
google_metadata_script_runner[846]: startup-script-url:
google_metadata_script_runner[846]: startup-script-url: There doesn't seem to be a GPU unit connected to your system. Aborting drivers installation.
```

siddharthab:
@bentsherman The test failures seem unrelated to this change. Please let me know if you would like to see anything more in this PR.

siddharthab:
@bentsherman Any thoughts?

bentsherman (approved):

siddharthab:
@pditommaso Bringing it to your attention in case you missed this PR. It is now waiting for your final approval.

pditommaso:
Testing against platform showcase pipelines

pditommaso:
ok, some tests fails because credentials are not added into foreign branches. let's merge anyway

PR #5408, opened 2024-10-17
Fix closure rendering in yaml and json config

pditommaso:
This PR fixed the rendering of closures in config file reporting as strings. 

For example 

```
{ 
   "process": {
        "memory": "{ 1.GB }"
    }
} 
``` 

or 

```
process:
  memory: '{ 1.GB }'
```

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/util/ConfigHelper.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/util/ConfigHelper.groovy
@@ -22,10 +22,10 @@ import groovy.json.JsonOutput
 import groovy.transform.CompileStatic
 import groovy.transform.PackageScope
 import groovy.util.logging.Slf4j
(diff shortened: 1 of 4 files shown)


modules/nextflow/src/main/groovy/nextflow/util/ConfigHelper.groovy
```diff
@@ -356,9 +356,32 @@ class ConfigHelper {
...
+
+    private static Object normaliseObject0(Object value) {
+        if( value instanceof Map ) {
+            final map = value as Map
+            final ret = new LinkedHashMap(map.size())
+            for( Map.Entry entry : map.entrySet() ) {
+                ret.put(entry.key, normaliseObject0(entry.value))
+            }
+            return ret
+        }
```
bentsherman, line 372: is there a case where you would have a closure in a list? I can't think of any

PR #5414, opened 2024-10-20
Remove invalid equals char

mr83-uw:
The equals character from nf-core modules/local/hicpro/dnase_mapping_stats.nf causes pipeline to crash on SGE executor.

Add "=" to list of invalid chars


pditommaso (commented):
Thanks for submitting this PR!

PR #5418, opened 2024-10-21
Make tests comply with strict syntax

bentsherman:
Follow up to #5336 , applying the new formal syntax also to e2e tests


bentsherman:
Closing in favor of #4744 and #4613

PR #5419, opened 2024-10-21
Document `subscribe` as a teminal operator

### LINKED ISSUE #5354: Operator `subscribe` returns wrong results when chained and hangs at the end

When `subscribe` is chained with another operator like `view`, some elements from the channel go to subscribe and some to `view`, but never both. As per the documentation, `subscribe` should return the source channel, thereby all elements should flow through all operators in the chain. The execution hangs after, which is a good thing in this case, because it prompts the user to look closer. Otherwise, the results would be incorrect.

```groovy
workflow {
    Channel
        .of('foo', 'bar', 'baz')
        .subscribe { println "Subscribe $it" }
        .view { println "View $it" }
}
```

### Program output 

```
View bar
Subscribe foo
Subscribe baz
^C # Had to manually kill the process
```

### Environment 

* Nextflow version: 24.04.4 build 5917
* Java version: OpenJDK Runtime Environment Zulu21.28+85-CA (build 21+35)
* Operating system: macOS

### Additional context

Seems to be a duplicate of #4436, but I wanted to call out the unexpected branching behavior.

bentsherman:
Close #5354 

Clarify that `subscribe` should be used only as a terminal operator.

We could also update the `subscribe()` API to return `void`, but that could break some existing pipeline. On the other hand, from the issue it sounds like it already causes the pipeline to hang, so perhaps it would be better to change the API so that there is a runtime error.

Doesn't matter much to me because I have plans for a "v2" operator library anyway

### DIFF
--- a/docs/reference/operator.md
+++ b/docs/reference/operator.md
@@ -1426,7 +1426,7 @@ The `subscribe` operator supports multiple types of event handlers:
 ```
 
 :::{note}
-Unlike most operators, `subscribe` does not emit any values -- it is a *terminal operator*. Instead, `subscribe` should be used only for *side effects* such as printing to the console, writing to a file, or making HTTP requests.
+Unlike most operators, `subscribe` is a *terminal operator* and does not emit any values. It should only be used for *side effects*, such as printing to the console, writing to a file, or making HTTP requests.
 :::
 
 Available options:
--- a/docs/reference/operator.md
+++ b/docs/reference/operator.md
@@ -1393,7 +1393,7 @@ See also: [countLines](#countlines)
 
 ## subscribe
 
-*Returns: the source channel*
+*Returns: nothing*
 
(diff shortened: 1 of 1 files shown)


docs/reference/operator.md
```diff
@@ -1425,6 +1425,10 @@ The `subscribe` operator supports multiple types of event handlers:
 :language: console
 ```
 
+:::{note}
+Unlike most operators, `subscribe` does not emit any values -- it is a *terminal operator*. Instead, `subscribe` should be used only for *side effects* such as printing to the console, writing to a file, or making HTTP requests.
```
christopher-hakkaart, line 1429:
```suggestion
Unlike most operators, `subscribe` is a *terminal operator* and does not emit any values. It should only be used for *side effects*, such as printing to the console, writing to a file, or making HTTP requests.
```
Suggestion

pditommaso (approved):

PR #5423, opened 2024-10-23
Use `conda-forge` and `bioconda` channels by default

ewels:
If `conda.channels` is not specified, set it to `conda-forge`,`bioconda` by default.

Quick edit, haven't adjusted tests or anything here, just wanted to raise for discussion.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/conda/CondaConfig.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/conda/CondaConfig.groovy
@@ -49,7 +49,7 @@ class CondaConfig extends LinkedHashMap {
     List<String> getChannels() {
         final value = get('channels')
         if( !value ) {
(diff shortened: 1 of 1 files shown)


ewels:
@pditommaso @bentsherman - I resurrected this old PR. Should be good to go now if we're happy with the behaviour?

bentsherman:
Can you remind me what's the reason for changing the defaults?

ewels:
As I remember it, there were two main reasons:

* We want simple package names to "just work". I don't _think_ that [this example](https://www.nextflow.io/docs/latest/conda.html#use-conda-package-names) in the docs works out of the box, for example, as it needs additional configuration in Nextflow to add the bioconda channel (pls correct me if I'm wrong)
* We don't want people to use the `defaults` channel by accident. By always providing channels to conda we should avoid this.

PR #5431, opened 2024-10-24
Document standard library types

bentsherman:
This PR is an initial attempt to flesh out the reference docs for standard types like lists, maps, sets, and strings. These types are used often in Nextflow, but users have difficulty using their APIs because we mostly delegate to the Groovy docs for these things.

Some notes:

- The full set of Groovy/Java APIs is HUGE, but most of them aren't needed in Nextflow or are redundant. I have tried to pick out a minimal set of types and methods that cover most use cases. We can always add more things, but every new thing should be justified.

- For now I'm just using Java types exactly as they are called. I doubt we'll have the exact same type nomenclature, but I think it will be similar enough that we can evolve it from Java as a starting point.

- Many of the Groovy APIs are not fully typed, especially with generics (e.g. list element type, map key/value types, closure param/return types). The Java APIs are (mostly) fully typed but I haven't bothered with generic types yet (e.g. for List / Map / Set) because it would just get ugly. I think this is fine for now, we can revisit when we build out the statically typed std lib.

  On that note, we'll likely want to swap out some Groovy APIs with our own implementations that are fully typed. This goes back to my first point about only taking what we need from the Groovy APIs.

- I really don't want to expose type hierarchies in Nextflow, but I have made one exception here for "Iterable", because there are three types of collections (Bag, List, Set) and they have mostly the same methods with some small variations. So it's easier to expose Iterable as a sort of shared interface or trait. I think this is fine, we'll need it anyway for things like `flatMap(splitCsv)`

- I have taken care to not add APIs that can lead to non-deterministic behavior (e.g. getting the "first" element of a set). Any collection methods that rely on order are only documented for List, and any unordered collection must go through `Iterable::toList()` to access these "ordered" methods. Note the big fat "danger" notice on `toList()`.


bentsherman:
@pditommaso @ewels this PR is a first iteration to document the "standard library" for Nextflow, so that people don't have to go digging through the Groovy docs to figure out how to use things like lists, maps, strings, etc

Would like to get this merged soon. I have documented methods that I feel confident should be part of the standard library, and we can always document more as needed

docs/script.md
```diff
@@ -106,22 +109,106 @@ When adding two maps, the first map is copied and then appended with the keys fr
...
+The `==` and `!=` operators can be used to test whether any two values are equal (or not equal):
+
+```nextflow
+assert 2 + 2 == 4
+assert [2, 2] != [4]
+assert 'two plus two' != 'four'
+```
+
+:::{tip}
+The `assert` keyword simply tests a condition and raises an error if the condition is false. Every assert that you see on this page should succeed if you execute it.
```
christopher-hakkaart, line 133:
```suggestion
The `assert` keyword simply tests a condition and raises an error if the condition is false. Every assert that you see on this page will succeed if executed.
```

bentsherman: I like that confidence 😅

christopher-hakkaart (approved):
Language looks good. The stdlib page is very long and it would be good to split it up, but we can leave that for another PR.

bentsherman:
@ewels do these docs make sense to you from a user perspective?

docs/reference/stdlib.md
```diff
@@ -194,111 +187,121 @@ The following constants are globally available in a Nextflow script for workflow
...
+```nextflow
+def vals = [1, 2, 3]
+println groovy.json.JsonOutput.toJson(vals)
+```
 
-(stdlib-types-duration)=
+The following sections describe the standard types provided by Nextflow.
 
-## Duration
+## Bag\<E\>
```
pditommaso, line 275: This is an internal structure, not sure it should be documented as public API

docs/reference/stdlib.md
```diff
@@ -322,30 +335,293 @@ assert a * 2 == b
 assert b / 2 == a
 ```
 
-The following methods are available for a `Duration` object:
+The following methods are available for a Duration:
 
-`getDays()`, `toDays()`
+`toDays() -> long`
```
pditommaso, line 340: Getter should be preferred because they can be accessed also as attributes

docs/reference/stdlib.md
bentsherman: It will need to be part of the public API to represent things like the values of a map or the return type of the `collect` operator

The standard collections are:
- Bag: unordered, duplicates allowed
- List: ordered, duplicates allowed
- Set: unordered, duplicates not allowed

docs/reference/stdlib.md
bentsherman: The getters like `getMillis()` are confusing because it's not clear if you're getting the millisecond representation of the whole duration or just the millisecond *component*, which is how it works in [JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMilliseconds) and [Python](https://docs.python.org/3/library/datetime.html#datetime.datetime.second). Whereas `toMillis()` is clear that it's the former

bentsherman: I added them back in a separate note so that it's clear we haven't removed them.

pditommaso (approved):

docs/reference/stdlib.md
pditommaso: Okey

Issue #5420, opened 2024-10-22
feature request: Make "/bin/bash" configurable


stianlagstad:
I would like to make `/bin/bash` configurable, either through setting an environment variable or through setting a configuration value. Would a PR for this be welcome? Is there anyone I should discuss this with first? Or are there things I should know before I start?

Prior work and comments on this:
- Issue: https://github.com/nextflow-io/nextflow/issues/1598. The issue suggests to use `/usr/bin/env bash` instead of `bin/bash`, but was rejected. Note the comment from Paolo: "The only solution that I can see is to make the shell path configurable via config setting (env var or config file) which defaults to /bin/bash and the user can change as needed."
- PR: https://github.com/nextflow-io/nextflow/pull/2341. A PR which suggested to change `/bin/bash` to `/usr/bin/env bash`. This was not accepted, and closed, due to the previous discussion in the above link.

My main motivations for this idea are:
1. To be able to use nextflow on systems where `/bin/bash` is not available, and where users can't easily setup a symlink or similar.
2. Get more reproducible workflows, as we can control which version of `bash` is used.
3. To avoid the many duplicate hardcoded strings "/bin/bash" present in the nextflow repository. Running the command `find . -name "*.groovy" -type f -exec grep -o "/bin/bash" {} + | wc -l` on the current master branch gives 106 results. If there were (many) fewer of these, then other packaging tools would be able to package nextflow and write a small patch which changes the occurrences of these strings.

Any input is appreciated. Thank you!

bentsherman:
Excluding tests, there are about 15 hardcoded references to `/bin/bash`, most of them related to launching the `.command.run` script on a particular executor

I believe the launcher script has certain assumptions baked in around `bash`, so it would make sense to point to a different version of bash but not necessarily a different shell

It could make sense as an environment variable or a config option such as `executor.bash`

PR #1614, opened 2020-05-29
Use current environment bash (#1598)

Emiller88:
Signed-off-by: Edmund Miller <edmund.a.miller@protonmail.com>

Close #1598 

Open to discussing the fix, just looking to help out and get involved!

PR #5432, opened 2024-10-25
Make Nextflow compatible with NixOS (or other systems without `/bin/bash`)

rollf:
UPDATE: After having created this PR I realized that I could do better (see comments below) and have also invested more time to understand the code base. Now, I (hopefully) have a version that fits well and is satisfying for everybody. Please refer to all the comments below before looking into the diff.

---
ORIGINAL PR description:
(This comment was made when 29df9b71a587c34e5e6248de4f972692193d6227 was HEAD.)

This PR allows to use `nextfow` on NixOS where `/bin/bash` does not exist. This is achieved by invoking `/usr/bin/env -S bash` where necessary. This approach (or similar) has been discussed previously:

* https://github.com/nextflow-io/nextflow/pull/2341
* https://github.com/nextflow-io/nextflow/issues/1598
* https://github.com/nextflow-io/nextflow/pull/1614
* https://github.com/nextflow-io/nextflow/issues/5420

However, compared the previous discussions/implementations, I specifically do *not* propose to change `/bin/bash` unless really necessary. By "necessary" I mean that `bash` is invoked on a local host system where it may not be available. Namely, when `<task>/.command.run` shall be invoked (always on the host) and when `<task>/.command.sh` is invoked in a non-container context (i.e. also on the host). I have *not* touched any other occurrences of `/bin/bash`, esp. not within the container.

My initial implementation used `which bash` instead if `/usr/bin/env -S bash` but I esteemed the latter to be more correct. Please refer to the git commit history/messages for some further insights.

I understand that it is possible for downstream consumers to work around `/bin/bash` (symlink was proposed in one of the linked issues above). However, I believe that my proposal/approach (=assume `/bin/bash` to be available everywhere except on the local host where `nextflow` is invoked) here is not too intrusive.

I'm happy  to change the the code as requested. I have not added any comments since I expect (at least) change requests. Please review this with an unbiased mind set. Multiple people have expressed their intention to use `nextflow` on systems without `/bin/bash`.

I have tested solely on NixOS.

--- 
Further context:

[Here](https://github.com/NixOS/nixpkgs/issues/350183) is the corresponding nixpkgs issue that motivates this PR.

[Here](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/interpreters/nextflow/default.nix#L36) is the current package definition for Nextflow in NixOS which causes trouble (see nixpkgs issue above).

[Here](https://github.com/rollf/nixpkgs/commit/ef7881768e7b8cebc63cbfb81ae389a0b37fe6df) is a new package definition assuming Nextflow would be compatible with NixOS out of the box.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -63,7 +63,11 @@ class BashWrapperBuilder {
 
     static final private ENDL = '\n'
 
+    static final private String LOCAL_BASH_BINARY = "which bash".execute().text.strip()
+
     static final public List<String> BASH
+
+    static final public List<String> HOST_BASH
 
     static private int level
 
@@ -85,6 +89,7 @@ class BashWrapperBuilder {
             log.warn "Invalid value for `NXF_DEBUG` variable: $str -- See http://www.nextflow.io/docs/latest/config.html#environment-variables"
         }
         BASH = Collections.unmodifiableList( level > 0 ? ['/bin/bash','-uex'] : ['/bin/bash','-ue'] )
+        HOST_BASH = Collections.unmodifiableList([LOCAL_BASH_BINARY, BASH[1]])
 
     }
 
@@ -292,7 +297,8 @@ class BashWrapperBuilder {
         /*
          * fetch the script interpreter i.e. BASH, Perl, Python, etc
          */
-        final interpreter = TaskProcessor.fetchInterpreter(script)
+        String interpreter = TaskProcessor.fetchInterpreter(script)
+        interpreter = containerBuilder ? interpreter : interpreter.replace('/bin/bash', LOCAL_BASH_BINARY)
 
         /*
          * append to the command script a prolog to capture the declared
@@ -571,7 +577,7 @@ class BashWrapperBuilder {
         final traceWrapper = isTraceRequired()
         if( traceWrapper ) {
             // executes the stub which in turn executes the target command
-            launcher = "/bin/bash ${fileStr(wrapperFile)} nxf_trace"
+            launcher = "${interpreter} ${fileStr(wrapperFile)} nxf_trace"
         }
         else {
             launcher = "${interpreter} ${fileStr(scriptFile)}"
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -575,7 +575,8 @@ class BashWrapperBuilder {
         final traceWrapper = isTraceRequired()
         if( traceWrapper ) {
             // executes the stub which in turn executes the target command
-            launcher = "/usr/bin/env -S bash ${fileStr(wrapperFile)} nxf_trace"
+            String traceInterpreter = runWithContainer ? "/bin/bash" : "/usr/bin/env bash"
+            launcher = "${traceInterpreter} ${fileStr(wrapperFile)} nxf_trace"
         }
         else {
             launcher = "${interpreter} ${fileStr(scriptFile)}"
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -295,8 +295,7 @@ class BashWrapperBuilder {
         /*
          * fetch the script interpreter i.e. BASH, Perl, Python, etc
          */
-        String fetchedInterpreter = TaskProcessor.fetchInterpreter(script)
-        final interpreter = runWithContainer ? fetchedInterpreter : fetchedInterpreter.replace('/bin/bash', '/usr/bin/env -S bash')
+        final interpreter = TaskProcessor.fetchInterpreter(script)
 
         /*
          * append to the command script a prolog to capture the declared
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -63,11 +63,9 @@ class BashWrapperBuilder {
 
     static final private ENDL = '\n'
 
-    static final private String LOCAL_BASH_BINARY = "which bash".execute().text.strip()
-
     static final public List<String> BASH
 
-    static final public List<String> HOST_BASH
+    static final public List<String> ENV_BASH
 
     static private int level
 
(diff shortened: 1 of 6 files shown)


rollf:
On another thought, iff `/usr/bin/env -S bash` is more acceptable than I think, one could be more intrusive, i.e. use it "everywhere" (where `BASH` is used) but make sure it is not used within container contexts. This seems to work for me and here would be another [commit](https://github.com/rollf/nextflow/commit/4e0176f034d34217e3ce6e1de17ce6a77c9d41d0) based on the changes in this PR (i.e. partially revert/invert them).

Either way, I feel it's always a bit ugly to use `/usr/bin/env -S bash` on the host and `/bin/bash` in the container because the way Nextflow is designed currently forwards the "host" interpreter to the container so one has to "manually" make sure `/bin/bash` is used in the container if `/usr/bin/env -S bash` got forwarded (but not for tracing where the initial invocation happens on the host, even for container contexts!). I'm not saying the design is wrong here, I just don't like the invocation of `.replace()` (see changes).

rollf:
Ah, I realize that I have used (for testing) `quay.io/nextflow/bash` as a docker image which actually *has* `/usr/bin/env` so the test was flawed. Playing around with this [image](https://github.com/bgruening/docker-busybox-bash) yields problem when tracing (because `/usr/bin/env` is not available). Good that I kept this PR as a draft then ^^. I'll update when I fixed the issue.

modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
```diff
@@ -65,6 +65,8 @@ class BashWrapperBuilder {
 
     static final public List<String> BASH
```
rollf, line 66: I suggest to rename `BASH` to `BIN_BASH` to make the difference clear.

modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
```diff
@@ -571,7 +574,8 @@ class BashWrapperBuilder {
         final traceWrapper = isTraceRequired()
         if( traceWrapper ) {
             // executes the stub which in turn executes the target command
-            launcher = "/bin/bash ${fileStr(wrapperFile)} nxf_trace"
+            String traceInterpreter = runWithContainer ? "/bin/bash" : "/usr/bin/env bash"
+            launcher = "${traceInterpreter} ${fileStr(wrapperFile)} nxf_trace"
```
rollf, line 578: No `-S` here becuase `bash` does not get any further options (think `-ue`)

modules/nextflow/src/main/groovy/nextflow/executor/local/LocalTaskHandler.groovy
```diff
@@ -131,7 +131,7 @@ class LocalTaskHandler extends TaskHandler implements FusionAwareTask {
     }
 
     protected ProcessBuilder localProcessBuilder() {
-        final cmd = new ArrayList<String>(BashWrapperBuilder.BASH) << wrapperFile.getName()
+        final cmd = new ArrayList<String>(BashWrapperBuilder.ENV_BASH) << wrapperFile.getName()
```
rollf, line 134: This makes sure that `.command.run` gets executed correctly on a host without `/bin/bash`.

modules/nextflow/src/main/groovy/nextflow/processor/TaskBean.groovy
```diff
@@ -133,7 +133,7 @@ class TaskBean implements Serializable, Cloneable {
         this.condaEnv = task.getCondaEnv()
         this.spackEnv = task.getSpackEnv()
         this.moduleNames = task.config.getModule()
-        this.shell = task.config.getShell() ?: BashWrapperBuilder.BASH
+        this.shell = task.config.getShell(task.isContainerEnabled()) ?: BashWrapperBuilder.BASH
```
rollf, line 136: At this point, `task` will always (?) have a `shell` since it comes with the process defaults. I have not toched the "else"-block here since I think it is actually not necessary (and obviously did not cause issues for me).

modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
```diff
@@ -375,10 +375,10 @@ class TaskConfig extends LazyMap implements Cloneable {
         return (List<String>) get('secret')
     }
 
-    List<String> getShell() {
+    List<String> getShell(boolean isContainerEnabled = true) {
```
rollf, line 378: I'm happy to adapt the unit tests but I don't know how :shrug: .

rollf:
(This comment was made when 1d9839e92f1509618cfd360fdfd410c60a2f57c2 was HEAD.)

After having created this PR I realized that I could do better (see comments below) and have also invested more time to understand the code base. Now, I (hopefully) have a version that fits well and is satisfying for everybody. In the end, my implementation is more intrusive that I initially thought but it feels right now. The final approach now is this: Right from the process/task (object) creation, consider whether the process/task will run inside a container. If not, use `/usr/bin/env -S bash`, otherwise keep `/bin/bash`. This, plus 2 minor tweaks for the execution of `.command.run` and for tracing allows to run `nextflow` on systems without `/bin/bash`.

I keep the original comments (below) /git history for documentation. I hope it is not too confusing.

Again, I'm happy to change the code as requested, manipulate the git history, etc. I have also added some review comments which include suggestions for improvements.

I have [improved](https://github.com/NixOS/nixpkgs/compare/master...rollf:nixpkgs:nextflow-trace) the nextflow-tests for NixOS. You may find some more details there, too. I made sure to test with a docker image that only has `/bin/bash` available. I did not find any quay.io image (I have roughly ~20 locally) that does not provide `/usr/bin/env`. I am *not* advocating for using `/usr/bin/env` inside the container. I mentioned this here so that it is clear why I specifically build a new image for the NixOS tests.

```
IMAGES=$(docker image ls | grep quay.io | awk -v OFS='' '{ print $1,":",$2 }')
for IMAGE in $IMAGES ; do echo $IMAGE ; docker run $IMAGE /usr/bin/env bash; done
```

bentsherman:
Collecting my thoughts here for now:

- this PR would make the host environment -- i.e. local non-container tasks -- always use `/usr/bin/env bash`
- `/usr/bin/env bash` is proven to work in the host environment, but not in containers, since `env` might not be available
- #4292 would make the `shell` directive apply to all bash commands in the task environment
- therefore one solution could be to merge #4292 and use this config to support NixOS:
```groovy
process.shell = ['/usr/bin/env', '-S', 'bash']
```

rollf:
@bentsherman Thanks for looking into this! We have touched the same code so it makes sense to coordinate. I like your suggestion. If #4292 is merged, I'll adapt my changes accordingly (and then, as you suggested, adapt the NixOS package itself or maybe see if any other mean is possible).

ewels:
@pditommaso will look into any additional dependencies on `/bin/bash` in the Nextflow ecosystems (Wave, Fusion, and others).

rollf:
I'm closing this PR in favor of #5684 .

PR #5434, opened 2024-10-25
Fix nf-tower plugin to upload logs when early failures 

### LINKED ISSUE #5433: Tower plugin is not pushing logs when error on validation
## Bug report 

When there is an error on validation (using nf-validation plugin) and cloudcache is enabled, the nf-tower plugin is not uploading the logs to the cloud cache. 

nf-tower plugin should upload the logs when workflow fails

Run a workflow with nf-validation plugin and nf-cloudcache such as the following:

NXF_CLOUDCACHE_PATH=az://xxxx
nexflow run https://github.com/adamrtalbot/nf-experiments -with-tower -r azure_seqera_platform_dies_no_logs

### Program output 

(Copy and paste here output produced by the failing execution. Please highlight it as a code block. Whenever possible upload the `.nextflow.log` file.)

### Environment 
* Nextflow version: 24.04.0 
* Java version: any
* Operating system: Linux
* Bash version: any

jorgee:
close #5433 
- Moving the start of LogsCheckpoit thread from onWorkflowBegin to onWorkflowCreate
- Implementing onWorkflowError

### DIFF
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsCheckpoint.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsCheckpoint.groovy
@@ -21,7 +21,9 @@ import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
 import nextflow.Session
 import nextflow.SysEnv
+import nextflow.processor.TaskHandler
 import nextflow.trace.TraceObserver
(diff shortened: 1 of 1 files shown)


bentsherman:
Good catch. While the `workflow.onComplete` handler is called regardless of success or error, the `onFlowError` hook in the trace observer is only called on success.

PR #5435, opened 2024-10-25
Use semver to pin only major/minor plugin versions

### LINKED ISSUE #4329: Plugins: Use any available cached version if `NXF_OFFLINE` is true

tom-seqera:
Closes #4329 

This makes two changes:
1. Allows using SemVer version constraints to specify plugin versions
2. Presents the directory of already downloaded plugins as an `UpdateRepository`, allowing the same SemVer version constraint to be applied offline as well.

The version range constraints are handled by the underlying SemVer library: https://github.com/zafarkhaja/jsemver?tab=readme-ov-file#range-expressions. Perhaps most usefully are tilde ranges (`~`), which allow pinning of the major or minor part of the version while automatically using any newer patch version. For example:
* `~1.2` is equivalent to `>=1.2.0 && <1.3.0`,
* `~1.2.1` is equivalent to `>=1.2.1 && <1.3.0`,
* `~1` is equivalent to `>=1.0.0 && <2.0.0`

In offline mode, it will look for the latest already downloaded plugin version which matches the constraint.


bentsherman:
At a first glance the changes make sense. Will need to try a manual test with some real plugins, ideally a published plugin with multiple major versions, and then review the code in more detail

@tom-seqera do you feel good about this approach? I thought you said you had some concerns about the local plugin provider but now I don't remember

tom-seqera:
> @tom-seqera do you feel good about this approach? I thought you said you had some concerns about the local plugin provider but now I don't remember

It's been a little while, but I think I found that some of the p4fj code (or maybe our wrapper code around it?) seems to generally assume that plugin update versions for a given plugin ID are only defined in a single repository and might not work properly if you tried to enable both the local cache and remote index at the same time.

It doesn't impact these specific changes because the code selects either the local cache _or_ the remote index depending on the offline flag, but will likely be something we need to handle if/when we harmonise plugin loading or improve support for multiple plugin sources.

bentsherman:
I tested the changes with some recent versions of [nf-amazon](https://github.com/nextflow-io/plugins/blob/99e96f6f6f041d5ef549a0f51024633d9705cb53/plugins.json#L509-L536):

```bash
make pack
build/releases/nextflow-25.01.0-edge-dist run hello -plugins nf-amazon@~2.9.0
```

It correctly downloads `nf-amazon@2.9.2`

Running again in offline mode:
```bash
NXF_OFFLINE=true build/releases/nextflow-25.01.0-edge-dist run hello -plugins nf-amazon@~2.9.0
```

It correctly uses `nf-amazon@2.9.2` from the local cache.

bentsherman (approved):
LGTM

@pditommaso ready to merge, let us know if you have any concerns

bentsherman:
I feel confident with these changes, so let's go ahead with this. That way we can play with it in the next edge release

pditommaso:
I wonder from where the `~` come from? Is inherited from pf4j?

bentsherman:
It comes from [jsemver](https://github.com/zafarkhaja/jsemver) which is used by pf4j

pditommaso:
I see. Bit annoying we have a different syntax for the same thing for nextflow version 

https://github.com/nextflow-io/nextflow/blob/130a9c819294222ffe8e233d15c1c30554cc993a/modules/nf-commons/src/test/nextflow/util/VersionNumberTest.groovy#L132-L141

tom-seqera:
The `~` syntax is a common convention for semver ranges. I think it originated in NPM but is now supported by semver implementations in multiple languages.

pditommaso:
Not against about it, I find inconvenient the inconsistency with the nextflow notation

bentsherman:
It might be for the best since the Nextflow version is not a semantic version

bentsherman:
In fact, I don't think we need the `+` at all for Nextflow versions. You really only need two forms:

```groovy
'24.10' // use stable release with latest patch
'24.10.2' // use specific patch release
```

And of course using `>=` or `<=` and so on to control the range. You don't really need the `+` because you could just not specify the patch number.

PR #5438, opened 2024-10-28
Document VS Code extension

bentsherman:
- Add new VS Code page
- Add custom pygments lexer for nextflow code
- Update code blocks to use nextflow lexer


docs/reference/channel.md
```diff
@@ -6,7 +6,7 @@
 
 ## empty
 
-The `channel.empty` factory method, by definition, creates a channel that doesn't emit any value.
+The `Channel.empty` factory method, by definition, creates a channel that doesn't emit any value.
```
pditommaso, line 9: Can this be avoided

bentsherman: This was just added for consistent syntax highlighting until we agree on which one to use across the board

pditommaso: yeah, but it's not nice to change it continuously in the docs. It was `Channel` originally, then become `channel` and now `Channel` again. Think it should stay `channel` at least until there's no consensus

pditommaso: 🙏

pditommaso (approved):

PR #5473, opened 2024-11-05
Fixing bug when execution with stub and no stub defined

jorgee:
When migrating resolve evaluation from TaskProcessor to TaskRun it was not checking if stub block was not defined.

Adding stub check and including the corresponding unit test.


modules/nextflow/src/main/groovy/nextflow/processor/TaskRun.groovy
```diff
@@ -790,7 +790,7 @@ class TaskRun implements Cloneable {
      * @param body A {@code BodyDef} object instance
      */
     void resolve(BodyDef body)  {
-        processor.session.stubRun
+        (processor.session.stubRun && config.getStubBlock())
```
pditommaso, line 793:
```suggestion
        processor.session.stubRun && config.getStubBlock()
```

pditommaso (approved):

PR #5474, opened 2024-11-05
Remove errorStrategy check in getMaxRetries to avoid possible deadloc…

### LINKED ISSUE #5464: Setting default maxRetries and errorStrategy and override for some processes
## Bug report 

(Please follow this template replacing the text between parentheses with the requested information)

I want to set errorStrategy and maxRetries in the nextflow config, and then for some processes to change the maxRetries with the aim that the errorStrategy should adjust accordingly

main.nf:
```
process A {
    maxRetries 4
    input:

    script:
    """
    exit 1
    """
}

process B {
    input:

    script:
    """
    exit 1
    """
}

workflow  {
    A()
    B()
}
```

nextflow.config:
```
profiles {
    local {
        process {
            executor = 'local'
            errorStrategy = {
                sleep(5 * Math.pow(2, task.attempt) as long)
                task.attempt <= process.maxRetries ? 'retry' : 'ignore'
            }
            maxRetries = 1
        }
    }

    other {
        process {
            maxRetries = 4
            errorStrategy = {
                sleep(60 * Math.pow(2, task.attempt) as long)
                task.attempt <= process.maxRetries ? 'retry' : 'terminate'
            }
        }
    }
}
```

### Program output 

The process maxRetries is ignored by the errorStrategy.

```
ubuntu@mcolpus-main:~/pipelines/nextflow_test_max_retries$ nextflow main.nf -profile local

 N E X T F L O W   ~  version 24.10.0

Launching `main.nf` [stupefied_ramanujan] DSL2 - revision: d107b1c91a

executor >  local (4)
[8e/c5cee3] A [100%] 2 of 2, failed: 2, retries: 1 ✔
[14/7baa91] B [100%] 2 of 2, failed: 2, retries: 1 ✔
[16/71c57f] NOTE: Process `B` terminated with an error exit status (1) -- Execution is retried (1)
[2d/4487c6] NOTE: Process `A` terminated with an error exit status (1) -- Execution is retried (1)
[8e/c5cee3] NOTE: Process `A` terminated with an error exit status (1) -- Error is ignored
[14/7baa91] NOTE: Process `B` terminated with an error exit status (1) -- Error is ignored
```

If the process error strategy is lower than the configs default then it will fail and not be ignored.

I also tried used `task.maxRetries` instead, but that crashed.
[nextflow.log](https://github.com/user-attachments/files/17619114/nextflow.log)

### Environment 

* Nextflow version: 24.10.0
* Java version: 17.0.12 2024-07-16
* Operating system: linux
* Bash version: 5.1.16(1)-release

jorgee:
close #5464
 
Remove errorStrategy check in getMaxRetries method. It avoids possible deadlock when using task.maxRetries in dynamic errorStrategy directives.

Default maxRetries is set to 1

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
@@ -345,7 +345,7 @@ class TaskConfig extends LazyMap implements Cloneable {
 
     int getMaxRetries() {
         def result = get('maxRetries')
-        result != null ? result as int : 1
+        result ? result as int : 1
     }
 
     int getMaxErrors() {
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
@@ -345,8 +345,7 @@ class TaskConfig extends LazyMap implements Cloneable {
 
     int getMaxRetries() {
         def result = get('maxRetries')
-        def defResult = getErrorStrategy() == ErrorStrategy.RETRY ? 1 : 0
-        result ? result as int : defResult
+        result != null ? result as int : 1
     }
 
     int getMaxErrors() {
--- a/modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy
@@ -126,7 +126,7 @@ class ProcessConfig implements Map<String,Object>, Cloneable {
             debug: false,
             cacheable: true,
             shell: BashWrapperBuilder.BASH,
-            maxRetries: 0,
+            maxRetries: 1,
             maxErrors: -1,
             errorStrategy: ErrorStrategy.TERMINATE
     ]
@@ -319,6 +319,7 @@ class ProcessConfig implements Map<String,Object>, Cloneable {
     BaseScript getOwnerScript() { ownerScript }
 
     TaskConfig createTaskConfig() {
+
         return new TaskConfig(configProperties)
     }
 
--- a/modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy
@@ -319,7 +319,6 @@ class ProcessConfig implements Map<String,Object>, Cloneable {
     BaseScript getOwnerScript() { ownerScript }
 
     TaskConfig createTaskConfig() {
-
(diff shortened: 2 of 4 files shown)


pditommaso:
A test is failing here 

```
~ Test 'stub-retry.nf' run failed
   + set -e
   + echo ''
   
   + /home/runner/work/nextflow/nextflow/nextflow -q run ../../stub-retry.nf -stub
   + tee stdout
   ERROR ~ Error executing process > 'stubtest'
   
   Caused by:
     Missing output file(s) `*.txt` expected by process `stubtest`
   
   
   Command executed:
   
     echo "Stubbing. Not creating file"
   
   Command exit status:
     0
   
   Command output:
     Stubbing. Not creating file
   
   Work dir:
     /home/runner/work/nextflow/nextflow/tests/checks/scratch/92/85bf921c434bf3554b86f5eb4fa660
   
   ++ grep INFO .nextflow.log
   ++ grep -c 'Submitted process > stubtest'
   + [[ 1 == 1 ]]
   ++ grep INFO .nextflow.log
   ++ grep -c 'Re-submitted process > stubtest'
   + [[ 0 == 1 ]]
   + false
```

jorgee:
Yes, It is failing because maxRetries is 0. I need to check why it is returning 0 instead of the default one. I have converted yo draft

modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
```diff
@@ -345,8 +345,7 @@ class TaskConfig extends LazyMap implements Cloneable {
 
     int getMaxRetries() {
         def result = get('maxRetries')
-        def defResult = getErrorStrategy() == ErrorStrategy.RETRY ? 1 : 0
-        result ? result as int : defResult
+        result != null ? result as int : 1
```
pditommaso, line 348:
```suggestion
        result ? result as int : 1
```

Better to keep as it was before

jorgee: I changed because result can be an integer and 0 is considered as false. It was breaking the test when number of retries is set to 0.

jorgee: I think it has not too much sense but there was this possibility and it was in the test. Or we keep null check or we change the test.

pditommaso: My reasoning here is to preserve the current behaviour that return `1` when max retries is set to `null` or `0`. The test was not taking into consideration the error strategy and therefore testing not the real usage `maxRetries` that's when `errorStrategy=='retry'`

PR #5488, opened 2024-11-08
Add further instructions for using the accelerator directive and additional options

adamrtalbot:
Extends the accelerator documentation to clarify that further requirements might be needed.

This wasn't obvious from the existing docs so I felt like it needed to be clarified.

### DIFF
--- a/docs/reference/process.md
+++ b/docs/reference/process.md
@@ -233,6 +233,10 @@ This directive is only used by certain executors. Refer to the {ref}`executor-pa
 :::
 
 :::{note}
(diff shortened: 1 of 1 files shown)


docs/reference/process.md
```diff
@@ -232,6 +232,10 @@ The above examples will request 4 GPUs of type `nvidia-tesla-k80`.
 This directive is only used by certain executors. Refer to the {ref}`executor-page` page to see which executors support this directive.
 :::
 
+:::{note}
+Additional options may be required to fully enable the use of accelerators. When using containers with GPUs, you must pass the GPU drivers through to the container. For Docker, this requires the option `--gpus all` in the docker run command. For Apptainer/Singularity, this requires the option `--nv`. The specific implementation details depend on the accelerator and container type being used.
```
pditommaso, line 236: um, should nextflow takes care of this?

PR #5489, opened 2024-11-08
Fix overlapping file lock exception

jorgee:
Fix conda packages with same name and content but different path.
Include a hash of the path in the environment prefix directory.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
@@ -22,6 +22,7 @@ import java.nio.file.Path
 import java.nio.file.Paths
 import java.util.concurrent.ConcurrentHashMap
 
+import com.google.common.hash.Hashing
 import groovy.transform.CompileStatic
 import groovy.transform.PackageScope
 import groovy.util.logging.Slf4j
@@ -32,7 +33,6 @@ import nextflow.file.FileMutex
 import nextflow.util.CacheHelper
 import nextflow.util.Duration
 import nextflow.util.Escape
-import org.yaml.snakeyaml.Yaml
 /**
  * Handle Conda environment creation and caching
  *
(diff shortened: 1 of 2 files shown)


ewels:
Tested locally with this:
```
nextflow run nf-core/rnaseq -r 3.17.0 -profile test,conda --outdir dir
```

Breaks on the released version and works fine with this PR build 🎉

ewels (approved):
If we could get this into a patch release that would be amazing 🙏🏻

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
```diff
@@ -192,7 +192,8 @@ class CondaCache {
                 if( yaml.name )
                     name = yaml.name
                 else
-                    name = path.baseName
+                    name = 'env-' + CacheHelper.hasher(path.toString()).hash().toString()
```
pditommaso, line 195: I think if two processes use the same conda env, it could raise the same issue

pditommaso: nvm

pditommaso:
Made a change to use normalised absolute path + using more compact sip hash algorithm. 

How this solution is going to be validated?

jorgee:
> Made a change to use normalised absolute path + using more compact sip hash algorithm.
> 
> How this solution is going to be validated?

I could try to add an integration tests that installs miniconda and runs the pipeline which reproduces the error. Do you think it is enough?

pditommaso:
I was more thinking running a real pipeline to replicate the problem. Maybe @ewels knows more

ewels:
I was trying to validate earlier this morning already, but couldn't get the original error to trigger any more 🙈

I'll clean up my env and try again shortly. Maybe @maxulysse or @pinin4fjords could also have a go?

maxulysse:
> I was trying to validate earlier this morning already, but couldn't get the original error to trigger any more 🙈
> 
> I'll clean up my env and try again shortly. Maybe @maxulysse or @pinin4fjords could also have a go?

I can definitively replicate the original error

pditommaso:
@maxulysse nice. Are you able to make a local build with this branch and try it?

maxulysse:
> @maxulysse nice. Are you able to make a local build with this branch and try it?

It's a success :rocket:

pditommaso:
Great!

ewels:
I was too slow, but I just checked on GitPod and was able to replicate the error and confirm that it's now fixed on `master` 👍🏻

PR #5496, opened 2024-11-11
Update install docs to reflect change from 'all' to 'dist'

kgalens:
Update nextflow docs to change `all` to `dist`. Unsure when this changed, but currently there is no longer an `all` distribution.


pditommaso (approved):
Well spotted!

pditommaso:
I've remove this 👉 f9d42844 warning because it does not apply anymore .

docs/install.md
```diff
@@ -119,15 +119,11 @@ NXF_VER=24.06.0-edge nextflow info
 
 ## Standalone distribution
 
-Nextflow has a set of {ref}`core plugins <plugins-core>` which are downloaded at runtime by default. There is also a standalone distribution (i.e. the `all` distribution) which comes pre-packaged with all core plugins. This distribution is mainly useful for offline environments.
+Nextflow has a set of {ref}`core plugins <plugins-core>` which are downloaded at runtime by default. There is also a standalone distribution (i.e. the `dist` distribution) which comes pre-packaged with all core plugins. This distribution is mainly useful for offline environments.
```
bentsherman, line 122: The `dist` doesn't include all core plugins anymore, should remove that part and/or specify which plugins are included

pditommaso: Good point, updated.

PR #5500, opened 2024-11-12
feat: RepositoryProvider.revision now public property

swingingsimian:
One line change to make the RepositoryProvider.revision a public property. The driver for this is to allow clients to better manage repository validation by allowing the RepositoryProvider to be the source of truth for the current revision that is set. 

An alternate approach would be to add revision validation within the RepositoryProvider itself, e.g. 
    `setRevision(String revision, boolean validateRevision = false)` 
Validation could then proceed by verifying the revision is present in getBranches or getTags.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy
@@ -65,12 +65,16 @@ abstract class RepositoryProvider {
     /**
      * The name of the commit/branch/tag
      */
-    String revision
+    protected String revision
 
     RepositoryProvider setCredentials(String userName, String password) {
         config.user = userName
         config.password = password
         return this
+    }
+
+    String getRevision() {
+        return this.revision
     }
 
     RepositoryProvider setRevision(String revision) {
--- a/modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy
@@ -78,6 +78,14 @@ abstract class RepositoryProvider {
         return this
     }
 
+    String getProject() {
+        return this.project
+    }
+
+    ProviderConfig getConfig() {
+        return this.config
+    }
+
     boolean hasCredentials() {
         getUser() && getPassword()
     }
--- a/modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy
@@ -65,7 +65,7 @@ abstract class RepositoryProvider {
     /**
      * The name of the commit/branch/tag
      */
-    protected String revision
+    String revision
 
     RepositoryProvider setCredentials(String userName, String password) {
         config.user = userName
--- a/modules/nextflow/src/test/groovy/nextflow/scm/RepositoryProviderTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/scm/RepositoryProviderTest.groovy
@@ -108,4 +108,15 @@ class RepositoryProviderTest extends Specification {
         and:
         1 * conn.setRequestProperty('Authorization', "Basic ${'foo:bar'.bytes.encodeBase64()}")
     }
+
+    def 'should have public revision property' () {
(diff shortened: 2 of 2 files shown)


modules/nextflow/src/test/groovy/nextflow/scm/RepositoryProviderTest.groovy
```diff
@@ -108,4 +108,15 @@ class RepositoryProviderTest extends Specification {
...
+
+    def 'should have public revision property' () {
+        given:
+        def provider = Spy(RepositoryProvider)
+        when:
+        provider.revision = 'branch_or_tag'
+        then:
+        provider.revision == 'branch_or_tag'
+        provider.hasProperty('revision') ? true : false
+    }
```
bentsherman, line 120: Do we need a unit test for this? I don't think the test will actually test anything because Spock can access protected members on a spy anyway

modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy
```diff
@@ -65,7 +65,7 @@ abstract class RepositoryProvider {
     /**
      * The name of the commit/branch/tag
      */
-    protected String revision
+    String revision
```
pditommaso, line 68: Instead of changing to public better to a getter to maintain the encapsulation. May even it be changed to `private visibility` (maybe). 

I would also add `project` and `config` for consistency

pditommaso (changes requested):

modules/nextflow/src/test/groovy/nextflow/scm/RepositoryProviderTest.groovy
swingingsimian: The property getter (i.e. `provider.revision == 'branch_or_tag'`), definitely raises a MissingMethodException exception if the field is protected. Although tbh I wasn't expecting that as the test is in the same package, so should have access. The `hasProperty` test was my attempt to assert that is is actually property and hence public.

This is however moot given Paolo's comments below which I will address. And this test will go away.

modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy
swingingsimian: Maybe my PR title is a little confusing, what I mean is enable public access the 'groovy way' via a property, which makes the underlying field private, not explicitly make the field public. i.e. taking advantage of groovy's accessor auto-generation/usage.

pditommaso: understand, my point is to keep protected (or private) and only add public getter so it can be accessed only for read but not for write

swingingsimian: `revision` already has a public setter `RepositoryProvider setRevision(String revision) `.  `project` and `config` do not, rather they are populated via the constructor, naturally, as they do not change as `revision` could. The current change to omit the modifer results in:

- private underlying field
- autogenerated public getter
- pre-existing explicitly defined public setter 

I will add public getters for project and config also.

modules/nextflow/src/test/groovy/nextflow/scm/RepositoryProviderTest.groovy
swingingsimian: Another option here would be to have test in another package to assert the public API. I don't currently see this pattern in the repo though, so will take you guidance on whether this is required of not.

modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy
pditommaso: Got your point still better keeping it `protected` and add `getRevision` for consistency with other getters

swingingsimian: Ah, <click> gotcha :)

swingingsimian: done

pditommaso: ty!

pditommaso (approved):

PR #5501, opened 2024-11-13
Fix problem that Nextflow Executions stuck on Kubernetes

Lehmann-Fabian:
When running Nextflow workflows on Kubernetes, workflows would often get stuck with Pods remaining in the Succeeded state without being removed, and the workflow itself would not progress.

After running a jstack on the Nextflow Java process, I identified a thread hanging at this line in TraceRecord.groovy:
[TraceRecord.groovy#L422](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy#L422).

To resolve this, I replaced the Groovy methods with native Java functions. After using this change in production for over six months, the problem has not recurred, so I am contributing it with confidence that it will benefit other Nextflow and Kubernetes users.

We experienced this issue on three different clusters, all running Ceph with Kubernetes.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
@@ -418,48 +418,47 @@ class TraceRecord implements Serializable {
      */
 
     TraceRecord parseTraceFile( Path file ) {
+        try (BufferedReader reader = new BufferedReader(new FileReader(file.toFile()))) {
 
-        final text = file.text
+            if( reader.readLine() != 'nextflow.trace/v2' )
+                return parseLegacy(file, file.readLines())
 
-        final lines = text.readLines()
-        if( !lines )
-            return this
-        if( lines[0] != 'nextflow.trace/v2' )
-            return parseLegacy(file, lines)
+            String line
+            while(( line = reader.readLine() ) != null ) {
+                final pair = line.tokenize('=')
+                final name = pair[0]
+                final value = pair[1]
+                if( value == null )
+                    continue
 
-        for( int i=0; i<lines.size(); i++ ) {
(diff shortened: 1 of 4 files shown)


bentsherman:
That line should be delegating to [this method](https://github.com/apache/groovy/blob/02a8d04cd3de2569fecd5e73b06239ae0ded032e/src/main/java/org/codehaus/groovy/runtime/IOGroovyMethods.java#L855-L873)

It is slightly different from your approach, maybe it is getting into an infinite loop

Lehmann-Fabian:
Yes, and the BufferedReader is created by: `Files.newBufferedReader(self, Charset.defaultCharset());`.
The jstack trace is always in this line: [IOGroovyMethods.java#L862](https://github.com/apache/groovy/blob/02a8d04cd3de2569fecd5e73b06239ae0ded032e/src/main/java/org/codehaus/groovy/runtime/IOGroovyMethods.java#L862)
There may be a timeout that is not caught in the distributed file system as an infinite loop would cause the answer StringBuilder to fill until OOM.
Anyway, the new implementation does not have the problem and is even a little bit more memory efficient, as it reads the trace file line by line ;)

modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
```diff
@@ -418,48 +418,50 @@ class TraceRecord implements Serializable {
      */
 
     TraceRecord parseTraceFile( Path file ) {
+        try (BufferedReader reader = new BufferedReader(new FileReader(file.toFile()))) {
```
pditommaso, line 421: I believe this is not going to work with object storage like AWS S3 file system

pditommaso:
Closing in favour of #5683

PR #5502, opened 2024-11-13
Fix bugs with workflow outputs

### LINKED ISSUE #5460: output block is needed for publishing outputs from workflow
## Bug report 

When using the second preview version for publishing outputs from workflows the `output` block is required for files to be actually published. Nothing needs to be set within the block, `output {}` is enough for files to be published. Omitting this line prevents any file from being published.

The documentation for [Customizing outputs](https://www.nextflow.io/docs/latest/workflow.html#customizing-outputs) clearly states that the `output` block is only needed for customizing the output and can be omitted when the default behaviour is good enough. But that doesn't seem to be the case, as omitting the `output` block prevents any file from being published.

workflow.nf
```
#!/usr/bin/env nextflow

nextflow.preview.output = true

process doStuff {
    output:
        path 'file.txt'

    script:
        """
        touch file.txt
        """
}


workflow {
    main:
        println "outputDir: $workflow.outputDir"
        out = doStuff()
        out.view { file -> "Out: ${file}" }

    publish:
        out >> "default"
    
}

// Uncomment the next line to actually publish the output file
// output {}
```

Run this file once and check with `ls` to see no `results` directory
```
NXF_VER=24.10.0 nextflow run workflow.nf
ls -lh
```

And then run it again after uncommenting the last line to see the `results` directory
```
NXF_VER=24.10.0 nextflow run workflow.nf
ls -lh
```

### Program output 

```
N E X T F L O W  ~  version 24.10.0
Launching `workflow.nf` [hopeful_sanger] DSL2 - revision: d4837e8320
WARN: WORKFLOW OUTPUT DSL IS A PREVIEW FEATURE - SYNTAX AND FUNCTIONALITY CAN CHANGE IN FUTURE RELEASES
outputDir: /absolute/path/to/current/directory/results
[be/c36072] Submitted process > doStuff
Out: /absolute/path/to/current/directory/work/be/c360721f8d99f992ba55284f903b45/file.txt
```

[.nextflow.log](https://github.com/user-attachments/files/17600872/nextflow.log)

### Environment 

* Nextflow version: 24.10.0
* Java version: openjdk 11.0.13 2021-10-19
* Operating system: Rocky Linux 8.10 (Green Obsidian)
* Bash version: GNU bash, version 4.4.20(1)-release

### LINKED ISSUE #5465: Output DSL not creating parent directory.

Using output dsl as specified in [docs](https://nextflow.io/docs/latest/workflow.html#migrating-from-first-preview) for creating a samplesheet of a channel. Throws an `ERROR ~ No such file or directory`, when I would expect it to just create the file with it's correct parent dirs.

```nextflow
#!/usr/bin/env nextflow

nextflow.preview.output = true
workflow {
    main:
    ch_fastq = Channel.of(
        [ [id: 1, name: 'sample 1'], '1a.fastq', '1b.fastq' ],
        [ [id: 2, name: 'sample 2'], '2a.fastq', '2b.fastq' ],
        [ [id: 3, name: 'sample 3'], '3a.fastq', '3b.fastq' ]
    )

    publish:
    ch_fastq >> 'fastq'

}

output {
    'fastq' {
        index {
            path 'index.csv'
        }
    }
}
``` 

```bash 
 NXF_VER=24.10.0 nextflow run . -output-dir testing
```

[dummy - repo](https://github.com/Joon-Klaps/Nextflow_sandbox/tree/dsl-output-error)

### Program output 

```stdout

 N E X T F L O W   ~  version 24.10.0

Launching `./main.nf` [ecstatic_shirley] DSL2 - revision: f3a7b194c5

WARN: WORKFLOW OUTPUT DSL IS A PREVIEW FEATURE - SYNTAX AND FUNCTIONALITY CAN CHANGE IN FUTURE RELEASES
ERROR ~ No such file or directory: /workspace/Nextflow_sandbox/testing/fastq/index.csv

```
[nextflow.log](https://github.com/user-attachments/files/17619663/nextflow.log)

### Environment 

* Nextflow version: 24.10.0
* Java version: 17.0.11-internal 2024-04-16
* Operating system: Linux
* Bash version: 5.1.16(1)-release

### LINKED ISSUE #5476: Dynamic publish path is not resolved against output directory

While trying out the workflow output definitions, I noticed some small inconsistencies on how the actual publishing happens between the different methods. 

This will publish to the `$launchDir` or `$projectDir` (not sure which one it is in this case) location instead of publishing relative to the output directory:

```nextflow
'test' { 
    path { txt -> "dir/${txt.name}" }
}
```

While this will publish relative to the output directory:

```nextflow
'test2' { 
    path { _txt -> { file -> "dir/${file}" } }
}
```

So when I executed both at the same time, the result was this:

```
.
├── dir
│   └── test.txt
│       └── test.txt -> /home/nvnieuwk/Documents/nextflow/issues/workflow-output-inconsistency/work/26/f7361be58d416641ec7e2226210613/test.txt
├── main.nf
├── results
│   └── dir
│       └── test2.txt -> /home/nvnieuwk/Documents/nextflow/issues/workflow-output-inconsistency/work/26/f7361be58d416641ec7e2226210613/test2.txt
└── work
    └── 26
        └── f7361be58d416641ec7e2226210613
            ├── test2.txt
            └── test.txt
```

I think it would be better if this behaviour would be consistent between both methods. Where all relative paths are published relative to the output directory and all absolute paths will be published to the absolute path instead. 

Execute the MRE in the dropdown menu below

<details>

<summary>MRE</summary>

```nextflow
nextflow.preview.output = true

process TEST {
    output:
    path("test.txt"), emit: test
    path("test2.txt"), emit: test2

    script:
    """
    touch test.txt
    touch test2.txt
    """
}

workflow {
    main:
    TEST()

    publish:
    TEST.out.test >> 'test'
    TEST.out.test2 >> 'test2'
}

output {
    'test' { 
        path { txt -> "dir/${txt.name}" }
    }
    'test2' { 
        path { _txt -> { file -> "dir/${file}" } }
    }
}
```

</details>

### Environment 

* Nextflow version: 24.10.0
* Java version: 17.0.3
* Operating system: Ubuntu

### LINKED ISSUE #5480: Bug in workflow output definitions with HTTP input file publishing

When trying to publish an HTTP file that has not been staged in a process, the pipeline fails with the following error:

```
ERROR ~ assert path
       |
       null
```

Simply running this script is enough to reproduce the error:

```nextflow
nextflow.preview.output = true

workflow {
    main:
    def test_out = Channel.fromPath("https://github.com/nf-core/test-datasets/raw/refs/heads/modules/data/genomics/sarscov2/genome/genome.fasta")

    publish:
    test_out >> 'test'
}

output {
    'test' {
        path { _txt -> { file -> "test/${file}" }}
    }
}
```

It does seem to succesfully resolve itself when I omitted the `output` block, but it didn't publish the file

### Program output 

[nextflow.log](https://github.com/user-attachments/files/17661028/nextflow.log)

### Environment 

* Nextflow version: 24.10.0
* Java version: 17.0.3
* Operating system: Ubuntu

### LINKED ISSUE #5543: outputDir can't be used as an implicit variable within a config
## Bug report 

outputDir can't be used as an implicit variable within a config.
This means it can't be used within `report` to set the location of the report dynamically.

```
outputDir = "output_test"
report {
    enabled = true
    file    = "${outputDir}/pipeline_info/execution_report.html"
}
```
should put the report inside the outputDir, but instead gives 
```
ERROR ~ Unknown config attribute `report.outputDir` 
```

### Environment 

* Nextflow version: 24.10.1

### Additional context

Confirmed as an issue by @bentsherman :

> I'm now seeing we forgot to add outputDir as an implicit variable in the config

bentsherman:
TODO:
- [x] Fix #5460
- [x] Fix #5465
- [x] Fix #5476 
- [x] Fix #5480
- [x] Fix #5543
- [x] Update unit tests, e2e tests where appropriate


bentsherman:
Tests are failing due to Wave?

```
~ Test 'fusion-symlink.nf' run failed
   + '[' -z *** ']'
   + echo initial run
   initial run
   + /home/runner/work/nextflow/nextflow/nextflow -q run ../../fusion-symlink.nf -c .config
   ERROR ~ Error executing process > 'CREATE'
   
   Caused by:
     Process `CREATE` terminated with an error exit status (125)
   
   
   Command executed:
   
     echo HELLO > data.txt
   
   Command exit status:
     125
   
   Command output:
     (empty)
   
   Command error:
     Unable to find image 'wave.seqera.io/wt/e7b9795c2b58/nextflow/tests:latest' locally
     latest: Pulling from wt/e7b9795c2b58/nextflow/tests
     a3ed95caeb02: Already exists
     1b3e7701db96: Already exists
     a3ed95caeb02: Already exists
     a3ed95caeb02: Already exists
     486f7ff2dd95: Already exists
     009b1e9adc43: Already exists
     5ca4cba2c0a3: Already exists
     2997c9ef2ef5: Already exists
     a3ed95caeb02: Already exists
     a3ed95caeb02: Already exists
     a3ed95caeb02: Already exists
     140836dbc286: Already exists
     eed831c78eec: Already exists
     ac6d42ea81bb: Already exists
     f9510fa3a08f: Already exists
     3dca4ae9f36c: Already exists
     a3ed95caeb02: Already exists
     ceb5650fa213: Pulling fs layer
     dd97694ce7a4: Pulling fs layer
     ceb5650fa213: Verifying Checksum
     ceb5650fa213: Download complete
     ceb5650fa213: Pull complete
     dd97694ce7a4: Verifying Checksum
     docker: filesystem layer verification failed for digest sha256:dd97694ce7a42d826212b59d13df70a643c0192533becd68d9e867f88297c996.
     See 'docker run --help'.
   
   Work dir:
     s3://nextflow-ci/work/89/87e1d9b4c45dedc5006c8f69003[733](https://github.com/nextflow-io/nextflow/actions/runs/12281095140/job/34269672962?pr=5502#step:5:734)
   
   Container:
     wave.seqera.io/wt/e7b9795c2b58/nextflow/tests:latest
```

pditommaso:
I think it's a Fusion issue @alberto-miranda

docs/workflow.md
```diff
@@ -607,14 +605,20 @@ You can customize the index file with additional directives, for example:
...
-    sep '\t'
-    mapper { meta, fq_1, fq_2 -> meta + [fastq_1: fq_1, fastq_2: fq_2] }
+    path 'samples.csv'
+    header true
+    sep '|'
 }
 ```
 
-This example will produce the same index file as above, but with the `name` column removed and with tabs instead of commas.
+This example will produce the following index file.
```
christopher-hakkaart, line 614:
```suggestion
This example will produce the following index file:
```

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
```diff
@@ -149,38 +201,46 @@ class PublishOp {
...
+                }
         }
 
         throw new IllegalArgumentException("Index file record must be a list, map, or file: ${value} [${value.class.simpleName}]")
     }
 
-    private Path normalizePath(Path path) {
+    private String normalizePath(Path path, targetDirOrClosure) {
+        if( targetDirOrClosure instanceof Closure )
+            return (targetDirOrClosure.call(path.getName()) as Path).normalize().toUriString()
```
pditommaso, line 238: It would always be better to use `FileHelper.asPath` over `as Path`. Even if it's doing be same under the hood, it would allow us to better keep where this conversion is done.

pditommaso (changes requested):
I some concerns in this PR both in terms of the complexity introduces by the need for dynamic resolution of closures and String/Path conversions that makes very hard to understand, and even more in the complexity of the resulting pipeline code, for example: 

https://github.com/nextflow-io/nextflow/pull/5502/files#diff-fd4cbeab566b734cd6e0070be77477ff1e70bc1374d4734ca01882273c0e468dR92-R105

One goal of this effort is to provide a more declarative model and less programmatic oriented. I feel this brings back the chaos of `saveAs` closures in the current `publishDir`

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
bentsherman: Agreed

tests/output-dsl.nf
```diff
@@ -16,75 +16,95 @@
...
+        'log': 'fastqc'
+      ]
+      return { filename ->
+        def ext = filename.tokenize('.').last()
+        def dir = dirs[ext]
+        dir != null
+          ? "${dir}/${filename}"
+          : "${filename}/${sample.id}"
+      }
+    }
```
bentsherman, line 105: @pditommaso regarding your main comment

The purpose of the dynamic path directive is to map a data structure, like a record, to a directory tree, so that the directory structure can be independent of the index file structure. This decoupling is crucial.

I agree that it is ugly. I would like to find a more declarative syntax for this that is sufficiently flexible. I have a few ideas cooking. But this nested closure is the way to do it in the second preview. So right now the goal is to fix the bugs so that it behaves as intended. The sooner we merge these fixes, the sooner I can start working on a third preview / final version.

pditommaso: Still struggling my head about this snipper. If i'm understanding correctly `sample` is a map holding a the `id` and other file paths. I guess the closure in then invoked on all map entries that are Path. Is that correct?

bentsherman: yes

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
```diff
@@ -98,26 +105,65 @@ class PublishOp {
             publisher.apply(files, sourceDir)
         }
 
-        // create record for index file
+        // append record to index file
         if( indexOpts ) {
             final record = indexOpts.mapper != null ? indexOpts.mapper.call(value) : value
-            final normalized = normalizePaths(record)
+            final normalized = normalizePaths(record, targetDirOrClosure)
```
pditommaso, line 111: Ok, it has not been introduced in this PR, this always though this is not scalable with a large number of files. It should be taken in consideration an stream like approach before it's released as stable e.g. [here](https://github.com/nextflow-io/nextflow/blob/fix-workflow-outputs/modules/nextflow/src/main/groovy/nextflow/cache/CacheDB.groovy#L50-L50)

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
```diff
@@ -18,47 +18,48 @@ package nextflow.extension
...
+
     private DataflowReadChannel source
 
     private Map opts
 
-    private Path targetDir
+    private String path
 
-    private Closure pathAs
+    private Closure dynamicPath
```
pditommaso, line 46: Maybe `dynamicResolver` or  `pathResolver` or just `resolver` is not explicit

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
```diff
@@ -149,38 +202,46 @@ class PublishOp {
...
+    private String normalizePath(Path path, targetDirOrClosure) {
+        if( targetDirOrClosure instanceof Closure )
+            return FileHelper.asPath(targetDirOrClosure.call(path.getName()).toString()).normalize().toUriString()
         final sourceDir = getTaskDir(path)
-        return targetDir.resolve(sourceDir.relativize(path)).normalize()
+        if( sourceDir == null )
+            return path.toUriString()
+        final targetDir = FileHelper.asPath(targetDirOrClosure.toString())
+        return targetDir.resolve(sourceDir.relativize(path)).normalize().toUriString()
     }
```
pditommaso, line 245: This become very "dense". Adding a javadoc and some comments explaining the logic of different branches would help preventing future headaches

pditommaso: Also wonder why returning a string instead of `Path` the underlying operation will not require converting again to a `Path`?

docs/workflow.md
```diff
@@ -523,7 +523,7 @@ This output block has the following effect:
 See [Reference](#reference) for all available directives in the output block.
 
 :::{tip}
-The output block is only needed if you want to customize the behavior of specific targets. If you are satisfied with the default behavior and don't need to customize anything, the output block can be omitted.
+The output block can be omitted if you are satisfied with the default behavior. However, as a best practice you should declare each output in the output block, even if you don't need to customize the publish path or create any index files.
```
christopher-hakkaart, line 526:
```suggestion
The output block can be omitted if you are satisfied with the default behavior. However, as a best practice, you should declare each output in the output block, even if you don't need to customize the publish path or create any index files.
```

tests/output-dsl.nf
bentsherman: Another idea based on our discussion today:

```groovy
    path { sample ->
      sample.fastqc >> 'log/'
      sample.bam >> 'align/'
      sample.bai >> 'align/'
      sample.quant >> "quant/${sample.id}"
    }
```

More like a builder DSL, but still requires a closure to access the record members.

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
bentsherman: Good point, I will address this in a follow-up PR

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
bentsherman: Good point, I was struggling to find a good name for it

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
bentsherman: Somehow I managed to convert a path to string and back to path twice in this function 😅 

I think I changed to return string to ensure that paths were serialized correctly. Taking a second look, I see that the json writer already handles this and the csv writer just needed a bit of extra logic to do the same.

Between those simplifications and some extra comments, hopefully the code is much easier to understand now

PR #5507, opened 2024-11-14
Warn when file() matches a collection of files

bentsherman:
Looking at the [standard library functions](https://nextflow.io/docs/latest/reference/stdlib.html#functions) with an eye towards static type checking, the only problematic function is `file()` which can return a single file or list of files based on the glob pattern.

Rather than introduce some new function, I think we can transition `file()` to be properly typed:
1. for now, warn if `file()` matches a collection of files
2. in the future, change the return type of `file()` to `Path` and raise an error if the glob pattern matches zero or multiple files

Note that the glob pattern is still useful for `file()` when you know it will match only one file but you don't want to spell it out. Raising an error when this assumption is wrong seems like the right approach to me, but of course we should keep it as a warning just for now.

Not sure whether you want to add a test for the warning, but you can see it in action here:
```groovy
// assumes there are no *.foo files in the launch directory
println file('*.foo')
println file('*.nf')
println files('*.foo')
println files('*.nf')
```


christopher-hakkaart:
Is it appropriate to document `files()` should be used only for a collection of files at the same time?

bentsherman:
Agreed, better to describe `files()` on its own terms

docs/reference/stdlib.md
```diff
@@ -207,7 +207,10 @@ The following functions are available in Nextflow scripts:
 : Stop the pipeline execution and return an exit code and optional error message.
 
 `file( filePattern, [options] )`
-: Get one or more files from a path or glob pattern. Returns a [Path](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Path.html) or list of Paths if there are multiple files.
+: :::{versionchanged} 24.11.0-edge
+  The `file()` function should be used only to match a single file. Use `files()` to match a collection of files. In the future, `file()` will raise an error if it does not match exactly one file.
+  :::
+: Get one or more files from a file name or glob pattern. Returns a [Path](#path) if there is one file, or a collection of Paths if there are multiple files.
```
christopher-hakkaart, line 213:
```suggestion
: Get one or more files from a file name or glob pattern. Returns a single [Path](#path) if there is one file and a collection of Paths if there are multiple files.
```

christopher-hakkaart:
I've added a minor suggestion. Otherwise, the docs look good.

modules/nextflow/src/main/groovy/nextflow/Nextflow.groovy
```diff
@@ -115,11 +113,16 @@ class Nextflow {
...
      * @return An instance of {@link Path} when a single file is matched or a list of {@link Path}s
      */
-    static file( Map options = null, def filePattern ) {
-
+    static file(Map options = null, def filePattern) {
         if( !filePattern )
-            throw new IllegalArgumentException("Argument of `file` function cannot be ${filePattern==null?'null':'empty'}")
+            throw new IllegalArgumentException("Argument of `file()` function cannot be ${filePattern==null?'null':'empty'}")
+        final result = file0(options, filePattern)
+        if( result instanceof Collection && result.size() != 1 && SysEnv.get('NXF_SYNTAX_PARSER') == 'v2' )
```
bentsherman, line 120: TODO: replace with `Const.getSyntaxParser()` after #5816 is merged

christopher-hakkaart:
Docs look good to me 👍

bentsherman:
Since the change from `file()` to `files()` is pretty simple, and runtime warnings tend to adversely affect non-developer users, I think we can do without this warning and simply add a type-checking error whenever the developer opts into static type checking in the future

bentsherman:
Reviving this PR.

We removed several runtime warnings since they were more appropriate in the linter / language server. However, after reflecting on it more, I think this one needs to be a runtime warning, since the glob pattern is only evaluated at runtime.

I kept it as a warning and enabled only in the strict syntax, so it shouldn't be too bothersome.

docs/reference/stdlib-namespaces.md
```diff
@@ -79,10 +79,13 @@ The global namespace contains globally available constants and functions.
   `type: String`
   : Type of paths returned, can be `'file'`, `'dir'` or `'any'` (default: `'file'`)
 
-: See also: {ref}`channel.fromPath <channel-path>`.
+: :::{note}
+  This function will return a collection if the glob pattern yields zero or multiple files. Use `files()` to get a collection of files.
```
christopher-hakkaart:
```suggestion
  This function returns a collection if the glob pattern yields zero or multiple files. Use `files()` to get a collection of files.
```

christopher-hakkaart:
```suggestion
  This function will return a collection if the glob pattern yields zero or multiple files. To get a collection of files, use `files()`.
```

christopher-hakkaart (commented):
Added a couple of suggestion. Feel free to accept or reject.

christopher-hakkaart (approved):
Docs look good 👍

pditommaso (approved):

PR #5508, opened 2024-11-14
Deprecate process `shell` block

bentsherman:
Based on some recent internal discussions. The original purpose of the `shell` block was to help distinguish between Nextflow and Bash variables in the process script. Since the VS Code extension now provides both syntax highlighting and error checking for variable names, the `shell` block isn't needed as much and can be phased out to promote consistent code.

I didn't bother to add a warning in the code because I will add one in the language server, but we can add one here if you want.

### DIFF
--- a/docs/process.md
+++ b/docs/process.md
@@ -212,6 +212,10 @@ Template scripts are generally discouraged due to the caveats described above. T
 
 ### Shell
 
+:::{deprecated} 24.11.0-edge
+Use the `script` block instead. Consider using the {ref}`VS Code extension <vscode-page>`, which provides syntax highlighting and error checking to distinguish Nextflow variables from Bash variables in the process script.
+:::
+
 The `shell` block is a string expression that defines the script that is executed by the process. It is an alternative to the {ref}`process-script` definition with one important difference: it uses the exclamation mark `!` character, instead of the usual dollar `$` character, to denote Nextflow variables.
 
 This way, it is possible to use both Nextflow and Bash variables in the same script without having to escape the latter, which makes process scripts easier to read and maintain. For example:
--- a/docs/vscode.md
+++ b/docs/vscode.md
@@ -1,3 +1,4 @@
+(vscode-page)=
 
 # VS Code integration
 


pditommaso:
It would make sense to add a deprecation warning (and a warning to disable it)

christopher-hakkaart:
Can we clarify the behavior of single (`'`) vs. double quotes (`"`) in this PR as well?

pbelmann:
Just my two cents: I use most of the time the terminal to work on Nextflow files, and I think it's still useful to easily distinguish between Nextflow and Bash variables.

bentsherman:
@pbelmann which terminal editor do you use? the language server is being integrated with other editors [here](https://github.com/nextflow-io/language-server/issues/56). I believe they already have emacs and neovim

bentsherman:
@christopher-hakkaart I will address single vs double quotes in the other PR, but I believe `shell:` is meant to be used always with single quotes so that the dollar sign isn't treated as a Nextflow variable

modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
```diff
@@ -640,8 +640,9 @@ class NextflowDSLImpl implements ASTTransformation {
                             readSource(stm,source,unit)
                             break
 
-                        case 'script':
                         case 'shell':
+                            log.warn "The `shell` block is deprecated, use `script` instead"
```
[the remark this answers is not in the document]
bentsherman: done

pbelmann:
> @pbelmann which terminal editor do you use? the language server is being integrated with other editors [here](https://github.com/nextflow-io/language-server/issues/56). I believe they already have emacs and neovim

I use Vim most of the time.

bentsherman:
I'm not sure if vim supports LSP but they just added the nextflow language server to [neovim](https://github.com/neovim/nvim-lspconfig/pull/3423)

modules/nextflow/src/main/groovy/nextflow/ast/NextflowDSLImpl.groovy
```diff
@@ -640,8 +641,9 @@ class NextflowDSLImpl implements ASTTransformation {
                             readSource(stm,source,unit)
                             break
 
-                        case 'script':
                         case 'shell':
+                            log.warn1 "Process $name > the `shell` block is deprecated, use `script` instead"
```
pditommaso, line 645: Minor improvement, instead of reporting this warning here, it could be done at TaskProcessor creation level, this guarantee that's called once and the process name is provided 

https://github.com/nextflow-io/nextflow/blob/c6dfef6188127e1b036923f1304795de57b38071/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy#L311-L311

bentsherman: done

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -309,6 +309,8 @@ class TaskProcessor {
         this.ownerScript = script
         this.config = config
         this.taskBody = taskBody
+        if( taskBody.isShell )
+            log.warn1 "Process ${name} > the `shell` block is deprecated, use `script` instead"
```
pditommaso, line 313:
```suggestion
            log.warn "Process ${name} > the `shell` block is deprecated, use `script` instead"
```

This class should create once of process, `log.warn` should be ok

pditommaso (approved):

Issue #5503, opened 2024-11-14
NXF_DISABLE_WAVE_SERVICE do not fully disable Wave


jordeu:
## Bug report 

I'm trying to run a pipeline with Fusion enabled but Wave disabled, but even if `NXF_DISABLE_WAVE_SERVICE=true` environment variable Nextflow it's still trying to use Wave and this error is reported:

```
Caused by:
  Unable to find any container with key: my-fusion-container:latest
java.lang.IllegalStateException: Unable to find any container with key: my-fusion-container:latest
    at io.seqera.wave.plugin.WaveClient.isContainerReady(WaveClient.groovy:638)
    at io.seqera.wave.plugin.resolver.WaveContainerResolver.isContainerReady(WaveContainerResolver.groovy:132)
    at org.codehaus.groovy.vmplugin.v8.IndyInterface.fromCache(IndyInterface.java:321)
    at nextflow.processor.TaskRun.isContainerReady(TaskRun.groovy:720)
```

The problem seems to be that is looking for the `my-fusion-container` image, but it fails because this image is only available locally.

### Expected behavior and actual behavior

That `WaveContainerResolver` is not used when it is disable, and then a local only image can be resolved.

### Steps to reproduce the problem

Replace `YOUR_BUCKET` for a bucket where you have access and run these bash script:

```bash
cat - >test.nf <<EOF
process hello {

  output:
  path 'hello.txt'

  script:
  '''
  echo "Hello world" > hello.txt
  '''
}

workflow {
  hello()
}
EOF

cat - >nextflow.config <<EOF
docker.enabled = true
process.container = 'my-fusion-container:latest'
fusion.enabled = true
fusion.exportStorageCredentials = true
workDir = 's3://YOUR_BUCKET/work'
EOF

docker pull ubuntu && docker tag ubuntu my-fusion-container

export NXF_DISABLE_WAVE_SERVICE=true

nextflow run test.nf
```

### Program output 

[.nextflow.log](https://github.com/user-attachments/files/17746157/default.nextflow.log)

### Environment 

```bash
$ nextflow info
  Version: 24.10.0 build 5928
  Created: 27-10-2024 18:36 UTC (19:36 CEST)
  System: Linux 5.15.0-124-generic
  Runtime: Groovy 4.0.23 on OpenJDK 64-Bit Server VM 21.0.2+13-jvmci-23.1-b30
  Encoding: UTF-8 (UTF-8)
```

pditommaso:
[...]

PR #5509, opened 2024-11-14
Fix isContainerReady when wave is disabled

pditommaso:
This PR fixes the access to the method [isContainerReady](https://github.com/nextflow-io/nextflow/blob/803f4ee386c63e945dfd3daf5cbef895428a456b/plugins/nf-wave/src/main/io/seqera/wave/plugin/resolver/WaveContainerResolver.groovy#L131-L131) when Wave is disabled. 

Solve #5503


pditommaso:
@jordeu likely you have already the test script and container in place, can you please validate it ?

jordeu (approved):
Tested and now it works as expected

pditommaso:
Thanks!

PR #5513, opened 2024-11-15
Removing secondary exceptions from console

jorgee:
- Unknown errors before session abort have been moved from log.error to log.debug to do not print them in the console.
- When session is cancelled and later aborted, do not call notifyError in abort (duplicate error notifications)

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/extension/DataflowHelper.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/extension/DataflowHelper.groovy
@@ -95,7 +95,7 @@ class DataflowHelper {
         @Override
         boolean onException(final DataflowProcessor processor, final Throwable t) {
             final e = t instanceof InvocationTargetException ? t.cause : t
-            OperatorImpl.log.error("@unknown", e)
+            OperatorImpl.log.debug("@unknown", e)
             session?.abort(e)
             return true;
         }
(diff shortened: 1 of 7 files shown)


modules/nextflow/src/main/groovy/nextflow/extension/DataflowHelper.groovy
```diff
@@ -95,7 +95,7 @@ class DataflowHelper {
         @Override
         boolean onException(final DataflowProcessor processor, final Throwable t) {
             final e = t instanceof InvocationTargetException ? t.cause : t
-            OperatorImpl.log.error("@unknown", e)
+            OperatorImpl.log.debug("@unknown", e)
```
pditommaso, line 98: Would not make more sense to rely only on `session.abort` to report this error?

jorgee: Do you mean removing them?. Or moving these logs to abort? Currently, the first exception reaching abort is the one that is printed.

pditommaso: I was thinking more removing, but I agree better keep as debug. It can be removed later

pditommaso:
I was thinking something like [this](https://github.com/nextflow-io/nextflow/pull/5518). Can you please give it a try

pditommaso:
Improved version at 👉  https://github.com/nextflow-io/nextflow/issues/5437

pditommaso:
Closing in favour https://github.com/nextflow-io/nextflow/pull/5518

PR #5514, opened 2024-11-15
Include job status report when failure and no exit code found

jorgee:
When a job fail and no exit code is found. It prints a dump of the queue however no details of the failed jobs are shown specially in Slurm, where the job is not shown in the queue dump. This dump is replaced by a job status report provided by a call to te queue system


bentsherman:
Closing due to age. Feel free to re-open if you get it working

PR #5517, opened 2024-11-18
Fix hash when CmdEvalParam as output

### LINKED ISSUE #5470: Pipeline uses cached process despite edited eval script
## Bug report 

- Despite editing an eval script, a cached version of the process will be used under certain circumstances

- Expected: editing an eval script (e.g. from `multiqc --version` to `multiqc --version | sed "s/version//" `) will be picked up as an edit and therefore a cached version won't be used.

- Actual behaviour: it works correctly if `resume = true` is stated within the main workflow script. However if `-resume` is provided on command line, or `resume = true` is given in nextflow.config, then the cached version is used after a change to the `eval` script

1. Generate nextflow.config and main.nf from below example
2. Run the pipeline
3. Edit the eval statement to 'multiqc --version | sed "s/version//"'
4. Rerun -> nothing changes about the output

```nextflow.config
resume = true
apptainer.enabled = true
apptainer.autoMounts = true
```

```main.nf
nextflow.enable.dsl=2
nextflow.preview.topic = true

process FOO {
    tag "${id}"
    container "quay.io/biocontainers/multiqc:1.14--pyhdfd78af_0"

    input:
    val(id)

    output:
    tuple val(task.process), eval('multiqc --version'), topic: versions

    script:
    """
    """
}

workflow {
    input_ch = Channel.from("Sample1") 

    FOO(input_ch)

    channel.topic('versions').view()
}
```

### Program output 

Before edit:
[FOO, multiqc, version 1.14]

After edit:
[FOO, multiqc, version 1.14]

Should be:
[FOO, multiqc,  1.14]   # Note that this behaviour is achieved by moving `resume = true` from `nextflow.config` to `main.nf`

### Environment 

* Nextflow version: version 24.10.0 build 5928
* Java version: openjdk 11.0.24 2024-07-16
* Operating system: Ubuntu 22.04.5 LTS

jorgee:
close  #5470

Adds the CmdEvalParams in the task outputs as a key to compute the hash

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -2211,6 +2211,13 @@ class TaskProcessor {
             keys.add( it.value )
         }
 
+        // add all eval commands (they are part of the script)
+        for( Map.Entry<OutParam,Object> it : task.outputs ) {
+            if( it.key instanceof CmdEvalParam ) {
+                keys.add(((CmdEvalParam) it.key).getTarget(task.context))
+            }
+        }
+
         // add all variable references in the task script but not declared as input/output
         def vars = getTaskGlobalVars(task)
         if( vars ) {
@@ -2245,6 +2252,7 @@ class TaskProcessor {
             }
         }
 
+        // add stub-run flag
         if( session.stubRun && task.config.getStubBlock() ) {
             keys.add('stub-run')
         }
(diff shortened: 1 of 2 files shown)


pditommaso:
This alter the assumption the task is idempotent. The task hash should only rely on the inputs not on the outputs

bentsherman:
Considering that eval is a shorthand for export an output env var in the script, it is essentially part of the script and therefore part of the task identity.

I think it is possible to evaluate the task outputs without "resolving" or "collecting" them, before computing the task hash. Then you can see if any output evals were declared and include them as if they were part of the script

pditommaso:
This needs to be marked as a cache breaking change in the changelog

pditommaso:
also I really not sure about this, because break the contract that task ID depends only on inputs + script

bentsherman:
The eval commands are part of the script

PR #5518, opened 2024-11-18
Fix secondary exceptions from console

pditommaso:
This PR is a an attempt to fix issues https://github.com/nextflow-io/nextflow/issues/5437 turning off the printing of logs once `abort` has been invoked

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -792,7 +792,7 @@ class Session implements ISession {
             log.debug "Session aborted -- Cause: ${cause?.message ?: cause ?: '-'}"
         aborted = true
         error = cause
-        LoggerHelper.skipConsole = true
+        LoggerHelper.aborted = true
         try {
             // log the dataflow network status
             def status = dumpNetworkStatus()
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -71,6 +71,7 @@ import nextflow.util.Barrier
 import nextflow.util.ConfigHelper
 import nextflow.util.Duration
 import nextflow.util.HistoryFile
+import nextflow.util.LoggerHelper
 import nextflow.util.NameGenerator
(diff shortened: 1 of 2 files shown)


jorgee:
Much more elegant solution than #5513

pditommaso:
Just tried to make it more generic. Does it solve the original issue?

jorgee:
> Just tried to make it more generic. Does it solve the original issue?

It prints less logs but there is a case where it is printing several messages. When there is task failure that cancels the session and later a failure that produces the abort. It prints the task failure, the abort error, and two nf-core messages due to the two error notifications. 
The duplicated error notification can be solved with what I did in the other [branch](https://github.com/nextflow-io/nextflow/blob/19a4f4e631db981a2101a2e78c09fae91f6967e7/modules/nextflow/src/main/groovy/nextflow/Session.groovy#L802C1-L806C15).

Regrading the error message when aborting after cancelling, not sure if you prefer to keep both errors or not. In any case it can disable the second error by setting the `LoggerHelper.aborted` flag also in the `Session.cancel()`.  

If you agree I can commit this two changes in this branch

pditommaso:
Do you mean [this](https://github.com/nextflow-io/nextflow/pull/5513/files#diff-bee88a999f8eecd3cec5fcad24b76736162be86626f89d746932f64928f24374R801-R808)? 

I'm a bit hesitant to chance the `notifyError` logic only to suppress extra logs. Unless there's something else  i'm missing i'll keep as it it

jorgee:
> Do you mean [this](https://github.com/nextflow-io/nextflow/pull/5513/files#diff-bee88a999f8eecd3cec5fcad24b76736162be86626f89d746932f64928f24374R801-R808)?
> 
Yes
> I'm a bit hesitant to chance the `notifyError` logic only to suppress extra logs. Unless there's something else i'm missing i'll keep as it it
More than logging is that in situations where there is a session cancellation and later an abort, notifyError is called twice. I just wonder if it fine or we should just notify once. The same for the logger if there is a session cancellation and then an abort should Nextflow print both errors or just the first? 

If nextflow should print and notify both errors then it is fine, no more changes are required.

pditommaso:
Let's go with this, and make another iteration if needed

PR #5520, opened 2024-11-19
Improve groupTuple docs with scatter/gather example

bentsherman:
This PR expands the groupTuple + groupKey example to better demonstrate how to do a nested scatter/gather with flatMap and groupTuple/groupKey.

It also explicitly unwraps the group key to avoid the issue reported in #4104


christopher-hakkaart (approved):
Nice catch

pditommaso (approved):

PR #5526, opened 2024-11-19
Update process snippets to comply with strict syntax

bentsherman:
The strict syntax is a bit more picky about the process `script:` section label ([source](https://nextflow.io/docs/latest/vscode.html#restricted-syntax)):

> The Nextflow language specification allows the `script:` label to be omitted only if there are no other sections

This PR updates process snippets throughout the docs to comply with this rule.

### DIFF
--- a/docs/cache-and-resume.md
+++ b/docs/cache-and-resume.md
@@ -148,6 +148,8 @@ process gather {
     input:
     tuple val(id), file(foo)
     tuple val(id), file(bar)
+
+    script:
     """
     merge_command $foo $bar
     """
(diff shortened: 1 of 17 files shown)


christopher-hakkaart:
Adding `script:` looks good to me. It does make the the `"""` vs `'''` more noticeable in examples. What do you think about standardizing these as `"""` (unless single quotes are required)?

bentsherman:
The rules around single vs double quotes:
- if you need to reference a Nextflow variable, you must use double quotes
- otherwise, if you need to reference a Bash variable, it's easier to use single quotes so that you don't have to escape the dollar sign
- otherwise, it doesn't matter

I have updated the process scripts to be more consistent, i.e. use double quotes except where it's cleaner to use single quotes

pditommaso (approved):

PR #5529, opened 2024-11-21
Fix PBSPRO queue status fetch

### LINKED ISSUE #5528: Cannot fetch PBSPRO queue status

No warnings should appear in the log file. It seems that the pipelines are unable to detect failing jobs, instead waiting for the timeout before retrying the task, which significantly increases the runtime. This behaviour has been observed in `nf-core/cutandrun` and `nf-core/scflow`.

I believe all pipelines/runs with `pbspro` as their executors should be affected.

### Program output 

```bash
WARN: [PBSPRO] queue status cannot be fetched
- cmd executed: bash -c set -o pipefail; qstat -f $( qstat -B | grep -E -v '(^Server|^---)' | awk -v ORS=' ' '{print "@"$1}' ) | { grep -E '(Job Id:|job_state =)' || true; }
- exit status : 143
- output      :
```

### Environment 

* Nextflow version: 24.10.1 
* Java version: 23.0.1
* Operating system: Red Hat Enterprise Linux (version 8.5) 

### Additional context

The problematic code snippet appears to be `qstat -B | grep -E -v '(^Server|^---)' | awk -v ORS=' ' '{print "@"$1}'`, which produces two outputs in my case- `@pbs` and `@`.

While `qstat -f @pbs` executes as expected, `qstat -f @` results in an error, which is the root cause of the issue.

HDash:
Closes #5528 

A simple conditional `if` statement to check if the output is empty fixes the issue.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/PbsProExecutor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/PbsProExecutor.groovy
@@ -104,7 +104,7 @@ class PbsProExecutor extends PbsExecutor {
         if( queue ) {
             cmd += queue
         } else {
-            cmd += '$( qstat -B | grep -E -v \'(^Server|^---)\' | awk -v ORS=\' \' \'{print \"@\"\$1}\' )'
+            cmd += '$( qstat -B | grep -E -v \'(^Server|^---)\' | awk -v ORS=\' \' \'{if (\$1 != \"\") print \"@\"\$1}\' )'
         }
         return ['bash','-c', "set -o pipefail; $cmd | { grep -E '(Job Id:|job_state =)' || true; }".toString()]
     }
--- a/modules/nextflow/src/test/groovy/nextflow/executor/PbsProExecutorTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/executor/PbsProExecutorTest.groovy
@@ -211,7 +211,7 @@ class PbsProExecutorTest extends Specification {
         def executor = [:] as PbsProExecutor
 
         expect:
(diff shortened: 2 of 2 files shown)


bentsherman:
I have a sneaking suspicion that this worked on whatever PBSPro cluster it was tested on, and maybe the status text changed across different versions. It would be good to have some clear documentation on how the status text is supposed to look, so that we can test against it, like here: https://github.com/nextflow-io/nextflow/blob/fa0e8e0f3d01e27dac60230d304dcec95dd1c2f5/modules/nextflow/src/test/groovy/nextflow/executor/PbsProExecutorTest.groovy#L219-L238

At the same time, I'm pretty sure the PBSPro executor was added at my request, to support my university cluster 😅 And I do remember seeing these "failed to fetch queue status" warnings but I never figured out why they were happening

HDash:
Hello @bentsherman,

You are correct—I’ve been encountering this issue on Imperial’s cluster. It seems that `qstat -B` returns a phantom empty server string, which causes an error when passed to `qstat -f`. I have detailed the output in the mentioned issue.

The job state parsing portion of the code, however, appears to work perfectly and functions as intended. The failure lies in fetching the job state strings.

With my fork, `qstat -f $( qstat -B | grep -E -v '(^Server|^---)' | awk -v ORS=' ' '{if ($1 != "") print "@"$1}' ) | { grep -E '(Job Id:|job_state =)' || true; }` outputs:

```bash
Job Id: 297959.pbs
    job_state = Q
```

Which is exactly what we want. My queued job is labelled appropriately.

bentsherman:
Closing since I don't feel confident about changing this logic without potentially breaking it for other users

If there is some public documentation that can confirm what the `qstat` output looks like in a universal way, then we could point to that documentation in the tests

PR #5533, opened 2024-11-21
Update machineType docs to reflect supported executors

adamrtalbot:
Process directive machineType was out of date. Updated it here.

### DIFF
--- a/docs/reference/process.md
+++ b/docs/reference/process.md
@@ -817,7 +817,7 @@ See also: [resourceLabels](#resourcelabels)
 :::{versionadded} 19.07.0
 :::
 
-The `machineType` can be used to specify a predefined Google Compute Platform [machine type](https://cloud.google.com/compute/docs/machine-types) when running using the {ref}`Google Life Sciences <google-lifesciences-executor>` executor.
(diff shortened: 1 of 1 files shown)


pditommaso (approved):

docs/reference/process.md
```diff
@@ -817,7 +817,7 @@ See also: [resourceLabels](#resourcelabels)
 :::{versionadded} 19.07.0
 :::
 
-The `machineType` can be used to specify a predefined Google Compute Platform [machine type](https://cloud.google.com/compute/docs/machine-types) when running using the {ref}`Google Life Sciences <google-lifesciences-executor>` executor.
+The `machineType` can be used to specify a predefined Google Compute Platform [machine type](https://cloud.google.com/compute/docs/machine-types) when running using the {ref}`Google Batch <google-batch-executor>` or {ref}`Google Life Sciences <google-lifesciences-executor>` executor, or when using the autopools feature of the {ref}`Azure Batch executor<azurebatch-executor>`.
 
 This directive is optional and if specified overrides the cpus and memory directives:
```
bentsherman, line 822: By the way, this sentence is not necessarily true. On Google Batch you can specify both the machineType and cpus/memory, which is useful for packing jobs onto a VM as a job array

Can't recall how it works with Azure Batch

adamrtalbot: I think it works the same way, using globs:

```nextflow
process MY_GOOGLE_PROCESS {
    machineType "n2*"
    cpus 2
    // uses an n2-standard-2
}

process MY_AZURE_PROCESS {
    machineType "standard_d*_v5"
    cpus 2
    // uses a standard_d2_v5
}
```

Would it be better to point people to the relevant documentation here?

bentsherman: Probably yes

PR #5534, opened 2024-11-22
Docs: Move `-params-file` config from CLI to config

christopher-hakkaart:
- Added params file section to the config page
- Reduced CLI documentation (not the place to describe JSON and YAML)
    - Added link to config page in the example

cc @FriederikeHanssen

@netlify /config#params-file

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -138,6 +138,8 @@ params {
 }
 ```
 
+See {ref}`cli-params` for information about how to modify these on the command line.
+
 (config-process)=
 
 ## Process configuration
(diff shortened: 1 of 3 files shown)


docs/config.md
```diff
@@ -331,3 +331,37 @@ workflow.onError = {
...
+nextflow run main.nf -params-file pipeline_params.yml
+```
+
+The `-params-file` option loads parameters for your Nextflow pipeline from a JSON or YAML file. Parameters defined in the file are equivalent to specifying them directly on the command line. For example:
+
+```bash
+nextflow run main.nf --alpha 1 --beta foo
+```
+
+The above parameters can be represented in YAML format:
```
ewels, line 351: Feels like this doesn't quite follow on / flow from the text above.
```suggestion
would be equivalent to the following in YAML format:
```

ewels (approved):
LGTM, thanks!

docs/config.md
christopher-hakkaart: It was feeling uneasy for me too. New flow proposed.

bentsherman (changes requested):
Params file should be described [here](https://nextflow.io/docs/latest/cli.html#pipeline-parameters) (on the Command line page) rather than the Config page. You can add a backlink to the Command line page from [here](https://nextflow.io/docs/latest/config.html#parameters)

bentsherman (approved):

pditommaso (approved):

PR #5536, opened 2024-11-22
Support not using a container in Azure Batch

adamrtalbot:
There are rare cases where a user needs to run a process on a bare VM on Azure Batch. Most notably, when using the Dragen which requires you to use a specific machine image that does not support a Docker container. 

Previously, Nextflow prevented this, but this PR removes that safeguard and replaces it with a warning. It's not a good idea, but it can unblock users if they need to use a bare VM.

If nothing else, this PR functions as proof it can be done if anyone else needs to use it.

### DIFF
--- a/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
+++ b/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
@@ -437,15 +437,18 @@ class AzBatchService implements Closeable {
 
         log.trace "[AZURE BATCH] Submitting task: $taskId, cpus=${task.config.getCpus()}, mem=${task.config.getMemory()?:'-'}, slots: $slots"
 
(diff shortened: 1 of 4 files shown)


tom-seqera:
> It's not a good idea, but it can unblock users if they need to use a bare VM.

Given this, would it be worth keeping the strict validation on by default and instead providing an override config option to turn it off (like `azure.requireContainer=false` or something) then updating the validation error message to explain that the option exists but you should only use it if you know what you're doing?

pditommaso:
My understanding is that some tasks my require a container, some other may not, therefore a global setting is sub-optimal. 

Also, could not be Dragen accessed by the container anyway? this is for example how it works with AWS Batch (the dragen path in the host is made accessible in the container context)

pditommaso:
> The official AMI provided by Illumina for Azure is not compatible with Docker

What this means exactly?

PR #5538, opened 2024-11-25
Fix overlapping in conda environment creation when equivalent but diff…

jorgee:
When two process uses the same environment file but they are specified in the 'conda' directive using different relative paths, it generates different entries in the 'condaPrefixPaths' hasmap with different keys but the same value. An it tries to concurrently build the same environment twice producing the the lock file overlapping error

To fix it, when the provided conda env is considered a file/directory path (as evaluated in 'createLocalCondaEnv'), it is normalized before inserting to the hasmap or checking if conda environment is already in the hashmap.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
@@ -359,8 +359,8 @@ class CondaCache {
      *      The {@link DataflowVariable} which hold (and pull) the local image file
      */
     @PackageScope
-    DataflowVariable<Path> getLazyImagePath(String condaEnv) {
-
+    DataflowVariable<Path> getLazyImagePath(String rawCondaEnv) {
+        def condaEnv = normalizeCondaEnv(rawCondaEnv)
         if( condaEnv in condaPrefixPaths ) {
             log.trace "${binaryName} found local environment `$condaEnv`"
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
```diff
@@ -359,8 +359,8 @@ class CondaCache {
      *      The {@link DataflowVariable} which hold (and pull) the local image file
      */
     @PackageScope
-    DataflowVariable<Path> getLazyImagePath(String condaEnv) {
-
+    DataflowVariable<Path> getLazyImagePath(String rawCondaEnv) {
+        def condaEnv = normalizeCondaEnv(rawCondaEnv)
```
pditommaso, line 363: Instead of introducing a new function, what about using [condaPrefixPath](https://github.com/nextflow-io/nextflow/blob/5462d7a6b3343c4513e2c16faac5de34d4756eda/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy#L189-L189) and making sure it return an absolute  and normalized path ?

jorgee: I have implemented this change in PR [#5540](https://github.com/nextflow-io/nextflow/pull/5540)

PR #5542, opened 2024-11-26
Update syntax docs

bentsherman:
Minor corrections to syntax rules for strings and the workaround for dynamic config includes.


christopher-hakkaart (approved):

docs/vscode.md
```diff
@@ -333,6 +301,62 @@ To ease the migration of existing scripts, the language server only reports warn
...
+
+Groovy supports a wide variety of strings, including multi-line strings, dynamic strings, slashy strings, multi-line dynamic slashy strings, and more.
+
+The Nextflow language specification supports single- and double-quoted strings, multi-line strings, and slashy strings.
+
+Slashy strings cannot be interpolated:
+
+```nextflow
+def id = 'SRA001'
+assert 'SRA001.fastq' ~= /${id}\.f(?:ast)?q/
```
pditommaso, line 314: wasn't said to keep slashy string (with interpolation) and deprecate dollar slash strings?

bentsherman: neither are supported. I am looking into slashy strings [here](https://github.com/nextflow-io/language-server/pull/51) but until I can get it to work, they are effectively not supported

pditommaso: think we should consider a lack in the lang server, but i'd keep in the lang reference

bentsherman: Dynamic slashy strings are not mentioned in the language reference. The effort to support them (https://github.com/nextflow-io/language-server/pull/51) is purely exploratory. There is no guarantee that we can support it or that it would be wise to do so. Meanwhile, a regular dynamic string can be used by simply replacing `\` with `\\`.

I will add it in the future if I feel confident that we can support it.

PR #5560, opened 2024-12-03
Fixing PLAT-974

jorgee:
The getRepositoryURL in SCM Azure provider is returning an incomplete URL when organization-name/project-name/repo-name is provided. It is mainly caused because the project-name seems optional and the repo-name is only included RepositoryProvider.project variable when project-name is not provided. 
Just a adding a flag to indicate when project-name and use it in the getRepositoryURL to include the repo name seems enough to fix the issue. 
I think in Nextflow the getRepositoryURL is mainly used for logging purposes and unit tests, because there was no error pulling the repositories using both options (with and without project-name)

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/scm/AzureRepositoryProvider.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/AzureRepositoryProvider.groovy
@@ -35,6 +35,7 @@ final class AzureRepositoryProvider extends RepositoryProvider {
 
     private String user
     private String repo
+    private boolean includesRepo; //Flag to indicate if project already include the repo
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/scm/AzureRepositoryProvider.groovy
```diff
@@ -44,10 +45,13 @@ final class AzureRepositoryProvider extends RepositoryProvider {
          */
         def tokens = project.tokenize('/')
         this.repo = tokens.removeLast()
+
```
pditommaso, line 48:
```suggestion
```

PR #5561, opened 2024-12-03
Ensure job is killed when exception in task status check

jorgee:
If there is some exception in the task status check code, notifyError is called and the job is removed from the running queue preventing to be killed when doing the cleanup at abort.

The modification ensures the job is killed when no controlled process failures are notified.


modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy
```diff
@@ -573,6 +579,16 @@ class TaskPollingMonitor implements TaskMonitor {
                 checkTaskStatus(handler)
             }
             catch (Throwable error) {
+                // At this point NF assumes job is not running, but there could be errors at monitoring that could leave a job running (#5516).
+                // In this case, NF needs to ensure the job is killed.
+                if( error !instanceof ProcessException && error !instanceof ProcessRetryableException && error !instanceof CloudSpotTerminationException
+                     && error !instanceof ProcessEvalException && error !instanceof FailedGuardException) {
```
pditommaso, line 585: This condition is a bit crazy

pditommaso: would not the kill invoked during the shutdown phase ?

jorgee: The kill in the shutdown (cleanup) is done only for the tasks in the running queue. But in handleException the task is going to be removed from this queue (handleException->evict->remove). This is the case happening in the issue and why the kill is added. If you prefer I could create another queue (e.g crashedTasks) and do the kill of this tasks at the cleanup too.

pditommaso: Basically you are saying it manages to launch the job but it fails to move to the running queue (?)

jorgee: This is what it was happening in the issue:
- Tasks are submitted correctly and added to the running queue
- TaskPollingMonitor periodically checks the running queue
- In one of this checks, there is an exception in the code of the check. The exception management removes from the running queue the task whose check has generated the exception ( it is in the code path handleException->evict->remove).
- Finally, the shutdown cleanup kills all the tasks except this tasks because it is not in the running queue.
The solution that I added is killing the task when a "non-controlled" exception is produced to avoid to keep the task running without monitoring.

pditommaso: Got it! Instead trying determining the "right" exception to kill the task I would add `kill` in any case, adding an internal state to avoid double killing e.g. 

```
private boolean volatile terminated  

final void terminate() {
   if( !terminated ) {
     terminated = true
     kill()
   }
}
```

And use `terminated` in place of  `kill`

pditommaso:
Depends on #5594

bentsherman (approved):

modules/nextflow/src/main/groovy/nextflow/processor/TaskHandler.groovy
```diff
@@ -77,10 +80,22 @@ abstract class TaskHandler {
...
+    /**
+     * Kill a job execution.
+     *
+     * @see #killTask()
+     */
+    void kill() {
+        if (!killed.getAndSet(true)) {
+            killTask()
+        }
+    }
```
pditommaso, line 97: @jorgee Let's add a unit test to validate this logic and merge 🙏

jorgee: I have added the unit test

PR #5564, opened 2024-12-03
Update bin, lib, template docs

christopher-hakkaart:
Draft, do not merge.


docs/structure.md
```diff
@@ -0,0 +1,156 @@
+(structure-page)=
+
+# Structure
```
pditommaso, line 3: What does "structure" mean?

christopher-hakkaart: "Structure" of the project directory. Information about the project-level `bin` and `lib` directories currently sits on the [Sharing pipelines](https://www.nextflow.io/docs/latest/sharing.html) page, and it's hard to find.

Maybe "Project directory" would be better? Or something else?

docs/index.md
```diff
@@ -78,6 +78,7 @@ module
 notifications
 secrets
 sharing
+structure
```
bentsherman, line 81: I would put before sharing

docs/structure.md
bentsherman: I think "Project Structure" would be best

bentsherman:
I'm not entirely sold on this, but let's see where you take it.

In other words, I'm not sure it's worth having an entire page dedicated to project-level bin/lib/templates when we are trying to encourage people to use module directories. And in the long-term, trying to move people away from the lib directory in favor of nextflow scripts and custom plugins.

But if this page can absorb other concepts related to project structure, e.g. from the CLI and sharing pages, it might be worthwhile

christopher-hakkaart:
Hmmm, yeah, good point.

I'm not completely committed to a whole "Project structure" page, but I think we do need to describe what project-level bin/lib/templates are and how to use them. It might be a good place to mention main.nf and nextflow.config too. Alternatively, I'd be fine with blending this into another page, I'm just not sure which one.

Either way, I'll modify the text to push developers to module-level module and template directories, and away from the lib directory in favor of nextflow scripts and custom plugins.

bentsherman:
Closing due to age. Feel free to re-open if you ever decide to finalize it. Otherwise there isn't much point in keeping it open

PR #5567, opened 2024-12-04
Fix Google Batch hang when internal error during scheduling

jorgee:
When there is an internal error in Google Batch the job during scheduling. Jobs go from queued to failed. This situation is happening when request a VM type which is not available in the selected region. In this situation, there are no tasks in the batch job and NF is blocked in an infinite loop caused by the following part of code.

[nextflow/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy](https://github.com/nextflow-io/nextflow/blob/62fc0017ffae1ea151581f738c17ab840649aa0c/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy#L448-L449)

In this PR, this part is modified including a job status check to avoid to continue checking when job is failed.

I still have pending to decide what to do if job do not have tasks and job status is null. I think we should abort but @bentsherman do you remember why you returned PENDING when there are no tasks in the job? Is the some initial state where GCP is creating the tasks and we should consider PENDING?

### DIFF
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
@@ -62,7 +62,7 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
 
     private static Pattern EXIT_CODE_REGEX = ~/exit code 500(\d\d)/
 
-    private static Pattern BATCH_ERROR_REGEX = ~/Batch Error: code/
+    private static final Pattern BATCH_ERROR_REGEX = ~/Batch Error: code/
 
     private GoogleBatchExecutor executor
 
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
@@ -444,11 +444,27 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
      * @return Retrieve the submitted task state
      */
     protected String getTaskState() {
-        final tasks = client.listTasks(jobId)
-        if( !tasks.iterator().hasNext() )
-            return 'PENDING'
-
         final now = System.currentTimeMillis()
+        final tasks = client.listTasks(jobId)
+        if( !tasks.iterator().hasNext() ) {
+            // if there are no tasks checks the job status
+            final jobStatus = client.getJobStatus(jobId);
+            final newState = jobStatus?.state as String
+            if (newState) {
+                taskState = newState
+                timestamp = now
+                if (newState == "FAILED"){
+                    final eventsCount = jobStatus.getStatusEventsCount()
+                    final lastEvent = eventsCount > 0 ? jobStatus.getStatusEvents(eventsCount - 1) : null
+                    if (lastEvent){
+                        log.warn1 "Batch job failure: ${lastEvent.getDescription()}"
+                    }
+                }
+                return taskState
+            } else {
+                return "PENDING"
+            }
+        }
         final delta =  now - timestamp;
         if( !taskState || delta >= 1_000) {
             final status = client.getTaskStatus(jobId, taskId)
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
@@ -62,6 +62,8 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
 
     private static Pattern EXIT_CODE_REGEX = ~/exit code 500(\d\d)/
 
+    private static Pattern BATCH_ERROR_REGEX = ~/Batch Error: code/
+
     private GoogleBatchExecutor executor
 
     private Path exitFile
@@ -535,7 +537,7 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
             log.debug "[GOOGLE BATCH] Process `${task.lazyName()}` - last event: ${lastEvent}; exit code: ${lastEvent?.taskExecution?.exitCode}"
 
             final error = lastEvent?.description
-            if( error && EXIT_CODE_REGEX.matcher(error).find() ) {
+            if( error && (EXIT_CODE_REGEX.matcher(error).find() || BATCH_ERROR_REGEX.matcher(error).find())) {
                 return new ProcessException(error)
             }
         }
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
@@ -475,8 +475,8 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
         return taskState
     }
 
-    String checkJobStatus() {
-        final jobStatus = client.getJobStatus(jobId);
+    protected String checkJobStatus() {
+        final jobStatus = client.getJobStatus(jobId)
         final newState = jobStatus?.state as String
         if (newState) {
             taskState = newState
@@ -533,11 +533,11 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
     protected Throwable getJobError() {
         try {
             final events = noTaskJobfailure ? client.getJobStatus(jobId).getStatusEventsList() : client.getTaskStatus(jobId, taskId).getStatusEventsList()
-            final lastEvent = events?.get( events.size() -1 )
+            final lastEvent = events?.get(events.size() - 1)
             log.debug "[GOOGLE BATCH] Process `${task.lazyName()}` - last event: ${lastEvent}; exit code: ${lastEvent?.taskExecution?.exitCode}"
 
             final error = lastEvent?.description
-            if( error && (EXIT_CODE_REGEX.matcher(error).find() || BATCH_ERROR_REGEX.matcher(error).find())) {
+            if( error && (EXIT_CODE_REGEX.matcher(error).find() || BATCH_ERROR_REGEX.matcher(error).find()) ) {
                 return new ProcessException(error)
             }
         }
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
@@ -60,7 +60,7 @@ import nextflow.trace.TraceRecord
 @CompileStatic
 class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
 
-    private static Pattern EXIT_CODE_REGEX = ~/exit code 500(\d\d)/
+    private static final Pattern EXIT_CODE_REGEX = ~/exit code 500(\d\d)/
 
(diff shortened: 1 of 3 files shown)


plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -444,11 +444,27 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
...
+                    final eventsCount = jobStatus.getStatusEventsCount()
+                    final lastEvent = eventsCount > 0 ? jobStatus.getStatusEvents(eventsCount - 1) : null
+                    if (lastEvent){
+                        log.warn1 "Batch job failure: ${lastEvent.getDescription()}"
+                    }
+                }
+                return taskState
+            } else {
+                return "PENDING"
+            }
```
pditommaso, line 466: What about isolating this into a method `checkJobStatus()`, adding some tests whenever possible?

plugins/nf-google/src/test/nextflow/cloud/google/batch/GoogleBatchTaskHandlerTest.groovy
```diff
@@ -581,4 +584,40 @@ class GoogleBatchTaskHandlerTest extends Specification {
         and:
         0 * client.deleteJob('job1') >> null
     }
+
+    JobStatus makeJobStatus(JobStatus.State state, String desc = null) {
+        def builder = JobStatus.newBuilder()
+        builder = builder.setState(state)
```
pditommaso, line 590:
```suggestion
        final builder = JobStatus.newBuilder().setState(state)
```

Minor, likely can be simplified like this

plugins/nf-google/src/test/nextflow/cloud/google/batch/GoogleBatchTaskHandlerTest.groovy
```diff
@@ -581,4 +584,40 @@ class GoogleBatchTaskHandlerTest extends Specification {
...
+        if (desc)
+            builder = builder.addStatusEvents(
+                StatusEvent.newBuilder()
+                    .setDescription(desc)
+            )
```
pditommaso, line 595:
```suggestion
            builder.addStatusEvents(
                StatusEvent.newBuilder()
                    .setDescription(desc)
            )
```


likely not needed

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
jorgee: I have isolated in a method and added a test. I have also modified jobError to manage the error in this situations. I see in this [part of the code](https://github.com/nextflow-io/nextflow/blob/fc6312856885fc818007a77d7e7e072e254bff32/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy#L519) that if the error is due to an spot termination, it is throwing a ProcessException. 

Currently, I only print the error in the log and it is managed by as unknown error. 
```
Caused by:
  Process `FAILURE (1)` terminated for an unknown reason -- Likely it has been terminated by the external system

```
I could do the same as with spot looking for 'Batch error', but not sure if it will be correct. @pditommaso do you think leaving this errors as unexpected are correct, or should I generate a ProcessException as in the case of spot terminations?

plugins/nf-google/src/test/nextflow/cloud/google/batch/GoogleBatchTaskHandlerTest.groovy
```diff
@@ -581,4 +584,39 @@ class GoogleBatchTaskHandlerTest extends Specification {
...
+        client.getJobStatus(jobId) >>> [
+            null,
+            makeJobStatus(JobStatus.State.FAILED, 'Scheduling Failed'),
+        ]
+        then:
+        handler.getTaskState() == "PENDING"
+        handler.getTaskState() == "FAILED"
+
+
+
```
pditommaso, line 620:
```suggestion
```

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
pditommaso: If the Google backend provides a more details error description could make sense to use the same approach as spot ie. [this](https://github.com/nextflow-io/nextflow/blob/51aea8f389adbaff8e820d226337c47587ad3c57/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy#L519-L520)

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -62,6 +62,8 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
 
     private static Pattern EXIT_CODE_REGEX = ~/exit code 500(\d\d)/
 
+    private static Pattern BATCH_ERROR_REGEX = ~/Batch Error: code/
```
pditommaso, line 65:
```suggestion
    private static final Pattern BATCH_ERROR_REGEX = ~/Batch Error: code/
```

also the previous one above (!)

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -97,6 +99,10 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
     private volatile CloudMachineInfo machineInfo
 
     private volatile long timestamp
+    /**
+     * Flag to indicate job has failed without tasks
```
pditommaso, line 103:
```suggestion
     * A flag to indicate that the job has failed without launching any tasks
```

pditommaso: Gpt suggestion 😄

pditommaso (approved):

bentsherman:
> I still have pending to decide what to do if job do not have tasks and job status is null. I think we should abort but @bentsherman do you remember why you returned PENDING when there are no tasks in the job? Is the some initial state where GCP is creating the tasks and we should consider PENDING?

Yes, initially the job might not have any tasks, but we need to check the task state now to support job arrays, so when there are no tasks we just wait for tasks to be created. Checking the job status for failure makes sense

pditommaso:
@bentsherman approve if you think it's good

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -468,6 +475,21 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
         return taskState
     }
 
+    String checkJobStatus() {
+        final jobStatus = client.getJobStatus(jobId);
```
bentsherman, line 479:
```suggestion
    protected String checkJobStatus() {
        final jobStatus = client.getJobStatus(jobId)
```

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -510,13 +532,12 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
 
     protected Throwable getJobError() {
         try {
-            final status = client.getTaskStatus(jobId, taskId)
-            final eventsCount = status.getStatusEventsCount()
-            final lastEvent = eventsCount > 0 ? status.getStatusEvents(eventsCount - 1) : null
+            final events = noTaskJobfailure ? client.getJobStatus(jobId).getStatusEventsList() : client.getTaskStatus(jobId, taskId).getStatusEventsList()
+            final lastEvent = events?.get( events.size() -1 )
```
bentsherman, line 536:
```suggestion
            final lastEvent = events?.get(events.size() - 1)
```

plugins/nf-google/src/main/nextflow/cloud/google/batch/client/BatchClient.groovy
```diff
@@ -123,6 +124,10 @@ class BatchClient {
         return describeTask(jobId, taskId).getStatus()
     }
 
+    JobStatus getJobStatus(String jobId){
+        return describeJob(jobId).getStatus();
```
bentsherman, line 128:
```suggestion
        return describeJob(jobId).getStatus()
```

plugins/nf-google/src/test/nextflow/cloud/google/batch/GoogleBatchTaskHandlerTest.groovy
```diff
@@ -581,4 +584,38 @@ class GoogleBatchTaskHandlerTest extends Specification {
...
         0 * client.deleteJob('job1') >> null
     }
+
+    JobStatus makeJobStatus(JobStatus.State state, String desc = null) {
+        final builder = JobStatus.newBuilder().setState(state)
+        if (desc)
+            builder.addStatusEvents(
+                StatusEvent.newBuilder()
+                    .setDescription(desc)
+            )
```
bentsherman, line 594:
```suggestion
        if( desc ) {
            builder.addStatusEvents(
                StatusEvent.newBuilder()
                    .setDescription(desc)
            )
        }
```

plugins/nf-google/src/main/nextflow/cloud/google/batch/client/BatchClient.groovy
```diff
@@ -123,6 +124,10 @@ class BatchClient {
         return describeTask(jobId, taskId).getStatus()
     }
 
+    JobStatus getJobStatus(String jobId){
```
bentsherman, line 127:
```suggestion
    JobStatus getJobStatus(String jobId) {
```

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -510,13 +532,12 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
 
     protected Throwable getJobError() {
         try {
-            final status = client.getTaskStatus(jobId, taskId)
-            final eventsCount = status.getStatusEventsCount()
-            final lastEvent = eventsCount > 0 ? status.getStatusEvents(eventsCount - 1) : null
+            final events = noTaskJobfailure ? client.getJobStatus(jobId).getStatusEventsList() : client.getTaskStatus(jobId, taskId).getStatusEventsList()
```
bentsherman, line 535:
```suggestion
            final events = noTaskJobfailure
                ? client.getJobStatus(jobId).getStatusEventsList()
                : client.getTaskStatus(jobId, taskId).getStatusEventsList()
```

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -97,6 +99,10 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
     private volatile CloudMachineInfo machineInfo
 
     private volatile long timestamp
+    /**
```
bentsherman, line 102:
```suggestion

    /**
```

bentsherman:
It would be nice to test against a real error case, but not sure if it can be easily replicated. Aside from that, the changes look good to me

bentsherman (approved):

pditommaso:
Good point, Esha reported the issue she has a test case

pditommaso:
Tested internally

PR #5578, opened 2024-12-05
Fail the run if publish thread pool times out

### LINKED ISSUE #5577: Fail the run if the publish thread pool times out
The entire purpose of a pipeline run is to produce a particular set of results. If any of these results are missing for any reason, the run should be considered a failure. Otherwise the user has a false sense of success when in fact some results may be missing.

When all jobs are complete, the run waits for these two thread pools to finish: https://github.com/nextflow-io/nextflow/blob/308d5a5e893a7b3691d58025cbf673d62e1d7481/modules/nextflow/src/main/groovy/nextflow/Session.groovy#L691-L693

By default they will hard shutdown after 12 hours: https://github.com/nextflow-io/nextflow/blob/308d5a5e893a7b3691d58025cbf673d62e1d7481/modules/nextflow/src/main/groovy/nextflow/util/ThreadPoolHelper.groovy#L47-L51

Currently only a warning is logged, but the run is still considered successful as long as all jobs completed successfully (or were ignored).

Instead, the run should fail in this case. As a compromise, we could fail the run only if `workflow.output.ignoreErrors` is false, so that the user can keep the existing behavior if they want. But the default behavior should be to not declare success until all results have been published.

bentsherman:
Close #5577


bentsherman:
@pditommaso ready for review. It will fail the run if `workflow.output.ignoreErrors` is false and the finalizer or publish pool times out

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -682,15 +682,23 @@ class Session implements ISession {
...
+            final publisherComplete = publishPoolManager?.shutdown(false)
+            if( !finalizerComplete || !publisherComplete ) {
+                final ignoreErrors = config.navigate('workflow.output.ignoreErrors')
+                if( !ignoreErrors )
+                    throw new AbortOperationException("Timed out while waiting to publish outputs")
+            }
+        }
+        else {
+            finalizePoolManager?.shutdown(true)
+            publishPoolManager?.shutdown(true)
```
bentsherman, line 695: Moving the shutdown here is effectively the same as before because `destroy()` is called right after `await()` in the script runner.

Although, `destroy()` is not called in the event of an abort, which means that previously the thread pools were only ever shut down when `aborted == false`.

So we can either keep these lines here, move them to `abort()`, or remove them entirely, since they were previously unreachable anyway

pditommaso:
Would have not been simpler to throw an exception [here](https://github.com/nextflow-io/nextflow/blob/237894a796a55fbf99317e777e3e5cc8dcf5635b/modules/nextflow/src/main/groovy/nextflow/util/ThreadPoolHelper.groovy#L47-L51) and rely on the abort mechanism?

bentsherman:
Probably would be simpler. In that case the `abort()` would be called in the middle of `destroy()`, is that expected to work?

pditommaso:
I think the `abort` will be called implicitly by throwing an exception. see here

https://github.com/nextflow-io/nextflow/blob/237894a796a55fbf99317e777e3e5cc8dcf5635b/modules/nextflow/src/main/groovy/nextflow/script/ScriptRunner.groovy#L150-L150

bentsherman:
Right, just wondering if `abort()` has some implicit assumption that `destroy()` has not been called yet. But from the code it looks safe. Simplified things a bit

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -689,8 +690,17 @@ class Session implements ISession {
...
             log.trace "Session > destroying"
             // shutdown thread pools
-            finalizePoolManager?.shutdown(aborted)
-            publishPoolManager?.shutdown(aborted)
+            try {
+                finalizePoolManager?.shutdown(aborted)
+                publishPoolManager?.shutdown(aborted)
+            }
+            catch( TimeoutException e ) {
+                final ignoreErrors = config.navigate('workflow.output.ignoreErrors', false)
```
pditommaso, line 698: Was this flag introduced to handle this case? Wonder if [strict](https://github.com/nextflow-io/nextflow/blob/5577-fail-on-publish-timeout/modules/nextflow/src/main/groovy/nextflow/NF.groovy#L63-L63) mode should not be used instead?

bentsherman: Actually the ignoring of publish errors can be controlled by either `nextflow.enable.strict` or `workflow.output.ignoreErrors`, so it would be consistent to check for both here.

I generally try to only use feature flags for language features though

pditommaso: ok, just made some minor changes

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -689,8 +690,17 @@ class Session implements ISession {
...
+                finalizePoolManager?.shutdownOrAbort(aborted,this)
+                publishPoolManager?.shutdownOrAbort(aborted,this)
+            }
+            catch( TimeoutException e ) {
+                final ignoreErrors = config.navigate('workflow.output.ignoreErrors', false)
+                if( !ignoreErrors )
+                    throw new AbortOperationException("Timed out while waiting to publish outputs")
+                else
+                    log.warn e.message
+            }
```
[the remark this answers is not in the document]
pditommaso: 💯 👉 00376aa3

PR #5600, opened 2024-12-11
Fix race condition when resuming an aborted run

### LINKED ISSUE #5595: Resume from crashed head job caused process and its downstream dependency to run at the same time
## Bug report 

Due to an LSF compute cluster outage, two of our pipelines ended up in a state where the nextflow process was unable to submit jobs to the cluster and crashed. However, the running processes completed successfully.

When we resumed the pipeline after the cluster had recovered from the crash, we saw nextflow 
- launch the processes that had completed successfully
- immediately conclude the process had completed and launch a process that depended on its output

The first launched process remained running for another ten minutes and wrote output files while its dependent process ran, leading to truncated output and a false negative result.

After some digging, I believe that what happened is related to the behaviour of the GridTaskHandler, specifically this line:

https://github.com/nextflow-io/nextflow/blob/18f7de132c0342102f7160ce97b2e68dc6b67[…]xtflow/src/main/groovy/nextflow/executor/GridTaskHandler.groovy

The function  `readExitStatus()` looks for a file called `.exitfile` and reads the exit code of the completed process from there. Since nextflow resumed a process that had successfully completed after the head job crashed, that file was already present in the work directory when the pipeline resumed and this caused a premature launch of the downstream process.

I think an easy fix would be to make sure the .exitfile is deleted before submitting a process, possibly just with an `rm -f .exitfile`

Expected behaviour: when resuming a pipeline, nextflow should wait for the resumed process to complete executing on the cluster.
Actual behaviour: nextflow immedately launches a downstream process while still running.

The following script illustrates the problem. It has a different output on a grid scheduler than when run locally.
A similar result can be achieved by removing the line `echo 0 > .exitcode`, launching nextflow on a grid scheduler, killing nextflow when the process has started, and then relaunching with -resume.

```
workflow {
    messWithExitCode | view
}


process messWithExitCode {
    beforeScript 'module load cluster/test'

    output:
    stdout

    script:
    """
    echo On a Grid scheduler, this line will print
    # the line below is just for demo purposes. On our grid cluster, this was achieved by
    # - letting nextflow launch the child process
    # - killing the nextflow executor
    # - resuming the workflow after the child process had completed
    echo 0 > .exitcode
    sleep 60s
    echo On a Grid scheduler, this line will not print
    """
}
```

### Program output 

On a grid scheduler, the above will only print
```
echo On a Grid scheduler, this line will print
```

### Environment 

* Nextflow version: 22.04.5
* Java version: 1.8.0
* Operating system: linux
* Bash version: 4.3.48(1)-release

### Additional context

N/A

bentsherman:
Close #5595 

When a Nextflow run is aborted for any reason, Nextflow will try to kill all jobs and save them to the cache as failed tasks, so that on a resumed run, the new jobs will use new task directories.

However, if Nextflow is unable to complete this cleanup for any reason, such as an abrupt crash, the task cache might not be updated. On a resumed run, the new task will be re-executed in the same directory as before:

https://github.com/nextflow-io/nextflow/blob/b23e42cd6dea0e4ce6bba19c13968546807be5f1/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy#L806-L826

What I think happens here is:
1. the cache entry is missing
2. Nextflow doesn't check whether the task dir exists
3. Nextflow proceeds to re-use the task dir with a new task

This can cause a race condition because if the begin and exit files are still present, Nextflow could submit a job, detect the begin/exit codes from the previous job, mark the job as completed (as long as the required outputs are still present), and launch downstream tasks on truncated data.

I think this can only happen on grid executors because the GridTaskHandler checks the begin/exit files before polling the scheduler. The cloud executors poll their API first, and the local executor doesn't check these files at all.

It should be possible to replicate this issue on a grid executor by cancelling a run with `-disable-jobs-cancellation` and allowing the jobs to complete before resuming the run. Which by the way is an unsafe property of that CLI option.

I'm trying to modify the local executor to simulate the issue but haven't been able to replicate yet

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -829,7 +829,7 @@ class TaskProcessor {
 
             final lock = lockManager.acquire(hash)
             try {
-                if( !exists && !workDir.mkdirs() )
+                if( !workDir.mkdirs() )
                     throw new IOException("Unable to create directory=$workDir -- check file system permissions")
             }
             finally {
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -803,16 +803,18 @@ class TaskProcessor {
         while( true ) {
             hash = HashBuilder.defaultHasher().putBytes(hash.asBytes()).putInt(tries).hash()
 
-            Path resumeDir = null
+            Path workDir = null
             boolean exists = false
             try {
                 final entry = session.cache.getTaskEntry(hash, this)
-                resumeDir = entry ? FileHelper.asPath(entry.trace.getWorkDir()) : null
-                if( resumeDir )
-                    exists = resumeDir.exists()
+                workDir = entry
+                    ? FileHelper.asPath(entry.trace.getWorkDir())
+                    : task.getWorkDirFor(hash)
+                if( workDir )
+                    exists = workDir.exists()
 
-                log.trace "[${safeTaskName(task)}] Cacheable folder=${resumeDir?.toUriString()} -- exists=$exists; try=$tries; shouldTryCache=$shouldTryCache; entry=$entry"
-                final cached = shouldTryCache && exists && entry.trace.isCompleted() && checkCachedOutput(task.clone(), resumeDir, hash, entry)
+                log.trace "[${safeTaskName(task)}] Cacheable folder=${workDir?.toUriString()} -- exists=$exists; try=$tries; shouldTryCache=$shouldTryCache; entry=$entry"
+                final cached = shouldTryCache && exists && entry && entry.trace.isCompleted() && checkCachedOutput(task.clone(), workDir, hash, entry)
                 if( cached )
                     break
             }
@@ -826,10 +828,7 @@ class TaskProcessor {
             }
 
             final lock = lockManager.acquire(hash)
-            final workDir = task.getWorkDirFor(hash)
             try {
-                if( resumeDir != workDir )
-                    exists = workDir.exists()
                 if( !exists && !workDir.mkdirs() )
                     throw new IOException("Unable to create directory=$workDir -- check file system permissions")
             }


bentsherman:
Alternative solution would be for the task to always delete `.command.begin` and `.exitcode` as a precaution. However I think we generally assume that a task should never reuse an existing work directory since the presence of old outputs could cause other problems, hence the current approach

pditommaso:
Not sure to understand on what resource the race condition happens?

bentsherman:
Maybe it's not so much a race condition as it is a cache corruption

Nextflow re-executes a task in the same directory as before, and because the previous `.command.begin` and `.exitcode` are present, Nextflow immediately thinks the task is finished

pditommaso:
Basically we are saying the task fails `0` in the `.exitcode` ?

bentsherman:
The task completes with exit code `0` because maybe Nextflow was previously killed before it was able to cancel the job. So the job runs to completion and writes the exit file. But there is no cache entry for this task so Nextflow assumes the task directory is safe to use.

pditommaso:
ok, got it

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -803,16 +803,18 @@ class TaskProcessor {
...
+                workDir = entry
+                    ? FileHelper.asPath(entry.trace.getWorkDir())
+                    : task.getWorkDirFor(hash)
+                if( workDir )
+                    exists = workDir.exists()
 
-                log.trace "[${safeTaskName(task)}] Cacheable folder=${resumeDir?.toUriString()} -- exists=$exists; try=$tries; shouldTryCache=$shouldTryCache; entry=$entry"
-                final cached = shouldTryCache && exists && entry.trace.isCompleted() && checkCachedOutput(task.clone(), resumeDir, hash, entry)
+                log.trace "[${safeTaskName(task)}] Cacheable folder=${workDir?.toUriString()} -- exists=$exists; try=$tries; shouldTryCache=$shouldTryCache; entry=$entry"
+                final cached = shouldTryCache && exists && entry && entry.trace.isCompleted() && checkCachedOutput(task.clone(), workDir, hash, entry)
```
pditommaso, line 817: This is one of that piece of code the lesser the changes the better, both to make it simple to review and to keep history readable

bentsherman: I know. This is about as simple as it gets while keeping the intent clear.

1. check for a cache entry and it's work dir, or compute the work dir if there is no cache entry
2. check whether the work dir exists
3. check for cached outputs if the cache entry and work dir exist and the task completed

and further down:
4. if the outputs are cached then use them
5. otherwise, if the work dir exists then use a new work dir
6. otherwise, create the work dir and use it

pditommaso: ok, getting better, but why adding `entry &&` in the `cached` condition? If the intent is to applied the same logic when the `entry` is missing should not the condition remain the same?

bentsherman: In the previous logic, it wouldn't check for cached outputs if the cache entry was missing. That is also how it behaves here. The `&& entry` is required to prevent a null reference exception on `entry.trace.isCompleted()`

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -826,11 +828,8 @@ class TaskProcessor {
             }
 
             final lock = lockManager.acquire(hash)
-            final workDir = task.getWorkDirFor(hash)
             try {
-                if( resumeDir != workDir )
-                    exists = workDir.exists()
-                if( !exists && !workDir.mkdirs() )
+                if( !workDir.mkdirs() )
```
bentsherman, line 832: I've been wondering about this lock over the task directory. I think the purpose is to prevent two tasks from using the same directory. But in that case maybe it should be over the previous try-catch block?

It should prevent two tasks from checking the same directory at the same time, because that is how a task determines whether to use the directory or try a different one

pditommaso: It may. however what I'm thinking that to solve this issue it should be assumed a new task run should *always* use a newly created directory. Not sure this logic satisfy it.

pditommaso:
Closing in favour https://github.com/nextflow-io/nextflow/pull/5682. let me know if you have doubts regarding that

bentsherman:
I think my approach makes the code easier to read, but either one should fix the original issue

PR #5610, opened 2024-12-13
Default revision

### LINKED ISSUE #4427: Set a default revision
## New feature

In the manifest it would be nice if we could default to a git tag or commit hash.

## Usage scenario 

```nextflow
manifest {
    version         = '1.0.0'
    defaultBranch   = '1.0.0'
    // defaultRevision = '1.0.0'
}
```

So we don't need a `master` branch that is just a "stable" release, instead we can have users checkout a version by default and they won't get a dev version by mistake.

## Suggest implementation 

1. Add a defaultRevision or modify the defaultBranch functionality to accept tags or commit hashs. The same functionality as `-r`
2. Set the default behavior to look for a git tag, and if there isn't one default to master.
3. If a manifest.version is set, then look for that git tag instead of using the defaultBranch on the git repo.

edmundmiller:
Closes #4427 

Need to clean up this PR a bit, and test it out in a real-world, not just in tests.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/config/Manifest.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/config/Manifest.groovy
@@ -58,8 +58,12 @@ class Manifest {
         target.defaultBranch
     }
 
+    String getDefaultRevision() {
+        target.defaultRevision ?: getVersion() ?: getDefaultBranch()
+    }
+
     String getDescription() {
-        target.description 
+        target.description
     }
 
     String getAuthor() {
@@ -139,6 +143,7 @@ class Manifest {
             .map(c -> c.toMap())
             .collect(Collectors.toList())
         result.defaultBranch = getDefaultBranch()
+        result.defaultRevision = getDefaultRevision()
         result.description = getDescription()
         result.homePage = homePage
         result.gitmodules = getGitmodules()
--- a/modules/nextflow/src/main/groovy/nextflow/config/Manifest.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/config/Manifest.groovy
@@ -55,12 +55,17 @@ class Manifest {
 
 
     String getDefaultBranch() {
-        target.defaultBranch
+        target.defaultBranch ?: 'master'
     }
 
-    // 1. if defaultRevision is set, use that
-    // 2. if version is set, use that
-    // 3. Fallback to defaultBranch
+    /**
+     * Gets the default revision to use
+     * Priority order:
+     * 1. defaultRevision if set
+     * 2. version if set
+     * 3. defaultBranch
+     * @return The revision to use
+     */
     String getDefaultRevision() {
         target.defaultRevision ?: getVersion() ?: getDefaultBranch()
     }
@@ -111,6 +116,10 @@ class Manifest {
         target.nextflowVersion
     }
 
+    /**
+     * Gets the version
+     * @return The version string or null if not set
+     */
     String getVersion() {
         target.version
     }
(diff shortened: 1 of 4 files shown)


modules/nextflow/src/main/groovy/nextflow/config/Manifest.groovy
```diff
@@ -55,11 +55,23 @@ class Manifest {
 
 
     String getDefaultBranch() {
-        target.defaultBranch
+        target.defaultBranch ?: 'master'
```
tom-seqera, line 58: A git repository has its own concept of a 'default' branch, which is usually `master` or `main` - but can be absolutely anything. If the manifest doesn't specify a default branch/revision, we want to use the default branch set in git.

So, rather than assuming 'master', this method needs to return null if no value is specified in the manifest file. This  allows the logic in `AssetManager.getDefaultBranch()` to interrogate git's default branch.

modules/nextflow/src/main/groovy/nextflow/config/Manifest.groovy
```diff
@@ -55,11 +55,23 @@ class Manifest {
...
+    /**
+     * Gets the default revision to use
+     * Priority order:
+     * 1. defaultRevision if set
+     * 2. version if set
+     * 3. defaultBranch
+     * @return The revision to use
+     */
+    String getDefaultRevision() {
+        target.defaultRevision ?: getVersion() ?: getDefaultBranch()
```
tom-seqera, line 70: I don't think 'version' should be in this list at all.

'defaultRevision'/'defaultBranch' are pointers to a (potentially) different branch/revision from the one checked out.

Version, if set, would usually represent the version of the checked out branch/revision, so I think including it in the list effectively means that whichever branch is checked out will become the default. It seems like this could result in surprising behaviour. Probably better only to rely on manifest fields which are explicitly related to defaults.

modules/nextflow/src/main/groovy/nextflow/config/Manifest.groovy
```diff
@@ -193,4 +210,100 @@ class Manifest {
...
+        // Development versions are always compatible
+        if (isDevelopmentVersion()) return true
+
+        // For release candidates and hotfixes, compare base versions
+        def v1 = version.replaceAll(/(-RC\d+|-dev|dev|-hotfix)$/, '')
+        def v2 = otherVersion.replaceAll(/(-RC\d+|-dev|dev|-hotfix)$/, '')
+
+        // Version should be greater than or equal to the other version
+        compareVersions(v1, v2) >= 0
+    }
```
tom-seqera, line 308: All of these version-related methods seem to be assuming/enforcing a very specific versioning scheme on pipelines. I don't think Nextflow can make these kind of assumptions/expectations since any organisation could be writing their own Nextflow pipelines and might use a different versioning scheme which is completely incompatible.

It doesn't look like any of these methods are used in this PR though, so I think they can just be removed.

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
```diff
@@ -937,9 +937,18 @@ class AssetManager {
         assert localPath
 
         def current = getCurrentRevision()
-        if( current != defaultBranch ) {
+        def defaultRev = getManifest().getDefaultRevision()
```
tom-seqera, line 940: This shouldn't be asking the manifest directly; it needs to use the `defaultBranch` logic in `AssetManager` to handle cases where no value is specified in the manifest.

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
```diff
@@ -937,9 +937,18 @@ class AssetManager {
...
+        if( current != defaultRev ) {
+            // NOTE This is the issue
             if( !revision ) {
-                throw new AbortOperationException("Project `$project` is currently stuck on revision: $current -- you need to explicitly specify a revision with the option `-r` in order to use it")
+                Ref head = git.getRepository().findRef(Constants.HEAD);
+
+                // try to resolve the the current object id to a tag name
+                Map<ObjectId, String> names = git.nameRev().addPrefix( "refs/tags/" ).add(head.objectId).call()
+                def tag = names.get( head.objectId ) ?: head.objectId.name()
+                if( current != tag ) {
```
tom-seqera, line 949: I'm not quite sure I understand this change.

It looks like the new logic boils down to "don't throw the error if the current revision is a tag (any tag)". Is that the intention?

modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
```diff
@@ -602,7 +547,7 @@ class AssetManagerTest extends Specification {
     }
 
     @Requires({System.getenv('NXF_GITHUB_ACCESS_TOKEN')})
-    def 'should identify default branch when downloading repo'() {
+    def "should use a default tag"() {
```
tom-seqera, line 550: This test name (`'should identify default branch when downloading repo'`) was correct and shouldn't be changed. It tests that if the repo manifest doesn't explicitly specify a default branch (or revision) then Nextflow can correctly identify the default branch configured at the git level. In the case of the 'socks' repo, this is 'main'.

modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
```diff
@@ -623,32 +568,470 @@ class AssetManagerTest extends Specification {
...
+        holder.build('foo/bar')
+
         then:
-        remote_head != null
-        !AssetManager.isRemoteBranch(remote_head)
+        holder.manifest.getDefaultRevision() == '1.0.0'
+        holder.manifest.getDefaultBranch() == 'master'
+    }
+
+    def "should use version as defaultRevision when available"() {
```
tom-seqera, line 616: As mentioned, I'm not sure 'version' should be involved in determining the default revision/branch.

modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
```diff
@@ -623,32 +568,470 @@ class AssetManagerTest extends Specification {
         noExceptionThrown()
     }
 
-    @Requires({System.getenv('NXF_GITHUB_ACCESS_TOKEN')})
-    def 'can filter remote branches'() {
```
tom-seqera, line 627: This test (`'can filter remote branches'`) should not be removed.

modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
```diff
@@ -71,216 +72,141 @@ class AssetManagerTest extends Specification {
...
             defaultBranch = config.getString('init', null, 'defaultBranch') ?: 'master'
         }
-        return defaultBranch
+        defaultBranch
     }
 
-    def testList() {
-
-        given:
+    def "should list available pipeline assets"() {
```
tom-seqera, line 87: While I agree with efforts to improve the naming and clarity of the tests, doing so alongside changes to behaviour changes introduces a lot of noise. I'd suggest splitting such changes into a separate PR which improves the tests without modifying any behaviour.

edmundmiller: I think this is a great idea. Let me do that first. This PR got convoluted because I wanted to make sure any changes I made were going to be backwards compatible 😄

pditommaso:
This looks a bit abandoned. Closing it for now.

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
edmundmiller: This just behaves really strangely currently. I'll try to iron out what the edge cases are in a follow-up PR

PR #5614, opened 2024-12-16
[FSN-13] feat(nf-tower): Retrieve and set Fusion license tokens in a process environment

alberto-miranda:
# Description

This PR adds a new `TowerFusionEnv` class to `nf-tower` that enables setting Platform-specific environment variables to Nextflow processes running with Fusion enabled. The class implements the `FusionEnv` interface and, as such, provides a `getEnvironment()` method that returns the relevant key-value pairs.

Currently, the provider is only being used to set the `FUSION_LICENSE_TOKEN` environment variable, the value of which is retrieved by sending a specific HTTP request to Platform in order to validate the run. Once the run is validated, Platform will reply with a corresponding token which will be injected into the process environment.

>[!note]
>Since fusion licenses are still not mandatory, if the license cannot be validated (or there's a configuration error affecting this validation process) a warning is emitted:
>
> ```console
>user@host$ nextflow run hello
>
 >N E X T F L O W   ~  version 24.11.0-edge
>
>Launching `https://github.com/nextflow-io/hello` [mighty_liskov] DSL2 - revision: afff16a9b4 [master]
>
>executor >  local (fusion enabled) (4)
>[ec/6b82ba] process > sayHello (4) [100%] 4 of 4 ✔
>Hello world!
>
>Hello world!
>
>Ciao world!
>
>Ciao world!
>
>Bonjour world!
>
>Bonjour world!
>
>Hola world!
>
>Hola world!
>
>WARN: Error retrieving Fusion license information: Missing personal access token -- Make sure there's a variable >TOWER_ACCESS_TOKEN in your environment
>WARN: Error retrieving Fusion license information: Missing personal access token -- Make sure there's a variable >TOWER_ACCESS_TOKEN in your environment
>WARN: Error retrieving Fusion license information: Missing personal access token -- Make sure there's a variable >TOWER_ACCESS_TOKEN in your environment
>WARN: Error retrieving Fusion license information: Missing personal access token -- Make sure there's a variable >TOWER_ACCESS_TOKEN in your environment
>```

## Testing
This PR can be tested by following [this guide](https://github.com/seqeralabs/fusion-team/blob/master/docs/license/license-e2e-manual-testing.md).


modules/nextflow/src/main/groovy/nextflow/fusion/FusionClient.groovy
```diff
@@ -0,0 +1,204 @@
...
+import java.util.concurrent.Executors
+import java.util.function.Predicate
+
+/**
+ * Client service for Fusion-related operations
+ *
+ * @author Alberto Miranda <alberto.miranda@seqera.io>
+ */
+@CompileStatic
+class FusionClient {
```
pditommaso, line 36: I'd prefer implementing this logic as part of [TowerClient](https://github.com/nextflow-io/nextflow/blob/main/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy#L57-L57) since we said Platform token is going to be a pre-requisite to use fusion

pditommaso: This should be doable having the Tower plugin implementing its own extension for `FusionEnv` see here

https://github.com/nextflow-io/nextflow/blob/main/modules/nextflow/src/main/groovy/nextflow/fusion/FusionEnvProvider.groovy#L40-L45

alberto-miranda: Got it, I separated it because I confused `TowerClient` as being part of the Wave plugin, but if it's part of the tower plugin it makes more sense to have it there.

alberto-miranda:
@pditommaso I have refactored the implementation into `TowerClient` as you suggested. A couple of comments:

1. I'm not very fond of making `TowerClient` implement `FusionEnv`: I'd rather have a dedicated `TowerFusionEnv` class with the `getEnvironment()` method and helpers, but then I cannot reuse the client itself or the `sendHttpMessage()` and co. functions. Any advice here?
2. It doesn't seem to be picking up `TowerClient` as an extension despite having defined it in `plugins/nf-tower/src/resources/META-INF/extensions.idx`. As such, it's not using it (e.g. [this should fail](https://github.com/nextflow-io/nextflow/blob/6fc8fbe5b4027c5bf96ecb6f094b526aa64524a1/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy#L85)) and it's not failing. Is there anything else I need to do to register the extension?

pditommaso:
> I'm not very fond of making TowerClient implement FusionEnv: I'd rather have a dedicated TowerFusionEnv class

I agree, sorry if I may have suggested in that direction. My point was to move this logic in the tower client module 

> then I cannot reuse the client itself or the sendHttpMessage()

alberto-miranda:
@pditommaso I refactored the implementation based on our discussions. The PR is ready to review.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
```diff
@@ -0,0 +1,266 @@
...
+     * @param opts the configuration options for Platform
+     * @param env the applicable environment variables
+     * @return the Platform access token
+     */
+    protected static String accessToken0(Map opts, Map<String, String> env) {
+        def token = env.get('TOWER_WORKFLOW_ID')
+            ? env.get('TOWER_ACCESS_TOKEN')
+            : opts.containsKey('accessToken') ? opts.accessToken as String : env.get('TOWER_ACCESS_TOKEN')
+        return token
+    }
```
alberto-miranda, line 196: This basically replicates what `TowerConfig` does in `nf-wave`, but I didn't want to add a explicit dependency on that.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
```diff
@@ -0,0 +1,266 @@
...
+     * @throws AbortOperationException if a Platform access token cannot be found
+     *
+     * @return The signed JWT token
+     */
+    protected String getLicenseToken(product, version) {
+        // FIXME(amiranda): Find out how to obtain the product and version
+        // Candidate: FusionConfig?
+
+        if (accessToken == null) {
+            throw new AbortOperationException("Missing personal access token -- Make sure there's a variable TOWER_ACCESS_TOKEN in your environment")
```
alberto-miranda, line 124: I'm not sure if this is the correct exception to throw in this context. Please double check.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
```diff
@@ -0,0 +1,266 @@
...
+     * @param config The Fusion configuration object
+     * @return A map of environment variables
+     */
+    @Override
+    Map<String, String> getEnvironment(String scheme, FusionConfig config) {
+
+        // TODO(amiranda): Hardcoded for now. We need to find out how to obtain
+        // the concrete product SKU and version. Candidate: FusionConfig?
+        final product = 'fusion'
+        final version = '2.4'
```
pditommaso, line 98: `FusionConfig` should be the right place

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
```diff
@@ -0,0 +1,266 @@
...
+
+        try {
+            final token = getLicenseToken(product, version)
+            return [
+                'FUSION_LICENSE_TOKEN': token,
```
pditommaso, line 103:
```suggestion
                FUSION_LICENSE_TOKEN: token,
```

Groovy magic

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
```diff
@@ -0,0 +1,266 @@
...
+            ]
+        } catch (Exception e) {
+            log.warn("Error retrieving Fusion license information: ${e.message}")
+            return [:]
```
pditommaso, line 107:
```suggestion
            return Map.of()
```

If generally preferable because returns always the same immutable empty map

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
pditommaso: Should be ok

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
```diff
@@ -0,0 +1,266 @@
...
+    protected String getLicenseToken(product, version) {
+        // FIXME(amiranda): Find out how to obtain the product and version
+        // Candidate: FusionConfig?
+
+        if (accessToken == null) {
+            throw new AbortOperationException("Missing personal access token -- Make sure there's a variable TOWER_ACCESS_TOKEN in your environment")
+        }
+
+        final req = HttpRequest.newBuilder()
+            .uri(URI.create("${endpoint}/${LICENSE_TOKEN_ENDPOINT}").normalize())
```
pditommaso, line 128: `LICENSE_TOKEN_ENDPOINT` should not be a fully qualified URI

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
```diff
@@ -0,0 +1,266 @@
...
+            .header('Authorization', "Bearer ${accessToken}")
+            .POST(
+                HttpRequest.BodyPublishers.ofString(
+                    new Gson().toJson(
+                        new LicenseTokenRequest(
+                            product: product,
+                            version: version
+                        ),
+                        LicenseTokenRequest.class
+                    ),
```
pditommaso, line 139: Built-in `JsonOutout.toJson(Map)` can be easier

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
pditommaso: Could make sense to turn those into `PlatformHelper` class shared across modules & plugins

pditommaso (changes requested):
Looks good, left some comments. Question, what happens when using Fusion and Tower plugin is not enabled?

alberto-miranda:
> Looks good, left some comments. Question, what happens when using Fusion and Tower plugin is not enabled?

Nothing special, if `nf-tower` is enabled but Fusion is not, the `TowerFusionEnv` itself is never instantiated, and there are no calls made to retrieve a token. The same happens if both are disabled, but in this case `nf-tower` is also never loaded.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
alberto-miranda: I don't follow: `LICENSE_TOKEN_ENDPOINT` is set to `license/token/`, so what am I missing? Could you elaborate?

pditommaso:
Will Fusion be able to run?

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
alberto-miranda: I agree. Do you want it done in the context of this PR?

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
alberto-miranda: I did it this way for symmetry with the [deserialization](https://github.com/nextflow-io/nextflow/blob/e7b10823f5937cfcf32c76f165651300e6216166/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy#L149) but I don't have a preference for one or the other.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
pditommaso: an endpoint is usually a fully qualified URI, see [here](https://github.com/nextflow-io/nextflow/blob/6183fdf9dbd9114f5ee86d11bc5e69cbd06501a6/plugins/nf-wave/src/main/io/seqera/wave/plugin/config/TowerConfig.groovy#L53-L53) for example

This looks like more a path, always is the constant really needed?

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
pditommaso: it could be better

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
pditommaso: I'm fine keeping gson, then it would be better to isolate into its own method

alberto-miranda:
> Will Fusion be able to run?

The license is not yet enforced in the Fusion side, so yes.

A side effect we didn't consider is that if `fusion` is run without `nf-tower` the license is not enforced currently (because `TowerFusionEnv` is not loaded in this case). Is this a use case that should concern us?

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
alberto-miranda: Ah, I get what you mean now. I can either rename it to something like `LICENSE_TOKEN_PATH` or just get rid of the constant altogether. I generally prefer to use constants in case we ever decide to change it in the future :shrug:.

alberto-miranda: Addressed with 20c8a535dca9cb55e84f19682cfe3031baa75309

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
alberto-miranda: Fixed with 25cd5ae9987a5bc518a0abed150997ca759b0991

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
alberto-miranda: Fixed with f32f0ca129e35e1305cb8389f0d83d527441ee16

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
alberto-miranda: Refactored with c1147cdc922a9a5423e1ea5131cbf7c5a70ce497. I created the `PlatformHelper` class under `modules/nextflow/src/main/groovy/nextflow/platform` since it seemed to fit with the existing conventions for common code. Please double check.

pditommaso:
> Even if not enforced by nextflow, fusion will still fail once licenses become mandatory, but the error received by the user will probably be uglier

and slower, ie. after the job started. we should consider adding a "default" `TowerFusionEnv` implementation of this execution reporting a warning message. 

The "default" `TowerFusionEnv` is "overridden" by the one provided by tower plugin when it's enabled 

See here for example 

https://github.com/nextflow-io/nextflow/blob/9eefd207655a962d3c945967a8f86e419cdcf5a4/plugins/nf-wave/src/main/io/seqera/wave/plugin/resolver/WaveContainerResolver.groovy#L41-L41

alberto-miranda:
> > Even if not enforced by nextflow, fusion will still fail once licenses become mandatory, but the error received by the user will probably be uglier
> 
> 
> 
> and slower, ie. after the job started. we should consider adding a "default" `TowerFusionEnv` implementation of this execution reporting a warning message. 
> 
> 
> 
> The "default" `TowerFusionEnv` is "overridden" by the one provided by tower plugin when it's enabled 
> 
> 
> 
> See here for example 
> 
> 
> 
> https://github.com/nextflow-io/nextflow/blob/9eefd207655a962d3c945967a8f86e419cdcf5a4/plugins/nf-wave/src/main/io/seqera/wave/plugin/resolver/WaveContainerResolver.groovy#L41-L41  

It's a great idea, but if I understand correctly the mechanism you propose, it prioritizes between implementations of a specific interface doesn't it? For our case, we need to override one implementer of `FusionEnv` (i.e. `TowerFusionEnv`) but maintain any other implementers such as `AwsFusionEnv`. Can we do that automatically?

pditommaso:
You are right, it may need some customisation in the plugins handling. Ignore it for now.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
jordeu: We'll extract these two values from the URL. 

Use the `FusionConfig.version()` for the `version`.

Add a method like this at `FusionConfig`
```
String product() {
        return "fusion"
}
```
and use it for the `product` variable.

In the future we can parse the product from the URL, like we are doing for the version.

jordeu (approved):

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    /**
+     * Create a JSON string representing a {@link LicenseTokenRequest} object
+     *
+     * @param product The product SKU
+     * @param version The version
+     * @return The resulting JSON string
+     */
+    private static String makeLicenseTokenRequest(String product, String version) {
```
stefanoboriero, line 238: Does this need to be static? It's private and it seems to be called only by non static method `getLicenseToken`

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
```diff
@@ -0,0 +1,257 @@
...
+        )
+    }
+
+    /**
+     * Parse a JSON string into a {@link LicenseTokenResponse} object
+     *
+     * @param stringHttpResponse The HttpResponse containing the JSON string
+     * @return The resulting LicenseTokenResponse object
+     */
+    private static LicenseTokenResponse parseLicenseTokenResponse(HttpResponse<String> resp) {
```
stefanoboriero, line 254: Does this need to be static? It's private and it seems to be called only by non static method `getLicenseToken`

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
```diff
@@ -0,0 +1,257 @@
...
+     *
+     * @param req The HttpRequest to send
+     * @return The HttpResponse received
+     */
+    private <T> HttpResponse<String> safeHttpSend(HttpRequest req, RetryPolicy<T> policy) {
+        return Failsafe.with(policy).get(
+            () -> {
+                log.debug "Request: method:=${req.method()}; uri:=${req.uri()}; request:=${req}"
+                final resp = httpClient.send(req, HttpResponse.BodyHandlers.ofString())
+                log.debug "Response: statusCode:=${resp.statusCode()}; body:=${resp.body()}"
```
stefanoboriero, line 225: This response is going to contain an authentication token right? Do we want to log it?

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
```diff
@@ -0,0 +1,257 @@
...
+@Slf4j
+@Extension
+@CompileStatic
+class TowerFusionEnv implements FusionEnv {
+
+    // The path relative to the Platform endpoint where license-scoped JWT tokens are obtained
+    private static final String LICENSE_TOKEN_PATH = 'license/token/'
+
+    // Server errors that should trigger a retry
+    private static final List<Integer> SERVER_ERRORS = [429, 500, 502, 503, 504]
```
stefanoboriero, line 47: I'm not sure how I feel about retrying 503 and 504 errors, if Seqera Platform is under heavy load wouldn't this make it worse?

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
```diff
@@ -0,0 +1,257 @@
...
+            log.warn("Error retrieving Fusion license information: ${e.message}")
+            return Map.of()
+        }
+    }
+
+    /**
+     * Send a request to Platform to obtain a license-scoped JWT for Fusion. The request is authenticated using the
+     * Platform access token provided in the configuration of the current session.
+     *
+     * @throws AbortOperationException if a Platform access token cannot be found
```
stefanoboriero, line 116: It also throws other exceptions if the token is not valid or malformed

stefanoboriero (approved):
Nice, left a few comments but nothing blocking

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
alberto-miranda: I made the function `static` because its execution is not linked to the object's lifetime and doesn't need any state from it. Why do you think it's necessary to make it a member function?

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
alberto-miranda: Good question. I'm not sure about what the preferred policy is in Nextflow logs (cc @pditommaso)

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
alberto-miranda: I took this from [Wave's client](https://github.com/nextflow-io/nextflow/blob/ee252173bf823d4bbef5522bd726ccd4df595a7d/plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy#L788-L795) to follow the existing approach. I don't have a specific preference :man_shrugging: (cc @pditommaso)

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
stefanoboriero: It's fine, I was wondering about it since it's something I don't usually do unless I need to access the static method from outside the class or from other static methods in the class. There's no "necessary" here in either direction, happy to resolve the thread

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
stefanoboriero: Thanks consider this resolved, I apparently do not have the ability to resolve myself :(

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
alberto-miranda: Got it. Thanks for the explanation!

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
pditommaso: That's why it's used an exponential backoff. Actually think it should be added 408 (connection timeout) to that list

stefanoboriero: Makes sense, consider this resolved :+1:

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
alberto-miranda: I actually didn't read this recommendation until now, but ended up doing something similar with 94f6075fa27cdba2db7d75eee5740cd9e6318d8b :sweat_smile:

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
alberto-miranda: Error 408 added with 6ce5be3dc2a152d2642aba4af779bf3254d8b2c4

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
alberto-miranda: Addressed with 1fed85918c016495c598fc805f4bbce4d1d3fce4 and refactored into a9849f55891aebac3bb04a4a75b98e18ad39e5a5

alberto-miranda:
As discussed this morning, I have added caching of license token responses with a9849f55891aebac3bb04a4a75b98e18ad39e5a5 and re-scoped the license-related error messages as `debug` instead of `warning`...

I'm not at all convinced that the caching is working as it should, so I would appreciate more experienced eyes looking into it.

alberto-miranda:
It looks like I broke some tests in the process. Fixing.

pditommaso:
Does this depends on Fusion 2.5?

jordeu:
> Does this depends on Fusion 2.5?

No, it's also supported on Fusion 2.4

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
pditommaso: Is the token temporary, right? then it should not be a problem

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionEnv.groovy
```diff
@@ -0,0 +1,299 @@
...
+            def resp = tokenCache.get(
+                key,
+                () -> sendRequest(req)
+            ) as LicenseTokenResponse
+
+            if( resp.expirationDate.before(new Date()) ) {
+                log.debug "Cached token already expired; refreshing"
+                resp = sendRequest(req)
+                tokenCache.put(key, resp)
+            }
```
pditommaso, line 146: I wonder why this is needed? when the `resp` should be automatically be evicted by the cache, and automatically invoke the `sendRequest` method

alberto-miranda: It's needed because the JWT token TTL in the cache (1 hour) and its expiration date (e.g. 1 day?) are not sync'ed, so it could happen that we get a token from the cache which was valid at the time of insertion but is now expired. As an alternative implementation we could have defined the token TTL in terms of its expiration date. I decided against it because:
1. It required parsing and validating the received token which seemed overkill at the time since it may bring with it extra dependencies.
2. The `Cache` implementation we are using only seems to allow setting a global TTL for all entries, but there's really no hard guarantee anywhere that says that all tokens should have the same expiration period.
If `1` is not not a problem and we use a different cache for `2`, we could certainly improve on this.

pditommaso: Understood, made a little change adding this comment as reminder 👉 61bb8b82. 

Then, two other points: 
1. should not the expiration check have a few seconds (mins) gap considering the latency to launch the Fusion job?. It could happen this delta `resp.expirationDate.before(new Date())` is shorter then then time required Fusion got executed or am I wrong? 
2. The `expirationDate` needs to be an `Instant`. `Date` object in Java are localised, with all the mess about TZ, etc`. Instead `Instant is absolute UTC time.

alberto-miranda: > Understood, made a little change adding this comment as reminder 👉 61bb8b82. 
> 

Great, thanks!

> 
> Then, two other points: 
> 
> 1. should not the expiration check have a few seconds (mins) gap considering the latency to launch the Fusion job?. It could happen this delta `resp.expirationDate.before(new Date())` is shorter then then time required Fusion got executed or am I wrong? 

IIRC `resp.expirationDate` is an absolute time point with time zone information so just verifying that it's valid according to the current "date" should be enough.

> 2. The `expirationDate` needs to be an `Instant`. `Date` object in Java are localised, with all the mess about TZ, etc`. Instead `Instant is absolute UTC time.  

Probably, I've not used Java in far too long and I was really confused with Dates, Durations, Instants, etc...

pditommaso: > Probably, I've not used Java in far too long and I was really confused with Dates, Durations, Instants

Yeah, it's definitively a mess. We need to change it both here and backend

modules/nextflow/src/main/groovy/nextflow/fusion/FusionEnvProvider.groovy
```diff
@@ -46,4 +49,32 @@ class FusionEnvProvider {
...
+        }
+        return result
+    }
+
+    protected Map<String,String> getFusionToken(String scheme, FusionConfig config) {
+        final providers = Plugins.getPriorityExtensions(FusionToken)
+        log.debug "Fusion token extensions=$providers"
+        final result = new HashMap<String,String>()
+        try {
+            final env = providers.first.getEnvironment(scheme,config)
```
bentsherman, line 70: Don't know how this got through, but this is causing CI errors. Should be:
```groovy
            final env = providers.first().getEnvironment(scheme,config)
```

pditommaso: Uhh tricky. `first()` is a Groovy extension method, `getFirst()` is Java 21 default method

pditommaso: Done a0ca6e4f

PR #5620, opened 2024-12-18
Add term 'workflow introspection' to docs

ewels:
I keep searching for this specific term and it's been removed. Apparently I wasn't the only one.

Adding it back so that hitting this term into the search box brings up the right page.

'cc @theJasonFan


theJasonFan:
Ship it 🚀

ewels:
Looks like Netlify bumped their default Python version to 3.13 and it broke the build. Pinned to 3.12 and the build is working now, CI failures are stale and can be ignored.

Issue #5626, opened 2024-12-27
docs: more examples of dynamic resource calculations 


nick-youngblut:
## New feature

https://www.nextflow.io/docs/latest/process.html#dynamic-task-resources provides just one example of dynamic resource calculations:

```
process foo {
    memory { 2.GB * task.attempt }
    time { 1.hour * task.attempt }

    errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
    maxRetries 3

    script:
    """
    your_command --here
    """
}
```

It does not show (or link to) more complex examples, which are non-trivial to implement, such as:

```
memory { 16.GB + input_file.size() / 10.GB }
```

The above calculation actually does not work: `ERROR ~ Unknown method invocation `div` on Long type`

So, it would helpful to show examples of:

1. using input file sizes in the calculations
2. perform 

## Usage scenario 

More complex dynamic resource allocation examples than just `memory { 2.GB * task.attempt }`.

## Suggest implementation 

More examples in the docs. I would create just a PR, but as you can see by my failed example above, I do not know how to actually perform some of the more complex resource allocation calculations.

### General comment on docs

As a more general reason for including such docs: users and developers obtain their information from ChatGPT, Claude, etc. So, the more publicly available Nextflow docs, the more helpful these LLM services will be in regards to Nextflow. 

For this reason, it is best to include more details than less.

If you are worried that the docs will be too cluttered for users to find the relevant docs, that issue is becoming less of a concern as users rely more on web search via ChatGPT, Claude (possible with MCP), Perplexity, etc. to find the specific relevant docs.

PR #5628, opened 2024-12-27
Added another example for dynamic-task-resources

nick-youngblut:
This pull request includes an update to the `docs/process.md` file to add a more detailed example for dynamic task resource allocation in Nextflow processes.

Documentation enhancements:

* [`docs/process.md`](diffhunk://#diff-aed6df2932484b499862e14d7e11cd39fd0bd2cb28a80b94c79446c984b9e9abR1306-R1323): Added a detailed example for dynamically allocating memory based on input file size and the number of task attempts. This example demonstrates how to set a minimum memory of 16 GB and increase it proportionally to the size of the `input_file` and the number of attempts (`task.attempt`). It also includes a note on converting `input_file.size()` to bytes to avoid errors.

Addressing issue https://github.com/nextflow-io/nextflow/issues/5626

### DIFF
--- a/docs/process.md
+++ b/docs/process.md
@@ -1303,6 +1303,24 @@ If the task execution fail reporting an exit status in the range between 137 and
 
 The directive {ref}`process-maxretries` set the maximum number of time the same task can be re-executed.
 
+A more elaborate example:
+
+```nextflow
+process foo {
+  memory { 
+    def baseMem = 16.GB
+    def additionalMem = input_file.size() / 1e9
+    baseMem + (additionalMem * task.attempt).GB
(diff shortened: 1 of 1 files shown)


bentsherman:
See also: #4545

nick-youngblut:
So, just close this one, given https://github.com/nextflow-io/nextflow/pull/4545? ... or maybe combine the examples?

bentsherman:
Well, the other one got held up because I wanted to make sure the example was actually realistic. So I would be keen to combine these into a single example, but if you think your example is representative then we should lean towards this one

PR #5631, opened 2025-01-02
Only close the cache db if it is not null.

robsyme:
This fixes a bug where the cache db is not closed if an exception occurs when opening it. This unhandled exception will cause Nextflow to fail and prevent the Global.cleanUp() from being called.


modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -1223,10 +1223,10 @@ class Session implements ISession {
             log.trace "Clean workdir complete"
         }
         catch( Exception e ) {
-            log.warn("Failed to cleanup work dir: ${workDir.toUriString()}")
+            log.warn("Failed to cleanup work dir: ${workDir.toUriString()}. Exception: ${e.message}")
         }
         finally {
-            db.close()
+            db?.close()
```
pditommaso, line 1229: This can be better managed using (new) try-with-resource statement ie. 

```
try ( CacheDB db = CacheFactory.create(uniqueId, runName).openForRead() ) {
  ..
}
catch( Exception e ) {
  .. 
}

File: modules/nextflow/src/main/groovy/nextflow/Session.groovy [hunk 2]
```diff
@@ -1223,10 +1223,10 @@ class Session implements ISession {
             log.trace "Clean workdir complete"
         }
         catch( Exception e ) {
-            log.warn("Failed to cleanup work dir: ${workDir.toUriString()}")
+            log.warn("Failed to cleanup work dir: ${workDir.toUriString()}. Exception: ${e.message}")
```
pditommaso, line 1226:
```suggestion
            log.warn("Failed to cleanup work dir: ${workDir.toUriString()}", e)
```

It could help tracing the exception as well

modules/nextflow/src/main/groovy/nextflow/Session.groovy
robsyme: Ah, I didn't know about try-with-resource! Lovely feature, no implemented in the latest commit.

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -1209,10 +1209,8 @@ class Session implements ISession {
         if( aborted || cancelled || error )
             return
 
-        CacheDB db = null
-        try {
+        try (CacheDB db = CacheFactory.create(uniqueId, runName).openForRead()) {
             log.trace "Cleaning-up workdir"
```
pditommaso, line 1213: It would make sense to move the `log` out of the `try`, then

robsyme: Great point, done

bentsherman:
There is a failing docs snippet, likely due to an environment mismatch. I will update that test in a separate PR.

Issue #5084, opened 2024-06-20
Plugin did not take subcommand options


sfc-gh-hyu:
I am trying to develop a plugin against nextflow, however, it seems that subcommand options is considered as a parent command options:

Here is a simple plugin:
```
class MyPlugin extends BasePlugin implements PluginAbstractExec{

    MyPlugin(PluginWrapper wrapper) {
        super(wrapper)
    }

    @Override
    List<String> getCommands() {
        [ 'run' ]
    }

    @Override
    int exec(String cmd, List<String> args) {
        System.out.println("command array: " + Arrays.toString(args))
        if( cmd == 'run' ) {
            return 0
        }
        else {
            System.err.println "Invalid command: ${cmd}"
            return 1
        }
    }
}
```
When I run the plugin
```
$ ./launch.sh plugin nf-foo:run
command array: [[]]
$ ./launch.sh plugin nf-foo:run --foo bar
Unknown option: --foo=bar -- Check the available commands and options and syntax with 'help'
```
Did I miss anything?

sfc-gh-hyu:
Ping on this. Is there any plan to fix this?

bentsherman:
@sfc-gh-hyu sorry for the delay, we will look into it when we have some time

tom-seqera:
@sfc-gh-hyu my understanding is that anything prefixed by `-` or `--` will currently be treated by nextflow as an option for nextflow itself. You can pass arguments to the plugin command without prefixes, like so:

```
$ ./launch.sh plugin nf-foo:run bar
command array: [[bar]]

$ ./launch.sh plugin nf-foo:run foo bar
command array: [[foo, bar]]
```

PR #5643, opened 2025-01-06
allows parameters when run a plugin subcommand

jagedn:
Current implementation doesn't allow to pass tagged parameters when running a cmd plugin, only positional parameters are allowed (see https://github.com/nextflow-io/nextflow/issues/5084 )

This PR is a proposal to allow "--" as DynamicParameter so plugins can parse them.

For example, once the plugin receives the `String[]args`,  it can use JCommander and parse the arguments similar as core is doing right now so running 

`nextflow plugin my-plugin:my-command 1234 --f memory,module,name`

can be the same as running 

`nextflow plugin my-plugin:my-command --f memory,module,name 1234`


pditommaso:
Thanks @jagedn! @bentsherman @tom-seqera thoughts?

bentsherman (approved):

bentsherman:
Should be fixed by the upstream changes

PR #5644, opened 2025-01-06
Update Google Batch docs to clarify spot retry behaviour

adamrtalbot:
I found the Google Batch spot documentation a bit hard to follow, I've updated the language here to see if it makes more sense.

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -829,7 +829,7 @@ The following settings are available for Google Cloud Batch:
 `google.batch.autoRetryExitCodes`
 : :::{versionadded} 24.07.0-edge
   :::
-: Defines the list of exit codes that will be automatically retried by Google Batch when `google.batch.maxSpotAttempts` is greater than 0 (default `[50001]`). Refer to the [Google Batch documentation](https://cloud.google.com/batch/docs/troubleshooting#reserved-exit-codes) for the list of retryable exit codes.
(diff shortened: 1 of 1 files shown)


bentsherman (approved):

christopher-hakkaart (approved):

bentsherman (changes requested):
Actually @pditommaso it looks like we aren't even using the auto-retry exit codes anymore. I think they were used here:

https://github.com/nextflow-io/nextflow/blob/cc0163ac0f994f43d6195c4819880a1ac1ad149b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy#L542-L545

But now there is just a regex matching all 500xx exit codes

pditommaso (approved):

bentsherman:
@pditommaso see my previous comment on this thread. Something has changed and these auto retry codes don't appear to be used anymore in the code

pditommaso:
Why not? what's the evidence for that?

bentsherman:
The `autoRetryExitCodes` is not used anywhere in the google batch task handler. In fact I can't find a commit where it was used, even the [original commit](https://github.com/nextflow-io/nextflow/commit/e562ce061c0133ab2d16d8601127d86e1aba8d55) did not use it.

pditommaso:
New bug!

PR #5645, opened 2025-01-06
Document differences between `join` operator and SQL join

### LINKED ISSUE #5635: docs: point out that join does not fully act like sql join
## New feature

```
left  = Channel.of( ['X', 1], ['X', 2], ['Z', 3], ['P', 7] )
right = Channel.of( ['X'] )
left.join(right).view()
```

The output:

```
[X, 1]
```

...instead of:

```
[X, 1], [X, 2]
```

...as would be the case for an SQL inner join. 

Only first match is kept instead of all matches. 

As far as I can tell, https://www.nextflow.io/docs/latest/reference/operator.html#join does not point out this aspect of the `join` operator; instead, the docs currently state:

> It is equivalent to an inner join in SQL, or an outer join when remainder is true.

## Suggest implementation 

Update the docs to point out that join is not "equivalent to an inner join in SQL", or modify the `join` operator to be equivalent to an SQL join.

bentsherman:
Close #5635

### DIFF
--- a/docs/reference/operator.md
+++ b/docs/reference/operator.md
@@ -743,7 +743,7 @@ By default, unmatched items are discarded. The `remainder` option can be used to
 ```
 
 :::{note}
-The `join` operator is similar to an *inner join* in SQL, or an *outer join* when `remainder` is `true`. The only difference is that `join` does not support duplicate keys, whereas a SQL join produces the cross-product of duplicate keys. The `combine` operator with the `by` option is equivalent to a SQL join.
+The `join` operator is similar to an SQL *inner join*, or an SQL *outer join* when `remainder` is `true`. The only difference is that `join` does not support duplicate keys, whereas an SQL join produces the cross-product of duplicate keys. The `combine` operator with the `by` option is equivalent to an SQL join.
 :::
 
 Available options:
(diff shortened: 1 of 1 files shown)


docs/reference/operator.md
```diff
@@ -742,6 +742,10 @@ By default, unmatched items are discarded. The `remainder` option can be used to
 :language: console
 ```
 
+:::{note}
+The `join` operator is similar to an *inner join* in SQL, or an *outer join* when `remainder` is `true`. The only difference is that `join` does not support duplicate keys, whereas a SQL join produces the cross-product of duplicate keys. The `combine` operator with the `by` option is equivalent to a SQL join.
```
christopher-hakkaart, line 746:
```suggestion
The `join` operator is similar to an SQL *inner join*, or an SQL *outer join* when `remainder` is `true`. The only difference is that `join` does not support duplicate keys, whereas an SQL join produces the cross-product of duplicate keys. The `combine` operator with the `by` option is equivalent to an SQL join.
```

christopher-hakkaart (commented):
"Join" is used a lot in the note. It would be good to be specific about Nextflow vs SQL joins. Otherwise looks good.

christopher-hakkaart (approved):

PR #5646, opened 2025-01-06
Fix flaky docs snippet

bentsherman:
The docs snippet for eval output uses `bash --version` since that is the intended use of eval output. However, it has proven hard to test since the exact bash version text can change between e.g. my local environment and CI. This PR makes the eval output example consistent with the stdout example.

See also: https://github.com/nextflow-io/nextflow/pull/5631#issuecomment-2573484502


christopher-hakkaart (approved):
LGTM

PR #5650, opened 2025-01-07
Handle attribute caching with Lustre filesystem

### LINKED ISSUE #5630: Nextflow loses track of SLURM jobs when the number of jobs is large
## Bug report 

The client experiencing an issue with Nextflow when running a pipeline that involves large-scale computations. Specifically, completed jobs are not being detected, which prevents new jobs from being launched, leaving the pipeline unable to finish. This seems related to a previously reported issue ([#2695](https://github.com/nextflow-io/nextflow/issues/2695), but the proposed solutions there have not resolved the problem in my case.

The problem occurs when running pipelines on a Lustre filesystem, where attribute caching might be causing issues. Attempts to resolve the issue by using alternative paths (e.g., non-Lustre storage) have not worked consistently, as the underlying filesystem also uses Lustre.

Nextflow should detect completed jobs and proceed to launch new ones, allowing the pipeline to finish.

Steps to Reproduce:

Run the pipeline [nf-coloc](https://github.com/NovoNordisk/nf-coloc) on a Lustre filesystem using the following parameters:
600 jobs, each taking ~1 hour.
Work directory set to /delete90d/<user>/nextflow/work/.
Monitor the job queue. Despite jobs completing successfully (verified via sacct), new jobs are not launched.
Observe logs where Nextflow repeatedly attempts to cancel jobs that have already ended.

### Program output 
Relevant excerpt from the log:

```
Dec-18 10:01:55.421 [main] WARN  n.processor.TaskPollingMonitor - Killing running tasks (380)
Dec-18 10:01:55.472 [main] TRACE n.executor.AbstractGridExecutor - Kill command: [scancel, 5773200, ...]
```

```
JobID           JobName  Partition    Account  AllocCPUS      State ExitCode 
------------ ---------- ---------- ---------- ---------- ---------- -------- 
5773200      nf-COLOC_+    compute 005.1 com+          1  COMPLETED      0:0
5773200.bat+      batch            005.1 com+          1  COMPLETED      0:0
5773200.ext+     extern            005.1 com+          1  COMPLETED      0:0
```

### Environment 

* Nextflow version: 23.10.1
* Cluster Environment: SLURM
* Filesystem: Lustre

### Additional context

The client also noticed that things are slightly better with the latest edge release and the fraction of jobs failing is smaller, but still not 0. 

### Troubleshooting Attempts:

- Following suggestions from issue #2695 (e.g., monitoring the directory with lsof) did not resolve the issue.
- Switching NXF_WORK to a different path temporarily resolved the issue, but subsequent runs failed as the new location also relied on Lustre.
- Potential code modification (suggested by a colleague) to address attribute caching was deemed unfeasible without upstream support.
-  Sci-dev has taken a look at this and worked with client and suggested it is a bug/feature but they were not sure provided the previous nextflow ticket

bentsherman:
Close #5630 

Lustre performs attribute caching similar to NFS, but the `stat` command on a Lustre filesystem returns `lustre` instead of `nfs`. This PR adds a check for Lustre so that the grid executors can handle attribute caching for both NFS and Lustre.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/GridTaskHandler.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/GridTaskHandler.groovy
@@ -319,7 +319,7 @@ class GridTaskHandler extends TaskHandler implements FusionAwareTask {
     protected Integer readExitStatus() {
 
         String workDirList = null
-        if( exitTimestampMillis1 && FileHelper.workDirIsNFS ) {
+        if( exitTimestampMillis1 && FileHelper.workDirIsSharedFS ) {
             /*
              * When the file is in a NFS folder in order to avoid false negative
              * list the content of the parent path to force refresh of NFS metadata
--- a/modules/nf-commons/src/main/nextflow/file/FileHelper.groovy
+++ b/modules/nf-commons/src/main/nextflow/file/FileHelper.groovy
@@ -468,14 +468,14 @@ class FileHelper {
      * @return The {@code true} when the path is a NFS mount {@code false} otherwise
      */
     @Memoized
-    static boolean isPathNFS(Path path) {
+    static boolean isPathSharedFS(Path path) {
         assert path
         if( path.getFileSystem() != FileSystems.getDefault() )
             return false
 
         final type = getPathFsType(path)
-        def result = type == 'nfs' || type == 'lustre'
-        log.debug "NFS path ($result): $path"
+        final result = type == 'nfs' || type == 'lustre'
+        log.debug "FS path type ($result): $path"
         return result
     }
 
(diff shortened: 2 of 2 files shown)


pditommaso:
Updated the method names and messages for consistency

burcarjo:
The same behavior occurs with distributed systems such as: CEPH which causes the executor process to hang waiting for updates to the .exitcode file.
Is it possible to make a similar patch for ceph systems?
My workdir is in:

stat -f -c %T .
**ceph**

bentsherman:
@burcarjo would you be willing to test the change locally? Follow the instructions [here](https://nextflow.io/docs/latest/developer/index.html#building-from-source) to build locally with your suggested change, and if it helps, we can merge it.

burcarjo:
I added the CEPH condition as shared file system in the function:  FileHelper->isPathSharedFS and then build it:
final result = type == 'nfs' || type == 'lustre' || type == 'ceph'

However I get the same results ( master executor remains hanged after some time).

I can note that the thread of execution never reaches the function safeExists , although the system is considered sharedFS

PR #4522, opened 2023-11-18
Preserve attributes of published files

jjfarrell:
When PublishDir copies pairs of files such as (cram and cram.crai) and (vcf.gz and vcf.gz.tbi). the atributes of the files are not copied which includes the file creation time.   Downstream software expects the index files (crai and tbi)  to have a later time than the corresponding cram and vcf.gz files). To preserve the creation time of the files, the PublishDir code was modified to copy the files attributes which preserves the original file creation time. #1489 described the issue.

PR #5660, opened 2025-01-09
Preserve attributes of published files (take 2)

bentsherman:
Continuation of #4522 

Allow COPY_ATTRIBUTES on S3 copy as a no-op, so that this attribute can be set generically when publishing files.

For some reason this issue did not surface in the previous PR but only once the PR was merged


pditommaso:
Tagging @jordeu for visibility? Does Fusion honour the copy of file timestamps?

pditommaso:
If the copy attributes is only honoured by local and shared file system, we should consider making this an opt-in feature, because it's not a portable behaviour.

In alternative, instead of copying the file stamps, it could be tried to copy the files ordering them by timestamp. Even if the timestamp would be different it may guarantee they are consistent with expectation of the tool reported in the issue (tho is should be verified the the timestamp is set when stating the upload or when it completes).

bentsherman:
I don't think re-ordering the publish operations will matter because they are published concurrently

pditommaso:
Yeah, ordering was more a random thought

pditommaso:
To recap here, since this rule cannot be generalised `StandardCopyOption.COPY_ATTRIBUTES` should be added only when enable a specific config (output?) option

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
```diff
@@ -509,13 +516,14 @@ class PublishDir {
...
+            FileHelper.movePath(source, destination, options as CopyOption[])
         }
         else if( mode == Mode.COPY ) {
-            FileHelper.copyPath(source, destination)
+            FileHelper.copyPath(source, destination, options as CopyOption[])
         }
         else if( mode == Mode.COPY_NO_FOLLOW ) {
-            FileHelper.copyPath(source, destination, LinkOption.NOFOLLOW_LINKS)
+            options.add(LinkOption.NOFOLLOW_LINKS)
+            FileHelper.copyPath(source, destination, options as CopyOption[])
```
bentsherman, line 526: Could make sense to refactor this as another config option called `workflow.output.followLinks` (default true) instead of a separate publish mode

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3FileSystemProvider.java
```diff
@@ -535,7 +535,7 @@ public void copy(Path source, Path target, CopyOption... options)
 		 * "copying directories is not yet supported: %s", target); // TODO
 		 */
 		ImmutableSet<CopyOption> actualOptions = ImmutableSet.copyOf(options);
-		verifySupportedOptions(EnumSet.of(StandardCopyOption.REPLACE_EXISTING),
+		verifySupportedOptions(EnumSet.of(StandardCopyOption.COPY_ATTRIBUTES, StandardCopyOption.REPLACE_EXISTING),
```
bentsherman, line 538: This change doesn't actually enable copy attributes for S3, just prevents a failure. We could remove it, I would just add a warning to the docs that copy attributes only works for loca/shared filesystem

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
```diff
@@ -498,6 +500,11 @@ class PublishDir {
     protected void processFileImpl( Path source, Path destination ) {
         log.trace "publishing file: $source -[$mode]-> $destination"
 
+        final options = new ArrayList<>(2)
+        final copyAttributes = session.config.navigate('workflow.output.copyAttributes') as Boolean
```
pditommaso, line 504: I fear this `as Boolean` is a false friend and returns `true` whatever this string is. Better to isolate this into its own method returning `StandardCopyOption[]` and test it

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
pditommaso: Maybe, mostly interested to close this tho

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3FileSystemProvider.java
pditommaso: Agree on remove it and add a warning that's only honoured by local/shared file systems

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
bentsherman: I have refactored to more closely match how boolean config options are typically loaded, for example [here](https://github.com/nextflow-io/nextflow/blob/49dca8bd7d2d432a229d857a77cead527da1644d/plugins/nf-google/src/main/nextflow/cloud/google/batch/client/BatchConfig.groovy#L81-L84)

In reality, many boolean config options across the codebase are loaded using `as Boolean`. I don't think we generally check for string values. In the future, the strict config loader and language server will prevent things like passing a string for a boolean anyway via type checking

pditommaso:
@bentsherman Am I wrong or there has been a regression after last changes?

bentsherman:
The last two commits with CI failures report errors with the bitbucket integration:

```
BitbucketRepositoryProviderTest > testReadContent FAILED
    org.spockframework.runtime.SpockTimeoutError at BitbucketRepositoryProviderTest.groovy:58

BitbucketRepositoryProviderTest > should return content URL FAILED
    org.spockframework.runtime.SpockTimeoutError at BitbucketRepositoryProviderTest.groovy:108

BitbucketRepositoryProviderTest > should read content on main and on separate branch FAILED
    org.spockframework.runtime.SpockTimeoutError at BitbucketRepositoryProviderTest.groovy:137
```

I don't see how it could be related to this PR, seems like something wrong with bitbucket creds or some networking issue

bentsherman:
Tests are passing now

pditommaso:
Nice

PR #5661, opened 2025-01-09
Fix CLI params normalization

### LINKED ISSUE #5655: CLI parameter case normalization/translation (kebab-case to camelCase) doesn't work at all since recent update

According to the documentation (and functionality in previous versions), nextflow should translate `--kebab-case` parameters into camelCase. There were some problems with this that I reported previously, which are stated to have been fixed in #4702, but now I'm not able to get it to work at all. In the version I'm using (24.11.0-edge build 5929), nothing at all happens when I pass `--kebab-case` parameters. 

Here is a very simple workflow:

```nextflow
params.thingOne = "nothing"
workflow {
  println(params)
  println(params.thingOne)
}
```

When I do this:

```console
nextflow run test.nf --thing-one newval
```

I would expect to see this output (according to #4702):

```
[thingOne:newval]
newval
```

But instead I see this:

```
[thing-one:newval, thingOne:nothing]
nothing
```

This actually appears to be a step *backward* from where we were before, because case translation worked then (despite other issues I may have had). 

### Program output 

see above

### Environment 

* Nextflow version: 24.11.0-edge build 5929
* Java version: openjdk 17.0.5 2022-10-18
* Operating system: Ubuntu 22.04.1 LTS  (also tried on mac)

bentsherman:
Close #5655

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
@@ -685,7 +685,7 @@ class CmdRun extends CmdBase implements HubOptions {
             addParam((Map)nested, key.substring(p+1), value, path, fullKey)
         }
         else {
-            params.put(key.replaceAll(DOT_ESCAPED,'.'), parseParamValue(value))
+            addParam0(params, key.replaceAll(DOT_ESCAPED,'.'), parseParamValue(value))
         }
     }
 
--- a/modules/nextflow/src/test/groovy/nextflow/cli/CmdRunTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/cli/CmdRunTest.groovy
@@ -99,16 +99,16 @@ class CmdRunTest extends Specification {
 
         when:
         def params = [:]
-        CmdRun.addParam0(params, 'alphaBeta', 1)
-        CmdRun.addParam0(params, 'alpha-beta', 10)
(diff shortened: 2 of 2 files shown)


pditommaso (changes requested):
There should be a test that validates the solution

pditommaso:
Not following much, the issue says 

```
I would expect to see this output (according to https://github.com/nextflow-io/nextflow/pull/4702):

[thingOne:newval]
newval
But instead I see this:

[thing-one:newval, thingOne:nothing]
nothing
```

I would expect, the test to validate the correct behaviour, instead only add quotes to some numeric values

bentsherman:
Check again, it also changed `addParam0` to `addParam` in order to test the full behavior

pditommaso:
Continue to not understand how the test validates the fix for the defect reported.

bentsherman:
Discussed internally. Here's the red-green analysis to show how the unit test captures the original problem.

Without the changes to CmdRun, the test fails as follows:
```
Condition not satisfied:

params['alphaBeta'] == 10
|     |             |
|     1             false
[alphaBeta:1, alpha-beta:10]

	at nextflow.cli.CmdRunTest.should convert cli params from kebab case to camel case(CmdRunTest.groovy:105)
```

Note that `alphaBeta` is set first, then `alpha-beta` which should be converted to `alphaBeta` and override the previous setting.

With the changes to CmdRun, the tests pass as shown in CI.

PR #5663, opened 2025-01-13
Add copy button selector

christopher-hakkaart:
Add a selector for a commands/snippets copy button. Turned off by default but can be added selectively:

```
```{code-block} console
:class: copyable
...
```


pditommaso (approved):

bentsherman:
@christopher-hakkaart do you have an idea of which snippets you would like to make copyable?

christopher-hakkaart:
Good point. I've added copy buttons to the install page as a POC and made a few minor language edits to the text.

PR #5664, opened 2025-01-13
Update your first script

christopher-hakkaart:
Update your first script page to a guide focusing on running a script, modifying and resuming a script, and customizing a parameter.

### DIFF
--- a/docs/your-first-script.md
+++ b/docs/your-first-script.md
@@ -87,7 +87,7 @@ This script defines two processes:
 
 The `splitString` output is emitted as a single element. The `flatten` operator splits this combined element so that each file is treated as a sole element.
 
-The outputs from both processes are published in the `results` directory. Each publish target is saved into a subdirectory of the `results` directory. By default, the target name is used as the directory name.
+The outputs from both processes are published in the `results` directory. Each publish target is saved into a subdirectory of the `results` directory. By default, the target names, that is, `lower` and `upper`, are used as the subdirectory names.
 
 To run your pipeline:
 
--- a/docs/your-first-script.md
+++ b/docs/your-first-script.md
@@ -2,35 +2,30 @@
 
 # Your first script
 
-Once you have Nextflow installed, you’re ready to run your first script. This guide details how to run a basic Nextflow pipeline. It includes:
+Once you have Nextflow installed, you’re ready to run your first script. This guide details fundamental skills to run a basic Nextflow pipeline. It includes:
 
 - Running a pipeline
 - Modifying and resuming a pipeline
 - Configuring a pipeline parameter
 
-**Prerequisites**
+<h3>Prerequisites</h3>
 
 You will need the following to get started:
 
-- Nextflow 24.10.0 or later. See {ref}`install-page` for instructions to install or update your version of Nextflow.
-
-:::{note}
-This guide uses the second preview of publish targets, introduced in Nextflow 24.10.0. Publish targets are intended to replace the `publishDir` directive.
-:::
+- Nextflow 22.10.0 or later. See {ref}`install-page` for instructions to install or update your version of Nextflow.
 
 ## Run a pipeline
 
 You will run a basic Nextflow pipeline that splits a string of text into two files and then converts lowercase letters to uppercase letters. You can see the pipeline here:
 
-```groovy
-// Enable output targets
-nextflow.preview.output = true
-
+```{code-block} groovy
+:class: copyable
 // Default parameter input
 params.str = "Hello world!"
 
 // splitString process
 process splitString {
+    publishDir "results/lower"
     
     input:
     val x
@@ -46,13 +41,14 @@ process splitString {
 
 // convertToUpper process
 process convertToUpper {
+    publishDir "results/upper"
     tag "$y"
 
     input:
     path y
 
     output:
-    path 'upper_chunk_*'
+    path 'upper_*'
 
     script:
     """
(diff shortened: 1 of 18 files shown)


docs/your-first-script.md
```diff
@@ -2,115 +2,226 @@
...
+- Configuring a pipeline parameter
+
+**Prerequisites**
+
+You will need the following to get started:
+
+- Nextflow 24.10.0 or later. See {ref}`install-page` for instructions to install or update your version of Nextflow.
+
+:::{note}
+This guide uses the second preview of publish targets, introduced in Nextflow 24.10.0. Publish targets are intended to replace the `publishDir` directive.
```
pditommaso, line 18: Think it should not be used a feature under dev in the getting started guide

pditommaso: Learning a new tool and starting a getting warning is evil. I think we should use document what's the current approach that's `publishDir`

christopher-hakkaart: Fair point - I've removed the targets and added publishDir.

pditommaso (approved):

docs/your-first-script.md
```diff
@@ -2,115 +2,209 @@
...
+
+- Running a pipeline
+- Modifying and resuming a pipeline
+- Configuring a pipeline parameter
+
+<h3>Prerequisites</h3>
+
+You will need the following to get started:
+
+- Nextflow 22.10.0 or later. See {ref}`install-page` for instructions to install or update your version of Nextflow.
```
bentsherman, line 15: Any reason why you used 22.10 here? We typically don't provide ongoing support for versions older than a year.

christopher-hakkaart: DSL1 > DSL2 and the need to enable DSL2 with `nextflow.enable.dsl=2`

I'm not committed to keeping it. I'll remove it.

bentsherman: I wouldn't go any older than 23.10

christopher-hakkaart: Updated 👍

bentsherman: Sorry, I had some further comments I wanted to make but haven't had time. Will try to review today or tomorrow

docs/your-first-script.md
```diff
@@ -2,115 +2,209 @@
...
+
+- Running a pipeline
+- Modifying and resuming a pipeline
+- Configuring a pipeline parameter
+
+<h3>Prerequisites</h3>
+
+You will need the following to get started:
+
+- Nextflow 23.10.0 or later. See {ref}`install-page` for instructions to install or update your version of Nextflow.
```
bentsherman, line 15: It's a bit awkward to mention this prerequisite when you already started by saying "once you have Nextflow installed...". Is there a way we could mention this once?

Maybe we don't need to give the minimum version? At this point the 23.10 is pretty arbitrary, if the only requirement is DSL2, that's been standard for so long that maybe we can assume that if the user goes through the Install page, they will have a sufficient version

bentsherman:
Apologies, wanted to make a bunch of small suggestions and it was easier to just commit them. Feel free to take or leave what you want. I mainly tried to improve the variable names and comments, and some minor language edits.

docs/your-first-script.md
```diff
@@ -2,115 +2,209 @@
...
 
 (getstarted-params)=
 
 ## Pipeline parameters
 
-Pipeline parameters are simply declared by prepending to a variable name the prefix `params`, separated by dot character. Their value can be specified on the command line by prefixing the parameter name with a double dash character, i.e. `--paramName`
+Nextflow looks for configuration settings in multiple locations when launched. Configuration settings are merged and, if conflicts exist, override each other in a set order. 
 
-For the sake of this tutorial, you can try to execute the previous example specifying a different input string parameter, as shown below:
+Parameters are a type of configuration. They are declared by prepending a variable name to the prefix `params`, separated by dot character. Parameters can be specified on the command line by prefixing the parameter name with a double dash character, for example, `--paramName`. Parameters specified on the command line override parameters specified in a main script.
```
bentsherman, line 181: > Parameters are a type of configuration.

I don't think we should try to conflate params and config because there is an important difference. Params should modify *what is executed* while config should modify *how it is executed* (i.e. without changing the result). There may be some small exceptions to this but as a general rule it holds.

Though it is typical to define params in the config currently, we intend to move away from the config being the source-of-truth for params.

Instead, I would introduce params as something like:

> Parameters are used to control the inputs to a pipeline.

The paragraph on configuration should be removed or moved to a dedicated section on configuration, if you think it would be an appropriate topic to introduce here. Happy to save that for another PR too

docs/your-first-script.md
christopher-hakkaart: Fair point about the repetition. I'll revise the text.

I would like to introduce a "Prerequisites" section to all guides and tutorials so that it's clear what is required to complete the guide. It's not as important here as all we expect is Nextflow, but as we add more guides that require more details (e.g., credentials, docker), I'd like to have some consistency.

docs/your-first-script.md
christopher-hakkaart: Fair point. Configuration isn't really used in this script so I've removed it completely.

docs/your-first-script.md
christopher-hakkaart: Update, I've removed the "once you have nextflow installed" text but kept the Prerequisites section. If you think it's too thin I could add another point about a text editor, and then point at the VS code or Dev Env page.

bentsherman: Sounds good to me. Let's merge!

PR #5672, opened 2025-01-15
Add fusion tip about AWS storage config

pditommaso:
This PR adds more details about the setting of AWS storage for Fusion

### DIFF
--- a/docs/fusion.md
+++ b/docs/fusion.md
@@ -101,6 +101,11 @@ Seqera Platform is able to automatically format and configure the NVMe instance 
 the option "Use Fast storage" when creating the Batch compute environment.
 :::
 
+:::{tip}
+In alternative to setting up a NVMe storage in you compute node, you can use an EBS volume type `gp3` selecting
+325 MiB/s (or more as throughput) and 100 GiB (or more as size). 
+:::
+
 #### AWS IAM permissions
 
 The AWS S3 bucket should be configured with the following IAM permissions:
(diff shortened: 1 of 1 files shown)


docs/fusion.md
```diff
@@ -101,6 +101,11 @@ Seqera Platform is able to automatically format and configure the NVMe instance
 the option "Use Fast storage" when creating the Batch compute environment.
 :::
 
+:::{tip}
+In alternative to setting up a NVMe storage in you compute node, you can use an EBS volume type `gp3` selecting
+325 MiB/s (or more as throughput) and 100 GiB (or more as size).
```
christopher-hakkaart, line 106:
```suggestion
As an alternative to configuring NVMe storage on your compute node, you can use an EBS `gp3` volume with a throughput of 325 MiB/s or more and a size of 100 GiB or more.
```
Please check meaning is retained.

docs/fusion.md
```diff
@@ -101,6 +101,11 @@ Seqera Platform is able to automatically format and configure the NVMe instance
 the option "Use Fast storage" when creating the Batch compute environment.
 :::
 
+:::{tip}
```
jordeu, line 104: I think it's important to stress that it's a good alternative but not as good as NVMe disk.

"While not as fast as NVMe storage, an EBS gp3 volume with a throughput of 325 MiB/s or more and a size of 100 GiB or more provides sufficient performance for many workloads on your compute node"

docs/fusion.md
jordeu: Looks good to me

PR #5676, opened 2025-01-17
Configure target java version alongside gradle toolchain

tom-seqera:
Configure the target Java version inside the gradle `java` block next to the toolchain info, rather than on a per-task basis. This propogates the information to all the variants, in particular the test fixtures.

Without this, a project depending on the test fixtures (such as a nextflow plugin) can, in some situations, fail to build with the following error:
```
- Variant 'testFixturesApiElements' declares a library for use during compile-time, packaged as a jar, and its dependencies declared externally:
    - Incompatible because this component declares a component, compatible with Java 21 and the consumer needed a component, compatible with Java 17
```

# Before this change

Using command `gradle :nextflow:outgoingVariants` to show the variants, the 'main' variants correctly declare that they require Java 17:
image

but the 'testFixtures' variants requires Java 21:
image

# After this change

All the Java variants require Java 17:
image
image

All the compiled bytecode still has class file format version 61 (Java 17).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,18 +67,16 @@ allprojects {
     apply plugin: 'jacoco'
 
     java {
+        // these settings apply to all jvm tooling, including groovy
         toolchain {
             languageVersion = JavaLanguageVersion.of(21)
         }
+        sourceCompatibility = 17
+        targetCompatibility = 17
     }
 
     compileJava {
         options.release.set(17)
-    }
-
-    tasks.withType(GroovyCompile) {
-        sourceCompatibility = 17
-        targetCompatibility = 17
     }
 
     idea {


pditommaso:
All green, switch to Ready to review and merge?

build.gradle
```diff
@@ -70,15 +70,8 @@ allprojects {
         toolchain {
             languageVersion = JavaLanguageVersion.of(21)
         }
-    }
-
-    compileJava {
-        options.release.set(17)
-    }
```
pditommaso, line 77: This should stay, otherwise Java version is inherited from toolchain.languageVersion

tom-seqera: True. Should it be applied to all the Java compile tasks?
```groovy
tasks.withType(JavaCompile).configureEach {
    options.release = 17
}
```

pditommaso: Believe so

build.gradle
```diff
@@ -70,15 +70,11 @@ allprojects {
...
-
-    tasks.withType(GroovyCompile) {
-        sourceCompatibility = '17'
-        targetCompatibility = '17'
+    tasks.withType(JavaCompile).configureEach {
+        options.release = 17        // javac --release
     }
```
pditommaso, line 78: How this syntax improves? afaik `withType(JavaCompile)` does not apply to Java code compiled via groovy compiler

pditommaso:
Solved by 6dd92e3a3. Feel free to re-open if it still does not solve your problem.

tom-seqera:
That change doesn't seem to have solved it: the `testFixtures` variant jars are still marked as requireing Java 21.

I did some digging and there's a few things going on here.

The relevant javac compiler options are:
* `--source` - this controls which Java language features can be used
* `--target` - this determines what class file format version the compiled bytecode will have
*  `--release` - this is equivalent to setting `--source` and `--target`, but also controls which Java API classes/methods can be used

## Source/target
in this version of Gradle, the intended way to set `--source` and `--target` is in the `java` block, which should be respected by all JVM-based tasks (including Groovy, Kotlin, etc):
```gradle
java {
    toolchain {
        languageVersion = JavaLanguageVersion.of(21)
    }
    sourceCompatibility = 17
    targetCompatibility = 17
}
```
I was able to verify that this does seem to work - all the .class files (including those generated from both .groovy and .java files) have the correct version (61). Unlike setting this on the compile tasks explicitly, this also correctly sets the required Java version on the variant jars (like test fixtures).

This change would fix the issue described in this PR.

## Release

While not needed for this fix, the `--release` compiler option is a bit more interesting. Gradle currently doesn't have a nice syntax for this. Before the changes in this PR, it is configured like so:
```gradle
compileJava {
    options.release.set(17)
}
```
This is slightly incorrect because it doesn't apply to the tests, so an improvement would be:
```gradle
tasks.withType(JavaCompile).configureEach {
    options.release = 17
}
```

Except of course when compiling with groovyc in 'joint' mode, neither approach has any effect (ie the existing config on master isn't doing anything either). We'd need to specify it on the Groovy compile tasks for it to be applied. Using groovyc directly, it would look like this:
```bash
groovyc -j -J-release=17 ...
```
Unfortunately, as far as I can tell, there doesn't seem to be a way to persuade Gradle to correctly pass this through in joint compilation mode - `options.forkOptions.jvmArgs` would seem logical, but I wasn't able to find a way to make it work.

In any case, since this already seems not to be working on master, it's probably a separate issue to the change needed for test fixtures.

pditommaso:
> That change doesn't seem to have solved it: the testFixtures variant jars are still marked as requireing Java 21.

Have you tried using `testCompileJava` similarly to `compileJava` 

```
  testCompileJava {
        options.release.set(17)
    }
```

pditommaso:
`Even better then! please add a comment to highlight `sourceCompatibility and `targetCompatibility` apply also to the groovy compiler and there's no need for groovy specific setting

pditommaso (approved):

PR #5684, opened 2025-01-20
Use `/usr/bin/env -S bash` instead of `/bin/bash` on the host system

rollf:
On systems without `/bin/bash` (e.g. NixOS), invoking tasks may fail because of some hard-coded references to `/bin/bash`. Those references are either 
- fine (because the task runs in a container where /bin/bash *is* available),
- don't matter (because the shebang is ignored)
- or cause failure (if `/bin/bash` cannot be invoked).

This PR adapts Nextflow for the last scenario. Instead of invoking `/bin/bash`, `/usr/bin/env -S bash` is used where necessary. This approach (or similar) has been discussed previously:

* https://github.com/nextflow-io/nextflow/pull/2341
* https://github.com/nextflow-io/nextflow/issues/1598
* https://github.com/nextflow-io/nextflow/pull/1614
* https://github.com/nextflow-io/nextflow/issues/5420

However, compared the previous discussions/implementations, I specifically do *not* propose to change `/bin/bash` unless really necessary. By "necessary" I mean that `bash` is invoked on a local host system where it may not be available. Namely, when `<task>/.command.run` shall be invoked (always on the host) and when `<task>/.command.sh` is invoked in a non-container context (i.e. also on the host). I have *not* touched any other occurrences of `/bin/bash`, esp. not within the container.

I'm happy  to change the the code as requested. I have not added any comments since I expect (at least) change requests. Please review this with an unbiased mind set. Multiple people have expressed their intention to use `nextflow` on systems without `/bin/bash`.

I have tested solely on NixOS.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -65,6 +65,8 @@ class BashWrapperBuilder {
 
     static final public List<String> BASH
 
+    static final public List<String> ENV_BASH
+
     static private int level
 
     @PackageScope
@@ -85,6 +87,7 @@ class BashWrapperBuilder {
             log.warn "Invalid value for `NXF_DEBUG` variable: $str -- See http://www.nextflow.io/docs/latest/config.html#environment-variables"
         }
         BASH = Collections.unmodifiableList( level > 0 ? ['/bin/bash','-uex'] : ['/bin/bash','-ue'] )
+        ENV_BASH = Collections.unmodifiableList(['/usr/bin/env', '-S', 'bash', BASH[1]])
 
     }
 
@@ -563,12 +566,14 @@ class BashWrapperBuilder {
         statsEnabled || fixOwnership()
     }
 
-    protected String shellPath() {
-        // keep the shell path as "/bin/bash" when a non-custom "shell" attribute is specified
-        // to not introduce unexpected changes due to the fact BASH is defined as "/bin/bash -eu" by default
-        return shell.is(BASH)
-            ? "/bin/bash"
-            : shell.join(' ')
+    protected String traceInterpreter() {
+        // if no specific shell was defined, do not use the default one (i.e. "...bash -ue[x]") neither because
+        // -ue[x] should not be used during tracing
+        if( shell.is(BASH) || shell.is(ENV_BASH) ) {
+            return shell.init().join(' ')
+        }
+        // otherwise, stick to the custom shell
+        return shell.join(' ')
     }
 
     protected String getLaunchCommand(String interpreter, String env) {
@@ -582,7 +587,7 @@ class BashWrapperBuilder {
         final traceWrapper = isTraceRequired()
         if( traceWrapper ) {
             // executes the stub which in turn executes the target command
-            launcher = "${shellPath()} ${fileStr(wrapperFile)} nxf_trace"
+            launcher = "${traceInterpreter()} ${fileStr(wrapperFile)} nxf_trace"
         }
         else {
             launcher = "${interpreter} ${fileStr(scriptFile)}"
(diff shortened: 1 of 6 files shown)


modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
```diff
@@ -85,6 +87,7 @@ class BashWrapperBuilder {
             log.warn "Invalid value for `NXF_DEBUG` variable: $str -- See http://www.nextflow.io/docs/latest/config.html#environment-variables"
         }
         BASH = Collections.unmodifiableList( level > 0 ? ['/bin/bash','-uex'] : ['/bin/bash','-ue'] )
+        ENV_BASH = Collections.unmodifiableList(['/usr/bin/env', '-S', 'bash', BASH[1]])
```
pditommaso, line 90: Why introducing this variable instead of relying on `process.shell`?

rollf: Thanks for looking into this. `process.shell` is `/bin/bash ...` [by default](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy#L128). Does this answer your question?

pditommaso: My point is that after https://github.com/nextflow-io/nextflow/pull/4292 you should be able to use `process.shell = ['/usr/bin/env, '-S', 'bash']` in your config.

rollf: "Yes", to my first or second interpretation? I feel like the are mutually exclusive. :shrug:

pditommaso: Yes, I suggest this PR is not necessary at all because one could set process.shell

rollf: [Disclaimer: I assume that the additional usage of `/usr/bin/env -S bash` on the host system (to invoke `.command.run` as well as for non-container task execution) is acceptable by the Nextflow maintainers and we just need to sort out the technical implementation. Note that the POSIX-compliant existence of `/usr/bin/env` is already [expected](https://github.com/nextflow-io/nextflow/blob/master/nextflow#L1) by Nextflow and please also consider the PR description above.]

`procoess.shell` cannot not be used for various reasons:
* The value of `process.shell` will be used for task execution in both the non-container as well as the container scenario. This means that one would need to set this to a value that works in the host as well as in the container env if one wants to freely disable/enable containerization. And this contradicts the whole point.
* Setting `process.shell =/usr/bin/env -S bash -ue` (because I want `-ue`) in the `nextflow.config` causes problems during tracing (because due to #4292, Nextflow will favor `process.shell`'s actual value [it's not `BASH`] and then use `-ue` which Nextflow specifically doesn't not expect for tracing)
* Even if it worked, that means that everybody without `/bin/bash` needs to adapt every pipeline they want to run. I'm not trying to fix a specific problem I have with a specific pipeline, I'd like to contribute here in order to solve the general problem.

modules/nextflow/src/main/groovy/nextflow/script/ProcessConfig.groovy
```diff
@@ -125,7 +125,6 @@ class ProcessConfig implements Map<String,Object>, Cloneable {
     static final Map<String,Object> DEFAULT_CONFIG = [
             debug: false,
             cacheable: true,
-            shell: BashWrapperBuilder.BASH,
```
rollf, line 128: The changes in this file and the next one make sure that Nextflow does not assume `/bin/bash` execution by default. Rather, it used `/bin/bash` when the configuration requests task containerization, otherwise, `/usr/bin/env -S bash` is used (i.e. when the task is to be executed on the host system itself).

modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
pditommaso: The use of `/usr/bin/env` is not viable because adds another level of indirection   of in the process execution making (even) more complex the composition of the launch command and the passing of shell option. 

I propose to address the problem of providing a custom Bash path in a more structural manner in this PR https://github.com/nextflow-io/nextflow/pull/5696

pditommaso:
Closing in favour of https://github.com/nextflow-io/nextflow/pull/5696

PR #5696, opened 2025-01-22
Make Bash path configurable

pditommaso:
This PR allows the customisation of Bash interpreter used by Nextflow go a user provided path defined via nextflow config file or an environment variable (to be done).

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/NF.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/NF.groovy
@@ -17,6 +17,7 @@
 package nextflow
 
 import groovy.runtime.metaclass.NextflowDelegatingMetaClass
+import groovy.transform.Memoized
 import nextflow.extension.CH
 import nextflow.plugin.extension.PluginExtensionProvider
 import nextflow.script.ExecutionStack
@@ -78,5 +79,10 @@ class NF {
 
     static boolean isTopicChannelEnabled() {
         NextflowMeta.instance.preview.topic
+    }
+
+    @Memoized
+    static String bash() {
+        session().config.navigate('nextflow.defaults.bash') ?: '/bin/bash'
     }
 }
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -26,6 +26,7 @@ import java.nio.file.Path
 import groovy.transform.CompileStatic
 import groovy.transform.PackageScope
 import groovy.util.logging.Slf4j
+import nextflow.NF
 import nextflow.SysEnv
 import nextflow.container.ContainerBuilder
 import nextflow.container.DockerBuilder
@@ -84,8 +85,7 @@ class BashWrapperBuilder {
         catch( Exception e ) {
             log.warn "Invalid value for `NXF_DEBUG` variable: $str -- See http://www.nextflow.io/docs/latest/config.html#environment-variables"
         }
-        BASH = Collections.unmodifiableList( level > 0 ? ['/bin/bash','-uex'] : ['/bin/bash','-ue'] )
-
+        BASH = Collections.unmodifiableList( level > 0 ? [NF.bash(), '-uex'] : [NF.bash(), '-ue'] )
     }
 
 
@@ -567,7 +567,7 @@ class BashWrapperBuilder {
         // keep the shell path as "/bin/bash" when a non-custom "shell" attribute is specified
         // to not introduce unexpected changes due to the fact BASH is defined as "/bin/bash -eu" by default
         return shell.is(BASH)
-            ? "/bin/bash"
+            ? NF.bash()
             : shell.join(' ')
     }
 
@@ -600,7 +600,7 @@ class BashWrapperBuilder {
             if( (env || needChangeTaskWorkDir) && !containerConfig.entrypointOverride() ) {
                 if( needChangeTaskWorkDir )
                     cmd = 'cd $NXF_TASK_WORKDIR; ' + cmd
-                cmd = "/bin/bash -c \"$cmd\""
+                cmd = "${NF.bash()} -c \"$cmd\""
             }
             launcher = containerBuilder.getRunCommand(cmd)
         }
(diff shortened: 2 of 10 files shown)


modules/nextflow/src/main/groovy/nextflow/NF.groovy
```diff
@@ -79,4 +80,9 @@ class NF {
     static boolean isTopicChannelEnabled() {
         NextflowMeta.instance.preview.topic
     }
+
+    @Memoized
+    static String bash() {
+        session().config.navigate('nextflow.defaults.bash') ?: '/bin/bash'
```
bentsherman, line 86: This config scope was renamed to `workflow`, I would just call it `workflow.bash`

pditommaso: let's discuss about this. I think we need to nextflow wide setting (more tied to runtime than workflow concept)

modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
```diff
@@ -84,8 +85,7 @@ class BashWrapperBuilder {
         catch( Exception e ) {
             log.warn "Invalid value for `NXF_DEBUG` variable: $str -- See http://www.nextflow.io/docs/latest/config.html#environment-variables"
         }
-        BASH = Collections.unmodifiableList( level > 0 ? ['/bin/bash','-uex'] : ['/bin/bash','-ue'] )
-
+        BASH = Collections.unmodifiableList( level > 0 ? [NF.bash(), '-uex'] : [NF.bash(), '-ue'] )
```
bentsherman, line 88: This use of `NF.bash()` makes me think that the user bash setting can only be the bash path itself and can't include extra arguments like `env -S bash` or `bash -o pipefail`. Will that be acceptable?

pditommaso: The goal should be to specify a custom Bash path not to alter the shell option. That's possible in some extent via `process.shell`

rollf, line 88: My understanding is that `BASH` is used for both running the task on the host as well as within the container. This means `NF.bash` should be something that is acceptable in both situations (OR: one has to change the profile and `NF.bash` at the same time if one needs to use different interpreters depending on whether the task runs on a host or on a container). I also understand that `/bin/bash` is the only official entry point to run docker containers.

Maybe for some context, here is a [test](https://github.com/rollf/nixpkgs/blob/nextflow-pull/nixos/tests/nextflow.nix) script that I am using to check whether the Nextflow package is running as expected on NixOS. I'm linking this here to show what my goal is/was with #5684. (The test is easy to understand without knowledge of Nix.). 5684 would allow me to pass this test.

I think even today, I could use an unpatched Nextflow on NixOS if I "simply" adapt `process.shell` everywhere. Tracing might not work. I'd like to improve on that situation.

pditommaso: > This means NF.bash should be something that is acceptable in both situations

Yes, that's correct. 


> I think even today, I could use an unpatched Nextflow on NixOS if I "simply" adapt process.shell everywhere. Tracing might not work. I'd like to improve on that situation.

my understanding is that in NixOS the Bash interpreter is located at `/usr/bin/bash` so this PR should solve the problem with a minimal setting.

 Is it not the case?

pditommaso: Got it. What's preventing creating a symlink to "normalise" it?

rollf:
Thank you for addressing my issues @pditommaso.

NixOS currently [patches](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/pkgs/by-name/ne/nextflow/package.nix#L36-L45) Nextflow so that it can run there even though `/bin/bash` is not available. This works great unless one wants to enable [tracing](https://github.com/NixOS/nixpkgs/issues/350183). Instead of improving the patch, I thought it might be nice to solve the issue upstream. Hence my motivation for #5432. I closed that PR in favor of #5684 (because the latter is more clean and I also updated my initial implementation since #4209 got merged in the meantime).

Now, #5684 got also closed and my approach will not be merged (reasoning is mentioned [here](https://github.com/nextflow-io/nextflow/pull/5684#discussion_r1925358345)) in favor of this PR here. I believe this PR here does not improve the situation for NixOS (see my reasoning [here](https://github.com/nextflow-io/nextflow/pull/5696#discussion_r1925455955) and [here](https://github.com/nextflow-io/nextflow/pull/5684#discussion_r1923749688)).

Thus, I see two options now
* Either: I'm wrong and this PR here does indeed improve the situation for NixOS in all 4 scenarios ([docker on/off x tracing on/off](https://github.com/rollf/nixpkgs/blob/nextflow-pull/nixos/tests/nextflow.nix#L57-L62)) with minimal additional patching/configuration upfront but without having to touch the actual pipeline and without having to change `process.shell` when enabling/disabling docker/tracing.
* Or: This PR here does not help and I thus suggest to not merge if it was only motivated by supporting NixOS (since Nextflow doesn't need it). The Nix package definition for Nextflow will then continue to patch vanilla Nextflow; the patch will of course be improved to allow the tracing scenario (i.e. simply derive a patch from #5684 and apply it).

I can test option 1) but would need advice how you envision this to be used. However, we might also directly agree on option 2).

(For option 2, it would make my life easier if this PR here does not get merged. But  this is of course entirely up to others to decide)

modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
rollf: A Nix package cannot create arbitrary files on the file system. Actually, older definitions of the Nextflow package used a [wrapper](https://github.com/NixOS/nixpkgs/blob/nixos-24.05/pkgs/development/interpreters/nextflow/default.nix#L74) to make `nextflow` *believe* that `/bin/bash` exists. However, using this wrapper comes with other drawbacks.

rollf:
@pditommaso could you have a look at my [comment](https://github.com/nextflow-io/nextflow/pull/5696#issuecomment-2607740437) above and give some feedback? Thanks.

bentsherman:
Closing due to age. It seems we weren't able to decide whether this change was worthwhile

rollf:
@bentsherman I think this is fine for the NixOS community. I mentioned above that IMHO this PR here would have worsened the situation for NixOS. Besides, a Nix-specific [solution](https://github.com/NixOS/nixpkgs/pull/491967) has been recently discussed/imlemented/merged.

PR #5698, opened 2025-01-22
Remove redundant slash from s3 URI

robsyme:
Before this fix, calling the `toUri` method on an instance of nextflow.cloud.aws.nio.S3Path would print a triple-slash path:

```
s3:///bucket-name/path/to/SampleSheet.csv
```

This fix removes the redundant slash when there is no endpoint specified.


robsyme:
Nevermind - I see it's part of the URI spec. Closing

Issue #5363, opened 2024-10-04
nextflow 24.0.4.4 never exit due to incomplete file transfer


divinomas-gh:
### Expected behavior and actual behavior

I have used nextflow 22.10.6.5843, which runs smoothly. After I updated my nextflow to v24.0.4.4, the same script hangs with some files not finished for transferring. The files to be transferred are totally around 50Gb.

### Program output 
Oct-03 12:42:12.157 [main] DEBUG nextflow.Session - Session await > all processes finished
Oct-03 12:42:17.082 [Task monitor] DEBUG n.processor.TaskPollingMonitor - <<< barrier arrives (monitor: slurm) - terminating tasks monitor poll loop
Oct-03 12:42:17.082 [main] DEBUG nextflow.Session - Session await > all barriers passed
Oct-03 12:42:17.093 [main] DEBUG nextflow.util.ThreadPoolManager - Thread pool 'TaskFinalizer' shutdown completed (hard=false)
Oct-03 12:42:22.095 [main] INFO  nextflow.util.ThreadPoolHelper - Waiting for file transfers to complete (7 files)
Oct-03 12:43:22.102 [main] DEBUG nextflow.util.ThreadPoolHelper - Waiting for file transfers to complete (7 files)
Oct-03 12:44:22.104 [main] DEBUG nextflow.util.ThreadPoolHelper - Waiting for file transfers to complete (6 files)
Oct-03 12:45:22.106 [main] DEBUG nextflow.util.ThreadPoolHelper - Waiting for file transfers to complete (6 files)
Oct-03 12:46:22.108 [main] DEBUG nextflow.util.ThreadPoolHelper - Waiting for file transfers to complete (6 files)
Oct-03 12:47:22.110 [main] DEBUG nextflow.util.ThreadPoolHelper - Waiting for file transfers to complete (6 files)
Oct-03 12:48:22.112 [main] DEBUG nextflow.util.ThreadPoolHelper - Waiting for file transfers to complete (4 files)
Oct-03 12:49:22.114 [main] DEBUG nextflow.util.ThreadPoolHelper - Waiting for file transfers to complete (3 files)
Oct-03 12:50:22.116 [main] DEBUG nextflow.util.ThreadPoolHelper - Waiting for file transfers to complete (3 files)
.......
.......
.......
Oct-04 00:41:23.430 [main] DEBUG nextflow.util.ThreadPoolHelper - Waiting for file transfers to complete (3 files)
[...]

PR #5700, opened 2025-01-23
Use unbound queue for transfer thread pool

pditommaso:
This PR changes the default setting `ThreadPoolBuilder` class so that: 
* by default uses an unbound queue instead in order to prevent the triggering of the rejection policy when exceeding the queue size 
* use [BlockingBlockingQueue](https://github.com/nextflow-io/nextflow/blob/075ef4ccbec7caf122b966ff89af56d5b0351e30/modules/nextflow/src/main/groovy/nextflow/util/BlockingBlockingQueue.groovy#L33-L33) when the queue size is specified 
* use the `AbortPolicy` as rejection policy to prevent the issue reported [here](https://github.com/nextflow-io/nextflow/issues/5363#issuecomment-2607211371)
* default to `-1`(unbound) queue size in the [ThreadPoolManager](https://github.com/nextflow-io/nextflow/blob/075ef4ccbec7caf122b966ff89af56d5b0351e30/modules/nextflow/src/main/groovy/nextflow/util/ThreadPoolManager.groovy#L42-L42) 

Solves #5363

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/util/ThreadPoolBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/util/ThreadPoolBuilder.groovy
@@ -106,7 +106,8 @@ class ThreadPoolBuilder {
 
     ThreadPoolBuilder withQueueSize(int size) {
         this.queueSize = size
-        this.workQueue = new LinkedBlockingQueue<Runnable>(size)
+        if( size>0 )
+            this.workQueue = new BlockingBlockingQueue<Runnable>(size)
         return this
     }
 
@@ -141,7 +142,7 @@ class ThreadPoolBuilder {
         if( workQueue==null )
             workQueue = new LinkedBlockingQueue<>()
         if( rejectionPolicy==null )
-            rejectionPolicy = new ThreadPoolExecutor.CallerRunsPolicy()
+            rejectionPolicy = new ThreadPoolExecutor.AbortPolicy()
         if( threadFactory==null )
             threadFactory = new CustomThreadFactory(name)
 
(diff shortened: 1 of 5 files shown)


modules/nextflow/src/main/groovy/nextflow/util/ThreadPoolBuilder.groovy
```diff
@@ -141,7 +142,7 @@ class ThreadPoolBuilder {
         if( workQueue==null )
             workQueue = new LinkedBlockingQueue<>()
         if( rejectionPolicy==null )
-            rejectionPolicy = new ThreadPoolExecutor.CallerRunsPolicy()
+            rejectionPolicy = new ThreadPoolExecutor.AbortPolicy()
```
jorgee, line 145: This will change the default rejection policy to FinalizePool, PublishDir, FilePorter, and S3ThreadPool. Is this the intention? It could cause side effects. The error was in the S3ThreadPool, we could add a ThreadPoolManager.withRejectionPolicy method and set the AbortPolicy just for the S3 cases.

modules/nextflow/src/main/groovy/nextflow/util/ThreadPoolBuilder.groovy
```diff
@@ -160,17 +161,14 @@ class ThreadPoolBuilder {
...
     static ThreadPoolExecutor io(String name=null) {
         io(10, 100, 10_000, name)
     }
 
-
     static ThreadPoolExecutor io(int min, int max, int queue, String name=null) {
         new ThreadPoolBuilder()
                 .withMinSize(min)
                 .withMaxSize(max)
-                .withQueueSize(queue)
```
jorgee, line 173: Why is this removed?

pditommaso: The idea was to default on unbound queue, but I agree that's inconsistent. Above all it's not used anymore, it should be removed

pditommaso:
> It was failing with the master and working with this PR

well done! 

>  I was thinking of adding it as validation or AWS test, but it could hang the CI if it fails

not sure it's needed

modules/nextflow/src/main/groovy/nextflow/util/ThreadPoolBuilder.groovy
pditommaso: Removed

jorgee (approved):

Issue #5694, opened 2025-01-22
Google Batch job arrays do not play nicely with wave containers built from conda


chris-mbiomics:
## Bug report 

Hi, thanks for developing nextflow.
While moving some nextflow pipelines to google batch, I have encountered an issue with wave containers. They don't seem to work with job arrays.

### Expected behavior and actual behavior

I would expect the job array to perform just as well with job arrays. When using wave containers in conjunction with google batch arrays, it crashes instead of executing the tasks. Note that the local execution with the "conda" profile, the google batch execution without job arrays and the batch execution with arrays and a fixed container works fine. Just the wave containers seem to be an issue for some reason.

### Steps to reproduce the problem

nextflow.config
```

profiles {
    google_batch {
        wave.enabled=true
        process {
            executor = "google-batch"
        }
    }

    google_batch_array {
        wave.enabled=true
        process {
            executor = "google-batch"
            array=3
        }
    }

    google_batch_array_container {
        process {
            executor = "google-batch"
            array=3
            container = "wave.seqera.io/wt/291260ae7922/wave/build:python--5c1506391ad123b7"
        }
    }

    conda {
        conda.enabled = true
    }
}

```

main.nf

```
process python_version {
    conda 'python'

    input:
    val x

    output:
    stdout

    """
    echo $x
    python --version
    """
}

workflow {
    python_version(Channel.of(1, 2, 3)) \
    | view
}
```

### Program output 

`$ nextflow run main.nf -profile google_batch_array`

 ```
N E X T F L O W   ~  version 24.10.4

Launching `main.nf` [exotic_wright] DSL2 - revision: 862a03c6dd

[-        ] process > python_version [  0%] 0 of 3
[0d/652204] process > python_version (2) [100%] 3 of 3
ERROR ~ Error executing process > 'python_version (1)'

Caused by:
  Process `python_version (1)` failed because the container image was not specified


Command executed:
                                                                                                                                                                                          array=( /mnt/disks/disk/8e/9245115770a8ff6b89358d5070d39c /mnt/disks/disk/32/74bd85903d0f45014e00b8bd9eabae /mnt/disks/disk/0d/6522043ca2a83a05e4ce0adbb239a7 )
  export nxf_array_task_dir=${array[BATCH_TASK_INDEX]}
[...]

PR #5704, opened 2025-01-23
Fix Conda directive is not honoured by task array

pditommaso:
This PR fixes the usage of `conda` directive with task arrays. 

Solves #5694

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskArrayCollector.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskArrayCollector.groovy
@@ -44,6 +44,7 @@ class TaskArrayCollector {
             'accelerator',
             'arch',
             'clusterOptions',
+            'conda',
             'cpus',
             'disk',
             'machineType',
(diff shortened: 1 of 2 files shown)


bentsherman:
I need to update the docs for this. I think I didn't include `conda` in the list because the conda environment is normally handled by the task script. So each task could have a different conda environment. But with Wave, the conda environment needs to be uniform for the process because Wave uses conda to resolve the container image beforehand

modules/nextflow/src/main/groovy/nextflow/processor/TaskArrayCollector.groovy
```diff
@@ -56,6 +55,9 @@ class TaskArrayCollector {
             // only needed for container-native executors and/or Fusion
             'container',
             'containerOptions',
+            // only needed when using Wave
+            'conda',
+            'spack',
```
pditommaso, line 60: Wave is not supporting anymore `spack`. Do you think it needed anycase?

bentsherman: oh, in that case it should not be needed. I will revert

Issue #1061, opened 2019-03-05
watchPath() does not work for files added/ updated from a different machine


pinin4fjords:
### Expected behavior and actual behavior

Expect that files updated/ added to a directory will trigger watchPath() and downstream operations regardless of the machine from which those updates are made.

Actual behaviour seems to be that file additions/ updates are only noticed when done from the same machine. 

### Steps to reproduce the problem

Make a workflow with a watchDir() like:

```
#!/usr/bin/env nextflow

// Watch the SDRF directory for incoming SDRF files

sdrfDir = '/path/to/sdrf_dir'

Channel
    .watchPath( "${sdrfDir}/*.sdrf.txt", 'create,modify' )
    .subscribe { println "SDRF file: $it" }

...
```
Then test it by adding files to the watched directory a) from the same machine and b) by submitting the 'watcher' Nextflow to a cluster node, and running file updates from the head.

In a) the file additions/ updates are detected, in b) they are not. If you then SSH to the cluster node which is running the watcher, and do file updates there, this will trigger the expected behaviour.

### Environment 

* Nextflow version: 19.03.0-SNAPSHOT.5058
* Java version: openjdk 10.0.2 2018-07-17
* Operating system: Linux

pditommaso:
This depends on the Java [watch api](https://docs.oracle.com/javase/tutorial/essential/io/notification.html#overview) which in turns depends by the file system implementation. On shared file system there may be tricky behaviour. 

I'm labelling as a known issue and for which it would desirable a better implementation.

apeltzer:
Ran into this again as well, can confirm that this is indeed not working on NFS shared network mounts unfortunately as expected. Our pipeline works either via `Channel.fromFile()` or via  `Channel.watchPath()`  - the latter just works on local storage and not on network file systems.

pditommaso:
This is a grey area. NF relies on Java file watch API which in turn relies on POSIX notify events, that not all NFS implementation support (any?) 

https://unix.stackexchange.com/questions/81663/why-are-inotify-events-different-on-an-nfs-mount

PR #5712, opened 2025-01-24
 Add note to document watchPath incompatibility with cloud URIs

theJasonFan:
Changes:
- Updates docs, adding a note to document watchPath incompatibility with cloud URIs.

### DIFF
--- a/docs/reference/channel.md
+++ b/docs/reference/channel.md
@@ -508,6 +508,11 @@ that it will cause your pipeline to run forever. Consider using the `take` or `u
 a certain condition is met (e.g. after receiving 10 files, receiving a file named `DONE`).
 :::
 
+:::{note}
+The `channel.watchPath` relies on the Java file watch API which and POSIX notify events.
+As such, `channel.watchPath` may not work with NFS network mounts and will not work for cloud provider URIs (e.g. `s3://.../*.txt`).
+:::
+
 See also: [channel.fromPath](#frompath) factory method.
 
 [glob]: http://docs.oracle.com/javase/tutorial/essential/io/fileOps.html#glob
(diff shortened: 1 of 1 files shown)


docs/reference/channel.md
```diff
@@ -508,6 +508,11 @@ that it will cause your pipeline to run forever. Consider using the `take` or `u
 a certain condition is met (e.g. after receiving 10 files, receiving a file named `DONE`).
 :::
 
+:::{note}
+The `channel.watchPath` relies on the Java file watch API which and POSIX notify events.
+As such, `channel.watchPath` may not work with NFS network mounts and will not work for cloud provider URIs (e.g. `s3://.../*.txt`).
```
bentsherman, line 513: What's the issue with NFS?

theJasonFan: Per Paulo's comment here: https://github.com/nextflow-io/nextflow/issues/1061#issuecomment-933253132

Perhaps there is something better than "may not work"... it may depend on specific implementations. Unfortunately, I don't have many systems to test specific scenarios.

pditommaso: As of that comment theres' a new implementation that's safe with NFS https://github.com/nextflow-io/nextflow/blob/19c40a4a89c6357d3c0ee1c524e503a1baeebe25/modules/nextflow/src/main/groovy/nextflow/file/DirWatcherV2.groovy#L42-L42

theJasonFan: Nice! It's just a small note about URIs then.

docs/reference/channel.md
```diff
@@ -508,6 +508,11 @@ that it will cause your pipeline to run forever. Consider using the `take` or `u
 a certain condition is met (e.g. after receiving 10 files, receiving a file named `DONE`).
 :::
 
+:::{note}
+The `channel.watchPath` relies on the Java file watch API which and POSIX notify events.
```
christopher-hakkaart, line 512:
```suggestion
The `channel.watchPath` relies on the Java file watch API and POSIX to notify events.
```
I don't think this point makes sense - is this still correct?

bentsherman: I will rewrite this note

pditommaso (approved):

PR #5715, opened 2025-01-27
Data lineage tracking (aka CID store)

pditommaso:
Tentative implementation for addressable data store (very basic POC so far).

### Update on 1 Mar 2025 from #5787 by @jorgee 

M1 Implementation of CID store for provenance

Changes:
- CID store is specified by workflow.data.store.location
- Workflow Hash is created based on the workflow and parameters description
- workflow, tasks and outputs metadata are stored in <cid.store.location>/.meta
- references to other cid metadata are cid://<workflow_hash|task_hash/output_target_path
- CID NIO Filesystem to access data based on CIS URLs
- nextflow cid command to log, show and get lineage from CID store metadata

Known Limitations:
- Outputs which are not published in absolutePaths or URLs which are not subfolders both the outputDir, we can not infer the relative output target path. They are not currently tracked in the CID store. We could create a hash for the parent directory of the URL or absolute path and use it as relative folder.


pditommaso:
@jorgee apologies, can latest changes be made as PR against this branch? so it will be much simpler do understand what's new for me

jorgee:
> @jorgee apologies, can latest changes be made as PR against this branch? so it will be much simpler do understand what's new for me

 I have reverted changed in this branch and created a new one in PR #5787

pditommaso:
The following exception is reported then using `nextflow cid log` but not config is provided 

```
Mar-01 16:16:26.201 [main] ERROR nextflow.cli.Launcher - @unknown
java.lang.NullPointerException: Cannot invoke "nextflow.data.cid.CidStore.getHistoryFile()" because "store" is null
	at nextflow.cli.CmdCid$CmdLog.printHistory(CmdCid.groovy:124)
	at nextflow.cli.CmdCid$CmdLog.apply(CmdCid.groovy:117)
	at nextflow.cli.CmdCid.run(CmdCid.groovy:81)
	at nextflow.cli.Launcher.run(Launcher.groovy:504)
	at nextflow.cli.Launcher.main(Launcher.groovy:659)
```

pditommaso:
Minor to do:
 
- [ ] Fix `cid log` exception when config is missing
- [ ] Add `nextflow cid -h` help

modules/nextflow/src/main/groovy/nextflow/data/cid/DefaultCidStore.groovy
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    Path getPath(){ location }
+
+    @Override
+    CidHistoryFile getHistoryFile(){
+        return new CidHistoryFile(metaLocation.resolve(HISTORY_FILE_NAME))
+    }
+
+    static Path getMetadataPath(DataConfig config){ config.store.location.resolve('.meta') }
```
pditommaso, line 79: This could be an attribute of `DataStoreOpts`

jorgee:
Where is it better to make the changes from your [comments] (https://github.com/nextflow-io/nextflow/pull/5715#issuecomment-2692286200). In this PR, in #5833 or a new one

pditommaso:
Let's do directly here, they should be small changes

pditommaso:
Considering the checksum can be computed in different ways, it should be tracked the algorithm used along with the checksum value. I've added a "task" checkbox about this in the comment above.

modules/nextflow/src/main/groovy/nextflow/data/cid/CidObserver.groovy
bentsherman, line 1: @jorgee you can use the `onWorkflowPublish` event to capture the output metadata (i.e. annotations) from the workflow outputs. This event comes from PublishOp if you want to see how it works. Alternatively, we can modify PublishOp to send the entire metadata for an output when it's done.

I might sketch a PR for this later if I have some time

bentsherman: Right now there is only this e2e test: https://github.com/nextflow-io/nextflow/blob/master/tests/output-dsl.nf

jorgee:
> Considering the checksum can be computed in different ways, it should be tracked the algorithm used along with the checksum value. I've added a "task" checkbox about this in the comment above.

I am considering adding the algorithm inside the checksum field like the container image digest.

`checksum="<algorithm>:<hash>"` 

However, the hashing algorithm used by Nexflow depends on the mode and type of data. I should replicate the same code to extract the algorithm, and if I am not wrong, it could also be a combination of algorithms. For instance, in the case of files or directories, it is using a sha-256 hash pass to the default murmur3_128 hasher. So at this moment, I will put something like checksum "nf-<mode>:<hash>", to indicate we are using the Nextflow hashing and the selected mode.

pditommaso:
Not understanding the rationale of the mode over the actual algorithm

pditommaso:
I'd considering using something like to avoid collapsing everything in the prefix (and make more extensible if needed) 

```
checksum: { value:<>, alg:<>, mode:<> }
```

modules/nextflow/src/main/groovy/nextflow/data/cid/CidStore.groovy
```diff
@@ -0,0 +1,44 @@
...
+ * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
+ */
+@CompileStatic
+interface CidStore {
+
+    void open(DataConfig config)
+
+    void save(String key, Object value)
+
+    void list(String key, Consumer<String> consumer)
```
jorgee, line 36: I was writing the JavaDoc for the interfaces and I have seen this function is not used.  @pditommaso, do you remember what was the purpose of this method? In the DefaultCidStore it is returning the files in the key, but currently it is just the .data.json file, because data is not stored in the CidStore. Not sure if it is needed anymore.

jorgee: Yes the list. I haven't modified it. It was as it is from the first commit. But it has never been used. So, I will remove it. I will add a similar one in case it is needed in the future.

pditommaso:
Minor this warning message should not container the object address 

```
WARN: Can't read CID history file: nextflow.data.cid.CidHistoryFile@43b172e3
```

pditommaso:
Some quick wins discussed this morning with @jorgee 

* Use `Instant` object instead of strings in model objects
* Unify nomenclature using `taskRun` and `workflowRun` instead of publishBy, runBy, etc
* Use `#outputs` fragment using run CID instead of a separate result CID
* Use `DataOutput` in place of WorkflowOutput and TaskOutput

modules/nextflow/src/main/groovy/nextflow/script/OutputDsl.groovy
```diff
@@ -171,6 +171,14 @@ class OutputDsl {
             setOption('tags', value)
         }
 
+        void annotations(Map value) {
+            setOption('annotations', value)
+        }
+
+        void annotations(Closure value) {
+            setOption('annotations', value)
+        }
```
bentsherman, line 180: TODO: decide whether to use tags or phase it out in favor of annotations

pditommaso:
Made a few changes to cleanup and simplify some commons patterns. 

As general comment try reducing average cyclomatic complexity using smaller method and simpler if-block using [if-guard pattern](https://medium.com/@scadge/if-statements-design-guard-clauses-might-be-all-you-need-67219a1a981a) (an example [here](https://github.com/nextflow-io/nextflow/pull/5715/commits/9b4addf45ad1fb3d2f69a6e86f0485a8f69e31ae)) 

Also it should be increased the test coverage, there are large part not tested. 

Screenshot 2025-04-05 at 22 30 16

pditommaso:
A few more comments 

1. we should think how to unify the model for inputs and outputs. Now are used both `List<DataPath>`, `List<Parameter>` and `Map<String, Object>` 
2. along the same manner it would be nice unify `mainScriptFile` and `otherScriptFile` into a single collection `scriptFiles` 
3. it would be useful to have a symmetry on the usage `/outputs` and `#outputs`  for tasks along the same manner of workflow

modules/nf-cid/src/main/nextflow/data/cid/CidHistoryFile.groovy
```diff
@@ -0,0 +1,170 @@
...
+                if (current.sessionId == id) {
+                    log.trace("Updating record for $id in CID history file ${FilesEx.toUriString(this.path)}")
+                    final newRecord = new CidHistoryRecord(current.timestamp, current.runName, current.sessionId, runCid)
+                    newHistory << newRecord.toString() << '\n'
+                } else {
+                    newHistory << line << '\n'
+                }
+            }
+            catch (IllegalArgumentException e) {
+                log.warn("Can't read CID history file: ${FilesEx.toUriString(this.path)}", e.message)
```
pditommaso, line 106: This should be review to make it work without the need for locks.

modules/nf-cid/src/main/nextflow/data/cid/CidObserver.groovy
```diff
@@ -0,0 +1,415 @@
...
+    void onFlowComplete(){
+        if (this.workflowResults){
+            workflowResults.createdAt = Instant.now()
+            final key = executionHash + SEPARATOR + 'outputs'
+            this.store.save(key, workflowResults)
+        }
+    }
+
+    protected String storeWorkflowRun() {
+        final normalizer = new PathNormalizer(session.workflowMetadata)
```
pditommaso, line 103: Likely this can be a class attribute to avoid creating a new instance for each task, see below

modules/nf-cid/src/main/nextflow/data/cid/model/DataOutput.groovy
```diff
@@ -0,0 +1,42 @@
...
+@CompileStatic
+class DataOutput implements CidSerializable {
+    String path
+    Checksum checksum
+    String source
+    String run
+    long size
+    Instant createdAt
+    Instant modifiedAt
+    Map annotations
```
pditommaso, line 41: Let's add a description for these fields

modules/nf-cid/src/main/nextflow/data/cid/CidHistoryFile.groovy
jorgee: I followed the same as the executions log file. Two executions could write the log. The other option is converting the `.history ` to a folder and have a file per run, the history log will be the list of the content of the files.

pditommaso: > The other option is converting the .history  to a folder and have a file per run, the history log will be the list of the content of the files

I was thinking the same

pditommaso:
[This test](https://github.com/nextflow-io/nextflow/blob/f1a87704c18076aa6985253cb3b5bf212d17368a/modules/nf-cid/src/test/nextflow/data/cid/CidObserverTest.groovy#L183-L183) fails on macOs this message: 

```
taskRunResult == taskDescription
|             |  |
|             |  nextflow.data.cid.model.TaskRun(e74471e2-05ee-46a0-98ca-6dfa409c0ce7, foo, nextflow.data.cid.model.Checksum(a71d06a9f50695c56a5b9933d1a38846, nextflow, standard), nextflow.data.cid.model.Checksum(1cc1165b39e97fc6b173579e907ecd66, nextflow, standard), [nextflow.data.cid.model.Parameter(FileInParam, file1, [cid://78567890/outputs/file1.txt]), nextflow.data.cid.model.Parameter(FileInParam, file2, [[path:file:///var/folders/rx/l2dmn8g951v161x04zshl0r80000gn/T/test1162005281900395322/file2.txt, checksum:[value:a64d2f35d0c2285cf3e61d6862771892, algorithm:nextflow, mode:standard]]]), nextflow.data.cid.model.Parameter(ValueInParam, id, value)], null, null, null, null, [:], [], cid://hash, null)
|             false
nextflow.data.cid.model.TaskRun(e74471e2-05ee-46a0-98ca-6dfa409c0ce7, foo, nextflow.data.cid.model.Checksum(a71d06a9f50695c56a5b9933d1a38846, nextflow, standard), nextflow.data.cid.model.Checksum(1cc1165b39e97fc6b173579e907ecd66, nextflow, standard), [nextflow.data.cid.model.Parameter(FileInParam, file1, [cid://78567890/outputs/file1.txt]), nextflow.data.cid.model.Parameter(FileInParam, file2, [[path:file:///private/var/folders/rx/l2dmn8g951v161x04zshl0r80000gn/T/test1162005281900395322/file2.txt, checksum:[value:68bb07bd21a74617443352c058bd743d, algorithm:nextflow, mode:standard]]]), nextflow.data.cid.model.Parameter(ValueInParam, id, value)], null, null, null, null, [:], [], cid://hash, null)

```


The problem is this 

```
file:///var/folders/rx/l2dmn8g951v161x04zshl0r80000gn/T/test1162005281900395322/file2.txt
```

vs 

```
file:///private/var/folders/rx/l2dmn8g951v161x04zshl0r80000gn/T/test1162005281900395322/file2.txt
```

Very likely is caused by the use of `toRealPath` but cannot find where it's invoked.

pditommaso:
Also this could introduce some unexpected side effect with symlinks. We may to introduce a specific method for it.  

https://github.com/nextflow-io/nextflow/blob/f1a87704c18076aa6985253cb3b5bf212d17368a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy#L1945-L1947

modules/nf-cid/src/main/nextflow/data/cid/fs/CidPath.groovy
```diff
@@ -0,0 +1,492 @@
...
+        return this.getTargetPath(true)
+    }
+
+    Path toTargetPath() {
+        return getTargetPath(true)
+    }
+
+    protected Path getTargetPath(boolean resultsAsPath=false){
+        return findTarget(fileSystem, filePath, resultsAsPath, parseChildrenFormFragment(fragment))
+    }
```
pditommaso, line 440: @jorgee is the `resultsAsPath` parameter really needed. I've noticed that almost always invoked with `true`, except here 

https://github.com/nextflow-io/nextflow/blob/a8765e6af393811b4861ad9885e2a59d692cc8ee/modules/nf-cid/src/main/nextflow/data/cid/fs/CidFileSystemProvider.groovy#L213-L213

if needed it may be better to split in two methods with different names that clarifies the semantic

pditommaso:
@jorgee can you have a look at these conflicts? I've tried but not 100% sure

jorgee:
> @jorgee can you have a look at these conflicts? I've tried but not 100% sure

Done, there was some functionality from workflow-outputs that I also added in the branch

pditommaso:
Wonder if the `resolvedConfig` should not be instead a `Map` 

https://github.com/nextflow-io/nextflow/blob/f7d84ba0f26b9d881482ab1a29cd9717933d2b44/modules/nf-cid/src/main/nextflow/data/cid/model/WorkflowRun.groovy#L50-L50

modules/nf-cid/src/main/nextflow/data/cid/fs/CidPath.groovy
jorgee: Yes, I will change it, but not sure how to call them yet.

jorgee:
I think the content is the same as session.config, but the secrets are removed. I will check if I can apply the secrets remove function on a copy of the session's config map.

pditommaso:
[This](https://github.com/nextflow-io/nextflow/blob/1c06e6932b7d559d477b75790449b1871cb6b9e4/modules/nextflow/src/main/groovy/nextflow/util/SecretHelper.groovy#L24-L24) should remove secrets from a map tho from a lineage point of view makes not much sense :/

modules/nf-cid/src/main/nextflow/data/cid/fs/CidPath.groovy
pditommaso: The second is a the pseudo path for `#outputs` ?

jorgee: Yes, the second is for the `#outputs` or other metadata fragment that is returned as in-memory file (`CidMetadataPath`). But it also returns the target path if the key is from a `DataOutput`. I have kept `getTargetPath()` when it was `false`, and I have included `getTargetOrMetadataPath()` when it was `true`. Let me know if you find a better name.

pditommaso: I think it's fine

jorgee:
Pushed minor fixes for the render. It was failing because of the change of the task inputs type (FileInParam is now path). There was also the problem of including forbidden characters in the Mermaid node id. Moreover, I have added a default value for the html file. Users only need to specify the data to render. 

I have also checked what happens when you pass a task run or workflow run LID.  In the case of task runs, it renders the task graph starting from the requested task and its predecessors based on file parameter dependencies. In the case of workflows, it renders the workflow and the input parameters. It was unintentional, but I will keep it, just an extra functionality for free.

 Finally, I have added the closest property in the validation errors when parsing the fragments or query strings.

pditommaso:
It looks in a good shape. I've made some tests and a a few little changes. Some notes:

1. better not use `.` in error messages. It may be confusing when preceding URIs or numbers. 
1. Timestamps have been changes to `OffsetDateTime` using the *current* time zone. This allow keep track of the current location time. ✅ 
1. Added `li` command shortcut. ✅ 
1. We may want to change `resolvedConfig` to `config` for simplicity 
3. We may want to change `XxxOutputs` and `#outputs` to `XxxOutput`. Along the same manner `inputs` to `input`
4. Find `describe` command verbose. What about `view` or `print` instead? 
5. Likely TaskRun should include the resolved task script 
6. Command `lineage find 'type=DataOutput'` work OK, but equivalent `channel.fromPath('lid:///?type=DataOutput')` seems not working 🔴

bentsherman:
> Find describe command verbose. What about view or print instead?

`view` would be consistent with the `view` command, especially if we merge #5966

> Command lineage find 'type=DataOutput' work OK, but equivalent channel.fromPath('lid:///?type=DataOutput') seems not working

I don't think this is meant to work? The query isn't really returning a single JSON but a collection of JSONs. This is why I suggested to remove the `lid://` prefix when making a query, to not give the illusion that it's a LID path

modules/nf-lineage/src/main/nextflow/lineage/model/DataOutput.groovy
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Model a base class for workflow and task outputs
+ *
+ * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
+ */
+@Canonical
+@CompileStatic
+class DataOutput implements LinSerializable {
+    /**
```
bentsherman, line 33: Is this class used for file outputs? If so then I would call it `FileOutput`, I'm having a hard time understanding what it's used for

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
```diff
@@ -214,7 +219,7 @@ class PublishOp {
             else {
                 log.warn "Invalid extension '${ext}' for index file '${indexPath}' -- should be CSV, JSON, or YAML"
             }
-            session.notifyFilePublish(indexPath)
+            session.notifyFilePublish(indexPath, null, publishOpts.tags as Map)
```
bentsherman, line 222: Should this be `annotations` instead of `tags`?

modules/nextflow/src/main/groovy/nextflow/processor/TaskId.groovy
```diff
@@ -38,6 +38,8 @@ class TaskId extends Number implements Comparable, Serializable, Cloneable {
 
     private final int value
 
+    int getValue() { value }
```
bentsherman, line 41: There is already `intValue()` for this, see below

modules/nf-lineage/src/main/nextflow/lineage/config/LineageConfig.groovy
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Model workflow data lineage config
+ *
+ * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
+ */
+@CompileStatic
+class LineageConfig {
+
+    final LineageStoreOpts store
```
bentsherman, line 31: If the only option under this scope is `location`, then I think we can shorten it to `workflow.lineage.store`

modules/nf-lineage/src/main/nextflow/lineage/model/WorkflowOutputs.groovy
```diff
@@ -0,0 +1,49 @@
...
+import java.time.OffsetDateTime
+
+/**
+ * Models the results of a workflow execution.
+ *
+ * @author Jorge Ejarque <jorge.ejarque@seqera.io
+ */
+@Canonical
+@CompileStatic
+class WorkflowOutputs implements LinSerializable {
```
bentsherman, line 32: Continuing the discussion here about making workflow runs Merkle-compliant:
1. Rename `WorkflowRun` to `WorkflowLaunch`
2. Rename this class to `WorkflowRun`
3. Rename the `workflowRun` field in this class to `launch`
4. Use the hash of this class as the "workflow run id" in the lineage log

This way, the `WorkflowRun` represents the "final" record that is created and the entrypoint into the lineage from the runs log. It resolves the issue where `lid://<hash>#outputs` requires a reverse lookup. Even if we never go full CID, it still makes sense to do it

bentsherman: We could do a similar thing for tasks, i.e. `TaskRun` -> `TaskLaunch` and `TaskOutputs` -> `TaskRun`

modules/nextflow/src/main/groovy/nextflow/cli/CmdLineage.groovy
```diff
@@ -0,0 +1,291 @@
...
+        if( matches )
+            msg += " -- Did you mean one of these?\n" + matches.collect { "  $it"}.join('\n')
+        throw new AbortOperationException(msg)
+    }
+
+    class CmdLog implements SubCmd {
+
+        @Override
+        String getName() {
+            return 'log'
```
bentsherman, line 155: If we rename `describe` to `view`, maybe we should also rename `log` to `list` to align more with the pipeline commands, and to not confuse with `nextflow log` or `.nextflow.log`

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
```diff
@@ -257,7 +262,7 @@ class PublishOp {
      */
     protected Object normalizePaths(value, targetResolver) {
         if( value instanceof Path ) {
-            return List.of(value.getBaseName(), normalizePath(value, targetResolver))
+            return normalizePath(value, targetResolver)
```
bentsherman, line 265: I tested with rnaseq-nf (`workflow-outputs-3` branch) and received the following workflow output structure:

```json
[
  {
    "type": "Collection",
    "name": "samples",
    "value": [ /* ... */ ]
  },
  {
    "type": "Collection",
    "name": "summary",
    "value": [
      "multiqc_report",
      "lid://494b7281ea2e02e985636dfbbcf6b8b3/multiqc_report.html"
    ]
  }
]
```

But this is not quite right. The `summary` output should just be a file. Wrapping in a list might be nice for a CSV file but it obscures the output here and also causes the LinObserver to infer the wrong type (Collection).

Applying this change produces the correct output structure:
```json
[
  {
    "type": "Collection",
    "name": "samples",
    "value": [ /* ... */ ]
  },
  {
    "type": "Path",
    "name": "summary",
    "value": "lid://2265a814fd1c205ecc5b629070d759e2/multiqc_report.html"
  }
]
```

modules/nf-lineage/src/main/nextflow/lineage/fs/LinPath.groovy
bentsherman, line 1: After playing around with the `lineage` command, I am skeptical about how much we are overloading this `lid` pseudo-filesystem. I thought it was just a nice add-on that we could experiment with, but now I think it's just getting in the way.

Currently there are three main uses for `lid` paths:

1. `lid://<hash>[#props]`: returns a metadata record or sub-path. This has no practical utility in a Nextflow script, not even for workflow outputs. Now that `#outputs` is a list, I can't access an output by name (e.g. `#outputs.samples`), which means I can't use `channel.fromPath()` to access an LID output in the same way as a samplesheet. So the LID output is no longer a drop-in replacement for samplesheets.

  On the command line, it would be simpler to just provide the hash and use `jq`:
  ```bash
  # before
  nextflow li describe lid://<hash>#params

  # after
  nextflow li describe <hash> | jq .params
  ```

  In a web interface like the platform, you'll use a graphical interface to navigate this metadata, so the fragment syntax is not needed there.

2. `lid:///?<name>=<value>&...`: used by the `find` command to retrieve a collection of metadata records. This also has no utility in a Nextflow script, because it is unrelated to domain-specific data like `#outputs`. It is only used by the `find` command, so the URI syntax is just getting in the way:

  ```bash
  # before
  # oops, forgot to escape the & ...
  nextflow li find lid:///?type=DataOutput&workflowRun=lid://2265a814fd1c205ecc5b629070d759e2

  # after
  nextflow li find type=DataOutput workflowRun=2265a814fd1c205ecc5b629070d759e2
  ```

3. `lid://<hash>/<path>`: returns a content-addressed file. This is the original use case and the only one that still makes sense as far as I can tell. I think this works perfectly both on the command line and in the Nextflow script/runtime.

Based on this analysis, I think we should ditch (1) and (2) entirely and use `lid://` only to refer to files.

Maybe we could use the fragment to refer to a specific output, e.g. `lid://<hash>#samples`. That would at least restore the original use case of passing a workflow output as input to a downstream pipeline.

modules/nf-lineage/src/main/nextflow/lineage/model/DataOutput.groovy
jorgee: This is used to describe the published files. They can be files or dirs, it is the reason why I called data but  no problem to change to file to be more coherent with the rest of the code.

jorgee:
> 8\. Command `lineage find 'type=DataOutput'` work OK, but equivalent `channel.fromPath('lid:///?type=DataOutput')` seems not working 🔴

The `?` is interpreted as a glob. But setting the glob: false option is also failing. It is not detecting is a global query. I will check it. However, it think the result in this case could be a bit undetermined. what do you expect to have in this case? the DataOutput descriptions or the real target paths?

modules/nf-lineage/src/main/nextflow/lineage/model/WorkflowOutputs.groovy
pditommaso: I'm still a bit hesitant about this change. I'd prefer to address some quick wins above, merge and keep this for next iteration/discussion

bentsherman: Once we release a first version and people start playing with it, it's going to be harder to make changes like renaming entities. This is why I want to resolve this before 25.04

pditommaso: No hurry, this is going to be a preview release

modules/nextflow/src/main/groovy/nextflow/cli/CmdLineage.groovy
pditommaso: Fair enough

modules/nextflow/src/main/groovy/nextflow/processor/TaskId.groovy
jorgee: removed in last commits

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
jorgee: Fixed in last commit

jorgee:
Based on previous comments, I have pushed some minor changes:
- `WorkflowRun.resolvedConfig` to `config` 
- `XxxOutputs` -> `XxxOutput`, `outputs` to `output` and `inputs` to `input`
- Command `describe` to `view` and `log` to `list`
- `TaskRun.scriptChckesum` -> resolved task script
- `DataOutput` -> `FileOutput`

bentsherman:
TODOs from our discussion

This PR:
- rename `log` to `list`
- rename `describe` to `view`

Separate PR(s);
- document `lineage` command
- add `channel.fromLineage()` factory to query metadata
- remove URL query syntax
  - use key-value pairs on command line
  - use key-value pairs in channel factory, e.g. `channel.fromLineage(foo: 'bar')`
- simplify URL fragment syntax
  - use `jq` on command line
  - use json-path in channel factory (I think `splitJson` operator already implements json-path)
  - use URL fragment to retrieve workflow output by name e.g. `channel.fromPath('lid://<hash>#samples')`

pditommaso:
Ok, I've moved the h2 stuff in the corresponding repo https://github.com/nextflow-io/nf-lineage-h2

jorgee:
> This PR:
> 
> * rename `log` to `list`
> * rename `describe` to `view`

This is already changed in the current PR

pditommaso:
Let's move ahead, then

modules/nf-lineage/src/main/nextflow/lineage/DefaultLinStore.groovy
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
+ */
+@Slf4j
+@CompileStatic
+class DefaultLinStore implements LinStore {
+
+    private static String HISTORY_FILE_NAME = ".history"
+    private static final String METADATA_FILE = '.data.json'
+    private static final String METADATA_PATH = '.meta'
```
bentsherman, line 45: @jorgee what is the rationale for the `.meta` subfolder? It looks like the only thing that is created in the `lineage` folder. Do you intend to store other things under `lineage` as well?

jorgee: @bentsherman, it was storing the output data and the metadata in the first implementation, but currently this sub-folder is not needed.

bentsherman: Okay, not urgent but something to consider in the final cleanup before 25.04

PR #5721, opened 2025-01-28
feat(azure-batch): Ignore pool already exists error

### LINKED ISSUE #5713: Ignore "pool exists" error on Azure Batch
## New feature

If you submit an Azure Batch pipeline which tries to create a pool when the pool already exists, you get the following error:

```
Caused by:
  Status code 409, "{
    "odata.metadata":"https://mybatch.eastus.batch.azure.com/$metadata#Microsoft.Azure.Batch.Protocol.Entities.Container.errors/@Element","code":"PoolExists","message":{
      "lang":"en-US","value":"The specified pool already exists.\nRequestId:d9475bc1-f9e5-492b-9114-6a05a8a57abc\nTime:2025-01-23T14:15:51.6526349Z"
    }
  }"

```

We can just ignore this error since the pool is being created.

## Usage scenario 

Related to #5575 

## Suggest implementation 

We can add the code here: https://github.com/nextflow-io/nextflow/blob/19c40a4a89c6357d3c0ee1c524e503a1baeebe25/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L926

or an explicit exception here: https://github.com/nextflow-io/nextflow/blob/19c40a4a89c6357d3c0ee1c524e503a1baeebe25/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L935-L947

adamrtalbot:
Ignores pool already exists error and continues. Fixes #5713

I can't get the mocks working for the tests :(


plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -711,77 +711,85 @@ class AzBatchService implements Closeable {
     }
 
     protected void createPool(AzVmPoolSpec spec) {
+        try {
```
bentsherman, line 714: Generally speaking, when a long function needs to be wrapped in a try-catch like this, we like to put the try-catch in a separate wrapper function:

```groovy
protected void safeCreatePool(AzVmPoolSpec spec) {
    try {
        createPool(spec)
    }
    catch( HttpResponseException e ) {
        // ...
    }
}
```

pditommaso:
This PR seems overkill compared the stated goal. Can you explain how you are trying to solved this problem, so we can assist you with this issue?

PR #5690, opened 2025-01-21
Add Google Batch NOT_FOUND error management

jorgee:
close #5422
close #5727

This PR includes a possible fix for the NotFoundException returned by the Google Batch API when getting some tasks.

When the client.getTaskStatus throws a NotFoundException it is caught and managed in the following way:
- If the retrieved from client.listTasks has several tasks. It tries to find the task and check the status 
- otherwise or if the task is not found in the list it retrieves the job status.
- If it is not able to neither get the task or job status it returns pending

Including a unit test producing the NotFoundException to validate the logic.

Two corner cases could not be correctly managed. 
1. When neither task status nor a job status is found, it is returning PENDING. I think that it could happen in the initial stage when Google Batch is creating the job and tasks. I am assuming that a task or job status will be received at some point in the execution. It is the same management that we did when no tasks were found in the job. 
2. When a task is not found in Google Batch, it belongs to a task array job and the job status is RUNNING or FAILED, we could get an incorrect task status. It is not very important in the case of RUNNING, but with FAILED we could get a task failed but, its state is unknown. According to the API documentation, the RUNNING or FAILED job states mean that at least one of the tasks is in this state. Job could be FAILED when there has been a job failure (such as the invalid type one we saw in #5550). So, I am assuming that when the task status is not found, it is not in the list and the Job is FAILED, It is due to a job failure and the task is also FAILED. Other alternatives that I have considered are: set the task to PENDING, but I think we could get a deadlock if it is job failure; or throw an exception to abort the execution. Other suggestions are welcome.

PR #5723, opened 2025-01-29
Alternative for managing task array status in Google Batch

jorgee:
This PR is an alternative to process task status in task arrays. 
In the current status, every time that getTaskState is called, we perform two calls to the Google Batch client, one to retrieve the list of tasks(`listTasks`) and another to get the task state (`getTaskStatus`). This second call is problematic because it generates the NotFoundException and it is also redundant because the first call already provides the task descriptions including their states.

The BatchClient has a HashMap as a cache of array task status. A new method is created to retrieve the status of a task belonging to a task array. Instead of making the call to the Google Batch API, it checks the status in the cache. When the status is not in the cache or is outdated, the `listTasks` method is called to update all the array tasks statuses. So, the rest of the array tasks do not require querying the Google Batch API again. 

When there is no task status, it returns null and fallbacks to the `getJobStatus.` This is the same as we were doing when no tasks were retrieved from the job or there was a `NotFoundException`.

The invalidation time is 1 second because it is the same as the one in `GoogleBatchTaskHandler`. Another alternative is setting it with the same value as the polling interval.

### DIFF
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/client/BatchClient.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/client/BatchClient.groovy
@@ -50,11 +50,12 @@ import groovy.util.logging.Slf4j
 @Slf4j
 @CompileStatic
 class BatchClient {
-
+    private static long TASK_STATE_INVALID_TIME = 1_000
     protected String projectId
     protected String location
     protected BatchServiceClient batchServiceClient
     protected BatchConfig config
+    private Map<String, TaskStatusRecord> arrayTaskStatus = new HashMap<String, TaskStatusRecord>()
 
     BatchClient(BatchConfig config) {
         this.config = config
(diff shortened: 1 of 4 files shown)


pditommaso:
Bit lost, is there any relation between this PR and #5690 ?

jorgee:
> Bit lost, is there any relation between this PR and #5690 ?

Yes, in #5690 we revert to the old job status management when task is not an array, but we kept the same for task arrays. In this PR, I also changed the way how task array statuses are queried to use the listTasks instead of getTaskStatus. This also avoids the NotFoundException because we do not call the problematic getTaskStatus.

pditommaso:
Is this ready for review?

plugins/nf-google/src/main/nextflow/cloud/google/batch/client/BatchClient.groovy
```diff
@@ -50,11 +50,12 @@ import groovy.util.logging.Slf4j
 @Slf4j
 @CompileStatic
 class BatchClient {
-
+    private static long TASK_STATE_INVALID_TIME = 1_000
```
pditommaso, line 53:
```suggestion
    private final static long TASK_STATE_INVALID_TIME = 1_000
```

plugins/nf-google/src/main/nextflow/cloud/google/batch/client/BatchClient.groovy
```diff
@@ -198,4 +199,40 @@ class BatchClient {
...
+            record = arrayTaskStatus.get(taskName)
+        }
+        return record?.status
+    }
+
+    private void updateArrayTasks(String jobId, long now){
+        for( Task t: listTasks(jobId) ){
+            arrayTaskStatus.put(t.name, new TaskStatusRecord(t.status, now))
+        }
+
```
pditommaso, line 220:
```suggestion
```

plugins/nf-google/src/main/nextflow/cloud/google/batch/client/BatchClient.groovy
```diff
@@ -198,4 +199,40 @@ class BatchClient {
...
+            log.debug("[GOOGLE BATCH] Updating tasks status for job $jobId")
+            updateArrayTasks(jobId, now)
+            record = arrayTaskStatus.get(taskName)
+        }
+        return record?.status
+    }
+
+    private void updateArrayTasks(String jobId, long now){
+        for( Task t: listTasks(jobId) ){
+            arrayTaskStatus.put(t.name, new TaskStatusRecord(t.status, now))
```
pditommaso, line 218: there can be any race condition on accessing the `arrayTaskStatus` map?

pditommaso:
Moving fwd this

PR #5730, opened 2025-01-31
Update HyperQueueExecutor.groovy

chaetognatha:
nextflow 24.10 and 24.11 edge produce errors when using this executor and the error seems to be due to the --log directive not being recognized by hq

from reviewing hq 0.21.0 submit command flags, it looks like the logs flag should be replaced with the stream flag

these are the errors in question:

```sh
ERROR ~ Error executing process > 'NFCORE_RNASEQ:RNASEQ:CAT_FASTQ (4)'                                                                                                                                               
                                                                                                                                                                                                                     
Caused by:                                                                                                                                                                                                           
  Failed to submit process to grid scheduler for execution                                                                                                                                                           
                                                                                                                                                                                                                     
                                                                                                                                                                                                                     
Command executed:                                                                                                                                                                                                    
                                                                                                                                                                                                                     
  hq --output-mode=quiet submit --directives=file .command.run                                                                                                                                                       
                                                                                                                                                                                                                     
Command exit status:                                                                                                                                                                                                 
  1                                                                                                                                                                                                                  
                                                                                                                                                                                                                     
Command output:                                                                                                                                                                                                      
  Could not parse directives from .command.run: You have used invalid parameter(s) after a #HQ directive.                                                                                                            
  error: unexpected argument '--log' found                                                                                                                                                                           
                                                                                                                                                                                                                     
  Usage: #HQ <hq submit parameters>   
```

### DIFF
--- a/docs/executor.md
+++ b/docs/executor.md
@@ -223,7 +223,7 @@ Resource requests and other job characteristics can be controlled via the follow
 - {ref}`process-memory`
 - {ref}`process-time`
 
-:::{note} As of Nextflow version 24.06.0-edge, HyperQueue version 0.17.0 or later is required.
+:::{note} HyperQueue version 0.20.0 or later is required.
 :::
 
 (k8s-executor)=
--- a/modules/nextflow/src/main/groovy/nextflow/executor/HyperQueueExecutor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/HyperQueueExecutor.groovy
@@ -59,7 +59,7 @@ class HyperQueueExecutor extends AbstractGridExecutor {
     protected List<String> getDirectives(TaskRun task, List<String> result) {
 
         result << '--name' << getJobNameFor(task)
-        result << '--log' << quote(task.workDir.resolve(TaskRun.CMD_LOG))
+        result << '--stream' << quote(task.workDir.resolve(TaskRun.CMD_LOG))
         result << '--cwd' << quote(task.workDir)
 
         // No enforcement, Hq just makes sure that the allocated value is below the limit
--- a/modules/nextflow/src/test/groovy/nextflow/executor/HyperQueueExecutorTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/executor/HyperQueueExecutorTest.groovy
@@ -117,7 +117,7 @@ class HyperQueueExecutorTest extends Specification {
         then:
         executor.getHeaders(task) == '''
             #HQ --name nf-task-1
-            #HQ --log /work/dir/.command.log
+            #HQ --stream /work/dir/.command.log
             #HQ --cwd /work/dir
             '''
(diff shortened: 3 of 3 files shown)


pditommaso (changes requested):
Make sure to update the corresponding test. Thanks! 

https://github.com/nextflow-io/nextflow/blob/0a29236e7f070378bc489f149fe3cd4ffd566e31/modules/nextflow/src/test/groovy/nextflow/executor/HyperQueueExecutorTest.groovy#L33-L33

pditommaso:
Please consider signing your commit as [explained here](https://github.com/nextflow-io/nextflow/blob/master/CONTRIBUTING.md#code-changes) to make pass the CDO validation bot.

pditommaso:
Also this should be updated accordingly 

https://github.com/nextflow-io/nextflow/blob/b5e8c4645b3eeb71dceb2106333c18eca59f13ed/docs/executor.md#L226-L226

chaetognatha:
> Also, if I'm understanding correctly this is a breaking change introduced by [HyperQueue 0.20.0](https://github.com/It4innovations/hyperqueue/releases/tag/v0.20.0)

this is correct!

pditommaso (approved):

christopher-hakkaart (approved):

chaetognatha:
thanks for the guidance @pditommaso :)

pditommaso:
You are welcome

yqshao:
Hi, I hit #5558 too, but the fix here doesn't seem to work for me (tested with hq 0.20 and nf 25.02.1 edge) -- jobs will fail with:

```
# hq job info $hq_job_id
Error: Error: Stream path /xxx is not a directory. 
```

I don't think `--stream` is meant as a drop-in replacement of `--log` (it expects an existing directory rather than a file, and seems to be intended for job arrays). Minimal examples of failure with `hq`:

```
touch log  && hq submit --stream $PWD/log -- echo 1 # failes existing file
rm -f log  && hq submit --stream $PWD/log -- echo 1 # failed with non-existing file
```

I believe we need `--stdout` instead:

https://it4innovations.github.io/hyperqueue/stable/jobs/streaming/#partial-redirection

spirali:
> I don't think `--stream` is meant as a drop-in replacement of `--log` (it expects an existing directory rather than a file, and seems to be intended for job arrays). Minimal examples of failure with `hq`:
 
Actually, `--log` was also used mainly for array jobs. It was designed for scenarios when you have many tasks and you do not want to create many stdio/stderr files for each tasks. This is also usecase for `--stream`. I have renamed it to `--stream` when we have changed how it works, because `--log` is really misleading name. If you do not want use postprocessing of these aggregated streams (via `hq output-log`) then it does not make sense to use  `--stream` nor `--log`.

Disclaimer: I am author of the `--log/--stream` functionality in HQ. I know almost nothing about nextflow.

PR #5745, opened 2025-02-05
Document temporary strict syntax for workflow handlers

bentsherman:
Related to #5261

Document the syntax for workflow handlers that satisfies the language server, until we implement a more concise syntax (depends on #4613).

### DIFF
--- a/docs/vscode.md
+++ b/docs/vscode.md
@@ -464,6 +464,34 @@ process greet {
 }
 ```
 
+**Workflow onComplete/onError handlers**
+
+{ref}`Workflow handlers <workflow-handlers>` (i.e. `workflow.onComplete` and `workflow.onError`) can be defined in several different ways in a script, but are typically defined as top-level statements and without an equals sign:
+
+```nextflow
+workflow.onComplete {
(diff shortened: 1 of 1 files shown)


docs/vscode.md
```diff
@@ -464,6 +464,34 @@ process greet {
...
+The Nextflow language specification does not allow statements to be mixed with script declarations, so workflow handlers must be defined in the entry workflow:
+
+```nextflow
+workflow {
+    // ...
+
+    workflow.onComplete = {
+        println "Pipeline completed at: $workflow.complete"
+        println "Execution status: ${ workflow.success ? 'OK' : 'failed' }"
+    }
```
pditommaso, line 487: Bit ugly nested `workflow`. In principle `onComplete` could be considered a `workflow` attribute, as such, be declarable outside the main `workflow` (and even desirable for better separation of concerns)

bentsherman: This is a temporary fix until we provide a better syntax. People keep asking me about it so it needs to be documented

pditommaso: Fair enough

PR #5747, opened 2025-02-05
Document config profile resolution priority

### LINKED ISSUE #5306: Overriding `resourceLimits`
Hello!

I am running an nf-core pipeline with `--profile test,self_hosted_runner` which both declare different `process.resourceLimits`, and I expect `self_hosted_runner` to be the one with priority and overriding `test`, but it's not the case.

In the `nextflow.config` we include configs in this order:

1. include `nfcore_custom.config`
2. `nfcore_custom.config` defines the profile `self_hosted_runner` which includes `self_hosted_runner.config`
3. we define the profile `test` which includes `test.config`

Is the order in which we include each config the one that has priority? or how does `resourceLimits` overriding work?

bentsherman:
Close #5306

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -294,6 +294,8 @@ Multiple configuration profiles can be specified by separating the profile names
 ```bash
 nextflow run <your script> -profile standard,cloud
 ```
+
+Config profiles are applied in the order in which they were defined in the config files.
 :::
 
 :::{danger}


docs/config.md
```diff
@@ -294,6 +294,8 @@ Multiple configuration profiles can be specified by separating the profile names
 ```bash
 nextflow run <your script> -profile standard,cloud
 ```
+
+Config profiles are applied in the order in which they were defined in the config files.
```
christopher-hakkaart, line 298:
```suggestion
Config profiles are applied in the order in which they were defined in the config file, regardless of the order they are listed in the run command.
```
I think the point here could be stressed for clarity. Not a blocker if you disagree.

christopher-hakkaart (approved):
Added a suggestion. Not a blocker if you disagree.

PR #4339, opened 2023-09-23
Prevent cache invalidation for task directives

pditommaso:
This PR prevents the cache invalidation then a process directive is modified, as introduced by PR https://github.com/nextflow-io/nextflow/issues/2382, with the exception of the `ext.*` extension directives. 

More context here https://github.com/nextflow-io/nextflow/issues/4326#issuecomment-1731183716. 

Solves #4326.

PR #5749, opened 2025-02-05
Improve documentation of task hash

### LINKED ISSUE #5417: Task ext directive not always included in task hash computation
## Bug report 

Only the variables from the task ext directive that are explicitly accessed in a process are used to compute the task hash.
This is at odds with the way the docs are written as these imply that any changes to the task ext directive should cause a change in the computation of the task hash.

#### Expected behavior

Changes to task ext directive will cause changes to the computed task hash.

#### Actual behavior

Changes to task ext directive do not cause changes to the computed task hash when no variable from the task ext directive is directly accessed in the process.

Please find a reprex in the following repo: https://github.com/trev-f/nf-task-ext-directive-hash-computation

### Program output 

```
Starting Run 1
Hello
world!
Hello
world!

Starting Run 2
Hello
world!
Hello
world!

Task hash diff:
1c1
< [explicitTaskExtAccess] cache hash: 9c235f212efac9273fab8a12f679a5ab; mode: STANDARD; entries: [
---
> [explicitTaskExtAccess] cache hash: c883dcfe8824349d780b12ac5d7acc2a; mode: STANDARD; entries: [
```

[run_1.log](https://github.com/user-attachments/files/17464919/run_1.log)

[run_2.log](https://github.com/user-attachments/files/17464925/run_2.log)

### Environment 

* Nextflow version: 24.04.4 build 5917
* Java version: "17.0.5" 2022-10-18 LTS
* Operating system: Linux

### Additional context

I think that all variables within the task ext directive should be used for computing the task hash, not just those which are explicitly accessed in the process `script` block.
This would allow for finer control over resuming that doesn't necessarily rely on changing the behavior of the process.

bentsherman:
Close #5417

### DIFF
--- a/docs/cache-and-resume.md
+++ b/docs/cache-and-resume.md
@@ -22,10 +22,10 @@ The task hash is computed from the following metadata:
 - Task {ref}`environment modules <process-module>` (if applicable)
 - Task {ref}`Conda environment <process-conda>` (if applicable)
 - Task {ref}`Spack environment <process-spack>` and {ref}`CPU architecture <process-arch>` (if applicable)
-- Task {ref}`process-ext` directive (if applicable)
 - Task {ref}`inputs <process-input>`
 - Task {ref}`script <process-script>`
 - Any global variables referenced in the task script
+- Any task {ref}`process-ext` properties referenced in the task script
 - Any {ref}`bundled scripts <bundling-executables>` used in the task script
 - Whether the task is a {ref}`stub run <process-stub>`
 - Task attempt


christopher-hakkaart (approved):

docs/cache-and-resume.md
```diff
@@ -22,10 +22,10 @@ The task hash is computed from the following metadata:
 - Task {ref}`environment modules <process-module>` (if applicable)
 - Task {ref}`Conda environment <process-conda>` (if applicable)
 - Task {ref}`Spack environment <process-spack>` and {ref}`CPU architecture <process-arch>` (if applicable)
-- Task {ref}`process-ext` directive (if applicable)
 - Task {ref}`inputs <process-input>`
 - Task {ref}`script <process-script>`
 - Any global variables referenced in the task script
+- Any task {ref}`process-ext` properties referenced in the task script
```
pditommaso, line 28: If I remember correctly there were some exceptions

bentsherman: As far as I can tell it is just the ext variables that are referenced in the script: https://github.com/nextflow-io/nextflow/blob/1c2ef7b837903b6d331b4ba360e9e10f7cac2da6/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy#L2319-L2330

pditommaso: ok, the exception is all `.ext.*` 😄 https://github.com/nextflow-io/nextflow/pull/4339

PR #5753, opened 2025-02-05
Fix documentation of process `path` input

### LINKED ISSUE #5565: If strings are passed to inputs staged as `path`s, the received value does not get implicitly converted to a string and written to a file
## Bug report 

Summary: The nextflow documentation claims that if strings are passed to inputs staged as `path`s, the received value will get implicitly converted to a string and written to a file. It looks like this is actually not the case for `path`s (though is the case for `file`s) and while looking at the code, it looks like this is possibly intentional, so the documentation may need updating.

From https://www.nextflow.io/docs/latest/reference/process.html#inputs:
> path( identifier | stageName )
> Declare a file input. The received value can be any type, and it will be staged into the task directory. **If the received value is not a file or collection of files, it is implicitly converted to a string and written to a file.**

(emphasis added)

```
process test {
    input:
        path input
    script:
    """
        cat ${input}
    """
}

workflow  {
    test("some text")
}

```

### Program output 

```
$ ~/nextflow test.nf 

 N E X T F L O W   ~  version 24.10.1

Launching `test.nf` [chaotic_leakey] DSL2 - revision: 61c6a9794f

ERROR ~ Error executing process > 'test'

Caused by:
  Not a valid path value: 'some text'

```

### Environment 

* Nextflow version: 24.10.1
* Java version: 21.0.4
* Operating system: [macOS, Linux, etc] macOS
* Bash version: (use the command `$SHELL --version`) 5.2.37

bentsherman:
Close #5565 

Just fixing a lazily-copied paragraph

### DIFF
--- a/docs/reference/process.md
+++ b/docs/reference/process.md
@@ -73,7 +73,7 @@ Additionally, the [directive values](#directives) for the given task can be acce
 
 `path( identifier | stageName )`
 
-: Declare a file input. The received value can be any type, and it will be staged into the task directory. If the received value is not a file or collection of files, it is implicitly converted to a string and written to a file.
+: Declare a file input. The received value should be a file or collection of files, and it will be staged into the task directory.
 
 : The argument can be an identifier or string. If an identifier, the received value will be made available to the process body as a variable. If a string, the received value will be staged into the task directory under the given alias.
 
--- a/docs/reference/process.md
+++ b/docs/reference/process.md
@@ -73,7 +73,7 @@ Additionally, the [directive values](#directives) for the given task can be acce
 
 `path( identifier | stageName )`
 
-: Declare a file input. The received value should be a file or collection of files, and it will be staged into the task directory.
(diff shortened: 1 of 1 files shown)


docs/reference/process.md
```diff
@@ -73,7 +73,7 @@ Additionally, the [directive values](#directives) for the given task can be acce
 
 `path( identifier | stageName )`
 
-: Declare a file input. The received value can be any type, and it will be staged into the task directory. If the received value is not a file or collection of files, it is implicitly converted to a string and written to a file.
+: Declare a file input. The received value should be a file or collection of files, and it will be staged into the task directory.
```
christopher-hakkaart, line 76:
```suggestion
: Declare a file input. The received value should be a file or collection of files and will be staged into the task directory.
```

christopher-hakkaart (approved):
Minor suggestion. Not a blocker if you disagree or the meaning has changed.

PR #5758, opened 2025-02-06
Document dynamic directive syntax with named arguments

bentsherman:
Based on a Slack thread, where someone was wondering how to make a dynamic directive with named arguments.


docs/process.md
```diff
@@ -1293,13 +1293,26 @@ process foo {
 }
 ```
 
-In the above example the {ref}`process-memory` and execution {ref}`process-time` limits are defined dynamically. The first time the process is executed the `task.attempt` is set to `1`, thus it will request a two GB of memory and one hour of maximum execution time.
+In the above example the {ref}`process-memory` and execution {ref}`process-time` limits are defined dynamically. The first time the process is executed the `task.attempt` is set to `1`, thus it will request 2 GB of memory and 1 hour of walltime.
 
-If the task execution fail reporting an exit status in the range between 137 and 140, the task is re-submitted (otherwise terminates immediately). This time the value of `task.attempt` is `2`, thus increasing the amount of the memory to four GB and the time to 2 hours, and so on.
+If the task execution fails with an exit status between 137 and 140, the task is re-executed, otherwise the run is terminated immediately. The re-executed task will have `task.attempt` set to `2`, and will request 4 GB of memory and 2 hours of walltime, and so on.
```
christopher-hakkaart, line 1298:
```suggestion
If the task execution fails with an exit status between 137 and 140, the task is re-executed; otherwise, the run is terminated immediately. The re-executed task will have `task.attempt` set to `2`, and will request 4 GB of memory and 2 hours of walltime.
```

docs/process.md
```diff
@@ -1293,13 +1293,26 @@ process foo {
...
 
-The directive {ref}`process-maxretries` set the maximum number of time the same task can be re-executed.
+The {ref}`process-maxretries` directive sets the maximum number of times the same task can be re-executed.
+
+:::{tip}
+Directives with named arguments like `accelerator` and `disk` must use a more verbose syntax when they are dynamic. For example:
```
christopher-hakkaart, line 1303:
```suggestion
Directives with named arguments, such as `accelerator` and `disk`, must use a more verbose syntax when they are dynamic. For example:
```

docs/process.md
```diff
@@ -1317,8 +1330,8 @@ process foo {
     """
 }
 ```
-In the above example, the {ref}`process-memory` is set according to previous trace record metrics. In the first attempt, when no trace metrics are available, it is set to one GB. In the subsequent attempts, it doubles the previously allocated memory. See {ref}`trace-report` for more information about trace records.
 
+In the above example, the {ref}`process-memory` is set according to previous trace record metrics. In the first attempt, when no trace metrics are available, it is set to one GB. In the subsequent attempts, it doubles the previously allocated memory. See {ref}`trace-report` for more information about trace records.
```
christopher-hakkaart, line 1334:
```suggestion
In the above example, the {ref}`process-memory` is set according to previous trace record metrics. In the first attempt, when no trace metrics are available, it is set to 1 GB. In the subsequent attempts, it doubles the previously allocated memory. See {ref}`trace-report` for more information about trace records.
```

christopher-hakkaart (changes requested):
Minor suggestions.

christopher-hakkaart (approved):
Looks good to me!

PR #5759, opened 2025-02-06
Move foreign file staging after storeDir evaluation

### LINKED ISSUE #5468: storeDir does not skip staging of inputs

I am expecting storeDir to help skip processes that are already performed. However I am finding that it does not skip staging of the input files when they are remote. If the input file is very large, it will delay the downstream steps of the pipeline to download something that takes up space on the local disk and isn't necessarily used, because the task is skipped.

```
process A {
cache true
storeDir "storeDir"
input:
tuple val(meta), path(x)
output:
path(x)
script:
"""
echo hi
"""
}

workflow {
remotepath="https://data.broadinstitute.org/Trinity/CTAT_RESOURCE_LIB/__genome_libs_StarFv1.10/Mouse_GRCm39_M31_CTAT_lib_Nov092022.source.tar.gz"
A([[:],remotepath])
}

```

### Program output 

If I run for the first time it will download and perform the task, as expected. If I immediately re-run with `-resume` it will skip the process immediately and the pipeline will complete very quickly, also as expected. However, when i remove the folder  `work/stage-*` and then try to resume, I get this:
```
N E X T F L O W  ~  version 23.10.1
Launching `main.nf` [nauseous_archimedes] DSL2 - revision: cc0b1bbdb5
[skipped  ] process > A [100%] 1 of 1, stored: 1 ✔
Staging foreign file: https://data.broadinstitute.org/Trinity/CTAT_RESOURCE_LIB/__genome_libs_StarFv1.10/Mouse_GRCm39_M31_CTAT_lib_Nov092022.source.tar.gz
[skipping] Stored process > A
```
This file is only 1.9 Gb which is big enough to notice a delay but it's not very long in the grand scheme of things. But in our pipeline we have a 30 Gb input reference file which takes anywhere from 4-22 minutes to download depending on IO speeds. Is there any way to skip staging? Seems like an unnecessary step in this context.

### Environment 

* Nextflow version: 23.10.1 
* Java version: 11
* Operating system:  Linux
* Bash version: GNU bash, version 4.2.46(2)-release (x86_64-redhat-linux-gnu)

jorgee:
close #5468

This PR modifies the foreign file staging management in TaskProcessor.makeTaskContextStage2 to return the staging batch and do the staging copy just before the hash computation. It allow to skip executions in storeDir without requiring to download inputs.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -639,7 +639,7 @@ class TaskProcessor {
         // -- map the inputs to a map and use to delegate closure values interpolation
         final secondPass = [:]
         int count = makeTaskContextStage1(task, secondPass, values)
-        makeTaskContextStage2(task, secondPass, count)
+        final batch = makeTaskContextStage2(task, secondPass, count)
 
         // verify that `when` guard, when specified, is satisfied
         if( !checkWhenGuard(task) )
@@ -652,6 +652,9 @@ class TaskProcessor {
         //    if true skip the execution and return the stored data
         if( checkStoredOutput(task) )
             return
+
+        // -- download foreign files
+        session.filePorter.transfer(batch)
 
         def hash = createTaskHashKey(task)
         checkCachedOrLaunchTask(task, hash, resumable)
@@ -2139,7 +2142,7 @@ class TaskProcessor {
         return count
     }
 
-    final protected void makeTaskContextStage2( TaskRun task, Map secondPass, int count ) {
+    final protected FilePorter.Batch makeTaskContextStage2( TaskRun task, Map secondPass, int count ) {
 
         final ctx = task.context
         final allNames = new HashMap<String,Integer>()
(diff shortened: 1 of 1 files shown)


modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -639,7 +639,7 @@ class TaskProcessor {
         // -- map the inputs to a map and use to delegate closure values interpolation
         final secondPass = [:]
         int count = makeTaskContextStage1(task, secondPass, values)
-        makeTaskContextStage2(task, secondPass, count)
+        final batch = makeTaskContextStage2(task, secondPass, count)
```
bentsherman, line 642: Let's rename to `foreignFiles` here and in the producing function

jorgee: Done

bentsherman:
@jorgee were you able to confirm that this change fixes the reported issue? using the test case they gave?

jorgee:
@bentsherman I tested with a slightly different case where output was different from the input. I have noted the input is the same as output in the example they provided. In this case the fix is not working because when the stage dir is removed, the file in the storeDir is also removed, and nextflow downloads again because the file does not exist.

bentsherman (approved):

PR #5765, opened 2025-02-07
Fix false error in variable visitor

### LINKED ISSUE #804: Variables within closures throw error that they are already defined

Using some groovy code before the script generation string or in the exec block causes a bug when a variable is declared in a closure. The example below contains a workaround by introducing a new variable and assigning it with the value that shall be used.

### Program output 

~~~~
N E X T F L O W  ~  version 0.30.2
Launching `./main.nf` [silly_austin] - revision: 34c2c42f4d
ERROR ~ Variable `z` already defined in the process scope @ line 30, column 20.
           int a =  y[z]
                      ^

1 error
~~~~

~~~~

Channel.from(1,2,3,4).into{ch_1_1; ch_1_2}
Channel.from(1,1,1,1,1,1).into{ch_2_1; ch_2_2}

process works {

    input:
    val x from ch_1_1.collect()
    val y from ch_2_1.collect()

    script:
    x.each{
        int z = it
        println (z)
        int a =  y[z]
        println a
    }
    """echo done"""
}

process works_not {

    input:
    val x from ch_1_2.collect()
    val y from ch_2_2.collect()

    script:
    x.each{ z ->
        println (z)
        int a =  y[z]
        println a
    }
    """echo done"""

}
~~~~

### Environment 

* Nextflow version: 0.30.2
* Java version: 1.8.0_171
* Operating system: Linux

bentsherman:
Close #804 

Test case:

```groovy
process VariableTesting {
    input:
    val foo

    exec:
    if( foo[0] == "a" )
        log.info "Found an A"
    def bar = foo.find { it == "d" }
}

workflow {
    inputs = Channel.from(["a", "b", "c"], ["d", "e", "f"])

    inputs | VariableTesting
}
```

Currently it fails with:
```
Script compilation error
- file : /path/to/main.nf
- cause: Variable `foo` already defined in the process scope @ line 28, column 15.
       def bar = foo.find {
                 ^

1 error
```

This is because the variable visitor was treating the `foo` in `foo.find` as a re-declaration of `foo` rather than merely a reference to it

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/ast/VariableVisitor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/ast/VariableVisitor.groovy
@@ -72,11 +72,13 @@ class VariableVisitor extends ClassCodeVisitorSupport {
     void visitDeclarationExpression(DeclarationExpression expr) {
         declaration = true
         try {
-            super.visitDeclarationExpression(expr)
+            visit(expr.getLeftExpression())
         }
         finally {
             declaration = false
         }
+
+        visit(expr.getRightExpression())
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


pditommaso (changes requested):
Can a unit test be added to validate this change?

pditommaso (approved):
👍

PR #5776, opened 2025-02-10
Task array improve

pditommaso:
This PR tries to avoid the use of side-effect to change the state of `TaskRun.isChild` field. 

Overall logic: 
1. set the field only when `createTaskArray` is invoked 
2. when a task is retried reset the attribute in the `TaskRun.clone` method
3. skip the `TaskArrayCollector` when it's a task retry in any case

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -2185,7 +2185,7 @@ class TaskProcessor {
         session.filePorter.transfer(batch)
     }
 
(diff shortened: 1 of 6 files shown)


modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -2340,12 +2340,12 @@ class TaskProcessor {
 
         makeTaskContextStage3(task, hash, folder)
 
+        // when no collector is define OR it's a task retry submit directly for execution
```
pditommaso, line 2343:
```suggestion
        // when no collector is define OR it's a task retry, then submit directly for execution
```

jorgee (approved):

Issue #5720, opened 2025-01-28
Unable to read from Oracle S3 buckets


mcolpus:
### Expected behavior and actual behavior

With nextflow 23.04.3 I can make a channel from an S3 OCI bucket just fine. But using 23.10.4 or later (I've tested with 24.10.4 as well) it gives an error, seemingly because it tries to use an Amazon endpoint.

### Steps to reproduce the problem

main.nf :
```
workflow {
    Channel
        .fromPath("s3://my-bucket/clean_reads/*fastq.gz")
        .take(1)
        .view()
}
```

nextflow.config
```
aws {
    accessKey = "*******************"
    secretKey = "*****************************"
    region = "uk-london-1"
    client {
        endpoint = "https://************.compat.objectstorage.uk-london-1.oraclecloud.com"
        s3PathStyleAccess = true
    }
}
```

### Program output 

When it works:
```
NXF_VER=23.04.3 nextflow run main.nf 

Nextflow 24.10.4 is available - Please consider updating your version to it
N E X T F L O W  ~  version 23.04.3
Launching `main.nf` [drunk_curran] DSL2 - revision: b1377beb96
/my-bucket/clean_reads/SAMD00000572.clean_1.fastq.gz
```

when it fails:
```
NXF_VER=23.10.4 nextflow run main.nf

Nextflow 24.10.4 is available - Please consider updating your version to it
N E X T F L O W  ~  version 23.10.4
Launching `main.nf` [admiring_yalow] DSL2 - revision: b1377beb96
ERROR ~ Unable to execute HTTP request: s3.uk-london-1.amazonaws.com

 -- Check '.nextflow.log' file for details
```
which makes it look like it's ignoring the endpoint provided. 

### Environment 

* Nextflow version: works on 23.04.3, fails on later
* Java version: 
openjdk 17.0.10 2024-01-16
OpenJDK Runtime Environment Temurin-17.0.10+7 (build 17.0.10+7)
OpenJDK 64-Bit Server VM Temurin-17.0.10+7 (build 17.0.10+7, mixed mode, sharing)
* Operating system: ubuntu 22.04
* Bash version: (use the command `$SHELL --version`): GNU bash, version 5.1.16(1)-releas

bentsherman:
Possibly related to #4732

I suspect there is a regression from the AWS config refactor that happened between 23.04 and 23.10

bobturneruk:
[...]

PR #5779, opened 2025-02-11
set ForceGlobalBucketAccess to false if region or endpoint provided, to enable non amazon S3 buckets

mcolpus:
fixes https://github.com/nextflow-io/nextflow/issues/5720

I've only tested with an oracle S3 bucket, but I assume it should still work with amazon provided the user doesn't select an incorrect region or endpoint as it will no longer automatically update.


plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3FileSystemProvider.java
```diff
@@ -835,7 +835,7 @@ protected S3FileSystem createFileSystem(URI uri, AwsConfig awsConfig) {
 		ClientConfiguration clientConfig = createClientConfig(props);
 
 		final String bucketName = S3Path.bucketName(uri);
-		final boolean global = bucketName!=null;
+		final boolean global = bucketName!=null && awsConfig.getRegion() == null && props.getProperty("endpoint") == null;
```
pditommaso, line 838: I'm not able to deduce what's the rationale of this

mcolpus: I've explained in more detail in the issue: https://github.com/nextflow-io/nextflow/issues/5720

The variable `global` controls withForceGlobalBucketAccessEnabled in `getS3Client`. When it's set to true the S3 client will automatically override the endpoint provided in the config to an amazon endpoint. The result being that nextflow cannot read from oracle buckets.
(The purpose of withForceGlobalBucketAccessEnabled seems to be to correct for a user who has selected the wrong region or does not know their region).

The logic here is that if the user has provided an endpoint/region, than to not override this.

I made this PR mainly to show that my issue did have a simple fix, but I imagine a better solution might involving adding another nextflow.config parameter?

pditommaso: Understand that. Just thinking to prevent unexpected side-effect would would for you using 

```
final boolean global = bucketName!=null && !awsConfig.getS3Config().isCustomEndpoint();
```

pditommaso: Nice

PR #5787, opened 2025-02-13
CID blockstore

jorgee:
M1 Implementation of CID store for provenance

Changes:
- CID store is specified by workflow.data.store.location
- Workflow Hash is created based on the workflow and parameters description
- workflow, tasks and outputs metadata are stored in <cid.store.location>/.meta
- references to other cid metadata are cid://<workflow_hash|task_hash/output_target_path
- CID NIO Filesystem to access data based on CIS URLs
- nextflow cid command to log, show and get lineage from CID store metadata

Known Limitations:
- Outputs which are not published in absolutePaths or URLs which are not subfolders both the outputDir, we can not infer the relative output target path. They are not currently tracked in the CID store. We could create a hash for the parent directory of the URL or absolute path and use it as relative folder.


modules/nextflow/src/main/groovy/nextflow/cli/CmdLog.groovy
```diff
@@ -203,6 +203,7 @@ class CmdLog extends CmdBase implements CacheBase {
                     .head('STATUS')
                     .head('REVISION ID')
                     .head('SESSION ID')
+                    .head('CID HASH')
```
pditommaso, line 206: Le't call this `RUN CID`

modules/nextflow/src/main/groovy/nextflow/data/cid/model/Workflow.groovy
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package nextflow.data.cid.model
+
+import groovy.transform.Canonical
+import groovy.transform.CompileStatic
+
+
+/**
+ *
```
pditommaso, line 25: Add a short description please

modules/nextflow/src/main/groovy/nextflow/data/config/DataConfig.groovy
```diff
@@ -37,7 +37,7 @@ class DataConfig {
 
     static DataConfig create(Session session) {
         if( session ) {
-            return new DataConfig(session.config.navigate('workflow.data') as Map ?: Map.of())
+            return new DataConfig(session.config.navigate('cid') as Map ?: Map.of())
```
pditommaso, line 40: I'd prefer to keep `workflow.data` at least for now

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -254,6 +259,18 @@ class Session implements ISession {
...
+    boolean getCidEnabled() { cidEnabled }
+
+    private HashCode executionHash
+
+    private CidStore cidStore
+
+    CidStore getCidStore() { cidStore }
+
+    String getExecutionHash() { executionHash }
+
```
pditommaso, line 273: I wonder if most of those could not be isolated in the CidObserver or a dedicated class, ideally here there should be only a reference to the observer

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -393,6 +410,12 @@ class Session implements ISession {
         // -- file porter config
         this.filePorter = new FilePorter(this)
 
+        if (config.cid) {
+            this.cidEnabled = true
+            this.cidStore = new DefaultCidStore()
```
pditommaso, line 415: it's fine for now, but ideally the reference to this should be taken via the plugin factory

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
```diff
@@ -146,7 +146,19 @@ class PublishDir {
...
+                log.warn("CID store is enabled but 'publishDir' is set to an absolute path ($resolvedPath). Outputs in this path will not published in the CID store")
+                this.path = FileHelper.toCanonicalPath(resolved)
+            }
+            else{
+                this.path = session.outputDir.resolve(resolvedPath)
+            }
+        }
+        else {
+            this.path = FileHelper.toCanonicalPath(resolved)
+        }
```
pditommaso, line 161: Is this really needed? The CidObserver is already using the source path to determine the  origin of the file. 

https://github.com/nextflow-io/nextflow/pull/5787/files#diff-fe1fbb6658e26a22fa620333e674faabe3fbd0388f68cf78eda2dd50aea216beR167-R180

jorgee: Removed. Not needed in the last version

modules/nextflow/src/main/groovy/nextflow/data/config/DataConfig.groovy
jorgee: Changed

modules/nextflow/src/main/groovy/nextflow/data/cid/model/Workflow.groovy
jorgee: Added

modules/nextflow/src/main/groovy/nextflow/Session.groovy
jorgee: I have moved the execution hash to the observer, however I have kept the cidStore because I get the store from the session in the CmdCid class.

modules/nextflow/src/main/groovy/nextflow/Session.groovy
jorgee: Do you mean something similar to the ExecutorFactory? I could get from Plugins.manager the classes implementing the CidStore interface and add the `workflow.data.store.type` property to indicate the class.

bentsherman: I think he means to use the same pattern as `CacheStore` and `CacheFactory` to load a CID backend through the plugin system

jorgee: Ok, I have used the same for the CidStore

pditommaso:
Looks excellent. Let's merge and continue in the baseline branch

PR #5799, opened 2025-02-18
feat(AzBatchService): Include cpu-shares and memory in docker run command

adamrtalbot:
The Azure Batch task did not include the `--cpu-shares` or `--memory` options of the docker run command, meaning resource allocation was only controlled by slots on the node. This PR introduces the additional options to the container run command, including `--cpu-shares` and `--memory`. The behaviour will now be more similar to AWS Batch which includes a soft limit on CPUs and hard limit on memory.


pditommaso (changes requested):
Nice. I'd prefer readability in this code, seem the befits of string builder is negligible in this scenario.

KH-NN:
Hi,

Just fyi. This PR has implications on AZCOPY. If you set memory to e.g. 4GB azcopy will crash when transferring multiple large files (50GB+). For large files we found that a minimum of 16GB is required to copy the files to the VM.

Due to code in [link](https://github.com/nextflow-io/nextflow/blob/75f1bc52bcbb0fd6d9680cc5fd7eaf18efb5428b/plugins/nf-azure/src/test/nextflow/executor/BashWrapperBuilderWithAzTest.groovy#L84) it will then try to generate a folder that also crashs. It took a bit of time to find this bug :)

PR #5803, opened 2025-02-20
feat(AzBatchTaskHandler): Include other output files in final upload statement

adamrtalbot:
The Azure Batch executor would sometimes fail to upload output files to storage, leaving failed jobs with no indication why. This PR adds the native Azure Batch output file directive for more files which will upload most of the auxiliary files to make it easier to debug failed pipelines.

I'm not 100% sure this will work, but I also can't see any major downside to including these files?


plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -563,6 +563,11 @@ class AzBatchService implements Closeable {
         List<OutputFile> result = new ArrayList<>(20)
         result << destFile(TaskRun.CMD_EXIT, task.workDir, sas)
         result << destFile(TaskRun.CMD_LOG, task.workDir, sas)
+        result << destFile(TaskRun.CMD_OUTFILE, task.workDir, sas)
+        result << destFile(TaskRun.CMD_ERRFILE, task.workDir, sas)
+        result << destFile(TaskRun.CMD_STAGE, task.workDir, sas)
+        result << destFile(TaskRun.CMD_TRACE, task.workDir, sas)
+        result << destFile(TaskRun.CMD_ENV, task.workDir, sas)
```
adamrtalbot, line 570: We might want to remove this one for security reasons.

adamrtalbot, line 570:
```suggestion
        result << destFile(TaskRun.CMD_ENV, task.workDir, sas)
```
We might want to remove this one for security reasons.

bentsherman: This file is used only to collect `env` and `eval` outputs, so there shouldn't be any security concern. Secrets are injected via `.command.run` or the cloud provider

adamrtalbot: We also use it to embed the temporary SAS: https://github.com/nextflow-io/nextflow/blob/azure_batch_upload_more_output_files/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzFileCopyStrategy.groovy#L57-L76

But this is short lived in general.

bentsherman: That code is added to the `.command.run`, not the `.command.env`

adamrtalbot: OK! Then nevermind!

PR #2099, opened 2021-05-07
Azure Batch executor: report error code and messages for failed Tasks.

manuelesimi:
Instead of reporting a generic error message, with these changes, Nextflow fetches the actual error code and message from the Batch API and returns them to the user.

For instance, this is the output if an invalid docker image is specified in the `container` option of a process:
<pre>
N E X T F L O W  ~  version 21.05.0-edge
Launching `main-azure.nf` [berserk_euler] - revision: 2086f6cebe
executor >  azurebatch (1)
[e9/06af2f] process > sayHelloInItalian (1) [  0%] 0 of 1
[-        ] process > sayHelloInFrench      -
[-        ] process > sayHelloInSpanish     -
[-        ] process > sayHelloInEnglish     -
[AZURE BATCH] Task failed. ERROR code: ContainerInvalidImage; message: One or more container images specified are invalid

Error executing process > 'sayHelloInItalian (1)'

Caused by:
  One or more container images specified are invalid

</pre>

Any other runtime error occurring in the task and detected by the Batch API would be reported as well.

PR #5806, opened 2025-02-21
fix (5794): Handle Azure error when using Fusion

adamrtalbot:
The error strategy was ignored when using Fusion on Azure, this PR fixes it by handling the error after reading the exitCode

I don't know if this is the right way to do it...but it works?


bentsherman:
Here's how aws batch handles it: https://github.com/nextflow-io/nextflow/blob/c3f67db2619f302589c5379a7851174bc7f76a54/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy#L257-L285

pditommaso:
Doesn't look an oversight, was made on purpose by https://github.com/nextflow-io/nextflow/pull/2099

jorgee:
Following what @alberto-miranda suggested, the reason why Azure Batch Tasks are different when enabling fusion is because of the cmd command set in this [part of the code](https://github.com/nextflow-io/nextflow/blob/d29286078cd01ade15874cb120db370d93c5a9ed/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L500C77-L500C82)

- For fusion: `/usr/bin/fusion bash /fusion/az/xxxx/.command.run`

- For non-fusion:  `sh -c 'bash .command.run 2>&1 | tee .command.log'`

The `| tee .command.log` is preventing Azure Batch to get the real exit code. I have reproduced what is happening with a small script that is returning an exit code 42.

```
$ sh -c 'bash test.sh'
$ echo $?
42
$ sh -c 'bash test.sh | tee -a .command.log'
$ echo $?
0
```
So, without fusion the Azure Batch Task exit code is always 0 without fusion, but the failure is detected later in the code checking the Nextflow task exit code. 

To make both executions equivalent, I think we should make the non-fusion cmd to get the real exist code with something like:

`bash -o pipefail -c 'bash .command.run 2>&1 | tee .command.log'`

I think @adamrtalbot solutions are fine to avoid exit code failures are recoverable but I do not have clear is why the rest of failures are Unrecoverable by default.

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchTaskHandler.groovy
```diff
@@ -115,13 +116,17 @@ class AzBatchTaskHandler extends TaskHandler implements FusionAwareTask {
...
+            final info = batchService.getTask(taskKey).executionInfo
+            task.exitStatus = info?.exitCode ?: readExitFile()
             task.stdout = outputFile
             task.stderr = errorFile
             status = TaskStatus.COMPLETED
-            final info = batchService.getTask(taskKey).executionInfo
-            if (info.result == BatchTaskExecutionResult.FAILURE)
-                task.error = new ProcessUnrecoverableException(info.failureInfo.message)
+            if (info.result == BatchTaskExecutionResult.FAILURE ) {
+                if (task.exitStatus == 0 || task.exitStatus == Integer.MAX_VALUE) {
```
pditommaso, line 125: The MAX_VALUE condition is managed here 

https://github.com/nextflow-io/nextflow/blob/076c8237afa51b219d140ccdcba121fd6e5f0e28/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy#L2385-L2385

jorgee: I didn't see it. Updated

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchTaskHandler.groovy
```diff
@@ -120,8 +121,15 @@ class AzBatchTaskHandler extends TaskHandler implements FusionAwareTask {
             task.stderr = errorFile
             status = TaskStatus.COMPLETED
             final info = batchService.getTask(taskKey).executionInfo
-            if (info.result == BatchTaskExecutionResult.FAILURE)
-                task.error = new ProcessUnrecoverableException(info.failureInfo.message)
+            if (info.result == BatchTaskExecutionResult.FAILURE) {
+                if (task.exitStatus != 0) {
+                    // If the exit status is not 0, throw a process failed exception and Nextflow will handle it with errorStrategy
+                    task.error = new ProcessFailedException("Task failed with exit code ${task.exitStatus}".toString())
```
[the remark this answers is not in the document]
jorgee: Updated

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchTaskHandler.groovy
```diff
@@ -115,13 +116,14 @@ class AzBatchTaskHandler extends TaskHandler implements FusionAwareTask {
...
             // finalize the task
-            task.exitStatus = readExitFile()
+            final info = batchService.getTask(taskKey).executionInfo
+            task.exitStatus = info?.exitCode ?: readExitFile()
             task.stdout = outputFile
             task.stderr = errorFile
             status = TaskStatus.COMPLETED
-            final info = batchService.getTask(taskKey).executionInfo
-            if (info.result == BatchTaskExecutionResult.FAILURE)
+            if (info.result == BatchTaskExecutionResult.FAILURE && task.exitStatus == 0) {
```
pditommaso, line 124: This is really counter intuitive, let's add a comment why throw the exception only when exitStatus == 0

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchTaskHandler.groovy
jorgee: I have reopened it because I have noted that if we do not manage the MAX_VALUE case here we will loose the error message provided by Azure. @pditommaso, if you agree I will revert the change

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchTaskHandler.groovy
jorgee: Added in reverted commit

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchTaskHandler.groovy
```diff
@@ -115,13 +116,17 @@ class AzBatchTaskHandler extends TaskHandler implements FusionAwareTask {
...
             task.stderr = errorFile
             status = TaskStatus.COMPLETED
-            final info = batchService.getTask(taskKey).executionInfo
-            if (info.result == BatchTaskExecutionResult.FAILURE)
-                task.error = new ProcessUnrecoverableException(info.failureInfo.message)
+            if( info.result == BatchTaskExecutionResult.FAILURE ){
+                if( task.exitStatus == 0 || task.exitStatus == Integer.MAX_VALUE) {
+                    // Consider unrecoverable failure when Task failure and exit code is 0 or does not exist (MAX_VALUE)
+                    task.error = new ProcessUnrecoverableException(info.failureInfo.message)
+                }
```
pditommaso, line 128: As a developer still do not understand why is thrown `ProcessUnrecoverableException` only when exitStatus == 0 or exitStatus == Integer.MAX_VALUE

jorgee: Do you mean the condition to send the exception or the exception type? 

The condition is the case when the exitcode is not enought to know why the task has failed and we have to get the message from Azure. I will update the comment to clarify it. 

Regarding the exception type, I have kept the type that was before, but I don't know why all failed jobs were considered `ProcessUrecoverableException`. [In my opinion](https://github.com/nextflow-io/nextflow/pull/5806#issuecomment-2690938730), it should be just `ProcessFailedException`, but I didn't change it to avoid possible unknown side effects.

pditommaso: Made a few tests: when a task fail with a non-zero exit status, the azure task info report `success`, not `failure`. 

I believe `failure` is returned when the task fails for some infra related problem e.g.  container, queue, etc. That should be the rationale of `ProcessUnrecoverableException`. 

That exception was added in this PR https://github.com/nextflow-io/nextflow/pull/2099. 

I'm still not able to understand what condition are you trying to handle with `task.exitStatus == 0 || task.exitStatus == Integer.MAX_VALUE`

jorgee: I am not getting the same results as in your tests. I have just run the following pipeline to generate a task with non-zero exit value.

```
workflow {
main:
    EXITCODE_TEST()
}

process EXITCODE_TEST {
    errorStrategy 'retry'
    """
    exit 42
    """
}
```
I have also added a print to check the received info results.

```
            status = TaskStatus.COMPLETED
            log.info("**** Task result is " + info.result)
            if( info.result == BatchTaskExecutionResult.FAILURE ){
                if( task.exitStatus == 0 || task.exitStatus == Integer.MAX_VALUE) {
                    // Consider unrecoverable failure when Task failure and exit code is 0 or does not exist (MAX_VALUE)
                    task.error = new ProcessUnrecoverableException(info.failureInfo.message)
                }
            }
```

This is what I get in when running the task

```
 N E X T F L O W   ~  version 25.01.0-edge

Launching `main.nf` [tiny_plateau] DSL2 - revision: 4b5489abc9

executor >  azurebatch (2)
[84/d84d88] process > EXITCODE_TEST [100%] 2 of 2, failed: 2, retries: 1 ✘
**** Task result is failure
**** Task result is failure
[f8/2ac2b2] NOTE: Process `EXITCODE_TEST` terminated with an error exit status (42) -- Execution is retried (1)
ERROR ~ Error executing process > 'EXITCODE_TEST'

Caused by:
  Process `EXITCODE_TEST` terminated with an error exit status (42)


Command executed:

  exit 42

Command exit status:
  42

Command output:
  (empty)

Work dir:
  az://test-nf-jorgee/work/84/d84d888c33248d815535955c3b81d1

Container:
  ubuntu

```
And this is the Azure batch task

image

jorgee: > I'm still not able to understand what condition are you trying to handle with `task.exitStatus == 0 || task.exitStatus == Integer.MAX_VALUE`

The condition is just to remove the failure cases that due to a non-zero exit codes. Before the PR, the non-fusion execution were always returning 0 as exit code because of the pipe in the execution command mentioned [above](https://github.com/nextflow-io/nextflow/pull/5806#issuecomment-2690938730). So, if there was a failure it was considered Unrecoverable. Now, that we changed the command a azure batch failure can happen if exit code is not zero (see my previous comment)

jorgee:
> Perhaps we should check the [ErrorCategory](https://learn.microsoft.com/en-us/java/api/com.azure.compute.batch.models.errorcategory?view=azure-java-preview) in the [BatchTaskFailureInfo](https://learn.microsoft.com/en-us/java/api/com.azure.compute.batch.models.batchtaskfailureinfo?view=azure-java-preview), with `USER_ERROR` = retry failure and `SERVER_ERROR` = unrecoverable.
Yes, I saw it but I looking at the possible errors [here](https://learn.microsoft.com/en-us/rest/api/batchservice/batch-status-and-error-codes?view=rest-batchservice-2024-07-01#BKMK_JobTaskError), I couldn't see a that all server errors are unrecoverable and user errors are recoverable. For instance, this one "BlobDownloadTimedOut (ServerError) Timeout encountered while downloading one of the specified Azure Blob(s)" looks like a retryable error. At the end, it was not the purpose of the PR.

pditommaso:
Then I think the condition should be 

```
            // finalize the task
            final info = batchService.getTask(taskKey).executionInfo
            task.exitStatus = info?.exitCode ?: readExitFile()
            task.stdout = outputFile
            task.stderr = errorFile
            status = TaskStatus.COMPLETED
            if( task.exitStatus == Integer.MAX_VALUE && info.failureInfo.message ){
                task.error = new ProcessException(info.failureInfo.message)
            }
```


Why? because if there's no exit status is not defined, likely it not even started and the error message is retrieved by via Azure API

pditommaso:
it could be interesting also `info.containerInfo.error` for doing something similar to aws batch https://github.com/nextflow-io/nextflow/blob/b6dbc1e718ffa014e76b3c4d9fcb416be217e9f9/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy#L274-L274

jorgee:
> Then I think the condition should be
> 
> ```
>             // finalize the task
>             final info = batchService.getTask(taskKey).executionInfo
>             task.exitStatus = info?.exitCode ?: readExitFile()
>             task.stdout = outputFile
>             task.stderr = errorFile
>             status = TaskStatus.COMPLETED
>             if( task.exitStatus == Integer.MAX_VALUE && info.failureInfo.message ){
>                 task.error = new ProcessException(info.failureInfo.message)
>             }
> ```
> 
> Why? because if there's no exit status is not defined, likely it not even started and the error message is retrieved by via Azure API

I think it should work. but we are ignoring any error that could happen after the nextflow task execution. In the Azure case, it is uploading the .exitcode .command.log. Exit code is received with the API, but we could loose .command.log.

pditommaso:
status from `.exitcode` should be included by this 

```
task.exitStatus = info?.exitCode ?: readExitFile()
``` 

if `info.exitCode == 0` (groovy truth is false) then fallback to  `readExitFile()`

jorgee:
I have checked the lastest suggestion from @pditommaso and it also works. I have pushed the changes.

PR #5807, opened 2025-02-21
Add known_hash file() option

edmundmiller:
## Overview
This PR adds a new file hash verification feature to Nextflow's `file()` operator. This feature allows users to verify file integrity by checking file hashes against expected values. This feature is particularly useful when working with downloaded files or when ensuring data integrity is critical.

## Implementation Details
- Added `FileHashVerifier` class that supports multiple hash algorithms:
  - MD5
  - SHA-1
  - SHA-256
  - SHA-384
  - SHA-512
- Extended the `file()` operator to accept a `known_hash` parameter in the format `algorithm:hash`
- Added comprehensive test coverage in `FileHashVerifierTest`
- Updated documentation in `working-with-files.md`

## Usage Example

```nextflow
// Verify a file with MD5
file('data.txt', known_hash: 'md5:d41d8cd98f00b204e9800998ecf8427e')
// Verify a file with SHA-256
file('data.txt', known_hash: 'sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855')
```

## Limitations
- Hash verification is not supported when using glob patterns that match multiple files
- Throws an `IllegalArgumentException` if:
  - Hash format is invalid
  - Hash algorithm is not supported
  - File hash doesn't match the expected value
  
  ## Inspiration
This feature was inspired by the [Pooch library](https://www.fatiando.org/pooch/latest/), which provides similar functionality for Python.

### DIFF
--- a/docs/working-with-files.md
+++ b/docs/working-with-files.md
@@ -12,6 +12,30 @@ myFile = file('some/path/to/my_file.file')
 
 The `file()` method can reference both files and directories, depending on what the string path refers to in the file system.
 
+You can also verify file integrity by providing a hash value:
+
+```nextflow
+// Verify a file with MD5
+file('some/path/to/my_file.file', known_hash: 'md5:d41d8cd98f00b204e9800998ecf8427e')
+
+// Verify a file with SHA-256
+file('some/path/to/my_file.file', known_hash: 'sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855')
+```
+
+The following hash algorithms are supported:
+
+-   MD5 (`md5:`)
+-   SHA-1 (`sha1:`)
+-   SHA-256 (`sha256:`)
+-   SHA-384 (`sha384:`)
+-   SHA-512 (`sha512:`)
+
+If the file's hash doesn't match the expected value, an error will be thrown.
+
+:::{note}
+Hash verification is not supported when using glob patterns that match multiple files.
+:::
+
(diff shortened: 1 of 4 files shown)


bentsherman:
I think Jordi took a crack at this a few years ago: #4415

I guess now we have two approaches we can compare 😅

docs/working-with-files.md
```diff
@@ -12,6 +12,30 @@ myFile = file('some/path/to/my_file.file')
 
 The `file()` method can reference both files and directories, depending on what the string path refers to in the file system.
 
+You can also verify file integrity by providing a hash value:
+
+```nextflow
+// Verify a file with MD5
+file('some/path/to/my_file.file', known_hash: 'md5:d41d8cd98f00b204e9800998ecf8427e')
```
pditommaso, line 19:
```suggestion
file('some/path/to/my_file.file', checksum: 'md5:d41d8cd98f00b204e9800998ecf8427e')
```

`checksum` can be preferable

bentsherman:
I think I like Jordi's approach better. The problem with this PR is that it verifies the checksum when you call `file()`, but `file()` should only give you a reference, it should not trigger a download.

@edmundmiller would you be willing to test Jordi's PR and build on it instead? If you can confirm that it works then that will make it easier to merge

pditommaso:
Does not this and https://github.com/nextflow-io/nextflow/pull/4415 overlaps with the CID effort? I'm inclined to close both in favour of the latter.

bentsherman:
The use case for these PRs is downloading an HTTP/FTP file and verifying its checksum. The user could alternatively "upload" that file to their CID store, access it using the `cid://` path, and get the same checksum guarantees. So they are complementary, but I would say the need for a `checksum` option in `file()` becomes less important if we have the CID store.

pditommaso:
Let's reconsider once after the CID feature has been rolled-out.

PR #5816, opened 2025-02-25
Config schema checking

bentsherman:
Redo of #4201 

This PR uses the config scope classes (in the new compiler module) to validate the config at runtime, for the `config` and `run` commands.

Validation occurs after plugins are loaded, and the validator uses the plugin system to collect all implementations of the `ConfigScope` interface, so plugins can use this interface to declare their config options to Nextflow.

I suggest we merge this in an upcoming release and advise plugin developers to implement this interface before 25.04, to make the transition to next stable as smooth as possible.

This work is also a prerequisite to recognizing third-party plugin config in VS Code.


docs/developer/plugins.md
```diff
@@ -108,16 +108,49 @@ You can then set this option in your config file:
...
+        return '''
+            The `myplugin` scope...
+            '''
+    }
+
+    @ConfigOption('''
+        Message to print to standard output when a run is initialized.
+        ''')
+    String createMessage
 }
```
bentsherman, line 149: Here is an example of how to declare config options for the schema checker. I want to make sure we're solid on this interface before we release it to plugin developers.

I was thinking it might be better to declare the scope name/description with annotations (e.g. `@Name` and `@Description`) instead of methods. That way the schema checker doesn't need to create an instance of each class, and then you don't need the no-arg constructor.

Similarly for config options, maybe better to provide the description in a separate `@Description` annotation...

modules/nextflow/src/main/groovy/nextflow/config/ConfigValidator.groovy
```diff
@@ -0,0 +1,92 @@
...
+@CompileStatic
+class ConfigValidator {
+
+    void validate(ConfigMap config) {
+        validate(config.toConfigObject())
+    }
+
+    void validate(ConfigObject config) {
+        final schema = getSchema()
+        final flatConfig = config.flatten()
```
bentsherman, line 39: This line is causing a stack overflow error with this [test case](https://github.com/nextflow-io/nextflow/blob/d539ccaccabbb651d1dbc09346a8bb64d9697f04/modules/nextflow/src/test/groovy/nextflow/cli/CmdConfigTest.groovy#L410-L449). Not sure why, but I'm guessing it's related to the nested `profiles` blocks

tom-seqera: It looks like only a single instance of `ConfigObject` is being created for both `profile` blocks, and then both the `test` and `debug` keys are being added to it. This results in an infinite loop while traversing.

bentsherman:
Looking deeper into the failing unit test:
```groovy
params.foo = 'baz'

profiles {
    test {
        params.foo = 'foo'
        profiles {
            debug {
                cleanup = false
            }
        }
    }
}
```

the test expects the following resolved config:
```groovy
params.foo = 'baz'

profiles {
   test {
      params.foo = 'foo'
   }
   debug {
      cleanup = false
   }
}
```

in other words, the nested `profiles` block is actually merged with the top `profiles` block, producing two profiles `test` and `debug`.

However, while it works with `nextflow config`, it doesn't work with `nextflow run`:
```console
$ nextflow run hello -profile debug

 N E X T F L O W   ~  version 25.01.0-edge

Unknown configuration profile: 'debug'
```

I think this is an artifact of the `ConfigSlurper` on which the v1 config parser was based. It was never a documented feature, and the v2 config parser doesn't support it. I think it just makes configs harder to interpret anyway, if profiles can modify other profiles.

The v2 config parser doesn't support nested profiles, so it just warns about a spurious config option:
```console
$ NXF_SYNTAX_PARSER=v2 ./build/releases/nextflow-25.01.0-edge-dist config -a 
WARN: Unrecognized config option 'profiles.debug.cleanup'
params.foo = 'baz'

profiles {
   test {
      params.foo = 'foo'
      profiles {
         debug {
            cleanup = false
         }
      }
   }
}
```

tests/checks/profiles.nf/.expected_profile_advanced.txt
```diff
@@ -1,5 +1,4 @@
-echo = true
-foo = 'bar'
+outputDir = 'results'
```
bentsherman, line 1: Had to pick some valid config option here. Alternatively we could leave the invalid options and include the warnings for them in the expected stdout

christopher-hakkaart:
The documentation looks good.

docs/developer/plugins.md
```diff
@@ -108,16 +108,43 @@ You can then set this option in your config file:
...
 // dot syntax
-myplugin.create.message = "I'm alive!"
+myplugin.createMessage = "I'm alive!"
 
-// closure syntax
+// block syntax
 myplugin {
-    create {
-        message = "I'm alive!"
+    createMessage = "I'm alive!"
```
pditommaso, line 115: What about structured options e.g. `wave.build.repository` ?

bentsherman: You can see how I handled that [here](https://github.com/nextflow-io/language-server/blob/main/modules/compiler/src/main/java/config/scopes/WaveConfig.java)

pditommaso: Ok, so it can handle nested objects

modules/nextflow/src/main/groovy/nextflow/config/ConfigParserFactory.groovy
```diff
@@ -33,7 +33,7 @@ import nextflow.config.parser.v2.ConfigParserV2
 class ConfigParserFactory {
 
     static ConfigParser create() {
-        final parser = SysEnv.get('NXF_SYNTAX_PARSER', 'v1')
+        final parser = Const.getSyntaxParser()
```
pditommaso, line 36: The use of `Const` for helper methods is mistake I've made at the beginning. Think it would better to no propagate it. `NF` could be a better place for this

pditommaso (changes requested):

modules/nextflow/src/main/groovy/nextflow/NF.groovy
```diff
@@ -32,6 +32,10 @@ class NF {
         return (Session)Global.session
     }
 
+    static String getSyntaxParser() {
```
pditommaso, line 35: Using `getSyntaxParserVer` would not be a bit more explicit? otherwise it seems it return the parser itself

modules/nextflow/src/main/groovy/nextflow/cli/CmdConfig.groovy
```diff
@@ -113,6 +116,10 @@ class CmdConfig extends CmdBase {
 
         final config = builder.buildConfigObject()
 
+        // -- validate config options
+        new ConfigValidator().validate(config)
```
pditommaso, line 120: Nit, what about adding a shortcut static `validate` (likely it would require to rename current `validate` to something else e.g. `validateConfig`

bentsherman: meh. it's not really needed

pditommaso (approved):

PR #5819, opened 2025-02-26
ci: Improve `fusion-symlink.nf` test

alberto-miranda:
## Description
The `fusion-symlink.nf` fails randomly due to what looks like a race condition. Since it was impossible to reproduce the problem in the a controlled environment, we are slightly modifying the test in an attempt to better diagnose the problem. To this end, this PR
extends the `fusion-symlink.nf` in two ways:
1. We add a retry protocol with an exponential backoff to the `nextflow fs cp [...]` commands used for validating the published data. The idea is that if there's indeed a race condition between the publishing and the validation, retrying the validation should eventually succeed. We ensure the test still fails so that we may detect when the potential race condition occurs.
2. If the error is not a publish/validation race condition, we still want to capture as much information from the running `fusion` instances. To that end we enable fusion's extended telemetry by setting the `FUSION_TRACING_DESTINATION` environment variable.

### DIFF
--- a/tests/checks/fusion-symlink.nf/.checks
+++ b/tests/checks/fusion-symlink.nf/.checks
@@ -1,5 +1,43 @@
 #!/bin/bash
 
+# Retry a command with exponential backoff. The function returns 0 if the command was successful
+# on its first attempt, 1 if the command failed after all attempts, and 2 if the command was successful
+# after one or more retries.
+function _retry {
+
+  if [[ $# -lt 4 ]]; then
+    echo "Usage: _retry <max_attempts> <initial_delay> <max_delay> <cmd>"
+    return 1
+  fi
+
+  local max_attempts="$1"; shift
+  local initial_delay="$1"; shift
+  local max_delay="$1"; shift
+  local cmd=( "$@" )
+  local attempt_num=1
+  local max_attempts=${max_attempts}
+  local max_delay=${max_delay}
+  local initial_delay=${initial_delay}
+  local exit_code=0
+
+  until "${cmd[@]}"; do
+    exit_code=2
+    if (( attempt_num == max_attempts )); then
+      echo "-- [$attempt_num/$max_attempts] attempt failed! No more attempts left."
+      return 1
+    fi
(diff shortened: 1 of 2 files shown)


alberto-miranda:
Could someone double check if there's a `.fusion/traces/` directory in any of the workdirs for the `fusion-symlink.nf` test? I would like to make sure the setting is working before we merge this, just in case...

pditommaso:
Where are supposed those files? 

```
»  aws s3 ls s3://nextflow-ci/work/53/7f3a251f902c946e3ad794a5642456/
                           PRE .fusion/
2025-02-25 18:39:13          0 
2025-02-25 18:39:25          0 .command.begin
2025-02-25 18:39:25          0 .command.err
2025-02-25 18:39:25          0 .command.log
2025-02-25 18:39:25          0 .command.out
2025-02-25 18:39:14       4156 .command.run
2025-02-25 18:39:14         38 .command.sh
2025-02-25 18:39:25          1 .exitcode
2025-02-25 18:39:26       9551 .fusion.log
2025-02-25 18:39:25          6 data.txt
```

alberto-miranda:
> Where are supposed those files?
> 
> ```
> »  aws s3 ls s3://nextflow-ci/work/53/7f3a251f902c946e3ad794a5642456/
>                            PRE .fusion/
> 2025-02-25 18:39:13          0 
> 2025-02-25 18:39:25          0 .command.begin
> 2025-02-25 18:39:25          0 .command.err
> 2025-02-25 18:39:25          0 .command.log
> 2025-02-25 18:39:25          0 .command.out
> 2025-02-25 18:39:14       4156 .command.run
> 2025-02-25 18:39:14         38 .command.sh
> 2025-02-25 18:39:25          1 .exitcode
> 2025-02-25 18:39:26       9551 .fusion.log
> 2025-02-25 18:39:25          6 data.txt
> ```

If I configured correctly, they should be under the `.fusion` directory shown in your listing. Can you try with:

```
aws s3 ls s3://nextflow-ci/work/53/7f3a251f902c946e3ad794a5642456/.fusion/traces/
```
Thanks!

PR #5830, opened 2025-02-28
Use parallelization in Gradle build

bentsherman:
Why weren't we doing this before 😆


pditommaso:
Good you are learning more Gradle 😄

PR #5833, opened 2025-03-01
Decouple cid store from session

pditommaso:
This PR double the CID store from the Session object. Tests needs to be fixed.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/data/cid/CidStoreFactory.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/data/cid/CidStoreFactory.groovy
@@ -51,7 +51,7 @@ abstract class CidStoreFactory implements ExtensionPoint {
     static CidStore getOrCreate(Session session) {
         if( instance || initialized )
             return instance
-        synchronized (session) {
+        synchronized (CidStoreFactory.class) {
             if( instance || initialized )
                 return instance
             initialized = true
@@ -59,6 +59,13 @@ abstract class CidStoreFactory implements ExtensionPoint {
             if( !config.enabled )
                 return null
             return instance = create(config)
+        }
+    }
+
+    static void clean(){
+        synchronized (CidStoreFactory.class) {
+            instance = null
+            initialized = false
         }
     }
 
--- a/modules/nextflow/src/main/groovy/nextflow/data/cid/CidStoreFactory.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/data/cid/CidStoreFactory.groovy
@@ -19,6 +19,7 @@ package nextflow.data.cid
 
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
+import nextflow.Session
 import nextflow.data.config.DataConfig
 import nextflow.plugin.Plugins
 import org.pf4j.ExtensionPoint
@@ -32,6 +33,10 @@ import org.pf4j.ExtensionPoint
 @CompileStatic
 abstract class CidStoreFactory implements ExtensionPoint {
 
+    private static CidStore instance
+
+    private static boolean initialized
+
     protected abstract CidStore newInstance(DataConfig config)
 
     static CidStore create(DataConfig config){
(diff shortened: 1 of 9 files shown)


modules/nextflow/src/main/groovy/nextflow/data/config/DataConfig.groovy
```diff
@@ -31,8 +31,11 @@ class DataConfig {
 
     final DataStoreOpts store
 
+    final boolean enabled
+
     DataConfig(Map opts) {
         this.store = new DataStoreOpts(opts.store as Map ?: Map.of())
+        this.enabled = opts.store != null
```
pditommaso, line 38: We may decided to an explicit logic to enable/disable it. Just a temporay flag

modules/nextflow/src/main/groovy/nextflow/data/cid/CidStoreFactory.groovy
```diff
@@ -32,6 +33,10 @@ import org.pf4j.ExtensionPoint
 @CompileStatic
 abstract class CidStoreFactory implements ExtensionPoint {
 
+    private static CidStore instance
+
+    private static boolean initialized
+
     protected abstract CidStore newInstance(DataConfig config)
 
     static CidStore create(DataConfig config){
```
jorgee, line 42:
```suggestion
    private static CidStore create(DataConfig config){
```

modules/nextflow/src/main/groovy/nextflow/data/cid/CidStoreFactory.groovy
```diff
@@ -43,4 +48,18 @@ abstract class CidStoreFactory implements ExtensionPoint {
...
+        synchronized (session) {
+            if( instance || initialized )
+                return instance
+            initialized = true
+            final config = DataConfig.create(session)
+            if( !config.enabled )
+                return null
+            return instance = create(config)
+        }
+    }
```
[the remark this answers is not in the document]
jorgee: Ok, about the double check. But is it not better to lock per class and not per session argument? It is assuming that session object is always the same. Not sure if there is a case that it is not true.

Forget the initialized comment, I've got it. It is also to avoid unnecessary creates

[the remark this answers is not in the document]
pditommaso: Not a strong opinion on this. In practical terms nextflow uses exactly one session, so does not matter. For testing purposes it may be slightly better. Feel free to update otherwise.

pditommaso:
@jorgee can you help solving the broken tests? 🙏

jorgee:
> @jorgee can you help solving the broken tests? 🙏

Yes, I am on it.

modules/nextflow/src/main/groovy/nextflow/data/cid/CidStoreFactory.groovy
```diff
@@ -62,4 +62,11 @@ abstract class CidStoreFactory implements ExtensionPoint {
         }
     }
 
+    static void clean(){
```
pditommaso, line 65: Let's mark it with `@TestOnly`, also `reset` may be a slightly better for consistency with other similar methods eg 

https://github.com/nextflow-io/nextflow/blob/138a1ff7e73089f516c44a11f7e72cf3a384ede7/modules/nextflow/src/main/groovy/nextflow/secret/SecretsLoader.groovy#L62-L64

jorgee: Changed

modules/nextflow/src/main/groovy/nextflow/data/config/DataConfig.groovy
jorgee: I have added the workflow.data.enabled flag to enable/disable it

jorgee:
I have solved broken tests, added the `workflow.data.enabled` flag and I have also managed the case where store is null in the `cid` command. I can be merged to cid-store from my side.

pditommaso:
Nice!

PR #5847, opened 2025-03-04
Add container platform to docker command

pditommaso:
This PR improves the handling of container platform in two ways: 

* Add explicitly the platform to be used to Docker and Podman commands. This is needed to solve the ambiguity on some OS (e.g. macOS) where a container can be executed both as ARM (using silicon) and AMD (via emulation) 
* Default to `linux/amd64` when no `process.arch` is specified for consistency with Wave defaults.

### DIFF
--- a/docs/process.md
+++ b/docs/process.md
@@ -473,27 +473,11 @@ workflow {
 }
 ```
 
-:::{versionadded} 23.09.0-edge
-:::
-
-By default, `path` inputs will accept any number of files and stage them accordingly. The `arity` option can be used to enforce the expected number of files, either as a number or a range.
-
-For example:
-
-```nextflow
-input:
-    path('one.txt', arity: '1')         // exactly one file is expected
-    path('pair_*.txt', arity: '2')      // exactly two files are expected
-    path('many_*.txt', arity: '1..*')   // one or more files are expected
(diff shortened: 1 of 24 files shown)


pditommaso:
@bentsherman what do you think about this? any side-effects that was not considered?

bentsherman (changes requested):
The changes make sense. Just need to document the new config option for docker and podman.

I wonder if there should be no default so that it is possible to omit the platform option. I'm not sure what the risk is of always specifying `--platform` (e.g. risk that the default is wrong and user doesn't know to change it). I will trust your judgement here

pditommaso:
> Just need to document the new config option for docker and podman.

It's not a setting we are exposing directly. It's inherited by `process.arch` definition.

I agree that would be better to not have default platform selection, however part of the inconsistency comes from the fact Wave defaults to `linux/amd64` and running on silicon it would result into an error because it would not find the container. 

At the same time i'm not confident to push `linux/amd64` as default when running on macOS as @evanfloden is suggesting, because some users can rely on native containers for mac. 

I'll change to default on amd64 only when using Wave.

pditommaso:
Ok, I've found cleaner solution

Issue #5570, opened 2024-12-04
Ability to dedicate accelerator directive (GPU) over range


adamrtalbot:
## New feature

When submitting to nodes with >1 GPU, Nextflow has very limited capabilities to split the work over separate GPUs.

## Usage scenario 

Let's imagine running on AWS Batch. We submit multiple GPU enabled tasks to the Batch service. AWS allocates them to a single, large instance with multiple GPUs, as per it's allocation strategy which prioritises the cheapest per CPU price.

In this instance, all tasks would be able to use all GPUs at the same time, leading to collisions and GPU memory issues.

We have some strategies to deal with this:

- Use a specific machine size which can only fit a single GPU enabled task
- Use environment variables such as NVIDIA_VISIBLE_DEVICES to point each task at a specific GPU
- Set maxForks to 1 to ensure only a single task is executed at once

However, we lack a way of saying "for a queue of x GPUs, assign each task to 1 GPU available"

What I want is each task to know which GPU we can use, then only use that GPU.

## Suggest implementation 

I don't actually have a good fix here. Perhaps using a [process array](https://www.nextflow.io/docs/latest/reference/process.html#array) with an index might help? Perhaps it's specific to each executor? But I feel like Nextflow could expose a variable to help us here.

bentsherman:
I think this needs to be addressed by the executor, i.e. AWS Batch or SLURM should allow multiple tasks to request 1 GPU each, pack them onto a multi-GPU node as it would for CPUs, and set the `NVIDIA_VISIBLE_DEVICES` for each task.

Otherwise Nextflow would basically have to become the executor by tracking the VM assignments for each task in order to figure out which GPUs are available at any given time.

FloWuenne:
@bentsherman SLURM is already doing this I believe? Did some testing yesterday and setting the following in the nextflow.config will tell SLURM to ask for 1 GPU for each task.

```
process{
   clusterOptions = '--gres=gpu:1'  // Request 1 GPU per task
} 
```

 If submitted to a multi-GPU node, different GPUs will be assigned but they will all be mapped to device 0 in each task. So `CUDA_VISIBLE_DEVICES` will be 0 inside all tasks running in parallel on different GPUs I believe.

bentsherman:
For SLURM it may depend on the individual cluster setup. The sysadmin can (probably) use cgroups to isolate GPUs just like you would for CPUs and memory, so that a job only sees the requested resources even if the underlying node has more.

[...]

PR #5850, opened 2025-03-04
Support accelerator directive for local executor

bentsherman:
When a pipeline runs multiple GPU-enabled tasks on the same node, each task will see all GPUs and will not try to coordinate which task should use which GPU.

NVIDIA provides the `CUDA_VISIBLE_DEVICES` environment variable to control which tasks can see which GPUs, and users generally have to manage this variable themselves. Some HPC schedulers can assign this variable automatically, or use cgroups to control GPU visibility at a lower level.

Nextflow should be able to manage this variable for the local executor, so that the user doesn't have to add complex pipeline logic to do the same. Running a GPU workload locally on a multi-GPU node is a common use case, so it is worth doing.

See the docs in the PR for usage.

To use with containers, you might have to add `CUDA_VISIBLE_DEVICES` to `docker.envWhitelist`.
* I don't remember if `CUDA_VISIBLE_DEVICES` works with containers or if you have to set `NVIDIA_VISIBLE_DEVICES`.
* I don't remember if you have to set `--gpus` for the docker command in order to use the GPUs at all, that can be set in `docker.runOptions`

See also: #5570


modules/nextflow/src/main/groovy/nextflow/processor/LocalPollingMonitor.groovy
```diff
@@ -241,12 +241,13 @@ class LocalPollingMonitor extends TaskPollingMonitor {
...
      *      {@code false} otherwise
      */
+    @Override
     protected boolean remove(TaskHandler handler) {
         final result = super.remove(handler)
         if( result ) {
             availCpus += cpus(handler)
             availMemory += mem(handler)
-            availGpus.release(((LocalTaskHandler) handler).gpuSlots)
+            availGpus.release(((LocalTaskHandler) handler).gpuSlots ?: [])
```
pditommaso, line 250: Nit, `List.of()` prevent creating a new class for each task

thealanjason:
@bentsherman This feature implemented in this PR would really help us use all the local GPUs without dealing with scheduling tasks on them manually. 

Do you know when this feature will be released? Or is it even planned?

bentsherman:
Right now I just put it out so that people can try it out, so I encourage you to try it with a local build of this PR. In principle we do want to have this, just haven't decided whether it should be part of `local` or a separate executor like `local-gpu`

thealanjason:
Hi @bentsherman or @pditommaso , I finally got some time to try this out. However, I was not able to compile nextflow from source. 

I used the following steps: 
1. Install Java (version 17) using SDKMAN
2. Clone the nextflow repository
3. Checkout this PR branch `local-gpu-executor`
4. `cd nextflow`
5. `make compile`

The error I get is as below: 
```
ajc@mbd:~/Work/git_EXT/nextflow$ make compile
./gradlew compile exportClasspath
> Task :nextflow:compileGroovy FAILED

[Incubating] Problems report is available at: file:///home/ajc/Work/git_EXT/nextflow/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':nextflow:compileGroovy'.
> Could not resolve all files for configuration ':nextflow:compileClasspath'.
   > Could not resolve com.github.nextflow-io.language-server:compiler:main-SNAPSHOT.
     Required by:
         project :nextflow
      > Could not resolve com.github.nextflow-io.language-server:compiler:main-SNAPSHOT.
         > Unable to load Maven meta-data from https://s3-eu-west-1.amazonaws.com/maven.seqera.io/releases/com/github/nextflow-io/language-server/compiler/main-SNAPSHOT/maven-metadata.xml.
            > Could not GET 'https://s3-eu-west-1.amazonaws.com/maven.seqera.io/releases/com/github/nextflow-io/language-server/compiler/main-SNAPSHOT/maven-metadata.xml'. Received status code 403 from server: Forbidden
      > Could not resolve com.github.nextflow-io.language-server:compiler:main-SNAPSHOT.
         > Unable to load Maven meta-data from https://s3-eu-west-1.amazonaws.com/maven.seqera.io/snapshots/com/github/nextflow-io/language-server/compiler/main-SNAPSHOT/maven-metadata.xml.
            > Could not GET 'https://s3-eu-west-1.amazonaws.com/maven.seqera.io/snapshots/com/github/nextflow-io/language-server/compiler/main-SNAPSHOT/maven-metadata.xml'. Received status code 403 from server: Forbidden

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/8.12.1/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.

BUILD FAILED in 4s
27 actionable tasks: 21 executed, 2 from cache, 4 up-to-date
make: *** [Makefile:31: compile] Error 1
```

I'm not really sure, why I receive the 403 status code during download. Do you have any ideas to fix this?

I would really want to try out this feature on our local GPU machines.

thealanjason:
Hi @bentsherman or @pditommaso, I'd be happy if you could have a look at this PR: 
https://github.com/nextflow-io/nextflow/pull/6189

thealanjason:
Hi @bentsherman or @pditommaso, could you please also have a look at this PR:
https://github.com/nextflow-io/nextflow/pull/6218

It proposes a fix to respect gpuIDs set in `CUDA_VISIBLE_DEVICES` before running `nextflow`.

bentsherman:
@thealanjason thank you again, you actually inspired me to improve the overall approach and make it more generic.

I removed the `executor.gpus` config setting and rely solely on the `CUDA_VISIBLE_DEVICES` environment variable to make GPUs visible to Nextflow. This is also easily extended to support other runtimes like AMD ROCm and HIP, since they follow the same convention.

@pditommaso I think this PR is ready for serious consideration. Using `CUDA_VISIBLE_DEVICES` (and its variants) for everything gives us a seamless way for GPU users to integrate with Nextflow. The `AcceleratorTracker` is a nice abstraction that can be extended to support new devices and strategies without complicating the local executor.

modules/nextflow/src/main/groovy/nextflow/processor/LocalPollingMonitor.groovy
```diff
@@ -189,6 +210,13 @@ class LocalPollingMonitor extends TaskPollingMonitor {
...
+        // accelerator ids must be assigned before the task is submitted
+        final taskAccelerators = accelerators(handler)
+        if ( taskAccelerators > 0 ) {
+            ((LocalTaskHandler) handler).acceleratorEnv = acceleratorTracker.name()
+            ((LocalTaskHandler) handler).acceleratorIds = acceleratorTracker.acquire(taskAccelerators)
+        }
+
         super.submit(handler)
         availCpus -= cpus(handler)
         availMemory -= mem(handler)
```
thealanjason: @bentsherman 
I know that this suggestion is not related to this PR, but to me it makes sense to subtract resources from the polling monitor before submitting the task.

This was also the reason why the original version of this PR did not work as expected.

It may avoid a race condition even if `submit` function from the `LocalPollingMonitor` is called asynchronously. 


```suggestion
        availCpus -= cpus(handler)
        availMemory -= mem(handler)
        super.submit(handler)
```

thealanjason (commented):
It's great that now NVIDIA, AMD, and HIP devices can be handled generically :)

docs/executor.md
```diff
@@ -241,6 +242,23 @@ The local executor supports two types of tasks:
 - Script tasks (processes with a `script` or `shell` block) - executed via a Bash wrapper
 - Native tasks (processes with an `exec` block) - executed directly in the JVM.
 
+### Accelerators
+
+:::{versionadded} 25.10.0
+:::
+
+The local executor can use the `accelerator` directive to allocate accelerators such as GPUs. To enable this feature, set one of the following environment variables based on the device that you are using:
```
christopher-hakkaart:
```suggestion
The local executor can use the `accelerator` directive to allocate accelerators, such as GPUs. To enable accelerators, set the environment variable that matches the device you are using:
```

docs/executor.md
```diff
@@ -241,6 +242,23 @@ The local executor supports two types of tasks:
...
+
+- `CUDA_VISIBLE_DEVICES` for [NVIDIA CUDA](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#cuda-environment-variables) applications
+
+- `HIP_VISIBLE_DEVICES` for [HIP](https://rocm.docs.amd.com/projects/HIP/en/docs-develop/reference/env_variables.html) applications
+
+- `ROCR_VISIBLE_DEVICES` for [AMD ROCm](https://rocm.docs.amd.com/en/latest/conceptual/gpu-isolation.html) applications
+
+Set the environment variable to a comma-separated list of device IDs that you want to make available to Nextflow. The device ID can be an integer or a UUID. For example, to use all GPUs on a node with four NVIDIA GPUs, set `CUDA_VISIBLE_DEVICES` to `0,1,2,3`.
```
christopher-hakkaart:
```suggestion
Set the environment variable to a comma-separated list of device IDs that you want Nextflow to access. You can use integers or UUIDs as device identifiers. For example, to use all GPUs on a node with four NVIDIA GPUs, set `CUDA_VISIBLE_DEVICES` to `0,1,2,3`.
```

docs/executor.md
```diff
@@ -241,6 +242,23 @@ The local executor supports two types of tasks:
...
+
+Each task that requests accelerators is allocated accordingly using the same environment variable. Continuing the above example, if four tasks each request one GPU, they will be executed with `CUDA_VISIBLE_DEVICES` set to `0`, `1`, `2`, and `3`, respectively.
```
christopher-hakkaart:
```suggestion
Nextflow uses the same environment variable to allocate accelerators for each task that requests them.

Continuing the above example, if four tasks each request one GPU, they will be executed with `CUDA_VISIBLE_DEVICES` set to `0`, `1`, `2`, and `3`, respectively.
```

modules/nextflow/src/main/groovy/nextflow/processor/LocalPollingMonitor.groovy
bentsherman: I see, good point. Applied your suggestion to the PR

ECM893:
Just commenting to say this would help on SO many of cloud compute deployments.

bentsherman:
@ECM893 do you typically use the local executor in cloud for GPUs? If so I'm curious what your process looks like

ECM893:
Yes.
I use nf-core docker based pipelines, in cloud instance VMs, in Azure and GCP.
I would like to be able to say "hey, here's a pool of 8 Graphics cards, attached to this machine, just use them as needed"

modules/nextflow/src/main/groovy/nextflow/processor/AcceleratorTracker.groovy
```diff
@@ -0,0 +1,80 @@
...
+@CompileStatic
+class AcceleratorTracker {
+
+    private static final List<String> DEVICE_ENV_NAMES = [
+        'CUDA_VISIBLE_DEVICES',
+        'HIP_VISIBLE_DEVICES',
+        'ROCR_VISIBLE_DEVICES'
+    ]
+
+    static AcceleratorTracker create() {
```
pditommaso, line 38: Worth adding a unit test for this logic

modules/nextflow/src/main/groovy/nextflow/processor/AcceleratorTracker.groovy
```diff
@@ -0,0 +1,80 @@
...
+    private final TrackingSemaphore semaphore
+
+    AcceleratorTracker(String name, List<String> ids) {
+        this.name = name
+        this.semaphore = new TrackingSemaphore(ids)
+    }
+
+    AcceleratorTracker() {
+        this(null, [])
+    }
```
pditommaso: Likely should be private since it's expected to created via `create`

modules/nextflow/src/main/groovy/nextflow/processor/LocalPollingMonitor.groovy
```diff
@@ -179,9 +196,13 @@ class LocalPollingMonitor extends TaskPollingMonitor {
         if( taskMemory>maxMemory)
             throw new ProcessUnrecoverableException("Process requirement exceeds available memory -- req: ${new MemoryUnit(taskMemory)}; avail: ${new MemoryUnit(maxMemory)}")
 
-        final result = super.canSubmit(handler) && taskCpus <= availCpus && taskMemory <= availMemory
+        final taskAccelerators = accelerators(handler)
+        if( taskAccelerators > acceleratorTracker.total() )
+            throw new ProcessUnrecoverableException("Process requirement exceeds available accelerators -- req: $taskAccelerators; avail: ${acceleratorTracker.total()}")
```
pditommaso, line 201: may be worth to isolate into a separate method to make the submit logic more readable

modules/nextflow/src/main/groovy/nextflow/processor/LocalPollingMonitor.groovy
```diff
@@ -194,9 +215,16 @@ class LocalPollingMonitor extends TaskPollingMonitor {
      */
     @Override
     protected void submit(TaskHandler handler) {
-        super.submit(handler)
         availCpus -= cpus(handler)
         availMemory -= mem(handler)
+
+        final taskAccelerators = accelerators(handler)
+        if( handler instanceof LocalTaskHandler && taskAccelerators > 0 ) {
```
pditommaso: LocalTaskHandler is not implicit type for the local executor?

pditommaso (changes requested):

docs/executor.md
```diff
@@ -241,6 +242,25 @@ The local executor supports two types of tasks:
...
 - Native tasks (processes with an `exec` block) - executed directly in the JVM.
 
+(local-accelerators)=
+
+### Accelerators
+
+:::{versionadded} 25.10.0
+:::
+
+The local executor can use the `accelerator` directive to allocate accelerators such as GPUs. To use accelerators, set the corresponding environment variable:
```
christopher-hakkaart:
```suggestion
The local executor can use the `accelerator` directive to allocate accelerators, such as GPUs. To use accelerators, set the corresponding environment variable:
```

docs/executor.md
```diff
@@ -241,6 +242,25 @@ The local executor supports two types of tasks:
...
+
+The local executor can use the `accelerator` directive to allocate accelerators, such as GPUs. To use accelerators, set the corresponding environment variable:
+
+- `CUDA_VISIBLE_DEVICES` for [NVIDIA CUDA](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#cuda-environment-variables) applications
+
+- `HIP_VISIBLE_DEVICES` for [HIP](https://rocm.docs.amd.com/projects/HIP/en/docs-develop/reference/env_variables.html) applications
+
+- `ROCR_VISIBLE_DEVICES` for [AMD ROCm](https://rocm.docs.amd.com/en/latest/conceptual/gpu-isolation.html) applications
+
+Set the environment variable to a comma-separated list of device IDs that you want Nextflow to access. Nextflow uses the same environment variable to allocate accelerators for each task that requests them.
```
christopher-hakkaart:
```suggestion
Set the environment variable to a comma-separated list of device IDs for Nextflow to access. Nextflow uses this same variable to allocate accelerators for tasks that request them.
```

christopher-hakkaart (approved):
I've added one suggestion. Please check meaning is retained. Otherwise docs looks. Approved.

modules/nextflow/src/main/groovy/nextflow/processor/LocalPollingMonitor.groovy
bentsherman: The handler can also be `NativeTaskHandler`

thealanjason:
Hi @bentsherman and @pditommaso 
Out of curiosity, I tried using the test case [here](https://github.com/thealanjason/showmehow/tree/gpu_test/usecase99) on the `local-gpu-executor` branch. Unfortunately, it does not work as intended when using the [`accelerator 1`](https://github.com/thealanjason/showmehow/blob/74ad8846c40a29279a0662ec78e9ec7ee09532bc/usecase99/SHOWME.how.nf#L28) directive.

The command I am using is `CUDA_VISIBLE_DEVICES=2,3 nextflow run SHOWME.how.nf`

My observation is in the environment variable `CUDA_VISIBLE_DEVICES`, ~some tasks receive only one value~, some others receive two, ~some others none~.

I'm not sure why this happens, but maybe some more tests are needed.

thealanjason:
image

bentsherman:
@thealanjason thanks for testing. I added an end-to-end test based on your use case, but I wan't able to replicate your issue.

Can you add `echo $CUDA_VISIBLE_DEVICES` to your task script so that we can see how each task is being allocated GPUs?

jorgee:
## Follow-up: CI integration-test failure + an accelerator allocation race

While looking into the failing CI checks I found two related issues and pushed fixes to the branch (`ebc3ff7`).

### 1. `local-gpu.nf` integration test was core-count dependent

`test_integration` / `test_parser_v2` were failing with:

```
~ Test 'local-gpu.nf' run failed
   .expected /dev/fd/63 differ: byte 29, line 2
```

Root cause: the test sets `CUDA_VISIBLE_DEVICES="0,1,2,3,4,5,6,7"` and runs 8 `accelerator 1` tasks, and `.expected` assumed every task gets a **distinct** id `0–7`. But the GitHub runners only have **4 cores** (`cpus=4; capacity=4` in the log), so the tasks run in waves and GPU ids get released and **recycled** — observed ids cluster at `0–3` and never reach `4–7`, so the sorted output never matches `.expected`. It would only pass on a host with ≥8 cores.

**Fix:** added `tests/local-gpu.config` pinning `executor.cpus`/`queueSize = 8` so the 8 tasks are guaranteed to run concurrently regardless of physical cores, and reworked `.checks` to assert the allocation in an order-independent way (8 distinct ids appear for the concurrent wave; `matmul2` ran, proving ids were released and reacquired) instead of comparing against a fixed `.expected` file (now removed).

Verified locally: on a 12-core box the monitor still ran at `cpus=8/capacity=8` (override took effect), `matmul1` got distinct GPUs `0–7`, and `matmul2` reused them after release.

### 2. Data race in `TrackingSemaphore`

`acquire()` runs on the *Task submitter* thread (`submit → acquire`) while `release()` runs on the *Task monitor* thread (`evict → remove → release`) — these run concurrently on the normal task-completion path. The original `release()` released the semaphore permit **before** updating the id map:

```groovy
semaphore.release(ids.size())
for( id in ids ) availIds.put(id, true)   // map updated *after* the permit is freed
```

So a waking `acquire()` could observe permits-available but the freed ids not-yet-marked, scan the map, and **return fewer device ids than requested** (a short/empty `CUDA_VISIBLE_DEVICES`). The plain `HashMap` also had no happens-before guarantee for visibility.

**Fix:**
- update the map **before** releasing the permit in `release()`;
- guard the map mutations with `synchronized(availIds)` (acquired *outside* the blocking `semaphore.acquire()` to avoid deadlocking a concurrent release);
- use a `LinkedHashMap` for deterministic, insertion-order allocation.

### Unit tests

Added `TrackingSemaphoreTest`, including a concurrent acquire/release test (4 threads × 2000 iterations) asserting no id is ever double-assigned and no permits leak. It fails against the original ordering and passes against the fix.

### Docs

- Targeted the feature at **26.10.0** and moved the migration note from the shipped `25-10.md` to a new `migrations/26-10.md`.
- Noted that the device variable must be **non-empty** (otherwise tasks requesting accelerators fail).
- Added a note that running in containers requires propagating the variable (`docker.envWhitelist`) and configuring GPU access (`docker.runOptions`, e.g. `--gpus`).

---

One item I considered but **did not** change: in `LocalPollingMonitor.submit()` the cpus/mem/accelerator allocations now happen before `super.submit()`, so a throw from `handler.submit()` would leak resources (since `remove()` won't compensate for a handler never added to the running queue). In practice the synchronous throw surface is narrow (`ProcessUnrecoverableException` for fs-type mismatch, which aborts the session anyway; `builder.start()` failures are async and captured), so I left it as-is — happy to add a guard if you'd prefer it be correct-by-construction.

jorgee (approved):
Included a commit with fixes for failing integration tests, a possible race issue, and docs with the new expected version (26.10). Everything else looks fine to me.

I have added the 26.10 migration notes page with the content that was previously included in the 25.10 migration notes; not sure if it should be included or if you want to do it later. Feel free to remove it.

modules/nextflow/src/main/groovy/nextflow/processor/LocalPollingMonitor.groovy
pditommaso, line 201: **Backward-incompatible change (please reconsider).**

When no device env var is set, `AcceleratorTracker.create()` returns the no-arg tracker with `total() == 0` and `name() == null`. So any task that declares `accelerator N` and runs on the local executor without `CUDA_VISIBLE_DEVICES`/`HIP_*`/`ROCR_*` will now hit this `throw` and abort the run.

Previously the local executor simply **ignored** the `accelerator` directive and ran the task. This is a very common case — nf-core and other pipelines routinely declare `accelerator` on GPU processes, and users test those pipelines locally on machines where these env vars aren't set. This turns previously-working local runs into hard failures.

Suggestion: preserve the legacy behavior when no pool was advertised, and only enforce the limit when one was:
```groovy
if( acceleratorTracker.name() != null && taskAccelerators > acceleratorTracker.total() )
    throw new ProcessUnrecoverableException(...)
```
and correspondingly skip the acquire in `submit()` when `acceleratorTracker.name() == null`.

modules/nextflow/src/main/groovy/nextflow/processor/LocalPollingMonitor.groovy
```diff
@@ -194,9 +215,16 @@ class LocalPollingMonitor extends TaskPollingMonitor {
      */
     @Override
     protected void submit(TaskHandler handler) {
-        super.submit(handler)
         availCpus -= cpus(handler)
```
pditommaso: **Resource leak introduced by the reorder.**

This moves the cpu/mem decrement to *before* `super.submit(handler)` (line 227). `super.submit` → `handler.submit()` can throw **synchronously** before the task is added to `runningQueue` — e.g. `buildTaskWrapper()` or `createLaunchProcessBuilder()` raising `ProcessUnrecoverableException` (non-POSIX work dir, Fusion filesystem misconfig).

When that happens, `submitPendingTasks` catches it → `evict` → `remove`, but `TaskPollingMonitor.remove` returns **false** (the handler was never added to `runningQueue`), so the `if(result)` branch in `LocalPollingMonitor.remove` never runs. Result: cpus, memory, and any acquired device IDs are decremented/acquired but **never released** — a permanent leak from the available pool. The original ordering (`super.submit` first, then decrement) avoided exactly this.

The accelerator acquire genuinely must happen before `super.submit` (the handler needs `acceleratorIds` set before `handler.submit()` builds the `ProcessBuilder`), so the fix isn't simply moving it back. Suggest: keep the cpu/mem decrement *after* `super.submit`, and release the acquired IDs if `super.submit` throws:
```groovy
protected void submit(TaskHandler handler) {
    final taskAccelerators = accelerators(handler)
    if( handler instanceof LocalTaskHandler && taskAccelerators > 0 ) {
        handler.acceleratorEnv = acceleratorTracker.name()
        handler.acceleratorIds = acceleratorTracker.acquire(taskAccelerators)
    }
    try {
        super.submit(handler)
    }
    catch( Throwable e ) {
        if( handler instanceof LocalTaskHandler && handler.acceleratorIds )
            acceleratorTracker.release(handler.acceleratorIds)
        throw e
    }
    availCpus -= cpus(handler)
    availMemory -= mem(handler)
}
```

modules/nextflow/src/main/groovy/nextflow/processor/AcceleratorTracker.groovy
```diff
@@ -0,0 +1,84 @@
...
+
+    static AcceleratorTracker create(Map<String,String> env) {
+        return DEVICE_ENV_NAMES.stream()
+            .filter(name -> env.containsKey(name))
+            .map((name) -> {
+                final ids = env.get(name).tokenize(',')
+                return new AcceleratorTracker(name, ids)
+            })
+            .findFirst().orElse(new AcceleratorTracker())
```
pditommaso, line 57: Minor: `create()` selects only the **first** matching env var in the fixed order CUDA → HIP → ROCR. If a user has more than one set, the others are silently ignored. Probably fine, but worth a one-line doc note so the precedence is explicit.

modules/nextflow/src/test/groovy/nextflow/util/TrackingSemaphoreTest.groovy
```diff
@@ -0,0 +1,106 @@
...
+package nextflow.util
+
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.Executors
+
+import spock.lang.Specification
+
+/**
+ *
+ * @author Jorge Ejarque <jorge.ejarque@seqera.io>
```
pditommaso: Nit: author is tagged as Jorge Ejarque here, while the class under test (`TrackingSemaphore`) and the rest of the PR are attributed to Ben Sherman. Worth aligning.

pditommaso (dismissed):
Thanks for this — the design is clean and `TrackingSemaphore` is well reasoned (acquire outside the lock, mark-available-before-release) and well tested, including the 2000-iteration concurrency test. Both launch paths get `prepareAccelerators` so Fusion is covered, and the docs note the container caveat.

I'm requesting changes on two points (inline):

1. **Backward-incompatible local behavior.** With no device env var set, `acceleratorTracker.total()` is 0, so any task declaring `accelerator N` now throws `ProcessUnrecoverableException` and aborts. Previously the local executor ignored the directive and ran the task. Pipelines (e.g. nf-core) that declare `accelerator` on GPU processes but are tested locally without `CUDA_VISIBLE_DEVICES` will break. Suggest falling back to legacy (ignore the directive) when `acceleratorTracker.name() == null`, and only enforce when a pool was actually advertised.

2. **Resource leak from the `submit()` reorder.** Moving the cpu/mem decrement (and the accelerator acquire) before `super.submit(handler)` means that if `super.submit` throws synchronously (e.g. `buildTaskWrapper()`/`createLaunchProcessBuilder()` raising for a non-POSIX work dir or Fusion misconfig), the task is never added to `runningQueue`, so `remove()` returns false and cpus/memory/device IDs are never released. The pre-PR ordering (`super.submit` first) was balanced precisely to avoid this.

Minor (non-blocking): inline notes on `AcceleratorTracker.create` first-match behavior and a stray author tag; also `tests/local-gpu.config` and `tests/checks/local-gpu.nf/.checks` are missing trailing newlines.

modules/nextflow/src/test/groovy/nextflow/util/TrackingSemaphoreTest.groovy
bentsherman: I changed it on purpose because Jorge generated this file

bentsherman:
(disregard that last commit, agent was a bit too quick to push :sweat_smile: )

bentsherman:
okay, review comments addressed

PR #5854, opened 2025-03-04
Fix bugs with v2 config parser

bentsherman:
This PR fixes a few issues I encountered while testing the v2 config parser against some nf-core configs

- make sure config file passes its params to included configs
- make sure profiles are evaluated only if they were enabled, to prevent unnecessary config inclusions (for [example](https://github.com/nf-core/configs/blob/master/nfcore_custom.config))
- factor out DSLs for `plugins`, `process`, and `profiles` block to make the code easier to read

This PR definitely should be merged before the next edge release, so that the v2 config parser is usable by nf-core

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/config/parser/v2/ConfigParserV2Test.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/config/parser/v2/ConfigParserV2Test.groovy
@@ -29,6 +29,7 @@ import nextflow.config.ConfigBuilder
 import nextflow.config.ConfigClosurePlaceholder
 import nextflow.util.Duration
 import nextflow.util.MemoryUnit
+import org.codehaus.groovy.control.CompilationFailedException
 import spock.lang.Specification
 
 /**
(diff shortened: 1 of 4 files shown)


modules/nextflow/src/test/groovy/nextflow/config/parser/v2/ConfigParserV2Test.groovy
```diff
@@ -202,31 +203,31 @@ class ConfigParserV2Test extends Specification {
 
         main. text = '''
             profiles {
-                includeConfig 'dir1/config'
-                includeConfig 'dir2/config'
-                includeConfig 'dir3/config'
+                proc1 {
+                    includeConfig 'dir1/config'
```
pditommaso, line 207: What about `includeConfig` outside `profiles` is that allowed?

bentsherman: yes of course. that is covered by other unit tests

pditommaso: Okey

pditommaso (approved):

PR #5856, opened 2025-03-05
Move troubleshooting to new docs section

christopher-hakkaart:
This PR aims to move current troubleshooting documentation to a new section. This section will be used to collect troubleshooting documentation in a standardized structure.


pditommaso:
i'd argue that as a user it's more useful to find the troubleshooting in the same (infra) context. This would not prevent having a central troubleshooting page linking individual sections

bentsherman:
I agree with Paolo

christopher-hakkaart:
To clarify, the cache and language server are okay, but you would prefer the storage and compute page to be put back on the infrastructure page?

bentsherman:
The problem is that many of these troubleshooting sections don't make as much sense when you pull them out of their proper context. We would need to re-evaluate every troubleshooting "item" to make it work on its own

christopher-hakkaart:
Fair enough. I'll convert back to draft and move the pages back into relevant sections.

docs/aws.md
```diff
@@ -478,7 +478,7 @@ The above snippet defines two volume mounts for the jobs executed in your pipeli
 
 <h4>Job queue not found</h4>
 
-**`JobQueue <YOUR_QUEUE> not found`**
+**`JobQueue <QUEUE> not found`**
```
pditommaso, line 481: Nit, as a user I find much more intuitive `YOUR something` to understand it's a placeholder that I need to replace

christopher-hakkaart: We're slowly moving towards a standardized placeholder across all doc sets. This is capitalized placeholders with underscores and no possessive adjectives. It's meant to make placeholders clearer and easier to scan. When a replacement action is required, we will start adding "Replace <PLACEHOLDER> with your XXXX" or something similar below the example. I didn't add a replacement instruction here as it's an example of what the error message looks like.

bentsherman:
Closing since we didn't reach consensus on this, and the language improvements were applied elsewhere

PR #5859, opened 2025-03-06
Fix Fusion symlink test

bentsherman:
This PR attempts to fix the race condition for `fusion-symlink.nf`. Each run of this test will now publish to a unique S3 directory using the session ID, which can be obtained after the first run using the `nextflow log` command.

### DIFF
--- a/tests/checks/fusion-symlink.nf/.checks
+++ b/tests/checks/fusion-symlink.nf/.checks
@@ -1,43 +1,5 @@
 #!/bin/bash
 
-# Retry a command with exponential backoff. The function returns 0 if the command was successful
-# on its first attempt, 1 if the command failed after all attempts, and 2 if the command was successful
-# after one or more retries.
-function _retry {
-
-  if [[ $# -lt 4 ]]; then
-    echo "Usage: _retry <max_attempts> <initial_delay> <max_delay> <cmd>"
-    return 1
-  fi
-
-  local max_attempts="$1"; shift
-  local initial_delay="$1"; shift
-  local max_delay="$1"; shift
-  local cmd=( "$@" )
-  local attempt_num=1
-  local max_attempts=${max_attempts}
-  local max_delay=${max_delay}
-  local initial_delay=${initial_delay}
-  local exit_code=0
-
-  until "${cmd[@]}"; do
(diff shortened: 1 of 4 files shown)


tests/checks/fusion-symlink.nf/.checks
```diff
@@ -49,12 +11,9 @@ fi
 #
 echo initial run
 $NXF_RUN -c .config
+RUN_ID=$($NXF_CMD log | tail -n1 | awk '{ print $7 }')
```
jordeu, line 14: Can we have concurrent `$NXF_RUN` executions in the same folder?

bentsherman: No, the test runner does one test at a time

bentsherman:
@pditommaso I'm assuming you have a retention policy set up for `s3://nextflow-ci` so that all of these published folders will be cleaned up automatically

jordeu (approved):

tests/checks/fusion-symlink.nf/.checks
pditommaso: if the goal is to isolate the tests, would not `uuidgen` be enough ?

bentsherman: sure, even better. I will update it

pditommaso (approved):

bentsherman:
Looks like `publish-s3-kms.nf` has the same problem 😅 

```
~ Test 'publish-s3-kms.nf' run failed
   + '[' -z *** ']'
   + /home/runner/work/nextflow/nextflow/nextflow -q run ../../publish-s3-kms.nf -c .config
   + tee .stdout
   ++ grep INFO .nextflow.log
   ++ grep -c 'Submitted process'
   + [[ 1 == 1 ]]
   ++ check_kms_key
   ++ aws s3api head-object --bucket nf-kms-xyz --key work/ci-test/publish-s3/HELLO.tsv
   ++ grep -c e5109f93-b42d-4c26-89ee-8b251029a41d
   
   An error occurred (404) when calling the HeadObject operation: Not Found
   + [[ 0 == 1 ]]
   + false
```

I will apply a similar fix to it

PR #5863, opened 2025-03-08
WIP: Add Pod options to make a container mounted volume read only

cgoina:
I found that in some cases when using k8s executor the job will try to write to a mounted volume, which I only need it to be readOnly. If I manually added the readOnly: true flag to the volumeMount option in the generated .command.yaml and then submitted the job to the k8s cluster manually the job would run, so this is what the PR tries to address. 

I still need to do some more testing but please let me know if there is interest in this work.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -151,9 +151,10 @@ developer/plugins
 
 ```{toctree}
 :hidden:
-:caption: Tutorials
+:caption: Guides
 :maxdepth: 1
 
-flux
+updating-spot-retries
 metrics
+flux
 ```
--- a/docs/updating-spot-retries.md
+++ b/docs/updating-spot-retries.md
@@ -1,0 +1,93 @@
+(spot-retries-page)=
+
+# Spot instance failures and retries
+
+This page describes changes in how Nextflow handles spot instance failures and retries on AWS and Google Cloud, the impact of those changes, and how to configure spot retry behavior for your pipelines. These changes apply to Nextflow 24.10 and later.
+
+## Retry behavior
+
+Up to version 24.10, Nextflow would silently retry spot instance failures up to `5` times when using AWS Batch or Google Batch. These retries were controlled by cloud-specific configuration parameters (e.g., `aws.batch.maxSpotAttempts`) and happened in cloud infrastructure without explicit visibility to Nextflow.
+
+<h3>Before Nextflow 24.10</h3>
+
+By default, Nextflow would instruct AWS and Google to automatically retry jobs lost to spot reclamation up to `5` times. Retries were handled by the cloud provider _within_ a Nextflow task. It was often unclear that tasks were restarted as there was no explicit message. Task runtimes and associated cloud costs were increased because they included the runtime of the reclaimed and retried tasks. Due to the high likelihood of reclamation before completion, long-running tasks running on spot instances frequently required retries, leading to inefficient allocation of resources and higher costs.
+
+<h3>After Nextflow 24.10</h3>
+
+The default spot reclamation retry setting changed to `0` on AWS and Google. By default, no _internal_ retries are attempted on these platforms. Spot reclamations lead to an immediate failure, exposed to Nextflow in the same way as other generic failures (for example, returning `exit code 1` on AWS). Nextflow treats these failures like any other job failure, unless a retry strategy is configured.
+
+## Impact on existing workflows
+
+If you rely on silent spot retries (the previous default behavior), you may now see more tasks fail with the following characteristics:
+
+- **AWS**: Generic failure with `exit code 1`. You may see messages indicating the host machine was terminated.
+- **Google**: Spot reclamation typically produces a specific code, but is now surfaced as a recognizable task failure in Nextflow logs.
+
+Since the default for spot retries is now `0`, you must actively enable a retry strategy if you want Nextflow to handle reclaimed spot instances automatically.
+
+## Possible actions
+
+There are four possible actions.
+
+### Do nothing
+
+If you do not configure anything, you will observe more pipeline failures when spot instances are reclaimed. This approach provides clearer visibility into failures. Failed tasks can be re-run with the `-resume` option. However, frequent task reclamation may lead to a higher failure rate and each retry requires manual intervention.
+
+:::{note}
+If you resume the pipeline using the resume option, it will pick up at the point the pipeline was interrupted and start with a retry of that task.
+:::
+
+### Re-enable spot retries
+
+You can re-enable spot retries at the provider level in your Nextflow configuration:
+
+```
+// nextflow.config
+aws {
+    batch {
+        maxSpotAttempts = 5
+    }
+}
+
+google {
+    batch {
+        maxSpotAttempts = 5
+    }
+}
+```
+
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -40,3 +40,4 @@ plugins-prod
 sandbox
 wave-tests
 x/*
+.DS_Store
```
cgoina, line 43: This gets generated on my mac so I thought I would add it

bentsherman:
Can you not do this through a pod `volumeMount` ? I believe that already has a readOnly option

cgoina:
@bentsherman I think using the persistent volume claim requires the persistent volume to exist on your kubernetes cluster. This doesn't require a persistent volume to exist. One would simply add:
```
            pod = [
               [hostPath: 'the path on the host',
                mountPath: 'path inside the pod',
                readOnly: true]
            ]
```
and this would mount the path as readOnly without trying to create any directories. We are using automounts and the path initially appears it's not there so it will try to create it but of course it fails (because the user does not have any write permissions) and the job fails as a result too.

bentsherman:
The hostPath generally only works when you're running a cluster on a single node. In a multi-node cluster it's harder to know what the appropriate host path should be since a pod could be submitted to any node. Are you sure that won't be a problem for your setup?

cgoina:
hostPath in our case is a shared NFS mounted path so it's available on all nodes from the k8s cluster. Also when the volume is already (NFS) mounted - it works. Unrelated to nextflow, but for example with docker swarm for example it was enough to mark the volume mount as "shared" and that was enough for swarm. If something like that is possible with k8s or if dynamic provisioning is an option I could try that (but I didn't see how to do that either).

.gitignore
pditommaso: Please do not include this in the PR

modules/nextflow/src/main/groovy/nextflow/k8s/model/PodHostMount.groovy
```diff
@@ -30,12 +30,29 @@ import groovy.transform.ToString
...
+    ]
+
     String hostPath
 
     String mountPath
 
-    PodHostMount(String host, String container) {
+    boolean readOnly
+
+    String type // must be empty or a value from VALID_TYPES
```
pditommaso, line 43: Is the type required by the readOnly capability or it's a separate extension?

modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
```diff
@@ -573,8 +573,16 @@ class PodSpecBuilder {
         // -- host path volumes
         for( PodHostMount entry : hostMounts ) {
             final name = nextVolName()
-            mounts << [name: name, mountPath: entry.mountPath]
-            volumes << [name: name, hostPath: [path: entry.hostPath]]
+            if (entry.readOnly) {
+                mounts << [name: name, mountPath: entry.mountPath, readOnly: true]
+            } else {
+                mounts << [name: name, mountPath: entry.mountPath]
+            }
```
pditommaso, line 580: Consider something like to improve readability  

```
if( entry.readOnly ) 
  mounts.readOnly = true
```

pditommaso (changes requested):

modules/nextflow/src/main/groovy/nextflow/k8s/model/PodHostMount.groovy
cgoina: No it's not - I can remove those - initially I thought I would need them but then found that the readOnly was possible and much simpler. Let me know what you prefer.

pditommaso: I'd suggest to remove then

modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
cgoina: this fails because of the compilestatic annotation. If you think the code is not very clear I can add a comment but IMO that is completely redundant because the if is very clear - plus the branch is short

pditommaso: then you can use 

```
if( entry.readOnly ) 
  mounts.put('readOnly', true)
```

cgoina: IMO this is less obvious than this `mounts << [name: name, mountPath: entry.mountPath, readOnly: true]` but I will go with it if you prefer it

pditommaso: Sounds good

cgoina: @pditommaso done

modules/nextflow/src/main/groovy/nextflow/k8s/model/PodSpecBuilder.groovy
```diff
@@ -573,7 +573,11 @@ class PodSpecBuilder {
         // -- host path volumes
         for( PodHostMount entry : hostMounts ) {
             final name = nextVolName()
-            mounts << [name: name, mountPath: entry.mountPath]
+            def nextMount = [name: name, mountPath: entry.mountPath]
+            if( entry.readOnly )
+                nextMount.readOnly = true
+
+            mounts << nextMount
```
pditommaso, line 580: Likely `nextMount` is not needed, and the change can be applied directly to `mounts`

pditommaso: Got it, you are right

pditommaso (commented):
It looks good. Think some docs should be added. Adding @christopher-hakkaart for visibility and guidance

docs/reference/process.md
```diff
@@ -1128,11 +1128,12 @@ The following options are available:
 : *Can be specified multiple times*
 : Defines an environment variable with the given name and value.
 
-`hostPath: '/host/absolute/path', mountPath: '</pod/absolute/path>'`
+`hostPath: '/host/absolute/path', mountPath: '</pod/absolute/path>' [, readOnly: true | false]`
 : :::{versionadded} 23.10.0
   :::
 : *Can be specified multiple times*
 : Allows creating [hostPath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) volume and access it with the specified `mountPath` in the pod.
+: The `readOnly` option can be used to mount the volume as read-only (default: `false`)
```
christopher-hakkaart, line 1136:
```suggestion
: The `readOnly` option can be used to mount the volume as read-only (default: `false`).
```

christopher-hakkaart:
Missing a full stop, otherwise docs look good.

pditommaso:
@cgoina This sounds good. Consider sign-off your commit to clear the DCO bot. See [here](https://github.com/nextflow-io/nextflow/blob/master/CONTRIBUTING.md#code-changes)

PR #5868, opened 2025-03-10
Fix task progress percentage with failed tasks

bentsherman:
Spun off from #4588 to provide a more focused change

Currently, when a task fails, it is included in the "complete" and "total" task count per process. As a result, the percentage actually increases when tasks fail, even though no progress has been made. It also distorts the total number of "real tasks" vs "task attempts" in the final log.

This PR removes the "failed" count from the complete/total task counts, and adds the "ignored" count in its place. The rationale is that every failed task will either be retried or ignored, or the run will terminate and it won't matter. Of these failures, only the ignored tasks should count towards progress. Retries that succeed will be marked as succeeded and count towards progress, otherwise they shouldn't.

To see the difference, run this pipeline with and without these changes:
```groovy
process foo {
  errorStrategy { sleep(5000) ; task.exitStatus==1 && task.attempt==1 ? 'retry' : 'ignore' }

  input:
  val x

  script:
  """
  sleep ${x * 3}
  exit 1
  """
}

workflow {
  Channel.of(1, 2, 3) | foo
}
```

This pipeline executes three tasks that fail, are retried, fail again, and ignored, with some delays to help you see the progress.

Without this PR, the process will appear to make progress as tasks fail and are retried. The final task count will be 6, even though there were only 3 "real tasks" to complete.

With this PR, the progress will remain at 0 until the tasks are ignored towards the end. The final task count will be 3, with the retried and ignored count on the side.

@jorgee if you agree with these changes, please approve so that I can merge. Let me know if you have any questions


pditommaso:
We need to make sure this is not going to impact Platform progress (also considering current redesign)

bentsherman:
It looks like getTotalCount() and getCompleteCount() are not sent to platform, in fact they are explicitly excluded:

https://github.com/nextflow-io/nextflow/blob/bb11d2e9f4aa8123790bb20cc58e13a6d64f7937/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerJsonGenerator.groovy#L61-L72

Which makes sense because platform needs the raw numbers anyway

jorgee:
@bentsherman, I have pushed the fix for the report stats

modules/nextflow/src/main/groovy/nextflow/trace/WorkflowStats.groovy
```diff
@@ -114,22 +114,17 @@ class WorkflowStats implements Cloneable {
     }
 
     float getSucceedPct() {
-        int tot = gtz(succeededCount + cachedCount + ignoredCount + failedCount)
+        int tot = gtz(succeededCount + cachedCount + failedCount)
```
bentsherman, line 117: This total should include the aborted count too right? Since this total is used four times I would just factor it into a separate `getTotalCount()` function

modules/nextflow/src/main/groovy/nextflow/trace/ProgressRecord.groovy
```diff
@@ -61,13 +61,15 @@ class ProgressRecord implements Cloneable {
         this.taskName = processName
     }
 
+    // failed tasks that were not retried are included in the total count
     int getTotalCount() {
-        pending+ submitted+ running+
-           succeeded+ failed+ cached+ stored + aborted
+        pending + submitted + running +
+           succeeded + failed - retries + cached + stored + aborted
```
jorgee, line 67:
```suggestion
           succeeded + ignored + failed - retries + cached + stored + aborted
```
Missing ignored

modules/nextflow/src/test/groovy/nextflow/trace/ProgressRecordTest.groovy
```diff
@@ -53,30 +53,32 @@ class ProgressRecordTest extends Specification {
...
-        def ABORTED = 8
+        def PENDING = 1
+        def SUBMITTED = 2
+        def RUNNING = 3
+        def SUCCEEDED = 4
+        def FAILED = 5
+        def IGNORED = 6
+        def CACHED = 7
+        def STORED = 8
+        def ABORTED = 9
```
jorgee, line 64:
```suggestion
        def ABORTED = 9
        def RETRIES = 1
```

modules/nextflow/src/test/groovy/nextflow/trace/ProgressRecordTest.groovy
```diff
@@ -53,30 +53,32 @@ class ProgressRecordTest extends Specification {
...
+        rec.ignored = IGNORED
+        rec.cached = CACHED
+        rec.stored = STORED
         rec.aborted = ABORTED
 
         then:
-        rec.getCompletedCount() == SUCCEEDED+ FAILED+ CACHED+ STORED
-        rec.getTotalCount() == PENDING+ SUBMITTED+ RUNNING + SUCCEEDED+ FAILED+ CACHED+ STORED+ ABORTED
+        rec.getCompletedCount() == SUCCEEDED + IGNORED + CACHED + STORED
+        rec.getTotalCount() == PENDING + SUBMITTED + RUNNING  + SUCCEEDED + IGNORED + CACHED + STORED + ABORTED
```
jorgee, line 81:
```suggestion
        rec.getTotalCount() == PENDING + SUBMITTED + RUNNING  + SUCCEEDED + IGNORED + FAILED - RETRIES CACHED + STORED + ABORTED
```

jorgee (changes requested):
There is a test failing adding suggestions to fix it. Please check if they are correct

jorgee (approved):

bentsherman:
To summarize:

- Every failed task is either ignored, retried, or terminates the pipeline

- Retried tasks should not count towards the "total" count because a retry is essentially a duplicate of the first attempt

- Only ignored tasks should be counted towards the "completed" count because they are the only kind of failed task that represents progress

- In the execution report, retries should not be included in the progress bar, so we show the ignored count in yellow and the "effective failed" count (failed minus ignored/retried) in red

bentsherman:
These changes have also been implemented in platform as part of the run detail enhancements for a coordinated release

PR #5870, opened 2025-03-10
Draft implementation for SQL-based CID store

pditommaso:
This PR drafts an implementation for a SQL-based CID store able to handle tags annotations. 

Likely we need to review some interfaces before merging this, including moving this component into its own module


jorgee:
Looking at this draft, I think it will be better to modify the CidStore interface to pass model objects instead of the JSON serialization. I think the storage must decide what is the best option to serialize it to the storage layer

pditommaso:
Yes, I thought the same. we can once the other PR is merged

pditommaso:
Closing in favour of #5889

PR #5885, opened 2025-03-13
PoC for CID store annotations and workflow outputs structure

jorgee:
This PR is a PoC for adding the annotations to metadata entries in the CID and restructure the workflow outputs in the way defined in output DSL

It is currently getting annotations from tags

Tested with a variation of the [e2e test](https://github.com/nextflow-io/nextflow/blob/master/tests/output-dsl.nf) with a small modification to include tags.
```
output {
  samples {
    path { sample ->
      ...
      }
    }
    tags (experimentId: "value", experimentDate: "date")
    index {
      ...
    }
  }
}
```
An example of is the generated WorkflowResults. `publishedData` is a list of all files published and outputs are how they are defined as records in the ouputsDsl
```
{
    "type": "WorkflowResults",
    "run": "cid://835f10672ae237225406de48672493a0",
    "outputs": {
        "samples": [
            {
                "id": "delta",
                "fastqc": "cid://835f10672ae237225406de48672493a0/fastqc/delta.fastqc.log",
                "quant": "cid://835f10672ae237225406de48672493a0/quant/delta"
            },
            {
                "id": "beta",
                "fastqc": "cid://835f10672ae237225406de48672493a0/fastqc/beta.fastqc.log",
                "quant": "cid://835f10672ae237225406de48672493a0/quant/beta"
            },
            {
                "id": "alpha",
                "fastqc": "cid://835f10672ae237225406de48672493a0/fastqc/alpha.fastqc.log",
                "quant": "cid://835f10672ae237225406de48672493a0/quant/alpha"
            }
        ]
    },
    "publishedData": [
        "cid://835f10672ae237225406de48672493a0/fastqc/delta.fastqc.log",
        "cid://835f10672ae237225406de48672493a0/quant/delta",
        "cid://835f10672ae237225406de48672493a0/fastqc/beta.fastqc.log",
        "cid://835f10672ae237225406de48672493a0/quant/beta",
        "cid://835f10672ae237225406de48672493a0/fastqc/alpha.fastqc.log",
        "cid://835f10672ae237225406de48672493a0/quant/alpha",
        "cid://835f10672ae237225406de48672493a0/samples.csv"
    ]
}
```
Workflow output files are annotated with the provided tags

```
$ nextflow cid show cid://835f10672ae237225406de48672493a0/fastqc/delta.fastqc.log
{
    "type": "WorkflowOutput",
    "path": "/home/jorgee/nextflow_tests/provenance-test/results/fastqc/delta.fastqc.log",
    "checksum": {
        "value": "f8406b93427367bd50bc2bdf34659aa3",
        "algorithm": "nextflow",
        "mode": "standard"
    },
    "source": "cid://a786558405845a775fc6218fa6aa7b03/delta.fastqc.log",
    "size": 6,
    "createdAt": 1741870059179,
    "modifiedAt": 1741870059179,
    "annotations": {
        "experimentId": "value",
        "experimentDate": "date"
    }
}
```

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -1142,8 +1142,7 @@ class Session implements ISession {
         for( TraceObserver observer : copy  ) {
             try {
                 observer.onFilePublish(destination, source)
-                if (annotations)
-                    observer.onFileAnnotation(destination, annotations)
+                observer.onFilePublish(destination, source, annotations)
             }
             catch( Exception e ) {
                 log.error "Failed to invoke observer on file publish: $observer", e
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -1141,7 +1141,6 @@ class Session implements ISession {
         def copy = new ArrayList<TraceObserver>(observers)
         for( TraceObserver observer : copy  ) {
             try {
-                observer.onFilePublish(destination, source)
                 observer.onFilePublish(destination, source, annotations)
             }
             catch( Exception e ) {
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -1126,10 +1126,10 @@ class Session implements ISession {
         }
     }
 
-    void notifyWorkflowPublish(Object value) {
+    void notifyWorkflowPublish(String name, Object value) {
         for( final observer : observers ) {
             try {
-                observer.onWorkflowPublish(value)
+                observer.onWorkflowPublish(name, value)
             }
             catch( Exception e ) {
                 log.error "Failed to invoke observer on workflow publish: $observer", e
@@ -1137,11 +1137,13 @@ class Session implements ISession {
         }
     }
 
-    void notifyFilePublish(Path destination, Path source=null) {
+    void notifyFilePublish(Path destination, Path source, Map annotations) {
         def copy = new ArrayList<TraceObserver>(observers)
         for( TraceObserver observer : copy  ) {
             try {
                 observer.onFilePublish(destination, source)
+                if (annotations)
+                    observer.onFileAnnotation(destination, annotations)
             }
             catch( Exception e ) {
                 log.error "Failed to invoke observer on file publish: $observer", e
--- a/modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
@@ -38,6 +38,8 @@ class PublishOp {
 
     private Session session
 
+    private String name
+
     private DataflowReadChannel source
 
     private Map opts
@@ -50,12 +52,11 @@ class PublishOp {
 
     private List indexRecords = []
 
-    private String name
-
     private volatile boolean complete
 
     PublishOp(Session session, String name, DataflowReadChannel source, Map opts) {
         this.session = session
+        this.name = name
         this.source = source
         this.opts = opts
         this.path = opts.path as String
@@ -63,7 +64,6 @@ class PublishOp {
             this.pathResolver = opts.pathResolver as Closure
         if( opts.index )
             this.indexOpts = new IndexOpts(session.outputDir, opts.index as Map)
-        this.name = name
     }
 
     boolean getComplete() { complete }
@@ -109,15 +109,11 @@ class PublishOp {
             final files = entry.value
             publisher.apply(files, sourceDir)
         }
-        // append record to index file
-        if( indexOpts ) {
-            final record = indexOpts.mapper != null ? indexOpts.mapper.call(value) : value
-            final normalized = normalizePaths(record, targetResolver)
-            log.trace "Normalized record for index file: ${normalized}"
-            indexRecords << normalized
-            // emit workflow publish event
-            session.notifyWorkflowPublish(name, normalized)
-        }
+
+        // append record to index
+        final normalized = normalizePaths(value, targetResolver)
+        log.trace "Normalized record for index file: ${normalized}"
+        indexRecords << normalized
     }
 
     /**
@@ -156,12 +152,21 @@ class PublishOp {
     }
 
     /**
-     * Once all values have been published, write the
-     * index file (if enabled).
+     * Once all values have been published, publish the index
+     * and write it to a file (if enabled).
      */
     protected void onComplete(nope) {
-        if( indexOpts && indexRecords.size() > 0 ) {
-            log.trace "Saving records to index file: ${indexRecords}"
+        // publish individual record if source is a value channel
+        final index = CH.isValue(source)
+            ? indexRecords.first()
+            : indexRecords
+
+        // publish workflow output
+        session.notifyWorkflowPublish(name, index)
+
+        // write index file
+        if( indexOpts && index ) {
+            log.trace "Saving records to index file: ${index}"
             final indexPath = indexOpts.path
             final ext = indexPath.getExtension()
             indexPath.parent.mkdirs()
--- a/modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
@@ -96,6 +96,10 @@ class PublishOp {
         final overrides = targetResolver instanceof Closure
             ? [saveAs: targetResolver]
             : [path: targetResolver]
+        if (opts.annotations instanceof Closure){
+            final annotations = opts.annotations as Closure
+            overrides.annotations = annotations.call(value) as Map
+        }
         final publisher = PublishDir.create(opts + overrides)
 
         // publish files
--- a/modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
@@ -54,7 +54,7 @@ class PublishOp {
 
     private volatile boolean complete
 
-    PublishOp(String name, Session session, DataflowReadChannel source, Map opts) {
+    PublishOp(Session session, String name, DataflowReadChannel source, Map opts) {
         this.session = session
         this.source = source
         this.opts = opts
--- a/modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
@@ -50,9 +50,11 @@ class PublishOp {
 
     private List indexRecords = []
 
+    private String name
+
     private volatile boolean complete
 
-    PublishOp(Session session, DataflowReadChannel source, Map opts) {
+    PublishOp(String name, Session session, DataflowReadChannel source, Map opts) {
         this.session = session
         this.source = source
         this.opts = opts
@@ -61,6 +63,7 @@ class PublishOp {
             this.pathResolver = opts.pathResolver as Closure
         if( opts.index )
             this.indexOpts = new IndexOpts(session.outputDir, opts.index as Map)
+        this.name = name
     }
 
     boolean getComplete() { complete }
@@ -89,9 +92,6 @@ class PublishOp {
         if( targetResolver == null )
             return
 
-        // emit workflow publish event
-        session.notifyWorkflowPublish(value)
-
         // create publisher
         final overrides = targetResolver instanceof Closure
             ? [saveAs: targetResolver]
@@ -105,13 +105,14 @@ class PublishOp {
             final files = entry.value
             publisher.apply(files, sourceDir)
         }
-
         // append record to index file
         if( indexOpts ) {
             final record = indexOpts.mapper != null ? indexOpts.mapper.call(value) : value
             final normalized = normalizePaths(record, targetResolver)
             log.trace "Normalized record for index file: ${normalized}"
             indexRecords << normalized
+            // emit workflow publish event
+            session.notifyWorkflowPublish(name, normalized)
         }
     }
 
@@ -169,7 +170,7 @@ class PublishOp {
             else {
                 log.warn "Invalid extension '${ext}' for index file '${indexPath}' -- should be 'csv' or 'json'"
             }
-            session.notifyFilePublish(indexPath)
+            session.notifyFilePublish(indexPath, null, opts.tags as Map)
         }
 
         log.trace "Publish operator complete"
--- a/modules/nextflow/src/main/groovy/nextflow/script/OutputDsl.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/OutputDsl.groovy
@@ -84,7 +84,7 @@ class OutputDsl {
             final opts = publishOptions(name, defaults, overrides)
 
             if( opts.enabled == null || opts.enabled )
-                ops << new PublishOp(session, CH.getReadChannel(mixed), opts).apply()
+                ops << new PublishOp(name, session, CH.getReadChannel(mixed), opts).apply()
         }
     }
 
--- a/modules/nextflow/src/main/groovy/nextflow/script/OutputDsl.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/OutputDsl.groovy
@@ -84,7 +84,7 @@ class OutputDsl {
             final opts = publishOptions(name, defaults, overrides)
 
             if( opts.enabled == null || opts.enabled )
-                ops << new PublishOp(name, session, CH.getReadChannel(mixed), opts).apply()
+                ops << new PublishOp(session, name, CH.getReadChannel(mixed), opts).apply()
         }
     }
 
@@ -169,6 +169,14 @@ class OutputDsl {
 
         void tags(Map value) {
             setOption('tags', value)
+        }
+
+        void annotations(Map value) {
+            setOption('annotations', value)
+        }
+
+        void annotations(Closure value) {
+            setOption('annotations', value)
         }
 
         private void setOption(String name, Object value) {
--- a/modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
@@ -125,9 +125,10 @@ trait TraceObserver {
     /**
      * Method that is invoked when a value is published from a channel.
      *
+     * @param name
      * @param value
      */
-    void onWorkflowPublish(Object value){}
+    void onWorkflowPublish(String name, Object value){}
 
     /**
      * Method that is invoke when an output file is published
@@ -150,4 +151,13 @@ trait TraceObserver {
     void onFilePublish(Path destination, Path source){
         onFilePublish(destination)
     }
+    /**
+     * Method that is invoked when a output file is annotated
+     * @param destination
+     *      The destination path at `publishDir` folder.
+     * @param annotations
+     *      The annotations attached to this file
+     */
+    void onFileAnnotation(Path destination, Map annotations){}
+
 }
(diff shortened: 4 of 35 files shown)


modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
```diff
@@ -50,9 +50,11 @@ class PublishOp {
 
     private List indexRecords = []
 
+    private String name
+
     private volatile boolean complete
 
-    PublishOp(Session session, DataflowReadChannel source, Map opts) {
+    PublishOp(String name, Session session, DataflowReadChannel source, Map opts) {
```
bentsherman, line 57:
```suggestion
    PublishOp(Session session, String name, DataflowReadChannel source, Map opts) {
```

modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
```diff
@@ -125,9 +125,10 @@ trait TraceObserver {
     /**
      * Method that is invoked when a value is published from a channel.
```
bentsherman, line 126:
```suggestion
     * Method that is invoked when a value is published to a workflow output.
```

modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
```diff
@@ -150,4 +151,13 @@ trait TraceObserver {
...
         onFilePublish(destination)
     }
+    /**
+     * Method that is invoked when a output file is annotated
+     * @param destination
+     *      The destination path at `publishDir` folder.
+     * @param annotations
+     *      The annotations attached to this file
+     */
+    void onFileAnnotation(Path destination, Map annotations){}
```
bentsherman, line 161:
```suggestion
    void onFilePublish(Path destination, Path source, Map annotations){
        onFilePublish(destination, source)
    }
```

modules/nf-cid/src/main/nextflow/data/cid/model/WorkflowResults.groovy
```diff
@@ -30,5 +30,6 @@ import groovy.transform.CompileStatic
 class WorkflowResults {
     DataType type
     String run
-    List outputs
+    Map<String, Object> outputs
+    List<String> publishedData
```
bentsherman, line 34:
```suggestion
    List<String> publishedFiles
```

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -1126,22 +1126,24 @@ class Session implements ISession {
...
     }
 
-    void notifyFilePublish(Path destination, Path source=null) {
+    void notifyFilePublish(Path destination, Path source, Map annotations) {
         def copy = new ArrayList<TraceObserver>(observers)
         for( TraceObserver observer : copy  ) {
             try {
                 observer.onFilePublish(destination, source)
+                if (annotations)
+                    observer.onFileAnnotation(destination, annotations)
```
bentsherman, line 1146:
```suggestion
                observer.onFilePublish(destination, source, annotations)
```

bentsherman (changes requested):
Looks great so far. Left a few minor suggestions. I will try to play with it when I have time

jorgee:
I have updated the code with annotations field in the output. It can be Map o as closure. The closure is evaluated per sample. So now we could support the case of adding sample information as annotation.

```
output {
  samples {
    path { sample ->... }

    annotations { sample ->
        return [experimentId: params.experimentId , sampleId : sample.id]
    }
    index {
      path 'samples.csv'
      header true
      sep ','
    }
  }
}
```

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
```diff
@@ -105,13 +109,14 @@ class PublishOp {
...
         }
-
         // append record to index file
         if( indexOpts ) {
             final record = indexOpts.mapper != null ? indexOpts.mapper.call(value) : value
             final normalized = normalizePaths(record, targetResolver)
             log.trace "Normalized record for index file: ${normalized}"
             indexRecords << normalized
+            // emit workflow publish event
+            session.notifyWorkflowPublish(name, normalized)
```
bentsherman, line 119: Something I've been thinking about saving the index metadata to the workflow result. Currently it will always be a list because of how this publish event works. If the source channel is a value channel, it will be saved as a JSON array with a single element rather than just a JSON object

So I would like to refactor this event to instead happen in the onComplete of the publish op, and include either (1) the one record as an object if the source was a value channel or (2) the entire list of index records if the source was a queue channel.

If you think you can implement that, feel free, otherwise I can do it myself

jorgee: I will do it. 
The other doubt I had in this part is about normalized paths. Originally, it was calling notifyWorkflowPublish per value without normalizing the path. It was just normalized for the index file. For testing purposes I moved the notification to this place. But I think it should notify values with normalized paths in all the cases, even the index file is not set. What do you think about it?

bentsherman: Yeah I wasn't sure originally whether to normalize the paths here or not, since a trace observer could use the onFilePublish events to recover the source-target mapping.

But as we integrate the index file with the cid store, I think it makes sense to normalize them in the index file.

And yes, I think we should do the onWorkflowPublish regardless of whether the index file is specified. The only complication is the index "mapper", but I plan to remove it anyway in the third preview, so you can remove it here if you want

bentsherman: I have updated the [workflow outputs](https://github.com/nextflow-io/nextflow/tree/workflow-outputs-3) branch based on our discussion. So you can just copy the changes from that branch if you want

modules/nextflow/src/main/groovy/nextflow/script/OutputDsl.groovy
```diff
@@ -171,6 +171,14 @@ class OutputDsl {
             setOption('tags', value)
         }
 
+        void annotations(Map value) {
+            setOption('annotations', value)
+        }
+
+        void annotations(Closure value) {
+            setOption('annotations', value)
+        }
```
bentsherman, line 180: TODO: decide whether to use `tags` or phase it out in favor of `annotations`

jorgee:
Last changes:
- Implemented support for retreiving output descriptions `Channel.fromPath("cid://<workflow_run_cid>/outputs")`
- Workflow output cid files are added within `<workflow_run_cid>/outputs `to avoid the ambiguity of having an outputs with publish dir `outputs`
- Outputs file is managed as an extension o CidPath that creates a InputStream ByteChannel from the metadata description
- When passed to a task it is treated as a foreign file and it will be serialized as a staging file
- Added a new field in WorkflowResults creationTime to avoid several serializalizations when staging the outputs

modules/nf-cid/src/main/nextflow/data/cid/CidUtils.groovy
```diff
@@ -0,0 +1,110 @@
+package nextflow.data.cid
+
+import groovy.util.logging.Slf4j
+import nextflow.data.cid.fs.CidPath
+import nextflow.data.cid.serde.CidSerializable
+
+import java.nio.file.Path
+
+@Slf4j
+class CidUtils {
```
pditommaso, line 10: Missing `@CompileStatic`

modules/nf-cid/src/main/nextflow/data/cid/fs/CidPath.groovy
```diff
@@ -97,7 +106,7 @@ class CidPath implements Path, RealPathAware {
     /**
      * Finds the target path of a CID path
      **/
-    protected static Path findTarget(CidFileSystem fs, String filePath, String[] childs=[]) throws Exception{
+    protected static Path findTarget(CidFileSystem fs, String filePath, boolean resultsAsPath, String[] childs=[]) throws Exception{
```
pditommaso, line 109: This method is quite large. Likely can be split in smaller unit e.g. 

```
if( object ) {
  findTargetForObject( .. _ 
}

if( resultsAsPath ) {
  .. etc
}
```

add add the corresponding tests

modules/nf-cid/src/main/nextflow/data/cid/fs/CidPath.groovy
```diff
@@ -385,3 +408,102 @@ class CidPath implements Path, RealPathAware {
     }
 
 }
+
+class ResultsSeekableByteChannel implements SeekableByteChannel {
```
pditommaso, line 412: Missing `@CompileStatic`. Also may be convenient to keep in its own source file

plugins/nf-cid-h2/src/main/nextflow/data/cid/h2/H2CidStore.groovy
```diff
@@ -121,11 +133,34 @@ class H2CidStore implements CidStore {
...
+        return results
+    }
+    /**
+     * JSON_MATCH implementation for h2
+     * @param jsonString
+     * @param queryString
+     * @return
+     */
+    static boolean matchesJsonQuery(String jsonString, String queryString) {
+        def json = new JsonSlurper().parseText(jsonString)
```
pditommaso, line 153: Let's prefer the new Json encoder whenever possible

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
jorgee: Copied to the branch

modules/nf-cid/src/main/nextflow/data/cid/CidUtils.groovy
jorgee: Fixed last commit.

modules/nf-cid/src/main/nextflow/data/cid/fs/CidPath.groovy
jorgee: Fixed

plugins/nf-cid-h2/src/main/nextflow/data/cid/h2/H2CidStore.groovy
jorgee: Fixed

jorgee:
Last changes:
- Use string iso 8601 format for date instead of millis. 
- `query` and `show` commands unified.
- Get rid of `publishedFiles `
- New `publishedBy` field in workflowOutputs to enable search for all outputs published by a workflow. 
- Pseudo fs: modified path to render metadata elements: 
      old: `cid://<hash>/outputs` 
      new: `cid://<hash>/#outputs` ( I couldn't use ? in fromPath because of pattern,  using URI fragment instead which has a better meaning). 
- Query params still used for general query e.g nexflow cid show "cid:///?type=WorkflowOutputs&publishedBy=...."
- There is still an undetermined situation. Should I return the description of the workflow run when fromPath("cid://<hash>/") or just when a fragment is requested?
- I have recovered the results CID due to the change of render paths. As we are not really doing content-addressing, I could reuse the runCID and update the description with the outputs and remove the results CID. What do you think about it?

modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
```diff
@@ -123,11 +123,18 @@ trait TraceObserver {
...
 
     /**
-     * Method that is invoked when a value is published from a channel.
+     * Method that is invoked when a workflow output is published.
      *
+     * @param name
+<<<<<<< HEAD
+     *      The name of the workflow output
+=======
+>>>>>>> 3289e9e30 (first POC output structure and annotations)
```
pditommaso, line 132: Oops, failed merge ☝️ (?)

modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
```diff
@@ -123,11 +123,18 @@ trait TraceObserver {
...
      * @param value
+     *      A list if the published channel was a queue channel,
+     *      otherwise an object if the channel was a value channel
      */
-    void onWorkflowPublish(Object value){}
+    void onWorkflowPublish(String name, Object value){}
```
pditommaso, line 137: Is this the workflow run name? is it really necessary ?

bentsherman: It is the name of the workflow output, of which there can be several

modules/nf-cid/src/main/nextflow/data/cid/CidUtils.groovy
```diff
@@ -0,0 +1,160 @@
...
+                return searchPath(store, key, parameters, children )
+            }
+        } catch(Throwable e){
+            log.debug("Exception querying $uri. $e.message")
+            return []
+        }
+
+    }
+
+    public static String[] getChildrenFormFragment(String fragment) {
```
pditommaso, line 65: Unit test should be added

modules/nf-cid/src/main/nextflow/data/cid/CidUtils.groovy
```diff
@@ -0,0 +1,160 @@
...
+                return fragment.split("\\.")
+            } else {
+                return [fragment] as String[]
+            }
+        } else {
+            return [] as String[]
+        }
+    }
+
+    private static List<Object> searchPath(CidStore store, String path, Map<String, String> params, String[] children = []) {
```
pditommaso, line 77: Unit test and doc should be added

modules/nf-cid/src/main/nextflow/data/cid/CidUtils.groovy
```diff
@@ -0,0 +1,160 @@
...
+                newChildren.addAll(children)
+                return searchPath(store, parent.toString(), params, newChildren as String[])
+            } else {
+                throw new FileNotFoundException("Cid object $path/${children ? children.join('/') : ''} not found.")
+            }
+        }
+        return results
+    }
+
+    protected static void treatObject(def object, Map<String, String> params, List<Object> results) {
```
pditommaso, line 107: unit test should be added

modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
bentsherman: Don't think that is needed because it is an undocumented preview feature (part of workflow outputs preview)

jorgee: It is the name given to the output in the workflow. We use it as key for the outputs map, and it is also used to refer to an specific output by cid://<hash>/#output.<name>

modules/nf-cid/src/main/nextflow/data/cid/CidObserver.groovy
```diff
@@ -215,8 +217,8 @@ class CidObserver implements TraceObserver {
                 checksum,
                 "$CID_PROT$task.hash",
                 attrs.size(),
-                attrs.creationTime().toMillis(),
-                attrs.lastModifiedTime().toMillis())
+                Instant.ofEpochMilli(attrs.creationTime().toMillis()).toString(),
+                Instant.ofEpochMilli(attrs.lastModifiedTime().toMillis()).toString())
```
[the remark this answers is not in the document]
jorgee: These values are from the real file attributes

modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
bentsherman: Regarding the API change, let's move the discussion to #5909 which is the source of truth for this change

modules/nf-cid/src/main/nextflow/data/cid/CidObserver.groovy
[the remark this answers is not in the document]
pditommaso: oops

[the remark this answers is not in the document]
pditommaso: You are definitively right. In this case better to use defensive approach and consider the timestamp could be `null`. it should never happen but considering it could support third-party remote files system, everything could happen. A smaller helper function may be useful

modules/nf-cid/src/main/nextflow/data/cid/fs/CidPath.groovy
jorgee: Tests added as part of the getTargetPath tests

modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
jorgee: Fixed, I don't know why IntelliJ didn't warn me about it.

modules/nf-cid/src/main/nextflow/data/cid/CidObserver.groovy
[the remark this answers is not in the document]
jorgee: I have added two helper methods CidUtils.toDate and CidUtils.toFileTime() to convert from FileTime to String and vice-versa. Null cases are written to 'N/A'.

modules/nf-cid/src/main/nextflow/data/cid/CidUtils.groovy
jorgee: Test and doc added

modules/nf-cid/src/main/nextflow/data/cid/CidUtils.groovy
jorgee: Test and doc added

modules/nf-cid/src/main/nextflow/data/cid/CidUtils.groovy
jorgee: test and doc added

modules/nf-cid/src/main/nextflow/data/cid/CidUtils.groovy
```diff
@@ -0,0 +1,160 @@
...
+                (object as Collection).forEach { treatObject(it, params, results) }
+            } else if (checkParams(object, params)) {
+                results.add(object)
+            }
+        } else {
+            results.add(object)
+        }
+    }
+
+    public static Map<String, String> parseQuery(String queryString) {
```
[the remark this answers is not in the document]
jorgee: test and doc added

modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
pditommaso: > Don't think that is needed because it is an undocumented preview feature

Fair enough

modules/nf-cid/src/main/nextflow/data/cid/CidObserver.groovy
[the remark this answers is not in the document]
pditommaso: Considering it's a data format, better using `null` for missing values.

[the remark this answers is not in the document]
pditommaso: I'm quite sure it can be controller with the GsonBuilder, but not a blocker for this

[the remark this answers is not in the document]
jorgee: Ok, changed

jorgee:
Also added `TaskResults` descriptions. It includes all the task outputs (files and values)  as well as a reference to the taskRun and the workflowRun. 

So, we can look for the task results of a certain task run with `nextflow cid show cid://?taskRun=<taskRunCid> ` or all the tasks executed by a workflow with `nextflow cid show cid://?runBy=<workflowRunCid>`

pditommaso:
Changed the constructor of GsonEncoder because, according gpt, `RuntimeTypeAdapterFactory` is not thread safe

pditommaso:
Added the serialization of nulls

pditommaso:
Merging this, and continue the discussion about open points in the baseline PR

PR #5887, opened 2025-03-14
Support prefetching plugins metadata from an http api

tom-seqera:
This adds an implementation of p4fj's `UpdateRepository` which can download plugin metadata from an http api.

It also adds a new interface `PrefetchUpdateRepository` which extends p4fj's interface to add a `prefetch()` method. When initialising plugins, Nextflow will call this prefetch method on any UpdateRepository which implements this interface, allowing the repository to optimise how it downloads plugin metadata (for example, only loading metadata for required plugins rather than for all plugins).

Use of the new http update repository can be enabled with the new `NXF_PLUGINS_INDEX_URL` environment variable. If set, this overrides the default `indexUrl` location of the `plugins.json` metadata file. If the supplied url ends in `.json` then Nextflow will use the existing update repository implementation, otherwise it will use the new http api implementation.

Unless this new env var is set, these changes should have no effect on Nextflow's behaviour.


modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
```diff
@@ -0,0 +1,140 @@
...
+import java.net.http.HttpClient
+import java.net.http.HttpRequest
+import java.net.http.HttpResponse
+
+/**
+ * Represents an update repository served via an HTTP api.
+ */
+@Slf4j
+@CompileStatic
+class HttpPluginRepository implements PrefetchUpdateRepository {
```
pditommaso, line 29: it would be nice to be better describe how this prefect logic works. Where the pre-fetch data is stored. How it's loaded, when it expires, etc.

modules/nf-commons/src/main/nextflow/plugin/PluginsFacade.groovy
```diff
@@ -95,6 +98,17 @@ class PluginsFacade implements PluginStateListener {
         }
     }
 
+    protected String getPluginsIndexUrl() {
+        final url = env.get('NXF_PLUGINS_URL')
+        if( url ) {
+            log.trace "Detected NXF_PLUGINS_URL=$url"
+            return url
+        } else {
+            log.trace "Using default plugins url"
```
pditommaso, line 107: We may want show warning when using a non-official index url, in the same vein of 

https://github.com/nextflow-io/nextflow/blob/3e7670f5486c1465bc3547448f27a1b7e0bdeaa0/modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy#L99-L112

modules/nf-commons/src/main/nextflow/plugin/PluginsFacade.groovy
```diff
@@ -95,6 +98,17 @@ class PluginsFacade implements PluginStateListener {
         }
     }
 
+    protected String getPluginsIndexUrl() {
+        final url = env.get('NXF_PLUGINS_URL')
```
pditommaso, line 102: Let be verbose here and use something like `NXF_PLUGINS_INDEX_URL`

pditommaso (changes requested):

modules/nf-commons/src/main/nextflow/plugin/PluginsFacade.groovy
```diff
@@ -95,6 +98,28 @@ class PluginsFacade implements PluginStateListener {
         }
     }
 
+    protected String getPluginsIndexUrl() {
+        final url = env.get('NXF_PLUGINS_INDEX_URL')
+        if( url ) {
```
pditommaso, line 103: Ideally there should not be any warning if the `url` is the current (github) index or the uri host is `*.seqera.io`

pditommaso (commented):
Not major comments apart if I'm not wrong there isnt any unit test for this pre-fetching mechanism ?

tom-seqera:
Added unit tests

pditommaso (changes requested):
Ok, i've made a couple of minor comments and changes.

modules/nf-commons/src/main/nextflow/plugin/PluginsFacade.groovy
```diff
@@ -320,27 +347,21 @@ class PluginsFacade implements PluginStateListener {
...
+        start([PluginSpec.parse(pluginId, defaultPlugins)])
     }
 
-    void start(PluginSpec plugin) {
-        if( isEmbedded() && defaultPlugins.hasPlugin(plugin.id) ) {
-            log.debug "Plugin 'start' is not required in embedded mode -- ignoring for plugin: $plugin.id"
-            return
+    void start(List<PluginSpec> specs) {
+        def split = specs.split { plugin -> isEmbedded() && defaultPlugins.hasPlugin(plugin.id) }
+        def (skip, toStart) = [split[0], split[1]]
```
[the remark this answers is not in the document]
tom-seqera: This is just preserving the existing behaviour which doesn't start default plugins in embedded mode. The only change is that it used to check start each plugin one-by-one but now we need to get a list of all the plugins in one go so that we can call prefetch()

tom-seqera:
I updated the warning to be clearer that using a custom index is currently experimental. The previous wording suggested that people should only use the main plugin index, but I think people will want the option to have alternative plugin indexes.

pditommaso:
Please revert [920ed83](https://github.com/nextflow-io/nextflow/pull/5887/commits/920ed830ab2aa4fe57b04032e739c6f4db4d7241), the expected message is the previous commit

PR #5889, opened 2025-03-16
Add CID H2 plugin 

pditommaso:
## Summary 

This PR adds a new CID implementation that uses H2 to store CID data. 

#### Get started 

Add the following settings in your `nextflow.config` file 

```
plugins {
  id 'nf-cid-h2'
}

workflow.data.store.location = "jdbc:h2:file:./.data/h2;AUTO_SERVER=TRUE"
workflow.data.enabled = true
```

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdCid.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdCid.groovy
@@ -29,6 +29,7 @@ import nextflow.plugin.Plugins
 import org.pf4j.ExtensionPoint
 
 /**
+ * CID command line interface
  *
  * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
  */
@@ -38,7 +39,7 @@ class CmdCid extends CmdBase implements UsageAware {
 
     private static final String NAME = 'cid'
 
-    interface CidOperation extends ExtensionPoint {
+    interface CidCommand extends ExtensionPoint {
         void log(ConfigMap config)
         void show(ConfigMap config, List<String> args)
         void lineage(ConfigMap config, List<String> args)
@@ -53,7 +54,7 @@ class CmdCid extends CmdBase implements UsageAware {
 
     private List<SubCmd> commands = new ArrayList<>()
 
-    private CidOperation operations
+    private CidCommand operation
 
     private ConfigMap config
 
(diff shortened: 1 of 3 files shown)


modules/nf-cid/src/main/nextflow/data/cid/CidStore.groovy
```diff
@@ -49,12 +47,6 @@ interface CidStore {
      */
     Object load(String key)
 
-    /**
-     * Get the CID store location path.
-     * @return CID store location path.
-     */
-    Path getPath()
```
pditommaso, line 56: It looks like this was used only for testing. Better to remove for the interface to keep more generic

modules/nf-cid/src/main/nextflow/data/config/DataStoreOpts.groovy
```diff
@@ -30,14 +26,10 @@ import groovy.transform.CompileStatic
 @CompileStatic
 class DataStoreOpts {
 
-    final Path location
-    final Path logLocation
+    final String location
```
pditommaso, line 29: Store location not necessarily needs to be a file system path. Changed to a generic `String`

modules/nf-cid/src/main/nextflow/data/config/DataStoreOpts.groovy
```diff
@@ -30,14 +26,10 @@ import groovy.transform.CompileStatic
 @CompileStatic
 class DataStoreOpts {
 
-    final Path location
-    final Path logLocation
```
pditommaso, line 34: It would be better to keep at level of the concrete implementation

plugins/nf-cidh2/src/resources/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,6 @@
+Manifest-Version: 1.0
+Plugin-Class: nextflow.data.cid.h2.H2CidPlugin
+Plugin-Id: nf-cidh2
```
bentsherman, line 3: I recommend calling it `nf-cid-h2`

jorgee:
I have tested with rnaseq-nf and working with h2 and default when using local disk. Defautl with s3 locations were not supported. I have pushed some changes in `DefaultCidStoreFactory` to allow s3, gs and az schemes

pditommaso:
@jorgee is this ready to go in your side?

jorgee (approved):

jorgee:
Sorry, I forgot to rename the plugin, just I commit missing

plugins/nf-cidh2/src/resources/META-INF/MANIFEST.MF
jorgee: Renamed in last commit

jorgee:
Now renamed plugin is pushed

plugins/nf-cidh2/src/resources/META-INF/MANIFEST.MF
pditommaso: Nice

pditommaso:
Ok, I've also replace the Service loader in the cid command with classic plugin

PR #5893, opened 2025-03-18
Add serde interfaces

pditommaso:
Draft implementation for a generic ser-deserialization mechanism for CID model. 

Inspired to https://github.com/seqeralabs/wave/blob/77bbb416e8ff2101a8ede397049c5022d1f1d1b1/src/main/groovy/io/seqera/wave/encoder/MoshiEncodeStrategy.groovy#L46-L46


jorgee:
Update:
- Changes in the `CidObserver`, `CidOperationImpl,` and `CidPath` to support the new interface.
- Restrict the encoder interface to encode/decode `CidSerializable` objects. 
- Fix tests

Note:  I have restricted the encoder interface because for CIDs we do not need a generic input type. Also when loading the object from the `CidStore` in `CidPath` or the `CidOperationImpl` classes, we don't know the exact model class that we expected, it is only known data is retrieved and the type field is inspected, we only know they are` CidSerializable` and the Groovy's `JsonSlurper` was not able to cast to an interface. So, I needed to add something to manage the polymorphism.  Currently, I have implemented a simple mechanism in the `decode` method including in the `DataType `enum the class to cast the retrieved object.  We could manage more generically with Moshi Adapters as done in Wave but we must add a new dependency and it is not required for the Cid case. I can do it later if we want to create a generic encoder to be used in other part in Nextflow.

pditommaso:
Just update the `cid-store` branch, let's solve the conflicts

pditommaso:
Went ahead with the suggestion for a generic serde. Main changes: 

* Move the base encoder interface and logic in `nf-commons` 
* Implement a json encoder based on Gson that's already included in the nextflow base classpath 
* Refactor `CidEncoder` based on `GsonEncoder` 
* Remove the need for an \*explicit* `type` field CID model objects. 
* Use consistently `CidEncoder` in place of `JsonOutput`.

modules/nf-commons/src/main/nextflow/serde/gson/RuntimeTypeAdapterFactory.java
```diff
@@ -0,0 +1,342 @@
...
+ * }
+ * }</pre>
+ *
+ * This class addresses this problem by adding type information to the serialized JSON and honoring
+ * that type information when the JSON is deserialized:
+ *
+ * <pre>{@code
+ * {
+ *   "bottomShape": {
+ *     "type": "Diamond",
```
jorgee, line 97: Is this long comment required? It is already in the link https://github.com/google/gson/blob/main/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java

pditommaso: Just copy & pasted from the original source. better to keep as it for avoidance of doubt of what it changed. In any case does not hurt

PR #5894, opened 2025-03-18
Fix false "process already defined" error

bentsherman:
This PR fixes a regression caused by #4613 .

The strict syntax requires some new include logic because it includes modules in a slightly different way. My original change was minimal but had a side-effect on the v1 script loader. So I went back and separated the logic more explicitly.

This change is easy to confirm with sarek:
```bash
make pack
./build/releases/nextflow-25.02.0-edge-dist inspect nf-core/sarek -r 3.4.2
```


pditommaso:
Nice, running e2e tests vs prod

pditommaso (approved):
Awesome

PR #5907, opened 2025-03-20
Extend observer with onFileStage

### LINKED ISSUE #5905: Detect when files are being staged
## New feature

## Use case 

## Suggested implementation 

```groovy
trait TraceObserver {
}
```

```groovy
enum FileTransferStatus {
    STARTED,
    FINISHED
}
trait TraceObserver {
        }
    }
}
```

rcannood:
Closes #5905


bentsherman:
Before we go down this road, we should see if we can simply preserve the remote path here:

https://github.com/nextflow-io/nextflow/blob/36153c68ffb11685d7085737404694ee613bdac8/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy#L1943

Then you should be able to use the `onProcessPending` event in the trace observer to get the original path via `handler.task.inputs`. That would be preferable to adding a new trace event

rcannood:
> Before we go down this road, we should see if we can simply preserve the remote path here:

Added to #5911 -- not sure whether this is what you had in mind.

Having said that, I wonder whether adding this function to the observer isn't also just a good addition, because it would make it super clear from the documentation how to get observe staging events.

bentsherman:
Closing in favor of #5911

PR #5909, opened 2025-03-20
Workflow outputs (third preview)

bentsherman:
This PR implements the third preview of the workflow output definition for Nextflow 25.04.

Changes are described in the docs, copied here for convenience:

- The `publish:` section can only be specified in the entry workflow.

- Workflow outputs in the `publish:` section are assigned instead of using the `>>` operator. The output name must be a valid identifier.

- By default, output files are published to the base output directory, rather than a subdirectory corresponding to the output name.

- The syntax for dynamic publish paths has changed. Instead of defining a closure that returns a closure with the `path` directive, the outer closure should use the `>>` operator to publish individual files.

- The `mapper` index directive has been removed. Use a `map` operator in the workflow body instead.

Changes not described in the docs:

- The `onWorkflowPublish` trace event has been modified to include the output name. It is emitted once for each output, rather than once for each channel value, and it is emitted regardless of whether the index file is enabled.


modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
```diff
@@ -123,11 +123,15 @@ trait TraceObserver {
...
+     * Method that is invoked when a workflow output is published.
      *
+     * @param name
+     *      The name of the workflow output
      * @param value
+     *      A list if the published channel was a queue channel,
+     *      otherwise an object if the channel was a value channel
      */
-    void onWorkflowPublish(Object value){}
+    void onWorkflowPublish(String name, Object value){}
```
bentsherman, line 134: @pditommaso regarding this change to `onWorkflowPublish`

Aside from the fact that it's a preview feature, there is a bigger issue I wanted to raise about the TraceObserver -- these overloads that we add to be backwards compatible don't actually work

Even if we add something like this:
```groovy
void onWorkflowPublish(Object value) {
    onWorkflowPublish(null, value)
}
```

If I build a plugin with Nextflow 24.10 and try to use it with 25.04, it will fail with an error like this:

```
ERROR ~ Receiver class nextflow.validation.ValidationObserver does not define or inherit an implementation of the resolved method 'abstract void onWorkflowPublish(java.lang.String,java.lang.Object)' of interface nextflow.trace.TraceObserver.
```

This is because the custom trace observer gets compiled against the 24.10 version of TraceObserver and run against the 25.04 version at runtime, but it doesn't receive the new method overload. I think it's a limitation of the Java/Groovy runtime.

So there is no point in adding these extra overloads. Plugins built with an older Nextflow will break no matter what we do

See: https://github.com/nextflow-io/nf-schema/issues/90

docs/workflow.md
```diff
@@ -538,35 +488,36 @@ workflow {
...
-            { filename -> "fastq/${sample.id}/${filename}" }
+            sample.fastq_1 >> "fastq/${sample.id}/"
+            sample.fastq_2 >> "fastq/${sample.id}/"
         }
     }
 }
 ```
 
-The inner closure will be applied to each file in the channel value, in this case `sample.fastq_1` and `sample.fastq_2`.
+Each `>>` specifies a *source file* and *publish target*. The source file should be a file or collection of files, and the publish target should be a directory or file name. If the publish target ends with a slash, it is treated as the directory in which source files are published. Otherwise, it is treated as the target filename of a source file. Only files that are published with the `>>` operator are saved to the output directory.
```
pditommaso, line 516: The publish target was not the output e.g. `samples` in this example? this looks to me more a "publish directory" or just "output directory"

pditommaso (commented):
While I find the syntax improved compared to the previous iteration, I think this version diluting the original idea of decoupling output publishing for the final output structure by having an intermediate model represented by "publish target" that were defined both at process and sub-workflow level. 

In this version this is essentially not possible any more, and the output needs to be wired channel by channel in the main workflow definition if I'm understanding correctly. 

Is there as nf-core pipeline adopting this version as reference?

bentsherman:
> I think this version diluting the original idea of decoupling output publishing for the final output structure by having an intermediate model represented by "publish target" that were defined both at process and sub-workflow level.
> 
> In this version this is essentially not possible any more, and the output needs to be wired channel by channel in the main workflow definition if I'm understanding correctly.

I've gone back and forth on this throughout each iteration. I was initially skeptical about making the user wire the outputs all the way to the top, because I thought it would be a ton of work for little value.

Two things made me change my mind:

1. In practice, workflows will have very few outputs. For example, even though rnaseq can run dozens of tools, it really only has a few outputs -- "samples", "summary" (e.g. multiqc), and maybe "counts matrix". You might have to `join` several outputs from different tools, but the rest of the wiring is pretty simple.

2. The original problems with `publishDir` are that (1) you have no high-level visibility of the workflow outputs and (2) you can't model output metadata, only files. Publishing channels instead of glob patterns solves (2), but if we still allow publishing from processes and subworkflows, we haven't solved (1).

As a user, I should be able to see all of my outputs in one place and easily trace them to upstream sources, which I can't do if any subcomponent can "contribute" to an output. It is analogous to people using params anywhere in their pipeline, instead of only in the entry workflow.

It does require writing a bit more code, but the improved readability is worth it. And I would point out that we spend much more time reading code than writing code, so we should design the language accordingly.

There is still an intermediate model, but instead of "publish target" it is called simply an "output" or "output declaration". It has similar syntax/semantics to a parameter.

> Is there as nf-core pipeline adopting this version as reference?

nf-core is reluctant to adopt any features as long as we keep them in "preview", which creates a minor chicken and egg problem for us 😅 

I have updated my [fetchngs PR](https://github.com/nf-core/fetchngs/pull/312/files#diff-6401496ba455b9488ffa902a6e4d7732b2c60ff2d77c5c3ef96b28a7ac7d3b28) to use this preview, so you can see how it would look there.

docs/workflow.md
bentsherman: I have changed the meaning of publish target in this iteration to match the overall syntactic/semantic changes

What was previously a "publish target" like `samples` is now simply an "output" or "output declaration".

Publish target now refers to the right-hand side of a publish statement e.g. `sample.fastq_1 >> 'fastq/'`

Essentially I have moved the concept of "publish target" into the path directive.

Basically every real-world use case I've seen when talking to users requires this fine-grained level of publishing, because of how they like to organize their output directory

modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
pditommaso: Yes, I saw the same in the past. I think it's caused by how groovy converts the trait into an interface behind the scene. 

We should consider to replace the trait with an interface + defaults method that were designed to handle this problem

pditommaso: Let's [see](https://github.com/nextflow-io/nextflow/pull/5963), locally seems not breaking anything quite incredibility

docs/reference/syntax.md
```diff
@@ -298,20 +298,20 @@ Enum types cannot be included across modules at this time.
 
 ### Output block
 
-The output block consists of one or more *target blocks*. A target block consists of a *target name* and one or more *target directives* for configuring the corresponding publish target:
+The output block consists of one or more *output declarations*. An output declaration consists of a name and one or more *output directives* for defining the output:
```
christopher-hakkaart, line 301:
```suggestion
The output block consists of one or more *output declarations*. Each output declaration consists of a name and one or more *output directives* for defining the output:
```

docs/workflow.md
```diff
@@ -14,49 +22,41 @@ workflow {
 }
 ```
 
-A *named workflow*, on the other hand, is a workflow that can be called from other workflows:
+### Parameters
+
+Parameters can be defined in the script with a default value that can be overridden from the CLI, params file, or config file. Params should only be used by the entry workflow:
```
christopher-hakkaart, line 27:
```suggestion
Parameters can be defined in the script with a default value that can be overridden from the CLI, params file, or config file. Parameters should only be used by the entry workflow:
```

docs/workflow.md
```diff
@@ -14,49 +22,41 @@ workflow {
...
-    else
-        bar(foo())
+    my_workflow()
 }
 ```
 
-:::{note}
-While params can also be used by named workflows, this practice is discouraged. Named workflows should receive their inputs explicitly through the `take:` section.
-:::
+The above example defines a workflow named `my_workflow` which can be called from another workflow as `my_workflow()`. Both `foo` and `bar` could be any other process or workflow.
```
christopher-hakkaart, line 55:
```suggestion
The above example defines a workflow named `my_workflow` which is called from another workflow as `my_workflow()`. Both `foo` and `bar` could be any other process or workflow.
```

docs/workflow.md
```diff
@@ -370,57 +368,55 @@ workflow {
...
 
-## Publishing outputs
+## Workflow outputs
 
 :::{versionadded} 24.04.0
 :::
 
 :::{versionchanged} 24.10.0
-A second preview version has been introduced. Read the [migration notes](#migrating-from-first-preview) for details.
+A second preview version was introduced. Read the [migration notes](#migrating-from-first-preview) for details.
```
christopher-hakkaart, line 377:
```suggestion
A second preview version was introduced. See the [migration notes](#migrating-from-first-preview) for details.
```

docs/workflow.md
```diff
@@ -370,57 +368,55 @@ workflow {
...
+:::
+
+:::{versionchanged} 25.04.0
+A third preview version was introduced. Read the [migration notes](#migrating-from-second-preview) for details.
```
christopher-hakkaart, line 381:
```suggestion
A third preview version was introduced. See [Migrating from second preview](#migrating-from-second-preview) for more details.
```

docs/workflow.md
```diff
@@ -433,11 +429,9 @@ nextflow run main.nf -output-dir 'my-results'
 outputDir = 'my-results'
 ```
 
-It defaults to `results` in the launch directory. All published outputs will be saved into this directory.
+It defaults to `results` in the launch directory.
```
christopher-hakkaart, line 432:
```suggestion
The default output directory is `results` in the launch directory.
```

docs/workflow.md
```diff
@@ -433,11 +429,9 @@ nextflow run main.nf -output-dir 'my-results'
...
 
-Each publish target is saved into a subdirectory of the output directory. By default, the target name is used as the directory name.
-
-For example, given the following publish targets:
+By default, all output files are published to this directory. Each output in the output block can define where files are published using the `path` directive. For example:
```
christopher-hakkaart, line 434:
```suggestion
By default, all output files are published to the output directory. Each output in the output block can define where files are published using the `path` directive. For example:
```

docs/workflow.md
```diff
@@ -538,35 +488,36 @@ workflow {
...
+Each `>>` specifies a *source file* and *publish target*. The source file should be a file or collection of files, and the publish target should be a directory or file name. If the publish target ends with a slash, it is treated as the directory in which source files are published. Otherwise, it is treated as the target filename of a source file. Only files that are published with the `>>` operator are saved to the output directory.
 
 ### Index files
 
-A publish target can create an index file of the values that were published. An index file preserves the structure of channel values, including metadata, which is simpler than encoding this information with directories and file names. The index file can be CSV (`.csv`) or JSON (`.json`).
+Each output can create an index file of the values that were published. An index file preserves the structure of channel values, including metadata, which is simpler than encoding this information with directories and file names. The index file can be CSV (`.csv`), JSON (`.json`), or YAML (`.yml`, `.yaml`). The channel values should be files, lists, or maps.
```
christopher-hakkaart, line 520:
```suggestion
Each output can create an index file of the values that were published. An index file preserves the structure of channel values, including metadata, which is simpler than encoding this information with directories and file names. The index file can be a CSV (`.csv`), JSON (`.json`), or YAML (`.yml`, `.yaml`) file. The channel values should be files, lists, or maps.
```

docs/workflow.md
pditommaso: Not sure it still makes sense to talk about "target" in this case. May be just saying "directory" it can result more clear. Otherwise I'm OK

pditommaso (approved):

docs/workflow.md
```diff
@@ -667,6 +603,38 @@ Additionally, the following options from the {ref}`workflow <config-workflow>` c
...
+
+The first preview of workflow publishing was introduced in 24.04. The second preview, introduced in 24.10, made the following breaking changes:
+
+- The process `publish:` section has been removed. Channels should be published only in workflows, ideally the entry workflow.
+
+- The `directory` output directive has been replaced with the `outputDir` config option and `-output-dir` command line option, which is `results` by default. The other directives such as `mode` have been replaced with config options under `workflow.output.*`.
+
+  In other words, only target blocks can be specified in the output block, but target blocks can still specify directives such as `mode`.
+
+- Target names cannot begin or end with a slash (`/`);
```
christopher-hakkaart, line 640:
```suggestion
- Target names cannot begin or end with a slash (`/`).
```

christopher-hakkaart (commented):
Added some minor comments. Feel free to tag me if you want me to look at it again.

docs/workflow.md
bentsherman: The right-hand side can refer to a file name or directory name, that is why I call it "target" instead of "directory"

PR #5911, opened 2025-03-21
Add origPath as an argument to FileHolder

rcannood:
Alternative solution to #5907 to implement #5905.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -1940,7 +1940,7 @@ class TaskProcessor {
             if( item instanceof Path || coerceToPath ) {
                 def path = normalizeToPath(item)
                 def target = executor.isForeignFile(path) ? batch.addToForeign(path) : path
-                def holder = new FileHolder(target)
+                def holder = new FileHolder(target, path)
                 files << holder
             }
             else {
--- a/modules/nf-commons/src/main/nextflow/file/FileHolder.groovy
+++ b/modules/nf-commons/src/main/nextflow/file/FileHolder.groovy
@@ -40,26 +40,31 @@ class FileHolder  {
 
     final String stageName
 
-    FileHolder( Path inputFile ) {
+    final Path origPath
+
+    FileHolder( Path inputFile, Path origPath = null ) {
         assert inputFile
         this.sourceObj = inputFile
         this.storePath = real(inputFile)
         this.stageName = norm(inputFile.getFileName())
+        this.origPath = origPath
     }
 
-    FileHolder( def origin, Path path ) {
+    FileHolder( def origin, Path path, Path origPath = null ) {
         assert origin != null
         assert path != null
 
         this.sourceObj = origin
         this.storePath = path
         this.stageName = norm(path.getFileName())
+        this.origPath = origPath
     }
 
-    protected FileHolder( def source, Path store, def stageName ) {
+    protected FileHolder( def source, Path store, def stageName, Path origPath = null ) {
         this.sourceObj = source
         this.storePath = store
         this.stageName = norm(stageName)
+        this.origPath = origPath
     }
 
     FileHolder withName( def stageName )  {
@@ -69,6 +74,8 @@ class FileHolder  {
     Path getStorePath() { storePath }
 
     String getStageName() { stageName }
+
+    Path getOrigPath() { origPath }
 
     @PackageScope
     static FileHolder get( def path, def name = null ) {


modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -1939,9 +1940,14 @@ class TaskProcessor {
...
-                def holder = new FileHolder(target)
-                files << holder
+                if (executor.isForeignFile(path)) {
+                    def target = batch.addToForeign(path)
+                    def holder = new RemoteFileHolder(target, path)
+                    files << holder
+                } else {
+                    def holder = new FileHolder(path)
+                    files << holder
+                }
```
bentsherman, line 1950: Why didn't you just do `new FileHolder(path, target)` ? You shouldn't need a new class

bentsherman: Sorry, just saw your extra comment

bentsherman: However I think my point still stands. You should be able to just do `new FileHolder(path, target)` and add a method to FileHolder to access the source path, e.g.:

```groovy
Path getSourcePath() {
  sourceObj instanceof Path ? sourceObj : null
}
```

rcannood: Np, I don't like the solution of introducing an extra class here either.

I just thought it'd be better to get your opinion first ^^

bentsherman:
Looks good. Thanks for your help 😄 

Do you understand how to use this in your setup? I will also try it with nf-prov. Let me know if you need help with how to test on your end

rcannood:
I wonder whether something needs to be updated in the constructors:

```groovy
    FileHolder( Path inputFile ) {
        assert inputFile
        this.sourceObj = inputFile
        this.storePath = real(inputFile)
        this.stageName = norm(inputFile.getFileName())
    }
    FileHolder( def origin, Path path ) {
        assert origin != null
        assert path != null
        this.sourceObj = origin
        this.storePath = path // ← does this need to become `real(path)`?
        this.stageName = norm(path.getFileName())
    }
    protected FileHolder( def source, Path store, def stageName ) {
        this.sourceObj = source
        this.storePath = store // ← does this need to become `real(store)`?
        this.stageName = norm(stageName)
    }
```

---

Cfr the use case: I'm working on a [nf-lamin](https://github.com/laminlabs/nf-lamin/tree/plugin?tab=readme-ov-file#usage) plugin which allows users to run pre-existing Nextflow workflows and create records in the LaminDB regarding:

* Which workflow was run
* Information about the run itself
* List of input files (e.g. s3://path/to/input/*.fastq.gz)
* List of output files (e.g. s3://path/to/output/*.bam)

This package will use the observer to detect various events and then pass that information on to LaminDB.

pditommaso:
I think you should have a look at this on-going effort https://github.com/nextflow-io/nextflow/pull/5715

bentsherman:
I think this PR will benefit all of us -- the provenance store, nf-prov, and nf-lamin. Fundamentally it's about being able to trace staged files to their remote source, which we currently cannot do

The only snag is that this PR currently will change the task hash. When a task has a foreign input file, the locally staged path is used for the hash, not the remote path. This might actually be better, as it would decouple the task hash from the foreign file staging. But I wonder if the locally staged path was used for a reason

bentsherman:
I have it working with nf-prov! https://github.com/nextflow-io/nf-prov/pull/45

You can see the before/after in the linked PR. Turns out it doesn't affect the task hash. I think this is because the FilePorter uses the original URL to construct the hash for the stage directory, and since the file is hashed by either size+timestamp (standard) or contents (deep), the remote file and staged file produce the same hash

@rcannood to your point about doing `real(path)`, that function is mainly used to handle inputs that came from an upstream task, so it should not be needed here.

rcannood:
I see, thanks!

@bentsherman Any updates on the state of this PR? :bow:

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -1942,7 +1942,7 @@ class TaskProcessor {
             if( item instanceof Path || coerceToPath ) {
                 def path = normalizeToPath(item)
                 def target = executor.isForeignFile(path) ? foreignFiles.addToForeign(path) : path
-                def holder = new FileHolder(target)
+                def holder = new FileHolder(path, target)
                 files << holder
```
pditommaso, line 1946: Checking this, and I'm bit concerned on the different logic of the two constructors 

```
    FileHolder( Path inputFile ) {
        assert inputFile
        this.sourceObj = inputFile
        this.storePath = real(inputFile)
        this.stageName = norm(inputFile.getFileName())
    }

    FileHolder( def origin, Path path ) {
        assert origin != null
        assert path != null

        this.sourceObj = origin
        this.storePath = path
        this.stageName = norm(path.getFileName())
    }
```

Above all, note the different logic for `storePath`. I believe this will change the handle on symlink above all. 

If the goal is to bring the source path with could try using the first constructor and adding a method to update the source path

rcannood: I understand your concern.

Would you rather this get implemented as follows?

```suggestion
                def holder = new FileHolder(target)
                holder.setSourceObj(origin)
```

pditommaso:
I've tried adding this 

```
    FileHolder withSourcePath(Path source) {
        new FileHolder(source, this.storePath, this.stageName)
    }
```

The problem however is that this invalid existing caches for pipelines depending on staged files

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
bentsherman: Any reason why we can't resolve symlinks in the two-arg constructor?

```groovy
    FileHolder( def origin, Path path ) {
        this.sourceObj = origin
        this.storePath = real(path)
        this.stageName = norm(path.getFileName())
    }
```

bentsherman:
> The problem however is that this invalid existing caches for pipelines depending on staged files

@pditommaso how did you test this?

bentsherman:
From our discussion today, we will merge after 25.04 to give plenty of time to test in the edge releases

rcannood:
> From our discussion today, we will merge after 25.04 to give plenty of time to test in the edge releases

Unfortunate, but understandable :)

Just to check on this, does something still need to be fixed or review after the latest few commits? :bow:

bentsherman:
> Considering your requirements it may be better to expose task input and output files in the [TaskEvent](https://github.com/nextflow-io/nextflow/blob/b58aec1709dceb8cea559d56d69d3b8983d61918/modules/nextflow/src/main/groovy/nextflow/trace/event/TaskEvent.groovy#L31-L31) class

Regardless of whether we do this, we still need to preserve the source path in `makeTaskContextStage1()` / `makeTaskContextStage2()`, as it is not saved anywhere else in the TaskRun. Exposing it through a better interface like `TaskEvent` is essentially a separate problem as the TaskEvent is downstream of the TaskRun.

> The problem remaining is that even changing only the sourceObj with an ad-hoc method, that attribute is the one used to compute the checksum for the input files

@pditommaso this PR does not change the sourceObj or the hashing strategy, so there should be no issue with cache invalidation

bentsherman:
Merging so that we can battle-test in the upcoming edge releases.

pditommaso:
Sorry folks, I had reverted this e0f5b260. I had explained above why this is a risky change hard to test if not hpc cluster. Also I believe FileHolder is an internal class that should not be used a public interface for this kind of use cases

PR #5913, opened 2025-03-21
nf-lang: Add type checking utils

bentsherman:
- Add basic type checking utilities from language server
- Cleanup implicit declaration analysis
- Cleanup process/workflow output name checking

### DIFF
--- a/modules/nf-lang/src/main/antlr/ConfigParser.g4
+++ b/modules/nf-lang/src/main/antlr/ConfigParser.g4
@@ -238,7 +238,7 @@ multipleAssignmentStatement
     ;
 
 assignmentStatement
-    :   left=expression nls
+    :   target=expression nls
         op=(ASSIGN
         |   ADD_ASSIGN
         |   SUB_ASSIGN
@@ -254,7 +254,7 @@ assignmentStatement
         |   POWER_ASSIGN
         |   ELVIS_ASSIGN
         ) nls
-        right=expression
+        source=expression
     ;
 
 // -- expression statement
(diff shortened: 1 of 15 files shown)


bentsherman:
@pditommaso it looks like I still can't merge without at least 1 approval, even though this PR is contained within nf-lang. It requests a review from @nextflow-io/lang as code owner, but I still can't merge without approval and I can't approve my own PR. So it seems like being a code owner doesn't actually give me permission to push to master

pditommaso (commented):
Would not be worth adding a unit test?

pditommaso:
If you are in hurry, I can merge this, however remain a good practice in include tests new or existing components. I understand this is impacting existing functionality, but it will at some points.

bentsherman:
Putting the tests aside for a moment, you shouldn't have to approve/merge this at all. I should be able to do it myself since it only touches nf-lang

pditommaso:
I think one approver requires someone different from the PR creator, irrespective of module ownership

bentsherman:
Yeah I can't approve my own PR even though I'm a code owner. We might need to make some exception, either for my username or for the lang team. Not sure what options are available in the settings

pditommaso:
At this point, make more sense get rid of the "lang" team and include more approvers in the pr

bentsherman:
Yeah it seems like the code owners thing isn't doing anything for us. But I need to be able to update nf-lang without approval for the time being. That was one of my requirements for moving nf-lang into the nextflow repo

pditommaso (approved):

pditommaso:
You should be able to merge now

bentsherman:
Thanks. I will add some unit tests in a separate PR

PR #5915, opened 2025-03-23
Fancy ANSI formatting for `lint` and `format` log outputs

ewels:
Nicer output from the new `nextflow format` and `nextflow lint` commands. @bentsherman you will probably want to clean up my crappy Groovy code, but I'm happy with the outputs now.

TODO:

- [ ] Replicate the `.gitignore` exclusions to `format`
- [ ] Refactor code to avoid duplication
- [ ] @bentsherman to clean up my crappy groovy code
- [ ] _[Maybe]: Improve `.gitignore` code to work when launching from anywhere in the project tree_

# Format

https://github.com/user-attachments/assets/a44a7049-689c-4e49-a953-16198f1963af

# Lint

https://github.com/user-attachments/assets/c73fc2bb-8f09-4125-99e1-86d985d572ae

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdFormat.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdFormat.groovy
@@ -187,7 +187,7 @@ class CmdFormat extends CmdBase {
         for( final message : errorMessages ) {
             if( message instanceof SyntaxErrorMessage ) {
                 final cause = message.getCause()
-                term.fg(Color.RED).a(Attribute.INTENSITY_BOLD).a("error").fg(Color.DEFAULT).a(": ")
+                term.fg(Color.RED).bold().a("error").fg(Color.DEFAULT).a(": ")
                 term.a("Failed to parse ${source.getName()}").a(Attribute.INTENSITY_BOLD_OFF)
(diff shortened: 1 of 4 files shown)


ewels:
Added some more functionality:

* Ability to output the lint results in different formats (full, concise, summary, json)
* Ability to `-exclude` one or more files / dirs / glob patterns
* Parsing and use of a `.gitignore` file if present

The `.gitignore` stuff is only in `lint` for now - it should be in both but I don't want to just copy it over, as there is more and more duplication between the two CLI files now. But I'm not really sure where I should put this shared code?

ewels:
Wonder if we can set up GitHub actions to use the JSON output to spit out [error messages](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-error-message) for inline PR notes about lint errors 👀 

I read somewhere that it's only possible to do 10 at a time from GitHub Actions (GitHub apps more), many links seem outdated, so may need a bit of testing to see how viable it is.

bentsherman:
- Remove `.gitignore` logic for now until we decide on an overall strategy (see #5934)
- Use "error listener" interface to separate output formats from core linting logic
- Re-use the ANSI error listener in the script loader
- Move file exclusion checker into nf-lang (it will be re-used by the language server)

Issue #5855, opened 2025-03-05
Hierarchical exported Mermaid DAG has cycles


kdesnos:
### Expected behavior and actual behavior

An (acyclic) hierachical DAG created with the nextflow DSL is exported into a DAG that contains a cycle.

This issue is related to #5843 which causes several nested workflows to be improperly exported into a mermaid graph when a given `sub_sub` workflow is instantiated by two distinct `sub1` and `sub2` workflows in the hierarchy. This causes all processes of the `sub2::sub_sub` workflow to be interpreted as being part of the `sub1::sub_sub` instance.

In this additional issue, a `combine` channel operator within the `sub_sub` workflow, which should appear twice in the exported DAG (as `sub1::sub_sub::combine` and `sub2::sub_sub::combine`, actually appears only once in the exported diagram. As a result, a cycle can be created in the exported DAG if the `sub2::sub_sub::combine` operator takes as an argument a channel resulting from `sub1:sub_sub`. To trigger the issue, a "global" channel with a first "combine" operator needs to be there (see following example).

### Steps to reproduce the problem

In this case, I'm defining two simple workflows:
* `main.nf`
* `sub.nf`

`sub.nf`: 
``` 
process a {
    input:
        val something

    output:
        stdout

    script: 
    """
        echo "I'm a $something"
    """
}

process b {
    input:
        val in

    output:
        stdout
    script:
    """
        echo "I'm a process me too"
    """
}

 // Without this "global" variable combine, everything works fine.
outchan0 = Channel.fromList([2000])
outchan1 = Channel.fromList([3000])
outchan2 = outchan1.combine(outchan0)

workflow sub_sub {
    take:
        sub_sub_input
    main:        
        o = a(sub_sub_input)
        o2 = b(o.combine(outchan2))
    emit:
        o = o2
}

workflow sub {
    take:
        sub_input
    main:
        ou = sub_sub(sub_input).o
    emit:
        ou = ou
}
```

`main.nf`:
```
include { sub as sub_1; sub as sub_2 } from './sub'


workflow {
    ch = Channel.fromList([1000])

    res1 = sub_1(ch).ou
    sub_2(res1)
}
```

### Program output 

The resulting cyclic DAG:
```mermaid
flowchart TB
    subgraph " "
    v0["Channel.fromList"]
    v1["Channel.fromList"]
    v3["Channel.fromList"]
    end
    subgraph sub_1
    subgraph sub_sub
    v4([a])
    v6([b])
[...]

PR #5918, opened 2025-03-25
Mermaid DAG export: Fix operator infer subgraph key

kdesnos:
# Summary
Fix and improve the corrupted export of DAGs in the mermaid format.
Fix issue #5855.

# Issue fixed
The issue comes from the way the _key_ of the hierarchical subgraphs for operators are inferred in the mermaid DAG printer.
Unlike regular process, the DAG given to the printer does not contain the complete hierarchical path (i.e. the _key_) to the channel operators. Because of this, the printer attempts to recover this information automatically prior to printing the subgraph, so as to print operators in the scope they were originally declared. 

The current version of the code fails to recover the correct subgraphs in many situations, as it assumes that:
* All interconnected channel operators share the same subgraph.
* The subgraph associated to a (group of) channel operator is an input (or output) process of the (first) operator (of the group).

Because of this issue, in the non-verbose mode where the DAG printer attempts to collapse all connected channel operators into a single operators, the failed association of operators to the right subgraph may lead to the creation of cycles in the exported DAG, as exemplified in #5855.

Unfortunately, as described in issue #5855, finding the original scope might not always be possible in ambiguous situations where two distinct pipelines lead to the exact same DAG.

# How
Here is a list of the changes made in the code:
* **Infer keys at the start**:  Because in the non-verbose mode, operators are collapsed at the start of the printing tasks, the channel operator subgraph keys need to be known from the start. A call to the `inferSubgraphKeys()` function is made at the start of the printer, and its result is stored in the `operatorSubgraphKeys` map for later use/update by the `collapseOperators()` and the `getNodeTree()` functions.
* **`collapseOperators()`**: Instead of assuming that all connected operators can be collapsed, the function now limit the collapsing to connected operators from the same subgraph. 
*  **`collapseSubgraph()`**: Function now returns the Node resulting from the collapse to make it possible to update the `operatorSubgraphKeys` map when a subgraph of operators is collapsed.
* **`getNodeTree`**:  Take the pre-computed subgraph keys as an input.
* **`inferSubgraphKeys()`**: The subgraph of each operator is searched individually, as interconnected operators may be in distinct subgraphs. The algorithm used to infer the correct subgraph is completely redesigned with three distinct strategies applied successively. 
  1. [Low complexity] Assuming the DAG nodes are (mostly) in topological order, if an operator is surrounded by two processus in the list which share a common subgraph key, then this operator must be in the same subgraph.
  2. [Med complexity] If an operator contains process of the same subgraph in both its input and output processus, then the operator must be in the same subgraph, or in a lower level of hierarchy nested within this subgraph. In case several pairs of processus satisfy this predicate, the most nested pair is used to determine the operator's subgraph.
  3. [Low complexity] Legacy technique associating the operator to an input or output process' subgraph.

The code is a bit longer than before, but the runtime remains negligible for a reasonably large DAG with several hundreds of nodes.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/dag/DAG.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/dag/DAG.groovy
@@ -30,6 +30,7 @@ import nextflow.NF
 import nextflow.extension.CH
 import nextflow.extension.DataflowHelper
 import nextflow.processor.TaskProcessor
+import nextflow.script.ExecutionStack
 import nextflow.script.params.DefaultInParam
 import nextflow.script.params.DefaultOutParam
 import nextflow.script.params.EachInParam
@@ -159,7 +160,8 @@ class DAG {
      */
     @PackageScope
     Vertex createVertex( Type type, String label, Object extra=null ) {
-        def result = new Vertex(type, label)
+        final workflow = ExecutionStack.workflow()?.name ?: ""
+        final result = new Vertex(type, label, workflow)
         if( extra instanceof TaskProcessor ) {
             result.process = extra
             result.operators = [ extra.operator ]
@@ -388,6 +390,11 @@ class DAG {
         String label
 
         /**
+         * The name of the enclosing workflow
+         */
+        String workflow
+
+        /**
          * The vertex type
          */
         Type type
@@ -409,10 +416,12 @@ class DAG {
          *
          * @param type A {@link Type} value
          * @param label A descriptive string to label this vertex
+         * @param workflow The name of the enclosing workflow
          */
-        Vertex( Type type, String label = null ) {
+        Vertex( Type type, String label = null, String workflow = null ) {
             assert type
             this.label = label
+            this.workflow = workflow
             this.type = type
         }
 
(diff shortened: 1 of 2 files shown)


kdesnos:
## Minor additional change:
I realized that as a result of #5860, parent and subgraph names were separated with a "::". To remain coherent with process path printed in other reports, the ":" separator should be used instead.

bentsherman:
@kdesnos thank you for looking into this. I always knew there were edge cases with this heuristic that I never bothered to chase down. You inspired me to finally solve the root problem -- saving the workflow context in the DAG itself.

It ended up being simpler than I expected. Basically, when a node is added to the DAG during the script execution, Nextflow knows the fully-qualified workflow context in which it's currently running. So we just save that to the DAG. This removes the need for any more heuristics to recover something when we can just save it from the beginning.

I tried it on your example and it works. Can you try my changes and let me know if they work on your end?

kdesnos:
Hi @bentsherman,
This works great, and is a much more straightforward solution indeed. I also had this option in mind, but in case the PR was not merged, I wanted to keep the possibility to have my DAG renderer as a plugin 😄.

Thanks for the time spent on this :)

bentsherman (approved):

PR #5711, opened 2025-01-24
Improve git tests

edmundmiller:
Preliminary #4427 for #5610

PR #5927, opened 2025-03-26
Update git tests

edmundmiller:
An updated attempt and cleaned up from #5711 

This is a PR that's pre-work for https://github.com/nextflow-io/nextflow/issues/4427. 

I cleaned up some of the tests in #5610 and @tom-seqera suggested splitting them into separate PRs

> While I agree with efforts to improve the naming and clarity of the tests, doing so alongside changes to behaviour introduces a lot of noise. I'd suggest splitting such changes into a separate PR which improves the tests without modifying any behaviour.

I'll follow this up with a PR introducing the features for #4427 after.

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
@@ -665,4 +665,173 @@ class AssetManagerTest extends Specification {
         !AssetManager.isRemoteBranch(local_master)
     }
 
+    @Tag("config")
+    def 'should handle manifest with gitmodules configuration'() {
+        given:
+        def config = '''
+            manifest {
+                gitmodules = ['module1', 'module2']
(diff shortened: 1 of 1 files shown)


modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
```diff
@@ -338,35 +337,22 @@ class AssetManagerTest extends Specification {
...
-
-        when:
-        manager = [:] as AssetManager
-        manager.createHubProvider('xxx')
-        then:
-        thrown(AbortOperationException)
+        repo instanceof RepositoryProvider
 
+        where:
+        providerName << ['github', 'bitbucket', 'gitlab']
```
pditommaso, line 351: It's not validating that for a given provider name is creating the expected provider class

pditommaso:
Looks like this pull request lost the momentum. Closing for now.

Issue #4669, opened 2024-01-17
Workflow inputs definition and schema


bentsherman:
Spun off from #2723 

Params can currently be defined in config files (including profiles), params files, CLI options, and the pipeline code itself. This creates the potential for much confusion around how these various sources are resolved (see #2662). Additionally, params are not typed, and while the CLI can cast command line params based on regular expressions, it can also backfire when e.g. a string param is given a value that "looks" like a number.

Instead, params should be defined in a single place with metadata such as type, default value, description, etc. Benefits are:
- single source of truth
- less ambiguity of how params are resolved
- ability to validate params based on type definition instead of regex guessing

The nf-core parameter schema (`nextflow_schema.json`) as well as the nf-validation plugin are excellent steps in this direction, and the solution may be to simply incorporate them into Nextflow.

For backwards compatibility, we may allow params to be set in config files and pipeline code, but this would essentially be overriding the default value rather than "defining" the param, and it should be discouraged in favor of putting everything in the parameter schema. That being said, it can be useful to set params from a profile, such as a test profile that provides some test data, so this use case should be supported.

The main question that I see is whether the schema should be in a separate JSON/YAML file (as it currently is in nf-core) or in the pipeline code as part of the top-level workflow definition.

- I would like the latter approach because it makes the top-level workflow more of a "unit" and makes it easier for IDE tooling to validate param references in the pipeline code. It would likely be less verbose than a JSON schema.
- On the other hand, a JSON schema can be parsed by external tools written in other languages whereas Nextflow scripts can only be parsed by Nextflow (and any IDE tooling)
- For what it's worth, Nextflow could export the workflow inputs definition to a JSON file for use with other tools, but then we have to keep it in sync with the pipeline code somehow

jspaezp:
Hello there! I wanted to bring this project to the attention!
https://pkl-lang.org

It was released by apple so I will have some support and I feel like it addresses really well the needs for schema validation and progressive amendment (+java library).

let me know what you think!

bentsherman:
What I've been thinking is that the `nextflow_schema.json` should be the source of truth instead of the config file, but config profiles should still be able to override the default value. I think that would give the best of both worls.

As for your R example, I think the best practice is to encode that convention in the process that calls the R script like so:

```groovy
"""
Rscript script.R --val ${params.Rscript_val ?: 'NA'}
"""
```

bentsherman:
[...]

PR #5929, opened 2025-03-31
Workflow params

bentsherman:
Related to #4669

This PR is the first step towards defining workflow params entirely in the pipeline script. It allows you to define a `params` block like so:

```groovy
params {
  input: Path
  save_intermeds: Boolean = false
}

workflow {
  // ...
}
```

Instead of assigning individual params. The advantage is that Nextflow can validate params natively once the `params` block is defined, because it guarantees that all params are declared in one place.

There is still some work required to make the validation work, but the high-level flow is:
1. User specifies params on the command line / params file
2. Config files can override script params or define "config params" which are only used by the config
3. When the params block is defined in the script, config params are ignored and only overrides from the command line / config are applied. If a script param was not specified and has no default value, an error is reported. If a CLI param was not already defined in the config or script, an error is reported

TODO:
- [x] Separate config params from CLI params to identify invalid params (i.e. params that weren't declared in the script or config)

### DIFF
--- a/docs/vscode.md
+++ b/docs/vscode.md
@@ -26,7 +26,7 @@ The language server parses scripts and config files according to the {ref}`Nextf
 
 When you hover over certain source code elements, such as variable names and function calls, the extension provides a tooltip with related information, such as the definition and/or documentation for the element.
 
-If a [Javadoc](https://en.wikipedia.org/wiki/Javadoc) comment is defined above a workflow, process, or function, the extension will include the contents of the comment in hover hints. The following is an example Javadoc comment:
+If a [Javadoc](https://en.wikipedia.org/wiki/Javadoc) comment is defined above a workflow, process, function, or parameter in a `params` block, the extension will include the contents of the comment in hover hints. The following is an example Javadoc comment:
 
 ```nextflow
 /**
--- a/docs/workflow.md
+++ b/docs/workflow.md
@@ -22,7 +22,60 @@ workflow {
 }
 ```
 
-### Parameters
+(workflow-params-def)=
+
+## Parameters
+
+Parameters can be declared in a Nextflow script with the `params` block or with *legacy* parameter declarations.
+
+### Params block
+
+:::{versionadded} 25.10.0
+:::
+
+:::{note}
+This feature requires the {ref}`strict syntax <strict-syntax-page>` to be enabled (`NXF_SYNTAX_PARSER=v2`).
+:::
+
+A script can declare parameters using the `params` block:
+
+```nextflow
+params {
+    // Path to input data.
+    input: Path
+
+    // Whether to save intermediate files.
+    save_intermeds: Boolean = false
+}
+```
+
+The following types can be used for parameters:
+
+- {ref}`stdlib-types-boolean`
+- {ref}`stdlib-types-float`
+- {ref}`stdlib-types-integer`
+- {ref}`stdlib-types-path`
+- {ref}`stdlib-types-string`
+
+Parameters can be used in the entry workflow:
+
+```nextflow
+workflow {
+    analyze(params.input, params.save_intermeds)
+}
+```
+
+:::{note}
+As a best practice, parameters should only be used directly in the entry workflow and passed to workflows and processes as explicit inputs.
+:::
+
+The default value can be overridden by the command line, params file, or config file. Parameters from multiple sources are resolved in the order described in {ref}`cli-params`. Parameters specified on the command line are converted to the appropriate type based on the corresponding type annotation.
+
+A parameter that doesn't specify a default value is a *required* param. If a required param is not given a value at runtime, the run will fail.
+
+(workflow-params-legacy)=
+
+### Legacy parameters
 
 Parameters can be declared by assigning a `params` property to a default value:
 
@@ -37,10 +90,6 @@ workflow {
         analyze(fake_input(), params.save_intermeds)
 }
 ```
-
-:::{note}
-As a best practice, `params` should be used only in the entry workflow and passed to workflows and processes as explicit inputs.
-:::
 
 The default value can be overridden by the command line, params file, or config file. Parameters from multiple sources are resolved in the order described in {ref}`cli-params`.
 
--- a/docs/migrations/25-10.md
+++ b/docs/migrations/25-10.md
@@ -8,6 +8,31 @@ This page summarizes the upcoming changes in Nextflow 25.10, which will be relea
 This page is a work in progress and will be updated as features are finalized. It should not be considered complete until the 25.10 release.
 :::
 
+## New features
+
+<h3>Workflow params</h3>
+
+The `params` block is a new way to declare pipeline parameters in a Nextflow script:
+
+```nextflow
+params {
+  // Path to input data.
+  input: Path
+
+  // Whether to save intermediate files.
+  save_intermeds: Boolean = false
+}
+
+workflow {
+  println "params.input = ${params.input}"
+  println "params.save_intermeds = ${params.save_intermeds}"
+}
+```
+
+This syntax allows you to declare all parameters in one place with explicit type annotations, and it allows Nextflow to validate parameters at runtime.
+
+See {ref}`workflow-params-def` for details.
+
 ## Enhancements
 
 <h3>New syntax for workflow handlers</h3>
@@ -73,5 +98,7 @@ This feature addresses previous inconsistencies in timestamp representations.
 - The AWS Java SDK used by Nextflow was upgraded from v1 to v2, which introduced some breaking changes to the `aws.client` config options. See {ref}`the guide <aws-java-sdk-v2-page>` for details.
 
 ## Deprecations
+
+- The legacy type detection of CLI parameters is disabled when using the strict syntax (`NXF_SYNTAX_PARSER=v2`). {ref}`Legacy parameters <workflow-params-legacy>` in the strict syntax should not rely on legacy type detection. Alternatively, use the new `params` block to convert CLI parameters based on their type annotations. Legacy type detection can be disabled globally by setting the environment variable `NXF_DISABLE_PARAMS_TYPE_DETECTION=true`.
 
 - The use of workflow handlers in the configuration file has been deprecated. You should define workflow handlers in the pipeline script or a plugin instead. See {ref}`config-workflow-handlers` for details.
--- a/docs/reference/syntax.md
+++ b/docs/reference/syntax.md
@@ -28,7 +28,8 @@ A Nextflow script may contain the following top-level declarations:
 - Shebang
 - Feature flags
 - Include declarations
-- Parameter declarations
+- Params block
+- Parameter declarations (legacy)
 - Workflow definitions
 - Process definitions
 - Function definitions
@@ -107,9 +108,22 @@ The following definitions can be included:
 - Processes
 - Named workflows
 
-### Parameter
+### Params block
 
-A parameter declaration is an assignment. The target should be a pipeline parameter and the source should be an expression:
+The params block consists of one or more *parameter declarations*. A parameter declaration consists of a name and an optional default value:
+
+```nextflow
+params {
+    input: Path
+    save_intermeds: Boolean = false
+}
+```
+
+Only one params block may be defined in a script.
+
+### Parameter (legacy)
+
+A legacy parameter declaration is an assignment. The target should be a pipeline parameter and the source should be an expression:
 
 ```nextflow
 params.message = 'Hello world!'
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -121,6 +121,16 @@ class Session implements ISession {
     ScriptBinding binding
 
     /**
+     * Params that were specified on the command line.
+     */
+    Map cliParams
+
+    /**
+     * Params that were specified in the configuration.
+     */
+    Map configParams
+
+    /**
      * Holds the configuration object
      */
     Map config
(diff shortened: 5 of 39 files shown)


ewels:
> This concise syntax will work only if we're certain we don't need anything else in the script.

I'm not entirely sure that this is the case, the schema is used for validation of more than just type. I know that some of these things can be handled with Records (eg. `enum` choices), but what about things like `pattern`, `min`/`max` and [`uniqueItems`](https://nextflow-io.github.io/nf-schema/latest/nextflow_schema/nextflow_schema_specification/#uniqueitems) etc?

> I thought maybe the help text would be useful for CLI help, but that could be provided through a Javadoc comment

I'd be curious to see how this might look - ideally for both `description` and `helptext` in one.

bentsherman:
Using the Javadoc comment is "better" in the sense that you only need to parse the script to produce the CLI help, you don't have to execute it, which i think would be excessive

bentsherman:
See also Pydantic: https://docs.pydantic.dev/latest/concepts/fields/#validate-default-values

Simple param:

```groovy
myparam: String = "default-value"
```

Full param:

```groovy
myparam: String = Field(default="default-value", pattern="/some.*regex/")
```

bentsherman:
Declaring params in the entry workflow means that you don't need the `params.` prefix anymore:

```groovy
workflow {
  params:
  input: Path
  save_intermeds: boolean = false

  main:
  println "input = ${input}"
  println "save_intermeds = ${save_intermeds}"
}
```

On the one hand I like that it makes params more like workflow takes. On the other hand, you still need the `params.` prefix to use params in the config, so I fear that the end result would just be more confusing?

That would suggest that the `params` block is needed just for consistency with the config. Maybe we could allow the short and long forms like Pydantic:

```groovy
params {
  input: Path {
    description '...'
    pattern '*.csv'
  }
  save_intermeds: boolean = false
}

workflow {
  println "input = ${params.input}"
  println "save_intermeds = ${params.save_intermeds}"
}
```

Though I always hesitate to add shortcuts if it makes the code less consistent

ewels:
+1 for the separate `params` block, I feel like for consistency that is easier to read and understand, also avoids confusion with `take`, which is doing quite a similar thing.

Not sure about the squiggly bracket syntax. I like the thinking, but it means that we now have three different types of syntax for them. Nothing for config, `:` for types and `=` for variables / others. I can see that being really annoying.

That said, the `Field()` syntax can be confusing in its own ways, see the [Pydantic docs](https://docs.pydantic.dev/latest/concepts/fields/):

> Using the `f: <type> = Field(...)` form can be confusing and might trick users into thinking f has a default value, while in reality it is still required.

bentsherman:
Yeah, confusing the `Field()` with a default value is a serious drawback

The block syntax appeals to me because it is consistent with workflow outputs:
```groovy
// fetchngs...
outputs {
  samples: Channel<Sample> {
    path '...'
    // ...
  }
}

// rnaseq...
params {
  input: List<Sample> {
    // ...
  }
}
```

Since we want to be able to match outputs to inputs for pipeline chaining, it makes sense to me that the syntax for inputs and outputs mirror each other.

The config is another issue. Let me think through that and write a separate comment...

bentsherman:
We could add the same params block syntax to config files if consistency is an issue. But I fear this might feel too "weird" in a configuration context.

The nice thing about config params is that they basically have to be simple values (numbers, strings, booleans, etc). So being able to declare the type isn't so important because it can be inferred from the default value.

Meanwhile, if we take the hybrid approach of generating a skeleton schema that the user can annotate manually as needed, we don't need to add new syntax to the config file to support things like validation and help text, because those can just be defined in the JSON schema

kenibrewer:
I really like this syntax proposal a lot:

```
params {
  input: Path {
    description '...'
    pattern '*.csv'
  }
  save_intermeds: boolean = false
}
```

This is likely in part because it mirrors Python/Pydantic but I think that's a good thing for us to emulate. Imitating the syntax of the most popular typed python extension will make it easier for folks to learn and feel like they can read and understand.

ewels:
If we're leaning more towards that syntax, we could arguably have _all_ JSON schema parameters covered. I think we need to make sure we're crystal clear on what we want to support.

For example, maybe no `description` as that's "decorative" and too verbose, so the example above becomes simply:

```nextflow
params {
  input: Path { pattern '*.csv' }
  save_intermeds: boolean = false
}
```

ewels:
@bentsherman - any ideas how the schema builder might be able to reach into Nextflow code to update these definitions based on changes made in the GUI / JSON?

bentsherman:
I don't think that will be possible. I think the flow will have to be:
1. Write params definition in main script
2. Generate skeleton JSON schema
3. Extend the schema by hand or via CLI wizard or upload to schema builder
4. Don't edit things that are sourced from the main script

Or if you use the schema builder from scratch, you have to update the params definition by hand (or not use it at all)

bentsherman:
The kind of discovery work you describe (i.e. figuring out the appropriate type of a param) is exactly the kind of thing that needs to happen in the pipeline code, so that the Nextflow compiler can verify param names and types as they are used in the entry workflow. You can't get that kind of validation in the schema builder.

Instead, the schema builder should be used to annotate a fixed set of params with things like help text, icons, form validation rules, etc. It should be primarily concerned with how params are accessed by external systems.

ewels:
I agree that the logic / definition should be in the pipeline code: at least, that should be the source of truth. I was mostly wondering if we could have some way to update the Nextflow code from the JSON schema builder, to have the best of both worlds. The schema builder has some nice beginner-friendly functionality in it, for example a GUI with a built-in regex tester for writing patterns, and a bunch of built-in help text.

Maybe this is something that we could do with @reftrace ? eg. From the Python CLI that launches the schema builder GUI, then go back and access the Nextflow code to edit it in place. Not sure if that's possible.

Or if we launch the schema GUI editor from Nextflow itself (with a local server etc) could there be a callback which is able to edit the Nextflow code? 🤔 We would know the param name and attributes..

bentsherman:
The furthest I would go is to generate a code block in the schema builder that the user can copy into their Nextflow pipeline if they want.. Automatically updating code from an external source is an anti-pattern in my view

ewels:
Suggestion from call: Would be nice to support single-line comments (`//`) in addition to groovy-doc multi-line comments. This drops the number of lines per-param from 4 to 2, which makes quite a bit of difference if there are a lot of parameters.

docs/migrations/25-04.md
```diff
@@ -37,6 +37,35 @@ The third preview of workflow outputs introduces the following breaking changes
...
+  save_intermeds = false
+}
+
+workflow {
+  println "params.input = ${params.input}"
+  println "params.save_intermeds = ${params.save_intermeds}"
+}
+```
+
+This syntax allows you to declare all parameters in one place, and allows Nextflow to validate parameters at runtime.
```
christopher-hakkaart, line 65:
```suggestion
This syntax allows you to declare all parameters in one place and allows Nextflow to validate them at runtime.
```

docs/vscode.md
```diff
@@ -26,6 +26,19 @@ The language server parses scripts and config files according to the {ref}`Nextf
 
 When you hover over certain source code elements, such as variable names and function calls, the extension provides a tooltip with related information, such as the definition and/or documentation for the element.
 
+If a [Javadoc](https://en.wikipedia.org/wiki/Javadoc) comment is defined above a workflow, process, function, type definition, or parameter in a `params` block, the extension will include the contents of the comment in hover hints. Here is an example Javadoc comment:
```
christopher-hakkaart, line 29:
```suggestion
If a [Javadoc](https://en.wikipedia.org/wiki/Javadoc) comment is defined above a workflow, process, function, type definition, or parameter in a `params` block, the extension will include the contents of the comment in hover hints. The following is an example Javadoc comment:
```

docs/workflow.md
```diff
@@ -22,18 +22,69 @@ workflow {
 }
 ```
 
-### Parameters
+(workflow-params-def)=
 
-Parameters can be defined in the script with a default value that can be overridden from the CLI, params file, or config file. Parameters should only be used by the entry workflow:
+## Parameters
+
+There are two ways to declare parameters in a Nextflow script: the `params` block and *legacy* parameters.
```
christopher-hakkaart, line 29:
```suggestion
Parameters in a Nextflow script can be declared in the `params` block and as *legacy* parameters.
```
Please check meaning is retained.

christopher-hakkaart (commented):
I've added some suggestions to the docs. Looking good.

christopher-hakkaart (approved):
Docs looking good 👍

docs/migrations/25-04.md
```diff
@@ -37,6 +37,35 @@ The third preview of workflow outputs introduces the following breaking changes
 
 See {ref}`workflow-output-def` to learn more about the workflow output definition.
 
+<h3>Workflow params (first preview)</h3>
+
+The `params` block is a new way to declare pipeline parameters in a Nextflow script:
```
pditommaso, line 42: there's already support for parameters. It should be highlighted better what's new with this preview

docs/migrations/25-04.md
```diff
@@ -37,6 +37,35 @@ The third preview of workflow outputs introduces the following breaking changes
...
+
+```nextflow
+nextflow.preview.params = true
+
+params {
+  /**
+   * Path to input data.
+   */
+  input
```
pditommaso, line 51: it should not be allowed the use of undefined values

docs/workflow.md
```diff
@@ -22,18 +22,69 @@ workflow {
...
+:::{note}
+This feature requires the `nextflow.preview.params` feature flag to be enabled.
+:::
+
+A script can declare parameters using the `params` block:
+
+```nextflow
+params {
+    /**
+     * Path to input data.
```
pditommaso, line 45: what determines this is a path?

docs/workflow.md
```diff
@@ -22,18 +22,69 @@ workflow {
...
+workflow {
+    if( params.input )
+        analyze(params.input, params.save_intermeds)
+    else
+        analyze(fake_input(), params.save_intermeds)
+}
+```
+
+:::{note}
+Named workflows should not use params. They should receive their inputs explicitly through the `take:` section.
```
pditommaso, line 68: Not sure about this. I think it should be maintained the use of parameters both in workflows and processes as should as they are immutable

pditommaso (approved):
I like the general idea, but there are some points that needs to be improved. 

I'm a bit concerned about this:

> Config files can override script params or define "config params" which are only used by the config

Maybe I'm misunderstanding, but I don't config params should be managed differently from script params

pditommaso (changes requested):
I believe, this needs some work (as per previous comment)

docs/migrations/25-04.md
bentsherman: The benefit is further down:

> This syntax allows you to declare all parameters in one place and allows Nextflow to validate them at runtime.

docs/migrations/25-04.md
bentsherman: This is how you specify a required parameter. When there is no default value and you don't specify a value at runtime, the run will fail

docs/workflow.md
bentsherman: To be clear, parameters can still be used everywhere. This is just a recommendation since params cannot be validated outside the entry workflow. I can document this more explicitly if you want

bentsherman:
> I'm a bit concerned about this:
> 
> > Config files can override script params or define "config params" which are only used by the config
> 
> Maybe I'm misunderstanding, but I don't config params should be managed differently from script params

Config params are still added to the script as before:

https://github.com/nextflow-io/nextflow/blob/0f799994e2d76d307b51c4136b05fcfb0664200f/modules/nextflow/src/main/groovy/nextflow/Session.groovy#L447-L450

However, if you use the new `params` block, in the script you will only be allowed to use params that you declared in the `params` block:

https://github.com/nextflow-io/nextflow/blob/0f799994e2d76d307b51c4136b05fcfb0664200f/modules/nextflow/src/main/groovy/nextflow/script/ParamsDsl.groovy#L53-L69

So you can still override script params from the config, but you can't declare a new param from the config and use it in the script, without also declaring it in the `params` block. Again, this only applies if you use the new `params` block

docs/workflow.md
bentsherman: I have reframed this note as a best practice recommendation, since it is still allowed

pditommaso:
I believe it would be better continuing iterating this before release it as preview

bentsherman:
From our recent discussions, we will introduce workflow params in 25.10 as a complete feature (this iteration + type annotations + JSON schema integration)

I have updated the PR accordingly:
- remove the feature flag (will be introduced in 25.10 as a complete feature)
- remove support for v1 parser (incentivize adoption of strict syntax)

I would like to go ahead and merge this part so that I can use it in other efforts around static types. The type annotations and JSON schema integrations will need to be separate PRs.

awgymer:
Just coming across this PR. Can I ask what the intention is for type-coercion of params (particularly if "custom" types are supported)?

Is the idea that it will be built in to nextflow that a `Path` is automatically converted from a string? Would this functionality also handle the `exists` checks and checking `dir`/`path`/`glob` as we currently support in the schema? 

It feels like we are trending towards "pydantic for nextflow" which I love the idea of but I'm not clear quite how much of the schema functionality is expected to be replaced here?

bentsherman:
That is the long-term goal, to absorb much of the validation logic currently done by nf-schema, and to use type annotations in Nextflow code instead of writing JSON schema (the JSON schema will be generated from Nextflow code instead).

But this PR is just adding the basic params block, no types yet. Adding the types and validation will be an effort unto itself...

bentsherman:
That just means that the new `params` block will only be available with the strict syntax, but the v1 parser will still be available.

bentsherman:
Went ahead and added type annotations for params in this PR, since it seemed weird to add the `params` block without the ability to specify types.

One thing we need to figure out is what to do with the type casting of CLI params. We don't know at this point whether the user is using the new params definition, so I think we would need to either disable it by default or somehow "undo" it during the params validation.

EDIT: Since the `params` block is only supported by the v2 parser, I just disable the CLI params type casting when v2 parser is enabled.

christopher-hakkaart (approved):

christopher-hakkaart:
Docs are awesome.

bentsherman:
@pditommaso ~~this is ready to merge~~ never mind, another test is failing...

muffato:
> A parameter that doesn't specify a default value is a *required* param.

Can I specify `null` as the default value to indicate an optional param ? Something like
```
params {
  index: Path = null
}
```

That would be like `Optional[Path]` in Python

bentsherman:
@muffato yes, you can default to `null`. We will also add nullable types so that you have to specify it as `index: Path? = null` if the param can be null. That will be in a separate PR.

bentsherman:
@pditommaso now tests are passing, ready for you

Issue #4286, opened 2023-09-10
GPU usage metrics


shenker:
## New feature

It would be extremely useful if GPU usage metrics were recorded for GPU tasks.

## Usage scenario

Using GPU resources efficiently on HPC is often a challenge. For example, basecalling Oxford Nanopore sequencing data using the dorado basecaller often takes quite a bit of tuning to get good performance on HPC, for the following reasons:
1) Duplex mode makes heavy use of random access over thousands of files, resulting in low GPU utilization if the shared filesystem cannot keep up. Being able to monitor GPU utilization would allow detecting and mitigating this issue.
2) Dorado exhibits widely different performance on different GPU hardware. HPC nodes often are equipped with heterogenous GPU hardware. When parallelizing dorado jobs, it would be useful to measure the relative performance gaps between different GPU hardware. This information could be used to fine-tune job GPU requirements. (In SLURM or other cluster managers it's usually possible to specify which GPU hardware you're willing to use for a job to, e.g., exclude very old Nvidia architectures that no longer offer acceptable performance for a particular task.)
3) Dorado is a heavy user of GPU VRAM, and crashes when it runs out. Monitoring VRAM usage would help users tune dorado parameters to optimize performance/VRAM usage and know which GPU hardware to request from the cluster manager.

These are very common issues when running dorado on HPC (there are tons of issues on dorado's bug tracker, see, e.g.: https://github.com/nanoporetech/dorado/issues/68, https://github.com/nanoporetech/dorado/issues/336, https://github.com/nanoporetech/dorado/issues/306). This is just a particular example, I imagine the same basic GPU metrics would be useful for most users running GPU tasks with nextflow.

## Suggest implementation 

An initial implementation could restrict itself to Nvidia GPUs, since those are overwhelmingly the most important for scientific computing.

Use `nvidia-settings` to record the following metrics:
- GPU name(s) (useful in a heterogenous cluster to see what GPU hardware a particular task is running on). The same architecture with different amounts of VRAM should be considered different models (e.g., “A100 40GB” should be distinguished from “A100 80GB”)
- GPU utilization (average)
- GPU memory (average used, total available)
- GPU memory utilization (% of time memory controller was busy)
 
(A quick google search turned up this list of ways to programmatically grab GPU metrics: https://unix.stackexchange.com/questions/38560/gpu-usage-monitoring-cuda)

[...]

PR #5930, opened 2025-03-31
Add gpu metrics to TraceRecord

FloWuenne:
# Description

This PR implements feature requests from #4286 .

The current implementation adds the following fields to the TraceRecord:
- gpu_model
- gpu_mem
- gpu_driver: 'str'
- -'%gpu'
- %gpu_mem
- avg_gpu_mem

This is accomplished by implementing a new `nxf_gpu_watch()` in  `command-trace.txt `, similar to `nxf_mem_watch()`. This allows to access GPU metrics in the Tracefile, by adding the following lines to the nextflow.config:
```
trace{
fields = 'gpu_model,gpu_mem,gpu_driver,%gpu,%gpu_mem,avg_gpu_mem'
}
```

# Limitations

A current limitation of this approach is that if using Nextflow with an executor (for example local) and a local GPU that is operating system wide, the usage metrics are not task specific but include background processes. For example, on my local Ubuntu machine with an `NVIDIA GeForce GTX 1650` even a cpu only task reported some GPU usage, due to background GPU processes utilizing the GPU. This should not be the case when running on instances where GPUs are assigned to specific tasks using accelerator, however I haven't tested this on multi-GPU machines on AWS or other cloud executors.

### DIFF
--- a/docs/reports.md
+++ b/docs/reports.md
@@ -314,6 +314,31 @@ The following table shows the fields that can be included in the execution repor
   :::
 : The name of the CPU model used to execute the task. This data is read from file `/proc/cpuinfo`.
 
+`gpu_model`
+: :::{versionadded} 25.02.1-edge
+  :::
+: The name/model of the GPU device used to execute the task. This data is read using the `nvidia-smi` command.
+
+`gpu_mem`
+: :::{versionadded} 25.02.1-edge
+  :::
+: The total GPU memory available on the device. This data is read using the `nvidia-smi` command.
+
+`gpu_driver`
+: :::{versionadded} 25.02.1-edge
+  :::
+: The NVIDIA driver version installed on the system. This data is read using the `nvidia-smi` command.
+
+`%gpu`
+: :::{versionadded} 25.02.1-edge
+  :::
+: The average GPU utilization percentage during task execution. This data is read using the `nvidia-smi` command.
+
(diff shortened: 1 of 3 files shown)


pditommaso:
This looks a nice step forward, I'm start wondering if it's not time we write our own tool to trace these metrics (including gpu). @jordeu @fntlnz thoughts?

jordeu:
> This looks a nice step forward, I'm start wondering if it's not time we write our own tool to trace these metrics (including gpu). @jordeu @fntlnz thoughts?

Absolutely. Our own tool will simplify Nextflow bash scripts and allow us to do many more things, like collecting metrics during the process's lifetime.

FloWuenne:
Since our own tool for performance metrics still seems a bit away though, do we want to move forward in the meantime with using nvidia-smi here and add more gpu controllers over time or are there major implementations here we want to change?

fntlnz:
> This looks a nice step forward, I'm start wondering if it's not time we write our own tool to trace these metrics (including gpu). @jordeu @fntlnz thoughts?  

Agreed, seeing the broader picture with the ability to drill down to the specific task would be very useful both to optimize pipelines and also to optimize their runtime scheduling.

FloWuenne:
Sounds good @pditommaso! Looking forward to see the new tracing tool soon!

PR #5935, opened 2025-04-02
Fix Syntax Error with MermaidRenderer

Lehmann-Fabian:
When I create an HTML of the DAG using the MermaidRenderer, I get a Syntax Error when opening the file.
Quoting the names solves this.


bentsherman:
Thanks Fabian. Which version did this happen for you? Trying to determine if it was a regression

bentsherman (approved):
Please allow me to update the branch or update it yourself so that I can merge it.

Keep in mind that every time another PR is merged to master, this PR has to be updated again. Because of our time difference we could get caught in an infinite loop. This is why it's easier to give me push access so that I can update the branch myself

PR #5937, opened 2025-04-02
Fix Nextflow console

bentsherman:


pditommaso:
This is good as stopgap, it would be nice to make it working with v2 parser as well.

Issue #5940, opened 2025-04-03
Merge of conflicting process names in logs


kdesnos:
Hi,

I'm not really sure this can be considered a bug, but since this "issue" puzzled me, I'm sharing it with everyone to see if it should be acted upon. 

## The issue
When the list of process names is printed in the `.nextflow.log` file, it may happen that two distinct processes, declared in two distinct nf files, which share the exact same name become only one in the list.
```Apr-03 09:55:00.694 [main] DEBUG nextflow.Session - Workflow process names [dsl2]: process_a, process_b```

This was the source of my puzzlement, as when adding a new process in a file, you expect to see it appear in this list.

## Step to reproduce
Declares processes in two files:
```
# file: main.nf
process a {
    output:
        stdout

    script: 
    """
        echo "I'm process main:a"
    """
}

```

```
# file: sub.nf
process a {
    output:
        stdout

    script: 
    """
        echo "I'm process sub:a"
    """
}
```

Execute the workflow, the following content will be printed in the logs:
```[main] DEBUG nextflow.Session - Workflow process names [dsl2]: a```

In case the processes are used in several hierarchical workflows, the "resolved" names will also be printed as such:
`a` for the use of `main.nf:a` in the "top level" workflow of `main.nf`, and `sub:a` for the use of `sub.nf:a` in the a `sub` workflow.

## The cause
The reason behind this behavior is that the list of process names returned by the static member function `ScriptMeta.allProcessNames()` returns a `Set<String>`, which cannot contain duplicates. In a scenario where a is used within the main and a subworkflow, the following elements will actually be added to the set:
* `a` : process `main.nf:a`
* `a` : process `sub.nf:a`
* `a` : Use of process `main.nf:a` within top-level workflow
* `sub:a` : Use of process `sub.nf:a` within a sub workflow.

Resulting in the following set `a, sub:a`.
In case `sub.nf:a` is not used directly within a workflow, but instead imported with an aliased name in another workflow (eg. `include {a as x} from ./sub.nf`), then the list would look like this `a, x`, with no clear trace of the multiple definitions of `a`.

## What can be done?
To answer what can be done, I should first explain why this is bothering me :) My objective* is to analyze the traces of pipelines run with Nextflow to try to build a model for predicting performances of future executions of this pipeline. To do that, I need to be able to associate each process execution from a report, identified by its "resolved" and possibly aliased name, with the corresponding process `file:process_name`. This objective is currently made impossible, notably because of the described issue.

[...]

PR #5944, opened 2025-04-04
Print process dictionary in logs (#5940)

kdesnos:
## What
Update the logs in the following manner:
1. Replace workflow process name list, which mixes defined process names and aliased process names
   ```
   [main] DEBUG nextflow.Session - Workflow process names [dsl2]: a
   ```
   with two lists: one of process definitions from each file, one of aliases used for each process.
   ```
   [main] DEBUG nextflow.Session - Workflow process definitions [dsl2]: main.nf [a, other_process, ...], sub.nf [a, ...]
   [main] DEBUG nextflow.Session - Workflow resolved process names: a[main.nf:a], sub:a[sub.nf:a], x[sub.nf:a]  
   ```

2. Complement the following log line:
   ```
   [main] DEBUG nextflow.processor.TaskProcessor - Starting process > x
   ```
   with typed list of inputs and outputs of the process.
   ```
   Starting process > x (type:ArgName, ..., default:$) -> (type:OutName, ...)
   ```

## Motivations
See #5940

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -873,10 +873,19 @@ class Session implements ISession {
             final ver = "dsl${NF.dsl1 ?'1' :'2'}"
             final processDefs = ScriptMeta.allProcessDefinitions()
             log.debug "Workflow process definitions [$ver]: ${processDefs.entrySet().collect{"${it.key} ${it.value}"}.join(', ')}"
-            final resolvedNames = ScriptMeta.allResolvedProcessNames()
+            def resolvedNames = ScriptMeta.allResolvedProcessNames()
+            // Add unaliased process names to the resolved names map
             final mainScriptMeta = ScriptMeta.get(script)
-            final unaliasedProcessNames = mainScriptMeta.getLocalProcessNames() // Process in main file are not aliased.
-            log.debug "Resolved process names: ${unaliasedProcessNames.collect{"$it ${mainScriptMeta.getScriptPath()}:$it"}.join(', ')} ${resolvedNames.entrySet().collect{"${it.key} ${it.value}"}.join(', ')}"         
+            mainScriptMeta.getLocalProcessNames().each { name ->
+                // The processes defined in the main script cannot be included in 
+                // other scripts, as it would create a circular inclusion.
+                // Hence, no need to check for the key existence in the map.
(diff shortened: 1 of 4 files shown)


kdesnos:
Oops.. I thought I'd been very cautious not to alter any existing behavior when modifying the code..

Unfortunately, I'm off for two weeks so I can't have a look at why the tests are failing at the moment. (Also, I'm not using a proper IDE, only building and running, so I probably need to setup a proper dev environment to debug these issue.)

kdesnos:
The PR is ready to be merged.

kdesnos:
I noticed that the "resolved" name of the included process may be ambiguous, in case the same alias is used in different workflow, for different original process. 

This is because when an include occurs in a workflow `sub`, say with process `p` being included as `toto`, the process is first cloned with name `toto`, before being re-cloned with the unique name `sub:toto`. If main workflow clones process `q` with the name `toto` then, the resolved name `toto` will appear twice, once for each workflow. The aliased process `toto` included in the main file, will not be recloned with a different name. Hence, the process `toto` will be executed, but it will be ambiguous whether it is the clone of `p`, or `q`.

I'll fix this issue before reopening the PR.

kdesnos:
To better motivate the novelty brought by this PR, a link to a video where I explain how I study Nextflow in my research project:
https://youtu.be/YbiVhuC3jx8?feature=shared&t=2197
The most interesting/relevant part of the video is at 36 min 37 sec.
Don't hesitate to contact me for more info :)

kdesnos:
Argh.
@bentsherman 
I'm actually counting on this feature (and the associated PR) to build a post-execution trace analysis mechanism. Early results demonstrate that I can generate automatically configurations files for future executions of the pipeline, where parameterized resource allocation are closer to actual needs than developers guesstimates. On a cluster managed with Slurm, this generally enables more efficient scheduling of the computations.

Early open source results are presented here : https://youtu.be/YbiVhuC3jx8?si=z-z2JR5X-MDRm7dk&t=2204 and after consolidating them with additional pipelines, a publication is envisioned.

Although I understand this might not be strictly needed in Nextflow's current roadmap, I believe the PR to be harmless and relatively lightweight to merge...

bentsherman:
That's great, but then we should pursue that feature directly rather than piggy-backing on a debug message. Consider submitting a new issue with your proposal and we can discuss how to best implement it in the runtime

PR #5953, opened 2025-04-08
Prefer `toList` over `collect` operator

bentsherman:
Promote `toList` in the operator summary instead of `collect`

- `toList` mirrors the `fromList` channel factory
- avoid confusion with [Iterable::collect()](https://nextflow.io/docs/latest/reference/stdlib.html#iterable-e)
- avoid weird behaviors with `collect` like flattening by default and emitting nothing instead of an empty list

### DIFF
--- a/docs/channel.md
+++ b/docs/channel.md
@@ -86,8 +86,6 @@ Commonly used operators include:
 
 - {ref}`operator-combine`: emit the combinations of two channels
 
-- {ref}`operator-collect`: collect the values from a channel into a list
-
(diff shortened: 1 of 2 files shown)


bentsherman:
Hmm. On second thought, we shouldn't do this. `toList` should not be recommended because it returns an ordered list whereas `collect` returns a bag (unordered collection), which is the correct behavior

I think we'll have to go with `collect` and just fix it in the v2 operator library

PR #5954, opened 2025-04-09
Add support for Fusion Snapshots

pditommaso:
This PR adds support for Fusion snapshots by adding a config setting: 

```
fusion.snapshots=true|false
```

When enabled Fusion snapshots restores automatically AWS Batch jobs interrupted by a spot reclamation. 

The number of restarts is determined by the setting `aws.batch.maxSpotAttempts` (default `5` when Fusion Snapshots is enabled)


bentsherman (changes requested):
- add config option to docs
- add config option to `FusionConfig.java` in nf-lang (use the same description as the docs)

bentsherman (approved):

pditommaso:
Adding the adjustment of spot retries when snapshots is enabled

PR #5956, opened 2025-04-09
Cid store improve coverage, fixes and others

jorgee:
This PR includes this changes

- [x] Improve coverage
- [x] add both taskRun and workflowRun in DataOutput
- [x] Symetry in usage of /outputs and #outputs in Tasks and Workflows
- [x] Unify main and other script files
- [x] Change default history log to remove file lock
- [x] Navigate in collections
- [ ] Property validator for better query errors
- [x] find command
- [ ] Unify inputs and outputs model


jorgee:
I have implemented a property validator that checks if the properties included in the queries and fragments are in the CID model classes. However, as there are some properties (such as annotations) that are maps that can contain whatever key, it is not possible to use to validate them. 
These could be a pro to use a list of predefined objects instead of maps in whatever place outputs and annotations.

PR #5958, opened 2025-04-10
Update VS Code docs page

bentsherman:
This PR updates the VS Code docs page to focus on the language server, while deferring to the extension page for things like available commands and configuration settings, as the Nextflow docs often gets out of sync with the extension.

The extension now has more and more features outside of the language server, e.g. Copilot for Nextflow and custom webviews. These features are evolving rapidly and it doesn't make sense to keep up with them in the Nextflow docs. For this reason also I prefer to simply point to the extension page as the source of truth.

I would consider renaming the page "Language server" and add a section for editors like VS Code, but the reality is that basically everyone uses VS Code, so it makes more sense to keep that the main thing and just focus on language server features.

Once the newer features are more stable, it could make sense to document them here.

### DIFF
--- a/docs/vscode.md
+++ b/docs/vscode.md
@@ -2,7 +2,9 @@
 
 # VS Code integration
 
-The [Nextflow VS Code extension](https://marketplace.visualstudio.com/items?itemName=nextflow.nextflow) provides IDE support for Nextflow pipelines.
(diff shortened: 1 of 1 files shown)


christopher-hakkaart (approved):
Agree with your points above. I've added some minor language suggestions for consideration but they are non blocking.

PR #5966, opened 2025-04-14
Replace `drop/list/view` commands with `pipelines` command

### LINKED ISSUE #5959: [Proposal] Move CLI commands for project management to subcommands
As we add several new commands like inspect, linting, formatting, data provenance, the CLI is getting crowded.

There are several commands for managing pipeline projects:
- `clone`
- `drop`
- `list`
- `pull`
- `view`

I think all of these commands could be combined into a single command e.g. `projects` or `repos`:
```bash
nextflow projects list
nextflow projects pull <project>
nextflow projects view <project>
...
```

Similar to what we are doing in #5715 

We can keep the existing commands as redirects and hide them in the CLI help, so it should be easy to transition.

bentsherman:
Close #5959 

This PR moves the `drop`, `list`, and `view` commands into a single `pipelines` command:
- these commands all pertain to managing the pipeline cache
- they are very generic names which can be confusing as we add more top-level commands
- `list` is too similar to `lint`, easy to misspell

I was originally going to move `clone` and `pull` as well, but I'm not sure if there's a clean way to support CLI options for these subcommands with JCommander. I refactored the main three commands to not have any extra options, but I don't think I can do this for `clone` and `pull`. Also, moving the `pull` command might be more disruptive anyway.

Long-term, I think we could introduce a CLI v2 called `nf` with (1) GNU-style options, (2) picocli over JCommander, and (3) a refactored command tree. So for now I'm happy to just merge this change as a short-term improvement.


pditommaso:
Let's postpone this, post 25.04. There's no hurry

bentsherman:
At that point we might as well not do it, we'll likely have CLI v2 by then

bentsherman:
Yeah, let's just do this in CLI v2

PR #5968, opened 2025-04-14
Support downloading plugins using ORAS

tom-seqera:
ORAS (OCI Registry As Storage) is a way to store arbitrary artifacts in OCI Registries (see https://oras.land/).

This change adds support for downloading Nextflow plugins stored using ORAS. When a plugin download url starts with the protocol `oras://` the download is handled by a new OrasPluginDownloader class instead of the default pf4j FileDownloader.

It also defines an ORAS artifact type (`application/vnd.nextflow.plugin+zip`) and checks that the metadata of the given ORAS url matches this type.

This change should have no impact on any other download urls or plugin download mechanism.

### DIFF
--- a/modules/nf-commons/build.gradle
+++ b/modules/nf-commons/build.gradle
@@ -35,6 +35,7 @@ dependencies {
     api 'dev.failsafe:failsafe:3.1.0'
     // patch gson dependency required by pf4j
     api 'com.google.code.gson:gson:2.10.1'
+    implementation 'land.oras:oras-java-sdk:0.2.4'
 
     /* testImplementation inherited from top gradle build file */
     testImplementation(testFixtures(project(":nextflow")))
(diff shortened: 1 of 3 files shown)


pditommaso:
What's the benefits of using oras protocol, instead of downloading the plugin assets via http(s) ?

tom-seqera:
One particularly nice thing is it allows the download url to be more readable, similar to a docker style tag (eg `oras://example.com/plugins/nf-hello:0.5.0`) rather than a long hash.

pditommaso:
Think it's not enough to justify the new dependency. It would be preferable nextflow remains agnostic on this

bentsherman:
@pditommaso wasn't there a similar debate on http vs oras for downloading wave singularity images? Do you remember which one we preferred for that and why?

pditommaso:
For container we use both. Oras when Singularity is pulling it natively, and https for legacy runs to download the container and locally and run it as a file.

bentsherman:
Okay so oras wasn't an issue with singularity because the singularity CLI handles the download 👍

ewels:
> One particularly nice thing is it allows the download url to be more readable

Does this refer to when the user specifies the plugin identifier in their config? Can we set up Nextflow so that `nf-hello:0.5.0` gets whatever the default base is? In which case I'm not sure that readability / oras vs http really matters, as the end user will never see the full URL.

tom-seqera:
Closed for: #6016

PR #5971, opened 2025-04-14
Separate CLI from runtime

bentsherman:
This PR splits the nextflow module into two modules:
- `nf-cli-v1`: contains only the Launcher and CLI classes
- `nextflow`: contains everything (i.e. the "runtime")

Benefits:
- Improves build time, editing CLI code doesn't rebuild the entire project
- Removes some circular dependencies, making the code easier to read/understand
- Library consumers like platform or plugins can just depend on the runtime and not the CLI
- Makes it easier to implement a CLI v2 in the future

Notes:

- I moved the CLI bits of `ConfigBuilder` into a separate class `ConfigCmdAdapter` in order to keep the core config builder in the runtime. This also provides a nice separation of concerns.

- I moved `K8sDriverLauncher` into `nf-cli-v1` so that the `nf-k8s` is CLI-agnostic. This means I have to included nf-k8s as a build-time dep for CLI v1, but CLI v2 will be able to load `nf-k8s` on-demand as a plugin.

- I moved `PluginExecAware` into `nextflow` so that plugins can declare CLI commands without depending on `nf-cli-v1`. I replaced the config loading here with a minimal config, but this might not be enough, so we might want to think about how much of the config logic should be replicated for plugin commands


bentsherman:
Due to the nature of this PR, I will keep it updated via rebase instead of merge commits so that the changes are as clear as possible.

bentsherman:
Your k8s plugin inspired me 😄

pditommaso:
I feared that 😆

bentsherman:
Jokes aside, it might be good to do the actual code changes in a separate PR so that the nf-runtime part is as clean as possible

The nf-runtime refactor was just the easiest way to reveal the circular dependencies

pditommaso:
I'll review post 25.04

bentsherman:
Refactored so that the CLI code is in a new module `nf-cli-v1`. Made the PR considerably smaller

Note the following plugins depend explicitly on the CLI code because they add CLI commands:
- nf-console
- nf-k8s
- nf-tower
- nf-wave

Might be possible to remove this dependency by moving some interfaces into the core runtime, but not a big deal either way

bentsherman:
I refactored the `nf-lineage` module to depend on `nextflow` and be required by `nf-cli-v1`. This removes the need for the `LinCommand` interface as the lineage command can simply use `LinCommandImpl` directly.

Couldn't quite do this for `nf-k8s` because the `K8sDriverLauncher` uses `CmdRun`, so that's a pretty hard dependency on the CLI v1. We could make `nf-k8s` a module instead of a plugin, but then it would be included at build-time instead of runtime.

modules/nextflow/build.gradle
```diff
@@ -41,14 +44,13 @@ dependencies {
...
-    api "com.beust:jcommander:1.35"
     api("com.esotericsoftware.kryo:kryo:2.24.0") { exclude group: 'com.esotericsoftware.minlog', module: 'minlog' }
-    api('org.iq80.leveldb:leveldb:0.12')
-    api('org.eclipse.jgit:org.eclipse.jgit:7.1.1.202505221757-r')
-    api ('javax.activation:activation:1.1.1')
-    api ('javax.mail:mail:1.4.7')
-    api ('org.yaml:snakeyaml:2.2')
-    api ('org.jsoup:jsoup:1.15.4')
+    api 'org.iq80.leveldb:leveldb:0.12'
+    api 'org.eclipse.jgit:org.eclipse.jgit:7.1.0.202411261347-r'
```
pditommaso: Was this downgraded on purpose? before it was `7.1.1.202505221757`

pditommaso:
Did I break anything? I see some tests failing. Also trying locally I get 

```
» ./launch.sh run hello

 N E X T F L O W   ~  version 25.06.0-edge

NOTE: Your local project version looks outdated - a different revision is available in the remote repository [2ce0b0e294]
Launching `https://github.com/nextflow-io/hello` [backstabbing_ride] DSL2 - revision: afff16a9b4 [master]

ERROR ~ No such variable: session
```

modules/nextflow/build.gradle
bentsherman: this is likely an incorrect merge commit, none of the versions should be changed. I will fix it

bentsherman:
> Did I break anything? I see some tests failing. Also trying locally I get

Actually that's my fault, looks like a regression from #6257 . I will submit a separate fix

bentsherman:
Found some possible solutions for decoupling nf-k8s, nf-tower, and nf-wave from CLI v1. See top comment for details.

bentsherman:
Now the problem is that `nf-tower` has implementations of commands defined in `nf-cli-v1`, so `nf-tower` is now coupled to the CLI v1.

I guess we'll have to move the `AuthCommand` and `LaunchCommand` interfaces into the core runtime.

bentsherman:
In testing this refactor I uncovered a bug with the test harness. I have submitted the fix separately (#7264) since it is a standalone issue

Temporarily included in this PR as well to get the CI to pass. I recommend we merge #7264 first and then this one

bentsherman:
Tests are now passing :heavy_check_mark:

modules/nextflow/src/main/groovy/nextflow/cli/AuthCommand.groovy
jorgee, line 1: This class and `LaunchCommand` are the only remaining `cli` classes in `nextflow `module. It is only implemented in `nf-seqera.` I see `nf-cli-v1` already depends on `nf-k8s` plugin and nf-lineage module. Is there any reason why we can't do the same here and include the `nf-seqera` dependency and just import the implementations?

bentsherman: We generally avoid this in order to not bloat the core binary

Just to demonstrate the size of the core JAR vs plugins:
```
$ du -hs .nextflow/framework/26.04.4/*
41M     .nextflow/framework/26.04.4/nextflow-26.04.4-one.jar
$ du -hs .nextflow/plugins/*
80M     .nextflow/plugins/nf-amazon-3.9.2
23M     .nextflow/plugins/nf-azure-1.22.6
52M     .nextflow/plugins/nf-google-1.27.3
3.7M    .nextflow/plugins/nf-seqera-0.21.0
7.0M    .nextflow/plugins/nf-tower-1.27.0
6.0M    .nextflow/plugins/nf-wave-1.21.0
```

The cloud plugins are the main problem because they have so many dependencies

nf-seqera is not so bad (these sizes are prior to merging with nf-tower), but I think we still try to keep the plugins decoupled as a best practice

nf-k8s was a practical choice -- since I doubt the CLI v2 will have a `kuberun` command, I figure it's fine to couple nf-k8s to the CLI v1. In fact it was not a plugin for a long time, so it's not much different from the historical state

plugins/nf-seqera/src/main/io/seqera/tower/plugin/BaseCommandImpl.groovy
```diff
@@ -50,9 +50,13 @@ class BaseCommandImpl {
         return apiEndpoint.replace('://api.', '://').replace('/api', '')
     }
 
-    protected Map readConfig() {
-        final builder = new ConfigBuilder().setHomeDir(Const.APP_HOME_DIR).setCurrentDir(Const.APP_HOME_DIR)
-        return builder.buildConfigObject().flatten()
+    protected ConfigObject readConfig() {
+        final configFile = Const.APP_HOME_DIR.resolve('nextflow.config')
```
jorgee: `APP_HOME_DIR` points to $HOME/.nextflow/.  Before the ConfigBuilder changes it was pointing to `$HOME/.nextflow/config` and now it will resolve to `$HOME/.nextflow/nextflow.config` at it doesn't exists.

jorgee (changes requested):
I have seen an issue with the `nextflow auth` and `launch`, It is storing/reading the tower details in a config file included in `$HOME/.nextflow/config`  with the changes it is poinitng to an incorrect file `$HOME/.nextflow/nextflow.config`. 

In the previous version version it worked because `ConfigBuilder` was always including `$HOME/.nextflow/config` by default. However, in the new version It has been removed from `ConfigBuilder`. I think there is a risk with this change, there could be some code (maybe in plugins) that rely on this default and with new version `$HOME/.nextflow/config` options will not be loaded by default. Is there a reason to not keep this default in the `ConfigBuilder`?

plugins/nf-seqera/src/main/io/seqera/tower/plugin/BaseCommandImpl.groovy
bentsherman: Must have been a premature simplification I thought I could do. I will try to revert it and preserve the existing behavior

bentsherman: Fixed

bentsherman:
@jorge regarding your concern about config defaults -- the ConfigBuilder is a generic utility for loading a set of config files into a single config map, while the ConfigCmdAdapter provides the environment shell used by CLI commands (CLI options, environment defaults, etc)

The ConfigBuilder should not encode these defaults so that it can be generic

Most plugins so far have relied on the session config instead of trying to load config files directly, so they should not be affected

pditommaso:
The direction here is right and the bulk of the diff is low-risk pure renames (`Cmd*`, `Launcher`, tests → `nf-cli-v1`) plus the necessary coupling breaks (`ConfigCmdAdapter` extraction, `AuthCommand`/`LaunchCommand`/`PluginExecAware` moved to the runtime, `K8sDriverLauncher` move, `LoggerHelper` decoupling).

My one request: **a few refactors rode along that aren't required by the split.** They inflate the review surface, add the only non-trivial risk, and — importantly — make it harder to cherry-pick or revert the module move independently later. I'd pull them into separate PRs so this one stays a mechanical, easy-to-bisect move.

### Refactors that can be dropped from this PR

| Rider change | Required for the split? | Note |
|---|---|---|
| **`ConfigBuilder` full `@CompileStatic` rewrite** + method renames (`buildConfig0`→`build`, `validate`→`checkUnresolvedConfig`), `Paths.get`→`Path.of`, setter reordering | ❌ no | The *extraction* of CLI logic into `ConfigCmdAdapter` is needed; the rewrite is not — and it smuggled in two behavior changes (below) |
| **`HubOptions` trait→`record`** + method renames (`hubUser`→`user()`, `getHubProvider`→`provider()`) and `cliOpts`→`hubOpts` param rename across `AssetManager` | ❌ no | The split only requires *moving* the class out of `nextflow.cli`. Move as-is; rename later if desired |
| **`SpuriousDeps` deletion** (dead `cmdCloud` reflection) | ❌ no | Unrelated dead-code cleanup; git reports it as a 56% "rename" into `LoggerOptions`, which is misleading (it's a delete + add) |

### Behavior changes buried in the `ConfigBuilder` rewrite

These are the reason the rewrite matters beyond churn — they are semantic changes, not required by the move, and are **not** compensated in `ConfigCmdAdapter`:

**1. `session` dropped from the guaranteed top-level scopes**
```groovy
// before
for( String name : ['env','session','params','process','executor'])
// after
for( final name : List.of('env','executor','params','process') )
```
Any code reading `config.session` without a null-guard now sees it absent when unset.

**2. Config binding no longer includes `SysEnv`/env for the v2 parser**
```groovy
// before: always
final binding = new HashMap(SysEnv.get()); binding.putAll(env)
// after: only for v1
if( NF.getSyntaxParserVersion() == 'v1' ) { binding.putAll(SysEnv.get()); binding.putAll(env) }
```
For the v2 config parser, environment variables are no longer available in the config binding — a real change to config resolution.

If both are intentional they deserve their own PR with dedicated tests and a changelog note; if not, they're regressions. Either way they shouldn't land silently inside a 114-file module move — there are currently no tests asserting the new behavior.

### Minor
- Copyright year regressions: `ConfigBuilderTest` changed `2013-2026`→`2013-2024`; new `HubOptions.groovy` header says `2013-2024`.
- `AssetManager.setupCredentials` rebuilds the record (`hubOpts = new HubOptions(hub, hubOpts.user())`) just to set the provider — functionally equivalent to the old `cliOpts.hubProvider = hub` but a confusing way to express it.

### Suggestion
Land this PR as the mechanical move + coupling breaks only. Extract (a) the `ConfigBuilder` CompileStatic rewrite + the two behavior changes and (b) the `HubOptions` record conversion into follow-up PRs. That keeps the surface minimal, removes essentially all the correctness risk, and makes the module split cleanly cherry-pickable/revertable on its own.

bentsherman:
Thanks for the review. I'm happy to move these changes into separate PRs, but they are *prerequisites*, not independent follow-ups. Otherwise I wouldn't have done them here in the first place.

Some classes need to be split so that the CLI bits can be moved into the CLI module. Otherwise you have to leave those classes entirely in the runtime module and the CLI isn't actually decoupled

I will submit separate PRs for these changes and address the specific feedback there

bentsherman:
Spun out each "rider change" into separate PRs:

- https://github.com/nextflow-io/nextflow/pull/7371
- https://github.com/nextflow-io/nextflow/pull/7372
- https://github.com/nextflow-io/nextflow/pull/7373

Marking this PR as draft until these three are merged, then I will update this PR accordingly

pditommaso:
Thanks for splitting these out, Ben — and sorry for the extra round trips. The result speaks for itself: the diff is down to 103 files and reads as an actual mechanical move now. `ConfigCmdAdapter` showing up as a pure 0-line rename is exactly what I was hoping for, and having the three prerequisites land separately means this one can be cherry-picked or reverted on its own.

I went back through the spin-outs and they all check out:

- **#7373** — both behavior changes are now declared up front and covered by tests (`ConfigBuilderTest:265` and `:278`). I verified them independently: nothing in the runtime reads a top-level `session` scope, and the v2 env change matches the language contract, since `ConfigDsl` deliberately exposes `env(String)` as a function rather than binding env vars as constants. Agreed on both.
- **#7372** — landing this as a plain immutable class with `getUser()`/`getProvider()` rather than a record was the right call, and `withProvider(hub)` with the comment reads much better than the old rebuild.
- **#7371** — clean.

The ADR is a genuinely good addition. Calling out the `WaveCmdEntry` regression yourself rather than letting it be discovered later is the right instinct.

A few new things I picked up on this pass, none of them fundamental:

### Worth fixing before merge

**`packing.gradle` — `ant.copy` → `"cp $source $target".execute()`.** Groovy's `String.execute()` doesn't wait for the process and ignores its exit status, and this sits in `packOne` on the release path — a downstream task could pick up a partial jar with nothing surfacing. I think the change was only needed to rename the artifact, so `ant.copy(file: source, tofile: ..., overwrite: true)` or `archiveBaseName = 'nextflow'` in the shadowJar block would do it safely.

**`nf-k8s` now ships inside the distribution fat jar.** Dropping the explicit `configurations = [...]` from `shadowJar` means it falls back to `runtimeClasspath`, which now includes the `api(project(':plugins:nf-k8s'))` dep. I probed it on the branch:

```
SHADOW_CONFIGS: [configuration ':nf-cli-v1:runtimeClasspath']
K8S_RELATED_IN_FATJAR: [nf-k8s-1.5.5.jar, bcpkix-jdk18on-1.84.jar, bcprov-jdk18on-1.84.jar]
```

Since nf-k8s is still distributed as a plugin too (`plugins-info.txt: nf-k8s@1.5.5`), the same classes end up in both the app and plugin classloaders, and `nextflow.k8s` becomes a split package. The ADR accepts the build-time dependency, which is fair, but this is the part that could actually bite — plugin version pinning silently becoming a no-op, or cross-classloader `ClassCastException`. Keeping `configurations` explicit and excluding nf-k8s would contain it.

To be clear though, dissolving the `lineageImplementation` hack is the cleanest win in the whole PR — that workaround has annoyed me for a long time.

**Test heap for the `nextflow` module.** The `test { minHeapSize/maxHeapSize }` block moved to `nf-cli-v1` rather than being duplicated, so `:nextflow:test` now resolves to `null` for both while the root build sets no default. The largest suite in the repo drops from an explicit 4 GB, and the module with a fraction of the tests inherits it. CI passes, so it fits today — looks unintentional rather than harmful. Probably belongs in the root `subprojects` block so it can't drift again.

**`CmdKubeRun` no longer calls `Plugins.init()` / `Plugins.start('nf-k8s')`.** The old path initialized the plugin system before resolving the extension; the new one constructs `K8sDriverLauncher` directly. Whether that matters depends on the `k8s` config scope, which is registered via the `nextflow.k8s.K8sConfig` extension point. Nothing in CI exercises `kuberun`, so this probably wants a manual smoke test before merge.

**`PluginExecAware` breaks third-party plugins twice** — package move *and* signature change. The ADR covers the signature; worth a changelog note covering both.

### Nits

- `LoggerOptions` has no class javadoc and no `@author`, unlike the rest of the codebase. With seven positional params and three same-typed booleans, the javadoc earns its keep here — callers have only parameter order to go on. Also slightly ironic that we landed `HubOptions` *away* from a record in #7372 and introduced one here.
- `AuthCommand.groovy` and `LaunchCommand.groovy` headers say `2013-2025`.
- `LaunchOptions` could use its own file rather than riding along in `LaunchCommand.groovy`.
- `-XDignore.symbol.file` went to `nf-cli-v1` (one internal-API import) while the four files that motivate it — `Session`, `AbortSignalException`, `ThreadPoolHelper`, `LocalPollingMonitor` — stayed in `nextflow`. It's inert either way since both modules compile through groovyc, so this is cosmetic.
- Stale comment in `packing.gradle`: "default cfg = runtime + httpfs + lineage", but lineage now arrives transitively.
- A handful of `api('g:a:v')` → `api 'g:a:v'` reformats snuck back in.

### CI

The four failures are all 90-minute timeouts on `test_aws`/`test_wave`, hanging in *"running rnaseq-nf with Fusion on AWS Batch"* after job submission — that's infrastructure, not your change. Everything this PR touches passed. But the run is from 7/27 and master has moved 16 commits since, so it needs a re-run before merge regardless.

Nothing here is structural. Happy to approve once the packaging and heap bits are sorted.

bentsherman:
No problem. This PR deserves plenty of scrutiny.

Updates:
- restored the `ant.copy` with `tofile:`
- moved test heap settings to top-level `build.gradle` -- this seemed sensible to me since `test` is already configured there
- added note about `PluginExecAware` package change to ADR -- I will document this as a breaking change in the 26.10 migration notes
- refactored `LoggerOptions` as a `@Canonical` class (matching existing codebase pattern) and added a brief javadoc
- moved `LaunchOptions` to its own file
- verified that `-XDignore.symbol.file` is inert, removed it
- fixed other nits you mentioned

Regarding your concerns about `nf-k8s` -- as far as I can tell, the current refactor is the cleanest way to support `kuberun` in CLI v1 while keeping nf-k8s as a plugin. The only other option I see is to change nf-k8s back to a core module like nf-lineage.

I have updated the ADR with the split-package consequence. I will also perform some manual tests to make sure everything still works

bentsherman:
Tested `kuberun` locally. Found a regression from #7372 and patched it here. With that fix, `kuberun` is able to use nf-k8s classes without any issues.

The only edge case is if a user explicitly declares nf-k8s in config:
```groovy
plugins { id 'nf-k8s' }
```

In this case, the config validator will warn about the `k8s` config scope being registered twice. No failure, config validates the same as before, and there is no reason to do this in the first place.

pditommaso:
Went through the updates and re-validated the parts that needed evidence. Everything from the last round is addressed — `ant.copy` with `tofile:`, heap settings consolidated in the root `build.gradle`, `LoggerOptions` as a `@Canonical` class with a javadoc, `LaunchOptions` in its own file, `-XDignore.symbol.file` gone, headers and gradle reformats cleaned up. All confirmed on `d080c32a4`.

**First, a correction to my last review.** I claimed the split package could make plugin version pinning a silent no-op. It can't, and you were right to push back. I resolved the executor at runtime with both copies present:

```
Extension executors providers=[K8sExecutor, K8sExecutor]
Replacing executor K8sExecutor with class nextflow.k8s.K8sExecutor
WINNER_CL=org.pf4j.PluginClassLoader@59bbb974
```

pf4j enumerates classpath extensions before plugin extensions, so the plugin copy always wins. Pinning holds and there's no cross-classloader hazard in practice.

### `PluginExecAware` — consider restoring the old package name

I built `:nf-cli-v1:shadowJar` from this branch and ran it against the published `nf-tower@1.29.1` straight from the registry:

```
INFO  org.pf4j.AbstractPluginManager - Start plugin 'nf-tower@1.29.1'
ERROR nextflow.cli.Launcher - @unknown
java.lang.NoClassDefFoundError: nextflow/cli/PluginExecAware
Caused by: java.lang.ClassNotFoundException: nextflow.cli.PluginExecAware
	at org.pf4j.PluginClassLoader.loadClass(PluginClassLoader.java:169)
```

`TowerPlugin` and `WavePlugin` implement the interface on the plugin class itself, so it resolves eagerly at plugin start and takes down the whole command — not just `nextflow plugin exec`. I scanned the published jar and `PluginExecAware` + `PluginAbstractExec` are the *only* two `nextflow.cli.*` classes it can no longer find; `Launcher`, `CmdAuth` and `CmdLaunch` all still resolve because `nf-cli-v1` kept the package.

So I tried moving both back to `nextflow.cli` — two renames plus five one-line imports, 8 lines total, the classes stay physically in the `nextflow` runtime module. Rebuilt, reran the same command with the same published plugin: it completes normally.

The package name isn't what buys the decoupling — the module is. And the runtime already hosts `nextflow.cli.AuthCommand`, `LaunchCommand` and `LaunchOptions`, so `nextflow.cli` is split across the two modules either way; keeping these two there is arguably more consistent than carving out `nextflow.plugin.cli`.

The signature change can stay as-is. Old implementors load fine and only fail if someone actually invokes `nextflow plugin <id>:<cmd>` — and since `PluginAbstractExec` is a trait, old subclasses carry baked-in `$Trait$Helper` calls with the old descriptor, so it's a `NoSuchMethodError` at invocation rather than at load. Narrow and loud, instead of total.

To be clear, I'm not worried about the plugin `VERSION` files — those get bumped at release time. The point is that the failure mode for *third-party* plugins you can't rebuild goes from "nothing works" to "one command is unavailable".

### Residual issue: `AuthCommand` / `LaunchCommand` nested → top-level

With the crash out of the way, this showed up underneath it:

```
ERROR org.pf4j.AbstractExtensionFinder - nextflow/cli/CmdLaunch$LaunchCommand
ERROR org.pf4j.AbstractExtensionFinder - nextflow/cli/CmdAuth$AuthCommand
```

Promoting `CmdAuth.AuthCommand` and `CmdLaunch.LaunchCommand` out of their enclosing classes breaks the same published plugin. pf4j logs and skips, so it's non-fatal, but nf-tower's `launch` and `auth` extensions silently disappear. Keeping the nested names as `interface AuthCommand extends nextflow.cli.AuthCommand` would shim it — old implementors would still satisfy the new type. Your call whether that's worth it.

### ADR needs two corrections

The `k8s` scope warning isn't limited to explicitly declaring the plugin. Plain `process.executor = 'k8s'` auto-loads nf-k8s via `PluginsFacade:520` and both copies register:

```
WARN nextflow.config.ConfigValidator - Plugin config scope `nextflow.k8s.K8sConfig` conflicts with existing scope: `k8s`
```

That's every k8s run. The mechanism is the shadow `AppendingTransformer`: `nextflow.k8s.K8sConfig` and `K8sExecutor` end up at lines 39–40 of the fat jar's `META-INF/extensions.idx`, so they register as system extensions on *every* run, k8s or not. If you'd rather not see it in the wild, deduping in `ConfigValidator.loadPluginScopes()` when the duplicate is the same class name is a couple of lines and covers future split packages too.

Second, the bundled copy silently substitutes for the plugin. With `NXF_PLUGINS_DEFAULT=false`, `executor = 'k8s'` still resolves from the app classloader — previously it wouldn't have. Benign, but worth a line in the consequences.

### Nits

- `modules/nf-lineage/src/resources/META-INF/extensions.idx` still says `2013-2025`.
- `Launcher.loggerOptions()` builds `LoggerOptions` with seven positional args including three booleans. `@Canonical` gives you the map constructor for free — `new LoggerOptions(ansiLog: …, background: …)` would make a transposition impossible.

### CI

Fresh run on `d080c32a4`, so my "needs a re-run" note is resolved. Unit tests green on 17 and 25; `test_integration`, `test_parser_v2`, `test_azure`, `test_google`, `test_docs` all pass. The four `test_aws`/`test_wave` 90-minute timeouts have the same infrastructure signature as before.

Nothing structural left from my side — the split reads clean. Whatever you decide on the package, worth updating the ADR wording before merge.

bentsherman:
- Restored `PluginExecAware` and `PluginAbstractExec` to their original packages. good catch, agree this is the right move

- I did not add anything for `AuthCommand` / `LaunchCommand`. There is never any reason to use an older `nf-tower` with a newer Nextflow version since a `nf-tower` version is pinned to each release. Documented in the ADR

- Documented the double-loading of nf-k8s via `process.executor = 'k8s'`. Updated `ConfigValidator` to suppress the duplicate-registration warning for nf-k8s

- Addressed other nits

pditommaso (approved):
All four addressed — re-validated on `bfbc2e90f` rather than taking them on trust:

- Rebuilt the fat jar and reran the case that was fatal before, with the published `nf-tower@1.29.1` pulled from the registry. Plugin starts, command exits 0. No `nextflow.plugin.cli` references left anywhere, docs import updated too.
- The `ConfigValidator` fix being keyed on the class name rather than special-casing nf-k8s is the right shape — genuine collisions between two different classes still warn. Confirmed zero `conflicts with existing scope` lines on `process.executor = 'k8s'`, with `k8s.namespace` still validating.
- On the `LoggerOptions` named args I checked propagation rather than just the compile, since a silent map-constructor mismatch under `@CompileStatic` would have been invisible: `-log` writes to the given path and `-trace nextflow.plugin` yields TRACE output. Both fine.
- ADR reads accurately now on all of it.

Fair call on `AuthCommand`/`LaunchCommand` — it fails loudly and only costs those two commands, and core plugins are rebuilt at release anyway.

Thanks Ben, and thanks for the patience across all the rounds on this one. Splitting out the three prerequisites is what made this reviewable, and the ADR is going to save someone a lot of archaeology later. Nice work.

pditommaso:
> the ADR is going to save someone a lot of archaeology later 

LOL

bentsherman:
Now we wait for the CI to be fixed...

pditommaso:
Vamos!

PR #5974, opened 2025-04-14
Document process/workflow recursion

bentsherman:
Document process and workflow recursion, based on the [original GitHub discussion](https://github.com/nextflow-io/nextflow/discussions/2521) and some personal examples.

I've been pushing this off because I expected to replace the preview version, but it's been over three years and that hasn't happened yet, so let's just document it. It will be nice to keep around for a while even after we replace it.

I have only documented the `recurse` method and not the `scan` part because `scan` is a bit more complicated to use and can be implemented with `recurse` anyway. If people come asking about it then I can document it later. Just trying to keep the docs simple for now.


docs/workflow.md
```diff
@@ -366,6 +366,54 @@ workflow {
...
+:language: nextflow
+```
+
+```{literalinclude} ../snippets/recurse-workflow.out
+:language: console
+```
+
+**Limitations**
+
+- The recursive process or workflow must have matching inputs and outputs, so that the outputs can be fed back into the inputs.
```
christopher-hakkaart, line 413:
```suggestion
- Recursive processes or workflows require that the structure and types of their outputs match those of their inputs. This is necessary to ensure that the outputs from each iteration can be correctly fed back as inputs for the next iteration.
```

docs/workflow.md
```diff
@@ -366,6 +366,54 @@ workflow {
...
+
+- Recursive workflows cannot use *reduction* operators such as `collect`, `reduce`, and `toList`, because these operators cause the recursion to hang forever after the first iteration.
```
christopher-hakkaart, line 415:
```suggestion
- Recursive workflows cannot use *reduction* operators such as `collect`, `reduce`, and `toList`, because these operators cause the recursion to hang indefinitely after the initial iteration.
```

docs/workflow.md
```diff
@@ -366,6 +366,54 @@ workflow {
...
+
+```{literalinclude} ../snippets/recurse-process.nf
+:language: nextflow
+```
+
+```{literalinclude} ../snippets/recurse-process.out
+:language: console
+```
+
+In the above example, the `count_down` process is invoked first with `params.start`, then again with the output of that invocation, and so on. The `until` method is used to limit the recursion based on a condition.
```
christopher-hakkaart, line 390:
```suggestion
In the above example, the `count_down` process is first invoked with the value `params.start`. On each subsequent iteration, the process is invoked again using the output from the previous iteration. The recursion continues until the specified condition is met, as defined by the `until` method, which terminates the recursion.
```

christopher-hakkaart (approved):
A found a couple of points unclear. I had a go an improving them. Approving as changes are optional and non-blocking.

bentsherman:
Thanks Chris, all good clarifications

PR #5975, opened 2025-04-14
Document release notes

bentsherman:
This PR adds curated release notes for each stable release of Nextflow going back to 24.04.

Notable differences:
- Only stable releases -- no patch releases or edge releases
- Cherry-picked from GitHub release notes
- Categorize by new features / enhancements / breaking changes / deprecations / misc
- Add descriptions and point to source docs where helpful

This section should make it easy for users to find all the changes for a given release summarized in one place, so that they can assess the cost/benefit of upgrading, without having to sift through all the noise in the GitHub release notes.

We might also be able to use these docs to preview the set of new features/changes in the upcoming stable release. although I'm still thinking about this one because (1) it will likely always be incomplete until right before the release and (2) we might not want to publicize upcoming features prematurely.

TODO:
- [x] 24.04
- [x] 24.10
- [x] 25.04 (preview)

### DIFF
--- a/docs/migrations/24-04.md
+++ b/docs/migrations/24-04.md
@@ -7,17 +7,17 @@
 
 <h3>Topic channels (first preview)</h3>
 
-A topic channel is a queue channel that can receive values from multiple sources based on a matching name or *topic*.
+A topic channel is a queue channel that can receive values from multiple sources, based on a matching name or *topic*.
 
-Topic channels are particularly useful for collecting metadata from various places in the pipeline, without needing to write all of the channel logic that is normally required (e.g. using the `mix` operator).
+Topic channels are particularly useful for collecting metadata from various points in a pipeline without writing all of the channel logic that is normally required (e.g., using the `mix` operator).
 
 See {ref}`channel-topic` for details.
 
 <h3>Process <code>eval</code> outputs</h3>
 
-A process `eval` output is a new type of process output which allows you to capture the standard output of an arbitrary shell command.
+A process `eval` output is a new type of process output that captures the standard output of an arbitrary shell command.
 
-The `eval` output is a useful shorthand for capturing command outputs, such as the version of a command line tool, without needing to execute the command in the task script and capturing it with a `path` or `env` output.
+This output type is a simpler way to capture command outputs, such as the version of a command-line tool, instead of using a `path` or `env` output.
 
 See {ref}`process-out-eval` for details.
 
@@ -25,13 +25,13 @@ See {ref}`process-out-eval` for details.
 
 The `resourceLimits` process directive allows you to define global limits on the resources requested by individual tasks.
 
-This directive is useful for ensuring that a task's resource requirements never exceed system-wide limits, especially when using {ref}`retry with dynamic resources <dynamic-task-resources>`, thus preventing the task from becoming unschedulable.
+This directive is useful for ensuring that a task's resource requirements never exceed system-wide limits. It is especially important for preventing tasks from becoming unschedulable when using {ref}`retry with dynamic resources <dynamic-task-resources>`.
 
 See {ref}`process-resourcelimits` for details.
 
 <h3>Job arrays</h3>
 
-The `array` directive allows a process to submit jobs as *job arrays*, which are supported by many batch schedulers as a way to submit many similar jobs in an efficient way.
+The `array` directive allows a process to submit jobs as *job arrays*, a feature used by many batch schedulers to submit many similar jobs in an efficient manner.
 
 :::{tip}
 On Google Batch, the `array` directive can be used in conjunction with the `cpus`, `memory`, and `machineType` directives to pack multiple tasks onto the same VM.
@@ -71,7 +71,7 @@ See {ref}`workflow-output-def` for details.
 
 <h3>Colored logs</h3>
 
-Nextflow's console output now features colored logs and improved organization to highglight the most important information about a pipeline while its running.
+Nextflow's console output now features colored logs and improved organization to highlight the most important information about a pipeline while it's running.
 
 <h3>AWS Fargate</h3>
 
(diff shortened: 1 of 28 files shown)


bentsherman:
You can see from the commits, I initially added more detail like code examples to the release notes and then removed them. I figure it's not really needed in most cases because each example already points to the relevant docs which usually have the exact same example.

One exception is workflow outputs, since I don't keep the code examples from older previews in the latest docs, so it's useful as a historical record

bentsherman:
I think I found an acceptable way to show release notes for the upcoming release.

- I marked it as "preview" and added a note at the top that the notes will remain incomplete until the actual release.

- I only documented features that have been merged to master, as their descriptions are likely to be stable. We can update these notes whenever we merge a feature. Once we have nightly builds, we can simply say "these features can be used in the latest nightly build."

I found writing this page to be very helpful as a high-level view of what we plan to release. I'm not too worried about making it public, since any feature merged to master is already public via GitHub PRs and the docs.

If it's too much visibility, we can simply remove it from the index so that it's hidden but still available for our own benefit.

bentsherman:
Trying some more intrusive changes -- moved the DSL1 and strict syntax guides under Migration notes, interleaved with the stable releases in chronological order. So the migration path goes naturally from bottom to top

This mixing is a bit messier than I'd prefer, but it should be a one-off thing, because language updates moving forward will be tied to stable releases

We could move the strict syntax and DSL1 guides to "Guides" at the bottom. The strict syntax guide is linked from many other places already, including the migration notes for 24.10 and 25.04.

christopher-hakkaart:
Love the overall direction. Adding a few high level thoughts here:

- This is a good level of content.
- It's distinct enough from the changelog.
- The "Preparing for strict syntax" does stick out in the side navigation, and it is more of a guide rather than notes. I'd move it to the Guides section and reference it as required. The DSL1 migration feels appropriate as it's a catch-all for DSL1 pipelines.
- I really like the consistency in headings for each set of notes (New features, enchancements, breaking changes ect).
- The Migration notes index page could be expanded to explain the purpose of this guides, and also reference the changelogs and preparing for stricy syntax pages.

I'll make some suggestions on langauge in the PR.

docs/migrations/index.md
```diff
@@ -0,0 +1,15 @@
+(migrations-page)=
+
+# Migration notes
+
+The following pages contain migration notes for each stable release of Nextflow:
```
christopher-hakkaart, line 5:
```suggestion
This section covers important information for migrating between stable Nextflow releases. It includes descriptions of new features, enhancements, breaking changes, depreciations, and other miscellaneous updates that may require action during a migration.

For a preview of upcoming changes, see the {ref}`preview <preview-page>` page. This preview is a working document that tracks the most recent modifications planned for the next stable release. As development progresses, the content may change.

Migration notes for stable releases:
```
Some musings, feel free to adapt or ignore.

docs/migrations/index.md
```diff
@@ -0,0 +1,15 @@
...
+
+```{toctree}
+:maxdepth: 1
+
+25-04
+strict-syntax
+24-10
+24-04
+dsl1
```
christopher-hakkaart, line 14: Could drop this out from this list and have another introduction explaining DSL1 is out of date.

docs/migrations/strict-syntax.md
```diff
@@ -1,30 +1,28 @@
...
 :::
 
+## Overview
+
 The strict syntax is a subset of DSL2. While DSL2 allows any Groovy syntax, the strict syntax allows only a subset of Groovy syntax for Nextflow scripts and config files. This new specification enables more specific error reporting, ensures more consistent code, and will allow the Nextflow language to evolve independently of Groovy.
 
 The strict syntax will eventually become the only way to write Nextflow code, and new language features will be implemented only in the strict syntax, with few exceptions. Therefore, it is important to prepare for the strict syntax in order to maintain compatibility with future versions of Nextflow and be able to use new language features.
 
-This section describes the key differences between the DSL2 and the strict syntax. In general, the amount of changes that are required depends on the amount of custom Groovy code in your scripts and config files.
+This page describes the key differences between the DSL2 and the strict syntax. In general, the amount of changes that are required depends on the amount of custom Groovy code in your scripts and config files.
```
christopher-hakkaart, line 21:
```suggestion
This page outlines the key differences between DSL2 and the strict syntax. The extent of required changes will vary depending on the level of custom Groovy code used within your scripts and configuration files.
```

docs/migrations/index.md
bentsherman: side question -- is depreciation vs deprecation just a British/American English thing?

bentsherman:
I removed the strict syntax page from the list. I initially moved it to Guides, but ultimately settled on keeping it in Updates so that it's more prominent, at least for now

christopher-hakkaart (changes requested):
I've added suggestions separately to (hopefully) make it easier for you to approve or decline. Mostly just subtle working changes for consistency or clarity.

docs/migrations/24-04.md
```diff
@@ -0,0 +1,134 @@
...
+
+The `eval` output is a useful shorthand for capturing command outputs, such as the version of a command line tool, without needing to execute the command in the task script and capturing it with a `path` or `env` output.
+
+See {ref}`process-out-eval` for details.
+
+<h3>Resource limits</h3>
+
+The `resourceLimits` process directive allows you to define global limits on the resources requested by individual tasks.
+
+This directive is useful for ensuring that a task's resource requirements never exceed system-wide limits, especially when using {ref}`retry with dynamic resources <dynamic-task-resources>`, thus preventing the task from becoming unschedulable.
```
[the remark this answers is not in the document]
bentsherman: Due to the conditioning of multiple teachers in school, I am physically unable to use "this" as a standalone noun 😅

docs/migrations/index.md
bentsherman: Okay. I think I'll just leave it as is

bentsherman:
Thanks for taking the time to make the individual suggestions 😄  I made a few tweaks but overall they were quite good

christopher-hakkaart:
No problem, It's easier to say "yes" vs "no" to specific changes rather than editing PRs manually.

Also, I agree with your point about using "this; I feel a similar way. I try to use it sparsely when it's either obvious or can be made obvious what "this" is, e.g., "This directive/change/operator. "Clarity always comes first, so if you think it's not clear, then we can swap it back or make it clearer.

christopher-hakkaart (approved):
It all looks and reads really well. Looking forward to seeing how it evolves/grows over time.

Issue #2695, opened 2022-03-04
Nextflow - SLURM job interface disconnect


nickhsmith:
## Bug report 

I am noticing a problem with the Nextflow-SLURM interface when processing large number of jobs. For example, I am running a workflow and have spawned ~10,000 tasks (50 samples * 150 intervals/sample) + downstream tasks) using the nf-core sarek workflow. However, I notice that my tasks are hanging, and as a result, nothing is being submitted to SLURM despite the jobs being completed.

### Expected behavior and actual behavior

Nextflow generates a task. Nextflow submits the task to SLURM. SLURM runs the job. SLURM finishes the job (creating `.exitcode` with value 0). Nextflow recognizes this and submits the next task.

This last step of recognizing a finished job seems to be failing.

### Steps to reproduce the problem

More tangibly, I have used `queueSize` to limit my workflow to 100 jobs at 1 time. Some of those jobs will finish (`.exitcode` exists with the value 0 in the local work dir (`work/76/69b9b4...`)) BUT these jobs are still considered RUNNING when I look at the `.nextflow.log` file (see last photo)

Looking through the log file. I see that the SLURM job ID `636541` is submitted at 15:17  on March 2
It finishes and creates the .exitcode file (with a successful 0 exit) at 18:10. SLURM thinks the job is completed at this point.
![slurm finished job](https://user-images.githubusercontent.com/7143411/156730428-cf5bf2e1-854e-43bf-9e3f-a5106ee4d95f.png)

but as of yesterday March 3 there is no completed status of this job, as a result SLURM won't start any of the remaining ~7000 jobs in the task scheduler)
```
Mar-03 05:26:48.603 [Task submitter] DEBUG n.processor.TaskPollingMonitor - %% executor slurm > tasks in the submission queue: 6714 -- tasks to be submitted are shown below
```
![working dir](https://user-images.githubusercontent.com/7143411/156730578-094babcf-6bee-447e-847f-0b7501f0bf6b.png)
![nextflow log](https://user-images.githubusercontent.com/7143411/156730611-526e69bb-d6ed-4686-b80d-d7477b12d142.png)

Any help would be appreciated, but is there a reason that the completed jobs aren't being recognized as such, or what I can do to force a re-recognition? Thanks in advance

### Environment 
`nextflow info`
  Version: 21.10.6 build 5660
  Created: 21-12-2021 16:55 UTC (17:55 CEST)
[...]

PR #5976, opened 2025-04-15
Solve issue with losing track of SLURM jobs using cephfs

riederd:
The changes solve an issue with losing track of SLURM jobs on systems
    using cephfs when the number of jobs is high.
    
    Test case: running a pipeline that creates thousands of jobs on a
    SLURM cluster with a queuesize of 600 and cephfs (reef 18.2.2) as
    shared filesystem for the nextflow workDirs.
    
    There are existing open issues and mentions of this problem on the
    git issue tracker as well i.e. #2695, #5630, #5650
    
    As @burcarjo already tried in #5650, adding `ceph` as shared filesystem
    is not solving the issue, as, at least in our tests, it never reaches
    the relevant code which should trigger a metadata refresh.
    
    The proposed fix does a metadata refresh in multiple locations of the
    code and in our tests we were able to run a pipeline with > 21k jobs without
    any hang. Without the proposed changes there was no way to get this
    pipeline to completion in a single run.
    
    As the proposed changes refresh the metadata nearly once per job in
    `GridTaskHandler.groovy` we also tried without refresh in this part
    however it resulted in stalling again.


riederd:
Yeah basically yes, I did run an ablation analysis and looked at the `-trace` output to identify possible places where we can try to refresh the metadata.

pditommaso:
I'm vey hesitant regarding this PR, having a large number of jobs, syncing continuously the file system can lead to a network congestion.

riederd:
I understand, in my tests the number of syncs/directory listings was roughly corresponding the number of submitted jobs (21k jobs ~ 20k refresh), we did not see any network issues.

 I can try to omit the `sync` call though and only do directory listing.

bentsherman:
Closing since we didn't reach consensus on this PR due to the added complexity

It might be better to explore these cephfs specific improvements in a plugin

Issue #5983, opened 2025-04-18
Update vm list sizes for Azure to current list, important VMs are missing.


rijssel:
## Bug report 

(Please follow this template by replacing the text between parentheses with the requested information)

### Expected behavior and actual behavior

When running nextflow I expect vm sizes that are available in Azure to be usable. When trying to use Standard_NC40ads_H100_v5 machines in swedencentral (often referenced by Azure to be the default location in Europe to use for GPU based workloads due to space issues in west europe) I get an error that the vmsize is not available there.

### Steps to reproduce the problem

Create a pool in Azure with NC40ads_H100 nodes and use queue: named_pool as config param in your azure batch config file.

### Program output 

Caused by:
  Unable to find size for VM name 'standard_nc40ads_h100_v5' and location 'swedencentral'

### Environment 

* Nextflow version: 24.10.5 build 5935 & 25.02.3-edge build 5941
* Java version: [?]
* Operating system: [macOS, Linux, etc]
* Bash version: (use the command `$SHELL --version`)

### Additional context

Its clear that in the file: nextflow\plugins\nf-azure\src\resources\nextflow\cloud\azure\vm-list-size-swedencentral.json the vmsize is not available. The last update was 10 months ago and Azure has a whole list of new vm sizes over all regions. Kindly request to run the update script to fill these files with fresh info again. If needed/handy I would do it myself and open a PR. Its blocking our workflow at the moment.

bentsherman:
Hi @rijssel , feel free to run the update script and submit a PR

rijssel:
> Hi [@rijssel](https://github.com/rijssel) , feel free to run the update script and submit a PR

@bentsherman Provided PR here: https://github.com/nextflow-io/nextflow/pull/5985. Let me know if there's anything else I can do to get this working :)

PR #5985, opened 2025-04-18
Update for VM sizes and regions required for Azure Batch

rijssel:
Updated the regions and VM sizes for Azure batch as mentioned in issue: [3004533060](https://github.com/nextflow-io/nextflow/issues/5983#issue-3004533060). 

Also a powershell script was added to obtain only the relevant regions, as the old bash script obtained all regions, even if they were not associated to a compute, causing the get VM size script to add additional unused JSONs.


bentsherman (approved):

pditommaso:
Worth running e2e tests 👉  de05ecc6

pditommaso:
Yes, the test was Platform

PR #5990, opened 2025-04-20
Use the same antlr4 dep as Groovy 4

bentsherman:
Use the same antlr4 runtime as Groovy 4: https://github.com/apache/groovy/blob/deb89c45afe270e09cacd6a79408bfd527b5747a/build.gradle#L85

Should revert the recent increase in build size likely caused by having two antlr runtimes in the JAR.

I also need to check a few things with the language server. But let's see how the tests respond


bentsherman:
Everything looks good with the language server. Ready to merge 😄

bentsherman:
This change reduced the language server JAR from ~23 to ~13 MB.

pditommaso:
Nice

PR #5994, opened 2025-04-22
Make executor.pollInterval findable.

christopher-hakkaart:
Searching for "poll" did not find `executor.pollInterval`. Adding "poll" to the description makes it findable.

cc @kenibrewer

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -673,7 +673,7 @@ The following settings are available:
 : Specifies Platform LSF *per-task* memory reserve mode. See {ref}`lsf-executor`.
 
 `executor.pollInterval`
-: Determines how often to check for process termination. Default varies for each executor (see below).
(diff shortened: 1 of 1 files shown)


docs/reference/config.md
```diff
@@ -673,7 +673,7 @@ The following settings are available:
 : Specifies Platform LSF *per-task* memory reserve mode. See {ref}`lsf-executor`.
 
 `executor.pollInterval`
-: Determines how often to check for process termination. Default varies for each executor (see below).
+: Determines how often to poll for process termination. Default varies for each executor (see below).
```
pditommaso, line 676: Seems both not accurate. I'd say something like "Controls how often the system checks if the process has terminated" or "Defines the polling frequency for process termination detection"

christopher-hakkaart: Good point. Suggestion applied.

pditommaso (approved):

PR #5995, opened 2025-04-22
nf-lang updates

bentsherman:
Minor cleanup to nf-lang discovered as part of https://github.com/nextflow-io/language-server/pull/110


modules/nf-lang/src/main/java/nextflow/script/control/Compiler.java
```diff
@@ -58,6 +60,27 @@ protected GroovyClassLoader classLoader() {
         return compilationUnit().getClassLoader();
     }
 
+    public SourceUnit newSourceUnit(File file) {
```
pditommaso, line 63: Nit, `newSomething` seems alluding something new (vs old). Something like "create", "make", etc may sound better

bentsherman: fair enough

pditommaso (approved):

PR #5996, opened 2025-04-22
feat: Set maximum job duration for Azure Batch

adamrtalbot:
Adds a maximum job duration for Azure Batch, after which all jobs will terminate which will terminate them automatically and remove them from quota. This acts as a killswitch to prevent Job quota being used up if Nextflow dies unexpectedly.

Signed-off-by: adamrtalbot <12817534+adamrtalbot@users.noreply.github.com>


pditommaso:
Is not the task already constrained by the container max wall time 👉 5c11a0d4? in any case, agree with Ben, we should avoid introducing yet another setting

bentsherman:
how about `jobIdleTimeout`

pditommaso:
Java doc for task says: 

> Set the maxWallClockTime property: The maximum elapsed time that the Task may run, measured from the time the Task starts. If the Task does not complete within the time limit, the Batch service terminates it. If this is not specified, there is no time limit on how long the Task may run.


Java doc for job says: 

> Set the maxWallClockTime property: The maximum elapsed time that the Job may run, measured from the time the Job is created. If the Job does not complete within the time limit, the Batch service terminates it and any Tasks that are still running. In this case, the termination reason will be MaxWallClockTimeExpiry. If this property is not specified, there is no time limit on how long the Job may run.


I understand the rationale 

> When Nextflow exits abruptly, it leaves them in active state which uses up quota and requires them to be manually deleted.

However this may break long running jobs (tho 7 days could be quite unusual). 

Enabling cleanup by default, would not improve the problem? 

https://github.com/nextflow-io/nextflow/blob/de05ecc6eb6d93343938ea27925e8a0f8e5dc171/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L995-L998

bentsherman:
This timeout is for when a pipeline is killed before it can cleanup the jobs

pditommaso:
Let's use `jobMaxWallClockTime` that consistent with Batch API with 30d as default

plugins/nf-azure/src/test/nextflow/cloud/azure/batch/AzBatchServiceTest.groovy
```diff
@@ -941,4 +945,42 @@ class AzBatchServiceTest extends Specification {
...
+        when:
+        // This is what happens inside createJob0 for setting constraints
+        def content = new BatchJobCreateContent('test-job', new BatchPoolInfo(poolId: 'test-pool'))
+        if (exec.getConfig().batch().jobMaxWallClockTime) {
+            final constraints = new BatchJobConstraints()
+            final long millis = exec.getConfig().batch().jobMaxWallClockTime.toMillis()
+            final java.time.Duration maxWallTime = java.time.Duration.ofMillis(millis)
+            constraints.setMaxWallClockTime(maxWallTime)
+            content.setConstraints(constraints)
+        }
```
adamrtalbot, line 964: This doesn't feel like a very good test...but I can't see how to mock the methods inside createJob0? Or is this an indication I should split up the method to a createJobConstraints and createJob0 method?

adamrtalbot: Done in a3dbc7c4beaf63f9a2a7c98587be55f5afe18858

christopher-hakkaart (approved):
Docs look great.

docs/reference/config.md
```diff
@@ -344,6 +344,11 @@ The following settings are available:
 `azure.batch.location`
 : The name of the batch service region, e.g. `westeurope` or `eastus2`. This is not needed when the endpoint is specified.
 
+`azure.batch.jobMaxWallClockTime`
+: :::{versionadded} 25.04.0-edge
```
pditommaso, line 348:
```suggestion
: :::{versionadded} 25.04.0
```

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -945,7 +961,11 @@ class AzBatchService implements Closeable {
                 apply(() -> client.updateJob(jobId, jobParameter))
             }
             catch (Exception e) {
-                log.warn "Unable to terminate Azure Batch job ${jobId} - Reason: ${e.message ?: e}"
+                if (e.message?.contains('Status code 409') && e.message?.contains('JobCompleted')) {
```
pditommaso, line 964: Using a static exception type is may be better, see for example 

https://github.com/nextflow-io/nextflow/blob/master/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L873-L879

docs/reference/config.md
```diff
@@ -344,6 +344,11 @@ The following settings are available:
...
+  :::
+: The maximum elapsed time that jobs may run, measured from the time they are created. If jobs do not complete within this time limit, the Batch service terminates them and any tasks still running (default: `7d`).
```
pditommaso, line 350:
```suggestion
: The maximum elapsed time that jobs may run, measured from the time they are created. If jobs do not complete within this time limit, the Batch service terminates them and any tasks still running (default: `30d`).
```

pditommaso (approved):
Neat

qup9:
I am dealing with stuck active jobs keeping the pool alive and Tower RUns in a "Running" state.:

I see why you cannot set the maxWallClockTime on Jobs because you don't know how long until the cluster even starts the job.

In my opinion, jobMaxWallClockTime should apply to tasks under nf-workflow-*; 

I am looking at nf-workflow-* Job;  State=Active: Max wall-clock time --> Unlimited; The Sub-Task: Max wall-clock time --> Unlimited
This is where the nextflow command is run; 
Setting maxWallClockTime should terminate the container via SIGTERM, SIGKILL; The container will cleanup if started with containerRunOptions --rm


When I look at a sub-job, ie job-4a5faaea9b733549beba-* State=Active; Max wall-clock time --> Unlimited and there are no Tasks
This keeps tower in a "running" state and the node in a busy state preventing scale-down.

Consulting the node, Several active nf- jobs with active tasks associated. I terminated those tasks and the node goes Idle and will deallocate.

If you can at least get the nf-workflow-* jobs under control, Tower will be happy and possibly the nodes. If you can respond to SIGTERM and Terminate child Jobs, that may work.

I have other active sub-jobs, ie job-* that have completed or no tasks or completed with failure tasks. No idea why the sub-jobs are not "Completed"

PR #6002, opened 2025-04-23
nf-lang: Add unit tests for formatting

bentsherman:
This PR adds more unit tests for formatting statements and expressions (and fixes a few bugs I found along the way)

As a bonus it also tests that these code snippets are parsed correctly

From now on, if someone reports a parsing/formatting bug, we need only add a regression test to this test suite


christopher-hakkaart (approved):
Docs look great.

PR #6003, opened 2025-04-24
Data lineage programmatic API

jorgee:
Initial fromLineage factory implementation


bentsherman:
I think we can have a nice unification of the CLI and programmatic API here

Viewing a single LID:
- CLI: `view lid://<hash>[/<path>]` -> use `jq` to inspect further
- API: `channel.fromLineage(`lid://<hash>[/<path>]')` -> use json-path param or `jsonPath` function to inspect further

Querying a collection of LIDs:
- CLI: `find <name=value> <name=value> ...`
- API: `channel.queryLineage(foo: 'bar', baz: 'qux')` -> returns queue channel of items, use `map` operator with `jsonPath` function to inspect further

This analysis suggests that the json-path functionality should be provided as a standalone function so that it can be used in different ways. You can add it to `Nextflow.groovy`.

It might also make sense to refactor `channel.fromLineage` as a function instead of a channel factory, e.g. `lineage(lid)`, because as a channel factory it will always return a value channel which is a needless constraint. The `queryLineage` factory makes sense because it can emit results while it's querying.

Finally, I think that `find` / `queryLineage` should really just use key-value pairs, rather than the URI query syntax which is unnecessary and error-prone

jorgee:
I have pushed what I showed today.

- Files published with `PublishDir` included in outputs
- `fromLinage` channel factory: It should return the same as CLI `view` command:
Usage: `Channel.fromLineage("lid://xxxx")`
-`queryLineage` channel factory: It return a channel with the `LinPaths` matching with the queryString. Almost the same as the CLI `find` command
Usage: `Channel.queryString("type=FileOutput&annotations.value=test")`

modules/nextflow/src/main/groovy/nextflow/Channel.groovy
```diff
@@ -657,4 +660,42 @@ class Channel  {
...
+
+    private static void fromLineage0(DataflowWriteChannel channel, String uri) {
+        final operation = Plugins.getExtension(LinChannelEx)
+        if( !operation )
+            throw new IllegalStateException("Unable to load lineage extensions.")
+        def future = CompletableFuture.runAsync( { operation.viewLineage(session, channel, new URI(uri)) } as Runnable)
+        future.exceptionally(this.&handlerException)
+    }
+
+    static DataflowWriteChannel queryLineage(String queryString) {
```
bentsherman, line 682: Let's just use a Map here instead of query string. I would apply the same change to the `find` command as well. There is no need to add the extra complexity of URL encoding

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -381,7 +381,7 @@ class Session implements ISession {
         this.dag = new DAG()
 
         // -- init output dir
-        this.outputDir = FileHelper.toCanonicalPath(config.outputDir ?: 'results')
+        this.outputDir = FileHelper.toCanonicalPath(config.outputDir ?: config.navigate('params.outdir') ?: 'results')
```
bentsherman, line 384: Let's keep this PR focused on the channel factories, we can address publishDir support separately

modules/nextflow/src/main/groovy/nextflow/file/PathVisitor.groovy
```diff
@@ -103,6 +103,27 @@ class PathVisitor {
...
+        else
+            applyGlobPattern0(filePattern)
+    }
+
+    private static boolean isQuery(Path filePattern) {
+        log.debug("Checking if query: $filePattern.class ")
+        return filePattern instanceof QueryablePath && (filePattern as QueryablePath).hasQuery()
+    }
+
+    private boolean applyQueryablePath0(QueryablePath path) {
```
bentsherman, line 118: We don't need to add query support to `fromPath` because it will already be supported through `queryLineage`. Besides I would like to get away from using URI query params everywhere.

The `queryLineage` factory should return a channel of metadata objects, so you can chain it with a `map` operator to extract the actual files

bentsherman: > Maybe the best is returning the lid, letting the user do what they want according to the query.

This is exactly it. The `lineage` function and `queryLineage` factory should return just the metadata descriptions. Users can compose these with other operators to do things like get the actual files.

Meanwhile, `fromPath` should only be used to get individual files.

jorgee:
Made commented changes:

- `fromLineage` factory -> to `lineage` operator (still using fragments to be merged with #6011)
- `queryLinage` factory and `find` command moved from query string to key-value map
- remove `fromPath` and `publishDir`

Usage example:

```
process ls {
        input:
                path('file')
        output:
                stdout
        script:
        """
        ls -l file
        """
}

workflow {
        // Print lineage metadata
        channel.of("lid://xxxx") | lineage | view
        
        // Get Ouputs containing a certain annotation experiment=test and use them in a task
        channel.queryLineage('type': 'FileOutput', 'annotations.key': 'experiment, 'annotations.value': params.value) | ls | view
 
        // Query lineage metadata and print
        channel.queryLineage('type': 'WorkflowOutput', 'workflowRun': 'lid://xxxx') | lineage | view

}

```

modules/nextflow/src/main/groovy/nextflow/Session.groovy
jorgee: removed

jorgee:
Last changes according last discussions:
- `lineage` moved from operator to function. Also returning objects instead of String with json serializations
- Removed support for query in `view` and `fromPath`. If user add a query in a LinPath URI, it will be ignored. It prints a warning message.
- Fragment is also validated at LinPath creation to detect errors earlier. Before, incorrect fragments were detected when accessing the LinPath.

pditommaso:
Give a try to this i'm not convinced about `lineage` either as function or operator. 

I find counterintuitive especially when using it with `queryLineage`. I would expect the latter to emit map object corresponding to the matched metadata. Could even be interesting give an option to decide the object format e.g. string (json), Path or Map

bentsherman:
From our discussion today:
- remove the `lineage` function
- refactor `queryLineage` to always return `Channel<Path>`
  - allows the following fields from `FileOutput` as named args
    - workflowRun
    - taskRun
    - annotations
  - return the real paths associated with `FileOutput` records
  - multiple annotations are treated as AND

A possible example usage:
```groovy
channel.queryLineage(workflowRun: '...', annotations: ['my-label', 'foo=bar'])
```

Focusing on `FileOutput` allows us to document a specific set of keys that can be used as named args to the factory.

jorgee:
> * return the real paths associated with `FileOutput` records

real path or the LinPath?. If returning LinPath, it is checking the integrity when accessing. I think it is a better option than just returning the real path.

pditommaso:
Indeed 👍

bentsherman:
Maybe I'm misunderstanding. If you return the LinPath, will that just be the JSON metadata-as-a-file or the actual file? Because the user will want to use the actual file

jorgee:
It is both, when the URI LinPath is pointing to a FileOutput, it is just a wrapper of the real file, including the checksum validation. When is pointing to other metadata type, it is the JSON metadata-as-a-file.

bentsherman:
Perfect 👍

jorgee:
Implemented latests comment:

- `lineage` function removed
- `queryLineage` refactored to return Channel<Path>. Allowed params: `workflowRun`, `taskRun`, and `labels`
- multiple labels are treated as AND
- Conflicts with master solved
- Documentation updated

Usage Example:

`channel.queryLineage(workflowRun: 'lid://xxxx', taskRun: 'lid://xxx', annotations: ['my-label', 'foo=bar'])`

bentsherman (approved):

Issue #6006, opened 2025-04-24
Offline mode (NXF_OFFLINE='true') prevents running pipelines from local file:/ bare Git repositories


nz-mtra:
### Expected behavior and actual behavior

When attempting to run a Nextflow pipeline using a local bare Git repository specified with the file:/ prefix (e.g., file:/path/to/my/pipeline.git), enabling offline mode with `NXF_OFFLINE='true'` causes Nextflow to fail prematurely. It throws an AbortOperationException stating that automatic downloads are disabled, even though the target is a local path that doesn't require remote downloading.

#### Expected behavior
Nextflow should recognize file:/ as a local resource, bypass the remote download check associated with being offline, and proceed to use the AssetManager to set up and run the pipeline from the specified local bare Git repository path.

#### Actual behavior
An AbortOperationException is thrown with a message similar to:
Unknown project 'local/pipelineName' -- NOTE: automatic download from remote repositories is disabled

### Steps to reproduce the problem

1. Make a clean installation of nextflow.
2. Set NXF_OFFLINE variable to true (`export NXF_OFFLINE='true'`)
3. Download a bare git repo of a pipeline. (`git clone git@github.com:nf-core/rnaseq.git --bare`)
4. Run the pipeline: `nextflow run 'file:/path/to/local_pipelines/rnaseq.git' -r 3.18.0 -profile test`

### Program output 
```
 N E X T F L O W   ~  version 24.10.6

Unknown project `local/rnaseq` -- NOTE: automatic download from remote repositories is disabled
```

[.nextflow.log](https://github.com/user-attachments/files/19895750/default.nextflow.log)

### Environment 

* Nextflow version: 24.10.6
* Java version: openjdk - 17.0.14
* Operating system: linux
* Bash version: GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)

### Additional context
Reproducibility seems to depend on the .nextflow.

This issue is on line 588 of [nextflow/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy](https://github.com/nextflow-io/nextflow/blob/453ab6a6d2bdae76a3372dbef697f6175f3c576c/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy#L588).
The AssetManager component is capable of handling these file:/ URIs correctly, but the current offline check blocks this unnecessarily. 

Running the same local bare Git pipeline with `NXF_OFFLINE='false'` works even though the server does not have internet.

[...]

PR #6007, opened 2025-04-24
Fix: Allow running pipelines from local file:/ Git repositories in offline mode

nz-mtra:
### Description
This PR fixes an issue where Nextflow's offline mode incorrectly prevented running pipelines specified via local Git repository URIs starting with file:/. The existing offline check would throw an exception before the AssetManager could handle the local path.

This change modifies the offline check to specifically permit file:/ URIs, allowing AssetManager to correctly set up the pipeline from the local clone even when offline mode is active.

### Related issue
Fixes [#6006](https://github.com/nextflow-io/nextflow/issues/6006)

### Changes made:
On line 588 of [nextflow/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy](https://github.com/nextflow-io/nextflow/blob/453ab6a6d2bdae76a3372dbef697f6175f3c576c/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy#L588), changed  `if( offline )` to `if( offline && !pipelineName.startsWith('file:/') )`.

This ensures the `AbortOperationException` is only thrown in offline mode when the pipelineName does not represent a local file:/ path.

### How to test
1. Clone the branch, go to branch, compile
2. Set NXF_OFFLINE variable to true (`export NXF_OFFLINE='true'`)
3. Download a bare git repo of a pipeline to a local. (Eg. `git clone git@github.com:nf-core/rnaseq.git --bare`)
4. Run the pipeline from the nextflow repo: `./launch.sh run 'file:/path/to/local_pipelines/rnaseq.git' -r 3.18.0 -profile test`


bentsherman:
Hi @nz-mtra , sorry for the delayed response. Can you add a test to verify this change? It could be a unit test in CmdRunTest or an end-to-end test in the `tests` folder, as long as it checks against a real local pipeline (can be a minimal dummy pipeline)

bentsherman:
Closing in favor of #7035

PR #6011, opened 2025-04-25
Remove URL fragment in lineage IDs

bentsherman:
This PR removes the use of URL fragments in LIDs, since `jq` can be used on the command line. The `#output` shortcut is replaced by adding both WorkflowLaunch and WorkflowRun to the history log.

- Rename `WorkflowRun` -> `WorkflowLaunch`
- Rename `WorkflowOutput` -> `WorkflowRun`
- Add launch LID / run LID to `nextflow lineage list`
- Remove `TaskOutput` since it is not used
- Remove the use of URL fragments


bentsherman:
This PR removes a lot of unnecessary complexity around the LID filesystem that I feel is getting in the way of core use cases.

There are a few tests still failing, mainly because the `view` command now goes through `LinPath` instead of `LinUtils.query()` to resolve the LID, and it is currently trying to return the actual path of FileOutputs instead of the metadata description.

bentsherman:
Before we move this forward, I think we should decide whether we actually want to embed the workflow output value in the metadata. Like I mentioned, a pipeline with a `samples` output could produce 100,000 samples, which would be saved as a giant list in the metadata.

Alternatively, the user can save the output to an index file (i.e. samplesheet) and just reference that index file in a downstream pipeline, like they already do. We could simply reference this index file in the metadata instead of the contents. I intend to explore this as part of the TraceObserverV2 proposal.

In that case, maybe we could drop the use of URI fragments entirely. Users already have the index file as a drop-in replacement for samplesheets, and traversing the metadata can already be done more effectively via `jq` or json-path.

So I think my main goal for the 25.04 release is to remove the URI syntax overloading (fragment, query params) in favor of something more familiar and flexible.

modules/nf-lineage/src/main/nextflow/lineage/model/TaskRun.groovy
```diff
@@ -64,20 +64,12 @@ class TaskRun implements LinSerializable {
      * Architecture defined in the Spack environment used for the task run
      */
     String architecture
-    /**
-     * Global variables defined in the task run
-     */
-    Map globalVars
```
jorgee, line 70: GlobalVars are used to calculate the hash. I think it shouldn't be removed

modules/nf-lineage/src/main/nextflow/lineage/LinObserver.groovy
```diff
@@ -259,17 +228,23 @@ class LinObserver implements TraceObserver {
             normalizer.normalizePath(task.getCondaEnv()),
             normalizer.normalizePath(task.getSpackEnv()),
             task.config?.getArchitecture()?.toString(),
-            task.processor.getTaskGlobalVars(task),
```
jorgee, line 262: Same as the TaskRun comment, I think it shouldn't be removed because it it used in the hash computation

jorgee:
> Alternatively, the user can save the output to an index file (i.e. samplesheet) and just reference that index file in a downstream pipeline, like they already do. We could simply reference this index file in the metadata instead of the contents. I intend to explore this as part of the TraceObserverV2 proposal.

AFAIK, the index file is not mandatory, users need to indicate they want to write the index file, right? In the case, they do not write it, what we should put as output? Maybe we should write the index file in all the cases and publish to the file if indicated or store as lineage metadata.

jorgee:
> There are a few tests still failing, mainly because the `view` command now goes through `LinPath` instead of `LinUtils.query()` to resolve the LID, and it is currently trying to return the actual path of FileOutputs instead of the metadata description.

I think when `view`, it should just try to load and check the fragment. In LinPath it is also checking if the path is a subpath of a description but it is not needed in this case.

modules/nf-lineage/src/main/nextflow/lineage/model/TaskRun.groovy
bentsherman: I was going to remove because using global vars is an anti-pattern and will likely become an error under the strict syntax -- but now I realize it will still be useful with the `diff` command. So we should keep it for now and remove it in a future version where the task hash does not include global variables

modules/nf-lineage/src/main/nextflow/lineage/fs/LinPath.groovy
```diff
@@ -302,7 +271,7 @@ class LinPath implements Path, LogicalDataPath {
     @Override
     Path getFileName() {
         final result = Path.of(filePath).getFileName()?.toString()
-        return result ? new LinPath( fragment, query, result, null) : null
+        return result ? new LinPath(query, fragment, result, null) : null
```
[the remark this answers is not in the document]
jorgee: Included in #6003.

bentsherman:
Update this PR based on the latest changes, focusing on removing the use of URL fragments entirely in the LID filesystem. This makes the user experience simpler (don't need to remember the `#output` hack) and removes a lot of complex code.

Every workflow execution now has a "launch LID" and "run LID". The former is added to the history log when the workflow begins, the latter is added when the workflow completes.

Here's a simple test you can run to get started:
```bash
rm -rf .lineage/
nextflow run rnaseq-nf -r lineage -profile conda -resume --labels foo,bar
nextflow lineage list
```

@jorgee if these changes make sense to you, can you help me fix the tests? I think they are the same ones as before

bentsherman:
Expected output:
```console
$ nextflow lineage list
TIMESTAMP               RUN NAME        SESSION ID                              LAUNCH LID                              RUN LID                               
2025-05-02 19:06:15 CDT stoic_shaw      bc79451f-c573-4b7d-8e7c-697be8d9cefc    lid://cd7197c02ab1250eafc2bf7499715e5f  lid://304c57e48ab6b324715ad2c5ba55b25e

$ nextflow lineage view lid://304c57e48ab6b324715ad2c5ba55b25e
{
  "type": "WorkflowRun",
  "createdAt": "2025-05-02T19:06:16.160559118-05:00",
  "workflowLaunch": "lid://cd7197c02ab1250eafc2bf7499715e5f",
  "output": [
    {
      "type": "Path",
      "name": "summary",
      "value": "lid://cd7197c02ab1250eafc2bf7499715e5f/summary/multiqc_report.html"
    },
    {
      "type": "Collection",
      "name": "samples",
      "value": "lid://cd7197c02ab1250eafc2bf7499715e5f/samples.json"
    }
  ]
}
```

modules/nf-lineage/src/main/nextflow/lineage/model/WorkflowRun.groovy
bentsherman, line 1: Now that we have the concept of a completed workflow run, it would be good to add the status to this entity so that you know whether the run failed or not

bentsherman:
I added the completion status to the WorkflowRun. It can be SUCCEEDED, FAILED, or CANCELLED, consistent with platform terminology

jorgee:
My main concern was about the two LIDs and the path of the workflow output files. They are based on the `LAUNCH LID` but the users could expect to have it in the `RUN LID`.

bentsherman:
Could it be done the other way? I would expected the final workflow run hash to include the workflow outputs as components, so there would be no way for a workflow output to refer to the final workflow run that is referring to it

PR #6014, opened 2025-04-28
Trace observer v2

bentsherman:
- Add new TraceObserverV2 interface, replacing method overloads with custom event types
- Revert TraceObserver to be compatible with Nextflow 24.10
- Update Session to accept v1 and v2 trace observers
- Update lineage observer to use TraceObserverV2
- Fix issues with workflow output annotations

TODO:
- [x] how to handle workflow output value vs index?

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -22,6 +22,7 @@ import java.nio.file.Paths
 import java.util.concurrent.ConcurrentLinkedQueue
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.Executors
+import java.util.function.Consumer
 
 import com.google.common.hash.HashCode
 import groovy.transform.CompileDynamic
@@ -253,7 +254,9 @@ class Session implements ISession {
 
     private int poolSize
 
-    private List observers = Collections.emptyList()
+    private List<TraceObserver> observersV1 = Collections.emptyList()
+
+    private List<TraceObserverV2> observersV2 = Collections.emptyList()
 
     private Closure errorAction
 
@@ -435,10 +438,9 @@ class Session implements ISession {
         this.disableRemoteBinDir = getExecConfigProp(null, 'disableRemoteBinDir', false)
         this.classesDir = FileHelper.createLocalDir()
         this.executorFactory = new ExecutorFactory(Plugins.manager)
-        this.observers = createObservers()
-        this.statsEnabled = observers.any { ob ->
-            (ob instanceof TraceObserver && ob.enableMetrics()) || (ob instanceof TraceObserverV2 && ob.enableMetrics())
-        }
+        this.observersV1 = createObservers()
+        this.observersV2 = createObserversV2()
+        this.statsEnabled = observersV1.any { ob -> ob.enableMetrics() } || observersV2.any { ob -> ob.enableMetrics() }
         this.workflowMetadata = new WorkflowMetadata(this, scriptFile)
 
         // configure script params
@@ -466,7 +468,7 @@ class Session implements ISession {
      * @return A list of {@link TraceObserver} objects or an empty list
      */
     @PackageScope
-    List createObservers() {
+    List<TraceObserver> createObservers() {
 
         final result = new ArrayList(10)
 
@@ -479,14 +481,18 @@ class Session implements ISession {
             result.addAll(f.create(this))
         }
 
+        return result
+    }
+
+    @PackageScope
+    List<TraceObserverV2> createObserversV2() {
+        final result = new ArrayList(10)
         for( TraceObserverFactoryV2 f : Plugins.getExtensions(TraceObserverFactoryV2) ) {
             log.debug "Observer factory (v2): ${f.class.simpleName}"
             result.addAll(f.create(this))
         }
-
         return result
     }
-
 
     /*
      * intercepts interruption signal i.e. CTRL+C
@@ -983,6 +989,7 @@ class Session implements ISession {
         errorMessage << message.toString()
         return false
     }
+
     /**
      * Register a shutdown hook to close services when the session terminates
      * @param Closure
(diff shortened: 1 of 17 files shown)


modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -248,7 +253,7 @@ class Session implements ISession {
 
     private int poolSize
 
-    private List<TraceObserver> observers = Collections.emptyList()
+    private List observers = Collections.emptyList()
```
pditommaso, line 256: Then it would be preferable to use another typed list `List<TraceObserverV2> observersV2`

modules/nextflow/src/main/groovy/nextflow/trace/event/WorkflowOutputEvent.groovy
```diff
@@ -0,0 +1,48 @@
...
+@CompileStatic
+class WorkflowOutputEvent {
+    /**
+     * The name of the workflow output.
+     */
+    String name
+    /**
+     * The type of the workflow output.
+     */
+    String type
```
pditommaso, line 39: What `type` can be? some examples could be useful

pditommaso (changes requested):

modules/nextflow/src/main/groovy/nextflow/trace/event/WorkflowOutputEvent.groovy
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * The value of the workflow output. It will be null if the index file was specified.
+     */
+    Object value
```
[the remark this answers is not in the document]
bentsherman: It can be anything, it's whatever the published channel contains

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -248,7 +254,9 @@ class Session implements ISession {
 
     private int poolSize
 
-    private List<TraceObserver> observers = Collections.emptyList()
+    private List<TraceObserver> observersV1 = Collections.emptyList()
```
pditommaso, line 257: Would make sense to mark as `@Deprecated`

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -248,7 +254,9 @@ class Session implements ISession {
...
+
+    private List<TraceObserverV2> observersV2 = Collections.emptyList()
```
pditommaso, line 259: Is not ` Collections.emptyList()` immutable?

modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -986,40 +1005,19 @@ class Session implements ISession {
...
-            final observer = observers.get(i)
-            try {
-                observer.onProcessCreate(process)
-            }
-            catch( Exception e ) {
-                log.debug(e.getMessage(), e)
-            }
-        }
+        notifyAll(observersV1, ob -> ob.onProcessCreate(process))
+        notifyAll(observersV2, ob -> ob.onProcessCreate(process))
```
pditommaso, line 1009: Neat!

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
```diff
@@ -189,42 +190,55 @@ class PublishOp {
...
 
     /**
-     * Once all values have been published, publish the index
-     * and write it to a file (if enabled).
+     * Get or resolve the annotations of a workflow output.
+     *
+     * @param annotations Map | Closure<Map>
+     * @param value
+     */
+    protected static Map getAnnotations(annotations, value) {
```
pditommaso, line 198: Still not fully sure about this. Wonder if it should not be a list of string and call it `labels` for consistency with process labels. this would also make it possible to propagate the process labels to the TaskRun model

bentsherman: You should explore that in a separate PR. This change is just fixing a bug with the current behavior

modules/nextflow/src/main/groovy/nextflow/Session.groovy
bentsherman: Yes, but `obsserversV2` is reassigned in `init()`

christopher-hakkaart (commented):
Docs look good 👍

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
jorgee: I think for the new model (with lists instead of maps), the list of strings will be easier to refer in a query, now we have to specify `annotations.key` and `annotations.value`. I can do it in a separate PR once it is merged.

modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
```diff
@@ -120,18 +122,17 @@ interface TraceObserver {
...
+     * NOTE: This method is no longer used.
      *
-     * @param name
-     *      The name of the workflow output
      * @param value
-     *      A list if the published channel was a queue channel,
-     *      otherwise an object if the channel was a value channel
      */
-    default void onWorkflowPublish(String name, Object value){}
+    @Deprecated
```
jorgee, line 134: If all the class is deprecated, is it still required?

jorgee (approved):
Looks fine with me, just a minor comment.

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
pditommaso: > You should explore that in a separate PR

Agree  

> I think for the new model (with lists instead of maps), the list of strings will be easier to refer in a query

let's do

modules/nextflow/src/main/groovy/nextflow/trace/TraceObserver.groovy
pditommaso: Yes, because almost all plugins still use it. We need to migrate them and give time to third party to do the same

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
jorgee: @pditommaso I assume that you also want to change `annotations` to `labels` in the Lineage model, right? 
In processes there are two directives `label` and `resourceLabels`. The one that you want to add in TaskRun model is the first one, isn't it? Or do you refer to another property?

bentsherman: Why would we save the process labels in the lineage model? They are used only to apply configuration like cpus, memory, gpus, etc. They are mostly execution details

pditommaso: Because it's metadata and to avoid introducing yer another notation other `label` and `resourceLabels`

jorgee: There is also `tag` that exist in process and publishDir. I think it is the best one to be reused. I remember we had this discussion and we decided to keep `annotations` for M1

pditommaso: moreover  .. tho tags has always been kind of experimental

bentsherman: What if the user doesn't want the process labels to be included? They might want to use these annotations for a different purpose and forcing them to include the process labels could muddle that purpose

PR #6016, opened 2025-04-29
Support downloading plugins from OCI registries

tom-seqera:
An alternative to #5968, this allows Nextflow to download plugin artifacts from an OCI registry using an http digest url rather than an `oras://` url.

The plugin download logic currently expects the downloaded file name to already be in the `$id-$version.zip` format, but a file downloaded using a digest url will have the digest as its name instead. The download logic has been update to detect this and rename the file accordingly.

### DIFF
--- a/modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
+++ b/modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
@@ -222,9 +222,10 @@ class PluginUpdater extends UpdateManager {
         Path downloaded = safeDownloadPlugin(id, version);
 
         // 3. rename if filename is sha digest
+        // when the plugin is downloaded from the (OCI) registry, it is named as "sha256:<CHECKSUM>"
+        // rename it to something meaningful using the target plugin path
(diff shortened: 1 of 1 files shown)


modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
```diff
@@ -221,13 +221,20 @@ class PluginUpdater extends UpdateManager {
         // 2. download to temporary location
         Path downloaded = safeDownloadPlugin(id, version);
 
-        // 3. unzip the content and delete downloaded file
+        // 3. rename if filename is sha digest
```
pditommaso, line 224: I'd try to provide a bit more context on the rationale and an example of the URI to make it more directly understandable

pditommaso (commented):
Made a minor comment, otherwise it looks good

modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
pditommaso: Did it

PR #6035, opened 2025-05-05
Improve output labels

pditommaso:
Improve output labels definition so that it also allows a single string value, for example 

```
logs { path 'logs/'; labels('foo') }
```

Moreover it also add `label` shortcut that allows repeating the label definition for consistency with the process definition, ie. 

```
logs { 
   path 'logs/'
   label 'foo' 
   label 'bar
}
```

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
@@ -192,24 +192,25 @@ class PublishOp {
      * Get or resolve the labels of a workflow output.
      *
      * @param labels List | Closure<List>
-     * @param value
+     * @param context
      */
-    protected static List<String> getLabels(labels, value) {
+    protected static List<String> getLabels(Object labels, Object context) {
         if( labels == null )
             return []
+        if( labels instanceof CharSequence )
+            return [labels.toString()]
         if( labels instanceof List<String> )
             return labels
         if( labels instanceof Closure ) {
             try {
-                final result = labels.call(value)
-                if( result instanceof List<String> )
-                    return result
-            } catch (Throwable e) {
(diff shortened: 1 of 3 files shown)


modules/nextflow/src/test/groovy/nextflow/extension/PublishOpTest.groovy
```diff
@@ -0,0 +1,40 @@
...
+    @Unroll
+    def 'should get labels' () {
+        expect:
+        PublishOp.getLabels(LBL, VAL) == EXPECTED
+
+        where:
+        LBL                 | VAL           | EXPECTED
+        null                | null          | []
+        'foo'               | null          | ['foo']
+        ['a','b','c']       | null          | ['a','b','c']
```
pditommaso, line 37: @jorgee can you please add a test for labels as a closure?

jorgee:
The possibility to define a label as a closure was to have the possibility of defining labels per sample. I also added the labels in the index file because it was not possible to run the closure. If we remove the possibility to define as closure we can also remove it.

bentsherman (approved):

modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
```diff
@@ -242,7 +217,7 @@ class PublishOp {
             else {
                 log.warn "Invalid extension '${ext}' for index file '${indexPath}' -- should be CSV, JSON, or YAML"
             }
-            session.notifyFilePublish(new FilePublishEvent(null, indexPath, indexOpts.labels))
+            session.notifyFilePublish(new FilePublishEvent(null, indexPath))
```
jorgee, line 220: I think we should publish the index file with the labels specified for the output

bentsherman: Good point, now that the closure is gone it would make sense. I will add this

PR #6038, opened 2025-05-05
Improve lineage find docs

### LINKED ISSUE #6036: Improve lineage find docs
The documentation for `lineage find` should be extended: 

https://github.com/nextflow-io/nextflow/blob/375dc17ce73d053390e78ee8d2e4bd8414562234/docs/reference/cli.md?plain=1#L742

it should be described be the search keys allowed and provided a few examples e.g. 

```
nextflow lineage find labels=bar workflowRun=lid://af9d92fbb675764b2c875b21a306b66b
```

jorgee:
close #6036

- Improve the `find` command docs including reference to the model and two examples.
- Remove usage print in some errors in `CmdLineage`
- Rename `labels` to `label` in `fromLineage`, `find` and model.


docs/reference/cli.md
```diff
@@ -736,11 +736,22 @@ $ nextflow lineage view lid://c914d714877cc5c882c55a5428b510b1#output
 You can use the [jq](https://jqlang.org/) command-line tool to apply further queries and transformations on the resulting lineage record.
 :::
 
-Find all lineage records that match a set of key-value pairs:
+Find all lineage records that match a set of key-value pairs.
```
christopher-hakkaart, line 739:
```suggestion
Find all lineage records that match a set of key-value pairs:
```

docs/reference/cli.md
```diff
@@ -736,11 +736,22 @@ $ nextflow lineage view lid://c914d714877cc5c882c55a5428b510b1#output
...
 
 ```console
 $ nextflow lineage find <key-1>=<value-1> <key-2>=<value-2> ...
 ```
+You can use as key any object property defined in the [Lineage metadata model](https://github.com/nextflow-io/nextflow/tree/master/modules/nf-lineage/src/main/nextflow/lineage/model). Additionally, the `type` key is used to refer to a metadata object class. 
+For instance, you can look for outputs generated by a workflow annotated with a certain label using the following command.
```
christopher-hakkaart, line 745:
```suggestion

Use any object property defined in the [Lineage metadata model](https://github.com/nextflow-io/nextflow/tree/master/modules/nf-lineage/src/main/nextflow/lineage/model) as a key. Use the `type` key to refer to a metadata object class:
```

docs/reference/cli.md
```diff
@@ -736,11 +736,22 @@ $ nextflow lineage view lid://c914d714877cc5c882c55a5428b510b1#output
...
 ```console
 $ nextflow lineage find <key-1>=<value-1> <key-2>=<value-2> ...
 ```
+You can use as key any object property defined in the [Lineage metadata model](https://github.com/nextflow-io/nextflow/tree/master/modules/nf-lineage/src/main/nextflow/lineage/model). Additionally, the `type` key is used to refer to a metadata object class. 
+For instance, you can look for outputs generated by a workflow annotated with a certain label using the following command. 
+
+```console
+$ nextflow lineage find type=FileOutput workflowRun=lid://c914d714877cc5c882c55a5428b510b1 label=foo
+```
+We can also use the `find` command to find all the tasks executed by a workflow using the command below:
```
christopher-hakkaart, line 750:
```suggestion
Find all the tasks executed by a workflow:
```

christopher-hakkaart (changes requested):
This section is examples so I've updated the text to be in line with the rest of the examples. If you want to explain something in more detail we can add it somewhere else.

pditommaso:
How do you see to keep `labels` only in the model class(es)?

jorgee:
> How do you see to keep `labels` only in the model class(es)?

I think it is not too complicated, just requiring to add an if statement when parsing the keys. What about the `fromLineage` parameter? It is a list. Do you also want to name it ` labels` or `label`?

pditommaso:
My reasoning is that anything is "user-facing" should be "label" (and take a label string or a list of string) for consistency with `process.label` and `output.label`. Instead the model should be `labels` for consistency with the data type

jorgee:
pushed changes  keeping `labels` in the lineage model, and supporting `List` and `String` in the `fromLineage` `label` property

docs/reference/channel.md
```diff
@@ -72,15 +72,15 @@ The `channel.fromLineage` factory creates a channel that emits files from the {r
...
     .view()
 ```
 
 The above snippet emits files published by the given workflow run that are labeled as `alpha` and `beta`.
 
 Available options:
 
-`labels`
+`label`
 : List of labels associated with the desired files.
```
bentsherman, line 84: Shouldn't this be `labels` if it's a list? I think `label` only makes sense on the command line since you have to specify it once for each label

Issue #6039, opened 2025-05-06
Not logging nextflow.config.ConfigBuilder trace when required


kdesnos:
## Bug report 

When starting nextflow with option `-trace nextflow.config.ConfigBuilder`, the corresponding traces are not printed in the log file as expected. It was thus possible to activate the trace successfully by activating traces for all clases with `-trace nextflow`

## Cause
After investigation, the problem appears because log.trace is only called in the static `resolveConfig()` method within the ConfigBuilder class. This method is only called by the `CmdRun.run()` method, guarded by a call to `log.isTraceEnabled()`. So, in order to activate this trace logging, both the current (nonobvious) configuration is currently needed: `-trace nextflow.config.ConfigBuilder,nextflow.cli.CmdRun`.

A PR for a fix is coming soon.

PR #6040, opened 2025-05-06
Fix config for Trace level logging within the ConfigBuilder class 

kdesnos:
Fix issue #6039

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
@@ -30,6 +30,7 @@ import groovy.json.JsonSlurper
 import groovy.transform.CompileStatic
 import groovy.transform.Memoized
 import groovy.util.logging.Slf4j
+import org.slf4j.LoggerFactory
 import groovyx.gpars.GParsConfig
 import nextflow.BuildInfo
 import nextflow.NF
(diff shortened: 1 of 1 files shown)


kdesnos:
The unit test error seems to occur in some code I didn't touch at all..

kdesnos:
Note that an alternative solution would simply be to remove the call to `isTraceEnable` in CmdRun and let ConfigBuilder decide on its own.

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -355,7 +356,8 @@ class CmdRun extends CmdBase implements HubOptions {
 
         final isTowerEnabled = config.navigate('tower.enabled') as Boolean
         final isDataEnabled = config.navigate("lineage.enabled") as Boolean
-        if( isTowerEnabled || isDataEnabled || log.isTraceEnabled() )
+
+        if( isTowerEnabled || isDataEnabled || LoggerFactory.getLogger(nextflow.config.ConfigBuilder).isTraceEnabled() )
```
pditommaso, line 360: I think we can just remove these conditions and resolve the config always

bentsherman: I never understood why we resolve it here anyway, when it is already resolved here

https://github.com/nextflow-io/nextflow/blob/2c9dd9b18a179b28a46ed998c165acc47107f74d/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy#L329-L333

pditommaso: Because this it's done to render it as a string using 

https://github.com/nextflow-io/nextflow/blob/e4af2308d5c1a4fd16438005565fefcf07891af7/modules/nextflow/src/main/groovy/nextflow/config/ConfigBuilder.groovy#L908-L909

maybe it can be done reusing the config object already build, but never looked into that

PR #6041, opened 2025-05-06
Data Lineage workflow outputs fixes

jorgee:
This PR includes some fixes about how the data lineage module works when there aren't outputs.

- Do not write the `#outputs` entry if there aren't workflow outputs
- lineage `view` command returns an empty list if there aren't workflow outputs

Other minor fixes:
 - Fixes the return of fragment including empty lists
 - Rename `LinCommand` methods `log` and `describe` to `list` and `view` and subclasses `CmdLog `and `CmdDescribe` to CmdList and CmdView

### DIFF
--- a/modules/nf-lineage/src/main/nextflow/lineage/LinObserver.groovy
+++ b/modules/nf-lineage/src/main/nextflow/lineage/LinObserver.groovy
@@ -123,7 +123,7 @@ class LinObserver implements TraceObserverV2 {
 
     @Override
     void onFlowComplete(){
-        if( this.workflowOutput && this.workflowOutput.output ){
+        if(workflowOutput?.output ){
             workflowOutput.createdAt = OffsetDateTime.now()
             final key = executionHash + '#output'
             this.store.save(key, workflowOutput)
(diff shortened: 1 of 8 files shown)


modules/nf-lineage/src/main/nextflow/lineage/LinObserver.groovy
```diff
@@ -123,7 +123,7 @@ class LinObserver implements TraceObserverV2 {
 
     @Override
     void onFlowComplete(){
-        if (this.workflowOutput){
+        if( this.workflowOutput && this.workflowOutput.output ){
```
pditommaso, line 126:
```suggestion
        if(workflowOutput?.output ){
```

nit ☝️

pditommaso (commented):
Just a couple of minor comments

jorgee:
Including also a change to improve the warning message when outputs come from a not supported publishDir.
- New `OutputRelativePathException` thrown when a file is published from an unsupported publish dir.
- Printing the warning message `WARN: Lineage for workflow output is not supported by publishDir directive` instead of the error

pditommaso (approved):

PR #6042, opened 2025-05-06
Make config parser v2 compatible with AssetManager

### LINKED ISSUE #6027: Make config parser v2 compatible with AssetManager
Spun off from https://github.com/nextflow-io/nextflow/pull/5998#discussion_r2070964571

The AssetManager reads the config in order to grab the manifest defaultBranch, and it does so in a way that is incompatible with the strict config parser.

Either the config parser should be updated to support the AssetManager, or `manifest.defaultBranch` should be deprecated in favor of using the git default branch (#5375).

bentsherman:
Close #6027 

When the asset manaager parses a config file as raw text, config includes are automatically disabled, and compilation errors are not printed, since they don't have filename information.

I also changed the asset manager's exception to a warning, so that the runtime proceeds to parse the config file like normal and prints the actual error messages. Otherwise, you would never see config errors because it would always be short-circuited by the asset manager.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/config/parser/v2/ConfigParserV2.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/config/parser/v2/ConfigParserV2.groovy
@@ -130,18 +130,23 @@ class ConfigParserV2 implements ConfigParser {
             return Bolts.toConfigObject(target)
         }
         catch( CompilationFailedException e ) {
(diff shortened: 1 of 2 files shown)


pditommaso (commented):
This looks good to me, in any case I would prefer to merge post 25.04 considering this is not urgent and battle test it in a edge release

PR #6044, opened 2025-05-07
Improve data lineage docs

ewels:


docs/reference/cli.md
```diff
@@ -726,14 +730,19 @@ View a lineage record.
...
 
 ```console
 $ nextflow lineage view lid://c914d714877cc5c882c55a5428b510b1#output
 ```
 
-:::{tip}
-You can use the [jq](https://jqlang.org/) command-line tool to apply further queries and transformations on the resulting lineage record.
+:::{warning}
+Only outputs described using the {ref}`workflow-outputs` syntax are captured in lineage records.
+If your pipeline only uses {ref}`publishDir` then an empty list will be returned.
```
bentsherman, line 745: @jorgee this would be a good place to explain how files published via `publishDir` can still be accessed as LIDs

jorgee: Great. I am going to add it.

jorgee, line 745:
```suggestion
If your pipeline only uses {ref}`publishDir` then an empty list will be returned.
However, files published via `publishDir` can be accessed using the Lineage ID in the format `lid://<workflow_lid>/<relative_path>`, where `<relative_path>` refers to the published file's path relative to the workflow's output directory. If the `publishDir` path is not a subpath of the output directory, the lineage metadata entry for these outputs will not be created, and a warning message will be printed. 
```

docs/reference/cli.md
```diff
@@ -705,6 +705,10 @@ $ nextflow lineage SUBCOMMAND [arg ..]
 
 The `lineage` command is used to inspect lineage metadata. Data lineage can be enabled by setting `lineage.enabled` to `true` in your Nextflow configuration (see the {ref}`config-lineage` config scope for details).
 
+:::{tip}
+Set `lineage.enabled` in your `~/.nextflow/config` file to enable it globally.
```
pditommaso, line 709: This is not useful unless providing a common lineage path

ewels: The way I struggled with this was that I added `lineage.enabled` to a pipeline config, and then ran that pipeline in a different location. It correctly used lineage and generated a `.lineage` directory, but when I tried to do anything with it I couldn't, because the `nextflow lineage` commands couldn't find a config to enable them.

@jorgee mentioned maybe making these commands specifically default to having lineage enabled, which would solve it.

I do wonder if it's still useful though. Globally enabling it means that lineage is always created, wherever you run. Otherwise you need to manually enable for each run, and when you want to use the lineage (eg. to debug failed resume) you can't. So I find this useful even without a common lineage path.

docs/reference/cli.md
ewels: Great - I wonder if it would make sense to mention that the base directory can be set with a config one-liner using `params.outdir`? This is also maybe getting a bit too long for an admonition.

I'll leave this for @christopher-hakkaart to make a decision on / clean up, if that's ok.

docs/reference/cli.md
```diff
@@ -726,14 +730,19 @@ View a lineage record.
 $ nextflow lineage view <lid>
 ```
 
+:::{tip}
+You can use the [jq](https://jqlang.org/) command-line tool to apply further queries and transformations on the resulting lineage record.
```
christopher-hakkaart, line 734:
```suggestion
You can use the [jq](https://jqlang.org/) command-line tool to further query and transform lineage records.
```

docs/reference/cli.md
christopher-hakkaart: I've edited this comment for language, please check the meaning is retained.

pditommaso:
I think we should mark this feature as "Preview" in this release.

jorgee:
> I think we should mark this feature as "Preview" in this release.

Do you mean adding a warning message when lineage is activated? or adding `workflow.preview` option? I think the second is not needed, right?

pditommaso:
Just in the docs

bentsherman:
The `lineage` command is already marked as experimental in the docs

bentsherman:
Not sure what you mean, there is no main lineage section

bentsherman:
Rather, the `lineage` command is the main section

bentsherman:
From our discussion today, I will add a guide for using data lineage so that we have something more comprehensive without bloating the CLI docs

docs/reference/cli.md
bentsherman: Moving this tip to the guide

ewels:
Are we happy to merge this PR and do the guide in a different one @bentsherman ?

bentsherman (approved):

bentsherman:
Ready to go. The guide covers all the basic usages, as well as the tips about global config and setting the output directory. It does not cover the use of labels because I don't have a real-world example on hand for that. We can work on that in a separate PR.

I was going to merge, but I'm being blocked by this failing google test. Not sure why it's failing on only some PRs.

PR #6047, opened 2025-05-07
Solve queueSize exceeded when using job arrays

### LINKED ISSUE #5920: Job Array with queueSize exceeds queueSize

For example: 
```
process.array = 30
executor.queueSize = 40
``` 

jorgee:
close #5920

- Modifying canSubmit to avoid to avoid exceed `executor.queueSize` parameter. 
- Printing warning when array size exceeds the queueSize. This can make the task array 'unsubmitable' (Not sure if we should abort the execution in this situation)
- Add unit tests

Tested with this pipeline

```
process test {
    array 10 
    input:
    val x

    """
    echo $x
    """
}

workflow {
   Channel.of(1..20) | test
}
```
With a config with  `awsbatch` executor and different `executor.queueSize` values:

- size 15: Only one job submitted at a time in AWS Batch.
- size 5: Prints a warning but run continues without the task array jobs submitted.


bentsherman:
- change the warning to an error, since the run will hang anyway
- make process maxForks aware of job arrays

bentsherman (approved):

modules/nextflow/src/main/groovy/nextflow/processor/TaskHandler.groovy
```diff
@@ -266,7 +273,7 @@ abstract class TaskHandler {
      */
     boolean canForkProcess() {
         final max = task.processor.maxForks
-        return !max ? true : task.processor.forksCount < max
+        return !max ? true : task.processor.forksCount + getForksCount() <= max
```
pditommaso, line 276: Should not be `task.processor.forksCount * getForksCount()`, I mean multiplied?

bentsherman: why would it be multiplied?

pditommaso (changes requested):

pditommaso:
Because `processor.forksCount` is related to `maxForks` directive, instead here the aim is to controller the max numbers of task runs

bentsherman:
So "forks" here just refers to the number of concurrent tasks, therefore I see no reason to call it something else in the task handler

PR #6059, opened 2025-05-09
Refactor deprecations as best practices

bentsherman:
Follow-up to #6055 . The following deprecation warnings have been demoted to "best practice recommendations:

- Using `params` outside the entry workflow
- Implicit closure `it` parameter
- Process `when` section

The language server will highlight these patterns as blue instead of yellow to indicate the decreased urgency.

Some of these patterns will become deprecated as part of static types, but there is no need to warn about them right now. Just trying to keep the set of strict errors/warnings as focused as possible.

### DIFF
--- a/docs/process.md
+++ b/docs/process.md
@@ -24,13 +24,13 @@ See {ref}`syntax-process` for a full description of the process syntax.
 
 ## Script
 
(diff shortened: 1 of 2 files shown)


christopher-hakkaart (approved):
Looks great. No comments.

PR #6060, opened 2025-05-09
Document required `NXF_EDGE=0` when updating from edge to stable

bentsherman:
Discovered this issue while trying to self-update from 25.03 edge to 25.04


christopher-hakkaart (approved):
Nice addition

PR #6064, opened 2025-05-09
Remove one of the line chomps from the lint command

ewels:
Noticed that `nextflow lint` is a little too aggressive and can start eating back on the terminal from before its own launch command.. 😬 

Before this PR:

CleanShot 2025-05-10 at 00 26 45

After this PR:

CleanShot 2025-05-10 at 00 27 39


modules/nextflow/src/main/groovy/nextflow/script/parser/v2/ErrorListener.groovy
```diff
@@ -90,7 +90,7 @@ class StandardErrorListener implements ErrorListener {
 
     @Override
     void onError(SyntaxException error, String filename, SourceUnit source) {
-        term.bold().a(filename).reset()
+        term.bold().a("error: ").a(filename).reset()
```
bentsherman, line 93: Let me know if you like how this looks and I'll merge. I wonder if it would be better to put the error/warning label after the location

ewels: Haven't checked, but yeah I was thinking after the file name - like a prefix to the error message:

```
Error: Something bad happened.
```

Maybe with some colour too? Need to look at it for that though.

bentsherman: Updated. Let me know how it looks

ewels:
I think that it'd be improved with:

* Prefixing the entire line - that way they all align and it's easy to scan up and down to see which are errors and which are warnings
* More colour to better differentiate them, also helps to see the borders between each message.

I've pushed updates to that effect - I've somewhat controversially gone for a background colour which _really_ makes it stand out:

CleanShot 2025-05-12 at 14 28 02@2x

In concise mode:

CleanShot 2025-05-12 at 14 30 59@2x

Is that too much? I can revert it to the same text but just regular Red / Yellow font colour instead of background colour if you prefer.

ewels:
Done 👍🏻  Had to pad `Warn ` with spaces to get alignment but I think it's ok:

CleanShot 2025-05-12 at 14 56 03@2x

pditommaso:
Cool, tempted to make .2 to show this :D

ewels:
Wondering if we should have a hanging block-colour to connect this to the context below 🤔 2 secs - will see if I can whip it up and send a screenshot

ewels:
Example with `-o extended` (where it's more obvious) - do we like the side bar?

CleanShot 2025-05-12 at 15 18 49@2x

Regular output mode:

CleanShot 2025-05-12 at 15 19 38@2x

pditommaso:
think the last is the nicest

ewels:
> think the last is the nicest

Agreed 👍🏻  That's what I pushed, so ready for merge if you're happy.

pditommaso:
Leave Ben to close last comment and merge

bentsherman (changes requested):
I don't really like the use of background colors, I would prefer using the foreground instead

ewels:
Done! Screenshots of various output types:

CleanShot 2025-05-12 at 17 57 53@2x

CleanShot 2025-05-12 at 17 58 25@2x

CleanShot 2025-05-12 at 17 58 47@2x

ewels:
@bentsherman should I copy this style over to the run time errors too?

bentsherman (approved):

pditommaso:
Think the best was this version https://github.com/nextflow-io/nextflow/pull/6064#issuecomment-2872459051

bentsherman:
I just don't like the use of background colors for this

pditommaso:
The outline style works better with multiline blocks, but on simple cases seems a bit broken 

Screenshot 2025-05-13 at 18 48 21

PR #6070, opened 2025-05-13
Sunset the Google LS executor

pditommaso:
This PR sunsets the Google Cloud LifeScience execution since the service is going to not eb available anymore as of July 2025 

https://cloud.google.com/life-sciences/docs/deprecations


docs/reference/config.md
```diff
@@ -956,52 +956,6 @@ The following settings are available for Cloud Life Sciences:
 `google.zone`
 : The Google Cloud zone where jobs are executed. Multiple zones can be provided as a comma-separated list. Cannot be used with the `google.region` option. See the [Google Cloud documentation](https://cloud.google.com/compute/docs/regions-zones/) for a list of available regions and zones.
 
-`google.lifeSciences.bootDiskSize`
-: Set the size of the virtual machine boot disk e.g `50.GB` (default: none).
```
bentsherman, line 960: The entire "Cloud Life Sciences" section here can be removed, and the "Cloud Batch" heading can be removed. Any config settings that are used by both life sciences and batch should be duplicated in the docs, so it should be safe to remove the life sciences copy

I would check the `google.storage` options because I think they are only used by life sciences, they can probably be removed

pditommaso: possibly, i'll mark as draft

pditommaso: Ok, a bunch of storage settings have been removed, still remains the ones about defining the client retry policy

PR #6074, opened 2025-05-13
Add check subcommand in lineage

jorgee:
This PR adds the check subcommand in the lineage CLI command.This command is used to validate the integrity of an output path.


pditommaso:
Nice Jorge! I've refactored a bit

PR #6082, opened 2025-05-15
Fix bug in generated Groovy code

bentsherman:
This PR fixes a bug when running a script with the v2 parser:

```groovy
workflow INIT {
    take:
    validate_params

    main:
    def ch_samples = Channel.empty()
    // do stuff

    emit:
    samples = ch_samples
}

workflow {
    INIT(true)
}
```

```console
$ NXF_SYNTAX_PARSER=v2 nextflow run main.nf

 N E X T F L O W   ~  version 25.04.2

Missing workflow output parameter: samples
```


pditommaso (changes requested):
A test capturing this condition should be added

PR #6083, opened 2025-05-15
Add Failsafe retry mechanism in K8s

### LINKED ISSUE #5829: K8s client does not retry some 5xx errors
> If it's returning a 500 error code then Nextflow should be retrying it:
> 
> [nextflow/modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy](https://github.com/nextflow-io/nextflow/blob/0aa76affe22c35fc5bc6a4b609fa9b78528859a9/modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy#L629-L639)
> 
> Lines 629 to 639 in [0aa76af](/nextflow-io/nextflow/commit/0aa76affe22c35fc5bc6a4b609fa9b78528859a9)
>  try { 
>      return makeRequestCall( method, path, body ) 
>  } catch ( K8sResponseException | SocketException | SocketTimeoutException e ) { 
>      if ( e instanceof K8sResponseException && e.response.code != 500 ) 
>          throw e 
>      if ( ++attempt > maxRetries ) 
>          throw e 
>      log.debug "[K8s] API request threw socket exception: $e.message for $method $path - Retrying request (attempt=$attempt)" 
>      final long delay = (Math.pow(3, attempt - 1) as long) * 250 
>      sleep( delay ) 
>  } 
> 
> I wonder if it is retrying but the tunnel disconnect lasts longer than the duration of the 8 retries. Seems plausible if it is a regular outage. Can you see the "API request threw socket exception..." messages in your log?
> 
> @bentsherman, apologies if this should be raised a separate question/issue but this issue and conversation seem to be closely related to question/issue I am having. 
> 
> I am running in a cluster that sometimes under heavy load does return temporary 503 errors and this causes nextflow to terminate and this can be seen in the logs:
> 
> ```
> `Request GET /api...returned an error code=503...`
> ```
> 
> I seems that similar codes have been switched to retryable for AWS Batch #4709.
> 
> I am using nextflow `24.04.3` and when the 503 occurs and the workflow exits the running processes are not terminated.  Testing with `24.12.0-edge` that has fix #5561 resolves that and terminates running processes even if 503 occurs.
> 
> Is there a reason why 503 is not also added as a retryable exit code as for AWS Batch? 
> 
> Is this something that could potentially be added or a `k8s` setting that would allow the user to specify which errors to treat as retryable `k8s.retryable_errors = [500, 502, 503 ]`?
> 
> 
>  

 _Originally posted by @crossthet in [#5604](https://github.com/nextflow-io/nextflow/issues/5604#issuecomment-2690632473)_

jorgee:
close #5829 

This PR replace the failure mechanism by `FailSafe` retry mechanism used in other parts in the code base.
Replace the current config `k8s.maxErrorRetry` with `k8s.retryPolicy`

### DIFF
--- a/plugins/nf-k8s/src/main/nextflow/k8s/client/ClientConfig.groovy
+++ b/plugins/nf-k8s/src/main/nextflow/k8s/client/ClientConfig.groovy
@@ -135,7 +135,7 @@ class ClientConfig {
         if( opts.retryPolicy )
             result.retryConfig = new K8sRetryConfig(opts.retryPolicy as Map)
         if( opts.maxErrorRetry )
-            log.warn("'k8s.maxErrorRetry' is deprecated. Change it to 'k8s.retryPolicy.maxAttempts'")
+            log.warn("'k8s.maxErrorRetry' is deprecated - change it to 'k8s.retryPolicy.maxAttempts'")
 
         return result
     }
--- a/plugins/nf-k8s/src/main/nextflow/k8s/client/ClientConfig.groovy
+++ b/plugins/nf-k8s/src/main/nextflow/k8s/client/ClientConfig.groovy
@@ -135,7 +135,7 @@ class ClientConfig {
         if( opts.retryPolicy )
             result.retryConfig = new K8sRetryConfig(opts.retryPolicy as Map)
         if( opts.maxErrorRetry )
-            log.warn("'k8s.maxErrorRetry' is deprecated - change it to 'k8s.retryPolicy.maxAttempts'")
+            log.warn("Config setting 'k8s.maxErrorRetry' is deprecated - change it to 'k8s.retryPolicy.maxAttempts'")
 
         return result
     }
--- a/plugins/nf-k8s/src/main/nextflow/k8s/client/ClientConfig.groovy
+++ b/plugins/nf-k8s/src/main/nextflow/k8s/client/ClientConfig.groovy
@@ -16,6 +16,7 @@
 
 package nextflow.k8s.client
 
+import groovy.util.logging.Slf4j
 import nextflow.util.Duration
 
 import javax.net.ssl.KeyManager
@@ -31,6 +32,7 @@ import groovy.transform.EqualsAndHashCode
  */
 @EqualsAndHashCode
 @CompileStatic
+@Slf4j
 class ClientConfig {
 
     boolean verifySsl
@@ -55,7 +57,7 @@ class ClientConfig {
 
     KeyManager[] keyManagers
 
-    Integer maxErrorRetry = 4
+    K8sRetryConfig retryConfig
 
     /**
      * Timeout when reading from Input stream when a connection is established to a resource.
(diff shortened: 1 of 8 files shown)


plugins/nf-k8s/src/main/nextflow/k8s/client/K8sClient.groovy
```diff
@@ -735,6 +730,61 @@ class K8sClient {
...
+        final cond = new Predicate<? extends Throwable>() {
+            @Override
+            boolean test(Throwable t) {
+                if ( t instanceof K8sResponseException && t.response.code in RETRY_CODES )
+                    return true
+                if( t instanceof IOException || t.cause instanceof IOException )
+                    return true
+                if( t instanceof TimeoutException || t.cause instanceof TimeoutException )
+                    return true
+                return false
```
pditommaso, line 775: I may be wrong, but looks like the condition is not the same 

https://github.com/nextflow-io/nextflow/pull/6083/files#diff-c10aa1211d34577bb39d669cb3a132d9d3d3595ab8631e95ab0199d74d1ae73dL631-L632

plugins/nf-k8s/src/main/nextflow/k8s/client/ClientConfig.groovy
```diff
@@ -130,9 +132,10 @@ class ClientConfig {
             result.clientKey = opts.clientKey.toString().decodeBase64()
         else if( opts.clientKeyFile )
             result.clientKey = Paths.get(opts.clientKeyFile.toString()).bytes
-
+        if( opts.retryPolicy )
+            result.retryConfig = new K8sRetryConfig(opts.retryPolicy as Map)
         if( opts.maxErrorRetry )
-            result.maxErrorRetry = opts.maxErrorRetry as Integer
+            log.warn("'k8s.maxErrorRetry' is deprecated. Change it to 'k8s.retryPolicy.maxAttempts'")
```
pditommaso, line 138:
```suggestion
            log.warn("'k8s.maxErrorRetry' is deprecated - change it to 'k8s.retryPolicy.maxAttempts'")
```

nit, usually messages use `-` as separator

christopher-hakkaart:
Documentation looks good 👍

plugins/nf-k8s/src/main/nextflow/k8s/client/K8sClient.groovy
```diff
@@ -735,6 +730,61 @@ class K8sClient {
...
+     *
+     * @param cond A predicate that determines when a retry should be triggered
+     * @return The {@link dev.failsafe.RetryPolicy} instance
+     */
+    protected <T> RetryPolicy<T> retryPolicy(Predicate<? extends Throwable> cond) {
+        final cfg = config.retryConfig
+        final listener = new EventListener<ExecutionAttemptedEvent<T>>() {
+            @Override
+            void accept(ExecutionAttemptedEvent<T> event) throws Throwable {
+                log.debug("Azure TooManyRequests response error - attempt: ${event.attemptCount}; reason: ${event.lastFailure.message}")
```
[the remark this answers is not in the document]
jorgee: Ups, I forgot to change this message

plugins/nf-k8s/src/main/nextflow/k8s/client/K8sClient.groovy
jorgee: Yes, the SocketException and SocketTimeoutException are subclasses of IOException, so the retries for these cases are included in this case. I have also included the cases retried in the cloud executors that are using the same mechanism (Google and Azure).

pditommaso: think the retry condition should not be altered

jorgee: Ok, I will put the same exceptions as before.

PR #6093, opened 2025-05-16
POC: Support pipeline archives instead of git repos or local paths

adamrtalbot:
This adds support for using archives of pipelines instead of git repos for remote use. This is useful because it bypasses remote git which can be problematic in secure environments or when hitting rate limits. To use, you can run `nextflow run https://url.com/path/to/archive.tar.gz` which will download the files to artefacts and run them. We could imagine something like Seqera Platform serving secure URLs to Git archives to allow users to use a repo without granting them access to the repo itself. The main downside of this is it might encourage bad practice, where users will use this in preference to a Git repo where they should do the majority of their work.

Signed-off-by: adamrtalbot <12817534+adamrtalbot@users.noreply.github.com>

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
@@ -564,6 +564,61 @@ class CmdRun extends CmdBase implements HubOptions {
         }
 
         /*
+         * Handle archive URLs directly
+         */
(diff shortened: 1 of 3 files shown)


bentsherman:
Between this and the plugin registry, we're just getting closer and closer to a pipeline registry...

bentsherman:
Closing, we can revisit this if/when we implement a pipeline registry

PR #6105, opened 2025-05-20
Force overwritting to trace file

### LINKED ISSUE #6090: Error: Failed to write to file: .command.trace
## Bug report 

When running the nf-core/proteinfamilies pipeline as a use-case, I was trying to see the Metrics resources on Seqera Platform (running -with-tower), but all barplots except from the Job Duration, were empty. Digging a little deeper, I figured that on every working folder there was this line in the .command.err and .command.log files: `.command.run: line 182: .command.trace: cannot overwrite existing file
Error: Failed to write to file: .command.trace`

The line above shouldn't be there, and the barplots should be populated accordingly on Seqera Platform.

nextflow run proteinfamilies -profile singularity,test --outdir results

Then go to any working folder and open the .command.err or .command.log files. The last line should read: `.command.run: line 182: .command.trace: cannot overwrite existing file
Error: Failed to write to file: .command.trace`

The .command.trace will be there however.

### Program output 

Nothing out of the ordinary here.

### Environment 

* Nextflow version: this behaviour starts at Version 25.03.0-edge, onwards, including the latest 25.04.2
* Java version: 
openjdk version "17.0.6" 2023-01-17 LTS
OpenJDK Runtime Environment Corretto-17.0.6.10.1 (build 17.0.6+10-LTS)
OpenJDK 64-Bit Server VM Corretto-17.0.6.10.1 (build 17.0.6+10-LTS, mixed mode, sharing)
* Operating system: Linux

### Additional context

Tried with both 3.2.0 and 3.2.1 nf-core tools versions.
This behaviour doesn't happen for another local non-nf-core pipeline I run in the same system.

jorgee:
close #6090 

This PR forces the overwrite of .command.trace to avoid the error writting the trace file when `bash -C ` (noclobber) is set in `process.shell`

### DIFF
--- a/modules/nextflow/src/main/resources/nextflow/executor/command-trace.txt
+++ b/modules/nextflow/src/main/resources/nextflow/executor/command-trace.txt
@@ -135,7 +135,7 @@ nxf_write_trace() {
         "syscr=${io_stat1[2]}" \
         "syscw=${io_stat1[3]}" \
         "read_bytes=${io_stat1[4]}" \
-        "write_bytes=${io_stat1[5]}" > "$trace_file" || >&2 echo "Error: Failed to write to file: $trace_file"
+        "write_bytes=${io_stat1[5]}" >| "$trace_file" || >&2 echo "Error: Failed to write to file: $trace_file"
 }
 
 nxf_trace_mac() {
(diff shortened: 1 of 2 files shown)


pditommaso (commented):
Make sense 👍

Issue #6112, opened 2025-05-22
Incorrect hash when staging repository directory


DriesSchaumont:
## Bug report 

@rcannood and I have been looking at an issue for some time now where runs are not being resumed. 
We've been able to [reproduce the problem](https://github.com/DriesSchaumont/nextflow-file-caching-example) and found that it has to do with `hashDirSha256`  in `HashBuilder.java`. It is being called for directories which are part of the repository, after checking `isAssetFile`: 
https://github.com/nextflow-io/nextflow/blob/c384b9d0c3edd26c743fabefd747485c17aab42b/modules/nf-commons/src/main/nextflow/util/HashBuilder.java#L278

In the `hashDirSha256` method, `Files.walkFileTree` is being used for calculating the hash of a directory by iterating over all files. 

https://github.com/nextflow-io/nextflow/blob/c384b9d0c3edd26c743fabefd747485c17aab42b/modules/nf-commons/src/main/nextflow/util/HashBuilder.java#L327

However, a problem arises because the order in which the items in the directory are visited is [not guaranteed]( https://docs.oracle.com/javase/tutorial/essential/io/walk.html#order). In case order of the files is different, the `hasher` calculates a different hash even when the contents of the directory is the same.

This bug is particularly hard to track down because depending on the kernel/filesystem combination, the order or iteration for `walkFileTree`  _may_ be the same. However, we've been able to use `zfs` to reliably reproduce the problem. 

Please find a patch that is provided here: https://github.com/nextflow-io/nextflow/compare/master...rcannood:nextflow:fix-hashing. We've applied it to the reproducible example and found that it resolves the issue.
### Expected behavior and actual behavior

When an asset (repository) directory is being used as input for a process, the hash is not consistent across runs.

### Steps to reproduce the problem

Please find a reproducable example [here](https://github.com/DriesSchaumont/nextflow-file-caching-example)

### Program output 

See also [here](https://github.com/DriesSchaumont/nextflow-file-caching-example)

### Environment 

* Nextflow version:  25.04.2 build 5947
* Groovy 4.0.26 on OpenJDK 64-Bit Server VM 17.0.15+6-Debian-1deb12u1
* Operating system: Linux 6.8.12-10-pve SMP PREEMPT_DYNAMIC PMX 6.8.12-10 (2025-04-18T07:39Z) x86_64 GNU/Linux
* Bash version: GNU bash, version 5.2.15(1)-release

### Additional context

(Add any other context about the problem here)

[...]

PR #6113, opened 2025-05-22
Fix hashing of asset directories

rcannood:
`Files.walkFileTree` does not guarantee any specific order for iterating over files within a directory:

> Considerations When Creating a FileVisitor
>
> A file tree is walked depth first, but you cannot make any assumptions about the iteration order that subdirectories are visited.

-- [source](https://docs.oracle.com/javase/tutorial/essential/io/walk.html)

This causes the sha of `HashBuilder.hashDirSha256` to be different in certain situations (#6112).

This PR fixes the issue by storing all of the (path, sha256) in an array list, sorting by path, and then passing them to the hasher.

Notes:

* The code in `preVisitDirectory` seems like it would cause a null pointer exception if the condition (`base != null`) is false -- so maybe `base` is always not null? What should happen in this code chuck if base does happen to be null?
* It's hard to add a unit test for this, since the issue only arises in very specific situations.

### DIFF
--- a/modules/nf-commons/src/main/nextflow/util/HashBuilder.java
+++ b/modules/nf-commons/src/main/nextflow/util/HashBuilder.java
@@ -353,21 +353,20 @@ public class HashBuilder {
                 @Override
                 public FileVisitResult preVisitDirectory(Path path, BasicFileAttributes attrs) {
                     log.trace("Hash sha-256 dir content [DIR] path={} - base={}", path, base);
-                    // the file relative base
+
+                    // The file path relative to the base directory
                     String entry = base != null ? base.relativize(path).toString() : path.toString();
-                    // ↓ wouldn't this cause a null pointer exception when base is null?
-                    String value = base.relativize(path).toString();
                     
-                    pathShas.put(entry, value);
+                    // Store the directory path
+                    pathShas.put(entry, null);
+
                     return FileVisitResult.CONTINUE;
                 }
             });
 
             // Hash the directory content. The map should be sorted by the natural ordering of the keys.
             for (Map.Entry<String, String> pair : pathShas.entrySet()) {
-                if (pair.getKey() != null) {
-                    hasher.putUnencodedChars(pair.getKey());
-                }
+                hasher.putUnencodedChars(pair.getKey());
                 if (pair.getValue() != null) {
                     hasher.putUnencodedChars(pair.getValue());
                 }
--- a/modules/nf-commons/src/main/nextflow/util/HashBuilder.java
+++ b/modules/nf-commons/src/main/nextflow/util/HashBuilder.java
@@ -352,7 +352,9 @@ public class HashBuilder {
                     // the file relative base
                     if( base!=null )
                         hasher.putUnencodedChars(base.relativize(path).toString());
+                    // ↓ wouldn't this cause a null pointer exception when base is null?
                     hasher.putUnencodedChars(base.relativize(path).toString());
+                    // it looks like this code block might not get triggered?
                     return FileVisitResult.CONTINUE;
                 }
             });
--- a/modules/nf-commons/src/main/nextflow/util/HashBuilder.java
+++ b/modules/nf-commons/src/main/nextflow/util/HashBuilder.java
@@ -25,6 +25,8 @@ import java.nio.file.ProviderMismatchException;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
@@ -324,16 +326,20 @@ public class HashBuilder {
 
     static protected Hasher hashDirSha256( Hasher hasher, Path dir, Path base ) {
         try {
+            // temporarily store the information for later hashing
+            ArrayList<String> strs = new ArrayList<>();
+
+            // walk the directory and hash all the files
+            // NOTE: the order of the files not be guaranteed on some file systems
             Files.walkFileTree(dir, new SimpleFileVisitor<Path>() {
                 public FileVisitResult visitFile(Path path, BasicFileAttributes attrs) throws IOException {
                     log.trace("Hash sha-256 dir content [FILE] path={} - base={}", path, base);
                     try {
                         // the file relative base
-                        if( base!=null )
-                            hasher.putUnencodedChars(base.relativize(path).toString());
+                        String prefix = base != null ? base.relativize(path).toString() + " " : "";
                         // the file content sha-256 checksum
                         String sha256 = sha256Cache.get(path);
-                        hasher.putUnencodedChars(sha256);
+                        strs.add(prefix + sha256);
                         return FileVisitResult.CONTINUE;
                     }
                     catch (ExecutionException t) {
@@ -350,6 +356,14 @@ public class HashBuilder {
                     return FileVisitResult.CONTINUE;
                 }
             });
+
+            // sort the file hashes
+            Collections.sort(strs);
+
+            // hash the directory content
+            for (String str : strs) {
+                hasher.putUnencodedChars(str);
+            }
         }
         catch (IOException t) {
             Throwable err = t.getCause()!=null ? t.getCause() : t;
--- a/modules/nf-commons/src/main/nextflow/util/HashBuilder.java
+++ b/modules/nf-commons/src/main/nextflow/util/HashBuilder.java
@@ -27,9 +27,9 @@ import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
-import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.Map;
+import java.util.HashMap;
+import java.util.TreeMap;
 import java.util.Set;
 import java.util.UUID;
 import java.util.Comparator;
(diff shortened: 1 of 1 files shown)


rcannood:
Hi everybody! Can I get some feedback from the Seqera team on this issue + PR? :bow:

modules/nf-commons/src/main/nextflow/util/HashBuilder.java
```diff
@@ -344,12 +353,25 @@ public FileVisitResult visitFile(Path path, BasicFileAttributes attrs) throws IO
...
+                    // ↓ wouldn't this cause a null pointer exception when base is null?
+                    String value = base.relativize(path).toString();
+                    
+                    pathShaPairs.add(Map.entry(entry, value));
                     return FileVisitResult.CONTINUE;
                 }
             });
+
+            // sort the file hashes
+            Collections.sort(pathShaPairs, Comparator.comparing(Map.Entry::getKey));
```
pditommaso: It would be preferable using `TreeMap` to avoid sorting the entries separately

pditommaso (changes requested):
Thanks for reporting this, very tricky. Let a comment to improve the implementation

modules/nf-commons/src/main/nextflow/util/HashBuilder.java
rcannood: Good point. I swapped the ArrayList for a TreeMap!

modules/nf-commons/src/main/nextflow/util/HashBuilder.java
```diff
@@ -324,36 +328,54 @@ static protected Hasher hashFileSha256( Hasher hasher, Path path, Path base ) {
...
+                @Override
                 public FileVisitResult preVisitDirectory(Path path, BasicFileAttributes attrs) {
                     log.trace("Hash sha-256 dir content [DIR] path={} - base={}", path, base);
                     // the file relative base
-                    if( base!=null )
-                        hasher.putUnencodedChars(base.relativize(path).toString());
-                    hasher.putUnencodedChars(base.relativize(path).toString());
+                    String entry = base != null ? base.relativize(path).toString() : path.toString();
+                    // ↓ wouldn't this cause a null pointer exception when base is null?
+                    String value = base.relativize(path).toString();
```
rcannood: What about the implementation of `preVisitDirectory`? Should we change it such that no NullPointerException gets thrown when `base` is `null`?

To be honest, in this case the value can probably just be set to `null` (so it doesn't get passed to the hasher) or equal to `entry` (so it's the same value as before).

```suggestion
                    String entry = base != null ? base.relativize(path).toString() : path.toString();
                    String value = null;
```


```suggestion
                    String entry = base != null ? base.relativize(path).toString() : path.toString();
                    String value = entry;
```

WDYT?

pditommaso: I think this snipper was a mistake, it can be assumed `base` is expected to be null otherwise we would have seen this exception

https://github.com/nextflow-io/nextflow/blob/1f05451f02971b05dbf14e59fd1af081ded5cf46/modules/nf-commons/src/main/nextflow/util/HashBuilder.java#L347-L349

rcannood: alright, I made a suggestion on how to resolve this

bentsherman:
I think this function can be simplified by simply applying `hashUnorderedCollection()` to the set of key-value pairs. I will give it a try...

bentsherman:
@rcannood let me know if my changes work for you

rcannood:
> @rcannood let me know if my changes work for you

Thanks @bentsherman ! I'll run it on my other partition when I'm back at my PC.

rcannood:
# Conclusion

The fix seems to work! I wanted to write a unit test which would reproduce the issue when run on a ZFS partition, but given that it needs a very specific set of conditions to reproduce reliably, I'm not sure whether there's much benefit.

# Reproducible example

(Provided that you have a spare drive lying around)

Interpreted from [DriesSchaumont/nextflow-file-caching-example](https://github.com/DriesSchaumont/nextflow-file-caching-example):

## Set up ZFS partition

```bash
# set up a zpool on an empty disk
sudo zpool create zfsnxf /dev/<your empty disk>
sudo chown -R $(id -u):$(id -g) /zfsnxf
cd /zfs-nxf
```

## Check whether master repository works

```bash
# build nextflow
git clone https://github.com/nextflow-io/nextflow.git nextflow_master
cd nextflow_master
./compile.sh

# check whether caching works
# important: make sure the nextflow home directory is on the zfs partition!
export NXF_HOME=/zfs-nxf/.nextflow_master_home

NEXTFLOW=/zfs-nxf/nextflow_master/launch.sh

echo "----- Run 1"
$NEXTFLOW run DriesSchaumont/nextflow-file-caching-example -r acecab40435f6d3064f5ffd7c7efa2c60048f463

echo "----- Drop repo"
$NEXTFLOW drop DriesSchaumont/nextflow-file-caching-example

echo "----- Run 2"
$NEXTFLOW run DriesSchaumont/nextflow-file-caching-example -r acecab40435f6d3064f5ffd7c7efa2c60048f463 -resume
```

This returns:

```
----- Run 1

 N E X T F L O W   ~  version 25.05.0-edge

Launching `https://github.com/DriesSchaumont/nextflow-file-caching-example` [reverent_venter] DSL2 - revision: acecab40435f6d3064f5ffd7c7efa2c60048f463

executor >  local (3)
[80/1dbcfe] foo (Fifi) [100%] 3 of 3 ✔
96357abec96194ec47a13c810bd84b492c732f2b  test_assets/template.txt

96357abec96194ec47a13c810bd84b492c732f2b  test_assets/template.txt

96357abec96194ec47a13c810bd84b492c732f2b  test_assets/template.txt


----- Drop repo
----- Run 2

 N E X T F L O W   ~  version 25.05.0-edge

Pulling DriesSchaumont/nextflow-file-caching-example ...
 downloaded from https://github.com/DriesSchaumont/nextflow-file-caching-example.git
Launching `https://github.com/DriesSchaumont/nextflow-file-caching-example` [elated_koch] DSL2 - revision: acecab40435f6d3064f5ffd7c7efa2c60048f463

executor >  local (3)
[0d/905944] foo (Loulou) [100%] 3 of 3 ✔
96357abec96194ec47a13c810bd84b492c732f2b  test_assets/template.txt

96357abec96194ec47a13c810bd84b492c732f2b  test_assets/template.txt

96357abec96194ec47a13c810bd84b492c732f2b  test_assets/template.txt
```




## Check whether the fix works

```bash
# build nextflow
git clone https://github.com/rcannood/nextflow.git nextflow_fix
cd nextflow_fix
git fetch
git checkout fix-hashing
./compile.sh

# check whether caching works
# important: make sure the nextflow home directory is on the zfs partition!
export NXF_HOME=/zfs-nxf/.nextflow_master_fix

NEXTFLOW=/zfs-nxf/nextflow_fix/launch.sh

echo "----- Run 1"
$NEXTFLOW run DriesSchaumont/nextflow-file-caching-example -r acecab40435f6d3064f5ffd7c7efa2c60048f463

echo "----- Drop repo"
$NEXTFLOW drop DriesSchaumont/nextflow-file-caching-example

echo "----- Run 2"
$NEXTFLOW run DriesSchaumont/nextflow-file-caching-example -r acecab40435f6d3064f5ffd7c7efa2c60048f463 -resume
```

This returns:

```
----- Run 1

 N E X T F L O W   ~  version 25.05.0-edge

Launching `https://github.com/DriesSchaumont/nextflow-file-caching-example` [gloomy_jones] DSL2 - revision: acecab40435f6d3064f5ffd7c7efa2c60048f463

executor >  local (3)
[d0/daafa6] foo (Loulou) [100%] 3 of 3 ✔
96357abec96194ec47a13c810bd84b492c732f2b  test_assets/template.txt

96357abec96194ec47a13c810bd84b492c732f2b  test_assets/template.txt

96357abec96194ec47a13c810bd84b492c732f2b  test_assets/template.txt


----- Drop repo
----- Run 2

 N E X T F L O W   ~  version 25.05.0-edge

Pulling DriesSchaumont/nextflow-file-caching-example ...
 downloaded from https://github.com/DriesSchaumont/nextflow-file-caching-example.git
Launching `https://github.com/DriesSchaumont/nextflow-file-caching-example` [irreverent_sammet] DSL2 - revision: acecab40435f6d3064f5ffd7c7efa2c60048f463

[24/ff07d6] foo (Fifi) [100%] 3 of 3, cached: 3 ✔
96357abec96194ec47a13c810bd84b492c732f2b  test_assets/template.txt

96357abec96194ec47a13c810bd84b492c732f2b  test_assets/template.txt

96357abec96194ec47a13c810bd84b492c732f2b  test_assets/template.txt
```

modules/nf-commons/src/main/nextflow/util/HashBuilder.java
```diff
@@ -324,41 +324,47 @@ static protected Hasher hashFileSha256( Hasher hasher, Path path, Path base ) {
...
-                    // the file relative base
-                    if( base!=null )
-                        hasher.putUnencodedChars(base.relativize(path).toString());
-                    hasher.putUnencodedChars(base.relativize(path).toString());
+                    entries.put(relativePath(path, base).toString(), null);
                     return FileVisitResult.CONTINUE;
                 }
             });
+
+            return hashUnorderedCollection(hasher, entries.entrySet(), HashMode.SHA256);
```
pditommaso: That's a good idea because bring down the complexity from O(n*log(n)) to 2*O(n). However it could be further optimised implementing the `hashUnorderedCollection` (which is quite simple) in the `walkFileTree` block (also the HashMap would not needed anymore). This could avoid avoid requiring a separate loop to compute the hash.

bentsherman: Good idea. I found a way to eliminate the intermediate hashmap and still re-use `hashUnorderedCollection()`

pditommaso:
While this implementation is correct (and nice the use of Files.walk with stream), I found preferable keep using `Files.walkFileTree` to keep the implementation easier to follow and avoid an extra check of `Files.isDirectory` (it could add an extra network roundtrip when using remote files, tho I agree it's not an expected scenario). 

It's also preferable to avoid changing the type of exception thrown in case of error. Closing in favour of https://github.com/nextflow-io/nextflow/pull/6197/files

PR #6114, opened 2025-05-22
fix test cases for validating shell

nttg8100:
+ fix test cases for validating shell in issue https://github.com/nf-core/rnaseq/issues/1554#

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy
@@ -377,6 +377,8 @@ class TaskConfig extends LazyMap implements Cloneable {
 
     List<String> getShell() {
         final value = get('shell')
+        println '>>> VALIDATE SHELL CALLED'
+        println value
         if( !value )
             return BashWrapperBuilder.BASH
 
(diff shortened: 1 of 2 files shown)


nttg8100:
> Not sure what's this PR aims for?
Hi @pditommaso 
There is an error when the user cutomize the process.shell with contains the "\n".
https://github.com/nf-core/rnaseq/blob/master/nextflow.config#L299C1-L307C1
```
process.shell = """\
bash

set -e # Exit if a tool returns a non-zero status/exit code
set -u # Treat unset variables and parameters as an error
set -o pipefail # Returns the status of the last command to exit with a non-zero status or zero if all successfully execute
"""
```

It will failed when you run the workflow with this process.shell
https://github.com/nf-core/rnaseq/issues/1554. I saw that on the test cases for this on the latest version of nextflow, it is commented and also not raise the error. So I created this PR, to uncomment them, also in the validation shell function, I make it to raise error.

nttg8100:
> It does not use an exception on purpose, otherwise it will break most of nf-core pipelines

I think it is already breaks any nf-core pipeline that has the process.shell has "/n" in the nextflow.config with the 25.04.2 nextflow version. Ewels gives a temporary solution to redefine the process.shell in a config file. Any suggestion?

nttg8100:
Hi @pditommaso, without raising the error, the pipeline still be failed. The error message will be liked this with "permission denied" but there is no issue related to permission at all, just about the configuration of process.shell.
https://github.com/nf-core/rnaseq/issues/1554

I think this PR should be merged into a new release to fix this issue.

PR #6121, opened 2025-05-26
Include retries in git repo requests

jorgee:
This PR include retries when performing request in git repositories.

The request errors are managed at `RepositoryProvider.checkResponse`, where some `AbortException` or `RateLimitExceededExceptions` are thrown for some HTTP error codes, or by I`OExceptions` thrown by the` URLConnection` methods. They include `SocketExceptions` and `50X` error codes. I have made all `IOExceptions` retried by default, but this could be modified by including other codes in the `checkResponse` method.


christopher-hakkaart (approved):
Documentation changes look great.

docs/git.md
```diff
@@ -47,6 +47,18 @@ The following configuration properties are supported for each provider configura
...
+
+`providers.<provider>.retryPolicy.jitter`
+: Jitter value when retrying failed requests (default: `0.25`).
+
+`providers.<provider>.retryPolicy.maxAttempts`
+: Max attempts when retrying failed requests (default: `10`).
+
+`providers.<provider>.retryPolicy.maxDelay`
+: Max delay when retrying failed requests (default: `90s`).
+
```
pditommaso, line 61: we are starting to have too many of those settings replicated for different components. Wonder if it would not be better having a common one

bentsherman: I think that would be best. I would be fine with a `retryPolicy` scope that controls all retry settings, or something like `retryPolicy.<name>.*` to control retry settings for specific

bentsherman: @jorgee I can finish this one so that you can focus on AWS SDK v2

bentsherman:
I have added `nextflow.retryPolicy` and used it for the following operations:
- git repository provider
- ~~k8s client (replaces `k8s.client.retryPolicy`)~~
- file publishing (replaces `workflow.output.retryPolicy`)

There are a few other places that use retry settings, namely the grid and cloud executors. In a future iteration I think we could consolidate them as well, perhaps using the `$<name>` syntax in the `executor` scope to apply different settings to different contexts

modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy
```diff
@@ -342,4 +363,52 @@ abstract class RepositoryProvider {
...
+    protected <T> T apply(CheckedSupplier<T> action) {
+        // define the retry condition
+        final cond = new Predicate<? extends Throwable>() {
+            @Override
+            boolean test(Throwable t) {
+                // checkResponse method can throw RateLimitExceedExceptions which are retried and AbortedException which not retried.
+                // Other methods can throw IOExceptions including for 50x response codes which are retried by default.
+                if( t instanceof RateLimitExceededException || t.cause instanceof RateLimitExceededException)
+                    return true
+                if( t instanceof IOException || t.cause instanceof IOException )
```
pditommaso: `IOException` is too generic. It would be preferable having a more granular control on http error codes [here](https://github.com/nextflow-io/nextflow/blob/4b4568065ef78af080d4f3e2116a75fdb381d0ab/modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy#L253-L253). 

However it can be difficult to achieve this in the current implementation. Wonder if it should be used a different http client (see WaveClient)

pditommaso:
This is the PR implementing the use of Java HttpClient https://github.com/nextflow-io/nextflow/pull/6188

docs/reference/config.md
```diff
@@ -994,18 +994,6 @@ The following settings are available:
...
-
-`k8s.retryPolicy.jitter`
-: Jitter value when retrying failed API requests (default: `0.25`).
-
-`k8s.retryPolicy.maxAttempts`
-: Max attempts when retrying failed API requests (default: `4`).
-
-`k8s.retryPolicy.maxDelay`
-: Max delay when retrying failed API requests (default: `90s`).
-
```
pditommaso: it may be better to exclude K8s from this since its' a separate executor/plugin, also to avoid a breaking change in the current settings.  @bentsherman what do you think?

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
```diff
@@ -436,37 +438,46 @@ class AssetManager {
...
-        getManifest0()
+        try {
+            def config = getConfig0()
+            def manifestOpts = config.manifest as Map ?: Collections.emptyMap()
+            return new Manifest(manifestOpts)
+        }
+        catch( Exception e ) {
+            log.warn "Cannot read project config -- Cause: ${e.message ?: e}"
+            return new Manifest(Collections.emptyMap())
+        }
```
pditommaso, line 450: Likely this is not needed anymore, because the retry strategy works at a lower level

docs/reference/config.md
bentsherman: We can add a fallback like we do [here](https://github.com/nextflow-io/nextflow/blob/d1e6836ea8bf85c26c0279a74e590593ce9849b2/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy#L234-L241) to preserve backward compatibility if you want

I figure we would eventually include other executors like aws / google / azure in this as well. Since they all boil down to http requests, it seems reasonable to use the same retry settings for all of them

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
bentsherman: what do you mean?

docs/reference/config.md
pditommaso: Not sure ewe should go with this unification. The retry policy could be very different depending the infra for example azure requires much more retries compared to AWS. same could be for K8s depending how is deployed. Think it's preferable to keep those independent and have a common one for nextflow core runtime

bentsherman: Okay :+1:

pditommaso: I've realised that having retry option in the nextflow.config results into a circular dependency because to read that you need to access the git repo for which you need the  retry policy. it would make more sense to keep in the scm file

pditommaso: or control via env variables

pditommaso: I'm going to have a try to `scm` file approach

bentsherman: You're right, and it probably doesn't make sense for users to put retry settings in their pipeline config anyway.

I suggest we read these settings from the user config at `$NXF_HOME/nextflow.config` and document that certain settings like `nextflow.retryPolicy` are only read from this config.

I would also move the [scm config](https://nextflow.io/docs/latest/git.html#git-configuration) into the user config and make it a first-class config scope, as it is currently treated separately from other scopes and is not handled by the new config parser

bentsherman: > I'm going to have a try to scm file approach

Using the scm file doesn't make sense because these settings are also used for file publishing. Better to make the scm config part of the user config

pditommaso:
Since there are several overlapping changes I've created a new PR. Closing this in favour of https://github.com/nextflow-io/nextflow/pull/6195

docs/reference/config.md
pditommaso: i'm open to alternatives about retry settings but `scm` cannot be moved in the user config, both to avoid circular deps and usage made by platform

bentsherman: okay, continuing discussion in the other PR

PR #6131, opened 2025-05-27
Improve documentation of splitCsv operator

### LINKED ISSUE #6120: Improve documentation on splitCsv
## New feature

The documentation at https://www.nextflow.io/docs/latest/reference/operator.html#splitcsv needs to be improved with an example of how to actually read in an CSV file. Surely the expected use case is not for users to type in their CSV text inline.

bentsherman:
Close #6120


christopher-hakkaart (approved):
Looks good.

PR #6135, opened 2025-05-28
Use Priority to (not) overwrite Executor

Lehmann-Fabian:
Since `ServiceName.important` is deprecated, I added support for the Priority annotation to work with the same behavior as other Nextflow Extension Points.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/ExecutorFactory.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/ExecutorFactory.groovy
@@ -22,6 +22,7 @@ import groovy.transform.PackageScope
 import groovy.util.logging.Slf4j
 import nextflow.Session
 import nextflow.executor.local.LocalExecutor
+import nextflow.plugin.Priority
 import nextflow.script.BodyDef
 import nextflow.script.ProcessConfig
 import nextflow.script.ScriptType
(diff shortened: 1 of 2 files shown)


bentsherman:
`Plugins.getPriorityExtensions()` is normally used to load plugins in a priority-aware manner, and should ideally be used here as well. Though I'm not sure why the plugin manager is provided here instead of just using `Plugins`

Lehmann-Fabian:
I think we can merge both constructors now. What do you think?

pditommaso:
The `ServiceName.important` is deprecated, but there's no plan to remove it. Therefore I don't see a parallel logic necessary

Lehmann-Fabian:
Thank you for the clarification. However, I generally assumed that you eventually will remove deprecated elements. Therefore, I believe it's good practice to support both the deprecated and the new logic in parallel. This allows plugins to continue using `important` as long as it's available and ensures a smooth transition once it's removed, maintaining backward compatibility. This new approach is also consistent with how other extension points are currently managed.
If there is no plan to deprecate the `important` feature, should I remove the `@Deprecated` annotation?

pditommaso:
Even if it's deprecated we are not going to remove this annotation option, we are also using internally for other plugins. I've added more docs about the rationale of the deprecation 👉  3a81a4a3

PR #6139, opened 2025-05-29
Ensure that groupKey objects are treated as their groupTarget for the purposes of joins and combines

robsyme:
When using operators like join or combine, if one side of the join uses a plain value (e.g., a string) and the other side uses a GroupKey (wrapping the same string), the join may fail to match them—even though they represent the same logical key.

This is because, although GroupKey implements `equals` and `hashCode` to delegate to its target, there are subtle issues in Groovy/Java equality and hash semantics, especially when comparing a GroupKey to a plain value from the other side. This can lead to unpredictable or inconsistent join results, as described in #4104.

```groovy
v = 'BATCH1'
gk = groupKey(v, 3)
println v == gk // false
println gk == v // true
```
This asymmetry can cause joins to miss matches, depending on which side emits the GroupKey. The ordering of items on left or right can depend on the order of items in channels, leading to difficult-to-debug timing issues. This issue affects the join and combine operators which will do these sorts of equality checks.

Both JoinOp and CombineOp use the `makeKey` utility (which creates a KeyPair) to extract keys from tuples for matching. By ensuring that all keys are "unwrapped" from GroupKey to their target value, we guarantee that:
A key `'BATCH1'` and a key `groupKey('BATCH1')` are treated as the same for the purposes of joining or combining.
This removes the asymmetry and inconsistency in key matching, regardless of which side emits a GroupKey or a plain value.


pditommaso:
In my view this could be managed, unwrapping temporary the key for the requirement comparison/indexing operation, but continuing to emit as wrapped GroupKey. This essential because this was designed to traverse a chain of operators

robsyme:
If so, we may have to consider the pathological case where two groupKeys are joined, but the groupKey contains different size values 😬

pditommaso:
How that would happen in practice?

bentsherman:
Now I hope you see why I would rather deal with this in operators v2

robsyme:
Ok, I've fixed it to ensure that the groupKey is preserved if combined with the target. I'll accept that preserving the key comes at the cost of some additional complexity 😬

bentsherman:
Closing since I don't think it's worth the added complexity. v2 operators coming soon...

PR #6146, opened 2025-05-30
Update config.md

nap23carc:
Fixing -C to -c in the "tip" section

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -12,7 +12,7 @@ When a pipeline script is launched, Nextflow looks for configuration files in mu
 4. Config files specified using the `-c <config-files>` option
 
 :::{tip}
-You can alternatively use the `-C <config-file>` option to specify a fixed set of configuration files and ignore all other files.
+You can alternatively use the `-c <config-file>` option to specify a fixed set of configuration files and ignore all other files.
 :::
 
 (config-syntax)=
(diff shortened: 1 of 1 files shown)


pditommaso:
Thanks, but case matters here and `-C` is different from `-c`

nap23carc:
@pditommaso I had an error when trying to use -C without any other configurations and assumed it was a typo. Glad to be corrected and learn a bit more about Nextflow. Thank you.

docs/config.md
```diff
@@ -12,7 +12,7 @@ When a pipeline script is launched, Nextflow looks for configuration files in mu
 4. Config files specified using the `-c <config-files>` option
 
 :::{tip}
-You can alternatively use the `-C <config-file>` option to specify a fixed set of configuration files and ignore all other files.
+You can alternatively use the `-c <config-file>` option to specify a fixed set of configuration files and ignore all other files.
```
adamrtalbot, line 15: This is isn't correct

1. `-c` Supplements config files, i.e. it adds them on top of other config files
2. `-C` Overrides config files, i.e. it replaces all other configuration

This tooltip refers to 2, `-C` which replaces the other config files. How about this to make the language more clear?

```suggestion
You can also use the `-C <config-file>` option to load a specific set of configuration files, ignoring any others.
```

adamrtalbot (changes requested):

bentsherman:
This PR seems to be based on a misunderstanding about the `-C` vs `-c` options. Closing without merging

PR #6156, opened 2025-06-03
Replace foo/bar with better names in docs

bentsherman:
This PR removes the use of `foo` and `bar` in the docs with better names, e.g. `hello` and `bye` for processes, `alpha` and `beta` for parameters, and so on.

### DIFF
--- a/docs/cache-and-resume.md
+++ b/docs/cache-and-resume.md
@@ -144,19 +144,19 @@ Sometimes a process needs to merge inputs from different sources. Consider the f
 
 ```nextflow
 workflow {
-    ch_foo = channel.of( ['1', '1.foo'], ['2', '2.foo'] )
-    ch_bar = channel.of( ['2', '2.bar'], ['1', '1.bar'] )
-    gather(ch_foo, ch_bar)
+    ch_bam = channel.of( ['1', '1.bam'], ['2', '2.bam'] )
+    ch_bai = channel.of( ['2', '2.bai'], ['1', '1.bai'] )
+    check_bam_bai(ch_bam, ch_bai)
 }
 
-process gather {
+process check_bam_bai {
     input:
-    tuple val(id), file(foo)
-    tuple val(id), file(bar)
+    tuple val(id), file(bam)
+    tuple val(id), file(bai)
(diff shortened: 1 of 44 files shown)


docs/cache-and-resume.md
```diff
@@ -144,19 +144,19 @@ Sometimes a process needs to merge inputs from different sources. Consider the f
...
     input:
-    tuple val(id), file(foo)
-    tuple val(id), file(bar)
+    tuple val(id), file(bam)
+    tuple val(id), file(bai)
 
     script:
     """
-    merge_command $foo $bar
+    check_bam_bai $bam $bai
```
christopher-hakkaart, line 159: Having `check_bam_bai` as the process name and the tool in the script block might be confusing.

docs/workflow.md
```diff
@@ -333,7 +340,7 @@ process foo {
     result = "$data world"
 }
 
-process bar {
+process toUpper {
```
christopher-hakkaart: I don't have strong opinions, but for some examples we use snake case and some we use camel case. Do you think we could standardize?

christopher-hakkaart (approved):
My comments aren't blocking so I've approved. I added two comments.

Comment 1: I find it confusing to have a process name the same as the tool in the script. E.g., `check_bam_bai`

Comment 2: Some process names use camel case and some use snake case. Can we standardize?

docs/cache-and-resume.md
bentsherman: I think this is actually a good practice since the process is essentially a wrapper around the tool, so no point in calling it something else. nf-core does this, although they also make the process name uppercase, but we don't use uppercase much in the docs

docs/workflow.md
bentsherman: It would be nice but I'd rather address that separately

bentsherman: Actually this looks like a quick fix, I'll go ahead and do it

christopher-hakkaart: Only if it's not a hassle, it can be wrapped up in another PR or slowly implemented over time.

Issue #6092, opened 2025-05-16
Support for Pixi


adamrtalbot:
## New feature

Pixi is a package manager equivalent to Conda which is growing in popuarity recently.

## Use case 

Equivalent to Conda, Mamba, Spack, etc. We can use it to set environments per process using a Pixi TOML.

## Suggested implementation 

Here's a hacky version inspired by https://tiendu.github.io/2025/05/10/nextflow-pixi-eng.html. We could probably implement something similar inside a process.

`main.nf`:
```groovy

process PIXI {

    debug true

    beforeScript '''
    [ -d .pixi ] && [ -f pixi.lock ] || pixi install
    eval "$(pixi shell-hook)"
    '''

    input:
        path pixi_lock

    output:
        stdout

    script:
    """
    cowpy "hello pixi"
    """
}

workflow {
    pixi_lock = Channel.fromPath("pixi.toml", checkIfExists: true)

    PIXI(pixi_lock)
}
```

`pixi.toml`
```toml
[project]
name = "test-pixi-env"

channels = ["conda-forge"]
platforms = ["linux-64", "osx-arm64", "win-64"]

[dependencies]
cowpy = "1.1.*"
python = "3.13.*"
```

```console
N E X T F L O W   ~  version 25.04.1

Launching `./main.nf` [nauseous_hoover] DSL2 - revision: 8f2d7021a8

executor >  local (1)
[39/fcb524] PIXI (1) [100%] 1 of 1 ✔
 ______________________________________________________ 
< Cowacter, eyes:default, tongue:False, thoughts:False >
 ------------------------------------------------------ 
     \   ^__^
      \  (oo)\_______
         (__)\       )\/\
           ||----w |
           ||     ||
```

bentsherman:
It should have it's own class

bguo068:
I use this variant in the config, so that the process can pick up the environment even it is run outside the project folder. It has been working for me either running on a grid with slurm or locally. 
```groovy
beforeScript = """export PIXI_PROJECT_MANIFEST=${projectDir}/pixi.toml && eval "\$(pixi shell-hook)" """
```

PR #6157, opened 2025-06-03
Add pixi support

edmundmiller:
- **test(#6092):Add Pixi environment support**
- **test(#6092): Add Pixi environment tests**
- **test(#6092): Add integration tests for PixiCache functionality**
- **test: Update sayHello process to use cowpy**

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -335,6 +335,7 @@ class BashWrapperBuilder {
         binding.before_script = getBeforeScriptSnippet()
         binding.conda_activate = getCondaActivateSnippet()
         binding.spack_activate = getSpackActivateSnippet()
+        binding.pixi_activate = getPixiActivateSnippet()
 
         /*
          * add the task environment
@@ -388,7 +389,7 @@ class BashWrapperBuilder {
         binding.fix_ownership = fixOwnership() ? "[ \${NXF_OWNER:=''} ] && (shopt -s extglob; GLOBIGNORE='..'; chown -fR --from root \$NXF_OWNER ${workDir}/{*,.*}) || true" : null
 
         binding.trace_script = isTraceRequired() ? getTraceScript(binding) : null
-        
+
         return binding
     }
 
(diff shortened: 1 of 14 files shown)


modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
```diff
@@ -554,6 +555,29 @@ class BashWrapperBuilder {
...
+            result += "eval \"\$(pixi shell-hook --shell bash)\" && "
+            result += "cd \"\$OLDPWD\"\n"
+        }
+        else {
+            // Direct activation from environment directory
+            result += "cd ${Escape.path(pixiEnv)} && "
+            result += "eval \"\$(pixi shell-hook --shell bash)\" && "
+            result += "cd \"\$OLDPWD\"\n"
+        }
+        return result
```
adamrtalbot, line 578: Where is OLDPWD defined? Doesn't this change to the projectDir? Should we overwrite projectDir with the pixiFile contents?

While it's uncessary duplication, I prefer copying the pixi lock file into the working dir and using it at source, so we have a record. This has the additional advantage of making the code more simple and supporting a `.pixi` or `pixi.lock` file (psuedocode):

```suggestion
        // assuming you have copied the pixi file to working dir
        if( pixiFile.exists() ) {
            result += "[ -d .pixi ] && [ -f pixi.lock ] || pixi install"
            result += "eval \"\$(pixi shell-hook --shell bash)\""
        }
        return result
```

Downside - massive duplication of text files in every work dir.

edmundmiller: Do we copy the conda lock currently into the workdir? I don't think so, but I should double-check.

It's not a bad idea, it would also be cool if Nextflow did that on a wave build...

pditommaso:
I'm start thinking it would be better having a different abstraction for tooling, something like 

```
package "samtools=1 bamtools=2", provider: "conda|pixi" 
```

pditommaso:
it could be modelled as this class used by wave 

https://github.com/seqeralabs/libseqera/blob/af9ace8ea40ab95f6317b26a33a76860149f89ae/wave-api/src/main/java/io/seqera/wave/api/PackagesSpec.java#L32-L54

pditommaso:
Closing in favour of #6342

Issue #5575, opened 2024-12-05
Wait for Azure Batch job quota to clear before submitting a new job


adamrtalbot:
## New feature

Occasionally, you can hit the Azure Batch job limit when running a pipeline, leading to this error:

```
Error executing process > 'NFCORE_VIRALRECON:ILLUMINA:VARIANTS_IVAR:IVAR_VARIANTS_TO_VCF (SAMPLE2_PE)'

Caused by:
  Status code 409, "{
    "odata.metadata":"https://name.region.batch.azure.com/$metadata#Microsoft.Azure.Batch.Protocol.Entities.Container.errors/@Element","code":"ActiveJobAndScheduleQuotaReached","message":{
      "lang":"en-US","value":"Active job and job schedule quota for the account has been reached.\nRequestId:d28008e3-4018-41cc-90ff-6d96a7952ea3\nTime:2024-12-05T02:43:12.5743776Z"
    }
  }"
```

We could catch this error and just wait for the job quota to clear. This could be configurable as an `azure.batch` config option.

## Usage scenario 

Set `azure.batch.behaviourOnJobLimit = 'retry'` and Azure will retry with a backoff until cancelled. Alternatively, set it to `error` and it will do the current behaviour and raise an error.

Raise a warning to the console so the user is aware.

This would allow pipelines to continue, albeit stuck in a queue. As jobs are cleared, the pipelines will continue.

## Suggest implementation 

tbc

Related: https://github.com/nextflow-io/nextflow/issues/4792

adamrtalbot:
Still relevant

PR #6159, opened 2025-06-04
feat: Azure Batch eagerly terminates jobs after all tasks have been submitted

### LINKED ISSUE #5839: Nextflow eagerly terminates Azure Batch jobs during execution
## New feature

Based on comment: https://github.com/nextflow-io/nextflow/pull/3927#discussion_r1189944527

Azure Batch jobs are terminated during pipeline closing down, see https://github.com/nextflow-io/nextflow/pull/3927.

During that PR, we discussed eagerly changing the setting of each job to `onAllTasksComplete = 'terminate'` but never looked further into it.

## Use case 

The main benefit of doing it this way is that if Nextflow fails to close properly (i.e. is killed), most active jobs will be marked as to be terminated and be removed from the active job quota.

## Suggested implementation 

> We just need an "onProcessClose" event to notify when all tasks are created. I've already added it in a number of PRs, like this one: https://github.com/nextflow-io/nextflow/pull/3818

The suggestion from the first PR which could be implemented ☝

adamrtalbot:
## Summary

Fixes Azure Batch "job leak" issue where jobs remain in Active state even after task completion, causing quota exhaustion and preventing multiple pipelines from running simultaneously.

**Problem**: Jobs consume quota slots unnecessarily, blocking other workflows  
**Solution**: Leverage Azure Batch's native auto-termination to release quota immediately when tasks complete

### Problem Addressed
Azure Batch has a limitation where jobs remain in an "Active" state even after all their tasks complete. This causes:
- **Quota exhaustion**: Active jobs count against Azure Batch service quotas
- **Pipeline blocking**: Multiple Nextflow pipelines can't run simultaneously due to quota limits
- **Resource waste**: Jobs consume quota slots unnecessarily

#### Job Auto-Termination Configuration
```groovy
// In AzBatchOpts.groovy
@ConfigOption
@Description("When the workflow completes, set all jobs to terminate on task completion (default: true)")
final Boolean terminateJobsOnCompletion
```

**Default behavior**: `terminateJobsOnCompletion = true` (enabled by default)

#### Job Termination Mechanism 
The service implements a two-phase termination approach:

##### Phase 1: Set Jobs to Auto-Terminate
```groovy
protected void terminateJobs() {
    for( String jobId : allJobIds.values() ) {
        final job = apply(() -> client.getJob(jobId))
        final poolInfo = job.poolInfo
        
        final jobParameter = new BatchJobUpdateContent()
            .setOnAllTasksComplete(OnAllBatchTasksComplete.TERMINATE_JOB)  // Key setting
            .setPoolInfo(poolInfo)
        
        apply(() -> client.updateJob(jobId, jobParameter))
    }
}
```

##### Phase 2: Cleanup on Workflow Completion
```groovy
@Override
void close() {
    // Terminate all jobs to prevent them from occupying quota
    if( config.batch().terminateJobsOnCompletion ) {
        terminateJobs()
    }
    
    // Delete all jobs (if configured)
    if( config.batch().deleteJobsOnCompletion ) {
        cleanupJobs()
    }
}
```

#### How Auto-Termination Works
- Uses Azure Batch's native `OnAllBatchTasksComplete.TERMINATE_JOB` setting
- This tells Azure Batch: "When all tasks in this job finish, automatically terminate the job"
- Jobs transition from "Active" → "Terminating" → "Completed"

#### Eager Termination Flow
1. **Job Creation**: Jobs created normally with pool assignments
2. **Task Submission**: Tasks submitted to jobs as usual  
3. **Workflow Completion**: When Nextflow workflow completes
4. **Batch Update**: Nextflow updates all existing jobs with `OnAllTasksComplete = TERMINATE_JOB`
5. **Auto-Termination**: Azure Batch automatically terminates jobs as their tasks complete
6. **Quota Release**: Terminated jobs no longer consume quota

#### Resource Management
- **Quota preservation**: Jobs don't consume quota after completion
- **Multiple pipelines**: Allows running multiple Nextflow workflows simultaneously
- **Clean resource usage**: Prevents resource leaks in Azure Batch

#### Operational Improvements
- **Zero user impact**: Completely transparent to workflow users
- **Backward compatible**: Can be disabled if needed
- **Automatic cleanup**: No manual intervention required

### Configuration Options

Users can control the behavior:
```groovy
azure {
    batch {
        terminateJobsOnCompletion = true   // Enable eager termination (default)
        deleteJobsOnCompletion = false    // Optionally delete jobs entirely
        deleteTasksOnCompletion = true    // Clean up individual tasks
    }
}
```

#### Job Lifecycle Management
- **Job Reuse**: Same job ID reused for same Process+PoolId combination
- **Pool Independence**: Each pool can have its own jobs
- **Batch Updates**: Uses Azure Batch `updateJob` API for termination setting

#### Error Handling
```groovy
catch (HttpResponseException e) {
    if (e.response.statusCode == 409) {
        log.debug "Azure Batch job ${jobId} already terminated, skipping termination"
    } else {
        log.warn "Unable to terminate Azure Batch job ${jobId} - Status: ${e.response.statusCode}"
    }
}
```

## Impact

This implementation provides an elegant solution to Azure Batch's job quota problem by leveraging Azure's native auto-termination feature. It ensures that jobs automatically terminate when their tasks complete, preventing quota exhaustion while maintaining full compatibility with existing workflows.

## Related

- Closes #5839

---

> [!NOTE]
> Implements eager Azure Batch job auto-termination after task submission with automatic job recreation on 409 conflicts; updates tests, logging, and config docs.
> 
> - **Azure Batch service (`plugins/nf-azure/.../AzBatchService.groovy`)**:
>   - Eager auto-termination:
>     - New `setJobTermination(jobId)` and `setAutoTerminateIfEnabled(jobId, taskId)`; `runTask` now calls them after submission.
>     - Removed end-of-workflow `terminateJobs()` usage from `close()`.
>   - Resilient task submission:
>     - New `submitTaskToJob(...)` handles `409` on auto-terminated jobs by `recreateJobForTask(...)` and resubmits.
>     - `recreateJobForTask(...)` synchronizes and updates `allJobIds` mapping.
>   - Other:
>     - Lower `destFile(...)` log level to `trace`; fix "creddentials" typo; minor comments/docs; preserve job constraints on create.
> - **Tests**:
>   - Added coverage for eager auto-termination and job recreation concurrency; removed close-time termination expectation; misc updates.
> - **Docs (`docs/reference/config.md`)**:
>   - Clarify `azure.batch.terminateJobsOnCompletion`: set jobs to terminate on task completion (default `true`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 899f0a2d149008dfeb1d9528b4cb92f9c4695f71. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchProcessObserver.groovy
```diff
@@ -0,0 +1,74 @@
...
+        // Check if auto-termination is enabled
+        if( !batchService?.config?.batch()?.terminateJobsOnCompletion ) {
+            log.trace "Azure Batch job auto-termination is disabled, skipping eager termination for process: ${processor.name}"
+            return
+        }
+
+        // Find and set auto-termination for all jobs associated with this processor
+        batchService.allJobIds.findAll { key, jobId ->
+            key.processor == processor
+        }.values().each { jobId ->
```
adamrtalbot: This feels a bit convoluted to me...is there an easier way?

pditommaso:
Retry now

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchProcessObserver.groovy
```diff
@@ -0,0 +1,74 @@
...
+    AzBatchProcessObserver(Session session) {
+        // Session not needed, but required by factory interface
+    }
+
+    /**
+     * Called when a process terminates (all tasks have been submitted).
+     * Sets Azure Batch jobs to auto-terminate when all tasks complete.
+     */
+    @Override
+    void onProcessTerminate(TaskProcessor processor) {
```
pditommaso, line 46: Why using an observer instead of having this logic in the task hander? 

https://github.com/nextflow-io/nextflow/blob/08c0027038cd99789dd790f2082f50870bc2de91/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchTaskHandler.groovy#L127-L127

adamrtalbot: It's not a task, it's a job! 

In Azure, job = queue.

adamrtalbot: See comment here: https://github.com/nextflow-io/nextflow/pull/3927#discussion_r1189797069

Basically, it needs to wait until the last task of a process has been submitted to Az Batch, then make the job terminate after completion.

pditommaso: Fair, but I still don't see a big value using the trace observer. Would not make more sense to keep this in the cleanup logic here 

https://github.com/nextflow-io/nextflow/blob/5839_azure_batch_jobs_terminate_upon_completion/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L1065-L1080


all related metadata should be accessible in the AzBatchService object

adamrtalbot: That is how it currently works, see [terminateJobsOnCompletion](https://github.com/nextflow-io/nextflow/blob/14a249166f6b9f7ceb8e850cf426ad02b2801ecb/docs/reference/config.md#L352-L353) and [deleteJobsOnCompletion](https://github.com/nextflow-io/nextflow/blob/14a249166f6b9f7ceb8e850cf426ad02b2801ecb/docs/reference/config.md#L326-L327). However it's causing us problems. If Nextflow dies, all jobs are left in active state which consumes a very limited quota. For some context, I had to argue with Azure support a lot to get 1000 active jobs as a quota, and with 1 process equaling 1 job you can use this up in a matter of days. Once you have done this, the only way to run Nextflow again is to go into your Azure Batch account and manually remove some jobs.

By aggressively going and setting jobs to automatically terminate, we can reduce the number of active jobs to all but the ones with running tasks when a Nextflow process dies. This is reducing the pressure on active jobs as much as possible. Between this and the 30 day cooldown, we should be at effectively zero active jobs for normal running which is what we should aim for.

pditommaso: Then not sure how much this approach will improve, most of times processes termination on pipeline completion, so if the execution is killed abruptly the behaviour will be more or less the same. 

I wonder instead if it could be a problem with the cleanup execution. Are you able to replicate the problem? do you have any execution logs for effected pipelines ?

adamrtalbot: > most of times processes termination on pipeline completion

Arrgghh that's frustrating.

> do you have any execution logs for effected pipelines ?

Attached one from last night.

[nf-5ZBk4BU4tGxbqU.log](https://github.com/user-attachments/files/20772237/nf-5ZBk4BU4tGxbqU.log)

bentsherman: It sounds like the earliest point at which we could safely terminate the azure job is right after all tasks for the corresponding process have been submitted to azure batch.

The TaskProcessor can signal when all tasks are "pending" (submitted to nextflow but not to azure) and when all tasks have completed, but not when all tasks are submitted to azure. I think the trace observer is the only way to do this, because it needs to watch the task submissions to figure out when all tasks have been submitted.

In fact, as I write this, even that isn't enough because you could have task retries. If I submit all the tasks, terminate the job, then a task fails and I need to retry it, I assume that wouldn't work if the job was already terminated? Therefore we actually do have to wait until all tasks are completed.

adamrtalbot: Damn, you're right. Back to the drawing board.

adamrtalbot: > when all tasks have completed

@bentsherman, is there an event I can listen to check that the tasks have been completed (failed or succeeded, but not waiting for a resume)? That would be the safest option.

ghislaindemael:
Hi !
Posting my comment here as indicated in Slack.
Our 'use case' is relatively simple, we just run big pipelines (aka with many jobs), and with completed jobs being deleted only when all tasks are completed (or cancelled), we reach our quota relatively quickly, allowing us to have only a few pipelines running with 2-3 active jobs instead of dozens.

Should we not care about resuming a workflow, since task outputs are stored outside of the workdir, a simple flag to delete the job once the last task has finished it's lifecycle would be all we need.

The downside is that to "resume" a workflow, me must find a way to tell Nextflow that task outputs already exist and are stored at X location, as well as supply them. Which leads to the fact that the easiest solution would be to relaunch the whole workflow should we have an error arise somewhere during processing.

ghislaindemael:
@adamrtalbot 

My error, indeed I meant as we publish the results outside of the workdir (e.g. in Blob Storage), we can query them from here and thus delete them from the Batch VMs to remove the load and free up jobs for the quota.

luanjot:
@adamrtalbot we are also looking for a solution for this. We have executions that have hundreds of jobs sometimes, so, even in proper executions without errors, we are limited to 2 or 3 parallel executions per batch account. In our case, at any given moment, we would have something like:

Run1:
200 jobs marked as finished, e.g. `[100%] X of X ✔`
2 jobs ongoing with pending tasks
20 jobs not started

Run2:
57 jobs marked as finished, e.g. `[100%] X of X ✔`
4 jobs ongoing with pending tasks
200 jobs not started

So, from Azure's perspective, we have 200+2+57+4=263 jobs ongoing. As the runs progress, we have more and more jobs open and we reach the limit very quickly.

We are seeing if we can modify / extend the `nf-azure` plugin to handle this by adding some sort of cron job that deletes the jobs marked as finished (with a ✔), but you seem to say that this might cause issues when resuming the tasks? Why is that? It seems to me that the job names are different when it resumes the execution, no?

luanjot:
So using a cron job that validates when all tasks have completed successfully would work, right? In this case, the tasks have already completed. Or can we add it to the bit that runs and "decides" that all the tasks have been completed?

luanjot:
I have too many active jobs because nextflow does not close them, not because they are actually active.

Any job that has been marked by nextflow with a ✔ is, in my opinion, finished and will not be re-used for anything never again, however, nextflow does not close it until the full execution of the run is finished. This is the behaviour that I think is incorrect. If the run takes 2 days to run, the first task that finished 47 hours ago is still marked as "active" in batch because Nextflow does not close it even though it will never be used again. I think it is Nextflow that is not using the Batch account properly.

luanjot:
> Unfortunately, there isn't a way of Nextflow knowing the future and determining if another task will be submitted to the job which makes it tricky to know when to close a job.

Then how does it "decide" to add the ✔?

pditommaso:
The check is shown when no more tasks for that process need to be executed ie. the process execution is complete

pditommaso:
It could be done with a TraceObserver(V2). If i'm not wrong you already made a pr for that

docs/reference/config.md
```diff
@@ -936,9 +938,9 @@ The following settings are available:
 
 : You can specify the network as a full or partial URL. For example, the following are all valid URLs:
 
-  - `https://www.googleapis.com/compute/v1/projects/{project}/global/networks/{network}`
-  - `projects/{project}/global/networks/{network}`
-  - `global/networks/{network}`
+- `https://www.googleapis.com/compute/v1/projects/{project}/global/networks/{network}`
+- `projects/{project}/global/networks/{network}`
+- `global/networks/{network}`
```
adamrtalbot: ARRRGGHHHH VSCODE STOP FORMATTING MAH STUFF.

plugins/nf-azure/src/test/nextflow/cloud/azure/config/AzureConfigTest.groovy
```diff
@@ -119,7 +119,7 @@ class AzureConfigTest extends Specification {
                                              location: LOCATION,
                                              autoPoolMode: true,
                                              allowPoolCreation: true,
-                                             terminateJobsOnCompletion: false,
+                                              terminateJobsOnCompletion: false,
```
adamrtalbot:
```suggestion
                                             terminateJobsOnCompletion: false,
```

pditommaso:
It looks neat but not getting the general logic. Can you provide some hint ?

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -443,14 +449,23 @@ class AzBatchService implements Closeable {
 
     protected String createJob0(String poolId, TaskRun task) {
         log.debug "[AZURE BATCH] created job for ${task.processor.name} with pool ${poolId}"
+
```
pditommaso:
```suggestion
```

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -977,30 +1043,53 @@ class AzBatchService implements Closeable {
...
      */
-    protected void terminateJobs() {
-        for( String jobId : allJobIds.values() ) {
-            try {
-                log.trace "Setting Azure job ${jobId} to terminate on completion"
-
-                final job = apply(() -> client.getJob(jobId))
-                final poolInfo = job.poolInfo
+    void setJobAutoTermination(String jobId) {
+        log.trace "Eagerly setting Azure Batch job ${jobId} to auto-terminate when all tasks complete"
```
adamrtalbot:
```suggestion
        log.trace "Setting Azure Batch job ${jobId} to auto-terminate when all tasks complete"
```

pditommaso:
@adamrtalbot i've updated the PR description for you 😎

pditommaso:
One thing that i', not getting, if the job is created with the flag TERMINATE_JOB on creation 

[OnAllBatchTasksComplete.TERMINATE_JOB](https://github.com/nextflow-io/nextflow/blob/5839_azure_batch_jobs_terminate_upon_completion/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L461-L466)

why it still needed to update the job with the same flag on completion? 

https://github.com/nextflow-io/nextflow/blob/5839_azure_batch_jobs_terminate_upon_completion/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L1068-L1071

pditommaso:
From my point of view it would be better to keep only the cleanup on jon creation (why it should fail?). 

The update on terminate does not scale well with a large number of tasks

ghislaindemael:
Hi @adamrtalbot 
Sorry for the delay, I was temporarily out of office.
Still needing some testing on the dev branch ?

ghislaindemael:
I'll see what I can do. Problem is that our project doesn't make it easy using a custom build of Nextflow.

ghislaindemael:
Well our use case is just freeing up the quota so we can have more pipelines running, so I guess this solves our problem.

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -568,9 +575,62 @@ class AzBatchService implements Closeable {
...
+     * The new job will also be configured for auto-termination.
+     */
+    protected String recreateJobForTask(String poolId, TaskRun task, String oldJobId, String taskId) {
+        log.debug "Job ${oldJobId} is in completed/terminating state, creating a new job for task ${taskId}"
+        final AzJobKey mapKey = new AzJobKey(task.processor, poolId)
+        allJobIds.remove(mapKey)
+        final String newJobId = createJob0(poolId, task)
+        allJobIds[mapKey] = newJobId
+        return newJobId
+    }
```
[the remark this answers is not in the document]
adamrtalbot: Solved with cc2955d1f7d990b9b3fce3d9ba7b827a1bf551b8

ghislaindemael:
Btw @adamrtalbot we successfully managed to cherry-pick these changes and package them in our modified version of nf-azure@1.11.0. And on execution the jobs are properly terminated once all tasks complete. So our problem is solved until the pipelines are upgraded to work with nextflow's next update

ghislaindemael:
Hi @adamrtalbot,

Sorry for the update, but we hit an edge case where if Nextflow submits a task to a job, but - due to compute limitations as all nodes are used - while this job stays Active, Azure still terminates the job when onAllBatchTasksComplete gets fired. This causes the pipeline to halt without any error thrown.

Our solution is, like you did at the start, to listen with a TraceObserver to onProcessTerminate, fired when all of the tasks are submitted and completed.

```
@Slf4j
class AzTraceObserver implements TraceObserver {

    private Session session;

    AzTraceObserver(Session session) {
        this.session = session;
    }

    /**
     * Fires when the process completes, aka when the last task is completed.
     * Terminates all Azure Batch jobs created by the processor
     */
    @Override
    void onProcessTerminate(TaskProcessor processor) {

        if( !(processor.executor instanceof AzBatchExecutor) ) {
            return
        }

        final executor = processor.executor as AzBatchExecutor
        final batchService = executor.batchService

        if( !batchService?.config?.batch()?.terminateJobsOnCompletion ) {
            log.trace "terminateJobsOnCompletion is disabled, skipping termination for process: ${processor.name}"
            return
        }

        batchService.allJobIds.findAll { key, jobId ->
            key.processor == processor
        }.values().each { jobId ->
            try {
                log.debug "Terminating job ${jobId} for completed process: ${processor.name}"
                batchService.terminateJob(jobId)
            }
            catch( Exception e ) {
                log.warn "Failed to terminate Batch job ${jobId} for process '${processor.name}' - ${e.message ?: e}"
            }
        }
    }

}
```

ghislaindemael:
> Hi @ghislaindemael
> 
> > Sorry for the update, but we hit an edge case where if Nextflow submits a task to a job, but - due to compute limitations as all nodes are used - while this job stays Active, Azure still terminates the job when onAllBatchTasksComplete gets fired. This causes the pipeline to halt without any error thrown.
> 
> I'm not sure I follow this.
> 
> 1. Nextflow creates jobs and adds tasks
> 2. The Azure nodes are occupied to the Azure tasks wait in the queue
> 3. The waiting tasks complete?
> 4. Because of the lag, there is a gap where the job is terminated
> 5. Eventually, new tasks are required to be submitted and Nextflow tries to submit them to the job
> 6. _At this point, Nextflow should identify the Azure Batch job is terminated and create a new one_ <- Would I be correct in saying this silently fails?
> 7. At pipeline termination, Nextflow terminates the remaining jobs, clearing out the quota.
> 
> So the overall objective is achieved, but Nextflow isn't handling the Job re-creation correctly (step 6). Is my understanding correct?

In our case : 
1. Nextflow submits a Job and jobs inside (for example 9 out of the 10 it should) and marks it to be terminated on tasks complete
2. The 9 tasks complete
3. Nextflow submits a 10th task but since all nodes are taken, it stays "Active"
4. The onAllBatchTasksComplete cleaner, that runs every X minutes, looks at the job, ignores the 10th task for an unknown reason, and seeing 9 completed tasks, terminates the job.
5. The TaskStarter (or process that starts tasks) sees that the job is completed so will ignore the fact that there's an active task inside
6. Nextflow stalls as it waits for the 10th task to be ran, which'll never happen.

ghislaindemael:
Update @adamrtalbot 

In my humble opinion, I would disable the auto-termination and just go with the termination with onProcessComplete.

This as it triggers only we are sure the job isn't needed anymore and we can't prevent we might hit a race condition where the job is picked up by onAllBatchTasksComplete cleaner but Nextflow seeing the job as Active still inserts a new task inside before job is set to "Terminating" and the cleaner doesn't recheck.

This leads to an 'orphaned' task, that will not be deleted by Azure until job is deleted. As according to [docs](https://learn.microsoft.com/en-us/rest/api/batchservice/task/add?view=rest-batchservice-2024-07-01&tabs=HTTP#taskconstraints) the maxWallClockTime only applied when tasks starts to run. Which won't happen in this case.

ghislaindemael:
I may allow myself to think the Azure engineers thought about this case, but you can't guarantee some race conditions not to happen on systems of these sizes. 
I want to say it is what we observe, which might not be fully what they intended. But is having a full maxExistenceClockTime field really useful for some extreme edge cases ?

docs/reference/config.md
```diff
@@ -438,7 +438,7 @@ The following settings are available:
 `azure.batch.terminateJobsOnCompletion`
 : :::{versionadded} 23.05.0-edge
   :::
-: When the workflow completes, set all jobs to terminate on task completion (default: `true`).
+: Set all jobs to terminate on task completion (default: `true`).
```
```suggestion
: If `true`, all jobs terminate on task completion (default: `true`).
```
Very minor suggestion. TLDR; we are working towards standardizing language for configuration options. Please check meaning is retained.

bentsherman:
Took some time to try to understand the changes here.

The trace observer makes sense and seems like a straightforward solution. But I'm concerned about the changes to AzBatchService.

If I understand correctly, a task could be submitted to a job that is already set to auto-terminate, so you added some logic to re-create the job and submit the task. But if you only set the job to auto-terminate after the process has terminated (all tasks completed), then this should never happen. So I would not expect this logic to be needed at all.

bentsherman:
> The only time this is a problem is when all tasks within a job complete, the job terminates, and the next task is submitted.

This part I don't understand. If a job is terminated only after all of its tasks have completed, then by definition it should never receive another task

Overall, I think we should focus on #5575 in order to at least stop the bleeding, then perhaps we could add the trace observer. Not sure that the job re-creation is worthwhile

bentsherman:
But why would it do that? Where is task D coming from and why is it being sent to Job A?

Is it a different process that happens to use the same job?

ghislaindemael:
> > This part I don't understand. If a job is terminated only after all of its tasks have completed, then by definition it should never receive another task
> 
> Nextflow creates Job A and adds task A, B and C to it.
> 
> They complete, Job A autocompletes.
> 
> Nextflow then creates task D and tries to add it to Job A.

For this, it is indeed, true, as for some pipelines that behave in an "accordion" manner (Process X generates N tasks and outputs, Process Y regroups N outputs in 1, so spawns only 1 task, etc...), we had some race conditions where for some reason Azure would auto-close the job, even though Nextflow hadn't scheduled all of them.

I guess we aren't the only ones to have this problem as Azure now displays this info for completed jobs. 
Screenshot 2026-03-03 at 22 06 06_heicx com_convert

At least in our case, since our fix to use onProcessTerminate, the error hasn't been seen anymore.

PR #6165, opened 2025-06-05
AWS SDK V2 Porting (without AWS NIO SPI)

jorgee:
This PR contains the changes to port the Amazon plugin to AWS SDK version 2.


jorgee:
It is ready for review

docs/reference/config.md
```diff
@@ -223,14 +223,21 @@ The following settings are available:
 : The maximum number of retry attempts for failed retryable requests (default: `-1`).
 
 `aws.client.protocol`
-: The protocol to use when connecting to AWS. Can be `http` or `https` (default: `'https'`).
+: :::{deprecated} 25.06.0-edge
+  :::
+: The protocol to use when connecting to AWS. Can be `http` or `https` (default: `'https'`). This option is not supported on AWS SDK v2.
```
pditommaso:
```suggestion
: This option is not supported as of version `25.06.0-edge`.
```

docs/reference/config.md
```diff
@@ -249,10 +256,14 @@ The following settings are available:
 : The name of the signature algorithm to use for signing requests made by the client.
 
 `aws.client.socketSendBufferSizeHint`
-: The Size hint (in bytes) for the low level TCP send buffer (default: `0`).
+: :::{deprecated} 25.06.0-edge
+  :::
+: The Size hint (in bytes) for the low level TCP send buffer (default: `0`). This option is not supported in AWS SDK v2.
```
pditommaso:
```suggestion
: This option is no longer supported as of version `25.06.0-edge`.
```

docs/reference/config.md
```diff
@@ -249,10 +256,14 @@ The following settings are available:
...
 
 `aws.client.socketRecvBufferSizeHint`
-: The Size hint (in bytes) for the low level TCP receive buffer (default: `0`).
+: :::{deprecated} 25.06.0-edge
+  :::
+: The Size hint (in bytes) for the low level TCP receive buffer (default: `0`). This option is not supported in AWS SDK v2.
```
pditommaso:
```suggestion
: This option is no longer supported as of version `25.06.0-edge`.
```

plugins/nf-amazon/src/main/nextflow/cloud/aws/AwsClientFactory.groovy
```diff
@@ -171,148 +159,110 @@ class AwsClientFactory {
...
-                .standard()
-                .withRegion(region)
-
-        final credentials = getCredentialsProvider0()
-        if( credentials )
-            builder.withCredentials(credentials)
-
-        return builder.build()
+    BatchClient getBatchClient() {
+        return BatchClient.builder().region(getRegionObj(region)).credentialsProvider(getCredentialsProvider0()).build()
```
pditommaso: nit, splitting on multiple lines can make it a bit more readable

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
```diff
@@ -421,75 +395,80 @@ public void multipartCopyObject(S3Path s3Source, S3Path s3Target, Long objectSiz
...
+		CompletedMultipartUpload completedUpload = CompletedMultipartUpload.builder()
+        .parts(completedParts)
+        .build();
+
+		CompleteMultipartUploadRequest completeRequest = CompleteMultipartUploadRequest.builder()
+        .bucket(targetBucketName)
+        .key(targetObjectKey)
+        .uploadId(uploadId)
+        .multipartUpload(completedUpload)
+        .build();
```
pditommaso: Some indentation can make it better readable

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
```diff
@@ -503,161 +482,111 @@ static CopyPartResult copyPart0(AmazonS3 client, CopyPartRequest request, S3Mult
...
-				Download it = transferManager() .download(source.getBucket(), sourceKey, newFile.toFile());
-				allDownloads.add(it);
-
-				return FileVisitResult.CONTINUE;
-			}
+        FileDownload downloadFile = transferManager().downloadFile(downloadFileRequest);
+        try{
+            downloadFile.completionFuture().get();
+        } catch (InterruptedException e){
+            log.debug("S3 download file: s3://{}/{} cancelled", source.getBucket(), source.getKey());
```
pditommaso:
```suggestion
            log.debug("S3 download file: s3://{}/{} interrupted", source.getBucket(), source.getKey());
```

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
```diff
@@ -503,161 +482,111 @@ static CopyPartResult copyPart0(AmazonS3 client, CopyPartRequest request, S3Mult
...
+			Thread.currentThread().interrupt();
+        } catch (ExecutionException e) {
+            log.debug("S3 download file: s3://{}/{} exception thrown", source.getBucket(), source.getKey());
+            throw new IOException(e.getCause());
```
pditommaso:
```suggestion
            String msg = String.format("Exception thrown with downloading S3 object s3://{}/{}", source.getBucket(), source.getKey());
            throw new IOException(msg, e);
```

Logging the error and re-throwing will result in double logging it, which can be confusing. 

Alternatively it could be done just 

```
throw e.getCause()
```

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
```diff
@@ -503,161 +482,111 @@ static CopyPartResult copyPart0(AmazonS3 client, CopyPartRequest request, S3Mult
...
-			log.debug("S3 download directory: s3://{}/{} interrupted", source.getBucket(), source.getKey());
+		DirectoryDownload downloadDirectory = transferManager().downloadDirectory(downloadDirRequest);
+        try{
+		    downloadDirectory.completionFuture().get();
+        } catch (InterruptedException e){
+            log.debug("S3 download directory: s3://{}/{} cancelled", source.getBucket(), source.getKey());
 			Thread.currentThread().interrupt();
-		}
+        } catch (ExecutionException e) {
+            log.debug("S3 deownload directory: s3://{}/{} exception thrown", source.getBucket(), source.getKey());
```
pditommaso: as above

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
```diff
@@ -503,161 +482,111 @@ static CopyPartResult copyPart0(AmazonS3 client, CopyPartRequest request, S3Mult
...
-	final private ThreadLocal<List<Tag>> uploadTags = new ThreadLocal<>();
+		Upload upload = transferManager().upload(UploadRequest.builder().putObjectRequest(req.build()).requestBody(AsyncRequestBody.fromFile(source)).build());
+        try{
+		    upload.completionFuture().get();
+        } catch (InterruptedException e){
+            log.debug("S3 upload file: s3://{}/{} cancelled", target.getBucket(), target.getKey());
+            Thread.currentThread().interrupt();
+        } catch (ExecutionException e) {
+            log.debug("S3 upload file: s3://{}/{} exception thrown", target.getBucket(), target.getKey());
+            throw new IOException(e.getCause());
```
pditommaso: as above

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
```diff
@@ -503,161 +482,111 @@ static CopyPartResult copyPart0(AmazonS3 client, CopyPartRequest request, S3Mult
...
+            log.debug("S3 upload directory: s3://{}/{} cancelled", target.getBucket(), target.getKey());
 			Thread.currentThread().interrupt();
-		}
-	}
-
-	String getObjectKmsKeyId(String bucketName, String key) {
-		return getObjectMetadata(bucketName,key).getSSEAwsKmsKeyId();
+        } catch (ExecutionException e) {
+            log.debug("S3 upload directory: s3://{}/{} exception thrown", target.getBucket(), target.getKey());
+            throw new IOException(e.getCause());
```
pditommaso: as above

jorgee:
I have found an issue with multi-part uploads when uploading large files. I move to draft until I fix it.

docs/reference/config.md
```diff
@@ -231,6 +234,11 @@ The following settings are available:
 `aws.client.proxyPort`
 : The port to use when connecting through a proxy.
 
+`aws.client.proxyScheme`
```
bentsherman, line 260: Why did you create a new config option for this instead of reusing `aws.client.protocol`? Or are they different things?

bentsherman: If they are different, I think it would be better to call it `proxyProtocol` instead of proxy scheme, unless there is some technical reason

bentsherman: Never mind, I see now it is called "scheme" by the SDK

jorgee: Yes, the `protocol` was an option in v1 which is not available in v2 to select the http or https enpoint of AWS services. Now, http is only valid whe there is a endpoint override. The `proxyScheme` is used to create the proxy endpoint that is required in the proxy configuration for sync clients.

jorgee:
In one of the changes that I did to support the Signer override and UserAgent is breaking the multipart upload. I changed the way about how the transfer manager's S3AsyncClient is created and the multipart upload with large files is having issues. It is creating too many parts and it can exhaust the heap memory or a time out when acquaring the connections. It is not happening if we use the default S3CrtAsyncClient but it does not allow to define the UserAgent and Signer as a clientoverride. I am still looking for a way to define them but I am wondering how relevant are these options. Is there a possibility to deprecate them?

pditommaso:
@jorgee can you make a quick summary about this effort? any blocking issues?

jorgee:
@pditommaso, this is the summary that I breifly explained to @bentsherman, yesterday.

I have been working on these two issues:

- Custom Signer: Java SDK v2 only support sAWS V4 signer either in netty and CRT. This is also supported by Minio and Ceph. I think a custom signer could be requires if a user uses a non up-to-date version of these tools. In Netty, there is the possibility to pass a custom signer, but there aren't other implementations in the SDK than V4. So the user should provide the implementation. In CRT, there is no possibility to override any HTTP advanced options, so neither Signer or user agent can be specified.

The solution that I have pushed in the latest commits is the following. In the case that a custom signer or user agent is specified, Nextflow will configure a Netty async client of the S3 Transfer Manager. A custom signer can be used including the implementation in the classpath and adding the FQDN of the custom signer class in the `signerOverride` option.
In this [issue](https://github.com/aws/aws-sdk-java/issues/2660), we have an old implementation that could be used in case we need to support V2 Signer.

- The multipart upload problems: They were mainly due to using the old default uploadPartSize of 100MB with async clients. The transfer manager threads are not limited by the executor threads like in the past, they are managed by the async client `maxConcurrency` that is much more larger than threads. So, more parts are managed in parallel creating OOM problems when using small instances such as 1-2GB. With the default async multipart block size (8MB), it is not happening. So, I have kept the `uploadPartSize` option and the other `upload*` options for the case that is still using the old programmatic multipart upload (`FileSystemProvider.newOutputStream` case). For the transfers managed by the new S3 transfer manager, I have added new options to configure the async client.

Now, I am running the benchmarks to see if there are differences with the changes, but the code is again ready for review.

pditommaso:
What's CRT? 😄

jorgee:
CRT is the AWS Common Runtime. It is a C implementations of the AWS Client.
https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/crt-based-s3-client.html

pditommaso:
OK, let's wait the result of the benchmark. However my advice is to focus on merging this PR with the core migration ignoring corner cases such Custom signer and user agent problem, and addressing them if needed in a separate PRs. 

Regarding the multipart upload problem, is it related with a specific config or a general one?

bentsherman:
Note that with the custom signer, if we don't support it in this PR then we have to deprecate it as it would no longer do anything

docs/reference/config.md
```diff
@@ -216,13 +216,31 @@ The following settings are available:
   :::
 : The retrieval tier to use when restoring objects from Glacier, one of [`Expedited`, `Standard`, `Bulk`].
 
+`aws.client.maxConcurrency`
+: :::{versionadded} 25.06.0-edge
+  :::
+: Overrides the maximum number of concurrent connections in S3 Async client used in the S3 Transfer Manager. This is automatically set according to `targetThroughputInGbps`. Modifying this value can affect the memory used by the S3 transfers. [More details](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3CrtAsyncClientBuilder.html#maxConcurrency(java.lang.Integer)).
```
```suggestion
: Overrides the maximum number of concurrent connections in S3 Async client used in the S3 Transfer Manager. This is automatically set according to `targetThroughputInGbps`. Modifying this value can affect the memory used by the S3 transfers. See [S3CrtAsyncClientBuilder](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3CrtAsyncClientBuilder.html)e details](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3CrtAsyncClientBuilder.html#maxConcurrency(java.lang.Integer)) for more details.
```

docs/reference/config.md
```diff
@@ -216,13 +216,31 @@ The following settings are available:
...
-: The maximum number of allowed open HTTP connections (default: `50`).
+: The maximum number of allowed open HTTP connections (default: `50`). In AWS SDK v2, different AWS clients are used. This will only affect the synchronous clients. For asynchronous clients used in the S3 Transfer Manager use `aws.client.maxConcurrency`.
 
 `aws.client.maxErrorRetry`
 : The maximum number of retry attempts for failed retryable requests (default: `-1`).
 
+`aws.client.maxNativeMemory`
+: :::{versionadded} 25.06.0-edge
+  :::
+: The maximum native memory used by the S3 asynchronous client for S3 transfers. This is automatically set according to `targetThroughputInGbps`. Modifying this value can affect the memory used by the S3 transfers. [More details](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3CrtAsyncClientBuilder.html#maxNativeMemoryLimitInBytes(java.lang.Long)).
```
```suggestion
: The maximum native memory used by the S3 asynchronous client for S3 transfers. This is automatically set according to `targetThroughputInGbps`. Modifying this value can affect the memory used by the S3 transfers. See [maxNativeMemoryLimitInBytes](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3CrtAsyncClientBuilder.html#maxNativeMemoryLimitInBytes(java.lang.Long)) for more details.
```

docs/reference/config.md
```diff
@@ -216,13 +216,31 @@ The following settings are available:
...
+
+`aws.client.minimumPartSize`
+: :::{versionadded} 25.06.0-edge
+:::
+: The minimum part size used by the S3 async client in multipart upload mode. (default: `8 MB`). [More details](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3CrtAsyncClientBuilder.html#minimumPartSizeInBytes(java.lang.Long)).
```
```suggestion
: The minimum part size used by the S3 async client in multipart upload mode. (default: `8 MB`). See [minimumPartSizeInBytes](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3CrtAsyncClientBuilder.html#minimumPartSizeInBytes(java.lang.Long)) for more details.
```

docs/reference/config.md
```diff
@@ -240,18 +263,25 @@ The following settings are available:
...
-: Use [Rrequester Pays](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RequesterPaysBuckets.html) for S3 buckets (default: `false`).
+: Use [Requester Pays](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RequesterPaysBuckets.html) for S3 buckets (default: `false`).
 
 `aws.client.s3PathStyleAccess`
 : Use the path-based access model to access objects in S3-compatible storage systems (default: `false`).
 
 `aws.client.signerOverride`
-: The name of the signature algorithm to use for signing requests made by the client.
+: The name of the signature algorithm to use for signing requests made by the client. 
+: Since 25.06.0-edge, the nf-amazon plugin is using the SDK V2, this option is deprecated and only AWS V4 Signer is supported. In case a custom signer is required, this option is still usable by providing the custom implementation in the classpath and specifying the custom signer fully qualified class name. The default S3 CRT-based Async client is not supporting custom signers, therefore Nextflow will use the S3 Netty Async client when this option is activated. This could cause performance degradation when transferring large files.
```
```suggestion
: Since 25.06.0-edge, the nf-amazon plugin is using the SDK V2. This option is deprecated and only AWS V4 Signer is supported. If a custom signer is required, this option is still usable by providing the custom implementation in the classpath and specifying the custom signer's fully qualified class name. The default S3 CRT-based Async client does not support custom signers. Nextflow will use the S3 Netty Async client when this option is activated. This could cause performance degradation when transferring large files.
```

docs/reference/config.md
```diff
@@ -240,18 +263,25 @@ The following settings are available:
...
 
 `aws.client.socketSendBufferSizeHint`
+: :::{deprecated} 25.06.0-edge
+  This option is no longer supported.
+  :::
 : The Size hint (in bytes) for the low level TCP send buffer (default: `0`).
 
 `aws.client.socketRecvBufferSizeHint`
+: :::{deprecated} 25.06.0-edge
+  This option is no longer supported.
```
christopher-hakkaart, line 293:
```suggestion
```

docs/reference/config.md
```diff
@@ -240,18 +263,25 @@ The following settings are available:
...
 : Use the path-based access model to access objects in S3-compatible storage systems (default: `false`).
 
 `aws.client.signerOverride`
-: The name of the signature algorithm to use for signing requests made by the client.
+: The name of the signature algorithm to use for signing requests made by the client. 
+: Since 25.06.0-edge, the nf-amazon plugin is using the SDK V2, this option is deprecated and only AWS V4 Signer is supported. In case a custom signer is required, this option is still usable by providing the custom implementation in the classpath and specifying the custom signer fully qualified class name. The default S3 CRT-based Async client is not supporting custom signers, therefore Nextflow will use the S3 Netty Async client when this option is activated. This could cause performance degradation when transferring large files.
 
 `aws.client.socketSendBufferSizeHint`
+: :::{deprecated} 25.06.0-edge
+  This option is no longer supported.
```
christopher-hakkaart, line 287:
```suggestion
```

docs/reference/config.md
```diff
@@ -216,13 +216,31 @@ The following settings are available:
...
+: The maximum native memory used by the S3 asynchronous client for S3 transfers. This is automatically set according to `targetThroughputInGbps`. Modifying this value can affect the memory used by the S3 transfers. [More details](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3CrtAsyncClientBuilder.html#maxNativeMemoryLimitInBytes(java.lang.Long)).
+
+`aws.client.minimumPartSize`
+: :::{versionadded} 25.06.0-edge
+:::
+: The minimum part size used by the S3 async client in multipart upload mode. (default: `8 MB`). [More details](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3CrtAsyncClientBuilder.html#minimumPartSizeInBytes(java.lang.Long)).
+
 `aws.client.protocol`
+: :::{deprecated} 25.06.0-edge
+  This option is no longer supported.
```
christopher-hakkaart, line 250:
```suggestion
```

docs/reference/config.md
```diff
@@ -265,20 +295,36 @@ The following settings are available:
   :::
 : The AWS KMS key Id to be used to encrypt files stored in the target S3 bucket.
 
+`aws.client.targetThroughputInGbps`
+: :::{versionadded} 25.06.0-edge
+  :::
+: The S3 Async client target network throughput in Gbps. This value is used to automatically set `maxConcurrency` and `maxNativeMemory`. (Default: `10`). [More details](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3CrtAsyncClientBuilder.html#targetThroughputInGbps(java.lang.Double)).
```
```suggestion
: The S3 Async client target network throughput in Gbps. This value is used to automatically set `maxConcurrency` and `maxNativeMemory`. (Default: `10`). See [targetThroughputInGbps](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3CrtAsyncClientBuilder.html#targetThroughputInGbps(java.lang.Double)) for more details.
```

docs/reference/config.md
jorgee: I have modified the comment in latest commit

jorgee:
I have pushed the changes to use only CRT version deprecating `signerOverride` and `userAgent` options.

docs/reference/config.md
bentsherman: I feel like we have to explicitly say this because otherwise it looks like it's merely "deprecated", which implies that it still works, but in reality it has been removed.

docs/reference/config.md
```diff
@@ -265,20 +295,38 @@ The following settings are available:
...
+
 `aws.client.userAgent`
+:::{deprecated} 25.06.0-edge
+  This option is no longer supported.
+  :::
 : The HTTP user agent header passed with all HTTP requests.
 
 `aws.client.uploadChunkSize`
-: The size of a single part in a multipart upload (default: `100 MB`).
+: The size of a single part in a multipart upload (default: `100 MB`). Since version 25.06.0-edge, this option only applies to uploads generated from OutputStream. Other uploads are managed by the S3 transfer manager.
```
bentsherman: Is it worth keeping these options if they are only applied to OutputStream's ? We try not to expose this interface directly to users, in favor of `Path`, so I would be inclined to remove these options.

plugins/nf-amazon/src/main/nextflow/cloud/aws/config/AwsS3Legacy.groovy
```diff
@@ -45,6 +45,9 @@ class AwsS3Legacy {
     static protected Map normalizeAwsClientConfig(Map<String,?> client) {
 
         normalizeMemUnit(client, 'uploadChunkSize');
+        normalizeMemUnit(client, 'minimumPartSize');
+        normalizeMemUnit(client, 'multipartThreshold');
+        normalizeMemUnit(client, 'maxNativeMemory');
```
bentsherman, line 50: Why are these new settings being handled in the legacy config? They should be handled separately, likely `AwsS3Config`

pditommaso: it should be considered to remove the [s3Legacy](https://github.com/nextflow-io/nextflow/blob/aws-sdk-v2-fs-impl/plugins/nf-amazon/src/main/nextflow/cloud/aws/config/AwsConfig.groovy#L50-L50) settings

docs/reference/config.md
```diff
@@ -216,13 +216,36 @@ The following settings are available:
   :::
 : The retrieval tier to use when restoring objects from Glacier, one of [`Expedited`, `Standard`, `Bulk`].
 
+`aws.client.maxConcurrency`
+: :::{versionadded} 25.06.0-edge
+  :::
+: The maximum number of concurrent S3 transfers when using the asynchronous S3 client. By default, this setting is determined by `aws.client.targetThroughputInGbps`. Modifying this value can affect the amount of memory used for S3 transfers.
```
pditommaso: As a nextflow user I have no idea what's the asynchronous S3. That's not something should be exposed in the documentation. If I have understand correctly this is related to the CRT vs Netty runtime. Is it correct?

docs/reference/config.md
jorgee: Do you mean removing from the documentation the `aws.client.upload*` options?

docs/reference/config.md
pditommaso: We talked. Better to document these are using used by the data transfer component.

plugins/nf-amazon/src/main/nextflow/cloud/aws/config/AwsS3Legacy.groovy
jorgee: If I am not wrong, In nextflow AWS configuration can be obtained from a properties file in the classpath ([code](https://github.com/nextflow-io/nextflow/blob/26da281d91af0d9072aeb21d0390e199d80c3bbb/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3FileSystemProvider.java#L730)) and s3legacy is used to override them by the values set in the Nextflow config. Removing s3Legacy would also imply removing the properties file and changing the code where it is used (mainly` S3ClientConfiguration` classes and `S3FileSystemProvider.createFilesystem(`)). Not a big change, but just to let you know.

docs/reference/config.md
bentsherman: I don't mind mentioning the async S3 client if there is also an option for selecting the async / sync client. But if we can make it so that Nextflow only uses one then we don't need to mention it in the docs

plugins/nf-amazon/src/main/nextflow/cloud/aws/config/AwsS3Legacy.groovy
bentsherman: I think I would be fine with removing the legacy s3 settings (the `upload*` settings) and using the AWS defaults (or providing some sensible fixed defaults). It just doesn't make sense to expose these settings anymore to the user when there are also new settings like `maxConcurrency` and `maxNativeMemory`

pditommaso: Tend to agree, but let's keep into next iteration

jorgee:
> It's also not clear to me how to select the async client vs sync client, as I don't see any explicit config option for this

The sync client is used in all api call except the S3 transfer manager actions ( S3 copies, uploads and downloads). For transfer manager there is no option to use the sync client. I didn't changed all calls to async because it implied a lot of changes in the code

jorgee:
I have added a subsection in the aws part of the documentation mentioning the most important changes in the v1 to v2 migration

plugins/nf-amazon/src/main/nextflow/cloud/aws/config/AwsS3Legacy.groovy
bentsherman: In that case let's go ahead and mark these options as "no longer supported" in the docs so as to not confuse users

jorgee: Then, I do not need to change the code in this part, just documentation?

bentsherman: Yes

jorgee: Done

docs/reference/config.md
jorgee: Marked as deprecated as commented in other review comment

docs/reference/config.md
jorgee: I changed the text to refer to S3 transfer manager client. I have also added [this section](https://github.com/nextflow-io/nextflow/blob/90a1457a55bf619259b8d47309c96b64d1025ae8/docs/aws.md?plain=1#L554) in the AWS part, to document when sync and async client is used.

pditommaso:
End to end tests fail badly. I've got this message 

```
ERROR ~ Type com.amazonaws.services.batch.model.KeyValuePair not present
```

unfortunately I'm not able to retried the full log

jorgee:
> End to end tests fail badly. I've got this message
> 
> ```
> ERROR ~ Type com.amazonaws.services.batch.model.KeyValuePair not present
> ```
> 
> unfortunately I'm not able to retried the full log

The package is from v1. I am going to check if there is some place where I forgot someplace where the package is used

jorgee:
I have realized the gradle tasks implementation inside nexflow that still uses v1. I am going to change it, but I think it shouldn't affect end-to-end tests.

pditommaso:
Yep, should be unrelated

pditommaso:
I've realised some method like `AwsBatchTaskHandle.configJobDefRequest` were returning a Builder object, that in SDK v2 is an immutable object, making hard for sub-classes (like nf-xpack) to extend the handler behaviour.

To address this, I've introduced a custom class to model the Job definition request https://github.com/nextflow-io/nextflow/blob/fd86ab41314ce1ec951f8e8cec77fddbe8e863c2/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/model/RegisterJobDefinitionModel.groovy#L34-L34. This allows to better bridge the new builder API model with the existing behaviour and minimise the changes.

pditommaso:
Merging this seems `com.amazonaws.services.batch.model.KeyValuePair` seems related to the xpack that requires to be updated. 

@jorgee let's discuss the plan for supporting the option `signerOverride` if needed.

pditommaso:
I saw an integration test into [another PR](https://github.com/nextflow-io/nextflow/actions/runs/16097538539/job/45422506455) failing with this error 

```
Jul-06 09:44:25.772 [main] DEBUG nextflow.Session - Session aborted -- Cause: The request content has fewer bytes than the specified content-length: 68 bytes.
Jul-06 09:44:25.793 [main] ERROR nextflow.cli.Launcher - @unknown
java.lang.IllegalStateException: The request content has fewer bytes than the specified content-length: 68 bytes.
	at software.amazon.awssdk.core.internal.io.SdkLengthAwareInputStream.read(SdkLengthAwareInputStream.java:82)
	at software.amazon.awssdk.core.io.SdkFilterInputStream.read(SdkFilterInputStream.java:66)
	at software.amazon.awssdk.http.auth.aws.internal.signer.io.ChecksumInputStream.read(ChecksumInputStream.java:52)
	at software.amazon.awssdk.http.auth.aws.internal.signer.chunkedencoding.ChunkedEncodedInputStream.read(ChunkedEncodedInputStream.java:136)
	at software.amazon.awssdk.http.auth.aws.internal.signer.chunkedencoding.ChunkedEncodedInputStream.getChunk(ChunkedEncodedInputStream.java:112)
	at software.amazon.awssdk.http.auth.aws.internal.signer.chunkedencoding.ChunkedEncodedInputStream.currentChunk(ChunkedEncodedInputStream.java:95)
	at software.amazon.awssdk.http.auth.aws.internal.signer.chunkedencoding.ChunkedEncodedInputStream.read(ChunkedEncodedInputStream.java:90)
	at java.base/java.io.FilterInputStream.read(FilterInputStream.java:132)
	at software.amazon.awssdk.core.internal.io.SdkLengthAwareInputStream.read(SdkLengthAwareInputStream.java:75)
	at org.apache.http.entity.InputStreamEntity.writeTo(InputStreamEntity.java:140)
	at software.amazon.awssdk.http.apache.internal.RepeatableInputStreamRequestEntity.writeTo(RepeatableInputStreamRequestEntity.java:157)
	at org.apache.http.impl.execchain.RequestEntityProxy.writeTo(RequestEntityProxy.java:121)
	at org.apache.http.impl.DefaultBHttpClientConnection.sendRequestEntity(DefaultBHttpClientConnection.java:156)
	at org.apache.http.impl.conn.CPoolProxy.sendRequestEntity(CPoolProxy.java:152)
	at org.apache.http.protocol.HttpRequestExecutor.doSendRequest(HttpRequestExecutor.java:238)
	at org.apache.http.protocol.HttpRequestExecutor.execute(HttpRequestExecutor.java:123)
	at org.apache.http.impl.execchain.MainClientExec.execute(MainClientExec.java:272)
	at org.apache.http.impl.execchain.ProtocolExec.execute(ProtocolExec.java:186)
	at org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:185)
	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:83)
	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:56)
	at software.amazon.awssdk.http.apache.internal.impl.ApacheSdkHttpClient.execute(ApacheSdkHttpClient.java:72)
	at software.amazon.awssdk.http.apache.ApacheHttpClient.execute(ApacheHttpClient.java:259)
	at software.amazon.awssdk.http.apache.ApacheHttpClient.access$600(ApacheHttpClient.java:104)
	at software.amazon.awssdk.http.apache.ApacheHttpClient$1.call(ApacheHttpClient.java:236)
	at software.amazon.awssdk.http.apache.ApacheHttpClient$1.call(ApacheHttpClient.java:233)
	at software.amazon.awssdk.core.internal.util.MetricUtils.measureDurationUnsafe(MetricUtils.java:102)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.MakeHttpRequestStage.executeHttpRequest(MakeHttpRequestStage.java:88)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.MakeHttpRequestStage.execute(MakeHttpRequestStage.java:64)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.MakeHttpRequestStage.execute(MakeHttpRequestStage.java:46)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptTimeoutTrackingStage.execute(ApiCallAttemptTimeoutTrackingStage.java:74)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptTimeoutTrackingStage.execute(ApiCallAttemptTimeoutTrackingStage.java:43)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.TimeoutExceptionHandlingStage.execute(TimeoutExceptionHandlingStage.java:79)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.TimeoutExceptionHandlingStage.execute(TimeoutExceptionHandlingStage.java:41)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptMetricCollectionStage.execute(ApiCallAttemptMetricCollectionStage.java:55)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptMetricCollectionStage.execute(ApiCallAttemptMetricCollectionStage.java:39)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.executeRequest(RetryableStage.java:93)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:56)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:36)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:53)
	at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:35)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.executeWithTimer(ApiCallTimeoutTrackingStage.java:82)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:62)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:43)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallMetricCollectionStage.execute(ApiCallMetricCollectionStage.java:50)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallMetricCollectionStage.execute(ApiCallMetricCollectionStage.java:32)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ExecutionFailureExceptionReportingStage.execute(ExecutionFailureExceptionReportingStage.java:37)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ExecutionFailureExceptionReportingStage.execute(ExecutionFailureExceptionReportingStage.java:26)
	at software.amazon.awssdk.core.internal.http.AmazonSyncHttpClient$RequestExecutionBuilderImpl.execute(AmazonSyncHttpClient.java:210)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.invoke(BaseSyncClientHandler.java:103)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.doExecute(BaseSyncClientHandler.java:173)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.lambda$execute$1(BaseSyncClientHandler.java:80)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.measureApiCallSuccess(BaseSyncClientHandler.java:182)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.execute(BaseSyncClientHandler.java:74)
	at software.amazon.awssdk.core.client.handler.SdkSyncClientHandler.execute(SdkSyncClientHandler.java:45)
	at software.amazon.awssdk.awscore.client.handler.AwsSyncClientHandler.execute(AwsSyncClientHandler.java:53)
	at software.amazon.awssdk.services.s3.DefaultS3Client.putObject(DefaultS3Client.java:11228)
	at software.amazon.awssdk.services.s3.DelegatingS3Client.lambda$putObject$86(DelegatingS3Client.java:9145)
	at software.amazon.awssdk.services.s3.internal.crossregion.S3CrossRegionSyncClient.invokeOperation(S3CrossRegionSyncClient.java:67)
	at software.amazon.awssdk.services.s3.DelegatingS3Client.putObject(DelegatingS3Client.java:9145)
	at nextflow.cloud.aws.nio.S3OutputStream.putObject(S3OutputStream.java:626)
	at nextflow.cloud.aws.nio.S3OutputStream.putObject(S3OutputStream.java:581)
	at nextflow.cloud.aws.nio.S3OutputStream.close(S3OutputStream.java:384)
	at java.base/java.io.FilterOutputStream.close(FilterOutputStream.java:188)
	at nextflow.extension.FilesEx.closeQuietly(FilesEx.groovy:689)
	at nextflow.cache.CloudCacheStore.close(CloudCacheStore.groovy:93)
	at nextflow.cache.CacheDB.close(CacheDB.groovy:266)
	at nextflow.Session.destroy(Session.groovy:724)
	at nextflow.script.ScriptRunner.shutdown(ScriptRunner.groovy:263)
	at nextflow.script.ScriptRunner.execute(ScriptRunner.groovy:147)
	at nextflow.cli.CmdRun.run(CmdRun.groovy:379)
	at nextflow.cli.Launcher.run(Launcher.groovy:513)
	at nextflow.cli.Launcher.main(Launcher.groovy:673)
```


Worth double checking

jorgee:
It could be caused because in v2 content-length is now part of the request instead of metadata. The strange part is that it is not always happening. I am going to try to find when it is reproduced.

PR #5484, opened 2024-11-08
[skip CI] [do not merge] Add 'cycle' operator

adamrtalbot:
This is a simple operator that loops over an range and applies it as an index to each channel.

I don't intend this to be merged, it's a little demo to remind myself how it works

Signed-off-by: adamrtalbot <12817534+adamrtalbot@users.noreply.github.com>

Hi! Thanks for contributing to Nextflow project.

When submitting a Pull Request please make sure to not include
in the changeset any modification in these files:

* `nextflow`
* `modules/nf-commons/src/main/nextflow/Const.groovy`

Also, please sign-off the DCO [1] to certify you are the author of the contribution
and you adhere to Nextflow open source license [2] adding a `Signed-off-by` line to
the contribution commit message. For more details check [3].

1. https://developercertificate.org/
2. https://github.com/nextflow-io/nextflow/blob/master/COPYING
3. https://github.com/apps/dco

PR #6166, opened 2025-06-05
Remove path hash in the conda environment

### LINKED ISSUE #6140: conda environments recreated on every run with Seqera Platform (-with-tower)

When using the conda directive for a process to point to a environment yaml file, the environment gets re-created in the cache directory on every run when using Seqera Platform.
We would expect that nextflow would use the existing conda environment in the cache directory.

This change in behaviour is due to the conda cache changes here: https://github.com/nextflow-io/nextflow/pull/5489

1. create a nextflow pipeline with the following process and environment file
```
process foo {
  conda '/some/path/my-env.yaml'

  script:
  """
  your_command --here
  """
}
```

```yml
name: my-env
channels:
  - conda-forge
  - bioconda
dependencies:
  - star=2.5.4a
  - bwa=0.7.15
```
2. Import this pipeline into Seqera Platform  - https://docs.seqera.io/platform-cloud/getting-started/quickstart-demo/add-pipelines
3. Configure the `nextflow.config' of pipeline to:
    * enable conda `conda.enabled = true`
    * set the `conda.cacheDir = /some/path/on/the/system`
4. Launch the pipeline via Seqera (in my case I am using a SLURM compute environment)
    * Seqera will:
      * set `NXF_ASSETS` to a location in the work directory `<basepath>/work/.nextflow/pipelines/<pipeline-id>`
      * launch nextflow on the server with `nextflow run <pipeline> ... -with-tower <seqera-url>`

### Program output 

```console
Monitor the execution with Seqera Platform using this URL: <seqera-url>
Creating env using conda: /home/user/<username>/nextflow/work/.nextflow/pipelines/f1f16f92/my-pipeline/envs/environment.yml [cache /home/user/<username>/nextflow/conda_cache/env-fe50773488a5e0e5-535eefbee5734a58afed514fbe57635c]
```

### Environment 

* Nextflow version: 24.10.5
* Java version: 17.0.14
* Operating system: Linux
* Bash version:  5.1.8

### Additional context

My troubleshooting has lead me to understand that Seqera Platform generates a new directory for nextflow assets for every pipeline launch. The new directory is placed in the `work/.nextflow/pipelines` directory. This directory has a new name for every pipeline launch.
In my case:
* `/home/user/<username>/nextflow/work/.nextflow/pipelines/f1f16f92/` - Run 1
* `/home/user/<username>/nextflow/work/.nextflow/pipelines/ee5ee5c/` - Run 2

This is problematic because the `CondaCache` class computes the cached environment directory based on a hash of the full path of the environment.yml file:

https://github.com/nextflow-io/nextflow/blob/a5b7a5d2fc146b6f73893b53413dc9b132ae299b/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy#L199-L203

* `/home/user/<username>/nextflow/work/.nextflow/pipelines/f1f16f92/my-pipeline/envs/environment.yml` becomes `env-fe50773488a5e0e5-535eefbee5734a58afed514fbe57635c`

Essentially there is no way for two pipeline runs in Seqera to use a cached conda environment because they will never have the same name.
I see this as an incompatibility between the way Seqera Platform uses nextflow, so I'm not 100% sure where the fix should lie.

A current workaround may be to specify conda requirements within a string (instead of file):
```
process foo {
  conda 'star=2.5.4a bwa=0.7.15'
  ...
}
```

An alternative way (within nextflow) may be to make the generation of the conda environment hash more configurable: (ie. only take path from project root).

jorgee:
close #6140

#5484 reported an `OverlappingFileLockException` when a workflow with two different conda environment files contained the same content. The file path was used as key in the cache map but the environment folder pointed by the keys was the same because of the hash of the content. It was producing the mentioned error. To fix it, #5489 changed the key and environment folder, including the hash of the path, but as side effect, it introduced the problem reported in #6140.

However, the path hash is not needed if we first calculate the content hash and use it as key of the cache map. This PR introduces this change.

Tested with the following  workflow, that produced the error in 24.10. 

```
workflow {
  Channel.from("world") | greet | view
  Channel.from("world") | greet2 | view
}

process greet {
  conda "${baseDir}/conda_1/env.yaml"
  input: val name
  output: stdout

  """
  echo hello $name
  """
}

process greet2 {
  conda "${baseDir}/conda_2/env.yaml"
  input: val name
  output: stdout

  """
  echo hello2 $name
  """
}
```
both `env.yaml` files had the following content

```
channels:
  - conda-forge
dependencies:
  - coreutils
```
and the following `netxflow.config`

```
 conda {
  enabled = true
  cacheDir = '/tmp/shared_conda_cache'
}
```
I have installed in different folders as indicated in the issue and the environment was cached.

### DIFF
--- a/docs/aws.md
+++ b/docs/aws.md
@@ -348,14 +348,13 @@ The grandparent directory of the `aws` tool will be mounted into the container a
 
 ### Docker installation
 
-Docker is required by Nextflow to execute tasks on AWS Batch. The **Amazon ECS-Optimized Amazon Linux 2 (AL2) x86_64 AMI** has Docker installed, however, if you create your AMI from a different AMI that does not have Docker installed, you will need to install it manually.
+Docker is required by Nextflow to execute tasks on AWS Batch. The **Amazon ECS-optimized Amazon Linux 2023 AMI** has Docker installed, however, if you create your AMI from a different AMI that does not have Docker installed, you will need to install it manually.
 
 The following snippet shows how to install Docker on an Amazon EC2 instance:
 
 ```bash
 # install Docker
 sudo yum update -y
-sudo amazon-linux-extras install docker
 sudo yum install docker
 
 # start the Docker service
@@ -365,7 +364,7 @@ sudo service docker start
 sudo usermod -a -G docker ec2-user
 ```
 
-You may have to reboot your instance for the changes to `ec2-user` to take effect.
+You must logging out and logging back in again to use the new `ec2-user` permissions.
 
 These steps must be done *before* creating the AMI from the current EC2 instance.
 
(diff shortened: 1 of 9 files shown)


jorgee:
Fix for https://github.com/nextflow-io/nextflow/issues/6140#issuecomment-2944456043 included.

pditommaso:
@jorgee does this solve also https://github.com/nextflow-io/nextflow/issues/6186?  can you please verify it?

jorgee:
Yes, it is solving #6186.

pditommaso:
Is it ready to go on your side?

jorgee:
Yes, it is ready

PR #6176, opened 2025-06-06
Add namespaces to nf-lang

bentsherman:
This PR formalizes the concept of "namespaces" in nf-lang to align with the current [docs](https://nextflow.io/docs/latest/reference/stdlib-namespaces.html). This allows namespaces to be treated separately in the language server

Also:
- Document `log` namespace
- Improve related channel docs


docs/process.md
```diff
@@ -776,9 +776,9 @@ The process `echo` is executed two times because the `x` channel emits only two
 2 and b
 ```
 
-A different semantic is applied when using a {ref}`value channel <channel-type-value>`. This kind of channel is created by the {ref}`channel.value <channel-value>` factory method or implicitly when a process is invoked with an argument that is not a channel. By definition, a value channel is bound to a single value and it can be read an unlimited number of times without consuming its content. Therefore, when mixing a value channel with one or more (queue) channels, it does not affect the process termination because the underlying value is applied repeatedly.
+When a {ref}`alue channel <channel-type-value>` is supplied as a process input alongside a queue channel, the process is executed for each value in the queue channel, and the value channel is re-used for each execution.
```
```suggestion
When a {ref}`value channel <channel-type-value>` is supplied as a process input alongside a queue channel, the process is executed for each value in the queue channel, and the value channel is re-used for each execution.
```

docs/channel.md
```diff
@@ -2,81 +2,64 @@
 
 # Channels
 
-Nextflow is based on the dataflow programming model in which processes communicate through channels.
+In Nextflow, **channels** are the key data structure used to facilitate the dataflow dependencies between each step (i.e. {ref}`process <process-page>`) in a pipeline.
```
christopher-hakkaart:
```suggestion
In Nextflow, **channels** are the key data structures that facilitate the dataflow dependencies between each step (i.e. {ref}`process <process-page>`) in a pipeline.
```

docs/channel.md
```diff
@@ -2,81 +2,64 @@
...
+(channel-type-queue)=
 
-## Channel types
+## Queue channels
 
-In Nextflow there are two kinds of channels: *queue channels* and *value channels*.
+A *queue channel* is a channel that *emits* an asynchronous sequence of values.
 
-(channel-type-queue)=
+A queue channel can be created by channel factories ({ref}`channel.of <channel-of>`, {ref}`channel.fromPath <channel-path>`, etc), operators ({ref}`operator-map`, {ref}`operator-filter`, etc), and processes (see {ref}`Process outputs <process-output>`).
```
christopher-hakkaart:
```suggestion
A queue channel can be created by channel factories (e.g., {ref}`channel.of <channel-of>` and {ref}`channel.fromPath <channel-path>`), operators (e.g., {ref}`operator-map` and  {ref}`operator-filter`), and processes (see {ref}`Process outputs <process-output>`).
```

christopher-hakkaart: Ect is unclear because it's not established which channel factories can create queue channels, so ect could mean all of them.

docs/channel.md
```diff
@@ -2,81 +2,64 @@
...
 
-A *value channel* can be bound (i.e. assigned) with one and only one value, and can be consumed any number of times by
-a process or an operator.
+A *value channel* is a channel that is *bound* to an asynchronous value.
 
-A value channel can be created with the {ref}`channel-value` factory method or by any operator that produces a single value
-({ref}`operator-first`, {ref}`operator-collect`, {ref}`operator-reduce`, etc). Additionally, a process will emit value
-channels if it is invoked with all value channels, including simple values which are implicitly wrapped in a value channel.
+A value channel can be created with the {ref}`channel.value <channel-value>` factory, certain operators
+({ref}`operator-collect`, {ref}`operator-reduce`, etc), and processes (under {ref}`certain conditions <process-out-singleton>`).
```
christopher-hakkaart:
```suggestion
(e.g., {ref}`operator-collect` and {ref}`operator-reduce`), and processes (under {ref}`certain conditions <process-out-singleton>`).
```

docs/process.md
```diff
@@ -776,9 +776,9 @@ The process `echo` is executed two times because the `x` channel emits only two
...
 
-To better understand this behavior, compare the previous example with the following one:
+For example, compare the previous example with the following one:
```
```suggestion
For example, compare the previous example with the following:
```

docs/process.md
```diff
@@ -1160,6 +1160,22 @@ In this example, the process is normally expected to produce an `output.txt` fil
...
+
+### Singleton outputs
+
+When a process is supplied with only value channels or regular values as inputs, or no inputs, it returns any outputs as value channels. For example:
+
+```{literalinclude} snippets/process-out-singleton.nf
+:language: nextflow
+```
+
+In the above example, the `echo` process is invoked with a regular value, which is wrapped in a value channel. As a result, `echo` returns a value channel and `greet` is executed three times.
```
```suggestion
In the above example, the `echo` process is invoked with a regular value that is wrapped in a value channel. As a result, `echo` returns a value channel and `greet` is executed three times.
```

docs/process.md
```diff
@@ -1160,6 +1160,22 @@ In this example, the process is normally expected to produce an `output.txt` fil
...
+
+If the call to `echo` were changed to `echo( channel.of('hello') )`, the process would instaed return a queue channel and `greet` would be executed only once.
```
```suggestion
If the call to `echo` was changed to `echo( channel.of('hello') )`, the process would instead return a queue channel, and `greet` would be executed only once.
```

docs/process.md
```diff
@@ -1160,6 +1160,22 @@ In this example, the process is normally expected to produce an `output.txt` fil
 While this option can be used with any process output, it cannot be applied to individual elements of a [tuple](#output-tuples-tuple) output. The entire tuple must be optional or not optional.
 :::
 
+(process-out-singleton)=
+
+### Singleton outputs
+
+When a process is supplied with only value channels or regular values as inputs, or no inputs, it returns any outputs as value channels. For example:
```
```suggestion
When a process is only supplied with value channels, regular values, or no inputs, it returns outputs as value channels. For example:
```

christopher-hakkaart: Please check that the meaning is retained.

docs/channel.md
bentsherman: Agreed, thanks

PR #6178, opened 2025-06-06
Allow docker registry to override fully-qualified container images

### LINKED ISSUE #6150: docker.registry needs to override process.container when a registry is used in the container URI
When you define your Nextflow process with a `process.container` that includes a container registry in the URI, Nextflow ignores the `docker.registry` setting and uses the registry listed in the `process.container`.

Example;

https://github.com/nf-core/scdownstream/blob/1512d6ff5e3f2484c36b1bd1d53a0f435a7b3ae4/modules/nf-core/cellbender/merge/main.nf#L8

```groovy
process CELLBENDER_MERGE {
    tag "$meta.id"
    label 'process_single'

    conda "${moduleDir}/environment.yml"
    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
        'oras://community.wave.seqera.io/library/cellbender:0.3.0--c4addb97ab2d83fe':
        'community.wave.seqera.io/library/cellbender:0.3.0--41318a055fc3aacb' }"


```

Here the `process.container` is `community.wave.seqera.io/library/cellbender:0.3.0--41318a055fc3aacb` where `community.wave.seqera.io` is the name of the Docker container registry.

However, when we set `docker.registry = 12345678900.dkr.ecr.us-east-1.aws.com` etc to point to our internal AWS ECR registry, expecting that *all* containers will be pulled from our internal registry, and that a missing container will error-out (prompting us to review and import the container to our ECR).

This is important to meet Infosec compliance requirements in environments that require all containers to be internally hosted.

Similarly, this impacted our CI / CD here as well https://github.com/seqeralabs/nf-canary/issues/32

I was aware that nf-core was moving towards usage of Wave containers, which we had prepared for with our ECR import methodologies, however, I was surprised to see that now nf-core pipelines appear to be using this syntax of hard-coded container registry in the pipeline which breaks the usage of `docker.registry` which we relied on to ensure we are getting our internal version of the container. @ewels 

It seems like the easiest solution would be for Nextflow to allow `docker.registry` to actually control the Docker registry used, and not get ignored when `process.container` includes a registry in the URI. If needed I can post some code snippets that show some methods used to programmatically parse the container URI to detect if a registry is used in the URI name (and then you would be able to strip it easier from the container used)

The inability to control where your container is getting pulled from due to `docker.registry` getting ignored seems like a behavioral oversight of some sort, maybe. What do you think? Thanks.

### Environment 

* Nextflow version: 24.10.4
* Operating system: Linux AWS

bentsherman:
Close #6150


pditommaso (commented):
What's the rationale having a separate list of registries? I'm more inclined having this as a flag and use it along with `docker.registry` when specified to override fully qualified container names, that's the core requirement.

bentsherman:
A simple flag would be nice but I'm not sure that you could reliably separate the registry part from the container name part without a list of known registries

However, we could just have a flag that fails the run if any fully-qualified container is specified, instead of trying to override. This is because their main requirement is ensuring that Nextflow doesn't try to use any containers outside of `docker.registry`

pditommaso:
I think it would be better to not overload too much the settings for the container registry, and just provide  a flag to extend the use of `docker.registry` to the case of fully qualified container names. 

I've provided an implementation in this PR https://github.com/nextflow-io/nextflow/pull/6205.

docs/reference/config.md
```diff
@@ -592,7 +592,30 @@ The following settings are available:
...
+: A list of registries to override when specifying a custom registry with `docker.registry`. For example:
+
+  ```groovy
+  docker.registryOverrides = [
+    'community.wave.seqera.io',
+    'quay.io'
+  ]
+  ```
+
+: Nextflow will use this list to modify fully-qualified container images to use the registry specified by `docker.registry`.
```
christopher-hakkaart, line 610:
```suggestion
: Nextflow will use this list to modify fully-qualified container images and use the registry specified by `docker.registry`.
```
Please check meaning is retained

docs/reference/config.md
```diff
@@ -592,7 +592,30 @@ The following settings are available:
...
+
+: :::{tip}
+  Use `nextflow inspect` to obtain the list of all containers used by a pipeline, from which you can determine the set of registries that need to be overridden.
```
christopher-hakkaart, line 613:
```suggestion
  To determine the set of registries that need to be overridden, use `nextflow inspect` to obtain the list of all containers used by a pipeline.
```

docs/reference/config.md
```diff
@@ -592,7 +592,30 @@ The following settings are available:
...
+  :::
+
+: :::{warning}
+  The pipeline will fail if a container image cannot be overridden when using this setting, in order to prevent Nextflow from pulling containers from outside the custom registry.
```
christopher-hakkaart, line 617:
```suggestion
  To prevent pulling containers from outside the custom registry, Nextflow will fail the pipeline if it cannot override a container image when using this setting.
```
Please check the meaning is retained.

pditommaso:
Closing in favour of #6205

PR #6205, opened 2025-06-20
Add container registryOverride option

pditommaso:
This PR adds an container level config that allows override the registry name then when container name specify a fully qualified repository name, 

for example: 

```
process.container = "quay.io/my/container:latest: 
```

and having in the config 

```
docker.registry = "community.wave.seqera.io" 
docker.registryOverride = true
```

the final resolved container used by the task execution is going to be 

```
community.wave.seqera.io/my/container:latest
```


pditommaso:
It could also be explored if it could make sense to have a more compact notation like: 

```
docker.registry "custom.repo.io", override: true
```

It should be checked if this is going to work when Wave is enabled.

bentsherman:
@stevekm would this work for you? See the changes to `ContainerHandlerTest` to understand how container images would be rewritten

ewels:
@bentsherman the additional levels of URI hierarchy that I mentioned is with GitLab, see [their docs](https://docs.gitlab.com/user/packages/container_registry/#use-container-images-from-the-container-registry):

```
registry.example.com/group/project/image:tag
```

`/project/` represents one extra layer compared to normal - I have a feeling it can also have more layers, with sub-projects.

There may be others as well. Not sure if it's relevant to this PR or not (I think not?), but thought I should flag it.

pditommaso:
Add some docs. Moving this option forward.

PR #6211, opened 2025-06-23
Sanitize labels in AWS batch

edmundmiller:
This should prevent errors such as https://github.com/nf-core/hic/issues/224.

I think this goes with the Nextflow ethos to keep the user from thinking about the nuances of different executors.

But I could also see this not being "expected".


plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
```diff
@@ -849,6 +849,70 @@ class AwsBatchTaskHandler extends TaskHandler implements BatchHandler<String,Job
...
+            final value = entry.getValue()
+
+            if (key != null && value != null) {
+                final sanitizedKey = sanitizeAwsBatchLabel(key.toString(), 128)
+                final sanitizedValue = sanitizeAwsBatchLabel(value.toString(), 256)
+
+                // Only add non-empty keys and values
+                if (sanitizedKey && sanitizedValue) {
+                    result.put(sanitizedKey, sanitizedValue)
+                }
```
adamrtalbot: This will silently drop keys and values that can't be sanitised, which is a bit bad in terms of UX.

At a minimum, I would raise a warning here, but I might be tempted to throw an error and tell the user to specify a valid label.

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
```diff
@@ -849,6 +849,70 @@ class AwsBatchTaskHandler extends TaskHandler implements BatchHandler<String,Job
...
+     * @param input The input string to sanitize
+     * @param maxLength The maximum allowed length
+     * @return The sanitized string
+     */
+    protected String sanitizeAwsBatchLabel(String input, int maxLength) {
+        if (!input) return input
+
+        // Replace invalid characters with underscores
+        // AWS Batch allows: letters, numbers, spaces, and: _ . : / = + - @
+        String sanitized = input.replaceAll(/[^a-zA-Z0-9\s_.\:\/=+\-@]/, '_')
```
adamrtalbot: Untested, but I think you can simplify this to:
```suggestion
        String sanitized = input.replaceAll(/[^a-zA-Z0-9-_]+/,'_')
```

you might also be able to fold the length into a single chain:
```suggestion
        String sanitized = input
            .trim()
            .take(maxLength)
            .replaceAll(/[^a-zA-Z0-9-_]+/, '_')
```

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
```diff
@@ -849,6 +849,70 @@ class AwsBatchTaskHandler extends TaskHandler implements BatchHandler<String,Job
...
+
+        // Remove leading/trailing underscores and spaces
+        sanitized = sanitized.replaceAll(/^[_\s]+|[_\s]+$/, '')
+
+        // Truncate if too long
+        if (sanitized.length() > maxLength) {
+            sanitized = sanitized.substring(0, maxLength)
+            // Remove trailing underscore/space after truncation
+            sanitized = sanitized.replaceAll(/[_\s]+$/, '')
+        }
```
adamrtalbot: You could probably simplify this a little bit by trimming first then replacing (see chain example above).

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
```diff
@@ -849,6 +849,70 @@ class AwsBatchTaskHandler extends TaskHandler implements BatchHandler<String,Job
...
+    protected Map<String, String> sanitizeAwsBatchLabels(Map<String, String> labels) {
+        if (!labels) return labels
+
+        final result = new LinkedHashMap<String, String>()
+
+        for (Map.Entry<String, String> entry : labels.entrySet()) {
+            final key = entry.getKey()
+            final value = entry.getValue()
+
+            if (key != null && value != null) {
```
adamrtalbot: If the value is `null`, this is skipped. Currently, when the item is `"item": null` is the aws tag silently dropped?

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
adamrtalbot: Of course changing the order affects the output - you may want to replaceAll then take based on what you want.

Searching through the codebase it tends to be a ternary with substring: https://github.com/nextflow-io/nextflow/blob/6231d76fd633587cc3e85a44ff642cc849a2f578/modules/nextflow/src/main/groovy/nextflow/executor/AbstractGridExecutor.groovy#L182-L183

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
edmundmiller: Do we want to simplify or make it readable? 

I'm all for clever code, but I want to make it clear what we're doing and trying to overcome here.

adamrtalbot: Replacing the same variable a bunch of times can be hard to read as well because it's difficult to tell what the variable is at any one time. I wouldn't stack all of these changes into one chain but I would consider simplifying the code a bit. The best idea is to break it down into discrete chunks of obvious code, although that's not always easy to do.

pditommaso:
Let's address the comments please

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
edmundmiller: See https://github.com/nextflow-io/nextflow/pull/6211/commits/9f5c2ea521fc25f8ee3bf0f2d30d994c5d4938e0

pditommaso:
Thinking more about this, not sure this should be done. A label string can be an identifier used by an external system reference some resources. If the label is changed silently will result likely in an inconsistent or an error hard to troubleshoot. I think it's better to hard fail in this case

edmundmiller:
I see your point. It's also not fun to point people to add a custom sanitize function to their config. 

What if we added it behind a config option for each executor(that had the feature)? We could leave it off by default. `aws.batch.sanitize`?

pditommaso:
More than yet another option it would be better to report a warning if the label is not matching AWS constraints and skip it (when strict mode is enabled throw an error)

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
```diff
@@ -864,6 +865,98 @@ class AwsBatchTaskHandler extends TaskHandler implements BatchHandler<String,Job
...
+     * - Allowed characters: letters, numbers, spaces, and: _ . : / = + - @
+     *
+     * @param labels The original resource labels map
+     * @return The labels map (unchanged in validation mode)
+     * @throws ProcessUnrecoverableException when strict mode is enabled and labels are invalid
+     */
+    protected Map<String, String> validateAwsBatchLabels(Map<String, String> labels) {
+        if (!labels) return labels
+
+        final strictMode = executor.session.config.navigate('nextflow.enable.strict', false)
```
pditommaso, line 889: there's NF.isStrictMode for this https://github.com/nextflow-io/nextflow/blob/d558ee3393883f1d10c8c19484a34b8a0afc3659/modules/nextflow/src/main/groovy/nextflow/NF.groovy#L71-L71

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
```diff
@@ -864,6 +865,98 @@ class AwsBatchTaskHandler extends TaskHandler implements BatchHandler<String,Job
...
+            final value = entry.getValue()
+
+            // Check for null keys or values and filter them out (not validation violations)
+            if (key == null) {
+                log.warn "AWS Batch label dropped due to null key: key=null, value=${value}"
+                continue
+            }
+            if (value == null) {
+                log.warn "AWS Batch label dropped due to null value: key=${key}, value=null"
+                continue
```
pditommaso, line 904: should not these be violations instead of warns?

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
bentsherman: `nextflow.enable.strict` is a feature flag, which means it's intended only for language features and is not appropriate for controlling runtime behavior. Also it will eventually be deprecated in favor of strict syntax + static type checking

I would create a new config option for strict runtime behaviors like this. Perhaps something similar to `nextflow.retryPolicy`

bentsherman:
Closing due to lack of response

PR #6209, opened 2025-06-23
Add support for Bitbucket API tokens

tcrespog:
- Add support for Bitbucket API tokens:
  - Contrary to the deprecated app passwords, the API tokens are used with `Authorization: Bearer ...` headers instead of `Authorization: Basic ...` headers.
  - They only seem to work with with `http://api.bitbucket.org/2.0` base URL instead of `https://bitbucket.org/api/2.0`.

PR #6212, opened 2025-06-24
Resolve auth for Git providers that use tokens

tcrespog:
Those Git providers that rely on tokens for authentication ([GitLab](https://github.com/nextflow-io/nextflow/blob/d1e6836ea8bf85c26c0279a74e590593ce9849b2/modules/nextflow/src/main/groovy/nextflow/scm/GitlabRepositoryProvider.groovy#L43-L46), [Gitea](https://github.com/nextflow-io/nextflow/blob/d1e6836ea8bf85c26c0279a74e590593ce9849b2/modules/nextflow/src/main/groovy/nextflow/scm/GiteaRepositoryProvider.groovy#L47-L51), [Bitbucket](https://github.com/nextflow-io/nextflow/pull/6209)) do not need username and password to be defined too.

This PR tweaks the logic that determines whether a Git provider has authentication properties configured or not to take into account the sole presence of the `token` field.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy
@@ -126,12 +126,14 @@ abstract class RepositoryProvider {
     }
 
     boolean hasCredentials() {
-        getUser() && getPassword()
+        return (getUser() && getPassword()) || getToken()
     }
 
     String getUser() { config?.user }
 
     String getPassword() { config?.password }
+
+    String getToken() { config?.token }
 
     /**
      * @return The name of the source hub service e.g. github or bitbucket


pditommaso:
It may be better to override it in the class that has support for it. I only see it supported by these 

https://github.com/nextflow-io/nextflow/blob/b334df1d38c4c151403d8dd84e625191ec07b639/modules/nextflow/src/main/groovy/nextflow/scm/GiteaRepositoryProvider.groovy#L45-L55

https://github.com/nextflow-io/nextflow/blob/b334df1d38c4c151403d8dd84e625191ec07b639/modules/nextflow/src/main/groovy/nextflow/scm/GitlabRepositoryProvider.groovy#L42-L51

tcrespog:
> It may be better to override it in the class that has support for it. I only see it supported by these


The idea of adding it in the parent class is that more may come, such as https://github.com/nextflow-io/nextflow/pull/6209.

pditommaso:
I assume this is addressed by https://github.com/nextflow-io/nextflow/pull/6209. @tcrespog let me know if that's not the case

tcrespog:
> I assume this is addressed by #6209. @tcrespog let me know if that's not the case

Not exactly, providers other than Bitbucket (especially GitLab) would need the logic implemented in this PR too. I'm re-opening the PR; do you want me to address [this](https://github.com/nextflow-io/nextflow/pull/6212#issuecomment-3012076463) comment or is it okay as it is?

pditommaso:
I see. Yes, it would be better to handle at specific provider level

pditommaso:
worth adding a unit test?

pditommaso (approved):

Issue #4591, opened 2023-12-15
Update 'noHttps' option for both Singularity and Apptainer 


pachiras:
Hi!
Recently both Singularity and Apptainer changed '--nohttps' option to '--no-https'.
I wish nextflow supports both options.
Thanks.

shd-piste:
This is still an issue and needs to be fixed.

PR #6216, opened 2025-06-25
Fix apptainer and singularity --no-https option

steven-pilger:
Fixes: #4591


bentsherman:
Thanks for contributing. Can you figure out which versions of Singularity / Apptainer renamed this option? If we merge this, we'll need to document that as a minimum requirement

shd-piste:
Looking into it 👀

Seems like the last entry in the changelog of singularity already talks about deprecating --nohttps:
https://github.com/apptainer/singularity/blob/9dceb4240c12b4cff1da94630d422a3422b39fcf/CHANGELOG.md?plain=1#L55
So that would be singularity version 3.8.4, while 'current' version is 3.8.7 from March 2022.

This is the same for apptainer as it was renamed around that time if I see that right. I can not find the commit that finally drops the support for --nohttps. Is this sufficient? Should this be documented somewhere?

pditommaso (approved):

pditommaso:
Thanks for contributing this patch

pditommaso:
Oops, didn't notice @bentsherman comment

PR #6218, opened 2025-06-25
local-gpu-executor repects CUDA_VISIBLE_DEVICES when set

thealanjason:
In  #5850, if `nextflow` is run with `CUDA_VISIBLE_DEVICES` environment variable set before hand, the gpuIDs in `CUDA_VISIBLE_DEVICES` are ignored when building the `availIds` Map. 

This PR proposes a fix.

When `executor.gpus` is greater than number of devices specified in `CUDA_VISIBLE_DEVICES`, an error is thrown.

If `CUDA_VISIBLE_DEVICES` is not set, then the `0..n` gpuIDs are used to build `availIds` as before.


thealanjason:
I don't know why this PR shows so many changes. 
Only the following lines have changed: https://github.com/nextflow-io/nextflow/pull/6218/files#diff-a47e3cf73541b82f0ca80db440d4cec588f6617d24328844f350adf7c76e0d5dR37-R52

bentsherman:
Incorporated into #5850

PR #6220, opened 2025-06-25
Upgrade to Groovy 5

bentsherman:
Initial attempt to upgrade to [Groovy 5](https://groovy-lang.org/releasenotes/groovy-5.0.html).

Benefits:
- new collection functions like `zip` (can be added to Nextflow standard library)
- pattern matching for `instanceof` (can be added to Nextflow language)
- use `_` for placeholder variables (can be added to Nextflow language)

Changes:

- Had to remove generic type arguments from `instanceof` checks as the Groovy parser doesn't allow it anymore

- Had to add `@Slf4j` to some inner classes to make `log` work in that class. not sure if this is a bug or just a breaking change 

Current issues:

- Groovy complained about a few implementers of `CacheFunnel`, even though I didn't see anything wrong. Converting CacheFunnel to Java fixed the issue for one but not the other

- Multiple issues with the type checker that I'm pretty sure are just Groovy bugs, particularly `CmdLog` and `DumpOp`

- The type checker does not recognize `target` in `TaskConfig` as a Map, seems like a bug. I wonder if it's an edge case since `target` is defined in super class as private and a Delegate

Overall it looks like Groovy 5 beta still has a few compiler bugs to sort out. But I'm excited to get it working so that I can bring the new pattern matching features into the Nextflow language.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdLog.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdLog.groovy
@@ -220,6 +220,7 @@ class CmdLog extends CmdBase implements CacheBase {
     /**
      * Wrap a {@link TraceRecord} instance as a {@link Map} or a {@link Binding} object
      */
+    @Slf4j
     private static class TraceAdaptor extends Binding {
 
         static private int MAX_LINES = 100
@@ -299,13 +300,14 @@ class CmdLog extends CmdBase implements CacheBase {
                 def result = new StringBuilder()
                 path.withReader { reader ->
                     String line
-                    while( (line=reader.readLine()) && c++<MAX_LINES ) {
+                    // MAX_LINES must be cast to `int` in order to satisfy type checker
+                    // seems to be caused by the delegate
+                    while( (line=reader.readLine()) && c++ < (int)MAX_LINES ) {
                         result << line << '\n'
                     }
                 }
 
                 result.toString() ?: TraceRecord.NA
-
             }
             catch( IOError e ) {
                 log.debug "Failed to fetch content for file: $path -- Cause: ${e.message ?: e}"
--- a/modules/nextflow/src/main/groovy/nextflow/extension/DumpOp.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/extension/DumpOp.groovy
@@ -51,7 +51,6 @@ class DumpOp {
 
     DumpOp(Map opts, Closure<String> renderer) {
         checkParams('dump', opts, PARAMS_DUMP)
-        this.source = source
         this.tag = opts.tag
         this.pretty = opts.pretty ?: false
         this.renderer = renderer
@@ -75,11 +74,8 @@ class DumpOp {
 
     DataflowWriteChannel apply() {
 
-        if( !isEnabled() ) {
-            if( source instanceof DataflowWriteChannel )
-                return (DataflowWriteChannel)source
+        if( !isEnabled() )
             throw new IllegalArgumentException("Illegal dump operator source channel")
-        }
 
         final target = CH.createBy(source)
         final events = new HashMap(2)
--- a/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
@@ -659,7 +659,13 @@ class AssetManager implements Closeable {
      * @return {@code true} when the SCM source points to a local file-system repository.
      */
     boolean isLocalScmSource() {
-        return provider instanceof LocalRepositoryProvider
+        // assign to a local before the `instanceof` check: under Groovy 5 the
+        // static compiler otherwise narrows the `provider` *field* type to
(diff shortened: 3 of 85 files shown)


pditommaso:
Worth following this discussion, it looks like there are not negligile perf issue with Groovy 5 and invoke dynamic https://github.com/apache/grails-core/issues/15293

bentsherman:
Interesting, it looks similar to the issues they had with Groovy 4.

It looks like one commenter avoided the perf issues by using `@CompileStatic` and converting code to Java. Since most of the Nextflow codebase is already statically compiled, I suspect we won't be affected as much. But we'll have to run a benchmark to know for sure

modules/nextflow/src/main/groovy/nextflow/cli/CmdLog.groovy
```diff
@@ -299,7 +300,8 @@ class CmdLog extends CmdBase implements CacheBase {
                 def result = new StringBuilder()
                 path.withReader { reader ->
                     String line
-                    while( (line=reader.readLine()) && c++<MAX_LINES ) {
+                    // ERROR: Cannot call java.lang.Integer#compareTo(java.lang.Integer) with arguments [java.lang.Object]
+                    while( (line=reader.readLine()) && c++ < MAX_LINES ) {
```
bentsherman: I tried changing the type of `c` and `MAX_LINES` to `Integer` but I get this type checking error regardless. Seems like a Groovy type checker bug

modules/nextflow/src/main/groovy/nextflow/extension/DumpOp.groovy
```diff
@@ -76,11 +76,12 @@ class DumpOp {
...
         if( !isEnabled() ) {
-            if( source instanceof DataflowWriteChannel )
-                return (DataflowWriteChannel)source
+            if( source instanceof DataflowWriteChannel dwc )
+                return dwc
             throw new IllegalArgumentException("Illegal dump operator source channel")
         }
 
+        // ERROR: Cannot find matching static method nextflow.extension.CH#createBy(groovyx.gpars.dataflow.DataflowWriteChannel)
         final target = CH.createBy(source)
```
bentsherman: Based on my limited knowledge of how the Groovy type checker does type inference with `instanceof`, it seems to be applying the `instanceof` check incorrectly, treating `source` as type `DataflowWriteChannel` through the entire method instead of just the if statement. Likely a Groovy bug

modules/nextflow/src/main/groovy/nextflow/cli/CmdLog.groovy
bentsherman: This seems to be caused by the presence of the delegate field, which can obscure other fields. I have seen issues like this before when using delegates. I submitted a ticket to the Groovy maintainers, but for now we can just do `(int) MAX_LINES` and move on.

modules/nextflow/src/main/groovy/nextflow/extension/DumpOp.groovy
bentsherman: Looking at the surrounding context, I don't think this instanceof check is actually needed. I submitted a ticket to the Groovy maintainers, but for now I think we can just remove this check

bentsherman:
Found sensible workarounds to all of the compile errors

Now the problem is failing tests

There seems to be an issue with Path semantics:
- `assert path` always fails and you have to do `assert path != null` instead
- lots of failures seem to be paths not being resolved correctly (e.g. tmp test dir vs launch dir)

I know that we modified Path behavior with some Groovy magic, so likely something is broken with that

bentsherman:
Groovy 5.0.7 released, though still waiting for Maven Central to be updated

Applied separate fix for Path truthiness. Will re-run CI when Maven is ready

bentsherman:
Relevant Groovy 5 commit: https://github.com/apache/groovy/commit/a1c006c974a3aca7f3d8c37c607535849e388580

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
```diff
@@ -659,7 +659,13 @@ class AssetManager implements Closeable {
...
      */
     boolean isLocalScmSource() {
-        return provider instanceof LocalRepositoryProvider
+        // assign to a local before the `instanceof` check: under Groovy 5 the
+        // static compiler otherwise narrows the `provider` *field* type to
+        // LocalRepositoryProvider class-wide, inserting bogus casts in other
+        // methods (e.g. clone/getRemoteRevisions/updateModules) that fail with
+        // ClassCastException for non-local providers.
+        final p = provider
+        return p instanceof LocalRepositoryProvider
```
bentsherman, line 668: Looks like a Groovy 5 bug, I will submit a bug report

bentsherman: https://issues.apache.org/jira/browse/GROOVY-12166

bentsherman:
Tests are finally passing!

TODO:
- [ ] Remove local var + instanceof workaround after Groovy bug is fixed
- [ ] Review changes to TaskConfig carefully

PR #6222, opened 2025-06-26
Updated wave.retryPolicy default values

munishchouhan:
This PR will update wave.retryPolicy default values to the latest ones
https://github.com/nextflow-io/nextflow/blob/72415db8935d225c8f5cdbca18c0ed6938c65852/plugins/nf-wave/src/main/io/seqera/wave/plugin/config/RetryOpts.groovy#L30-L33


bentsherman:
Should we just remove the wave `RetryOpts` and use the common `RetryConfig` instead?

pditommaso:
Not sure it's worth, it's already used under the `wave` scope and it may be better to use `nextflow.retryPolicy` only for core components

bentsherman:
I was only talking about reusing the `RetryConfig` class, but without reusing `nextflow.retryPolicy`

pditommaso:
Got it, but still don't see much valuable only for that. I was thinking having a better abstraction for the retry logic other than the config that's getter replicated in severable places. In that cases it would make more sense, but I don't see as an urgent thing.

PR #6228, opened 2025-06-27
Update stalebot

bentsherman:
The stalebot seems to be more trouble than it's worth. I find myself having to undo the stalebot in many cases, although sometimes it's useful to flag issues that should be closed.

I suggest we either remove it or disable the ability to close issues, so that it can only mark issues as stale.

If we do keep it, we might as well update to the [stale action](https://github.com/actions/stale) since the stalebot we are using is deprecated. The README has some example configurations.


bentsherman:
Added stale action which only marks issues and PRs as stale, does not comment or close.

See the [README](https://github.com/actions/stale?tab=readme-ov-file#all-options) for the full list of options

pditommaso (approved):
Makes sense!

PR #6234, opened 2025-06-30
fixup: Azure Batch disk slot calculation demoted to debug

adamrtalbot:
A logging statement was inadvertently left as warning and not debug. It's just information, not required on every run so this PR demotes it to debug.


pditommaso:
In which version has this been introduced ?

Issue #5943, opened 2025-04-03
Usage of secrets in Workflow Script on AWS Batch and Platform


FriederikeHanssen:
## Bug report 

- Calling workflow secrets in the onComplete block does not work locally and on AWS Batch (submitted with Platform)
- Calling workflow secrets in the `workflow` body does not appear to work on AWS Batch (submitted with Platform)

### Expected behavior and actual behavior

I would like to use secrets as described [here](https://www.nextflow.io/docs/latest/secrets.html#pipeline-script) in both the workflow and onComplete block.

### Steps to reproduce the problem

I created a small example here:  https://github.com/FriederikeHanssen/mre/blob/secrets/main.nf

For testing locally:

1. Set a secret: ` nextflow secrets set RIKE_SECRET "Hello world"`
2. Run pipeline: (tested with both, version that originally released it and newer version):
              - `NXF_VER=24.03.0-edge nextflow run main.nf` 
              - `NXF_VER=24.10.5 nextflow run .`

For testing on Platform + AWS Batch

1. Set up a workspace etc. Define a RIKE_SECRET. Add the pipeline to the launchpad, ensure it is part of the Workflow secrets.
2. Run pipeline

### Program output 

Locally:
```
❯ nextflow run .

 N E X T F L O W   ~  version 24.10.5

Launching `./main.nf` [special_montalcini] DSL2 - revision: d462ac85b4

The secret in Workflow is: Hello world
ERROR ~ Failed to invoke `workflow.onComplete` event handler

 -- Check script 'main.nf' at line: 6 or see '.nextflow.log' file for more details
```

AWS Batch & platform:

```
The secret in Workflow is: null

ERROR ~ Failed to invoke `workflow.onComplete` event handler
-- Check script '.nextflow/assets/FriederikeHanssen/mre/main.nf' at line: 6 or see 'nf-1Sc7D9V8WrFVRF.log' file for more details

```

### Environment 

* Nextflow version: 24.10.5 & 24.03.0-edge
* Java version: [?]
* Operating system: macOS & AWS Batch
* Bash version: zsh 5.9 (arm64-apple-darwin24.0)

jpfeuffer:
Related to #4177 and the associated feature request. We are also desperately looking for that functionality.
I think at least the [docs](https://www.nextflow.io/docs/latest/secrets.html#pipeline-script) should be fixed in the meantime.

bentsherman:
[...]

Issue #5312, opened 2024-09-18
Using secrets in `includeConfig` doesn't work with AWS secrets


bentsherman:
We added the ability to use secrets in the `includeConfig` path in #4840 . This works for local secrets, but not AWS secrets, because AWS secrets requires the xpack-amzn plugin and the config is resolved before plugins are loaded, since the config may include additional plugins.

To fix this I think we would need to load plugins in two phases, adding an "early" phase which happens before the config is resolved.

Not sure how much demand there is for this use case, but it would include any AWS Batch run in Seqera Platform that loads a private config using a secret credential.

cc @markpanganiban

pditommaso:
Bit ugly, because plugins depends on the config 🤦

pditommaso:
Likely assuming the AWS config does not rely in the config setting it's possible to load the AWS plugin to resolve the secrets

pditommaso:
There are two possible approaches: 
1. Fetching the branch to be used via the provider API, likely [here](https://github.com/nextflow-io/nextflow/blob/349f4f6f3185c9b47d6777dd1c5ecadc348ba8c5/modules/nextflow/src/main/groovy/nextflow/scm/GiteaRepositoryProvider.groovy#L86-L95) 
2. or using `git remote` via jgit (see [here](https://github.com/nextflow-io/nextflow/blob/349f4f6f3185c9b47d6777dd1c5ecadc348ba8c5/modules/nextflow/src/main/groovy/nextflow/scm/GiteaRepositoryProvider.groovy#L86-L95))

Tagging @tom-seqera for visibility

ewels:
Bonus functionality that could be unlocked if we can get AWS credentials in for config loading: the ability to load a Nextflow config file from `s3://` (in addition to local / `http` etc).

adamrtalbot:
> Not an expert but I have the feeling that not even access to secrets in pipeline scripts is possible without xpack-amzn being loaded before discovering the secrets providers.
> At least I cannot get secrets to work on the head node (with AWS batch) which is a serious limitation for us.
> No secrets in onComplete, no conditional steps based on secrets, etc.
> Happy to open another issue though, if it is mistakenly unrelated.

This isn't currently supported but there is a feature request [here](https://seqera.canny.io/feature-requests/p/enable-read-access-to-pipeline-secrets-in-the-head-node).

bentsherman:
[...]

PR #6249, opened 2025-07-03
Fix plugin secrets in config

### LINKED ISSUE #5494: Can no longer load SecretsProviders from plugins
## Bug report 

`SecretsProvider`s can no longer be loaded from nextflow plugins

In 23.10, running Nextflow with a plugin which contains a custom `SecretsProvider` would be able to use the provider from the plugin. Since 24.04, the `LocalSecretsProvider` is always used, even if a plugin provides a custom implementation with a higher `@Priority`. 

This seems to be because `SecretsLoader.secretContext()` is called by the `ConfigBuilder` during reading of the nextflow config. This happens before plugins have been loaded, so it will only find the `LocalSecretsProvider` and not any providers from the plugin(s).

bentsherman:
Fix #5494 #5943

This PR fixes support for plugin secrets (i.e. AWS secrets) in the config by loading the config before resolving plugins. If the config attempts to use secrets, it will be reloaded with the complete set of secrets providers.

This is possible because plugin config must be static, unlike other config which can be dynamic and reference secrets. Therefore, we can simply load the config to extract the list of plugins (and other config settings like `tower.enabled` and `process.executor` which can trigger additional core plugins.

To minimize the impact of loading the config twice, we only reload the config if we detect that secrets were used by the config on the first load. This way, the double config load should only affect users who want to use secrets in the config.

In summary:
1. Load config with "shim" secrets provider which only tracks whether secrets are used by config
2. Load plugins
3. Load secrets provider
4. Reload config if secrets were used in config


bentsherman:
Here is how I've been testing:
1. Build with `make pack`
2. Upload dist binary to s3
3. Launch this pipeline in platform: https://github.com/bentsherman/nf-tests/tree/plugin-secrets
4. Download dist binary from s3 and replace default `nextflow` in pre-run script

However, it's not loading the `xpack-amzn` plugin:
```
Jul-03 19:06:49.448 [main] DEBUG nextflow.plugin.PluginsFacade - Plugins resolved requirement=[nf-tower@1.12.0, nf-amazon@2.15.0, nf-wave@1.13.0, nf-cloudcache@0.4.3]
```

It's at least loading `nf-amazon` since the workDir is s3, but I can't remember how to trigger the xpack plugin

bentsherman:
I wonder if the issue is that I'm using the dist release instead of the standard release

pditommaso (commented):
This looks smart. Moving to draft until the problem with the plugins is addressed. Also unit test and integration test should be added (it may be convenience setup a test GH repo for this)

bentsherman:
It turns out the issue was that I was using the single-VM CE, which doesn't support AWS secrets. It works with AWS Batch.

Regarding testing, I don't think we'll need a git repo unless you want to do an e2e test with platform. I'll add a local e2e test that just tests the detection of config secrets -> config reloading.

bentsherman:
One minor issue with this double-loading approach is that you can't use secrets in a config include (#5312). On the first load, the secret will be null and the include will likely fail, which means the run will fail

However, you should be able to work around this using a dynamic include:
```groovy
includeConfig secrets.GITHUB_TOKEN
    ? "https://raw.githubusercontent.com/markpanganiban/nf-test/master/ubuntu.config?token=${secrets.GITHUB_TOKEN}"
    : '/dev/null'
```

On the first load, it will access the secret, triggering a reload, but won't attempt to include the secured config. On the second load it will include the config.

I will add this to the docs on config secrets

docs/secrets.md
```diff
@@ -48,6 +50,21 @@ The above snippet accesses the secrets `MY_ACCESS_KEY` and `MY_SECRET_KEY` and a
 Secrets cannot be assigned to pipeline parameters.
 :::
 
+:::{versionadded} 25.10.0
+:::
+
+In order to enable the use of AWS secrets in configuration files, Nextflow loads the configuration once without secrets enabled, and reloads the configuration only if secrets were accessed. This is because the AWS secret integration is part of the `nf-amazon` plugin, and plugins are resolved after the configuration is loaded (since plugins can be specified in the configuration). As a result, config secrets must be used in a way that does not cause the config resolution to fail when secrets are not present.
```
christopher-hakkaart:
```suggestion
Nextflow supports AWS secrets in configuration files through the `nf-amazon` plugin. However, due to the way plugins are loaded, there are specific considerations when using secrets in configuration files:

- **Initial config load**: Nextflow first loads the configuration _without_ secrets enabled.
- **Plugin resolution**: The `nf-amazon` plugin is resolved after the initial configuration load. This is because plugins themselves can be specified in the configuration.
- **Secret reloading**: If secrets are accessed during configuration and the initial load succeeds, Nextflow will reload the configuration with secrets enabled.

As a result, config secrets must be used in a way that does not cause the config resolution to fail when secrets are not present.
```
Suggestion to make these into bullet points. Please check meaning is retained.

bentsherman:
The e2e test that I added is failing in GitHub, but not sure why because it works for me locally

pditommaso:
For me does not work locally 

```
» cd tests/checks
» b qrun.sh config-secrets.nf/
> Running test: config-secrets.nf

~ Test 'config-secrets.nf' run failed
   + set -e
   + /Users/pditommaso/Projects/nextflow/launch.sh secrets set MY_SECRET hello-world
   + /Users/pditommaso/Projects/nextflow/launch.sh -q run ../../config-secrets.nf -c ../../config-secrets.config
   + tee stdout
   + grep 'Config file used secrets -- reloading config with secrets provider'
   + false

Jul-15 17:54:46.610 [main] DEBUG nextflow.cli.Launcher - $> /Users/pditommaso/Projects/nextflow/launch.sh -q run ../../config-secrets.nf -c ../../config-secrets.config
Jul-15 17:54:46.634 [main] INFO  nextflow.cli.CmdRun - N E X T F L O W  ~  version 25.06.0-edge
Jul-15 17:54:46.646 [main] DEBUG nextflow.plugin.PluginsFacade - Setting up plugin manager > mode=dev; embedded=false; plugins-dir=/Users/pditommaso/Projects/nextflow/plugins; core-plugins: nf-amazon@3.0.0,nf-azure@1.18.0,nf-cloudcache@0.4.3,nf-codecommit@0.2.3,nf-console@1.2.1,nf-google@1.22.1,nf-k8s@1.1.0,nf-tower@1.13.0,nf-wave@1.14.0

```

bentsherman:
I see now, I wasn't testing it correctly on local. The e2e test runs with `-q` which suppresses `log.info` but not `println`. Updated the testing docs

docs/secrets.md
```diff
@@ -13,6 +13,8 @@ This feature allows you to decouple the use of secrets in your pipelines from th
 
 When a pipeline is launched, Nextflow injects the secrets into the run without leaking them into temporary execution files. Secrets are provided to tasks as environment variables.
 
+Secrets can be used with the local executor and the grid executors (e.g., Slurm or Grid Engine). Secrets can be used with the AWS Batch executor when launched from [Seqera Platform](https://seqera.io/blog/pipeline-secrets-secure-handling-of-sensitive-information-in-tower/).
```
christopher-hakkaart:
```suggestion
Secrets can be used with the local executor and grid executors (e.g., Slurm or Grid Engine). Secrets can be used with the AWS Batch executor when launched from [Seqera Platform](https://seqera.io/blog/pipeline-secrets-secure-handling-of-sensitive-information-in-tower/).
```

docs/secrets.md
```diff
@@ -45,9 +47,32 @@ aws {
 The above snippet accesses the secrets `MY_ACCESS_KEY` and `MY_SECRET_KEY` and assigns them to the corresponding AWS config settings.
 
 :::{warning}
-Secrets cannot be assigned to pipeline parameters.
+Secrets should not be assigned to pipeline parameters, as they can be easily leaked by the pipeline.
```
christopher-hakkaart:
```suggestion
Secrets should not be assigned to pipeline parameters, as they can be leaked by the pipeline.
```
Highly optional suggestion

christopher-hakkaart (commented):
Some very minor suggestions. Feel free to ignore if the meaning changes.

docs/developer/index.md
```diff
@@ -137,6 +137,13 @@ cd tests/checks
 ./qrun.sh
 ```
 
+To run a specific integration test:
+
+```bash
+cd tests/checks
+./qrun.sh <folder>
```
christopher-hakkaart:
```suggestion
./qrun.sh <FOLDER>
```

pditommaso:
@bentsherman can you have a look at the conflicts? I've tried but not sure how address them

pditommaso:
@claude review

pditommaso:
Ok, thread safety is not relevant because this code is not expected to run in parallel manner. Added more docs and renamed `ConfigNullProvider` to `EmptySecretProvider` to avoid confusing it with `NullProvider`

pditommaso (approved):

robsyme:
This is great! A huge bonus is that now we can do `includeConfig` from remote sources, e.g.

```groovy
plugins { id 'nf-amazon' }

// Force Phase 2 reload by accessing a secret (even if it doesn't exist)
params.dummy = secrets.DUMMY_SECRET ?: 'default'

// Use conditional to defer S3 access until Phase 2 when plugins are loaded
includeConfig nextflow.plugin.Plugins.getManager()
    ? 's3://my-bucket/path/to/nextflow.config'
    : '/dev/null'
```

Is there a more elegant way to force that second reload of configuration?

bentsherman:
I was mainly thinking about config includes that reference a secret like the github token. If you don't actually need a secret then you'll need to use a dummy as in your example. But you can use that to simplify your include:

```groovy
includeConfig secrets.DUMMY_SECRET
    ? 's3://my-bucket/path/to/nextflow.config'
    : '/dev/null'
```

Accessing internal classes like that probably shouldn't be allowed anyway

robsyme:
This is great. We can store the S3 path in a workspace secret and then just have configuration:

```groovy
includeConfig secrets.WORKSPACE_CONFIG ?: '/dev/null'
```

PR #6254, opened 2025-07-04
Fix bugs with workflow output and lineage

bentsherman:
This PR fixes two bugs related to workflow outputs and lineage:

1. Nested maps in published outputs were not properly normalized for the lineage / index file.
2. Publishing a workflow output with a typical structure (list of maps containing files) causes a stack overflow error

@jorgee I have added a unit test demonstrating the stack overflow error with the LinEncoder. Can you recommend a fix?

TODO:
- [ ] fix stack overflow error
- [ ] document how non-task output files are handled


jorgee:
The encoding of the Path class with GSON was creating a infinity loop creating the stackoverflow. I have added a GSON `TypeAdapter` to manage Path writes and reads to/from JSON as strings and seems to fix the issue.

bentsherman:
@jorge thanks, can you approve it for your side? then I will add some docs and merge

PR #6256, opened 2025-07-07
Ignore secrets in stub run

### LINKED ISSUE #6251: Stub runs & Secrets
## New feature

When running a pipeline with `-stub-run` , it would be helpful to be  able to ignore secrets that are set in a process.

## Use case 

We are using stub to quickly prototype something and to check the channel logic, where we typically wouldn't want to do the actual tool call that requires the secret. Allowing to ignore the secrets directive would be very helpful so we don't need to setup mock secrets for testing

bentsherman:
Close #6251 

Not 100% sure about this yet, just trying it out. Need to consider whether there are use cases where secrets would still be desired in stub run


bentsherman:
Based on discussion in #6251 , I think it would be better to have a config scope for stub-related options, so that the user can toggle individual features based on their needs (conda, containers, secrets, etc)

PR #6260, opened 2025-07-09
Include support for git-remote-s3

jorgee:
S3 can be used to store git repositories using the [git-remote-s3](https://github.com/awslabs/git-remote-s3) implementation of [remote helper](https://git-scm.com/docs/gitremote-helpers).

This branch provides support for this S3 repositories as part of the `nf-amazon` plugin. 
It implements a JGit `Transport ` with the Fetch and Push connection as well as the `RepositoryProvider`, `ProviderConfig` `RepositoryFactory` required to support s3 git repositories.
The fetch connection is used to clone and pull branches of the repository in the clone, pull and run commands
The push connection is able to push branches. 

Current Limitations:
- ~Due to a limitation in the access rights of some JGit transport classes, when a branch is pushed it is correctly stored in the remote repository but the local repo refs are not updated with the reference to the remote ref (origin/branch)~
- No LSF and submodules supported.
- Bundle size limited to 5GB (putObject operation) 

TODO:
- [x] Try to fix limitations
- [x] Get defaults from aws config
- [x] Test pull when branch is detached
- [x] Test push non up-to-date branch
- [x] Add unit tests


jorgee:
I have added a prototype implementation of a push command to enable to push a local directory to a remote repository. It uses the jgit, so in theory, it should work with whatever git repository but I have just tested it with s3-git-remote.

```
Usage: push [options] Path to push
  Options:
    -h, -help
       Print the command usage
       Default: false
    -max-size
       Maximum file size in MB to push without confirmation (default: 10)
       Default: 10
    -message, -m
       Commit message
       Default: Push from nextflow
    -repo
       Defines the repository to push to
    -r, -revision
       Revision of the project to push 
       Default: main
```
Two main use cases:

**1- The specified folder is not a git repository:**
For instance running `nextflow push new-repo -repo s3://s3-git-remote/repos/new-repo` where new-repo is not a local git repo.
 It inits git in the `new-repo`, creates a .gitignore with (.nextflow* and work), look for big files and ask the user to include in the `.gitignore`, creates a branch, commits everything else to this branch and pushes to` s3://s3-git-remote/repos/new-repo`

After the push a user can run `nextflow run s3://s3-git-remote/repos/new-repo` and run the main.nf contained in the repo.

**2- The folder is already a git repo:** 
If the user has specified `-revision` and `-repo`, it checks if the current folder has defined the `repo` as remote and it is currently check out to the revision. 

If the user has not specified anything, it uses the remote and revision currently set in the git folder. 
For instance, if you run `nextflow push new-repo`, it automatically pushes all the changes and new files to `s3://s3-git-remote/repos/new-repo` and `main` branch.

pditommaso:
Is the command with the repo option in the example correct? I'd have expected 

```
nextflow push s3://s3-git-remote/repos/new-repo -repo new-repo
```

instead of 

```
nextflow push new-repo -repo s3://s3-git-remote/repos/new-repo
```


also because later it shows the one without repo 

```
nextflow run s3://s3-git-remote/repos/new-repo
```

plugins/nf-amazon/src/resources/META-INF/extensions.idx
```diff
@@ -0,0 +1,22 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+nextflow.cloud.aws.batch.AwsBatchExecutor
+nextflow.cloud.aws.util.S3PathSerializer
+nextflow.cloud.aws.util.S3PathFactory
+nextflow.cloud.aws.fusion.AwsFusionEnv
+nextflow.cloud.aws.mail.AwsMailProvider
+nextflow.cloud.aws.scm.S3RepositoryFactory
```
pditommaso: This should not be needed anymore using the new plugin system. Consider synching this branch with master

jorgee:
> Is the command with the repo option in the example correct? I'd have expected
> 
> ```
> nextflow push s3://s3-git-remote/repos/new-repo -repo new-repo
> ```
> 
> instead of
> 
> ```
> nextflow push new-repo -repo s3://s3-git-remote/repos/new-repo
> ```
> 
> also because later it shows the one without repo
> 
> ```
> nextflow run s3://s3-git-remote/repos/new-repo
> ```

In the current implementation, the main argument is the path that we want to push. I could do as you propose assuming that the path to push is the current working dir `.`

As commented before, there are two cases: when the folder is already a local git repo (contains .git) or not. The one without repo is when .git already exists and you want to update with current changes. It will push to the repository in origin and current branch.

plugins/nf-amazon/src/resources/META-INF/extensions.idx
jorgee: Removed

jorgee:
@claude  review

jorgee:
Updates:
- As suggested by @pditommaso and Claude, I have put the repo as the first argument and use the current working directory as the default directory. To specify another directory, use the `-directory `option.

```
#Push current working dir to 's3://s3-git-remote/repos/new-repo'
nextflow push s3://s3-git-remote/repos/new-repo 

#Push 'path/to/local/dir' to 's3://s3-git-remote/repos/new-repo'
nextflow push s3://s3-git-remote/repos/new-repo -d path/to/local/dir 

#Push current working dir to origin defined in .git. Fail if no .git
nextflow push 
```
- Regarding Claude comment:
1. `RemoteRefUpdate`: the RemoteRefUpdate.setStatus method is package-private and the JGit Jar is signed. No public API to perform it, and no possibility to add PushConnection to the same package. So, no other way to implement the required remote reference update. 

2. `S3RepositoryProvide.readBytes` is inefficient: git-remote-s3 stores tasks as bundles, requiring downloading a bundle per branch, no way to access a single file. We need to download a bundle with the required branch. The inefficiency is for the first clone. The readBytes is invoked before cloning to check if it is a valid repo, and the used localpath where to clone is in the AssetManager. No way to know if the localPath at RepositoryProvider. A different implementation would require modifying the common interaction between at AssetManager and RepositoryProvider. 

 - Added More Unit testing and a validation test in AWS, which pushes a pipeline in a local folder to S3 remote and runs it later using the S3 URL.

pditommaso:
This PR is awesome! I'd to propose a few improvements: 

- [x] not fully convinced about the automatic commit (at least not as default). It may be better to add specific option to enable auto-commit e.g. `-a` (auto) or `-c` (commit)
- [x] along the same manner for auto init. maybe it could be included in the auto option above. But let's discuss later about this, and keep as it is for now 
- [x] Require as mandatory `nextflow.config` on `push` operation. If missing it should stop. 
- [x] Let's use `.nextflow*` and `work` as default for `.gitignore` 
- [x] no need for `-directory`, it should push the current project directory

pditommaso:
Also, it may be needed to investigate a bit more the support for tag and branches. Not sure the current impl works right now 

```
https://github.com/nextflow-io/nextflow/blob/a0d2cfe814076edabddaf567af14824a92c2fa85/plugins/nf-amazon/src/main/nextflow/cloud/aws/scm/S3RepositoryProvider.groovy#L198-L216
```

For example in my tests it's only reporting the main branch, when I've pushed other tags 

```
nextflow info s3://nextflow-ci/git-repo
 project name: nextflow-ci/git-repo
 repository  : s3://nextflow-ci/git-repo
 local path  : /Users/pditommaso/.nextflow/assets/nextflow-ci/git-repo
 main script : main.nf
 revision    : * main (default)
```

ewels:
Before we merge this I'd like to think through potential UX issues. I think it's looking very solid on the technical side but I still have some concerns about developer anti-patterns and risks. I will get to this ASAP, it's second on my list. So hopefully tomorrow / start of next week. But wanted to flag on this PR for visibility.

jorgee:
@pditommaso I have addressed most of the changes that you suggested. But what's the reason to stop pushing if nextflow.config doesn't exist? Should we also check if there is any .nf file in the folder?

Regarding `nextflow info`, I have seen that this command shows the local branches in the assets folder. When doing `nextflow run s3://xxx` it only downloads a branch. However, if you do a `nextflow pull s3://xxx` it downloads the full repo and you can see all the branches in `nextflow info`

pditommaso:
> But what's the reason to stop pushing if `nextflow.config` doesn't exist?

Because without that if fails to pull/run the pipelines. It's not specific to the new storage, it happens because nextflow tries to check the existence of `nextflow.config` when pulling the code via Git 


> Regarding nextflow info, I have seen that this command shows the local branches in the assets folder. When doing nextflow run s3://xxx it only downloads a branch. 

I wonder if it should be downloaded implicitly all repo for any git related command

adr/20251111-s3-git-repository-provider.md
pditommaso, line 1: @jorgee I've added a tech document regarding this implementation. Please review and to see if it's technically accurate

jorgee: Looks fine for me. Very accurate!

bentsherman:
Closing since we decided not to move forward with this approach. We can always re-open if we decide to investigate again

PR #6267, opened 2025-07-10
Prevent using Azure Low Priority VMs when using Batch Managed mode

### LINKED ISSUE #6258: Deprecation of Azure Low Priority VMs
Azure Low Priority VMs are going away: https://learn.microsoft.com/en-us/azure/batch/low-priority-vms-retirement-migration-guide

For Nextflow, this means a user using Azure low priority VM pools in a Batch Managed account will fail.

We should handle this with a clear error message or failure so a user understands the issue and knows how to deal with it.

adamrtalbot:
Azure Low Priority VMs are deprecated in Batch Managed pool allocation mode but still supported in User Subscription mode. This adds validation to:
- Block low priority VMs for Batch Managed accounts with clear error message
- Allow low priority VMs for User Subscription accounts
- Gracefully handle unknown allocation modes with warnings

This required adding a `subscriptionId` to the `azure.batch` config to enable creation of the management client. If not supplied Nextflow will raise a warning and continue. This does open us up to being able to do more with Azure Batch because we have access to the management API.

Fix Azure Low Priority VM deprecation for Batch Managed accounts (#6258)

### DIFF
--- a/plugins/nf-azure/build.gradle
+++ b/plugins/nf-azure/build.gradle
@@ -46,6 +46,9 @@ dependencies {
     api('com.azure:azure-identity:1.15.1') {
         exclude group: 'org.slf4j', module: 'slf4j-api'
     }
+    api('com.azure.resourcemanager:azure-resourcemanager-batch:1.1.0-beta.4') {
+        exclude group: 'org.slf4j', module: 'slf4j-api'
+    }
 
     // address security vulnerabilities
     runtimeOnly 'io.netty:netty-handler:4.1.118.Final'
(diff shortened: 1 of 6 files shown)


pditommaso:
Not sure this makes sense considering "low priority" feature is going to be retired on 30th Sept, 2025 by Microsoft

bentsherman:
Apparently the `lowPriority` setting is also used to enable spot VMs which are the new replacement :facepalm:

pditommaso:
> Apparently the lowPriority setting is also used to enable spot VMs which are the new replacement

Umm, very confusing. We may still consider to  call `spot` in Nextflow and align it with other cloud providers. In any case this PR needs a detailed docs to highlight what's new and how to use the new spot instances  

https://learn.microsoft.com/en-us/azure/batch/batch-spot-vms

PR #6270, opened 2025-07-11
Use ECS properties instead of container properties

jorgee:
Closes https://github.com/nextflow-io/nextflow/issues/5096

Ports the @tom-seqera implementation in #5391 to AWS SDK v2 implementation.

### DIFF
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
@@ -16,7 +16,7 @@
 
 package nextflow.cloud.aws.batch
 
-
+import static nextflow.cloud.aws.batch.AwsBatchHelper.*
 import static nextflow.cloud.aws.batch.AwsContainerOptionsMapper.*
 
 import java.nio.file.Path
@@ -51,13 +51,13 @@ import nextflow.util.TestOnly
 import software.amazon.awssdk.services.batch.BatchClient
 import software.amazon.awssdk.services.batch.model.ArrayProperties
 import software.amazon.awssdk.services.batch.model.AssignPublicIp
-import software.amazon.awssdk.services.batch.model.AttemptContainerDetail
+import software.amazon.awssdk.services.batch.model.AttemptDetail
 import software.amazon.awssdk.services.batch.model.BatchException
 import software.amazon.awssdk.services.batch.model.ClientException
-import software.amazon.awssdk.services.batch.model.ContainerOverrides
 import software.amazon.awssdk.services.batch.model.DescribeJobDefinitionsRequest
 import software.amazon.awssdk.services.batch.model.DescribeJobDefinitionsResponse
 import software.amazon.awssdk.services.batch.model.DescribeJobsRequest
+import software.amazon.awssdk.services.batch.model.EcsPropertiesOverride
 import software.amazon.awssdk.services.batch.model.EphemeralStorage
 import software.amazon.awssdk.services.batch.model.EvaluateOnExit
 import software.amazon.awssdk.services.batch.model.Host
@@ -79,6 +79,8 @@ import software.amazon.awssdk.services.batch.model.RetryStrategy
 import software.amazon.awssdk.services.batch.model.RuntimePlatform
 import software.amazon.awssdk.services.batch.model.SubmitJobRequest
 import software.amazon.awssdk.services.batch.model.SubmitJobResponse
+import software.amazon.awssdk.services.batch.model.TaskContainerOverrides
+import software.amazon.awssdk.services.batch.model.TaskPropertiesOverride
 import software.amazon.awssdk.services.batch.model.TerminateJobRequest
 import software.amazon.awssdk.services.batch.model.Volume
 /**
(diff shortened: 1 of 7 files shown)


pditommaso:
Nice, let's test using `[e2e prod]` in the commit message

pditommaso:
@jorgee aws tests are failing

jorgee:
It seems the tests are failing because of the job definition reuse. The old definition is using containerProperties and the new one is trying to override with ecsProperties. If I am not wrong, the job definition is reused if container name and nf-token is the same, isn't it? So, I think we should include something else to the token to ensure both are using the same container or ecs properties.

jorgee:
Token was basically hashing the result of `containerProperties.toString`. I have added a fake property `ecsMode=true` in the new version. If there is a job definition with same values but with the previous version, the token will be different and it will generate a new revision. Now, the job is failing because of CI user role permissions as in the other CI pipelines.

pditommaso:
Launching some e2e tests

pditommaso:
@jorgee would not this impact also the xpack plugin for amazon?

jorgee:
> @jorgee would not this impact also the xpack plugin for amazon?

The xpack is overriding the `configJobDefRequest` method that is not changed. The changes are internal in the `RegisterJobDefinitionModel.toBatchRequest` method where it sets the ecs properties. But let me do a test to see if I miss something.

jorgee:
When testing with xpack I have got this error. Not sure yet if it is related to this change or happening also with the master. 

```
java.lang.ClassCastException: class software.amazon.awssdk.services.batch.model.MountPoint cannot be cast to class software.amazon.awssdk.services.batch.model.MountPoint (software.amazon.awssdk.services.batch.model.MountPoint is in unnamed module of loader org.pf4j.PluginClassLoader @4ea17147; software.amazon.awssdk.services.batch.model.MountPoint is in unnamed module of loader org.pf4j.PluginClassLoader @9a20cbd)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1597)
        at software.amazon.awssdk.services.batch.model.MountPointsCopier.copy(MountPointsCopier.java:32)
        at software.amazon.awssdk.services.batch.model.ContainerProperties$BuilderImpl.mountPoints(ContainerProperties.java:2578)
        at nextflow.cloud.aws.batch.model.ContainerPropertiesModel.toBatchContainerProperties(ContainerPropertiesModel.groovy:254)
```

jorgee:
It was an xpack gradle problem. Working after applying the fix.

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy
```diff
@@ -246,7 +248,8 @@ class AwsBatchTaskHandler extends TaskHandler implements BatchHandler<String,Job
             checkIfUnschedulable(job)
         // fetch the task arn
         if( !taskArn )
-            taskArn = job?.container()?.taskArn()
+            taskArn = getTaskProperties(job)?.taskRoleArn()
+                ?: job?.container()?.taskArn()
```
pditommaso, line 252: Why now check first `getTaskProperties(job)` and then `job?.container()?.taskArn()` ?

jorgee: I got it from Tom's PR #5391. I think it is needed because of the reuse of job definitions.  If the job was defined with ECS, the information is inside `TaskProperties`. If it is reusing an old definition with container properties, the `taskProperties` are empty and you need to get in the `Container`. It is the same for other job properties. 

Maybe, with the change I did to do not reuse jobs definitions with container properties, it is not needed (https://github.com/nextflow-io/nextflow/pull/6270/commits/54ee92bf69a71dd49f549c79e9f2c0d79c2f3eb1). However, I am not 100% sure because the hash is only applied when container properties are defined, and there could be cases were old definitions are still reused. Maybe the best is just add a comment to explain it.

pditommaso: Not sure to follow, if it's an improvement it should go into a separate PR and keep this as a mere SDK upgrade

jorgee: It is not an improvement. When using ECS properties in the task definition (the goal of the PR), some properties that were obtained before from the container are now inside TaskProperties. This is why we check them first in  `getTaskProperties(job)`, it is the expected place with the new definition. We have a fallback to the old container properties because of the job definition reusage. If a job submission is reused with the old container properties, the task properties will be empty and we should get them from the container properties as before.

pditommaso: ok, triggered e2e tests

jorgee:
It seems that something is broken with the e2e test triggering condition. It has not triggered the e2e tests.

jorgee:
The problem was the git commit message was returning a string with escapes due to `%q` option in 
`echo "commit_message=$(printf "%q" "$COMMIT_MESSAGE")" | head -n 1 >> $GITHUB_OUTPUT`
The logs of the get_commit_message show the escaped string: 
```
GitHub event=pull_request
Commit message=\[e2e\ prod\]\ Merge\ branch\ \'master\'\ into\ migrate-ecs-properties
```
I have changed `%q` to `%s `to generate a plain string.

bentsherman:
Closing for now until there is a clear need or the current settings get deprecated

TechIsCool:
I am disappointed to see this marked as closed, our internal platform has multiple AWS Batch/ECS workloads that could benefit from sidecar containers in the same task as the primary workload container. 

For us, this is not about preferring a newer API surface, it is about enabling multi-container task definitions for real use cases like mounting the newest version of AWS CLI, logging and observability patterns. Without ECS properties, these workloads either cannot be expressed.

PR #6271, opened 2025-07-12
Unify nf-lang config scopes with runtime classes

bentsherman:
This PR moves the config schema annotations into the runtime so that we don't have to update config options in both places.

Plugin config scopes are now defined in the plugins themselves. Core plugins are typically only loaded when they are actually used (i.e. the awsbatch executor), so when Nextflow validates config at runtime, it will skip validation of core plugin config if the core plugin isn't loaded. This allows the `lint` command to work the same as before.

I will update the language server to extract this metadata at build-time with a custom Gradle task. For third-party plugins we will need to publish this metadata to the plugin registry, but that can be done later.

I have removed several config options that were undocumented or appeared to be legacy options. We can preserve them if needed. See my comments below.

TODO:
- [x] Update tests
- [x] Publish JSON artifact for language server


christopher-hakkaart (commented):
Docs look good. No comments.

modules/nextflow/src/main/groovy/nextflow/container/ContainerConfig.groovy
```diff
@@ -17,129 +17,51 @@
...
-        this.sysEnv = env
+    default boolean canRunOciImage() {
+        return false
     }
 
-    boolean isEnabled() {
-        return get('enabled')?.toString() == 'true'
+    default boolean entrypointOverride() {
+        return ContainerHelper.entrypointOverride()
     }
```
bentsherman, line 35: The `entrypointOverride` is a hidden config option that can be set for `docker`, `k8s`, and the other container config scopes.

I made it so that it can only be enabled by setting the `NXF_CONTAINER_ENTRYPOINT_OVERRIDE` environment variable, since that seems simpler than adding a config option for every container scope.

If we want to preserve this behavior, I suggest we document this env var and add a note to the 25.10 migration notes.

modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
```diff
@@ -728,22 +733,17 @@ class BashWrapperBuilder {
...
-        }
 
-        if( containerConfig.containsKey('kill') )
-            builder.params(kill: containerConfig.kill)
+        if( containerConfig.getKill() != null )
+            builder.params(kill: containerConfig.getKill())
 
-        if( containerConfig.writableInputMounts==false )
+        if( containerConfig.writableInputMounts()==false )
             builder.params(readOnlyInputs: true)
```
bentsherman: It looks like some container scopes can specify `writableInputMounts = false` to mount folders into the container as read-only. There are tests for this option for apptainer, charliecloud, docker, podman, and singularity... so basically all of them.

If we want to keep this behavior, I suggest we add config option for each container scope where it's relevant. Or we could add an environment var.

plugins/nf-k8s/src/main/nextflow/k8s/K8sConfig.groovy
```diff
@@ -89,43 +248,30 @@ class K8sConfig implements Map<String,Object> {
...
-
-    K8sDebug getDebug() {
-        new K8sDebug( (Map<String,Object>)get('debug') )
+        pod.getAnnotations()
     }
 
     boolean getCleanup(boolean defValue=true) {
-        target.cleanup == null ? defValue : Boolean.valueOf( target.cleanup as String )
+        defValue
     }
```
bentsherman: There is a hidden option `k8s.cleanup` which is enabled by default, but can be disabled to not delete any pods. Normally only successful pods are deleted. Not sure if we still need this setting.

plugins/nf-k8s/src/main/nextflow/k8s/K8sConfig.groovy
```diff
@@ -89,43 +248,30 @@ class K8sConfig implements Map<String,Object> {
...
 
     String getUserName() {
-        target.userName ?: System.properties.get('user.name')
+        System.properties.get('user.name')
     }
```
bentsherman: There is a hidden config option `k8s.userName` to control the user name, which is used to set the default launchDir. Since we can set launchDir directly I'm not sure that we need this option.

plugins/nf-k8s/src/main/nextflow/k8s/K8sConfig.groovy
```diff
@@ -140,74 +286,36 @@ class K8sConfig implements Map<String,Object> {
...
 
     String getNextflowImageName() {
-        final defImage = "nextflow/nextflow:${BuildInfo.version}"
-        return target.navigate('nextflow.image', defImage)
-    }
-
-    boolean getAutoMountHostPaths() {
-        Boolean.valueOf( target.autoMountHostPaths as String )
+        return "nextflow/nextflow:${BuildInfo.version}"
     }
```
bentsherman, line 318: There is a hidden option `k8s.nextflow.image` which controls the nextflow container image used by `kuberun`. This can already be controlled via CLI option so I don't think we need a config option for it.

plugins/nf-k8s/src/main/nextflow/k8s/K8sConfig.groovy
```diff
@@ -217,50 +325,27 @@ class K8sConfig implements Map<String,Object> {
...
     ClientConfig getClient() {
 
-        final result = ( target.client instanceof Map
-                ? clientFromNextflow(target.client as Map, target.namespace as String, target.serviceAccount as String)
-                : clientDiscovery(target.context as String, target.namespace as String, target.serviceAccount as String)
-        )
-
-        if( target.httpConnectTimeout )
-            result.httpConnectTimeout = target.httpConnectTimeout as Duration
+        final result = clientDiscovery(context, namespace, serviceAccount)
```
bentsherman: There is a hidden scope `k8s.client` which can be used to define settings for the K8s client directly in Nextflow config. I think the best practice is to define these settings in `.kube/config`. If we want to keep this approach we should explicitly document it.

modules/nf-lineage/src/main/nextflow/lineage/cli/LinCommandImpl.groovy
```diff
@@ -193,8 +192,8 @@ class LinCommandImpl implements CmdLineage.LinCommand {
     }
 
     @Override
-    void check(ConfigMap config, List<String> args) {
-        final store = LinStoreFactory.getOrCreate(new Session(config))
+    void check(Map config, List<String> args) {
+        final store = LinStoreFactory.getOrCreate(LineageConfig.create(config), false)
```
bentsherman: @jorgee I refactored the `lineage` commands to not create a session since it was only used to create the `LineageConfig` and now isn't needed. However there are two failing unit tests:

```
LinCommandImplTest > should print error store is not found in diff FAILED
    org.spockframework.runtime.SpockComparisonFailure at LinCommandImplTest.groovy:467

LinCommandImplTest > should print correct validate path FAILED
    nextflow.exception.AbortOperationException at LinCommandImplTest.groovy:527
```

I wonder if it's a side effect of not creating the session? Do you know if the session is still needed (e.g. by the LID filesystem) for other purposes?

It would be nice to avoid creating a session since we aren't running a pipeline here, but if it requires deeper changes we could defer to a separate PR.

modules/nf-lineage/src/main/nextflow/lineage/cli/LinCommandImpl.groovy
```diff
@@ -51,9 +51,8 @@ class LinCommandImpl implements CmdLineage.LinCommand {
     private static final String ERR_NOT_LOADED = 'Error lineage store not loaded - Check Nextflow configuration'
     
     @Override
-    void list(ConfigMap config) {
-        final session = new Session(config)
-        final store = LinStoreFactory.getOrCreate(session)
+    void list(Map config) {
+        final store = LinStoreFactory.getOrCreate(LineageConfig.create(config), false)
```
jorgee: Why is this false? This is the cause of one of the failing tests. I think it should be true.

modules/nf-lineage/src/main/nextflow/lineage/LinStoreFactory.groovy
```diff
@@ -51,15 +51,14 @@ abstract class LinStoreFactory implements ExtensionPoint {
         return factory.newInstance(config)
     }
 
-    static LinStore getOrCreate(Session session) {
+    static LinStore getOrCreate(LineageConfig config, boolean requireConfig=true) {
```
jorgee: what's the purpose of `requireConfig`?

modules/nf-lineage/src/main/nextflow/lineage/LinStoreFactory.groovy
```diff
@@ -51,15 +51,14 @@ abstract class LinStoreFactory implements ExtensionPoint {
...
         if( instance || initialized )
             return instance
         synchronized (LinStoreFactory.class) {
             if( instance || initialized )
                 return instance
             initialized = true
-            final config = LineageConfig.create(session)
-            if( !config.enabled )
+            if( requireConfig && !config.enabled )
```
jorgee: I think this statement is not correct. If lineage is not enabled, it should return null for whatever `requireConfig` value.  When running the commands, `requireConfig` is `false`, and it always creates the config.

modules/nf-lineage/src/main/nextflow/lineage/cli/LinCommandImpl.groovy
jorgee: I had a look and the problems are not related with session. 

The first test is failing because the `requireConfig`. As it is`false` for the commands, it always creates the store and it is not printing the excepted error. The purpose of the test was to check if there wasn't configuration, lineage is not enabled and it should print the error. It seems with the changes, you are always enabling lineage in commands despite what it is disabled in the config. Is this what you wanted to do? If it is, we just need to modify the test. 

The second test is getting the default location `.lineage` instead of the location in the configuration. I am going to debug more to see which change is affecting this test.

jorgee: I was wrong, the error of the second test (`should print correct validate path`) is related to the session.  The filesystem is getting the configuration from the session, and this is why it is  using the default location. 

I have tried to replace the `create` call in the `LidFileSystem` constructor by `getOrCreate` but then several Lin FS tests are failing because they are reusing the store. I should modify them to reset the store at the end of each test. I think it is worth to do in another PR.

bentsherman: I reverted the lineage-related changes for now. I was attempting to make two improvements:
- allow the `lineage` command to be used without requiring `lineage.enabled = true`
- don't create a session in the `lineage` commands

My approach might have also affected the ability to specify a custom store location. Anyway, I will create a PR for these improvements later when we have time to look at them

pditommaso:
Got some tests failing locally, it looks like other in the CI. 

```
GraphObserverTest > should write an svg file FAILED
    groovy.lang.GroovyRuntimeException at GraphObserverTest.groovy:140

GraphObserverTest > should write a png file FAILED
    groovy.lang.GroovyRuntimeException at GraphObserverTest.groovy:158

GraphObserverTest > should write a pdf file FAILED
    groovy.lang.GroovyRuntimeException at GraphObserverTest.groovy:175
```

It would be great adding a detailed description about: 
- how the config object are expected to be defined 
- how the docs is expected to be provided in the config object  
- how the annotation are turned into markdone docs (are they?)
- how the annotation are turned into metadata 

This could be a great addition to the developer docs

bentsherman:
Looks like those graphviz tests were skipped for me because I didn't have graphviz installed, and the error wasn't caught by type checking because it wasn't enabled for the class. Should be resolved now, as well as the other failing CI test

I will add a page to the developer docs

bentsherman:
Tests are passing now and I added a page to the developer docs

I also added the JSON and Markdown renderers. They aren't used by Nextflow itself. The JSON renderer is used by the language server, but it will also be needed by plugins in the future so it makes sense to keep the core logic in Nextflow.

modules/nextflow/src/main/groovy/nextflow/container/ContainerConfig.groovy
pditommaso: Let's document it but report in any as deprecated feature

modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
pditommaso: Let's investigate in relation to https://github.com/nextflow-io/nextflow/pull/3590

plugins/nf-k8s/src/main/nextflow/k8s/K8sConfig.groovy
pditommaso: Let's keep and document properly

plugins/nf-k8s/src/main/nextflow/k8s/K8sConfig.groovy
pditommaso: Let's keep as well

plugins/nf-k8s/src/main/nextflow/k8s/K8sConfig.groovy
pditommaso: It may be useful for debugging, etc. It may be useful to keep it

modules/nextflow/src/main/groovy/nextflow/container/ContainerConfig.groovy
bentsherman: Related: https://github.com/nextflow-io/nextflow/commit/b3a4bf857cef0a8b16814baf9e13aa9296ca208a

modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
bentsherman: Looks like this was originally added for docker in these commits: https://github.com/nextflow-io/nextflow/commit/a6620d9a54396dc3cd4bb23a69f8b3c1dc33761a https://github.com/nextflow-io/nextflow/commit/35b79f187ebcb444c462f0fcb7e999f4909f862f

And was later added to charliecloud in #3590

The purpose was to specify read-only mounts separately from the work-dir mount, to prevent the container from mounting the common root and having access to e.g. an entire filesystem

bentsherman: I think I will add explicit config options for `writableInputMounts` just for docker and charliecloud for now, and if people request it for other containers we can add it

pditommaso:
Wonder why the config classes need to be specified in the plugin extension files?

bentsherman:
Because they extend `ConfigScope` which is an extension point. Then `ConfigValidator` uses the plugin system to find all config scopes

pditommaso:
Got it. Could not make sense to use classic Java service loader to avoid to maintain extension.idx files?

bentsherman:
I'm fine with using the ServiceLoader if it's something you want to apply to the rest of the codebase, otherwise I would rather stick to the existing system. Maybe the ServiceLoader approach should be investigated separately

pditommaso:
My rationale was more that config scope are not strictly plugins, however they may be part of a plugin. Therefore make sense to keep it as it is. 

Think several tests are still failing

bentsherman:
The CI failures are from the plugin registry

pditommaso:
let's disable that test(s) temporary

pditommaso (approved):
Great job! let's go

PR #6281, opened 2025-07-16
Fix submitted count when aborting submitted tasks

### LINKED ISSUE #6280: Submitted count is incorrect when aborting a run
When a run is aborted (e.g. cancelling a run from Seqera Platform), all submitted/running tasks are killed, marked as "aborted", and the workflow stats are updated accordingly:

https://github.com/nextflow-io/nextflow/blob/99f9edf1ba1359041176cf7688484fe15adcc45e/modules/nextflow/src/main/groovy/nextflow/trace/WorkflowStats.groovy#L348-L358

However, `markCompleted` always decrements the running count regardless of whether the completed task was running or submitted. If there are submitted tasks that were aborted, the submitted count will be too high and the running count will be too low (or even negative).

I see two solutions:

1. Include the previous state in the call to `WorkflowStats::markCompleted()` so that it knows which count to decrement
2. Add a function `markAborted()` and call it before updating the task state in TaskPollingMonitor, so that it can see the previous state.

bentsherman:
Close #6280

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/trace/WorkflowStats.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/trace/WorkflowStats.groovy
@@ -26,6 +26,7 @@ import groovy.util.logging.Slf4j
 import nextflow.processor.ErrorStrategy
 import nextflow.processor.TaskProcessor
 import nextflow.processor.TaskRun
+import nextflow.processor.TaskStatus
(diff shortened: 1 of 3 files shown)


swingingsimian:
@bentsherman also note: https://github.com/nextflow-io/nextflow/issues/6280#issuecomment-3083062000

jorgee (approved):

PR #6282, opened 2025-07-16
Fix formatting of channel factories

bentsherman:
This PR updates the formatter to work with channel factories using the `channel` namespace instead of the `Channel` type. It ensures that a channel factory called via `channel` is kept on a single line instead of being wrapped across two lines.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -811,8 +811,10 @@ class Session implements ISession {
             // dump threads status
             if( log.isTraceEnabled() )
                 log.trace(SysHelper.dumpThreads())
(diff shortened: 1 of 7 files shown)


bentsherman:
Reverting to draft. Need to move helper functions to `Types` instead of `ASTUtils`

jorgee:
Could you provide some context about the change in the description?

bentsherman:
updated

jorgee:
If I understood correctly,  in the case of` Channel `(that is deprecated), the formatting rule is not applied and the factory is splitted in a separate line. Is this the expected behaviour?

bentsherman:
yes, no need to maintain the old formatting rule since it's deprecated anyway

jorgee (approved):

PR #6302, opened 2025-07-23
Add manual install instructions to offline mode

ammaraziz:
This pull request adds extra information on how to manually download and install a plugin. Sometimes it's not possible to have access to an online machine with nextflow, therefore manually installing the plugin is required.


bentsherman:
Sorry, but we already have instructions for downloading plugins for offline use. If you can download and unzip a plugin, you can install it with nextflow on that machine and transfer the `.nextflow` directory to the offline machine.

This keeps you abstracted from the details of how plugins are managed under the hood, which could change and invalidate these instructions

ammaraziz:
I can see your point of view but I disagree with it. 

My situation was that I had a machine with internet but no ability to install nextflow and a machine without internet that had nextflow. Therefore the manual installation method I describe above was required.

In addition copying the plugins folder will override the contents of your existing installation. My method adds to it instead.

PR #6303, opened 2025-07-23
Fix new plugins registry download when OCI repository requires tokens

jorgee:
Some OCI repositories despite artifact are public they do not allow direct HTTP downloads. They require to acquire a temporal token to preform the download.

This PR extends the pf4j simple file download to allow this type of downloads

### DIFF
--- a/modules/nf-commons/src/main/nextflow/plugin/OciAwareFileDownloader.groovy
+++ b/modules/nf-commons/src/main/nextflow/plugin/OciAwareFileDownloader.groovy
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2013-2025, Seqera Labs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


bentsherman:
Are you able to test this with the real plugin registry? Even just a manual test for now would be fine

jorgee:
Yes, tested with https://cr.stage-seqera.io/ and https://public.cr.stage-seqera.io/. I was thinking to add a test with a real blob but I finally did it with wiremock because I was afraid that the blob could be removed and tests could start to fail and cause problems in the future. I could add it if there is an stable artifact. Maybe one of the core  plugins added in https://cr.stage-seqera.io/?

pditommaso:
Let's add an integration test using the current dev registry, truth that's a bit unstable but a good reminder to close it asap

modules/nf-commons/src/main/nextflow/plugin/OciAwareFileDownloader.groovy
```diff
@@ -0,0 +1,103 @@
...
+
+        String path = fileUrl.getPath();
+        String fileName = path.substring(path.lastIndexOf('/') + 1);
+        Path file = destination.resolve(fileName);
+        HttpURLConnection conn = (HttpURLConnection) fileUrl.openConnection()
+        conn.instanceFollowRedirects = true
+
+        if (conn.responseCode == HttpURLConnection.HTTP_UNAUTHORIZED) {
+            def wwwAuth = conn.getHeaderField("WWW-Authenticate")
+            if (wwwAuth?.contains("Bearer")) {
```
pditommaso, line 58: Note that this may also be `Basic`. It may be worth considering adopting the ORAS client to avoid to maintain all this burden. My initial assumption was to avoid that because the registry it was expected to not require any auth.

PR #6312, opened 2025-07-24
Register fully-qualified process names at compile-time

bentsherman:
This PR extends the strict parser to resolve fully-qualified process names via static analysis -- including conditional processes.

This should eliminate any false warnings when validating process selectors. Thus we can deprecate the `nextflow.enable.configProcessNamesValidation` feature flag and move towards requiring all feature flags to be defined in the script.

### DIFF
--- a/docs/reference/feature-flags.md
+++ b/docs/reference/feature-flags.md
@@ -2,26 +2,23 @@
 
 # Feature flags
 
-Feature flags are used to introduce experimental or other opt-in features. They can be specified in the pipeline script or the configuration file.
+Feature flags are used to introduce experimental or other opt-in features. They must be specified in the pipeline script.
 
 `nextflow.enable.configProcessNamesValidation`
+: :::{deprecated} 25.10.0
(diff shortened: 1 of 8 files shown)


docs/reference/feature-flags.md
```diff
@@ -2,26 +2,23 @@
 
 # Feature flags
 
-Feature flags are used to introduce experimental or other opt-in features. They can be specified in the pipeline script or the configuration file.
+Feature flags are used to introduce experimental or other opt-in features. They must be specified in the pipeline script.
 
 `nextflow.enable.configProcessNamesValidation`
+: :::{deprecated} 25.10.0
+  Use the {ref}`strict syntax <strict-syntax-page>` instead. It will validate process selectors without producing false warnings.
```
christopher-hakkaart:
```suggestion
  Use the {ref}`strict syntax <strict-syntax-page>` instead. It validates process selectors without producing false warnings.
```

christopher-hakkaart (approved):
One minor suggestion. Otherwise docs look great.

PR #6313, opened 2025-07-24
Improve Fusion license error message

bentsherman:
This PR clarifies the error message when a platform access token is not provided, that it is required when using Fusion, with a link to the licensing docs.

Consider also lowering this error to a warning, if not for `master`, at least for previous releases that were backported.

### DIFF
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionToken.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionToken.groovy
@@ -107,7 +107,7 @@ class TowerFusionToken implements FusionToken {
         if( !endpoint )
             throw new IllegalArgumentException("Missing Seqera Platform endpoint")
         if( !accessToken )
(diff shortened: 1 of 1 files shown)


drewdipalma:
The updated error message looks good to me.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerFusionToken.groovy
```diff
@@ -107,7 +107,7 @@ class TowerFusionToken implements FusionToken {
         if( !endpoint )
             throw new IllegalArgumentException("Missing Seqera Platform endpoint")
         if( !accessToken )
-            throw new IllegalArgumentException("Missing Seqera Platform access token")
+            throw new IllegalArgumentException("A Seqera Platform access token is required to use Fusion -- see https://docs.seqera.io/fusion/licensing for more information")
```
pditommaso:
```suggestion
            throw new IllegalArgumentException("Seqera Platform access token is required to use Fusion -- see https://docs.seqera.io/fusion/licensing for more information")
```

PR #6314, opened 2025-07-28
Solve Job arrays exceeding queuesize

### LINKED ISSUE #5920: Job Array with queueSize exceeds queueSize

jorgee:
close #5920

Modifying canSubmit to avoid to avoid exceed executor.queueSize parameter.
Printing warning when array size exceeds the queueSize. This can make the task array 'unsubmitable' (Not sure if we should abort the execution in this situation)
Add unit tests
Tested with this pipeline

```
process test {
    array 10 
    input:
    val x

    """
    echo $x
    """
}

workflow {
   Channel.of(1..20) | test
}
```
With a config with awsbatch executor and different executor.queueSize values:

size 15: Only one job submitted at a time in AWS Batch.
size 5: Prints a warning but run continues without the task array jobs submitted.


bentsherman (approved):

pditommaso:
I went for a simpler solution for this problem https://github.com/nextflow-io/nextflow/pull/6345

bentsherman:
Your PR does not update the forks count correctly. This PR should be restored

PR #6316, opened 2025-07-29
Update to plugin registry V1 

jorgee:
This PR includes two changes to integrate Nextflow with plugin registry updates:
1. Update plugins download to the  plugin registry API v1
`POST /plugins/collect` ->  `GET /v1/plugins/dependencies` 
2. Update OciAwareFileDownloader:
 - Fix automatic redirection problems (changing protocol http-> https and getting the correct filename) by doing a manual redirection.
- Refactor with Java HTTP client


jorgee:
@pditommaso and @bentsherman 
When doing this change, I had a doubt about how to manage the api path inside nextflow. Currently, we require to set the `NXF_PLUGINS_INDEX_URL=<plugins-host>/api` and Nextflow adds the path `/plugins/collect` to get the plugins. 

For the changes for v1, I have added the 'v1' in the path to do not change the environment variable. However, there are other options to manage it:
1-  `NXF_PLUGINS_INDEX_URL=<plugins-host>/api/v1` only add `/plugins/...`
2-  `NXF_PLUGINS_INDEX_URL=<plugins-host>` and nextflow add the full api path `/api/v1/....` 

I think (2) makes more sense, but I would like to know your opinion. 

The same is happening for the gradle plugin and the nf-plugin-template. Depending on what we include in the URL, we need to update the template, the plugin or both.

pditommaso:
The "correct" way it to keep as parameter api root path ie. `https://<HOST>/api` (without slash) and append `/v1/etc`

jorgee:
> The "correct" way it to keep as parameter api root path ie. `https://<HOST>/api` (without slash) and append `/v1/etc`

Then, this PR does not require more changes and I will create a PR in the gradle plugin to support v1.

pditommaso:
Worth adding a retry logic e.g 

https://github.com/nextflow-io/nextflow/blob/9a70cdbb54b9cd972b9e32e0404b9d22b7856b24/modules/nextflow/src/main/groovy/nextflow/scm/RepositoryProvider.groovy#L497-L497

I've made a little library to wrap the FailSafe boilerplate into a reusable component, see here https://github.com/seqeralabs/libseqera/tree/master/lib-retry

jorgee:
There is already a retry logic with FailSafe [here](https://github.com/nextflow-io/nextflow/blame/adc87cc94f1936e5a5febbbc65f5ba48696731fe/modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy#L120-L124). I see it is not using the common RetryConfig, defaults are hardcoded. I will modify to use it. Or do you prefer to replace this retry logic by lib-retry?

pditommaso:
Maybe i'm missing something, what about these sends 

https://github.com/nextflow-io/nextflow/blob/adc87cc94f1936e5a5febbbc65f5ba48696731fe/modules/nf-commons/src/main/nextflow/plugin/OciAwareFileDownloader.groovy#L66-L66

https://github.com/nextflow-io/nextflow/blob/adc87cc94f1936e5a5febbbc65f5ba48696731fe/modules/nf-commons/src/main/nextflow/plugin/OciAwareFileDownloader.groovy#L129-L129

jorgee:
Ah! yes, you are right. I was just looking at the fetch part.

pditommaso:
Then let's adopt the new lib consistently at least here

jorgee:
I have created a common `RetryableHttpClient` that wraps the Java Http Client with the Sequera lib-retry lib. I have used it in the fetch call in `HttpPluginRepository` and the download calls in `OciAwareFileDownloader`. With this change, all calls are retried for the retriable error codes (the old fail-safe mechanism in fetch was not doing it). 

The retry configuration for these calls is obtained from the common retry configuration `nextflow.retryConfig` using RetryConfig. As it is not in the nf-commons package and it depends on other `nextflow `module classes, I have created the `IRetryConfig` interface that is implemented by `RetryConfig` and I have also added a new method in `ISession` to get the common retry configuration. 

I tried to use the lib-retry `Retryable.Config` interface instead of using the `IRetryConfig`, but `RetryConfig` was using `nextflow.utils. Duration` instead of Java Duration and I wasn't sure if changing the type could introduce side effects, so I discarded this option.

modules/nf-commons/src/main/nextflow/ISession.groovy
```diff
@@ -66,4 +68,7 @@ interface ISession {
 
     String getCommitId()
 
+    IRetryConfig getCommonRetryConfig()
```
bentsherman: Instead of adding this method, just use `getConfig()` and access the common retry settings as:
```groovy
new RetryConfig( config.navigate('nextflow.retryPolicy') as Map ?: Collections.emptyMap() ) 
```

bentsherman: Actually you can just do `RetryConfig.config()`

modules/nf-commons/src/main/nextflow/util/IRetryConfig.groovy
bentsherman, line 1: If the long-term goal is to use lib-retry throughout the codebase, then all of the retry config classes could be refactored to use Java Duration instead of Nextflow Duration. The constructor would expect a Nextflow Duration and convert to a Java Duration. Perhaps you could do that here and have `RetryConfig` implement `Retryable.Config` directly without the `IRetryConfig` interface

bentsherman (changes requested):
Please allow me to review the config classes before merging. I need to make sure everything will work with the config refactor in #6271

modules/nf-commons/src/main/nextflow/ISession.groovy
jorgee: `RetryConfig `is not in the `nf-commons `module (the module where the plugin classes are). It is the `nextflow` module. It also imports a class from `nextflow.config`. I tried to avoid moving the class. If you consider that it is better to move than implement the interface, I can check how complicated it is to avoid this import.

bentsherman: I see now. I think I would prefer to just move the retryable http client into `nextflow`. Once we move the CLI into its own module, we might just merge `nextflow` and `nf-commons` anyway

bentsherman: I didn't realize the plugin code was also in nf-commons. In that case I think we should just move `RetryConfig` into nf-commons

jorgee: I have removed `IRetryConfig` and moved `RetryConfig` to `nf-commons` as well as some imported classes `ConfigHelper` and ` ConfigClosurePlaceHolder`. I also needed to make small changes because extensions from `FileEx` and `Bolts` were not detected in `nf-commons.`

bentsherman: It might be easier to just move `valueOf` and `toType` into `RetryConfig` for now. Those methods were added with RetryConfig and are only used by it. It we need to generalize we can do that later

jorgee: What about moving them to Bolts? `valueOf` looks like an extension of `Map` such as `navigate` but falling back to SysEnv variable before returning the default.

bentsherman: That could work. In that case I would rename it to something like `navigateWithEnv` since `valueOf` has a different meaning for other classes (e.g. `String.valueOf()`)

jorgee: When testing the latest changes with `RetryConfig`, I have noted that the session is not created when fetching the plugins, so we can not get the `nextflow.retry` configuration. 
```
WARN: Missing nextflow session - using default retry config
Launching `https://github.com/nextflow-io/hello` [gloomy_allen] DSL2 - revision: 2ce0b0e294 [master]

Downloading plugin nf-hello@0.5.0
WARN: Missing nextflow session - using default retry config
WARN: Missing nextflow session - using default retry config
```
Therefore, the `IRetryConfig` and the `ISession.getCommonRetryConfig` added for this purpose do not make sense. Passing the config until the `HttpPluginRepository` requires a lot of changes. As this PR is to update to the V1 endpoint of the plugin registry, I have reverted the changes for getting the common retry configuration. The plugin registry calls are now configured with the lib-retry `Retryable.Config` defaults. This will make the PR  smaller, which is better to backport to previous versions.

pditommaso:
### Latest Updates - Plugin HTTP Infrastructure Refactoring

#### Key Changes

**RetryConfig Enhancement & Movement:**
- Move RetryConfig from nextflow to nf-commons module for better reusability
- Add support for multiplier parameter in retry configuration
- Replace navigate() dependency with static compilation-friendly getNestedConfig() helper
- Add NXF_RETRY_POLICY_MULTIPLIER environment variable (default: 2.0)

**HTTP Client Modernization:**
- Replace custom HttpRetryableClient with external io.seqera:lib-httpx library
- Update HttpPluginRepository to use HxClient for better reliability
- Enhanced OciAwareFileDownloader with improved authentication and error handling

**Documentation:**
- Add NXF_PLUGINS_REGISTRY_URL and NXF_RETRY_POLICY_MULTIPLIER to env-vars.md
- Document retry policy multiplier configuration for exponential backoff

**Dependencies:**
- Add io.seqera:lib-httpx:1.0.0 and io.seqera:lib-retry:1.2.0
- Remove custom HTTP retry implementation in favor of external libraries

This refactoring improves the plugin system's HTTP infrastructure reliability while maintaining backward compatibility and adding new configuration flexibility.

PR #6321, opened 2025-07-31
Add s3 checksum algorithm

jorgee:
Add checksum algorithm to upload and copy S3 class. (#3585)

Notes:
- aws CLI only supports --checksum-algorithm in high-level calls cp, mv, sync in versions 2.18+ 
- S5cmd doesn't support the definition of checksum algorithm
- Fusion doesn't support checksum algorithm ( I haven't seen a FUSION_AWS_xx env var for defining the checksum)


bentsherman:
Closing since this was blocked by Fusion config. We can re-open once that is implemented

PR #6323, opened 2025-08-02
Reset terminal color after lint success/failure message

### LINKED ISSUE #6127: nextflow lint success message does not reset color
After running `nextflow lint` on a pipeline throwing zero errors, the green color from the success message persists in subsequent console lines.

nextflow-25.04.2-dist, java openjdk-17.0.3, rocky linux 9, bash v5.1.8

samuell:
This is a suggested fix for #6127 

I have not managed to build and test this with Gradle just yet, but seems it should work based on surrounding code, which does append the `.reset()` after changing color.


bentsherman:
It might be safer just to reset before the final print instead of handling each case:

```groovy
        term.reset()
        AnsiConsole.out.print(term)
        AnsiConsole.out.flush()
```

Does this also work?

pditommaso (approved):
It looks safe. Thanks for the contribution

PR #6324, opened 2025-08-02
Check for duplicate process/workflow calls in strict syntax

bentsherman:
This PR adds the check for duplicate process/workflow calls to the strict syntax.

Currently the check is performed at runtime:

https://github.com/nextflow-io/nextflow/blob/6f6a4cd22a275ddfba062090aaf9401e96d313da/modules/nextflow/src/main/groovy/nextflow/script/BindableDef.groovy#L43-L47

With the strict syntax, we can now catch this error much earlier at compile-time.

It looks like the runtime check only applies to duplicate process calls. There is no equivalent check for duplicate workflow calls, but I believe there should be, since duplicate workflows would lead to processes with the same fully-qualified name.

### DIFF
--- a/modules/nf-lang/src/main/java/nextflow/script/control/VariableScopeVisitor.java
+++ b/modules/nf-lang/src/main/java/nextflow/script/control/VariableScopeVisitor.java
@@ -16,7 +16,9 @@
 package nextflow.script.control;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import nextflow.script.ast.ASTNodeMarker;
 import nextflow.script.ast.AssignmentExpression;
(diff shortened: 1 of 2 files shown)


jorgee:
I have seen it is just checking if the process has been visited before in the workflow. In cases where the process is called inside an if-else statement, it will produce an error and it should be valid.
The following workflow should be valid and it is getting an error (If I correctly tested it)

```
nextflow.enable.strict=true
params.a = 1
process hello {
                script:
                """
                echo hello
                """
}
workflow {
      if (params.a == 1){
                hello()
      }else{
                hello()
      }
}
```
 When running with the PR: 
```
$ NXF_SYNTAX_PARSER=v2 ./launch.sh run a.nf

 N E X T F L O W   ~  version 25.06.0-edge

Launching `a.nf` [thirsty_lalande] DSL2 - revision: bb1409e71b

Error a.nf:13:3: Process `hello` cannot be called more than once in the same workflow -- use an alias instead
│  13 | 		hello()
╰     |   ^^^^^^^


ERROR ~ Script compilation failed
```

With the latest edge is working
 
```
$ NXF_VER=25.07.0-edge NXF_SYNTAX_PARSER=v2 nextflow run a.nf

 N E X T F L O W   ~  version 25.07.0-edge

Launching `a.nf` [small_hoover] DSL2 - revision: bb1409e71b

executor >  local (1)
[29/7a91f4] hello | 1 of 1 ✔
```

bentsherman:
Good catch. I would need to implement some live-variable analysis to do this properly in static analysis. Not worth doing right now. Thanks for testing

PR #6330, opened 2025-08-06
Fix GString equality checks with String

bentsherman:
This PR improves the semantics of GStrings and lazy expressions in the strict syntax.

- GStrings are coerced into Strings by appending `.toString()` under the hood, so that equality checks like `"${'test'}" == 'test'` are correct.

- Lazy expressions in process inputs/outputs/directives are handled more consistently. Before, only GStrings and property expressions were lazily evaluated. Now, any expression that references a variable is wrapped in a closure.


docs/process.md
```diff
@@ -1276,14 +1276,19 @@ All directives can be assigned a dynamic value except the following:
...
 ```nextflow
-queue "${ entries > 100 ? 'long' : 'short' }"
+queue (entries > 100 ? 'long' : 'short')
 ```
 
-Note, however, that the latter syntax can be used both for a directive's main argument (as in the above example) and for a directive's optional named attributes, whereas the closure syntax is only resolved dynamically for a directive's main argument.
+Nextflow will evaluate this directive dynamically if it references task inputs. Directives that use an explicit closure are still resolved dynamically.
+
+:::{note}
+Process configuration options must still be specified with a closure in order to be dynamic.
```
```suggestion
Process configuration options must be specified with a closure to be dynamic.
```

docs/process.md
```diff
@@ -1276,14 +1276,19 @@ All directives can be assigned a dynamic value except the following:
 - {ref}`process-label`
 - {ref}`process-maxforks`
 
-:::{tip}
-Assigning a string value with one or more variables is always resolved in a dynamic manner, and therefore is equivalent to the above syntax. For example, the above directive can also be written as:
+:::{versionadded} 25.10
+:::
+
+Dynamic directives do not need to be wrapped in a closure when using the {ref}`strict syntax <strict-syntax-page>`:
```
```suggestion
Dynamic directives do not need to be wrapped in closures when using the {ref}`strict syntax <strict-syntax-page>`:
```
Treating Dynamic directives as plural

docs/process.md
```diff
@@ -1276,14 +1276,19 @@ All directives can be assigned a dynamic value except the following:
...
 
 ```nextflow
-queue "${ entries > 100 ? 'long' : 'short' }"
+queue (entries > 100 ? 'long' : 'short')
 ```
 
-Note, however, that the latter syntax can be used both for a directive's main argument (as in the above example) and for a directive's optional named attributes, whereas the closure syntax is only resolved dynamically for a directive's main argument.
+Nextflow will evaluate this directive dynamically if it references task inputs. Directives that use an explicit closure are still resolved dynamically.
```
```suggestion
Nextflow will evaluate directives dynamically if they reference task inputs. Directives that use an explicit closure are still resolved dynamically.
```

christopher-hakkaart (approved):
A couple of non-blocking suggestions. Please check meaning is retained.

PR #6332, opened 2025-08-07
WIP: Capture all comments in parser

ErikDanielsson:
This draft PR aims to add full support for finding comments in Nextflow source code and attaching them to the appropriate AST node. The additions to the parser are mirrored by the formatter classes used by the language server with the goal of addressing issues with the formatter removing comments.


bentsherman:
Notes from today's meeting

We will investigate another approach in which the CST is converted directly into a formatting IR and used by the formatter.

We can also link CST nodes to AST nodes as needed to retrieve semantic information. But in the first iteration let's see how far we can get with just CST -> formatting IR -> formatter.

bentsherman:
Closing in favor of #7346

Issue #1526, opened 2020-03-10
nix/guix environments


Emiller88:
## New feature

Hello! I'm thinking about implementing nix/guix environments for a process.

Similar to `conda`, `docker`, etc. 

I found this issue, I think the problem with it was the actual building of nextflow.

https://github.com/nextflow-io/nextflow/issues/1395

This is not for the nextflow builds but the individual process

## Usage scenario 

```nextflow
process foo {
  nix-env 'bwa samtools multiqc'

  '''
  your_command --here
  '''
}
```
## Suggest implementation 

I'll need to look under the hood at the conda implementation.

pditommaso:
How the linked issue is related to this requirement?

Emiller88:
I was explaining how the two were different issues, since the last issue was closed. And to show that I couldn't find a previous issue for this topic.

pditommaso:
Looks stalled. Closing.

lindenb:
Can you please reopen this?

The support for guix would be very useful.  Our sysadmins told us that the "conda partition" is full, they'd like us to use guix instead of the conda/mamba/micromamba ecosystem.

pditommaso:
The use of containers would not be an option for you?

lindenb:
@pditommaso well, our server doesn't allow us to build apptainer images , so we have to build the images on our personal laptop and then we have upload the image on the server. That's not an ideal solution when you want to explore/test things. Meanwhile guix looks very fast.

I don't have much experience with guix. For now I just know how to install a package. But it looks like much more powerful than conda and, as far as I understand, you can "code" your package with a scheme/lisp syntax, and generate a package just like a git project (?) with diff/add/remove etc....

Our sysadmins love this system. They used to work with physicists but now we - the biologists -  are coming with our conda-stuff and it looks like it's a problem.

PR #6342, opened 2025-08-14
Package directive

edmundmiller:
PoC of a package directive.

Moved conda into a plugin, and the Pixi code from #6157. Hoping that by making them plugins, we can support things like #1526 in the community, etc.

I also wanted to keep `conda` for now(it might throw a deprecation, but it shouldn't yet) and tried to gate it behind a feature flag.


docs/package.md
```diff
@@ -0,0 +1,369 @@
...
+```
+
+## Supported Providers
+
+### Conda
+
+The conda provider supports:
+- Package specifications with version constraints
+- Custom channels
+- Environment files (`.yml`, `.yaml`)
```
pditommaso, line 178: I'm tempted to get rid of env files with the new directive. Tool requirements should be strictly the ones required by the task, not bloated env files

docs/conda.md
```diff
@@ -191,6 +195,49 @@ process hello {
...
+```groovy
+// nextflow.config
+nextflow.preview.package = true
+```
+
+The unified system provides:
+- Consistent interface across different package managers
+- Plugin-based architecture for extensibility
+- Better integration with containerization platforms
+- Support for multiple package managers (conda, pixi, etc.)
```
pditommaso, line 239: Think we should be explicit about which package managers are supported. ideally also pak or install2.r for R

pditommaso (commented):
Well done, seems going in the right direction. need to look more in deep for some low level details

docs/conda.md
edmundmiller: Agreed! Do you just want to support the package managers that wave supports?

I also thought this might be a nice abstraction for the core container directives, but I wanted to keep the scope tight.

pditommaso: Yep. we should keep aligned with Wave

docs/package.md
edmundmiller: Which part of env files do you not like? 
1. The extra file
2. The separation of the packages from the process

People having bloated Nextflow processes is just a part of migrating old scripts to Nextflow. Because their scripts, prior the "business" logic, was mixed with the actual implementation logic, it can be a pain to separate out.

Task batching or grouping would help people organize the logic; sometimes, it's a performance concern, like piping BWA into Samtools. 

I personally like the environment.yamls, and I find once you get over three packages(which happens in an R script) and add in the channels, you can't read the `conda` directive. 😅 


So, compromise, what if we made the environment.yml magically picked up like Dockerfiles are with wave? I just want to get rid of the `conda "${moduleDir/environment.yml}"` line.

docs/conda.md
edmundmiller: Started to implement this, but was waiting on the wave support, so we'll just make a separate follow-up PR to add the wave package management after [wave settles](https://github.com/seqeralabs/wave/issues/873).

docs/package.md
thealanjason: I would also suggest keeping the environment.yml. 
Reasoning: Within the bioinformatics world, there might be well developed tools and each process might need only one or two of them. 

However, we are currently using nextflow as a workflow orchestration tool outside the in a computational modeling and uncertainty quantification space. Here, some `process` are developed by individuals in on their own first and then integrated into the workflow. In this case the individual builds up their `environment.yml` file and it would be easy to carry this over directly into the workflow. Converting each `environment.yml` to the style of the package description could be a pain in the neck.

bentsherman:
Closing since this effort lost momentum. Feel free to re-open if you decide to pick it back up

PR #6345, opened 2025-08-14
Fix Job arrays exceeding queueSize (#5920)

### LINKED ISSUE #5920: Job Array with queueSize exceeds queueSize

pditommaso:
## Summary

Fixes issue where job arrays bypassed `executor.queueSize` limits, causing resource overallocation. Previously, a job array would count as only 1 task against the queue limit regardless of its actual array size.

**Before:** `process.array = 30` + `executor.queueSize = 40` → up to 70 concurrent tasks  
**After:** `process.array = 30` + `executor.queueSize = 40` → max 40 concurrent tasks (respects queue limit)

## Root Cause

The original `canSubmit()` method only checked `runningQueue.size() < capacity`, treating job arrays as single tasks. This allowed arrays to exceed intended resource limits.

### Core Changes
- **Queue Slot Accounting**: Job arrays now consume slots equal to their array size
- **Validation**: Throws `IllegalArgumentException` if array size exceeds total queue capacity  
- **Clean Architecture**: Extracted complex logic into `checkQueueCapacity()` helper while preserving original ternary structure

### Implementation Details
The `checkQueueCapacity()` method:
1. Calculates task slots (1 for regular tasks, array size for job arrays)
2. Validates array size doesn't exceed total capacity
3. Checks if remaining capacity can accommodate the task

## Test Coverage

Added comprehensive tests using Spock's parameterized testing with 12 new test cases covering:
- Exception validation for oversized arrays (3 test cases)
- Queue capacity accounting with various scenarios (9 test cases)
- Inheritance verification for ParallelPollingMonitor

## Backward Compatibility

✅ **Fully backward compatible** - no breaking changes:
- Regular tasks continue to work unchanged (consume 1 slot)
- Unlimited capacity (`executor.queueSize = 0`) works as before
- All existing configurations remain valid
- Only affects oversized job arrays that were previously misconfigured

## Related

- Closes #5920  
- Related to reverted PR #6314 (this implementation avoids the issues that caused the revert)

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy
@@ -198,6 +198,47 @@ class TaskPollingMonitor implements TaskMonitor {
 
 
     /**
+     * Get the number of queue slots consumed by a task handler.
+     * Regular tasks consume 1 slot, job arrays consume slots equal to their array size.
+     *
+     * @param handler A {@link TaskHandler} for the task
+     * @return The number of queue slots consumed by this task
+     */
+    private int getTaskSlots(TaskHandler handler) {
+        return handler.task instanceof TaskArrayRun ?
+            ((TaskArrayRun) handler.task).getArraySize() : 1
+    }
+
+    /**
+     * Validates that a task handler can be submitted based on queue capacity constraints.
+     * Performs validation for job arrays that exceed queue capacity.
+     *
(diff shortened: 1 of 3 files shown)


TheBil99:
Hey! I am not sure this has already been discussed elsewhere, but does the executor.queueSize work when the array size is larger than the queue size? For instance say i have an array of 1000 and executor.queueSize is 10, shouldn't queue size work similarly to a throttle in slurm? I tried it and it didn't work, I was wondering if you have already something in place for this case. 

Thank you!

PR #6346, opened 2025-08-15
BREAKING CHANGE: Improve eval output hash with semantic names instead of raw commands

### LINKED ISSUE #5470: Pipeline uses cached process despite edited eval script

pditommaso:
## Summary

This PR fixes issue #5470 by implementing a more robust approach to including eval output commands in task hash calculation. The solution uses semantic parameter names paired with command values, creating a symmetric pattern with input parameter hashing.

## Problem Statement

Issue #5470 identified that changes to eval output commands were not properly invalidating task cache, leading to incorrect cached results. The previous fix (commit b0fe0a9d) was reverted because it included raw bash commands directly in the hash.

## Key Improvements Over Reverted Approach

- **Semantic names**: Uses `nxf_out_eval_*` parameter names instead of raw bash commands for better readability
- **Symmetric pattern**: Follows the same `name + value` approach as input parameter hashing  
- **Deterministic ordering**: Sorts eval outputs by name for consistent hash generation across JVM runs
- **Testable architecture**: Separates logic into dedicated `computeEvalOutputsContent()` method
- **Better documentation**: Comprehensive comments explaining rationale and implementation

## Breaking Change Notice

⚠️ **BREAKING CHANGE**: This change will invalidate existing task cache entries that use output eval parameters, requiring re-execution of those tasks.

The cache invalidation is **intentional and necessary** to ensure proper cache behavior when eval output definitions change.

## Code Changes

- Added eval output hash calculation in `TaskProcessor.createTaskHashKey()` 
- Implemented `computeEvalOutputsContent()` method with sorting and proper formatting
- Added comprehensive unit tests for deterministic behavior verification

Fixes #5470

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -2211,6 +2211,13 @@ class TaskProcessor {
             keys.add( it.value )
         }
 
+        // add eval output commands to the hash for proper cache invalidation (fixes issue #5470)
+        final outEvals = task.getOutputEvals()
+        if( outEvals ) {
+            keys.add("eval_outputs")
+            keys.add(computeEvalOutputsContent(outEvals))
+        }
+
         // add all variable references in the task script but not declared as input/output
         def vars = getTaskGlobalVars(task)
         if( vars ) {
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -2614,4 +2621,42 @@ class TaskProcessor {
...
+        
+        final result = new StringBuilder()
+        
+        // Sort entries by key for deterministic ordering. This ensures that the same set of
+        // eval outputs always produces the same hash regardless of map iteration order,
+        // which is critical for cache consistency across different JVM runs.
+        // Without sorting, HashMap iteration order can vary between executions, leading to
+        // different cache keys for identical eval output configurations and causing
+        // unnecessary cache misses and task re-execution
+        final sortedEntries = outEvals.entrySet().sort { a, b -> a.key.compareTo(b.key) }
```
bentsherman, line 2645: @pditommaso instead of sorting the entries, it would be better to hash them as an unordered collection (e.g. `ArrayBag`)

pditommaso: What would be the difference?

bentsherman: It would just be a more efficient hash calculation since you wouldn't need to sort the entries. Like we did with the directory hash

pditommaso: Considering the average number of entries (< 10) not sure it's something it could have an impact, but not against that. Feel free to make a PR if you think it's important.

PR #6358, opened 2025-08-21
install.md - SDKMan install fix in

rpomaris:
Need to initialize SDKMan to use; otherwise Java install doesn't work

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -27,13 +27,15 @@ If you don't have a compatible version of Java installed, it is recommended that
 
 To install Java with SDKMAN:
 
-1. [Install SDKMAN](https://sdkman.io/install):
+1. [Install and Initialize SDKMAN](https://sdkman.io/install):
 
     ```{code-block} bash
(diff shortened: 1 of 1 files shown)


bentsherman:
This requirement is fulfilled by step 2 -- "open a new terminal"

rpomaris:
@bentsherman - thanks for looking into this. Opening a new terminal did not resolve the issue for me. The [SDKMan install instructions](https://sdkman.io/install/) has users execute the Initialization step. This worked for me.

bentsherman:
That may be an issue/incompatibility between your environment and sdkman. The sdkman installer says this at the end:
```
Please open a new terminal, or run the following in the existing one:

    source "/home/bent/.sdkman/bin/sdkman-init.sh"
```

Indicating that opening a new terminal should work. I suppose it could be a bug with sdkman

rpomaris:
Ah, got it. Thanks. I may have missed that in the output from the installation.

PR #6359, opened 2025-08-21
Fix false warning for map config options

bentsherman:
This PR fixes a false warning reported by the strict config parser for config options that take a map, such as `process.resourceLimits`:

```
WARN: Unrecognized config option 'process.resourceLimits.cpus'
```

It has to do with how the config object is flattened. Instead, the validator should recognize it as a map property of the config option `process.resourceLimits`.


modules/nextflow/src/main/groovy/nextflow/config/ConfigValidator.groovy
```diff
@@ -155,6 +157,32 @@ class ConfigValidator {
...
+     * property thereof.
+     *
+     * @param names
+     */
+    private boolean isMapOption(List<String> names) {
+        return isMapOption0(SchemaNode.ROOT, names)
+            || isMapOption0(pluginScopes, names)
+    }
+
+    private static boolean isMapOption0(SchemaNode.Scope scope, List<String> names) {
```
jorgee, line 171: It is difficult to understand what this method is doing without some context. Is `names` providing the path to get the property and it is navigating through the schema nodes starting from `scope` ? Maybe a comment will help to better understand the method.

bentsherman: `names` is the config option split into individual names, e.g. `process.resourceLimits` -> `[process, resourceLimits]`

modules/nextflow/src/main/groovy/nextflow/config/ConfigValidator.groovy
```diff
@@ -155,6 +157,32 @@ class ConfigValidator {
             && !pluginScopes.children().containsKey(name)
     }
 
+    /**
+     * Determine whether a config option is a map option or a
+     * property thereof.
+     *
+     * @param names
```
jorgee:
```suggestion
     * @param names Config option split into individual names, e.g. 'process.resourceLimits' -> [process, resourceLimits]
```

modules/nextflow/src/main/groovy/nextflow/config/ConfigValidator.groovy
```diff
@@ -155,6 +157,32 @@ class ConfigValidator {
...
+            || isMapOption0(pluginScopes, names)
+    }
+
+    private static boolean isMapOption0(SchemaNode.Scope scope, List<String> names) {
+        SchemaNode node = scope
+        for( final name : names ) {
+            if( node instanceof SchemaNode.Scope )
+                node = node.children().get(name)
+            else if( node instanceof SchemaNode.Placeholder )
+                node = node.scope()
```
pditommaso, line 177: what is using `node` after it has been assigned?

bentsherman: the next iteration of the loop

pditommaso: Got it

pditommaso (approved):

PR #6364, opened 2025-08-26
Fix unstage controls in command.run when using storeDir

### LINKED ISSUE #6311: storeDir : .command.out' are the same file
Hi,
I found a bug in nextflow v25.04.3.

I get these errors in the log of many processes:

cp: '.command.out' and '/scratch/analysis/work/61/9a68b6a99e3766137c31971f72db9a/.command.out' are the same file
cp: '.command.err' and '/scratch/analysis/work/61/9a68b6a99e3766137c31971f72db9a/.command.err' are the same file

The cp commands in the .command.run script are doing this. Not sure why.

jorgee:
close #6311 

`unstage_controls` was set in .`command.run` using the same condition as in `unstage_outputs`. It was producing an "are the same file" error message in `.command.log` when using the `storeDir`. 
By default `targetDir` is set to the working directory, but when user defines a `storeDir `, the `targetDir` is set to `storeDir`. The condition `shouldUnstageOutputs` checks if `targetDir` is different from the working directory, and the fucntion to generate the `unstage_outputs` create the commands to copy the outputs to `targetDir`. However, control logs (.command.err, .command.out, .command.trace and .command.env) are always copied to the working directory using the function `copyFileToWorkDir`. For this reason, the same condition cannot be applied.

This PR changes the condition when the `unstage_controls` is set. A new method `shouldUnstageControls` is created to differentiate the condition when checking the outputs or controls. By default, it is set to false, the unstage controls will be applied if scratch is defined. There is a special case for Azure, where both controls and outputs are forced to be unstaged

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -160,6 +160,14 @@ class BashWrapperBuilder {
         return targetDir && workDir!=targetDir
     }
 
+    /**
+     * Template method that allows controlling if it's required to unstage
+     * task control files (.command.out, .command.err, .command.trace, .command.env)
+     *
+     * See also https://github.com/nextflow-io/nextflow/pull/6364
+     * 
+     * @return false by default; executors may override to implement their own logic
+     */
     protected boolean shouldUnstageControls() {
         return false
     }


modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
```diff
@@ -160,6 +160,10 @@ class BashWrapperBuilder {
         return targetDir && workDir!=targetDir
     }
 
+    protected boolean shouldUnstageControls() {
+        return false
```
pditommaso, line 172: I wonder if this should not default to `shouldUnstageOutputs()`

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchScriptLauncher.groovy
```diff
@@ -33,4 +33,9 @@ class AzBatchScriptLauncher extends BashWrapperBuilder {
     protected boolean shouldUnstageOutputs() {
         return true
     }
+
+    @Override
+    protected boolean shouldUnstageControls() {
+        return true
```
pditommaso, line 39: why is this only azure specific ?

modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
jorgee: No, the issue was caused because it was using `shouldUnstageOutputs` in both  outputs and controls. If we default to `shouldUnstageOutputs` we will not solve the issue.

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchScriptLauncher.groovy
jorgee: The `shouldUnstageOutputs` is also azure specific. It was always unstaging outputs and controls. In a first implementation, I removed the specific Azure behaviour for controls but tests were failing, so I decided to keep it.

modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
pditommaso: Not sure Jorge, especially this 

> However, control logs (.command.err, .command.out, etc.) are always copied to the working directory.

Only the `.command.log` and `.command.begin` and `.exitcode` are created directly in the (target) work dir. The stage controls are these below and follow the usual logic as for output files 

https://github.com/nextflow-io/nextflow/blob/e89b6e0430e848a64164d72cb64cf52de4dc907c/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy#L773-L778

jorgee: I have added the integration test as commented in the last meeting. 
I have also updated the description of the PR I think it was not clear. The files` .command.log,`, `.command begin` and `.exitcode` are not managed in the `unstage_controls` part, they were not affected with the issue. The logic for controls looks similar that for outputs, but they use different functions to generate the commands. Controls use the `copyFileToWorkDir`, so they are "always" copied to working dir. For outputs, it uses `getUnstageOutputFilesScript(outputFiles,targetDir)` where `targetDir` can be `storeDir` or working dir depending on the case.

pditommaso (approved):

PR #6367, opened 2025-08-27
Support running nextflow script in cygwin

markwoon:
Support running in cygwin by translating cygwin-style paths (starting with /cygrdrive/) to windows-style paths.

### DIFF
--- a/nextflow
+++ b/nextflow
@@ -15,17 +15,16 @@
 #  limitations under the License.
 
 [[ "$NXF_DEBUG" == 'x' ]] && set -x
-
-home=$HOME
+NXF_VER=${NXF_VER:-'25.07.0-edge'}
+NXF_ORG=${NXF_ORG:-'nextflow-io'}
 if [[ "$HOME" == "/cygdrive/"* ]]; then
   drive_letter="${HOME:10:1}"
   replacement="$drive_letter:"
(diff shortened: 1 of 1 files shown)


nextflow
```diff
@@ -15,9 +15,17 @@
...
+if [[ "$HOME" == "/cygdrive/"* ]]; then
+  drive_letter="${HOME:10:1}"
+  replacement="$drive_letter:"
+  home="${HOME/\/cygdrive\/$drive_letter/$replacement}"
+fi
+
 NXF_VER=${NXF_VER:-'25.07.0-edge'}
 NXF_ORG=${NXF_ORG:-'nextflow-io'}
-NXF_HOME=${NXF_HOME:-$HOME/.nextflow}
+NXF_HOME=${NXF_HOME:-$home/.nextflow}
```
pditommaso: Why using `home` instead of `HOME`? think this will break default launcher

markwoon: If you look at line 19 of my commit, you'll see that `home` is a local variable.

The whole point is to rewrite cygwin mounts as windows mounts so that the script works.

Would you prefer if I named it something else?  Maybe `normalized_home`?

pditommaso: I see. indeed it would be preferable to use a different name. what about using `NXF_HOME` (if not defined yet)?

markwoon:
How about this?

PR #6369, opened 2025-08-28
Limit S3 client connections when using virtual threads

### LINKED ISSUE #4995: Improve stability of virtual threads at scale

jorgee:
close #4995
In SDK v1, S3 tranfers are managed by the transfer manager and a pool of threads to the transfers with a sync client. In SDK v2, the transfer manager is using an async client and the pool of threads is just used for preliminar work but not the tranfers.  However, previous steps in publish dir ( check if is a directory, etc.) are using the sync client producing the same timeout errors.

This PR make the changes to limit the resources when using virtual threads is S3 transfers. A semaphore with a permit equal to the client's max connections is set in the Netflow's S3 client. It avoids to perform too many concurrent calls in the S3 client.

Tested with a pipeline with 15 tasks that generate 1000 files of 25 MB each one. It generated timeout errors when using in both 25.04 (SDK v1 and master (SDK v2). They disappear with this PR.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/util/ThreadPoolManager.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/util/ThreadPoolManager.groovy
@@ -105,7 +105,6 @@ class ThreadPoolManager {
 
     // this class is required to avoid failures when using Java version < 21
     private static class VirtualThreadFactoryBuilder {
-
         static ThreadFactory create(String name) {
             return Thread.ofVirtual().name(name).factory()
         }


bentsherman:
I have removed the config options we discussed and updated the docs to show the desired state.

- The sync client should use the semaphore to limit virtual threads based on the maxConnections. All calls to the sync client are wrapped with `runWithPermit` to enforce this

- The async client should limit itself internally based on either the maxConnections or target throughput. Neither of these settings are specified by default in `AwsS3Config`, so the default behavior should be that the async client uses a target throughput of 10 Gbps

- The sync client uses a default maxConnections of 50 to ensure that it always has a connection limit

jorgee:
As commented last week, I got some `OutOfMemory` issues when using virtual threads with small heap sizes (<2GB). So, I have inspected a heap dump for 15K file uploads and as you can see in the following image, the main memory consumption comes from two parts:
image

1. The XML deserializer used internally by the AWS Sync client is storing the XMLInputFactory using thread locals, a instance of this is replicated in each thread. These objects are keeping the XML request' responses, so the retained size of these objects is more than 500MB in this case. For the case of the issue (~60K transfer), it could be 2GB.
  
2. The async request handling used in the trasnsfer manager client is also consuming a lot of heap space with large number of transfers. For instance, in the case of the following figure, it is doing a bucket-to-bucket copy but, as the size exceeds the multipart-threshold, so a copy is converted to several multi-part copy requests. For 15K copies, it is creating more than 50k async requests. Each request is using 12KB, so it is consuming 600MB. In the case of the issue (60K), it could consume another 2GB.
  
image

As they are internal to the AWS SDK implementation, I think we can not currently fix it. So, I think the easier option would be to limit the number of publish transfers, as it is done in the file porter. We could have the same issues for other providers. However, in case we don't want to limit it, other options that could work is moving the S3 sync to async as well as adding a limit of concurrent async operations submitted to the async client, to limit the number of AsyncHandlers. 
@bentsherman, @pditommaso, what do you think?

pditommaso:
not sure it's worth optimising for virtual threads until they are not property supported by the AWS sdk

bentsherman:
@jorgee is it possible to use a sync client for S3 transfers? then we could apply the semaphore to it and ensure that there are no more than a few hundred concurrent requests by limiting `aws.client.maxConnections`.

Right now it seems that the async client is allocating thousands of request handlers even with a smaller max concurrency, is that right?

It seems like we should move towards sync clients anyway so that we can just control it with virtual threads + semaphore.

It will be nice if the AWS SDK can provide the target throughput functionality using virtual threads, but we also need to find some kind of solution since many customers have been asking for it, and we have no control over AWS timeline.

jorgee:
The async client is mandatory for the S3 transfer manager in AWS SDKv2. We should manage multi-part stuff in Nextflow again, and I think in sdk v1 we also managed some transfers with S3 transfer manager. So we should also rewrite them to not use the transfer manager.

bentsherman:
Well, I don't know if we should go that far. I doubt we'll be able to manage the transfers as well as the transfer manager.

I have an idea that I'd like to try. I'll push it if it works

bentsherman:
I managed to wrap the S3 transfer manager in a synchronous interface. It just calls the async client and immediately waits for the result. I also wrapped these calls in the semaphore limiter so that the async client is controlled directly by `aws.client.maxConnections`.

This should ensure that the async client doesn't create too many requests when using virtual threads.

I also kept the target throughput setting so that users can control the maximum throughput, but it will still be limited by maxConnections.

If AWS ever improves the SDK to support virtual threads natively, we need only remove the wrapper class while the config options remain the same. This way we don't have to wait for AWS to give us a solution to customer issues.

bentsherman:
@jorgee can you test the PR with your benchmark and see if the OOM errors are resolved?

jorgee:
I will do. I will also try to test with this AWS SDK v2 preview version https://github.com/aws/aws-sdk-java-v2/issues/6268

jorgee:
Some updates regarding Virtual threads. 
1- The master implementation was not using Virtual Threads at least for publishDir. It was due to the CustomThreadFactory which was creating a platform thread per publishDir. I noticed it because I got this error when trying to transfer more than 30K files.
`Server VM warning: INFO: os::commit_memory(0x000076957f9ff000, 16384, 0) failed; error='Not enough space' (errno=12) [259,030s][warning][os,thread] Failed to start thread "Unknown thread" - pthread_create failed (EAGAIN) for attributes: stacksize: 1024k, guardsize: 0k, detached.`
This means that it was trying to create a pthread per publishDir thread. After the change to a Virtual thread factory, I can see 30K VirtualThread objects but not the platform threads.

2- From the heap memory usage, the proposed @bentsherman change was limiting the memory because CRT handlers but not the XMLInputFactory. I finally found a solution by changing the stax implementation to [woodstox ](https://github.com/FasterXML/woodstox)

So the current implementation has solved the memory issues for uploads and copies. For downloads, we just need to merge with PR #6402 because few big downloads can increase a lot the heap. @bentsherman, @pditommaso do you want to keep it in separate PRs or merge both in one?

pditommaso:
@claude review

pditommaso:
Smaller, well scope pr is better IMO

pditommaso:
The invasion of bot

jorgee:
> The `deleteObject` method isn't wrapped with `runWithPermit()`, unlike other S3 client operations in this class. This means delete operations bypass the semaphore-based connection limiting, which defeats the purpose of controlling concurrent S3 connections when using virtual threads.
 
good catch cursor! Fixed

docs/guides/aws-java-sdk-v2.md
```diff
@@ -16,15 +16,17 @@ The HTTP client in SDK v2 does not support overriding certain advanced HTTP opti
...
 - `aws.client.userAgent`
 
-## S3 transfer manager
+## S3 concurrency
 
-The *S3 transfer manager* is a subsystem of SDK v2 that handles S3 uploads and downloads.
+You can use the `aws.client.maxConnections` config option to control the maximum number of concurrent HTTP connections to S3.
 
-You can configure the concurrency and throughput of the S3 transfer manager manually using the `aws.client.maxConcurrency` and `aws.client.maxNativeMemory` configuration options. Alternatively, you can use the `aws.client.targetThroughputInGbps` option to set both values automatically based on a target throughput.
+You can also use the `aws.client.targetThroughputInGbps` option to control the concurrency of S3 uploads and downloads specifically, based on the available network bandwidth. This setting is `10` by default, which means that Nextflow performs S3 transfers concurrently up to 10 Gbps of network throughput, up to the maximum connection limit.
```
christopher-hakkaart:
```suggestion
Use the `aws.client.targetThroughputInGbps` option to control the concurrency of S3 uploads and downloads based on the available network bandwidth. This setting defaults to `10`, which allows Nextflow to perform concurrent S3 transfers up to 10 Gbps of network throughput, limited by the maximum connection count.
```

docs/guides/aws-java-sdk-v2.md
```diff
@@ -16,15 +16,17 @@ The HTTP client in SDK v2 does not support overriding certain advanced HTTP opti
...
+
+Use these settings with virtual threads to achieve optimal performance for your environment. Increasing these settings beyond their defaults may improve performance for large runs. You can enable virtual threads by setting the `NXF_ENABLE_VIRTUAL_THREADS` environment variable to `true`.
```
christopher-hakkaart:
```suggestion
Use these settings with virtual threads to achieve optimal performance for your environment. Increasing these settings beyond their defaults may improve performance for large runs. To enable virtual threads, set the `NXF_ENABLE_VIRTUAL_THREADS` environment variable to `true`.
```

docs/guides/aws-java-sdk-v2.md
```diff
@@ -16,15 +16,17 @@ The HTTP client in SDK v2 does not support overriding certain advanced HTTP opti
...
 
 ## Multi-part uploads
 
-Multi-part uploads are handled by the S3 transfer manager. You can use the `aws.client.minimumPartSize` and `aws.client.multipartThreshold` config options to control when and how multi-part uploads are performed.
+Nextflow uploads large files to S3 as multi-part uploads. You can use the `aws.client.minimumPartSize` and `aws.client.multipartThreshold` config options to control when and how multi-part uploads are performed.
```
christopher-hakkaart:
```suggestion
Nextflow uploads large files to S3 as multi-part uploads. Use the `aws.client.minimumPartSize` and `aws.client.multipartThreshold` configuration options to control when and how multi-part uploads are performed.
```

christopher-hakkaart (commented):
Added some suggestions. Overall the docs are great. These suggestions change the tense to make them active and present, but otherwise awesome.

docs/reference/config.md
```diff
@@ -167,34 +167,24 @@ The following settings are available:
...
-`aws.client.maxNativeMemory`
-: :::{versionadded} 25.06.0-edge
-  :::
-: The maximum native memory used by the S3 transfer manager. By default, this setting is determined by `aws.client.targetThroughputInGbps`.
-
 `aws.client.minimumPartSize`
 : :::{versionadded} 25.06.0-edge
   :::
-: The minimum part size used by the S3 transfer manager for multi-part uploads (default: `8 MB`).
+: The minimum part size used for multi-part uploads to S3 (default: `8 MB`).
```
bentsherman: @jorgee do these multi-part settings apply only to uploads or also to downloads? I keep getting confused about this, it seems like they are used for both

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
```diff
@@ -285,15 +301,14 @@ public String getObjectKmsKeyId(String bucketName, String key) {
...
+		return runWithPermit(() -> client.listObjectsV2Paginator(request));
 	}
 
 	// ===== transfer manager section =====
 
 	synchronized ExtendedS3TransferManager transferManager() {
 		if( transferManager == null ) {
-			log.debug("Creating S3 transfer manager pool - max-treads={};", transferManagerThreads);
-			transferPool = ThreadPoolManager.create("S3TransferManager", transferManagerThreads);
+			transferPool = ThreadPoolManager.create("S3TransferManager");
```
bentsherman: @jorgee I am restoring the `maxConcurrency` and `maxNativeMemory`, and removing the semaphore from the S3 transfers. Do we still need to limit the transfer manager thread pool size here? I figured the async client already limits the number of concurrent transfers

docs/reference/config.md
jorgee: They are applied to uploads and copies

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
jorgee: Fine with restoring the maxConcurrency and maxNativeMemory, and we don't need to limit the transfer manager thread pool. 
However I tried to remove the semaphore for s3 transfers when benchmarking the virtual threads. If you see the result, I got a heap space error when copying 1000 files of 1GB in the large instance. It is basically caused by the number of Native handlers as I saw in  https://github.com/nextflow-io/nextflow/pull/6369#issuecomment-3242018450 The benefit of removing the limit is very small and just for uploading small files. This is why I didn't remove the limit.
image

bentsherman: I see. But we also agreed that we should use separate limits for the S3 client vs the transfer manager right? So maybe I should just add a semaphore to the ExtendedS3TransferManager

docs/reference/config.md
bentsherman: But I see that this setting is being used for downloads as well, such as `AwsS3Config::checkDownloadBufferParams()` and `ExtendedS3TransferManager::setDownloadBufferProperties()` ?

jorgee: Ah its true. It is also used in downloads to setup the initial buffer size, that by default is 10 times the part size. So it is used in all transfers

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
bentsherman: Also, I was concerned that putting a hard limit on the S3 transfer manager might bottleneck the case of performing many s3-to-s3 copies in parallel

So when you did the benchmark that showed 50% reduction in publishing time, that was with a limit of 50 concurrent transfers?

docs/reference/config.md
bentsherman: Also, I thought the multipart threshold was 5 GB but here it is 8 MB. I must be missing something

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
jorgee: Yes, it was with the default 50 concurrent transfers

docs/reference/config.md
bentsherman: Because if the threshold is only 8 MB then that would mean that our benchmarks with 1 GB files were using multipart

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
bentsherman: Okay, maybe we can bump the default to 100 :eyes:

bentsherman: You already have a semaphore for downloads that is based on the max download buffer. But my understanding is that uploads and copies don't have these buffers, they just have these large native handlers. So I guess we should just have one semaphore for the downloads and another for the copies and uploads

docs/reference/config.md
jorgee: Yes, in v2 the 1GB was using multipart. There are two options one for the threshold and anothe for the minimum part. By thedault both are the same 8MB. What I am not sure is in v1 version. I saw at someplace a check for 5GB when uploading but I can't find it.

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
jorgee: Yes, they don't have these buffers, in copies and uploads we could have the handlers problem. So, another semaphore

bentsherman:
@jorgee I have updated everything based on what we discussed and all tests are passing. Let me know if everything looks good

Here is what we have now:

- Limit S3 API calls with semaphore when using virtual threads
  - allows the FilePorter and PublishDir to use virtual threads without exhausting the aws maxConnections

- Replace `transferManagerThreads` config option with semaphores
  - transfer manager uses semaphore to limit copies and uploads to prevent out-of-memory error
  - transfer manager already uses semaphore to limit downloads based on download buffer size

- Remove `maxConcurrency` and `maxNativeMemory` from docs
  - they will remain as hidden options for now

- Make docs more user friendly
  - use aws targetThroughputInGbps to limit S3 transfers
  - use aws maxConnections to limit all other S3 API calls

bentsherman:
I'm thinking to split this PR into two:
1. Limit S3 sync client with semaphore when using virtual threads
2. Clean up S3 transfer manager settings

I think everything is correct as is, but this separation might be better for the commit history

jorgee:
Changes look good to me. I will repeat some beanchmark on Monday to ensure nothing is broken.

bentsherman:
Thanks. Even if the virtual threads don't improve the performance much, it makes the user experience much simpler (no need to configure FilePorter / PublishDir thread pools), so it will be worthwhile

modules/nextflow/build.gradle
```diff
@@ -53,6 +53,7 @@ dependencies {
     api 'org.pf4j:pf4j:3.12.0'
     api 'dev.failsafe:failsafe:3.1.0'
     api 'io.seqera:lib-trace:0.1.0'
+    api 'com.fasterxml.woodstox:woodstox-core:7.1.1'
```
pditommaso, line 56: More details about this? if it's used by amazon plugin could be kept there?

bentsherman: @jorgee can we move the `WstxInputFactory` configuration to nf-amazon?

bentsherman: also seems strange that the `resources` directory is under `src/main` in the nextflow module but under `src` in nf-amazon

jorgee: I added the WstxInputFactory in nextflow because I got errors when added in nf-amazon, but I didn't note the location of the resources folder. Maybe it was the cause. I will try again.

jorgee: I have reviewed it. The resources is under `src/main`, but the `build.gradle` is pointing to the correct path for resources. In the build folder, everything seems correct, and the `META-INF/services` is in the classpath of the plugin. However, it is not loading the WstxInputFactory if I move the dependency and resource file to the plugin. It loads the xerces one (default), that keeps the last message in the ThreadLocal. I think it is due to the pf4j class loaders which is very difficult to debug.

pditommaso: ok, we can live with that.

modules/nextflow/src/main/groovy/nextflow/util/ThreadPoolManager.groovy
```diff
@@ -89,9 +90,25 @@ class ThreadPoolManager {
...
+
+    protected ExecutorService virtualThreadService() {
+        final poolName = name ?: "nf-thread-pool-${poolCount.getAndIncrement()}".toString()
+        final pool = VirtualThreadFactoryBuilder.create(poolName)
+        return Executors.newThreadPerTaskExecutor(pool)
+    }
+
+    // this class is required to avoid failures when using Java version < 21
+    private static class VirtualThreadFactoryBuilder {
+
```
pditommaso:
```suggestion
```

pditommaso (approved):

PR #6372, opened 2025-08-29
Add Value type for dataflow values (value channels)

bentsherman:
This PR adds the `Value` type to the Nextflow type system to distinguish value channels from queue channels.

It renames "queue channels" and value channels as "dataflow channels" and "dataflow values", or "channels and values" for short. This language is more clear and concise, since "queue channel" is somewhat redundant and "value channel" is somewhat contradictory.

There are no changes to runtime behavior aside from allowing the use of `Value` in type annotations. It is primarily a change to docs and terminology.

This change is still under discussion. We already have the `Channel` type, but we need an additional type to distinguish value channels from queue channels. `Value` was the resounding favorite in an internal poll (other options were `Async`, `Future`, `ValueChannel`).

So if we go with `Channel` and `Value`, it makes sense to update the vocabulary in the docs to match, hence the proposal of "dataflow channel" and "dataflow value".

### DIFF
--- a/docs/channel.md
+++ b/docs/channel.md
@@ -1,54 +1,30 @@
-(channel-page)=
+(dataflow-page)=
 
-# Channels
+# Dataflow
 
-In Nextflow, **channels** are the key data structures that facilitate the dataflow dependencies between each step (i.e. {ref}`process <process-page>`) in a pipeline.
+Nextflow uses a **dataflow** programming model to define workflows in a declarative manner. {ref}`Processes <process-page>` in a pipeline are connected to each other by *dataflow channels* and *dataflow values*.
 
-There are two kinds of channels, *queue channels* and *value channels*. Channels are created using *channel factories* and transformed using *channel operators*.
+(dataflow-type-channel)=
 
-(channel-type-queue)=
+## Channels
 
-## Queue channels
+A *dataflow channel*, or *channel* for short, is an asynchronous sequence of values.
 
-A *queue channel* is a channel that *emits* an asynchronous sequence of values.
-
-A queue channel can be created by channel factories (e.g., {ref}`channel.of <channel-of>` and {ref}`channel.fromPath <channel-path>`), operators (e.g., {ref}`operator-map` and  {ref}`operator-filter`), and processes (see {ref}`Process outputs <process-output>`).
-
-The values in a queue channel cannot be accessed directly -- they can only be accessed by passing the channel as input to an operator or process. For example:
+The values in a channel cannot be accessed directly -- they can only be accessed by passing the channel as input to an operator or process. For example:
 
 ```nextflow
-channel.of(1, 2, 3).view { v -> "queue channel emits ${v}" }
+channel.of(1, 2, 3).view { v -> "channel emits ${v}" }
 ```
 
 ```console
-queue channel emits 1
-queue channel emits 2
-queue channel emits 3
+channel emits 1
+channel emits 2
+channel emits 3
 ```
 
-(channel-type-value)=
+### Factories
 
-## Value channels
-
-A *value channel* is a channel that is *bound* to an asynchronous value.
-
-A value channel can be created with the {ref}`channel.value <channel-value>` factory, certain operators (e.g., {ref}`operator-collect` and {ref}`operator-reduce`), and processes (under {ref}`certain conditions <process-out-singleton>`).
-
-The value in a value channel cannot be accessed directly -- it can only be accessed by passing the channel as input to an operator or process. For example:
-
-```nextflow
-channel.value(1).view { v -> "value channel is ${v}" }
-```
-
-```console
-value channel is 1
-```
-
-## Channel factories
-
-Channel factories are functions that create channels from regular values.
-
-The `channel.fromPath()` factory creates a channel from a file name or glob pattern, similar to the `files()` function:
+A channel can be created by factories in the `channel` namespace. For example, the `channel.fromPath()` factory creates a channel from a file name or glob pattern, similar to the `files()` function:
 
 ```nextflow
 channel.fromPath('input/*.txt').view()
@@ -56,7 +32,7 @@ channel.fromPath('input/*.txt').view()
 
 See {ref}`channel-factory` for the full list of channel factories.
 
-## Operators
+### Operators
 
 Channel operators, or *operators* for short, are functions that consume and produce channels. Because channels are asynchronous, operators are necessary to manipulate the values in a channel. Operators are particularly useful for implementing glue logic between processes.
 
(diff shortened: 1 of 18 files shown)


docs/channel.md
```diff
@@ -1,62 +1,38 @@
-(channel-page)=
+(dataflow-page)=
 
-# Channels
+# Dataflow
 
-In Nextflow, **channels** are the key data structures that facilitate the dataflow dependencies between each step (i.e. {ref}`process <process-page>`) in a pipeline.
+Nextflow uses a **dataflow** programming model to define workflows in a declarative manner. {ref}`Processes <process-page>` in a pipeline are connected to each other by *dataflow channels* and *dataflow values*.
```
christopher-hakkaart:
```suggestion
Nextflow uses a **dataflow** programming model to define workflows declaratively. In this model, {ref}`processes <process-page>` within a pipeline connect to each other through *dataflow channels* and *dataflow values*.
```
This feels more natural to me, but might change the intended meaning.

docs/channel.md
```diff
@@ -1,62 +1,38 @@
...
 
-There are two kinds of channels, *queue channels* and *value channels*. Channels are created using *channel factories* and transformed using *channel operators*.
+(dataflow-type-channel)=
 
-(channel-type-queue)=
+## Channels
 
-## Queue channels
+A *dataflow channel*, or *channel* for short, is an asynchronous sequence of values.
```
christopher-hakkaart:
```suggestion
A *dataflow channel* (or simply *channel*) is an asynchronous sequence of values.
```

docs/channel.md
```diff
@@ -1,62 +1,38 @@
...
-(channel-type-value)=
-
-## Value channels
-
-A *value channel* is a channel that is *bound* to an asynchronous value.
-
-A value channel can be created with the {ref}`channel.value <channel-value>` factory, certain operators (e.g., {ref}`operator-collect` and {ref}`operator-reduce`), and processes (under {ref}`certain conditions <process-out-singleton>`).
-
-The value in a value channel cannot be accessed directly -- it can only be accessed by passing the channel as input to an operator or process. For example:
+The values in a channel cannot be accessed directly -- they can only be accessed by passing the channel as input to an operator or process. For example:
```
christopher-hakkaart:
```suggestion
The values in a channel cannot be accessed directly and must be passed as an input to an operator or process. For example:
```

docs/channel.md
```diff
@@ -81,3 +57,23 @@ Commonly used operators include:
...
 
 See {ref}`operator-page` for the full list of operators.
+
+(dataflow-type-value)=
+
+## Values
+
+A *dataflow value* is an asynchronous value.
+
+A dataflow value can be created with the {ref}`channel.value <channel-value>` factory, or by a process (under {ref}`certain conditions <process-out-singleton>`).
```
christopher-hakkaart:
```suggestion
A dataflow value can be created using the {ref}`channel.value <channel-value>` factory, or by a process (under {ref}`certain conditions <process-out-singleton>`).
```

docs/channel.md
```diff
@@ -81,3 +57,23 @@ Commonly used operators include:
...
+
+The value in a dataflow value cannot be accessed directly -- it can only be accessed by passing the channel as input to an operator or process. For example:
```
christopher-hakkaart:
```suggestion
A dataflow value cannot be accessed directly and must be passed as input to an operator or process. For example:
```

docs/migrations/25-10.md
```diff
@@ -72,6 +72,8 @@ Type annotations can be appended with `?` to denote that the value can be `null`
 def x_opt: String? = null
 ```
 
+Queue channels are represented in the type system as `Channel`, while value channels are represented as `Value`. As a result, the terminology for channels has been updated to be more clear and concise. Queue channels are now called *dataflow channels* (or *channels* for short) and value channels are now called *dataflow values*. See {ref}`dataflow-page` for more information.
```
christopher-hakkaart:
```suggestion
In the type system, queue channels are represented as `Channel`, while value channels are represented as `Value`. To make the terminology clearer and more concise, queue channels are now called *dataflow channels* (or simply *channels*), and value channels are now called *dataflow values*. See {ref}`dataflow-page` for more information.
```

docs/overview.md
```diff
@@ -63,9 +61,9 @@ workflow {
...
 
-When the workflow is started, it will create two processes and one channel (`query_ch`) and it will link all of them. Both processes will be started at the same time and they will listen to their respective input channels. Whenever `blast_search` emits a value, `extract_top_hits` will receive it (i.e. `extract_top_hits` consumes the channel in a *reactive* way).
+When the workflow is executed, it creates two processes (`blast_search` and `extract_top_hits`) connected by the channel `query_ch`. Each process executes a task and emits a value for each input that it receives. Whenever `blast_search` emits a value, `extract_top_hits` receives it through the `query_ch` channel.
 
-Read the {ref}`Channel <channel-page>` and {ref}`Process <process-page>` sections to learn more about these features.
+Read {ref}`process-page`, {ref}`dataflow-page`, and {ref}`workflow-page` to learn more about these features.
```
christopher-hakkaart:
```suggestion
See {ref}`process-page`, {ref}`dataflow-page`, and {ref}`workflow-page` to learn more about these features.
```

docs/process.md
```diff
@@ -735,18 +735,20 @@ When multiple repeaters are defined, the process is executed for each *combinati
...
 
-### Multiple input channels
+### Multiple inputs
 
-A key feature of processes is the ability to handle inputs from multiple channels.
+A process can declare multiple inputs, which allows it to accept inputs from multiple dataflow sources.
 
-When two or more channels are declared as process inputs, the process waits until there is a complete input configuration, i.e. until it receives a value from each input channel. When this condition is satisfied, the process consumes a value from each channel and launches a new task, repeating this logic until one or more channels are empty.
+:::{warning}
+When calling a process with multiple inputs, make sure to not supply more than one channel -- all other inputs should be dataflow values. Invoking a process with multiple channels can lead to {ref}`non-deterministic behavior <cache-nondeterministic-inputs>`.
```
```suggestion
Do not supply more than one channel when calling a process with multiple inputs. Invoking a process with multiple channels can lead to {ref}`non-deterministic behavior <cache-nondeterministic-inputs>`. All additional inputs should be dataflow values. 
```

docs/process.md
```diff
@@ -762,62 +764,28 @@ process echo {
...
-
 workflow {
   x = channel.value(1)
   y = channel.of('a', 'b', 'c')
   echo(x, y)
 }
 ```
 
-The above example executes the `echo` process three times because `x` is a value channel, therefore its value can be read as many times as needed. The process termination is determined by the contents of `y`. It outputs:
+The above example executes the `echo` process three times. Because `x` is a dataflow value, it is used once for each value in `y`. It outputs:
```
```suggestion
The above example executes the `echo` process three times. The dataflow value `x` is used once for each value in `y`. It outputs:
```

christopher-hakkaart (approved):
Docs look good. I've added a few suggestions that I think help with flow. Take or leave what you like.

I would like to change some of the voice and tense on some of these pages. But I'll leave that to another PR.

bentsherman:
Thanks Chris. What do you think about the terminology shift overall? Referring to value channels as "dataflow values" instead of channels? Do you think it makes things more or less clear?

One implication is that describing workflows as "processes connected by channels" isn't strictly true, since the connections could be through dataflow values, which in this PR we're saying are not channels. But "processes connected by channels" is still mostly true and probably how most people will still talk about it

christopher-hakkaart:
My 0.02 cents worth here is that I like being able to say channels and values, and I think this would make sense moving forward (I think most will instinctively shorten to "value" from "dataflow values").

It is a fundamental shift in the way we talk/describe Nextflow. It might create some friction initially but I think it would be a net gain over time.

The  "processes connected by channels" not being 100% true concerns me a little bit. It would be ideal to tidy this somehow, but I'm not sure if there is a clean solution. It would likely require more renaming, which is just going to move the problem around.

PR #6375, opened 2025-09-02
Fix false error for executor.jobName config option

bentsherman:
This PR fixes a false error for the `executor.jobName` config option.

I originally added a check for invalid dynamic config options, since only process config options can be dynamic. However, there are some config options that explicitly expect a closure, like `executor.jobName`.

Instead of adding another special case for this, I will move this check into the language server as part of type checking. So this PR just removes the check in nf-lang to address the false error.

### DIFF
--- a/modules/nf-lang/src/main/java/nextflow/config/control/VariableScopeVisitor.java
+++ b/modules/nf-lang/src/main/java/nextflow/config/control/VariableScopeVisitor.java
@@ -119,12 +119,8 @@ class VariableScopeVisitor extends ConfigVisitorSupport {
         var scopes = currentConfigScopes();
         inProcess = !scopes.isEmpty() && "process".equals(scopes.get(0));
         inClosure = node.value instanceof ClosureExpression;
-        if( inClosure ) {
-            if( isWorkflowHandler(scopes, node) )
-                vsc.addWarning("The use of workflow handlers in the config is deprecated -- use the entry workflow or a plugin instead", String.join(".", node.names), node);
-            else if( !inProcess )
-                vsc.addError("Dynamic config options are only allowed in the `process` scope", node);
-        }
+        if( isWorkflowHandler(scopes, node) )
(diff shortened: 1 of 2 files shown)


bentsherman:
TODO: `executor.jobName` also needs to support the process DSL like `task`

bentsherman:
@jorgee this is ready for review

modules/nf-lang/src/main/java/nextflow/config/control/VariableScopeVisitor.java
```diff
@@ -117,34 +117,41 @@ public void visitConfigAssign(ConfigAssignNode node) {
...
-                vsc.addWarning("The use of workflow handlers in the config is deprecated -- use the entry workflow or a plugin instead", String.join(".", node.names), node);
-            else if( !inProcess )
-                vsc.addError("Dynamic config options are only allowed in the `process` scope", node);
-        }
+        if( isWorkflowHandler(scopes, node) )
+            vsc.addWarning("The use of workflow handlers in the config is deprecated -- use the entry workflow or a plugin instead", String.join(".", node.names), node);
         if( inClosure ) {
             vsc.pushScope(ScriptDsl.class);
-            if( inProcess )
+            if( inProcessScope )
```
jorgee, line 126: `InProcessScope` can be true for a `process` scope or for `executor.jobName`. In the other cases, it was used with a closure check, so I do not see a problem. In this case, you are pushing a scope of the `ProcessDsl`, and there could be a case where the scope is not a process. Is there an error here or is it fine?  I am not familiar with this part of the code to assess if it was correct. If it is correct, it would be good to add a comment to clarify it.

bentsherman: `executor.jobName` is a special case because it can use the ProcessDsl in the closure. I will add some comments to clarify this

PR #6376, opened 2025-09-02
Bump jgit:7.1.1.202505221757-r

bentsherman:


bentsherman:
The test-e2e action is passing but it only took 3 minutes to execute. Does it not wait for the pipelines to finish in platform?

pditommaso:
> Does it not wait for the pipelines to finish in platform?

nope, it only launches the executions

PR #6378, opened 2025-09-02
feat: remove footgun from TaskPath

nh13:


pditommaso:
Hi! More context would be appreciated

PR #6379, opened 2025-09-03
Update core plugins to use Nextflow Gradle plugin

bentsherman:
This PR updates the core plugins to use the [Nextflow Gradle plugin](https://github.com/nextflow-io/nextflow-plugin-gradle) and removes the old Gradle code for publishing plugins.

It is currently building successfully, but many tests are failing with plugin extensions. I think there is still something missing to configure the local build to provide plugins during testing. Maybe something I deleted in `plugins/build.gradle`.


pditommaso:
All broken :/

bentsherman:
The rest is up to you :smile:

PR #6380, opened 2025-09-04
WIP: First attempt at auth0 flow for `nextflow auth login`

ewels:
This PR introduces a new `nextflow auth login` command that enables OAuth2 authentication with Seqera Platform using Auth0.

## Features

- **OAuth2/PKCE Flow**: Implements secure PKCE (Proof Key for Code Exchange) authentication flow
- **Browser-based Login**: Opens the user's default browser for authentication
- **Local Callback Server**: Runs a temporary local server on port 8085 to handle the OAuth callback
- **Flexible API Endpoints**: Prompts user for API URL (default `https://api.cloud.seqera.io`) - if not recognised, skips auth0 flow and asks user to manually enter a PAT.

## Usage

```bash
# Authenticate with default Seqera Cloud
nextflow auth login

# Will prompt for API URL (defaults to https://api.cloud.seqera.io)
# Opens browser for authentication
# Saves credentials to ~/.config/seqera/credentials (or $XDG_CONFIG_HOME/seqera/credentials)
```

## Implementation Details

- Uses Auth0 domain: `seqera-development.eu.auth0.com`
- Implements PKCE flow for enhanced security (no client secrets required)
- Cross-platform browser opening with fallbacks

## Configuration

The Auth0 client ID is currently set to a placeholder and will need to be updated with the actual production value before merge.

## Related

This implementation aligns with the authentication flow used in the [VSCode Nextflow extension](https://github.com/nextflow-io/vscode-language-nextflow/pull/132).

🤖 Generated with help from [Claude Code](https://claude.ai/code)

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdAuth.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdAuth.groovy
@@ -1,0 +1,409 @@
+/*
+ * Copyright 2013-2025, Seqera Labs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package nextflow.cli
+
+import com.beust.jcommander.Parameter
+import com.beust.jcommander.Parameters
+import groovy.transform.CompileStatic
+import groovy.util.logging.Slf4j
+import groovy.yaml.YamlBuilder
+import groovy.yaml.YamlSlurper
+import nextflow.exception.AbortOperationException
+
+import java.awt.Desktop
+import java.net.ServerSocket
+import java.net.URI
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
+import java.security.MessageDigest
+import java.security.SecureRandom
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.TimeUnit
+
+/**
+ * Implements the {@code auth} command
+ *
+ * @author Phil Ewels <phil.ewels@seqera.io>
+ */
+@Slf4j
+@CompileStatic
+@Parameters(commandDescription = "Manage Seqera Platform authentication")
+class CmdAuth extends CmdBase implements UsageAware {
+
+    interface SubCmd {
+        String getName()
+        void apply(List<String> result)
+        void usage(List<String> result)
+    }
+
+    static public final String NAME = 'auth'
+
+    private List<SubCmd> commands = []
+
+    String getName() {
+        return NAME
+    }
+
+    @Parameter(hidden = true)
+    List<String> args
+
+    CmdAuth() {
+        commands.add(new LoginCmd())
+    }
+
+    void usage() {
+        usage(args)
+    }
+
+    void usage(List<String> args) {
+        List<String> result = []
+        if (!args) {
+            result << this.getClass().getAnnotation(Parameters).commandDescription()
+            result << 'Usage: nextflow auth <sub-command> [options]'
+            result << ''
+            result << 'Commands:'
+            commands.collect { it.name }.sort().each { result << "  $it".toString() }
+            result << ''
+        } else {
+            def sub = commands.find { it.name == args[0] }
+            if (sub)
+                sub.usage(result)
+            else {
+                throw new AbortOperationException("Unknown auth sub-command: ${args[0]}")
+            }
+        }
+        println result.join('\n').toString()
+    }
+
+    @Override
+    void run() {
+        if (!args) {
+            usage()
+            return
+        }
+
+        try {
+            getCmd(args).apply(args.drop(1))
+        } catch (Exception e) {
+            throw new AbortOperationException(e.message)
+        }
+    }
+
+    protected SubCmd getCmd(List<String> args) {
+        def cmd = commands.find { it.name == args[0] }
+        if (cmd) {
+            return cmd
+        }
+
+        def matches = commands.collect { it.name }.closest(args[0])
+        def msg = "Unknown auth sub-command: ${args[0]}"
+        if (matches)
+            msg += " -- Did you mean one of these?\n" + matches.collect { "  $it" }.join('\n')
+        throw new AbortOperationException(msg)
+    }
+
+    class LoginCmd implements SubCmd {
+
+        private static final String AUTH0_DOMAIN = "seqera-development.eu.auth0.com"
+        private static final String AUTH0_CLIENT_ID = "ENTER_CLIENT_ID_HERE_WHEN_WE_HAVE_IT"
+        private static final String AUTH0_AUDIENCE = "platform"
+        private static final int CALLBACK_PORT = 8085
+
+        @Override
+        String getName() { 'login' }
+
+        @Override
+        void apply(List<String> args) {
+            if (args.size() > 0) {
+                throw new AbortOperationException("Too many arguments for login command")
+            }
+
+            println "Nextflow authentication with Seqera Platform"
+            println ""
+
+            // Prompt user for API URL
+            def apiUrl = promptForApiUrl()
+
+            try {
+                performAuth0Login(apiUrl)
+            } catch (Exception e) {
+                log.debug("Authentication failed", e)
+                throw new AbortOperationException("Authentication failed: ${e.message}")
+            }
+        }
+
+        private String promptForApiUrl() {
+            System.out.print("Seqera Platform API URL [Default Seqera Cloud, https://api.cloud.seqera.io]: ")
+            System.out.flush()
+
+            def reader = new BufferedReader(new InputStreamReader(System.in))
+            def input = reader.readLine()?.trim()
+
+            return input?.isEmpty() || input == null ? 'https://api.cloud.seqera.io' : input
+        }
+
+        private void performAuth0Login(String apiUrl) {
+            println "Opening browser for authentication..."
+
+            // Generate PKCE parameters
+            def codeVerifier = generateCodeVerifier()
+            def codeChallenge = generateCodeChallenge(codeVerifier)
+            def state = generateState()
+
+            // Start local server for callback
+            def callbackFuture = startCallbackServer(state)
+
+            // Build authorization URL
+            def authUrl = buildAuthUrl(codeChallenge, state)
+
+            // Open browser
+            openBrowser(authUrl)
+
+            println "Waiting for authentication to complete..."
+
+            try {
+                // Wait for callback with timeout
+                def authCode = callbackFuture.get(5, TimeUnit.MINUTES)
+
+                // Exchange code for token
+                def tokenData = exchangeCodeForToken(authCode, codeVerifier)
+
+                // Save credentials
+                saveCredentials(tokenData, apiUrl)
+
+                println "Authentication successful! Credentials saved to ${getCredentialsPath()}"
+
+            } catch (Exception e) {
+                throw new RuntimeException("Authentication timeout or failed: ${e.message}", e)
+            }
+        }
+
+        private String generateCodeVerifier() {
+            def random = new SecureRandom()
+            def bytes = new byte[32]
+            random.nextBytes(bytes)
+            return Base64.urlEncoder.withoutPadding().encodeToString(bytes)
+        }
+
+        private String generateCodeChallenge(String codeVerifier) {
+            def digest = MessageDigest.getInstance("SHA-256")
+            def hash = digest.digest(codeVerifier.getBytes("UTF-8"))
+            return Base64.urlEncoder.withoutPadding().encodeToString(hash)
+        }
+
+        private String generateState() {
+            def random = new SecureRandom()
+            def bytes = new byte[16]
+            random.nextBytes(bytes)
+            return Base64.urlEncoder.withoutPadding().encodeToString(bytes)
+        }
+
+        private CompletableFuture<String> startCallbackServer(String expectedState) {
+            def future = new CompletableFuture<String>()
+
+            Thread.start {
+                try {
+                    def server = new ServerSocket(CALLBACK_PORT)
+                    server.soTimeout = 300000 // 5 minutes
+
+                    def socket = server.accept()
+                    def input = new BufferedReader(new InputStreamReader(socket.inputStream))
+                    def output = new PrintWriter(socket.outputStream)
+
+                    def requestLine = input.readLine()
+                    if (requestLine?.startsWith("GET /callback")) {
+                        def query = requestLine.split("\\?")[1]?.split(" ")[0]
+                        def params = parseQueryParams(query)
+
+                        if (params['state'] != expectedState) {
+                            throw new RuntimeException("Invalid state parameter")
+                        }
+
+                        if (params['error']) {
+                            throw new RuntimeException("Auth error: ${params['error']} - ${params['error_description']}")
+                        }
+
+                        def code = params['code']
+                        if (!code) {
+                            throw new RuntimeException("No authorization code received")
+                        }
+
+                        // Send success response
+                        output.println("HTTP/1.1 200 OK")
+                        output.println("Content-Type: text/html")
+                        output.println()
+                        output.println("<html><body><h1>Authentication successful!</h1><p>You can close this window.</p></body></html>")
+                        output.flush()
+
+                        future.complete(code)
+                    } else {
+                        future.completeExceptionally(new RuntimeException("Invalid callback request"))
+                    }
+
+                    socket.close()
+                    server.close()
+
+                } catch (Exception e) {
+                    future.completeExceptionally(e)
+                }
+            }
+
+            return future
+        }
+
+        private Map<String, String> parseQueryParams(String query) {
+            Map<String, String> params = [:]
+            if (query) {
+                query.split('&').each { param ->
+                    def parts = param.split('=', 2)
+                    if (parts.length == 2) {
+                        params[URLDecoder.decode(parts[0], "UTF-8")] = URLDecoder.decode(parts[1], "UTF-8")
+                    }
+                }
+            }
+            return params
+        }
+
+        private String buildAuthUrl(String codeChallenge, String state) {
+            def params = [
+                'response_type': 'code',
+                'client_id': AUTH0_CLIENT_ID,
+                'redirect_uri': "http://localhost:${CALLBACK_PORT}/callback",
+                'scope': 'openid profile email offline_access',
+                'audience': AUTH0_AUDIENCE,
+                'code_challenge': codeChallenge,
+                'code_challenge_method': 'S256',
+                'state': state,
+                'prompt': 'login'
+            ]
+
+            def query = params.collect { k, v ->
+                "${URLEncoder.encode(k.toString(), 'UTF-8')}=${URLEncoder.encode(v.toString(), 'UTF-8')}"
+            }.join('&')
+
+            return "https://${AUTH0_DOMAIN}/authorize?${query}"
+        }
+
+        private void openBrowser(String url) {
+            try {
+                if (Desktop.isDesktopSupported()) {
+                    Desktop.desktop.browse(new URI(url))
+                } else {
+                    // Fallback for systems without Desktop support
+                    def os = System.getProperty("os.name").toLowerCase()
+                    if (os.contains("win")) {
+                        Runtime.runtime.exec("rundll32 url.dll,FileProtocolHandler ${url}")
+                    } else if (os.contains("mac")) {
+                        Runtime.runtime.exec("open ${url}")
+                    } else {
+                        Runtime.runtime.exec("xdg-open ${url}")
+                    }
+                }
+            } catch (Exception e) {
+                println "Could not open browser automatically. Please visit: ${url}"
+            }
+        }
+
+        private Map exchangeCodeForToken(String authCode, String codeVerifier) {
+            def tokenUrl = "https://${AUTH0_DOMAIN}/oauth/token"
+
+            def params = [
+                'grant_type': 'authorization_code',
+                'client_id': AUTH0_CLIENT_ID,
+                'code': authCode,
+                'redirect_uri': "http://localhost:${CALLBACK_PORT}/callback",
+                'code_verifier': codeVerifier
+            ]
+
+            def postData = params.collect { k, v ->
+                "${URLEncoder.encode(k.toString(), 'UTF-8')}=${URLEncoder.encode(v.toString(), 'UTF-8')}"
+            }.join('&')
+
+            def connection = new URL(tokenUrl).openConnection() as HttpURLConnection
+            connection.requestMethod = 'POST'
+            connection.setRequestProperty('Content-Type', 'application/x-www-form-urlencoded')
+            connection.doOutput = true
+
+            connection.outputStream.withWriter { writer ->
+                writer.write(postData)
+            }
+
+            if (connection.responseCode != 200) {
+                def error = connection.errorStream?.text ?: "HTTP ${connection.responseCode}"
+                throw new RuntimeException("Token exchange failed: ${error}")
+            }
+
+            def response = connection.inputStream.text
+            def json = new groovy.json.JsonSlurper().parseText(response)
+            return json as Map
+        }
+
+        private void saveCredentials(Map tokenData, String apiUrl) {
+            def xdgConfigHome = System.getenv("XDG_CONFIG_HOME")
+            def credentialsDir = xdgConfigHome ?
+                Paths.get(xdgConfigHome, "seqera") :
+                Paths.get(System.getProperty("user.home"), ".config", "seqera")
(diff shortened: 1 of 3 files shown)


ewels:
Struggling to get the tests to pass. Even `master` is failing for me locally, whilst passing on GitHub. Not sure what's wrong.

ewels:
@claude Please review this pull request and provide feedback on:

- Code quality and best practices
- Potential bugs or issues
- Performance considerations
- Security concerns
- Test coverage

Use the repository's CLAUDE.md for guidance on style and conventions. Be constructive and helpful in your feedback.

Use `gh pr comment` with your Bash tool to leave your review as a comment on the PR.

pditommaso:
Let's start refactoring this as `nf-seqera` plugin (in the `plugins` directory, same as `nf-tower`)

jorgee:
I get confused a bit on how -url is managed. 
In usage examples, it seems the `-url` is used to support login for enterprise, but inside the code it is also used to decide to use cloud dev, stage or prod. 
To manage it, the code has a couple of maps with the api and Auth0 URLs, and Auth0 client Id for all environments hardcoded in the code. There is also a complex sequence to check if the URL is from platform (cloud) or enterprise and if it is enterprise then uses PAT instead of Auth0.
 
I think it is better to simplify it by -url is for enterprise with PAT and no url is for cloud. Moreover, instead of hardcoding all environments, keep just the production values as default and allow to test in other environments using env variables. In fact, we already have the TOWER_API_ENDPOINT, we just would need to add the ones for auth endpoint and client id. 

What do you think about it?

modules/nextflow/src/main/groovy/nextflow/cli/CmdAuth.groovy
```diff
@@ -0,0 +1,1324 @@
...
+                // Only delete PAT from platform if this is a cloud endpoint
+                if (isCloudEndpoint(apiUrl)) {
+                    def tokenId = decodeTokenId(existingToken as String)
+                    deleteTokenViaApi(existingToken as String, apiUrl, tokenId)
+                } else {
+                    println " - Enterprise installation detected - PAT will not be deleted from platform."
+                }
+
+                removeAuthFromConfig()
+
```
jorgee: What happens if there was a previous token already defined in the config. Could it be deleted by mistake?

ewels: Yes it will be deleted, by design. Maybe we could add a confirmation prompt?

ewels:
@jorgee I think that we still need the logic to detect the 3 prod / stage / dev URLs for cloud - it's needed to trigger the logic to _try_ the Auth0 flow rather than just booting people to use a PAT.

If we have that logic in the code, then I don't really see a reason to not also include the key and auth0 URL personally, it's only a handful of extra lines of code. And it'd be a swap for env vars. Given that I don't envision ever needing to test with any other auth0 URLs I'm not really sure it'd be any simpler..?

jorgee:
It seems login is not working well when there is a profile with tower properties. I had already defined a profile and it was removed from the profile and wrote a new one. I will debug it, but I was wondering if it will be easier to write everything in a separate file and just add the include it in the .nextflow/config. We will just need to add or remove the line instead of looking for the tower config.

jorgee:
Pushed changes to move Auth functionality to nf-tower

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
```diff
@@ -0,0 +1,1108 @@
...
+                performAuth0Login(apiUrl, endpointInfo.environment as String)
+            } catch( Exception e ) {
+                log.debug("Authentication failed", e)
+                println ""
+                throw new AbortOperationException("${e.message}")
+            }
+        } else {
+            // Enterprise endpoint - use PAT authentication
+            handleEnterpriseAuth(apiUrl)
+        }
```
jorgee, line 89: This is what I referred to in my previous comment https://github.com/nextflow-io/nextflow/pull/6380#issuecomment-3337805214.
In `normalizeApi`, it is converting http to https and, in `getCloudEndpoint`, it is checking if the provided apiUrl is in a legacy format '/api' to compare it with the dev, stage and prod endpoints
Assuming that `url ` flag is just for login in enterprise, it could be simplify by
```
if (apiUrl)
 handleEnterpriseAuth(apiUrl)
else
  performAuth0Login()
```
No need to `normalizeApi` function nor `getCloudEndpointInfo`. If the problem is providing the Auth0 url and clientId when testing we could be kept in a map api_endpoint -> auth0_Info and just use the TOWER_ENDPOINT_API to retrive them as it is used in the rest of the nf-tower calls.

jorgee: Forget this comment it is also used for logout. It will not simplify anything

jorgee:
Main changes from @ewels version:

- CmdAuth implementation moved to nf-tower
- Instead of writing directly to .nextflow/config writes in .login and includes it in Config. Avoid issues removing existing configuration.
- Read config using the ConfigBuilder instead of the configSlurper
- HttpURLConnection changed to HxClient as the TowerClient
- Other minor refactors for duplicated code.

ewels:
@jorgee I did some testing and added some small tweaks:

* Catch user cancellation of auth flow cleanly
     * I did cmd+c when testing and had some weird hanging behaviour, not sure if this is overly paranoid but it seemed to fix it for me
* Add confirmation when running `auth logout`
    * Deleting the PAT remotely is destructive, so I think it's good to prompt for confirmation. I've been meaning to add this for a while
* Rename `.login` to `seqera_auth.config`
    * I'd prefer to be explicit here, we're only adding auth for Seqera Platform, so let's call it that

Hope this is ok! All other testing was fine, so I'm good with your changes. If you're happy with mine, then I think we should be good to merge.

Phil

pditommaso:
@claude review

pditommaso:
> Failed with exit code 128

😆

pditommaso:
Color are nice, but there are too many in this screen (I could at 8 different ones ..)? 

Screenshot 2025-10-07 at 17 06 30

ewels:
@pditommaso I'm not sure I understand? This PR uses standard config, there's nothing bespoke here:

* Creates `~/.nextflow/seqera_auth.config`
* Adds `includeConfig 'seqera_auth.config'` to `~/.nextflow/config`

Nextflow then loads `tower.accessToken` exactly as normal. No change to how env vars are used, or Nextflow config is parsed.

Example `seqera_auth.config` file:

```nextflow
// Seqera Platform configuration
tower {
    accessToken = 'eyJ........'
    endpoint = 'https://api.cloud.stage-seqera.io'
    enabled = true
    workspaceId = '123456789012345'  // org / workspace [Full name of workspace]
}
```

pditommaso:
I was missing this! 

> Adds includeConfig 'seqera_auth.config' to ~/.nextflow/config

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
```diff
@@ -0,0 +1,1354 @@
...
+        final json = new JsonSlurper().parseText(response.body()) as Map
+        return json.user as Map
+    }
+
+    protected Path getConfigFile() {
+        return Const.APP_HOME_DIR.resolve('config')
+    }
+
+    protected Path getAuthFile() {
+        return Const.APP_HOME_DIR.resolve('seqera_auth.config')
```
pditommaso: We tend to use kebab-case over snake_case, `seqera-auth.config` would be preferable

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
```diff
@@ -0,0 +1,1354 @@
...
+        ],
+        'https://api.cloud.stage-seqera.io': [
+            domain  : 'seqera-stage.eu.auth0.com',
+            clientId: '60cPDjI6YhoTPjyMTIBjGtxatSUwWswB'
+        ],
+        'https://api.cloud.seqera.io' : [
+            domain  : 'seqera.eu.auth0.com',
+            clientId: 'FxCM8EJ76nNeHUDidSHkZfT8VtsrhHeL'
+        ]
+    ]
```
pditommaso: Let's use existing library instead of having a custom one only for this 

https://github.com/nextflow-io/nextflow/blob/main/modules/nextflow/src/main/groovy/nextflow/platform/PlatformHelper.groovy#L23-L29

pditommaso: It would better to add to `PlatformHelper` a method `getAuthClientId` following the same pattern

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
```diff
@@ -0,0 +1,1354 @@
...
+        }
+
+        return builder.build()
+    }
+
+    @Override
+    void login(String apiUrl) {
+
+        // Check if TOWER_ACCESS_TOKEN environment variable is set
+        final envToken = SysEnv.get('TOWER_ACCESS_TOKEN')
```
pditommaso, line 53: Better to use `Platform.getAccessToken` 

https://github.com/nextflow-io/nextflow/blob/main/modules/nextflow/src/main/groovy/nextflow/platform/PlatformHelper.groovy#L39-L44

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
```diff
@@ -0,0 +1,1354 @@
...
+
+        // Wait for Enter key with proper interrupt handling
+        try {
+            final input = System.in.read()
+            if( input == -1 ) {
+                throw new AbortOperationException("Authentication cancelled")
+            }
+        } catch( InterruptedException e ) {
+            Thread.currentThread().interrupt()
+            throw new AbortOperationException("Authentication cancelled")
```
pditommaso, line 111: This should not be needed, the ` Thread.currentThread().interrupt()` cause the blocking thread to throw an exception

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
```diff
@@ -0,0 +1,1354 @@
...
+        } else {
+            // Linux and other Unix-like systems
+            final browsers = ["xdg-open", "firefox", "google-chrome", "chromium", "safari"]
+            for( browser in browsers ) {
+                try {
+                    new ProcessBuilder(browser, urlWithCode).start()
+                    browserOpened = true
+                    break
+                } catch( Exception ignored ) {
+                    // Try next browser
```
pditommaso, line 195: Always better to log the error for troubleshooting, even if expected

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
```diff
@@ -0,0 +1,1354 @@
...
+        // Save to config
+        saveAuthToConfig(pat, apiUrl)
+        ColorUtil.printColored("Personal Access Token saved to Nextflow auth config (${getAuthFile().toString()})", "green")
+    }
+
+    private String getWebUrlFromApiEndpoint(String apiEndpoint) {
+        // Convert API endpoint to web URL
+        // e.g., https://api.cloud.seqera.io -> https://cloud.seqera.io
+        //      https://cloud.seqera.io/api -> https://cloud.seqera.io
+        return apiEndpoint.replace('://api.', '://').replace('/api', '')
```
pditommaso, line 277: I would avoid this, custom can use unpredictable URI patterns

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
```diff
@@ -0,0 +1,1354 @@
...
+
+        if( !existingToken ) {
+            ColorUtil.printColored("WARN: No authentication token found in auth file.", "yellow bold")
+            println "Removing file: ${ColorUtil.colorize(authFile.toString(), 'magenta')}"
+            removeAuthFromConfig()
+            return
+        }
+
+        // Check if TOWER_ACCESS_TOKEN environment variable is set
+        final envToken = SysEnv.get('TOWER_ACCESS_TOKEN')
```
pditommaso, line 393: Use `PlatformHelper.getAccessToken`

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
```diff
@@ -0,0 +1,1354 @@
...
+        }
+
+        return [
+            value     : effectiveValue,
+            source    : source,
+            fromConfig: configValue != null || authValue != null,
+            fromEnv   : envValue != null,
+            isDefault : !authValue && !configValue && !envValue
+        ]
+    }
```
pditommaso: This looks a bit magic. Is there a solid unit test for it ?

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
```diff
@@ -0,0 +1,1354 @@
...
+            final client = createHttpClient()
+            final request = HttpRequest.newBuilder()
+                .uri(URI.create("${endpoint}/service-info"))
+                .GET()
+                .build()
+
+            final response = client.send(request, HttpResponse.BodyHandlers.ofString())
+            return response.statusCode() == 200
+        } catch( Exception e ) {
+            return false
```
pditommaso, line 1130: Log the exception for debugging

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
```diff
@@ -0,0 +1,1354 @@
...
+                return [
+                    orgName: ws.orgName,
+                    workspaceName: ws.workspaceName,
+                    workspaceFullName: ws.workspaceFullName
+                ]
+            }
+
+            return null
+        } catch (Exception e) {
+            return null
```
pditommaso, line 1179: Log the exception

pditommaso (changes requested):

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
ewels: Done in 0ae13cf

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
ewels: Done in 8f3e2eb9f

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
ewels: I don't think that we want to - this is just throwing a warning to the terminal if the env var is set, it's not actually using the env var. This won't work with `Platform.getAccessToken` as that returns either the env var or the config var.

However, whilst looking at this I did see some other code where this fix _should_ be implemented - fixed in cf05f3ffa

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
ewels: As above, this is checking specifically for the env var in order to show a warning. It's not using the value. See other comment reply and refactoring from elsewhere in cf05f3ffa

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
ewels: Removed in bb0ebe868

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
ewels: Added in 7c0c8e432 👍🏻

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
ewels: I don't think that there is any other way to get the web / html URLs without this. See Slack thread for further discussion.

Note that I took this approach from the Tower CLI, which [does the same thing](https://github.com/seqeralabs/tower-cli/blob/c420fcf5eead91ec4fd7cc6acf8734dc404f0d2d/src/main/java/io/seqera/tower/cli/commands/AbstractApiCmd.java#L430-L435): 

```java
    protected String serverUrl() {
        if (serverUrl == null) {
            serverUrl = app().url.replaceFirst("api\\.", "").replaceFirst("/api", "");
        }
        return serverUrl;
    }
```

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
ewels: Yeah this function grew a bit over time. The tests are pretty extensive though:

https://github.com/nextflow-io/nextflow/blob/7c0c8e43209805a2240baa38f5b8928c878629cf/plugins/nf-tower/src/test/io/seqera/tower/plugin/cli/AuthCommandImplTest.groovy#L354-L496

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
ewels: Done in 7c0c8e432

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
ewels: Done in 7c0c8e432

ewels:
@pditommaso thanks for the review - I believe that I have addressed all of your comments now.

I have leaned heavily into using `PlatformHelper` to fetch config attributes wherever they are used, so this acts as the single source of truth now.

There are two exceptions to this:
 - The `auth stats` table also grabs config manually so that it can show where things are defined (third table column says either `nextflow config` or `env var $TOWER_xx` - useful for debugging)
 - Several commands warn if an env var is found, as the recommended method is to use the Nextflow config and having both could result in confusing behaviour. These warnings grab the env var in order to tell if it's defined or not, but don't actually use its value.

Hope that makes sense and my edits are ok! 🙌🏻

modules/nextflow/src/main/groovy/nextflow/cli/CmdAuth.groovy
```diff
@@ -0,0 +1,273 @@
...
+
+
+        @Override
+        void usage(List<String> result) {
+            result << 'Authenticate with Seqera Platform'
+            result << "Usage: nextflow auth $name [-u <endpoint>]".toString()
+            result << ''
+            result << 'Options:'
+            result << '  -u, -url <endpoint>    Seqera Platform API endpoint (default: https://api.cloud.seqera.io)'
```
pditommaso: ideally this could come from PlatformHelper.getEndpoint

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
```diff
@@ -0,0 +1,1337 @@
...
+
+@Slf4j
+@CompileStatic
+class AuthCommandImpl implements CmdAuth.AuthCommand {
+
+    static final int API_TIMEOUT_MS = 10000
+    static final int AUTH_POLL_TIMEOUT_RETRIES = 60
+    static final int AUTH_POLL_INTERVAL_SECONDS = 5
+    static final int WORKSPACE_SELECTION_THRESHOLD = 8  // Max workspaces to show in single list; above this uses org-first selection
+    private static final String DEFAULT_API_ENDPOINT = 'https://api.cloud.seqera.io'
```
pditommaso: This should not be needed, as use instead `PlatformHelper.getEndpoint` as source of truth

modules/nextflow/src/main/groovy/nextflow/platform/PlatformHelper.groovy
```diff
@@ -27,6 +27,44 @@ class PlatformHelper {
...
+        switch(endpoint) {
+            case 'https://api.cloud.dev-seqera.io':
+                return 'seqera-development.eu.auth0.com'
+            case 'https://api.cloud.stage-seqera.io':
+                return 'seqera-stage.eu.auth0.com'
+            case 'https://api.cloud.seqera.io':
+                return 'seqera.eu.auth0.com'
+            default:
+                return null
+        }
```
pditommaso, line 49: This and (and getAuthClientId) should use an env variable and fallback to prod host as default e.g 

```
return SysEnv.get('TOWER_AUTH_DOMAIN') ?: 'seqera.eu.auth0.com'
```

Both for for internal usage third-parties on-prem installation

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
ewels: Parking for now - will hopefully get a new API field to provide this value directly in the near-ish future.

modules/nextflow/src/main/groovy/nextflow/platform/PlatformHelper.groovy
ewels: So you mean introduce new env vars for this purpose, for local testing?

Can do, actually I originally did it this way. It's just a bit annoying because you need to set API URL, auth domain and auth key every time you want to switch. But I guess it's not so many people wanting to test on staging.

ewels: Oh and you mean for third-parties to auth with their own Auth0? Good point.. Ok 👍🏻

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
ewels: Right, leftover from when I refactored last night sorry 👍🏻

ewels:
> it would also be nice reduce the number of colors [(ma](https://github.com/nextflow-io/nextflow/pull/6380#issuecomment-3377348518)

Not sure if you tested again after my changes, but I did make a start on toning the colours down a bit in 84c35847ef4c2abf0d2099f49eb210ed95b2afad already.

Main thing is that on my terminal, `cyan` is very similar to the default colour. You have the default terminal to bright green though, so the two look very different. I switched many instances of `cyan` out so it should look a bit better already.

I can do more to further reduce use of colour if you like. Note that `NO_COLOR` and the usual ANSI flags should work to totally remove colour if you want.

modules/nextflow/src/main/groovy/nextflow/cli/CmdAuth.groovy
ewels: Done in 1adf5223b

plugins/nf-tower/src/main/io/seqera/tower/plugin/cli/AuthCommandImpl.groovy
ewels: Done in 95caa00bc

modules/nextflow/src/main/groovy/nextflow/platform/PlatformHelper.groovy
```diff
@@ -27,6 +27,44 @@ class PlatformHelper {
...
     }
 
+    /**
+     * Get the Auth0 domain for a given Platform API endpoint
+     *
+     * @param endpoint the Platform API endpoint
+     * @return the Auth0 domain, or null if not a cloud endpoint
+     */
+    static String getAuthDomain(String endpoint) {
+        switch(endpoint) {
```
ewels, line 38: @pditommaso is it sufficient to add this env var and leave the other values hardcoded for convenience like this?

```suggestion
        switch(endpoint) {
            case SysEnv.get('TOWER_AUTH_DOMAIN'):
                return SysEnv.get('TOWER_AUTH_DOMAIN')
```

modules/nextflow/src/main/groovy/nextflow/platform/PlatformHelper.groovy
ewels: See https://github.com/nextflow-io/nextflow/pull/6380#discussion_r2415054570

pditommaso:
ok, i'll check again the colours. Do you mind to change the base branch to a non-master for a last cleanup round?

ewels:
Changed the PR target to a new `nextflow-auth` branch 👍🏻  (you should also be able to push directly to my fork to keep on this PR if you want).

Saw you merged the `TOWER_AUTH_DOMAIN` suggestion - should probably also comparable for `TOWER_AUTH_ID` and document both.

PR #6387, opened 2025-09-06
Refactor TowerClient onFlowComplete method for improved ordering

pditommaso:
## Summary
- Remove unused `terminated` field that was set but never used
- Reorganize `onFlowComplete` method to ensure sender operations complete before workflow completion notification  
- Move reports publishing and completion after workflow completion notification for better sequencing

## Test plan
- [ ] Verify existing TowerClient tests pass
- [ ] Confirm workflow completion flow works correctly with Tower integration
- [ ] Check that reports are properly published after completion

### DIFF
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
@@ -395,22 +395,22 @@ class TowerClient implements TraceObserverV2 {
      */
     @Override
     void onFlowComplete() {
+        // publish runtime reports
+        reports.publishRuntimeReports()
+        // submit the completion record
         if( sender ) {
-            // submit the completion record
             events << new ProcessEvent(completed: true)
             // wait the submission of pending events
             sender.join()
         }
+        // wait and flush reports content
+        reports.flowComplete()
         // notify the workflow completion
         if( workflowId ) {
             final req = makeCompleteReq(session)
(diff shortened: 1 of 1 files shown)


bentsherman:
@claude please review

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
```diff
@@ -397,22 +395,22 @@ class TowerClient implements TraceObserverV2 {
...
-        terminated = true
         if( workflowId ) {
             final req = makeCompleteReq(session)
             final resp = sendHttpMessage(urlTraceComplete, req, 'PUT')
             logHttpResponse(urlTraceComplete, resp)
         }
+        // publish runtime reports
+        reports.publishRuntimeReports()
+        // wait and flush reports content
+        reports.flowComplete()
```
bentsherman: why did you move the reports publishing after the completion event?

pditommaso: my rationale was the reports should be created once the execution is done. but think more this event signal the completion by definition, therefore does not really depend on this, on the other hand platform can requires some of those reports, therefore it may be better to have the reports generation as the first thing in this method

pditommaso: I've restored as it was

PR #6381, opened 2025-09-04
Implement Process Entry Execution with Command-Line Parameter Mapping (aka module execution) 

pditommaso:
## Summary

This PR implements a new feature that allows direct execution of individual Nextflow processes without requiring explicit workflow definitions. Users can now run processes directly from the command line with automatic parameter mapping.

### Key Features

- **Automatic Single Process Execution**: Scripts with one process run automatically
- **Automatic Multi-Process Execution**: Scripts with multiple processes run the first process automatically  
- **Command-Line Parameter Mapping**: Automatic mapping of `--param value` arguments to process inputs
- **Full Input Type Support**: Supports `val`, `path`, `env`, `tuple`, and `each` input types

### Single Process Script

Create a script with one process (auto-execution):

```groovy
#!/usr/bin/env nextflow

process analyzeData {
    debug true
    
    input:
    val sampleName
    path inputFile
    
    script:
    """
    echo "Analyzing sample: ${sampleName}"
    echo "Input file: ${inputFile}"
    wc -l ${inputFile}
    """
}
```

**Execute with:**
```bash
nextflow run analyze.nf --sampleName "sample1" --inputFile "data.txt"
```

### Multi-Process Script (Auto-runs First Process)

Scripts with multiple processes automatically execute the first process defined:

```groovy
#!/usr/bin/env nextflow

process preprocessData {
    debug true
    input: 
    path inputFile
    val quality
    script: """
    echo "Preprocessing ${inputFile} with quality ${quality}"
    head -n 5 "${inputFile}"
    """
}

process analyzeResults {
    debug true
    input: 
    val experimentId
    path resultsFile
    script: """
    echo "Analyzing experiment ${experimentId}"
    tail -n 3 "${resultsFile}"
    """
}
```

**Execute with:**
```bash
# Automatically runs preprocessData (first process)
nextflow run multi.nf --inputFile "data.txt" --quality "high"
```

### Comprehensive Input Types Example

```groovy
#!/usr/bin/env nextflow

process complexProcess {
    debug true
    
    input:
    val sampleId
    path inputFile
    env DATABASE_URL
    tuple val(batchId), val(priority) 
    each item
    
    script:
    """
    echo "Sample: ${sampleId}"
    echo "File: ${inputFile}"  
    echo "Database: ${DATABASE_URL}"
    echo "Batch: ${batchId}, Priority: ${priority}"
    echo "Item: ${item}"
    """
}
```

**Execute with:**
```bash
[...]

PR #6368, opened 2025-08-27
Typed processe

bentsherman:
This PR introduces a new syntax for process which uses typed inputs and outputs. The existing syntax is still supported.

This PR refactors several large classes -- namely `ProcessConfig` and `TaskProcessor` -- to better separate concerns and enable a v1 / v2 model for process inputs/outputs. When moving existing code to new files, I try to change it as little as possible to not break anything.

### ProcessConfig refactor

The following new classes were spun out of `ProcessConfig`:

- `ProcessConfigV1` / `ProcessConfigV2` extend `ProcessConfig` with the declared inputs / outputs based on legacy (v1) or typed (v2) semantics

- `ProcessDslV1` / `ProcessDslV2` are builder DSLs for legacy / typed process definitions 

- `ProcessConfigBuilder` is an adapter for applying process configuration to a process definition

- `ProcessBuilder` is the base builder class used by the above builders

### TaskProcessor refactor

The following new classes were spun out of `TaskProcessor`:

- `TaskInputResolver` implements the input file resolution from `makeTaskContextStage2()`

- `TaskOutputResolver` implements the task output resolution logic for typed processes

- `TaskEnvCollector` implements the output env/eval resolution from `collectOutEnvMap()`

- `TaskFileCollector` implements the output file resolution from `collectOutFiles()`

### Typed inputs / outputs

The following new classes implement the new behavior for typed inputs / outputs:

- `ProcessInputs` and `ProcessOutputs` replace `InputsList` and `OutputsList` from the v1 model

- `ProcessInput` and `ProcessOutput` replace all `InParam` and `OutParam` classes from the v1 model

- `ProcessFileInput` and `ProcessFileOutput` replace `FileInParam` and `FileOutParam` in the v1 model

### Backwards compatibility

The runtime supports both legacy (v1) and typed (v2) processes by creating the ProcessDef with either a ProcessConfigV1 or ProcessConfigV2.

ProcessDef, TaskProcessor, and TaskRun check this type to determine whether to use v1 or v2 semantics. An `instanceof` check is performed at these decision points:

```groovy
if( config instanceof ProcessConfigV1 )
    // use legacy inputs/outputs
if( config instanceof ProcessConfigV2 )
    // use typed inputs/outputs
```

Based on initial work in #4553

TODO:
- [ ] update docs
- [ ] update tests
- [ ] add e2e tests

PR #6391, opened 2025-09-09
Simplify process entry execution

bentsherman:
This PR simplifies the parsing of process inputs in #6381 by applying a refactor that was originally developed for #6368 .

Instead of creating an on-the-fly closure delegate in the process entry handler, it is much simpler to use the process inputs after they are parsed into `InParam`s.

The problem is that the process inputs normally aren't constructed until the process is invoked by a workflow. Instead, we need to construct them when the ProcessDef is constructed.

At the same time, the process config still needs to be applied at invocation, since config settings can use `withName` selectors which depend on the execution context.

To do this, we factor out several classes from ProcessConfig:

- `ProcessDsl`: defines the DSL for inputs, outputs, and directives
- `ProcessConfigBuilder`: provides functions for applying process config settings
- `ProcessBuilder` base class containing shared logic for these two classes

This allows us to build the ProcessDef and resolve the process body immediately, while still applying the process config only after the process is invoked. It also greatly simplifies the process entry handler.

NOTE: In #6368 , ProcessDsl becomes ProcessDslV1, and a separate ProcessDslV2 is defined for the static type syntax.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/BaseScript.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/BaseScript.groovy
@@ -25,7 +25,7 @@ import nextflow.NF
 import nextflow.NextflowMeta
 import nextflow.Session
 import nextflow.exception.AbortOperationException
-import nextflow.script.dsl.ProcessDsl
+import nextflow.script.dsl.ProcessDslV1
 import nextflow.secret.SecretsLoader
 
 /**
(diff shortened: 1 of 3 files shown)


ewels:
Would love to see CLI syntax to select process by name, not just file path. Command gets much cleaner, also means that there doesn't need to be a one-process-per-file limitation.

eg. instead of:
```bash
nextflow run https://github.com/nf-core/modules/blob/master/modules/nf-core/samtools/index/main.nf --input mysample.bam
```

Could then have:
```bash
nextflow run nf-core/modules:SAMTOOLS_INDEX --input mysample.bam
```

bentsherman:
Closing in favor of #6368

Issue #6394, opened 2025-09-10
Improve the new workflow output feature by making it easier to give a header to CSV index files


cyrilcros:
## New feature

Relating to https://www.nextflow.io/docs/latest/tutorials/workflow-outputs.html#generating-an-index-file

Let's say I have an output channel that yields values like `output: tuple val(meta), path(my_output.txt),path(my_output.log)` where `meta` is like `[id: foo, date: today]`. Right now that can be saved as a CSV index file. Unfortunately setting `header = true` is finicky, it will error out because it only wants maps.

## Use case 

You can get as a workflow output a CSV sample sheet that's in a nice flat tabular format ready to be ingested into a downstream workflow or Python/R/whatever for analysis. 

## Suggested implementation 

Let's say you have a block like 

```
output {
    save_my_output {
        path {meta, _result, _log -> "${meta.date}_${meta.id}" }
        index {
            path 'outs.csv'
            header true
        }
    }
}
```

That would error out with header true because I am not providing a proper map with `[id: foo, date: today], output.txt, output.log`. Instead I would like the first key `meta` to be flattened if needed and the other column names to be used to automatically get columns `id, date, result, log` based on the closure argument variables.
Maybe as an option called `parse_path_closure`.

Thank you!

bentsherman:
The best practice here is to transform the tuples into maps when publishing the channel:

```groovy
workflow {
  main:
  my_output_ch = // ...

  publish:
  my_output = my_output_ch.map { meta, result, log -> meta + [result: result, log: log] }
}

output {
  my_output {
    path { sample -> "${sample.date}_${sample.id}" }
    index {
      path 'outputs.csv'
      header true
    }
  }
}
```

This avoids the tuple issue, but it also (1) allows you to define column names for the file outputs and (2) sets you up for record types in the future.

Once we have record types, you'll be able to replace tuple channels with record channels throughout the entire pipeline, so you won't need this extra conversion in the `publish:` section. But I think this is the best approach in the interim.

cyrilcros:
Thank you for the answer, I hadn't realized you could use `map` in the `publish:` section.

PR #6397, opened 2025-09-11
Update workflow-outputs.md to explain how to deal with meta maps

cyrilcros:
Documenting the kind answer I got in https://github.com/nextflow-io/nextflow/issues/6394

- Explaining how to deal with a `meta`  map
- fixing a random typo.

### DIFF
--- a/docs/tutorials/workflow-outputs.md
+++ b/docs/tutorials/workflow-outputs.md
@@ -214,7 +214,7 @@ results
     └── spleen
 ```
 
-We can achieve this directory structure by customzing the `output` block.
+We can achieve this directory structure by customizing the `output` block.
 
 First, update the `FASTQC` and `QUANT` processes to also emit the sample ID alongside the output files:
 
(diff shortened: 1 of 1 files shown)


bentsherman:
We already show how to transform the tuples into maps in the code example before your edits:

```groovy
    samples_ch = RNASEQ.out.fastqc
        .join(RNASEQ.out.quant)
        .map { id, fastqc, quant ->
            [id: id, fastqc: fastqc, quant: quant]
        }
```

cyrilcros:
Yes, personally I did not realize how to wrap a previously written workflow producing a `tuple val(meta), path(foo),...` into the new output syntax after reading the docs. Obviously feel free to close if this is redundant. I get what features the new output system bring with a simple example like `sample_ch`. I am more wondering in what direction you want to take the new output system when you chain workflows, and if I should go for `map`-style outputs until you have a record type in place.

bentsherman:
Thanks for the suggestion in any case. We're planning to make another pass on the workflow output guide, so I might find a way to include a specific example with a meta map.

I think you might as well transform tuple channels into map channels at the end, if you go ahead with workflow outputs, because it gives you a better index file. But it will also set you up nicely for record types.

Record types will be necessary to enable pipeline chaining -- basically, an output will specify a record type, which will allow it to be chained with a samplesheet input in a downstream pipeline:

```groovy
// pipeline A
output {
  samples: Channel<Sample> {
    index { path 'samples.json' }
  }
}

// pipeline B
params {
  input: List<Sample> // loaded from samples.json with matching schema
}

// shared
record Sample {
  id: String
  fastq_1: Path
  fastq_2: Path
}
```

This is just a sketch, but hopefully it shows how the record type is essential for mapping complex inputs/outputs like samplesheets.

PR #6408, opened 2025-09-16
Fix false error with storeDir and val outputs

### LINKED ISSUE #5785: topic channel output breaks storeDir

**Expected**

- including a storeDir directive in a process will catch only file outputs. I expected it to ignore topic channels since they are collected elsewhere.

**Actual**
 
- storeDir doens't ignore topic channels and fails to store any files:

```

nextflow.preview.topic = true

process FOO {

	storeDir 'output'

	output:
	path "output.txt"
	tuple val(task.process), val('grep'), eval('grep --version | head -n 1'), topic: versions

	script:
	"""
	touch output.txt
	"""
}

workflow {
	FOO()
}

```

### Program output 

```
WARN: [FOO] StoreDir can only be used when using 'file' outputs

ERROR ~ No such file or directory: /home/cormac/sandbox/output/.command.env
```

### Environment 

* Nextflow version: 24.10.4.5934
* Java version: openjdk version "21.0.6" 2025-01-21
* Operating system: Linux

bentsherman:
Close #5785 

This PR fixes a false positive error with storeDir that causes it to be overly strict with value outputs.

Currently storeDir only works with value outputs of the form `val(<name>)`, where `<name>` is defined as an input or in the script block. However, any value output should be allowed, because value outputs are always evaluated and do not depend on the storeDir or resume cache.

Test case:

```groovy
process FOO {
    storeDir 'store'

    output:
    tuple val(task.process), val('grep'), path('version.txt')

    script:
    """
    grep --version | head -n 1 > version.txt
    """
}

workflow {
    FOO()
}
```

Currently, the two value outputs above will caused the task to not be stored. With this PR, the task is stored correctly.


jorgee:
It would be good to add the test case as integration test

bentsherman:
TODO: update `dynamic-storedir.nf` to include this case

Issue #6405, opened 2025-09-15
"Session aborted -- Cause: Strings must not be null" error when specifying non-existent entry


robsyme:
# Bug report

## Expected behavior and actual behavior

**Expected behavior:** When specifying a non-existent workflow entry name with the `-entry` option, Nextflow should display a clear error message like "Unknown workflow entry name: foo" with helpful suggestions for similar workflow names.

**Actual behavior:** Nextflow crashes with a cryptic "Session aborted -- Cause: Strings must not be null" error instead of providing a helpful error message about the invalid entry name.

## Steps to reproduce the problem

1. Use any simple Nextflow pipeline (e.g., `nextflow-io/hello`)
2. Run with a non-existent entry name:
   ```bash
   nextflow run hello -entry foo
   ```

This can be reproduced with any pipeline by specifying an invalid entry name.

## Program output

```
Sep-15 16:55:32.400 [main] DEBUG nextflow.Session - Session aborted -- Cause: Strings must not be null
```

The complete log shows the session aborts during script parsing rather than providing a user-friendly error message about the invalid workflow entry name.

## Environment

* Nextflow version: 25.04.3
* Java version: OpenJDK 64-Bit Server VM 21.0.7+6-LTS
* Operating system: Linux/macOS (reproduced on both)
* Bash version: N/A (not bash-related)

## Additional context

**Root cause analysis:**

The error occurs in the workflow entry name validation logic in `BaseScript.groovy:180-186`. When Nextflow tries to provide helpful suggestions for the invalid entry name using the `closest()` method from `Bolts.groovy:739`, a Guava Preconditions validation fails with "Strings must not be null".

The issue is in this code path:
```groovy
// BaseScript.groovy line ~183
final guess = allNames.closest(binding.entryName)
```

The `closest()` method calls `StringUtils.getLevenshteinDistance()` but appears to encounter null values in the workflow names collection, triggering a Guava validation failure.

**Impact:**
- Poor user experience with cryptic error messages
- Difficulty debugging workflow entry issues
- Session abort instead of graceful error handling

**Suggested fix:**
Add null-safety checks in the `closest()` method or the workflow name suggestion logic to handle cases where workflow metadata might contain null values.

This affects the core user experience when making common mistakes like typos in entry point names.

gwright99:
[...]

PR #6409, opened 2025-09-16
Add unit test for invalid workflow entry name error handling

pditommaso:
## Summary

This PR adds a comprehensive unit test to validate the fix implemented in PR #6404 that prevents "Strings must not be null" errors when an invalid workflow entry name is specified via the `-entry` parameter.

**Related Issues:**
- Fixes/Tests #6405 
- Related to PR #6404

## Background

The original issue (#6405) occurred when running:
```bash
nextflow run script.nf -entry invalidEntryName
```

Instead of showing a helpful error message like "Unknown workflow entry name: invalidEntryName", users would see a confusing "Strings must not be null" error. This was fixed in PR #6404, and this PR adds comprehensive tests to prevent regression.

### 🧪 **New Unit Test** (`ScriptDslTest.groovy`)
- Added `should show proper error message for invalid entry name` test
- Validates that invalid entry names throw `IllegalArgumentException` with proper error message
- Ensures the confusing "Strings must not be null" error is never shown
- Tests the workflow name suggestion feature works correctly

### 🔧 **Enhanced Fix** (`ScriptMeta.groovy`)
- Extended the null-check fix to also cover the imports section in `getWorkflowNames()`
- This ensures comprehensive null filtering for both local workflow definitions and imported workflows
- Line 264: Added `&& item.name` check for imported workflows (matching the existing fix for local definitions)

## Test Validation

The test has been validated by:
1. ✅ **Passes with the fix** - Test passes when null checks are present
2. ❌ **Fails without the fix** - Test correctly fails when null checks are removed, proving it detects the regression
3. 🔄 **Fix restored** - Test passes again when null checks are restored

## Test Details

```groovy
def 'should show proper error message for invalid entry name' () {
    when:
    dsl_eval('invalidEntry', '''
        workflow validWorkflow {
          /echo valid/
        }
        
        workflow {
          /echo default/
        }
    ''')
    
    then:
    def err = thrown(IllegalArgumentException)
    err.message.contains('Unknown workflow entry name: invalidEntry')
    \!err.message.contains('Strings must not be null')
}
```

## Impact

- ✅ Prevents regression of the "Strings must not be null" bug from #6405
- ✅ Ensures proper error messages for invalid workflow entry names
- ✅ Maintains backward compatibility
- ✅ No breaking changes

This test will catch any future regressions and ensures users always receive clear, actionable error messages when specifying invalid workflow entry names.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy
@@ -261,7 +261,7 @@ class ScriptMeta {
         }
         // processes from imports
         for( def item: imports.values() ) {
-            if( item instanceof WorkflowDef && item.name )
+            if( item instanceof WorkflowDef )
                 result.add(item.name)
         }
         return result
--- a/modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy
@@ -256,12 +256,12 @@ class ScriptMeta {
         final result = new HashSet(definitions.size() + imports.size())
         // local definitions
         for( def item : definitions.values() ) {
-            if( item instanceof WorkflowDef && item.name != null )
+            if( item instanceof WorkflowDef && item.name )
                 result.add(item.name)
         }
         // processes from imports
         for( def item: imports.values() ) {
-            if( item instanceof WorkflowDef )
+            if( item instanceof WorkflowDef && item.name )
                 result.add(item.name)
         }
         return result
--- a/modules/nextflow/src/test/groovy/nextflow/script/ScriptDslTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/script/ScriptDslTest.groovy
@@ -642,7 +642,6 @@ class ScriptDslTest extends Dsl2Spec {
         then:
         def err = thrown(IllegalArgumentException)
         err.message.contains('Unknown workflow entry name: invalidEntry')
-        !err.message.contains('Strings must not be null')
     }
 
 }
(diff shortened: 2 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy
```diff
@@ -256,12 +256,12 @@ class ScriptMeta {
...
         // local definitions
         for( def item : definitions.values() ) {
-            if( item instanceof WorkflowDef && item.name != null )
+            if( item instanceof WorkflowDef && item.name )
                 result.add(item.name)
         }
         // processes from imports
         for( def item: imports.values() ) {
-            if( item instanceof WorkflowDef )
+            if( item instanceof WorkflowDef && item.name )
```
bentsherman:
```suggestion
            if( item instanceof WorkflowDef )
```

This is not needed

modules/nextflow/src/test/groovy/nextflow/script/ScriptDslTest.groovy
```diff
@@ -626,4 +626,23 @@ class ScriptDslTest extends Dsl2Spec {
...
+        
+        workflow {
+          /println 'default'/
+        }
+        ''')
+        
+        then:
+        def err = thrown(IllegalArgumentException)
+        err.message.contains('Unknown workflow entry name: invalidEntry')
+        !err.message.contains('Strings must not be null')
```
bentsherman:
```suggestion
```

This seems like an extraneous check

modules/nextflow/src/main/groovy/nextflow/script/ScriptMeta.groovy
pditommaso: Not strictly needed but still a good defensive approach.  Why omitting?

bentsherman: what are you defending against? it is impossible to import an entry workflow from another module

pditommaso: I don't fully agree, but it's really a minor point

Issue #6412, opened 2025-09-18
`GoogleBatchExecutor.groovy` does not respect `executor.$google_batch.[queueSize, pollInterval]`


mikesivley:
### Expected behavior and actual behavior

Setting `executor.$google_batch.queueSize` or `executor.$google_batch.pollInterval` in `nextflow.config` should override the [defaults](https://www.nextflow.io/docs/latest/reference/config.html#executor-specific-defaults).

These settings are ignored by GoogleBatchExecutor. The defaults are [hardcoded](https://github.com/nextflow-io/nextflow/blob/master/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchExecutor.groovy#L119) in the call to `TaskPollingMonitor.create`.

### Environment 

* Nextflow version: 25.04.6.5954 and 25.08.0-edge.5955
* Java version: openjdk 17.0.15-internal 2025-04-15
* Operating system: Linux
* Bash version: 5.1.16(1)-release (x86_64-pc-linux-gnu)

### Additional context

I modified the `GoogleBatchExecutor.groovy` to hardcode the `queueSize` value to `0` and confirmed that submissions to Google Batch are now unbounded by nextflow; exceeding 1000 concurrent jobs. I have not (yet) raised a PR to properly set the `queueSize` argument for `TaskPollingMonitor.create` according to `executor.$google_batch.queueSize`.

pditommaso:
The correct syntax is the following 

```
executor {
  '$google-batch' {
    queueSize = 10  
  }
}
```

The log indeed reports:  

```
Sep-21 16:47:07.311 [main] DEBUG n.processor.TaskPollingMonitor - Creating task monitor for executor 'google-batch' > capacity: 10; pollInterval: 10s; dumpInterval: 5m 

```

Closing this issue. Feel free to comment or re-open if needed

adamrtalbot:
If the format is invalid, maybe we should raise a warning when a user provides an invalid config item here?

```
executor.$google_batch.queueSize

WARN: 'executor.$google_batch' is not a valid config item, did you mean 'executor.$google-batch'?
```

mikesivley:
Thanks @pditommaso. I wasn't able to find any documentation on the proper syntax for the google-batch executor configuration, but my reason for using `$google_batch` is because `$google-batch` raises an error. I see from your comment that it should have been `'$google-batch'`

[...]

PR #6413, opened 2025-09-18
Fix hardcoded queue size in cloud executors

ejseqera:
Fix for #6412 where hardcoded queue size values in cloud executors that were ignoring Nextflow configuration settings.

- **Google Batch**: Now uses `config.getQueueSize(name, 1000)` instead of hardcoded 1000
- **Azure Batch**: Now uses `config.getQueueSize(name, 1000)` instead of hardcoded 1000  
- **Kubernetes**: Now uses `config.getQueueSize(name, 100)` instead of hardcoded 100

Previously, these executors ignored `executor.queueSize` or `executor.$<executor>.queueSize` settings, making config to throttle jobs ineffective.

This replaces hardcoded values with explicit configuration lookup while preserving original defaults that were set, [similar to AWS Batch](https://github.com/nextflow-io/nextflow/blob/24314b112f23d5ae17da9646f4442aee862e1392/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy#L198).
- Calls `config.getQueueSize(name, defQueueSize)`
- The `defQueueSize` parameter serves as the fallback default if no config is found/set
- Maintain configuration hierarchy: `executor.$<executor>.queueSize` → `executor.queueSize` → `NXF_EXECUTOR_QUEUESIZE `→ `defQueueSize`

### DIFF
--- a/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchExecutor.groovy
+++ b/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchExecutor.groovy
@@ -129,7 +129,8 @@ class AzBatchExecutor extends Executor implements ExtensionPoint {
 
     @Override
     protected TaskMonitor createTaskMonitor() {
-        TaskPollingMonitor.create(session, config, name, 1000, Duration.of('10 sec'))
+        final capacity = config.getQueueSize(name, 1000)
+        TaskPollingMonitor.create(session, config, name, capacity, Duration.of('10 sec'))
     }
 
     @Override
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchExecutor.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchExecutor.groovy
@@ -116,7 +116,8 @@ class GoogleBatchExecutor extends Executor implements ExtensionPoint, TaskArrayE
 
     @Override
     protected TaskMonitor createTaskMonitor() {
-        TaskPollingMonitor.create(session, config, name, 1000, Duration.of('10 sec'))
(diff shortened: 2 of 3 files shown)


pditommaso:
Not sure what this improves because `config.getQueueSize` is invoked as first thing by the create method 

https://github.com/nextflow-io/nextflow/blob/55f09132b540d8763c33bb95760a11099e098d54/modules/nextflow/src/main/groovy/nextflow/processor/TaskPollingMonitor.groovy#L169-L169

joshiesh:
@pditommaso Ah you're right, but then simplifying it to this should handle the config lookup issue correctly right?
```
TaskPollingMonitor.create(session, config, name, Duration.of('10 sec'))
```

mikesivley:
In my attempt to resolve the throughput issue, I did have to update my nextflow version from 25.04.6. to the latest edge release so that I could modify and install the latest plugin from `master`. It's possible that this issue had already been resolved in `master` and that the nextflow update itself is what solved my issue, not my change to the plugin. I have not tested the behavior of 25.08.0-edge without the modified plugin.

Indeed, the lines within `create` to fetch the config values were added by your August 7th commit, but lacking from the last stable release. All is well in the edge release.

mikesivley:
@ejseqera my [original issue](https://github.com/nextflow-io/nextflow/issues/6412) can be closed, but thank you for engaging with it so quickly!

pditommaso:
Unless I didn't understood correctly the problem, this should not be needed 

https://github.com/nextflow-io/nextflow/issues/6412#issuecomment-3316045824

pditommaso:
Closing as not required. Please reopen or comment if the solution in https://github.com/nextflow-io/nextflow/issues/6412 does not solve for you

PR #6426, opened 2025-09-25
Fix nf-tower to avoid skiping logs backup when no local cloud cache   

jorgee:
This PR fixes the problem where logs are not completely uploaded to the Platform at the end of the pipeline execution. The problem is produced because Platform is now using cloud cache, and the nf-tower cache backup command executed at the end of execution is skipping the logs upload when the local cache folder doesn't exist. It is printing a debug message, but it is not visible in the stdout. Moreover, the logs of this command were not uploaded to the Platform. So, it was not an easy way to see the error. 

This PR fixes the problem by uploading the logs first, moving the cache upload to the end. So, if the local cache folder does not exist, it will just skip the cache backup. 
The PR also improves logging. When the cloud cache path is not defined and the local folder doesn't exist, it prints a warning instead of a debug message.  In that case, the message will be printed in the stdout, making the problem visible to the user. When the cloud cache path is defined, we keep the debug message as the skip is expected, and no need to warn the user.

### DIFF
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/CacheManager.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/CacheManager.groovy
@@ -107,6 +107,27 @@ class CacheManager {
         if( !remoteWorkDir || !sessionUuid )
             return
 
+        if( Files.exists(localCachePath) ) {
+            // upload nextflow cache metadata
+            try {
+                log.info "Saving cache: ${localCachePath.toUriString()} => ${remoteCachePath.toUriString()}"
+                remoteCachePath.deleteDir()
+                remoteCachePath.parent.mkdirs()
+                FilesEx.copyTo(localCachePath, remoteCachePath)
+            }
+            catch( Throwable e ) {
+                log.warn "Failed to backup resume metadata to remote store path: ${remoteCachePath.toUriString()} — cause: ${e}", e
+            }
+        } else {
+            if( cloudCache ){
+                // Just log as debug message when localpath doesn't exist and cloud cache path is defined.
+                // It is likely the user has use cloud cache.
+                log.debug "Local cache path does not exist: $localCachePath — skipping cache backup"
+            } else {
+                // Print warning when localCachePath doesn't exists and cloud cache path is not defined.
+                log.warn "Local cache path does not exist: $localCachePath — skipping cache backup"
+            }
+        }
         // — upload out file
         try {
             if( localOutFile?.exists() )
@@ -146,29 +167,6 @@ class CacheManager {
         }
         catch (Throwable e) {
             log.warn "Unable to upload tower reprts file: $localTowerReports — reason: ${e.message ?: e}", e
-        }
-
-        if( !Files.exists(localCachePath) ) {
-            if( cloudCache ){
-                // Just log as debug message when localpath doesn't exist and cloud cache path is defined.
-                // It is likely the user has use cloud cache.
-                log.debug "Local cache path does not exist: $localCachePath — skipping cache backup"
-            } else {
-                // Print warning when localCachePath doesn't exists and cloud cache path is not defined.
-                log.warn "Local cache path does not exist: $localCachePath — skipping cache backup"
-            }
-            return
-        }
-
-        // upload nextflow cache metadata
-        try {
-            log.info "Saving cache: ${localCachePath.toUriString()} => ${remoteCachePath.toUriString()}"
-            remoteCachePath.deleteDir()
-            remoteCachePath.parent.mkdirs()
-            FilesEx.copyTo(localCachePath, remoteCachePath)
-        }
-        catch (Throwable e) {
-            log.warn "Failed to backup resume metadata to remote store path: ${remoteCachePath.toUriString()} — cause: ${e}", e
         }
     }
 
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/CacheManager.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/CacheManager.groovy
@@ -48,6 +48,7 @@ class CacheManager {
     @PackageScope Path localTowerConfig
     @PackageScope Path localTowerReports
     @PackageScope Path remoteWorkDir
+    @PackageScope boolean cloudCache = false
 
     @PackageScope Path getRemoteCachePath() { remoteWorkDir.resolve(".nextflow/cache/${sessionUuid}") }
     @PackageScope Path getRemoteOutFile() { remoteWorkDir.resolve(localOutFile.getName()) }
@@ -78,6 +79,8 @@ class CacheManager {
             localTowerConfig = Paths.get(env.TOWER_CONFIG_FILE)
         if( env.TOWER_REPORTS_FILE )
             localTowerReports = Paths.get(env.TOWER_REPORTS_FILE)
+        if( env.NXF_CLOUDCACHE_PATH )
+            cloudCache = true
     }
 
     protected void restoreCacheFiles() {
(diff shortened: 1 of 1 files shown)


pditommaso:
What am i'm missing to understand why the cache is not existing? That's supposed to be a s3 path like `s3://bucket-name/some/cache` right?

jorgee:
The cache manager uses `Const.appCacheDir` https://github.com/nextflow-io/nextflow/blob/defa8a6c8fe399df67b36e57f9c7b77704366aab/plugins/nf-tower/src/main/io/seqera/tower/plugin/CacheManager.groovy#L70
It gets NXF_CACHE_DIR, but the job defines NXF_CLOUDCACHE_PATH instead of NXF_CACHE_DIR. If I am not wrong, the backup is not needed if using cloud cache. Is it correct?

pditommaso:
`NXF_CACHE_DIR` is only used for testing. `NXF_CLOUDCACHE_PATH` is the s3 path provided by Platform where to keep cache metadata. The cache metadata are the leveldb files that are in any case written local, and therefore needs to be uploaded to NXF_CLOUDCACHE_PATH on completion and restored before resuming 

See this test https://github.com/nextflow-io/nextflow/blob/a5756da3e8a8bdfa55bc6ef36c56cd5e33454545/validation/awsbatch.sh#L46-L62

jorgee:
The `.netxflow/cache/uuid ` folder is created in `DefaultCacheStore` but if you run with` nf-cloudcache` you use `CloudCacheStore` that write directly to the cloud cache path. 

https://github.com/nextflow-io/nextflow/blob/4b00170a30cc23dc3e2dde35ba16cc8d25110190/modules/nextflow/src/main/groovy/nextflow/cache/DefaultCacheStore.groovy#L63-L74

pditommaso:
so my point is, why `localCachePath` does not exist? is not that unexpected?

bentsherman:
It's an unfortunate duplication but `NXF_CACHE_DIR` is only used to configure the local cache path while `NXF_CLOUDCACHE_PATH` is only used to configure the cloudcache path. In theory it could be one setting that delegates to local or cloudcache based on the URI scheme but we didn't do it that way

plugins/nf-tower/src/main/io/seqera/tower/plugin/CacheManager.groovy
```diff
@@ -104,22 +107,6 @@ class CacheManager {
...
-        // upload nextflow cache metadata
-        try {
-            log.info "Saving cache: ${localCachePath.toUriString()} => ${remoteCachePath.toUriString()}"
-            remoteCachePath.deleteDir()
-            remoteCachePath.parent.mkdirs()
-            FilesEx.copyTo(localCachePath, remoteCachePath)
-        }
-        catch (Throwable e) {
-            log.warn "Failed to backup resume metadata to remote store path: ${remoteCachePath.toUriString()} — cause: ${e}", e
-        }
```
pditommaso, line 121: Is not enough wrapping this into a `if( Files.exists(localCachePath) ) {`

jorgee: It is just wrapping if we do not want to warn/log when localCachePath doesn't exist. If you prefer to do not move the code I can just implement it with an if-else.

pditommaso: let's do the if-else approach

jorgee: changed

plugins/nf-tower/src/main/io/seqera/tower/plugin/CacheManager.groovy
```diff
@@ -104,22 +107,27 @@ class CacheManager {
...
+            }
+        } else {
+            if( cloudCache ){
+                // Just log as debug message when localpath doesn't exist and cloud cache path is defined.
+                // It is likely the user has use cloud cache.
+                log.debug "Local cache path does not exist: $localCachePath — skipping cache backup"
+            } else {
+                // Print warning when localCachePath doesn't exists and cloud cache path is not defined.
+                log.warn "Local cache path does not exist: $localCachePath — skipping cache backup"
+            }
```
pditommaso, line 129: let's make this simpler and use `log.info` for both

jorgee: Don't you think this message is going to warn the users? I did the differentiation to avoid printing a message in the output of the pipeline when the skip is expected.

pditommaso: That's only used by Platform, if I'm not wrong

jorgee: Yes, but if we do not do the distinction, platform users will see a message like "Local cache path does not exist: .nextflow/cache/xxxx — skipping cache backup" at the end of every pipeline execution. If I were a user, I would be worried about it.

pditommaso: Only in case of failure, no?

jorgee: Let me explain my reasoning in more detail:
The command is in the EXIT trap; it will always be executed.
Plaform executions always use nf-cloudcache, so it uses CloudCacheStore instead of DefaultCacheStore. 
As it is not using the DefaultCache store, the .`nextflow/cache/xxx` folder will never be created. 
Therefore, all platform executions will get this message. 
I am wrong at some point?

pditommaso: ~~The `.nextflow/cache` is created in normal conditions even with cloudcache, and therefore needs to be upload to the S3 path.~~

My understanding that in case of failure it may not exist, in this case make sense report an info message. 

Not understanding why this change is done (at least only) in `CacheManager`, when stating this, with ClaudeCache, `LogsHandler` is used instead 

https://github.com/nextflow-io/nextflow/blob/fix-tower-cache-manager-logs-upload-skip/plugins/nf-tower/src/main/io/seqera/tower/plugin/CacheCommand.groovy#L45-L54

jorgee: Ah, I get your point, this is the real issue. I saw the stacktrace and I went directly to CacheManager without noting it. There should be a difference between how the session is created when executing the CmdRrun or CacheCommand. I will figure out what.

jorgee: And here is the difference

https://github.com/nextflow-io/nextflow/blob/66cfeac9b0573249eb501669de8a81ed80027656/modules/nextflow/src/main/groovy/nextflow/config/ConfigBuilder.groovy#L830C5-L839C6

NXF_CLOUDCACHE_PATH is only evaluated in configRunOptions that is only called when CmdRun. In fact the env variables are only evaluated at CmdRun but not in the other cases.

pditommaso: oh, therefore `getSession().cloudCachePath` is not set!

pditommaso: @jorgee I was wrong in [this comment](https://github.com/nextflow-io/nextflow/pull/6426#discussion_r2381814158).

When using cloudcache the `.nextflow/cache` directory is not supposed to exists, because it's replace by the cache on S3 😮‍💨

pditommaso:
I went for a dirty patch https://github.com/nextflow-io/nextflow/pull/6432

pditommaso:
Closing in favour of https://github.com/nextflow-io/nextflow/pull/6432

PR #6428, opened 2025-09-25
Determine Azure directories more accurately using metadata and blob properties

### LINKED ISSUE #6427: Azure path incorrectly determined to be directory by trailing slash
## Bug report 

If using nf-schema, Nextflow will determine an Azure path to be a directory if it's missing a trailing slash:

```
ERROR ~ Validation of pipeline parameters failed!

The following invalid input values have been detected:

* --outdir (az://pipeline/output): 'az://pipeline/output' is not a directory, but a file

 -- Check script '/home/username/.nextflow/assets/nf-core/rnaseq/subworkflows/nf-core/utils_nfschema_plugin/main.nf' at line: 39 or see '.nextflow.log' file for more details
```

nf-schema checks `Nextflow.file.directory` for whether a file is a directory, but for Azure this is inaccurate because it's just checking for the trailing slash:

https://github.com/nextflow-io/nextflow/blob/a5756da3e8a8bdfa55bc6ef36c56cd5e33454545/plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzFileAttributes.groovy#L61

For most situations, this is fine but when using nf-schema this becomes an issue because it halts the pipeline.

Instead, we should be able to have a more robust way of checking if the file is a directory on Azure, it's probably possible using the SDKs.

I will add a bypass for nf-schema.

`main.nf`:
```
// Print the parameters for inspection
log.info """
Parameters:
s3_file: ${params.s3_file ?: 'not set'}
gs_file: ${params.gs_file ?: 'not set'}
az_file: ${params.az_file ?: 'not set'}
"""

workflow {
    println "Cloud path validation example completed successfully!"
}
```

`nextflow_schema.json`:
```
{
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "$id": "https://raw.githubusercontent.com/nf-core/nf-schema/master/examples/cloudpath/nextflow_schema.json",
    "title": "Cloud path validation example",
    "description": "Example demonstrating cloud path validation",
    "type": "object",
    "properties": {
        "s3_file": {
            "type": "string",
            "format": "file-path",
            "description": "An S3 file path"
        },
        "gs_file": {
            "type": "string",
            "format": "file-path",
            "description": "A Google Cloud Storage file path"
        },
        "az_file": {
            "type": "string",
            "format": "file-path",
            "description": "An Azure Blob Storage file path"
        }
    }
}
```

### Program output 

(Copy and paste the output produced by the failing execution. Please highlight it as a code block. Whenever possible upload the `.nextflow.log` file.)

### Environment 

* Nextflow version: 25.04.7 build 5955 
* nf-schema version: 2.5.1
* Java version: openjdk version "17.0.3" 2022-04-19
* Operating system: macOS
* Bash version: zsh

### Additional context

Slack thread: https://nextflow.slack.com/archives/C02TPRRLCF4/p1758724547528979?thread_ts=1727918072.021049&cid=C02TPRRLCF4
nf-schema issue: https://github.com/nextflow-io/nf-schema/issues/16

adamrtalbot:
## Problem

On Azure, `AzPath.isDirectory()` previously relied only on whether the path string ended in `/`. A real virtual directory referenced without that suffix, such as `az://pipeline/output`, was therefore reported as a file.

nf-schema validates `directory-path` values through `Nextflow.file(value).isDirectory()`, so `--outdir az://pipeline/output` could be rejected even when the prefix contained output files.

Fixes #6427

## Fix

- For a slashless path, `AzPath.isDirectory()` resolves the existing file attributes and uses their directory status.
- Paths created from a listing reuse their cached attributes without consuming the normal one-shot cache. A bare slashless path stores its resolved attributes on the path instance, allowing repeated `isDirectory()` calls to reuse them. A later `readAttributes()` call reuses and consumes that cache; a subsequent `isDirectory()` lookup may therefore query storage again.
- Relative paths return `false` without attempting an invalid container lookup.
- Hierarchical blob listings request metadata, allowing `hdi_isfolder=true` markers to reach `AzFileAttributes`.
- Direct blob properties and listing results recognise an `hdi_isfolder=true` marker as a directory only when its content length is zero.

The underlying attribute resolution detects virtual directories from child blobs, explicit trailing-slash markers, and Azure Data Lake Storage Gen2 `hdi_isfolder` metadata. Lookup failures continue to propagate consistently with `readAttributes()`.

## Validation

- `./gradlew :plugins:nf-azure:test --rerun-tasks --console=plain`
- `AzNioTest` independently exercises both `AzPath.isDirectory()` and `Files.isDirectory()` for a virtual directory referenced without a trailing slash.
- `git diff --check`


plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzFileAttributes.groovy
```diff
@@ -58,23 +58,42 @@ class AzFileAttributes implements BasicFileAttributes {
...
+            // Without metadata, default to treating as file
+            // This aligns with Azure SDK's approach where explicit directory markers are required
+            directory = false
+            size = props.getBlobSize()
         }
     }
 
     AzFileAttributes(String containerName, BlobItem item) {
         objectId = "/${containerName}/${item.name}"
-        directory = item.name.endsWith('/')
```
adamrtalbot, line 73: @bentsherman what will break if we incorrectly classify something as a file when it's a pseudo-directory? What's the impact of getting it wrong?

plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzPath.groovy
```diff
@@ -94,7 +94,34 @@ class AzPath implements Path {
...
+                def opts = new com.azure.storage.blob.models.ListBlobsOptions().setPrefix(prefix).setMaxResultsPerPage(1)
+                def hasChildren = containerClient.listBlobs(opts, null).stream().findFirst().isPresent()
+
+                if (hasChildren) {
+                    return true
+                }
+            }
+        }
+
+        return false
```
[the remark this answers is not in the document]
adamrtalbot: as opposed to false positives that cause entire pipelines to fail 🙄 

> It also has a logic flaw that can lead to incorrect directory detection 

Gonna need details Cursor.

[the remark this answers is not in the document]
adamrtalbot: That's literally the point of this PR.

plugins/nf-azure/src/test/nextflow/cloud/azure/nio/AzNioTest.groovy
```diff
@@ -944,4 +944,25 @@ class AzNioTest extends Specification implements AzBaseSpec {
         deleteBucket(bucket1)
     }
 
+    def 'should detect a virtual directory without a trailing slash (#6427)' () {
```
adamrtalbot:
```suggestion
    def 'should detect a virtual directory without a trailing slash' () {
```

plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzPath.groovy
```diff
@@ -94,7 +96,19 @@ class AzPath implements Path {
     }
 
     boolean isDirectory() {
-        return directory
+        if( directory )
+            return true
+        // A path without a trailing slash may still be a (virtual) directory in blob storage.
```
adamrtalbot:
```suggestion
```

plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzPath.groovy
```diff
@@ -94,7 +96,19 @@ class AzPath implements Path {
...
+        // Consult the file system, which resolves this via the blob metadata and listing (see
+        // AzFileSystem.readAttributes), instead of relying on the trailing slash alone (#6427).
```
adamrtalbot:
```suggestion
        // Consult the file system, which resolves this via the blob metadata and listing (see
        // AzFileSystem.readAttributes), instead of relying on the trailing slash alone.
```

plugins/nf-azure/src/test/nextflow/cloud/azure/nio/AzPathTest.groovy
```diff
@@ -331,4 +331,22 @@ class AzPathTest extends Specification {
 
     }
 
+    def 'isDirectory should consult the attributes for a path without a trailing slash (#6427)'() {
```
adamrtalbot:
```suggestion
    def 'isDirectory should consult the attributes for a path without a trailing slash'() {
```

plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzFileAttributes.groovy
```diff
@@ -61,10 +61,14 @@ class AzFileAttributes implements BasicFileAttributes {
         directory = client.blobName.endsWith('/')
         size = props.getBlobSize()
 
-        // Support for Azure Data Lake Storage Gen2 with hierarchical namespace enabled
+        // Support for Azure Data Lake Storage Gen2 with hierarchical namespace enabled.
+        // Honor the `hdi_isfolder` marker regardless of the reported blob size: some Storage
+        // account versions report a non-zero size for the directory marker blob, so gating on
+        // `size == 0` would misclassify it as a file.
```
adamrtalbot:
```suggestion
        // Support for Azure Data Lake Storage Gen2 with hierarchical namespace enabled.
```

bentsherman:
can you summarize the latest updates?

pditommaso (commented):
Nice, focused fix — reusing `readAttributes`/`guessPath` instead of duplicating the detection logic is the right call, and the `hdi_isfolder` `String`→`boolean` coercion catch (`"false"` no longer coercing to `true`) is a genuine bug fix.

A few things worth addressing before merge:

### `isDirectory()` consumes the one-shot attribute cache

`fs.readAttributes(this)` calls `attributesCache()`, which returns *and nulls* the cached attributes. So for a slashless path that came from a listing and is actually a **file**, `isDirectory()` now drains that cache, and a later `readAttributes()`/`size()` re-hits the network — an extra round-trip that didn't exist before.

Reading the field directly avoids that, and also keeps `isDirectory()` zero-I/O for any path that already carries attributes (listing-sourced or trailing-slash). The only case that then touches the network is a bare slashless string path — exactly the #6427 scenario, where the lookup is unavoidable (object stores have no real directories):

```groovy
boolean isDirectory() {
    if( directory )
        return true
    if( directoryResolved != null )
        return directoryResolved
    // prefer already-cached attributes (from a listing) without consuming the
    // one-shot cache; only hit the network for a bare slashless path
    final attrs = attributes ?: fs.readAttributes(this)
    return directoryResolved = (attrs != null && attrs.isDirectory())
}
```

### Description contradicts the code on error handling

The description says lookup errors are *"caught and treated as 'not a directory'"* and that `isDirectory()` *"degrades gracefully to false rather than throwing"*. But there's no try/catch, and the new test `isDirectory should propagate the error when the attribute lookup fails` asserts `thrown(IllegalArgumentException)` — errors propagate. Since nf-schema calls this during validation, a non-404 error would throw out of validation rather than fail cleanly. Either add the catch to match the promised behavior, or fix the description so they agree.

### Stale description — `resolve`/`resolveSibling` changes aren't in the diff

The "Additional fixes surfaced during validation" section describes reworking `resolve`/`resolveSibling` to build on the current file system, but those methods are unchanged from master in this branch. Drop that section, or re-add the change if it was intended.

### Nit: `hdi_isfolder` on `BlobItem` needs metadata retrieval enabled

```groovy
directory = item.name.endsWith('/') || item.getMetadata()?.get("hdi_isfolder") == "true"
```

Azure omits blob metadata from list results unless `ListBlobsOptions` requests it (`BlobListDetails.setRetrieveMetadata(true)`). If the listings producing these `BlobItem`s don't request it, `getMetadata()` is null and this branch silently falls back to trailing-slash — i.e. effectively dead. Worth confirming the listing options request metadata where directory detection matters.

### Minor: per-instance memoization can go stale

`directoryResolved` is memoized for the object's lifetime, so a path resolved as non-directory won't reflect a blob created afterward. `AzPath` instances are short-lived, so low risk — just noting the semantics.

pditommaso:
All five points from my previous review are addressed — thanks. I re-checked at `9a308622b`:

- `isDirectory()` now reads the `attributes` field and stores the resolved attrs back, so a later `readAttributes()` reuses them instead of re-hitting the network ✅
- description and code agree on error handling (failures propagate) ✅
- the stale `resolve`/`resolveSibling` section is gone; both methods are identical to master ✅
- metadata retrieval is enabled on the hierarchical listing, so the `hdi_isfolder` branch is actually reachable ✅ — I decompiled `azure-storage-blob-12.30.0` to confirm the old one-arg `listBlobsByHierarchy(prefix)` is exactly `("/", new ListBlobsOptions().setPrefix(prefix), null)`, so `setRetrieveMetadata(true)` is the only behavioural delta
- the resolved result now lives in the one-shot cache rather than being memoized for the object's lifetime ✅

I ran `:plugins:nf-azure:test` against a real storage account: `AzPathTest` 70/70 green including the 7 new cases, and `should detect a virtual directory without a trailing slash` passes live in 0.48s — so the #6427 fix is verified end-to-end through both entry points. The one failure, `should list root directory` (`SpockTimeoutError` at `AzNioTest.groovy:662`), reproduces identically on master and is an artifact of my test account holding 124 containers while that test walks the whole tree under a 30s timeout. Not related to this PR.

Two remaining points, neither blocking:

### Dropping the `size == 0` guard on `hdi_isfolder`

`4d8a69b6b` removed the size check on the premise that ADLS Gen2 can report a non-zero content length for a directory marker. I couldn't corroborate that — HNS directories return `Content-Length: 0` over the Blob endpoint as far as I can tell. The guard was cheap defense in depth: a regular non-empty blob that happens to carry `hdi_isfolder=true` is now reported as a directory with `size = 0`, which stages as a silently empty input rather than failing loudly. Since blob metadata is user-writable, could you either link a reference for the non-zero-length marker case, or keep the guard?

### `isDirectory()` is no longer I/O-free

Groovy property access resolves `azPath.directory` to `isDirectory()`, which is why `AzPathTest` had to switch to `path.@directory`. The only property use left in the codebase is the write at `AzFileSystem.groovy:328`, so nothing regressed — but a short comment on the field (or on `isDirectory()`) would save the next reader from adding an accidental network round-trip.

Minor nits:

- The description says "repeated `isDirectory()` calls do not repeat the storage lookup". That holds only until a `readAttributes()` consumes the one-shot cache; the following `isDirectory()` goes back to the network. Worth softening the wording.
- The `isDirectory should propagate the error when the attribute lookup fails` test was dropped, so the propagation behaviour the description now promises has no coverage.
- In `should validate blob constructor and cached attributes`, pre-seeding `AzFileAttributes(directory: IS_DIRECTORY)` makes the `path.isDirectory() == IS_DIRECTORY` assertion tautological. Harmless given the dedicated new tests, just noting it.

pditommaso:
Build looks broken

pditommaso (commented):
Re-validated at `bffd97afe` against a real Azure storage account. The fix works: `should detect a virtual directory without a trailing slash` fails on master and passes here, and `AzPathTest` is 72/72 green. The one suite failure, `AzNioTest > should list root directory` (30s `SpockTimeoutError`), reproduces identically on master — pre-existing, unrelated to this PR.

The `AzFileAttributes` half of the diff is good and I'd take it as-is: the `hdi_isfolder == 'true'` comparison fixes a genuine Groovy truth-coercion bug (`"false"` previously coerced to `true`), the `size == 0` guard is back in `3b38ba956`, and `BlobListDetails.setRetrieveMetadata(true)` is what makes the marker branch reachable in listings at all.

But the `AzPath.isDirectory()` change breaks `equals()`, and I think there's a much smaller fix available.

## Blocking: making `isDirectory()` do I/O breaks `AzPath.equals()`

`isDirectory()` is the Groovy getter for the `directory` property, and `@EqualsAndHashCode(includes = 'fs,path,directory', includeFields = true)` generates an `equals()` that compares **`this.@directory` (raw field) against `other.isDirectory()` (getter)**. From `javap -c` on the class built from this branch:

```
119: aload_0
120: getfield      #169   // Field directory:Z       <- this:  raw field
123: invokestatic  #192   // Method java/lang/Boolean.valueOf:(Z)Ljava/lang/Boolean;
126: aload_2
127: invokevirtual #427   // Method isDirectory:()Z  <- other: getter, now hits the network
130: invokestatic  #192   // Method java/lang/Boolean.valueOf:(Z)Ljava/lang/Boolean;
133: invokestatic  #78    // Method ScriptBytecodeAdapter.compareEqual
```

`hashCode()` uses `getfield directory` on both sides, so only `equals()` changes. Three consequences — I confirmed each with probe tests that **pass on master and fail on this branch**:

1. **`equals()` now performs an Azure REST round-trip** (`getProperties`, and `listBlobs` via `guessPath` on 404) and can throw `BlobStorageException`/`NoSuchFileException` out of `equals()`.
2. **`equals()` is asymmetric** — `a.equals(b) != b.equals(a)` when one side carries listing-sourced attributes and the other doesn't.
3. **The equals/hashCode contract is violated** — `hashCode` says two paths are in the same bucket, `equals` says they differ, so a `HashMap` lookup with an equal key can miss.

Point 3 isn't theoretical; `AzPath` is routinely a hash key on the task path:

- `TaskProcessor.groovy:1827` `Map<Path,FileHolder> holders` → `TaskInputResolver.groovy:132` `holders.containsKey(value)`
- `PublishDir.groovy:70` `Map<Path,Boolean> makeCache` → `PublishDir.groovy:559` `makeCache.containsKey(dir)`
- `TaskRun.outputFiles` / `TaskProcessor.getPublishFiles` — `Set<Path>`

Reproducer (drop into `plugins/nf-azure/src/test/nextflow/cloud/azure/nio/`, no credentials needed — all three pass on master, all three fail here):

```groovy
package nextflow.cloud.azure.nio

import spock.lang.Specification

class AzPathEqualsProbeTest extends Specification {

    def 'equals should not trigger a remote attribute lookup'() {
        given:
        def fs = Mock(AzFileSystem); fs.getContainerName() >> 'pipeline'
        def a = new AzPath(fs, '/pipeline/output')
        def b = new AzPath(fs, '/pipeline/output')
        when:
        a.equals(b)
        then:
        0 * fs.readAttributes(_)
    }

    def 'equals should be symmetric'() {
        given:
        def fs = Mock(AzFileSystem); fs.getContainerName() >> 'pipeline'
        def a = new AzPath(fs, '/pipeline/output')
        a.setAttributes(new AzFileAttributes(directory: true))   // listing-sourced: is a directory
        def b = new AzPath(fs, '/pipeline/output')               // bare path
        fs.readAttributes(_) >> new AzFileAttributes(directory: false)
        expect:
        a.equals(b) == b.equals(a)
    }

    def 'equal paths must be findable in a HashMap'() {
        given:
        def fs = Mock(AzFileSystem); fs.getContainerName() >> 'pipeline'
        def key = new AzPath(fs, '/pipeline/output')
        def lookup = new AzPath(fs, '/pipeline/output')
        lookup.setAttributes(new AzFileAttributes(directory: true))
        fs.readAttributes(_) >> new AzFileAttributes(directory: true)
        when:
        def map = [:]; map.put(key, 'value')
        then:
        map.get(lookup) == 'value'
    }
}
```

## Suggestion: the provider already gets this right

I probed master directly. For a container holding only `output/file.txt`, referenced as `az://container/output`:

```
Files.isDirectory = true    AzPath.isDirectory = false
```

`AzFileSystemProvider.readAttributes` → `readBlobAttrs0` → `guessPath` **already** resolves slashless virtual directories correctly, on master, with no changes. The whole of #6427 is that `AzPath` declares a public `isDirectory()` which shadows the `FilesEx.isDirectory(Path)` extension method and returns the trailing-slash answer instead — and that's exactly what nf-schema's `Nextflow.file(value).isDirectory()` resolves to under Groovy dynamic dispatch.

So deleting (or renaming) `AzPath.isDirectory()` lets `azPath.isDirectory()` fall through to `FilesEx.isDirectory` → `Files.isDirectory` → the provider, which fixes #6427 with:

- no new network call in `isDirectory()` beyond what `Files.isDirectory()` already does,
- no change to `equals()` — `@EqualsAndHashCode` goes back to reading the field on both sides,
- no one-shot-cache interaction to reason about.

I checked the callers: nothing outside `AzPath` itself invokes it under `@CompileStatic` — internal code (`blobClient()`, the constructors, `AzFileSystem.groovy:332`) reads or writes the `directory` field directly, and `AzFileSystem.groovy:378` is `GuessPath.directory`, not `AzPath`. `AzNioTest`'s new case would need `Files.isDirectory(...)` on both assertions, which is the behaviour users actually get.

Worth trying that first — if it holds, the `AzPath` and `AzFileSystem.readAttributes` changes drop out of this PR entirely and you're left with just the `AzFileAttributes` + listing-metadata improvements, which I have no concerns about.

## Minor

- The description still says "repeated `isDirectory()` calls reuse them" — accurate only until a `readAttributes()` consumes the one-shot cache, which the description does now cover further down. Fine either way if the `AzPath` change survives.

pditommaso:
I've put the alternative from my review up as #7465, stacked on this branch so the diff is only the delta.

It drops `AzPath.isDirectory()` instead of teaching it to consult storage. `AzFileSystemProvider.readAttributes` → `guessPath` already resolves slashless virtual directories on master — the only reason `Nextflow.file('az://container/output').isDirectory()` returns `false` is that `AzPath` declares a public `isDirectory()` that shadows the `FilesEx.isDirectory(Path)` extension method. Removing it fixes #6427 and keeps `equals()`/`hashCode()` reading the raw field, so the regression I flagged goes away. Your `AzFileAttributes` and listing-metadata changes are untouched — they're independently correct and stay.

Verified against a real storage account: `should detect a virtual directory without a trailing slash` passes in 0.33s and fails on master; the three `equals()` regression tests fail on this branch and pass on #7465. The only suite failure is `should list root directory`, which times out identically on master.

@adamrtalbot happy for you to cherry-pick it into this branch and close #7465 if you'd rather keep it as one PR — whatever's less friction.

PR #6432, opened 2025-09-26
Fix cache command to use SysEnv for cloudcache path check

pditommaso:
## Summary
- Fix cache command to use `SysEnv.get('NXF_CLOUDCACHE_PATH')` instead of `session.cloudCachePath`
- Add debug logging to distinguish between CacheManager and LogsHandler usage paths
- Improves handling when command is used via Seqera Platform

## Changes
- Use `SysEnv.get('NXF_CLOUDCACHE_PATH')` for environment variable check
- Add debug logs to show which backup method is being used
- Add explanatory comment about why direct env var access is used

## Context  
The `session.cloudCachePath` requires initialization via CmdRun execution, but this command is primarily used by Seqera Platform which provides the cache path via environment variable. Using `SysEnv` directly avoids the initialization dependency.

### DIFF
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/CacheCommand.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/CacheCommand.groovy
@@ -19,6 +19,7 @@ package io.seqera.tower.plugin
 
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
+import nextflow.SysEnv
 import nextflow.cli.PluginAbstractExec
 /**
  * Implements nextflow cache and restore commands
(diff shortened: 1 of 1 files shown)


plugins/nf-tower/src/main/io/seqera/tower/plugin/CacheCommand.groovy
```diff
@@ -43,12 +44,17 @@ class CacheCommand implements PluginAbstractExec {
     }
 
     protected void cacheBackup() {
-        log.debug "Running Nextflow cache backup"
-        if( !getSession().cloudCachePath ) {
+        // note: use directly NXF_CLOUDCACHE_PATH instead of `session.cloudCachePath`
+        // because the latter required to be initialized via the execution
+        // CmdRun. However this command is only executed to be used  via Seqera Platform
+        // that's providing the cache path via the env variable
+        if( !SysEnv.get('NXF_CLOUDCACHE_PATH') ) {
```
jorgee, line 51:
```suggestion
        if( !getSession().cloudCachePath && !SysEnv.get('NXF_CLOUDCACHE_PATH') ) {
```
Why we do not check both? We would detect either if it has been set in nextflow.config or env var. It will also make Cursor happy

pditommaso: Fair

pditommaso: Done a88e2cdf

PR #6434, opened 2025-09-29
Add support for comma-separated file paths in ProcessEntryHandler

pditommaso:
## Summary
Enhances ProcessEntryHandler to support comma-separated file paths for the `file` input type, enabling users to pass multiple files in a single parameter.

## Changes
- Add `parseFileInput()` helper method to handle comma-separated values
- Support both String and GString inputs  
- Return `List<Path>` for multiple files, single `Path` for one file
- Handle whitespace trimming and empty string filtering
- Maintain backward compatibility with existing single file behavior
- Add comprehensive unit tests using Spock parameterized testing

## Usage
```bash
# Single file (existing behavior)
nextflow run script.nf --input file.txt

# Multiple files (new functionality)  
nextflow run script.nf --input "file1.txt,file2.txt,file3.txt"

# Handles whitespace gracefully
nextflow run script.nf --input " file1.txt , file2.txt "
```

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
@@ -341,6 +341,27 @@ class ProcessEntryHandler {
     }
     
     /**
+     * Parses file input handling comma-separated values.
+     * If the input contains commas, splits and returns a list of files.
+     * Otherwise returns a single file object.
+     * 
+     * @param fileInput String representation of file path(s)
+     * @return Single file object or List of file objects
+     */
+    protected Object parseFileInput(String fileInput) {
+        if (fileInput.contains(',')) {
+            // Split by comma, trim whitespace, and convert each to a file
+            return fileInput.tokenize(',')
+                .collect { it.trim() }
+                .findAll { !it.isEmpty() }
+                .collect { Nextflow.file(it) }
+        } else {
+            // Single file case - existing behavior
+            return Nextflow.file(fileInput)
+        }
+    }
+    
+    /**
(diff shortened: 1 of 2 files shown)


bentsherman:
Probably better to support space-separated files since it will usually be a glob pattern like `file*.txt`

pditommaso:
That can become tricky to handle file names with blanks

bentsherman:
The same problem exists for a comma-separated list -- you have to quote or escape things either way

pditommaso:
Nextflow CLI handles blank separated values as independent value. I mean if you have: a1.txt, a2.txt, and a3.txt  


and run: 

```
nextflow run --foo a* 
```

`foo` will be `a1.txt`  

instead, if you wrap in quotes: 

```
nextflow run --foo "a*"
```

`foo` will be `a*`

bentsherman:
I see, so it's an issue with the CLI parsing. Should be something we can address in CLI v2 then

pditommaso:
Yep

modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
```diff
@@ -360,8 +381,8 @@ class ProcessEntryHandler {
         switch( paramType ) {
             case 'path':
             case 'file':
-                if( paramValue instanceof String ) {
-                    return Nextflow.file(paramValue)
+                if( paramValue instanceof String || paramValue instanceof GString ) {
```
bentsherman, line 384: How would the param value be a GString? Isn't it coming from the command line?

pditommaso: yeah, likely useless, but does not hurt

PR #6440, opened 2025-10-02
Add rnaseq-nf tutorial

christopher-hakkaart:
Overview:
- Add a rnaseq-nf tutorial to act as a prerequisite for other guides and tutorials that reference it heavily.

TODO:
- Fix mermaid diagrams

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -171,6 +171,7 @@ developer/packages
 :caption: Tutorials
 :maxdepth: 1
 
+tutorials/rnaseq-nf
 tutorials/data-lineage
 tutorials/workflow-outputs
 tutorials/metrics
--- a/docs/tutorials/rnaseq-nf.md
+++ b/docs/tutorials/rnaseq-nf.md
@@ -1,4 +1,4 @@
-
+(rnaseq-nf-page)=
 
 # rnaseq-nf
 
--- a/docs/tutorials/rnaseq-nf.md
+++ b/docs/tutorials/rnaseq-nf.md
@@ -1,0 +1,239 @@
+
+
+# rnaseq-nf
+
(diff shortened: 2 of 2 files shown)


christopher-hakkaart:
@bentsherman - This is what I was thinking for a precursor for many of the guides/tutorials in development. I've tried to keep it reasonably high level, but enough to explain what runs, and where it can be found.

I was thinking about linking to the specific files in the repo, i.e., when referring to the entry workflow or module, link to those in GitHub.

Let me know what you think.

bentsherman:
Thanks Chris, looks good so far.

I just updated the rnaseq-nf source code to align better with the initial version of the workflow outputs tutorial, and I updated this PR accordingly.

However now I think I might have updated rnaseq-nf "too much" by joining the fastqc/quant results into a single channel. That ends up making the workflow outputs tutorial "too simple." So I might revert rnaseq-nf back to having separate fastqc/quant channels so that the tutorial can still show how to join them to prepare for workflow outputs. Need to think a bit more about this

christopher-hakkaart:
Nice! The changes look great. I see what you mean about the separate fastqc/quant channels. I think it would be beneficial to still display this, as it's likely a common use case and could be considered a learning outcome. 

I'm going to explore presenting code in the Migrating to workflow outputs as separate code blocks or tabs, labeled or tagged to indicate their origin. It might look terrible, be unclear, and/or disrupt the flow, so I may not proceed with it further. I'm just flagging that I may create a draft PR to show what it looks like.

bentsherman:
rnaseq-nf and tutorials should be up to date now with separate fastqc/quant channels

christopher-hakkaart:
It's all looking really good. Did you want to let it sit and think about it some more? Or are we close to merging?

bentsherman:
Yeah I'm happy to merge

PR #6443, opened 2025-10-02
Add staging events to the TraceObserverV2

### LINKED ISSUE #5905: Detect when files are being staged

rcannood:
This PR implements file staging event notifications for `TraceObserverV2`, allowing plugins to track when remote files are staged to the local work directory. This resolves #5905 by providing observers access to both the original remote file paths and their staged local paths.

## Background

Issue #5905 identified a need for provenance tracking plugins (specifically nf-lamin) to access the original remote paths of input files, not just their staged local paths. Two previous attempts were made to solve this:

- **#5907**: Added `onFileStage` to `TraceObserver` - rejected in favor of #5911
- **#5911**: Modified `FileHolder` to include remote paths - merged but then **reverted** due to concerns about cache invalidation and risk to core functionality

I met up with @bentsherman at the beginning of September. He said he wanted to have another shot at resolving this issue, and that I could reach out in a few weeks time if no progress was made.

## Solution

This implementation is a simpler solution in comparison to the two previous PRs. I added this to the TraceObserverV2:

- **`onFileStaged(FileStagingEvent event)`** - Called after a file has been successfully staged from a remote location

The `FileStagingEvent` provides:
- `source`: The original remote path (e.g., `s3://bucket/file.txt`, `https://example.com/data.csv`)
- `target`: The staged local path in the work directory (e.g., `work/.../file.txt`)

## Usage Example

For plugins that need to track file provenance (like nf-lamin):

```groovy
class ProvenanceObserver implements TraceObserverV2 {
    
    @Override
    void onFileStaged(FileStagingEvent event) {
        // event.source = original remote path
        // event.target = staged local path
        log.info "File staged: ${event.source} -> ${event.target}"
        registerFileProvenance(event.source, event.target)
    }
}
```

## Related Issues

- Resolves #5905
- Supersedes #5907 (rejected)
- Supersedes #5911 (reverted)

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -77,6 +77,7 @@ import nextflow.trace.TraceObserverV2
 import nextflow.trace.TraceRecord
 import nextflow.trace.WorkflowStatsObserver
 import nextflow.trace.event.FilePublishEvent
+import nextflow.trace.event.FileStagingEvent
 import nextflow.trace.event.TaskEvent
 import nextflow.trace.event.WorkflowOutputEvent
 import nextflow.util.Barrier
@@ -1114,6 +1115,10 @@ class Session implements ISession {
     void notifyFilePublish(FilePublishEvent event) {
         notifyEvent(observersV1, ob -> ob.onFilePublish(event.target, event.source))
         notifyEvent(observersV2, ob -> ob.onFilePublish(event))
+    }
+
+    void notifyFileStaged(FileStagingEvent event) {
+        notifyEvent(observersV2, ob -> ob.onFileStaged(event))
     }
 
     void notifyFlowComplete() {
--- a/modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy
@@ -39,6 +39,7 @@ import groovy.transform.ToString
 import groovy.util.logging.Slf4j
 import nextflow.Session
 import nextflow.exception.ProcessStageException
+import nextflow.trace.event.FileStagingEvent
 import nextflow.extension.FilesEx
 import nextflow.util.CacheHelper
 import nextflow.util.Duration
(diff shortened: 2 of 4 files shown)


bentsherman:
Trying my idea again #6460

rcannood:
@pditommaso:

> Likely those events should be emitter *async* in this context for two reasons:

Thanks for the review, and I totally agree with you.

However, this PR mimics the event handling system that is in place for other events (e.g. [`notifyFilePublish`](https://github.com/rcannood/nextflow/blob/b6e17d2a68496049acde74744482b1f13da135fe/modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy#L585)). If we would like to make the notification system asynchronous, this is something [`notifyEvent`](https://github.com/rcannood/nextflow/blob/b6e17d2a68496049acde74744482b1f13da135fe/modules/nextflow/src/main/groovy/nextflow/Session.groovy#L1152-L1162) should probably take care of, right?

Edit: I created #6471

----

@bentsherman:
> I still prefer to enable this through the task events rather than adding a new event. I just haven't had time to prepare a new PR. I will try to have something ready this week or next week. I would at least like to try that before we add a whole new event.
> 
> Trying my idea again https://github.com/nextflow-io/nextflow/pull/6460

Alright, thank you for taking the time to take a look at this! Despite this new PR, would you agree that it might still make sense to merge this PR as well, since `onFileStaged` nicely mirrors `onFilePublish`?

bentsherman:
I'm still considering the merits of having an event for file staging. I don't really see it as a mirror of file publishing. File staging is just an implementation detail, whereas file publishing is "essential" to understand the outputs of a run and their provenance.

rcannood:
Thanks for taking this PR into consideration! 

From my perspective, the file staging is also essential to understand the inputs of a run and their provenance :sweat_smile: This PR is the only thing that is currently blocking me from implementing decent data provenance for Nextflow workflows with LaminDB using nf-lamin. For now, all of the Nextflow runs result in Lamin Runs being created with only outputs and no inputs, which goes a bit against the whole LaminDB philosophy.

Based on the proposed implementation, would you expect any significant slowdowns or breaking changes with existing functionality?

bentsherman:
> From my perspective, the file staging is also essential to understand the inputs of a run and their provenance

Of course, but file staging is tied to the task lifecycle in a way that file publishing is not. Output files can be published separately from the originating task. But remote input files are only staged in the context of a task getting ready to run, so there isn't as much of a need for a separate "file staged" event.

> Based on the proposed implementation, would you expect any significant slowdowns or breaking changes with existing functionality?

I suppose there is a small cost to adding new workflow events, but it should be pretty minimal. I'm more concerned about adding an event for something that is more of an implementation detail

For example, you wouldn't really be able to use the "file staged" event on its own. You would always need to use it in concert with the task events, because not all input files need to be staged. That suggests that it shouldn't be a standalone event at all

bentsherman:
Closing in favor of #6460

PR #6456, opened 2025-10-07
add a meta annotation in process

### LINKED ISSUE #6453: Feature request : add a description field in `process`
## New feature

Hi it would be cool to have a place to put a short/long description of each `process` , so we can have a human description of each step in the directed **graph**. Something like the *.yml files in the module files the nf-core project

## Use case 

```nextflow
process SAMTOOLS_FAIDX {
shortDescription = "index a reference genome"
(...)
}
description = "index a reference genome, creates a .fai file that can be used to quickly retrieve sequences"
(...)
}
```

## Suggested implementation 

see above

jagedn:
Add a new "meta" annotation in process where users can add metadatas to them

a description metadata is implemented and used in DAG 

closes #6453 

imagen
imagen
imagen

### DIFF
--- a/plugins/nf-google/build.gradle
+++ b/plugins/nf-google/build.gradle
@@ -58,6 +58,9 @@ dependencies {
     api 'com.google.cloud:google-cloud-nio:0.128.5'
     api 'com.google.cloud:google-cloud-storage:2.58.0'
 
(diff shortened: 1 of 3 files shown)


bentsherman:
We actually discussed this recently, but we plan to take a different approach: https://github.com/nextflow-io/nextflow/issues/6453#issuecomment-3377125017

jagedn:
great, 

thanks for the info @bentsherman , I will close the PR

PR #6460, opened 2025-10-08
Track remote input file in FileHolder

### LINKED ISSUE #5905: Detect when files are being staged

bentsherman:
Close #5905

This PR preserves the source path of remote input files in the FileHolder.

Here are the key changes:

TaskProcessor:
```diff
             if( item instanceof Path || coerceToPath ) {
                 final path = resolvePath(item)
                 final target = executor.isForeignFile(path) ? foreignFiles.addToForeign(path) : path
-                final holder = new FileHolder(target)
+                final holder = new FileHolder(path, target)
                 files << holder
             }
```

FileHolder:
```diff
     FileHolder( Path path ) {
         this.sourceObj = path
         this.storePath = real(path)
         this.stageName = norm(path.getFileName())
     }
 
     FileHolder( def origin, Path path ) {
         this.sourceObj = origin
-        this.storePath = path
+        this.storePath = real(path)
         this.stageName = norm(path.getFileName())
     }
```

The two-arg constructor is updated to match the behavior of the one-arg constructor using the `real()` method:

```groovy
    static private Path real( Path path ) {
        try {
            // main reason for this is to resolve symlinks to real file location
            // hence apply only for default file system
            // note: also for Google Cloud storage path it may convert to relative path
            // it may return invalid (relative) paths therefore do not apply it
            return path.getFileSystem() == FileSystems.default ? path.toRealPath() : path
        }
        catch( Exception e ) {
            log.trace "Unable to get real path for: $path"
            return path
        }
    }
```

The `real()` method only resolves symlinks for local files, so path resolution should behave the same way as before for both local and remote files.

A future improvement would be to forward this information to a less internal data structure in `TaskEvent`. But this change is necessary either way to preserve the remote input path in the `TaskRun`.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -1946,7 +1946,7 @@ class TaskProcessor {
             if( item instanceof Path || coerceToPath ) {
                 final path = resolvePath(item)
                 final target = executor.isForeignFile(path) ? foreignFiles.addToForeign(path) : path
-                final holder = new FileHolder(target)
+                final holder = new FileHolder(path, target)
                 files << holder
             }
             else {
--- a/modules/nf-commons/src/main/nextflow/file/FileHolder.groovy
+++ b/modules/nf-commons/src/main/nextflow/file/FileHolder.groovy
@@ -20,11 +20,15 @@ package nextflow.file
 import java.nio.file.FileSystems
 import java.nio.file.Path
 
+import com.google.common.hash.Hasher
 import groovy.transform.CompileStatic
 import groovy.transform.EqualsAndHashCode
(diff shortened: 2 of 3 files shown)


bentsherman:
This is the only other place where the two-arg constructor is used:
```groovy
        /*
         * default case, convert the input object to a string and save
         * to a local file
         */
        def source = input?.toString() ?: ''
        def result = Nextflow.tempFile(altName)
        result.text = source
        return new FileHolder(source, result)
```

This is for when a `file` input converts a string value into a temp file with the string content. It should not be affected by the change to the two-arg constructor

bentsherman:
Folding into #6490

PR #6471, opened 2025-10-11
Make notifyEvent asynchronous

rcannood:
Makes the notifyEvent functionality asynchronous, as requested by @pditommaso in #6443:

> This seems going into the right direction. My only connect is about the notification events, here
> 
> https://github.com/nextflow-io/nextflow/pull/6443/files#diff-cb4252ec3616353913180e8e6eb95c249b4d591daf4143283e543be188568685R105-R111
> 
> Likely those events should be emitter *async* in this context for two reasons:
> 
> 1. Make sure the target path is accessible when the event is received
> 
> 2. Avoid that core runtime performance could be impacted by compute/data intensive that could be implemented by third-parties plugins

---

Update: I had to change the return type of `notifyEvent` so `notifyFlowComplete` and `notifyError` can wait until all futures are resolved before continuing.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -1118,8 +1118,11 @@ class Session implements ISession {
     }
 
     void notifyFlowComplete() {
-        notifyEvent(observersV1, ob -> ob.onFlowComplete())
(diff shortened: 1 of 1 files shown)


bentsherman:
Closing without merging since we didn't add the onFileStage event

PR #6482, opened 2025-10-13
Type checking support

bentsherman:
This PR adds a few things to nf-lang to help enable type checking in the language server.

The full type checker will eventually be incorporated into nf-lang, but for now it will reside in the language server while we continue to develop it. This way I can release improvements to the type checker without requiring a Nextflow patch release

No unit tests because these changes are covered by language server unit tests for now.

Summary of changes:

- Save method overloads in `VariableScopeVisitor` when a method call matches multiple possible methods (used by type checker to select method or provide appropriate error message)

- Add `@NamedParams` annotation to standard library functions (used by type checker to validate named params)

- Fix resolution of output types for processes and workflows

- Resolve `params` block during name checking (was previously done only by language server)

- Minor bug fixes

### DIFF
--- a/modules/nf-lang/src/main/java/nextflow/script/control/ResolveVisitor.java
+++ b/modules/nf-lang/src/main/java/nextflow/script/control/ResolveVisitor.java
@@ -16,6 +16,7 @@
 package nextflow.script.control;
 
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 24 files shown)


modules/nf-lang/src/main/java/nextflow/script/control/ResolveVisitor.java
```diff
@@ -150,7 +150,7 @@ private boolean resolveGenericsType(GenericsType genericsType) {
         if( genericsType.isResolved() )
             return true;
         var type = genericsType.getType();
-        resolveOrFail(type, genericsType);
+        resolve(type);
```
jorgee: Why is the failure removed? Also ignoring the return. Not understanding this part.

bentsherman: It was causing an unhelpful error when a generic type argument couldn't be resolved. But now I remember I need to propagate this error up the stack and still report it, so I will add that

bentsherman: This code was inherited from the Groovy compiler and Groovy's handling of generics seems to be spotty. This code is now more robust for checking both types and generic type arguments.

Issue #6481, opened 2025-10-13
Nextflow doesn't capture Fusion exit codes from Google Batch API


ejseqera:
## Bug report 
When running Google Batch jobs with Fusion enabled, tasks that fail with fusion exit code 174/175 are not properly reported by Nextflow. Instead of showing the actual exit code (175), Nextflow displays the status as `-` or unknown.

The `GoogleBatchTaskHandler` [only reads the exit code from the `.exitcode` file](https://github.com/nextflow-io/nextflow/blob/master/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy#L548) and does not use the exit code provided by the Google Batch API response, unlike the AWS Batch and Azure Batch implementations. 
```
task.exitStatus = readExitFile()
```
When Fusion fails with exit code 175, the `.exitcode` file may not be written, causing `readExitFile()` to return `Integer.MAX_VALUE`, which displays as `-` in the NF log. The [Google Batch API does provide the exit code via](https://github.com/nextflow-io/nextflow/blob/master/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy#L574) `lastEvent?.taskExecution?.exitCode` but it's currently only logged and not used to set the task exit status.

#### Expected behaviour:
When a task fails with exit code 175 (or 174, or any other exit code returned by the batch API), Nextflow should:

1. Capture and display the actual exit code from the Google Batch API
2. Report the task as failed with the correct exit status
3. Behave consistently with other AWS Batch and Azure Batch executors

##### Actual behaviour:

1. Task exits with Fusion code 175 in Google Batch
2. Nextflow reports the exit status as `-` (which represents `Integer.MAX_VALUE` internally)
3. The actual exit code is logged but not used 

### Steps to reproduce the problem

1. Run a workflow with `exit 175` in the process script on Google Batch
2. Batch job will return exit code and fail
3. Nextflow will fail to return this exit code

### Program output 

```
[...]

Issue #6445, opened 2025-10-03
Optimize exit code handling by relying on scheduler status for successful executions


pditommaso:
## Background

Currently, Nextflow reads the `.exitcode` file from the work directory to determine task completion status. PR #6442 improved K8s error handling by prioritizing the scheduler's exit status for failed executions (e.g., OOMKilled, pod eviction), but still falls back to reading the `.exitcode` file for successful executions.

## Proposed Optimization

For successful task executions (scheduler exit status == 0), we should rely solely on the scheduler's reported exit status and bypass reading the `.exitcode` file entirely.

## Benefits

- **Reduced I/O pressure**: Eliminates one file read operation per successful task
- **Better scalability**: Particularly beneficial for workloads with many fine-grain jobs
- **Lower storage costs**: Reduces remote file storage access (S3, Azure Blob, GCS, etc.)
- **Improved performance**: Faster task completion acknowledgment

## Implementation Considerations

This optimization should be evaluated across all executor types:
- K8s (nf-k8s)
- AWS Batch (nf-amazon)
- Azure Batch (nf-azure)
- Google Batch (nf-google)
- Other cloud executors

## Related Work

- PR #6442: Improved K8s exit code handling for error cases
- Issue #6436: Original OOM handling issue

The current PR establishes the pattern of prioritizing scheduler exit status for errors. This issue proposes extending that approach to successful executions as well.

bentsherman:
@jorgee please feel free to tackle this when you have some time

BioWilko:
Just to check, would this approach fall back to `.exitcode` in the case that the scheduler resource no longer exists? e.g. if the k8s control plane kills a job prior to nextflow checking with the scheduler?

jorgee:
If I am not wrong, this is caused by the groovy truth for integers, where 0 is considered false, isn't it?

muffato:
Hello. The messages imply this is only for cloud schedulers, but I'd like to add that it seems valid for HPC schedulers too.
On HPC under certain circumstances, the exit code from the task is a misleading 0, or is missing (cf #2847). In those cases, the job scheduler can be queried and tell whether the task exited successfully or not.

Matthieu

PR #6484, opened 2025-10-14
Optimize exit code handling by relying on scheduler status for successful executions

jorgee:
## Summary

This PR unifies exit code handling behavior across all cloud provider executors (AWS Batch, Azure Batch, Google Batch, and Kubernetes). Previously, different executors had inconsistent approaches to obtaining task exit codes, which led to issues like missing Fusion exit codes (#6481) and unnecessary I/O overhead (#6445).

### Unified Exit Code Strategy

All cloud providers now follow a consistent two-step approach:

1. **Primary**: Get exit code from the scheduler/cloud API
2. **Fallback**: Read `.exitcode` file only when API returns `null` (not when it returns `0`)

### Fixes by Provider

| Provider | Before | After |
|----------|--------|-------|
| **Google Batch** | Read exit code only from `.exitcode` file | Get from API (`taskExecution.exitCode`), fallback to file if null |
| **AWS Batch** | API with fallback to file when exit code was 0 or null | API with fallback only when null |
| **Azure Batch** | API with fallback to file when exit code was 0 or null | API with fallback only when null |
| **Kubernetes** | API with fallback to file when exit code was 0 or null | API with fallback only when null |

### Benefits

- **Correct exit codes**: Fusion exit codes (174/175) and other scheduler-reported failures are now properly captured from Google Batch API
- **Reduced I/O**: Eliminates unnecessary `.exitcode` file reads for successful tasks (exit code 0)
- **Better scalability**: Particularly beneficial for workloads with many fine-grained jobs
- **Lower storage costs**: Reduces remote file storage access (S3, Azure Blob, GCS)
- **Consistent behavior**: All cloud executors now behave the same way

## Closes

- #6481 - Nextflow doesn't capture Fusion exit codes from Google Batch API
- #6445 - Optimize exit code handling by relying on scheduler status for successful executions

## Test Coverage

Added comprehensive unit tests for all affected handlers verifying:
- Exit code from API is used when available
- Fallback to `.exitcode` file when API returns null
- No fallback when API returns 0 (success)
- Error exit codes (non-zero) are correctly captured

### DIFF
--- a/plugins/nf-google/src/test/nextflow/cloud/google/batch/GoogleBatchTaskHandlerTest.groovy
+++ b/plugins/nf-google/src/test/nextflow/cloud/google/batch/GoogleBatchTaskHandlerTest.groovy
@@ -732,13 +732,43 @@ class GoogleBatchTaskHandlerTest extends Specification {
         result == RESULT
 
         where:
-        TASK_STATE              | DESC                                                          | EXIT_CODE | ARRAY_CHILD    | TASK_STATUS                               | EXIT_STATUS       | RESULT    | TASK_ERROR
-        TaskStatus.State.FAILED | "Error"                                                       | null      | true           | nextflow.processor.TaskStatus.COMPLETED   | 0                 | true      | null
-        null                    | "Error"                                                       | null      | true           | nextflow.processor.TaskStatus.COMPLETED   | 1                 | true      | null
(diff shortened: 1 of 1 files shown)


jorgee:
@bentsherman @pditommaso  the google batch task handler is directly getting the exit code form the file (#6481). This PR also removes the fallback to .exitcode when the exit code is 0. It is not a big PR but I was wondering if you prefer to split it in two PRs to facilitate backport to stable versions. The fix for  #6481 is just the first commit ea1aa48875fb714ecc9de704c7261bf2ab131fc9

pditommaso (changes requested):
This looks good, but i'd keep post 25.10

bentsherman:
@jorgee we talked and agreed to include the google and k8s fixes in 25.10 and merge the rest of this PR in the next edge release

PR #6487, opened 2025-10-14
Bring workflow outputs out of preview

bentsherman:
This PR makes workflow outputs final in the Nextflow language.

### DIFF
--- a/docs/migrations/25-10.md
+++ b/docs/migrations/25-10.md
@@ -37,9 +37,9 @@ See {ref}`workflow-params-def` for details.
 
 <h3>Workflow outputs (out of preview)</h3>
 
-{ref}`Workflow outputs <workflow-output-def>`, introduced in Nextflow 24.04 as a preview feature, have been brought out of preview, which means that they can be used without the `nextflow.preview.output` feature flag.
+{ref}`Workflow outputs <workflow-output-def>`, introduced in Nextflow 24.04 as a preview feature, have been brought out of preview and can now be used without the `nextflow.preview.output` feature flag.
 
-The syntax has not changed since the third preview, with one exception. Each output declaration in the `output` block can optionally specify a type annotation:
+The final version adds the ability to specify a type annotation for each output declaration:
 
 ```nextflow
 workflow {
@@ -59,6 +59,8 @@ output {
     }
 }
 ```
+
+See {ref}`migrating-workflow-outputs` to get started.
 
 <h3>Type annotations</h3>
 
--- a/docs/workflow.md
+++ b/docs/workflow.md
@@ -474,19 +474,8 @@ Workflows can also be invoked recursively:
 
 ## Workflow outputs
 
-:::{versionadded} 24.04.0
-:::
-
-:::{versionchanged} 24.10.0
-A second preview version was introduced. See the {ref}`migration notes <workflow-outputs-second-preview>` for details.
-:::
-
-:::{versionchanged} 25.04.0
-A third preview version was introduced. See the {ref}`migration notes <workflow-outputs-third-preview>` for details.
-:::
-
-:::{versionchanged} 25.10.0
-Workflow outputs are out of preview. See the {ref}`migration notes <workflow-outputs-final>` for details.
+:::{versionadded} 25.10.0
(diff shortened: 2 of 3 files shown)


docs/workflow.md
```diff
@@ -485,8 +485,8 @@ A second preview version was introduced. See the {ref}`migration notes <workflow
 A third preview version was introduced. See the {ref}`migration notes <workflow-outputs-third-preview>` for details.
 :::
 
-:::{note}
-This feature requires the `nextflow.preview.output` feature flag to be enabled.
+:::{versionchanged} 25.10.0
+Workflow outputs are out of preview. See the {ref}`migration notes <workflow-outputs-final>` for details.
```
bentsherman: @christopher-hakkaart now that workflow outputs are output of preview, I wonder if we can remove some of these version callouts. Maybe just keep this one and the first one, to show when it was first introduced vs finalized?

docs/migrations/25-10.md
```diff
@@ -33,6 +33,33 @@ This syntax allows you to declare all parameters in one place with explicit type
 
 See {ref}`workflow-params-def` for details.
 
+(workflow-outputs-final)=
+
+<h3>Workflow outputs (out of preview)</h3>
+
+{ref}`Workflow outputs <workflow-output-def>`, introduced in Nextflow 24.04 as a preview feature, have been brought out of preview, which means that they can be used without the `nextflow.preview.output` feature flag.
```
christopher-hakkaart:
```suggestion
{ref}`Workflow outputs <workflow-output-def>`, introduced in Nextflow 24.04 as a preview feature, have been brought out of preview and can now be used without the `nextflow.preview.output` feature flag.
```

docs/migrations/25-10.md
```diff
@@ -33,6 +33,33 @@ This syntax allows you to declare all parameters in one place with explicit type
...
+
+The syntax has not changed since the third preview, with one exception. Each output declaration in the `output` block can optionally specify a type annotation:
```
christopher-hakkaart:
```suggestion
Only one syntax change has been made since the third preview. Output declarations in the `output` block can now optionally include type annotations.:
```
I think it would be better to be upfront here and say there was one change. Show it, and then say everything else remains the same.

docs/migrations/25-10.md
```diff
@@ -33,6 +33,33 @@ This syntax allows you to declare all parameters in one place with explicit type
...
+output {
+    samples: Channel<Map> {
+        path '.'
+        index {
+            path 'samples.csv'
+        }
+    }
+}
+```
+
```
christopher-hakkaart, line 62:
```suggestion

All other syntax remains unchanged since the third preview.

```

docs/reference/feature-flags.md
```diff
@@ -46,8 +46,10 @@ Feature flags are used to introduce experimental or other opt-in features. They
 `nextflow.preview.output`
 : :::{versionadded} 24.04.0
   :::
-: *Experimental: may change in a future release.*
-: When `true`, enables the use of the {ref}`workflow output definition <workflow-output-def>`.
+: :::{deprecated} 25.10.0
+  This feature flag is no longer required to use workflow outputs.
```
christopher-hakkaart, line 50:
```suggestion
Workflow outputs no longer require this feature flag.
```

docs/workflow.md
christopher-hakkaart: Nice idea. It would be best to reduce these to help reduce the visual clutter.

Another solution might be to add them all to a single admonition that lists the previews.

```suggestion
:::{versionchanged} 25.10.0
Workflow outputs are out of preview. See the {ref}`migration notes <workflow-outputs-final>` for details.

Version history:

- New in version 24.04.0
- Changed in version 24.10.0: A second preview version was introduced. See the {ref}`migration notes <workflow-outputs-second-preview>` for details.
- Changed in version 25.04.0: A third preview version was introduced. See the {ref}`migration notes <workflow-outputs-third-preview>` for details.
```

I'm fine with either deleting or finding a solution like this to link out. Another alternative here would be to make a single admonition with the points rather than including them all in one.

christopher-hakkaart (approved):
Added some suggestions. Non-blocking.

The versioning admonitions clutter the page, and now it's coming out of preview they are not as important. I've made a suggestion to list the changes in a single admonition, but this could also be split into two.

PR #6490, opened 2025-10-16
Track remote input files in lineage

jorgee:
Replaces the staged path stored in the lineage TaskRun entry for task input files by the source path.


jorgee:
Just in case #6460 is merged, this would be good to include it to the PR

bentsherman:
Just tested with a small pipeline that stages some remote files and it correctly shows the original URL in the TaskRun record. Let's go

PR #6499, opened 2025-10-22
Fix no secrets in AWS Batch jobs

jorgee:
This PR include the secrets to the ContainerProperties builder used for AWS Batch jobs. They were missed during the AWS SDK v2 refactor.


just-simon-5:
Hello!

Can I ask a question on this?

I currently have an AWS job definition defined in terraform containing secrets.

I want to replace this job definition with one created by Nextflow via configuration settings. 

I have successfully managed to find most of what I need in the [aws configuration documentation](https://www.nextflow.io/docs/latest/reference/config.html#aws) however I cannot seem to workout how to tell Nextflow which secret to use.

I imagine having to provide a "name" and "valueFrom" (the latter being an ARN).

I am launching this from Seqera Platform and have configured the launchpad with the correct workspace secrets.

Any guidance would be super helpful!

Thanks,
Simon

PR #6501, opened 2025-10-22
docs: Update migration notes

christopher-hakkaart:
- Update migration notes to bring 25.10 out of preview


bentsherman:
Whoops, looks like we had the same idea around the same time :smile: 

I had to go through it anyway to check for any missing updates

Closing in favor of #6502

PR #6504, opened 2025-10-23
docs: Update feature flags page

christopher-hakkaart:
- Standardize reference language
- Expand admonitions about preview features out of previews


christopher-hakkaart:
There are a lot of warnings/deprecation admonitions. My motivation here was to be very clear that things will start to fail with the newer Nextflow versions, but looking at the preview, the descriptions may be too long and repetitive.

PR #6506, opened 2025-10-23
docs: Update task directives

christopher-hakkaart:
- Consistency work for the process reference task directives section.

### DIFF
--- a/modules/nextflow/build.gradle
+++ b/modules/nextflow/build.gradle
@@ -38,7 +38,7 @@ dependencies {
     api "ch.qos.logback:logback-core:1.5.20"
     api "org.codehaus.gpars:gpars:1.2.1"
     api("ch.artecat.grengine:grengine:3.0.2") { exclude group: 'org.codehaus.groovy' }
-    api "commons-lang:commons-lang:2.6"
+    api "org.apache.commons:commons-lang3:3.18.0"
     api "commons-codec:commons-codec:1.15"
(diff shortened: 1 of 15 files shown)


docs/reference/process.md
```diff
@@ -12,8 +12,7 @@ The following task properties are defined in the process body:
 : The current task attempt.
 
 `task.exitStatus`
-: The exit code of the task script. Only applicable for processes with a `script:` or `shell:` block.
-: Since the exit code is only available after the task has been executed, it can only be used by certain process directives such as [errorStrategy](#errorstrategy).
+: The exit code returned by the task script. Only applicable for processes with a `script:` or `shell:` block. Since the exit code is only available after the task has been executed, it can only be used by certain process directives (e.g., [errorStrategy](#errorstrategy)).
```
bentsherman: In cases like this where there is a long description, I like to break it up so that the first paragraph is short and just says what the thing is, with additional paragraphs as needed to give extra details

That way it's easy to get the gist of what the thing is and skip the fine print if you want to

Although I can see how in this case the first paragraph is actually two sentences... sometimes that happens too

christopher-hakkaart: For 2-3 sentences, I like to condense them into one paragraph, but also happy to keep it as two.

Here, it's fine as there is a difference in what is being defined/explained.

- Paragraph one describes the option
- Paragraph two explains it.

Happy to split these to two again.

bentsherman (approved):

PR #6507, opened 2025-10-24
docs: Add action verb to feature flag

christopher-hakkaart:
- Very small PR to standardize action verbs for feature flags as they "trigger" a behavior


bentsherman:
Folding into #6506

PR #6508, opened 2025-10-24
feat: Add per-task Fusion control via disk directive

edmundmiller:
## Summary

Per-task Fusion control via `disk fusion: true/false` - a clearer alternative to `scratch false`.

## Behavior Matrix

| scratch | disk.fusion | fusion.enabled | Result | Warning |
|---------|-------------|----------------|--------|---------|
| `true` | `true` | any | scratch | ⚠️ YES |
| `false` | `true` | any | fusion | NO |
| `true` | `false` | any | scratch | NO |
| `true` | null | `true` | scratch | ⚠️ YES |
| null | `true` | any | fusion | NO |
| null | `false` | any | local | NO |

**Precedence:** `scratch=true` > `disk.fusion` > `fusion.enabled`

## Syntax

```groovy
process myTask {
    disk fusion: false  // Disable Fusion for this task
    disk request: '100 GB', fusion: true  // With disk size
    disk request: '375 GB', type: 'local-ssd', fusion: false  // With GCP disk type
}
```

## Notes

- ✅ Backward compatible - `scratch false` still works
- ⚠️ AWS Batch: Only `disk.fusion` works (`disk.type`/`request` ignored - existing behavior)

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/executor/res/DiskResourceTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/executor/res/DiskResourceTest.groovy
@@ -35,17 +35,23 @@ class DiskResourceTest extends Specification {
         then:
         disk.request == REQ
         disk.type == TYPE
+        disk.fusion == FUSION
 
         where:
-        VALUE                                  | REQ      | TYPE
-        _100_GB                                | _100_GB  | null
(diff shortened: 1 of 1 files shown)


bentsherman:
Closing since we didn't reach consensus on this

PR #6523, opened 2025-10-30
Change log level from warn1 to debug1 for K8s Job Pod Warnings

### LINKED ISSUE #6331: Mute k8s job pod scheduling warnings by default
## New feature

Currently when many processes are queued via k8s jobs Nextflow will print warnings for every job unable to schedule a pod e.g.
```
WARN: K8s Job nf-1858f850f1e7e854a6f4695ace684a7c-df144 does not have pod - Not yet scheduled?
WARN: K8s Job nf-24a1c789dc837e54b34c16a08660365d-67a1c does not have pod - Not yet scheduled?
WARN: K8s Job nf-b3ed3f9916b43af5a8196597e9044bdb-48e63 does not have pod - Not yet scheduled?
WARN: K8s Job nf-ec282416a6cd7d600131a0d2685c7951-175d3 does not have pod - Not yet scheduled?
```
There are cases where this is a useful message but the majority of the time it just means that you don't have a large enough cluster to schedule all pods simultaneously which is fine and normal and shouldn't lead to a warning.

Could this message be hidden by default? it makes watching Nextflow in k8s via the command line quite unpleasant since stderr is totally full of these messages and you only occasionally get to see the status of the pipeline.

BioWilko:
Closes #6331

Currently the K8S executor spams stdout when jobs don't have a pod yet, this is a perfectly normal state of affairs in most cases and just means you don't have enough CPUs / memory available currently, suggest muting this by default!

One issue I could see this causing would be that it may be more difficult for users to detect when there's a real reason the pod hasn't been created e.g. an invalid container string, may be worth adding specific handling for this.

### DIFF
--- a/plugins/nf-k8s/src/main/nextflow/k8s/client/K8sClient.groovy
+++ b/plugins/nf-k8s/src/main/nextflow/k8s/client/K8sClient.groovy
@@ -441,7 +441,7 @@ class K8sClient {
             throw new ProcessFailedException("K8s Job $jobName execution failed: $message")
         }
 
-        log.warn1("K8s Job $jobName does not have pod - Not yet scheduled?")
+        log.debug1("K8s Job $jobName does not have pod - Not yet scheduled?")
         return Collections.emptyMap()
     }
 


pditommaso (approved):
Make sense. Thanks for contributing

PR #6525, opened 2025-10-30
docs: Running pipelines CLI 

christopher-hakkaart:
- Proposal structure for basic usage descriptions

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -302,7 +302,7 @@ See {ref}`cli-config` for more information.
 
 ### Process inspection
 
-:::{versionadded} 23.09.0-edge
+:::{versionadded} 23.10.0
 :::
 
 The `inspect` command analyzes process settings in a pipeline without executing it. It outputs container information in JSON or Nextflow configuration format.
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -26,7 +26,7 @@ Available options:
 : Comma-separated list of configuration files which are added to the configuration set.
 
 `-d, -dockerize`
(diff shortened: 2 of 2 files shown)


bentsherman:
Thanks Chris, the CLI page is looking much better. It gives a much more comprehensive overview of the available commands. Now I think we just need to whittle down the most important use cases

I think part of the reason why the page still felt reference-heavy is because of the exhaustive description of every top-level CLI option. I always found this excessive

So I have removed it and moved the important bits to other places. Added a few details to the reference docs, moved the version section after the help section, and moved the config sections to the new config section you added

I think the hard/soft config override behavior is probably worth keeping since it's an important distinction, I just copied/pasted but you might want to adjust it to make it fit better in your new section

christopher-hakkaart:
Thanks @bentsherman - this looks great.

Dropping the options makes a big difference. I've reshuffled the configuration section and widened the description to include options and commands.

Overall, I'm really happy with the structure and framing. I would leave it as is for now and adjust it based on community feedback. I'm also happy to follow your lead if you see anything that you'd like to cut or add if you think it's missing.

bentsherman:
Okay I'll go ahead and merge

Now that the main CLI page is more comprehensive, I wonder if we could move some of the examples from the reference page to the main page

That might make the reference page less cluttered and the examples more visible. Though I still see some value in at least having some basic examples on the reference page

bentsherman (approved):

christopher-hakkaart:
Thanks Ben - I've made a task for myself to follow up the examples and look for opportunities to thin/move them around

PR #6526, opened 2025-10-30
docs: Add launch and auth to CLI page

christopher-hakkaart:
- Add launch and auth to CLI page


christopher-hakkaart:
Folding into #6525

PR #6539, opened 2025-11-04
Remove unnecessary runtime checks with strict syntax

### LINKED ISSUE #6533: "Process is already defined" error with nested subworkflows

When a subworkflow is used more than once in an workflow, and it is sometimes included from a different subworkflow, it raises the error
```
A process named '...' is already defined or included in script
```

decompress [this MWE](https://github.com/user-attachments/files/23305873/mwe.zip)

Run 
```
NXF_SYNTAX_PARSER=v2 nextflow run .
```

### Program output 

```

Launching `./main.nf` [boring_rutherford] DSL2 - revision: bb66e6056d

A process named 'PROCESS' is already defined or included in script: ./mwe/subworkflow.nf
```

### Environment 

* Nextflow version: 25.10.0 with NXF_SYNTAX_PARSER=v2
* Java version: 21.0.2
* Operating system: macOS
* Bash version: GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)

### Additional context

Note that if i move all the subworkflows to be defined in main.nf, instead of in separate files, the workflow runs successfully. Also, the error only occurs with `NXF_SYNTAX_PARSER=v2`

bentsherman:
Close #6533 

I haven't figured out exactly what causes this error, but I believe it is related to how the v2 parser loads scripts differently than the v1 parser

Since the v2 parser checks for conflicting declarations at compile-time, these runtime checks aren't needed anyway. Disabling these checks for the v2 parser allows the MWE to run successfully. There is already a unit test for this in `ScriptResolveTest`.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1,333 +1,176 @@
 (cli-page)=
 
-# Command line
+# Command line interface
 
-Nextflow provides a robust command line interface (CLI) for the management and execution of pipelines.
+Nextflow provides a robust command line interface (CLI) for the management and execution of pipelines. This page explains the key concepts and common usage patterns for the CLI.
 
-Simply run `nextflow` with no options or `nextflow -h` to see the list of available top-level options and commands. See {ref}`cli-reference` for the full list of subcommands with examples.
+For the complete reference of all commands, subcommands, and options, see {ref}`cli-reference`.
 
 :::{note}
-Nextflow options use a single dash prefix, e.g. `-resume`. Do not confuse with double dash notation, e.g. `--resume`, which is instead used for {ref}`Pipeline parameters <cli-params>`.
+Nextflow uses two types of command line flags:
+- Nextflow options use a single dash (e.g., `-log`) and modify Nextflow's behavior.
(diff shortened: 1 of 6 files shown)


modules/nextflow/src/main/groovy/nextflow/script/parser/v2/ScriptLoaderV2.groovy
```diff
@@ -156,7 +156,6 @@ class ScriptLoaderV2 implements ScriptLoader {
             final meta = ScriptMeta.get(script)
             meta.setScriptPath(path)
             meta.setModule(module)
-            meta.validate()
```
jorgee, line 159: Is it required to remove the validate()? I think only with the checkComponentName changes are enough. Or do you want to remove any validation at runtime for V2 parser?

bentsherman: One case checks for conflicting imports vs definitions in a module, while the other case checks for conflicting definitions in the same module. So I figure we might as well remove both since they are both handled at compile-time

PR #6542, opened 2025-11-05
Fix false config validation warnings for process selectors

### LINKED ISSUE #6532: Unrecognized config options when withName selector has special characters

When I use a withName process selector that has special characters (e.g. `'.*TASK.*'`), I get a warning about unknown config options:
e.g. 
```
WARN: Unrecognized config option 'process.*TASK.*.ext.args'
```

main.nf
```
workflow {
    TASK(1).view()
}

process TASK {
    input:
    val thing

    output:
    stdout

    script:
    def args = task.ext.args ?: 'nothing'
    def prefix = task.ext.prefix ?: 'default'
    """
    echo "${thing} ${args}" | tee "${prefix}.txt"
    """
}
```

nextflow.config
```
resume = true

process {
    withName: '.*TASK.*' {
        ext.args   = 'hello'
        ext.prefix = 'prefix'
    }
}
```

command
```
NXF_SYNTAX_PARSER=v2 nextflow run .
```

The issue goes away if the special chars are removed from the withName selector. For example, if the nextflow.config is replaced by this:
```
resume = true

process {
    withName: 'TASK' {
        ext.args   = 'hello'
        ext.prefix = 'prefix'
    }
}
```

### Program output 

```

Launching `main.nf` [exotic_torricelli] DSL2 - revision: eeea9e75bc

WARN: Unrecognized config option 'process.*TASK.*.ext.args'
WARN: Unrecognized config option 'process.*TASK.*.ext.prefix'
executor >  local (1)
[b0/34fb22] process > TASK [100%] 1 of 1 ✔
1 hello
```

### Environment 

* Nextflow version: 25.10.0 with NXF_SYNTAX_PARSER=v2
* Java version: 21.0.2
* Operating system: macOS
* Bash version: GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)

bentsherman:
Close #6532 

I just refactored the validation logic to match the [ConfigSpecVisitor](https://github.com/nextflow-io/language-server/blob/main/src/main/java/nextflow/lsp/services/config/ConfigSpecVisitor.java) in the language server, instead of flattening the config object which creates all sorts of edge cases.

Also added some unit tests to improve test coverage.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1,333 +1,176 @@
 (cli-page)=
 
-# Command line
+# Command line interface
 
-Nextflow provides a robust command line interface (CLI) for the management and execution of pipelines.
+Nextflow provides a robust command line interface (CLI) for the management and execution of pipelines. This page explains the key concepts and common usage patterns for the CLI.
 
-Simply run `nextflow` with no options or `nextflow -h` to see the list of available top-level options and commands. See {ref}`cli-reference` for the full list of subcommands with examples.
+For the complete reference of all commands, subcommands, and options, see {ref}`cli-reference`.
 
 :::{note}
-Nextflow options use a single dash prefix, e.g. `-resume`. Do not confuse with double dash notation, e.g. `--resume`, which is instead used for {ref}`Pipeline parameters <cli-params>`.
+Nextflow uses two types of command line flags:
+- Nextflow options use a single dash (e.g., `-log`) and modify Nextflow's behavior.
+- Pipeline parameters use a double dash (e.g., `--input`) and are passed to your pipeline script.
 :::
 
-## Basic usage
+## Pipeline execution
 
-### Hard configuration override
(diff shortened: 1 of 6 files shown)


modules/nextflow/src/main/groovy/nextflow/config/ConfigValidator.groovy
```diff
@@ -89,35 +89,35 @@ class ConfigValidator {
...
+     * @param scopes
+     */
+    void validate(Map<String,?> config, List<String> scopes=[]) {
+        for( final entry : config.entrySet() ) {
+            final key = entry.key
+            final value = entry.value
+
+            final names = scopes + [key]
+
+            if( names.size() == 2 && names.first() == 'profiles' )
```
jorgee, line 105: This statement is slightly different that what was before. It could introduce a side effect. Is it possible to have a case where you have two scopes and first is profile?

bentsherman: It's the same behavior, just optimizing it while I'm here. A profile block always has the form `profiles` -> `<profile_name>`

jorgee (approved):

PR #6553, opened 2025-11-11
Fix access to public S3 buckets from restricted Instance/Job Roles

### LINKED ISSUE #4732: Input data from misconfigured public S3 buckets when using AWS credentials
## Usage scenario 

Many commonly used public datasets are stored on public S3 buckets which are configured in a way that only allows access using anonymous AWS credentials (e.g. [AWS iGenomes](https://ewels.github.io/AWS-iGenomes/)). AWS supports anonymous access to S3 when AWS credentials are configured using the `--no-sign-request` AWS CLI flag.

On the other hand, when running on AWS Batch, AWS credentials are required to access all AWS resources & services needed to run on a private AWS Batch cluster.

Nextflow allows configuring a single set of AWS credentials (or role) to be used during runtime, and accesses all S3 URLs provided to channels the same way, meaning you cannot use public datasets as mentioned above when running on your own private AWS Batch cluster, and also not in conjunction with datasets stored in private S3 buckets.

## Suggest implementation 

My suggestion is to add support for a new option flag to the `fromPath` channel factory (and maybe others as well), which when set to true, given an S3 URL as the path, uses the `--no-sign-request` flag when generating the AWS CLI command that pulls the data from the given S3 bucket.

This way the user will have the granularity required in order to access both public & private S3 buckets during the same run, while running on a private AWS Batch cluster as well, regardless of how the public S3 buckets are configured.

jorgee:
close #4732

Depite public s3 buckets can be accessed without s3 credentials, the access to this buckets using the AWS SDK or CLI can fail if we get the credentials from instance or job roles that only allow to access a certain private s3 buckets. It produces errors when running pipelines combining public s3 buckets with private buckets. 

The `aws.client.anonymous` option does not solve the issue by two reasons:
1- It only aplies to SDK actions at head node, and some stage-in operations are happening at the AWS Batch job that uses the AWS CLI.
2- It is aplied to all the clients. If applied the head job cannot access private buckets. 

This PR implements a fallback mechanism for S3 download operations in the CLI and SDK.
- In the case of the task stage in with CLI, Nexflow tries first with normal call and if fails by AccessDenied or Forbidden it tries adding --no-sign-request.
- In the case of the s3 client, we are creating a client per bucket. When creates the client it tries to access the bucket with the credentials and if fails it tries to access anonymously. If it's able to access with anonymous credentials, it configures the bucket client with anonymous credentials.

I think this approach can deprecate the `aws.client.anonymous` flag

### DIFF
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
@@ -109,13 +109,12 @@ public class S3Client {
 	private void checkAndFallbackToAnonymousIfNeeded(String bucketName, S3SyncClientConfiguration clientConfig) {
 		try {
 			// Try to access the bucket with current credentials
-			runWithPermit(() -> client.headBucket(HeadBucketRequest.builder().bucket(bucketName).build()));
+			client.headBucket(HeadBucketRequest.builder().bucket(bucketName).build());
 			log.trace("Bucket {} is accessible with current credentials", bucketName);
 		} catch (S3Exception e) {
 			// Check if it's an access denied error (403) or forbidden
 			if (e.statusCode() == 403 || e.statusCode() == 401) {
 				log.debug("Access denied to bucket {} with current credentials (status: {}), checking if bucket is public", bucketName, e.statusCode());
-
 				checkAndManagePublicBucket(bucketName, clientConfig);
 			} else {
 				// Other errors (like bucket doesn't exist) should be handled by caller


plugins/nf-amazon/src/main/nextflow/cloud/aws/AwsClientFactory.groovy
```diff
@@ -104,7 +106,8 @@ class AwsClientFactory {
                 ?: SysEnv.get('AWS_REGION')
                 ?: SysEnv.get('AWS_DEFAULT_REGION')
                 ?: fetchRegion()
-
+        this.anonymous = anonymous ?:
+            config.s3Config.anonymous
```
pditommaso, line 110:
```suggestion
        this.anonymous ?=config.s3Config.anonymous
```

there's a groovy operator for that! 😄

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
```diff
@@ -84,16 +84,80 @@ public class S3Client {
...
+
+			// Try to access the bucket anonymously
+            anonymousClient.headBucket(HeadBucketRequest.builder().bucket(bucketName).build());
+
+            // If no error update client
+            updateClient(anonymousFactory, anonymousClient);
+		} catch (Exception e) {
+			log.trace("Bucket {} is not publicly accessible: {}", bucketName, e.getMessage());
+			if(anonymousClient != null)
+                anonymousClient.close();
```
pditommaso, line 151:
```suggestion
            anonymousClient?close();
```

plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsBatchFileCopyStrategyTest.groovy
```diff
@@ -202,11 +202,28 @@ class AwsBatchFileCopyStrategyTest extends Specification {
...
+                    
+                        if ! output=$(aws "${args[@]}" 2>&1); then
+                            if echo "$output" | grep -Eq "(AccessDenied|Forbidden|403)"; then
+                                echo "Access denied, retrying unsigned request..."
+                                aws --no-sign-request "${args[@]}"  
+                            else
+                                echo "$output"
+                                return 1
+                            fi
+                        else
```
pditommaso, line 225: Theres' no way to avoid the fallback logic and make this predictable?

plugins/nf-amazon/src/main/nextflow/cloud/aws/AwsClientFactory.groovy
jorgee: Note `anonymous` is the method argument and it is different from the object property `this.anonymous`

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
jorgee: It is a java class. As far as I know it is not posible to use `?`

pditommaso: You are right

plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsBatchFileCopyStrategyTest.groovy
jorgee: It is tricky, the only way to check if it is public in this condition is just try to access. We could reduce the number of fallbacks by using the try we do in the `ls` for the `cp`, so if the `ls` has worked with` --no-sign-request`, we will use it in the `cp`. It will be a complex code but I will just fallback once per file.

jorgee: This could be the alternative only falling back for `ls`
```
      nxf_s3_download() {
            local source=\$1
            local target=\$2
            local file_name=\$(basename \$1)
            local opts=(--only-show-errors)
            local ls_output
            if ! ls_output=\$($cli s3 ls \$source 2>&1); then
                if echo "\$ls_output" | grep -Eq "(AccessDenied|Forbidden|403)"; then
                    echo "Access denied, retrying unsigned request..."
                    if ! ls_output=\$($cli s3 ls --no-sign-request \$source 2>&1); then
                        echo \$ls_output
                        return 1
                    else
                        opts+=(--no-sign-request)
                    fi
                else
                    echo \$ls_output
                    return 1
                fi
            fi   
                    
            local is_dir=\$(echo \$ls_output | grep -F "PRE \${file_name}/" -c)
            
            if [[ \$is_dir == 1 ]]; then
                opts+=(--recursive)
            fi
            $cli s3 cp "\${opts[@]}" "\$source" "\$target"
        }
```

pditommaso:
In my view we should rethink the settings for S3 client and moving to a model in which it's possible to declare the settings a bucket level e.g. region, anonymous, (likely secrets, etc). 

The model having a client instance for bucket was done having this approach in mind. This would make it possible to have predictable settings for each bucket instead of having global anonymous and fallback attempts

jorgee:
Do you mean something like we do for executors?. Keep current `aws.client.<option>` for general and have specific bucket client configuration with something like `aws.cli.$bucket_name.<option>` that overwrite the general option for a specific bucket.

pditommaso:
Yes, exactly. Maybe something more self-describing as syntax e.g. `aws.s3.<BUCKET>`.

bentsherman:
In that case I would call it `aws.client.buckets.<bucket>`. You can follow the same pattern used for `azure.batch.pools`

bentsherman:
From today's discussion, we won't try to backport a fix since it never worked anyway, Instead we will focus on enabling per-bucket credentials both for Nextflow and Fusion

bentsherman:
Closing in favor of #6584

PR #6558, opened 2025-11-12
Fix .command.stage file write when using Google Cloud

### LINKED ISSUE #5888: Staging script not found when running on Google Batch
## Bug report 

When running Nextflow on Google Batch with gcsfuse mounted directories, and attemping to stage in many input files for a task, the task will fail with the following error:
```
Error executing process > 'COPY_FILES'

Caused by:
  No such file or directory: /mnt/disks/nf-tower-test-eu-1/scratch/1rnFXXfWvpW08n/f7/b2319081b9b7957733eebb846c0fd3/.command.stage

```
This appears to be related to the fix implemented in [#4282](https://github.com/nextflow-io/nextflow/pull/4282) and reported in [#4279](https://github.com/nextflow-io/nextflow/issues/4279), which was intended to disable the separate staging script for remote object storage entirely. However, the fix doesn't properly work on Google Batch.

1. Run a Nextflow pipeline on GCP with a task that stages in many input files (e.g., ~1000 or more files
```bash
# create random files
for ((n=0;n<6000;n++)); do touch dummy_file_${n}.txt; done

# sync to a GCS bucket
gsutil -m rsync -r ./ gs://nf-tower-test-eu-1/esha/many_files_test/
```
One process workflow:
```
process COPY_FILES {
    input:
        path files
    output:
        path("outdir", type: 'dir')
    script:
    """
    mkdir -p outdir
    for f in ${files}; do
        cp \$f outdir/
    done
    """
}

workflow {
    Channel.fromPath(params.input).collect()
    | COPY_FILES()
}
```
3. Instead of staging in the directory, stage in each individual file which will result in a large `.command.run` exceeding 1MB.
4. The task fails because it tries to access the .command.stage file which isn't properly created or accessible

### Environment 

* Nextflow version: 24.10.5
* Seqera Platform Cloud Version 24.3.0-cycle4_803f393

### Additional context

jorgee:
close #5888

.command.stage file is created when stage script is larger than 1MB and stage file is in the default filesystem. In the case of Google Batch executor, the stage file is /mnt/xxx/ because of the gcsfuse. Later, it tries to write the file that doesn't exist in the head node, so it fails. It is mainly due to the targetStage file returns the mount path instead of the remote path as for .command.run, etc.

In this PR, I have fixed the error by returning he remote version of the stage file instead of the localone when requesting the targetStageFile.


bentsherman:
Ideally it should not write the stage script at all for google batch. That was the intention of this check:

https://github.com/nextflow-io/nextflow/blob/53c28b85aa941624d9734551a382348e8fb48e55/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy#L275-L282

But I didn't consider that the google batch script launcher overwrites the task workDir to be the local container-mounted path. So I'm not sure if there is a clean way to fix this check

jorgee:
I fix changing this because it was the simpler one that implied less modifications. The easy approaches I was considering is  overriding the `stageCommand` method (the one that you mention) in the Google case with just the 'else' part, or encapsulate the check in a function and override to return false for the Google batch case. Do you think one of these is more suitable?

bentsherman:
I feel like the cleanest solution is to add a method to `Executor` that specifies whether to use the stage script. It should be `false` by default and the `AbstractGridExecutor` should override it to be `true`.

For that I think you'll have to add a field to the `TaskBean` since the `BashWrapperBuilder` doesn't have a reference to the task executor

jorgee:
I have finally implemented the fix by adding the `stageFileEnabled` flag in the `BashWrapperBuilder` which is only set in the `GridTaskHandler`. To pass something from the Executor to the BashWrapperBuilder, we need to add fields in TaskRun and TaskBean. I see it cleaner and simpler.

Issue #5390, opened 2024-10-10
Provide google storage object as input to -params-file


peter-fsdx:
## New feature

Same request as this ticket. Would like to be able to pass a google storage path to the `-params-file` option. Currently this does not seem to work like it does for AWS S3 (https://github.com/nextflow-io/nextflow/issues/1412)

```
nextflow test.nf -params-file gs://MYBUCKET/test.json

 N E X T F L O W   ~  version 24.04.4

ERROR ~ Cannot initialize GsPathFactory: missing session

 -- Check '.nextflow.log' file for details
ERROR conda.cli.main_run:execute(124): `conda run nextflow test.nf -params-file gs://MYBUCKET/test.json` failed. (See above for error)
```

## Usage scenario 

Running nextflow in a cloud environment want to be able to pass the parameters to the pipeline without having a script to copy the file from cloud storage to the VM's disk.

## Suggest implementation 

Similar to what appears to be the S3 implementation, nextflow already is able to read files from google storage elsewhere in the code. Java/Groovy is not my strong suit but it appears that nextflow is trying to read the params file before the global session is set. Failing [here](https://github.com/nextflow-io/nextflow/blob/4c6f2e8541afae3a3960b04e71aaa62f744c1866/plugins/nf-google/src/main/nextflow/cloud/google/util/GsPathFactory.groovy#L45-L50)

PR #6568, opened 2025-11-14
Revert Plugins.init() at run command to original location

### LINKED ISSUE #6566: nf-amazon plugin not available for latest 25.10.0 release

When passing an s3 path for the `-params-file` parameter, I'd expect nextflow to use nf-amazon plugin to get it. This works with version 25.04.8 but not in 25.10.0 anymore.

```{bash}
nextflow run https://github.com/seqeralabs/nf-hello-world -params-file s3://example_bucket/hello/params.json
```

### Program output 

```

WARN: Unable to start plugin 'nf-amazon' required by s3://example_bucket/hello/params.json
ERROR ~ Missing plugin 'nf-amazon' required to read file: s3://example_bucket/hello/params.json

```
> I get the same error even if I use an actual s3 path so the path itself doesn't matter....

### Environment 

* Nextflow version: 25.10.0 
* Java version: openjdk version "17.0.6" 2023-01-17 LTS
* Operating system: macOS

### Additional context

If I downgrade to nextflow version `25.04.8` then it works correctly (with the above command it then complains that the s3 path doesn't exist but thats a valid error)

jorgee:
close #6566 

Commit 9e4c975 moved the Plugins.init() later in the code- Plugins.init() initializes the plugins manager, so autoStartIfMissing call done  before loading the configuration will fail. It includes getting the Repository providers implemented in plugins or downloading remote param-files. 

This PR moves again the Plugins.init() to begining of run command. This change doesn't affect secrets loading (tested in platform with `bentsherman/nf-tests -r plugin-secrets`). The first time that the config is loaded, it uses the EmptySecretsProvider, it doesn't affect if a plugin has been loaded before.


bentsherman:
Before we do this, I would like to understand why the params-file can be an S3 path but not an Azure path

It doesn't seem like either of these use cases were ever intentionally supported, considering that the params file is loaded before the config, which means the params file shouldn't be able to access remote storage

bentsherman:
It never worked for google storage either: #5390

jorgee:
The difference is in the path factory. For S3, nextflow gets the config from Global.config. For Google and Azure, it gets from Global.session.config, and they return a exception if session is null. At the point of loading the params with S3, the Global.config is null, but as it is not failing, it just create an empty Map and works with the env vars, sso, instace role, or whatever is configured in the process running the run command.  

We can do the same for the other cases and it will work. However, I see a side effect for S3 and Azure. File systems are storing the client used to access the cloud storage. In these cases, file systems are set per bucket and there is a hashmap to cache file systems based on bucket name. So, if there are specific configurations in the nextflow config file for the bucket used in the params file, they will never be applied. It will not happen in Google because the key file system cache map includes the configuration.

pditommaso:
Quick patch for 25.10.1 and separate task trying to have a common approach to this problem?

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -324,6 +324,7 @@ class CmdRun extends CmdBase implements HubOptions {
         printBanner()
 
         // -- resolve main script
+        Plugins.init()
```
bentsherman, line 326:
```suggestion
        Plugins.init()

        // -- resolve main script
```

PR #6569, opened 2025-11-14
Fix false error with task ext blocks in strict syntax

### LINKED ISSUE #4903: Unexpected concatenation of processes ext {} map in profiles directive
Hello,

First of all, I do not know if this is a bug or a feature or a bad usage of Nextflow, but it does looks like a bug to me!

## Bug report 

I have a file names `profiles.config` where I define the profiles for my workflow.
I have a `test` profile where I am configuring all my process so they can work on a minimal input.

Let's say that I have this in my `profiles.config:`
```
profiles {
  test {
    process {
      withName:'FOO' {
        ext {
          foo = 'foo'
        }
      }
      withName:'BAR' {
        ext {
          bar = 'bar'
        }
      }
      withName:'DOES_NOT_MATCH_ANYTHING' {
        ext {
          test = 'test'
        }
      }
    }
  }
}
```

**The expected behavior** is that I will only have access to `task.ext.foo` in the FOO process & `task.ext.bar` in the BAR process.

**The actual behavior** is that I have access `task.ext.foo`, `task.ext.bar` and `task.ext.test` in the FOO and BAR process.
It seems that when multiple `ext {}` are used, they are concatenated and then attributed to each process calling `ext {}`.

Using `ext.<>` rather than `ext {}` or only one `ext {}` for a process and `ext.<>` for other processes produce the expected behavior. So, the following works as expected:
```
profiles {
  test {
    process {
      withName:'FOO' {
        ext.foo = 'foo'
      }
      withName:'BAR' {
       ext.bar = 'bar'
      }
      withName:'DOES_NOT_MATCH_ANYTHING' {
        ext.test = 'test'
      }
    }
  }
}
```

```
profiles {
  test {
    process {
      withName:'FOO' {
        ext {
          foo = 'foo'
        }
      }
      withName:'BAR' {
        ext.bar = 'bar'
      }
      withName:'DOES_NOT_MATCH_ANYTHING' {
        ext.test = 'test'
      }
    }
  }
}
```

Also, using the `process {}` outside `profiles {}` produce the expected behavior:
```
process {
  withName:'FOO' {
    ext {
      foo = 'foo'
    }
  }
  withName:'BAR' {
    ext {
      bar = 'bar'
    }
  }
  withName:'DOES_NOT_MATCH_ANYTHING' {
    ext {
      test = 'test'
    }
  }
}
```
### Environment 

* Nextflow version: 23.05.0
* Java version: [?]
* Operating system: Linux
* Bash version: GNU bash, version 4.2.46(2)-release (x86_64-redhat-linux-gnu)

bentsherman:
Close #4903 

This PR fixes an issue with process `ext` blocks specified in a config profile

The original issue was about `ext` blocks under different selectors being merged when using the old syntax parser. The strict parser doesn't have this problem, but had some other issues that this PR addresses.

Since the old config parser is very complicated and fragile, and will be deprecated in 26.04, I have opted to fix the issues only in the strict parser.


pditommaso:
I've tried to backport to 25.10.1 but got a conflict. skipping for now.

PR #6572, opened 2025-11-14
Move task hashing logic to separate class

bentsherman:
This PR moves the task hashing logic from `TaskProcessor` to a separate class `TaskHasher`

- Isolates the task hashing logic, making it easier to understand and test
- Makes the TaskProcessor easier to read

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskHandler.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskHandler.groovy
@@ -221,6 +221,8 @@ abstract class TaskHandler {
         record.env = task.getEnvironmentStr()
         record.executorName = task.processor.executor.getName()
         record.containerMeta = task.containerMeta()
+        record.accelerator = task.config.getAccelerator()?.request
+        record.accelerator_type = task.config.getAccelerator()?.type
 
         if( isCompleted() ) {
             record.error_action = task.errorAction?.toString()
--- a/modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
@@ -103,7 +103,9 @@ class TraceRecord implements Serializable {
             vol_ctxt: 'num',        // -- /proc/$pid/status field 'voluntary_ctxt_switches'
             inv_ctxt: 'num',        // -- /proc/$pid/status field 'nonvoluntary_ctxt_switches'
             hostname: 'str',
-            cpu_model:  'str'
+            cpu_model:  'str',
+            accelerator: 'num',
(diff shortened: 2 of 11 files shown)


modules/nextflow/src/test/groovy/nextflow/processor/TaskProcessorTest.groovy
```diff
@@ -300,63 +300,6 @@ class TaskProcessorTest extends Specification {
 
     }
 
-
-    def 'should get bin files in the script command' () {
```
jorgee, line 304: Why this test has been removed. It doesn't seem to be related to TaskHasher.

modules/nextflow/src/test/groovy/nextflow/processor/TaskProcessorTest.groovy
```diff
@@ -387,34 +330,6 @@ class TaskProcessorTest extends Specification {
 
     }
 
-    def 'should get task directive vars' () {
```
jorgee, line 390: Same as before, it doesn't seem to be related with task hasher.

modules/nextflow/src/test/groovy/nextflow/processor/TaskProcessorTest.groovy
bentsherman: Both of these tests were moved to the TaskHasherTest and refactored. This is because the bin scripts and task directive vars are only used to calculate the task hash.

I believe the only other place these methods are used is in the LinObserver, and that's only to preserve the hash components in the TaskRun lineage record

pditommaso (changes requested):
Bit against reshuffling code without a clear motivation.

bentsherman:
The motivation is in the PR description

modules/nextflow/src/main/groovy/nextflow/processor/TaskHasher.groovy
```diff
@@ -0,0 +1,257 @@
...
+        }
+
+        return result.toString()
+    }
+
+    /**
+     * Get the mapping of global variables that were referenced by
+     * the task script, excluding references to `task.ext`.
+     */
+    Map<String,Object> getTaskGlobalVars() {
```
jorgee, line 184: These 3 functions `getTaskGlobalVar`, `getTaskExtensionDirectiveVars` and `getTaskBinEntries` are just using the `TaskRun` data and they are called from `TaskHasher` and `LinObserver.` Is it not better to move to `TaskRun` instead  `TaskHasher`?

bentsherman: Thanks for the suggestion. I think it makes more sense to keep these methods in the TaskHasher because they pertain specifically to the task hash. The goal is to isolate the concept of the task hash

PR #6574, opened 2025-11-15
Improve support for pipeline chaining

bentsherman:
This PR adds several features to facilitate pipeline chaining based on workflow inputs/outputs

- [x] Print JSON representation of workflow outputs at end of run
- [x] Allow code snippets to implicitly declare params and outputs ?
- [x] Allow JSON params to be supplied via stdin ?

Demo project: [bentsherman/pipeline-chaining-demo](https://github.com/bentsherman/pipeline-chaining-demo)

The JSON output is very similar to the lineage record (`lid://<workflow-run-hash>#output`), so maybe we could use that instead. The lineage record has the advantage of using LIDs for files, but the JSON structure doesn't quite match the params file.

With a bit more magic we could even make the pipeline chain as simple as a Bash pipeline:

```bash
# print only output JSON to stdout when params are supplied via stdin
cat params.json | nf run fetchngs | nf run rnaseq --transcriptome [...]
```

But the main thing for now is the basic reading/writing of inputs/outputs, then enhanced validation with schemas and record types.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdKubeRun.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdKubeRun.groovy
@@ -88,7 +88,7 @@ class CmdKubeRun extends CmdRun {
     @Override
     void run() {
         final scriptArgs = (args?.size()>1 ? args[1..-1] : []) as List<String>
-        final pipeline = stdin ? '-' : ( args ? args[0] : null )
+        final pipeline = args ? args[0] : null
         if( !pipeline )
             throw new AbortOperationException("No project name was specified")
         if( hasAnsiLogFlag() )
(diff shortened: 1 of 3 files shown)


bentsherman:
The problem with the stdin params right now is that in the case of a 3-step chain, the second step blocks to wait for stdin and the third step skips the stdin because it is not "available". This leads to a lock error since the first and third run try to lock the history file.

If I skip the `System.in.available()` check, then the run hangs when there is actually no stdin.

I'll have to see how composable CLI tools normally handle this. I assume each tool has to initialize stdout before waiting for stdin, in order to signal to the next tool in the pipeline that stdin is coming

bentsherman:
Notes:

- Printing the workflow outputs JSON seems like a good idea, but probably should be opt-in via CLI option

- The implicit params / output blocks for code snippets was considered too "magical", and it can't support full type validation. Especially with LLMs writing code, I don't think this is needed

- Receiving params from stdin is a fun shorthand but not really needed right now

TODO: trim PR scope to just the workflow outputs JSON with CLI option

bentsherman:
Closing in favor of #6875 . We can revisit if we decide to enable params via stdin

PR #6581, opened 2025-11-19
Add `listDirectory()` to Path type and deprecate `listFiles()`

### LINKED ISSUE #6152: Path `listFiles` return type not accepted by process `path` input
## Bug report 

I have a channel emitting a directory path, on which I call `.listFiles()`. However the class of the output collection of paths is `Ljava.nio.file.Path`, while a process input `path` seems to expect other kinds of collection (works with `java.util.ArrayList` and `nextflow.util.ArrayTuple`).

Should not require an extra conversion step (need to wrap in `tuple` or `as List`).

Create a directory with two files:

```bash
mkdir testdir
touch testdir/{a,b}.txt
```

Run the following nextflow script:

```nextflow
process take_files {
    input:
    path(infiles)

    output:
    stdout

    script:
    """
    ls *
    """
}

workflow {
    Channel.fromPath("testdir", type:"dir", checkIfExists: true)
    | map { it.listFiles() }
    | take_files
    | view
}
```

### Program output 

```stderr
ERROR ~ Error executing process > 'take_files (1)'

Caused by:
  Not a valid path value type: [Ljava.nio.file.Path; ([/home/Gullumluvl/testdir/a.txt, /home/Gullumluvl/testdir/b.txt])
```

### Environment 

* Nextflow version: 25.03.1-edge 
* Java version: OpenJDK 64-Bit Server VM 21.0.7+6-Debian-1
* Operating system: Debian 13 Linux 6.12.27-amd64

bentsherman:
Fix #6152 

The `listFiles()` method is problematic because the implementation returns a `Path[]` instead of a `Collection<Path>`. The array type supports most of the same functionality, but not all, leading to weird bugs such as not being able to supply as a process path input.

Additionally, TaskPath does not implement `listFiles()` correctly, which leads to strange errors when calling `listFiles()` on a task input in the task body.

- Introduce `listDirectory()` as a replacement for `listFiles()` that doesn't have the above issues

- Deprecate `listFiles()` in the Nextflow language (doesn't need to be deprecated in the runtime)

- Clean up documentation of Path type


pditommaso:
This can introduce breaking changes in third parties pipelines. Not sure the change is worth

bentsherman:
The problem is that these methods are documented in the Nextflow language as returning a list instead of an array:

`list() -> List<String>`
> Returns the first-level elements (files and directories) of a directory as a list of strings.

`listFiles() -> List<Path>`
> Returns the first-level elements (files and directories) of a directory as a list of Paths.

Because arrays don't exist in the Nextflow type system

It looks like everyone is wrapping `listFiles()` with an extra call like `.toList()` or `as List<Path>` in order to use it like an actual list. So for them it should not be breaking, it will just enable additional functionality

christopher-hakkaart (approved):

pditommaso (changes requested):
I understand the point about type system, however the array type has been used to keep aligned with the `File` version in Groovy SDK, until there isn't a clear boundary between nextflow and groovy and java API this change little. 

Secondly and even this kind of changes are extremely annoying for users because may introduce subtle bugs hard to detect, and confuse agents. 

If this change is really needed better to deprecated, document and induce and new API

Issue #6447, opened 2025-10-03
can't create parent directories when stageOutMode is rsync


anoronh4:
### Expected behavior and actual behavior

When using storeDir and scratch at the same time, i would think that storeDir would be able to create any parent folders that don't already exist so that rsync doesn't fail. 

### Steps to reproduce the problem

i have the following test script:
```
process A {

storeDir "./folder1/folder2/"
stageOutMode 'rsync'

input:
val(x)

output:
path("*.txt")

script:
"""
touch ${x}.txt
"""

}


workflow {
A("input")
}
```
changing stageOutMode to copy or move will not result in this error. nxf_fs_rsync in the .command.run is probably missing some mkdir commands. in fact, nxf_fs_move and nxf_fs_copy do have mkdir -p commands, and nxf_fs_rsync is missing them.

### Program output 

```
ERROR ~ Error executing process > 'A'

Caused by:
  Process `A` terminated with an error exit status (3)


Command executed:

  touch input.txt

Command exit status:
  3

Command output:
  (empty)

Command error:
  rsync: change_dir#3 "/data/noronhaa/forte_recreate/workspace/demo_rsync/folder1" failed: No such file or directory (2)
  rsync error: errors selecting input/output files, dirs (code 3) at main.c(702) [Receiver=3.1.3]
  cp: '.command.out' and '.command.out' are the same file
  cp: '.command.err' and '.command.err' are the same file
```

### Environment 

* Nextflow version: 25.04.7 
* Java version: 23
* Operating system: Linux
* Bash version: GNU bash, version 4.4.20(1)-release (x86_64-redhat-linux-gnu)

bentsherman:
Does `rsync` not create parent directories automatically? I thought it did

anoronh4:
apparently not. are you not able to reproduce with my example? i think it will create the directory structure that exists in the source location, but in my example ./folder1/folder2/ does not exist in the source, and i want that to be my destination.

maybe worth mentioning, i have rsync version 3.1.3, not sure if there's another version of rsync that works differently.

anoronh4:
may be related to #6214

PR #6587, opened 2025-11-19
Fix issues with rsync stage-out mode

### LINKED ISSUE #6214: Using `storeDir` with `process.stageOutMode = "rsync"` creates an invalid cache dir

Using the `storeDir` directive in combination with the `rsync` `stageOutMode` is unable to create a valid storage directory which causes the pipeline to fail. Everything works fine when using the directive with a `copy` `stageOutMode`

Expected behaviour (`copy` with `storeDir 'example'`):

```bash
$ tree example/
example/
├── 1.txt
├── 2.txt
├── 3.txt
├── 4.txt
└── 5.txt
```

Actual behaviour (`rsync` with `storeDir 'example'`):

```bash
$ ll
...
-rw-rw-r--  1 nvnieuwk nvnieuwk    42 Jun 25 10:44 example
...
```

A file has been created with the name of the `storeDir` instead of a directory containing the correct files

**main.nf:**
```groovy
workflow {
    def numbers = Channel.of(1, 2, 3, 4, 5)
    EXAMPLE(numbers)

}

process EXAMPLE {
    storeDir 'example'
    tag "$number"

    input:
    val number

    output:
    path('*.txt')

    script:
    """
    echo "This is an example process with number: ${number}" > ${number}.txt
    """
}
```

**nextflow.config:**

```groovy
params.mode = "rsync"
process.stageOutMode = params.mode
```

The issue can be replicated by running the following command in a directory with the files presented above:

```bash
rm -rf example && nextflow run main.nf --mode rsync
```

Note that this method does work when the `storeDir` is already filled with the correct files, that's why you should remove it every time you run this command

The expected behaviour can be replicated by running the following command

```bash
rm -rf example && nextflow run main.nf --mode copy
```

### Program output 

```
ERROR ~ Error executing process > 'EXAMPLE (3)'

Caused by:
  Missing output file(s) `*.txt` expected by process `EXAMPLE (3)`


Command executed:

  echo "This is an example process with number: 3" > 3.txt

Command exit status:
  0

Command output:
  (empty)

Work dir:
  /home/nvnieuwk/Documents/nextflow/issues/rsync-storeDir/work/93/2feb18d4f824a9d9c247c773b31e15

Tip: when you have fixed the problem you can continue the execution adding the option `-resume` to the run command line

```

### Environment 

* Nextflow version: 25.04.2 (but also noticed in older versions) 
* Java version: 17.0.3
* Operating system: Linux
* Bash version: 5.0.17

bentsherman:
Close #6214 #6447 

This PR fixes the `nxf_fs_rsync()` function in the `.command.run` template

The problem was that `rsync` does not actually initialize the target directory prior to copying, whereas I think we assumed that it did. This led to multiple issues:

- sometimes rsync will copy the source file "as" the target rather than "into" the target (#6214)

- rsync will fail if the target parent directory doesn't already exist (#6447)

### DIFF
--- a/docs/reference/process.md
+++ b/docs/reference/process.md
@@ -1656,17 +1656,17 @@ The `stageOutMode` directive defines how output files are staged out from the sc
 : Output files are copied from the scratch directory to the work directory.
 
 `'fcp'`
-: :::{versionadded} 23.02.0-edge
+: :::{versionadded} 23.04.0
   :::
-: Output files are copied from the scratch directory to the work directory by using the [fcp](https://github.com/Svetlitski/fcp) utility (note: it must be available in your cluster computing nodes).
+: Output files are copied from the scratch directory to the work directory by using the [fcp](https://github.com/Svetlitski/fcp) utility (note: it must be available in the task environment).
 
 `'move'`
 : Output files are moved from the scratch directory to the work directory.
 
 `'rclone'`
-: :::{versionadded} 23.01.0-edge
+: :::{versionadded} 23.04.0
   :::
-: Output files are copied from the scratch directory to the work directory by using the [rclone](https://rclone.org) utility (note: it must be available in your cluster computing nodes).
+: Output files are copied from the scratch directory to the work directory by using the [rclone](https://rclone.org) utility (note: it must be available in the task environment).
 
 `'rsync'`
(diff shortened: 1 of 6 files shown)


christopher-hakkaart (approved):
Docs look good 👍

PR #6591, opened 2025-11-20
Check Nextflow version before loading plugins

### LINKED ISSUE #6540: Nextflow version check happens after loading the plugins
## Bug report 

Manifest has (cf https://github.com/nf-core/rnavar/blob/dev/nextflow.config#L349)
```
    nextflowVersion = '!>=25.10.0'
```

Yes I can see this when I run on 25.04.8, when I would expect the pipeline to stop

```
Nextflow 25.10.0 is available - Please consider updating your version to it
N E X T F L O W  ~  version 25.04.8
Pulling nf-core/rnavar ...
 downloaded from https://github.com/nf-core/rnavar.git
Launching `https://github.com/nf-core/rnavar` [nostalgic_morse] DSL2 - revision: 9bcf75c395 [dev]
Downloading plugin nf-schema@2.5.1
Downloading plugin nf-core-utils@0.4.0
ERROR ~ Plugin nf-core-utils with version @0.4.0 does not exist in the repository
```

Error is misleading and due to the fact that this plugin minimal Nextflow version is 25.10.0.

I would expect the pipeline not to start, and after talking with @bentsherman we believe it's caused by the fact that this version check happens after the plugins are loaded.

(Provide a test case that reproduces the problem either with a self-contained script or GitHub repository)

### Program output 

(Copy and paste the output produced by the failing execution. Please highlight it as a code block. Whenever possible upload the `.nextflow.log` file.)

### Environment 

* Nextflow version: [?] 
* Java version: [?]

### Additional context

bentsherman:
Close #6540 

This PR moves the Nextflow version check (`manifest.nextflowVersion`) so that it happens as early as possible -- immediately after the config is loaded

This will improve the error message for errors that are downstream of a version mismatch. For example, in the linked issue, a pipeline requiring >=25.10 was run with 25.04, and because it was also loading a plugin that requires >=25.10, the resulting error was about the plugin not existing rather than the Nextflow version being incorrect


modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -677,6 +682,45 @@ class CmdRun extends CmdBase implements HubOptions {
         return result
     }
 
+    /**
```
jorgee, line 685: Shouldn't be better to create an independent class in nf-commons, in case it is required for other commands or plugins in the future?

bentsherman: Good idea, likely it will be used by both CLI v1 and CLI v2

CmdRun also has logic for (1) resolving the script file and (2) resolving CLI params. These could also be moved into helper classes in the future for the same reason. I will keep this in mind for CLI v2

bentsherman: @jorgee on second thought, it might be better to keep the logic in CmdRun for now. I'm not sure at this point what the right interface should be for a separate class -- should it take the config map? or a Manifest? or a version string? The answer will become clear when new features like CLI v2 need to re-use this logic.

jorgee: Ok, that's fine. I see the Manifest class also requires the config scope; it seems more difficult to isolate in an independent class.

Issue #6604, opened 2025-11-24
Repository assets which are not a sibling of the executed workflow are not correctly identified.


DriesSchaumont:
## Bug report 

When executing a workflow which resides in a subdirectory (using `main-script`), other repository assets are not recognized when they are not a child from the same subdirectory. This causes cache invalidation when the repository is fetched again (as is often the case on remote executors like `batch`). 

### Expected behavior and actual behavior 

When an input is not recognized as being an assets, the default cache strategy is applied instead of using a `sha256` checksum or the commit ID (~ `deep` mode). With the default strategy the metadata of the file is used, but because the repository often cloned with each run, resume never functions because the timestamp of the asset has changed.

The main switch for determining wether an input is an asset is the [`isAsset` function](https://github.com/nextflow-io/nextflow/blob/6c0813765cbeb6ca54fcd1d9d2364b3e06ef472b/modules/nf-commons/src/main/nextflow/util/HashBuilder.java#L514). It checks if the a certain file starts with the base directory (`session.getBaseDir()`). However, this base directory points to wherever the main script resides and not the root of the repository. `isAsset` is being used as a switch to determine wether to use the file contents or the user specified hashing strategy (no assets) or to use the file contents regardless of the chosen strategy (assets): https://github.com/nextflow-io/nextflow/blob/6c0813765cbeb6ca54fcd1d9d2364b3e06ef472b/modules/nf-commons/src/main/nextflow/util/HashBuilder.java#L265  

An asset should always be recognized as being part of the repository; regardless of its location. A proposed solution can be found at https://github.com/nextflow-io/nextflow/pull/6605. 

### Steps to reproduce the problem

As a sidenote: when the metadata is used for hashing a file, the[`hashFileMetadata`](url) method is called, which outputs `Hashing file meta:` to the logs; while when the file contents is used a log message of `Hash asset file sha-256` is output by e.g. [`hashFileSha256Impl0`](https://github.com/nextflow-io/nextflow/blob/6c0813765cbeb6ca54fcd1d9d2364b3e06ef472b/modules/nf-commons/src/main/nextflow/util/HashBuilder.java#L383C29-L383C48).

In order to reproduce the problem, I've created a repository here: https://github.com/DriesSchaumont/nextflow-subfolder-basedir-asset-caching

#### Cleaning the environment
```bash
[...]

PR #6605, opened 2025-11-24
Fix asset detection when item is not a sibling of the executed script

DriesSchaumont:
See #6604 

Nextflow uses the contents of a file to determine its hash regardless of the requested hashing strategy when the item is part of the git repository. This is because a different clone of the repository (as often happpens on remote executors like `batch`) changes the item's timestamp, which invalidates the cache when using its metadata.

When executing a workflow which resides in a subdirectory (using main-script), repository assets are not recognized when they are not a child from the same subdirectory where the script resides. This is because  `isAssetFile` uses `session.getBaseDir()` as a point of reference for where the assets are stored, but this is not the root of the repository.  

This PR changes the check for `isAssetFile` to use the asset root directory as defined by  `nextflow.Const.DEFAULT_ROOT`.

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/scm/AzureRepositoryProviderTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/scm/AzureRepositoryProviderTest.groovy
@@ -189,6 +189,7 @@ class AzureRepositoryProviderTest extends Specification {
                 new RepositoryProvider.BranchInfo('dev', 'cc0ca18640a5c995231e22d91f1527d5155d024b'),
                 new RepositoryProvider.BranchInfo('feature-x', '13456a001ba5a27d643755614ab8e814d94ef888'),
                 new RepositoryProvider.BranchInfo('master', 'a207636e419f18c4b8c8586b00e329ab4788a7f5'),
+                new RepositoryProvider.BranchInfo('test/branch+with&special-chars','a207636e419f18c4b8c8586b00e329ab4788a7f5')
         ]
     }
 
--- a/plugins/nf-k8s/src/main/nextflow/k8s/K8sTaskHandler.groovy
+++ b/plugins/nf-k8s/src/main/nextflow/k8s/K8sTaskHandler.groovy
@@ -416,8 +416,10 @@ class K8sTaskHandler extends TaskHandler implements FusionAwareTask {
 
     @Override
     boolean checkIfCompleted() {
-        if( !podName ) throw new IllegalStateException("Missing K8s ${resourceType.lower()} name - cannot check if complete")
-        def state = getState()
+        if( !podName )
+            throw new IllegalStateException("Missing K8s ${resourceType.lower()} name - cannot check if complete")
+
+        final state = getState()
         if( state && state.terminated ) {
             if( state.nodeTermination instanceof NodeTerminationException ||
                 state.nodeTermination instanceof PodUnschedulableException ) {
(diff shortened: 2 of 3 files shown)


pditommaso (changes requested):
Please provide context in the PR description and unit testing to validate the (new) `isAssetFile` semantic

DriesSchaumont:
> Please provide context in the PR description and unit testing to validate the (new) `isAssetFile` semantic

Thanks for the quick review! Updated the PR and open for new suggestions.

modules/nf-commons/src/main/nextflow/util/HashBuilder.java
```diff
@@ -509,9 +512,13 @@ static private byte[] sumBytes(byte[] resultBytes, byte[] nextBytes) {
...
      *
      * @param path
+     *      The item to check.
+     * @param asset_root
+     *      Location where assets are being stored.
      * @return
+     *      Whether or not `path` is included in the pipeline Git repository.  
      */
-    static protected boolean isAssetFile(Path path) {
+    static protected boolean isAssetFile(Path path, File asset_root) {
```
jorgee:
```suggestion
    static protected boolean isAssetFile(Path path, File assetsRoot) {
```

modules/nf-commons/src/main/nextflow/util/HashBuilder.java
```diff
@@ -522,7 +529,7 @@ static protected boolean isAssetFile(Path path) {
         if( session.getBaseDir().getFileSystem()!=path.getFileSystem() )
             return false;
         // if the file is in the same directory as the base dir it's a asset by definition
-        return path.startsWith(session.getBaseDir());
+        return path.startsWith(session.getBaseDir()) || path.startsWith(asset_root.toPath());
```
jorgee:
```suggestion
        return path.startsWith(session.getBaseDir()) || path.startsWith(assetsRoot.toPath());
```

modules/nf-commons/src/main/nextflow/util/HashBuilder.java
```diff
@@ -509,9 +512,13 @@ static private byte[] sumBytes(byte[] resultBytes, byte[] nextBytes) {
...
      *
      * @param path
+     *      The item to check.
+     * @param assetRoot
+     *      Location where assets are being stored.
      * @return
+     *      Whether or not `path` is included in the pipeline Git repository.  
      */
-    static protected boolean isAssetFile(Path path) {
+    static protected boolean isAssetFile(Path path, File assetRoot) {
```
jorgee, line 537: I think it would be more correct to use `session.workflowMetadata?.projectDir` instead of the DEFAULT_ROOT. It is unlikely that a file points to a different project asset, but it could happen.

DriesSchaumont: Thank you for the suggestion! Do `projectDir` and `baseDir` not both point to the same folder (i.e. where the main script is located, even when that is a subfolder in the repository)?

In the context of the config it seems to be (but I am unsure about `session.workflowMetadata?.projectDir`):
https://github.com/nextflow-io/nextflow/blob/0f86933ad511a15fdb9e6ea80fd12332e590c737/modules/nf-lang/src/main/java/nextflow/config/dsl/ConfigDsl.java#L37

jorgee: I have checked how the hash is built later, and it is getting the relative path from baseDir to build the hash. There are no bug difference if we consider assets from project dir or DEFAULT_ROOT. So, current solution is also valid.

jorgee (approved):

PR #6631, opened 2025-12-03
Document best practices for script and config params

bentsherman:
- Update config params example to use realistic config params
- Document best practices for declaring params in script vs config vs overriding via config profile
- Add related notes to `params` block migration for rnaseq-nf

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/local/LocalTaskHandler.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/local/LocalTaskHandler.groovy
@@ -218,6 +218,7 @@ class LocalTaskHandler extends TaskHandler implements FusionAwareTask {
              */
             if( elapsedTimeMillis() > wallTimeMillis ) {
                 destroy()
+                task.exitStatus = process.exitValue()
                 task.stdout = outputFile
                 task.stderr = errorFile
                 task.error = new ProcessException("Process exceeded running time limit (${task.config.getTime()})")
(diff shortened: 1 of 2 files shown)


docs/config.md
```diff
@@ -129,21 +129,45 @@ The following functions are globally available in a Nextflow configuration file:
...
 }
 ```
 
-See {ref}`cli-params` for information about how to specify pipeline parameters.
-
 :::{note}
 When including a config file, the included config is evaluated with the parameters that are defined before the include. Parameters defined after the include are not visible to the included config.
 :::
 
+As a best practice, parameters should be declared in the config file only if they are used by other config options. If a parameter is used in the script, it should be declared in the script and overridden in config profiles if necessary. For example:
```
christopher-hakkaart:
```suggestion
As a best practice, declare parameters in the config file only if they are used by other config options. If a parameter is used within the script, declare it there and override it in config profiles as needed. For example:
```

docs/tutorials/static-types.md
```diff
@@ -123,6 +123,14 @@ params {
 
 See {ref}`workflow-params-def` for more information about the `params` block.
 
+:::{note}
+Parameters that are only used in the config file should be declared in the config file, not the script. Since rnaseq-nf does not have any such parameters, all parameters are declared in the script. See {ref}`config-params` for more information.
```
christopher-hakkaart:
```suggestion
Parameters used only in the config file should be declared there, not in the script. Since rnaseq-nf has no such parameters, all parameters are declared in the script. See {ref}`config-params` for more information.
```

docs/tutorials/static-types.md
```diff
@@ -123,6 +123,14 @@ params {
...
+:::
+
+:::{tip}
+The rnaseq-nf pipeline initializes the `reads` and `transcriptome` parameters to a test dataset by default because it is designed to be run as a toy example. In practice, these defaults should be defined in a config profile (e.g., `test`).
```
christopher-hakkaart:
```suggestion
The rnaseq-nf pipeline initializes the `reads` and `transcriptome` parameters to a test dataset by default, as it is designed as a toy example. In practice, defaults for test data should be defined in a config profile (e.g., `test`).
```

PR #6633, opened 2025-12-03
Add runtimeClassName to the pod options

### LINKED ISSUE #6632: Add runtimeClassName support for pod specs in the K8s plugin
## New feature

The K8s executor does not support specifying the [runtime class](https://kubernetes.io/docs/concepts/containers/runtime-class/) for pods. This is needed when a K8s cluster provides multiple container runtime configurations. 

## Use case 

A very common use case consists in accessing a K8s cluster with a separate runtime configurations for containers that access *NVIDIA* GPUs, where running the pod with the default runtime would cause an error as it is not configured for this purpose. 

## Suggested implementation 

The suggested implementation consists in adding `runtimeClassName` to the `pod` directive settings.
For example:

```groovy
process {
    executor = 'k8s'

    accelerator {
        limit = 2
        type = "nvidia.com/gpu"
    }

    pod = [ runtimeClassName: 'nvidia' ]
}
```

alexpilotti:
Close #6632 

This PR adds `runtimeClassName` to the K8s pod options in order to support scenarios where the desired [container runtime configuration](https://kubernetes.io/docs/concepts/containers/runtime-class/) is not the default one as it often happens when using NVIDIA GPUs, although it can apply to other scenarios as well (e.g., kata containers). 

This PR includes also the corresponding updates in the documentation.

An initial commit is also included to remove unnecessary trailing whitespaces from the files affected by this PR to avoid unrelated changes in the subsequent commits, especially when using editors configured to automatically remove such whitespaces, as it is common practice.

Here's a typical configuration file usage example:

```groovy
process {
    executor = 'k8s'

    accelerator {
        limit = 2
        type = "nvidia.com/gpu"
    }

    pod = [ runtimeClassName: 'nvidia' ]
}
```

### DIFF
--- a/docs/amazons3.md
+++ b/docs/amazons3.md
@@ -47,9 +47,9 @@ If the access credentials are not found in the above file, Nextflow looks for AW
 2. The environment variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`
 3. The environment variables `AWS_ACCESS_KEY` and `AWS_SECRET_KEY`
 4. The profile in the AWS credentials file located at `~/.aws/credentials`
-    - Uses the `default `profile or the environment variable `AWS_PROFILE` if set
+    - Uses the `default` profile or the environment variable `AWS_PROFILE` if set
 5. The profile in the AWS client configuration file located at `~/.aws/config`
-    - Uses the `default `profile or the environment variable `AWS_PROFILE` if set
+    - Uses the `default` profile or the environment variable `AWS_PROFILE` if set
 6. The temporary AWS credentials provided by an IAM instance role. See [IAM Roles](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html) documentation for details.
 
 More information regarding [AWS Security Credentials](http://docs.aws.amazon.com/general/latest/gr/aws-security-credentials.html) are available in the AWS documentation.
--- a/docs/aws.md
+++ b/docs/aws.md
@@ -154,19 +154,19 @@ See the [bucket policy documentation](https://docs.aws.amazon.com/config/latest/
 
 ## AWS Batch
 
-[AWS Batch](https://aws.amazon.com/batch/) is a managed computing service that allows the execution of containerised workloads in the AWS cloud infrastructure. It dynamically provisions the optimal quantity and type of compute resources (e.g., CPU or memory optimized compute resources) based on the volume and specific resource requirements of the jobs submitted.
+[AWS Batch](https://aws.amazon.com/batch/) is a managed computing service that enables the execution of containerized workloads on AWS cloud infrastructure. It dynamically provisions the optimal quantity and type of compute resources (e.g., CPU or memory optimized compute resources) based on the volume and specific resource requirements of the submitted jobs.
 
-Nextflow provides built-in support for AWS Batch, allowing the seamless deployment of Nextflow pipelines in the cloud, in which tasks are offloaded as Batch jobs.
 
-Read the {ref}`AWS Batch executor <awsbatch-executor>` section to learn more about the `awsbatch` executor in Nextflow.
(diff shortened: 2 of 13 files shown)


plugins/nf-k8s/VERSION
bentsherman, line 1: You don't need to bump the version, this will happen automatically when we cut a new release

plugins/nf-k8s/changelog.txt
```diff
@@ -1,5 +1,8 @@
 nf-k8s changelog
 ===================
+1.3.1 - 03 Dec 2025
+- Add runtimeClassName support for pod specs in the K8s plugin (#6633) [aa1a9707b]
+
```
bentsherman: This can be reverted

plugins/nf-k8s/src/main/nextflow/k8s/model/PodSpecBuilder.groovy
```diff
@@ -599,6 +607,9 @@ class PodSpecBuilder {
         if( ttlSecondsAfterFinished != null )
             spec.ttlSecondsAfterFinished = ttlSecondsAfterFinished
 
+        if( runtimeClassName )
+            spec.runtimeClassName = runtimeClassName
```
bentsherman: Can the runtime class be specified in the Job spec? I assume the job would pick it up from the pod spec

alexpilotti: No, only in the pod spec, also when included in a job, but not in the job spec itself. Good question, I checked that up before sending the patch :)

plugins/nf-k8s/changelog.txt
alexpilotti: Done, I removed the last commit

plugins/nf-k8s/src/main/nextflow/k8s/model/PodSpecBuilder.groovy
bentsherman: Okay then this part can be removed because you already added the runtime class to the pod spec in the `build()` method

alexpilotti: Ops! Fixed, thanks!

bentsherman:
@alexpilotti does the PR still work on your end after the changes from the review?

alexpilotti:
> @alexpilotti does the PR still work on your end after the changes from the review?

Yes, just checked again with a fresh build, all good. Tests work as well.

alexpilotti:
> > @alexpilotti does the PR still work on your end after the changes from the review?
> 
> Yes, just checked again with a fresh build, all good. Tests work as well.

For reference, here's a `.command.yaml` from the last run, including the `runtimeClassName` spec setting:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nf-255469281f5726df29a2fa28462202f2-a1608
  namespace: nf-core
  labels: {nextflow.io/processName: NFCORE_CUDATEST_CUDATEST_NVIDIASMI, nextflow.io/runName: pensive_carlsson,
    nextflow.io/sessionId: uuid-5b9c1e91-9a08-4b7f-8efc-839ea2c1e377, nextflow.io/app: nextflow,
    nextflow.io/taskName: NFCORE_CUDATEST_CUDATEST_NVIDIASMI}
spec:
  restartPolicy: Never
  containers:
  - name: nf-255469281f5726df29a2fa28462202f2-a1608
    image: registry.nf-core:5000/nf-core-cuda
    args:
    - /bin/bash
    - -ue
    - -c
    - !!org.codehaus.groovy.runtime.GStringImpl {}
    resources:
      requests: {cpu: 4, memory: 16384Mi, nvidia.com/gpu: 2}
      limits: {memory: 16384Mi, nvidia.com/gpu: 2}
    volumeMounts:
    - {name: vol-1, mountPath: /workspace}
  activeDeadlineSeconds: 14400
  runtimeClassName: nvidia
  volumes:
  - name: vol-1
    persistentVolumeClaim: {claimName: nf-core-data-pvc}
```

PR #6637, opened 2025-12-04
docs: Add longer NXF_SYNTAX_PARSER descriptions

christopher-hakkaart:


pditommaso (approved):

docs/migrations/25-10.md
```diff
@@ -8,6 +8,14 @@
 
 <h3>Workflow params</h3>
 
+:::note
```
bentsherman, line 11: Should be `:::{note}`

docs/migrations/25-10.md
```diff
@@ -8,6 +8,14 @@
...
+To use workflow params, set the `NXF_SYNTAX_PARSER` environment variable to `v2`:
```
bentsherman, line 12: The wording kinda suggests that this is all you need to do in order to use workflow params, maybe better to phrase it the same way that you did in `workflow.md`

bentsherman: Also this note needs to be added to "Type annotations" and "New syntax for workflow handlers"

docs/process-typed.md
```diff
@@ -24,9 +24,14 @@ process hello {
 }
 ```
 
-In order to use this feature, you'll need to:
+To use this feature:
+
+1. Set the `NXF_SYNTAX_PARSER` environment variable to `v2`:
```
bentsherman, line 29: I would still phrase it as "Enable the strict syntax by setting..." so that there is a link to the strict syntax. Enabling this flag in practice will require other changes to pipeline code which are explained on that page

docs/strict-syntax.md
```diff
@@ -9,7 +9,13 @@ If you are still using DSL1, see {ref}`dsl1-page` to learn how to migrate your N
...
 
 :::{versionadded} 25.04.0
-The strict syntax can be enabled in Nextflow by setting the environment variable `NXF_SYNTAX_PARSER=v2`.
+To enable strict syntax, set the `NXF_SYNTAX_PARSER` environment variable to `v2`:
+
+```bash
+export NXF_SYNTAX_PARSER=v2
+```
+
+Without this environment variable set, Nextflow will not enforce strict syntax rules. Add this to your shell profile (e.g., `~/.bashrc`, `~/.zshrc`) or set it before you run Nextflow commands.
```
bentsherman, line 18: I'm not sure we should recommend setting this flag globally, although it may be sensible for some. I suspect most people will need to toggle it at a pipeline-level for the time being.

PR #6642, opened 2025-12-08
docs: Update config page

christopher-hakkaart:
- Refining language in the config page to make it more concise and user-focused.
- Remove reference from conceptual config page. Exists elsewhere.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -19,7 +19,7 @@ You can use the `-C <config-file>` option to specify a fixed set of configuratio
 
 ## Syntax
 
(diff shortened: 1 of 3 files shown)


bentsherman (approved):

bentsherman:
Thanks Chris. The config page was a bit too reference-y, now it should be more user-friendly.

I had to revert a few things:

- "Constants and functions" can't be removed because it is different from the script standard library. It doesn't really fit in anywhere else so for now I left it here, since it's a short list

- Some instances of active voice felt unnatural or slightly incorrect, so I kept it passive in those cases. Otherwise I'm happy to prefer active voice if you think it's better

Issue #6623, opened 2025-12-02
Params in included profile are not applied when using strict syntax


bentsherman:
## Bug report 

```groovy
// main.nf
workflow {
  print "${params.reads}"
}
```

```groovy
// nextflow.config
params {
  reads = null
}

profiles {
  test {
    includeConfig 'test.config'
  }
}
```

```groovy
// test.config
params.reads = '...'
```

### Expected behavior and actual behavior

The value of `params.reads` from the `test` profile should be applied, but it isn't.

### Steps to reproduce the problem

```console
$ NXF_SYNTAX_PARSER=v2 nextflow run main.nf -profile test
null
```

### Additional context

- The error doesn't happen when `params.reads` is not explicitly set in the main config

- The error doesn't happen when the profile is defined in the main config instead of being included from another config

bentsherman:
Found a more precise minimal example thanks to #6635 . The issue seems to be independent of profiles and the script, and doesn't happen with nested params:

```groovy
// nextflow.config
params.reads = null
params.inner.something = null

includeConfig 'test.config'
```

```groovy
// test.config
params.reads = 'not null'
params.inner.something = 'not null'
```

```console
$ NXF_SYNTAX_PARSER=v2 nextflow config
params {
   reads = null
   inner {
      something = 'not null'
   }
}
```

bentsherman:
Likely related to this conditional block:

https://github.com/nextflow-io/nextflow/blob/bb066969611219a73d0c0366d77e74f85991d0d5/modules/nextflow/src/main/groovy/nextflow/config/parser/v2/ConfigDsl.groovy#L120-L127

Which was added in #6377 in an attempt to fix CLI param overrides. So we this logic to allow included config params to override main config params, but not allow config params to override CLI params... 😵

PR #6643, opened 2025-12-09
Enable v2 syntax parser by default

bentsherman:
- Set `NXF_SYNTAX_PARSER` to `v2` by default
- Deprecate the `nextflow.enable.strict` feature flag
- Update unit tests to work with strict syntax
- Fix a few bugs with the strict syntax along the way
- Consolidate test harness for parsing/loading/running scripts into `ScriptHelper`

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/WorkflowDef.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/WorkflowDef.groovy
@@ -195,15 +195,19 @@ class WorkflowDef extends BindableDef implements ChainableDef, IterableDef, Exec
     }
 
     private Object run0(Object[] args) {
+        // add inputs to workflow binding
         collectInputs(binding, args)
+        // execute the workflow
         final closure = body.closure
         closure.setDelegate(binding)
         closure.setResolveStrategy(Closure.DELEGATE_FIRST)
         final result = closure.call()
         if( name == null ) {
+            // return the last statement if entry workflow (used for testing)
             return result
         }
         else {
+            // otherwise collect the outputs from the workflow binding
             output = collectOutputs(declaredOutputs)
             return output
         }
--- a/modules/nextflow/src/main/groovy/nextflow/script/parser/v2/ScriptLoaderV2.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/parser/v2/ScriptLoaderV2.groovy
@@ -95,6 +95,7 @@ class ScriptLoaderV2 implements ScriptLoader {
     ScriptLoaderV2 runScript() {
         assert session
         assert mainScript
+        // capture the last statement of the snippet or entry workflow (used for testing)
         this.result = mainScript.run()
         return this
     }
--- a/modules/nextflow/src/testFixtures/groovy/test/ScriptHelper.groovy
+++ b/modules/nextflow/src/testFixtures/groovy/test/ScriptHelper.groovy
@@ -45,10 +45,26 @@ import nextflow.script.ScriptType
  */
 class ScriptHelper {
 
+    /**
+     * Load a config from source text.
+     *
+     * @param text
+     */
     static Map loadConfig(String text) {
         return ConfigParserFactory.create().parse(text)
     }
 
+    /**
+     * Load a script from source text.
+     *
+     * This function compiles and executes the script without
+     * running the pipeline. The entry workflow is executed
+     * (i.e. to construct the workflow DAG) unless opts.module
+     * is enabled.
+     *
+     * @param opts
+     * @param text
+     */
     static BaseScript loadScript(Map opts = [:], String text) {
         def session = opts.config ? new MockSession(opts.config as Map) : new MockSession()
         session.setBinding(new ScriptBinding())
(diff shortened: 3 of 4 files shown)


bentsherman:
@claude please review

docs/strict-syntax.md
```diff
@@ -16,6 +16,10 @@ export NXF_SYNTAX_PARSER=v2
 ```
 :::
 
+:::{versionchanged} 26.04.0
+The strict syntax is enabled by default. It can be disabled by setting the environment variable `NXF_SYNTAX_PARSER=v1`.
```
christopher-hakkaart:
```suggestion
The strict syntax is enabled by default. You can disable it by setting the environment variable `NXF_SYNTAX_PARSER=v1`.
```

modules/nextflow/src/main/groovy/nextflow/NF.groovy
jorgee, line 71: This is related to strict syntax, right? I think it is set at CmdRun and nothing is changed

jorgee: https://github.com/nextflow-io/nextflow/blob/70e4aa33c2a1706a26fadbbc5608c104bbaaa4ac/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy#L494

bentsherman: The `nextflow.enable.strict` feature flag is something different. In fact this PR deprecates it because it is no longer needed with strict syntax enabled

ewels:
Related: A check on how many nf-core pipelines will break when we release this 😅 

https://github.com/ewels/strict-syntax-health

bentsherman:
Better let them know it's coming!

ewels:
That's the idea! Can't do much without https://github.com/nextflow-io/nextflow/issues/6623 being fixed though.

How does the timing look for merging? Is it worth moving the above fix out into its own PR for a quicker merge? We can't make the template strict-syntax compliant without that, which means that no nf-core pipelines can update until it's fixed.

bentsherman:
> I think this PR mixes two things: stabilising/improving the strict parser + promoting v2 as default

These goals are fundamentally intertwined. Making v2 the default makes it so that all of the unit tests use the v2 parser, which surfaces any remaining gaps.

For example, #6623 would have been caught by a unit test in `ConfigBuilderTest` if it used the v2 parser. To fix this issue by itself, we would need to duplicate the unit test or change it to use the v2 parser. But we're already planning to make v2 the default in 26.04 , so we might as well make all unit tests use the v2 parser.

So I see little difference between merging the two changes together vs separately. Battle-testing is what edge is for. We're already planning to make v2 the default in 26.04 so we might as well do it now

ewels:
Whilst it scares me, I generally agree with Ben here..

> battle test it into an edge

In order to really battle test it in an edge release, it needs to be the default in the edge release.

pditommaso:
Why? would not be enough to enable v2 in the nf-core tests suite?

ewels:
We can do that (though it's not trivial as it requires a template update, and we try not to push these too frequently), but I'm more thinking about people beyond nf-core.

If it's going to be default in the next stable release, I'd prefer to make it default in an edge release ASAP and then shout about it a lot. There's no way that this isn't going to cause _some_ disruption when it comes out, so the longer it sits in an edge release the better.

pditommaso:
> though it's not trivial as it requires a template update

Really, for setting a env variable? 

We can accelerate the cadence of edge releases for testing this more frequently, but from my point of view the switch should be done when there's any code/structural change.

ewels:
> Really, for setting a env variable?

Yes, because we would need to update the GitHub actions workflow scripts to do this. That involves a change to the pipeline template, which needs a release of tools, automatic sync, and for every pipeline to fix template merge conflicts, merge the automated template update PR, and do a release of their pipeline. This process typically takes several months from start to end.

We have talked about switching to using centralised / shared GitHub actions scripts, but this hasn't happened yet. And if we did that it probably wouldn't go down well, as flipping on the ENV var would immediately break all CI and no-one would be able to merge any PRs.

Releasing it as default in an edge release avoids all of this. We already have CI jobs in all pipelines that run on edge, don't block PRs but do alert authors to the fact that there is an upcoming breaking change.

jorgee:
@Ben, regarding the ConfigBuilder issue I commented to you yesterday in the call, [this](https://github.com/nextflow-io/nextflow/blob/70e4aa33c2a1706a26fadbbc5608c104bbaaa4ac/modules/nextflow/src/main/groovy/nextflow/config/ConfigBuilder.groovy#L755) is the part of code I was referring to:

This is using `NF.strictMode`, which currently only reads the value of the strict feature flag. But in `CmdRun`, the flag is set after creating the config. So, it will never apply the strict mode at this point. 

I think `NF.isStrictMode()` should be implemented as follows to ensure strict mode is applied either when using v2 parsing or using the feature flag.
```
    static boolean isStrictMode() {
        NextflowMeta.instance.isStrictModeEnabled() || isSyntaxParserV2()
    }
```

jorgee:
Apart from my previous comment, I would also be in favor to merge and release in an edge. In case we get terrible feedback, the changes seem to be valid for both parsers, so reverting to V1 again will be just modifying the line to set the default again to V1 and the corresponding docs.

bentsherman:
> This is using NF.strictMode, which currently only reads the value of the strict feature flag. But in CmdRun, the flag is set after creating the config. So, it will never apply the strict mode at this point.

@jorgee it's fine, the strict check in `mergeMaps` is not needed with the v2 parser. It will just be removed when we remove the strict feature flag

bentsherman:
We are now tracking strict syntax errors across all nf-core pipelines / modules / subworkflows:

https://github.com/nf-core/strict-syntax-health

(thanks a ton @ewels !)

I will continue to monitor the results and do what I can to ease the transition. I can make sure that error messages are clear, but at the end of the day it's up to pipeline developers to update their code. They can always set `NXF_SYNTAX_PARSER=v1` to buy more time.

Looking over the error reports, I think most pipelines can be brought into compliance without too much hassle.

PR #6645, opened 2025-12-10
docs: Update reports page

christopher-hakkaart:
- Styles reports page as high-level explanation and examples
- Update screenshots (now showing new logo)
- Move reference definition lists to the reference section/page

### DIFF
--- a/docs/reports.md
+++ b/docs/reports.md
@@ -2,7 +2,7 @@
 
 # Reports
 
-Nextflow provides several reporting mechanisms to monitor, debug, and analyze pipeline runs:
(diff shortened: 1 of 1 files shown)


bentsherman:
Thanks Chris. I moved some things around, everything looks much better with your refresh.

Some food for thought: we could put a note at the top of the Reports page to the effect of "these reports can only be used to analyze runs after they have completed, use a service like Seqera Platform for real-time monitoring"

christopher-hakkaart:
Thanks Ben - I've made a ticket about adding a note at the top of the page so it's not lost in the PR. I'll need to think about balanced language.

Issue #5924, opened 2025-03-26
Graceful fallback when `array` is used on unsupported executors


itrujnara:
## New feature

Currently, config parsing always fails if the `array` process directive is used in an executor that does not support job arrays (most importantly `local`). It would be reasonable to have Nextflow fall back to running single processes instead and print a warning to `stderr.` If needed, this behavior can be controlled with an environment variable.

## Use case 

Some pipelines have processes that take very little time to run (< 5 minutes). In that case, using job arrays is a reasonable default behavior. However, it is impossible to set directly, as it will fail on executors without job arrays.

## Suggested implementation 

As described above.

bentsherman:
I would just label the processes for which you want to use job arrays and use withLabel to set the array directive for those processes. It's generally better to be specific in that way rather than enabling job arrays across the board

itrujnara:
I'm not sure this fully addresses the issue. If I set the following in the config
```
withLabel: 'very_short' {
    array = 10
}
```
it will still fail if I run the pipeline with the `local` executor. Is there an idiomatic way to have such a pipeline ignore the array directive on a local machine but follow it on a Slurm cluster?

bentsherman:
I assume you have two profiles for local and slurm? In that case you can just put the array config in the slurm profile

itrujnara:
That's the solution I have implemented for the time being. The pipeline is nf-core, so it needs to be as portable as possible. I hoped I could somehow make using arrays the default behavior on grid executors, but it seems it's not possible for the time being.

bentsherman:
I would caution against making job arrays the default across the entire pipeline, because there are cases where it can actually hurt you.

For example, the first step in the pipeline will probably receive a deluge of tasks all at once because it's just loading the inputs. But a later step might receive tasks at a slower rate as some upstream tasks take longer than others. In the latter case, it might not be so important to batch the job submissions if they are already slow, and enabling job arrays might just needlessly delay job submissions while waiting for a full array to submit

That's just my intuition, but in reality it would depend on the actual submit rates vs the capacity of the scheduler. I do wonder whether this issue would come up in practice or not

anoronh4:
@bentsherman regarding the usage of array for a whole pipeline, I’ve noticed that when using the array directive, tasks can stay pending for a long time if the array doesn’t fill up or indefinitely if a poison pill isn’t sent (e.g. with watchPath). It may not matter if array is set for specific processes or for the whole pipeline.

[...]

PR #6647, opened 2025-12-10
Enhance TaskArrayCollector to support configurable timeout-based array submission

### LINKED ISSUE #6465: Enhance TaskArrayCollector to support configurable timeout-based array submission
Currently, TaskArrayCollector submits a job array only when the array reaches its configured size. In pipelines with slower task generation or variable workloads, small batches may remain in memory indefinitely if the array never fills, delaying execution.  This is especially a problem in pipelines with hanging channels, like channels made by watchPath.  We have a few pipelines that run forever, watching a samplesheet for new entries, and processing them as they come in.  We would like to use array jobs to reduce strain on the slurm scheduler as much as possible as this has been a problem on our cluster.

Below is a demonstration of the issue:

```
#!/usr/bin/env nextflow
params.array_size = 10
params.sleep_time = 30
params.outdir = "results"
params.input_dir = "input_files"
params.partition = 'testpartition'

process DUMMY_TASK {
    executor 'slurm'
    queue 'testpartition'
    array params.array_size
    cpus 1
    memory '1 GB'
    time '5m'
    debug true
    tag "task_${task.index}"
    
    input:
    path input_file
    
    output:
    path "${input_file.baseName}_output.txt"
    
    script:
    """
    echo "SLURM_ARRAY_JOB_ID: \${SLURM_ARRAY_JOB_ID:-'not_set'}" >> ${input_file.baseName}_output.txt
    # Simulate processing
    sleep ${params.sleep_time}
    """
}

workflow {
    input_files = Channel.watchPath("${params.input_dir}/*.txt")

    DUMMY_TASK(input_files)

    DUMMY_TASK.out.view { file ->
        "[${new Date().format('HH:mm:ss')}] Processed: ${file.name}"
    }
} 

```

Then run 

```for i in {1..16};do echo "Input data for task $i" > input_files/task_${i}.txt && echo "Created input_files/task_${i}.txt"; sleep .5;done```

In the same directory.   

Currently nextflow will hang at 

```[be/e4bc7e] process > DUMMY_TASK (task_15) [100%] 10 of 10```

There should be 16 jobs as we created 16 files. It is waiting for 4 more tasks to reach the array size of 10 before it submits the array job. If these 4 never come, then these 6 tasks will never get submitted as jobs.

My proposal is to create a maximum amount of time to wait for more jobs before submitting the array of jobs. Then the pipeline would work like this:

The first 10 jobs are submitted as an array job as expected. The next 6 jobs are added into the task array. Then, the time specified by executorSubmitTimeout elapses. At this point the remaining 6 jobs are put into a job array and submitted to the cluster.

My WIP fork currently has this implementation: 
https://github.com/johnoooh/nextflow/blob/feature/slurm_array_timer/modules/nextflow/src/main/groovy/nextflow/processor/TaskArrayCollector.groovy
It still needs a few changes.  The parameter needs to be added to the config.  

I know this is a pretty specific issue, but I don't think it's an unreasonable addition.  Let me know if you foresee any issues with this approach?

This discussion was started here #5924

johnoooh:
Currently, TaskArrayCollector submits a job array only when the array reaches its configured size. In pipelines with slower task generation or variable workloads, small batches may remain in memory indefinitely if the array never fills, delaying execution. This is especially a problem in pipelines with hanging channels, like channels made by watchPath. We have a few pipelines that run forever, watching a samplesheet for new entries, and processing them as they come in. We would like to use array jobs to reduce strain on the slurm scheduler as much as possible as this has been a problem on our cluster.

My proposal is to create a maximum amount of time to wait for more jobs before submitting the array of jobs. Then the pipeline would work like this:

The first 10 jobs are submitted as an array job as expected. The next 6 jobs are added into the task array. Then, the time specified by executorSubmitTimeout elapses. At this point the remaining 6 jobs are put into a job array and submitted to the cluster.

Discussion of this feature was originally started here:
https://github.com/nextflow-io/nextflow/issues/5924

Then I further described the issue here: https://github.com/nextflow-io/nextflow/issues/6465

I'm currently having an issue with java.util.ConcurrentModificationException when running hundreds of jobs.


bentsherman:
@johnoooh I went ahead and cleaned up the PR to get it ready for merging. I also pushed a temporary fix that might fix the error you are getting, let me know if it works

PR #6649, opened 2025-12-11
Refactor Google Batch getExitCode to imperative style

pditommaso:
## Summary

- Refactored `getExitCode()` method in `GoogleBatchTaskHandler` from functional stream-based style to imperative for-loop to fix IntelliJ syntax errors with lambda expressions in Groovy
- Added `StatusEvent` import to support the refactored code
- Added comprehensive unit test using Spock data-driven `where` clause covering 8 scenarios
- Added detailed inline comments explaining the logic flow

### GoogleBatchTaskHandler.groovy
- Replaced stream/filter/max/map chain with explicit for-loop iteration
- Added comments explaining:
  - Why events are iterated (not guaranteed chronological order)
  - How the latest event is found (by comparing timestamps)
  - Fallback behavior when no exit code found from API

### GoogleBatchTaskHandlerTest.groovy
- Added helper methods `makeStatusEventWithTime()` and `makeTaskStatusWithEvents()`
- Added data-driven test `should get exit code from latest task execution event` with 8 test cases:
  - null task status - fallback to file
  - empty events list - fallback to file
  - single event with exit code - returns exit code
  - single event with non-zero exit code - returns exit code
  - event without task execution - fallback to file
  - multiple events - returns latest by timestamp
  - multiple events out of order - correctly finds latest
  - mixed events (with/without execution) - filters correctly

## Test plan

- [x] Unit tests pass: `./gradlew :plugins:nf-google:test --tests GoogleBatchTaskHandlerTest`
- [x] New test covers all edge cases for exit code retrieval

### DIFF
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
@@ -47,6 +47,7 @@ import nextflow.exception.ProcessUnrecoverableException
 import nextflow.executor.BashWrapperBuilder
 import nextflow.executor.res.DiskResource
 import nextflow.fusion.FusionAwareTask
+import nextflow.fusion.FusionConfig
 import nextflow.fusion.FusionScriptLauncher
 import nextflow.processor.TaskArrayRun
 import nextflow.processor.TaskConfig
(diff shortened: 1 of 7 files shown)


pditommaso:
This was rewritten because intellij was reporting (false positive) compilation errors, still annoying tho.

bentsherman:
Do the errors prevent you from building?

pditommaso:
obviously no, tests were passing

bentsherman:
I wonder if the errors go away by replacing the lambdas with closures. Even though it's not desirable, I think Groovy is compiling lambdas the same way as closures anyway, so you might as well use the closure if it removes the error

pditommaso:
Likely yes, however it's not worth spending more time on this (and i also find a bit more readable the imperative approach)

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -586,17 +587,31 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
...
+                    final long eventTime = ev.getEventTime().getSeconds()
+                    if( eventTime > latestTime ) {
+                        latestTime = eventTime
+                        latestEvent = ev
+                    }
+                }
+            }
+            // Return the exit code from the most recent task execution event
+            if( latestEvent != null ) {
+                return latestEvent.getTaskExecution().getExitCode()
```
jorgee: There is a corner case where if there is a task execution event but with exit code is null. In this case, it will not fallback to readExitFile. Not sure if it is a possible case, but better to protect.
```suggestion
            if( latestEvent?.getTaskExecution()?.getExitCode() != null ) {
                return latestEvent.getTaskExecution().getExitCode()
```

pditommaso: Nice

PR #6650, opened 2025-12-11
Module system

pditommaso:
## Summary

Full implementation of the Nextflow Module System, enabling first-class support for installing, managing, publishing, and running modules from a centralized registry.

### CLI commands (`nextflow module <subcommand>`)

- **`module install`** — Download and install a module from the registry to the local `modules/` directory with checksum validation and integrity tracking. Supports `-version` and `-force` flags.
- **`module run`** — Execute a module directly from the registry without pre-installing. Auto-installs if needed, then delegates to the standard `run` command.
- **`module list`** — List all locally installed modules with version and integrity status (`OK` / `MODIFIED` / `LOCAL` / `CORRUPTED`). Supports table or JSON output.
- **`module remove`** — Remove installed modules with integrity checks to prevent accidental data loss. Supports `-force` and `-keep-files` flags.
- **`module search`** — Query the registry for modules matching a search term. Returns formatted results or JSON with metadata and relevance scores.
- **`module info`** — Display detailed module metadata including description, authors, tools, input/output parameter schema, and usage command templates. Smart parameter inference for nf-core modules.
- **`module publish`** — Publish a local module to the registry. Validates structure (requires `main.nf`, `meta.yml`, `README.md`), enforces size limits, bundles as tar.gz, computes checksum, and uploads via authenticated API. Supports `-dry-run`.

### Registry client (`ModuleRegistryClient`)

REST API client with built-in retry logic and multi-registry fallback:
- Fetch module metadata and specific release details
- Download module bundles with SHA-256 integrity validation
- Search modules by keyword
- Publish module bundles (authenticated via `NXF_REGISTRY_TOKEN` or config)

### Local storage management (`ModuleStorage`)

File-based storage in `modules/scope/name/` with `.module-info` marker files:
- Install/extract downloaded bundles with checksum tracking
- Detect local modifications via integrity checking
- Create tar.gz bundles for publishing (excluding internal metadata)
- Atomic operations with cleanup on failure

### Module resolution (`ModuleResolver`)

Coordinates between registry and local storage:
- Resolve module references to local paths with auto-install
- Version validation and auto-upgrade on mismatch
- Integrity warnings for locally modified modules

### DSL integration

New syntax for remote module inclusion:
```nextflow
include { myWorkflow } from 'scope/name'
include { myProcess } from 'scope/name@1.2.0'
```
Pluggable resolution via `RemoteModuleResolver` SPI, allowing third-party registries.

### Configuration (`registry` scope)

```groovy
registry {
    url = 'https://registry.nextflow.io/api'
    apiKey = '...'  // or NXF_REGISTRY_TOKEN env var
}
```
Supports multiple registries with fallback.

### Data models

- **`ModuleReference`** — Parses and validates `scope/name` format
- **`ModuleSpec`** — YAML parser for `meta.yml` with validation (name, version, description, license)
- **`InstalledModule`** — Tracks installed state, version, checksum, and integrity status
- **`ModuleChecksum`** — SHA-256 directory checksums for integrity verification

### Security

- Archive integrity validation and zip-slip prevention
- Bundle size limits (1MB)
- Checksum verification on download

### Documentation and design

- ADR documenting the full specification (`adr/20251114-module-system.md`)
- CLI reference docs (`docs/cli.md`, `docs/reference/cli.md`)
- Module user guide (`docs/module.md`)
- Implementation specs and registry API contract (`specs/251117-module-system/`)

### Testing

Comprehensive Spock test coverage for all components (~4,000 lines of tests):
- All 7 CLI subcommands
- Registry client, storage, resolver, checksum, spec parsing
- Module reference validation and include definition integration

### DIFF
--- a/adr/20251114-module-system.md
+++ b/adr/20251114-module-system.md
@@ -1,7 +1,7 @@
 # Module System for Nextflow
 
 - Authors: Paolo Di Tommaso
-- Status: draft
+- Status: approved
 - Date: 2025-01-06
 - Tags: modules, dsl, registry, versioning, architecture
 - Version: 2.7


bentsherman:
The module spec schema will be defined in the `nextflow-io/schemas` repo. Here is the PR for it: https://github.com/nextflow-io/schemas/pull/10

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1157 @@
...
+- Date: 2025-01-06
+- Tags: modules, dsl, registry, versioning, architecture
+- Version: 2.2
+
+## Updates
+
+### Version 2.2 (2025-01-06)
+- **Structured tool arguments**: Added `args` property to `tools` section for type-safe argument configuration
+- **New implicit variables**: `tools.<toolname>.args.<argname>` returns formatted flag+value; `tools.<toolname>.args` returns all args concatenated
+- **Deprecation**: `ext.args`, `ext.args2`, `ext.args3` pattern deprecated in favor of structured tool arguments
```
bentsherman: Other common ext properties include `ext.prefix` and `ext.suffix`, for example: https://github.com/nf-core/modules/blob/master/modules/nf-core/bedtools/intersect/main.nf

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1157 @@
...
+### 1. Remote Module Inclusion
+
+**DSL Syntax**:
+```groovy
+// Import from registry (scoped module name, detected by @scope prefix)
+include { BWA_ALIGN } from '@nf-core/bwa-align'
+
+// Existing file-based includes remain supported
+include { MY_PROCESS } from './modules/my-process.nf'
+```
```
bentsherman, line 73: This should be out-of-scope for the first iteration. Let's focus on the `modules` command and module spec generation

jorgee: If I have understood correctly, at parse time `@nf-core/bwa-align `will be translated to `modules/@scope/name@version/main.nf`. 
If it is not supported in first iteration, users must write the include statements providing the module path taking into account the version and keep the coherence of these include statements with the versions of `nextflow.config`. 
I think it is also required for defining modules importing other modules. If this is not supported, modules must be written using include with paths.

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1157 @@
...
+
+**Module Naming**: NPM-style scoped packages `@scope/name` (e.g., `@nf-core/salmon`, `@myorg/custom`). Unscoped names (eg. local paths) supported for legacy compatibility. No nested paths with the module are allowed - each module must have a `main.nf` as the entry point.
+
+**Version Resolution**: Module versions pinned in `nextflow.config`. If not specified, use the latest available locally in `modules/` directory, or downloaded and cached in the `modules/` directory.
+
+**Resolution Order**:
+1. Check `nextflow.config` for pinned version
+2. Check local `modules/@scope/name/` for any cached version
```
jorgee: Are module versions stored in subdirectories of name (`modules/@scope/name/@version`) or in the folder name (`modules/@scope/name@version`). 

Here seems a subdirectory but in other parts of the document version is included in the folder name.
```suggestion
2. Check local `modules/@scope/name@*` for any cached version
```

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1157 @@
...
+3. Query registry for latest version if not found
+4. Warn if transitive dependencies are not pinned
+
+**Resolution Timing**: Modules resolved at workflow parse time (after plugin resolution at startup).
+
+**Local Storage**: Downloaded modules stored in `modules/@scope/name@version/` directory in project root (not global cache). Each module must contain a `main.nf` file as the required entry point. It is intended that module source code will be committed to the pipeline git repository.
```
jorgee: There could be cases where the version in cached modules does not match the `nextflow.config` (e.g users upgrade the module in config but do not commit the module's files)
What should it do in these cases? warn the user and download the new version or fail?

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1157 @@
...
+
+**Arguments**:
+- `[scope/name]`: Optional module identifier. If omitted, installs all modules from config
+
+**Options**:
+- `-version <ver>`: Install a specific version (default: latest)
+- `-force`: Re-download even if already installed locally
+
+**Behavior**:
+1. Resolves the module version from `nextflow.config` or queries registry for latest
```
jorgee:
```suggestion
1. If `-version' not specified, resolves the module version from `nextflow.config` or queries registry for latest
```

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1157 @@
...
+2. Downloads the module archive from the registry
+3. Extracts to `modules/@scope/name@version/` directory
+4. Recursively installs transitive dependencies declared in `meta.yaml`
+5. Updates `nextflow.config` if installing a new module not already configured
```
jorgee: Should it also include the installed transitive dependencies in `nexflow.config`, or just the module requested to install.

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1157 @@
...
+Discussion/request goes back to at least 2019, see GitHub issues [#1376](https://github.com/nextflow-io/nextflow/issues/1376), [#1463](https://github.com/nextflow-io/nextflow/issues/1463) and [#4122](https://github.com/nextflow-io/nextflow/issues/4112).
+
+## Decision
+
+Implement a module system with four core capabilities:
+
+1. **Remote module inclusion** via registry
+2. **Semantic versioning** with dependency resolution
+3. **Unified Nextflow Registry** (rebrand existing plugin registry)
+4. **First-class CLI support** (pull, push, search, run)
```
jorgee:
```suggestion
4. **First-class CLI support** (install, publish, search, list, remove, freeze, run)
```

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1157 @@
...
+        └── custom-process@2.0.0/
+            ├── meta.yaml
+            └── main.nf             # Required entry point
+```
+
+## Implementation Strategy
+
+**Phase 1**: Module manifest schema, local module loading, validation tools
+
+**Phase 2**: Extend plugin registry for modules, implement caching, add `pull` and `search` commands
```
jorgee:
```suggestion
**Phase 2**: Extend plugin registry for modules, implement caching, add `install` and `search` commands
```

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1157 @@
...
+
+**Phase 3**: Extend DSL parser for `from module` syntax, implement dependency resolution from meta.yaml
+
+**Phase 4**: Implement `push` command with authentication and `run` command
```
jorgee:
```suggestion
**Phase 4**: Implement `publish` command with authentication and `run` command
```

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1157 @@
...
+
+**Dependency Resolution Flow**:
+1. Parse `include` statements → extract module names (e.g., `@nf-core/bwa-align`)
+2. For each module:
+   a. Check `nextflow.config` modules section for pinned version
+   b. If not pinned: check local `modules/@scope/name/` for any cached version (use latest local)
+   c. If not found locally: query registry for latest version
+   d. Warn if module not pinned in config (especially transitive dependencies)
+3. Download missing modules to `modules/@scope/name@version/`
+4. Read module's `meta.yaml` → resolve transitive dependencies recursively
```
jorgee: Modules meta includes required Nextflow version and plugins. At this point, I think it should validate the required nextflow version and the required plugins (loading them if missing)

adr/20251114-module-system.md
bentsherman: Yes, likely module versioning would also be out of scope. The primary motivation initially was to allow remote module execution:
```bash
nextflow module run @nf-core/bwa-align
```

This can be done without remote module inclusion. We can probably still implement module versioning just for installing and executing modules

adr/20251114-module-system.md
pditommaso: Yes, the version should contribute to the directory path. Instead think we are not going to use `@` as scope prefix in the directory structure (only as conventional prefix in the nextflow syntax)

adr/20251114-module-system.md
pditommaso: Likely it should be added a command to detect/solve these kind of inconsistency but I would avoid warning too verbosely

adr/20251114-module-system.md
pditommaso: There's no need to delay module versioning, there's already an initial implementation for the registry backend

adr/20251114-module-system.md
pditommaso: Updated 5b1d4e503

pditommaso:
## Summary of Recent Changes

### Major Simplification: Module Storage and Verification Model

The module system has been significantly simplified by making the **registry the source of truth** for checksums:

**Config Syntax Simplified** - Removed checksums from config, version-only format:
```groovy
modules {
    '@nf-core/salmon' = '1.1.0'
    '@nf-core/bwa-align' = '1.2.0'
}
```

**Local Storage Structure** - Single version per module (no version in path), with `.checksum` file for integrity verification:
```
modules/
└── @nf-core/
    └── bwa-align/
        ├── .checksum           # Cached registry checksum for fast validation
        ├── meta.yaml
        └── main.nf
```

**Verification Flow**:
1. Install: Download module → store `.checksum` from registry's `X-Checksum` header
2. On run: Compute local checksum → compare with `.checksum`
3. Match → proceed (no registry call needed)
4. Mismatch → report warning (user must explicitly reinstall)

**`freeze` Command Simplified** - Only pins versions, no checksums in config

### Other Updates

- **Module API** updated to match actual registry implementation (`/api/modules/{name}` pattern)
- **Tool Arguments CLI** support added: `--tools:<tool>:<arg> <value>` syntax
- **Open Questions** section added for pending design decisions
- **Terminology** standardized: "Nextflow registry" instead of "plugin registry"
- **Various fixes**: consistent `sha256:` prefix, `license` spelling, `authors` plural form, path formats

adr/20251114-module-system.md
pditommaso: Thinking more about this, I realised that's useless having multiple local versions. It should only cloned the required version. Therefore the path should be `modules/@scope/name`

jorgee: Agree. 
Moreover, related to @Ben [comment](https://github.com/nextflow-io/nextflow/pull/6650#discussion_r2666575811), it will also help to decouple the DSL part implementation with the rest. For testing, we could include in the code as `include { BWA_ALIGN } from './modules/nf-core/bwa-align'` instead of `include { BWA_ALIGN } from '@nf-core/bwa-align'` and it should work

pditommaso:
## Version 2.4 Changes

Updated the ADR to **remove transitive dependency resolution** from the initial implementation scope. This simplifies the module system by requiring explicit dependency declarations.

### Summary of Changes

- **Removed `freeze` command** from CLI - no longer needed without transitive dependency management
- **Simplified install behavior** - removed automatic recursive installation of transitive dependencies
- **Updated `remove` command** - removed orphaned transitive dependency cleanup
- **Simplified dependency resolution flow** - modules are resolved directly from `nextflow.config` declarations only
- **Updated rationale and consequences sections** to reflect the simplified model

### Rationale

Transitive dependency resolution adds significant complexity and will not be implemented in this initial stage. Each module's dependencies must be explicitly declared in `nextflow.config`, giving users full control and visibility over what modules are installed.

This change also affects the spec at `specs/251117-module-system/spec.md` which was updated to remove:
- Transitive dependency acceptance scenarios
- FR-010, FR-011 (transitive dependency requirements)
- `freeze` command from FR-019
- SC-008 (transitive dependencies success criteria)

adr/20251114-module-system.md
bentsherman: I still think remote module inclusion should be excluded for now. This way we don't have to update the language / runtime to deal with remote includes

I have attempted to remove these bits in #6723

The module CLI commands all work the same way. The `module run` command can use a global cache to download and execute remote modules on-the-fly

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+- **MINOR**: New processes, backward-compatible enhancements
+- **PATCH**: Bug fixes, documentation updates
+
+**Workflow Configuration** (`nextflow.config`):
+```groovy
+// Module versions (exact versions only, no ranges)
+modules {
+    '@nf-core/salmon' = '1.1.0'
+    '@nf-core/bwa-align' = '1.2.0'
+}
```
bentsherman: The problem with specifying modules in the config is that the user can override config at runtime

I have altered the ADR to store these in `nextflow_spec.json` instead in #6723

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+4. **First-class CLI support** (install, publish, search, list, remove, run)
+
+## Core Capabilities
+
+### 1. Remote Module Inclusion
+
+**DSL Syntax**:
+```groovy
+// Import from registry (scoped module name, detected by @scope prefix)
+include { BWA_ALIGN } from '@nf-core/bwa-align'
```
ewels: Could we include new syntax here to handle versions? I think that would be preferable to doing it in the config.

```nextflow
// Import from registry (scoped module name, detected by @scope prefix)
include { BWA_ALIGN } from '@nf-core/bwa-align', version: "1.1.2"
```

If it's in the config, users of the pipeline could override it which would break reproducibility. It's also two intrinsically linked things but specified in different files, difficult to remember and coordinate.

Doing it as separate to the module name like this means that paths aren't affected by version bumps, which I think was part of your concern @pditommaso ?

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+  modules:                        # Required modules (version constraints)
+    - nf-core/samtools/view@>=1.0.0,<2.0.0
+    - nf-core/samtools/sort@>=2.1.0,<2.2.0
+```
+
+**Version Constraints** (unified `name@constraint` syntax):
+- `name`: Any version (latest)
+- `name@1.2.3`: Exact version
+- `name@>=1.2.3`: Greater or equal
+- `name@>=1.2.3,<2.0.0`: Range (comma-separated)
```
ewels, line 136: Minor: I don't love the use of `@` here, as it's used to mean something different elsewhere (scope).

Given that package names presumably can't include special characters, is it actually needed at all?

```suggestion
- `name=1.2.3`: Exact version
- `name>=1.2.3`: Greater or equal
- `name>=1.2.3,<2.0.0`: Range (comma-separated)
```

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+4. Apply resolution rules (see below)
+
+**Resolution Rules**:
+
+| Local State | Declared Version | Action |
+|-------------|------------------|--------|
+| Missing | Any | Download declared version (or latest if not declared) |
+| Exists, checksum valid | Same as declared | Use local module |
+| Exists, checksum valid | Different from declared | **Replace** local with declared version |
+| Exists, checksum mismatch | Same as declared | **Warn**: module was locally modified, do not override |
```
ewels: If this is also at runtime, I think it's important to keep this as a warning (as written) and not an error. The person running the pipeline may not be able to edit the source code to fix themselves, so it's good to not block execution.

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+```
+GET  /api/modules?query=<text>               # Search modules (semantic search)
+GET  /api/modules/{name}                     # Get module + latest release
+GET  /api/modules/{name}/releases            # List all releases
+GET  /api/modules/{name}/{version}           # Get specific release
+GET  /api/modules/{name}/{version}/download  # Download module bundle
+POST /api/modules/{name}                     # Publish module version (authenticated)
+```
+
+Note: The `{name}` parameter includes the namespace prefix (e.g., "nf-core/fastqc").
```
ewels, line 187: Will slashes in the module name cause problems here? eg. a module called `@foo/bar/baz` ?

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+4. If local module was modified (checksum mismatch): warns and aborts unless `-force` is used
+5. Downloads the module archive from the registry
+6. Extracts to `modules/@scope/name/` directory
+7. Stores `.checksum` file from registry's X-Checksum response header
+8. Updates `nextflow.config` if installing a new module not already configured
+
+**Example**:
+```bash
+nextflow module install                      # Install all from config
+nextflow module install nf-core/bwa-align    # Install specific module (latest)
```
ewels, line 305: This downloads the files, but presumably doesn't do anything with imports, so they won't be used - is that right? I wonder how much value this brings.

In nf-core we give this option partly because `nf-core module install` gives a fuzzy search so it's a quick way to find a module that you know exists. Then it prompts you post-install with the include text to copy + paste into your nextflow script.

If this command doesn't have the fuzzy interactive prompt, I'm not sure that specifying any arguments at all really brings much value. Might be easier to just have the install from config (behaviour more like `npm install` then).

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+- Module name (`@scope/name`)
+- Configured version (from `nextflow.config`)
+- Installed version (from `modules/` directory)
+- Latest available version (from registry)
+- Status indicator (up-to-date, outdated, missing, not configured)
+
+**Example**:
+```bash
+nextflow module list
+nextflow module list -outdated
```
ewels, line 328: Can we have `nextflow module update` to go with this, for bulk updating modules? (Both module files and config / imports).

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+```bash
+nextflow module remove nf-core/bwa-align
+nextflow module remove myorg/custom -keep-files
+```
+
+---
+
+#### `nextflow module publish scope/name`
+
+Publish a module to the Nextflow registry, making it available for others to install. Requires authentication via API key and appropriate permissions for the target scope.
```
ewels, line 356: Is the idea that this is run from a within a module directory? (be it in a pipeline or otherwise)

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+  - nf-core community
+license: MIT
+
+requires:
+  nextflow: ">=24.04.0"
+  plugins:
+    - nf-amazon@2.0.0
+  modules:
+    - nf-core/samtools/view@>=1.0.0,<2.0.0
+    - nf-core/samtools/sort@>=2.1.0,<2.2.0
```
ewels: I think we said that we removed transitive deps, so no longer needed?

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+```
+
+**Module Spec extension** (`meta.yaml`):
+```yaml
+name: nf-core/bwa-align
+version: 1.2.4              # This module's version
+description: Align reads using BWA-MEM
+authors:
+  - nf-core community
+license: MIT
```
ewels, line 406: Need to be clear about module license vs. tool license here. Maybe worth naming explicitly?

```suggestion
module_license: MIT
```

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+```yaml
+tools:
+  - bwa:
+      description: BWA aligner
+      homepage: http://bio-bwa.sourceforge.net/
+      args:
+        K:
+          flag: "-K"
+          type: integer
+          description: "Process INT input bases in each batch"
```
ewels: If possible, I like having an additional optional `help` key for longer descriptive text + usage tipes etc (vs. `description` which should be brief).

I think `description` should be a required key.

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+3. **Module version configuration**: Should pipeline module versions be specified in `nextflow.config` or in a dedicated pipeline spec file (e.g., `pipeline.yaml`)?
+
+---
+
+## Appendix A: Module Metadata Schema Specification
+
+This appendix defines the JSON schema for module `meta.yaml` files. The schema maintains backward compatibility with existing nf-core module metadata patterns while supporting the new Nextflow module system features.
+
+**Schema File:** [module-spec-schema.json](module-spec-schema.json)
+**Published URL:** `https://registry.nextflow.io/schemas/module-spec/v1.0.0`
```
ewels, line 554: This should be kept in https://github.com/nextflow-io/schemas alongside the other schemas that we're building.

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+  nextflow: ">=24.04.0"
+  plugins:
+    - nf-amazon@2.0.0
+    - nf-wave@>=1.5.0
+  modules:
+    - nf-core/fastqc@>=1.0.0
+    - nf-core/samtools/sort@>=2.1.0,<3.0.0
+    - bwa/mem
+  workflows:
+    - nf-core/fastq-align-bwa@1.0.0
```
ewels: Not needed if modules only (no subworkflows) and no transitive deps

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+```yaml
+requires:
+  workflows:
+    - nf-core/fastq-align-bwa@1.0.0       # registry workflow
+    - my-local-workflow                    # local workflow
+```
+
+**Resolution:**
+1. The resolver looks up dependencies locally first, then in configured registries
+2. Pinned versions are recorded in `nextflow.config` for reproducibility
```
ewels: More transitive deps stuff

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+  - ch_index:
+      description: BWA index files
+      type: file
+
+output:
+  - bam:
+      description: Aligned BAM files
+  - versions:
+      description: Software versions
+```
```
ewels: transitive

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+version: "1.0.0"                    # Added version
+description: Align reads using BWA-MEM
+keywords:
+  - alignment
+  - bwa
+license: MIT                        # Added module license
+requires:                           # Added requirements
+  nextflow: ">=24.04.0"
+  modules:                          # Added module dependencies (if any)
+    - nf-core/samtools/sort@>=1.0.0
```
ewels: transitive

adr/20251114-module-system.md
pditommaso: We have done quite a good work to distill the requirements to core needs and use cases. I'm not getting why it should stripped the support remote module. Without that it would become just a re-implementation of existing nf-core module tool. Remote inclusion is an essential part of this feature.

adr/20251114-module-system.md
pditommaso: Think it has been convert with the current version.

adr/20251114-module-system.md
pditommaso: We have stripped transitive deps

adr/20251114-module-system.md
pditommaso: You were suggesting to amend this?

adr/20251114-module-system.md
pditommaso: I agree config is not the best place, but it's not the same that's happening for plugins? My take is the we should managed in the same manner

adr/20251114-module-system.md
pditommaso: This was excluded for the same reason we do not add the version in the inclusion for plugins. You will soon have many different versions across different module inclusions opening a dependencies version resolution hell. 

Likely this is one of the reasons why it's not commonly done in other programming lang.

adr/20251114-module-system.md
pditommaso: Indeed, it's a warning

adr/20251114-module-system.md
pditommaso: It's expected to be encoded.

adr/20251114-module-system.md
pditommaso: Yep

adr/20251114-module-system.md
bentsherman: Plugins need to be able to be customized at runtime, modules do not. That is the difference

adr/20251114-module-system.md
bentsherman: I think if we exclude remote module inclusion for now then there is no need for `nextflow module install`, because it is intended to be used in a setting where module code is not baked into the repo

But what do you mean about "not doing anything with imports"?

adr/20251114-module-system.md
bentsherman: I think so. I think this means that monorepos like `nf-core/modules` will continue to be the source of truth for developing modules, but not for installing them in a pipeline

adr/20251114-module-system.md
bentsherman: We could specify the tool license under the `tools` section

adr/20251114-module-system.md
ewels: Fair point, and I guess other programming languages do define versions in a top-level config file.

I wonder if we make a new top-level file for this then, to avoid the issues with defining versions in the config.. 🤔  (I suppose nf-core already has `modules.json` that does this).

adr/20251114-module-system.md
ewels: Also means that if a developer is making one pipeline with a set of modules within, they can edit the modules in-place, then `cd` into those directories and publish updates, without maintaining a separate repo just for modules.

adr/20251114-module-system.md
ewels: Yup, would still be good to improve the terminology at this top level one as well though I think to discriminate.

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+
+**Limitations:**
+- No documentation of available arguments
+- No validation or type checking
+- Unclear which `ext.argsN` maps to which tool
+- No IDE autocompletion support
+
+### New Pattern: Structured Tool Arguments
+
+Modules declare available arguments in `meta.yaml` under each tool's `args` property:
```
ewels: I think it would be good to be explicit on this point. I have described this to a few people now and almost everyone has had the same question / fear, that maintainers will be forced to maintain 1:1 parity with the upstream tool.

```suggestion
Modules declare available arguments in `meta.yaml` under each tool's `args` property.
This list does _not_ need to be exhaustive. It should include any arguments known to be used by pipelines or that could be expected to be used by users. However, arguments can still be specified in the config even if not defined in this file, so absence does not prevent use.
```

adr/20251114-module-system.md
bentsherman: Yeah, but I guess if you have a pipeline with some local modules, you don't really need to publish them unless you intend to share them.

adr/20251114-module-system.md
bentsherman: But you could argue the same for many of the top-level fields, like name and description. It seems inconsistent to apply it only to the license

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1184 @@
...
+   a. Check `nextflow.config` modules section for declared version
+   b. Check local `modules/@scope/name/` exists
+   c. Verify local module integrity against `.checksum` file
+   d. Apply resolution rules:
+      - Missing → download declared version from registry
+      - Exists, checksum valid, same version → use local
+      - Exists, checksum valid, different version → replace with declared version
+      - Exists, checksum mismatch → warn and do NOT override (local changes detected)
+3. On download: store module to `modules/@scope/name/` with `.checksum` file
+4. Parse module's `main.nf` file → make processes available
```
jorgee: Modules have Nextflow version and plugin requirements, they must be taken into consideration in this flow before parsing the module.
```suggestion
4. Read `meta.yaml` file:
    a. Validates Nextflow requirement → Fail if not fulfilled
    b. Load Pluign requirements if not exist.
5. Parse module's `main.nf` file → make processes available```

File: adr/20251114-module-system.md (hunk 10 above)
Comment (jorgee) [reply adr/20251114-module-system.md]: Yes. The code of the comment is outdated, so I have created the suggestion in the current code.
https://github.com/nextflow-io/nextflow/pull/6650/changes#r2711786212

File: adr/20251114-module-system.md (hunk 23 above)
Comment (pditommaso) [reply adr/20251114-module-system.md]: That's true for modules, but considering this spec could cover also sub-workflows then it makes sense because it can be dependencies both with modules and sub-workflows. clarifying would solve?

File: adr/20251114-module-system.md (hunk 27 above)
Comment (pditommaso) [reply adr/20251114-module-system.md]: Make sense

File: adr/20251114-module-system.md (hunk 12 above)
Comment (pditommaso) [reply adr/20251114-module-system.md]: Then, let's consider version pinning in the inclusion resolved, and move top level file into this thread https://github.com/nextflow-io/nextflow/pull/6650#discussion_r2696231247

File: adr/20251114-module-system.md (hunk 11 above)
Comment (pditommaso) [reply adr/20251114-module-system.md]: Think you raised a point, however I believe compared to other lang with nextflow is desirable to being able to control module version at config time. Even more with [interface modules](https://github.com/nextflow-io/nextflow/pull/6737) we are going toward a system in which modules (and therefore versions) can be fully decided at runtime. 

This makes me thing nextflow.config is the right place to keep them.

File: adr/20251114-module-system.md (hunk 27 above)
Comment (pditommaso) [reply adr/20251114-module-system.md]: Apologies Phil, I removed `However, arguments can still be specified in the config even if not defined in this file, so absence does not prevent use.`. I believe we need go beyond ext.args in the config. it's really a bad back that makes it impossible to have visibility/control over the module parametrisation

File: adr/20251114-module-system.md (hunk 13 above)
Comment (pditommaso) [reply adr/20251114-module-system.md:136]: This the syntax we are using for plugins (inherited from the [underlying framework](https://pf4j.org/doc/plugins.html)). I think it's preferable to *not* have different notations for plugins and modules

File: adr/20251114-module-system.md (hunk 23 above)
Comment (bentsherman) [reply adr/20251114-module-system.md]: I don't think it's a given that we would use the module spec also for sub-workflows, given that the module spec is currently designed for a single process (e.g. the tools list)

So we should not try to be proactive here since we might not go in this direction

File: adr/20251114-module-system.md (hunk 11 above)
Comment (bentsherman) [reply adr/20251114-module-system.md]: Let's not base the module design on other proposed features that are still speculative. The bottom line is that module versions are currently baked into the pipeline definition and we have not committed to changing that in any way. 

I don't think I have seen a programming language where code dependencies are mixed with runtime configuration, except perhaps some crazy dotnet stuff. So the idea of hot-swapping modules at runtime needs a lot more scrutiny

File: adr/20251114-module-system.md (hunk 13 above)
Comment (bentsherman) [reply adr/20251114-module-system.md:136]: I agree with Paolo here. Maybe we should remove the use of `@` for scope, since I'm not sure that we need it. You could distinguish between local and remote modules by using `./` for local

File: adr/20251114-module-system.md (hunk 27 above)
Comment (bentsherman) [reply adr/20251114-module-system.md]: I think Phil's point is that `tools.<tool>.args` can still be specified in the config, and it can specify args that aren't documented in the module spec. Not sure what is hacky about that. These two things are absolutely necessary if you want anyone to use this feature

File: adr/20251114-module-system.md (hunk 2 above)
Comment (bentsherman) [reply adr/20251114-module-system.md:73]: No, it is not essential. It has nothing to do with the original motivation of remote module execution.

For years we said we wouldn't do remote module includes, and now suddenly it's the most important thing we should be building? I have still not heard a good argument for why we need to do it right now.

File: adr/20251114-module-system.md [hunk 29]
```diff
@@ -0,0 +1,1188 @@
...
+version: 1.2.4              # This module's version
+description: Align reads using BWA-MEM
+authors:
+  - nf-core community
+license: MIT
+
+requires:
+  nextflow: ">=24.04.0"
+  plugins:
+    - nf-amazon@2.0.0
```
bentsherman: Plugin dependencies can also be removed, since modules basically never use plugins.

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1188 @@
...
+- Authentication required for publishing
+- Support for private registries
+
+**Integration with Plugin System**:
+- Modules can declare plugin dependencies in meta.yaml
+- Both plugins and modules query same registry
+- Single authentication system
+- Separate cache locations: `$NXF_HOME/plugins/` (global) vs `modules/` (per-project)
+
+## Tool Arguments Configuration
```
bentsherman: Nit: seems like this section should be before "implementation strategy" and "technical details"

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1188 @@
...
+// Config file
+withName: 'BWA_MEM' {
+    ext.args = "-K 100000000 -Y -B 3 -R ${meta.read_group}"
+    ext.args2 = "--output-fmt cram"
+}
+
+// Module script
+def args = task.ext.args ?: ''
+def args2 = task.ext.args2 ?: ''
+bwa mem $args -t $task.cpus $index $reads | samtools sort $args2 -o out.bam -
```
bentsherman: This section on tool args proposes to replace all `ext` properties, but only addresses a subset of `ext` usage i.e. `ext.args`

For the proposal to be sound, it needs to either (1) only claim to replace `ext.args` or (2) address the other major uses of `ext` (even if they aren't related to tool args)

Here are the other major uses of `ext` based on a review of nf-core/modules:
- `ext.prefix`
- `ext.suffix`
- `ext.when`

And here is an nf-core module that uses all three: [bedtools/intersect](https://github.com/nf-core/modules/blob/master/modules/nf-core/bedtools/intersect/main.nf)

For `ext.when` , the remedy is to use params and if statements (and records to keep it clean)

For `ext.prefix` and `ext.suffix`, it is more tricky because they are often set dynamically with a closure, for example:
```groovy
process {
    withName:'BEDTOOLS_INTERSECT' {
        ext.prefix = { "${meta.my_custom_field}" }
    }
}
```

This use case is extremely hacky and extremely popular :sweat_smile: I wasn't aware of it for a long time because you don't see it in pipeline repos. It is a common way to use custom metadata (like `meta.my_custom_field`) at runtime, without passing it through all of the workflow logic as a param.

@ewels do you know anything else about this use case? I'm not really sure how to handle it if we move away from `ext`. It does reek of "encoding metadata in file paths" which we are trying to avoid, so I wonder if we could just discard it

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1188 @@
...
+```
+
+### Script Usage
+
+In module scripts, access arguments via the `tools` implicit variable:
+
+```groovy
+// tools.bwa.args.K → "-K 100000000"
+// tools.bwa.args.Y → "-Y"
+// tools.bwa.args   → "-K 100000000 -Y"  (all args concatenated)
```
bentsherman: Still not sure about this from a type checking perspective. You have `tools.bwa.args` which is a String, but secretly it's also a Map of key-values for individual args. The latter might not even be needed, since usually the tool args are just passed verbatim to the tool (as in the `bwa | samtools` example below)

bentsherman: I found some nf-core modules where the individual arg key-values might be useful:

- [bcftools/annotate](https://github.com/nf-core/modules/blob/master/modules/nf-core/bcftools/annotate/main.nf)
- [muscle](https://github.com/nf-core/modules/blob/master/modules/nf-core/muscle/main.nf)

Many nf-core modules do it, but these are some of the most extreme examples. It would be good for the ADR to show whether / how examples like these would be improved by using the tool args

adr/20251114-module-system.md
bentsherman: Another approach we discussed today is "process params":

```groovy
process BEDTOOLS_INTERSECT {
    input:
    (meta: Map, intervals1: Path, intervals2: Path, chrom_sizes: Path) : Record

    params:
    args: String = '--foo'
    prefix: String = meta.id
    suffix: String = intervals1.extension

    // ...
}
```

```groovy
process {
    withName: BEDTOOLS_INTERSECT {
        params.args = '--foo-bar'
    }
}
```

This approach would address all uses of `ext` instead of only `ext.args`.

An even lighter approach (no new syntax required) would be to move the ext inputs into the record input, and have the pipeline inject them in a map operator, as described in https://github.com/nf-core/methylseq/pull/593#discussion_r2693592007 .

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1188 @@
...
+
+```yaml
+tools:
+  - bwa:
+      description: BWA aligner
+      homepage: http://bio-bwa.sourceforge.net/
+      args:
+        K:
+          flag: "-K"
+          type: integer
```
bentsherman: Regardless of the ongoing discussion around specifying tool args in the language, I think we should definitely add the ability to document tool args in the module spec. At the very least, it is an optional / progressive enhancement for users and agents.

I only wonder how far we should go in trying to model arguments like typed parameters. CLI arguments can be quite messy, so adding layers of abstraction might only make things worse.

Modeling the type should be okay as long as we keep it simple -- strings, numbers, booleans. Beyond that it's probably better to rely on the description.

Modelling the name is trickier, since you can have a short name, long name, single dash or double dash. Might be best for the args to just be a list instead of a map:

```yaml
      args:
        - name: "--output"
          shortName: "-o"
          type: string
          description: "Output file path"
```

adr/20251114-module-system.md
pditommaso: Marking as solved.

adr/20251114-module-system.md
pditommaso: ok, i've removed all modules and plugins version. Let's focus just on module spec 474eb758

adr/20251114-module-system.md
pditommaso: Solved 474eb758

adr/20251114-module-system.md
pditommaso: I think this can be easily covered by `params` as well. They should be declared in the spec as for the inputs. This could allow to define the description 

```
params:
  - name: "output" 
    description: "Output file path"
```

I feel having both `params` and `args` we are going to introduce too many overlapping concepts

bentsherman: I guess they would be complementary, because you would have one param called `args` and then the tool args spec would give some hints about what you can put in `params.args`

But I agree it could still be confusing even if they are complementary

adr/20251114-module-system.md
```diff
@@ -415,7 +414,7 @@ Everything within the module directory should be uploaded. Module bundle should
 ```
 my-module/
 ├── main.nf      # Required: entry point for module
-├── meta.yaml    # Optional: Module spec (metadata, dependencies, I/O specs)
+├── meta.yaml    # Optional: Module spec (metadata, I/O specs)
```
jorgee: Should it be optional? I think it is the only place where version is provided

pditommaso: Indeed, it should be mandatory for publishing

adr/20251114-module-system.md
pditommaso: Indeed, i've update the docs with `params` and remove tools `args`

adr/20251114-module-system.md
pditommaso: Think params solve this. Do you agree?

pditommaso:
## Version 2.5 Updates (2026-01-23)

### Major Changes

**Module Parameters** - Replaced the structured tool arguments (`tools.<toolname>.args`) approach with a simpler, more general module parameters system:

- Parameters are now defined in `meta.yaml` under a `params` section
- Each parameter has: `name` (required), `type` (optional), `description` (optional), `example` (optional)
- Parameters use standard Nextflow `params` syntax in scripts and CLI (`--<param_name>`)
- Removes the complexity of tool-specific argument mapping

**Example:**
```yaml
params:
  - name: batch_size
    type: integer
    description: "Process INT input bases in each batch"
    example: 100000000

  - name: use_soft_clipping
    type: boolean
    description: "Use soft clipping for supplementary alignments"
```

**Simplified Tools Section** - The `tools` section now only documents tool metadata (description, homepage, license, identifier). The `args` property has been removed.

**Simplified `requires` Block** - Now only contains `nextflow` version constraint. Removed `plugins`, `modules`, and `subworkflows` sub-properties.

**Process Modules Focus** - Spec is now focused on process modules only; sub-workflow references removed.

### Rationale

The module parameters approach is simpler and more aligned with existing Nextflow patterns:
- Uses familiar `params` variable instead of introducing new `tools.*.args` syntax
- Parameters are module-level, not tool-level, giving module authors flexibility
- Cleaner separation between tool documentation and module configuration

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1044 @@
...
+**Version Format**: MAJOR.MINOR.PATCH
+- **MAJOR**: Breaking changes to process signatures, inputs, or outputs
+- **MINOR**: New processes, backward-compatible enhancements
+- **PATCH**: Bug fixes, documentation updates
+
+**Workflow Configuration** (`nextflow.config`):
+```groovy
+// Module versions (exact versions only, no ranges)
+modules {
+    '@nf-core/salmon' = '1.1.0'
```
jorgee: The use of `@` seems a bit unclear. It is used in config, storage, and in the remote module inclusion, but not in commands and the registry API. 

It seems the jCommander does not allow passing a parameter starting with `@`. It is reserved for some expansion stuff. When I use it, jCommander uses the string after `@` as a path to read.

adr/20251114-module-system.md
bentsherman: Yes, at least to replace the `tools.<tools>.args` syntax.

The documenting tool CLI args in the module spec as a hint could still be useful on its own, even if it's not an exhaustive list. But the tool spec can also specify a documentation URL, so maybe users and agents could just use that to find CLI args.

adr/20251114-module-system.md
bentsherman: Yeah, I don't think we actually need the `@` prefix to signify scope. You could just use `nf-core/salmon` to refer to a remote module and `./path/to/local/module` for local modules

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1044 @@
...
+|-------------|------------------|--------|
+| Missing | Any | Download declared version (or latest if not declared) |
+| Exists, checksum valid | Same as declared | Use local module |
+| Exists, checksum valid | Different from declared | **Replace** local with declared version |
+| Exists, checksum mismatch | Same as declared | **Warn**: module was locally modified, do not override |
+| Exists, checksum mismatch | Different from declared | **Warn**: locally modified, will NOT replace unless `-force` is used |
+
+**Key Behaviors**:
+- **Version change**: When the declared version differs from the installed version (and local is unmodified), the local module is automatically replaced with the declared version
+- **Local modification**: When the local module content was manually changed (checksum mismatch with `.checksum`), Nextflow warns and does NOT override to prevent accidental loss of local changes
```
jorgee: I think this checksum validation could be tricky. In the registry, the checksum is calculated in the tar.gz bundle file. The local download is the uncompressed tar.gz bundle in a folder. 
To make both checksums comparable, we should control how bundles are created to apply the checksum on a reproducible tar.gz package. I think the order on how file are packed in the tar and some headers could change even if the content is not modified.

jorgee: I think we cannot rely in the way that a user created a the tar.gz bundle when publishing the package. It would be better to use the registry checksum just to validate the integrity in the download and store in the `.checksum` the hash of the files content once uncompressed and a in a sorted order

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1044 @@
...
+- `--<param_name> <value>`: Configure module parameters (validated against meta.yaml schema)
+- All standard `nextflow run` options (e.g., `-profile`, `-work-dir`, `-resume`, etc.)
+
+**Behavior**:
+1. Checks if module is installed locally; if not, downloads from registry
+2. Parses the module's `main.nf` to identify the main process and its input declarations
+3. Validates command-line arguments against the process input schema
+4. Validates parameters against the `params` schema in `meta.yaml`
+5. Generates an implicit workflow that wires CLI arguments to process inputs
+6. Executes the workflow using standard Nextflow runtime
```
bentsherman:
```suggestion
6. Executes the workflow using standard Nextflow runtime
7. Prints JSON of process outputs
```

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1044 @@
...
+    --input 'data/*.fastq.gz' \
+    -profile docker \
+    -resume
+
+# Run with work directory and output specification
+nextflow module run nf-core/salmon \
+    --reads reads.fq \
+    --index salmon_index \
+    -work-dir /tmp/work \
+    --outdir results/
```
bentsherman:
```suggestion
    -output-dir results/
```

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1044 @@
...
+```bash
+nextflow module publish myorg/my-process
+nextflow module publish myorg/my-process -dry-run
+```
+
+**General Notes**:
+- All commands respect the `registry.url` configuration for custom registries
+- Modules are automatically downloaded on `nextflow run` if missing but configured
+
+## Module Structure
```
bentsherman, line 387:
```suggestion
## Module Structure

A module should define exactly one process. It may optionally define an entry workflow and any number of helper functions.
```

bentsherman:
Trying to resolve some differences between the module schema in this PR and the one in https://github.com/nextflow-io/schemas/pull/10

I think the only substantive change is adding the `topics` section to this PR. Everything else is just refactoring the JSON schema definitions for clarity

Overall, the schema in this PR seems to have better descriptions and validation constraints, as well as examples. So I'm happy to absorb those changes in the schemas repo once the ADR is merged.

adr/20251114-module-system.md
pditommaso: However it's needed in the include syntax to distinguish from existing modules eg. `include { FASTQC } from '@nf-core/fastqc'`. Alongside it should be used in the local path `./modules/@nf-core/fastqc`. 

We can drop in the CLI commands since implicitly will refers to "managed" modules

I'd keep in the config for consistency with `include` syntax

adr/20251114-module-system.md
pditommaso: The nextflow command should create the tar, and likely use the same library both nextflow side and backend side. Think the trick is to have a predictable traversal order and ignore file timestamps

adr/20251114-module-system.md
pditommaso: Good point, tho I would not require explicitly the output format. Likely json it should be the an output option

adr/20251114-module-system.md
pditommaso: An entry workflow to run itself? could not this conflict with the automatic process inputs mapping?

pditommaso:
> I think the only substantive change is adding the topics section to this PR. Everything else is just refactoring the JSON schema definitions for clarity

Good, let's add it

adr/20251114-module-system.md
bentsherman: My idea was that if the module defines an entry workflow and `params` block, then the `module run` command can just use that, no need to generate implicit ones. That way users can customize the standalone module execution if they want

In fact I think this is what it already does, just a matter of saying that it is allowed in a module

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1044 @@
...
+```
+
+---
+
+#### `nextflow module publish scope/name`
+
+Publish a module to the Nextflow registry, making it available for others to install. Requires authentication via API key and appropriate permissions for the target scope.
+
+**Arguments**:
+- `scope/name`: Module identifier to publish (required)
```
jorgee, line 359: Are we expecting that the module to publish is in `module/@scope/name` or are we going to allow to publish from whatever path? The `name` property in the `meta.yml` should also include the `scope/path`

adr/20251114-module-system.md
```diff
@@ -0,0 +1,1044 @@
...
+
+**Options**:
+- `-registry <url>`: Target registry URL (default: `registry.nextflow.io`)
+- `-tag <tag>`: Additional tags for discoverability
```
jorgee, line 363: Are not the same as `keywords` in `meta.yml`?

adr/20251114-module-system.md
pditommaso: Yes, it needs to be predictable path (including the name)

adr/20251114-module-system.md
pditommaso: Useless, let's remove `keywords` option

jorgee: Do you mean removing this tag option? or keywords in meta.yml?

pditommaso: Yes, ~~this option~~ the tag option

adr/20251114-module-system.md
pditommaso: Ok, let's move this discussion into a separate spec

adr/20251114-module-system.md
pditommaso: Basically explicit wrapper vs  implicit one on which dev has full control. It's a good idea but does not bring back the problem what's an what's a parameter? ideally the module should expose the process inputs as THE module input

pditommaso:
## ADR Update: Version 2.6

**Date:** 2026-01-28

### Changes
- **Removed module parameters**: Module parameters specification has been moved to a separate spec document.

This simplifies the ADR by keeping it focused on the core module system design, while the module parameters details are now maintained in a dedicated specification.

adr/20251114-module-system.md
bentsherman: > ideally the module should expose the process inputs as THE module input

Yes, it should do that by default. But if you define custom params / entry workflow, you end up defining the mapping of params -> process inputs when you call the process. So the process inputs wouldn't need to be exposed at all in this case, just the usual `params` block

pditommaso: Let's consider this as part of module params discussion https://github.com/nextflow-io/nextflow/pull/6769

bentsherman (changes requested):
Summarizing my remaining issues:

- **Remote module inclusion**: Move to a separate ADR. This idea needs to be investigated further on its own. With this ADR, users can already publish/install/remove remote modules at development-time through the module CLI.

- **Module versions in config**: Since the config is user-configurable while modules are inherent to the pipeline definition, module versions should be managed in a separate file, such as the pipeline spec (`nextflow_spec.json`). User-configurable modules can be explored separately, such as part of interface modules (#6737).

docs/reference/cli.md
```diff
@@ -1125,6 +1125,213 @@ $ nextflow log tiny_leavitt -F 'process =~ /split_letters/'
...
+  # Install specific version
+  $ nextflow module install nf-core/fastqc -version 1.0.0
+
+  # Force reinstall over local modifications
+  $ nextflow module install nf-core/fastqc -force
+  ```
+
+(cli-module-run)=
+
+`run [options] [scope/name] [--<input_name> <input-value>]`
```
ewels: I think we need more docs here on what `--<input_name>` is and how to know what's available.

docs/cli.md
```diff
@@ -262,7 +262,135 @@ $ nextflow secrets set AWS_ACCESS_KEY_ID
...
+$ nextflow module run nf-core/fastqc --input 'data/*.fastq.gz'
+$ nextflow module run nf-core/fastqc --input 'data/*.fastq.gz' -version 1.0.0
+```
+
+The command accepts all standard Nextflow execution options (`-profile`, `-resume`, etc.):
+
+```console
+$ nextflow module run nf-core/salmon \
+    --reads reads.fq \
+    --index salmon_index \
```
ewels: As above, it's quite unclear where these came from and how to know what can be specified here.

jorgee: `nextflow module info [scope/name] ` provides this information. 

Is it enough adding a sentence in docs or usage such as '_run `nextflow module info <scope/name>` to get more information about the expected inputs_' ?

adr/20251114-module-system.md
```diff
@@ -0,0 +1,897 @@
...
+    // url = [ 'https://custom.registry.com', 'https://registry.nextflow.io' ]
+
+    auth {
+        'registry.nextflow.io' = '${NXF_REGISTRY_TOKEN}'
+        'npm.myorg.com' = '${MYORG_TOKEN}'
+    }
+}
+```
+
+**Module Spec** (`meta.yaml`):
```
bentsherman: Note that [nf-core/modules](https://github.com/nf-core/modules) uses `meta.yml`, not `meta.yaml`. Do we want to allow both?

On the other hand, I was thinking that we should call it `spec.yaml` to align with the name (*module spec*). Even if we support the existing nf-core module schema, it could help differentiate nf-core modules from registry modules

modules/nf-lang/src/main/java/nextflow/script/control/ResolveIncludeVisitor.java
```diff
@@ -83,7 +83,10 @@ public void visitInclude(IncludeNode node) {
             setPlaceholderTargets(node);
             return;
         }
-        var includeUri = getIncludeUri(uri, source);
+
+        var isRemoteModule = ModuleResolver.isRemoteModule(source);
+        var parent = isRemoteModule ? Path.of("modules") : Path.of(uri).getParent();
```
bentsherman, line 90: This will probably not work for the language server, since the language server can manage multiple workspaces at once. I will need to pass a workspace root into this visitor.

I will look into this separately, no need to block this PR

adr/20251114-module-system.md
bentsherman: I see now that the implementation actually uses meta.yml. I updated the ADR and code comments accordingly

I don't care that much about `meta.yml` vs `spec.yml`, so let's just leave as is

adr/20251114-module-system.md
```diff
@@ -0,0 +1,897 @@
+# Module System for Nextflow
+
+- Authors: Paolo Di Tommaso
+- Status: draft
```
pditommaso:
```suggestion
- Status: approved
```

christopher-hakkaart:
Thanks @bentsherman - will check out the docs today

PR #6651, opened 2025-12-11
Fix String.format error when plugin URL contains percent chars

pditommaso:
## Summary
- Escape `%` characters to `%%` in error messages thrown by `HttpPluginRepository` when a connection exception occurs
- This prevents `String.format` interpretation errors in `PluginRuntimeException` when the URI contains URL-encoded characters (e.g., `%20` for spaces)
- Added test case to verify error messages with percent-encoded URLs are handled correctly

## Test plan
- [x] New test case `handle prefetch error with percent chars in error message` passes
- [x] All existing `HttpPluginRepositoryTest` tests pass

### DIFF
--- a/modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
+++ b/modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
@@ -137,9 +137,7 @@ class HttpPluginRepository implements PrefetchUpdateRepository {
             throw e
         }
         catch (Exception e) {
-            // Note: escape '%' chars to avoid String.format interpretation in PluginRuntimeException
-            final msg = "Unable to connect to ${uri} - cause: ${e.message}".replace('%','%%')
-            throw new PluginRuntimeException(e, msg)
+            throw new PluginRuntimeException(e, "Unable to connect to %s - cause: %s", uri, e.message)
         }
     }
 


modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
```diff
@@ -137,7 +137,9 @@ class HttpPluginRepository implements PrefetchUpdateRepository {
             throw e
         }
         catch (Exception e) {
-            throw new PluginRuntimeException(e, "Unable to connect to ${uri} - cause: ${e.message}")
+            // Note: escape '%' chars to avoid String.format interpretation in PluginRuntimeException
+            final msg = "Unable to connect to ${uri} - cause: ${e.message}".replace('%','%%')
+            throw new PluginRuntimeException(e, msg)
```
bentsherman: Where is `String.format()` being called? I don't see the problem

pditommaso:
```
    public PluginRuntimeException(Throwable cause, String message, Object... args) {
        super(StringUtils.format(message, args), cause);
    }
```

bentsherman: Okay then it seems like the proper way to use this exception is:

```suggestion
            throw new PluginRuntimeException(e, "Unable to connect to %s - cause: %s", uri, e.message)
```

pditommaso: make sense

PR #6654, opened 2025-12-11
Fix leading **/ glob to match zero-or-more directories

### LINKED ISSUE #6653: Leading **/ glob pattern does not match files at root level

When using `files("path/**/file.txt")`, I expect the `**/` glob to match files at **all** directory levels, including at the root of `path/`. However, the `**/` pattern only matches files that are at least one directory deep.

**Expected:** `files("data/**/data.txt")` should match both `data/data.txt` and `data/subdir/data.txt`

**Actual:** Only `data/subdir/data.txt` is matched; `data/data.txt` at the root level is missed.

Create the following directory structure:
```
data/
├── data.txt
└── subdir
    └── data.txt
```

```bash
mkdir -p data/subdir
echo "root level" > data/data.txt
echo "subdirectory level" > data/subdir/data.txt
```

Create `main.nf`:
```nextflow
#!/usr/bin/env nextflow

workflow {
    def folder = file("data")

    // This should find data.txt at BOTH root and subdirectory levels
    def results = files("${folder}/**/data.txt")

    log.info "Pattern: ${folder}/**/data.txt"
    log.info "Found: ${results}"
    log.info "Expected: [data/data.txt, data/subdir/data.txt]"
}
```

Run with:
```bash
nextflow run main.nf
```

### Program output

```
N E X T F L O W  ~  version 25.10.2

Launching `main.nf` [prickly_linnaeus] DSL2 - revision: a272970ba6

Pattern: /path/to/data/**/data.txt
Found: [/path/to/data/subdir/data.txt]
Expected: [data/data.txt, data/subdir/data.txt]
```

Note that `data/data.txt` at the root level is **not** included in the results.

### Environment

* Nextflow version: 25.10.2.10555
* Java version: openjdk 21.0.8
* Operating system: macOS (Darwin 25.1.0)
* Bash version: zsh 5.9

### Additional context

The root cause is in `FileHelper.visitFiles()`. When a pattern like `path/**/foo.txt` is parsed by `FilePatternSplitter`, it becomes:
- folder: `path/`
- pattern: `**/foo.txt`

Java's `PathMatcher` with glob syntax requires `**/` to match **at least one** directory component. The pattern `**/foo.txt` will match `bar/foo.txt` but not `foo.txt` at the root level.

**Workaround:** Use `{,**/}` instead of `**/` to explicitly match "zero or more directories":
```nextflow
def results = files("${folder}/{,**/}data.txt")
```

**Suggested fix:** Transform leading `**/` patterns to `{,**/}` in `FileHelper.visitFiles()` so that users get the expected behavior without needing the unintuitive workaround.

robsyme:
## Summary

- Transform leading `**/` glob patterns to `{,**/}` in `FileHelper.visitFiles()` so they match files at the root level (zero directories) as well as in subdirectories

## Test plan

- [x] Added test case `should match files at root and subdirectories with double-star glob`
- [x] Verified existing `visitFiles` tests still pass
- [ ] Run full test suite

Closes #6653

### DIFF
--- a/modules/nf-commons/src/main/nextflow/file/FileHelper.groovy
+++ b/modules/nf-commons/src/main/nextflow/file/FileHelper.groovy
@@ -859,7 +859,13 @@ class FileHelper {
         final syntax = options.syntax ?: 'glob'
         final relative = options.relative == true
 
-        final matcher = getPathMatcherFor("$syntax:${filePattern}", folder.fileSystem)
+        // Transform leading ** to match zero-or-more directories (not just one-or-more)
+        // Java's glob ** at pattern start requires at least one directory component,
+        // but users expect **/foo to also match foo at the root level
+        final adjustedPattern = filePattern.startsWith('**/')
+            ? "{,**/}" + filePattern.substring(3)
+            : filePattern
+        final matcher = getPathMatcherFor("$syntax:${adjustedPattern}", folder.fileSystem)
         final singleParam = action.getMaximumNumberOfParameters() == 1
 
         Files.walkFileTree(folder, walkOptions, Integer.MAX_VALUE, new SimpleFileVisitor<Path>() {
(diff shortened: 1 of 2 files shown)


robsyme:
@claude review please

pditommaso:
Closing since stating https://github.com/nextflow-io/nextflow/issues/6653 is not planned anymore

PR #6579, opened 2025-11-17
6570 slurm job state query

tcutts:
This patch adds support for the squeue --only-job-state option, which allows admins of recent SLURM versions to enable job state caching, and thereby reduce the RPC call load on SLURM.  Because not all SLURM versions support this, and it does cause other filtering options (-p and -u) to be ignored, this has a new config parameter for the executor, onlyJobState, which defaults to "false" to stay the same as current nextflow behaviour.

I have written unit tests for the code, included in the PR, plus some docs for the users.

I have run the code on AstraZeneca's SLURM cluster, and it appears to behave correctly, at least for a very simple test workflow.

PR #6659, opened 2025-12-13
6570 slurm job state query (clean version)

### LINKED ISSUE #6570: Accelerate SLURM job state queries with --only-job-state option to squeue
## New feature

Make SLURM requests more efficient by [optionally] using the --only-job-state option to squeue

## Use case 

SLURM supported releases (24 and 25) introduced the ability to return just the job state information much more quickly.  Nextflow should use this capability if it's available in order to reduce the RPC burden on SLURM itself and accelerate scheduling. 

## Suggested implementation 

All that needs doing is changing the squeue command line in the executor plugin to:

`squeue --noheader --only-job-state -o "%i %t" -t all`

(i.e. one extra option).  The format of the output is _almost_ identical; I think the only difference is whether the % syntax for job arrays is emitted (i.e. how many jobs in the job array can be run at once, something I don't think nextflow makes use of)

If the SLURM administrator combines this with adding the enable_job_state_cache configuration in SLURM, it will mean that nextflow job state queries will no longer use the RPC mechanism which causes such performance difficulties for naive workflows.

Obviously there will need to be a bit more code than that; either to make the precise command configurable or to test the SLURM version for whether that option is available, and use it if it is.

tcutts:
This patch adds support for the squeue --only-job-state option, which allows admins of recent SLURM versions to enable job state caching, and thereby reduce the RPC call load on SLURM. Because not all SLURM versions support this, and it does cause other filtering options (-p and -u) to be ignored, this has a new config parameter for the executor, onlyJobState, which defaults to "false" to stay the same as current nextflow behaviour.

I have written unit tests for the code, included in the PR, plus some docs for the users.

I have run the code on AstraZeneca's SLURM cluster, and it appears to behave correctly, at least for a very simple test workflow.

Apologies for having to redo the pull request, I made a complete mess of git in my fork when trying to amend the commit logs and attest them properly, and the pull request suddenly seemed to be trying to make thousands of changes.  This is a new clean version with just the correct 5 commits.

Old PR #6579
close #6570

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -1594,6 +1594,11 @@ The following settings are available:
   :::
 : Forcefully apply compression option to all layers, including already existing layers (default: `false`).
 
+`wave.build.conda.baseImage`
+: :::{versionadded} 25.12.0-edge
+  :::
+: The base image for the final stage in multi-stage Conda container builds (default: `ubuntu:24.04`). This option only applies when using `wave.build.template` set to `conda/micromamba:v2` or `conda/pixi:v1`.
+
 `wave.build.conda.basePackages`
 : One or more Conda packages to be always added in the resulting container (default: `conda-forge::procps-ng`).
 
@@ -1606,6 +1611,16 @@ The following settings are available:
 `wave.build.repository`
 : The container repository where images built by Wave are uploaded.
 : The corresponding credentials must be provided in your Seqera Platform account.
+
+`wave.build.template`
+: :::{versionadded} 25.12.0-edge
+  :::
+: The build template to use for container builds (default: `conda/micromamba:v1`). Supported values:
+: - `conda/micromamba:v1`: Standard Micromamba 1.x single-stage build. Default when unspecified.
+: - `conda/micromamba:v2`: Micromamba 2.x with multi-stage builds.
+: - `conda/pixi:v1`: Pixi package manager with multi-stage builds for optimized image sizes.
+: - `cran/installr:v1`: R/CRAN packages using installr.
+: Multi-stage templates produce smaller images by excluding build tools from the final image.
 
 `wave.httpClient.connectTimeout`
 : :::{versionadded} 22.06.0-edge
--- a/docs/wave.md
+++ b/docs/wave.md
@@ -92,6 +92,40 @@ conda.channels = 'conda-forge,bioconda'
 
 Packages from the [Python Package Index](https://pypi.org/) can also be added to a Conda `environment.yml` file. See {ref}`Conda and PyPI <conda-pypi>` for more information.
 
+(wave-build-templates)=
+
+### Build templates
+
+:::{versionadded} 25.12.0-edge
+:::
+
+Wave supports different build templates for creating container images from Conda packages. Build templates control how packages are installed and how the final container image is structured.
+
+Multi-stage build templates (`conda/pixi:v1` and `conda/micromamba:v2`) offer several advantages:
+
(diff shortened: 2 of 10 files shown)


docs/executor.md
```diff
@@ -446,6 +446,10 @@ Nextflow does not provide direct support for SLURM multi-clusters. If you need t
 Some SLURM clusters require memory allocations to be specified with `--mem-per-cpu` instead of `--mem`. You can specify `executor.perCpuMemAllocation = true` in the Nextflow configuration to enable this behavior. Nextflow will automatically compute the memory per CPU for each task (by default 1 CPU is used).
 :::
 
+:::{versionadded} 25.12.0-edge
+Since SLURM 24, `squeue` supports an `--only-job-state` option that ignores the partition (`-p`) or user (`-u`) filters. To enable this behavior, specify `executor.$slurm.onlyJobState = true` in your Nextflow configuration. If `SchedulerParameters=enable_job_state_cache` is enabled, you can expect improved Nextflow performance and reduced load on the SLURM controller. More information about this feature in the [SLURM configuration](https://slurm.schedmd.com/slurm.conf.html#OPT_enable_job_state_cache) and the [`squeue` documentation](https://slurm.schedmd.com/squeue.html#OPT_only-job-state).
```
christopher-hakkaart:
```suggestion
Since SLURM 24, `squeue` supports an `--only-job-state` option that ignores the partition (`-p`) or user (`-u`) filters. To enable this behavior, specify `executor.$slurm.onlyJobState = true` in your Nextflow configuration. If `SchedulerParameters=enable_job_state_cache` is enabled, you can expect improved Nextflow performance and reduced load on the SLURM controller. See [`enable_job_state_cache`](https://slurm.schedmd.com/slurm.conf.html#OPT_enable_job_state_cache) and [`--only-job-state`](https://slurm.schedmd.com/squeue.html#OPT_only-job-state) for more information.
```

christopher-hakkaart: I've trimmed some of the words and called out the config options specifically. Optional if you disagree.

docs/reference/config.md
```diff
@@ -715,6 +715,13 @@ The following settings are available:
 : *Used only by the {ref}`slurm-executor` executor.*
 : When `true`, memory allocations for SLURM jobs are specified as `--mem-per-cpu <task.memory / task.cpus>` instead of `--mem <task.memory>`.
 
+`executor.onlyJobState`
+: :::{versionadded} 25.12.0-edge
+  :::
+: *Used only by the {ref}`slurm-executor` executor.*
+: *Requires SLURM 24.05 or later.*
+: When `true`, job status queries use `squeue --only-job-state` without partition (`-p`) or user (`-u`) filters. This can reduce the load on the SLURM controller, especially if your SLURM administrator has enabled `SchedulerParameters=enable_job_state_cache` in the [SLURM configuration](https://slurm.schedmd.com/slurm.conf.html#OPT_enable_job_state_cache). More information in the [`squeue` documentation](https://slurm.schedmd.com/squeue.html#OPT_only-job-state). (default: `false`).
```
```suggestion
: When `true`, job status queries use `squeue --only-job-state` without partition (`-p`) or user (`-u`) filters. This can reduce the load on the SLURM controller, especially if your SLURM administrator has enabled `SchedulerParameters=enable_job_state_cache` in your [SLURM configuration](https://slurm.schedmd.com/slurm.conf.html#OPT_enable_job_state_cache). See [`--only-job-state`](https://slurm.schedmd.com/squeue.html#OPT_only-job-state) for more information (default: `false`).
```

PR #6661, opened 2025-12-15
Fix ADR directory structure to match implementation

pditommaso:
## Summary
- Updates the multi-revision asset management ADR to reflect the actual directory structure used in the implementation
- Fixes the documented paths:
  - `bare/` instead of `.nextflow/bare_repo/`
  - `commits/` instead of `.nextflow/commits/`
  - Legacy repo at separate location `~/.nextflow/assets/<project>/` instead of inside `.repos` directory

## Test plan
- [x] Verified paths match `MultiRevisionRepositoryStrategy.groovy` constants

### DIFF
--- a/adr/20251205-multi-revision-asset-management-strategy-pattern.md
+++ b/adr/20251205-multi-revision-asset-management-strategy-pattern.md
@@ -75,76 +75,68 @@ Use a bare repository for storage and create clones for each commit, tracking th
 
 **Decision**: Initially implemented but later refined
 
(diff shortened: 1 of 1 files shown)


pditommaso:
@jorgee reading this, it seems option 4 (Strategy Pattern for Backward Compatibility) is an alternative implementation option. Should be included with the option 3 (the one that was implemented) ?

jorgee:
yes, 3 and 4 can be merged in one.

pditommaso:
Ok, have a look at the updated docs and it everything makes sense

PR #6665, opened 2025-12-16
Support conda lock files with any extension and remote URLs

### LINKED ISSUE #5583: Conda: Support `.lock` file extensions.
The `conda` process takes a string input that can be either one or more package names, or a conda environment / lock file. To determine which of these two the string is, we check the file extension:

https://github.com/nextflow-io/nextflow/blob/5c5a810abe0da342c953de2c2909fb903ba716e3/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy#L160-L167

This generally works well as until now almost all usage has been with files called `environment.yml`, by convention. However, since Wave has started supporting Conda lock files, this convention has become weaker. Wave itself generates downloads called `condalock` (no extension) and the Conda ecosystem often uses `.lock` file extensions by loose convention. These are interpreted as package names by Nextflow and the Conda environment resolution fails.

Instead I'd propose one of two different approaches:

1. Try to match the `conda` string against local file paths, treat as package names only if no local files match (eg. accept any file extension)
2. Explicitly add support for `*.lock` and `condalock` file names.

Originally raised by @pinin4fjords in https://github.com/nextflow-io/nextflow/issues/5453 - see original issue for examples of current failure / error messages. Splitting that issue into two to keep discussion focussed.

See also https://github.com/nextflow-io/nextflow/issues/5582 about handling remote files.

ewels:
From https://github.com/nextflow-io/nextflow/issues/5583#issuecomment-2875484270:

> @pinin4fjords: Conda lock files DO work, but they can't currently have the '.lock' extension. They must be .txt, and can't be remote files.

## PR Goals:

- Support remote URIs for conda lock files
- Detect conda lock files by looking at file content, instead of relying on file extension

<details>
This change allows conda lock files to be used with any file extension (e.g., .lock, .txt, or no extension) as long as they contain the `@EXPLICIT` marker in the first 20 lines. Additionally, remote conda lock files can now be specified using http/https URLs.

Changes:
- Add isLockFile() method to detect lock files by `@EXPLICIT` marker
- Add isRemoteFile() method to check for http/https URLs
- Add stageRemoteFile() method to download remote files locally
- Modify condaPrefixPath() to handle remote files and lock files with any extension
- Modify createLocalCondaEnv0() to detect and handle lock files properly
- Add tests for new functionality
- Update documentation to reflect new behavior

Fixes #5583
</details>


docs/conda.md
```diff
@@ -163,10 +163,46 @@ https://conda.anaconda.org/conda-forge/linux-64/libgcc-ng-13.2.0-h77fa898_7.cond
...
+  conda '/path/to/spec-file.lock'
+
+  script:
+  """
+  your_command --here
+  """
+}
+```
+
+Conda lock files can also be specified using a remote URL (HTTP, HTTPS, S3, Google Cloud Storage, Azure Blob Storage, etc.):
```
christopher-hakkaart:
```suggestion
You can also specify Conda lock files using a remote URL (HTTP, HTTPS, S3, Google Cloud Storage, Azure Blob Storage, etc.):
```

docs/conda.md
```diff
@@ -163,10 +163,46 @@ https://conda.anaconda.org/conda-forge/linux-64/libgcc-ng-13.2.0-h77fa898_7.cond
...
+```nextflow
+process hello {
+  conda 's3://my-bucket/conda/env.lock'
+
+  script:
+  """
+  your_command --here
+  """
+}
+```
```
ewels: I think it's probably a bit excessive to spell out the s3 stuff as well, this is probably fine:
````suggestion
You can also specify Conda lock files using a remote URL:

```nextflow
process hello {
  conda 'https://wave.seqera.io/v1alpha1/builds/bd-12345/condalock'

  script:
  """
  your_command --here
  """
}
```

````

**Comment by pditommaso** [comment]
It looks like this PR tries to do too many things (parsing content, formatting env (?), supporting remote uris, etc), please scope clearly the goals.

**Comment by ewels** [comment]
Tested successfully in GitHub codespaces, working as expected:

<details><summary><code>main.nf</code></summary>

```nextflow
#!/usr/bin/env nextflow

params.input = "data/*.fq.gz"

workflow {
    channel.fromPath(params.input)
        .set { reads }

    FASTQC(reads)
}

process FASTQC {
    conda "https://wave.seqera.io/v1alpha1/builds/bd-af7a5314d5015c29_1/condalock"

    input:
    path reads

    output:
    path "*.html"
    path "*.zip"

    script:
    """
    fastqc ${reads}
    """
}
```

With a `nextflow.config` with `conda.enabled = true`
</details>

I tried locally at first but it failed, because the lockfile was for a linux/arch64 Docker build so didn't work on my apple silicon ARM processor.

ewels:
> It looks like this PR tries to do too many things (parsing content, formatting env (?), supporting remote uris, etc), please scope clearly the goals.

You basically described the goals 😅 - though I don't think it does any formatting?

Goals:
* Support remote URIs for conda lock files
* Detect conda lock files by looking at file content, instead of relying on file extension

ewels:
I updated the PR description to try to make this clearer @pditommaso. I'm not really sure how I can reduce the scope of the code changes without missing the goals. Suggestions welcome!

pinin4fjords:
I think the scope here is appropriate. The two goals are inherently linked - remote conda lock files (like Wave's `condalock` endpoint) don't have `.txt` extensions, so content-based detection (`@EXPLICIT` marker) is required to make remote URLs work at all.

The implementation is clean: three small helper methods (`isLockFile`, `isRemoteFile`, `stageRemoteFile`) plus integration into the existing flow. The test coverage is thorough.

This unblocks a real user pain point from #5583 - conda lock files are increasingly common in nf-core and other pipelines, and the current `.txt` extension requirement is confusing and limiting.

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
```diff
@@ -166,6 +168,62 @@ class CondaCache {
...
+    boolean isLockFile(String content) {
+        if( !content )
+            return false
+        final lines = content.readLines()
+        final limit = Math.min(20, lines.size())
+        for( int i = 0; i < limit; i++ ) {
+            if( lines[i].trim() == '@EXPLICIT' )
+                return true
+        }
+        return false
```
pditommaso: This should use a reader (ie. read line by line) to avoid loading all file in memory to parse the header

ewels:
Ok, I reverted that part of the PR and added an admonition to the docs explaining that environment.yml files have partial remote support (http) but that's all. 

The behavior is now:
* Remote YAML files (http/https + .yml/.yaml): URL passed directly to conda/mamba
* Remote lock files: Throws error (not supported)
* Local lock files: Work with any extension (.lock, .txt, no extension, etc.)

modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
ewels: Done in 5debd49

ewels:
@pditommaso should be good to go now.

pditommaso:
Closing in favour of #6714 6713

Issue #6668, opened 2025-12-17
Improve S3 filesystem error messages when credentials are missing or access is denied


pinin4fjords:
## Summary

The S3 filesystem provider propagates raw AWS SDK exceptions when operations fail due to credential or access issues. This results in cryptic error messages that don't help users understand what went wrong, and prevents downstream tools from handling access errors appropriately.

### Example Error

When a user's AWS credentials cannot access an S3 path, they see errors like:

```
could not validate file format of 's3://annotation-cache/snpeff_cache/': Unable to marshall request to JSON: Key cannot be empty
```

A more helpful message would be:

```
Cannot access S3 path 's3://annotation-cache/snpeff_cache/': AWS credentials missing or access denied
```

## Affected Use Cases

This affects nf-core pipelines that have default S3 paths for resources (e.g., annotation caches, iGenomes). Users with their own AWS credentials configured for different accounts see confusing errors when pipeline validation checks these paths.

Related user report: nf-core/sarek#2079

## Impact on nf-schema Validation

Tools like nf-schema need to distinguish between:
1. **Access errors** (credentials missing, permission denied) → should skip validation gracefully
2. **Format errors** (path is malformed, wrong type) → should fail validation

Currently, nf-schema receives generic exceptions and cannot make this distinction. With proper exception handling, nf-schema could:
- Catch `AccessDeniedException` → skip path validation gracefully (user may not need this path)
- Let other validation (format checks, required fields) proceed normally

This allows nf-schema to bypass existence/accessibility validation on access errors without bypassing format validation.

## Root Cause

The `S3ObjectSummaryLookup.lookup()` method (`plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/util/S3ObjectSummaryLookup.java`) calls AWS SDK methods without catching SDK exceptions:

- Line 55: `client.getObjectMetadata(s3Path.getBucket(), "")` 
- Line 84: `client.listObjects(request.build())`

These methods can throw AWS SDK exceptions (`SdkClientException`, `S3Exception`) that propagate uncaught.

The `S3ObjectSummaryLookup.lookup()` method signature only declares `NoSuchFileException`:
```java
public S3Object lookup(S3Path s3Path) throws NoSuchFileException
```

## Proposed Solution

Catch AWS SDK exceptions in `S3ObjectSummaryLookup.lookup()` and translate them to standard Java NIO exceptions with clear messages:

1. Catch `S3Exception` and check status code:
[...]

PR #6669, opened 2025-12-17
Improve S3 error messages for credential and access issues

pinin4fjords:
## Summary

This PR illustrates a potential approach to translating AWS SDK exceptions to standard Java NIO exceptions with clearer error messages. **It is not intended to be merged as-is** - we're looking for guidance on a better solution.

### The Problem

Users see cryptic errors like:
```
could not validate file format of 's3://annotation-cache/snpeff_cache/': Unable to marshall request to JSON: Key cannot be empty
```

When they should see something like:
```
Access denied to S3 path - check AWS credentials and permissions
```

### What This PR Demonstrates

- Exception translation in `lookup()` and `getS3ObjectMetadata()` methods
- Mapping HTTP status codes (401/403 → `AccessDeniedException`, 404 → `NoSuchFileException`)
- Wrapping other SDK errors as `IOException`

## ⚠️ Why This Approach Is Problematic

The current implementation relies on fragile string matching to detect credential errors:

```java
message.contains("Unable to load credentials") ||
message.contains("Unable to marshall request")
```

**This is not a viable solution** because:
- It depends on error message text that could change between SDK versions
- The `"Unable to marshall request"` pattern catches serialization errors that *coincidentally* appear when credentials are misconfigured - this is not a reliable indicator
- It will misclassify errors and break unpredictably

## What We Need

We're looking for guidance on how to properly detect credential and access issues from AWS SDK exceptions. Is there a more robust approach using exception types, error codes, or the SDK's error handling facilities?

## Context

See issue #6668 for the full problem description. The goal is to allow downstream tools like nf-schema to distinguish access errors from other failures.

### DIFF
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/util/S3ObjectSummaryLookup.java
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/util/S3ObjectSummaryLookup.java
@@ -137,7 +137,9 @@ public class S3ObjectSummaryLookup {
 
         // Check for client-side credential errors based on message patterns
         String message = e.getMessage();
-        if (message != null && message.contains("Unable to load credentials")) {
+        if (message != null && (
+                message.contains("Unable to load credentials") ||
+                message.contains("Unable to marshall request"))) {
             return new AccessDeniedException(path, null,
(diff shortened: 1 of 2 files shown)


pinin4fjords:
Proper fix by @jorgee at https://github.com/nextflow-io/nextflow/pull/6706

PR #6672, opened 2025-12-18
Fix WaveClient sending Bearer token to public S3 URLs

### LINKED ISSUE #6671: WaveClient sends Bearer token to public S3 URLs, causing authentication failures
## Bug Description

When using a public AWS S3 URL as `containerConfigUrl` (e.g., for Fusion container config), the Wave plugin sends a Bearer token (Tower/Platform JWT) with the request. AWS S3 does not support Bearer token authentication and returns a 400 error: "Unsupported Authorization Type".

## Error Message

```
Unexpected response for containerContainerConfigUrl 'https://s3.eu-west-2.amazonaws.com/fusion-develop/snapshots/.../fusion-amd64.json': [400]
<Error>
  <Code>InvalidArgument</Code>
  <Message>Unsupported Authorization Type</Message>
  <ArgumentName>Authorization</ArgumentName>
  <ArgumentValue>Bearer eyJhbGciOiJIUzI1NiJ9...</ArgumentValue>
</Error>
```

## Root Cause

The `WaveClient` creates a single `HxClient` instance with a Bearer token configured globally:

https://github.com/nextflow-io/nextflow/blob/cdc7a5869/plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy#L145-L155

```groovy
protected HxClient newHttpClient() {
    return HxClient.newBuilder()
            .bearerToken(tower.accessToken)  // Bearer token applied to ALL requests
            // ...
            .build()
}
```

This same client is used for fetching container configs from external URLs:

https://github.com/nextflow-io/nextflow/blob/cdc7a5869/plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy#L397-L413

```groovy
protected ContainerConfig fetchContainerConfig(URL configUrl) {
    final req = HttpRequest.newBuilder()
            .uri(configUrl.toURI())
            .GET()
            .build()
    final resp = httpClient.sendAsString(req)  // Bearer token sent regardless of URL
    // ...
}
```

When the container config URL points to a public S3 bucket, the Bearer token is incorrectly sent, and S3 rejects it.

## Steps to Reproduce

1. Configure Nextflow with Wave and Tower/Platform integration
2. Set a custom Fusion container config URL pointing to a public S3 bucket:
   ```
   fusion.containerConfigUrl = 'https://s3.eu-west-2.amazonaws.com/bucket/path/fusion-amd64.json'
   ```
3. Run a workflow with Fusion enabled
4. Observe the authentication error

## Expected Behavior

Public URLs (especially S3 URLs) should be fetched without the Bearer token, since:
- Container configs are public resources
- S3 doesn't support Bearer authentication
- The Bearer token is only valid for Wave/Tower API endpoints

## Test URL

This public S3 URL can be used to reproduce and test the issue:

```
https://s3.eu-west-2.amazonaws.com/fusion-develop/snapshots/2025/12/17/174823056/fusion-amd64.json
```

## Environment

- Nextflow version: master (cdc7a5869)
- Wave plugin version: latest
- Platform: AWS with S3 Fusion config URL

pditommaso:
## Summary
- Fix WaveClient sending Bearer token to external URLs (e.g., public S3 buckets) when fetching container configs
- Add dedicated `plainHttpClient()` without authentication for external resource requests
- S3 and similar services reject requests with unsupported Authorization headers

## Changes
- Added `plainHttpClient()` factory method with `@Memoized` annotation for lazy initialization
- Updated `fetchContainerConfig()` to use the plain client instead of the authenticated client
- Added Javadoc to explain the rationale for each HTTP client
- Added unit test that verifies container config fetch works without Bearer token

Fixes #6671

## Test plan
- [x] Unit test added that simulates S3 behavior (rejects Bearer tokens)
- [x] All existing nf-wave tests pass

### DIFF
--- a/plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
+++ b/plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
@@ -142,6 +142,12 @@ class WaveClient {
     /* only for testing */
     protected WaveClient() { }
 
+    /**
+     * Creates the main HTTP client for Wave/Tower API communication.
+     * This client includes Bearer token authentication for secure API calls.
+     *
+     * @return An {@link HxClient} configured with authentication tokens
+     */
     protected HxClient newHttpClient() {
         final refreshUrl = tower.refreshToken ? "${tower.endpoint}/oauth/access_token" : null
         return HxClient.newBuilder()
@@ -154,6 +160,11 @@ class WaveClient {
                 .build()
     }
 
+    /**
+     * Creates the underlying Java HTTP client with common configuration.
+     *
+     * @return A configured {@link HttpClient} instance
+     */
     protected HttpClient newHttpClient0() {
         final builder = HttpClient.newBuilder()
                 .version(HttpClient.Version.HTTP_1_1)
@@ -164,6 +175,24 @@ class WaveClient {
             builder.executor(Executors.newVirtualThreadPerTaskExecutor())
         // build and return the new client
         return builder.build()
+    }
+
+    /**
+     * Creates an HTTP client without authentication for fetching external resources.
+     * <p>
+     * This client is used for requests to external URLs (e.g., public S3 buckets)
+     * that do not require or support Bearer token authentication. Services like
+     * AWS S3 reject requests with unsupported Authorization headers.
+     *
+     * @return An {@link HxClient} without authentication configuration
+     * @see #newHttpClient() for authenticated Wave/Tower API calls
+     */
+    @Memoized
+    protected HxClient plainHttpClient() {
+        return HxClient.newBuilder()
+                .httpClient(newHttpClient0())
+                .retryConfig(config.retryOpts())
+                .build()
     }
 
     WaveConfig config() { return config }
@@ -402,7 +431,7 @@ class WaveClient {
                 .GET()
                 .build()
 
-        final resp = httpClient.sendAsString(req)
+        final resp = plainHttpClient().sendAsString(req)
         final code = resp.statusCode()
         final body = resp.body()
         if( code>=200 && code<400 ) {
--- a/plugins/nf-wave/src/test/io/seqera/wave/plugin/WaveClientTest.groovy
+++ b/plugins/nf-wave/src/test/io/seqera/wave/plugin/WaveClientTest.groovy
@@ -1410,4 +1410,47 @@ class WaveClientTest extends Specification {
             cached: resp.cached )
     }
 
+    def 'should fetch container config without bearer token' () {
+        given: 'a server that rejects requests with Authorization header (like S3)'
+        def configJson = JsonOutput.toJson(new ContainerConfig(entrypoint: ['test.sh']))
+        HttpHandler handler = { HttpExchange exchange ->
+            // Check if Authorization header is present (like S3 would reject Bearer tokens)
+            def authHeader = exchange.requestHeaders.getFirst('Authorization')
+            if( authHeader ) {
+                // S3 returns 400 for unsupported auth types
+                def errorMsg = "Unsupported Authorization Type: ${authHeader}"
+                exchange.sendResponseHeaders(400, errorMsg.size())
+                exchange.getResponseBody() << errorMsg
+                exchange.getResponseBody().close()
+            }
+            else {
+                exchange.getResponseHeaders().add("Content-Type", "application/json")
+                exchange.sendResponseHeaders(200, configJson.size())
+                exchange.getResponseBody() << configJson
+                exchange.getResponseBody().close()
+            }
+        }
+
+        HttpServer server = HttpServer.create(new InetSocketAddress(9902), 0)
+        server.createContext("/", handler)
+        server.start()
+
+        and: 'a wave client configured with tower access token'
+        def config = [
+            wave: [containerConfigUrl: 'http://localhost:9902/config.json'],
+            tower: [accessToken: 'test-jwt-token', endpoint: 'https://tower.nf']
+        ]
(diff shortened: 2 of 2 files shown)


plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
```diff
@@ -402,7 +431,7 @@ class WaveClient {
                 .GET()
                 .build()
 
-        final resp = httpClient.sendAsString(req)
+        final resp = plainHttpClient().sendAsString(req)
```
jorgee, line 434: Should we always make this request without token? In the default case with the wave URL, isn't required?

pditommaso:
@jorgee Good question. After reviewing both codebases, here's what I found:

**The Bearer token IS used by Wave backend**, but for a different purpose than fetching the container config URL.

### How the token flow works:

1. **Nextflow WaveClient** sends the Tower access token to Wave API endpoints (like `/v1alpha2/container`)

2. **Wave backend validates it against Tower/Platform** - it forwards the token to the Tower `/user-info` endpoint to verify it's valid and retrieve the user identity (`UserServiceImpl.getUserByAccessToken`)

3. **What the token enables**: Once validated, the `PlatformId` (user info, workspace, token) is used for:
   - Fetching registry credentials from Tower
   - Accessing workflow launch info
   - Rate limiting and metrics tied to user identity

### Why `plainHttpClient()` is correct here:

The `containerConfig` URL returned by Wave points to an **external resource** (like a public S3 bucket), not to the Wave API itself. This URL doesn't need/accept the Tower Bearer token - S3 uses different auth mechanisms (signed URLs, IAM, etc.).

The Wave API request that returns this URL already used the authenticated client. The subsequent fetch of the config content from the external URL should be unauthenticated.

So to answer your question: No, the token is not required for fetching the container config content from external URLs. The token is only needed for Wave API calls, which still use `httpClient()` with authentication.

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
pditommaso: I had review this, and wave uses that token. See https://github.com/nextflow-io/nextflow/pull/6672#issuecomment-3669915328

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
```diff
@@ -166,6 +177,24 @@ class WaveClient {
...
+     * AWS S3 reject requests with unsupported Authorization headers.
+     *
+     * @return An {@link HxClient} without authentication configuration
+     * @see #newHttpClient() for authenticated Wave/Tower API calls
+     */
+    @Memoized
+    protected HxClient plainHttpClient() {
+        return HxClient.newBuilder()
+                .httpClient(newHttpClient0())
+                .retryConfig(config.retryOpts())
```
jorgee, line 194:
```suggestion
                .retryConfig(config.retryOpts())
                .followRedirects(HttpClient.Redirect.NORMAL)
```

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
```diff
@@ -402,7 +431,7 @@ class WaveClient {
                 .GET()
                 .build()
 
-        final resp = httpClient.sendAsString(req)
+        final resp = plainHttpClient().sendAsString(req)
         final code = resp.statusCode()
         final body = resp.body()
         if( code>=200 && code<400 ) {
```
jorgee, line 437:
```suggestion
        if( code>=200 && code<300 ) {
```

jorgee (changes requested):
I was testing the provided URL and it was answering with 301 it was considered a valid status code and then there was a failure because it was expecting a JSON but S3 respond with an XML

```
Caused by:
  Expected BEGIN_OBJECT but was STRING at line 1 column 1 path $
  See https://github.com/google/gson/blob/main/Troubleshooting.md#unexpected-json-structure
```

As it is not doing a Wave client request, I suggest to do the following changes:
1 - configure the `plainHttpClient` with followRedirection(NORMAL) 
2 - manage redirection status code as error to avoid the JSON error problem. 

Not sure for (1), but at least we should do (2)

plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy
pditommaso: Quite sure this is the default

pditommaso:
The redirect is handled by the http client by default. Just tried my branch and it's ok

jorgee:
Maybe I get confused, but HxClient is not HxClient a wrapper of HttpClient. In that libray, the default for followredirects is NEVER https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.html#followRedirects()

But, either NORMAL is the default or not, if the redirect has been followed, we shouldn't receive 3XX. So I think 3XX should be considered an error. This is the reason why I was getting the JSON parsing error when trying with "fusion.containerConfigUrl = 'https://s3.eu-west-2.amazonaws.com/bucket/path/fusion-amd64.json'

pditommaso:
So why (integration) tests pass ?

jorgee:
Ok, I see what is happening. The URL in the issue is fake. As it is a regional S3, the AWS is returning 301 to redirect to global endpoint. It is what produces the error in my case. With a correct URL, it is working.

pditommaso:
I assume this is ok in the correct form then

jorgee (approved):

jorgee:
Yes

PR #6675, opened 2025-12-18
Allow collection-type params to be loaded from files

bentsherman:
This PR adds the ability to declare a param as a collection type:

```groovy
params {
  input: List<Map>
}

workflow {
  println params.input*.id.join(',')
}
```

And supply a CSV, JSON, or YAML file at runtime:

```console
$ cat input.json
[
  { "id": "1", "fastq": "data/1.fastq" },
  { "id": "2", "fastq": "data/2.fastq" },
  { "id": "3", "fastq": "data/3.fastq" }
]

$ nextflow run main.nf --input input.json
1,2,3
```

Nextflow interprets the string as a file path and attempts to load the file contents based on the file extension. A runtime error is reported if the file contents are not compatible with the declared type.

Future work will extend the type validation with record types.

(spun off from #6574)

PR #6676, opened 2025-12-19
Extend collection param from file to support Iterable, Bag and Map

jorgee:
Extends #6675 to support also `Iterable,` `Bag` and `Map` standard types

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/ParamsDsl.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/ParamsDsl.groovy
@@ -16,9 +16,6 @@
 
 package nextflow.script
 
-import nextflow.script.types.Bag
-import nextflow.util.ArrayBag
-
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/script/ParamsDsl.groovy
```diff
@@ -117,7 +120,7 @@ class ParamsDsl {
             return MemoryUnit.of(str)
         }
 
-        if( Collection.class.isAssignableFrom(decl.type) ) {
+        if( Collection.class.isAssignableFrom(decl.type) || decl.type == Iterable.class || decl.type == Map.class) {
```
jorgee: I used `decl.type == Iterable.class` instead of I`terable.class.isAssignableFrom(decl.type)`, because `Path` implements` Iterable<Path>`, therefore inputs declared as `Path` were fulfilling this condition

PR #6353, opened 2025-08-20
Add opt-in patch for unordered directory hashing

pditommaso:
## Summary

This PR implements improved order-independent hashing for directories and unordered collections, controlled by the `NXF_PATCH_UNORDERED_DIR` environment variable.

### Key Features

- **Order-independent directory hashing**: Directory traversal order no longer affects hash values
- **Consistent unordered collection hashing**: Sets and Bags produce consistent hashes regardless of internal ordering
- **Backward compatible**: Feature is disabled by default, enabled via `NXF_PATCH_UNORDERED_DIR=true`
- **Addresses edge cases**: Fixes issue where directories with similar contents could have the same hash (relates to nextflow-io/nextflow#6198)

### Technical Implementation

- Uses commutative byte addition for order independence in both directory and collection hashing
- Implements separate "patched" methods: `hashDirSha256Patched()` and `hashUnorderedCollectionPatched()`
- Maintains full backward compatibility - default behavior unchanged
- Added comprehensive tests covering both default and patched behaviors

### Usage

To enable the patch:
```bash
export NXF_PATCH_UNORDERED_DIR=true
nextflow run your-pipeline.nf
```

### Test Coverage

- ✅ Original tests continue to pass (backward compatibility)
- ✅ New tests verify order independence for directories with same content but different creation order
- ✅ Tests verify different content yields different hashes (no false positives)
- ✅ Comprehensive SysEnv testing for environment variable control

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #6679, opened 2025-12-19
Record types

bentsherman:
This PR implements the first preview of record types:

- Add `Record` type to standard types
- Add `record()` function to create a record from named arguments
- Add record types via `record` keyword for validating records
- Add ability to destructure a record in a process input
- Fix hashing of records
- Add ability to include records across modules
- Add ability to stage record type

**Cache breaking**: This PR changes the way that map values are hashed. It now includes both the keys and values instead of only the values. It also prioritizes `CacheFunnel` over `Map`, so that custom classes can implement `Map` while also implementing `CacheFunnal` in order to define custom hashing behavior.

### DIFF
--- a/adr/20260306-record-types.md
+++ b/adr/20260306-record-types.md
@@ -1,0 +1,435 @@
+# Record types
+
+- Authors: Ben Sherman
+- Status: accepted
+- Deciders: Ben Sherman, Paolo Di Tommaso
+- Date: 2026-03-06
+- Tags: lang, static-types
+
+## Summary
+
+Provide a way to model composite data types in the Nextflow language.
+
+## Problem Statement
+
+Nextflow pipelines typically need a way to model *aggregates*, or "data that travels together", such as a paired-end read consisting of a sample ID and two FASTQ files.
+
+Primary use cases:
+
+- Model complex pipeline inputs and outputs (e.g. samplesheets as collections of records)
+
+- Generate JSON schemas for pipeline inputs and outputs from source code (e.g. to facilitate pipeline chaining with external validation)
+
+- Model directory outputs as records, enabling more fine-grained validation
+
+## Goals
+
+- Introduce a alternative data structure to tuples that allows users to model their data domain more precisely.
+
+- Introduce a way to validate custom data types at compile-time while keeping pipeline code concise and readable.
+
+## Non-goals
+
+- Removing support for tuples -- tuples should continue to work as before
+
+- Introducing new dataflow operators or changing existing ones -- any changes to operators will be addressed in future efforts as needed
+
+- Introducing type inheritance -- the Nextflow type system avoids type inheritance as much as possible in order to not introduce unnecessary complexity
+
+- Introducing object methods -- this can be handled by standalone functions for now, and may be improved in a future effort (e.g. namespaces)
+
+## Decision
+
+Add support for **records**, an alternative data structure to tuples, and **record types**, a way to declare custom types that can be applied to records a la carte.
+
+## Core Capabilities
+
+### Records
+
+Records are an attempt to take the best qualities of maps and custom classes while avoiding the downsides.
+
+A record can be created using the `record()` function:
+
+```groovy
+sample = record(
+    id: '1',
+    fastq_1: file('1_1.fastq'),
+    fastq_2: file('1_2.fastq')
+)
+
+sample.id = '2' // error: record cannot be modified
+sample += record(id: '2') // ok
+
+println sample.id
+```
+
+This function effectively creates an immutable map (`Map<String,?>`):
+
+- The keys are just field names
+- The values can have any type
+- The record can’t be modified -- use the `+` operator instead
+
+Records can have arbitrary fields, unlike custom classes, which makes them easy to use with dataflow operators.
+
+For example, a future version of the `join` operator could join records as follows:
+
+```groovy
+ch_bam = channel.of( record(id: '1', bam: file('1.bam')) )
+ch_bai = channel.of( record(id: '1', bai: file('1.bai')) )
+
+ch_bam.join(ch_bai, by: 'id').view()
+
+// -> record(id: '1', bam: file('1.bam'), bai: file('1.bai'))
+```
+
+Whereas tuples are joined by a matching index, records would be joined by a matching key.
+
+### Record types
+
+A record type is a user-defined type that consists of a name and a set of fields:
+
+```groovy
+record Sample {
+    id: String
+    fastq_1: Path
+    fastq_2: Path?
+}
+```
+
+Fields in a record type are declared the same way as [typed parameters](https://nextflow.io/docs/latest/workflow.html#typed-parameters). All [standard types](https://nextflow.io/docs/latest/reference/stdlib-types.html) can be used. Fields can be marked as optional by appending a `?` to the field type.
+
+The purpose of a record type is to specify a *minimum set of requirements* for a record *in a particular context*. A record created with the `record()` function simply has the type `Record` -- it makes no guarantees about which fields it provides. A *record type* can be used (e.g. in a workflow input) to make a stronger guarantee.
+
+For example:
+
+```groovy
+workflow RNASEQ {
+    take:
+    samples: Channel<Sample>
+
+    main:
+    // ...
+}
+
+workflow {
+    ch_samples = channel.of( record(id: '1', fastq_2: file('1_2.fastq')) )
+    RNASEQ(ch_samples) // error: `ch_samples` is missing `fastq_1` field required by Sample
+}
+```
+
+This workflow definition specifies that the `samples` input should be a channel of records, where each record has at minimum the fields specified by the `Sample` record type. The records can still have additional fields, but only the fields in `Sample` are guaranteed to be present.
+
+In other words, records are *duck-typed*. Duck-typing semantics are used whenever a record is validated against a record type:
+
+- Supplying a record argument to a workflow, process, or function with a record type input (as shown above)
+
+- Casting a record to a record type (e.g. `record(...) as Sample`)
+
+Record types can be included across modules:
+
+```groovy
+include { Sample } from './module'
+```
+
+Because of duck-typing, two record types with the same fields and field types are effectively equivalent, even if they have different names:
+
+```groovy
+record Sample {
+    id: String
+    fastq_1: Path
+    fastq_2: Path?
+}
+
+record FastqPair {
+    id: String
+    fastq_1: Path
+    fastq_2: Path?
+}
+```
+
+This makes it easier to compose modules and workflows that use their own record types.
+
+### Process inputs
+
+When a record is supplied as input to a process, the process needs to know how to stage input files from the record, like it does with the `path` qualifier.
+
+Typed processes can stage inputs using the `stage:` section, but ideally the files in a record should be automatically detected and staged.
+
+A typed process can declare a record using an *inline record type*:
+
+```groovy
+process FASTQC {
+    input:
+    sample: Record {
+        id: String
+        fastq_1: Path
+        fastq_2: Path
+    }
+
+    // ...
+}
+```
+
+All record fields that are a `Path` or `Path` collection (e.g. `Set<Path>`) are automatically staged. The record itself is declared in the process body as `sample`, like any other input, and record fields are accessed as `sample.id`, `sample.fastq_1`, and so on.
+
+A typed process can also use an explicit record type to achieve the same behavior:
+
+```groovy
+process FASTQC {
+    input:
+    sample: FastqPair
+
+    // ...
+}
+
+record FastqPair {
+    id: String
+    fastq_1: Path
+    fastq_2: Path
+}
+```
+
+The only difference between these two aprooaches is that the `FastqPair` type can be used elsewhere in pipeline code because it is declared externally.
+
+### Process outputs
+
+Typed processes can declare outputs with arbitrary expressions, so no new syntax is required to support record outputs. Simply use the `record()` function to create a record:
+
+```groovy
+process FASTQC {
+  // ...
+
+  output:
+  record(id: id, fastqc: file('fastqc_logs'))
+
+  // ...
+}
+```
+
+The type of this output is an *implicit* record type that is inferred from the code: `Record { id: String ; fastqc: Path }`.
+
+## Alternatives
+
+### Custom classes
+
+Define Groovy-style classes (see #2085) and use them to model composite data:
+
+```groovy
+@nextflow.io.ValueObject
+class Sample {
+    String id
+    Path fastq_1
+    Path fastq_2
+}
+
+workflow {
+    sample = new Sample('1', file('1_1.fastq'), file('1_2.fastq'))
+
+    println sample.id
+}
+```
+
+This approach can be used, but in practice it requires a lot of extra dataflow logic around process calls to convert between custom types and tuples, because processes don’t know how to stage input files from custom types.
+
+Custom classes are not very flexible. For example, joining two channels of custom classes would be more complicated than joining two tuples by a matching key, because you would need to define an additional class for the “joined” type and explicitly construct it from the two joining classes.
+
+### Maps
+
+Use maps to create composite data structures dynamically (see #2127):
+
+```groovy
+sample = [
+    id: '1',
+    fastq_1: file('1_1.fastq'),
+    fastq_2: file('1_2.fastq')
+]
+
+println sample.id
+```
+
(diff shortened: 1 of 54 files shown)


bentsherman:
This PR does not make any changes to operators, so you need some extra logic in some cases to pass records through an operator.

For example, given two record channels that you want to join:
```groovy
samples_ch = fastqc_ch.join(quant_ch, by: 'id')
```

The above syntax is what we'd like to do, but `join` doesn't work with records yet, so you have to wrap the records as tuples before the join and unwrap after the join:

```groovy
samples_ch = fastqc_ch.map { r -> tuple(r.id, r) }
    .join( quant_ch.map { r -> tuple(r.id, r) } )
    .map { id, r1, r2 -> r1 + r2 }
```

docs/tutorials/records.md
```diff
@@ -0,0 +1,325 @@
+(migrating-records)=
+
+# Migrating to records
+
+Nextflow 26.04 introduces the ability to use *records* and *record types* in a Nextflow pipeline. This tutorial demonstrates how to migrate to records using the [rnaseq-nf](https://github.com/nextflow-io/rnaseq-nf) pipeline as an example.
```
christopher-hakkaart:
```suggestion
Nextflow 26.04 introduces support for *records* and *record types* in a Nextflow pipeline. This tutorial demonstrates how to migrate to records using the [rnaseq-nf](https://github.com/nextflow-io/rnaseq-nf) pipeline as an example.
```

docs/tutorials/records.md
```diff
@@ -0,0 +1,325 @@
...
+
+```nextflow
+sample = tuple([id: '1'], file('1_1.fastq'), file('1_2.fastq'))
+
+channel.of(sample).view { meta, fastq_1, fastq_2 ->
+    "id: ${meta.id}, fastq_1: ${fastq_1.name}, fastq_2: ${fastq_2.name}"
+}
+```
+
+Tuples are easy to use, but have several downsides:
```
christopher-hakkaart:
```suggestion
Tuples are easy to use, but have several limitations:
```
Suggestion only

docs/tutorials/records.md
```diff
@@ -0,0 +1,325 @@
...
+
+- **Brittle data structures**: Tuples must be repeatedly unpacked and repacked whenever they pass through a channel operator. The tuple can instead be passed as a single argument, in which case tuple elements must be accessed by index. Either way, the user has to remember the names and ordering of the tuple elements.
```
christopher-hakkaart:
```suggestion
- **Brittle data structures**: Tuples must be repeatedly unpacked and repacked whenever they pass through a channel operator. If the tuple is passed as a single argument, the tuple elements must be accessed by index. In both cases, the user must remember the names and the order of the tuple elements.
```

docs/tutorials/records.md
```diff
@@ -0,0 +1,325 @@
...
+
+- **Tight coupling between processes and workflow logic**: When calling a process, the tuple elements must match the process inputs with the same ordering. Getting this wrong can lead to runtime errors that are hard to debug.
+
+- **Limited type checking**: The type checker can validate whether an incoming tuple has the right number of elements and the right element types, but it cannot validate the semantic name of each element, because tuple elements don't have names. For example, the type `Tuple<String,Path,Path>` does not tell you what any of the elements refer to, only their types. As a result, there are many kinds of semantic errors that cannot be caught by the type checker.
```
christopher-hakkaart:
```suggestion
- **Limited type checking**: The type checker can validate whether an incoming tuple has the right number of elements and the right element types. However, it cannot validate the semantic name of each element, as tuple elements don't have names. For example, the type `Tuple<String,Path,Path>` tells you nothing about what any of the elements refer to, only their types. As a result, there are many kinds of semantic errors that cannot be caught by the type checker.
```

docs/tutorials/records.md
```diff
@@ -0,0 +1,325 @@
...
+
+Records are a new data type for modelling complex data structures, as an alternative to tuples:
+
+- **Meaningful data structures**: Record fields are accessed by name instead of index, allowing you to think about your data in terms of meaningful names instead of arbitrary indices.
```
christopher-hakkaart:
```suggestion
- **Meaningful data structures**: Record fields are accessed by name instead of index. They allow you to think about your data in terms of meaningful names instead of arbitrary indices.
```

docs/tutorials/records.md
```diff
@@ -0,0 +1,325 @@
...
+
+- **Flexible workflow logic**: Records are *duck-typed*, which means that a record can be used as an input as long as it meets the minimum requirements of that input. This makes it much easier to keep related data together in a workflow without requiring additional logic to translate between channels and process inputs/outputs.
```
christopher-hakkaart:
```suggestion
- **Flexible workflow logic**: Records are *duck-typed*, which means that a record can be used as an input as long as it meets the minimum requirements of that input. This makes it easier to keep related data together in a workflow without requiring additional logic to translate between channels and process inputs/outputs.
```

docs/tutorials/records.md
```diff
@@ -0,0 +1,325 @@
...
+
+- **Robust type checking**: The type checker can validate workflow logic much more thoroughly when using records. For example, it can validate whether an incoming record has all field names and types required by a process input.
+
+Records allow you to model your data more precisely as it flows through your pipeline. It allows your pipeline code to focus on what matters, and not worry about things that don't matter, making it easier to read and understand.
```
christopher-hakkaart:
```suggestion
Records allow you to model your data more precisely as it flows through your pipeline.
```
IMO - Second part was more promotional

docs/tutorials/records.md
```diff
@@ -0,0 +1,325 @@
...
+
+Records and record types work best with the [Nextflow language server](https://github.com/nextflow-io/language-server) and [Nextflow VS Code extension](https://marketplace.visualstudio.com/items?itemName=nextflow.nextflow).
+
+:::{tip}
+See {ref}`devenv-page` for instructions on how to setup VS Code and the Nextflow extension.
+:::
+
+### Type checking
+
+The language server can check your code for type-related errors, which is especially useful when using records. For example, it can validate that a channel of records has all required fields when it is passed as input to a process.
```
christopher-hakkaart:
```suggestion
The language server can check your code for type-related errors. For example, it can validate that a channel of records has all the required fields when it is passed as input to a process.
```

docs/tutorials/records.md
```diff
@@ -0,0 +1,325 @@
...
+
+### Automatic migration
+
+The **Convert script to static types** command described in {ref}`migrating-static-types` automatically converts process inputs and outputs to records when the language version is set to 26.04.
```
christopher-hakkaart:
```suggestion
The **Convert script to static types** command automatically converts process inputs and outputs to records when the language version is set to 26.04. See {ref}`migrating-static-types` for more information.
```

docs/tutorials/records.md
```diff
@@ -0,0 +1,325 @@
...
+    output:
+    record(
+        meta: sample.meta,
+        bam: file('*.bam'),
+        bai: file('*.bai')
+    )
+}
+```
+
+The outputs must satisfy the following conditions in order to be converted:
```
christopher-hakkaart:
```suggestion
The outputs must satisfy the following conditions to be converted:
```

docs/tutorials/records.md
```diff
@@ -0,0 +1,325 @@
...
+    quant_ch = QUANT(reads_ch, index)
+    samples_ch = fastqc_ch.join(quant_ch)
+
+    emit:
+    samples : Channel<Tuple<String,Path,Path>> = samples_ch
+    index   : Value<Path> = index
+}
+```
+
+The `reads_ch` input needs to be refactored as a channel of records. While we could use `Channel<Record>`, it is better to use a *record type* that specifies the required fields. To do this, we need to examine the processes that use `ch_reads` and construct a record type that encapsulates those uses.
```
christopher-hakkaart:
```suggestion
The `reads_ch` input needs to be refactored as a channel of records. While you could use `Channel<Record>`, it is better to use a *record type* that specifies the required fields. To do this, you need to examine the processes that use `ch_reads` and construct a record type that encapsulates those uses.
```

docs/tutorials/records.md
```diff
@@ -0,0 +1,325 @@
...
+
+The `reads_ch` input is used by `FASTQC` and `QUANT`, which both have the following input declaration:
+
+```nextflow
+    input:
+    (id, fastq_1, fastq_2): Tuple<String, Path, Path>
+```
+
+Therefore, we can construct a record type that models these requirements. Update the `reads_ch` input as follows:
```
christopher-hakkaart:
```suggestion
Therefore, you can construct a record type that models these requirements. Update the `reads_ch` input as follows:
```

modules/nf-commons/src/main/nextflow/util/RecordMap.java
```diff
@@ -0,0 +1,67 @@
...
+
+    // @Override
+    // public Object put(String key, Object value) {
+    //     throw new UnsupportedOperationException();
+    // }
+
+    // @Override
+    // public void putAll(Map<? extends String, ? extends Object> m) {
+    //     throw new UnsupportedOperationException();
+    // }
```
pditommaso: The Javadoc correctly states records should be immutable, but `put()` and `putAll()` are commented out, making `RecordMap` fully mutable. Since records flow through shared channels across processes, this is a data race risk.

Uncomment the `put()` and `putAll()` overrides to enforce immutability. If mutability is needed during construction, consider using a builder pattern or populating via the `Map` constructor, then having all mutation methods throw `UnsupportedOperationException`.

modules/nf-lang/src/main/java/nextflow/script/parser/ScriptAstBuilder.java
```diff
@@ -496,40 +531,79 @@ private Parameter processInput(ProcessInputContext ctx) {
...
+        var recordNode = ast( new RecordNode(nextRecordName()), ctx );
+        if( ctx.recordBody() != null )
+            recordBody(ctx.recordBody(), recordNode);
+        else
+            collectSyntaxError(new SyntaxException("Missing record body", recordNode));
+
+        return ast( new Parameter(recordNode.getPlainNodeReference(), name), ctx );
+    }
+
+    private static AtomicInteger nextRecordId = new AtomicInteger(1);
```
pditommaso, line 573: `nextRecordId` is a `static AtomicInteger` shared across all `ScriptAstBuilder` instances. In a long-running process (language server, test suite), the counter never resets and produces different anonymous record names (`__Record_1`, `__Record_2`, ...) depending on compilation order. This could affect caching or reproducibility.

Consider making this an instance field, or including the source file/module name in the generated name to avoid cross-module collisions and ensure deterministic naming.

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -1822,16 +1844,25 @@ class TaskProcessor {
...
         }
         task.context.put(param.getName(), value)
         task.setInput(param, value)
     }
 
+    private static boolean isAssignableFrom(Class targetType, Class sourceType) {
+        if( Record.class.isAssignableFrom(targetType) && Record.class.isAssignableFrom(sourceType) )
+            return true
+        return targetType.isAssignableFrom(sourceType)
+    }
```
pditommaso, line 1866: This treats *any* `Record` as assignable to *any* other `Record`, completely bypassing field-level validation. A record with `{id: String}` will silently pass where `{id: String, fastq: Path}` is expected — the error only surfaces later when `sample.fastq` is accessed.

Consider at least validating that the source record contains all field names expected by the target type, and logging a warning for missing fields.

modules/nextflow/src/main/groovy/nextflow/processor/TaskInputResolver.groovy
```diff
@@ -89,17 +89,40 @@ class TaskInputResolver {
...
+        if( value instanceof Path ) {
+            return normalizePath(value, holders)
+        }
+
+        if( value instanceof Collection ) {
+            return value.collect { el -> normalizeValue(el, holders) }
+        }
+
+        if( value instanceof Map ) {
+            return value.collectEntries { k, v -> [k, normalizeValue(v, holders)] }
```
pditommaso: When `value instanceof Map` (which includes `RecordMap`), `collectEntries` returns a plain `LinkedHashMap`, losing the `RecordMap` type. After path normalization, the value is no longer a `Record` instance.

Consider handling `RecordMap` explicitly before the generic `Map` branch:
```groovy
if( value instanceof RecordMap ) {
    return new RecordMap(value.collectEntries { k, v -> [k, normalizeValue(v, holders)] })
}
```

modules/nextflow/src/main/groovy/nextflow/script/TypeDef.groovy
```diff
@@ -0,0 +1,64 @@
...
+
+    @Override
+    String getType() { 'type' }
+
+    @Override
+    String getName() { alias }
+
+    @Override
+    Object invoke_a(Object[] args) {
+        throw new UnsupportedOperationException()
```
pditommaso, line 50: `invoke_a()` throws a bare `UnsupportedOperationException` with no message. If a user accidentally invokes a type definition, they'll get an unhelpful stack trace.

Suggest adding a descriptive message:
```groovy
throw new UnsupportedOperationException("Type '${alias}' cannot be invoked as a function")
```

pditommaso (commented):
Code review of the record types feature. Overall well-structured implementation following existing codebase patterns. Flagging 5 issues ranging from a mutability/data-race risk to minor UX improvements.

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
bentsherman: These record types are validated by the type checker, which is currently only available in the language server.

I looked into adding a basic runtime check for the field names, but the problem is that the information about nullable fields is not available at this point, so a runtime check would produce many false warnings

So I think the best we can do is to rely on the language server for now and later move the type checker into nf-lang. Ideally before 26.10.

modules/nextflow/src/main/groovy/nextflow/script/TypeDef.groovy
bentsherman: This case is also validated at compile-time, so `TypeDef::invoke_a()` is effectively unreachable from Nextflow code. If it could be reached then I would consider that a bug. I added a test to `ScriptTypesTest` to validate this case

modules/nf-commons/src/main/nextflow/util/RecordMap.java
bentsherman: I think I commented these temporarily because there were some places in the codebase that were automatically treating them like mutable maps. Better to uncomment these and flush out those corner cases...

modules/nf-lang/src/main/java/nextflow/script/parser/ScriptAstBuilder.java
bentsherman: These anonymous record names should not affect resumability, since only the record values are used for hashing. And I doubt even a long-running language server session would reach the ~2 billion parses required to overload the counter 😅

Still, it should be feasible to just make it an instance variable. I will try this

adr/20260306-record-types.md
```diff
@@ -0,0 +1,435 @@
...
+
+process BAR {
+
+    input:
+    sample: Record {
+        meta: Map
+        fna: Path
+        faa: Path
+        tbl: Path
+    }
```
muffato, line 415: Hi. Should `meta` be added to the output record of `PROKKA` for it to be compatible with `FOO` and `BAR` ? The Google doc consistently had `id` across all PROKKA input, PROKKA output, FOO input and BAR input.

bentsherman: Good catch. I forgot to replace id with meta in the prokka output here. I will fix it in a subsequent PR

pditommaso:
Does this change the hashing irrespective the use of record types? if so could be confined only when the feature is enabled?

bentsherman:
The hashing change was to hash maps by their keys and values instead of only their values

It made more sense to apply this change for all records and maps because it fixes a bug (#4916) where e.g. a meta-map key would be renamed but the task wouldn't re-execute

pditommaso:
Considering the impact breaking the cache in production deployment it may be worth considering making this configurable. Default should use the new version, however it could be possible to back to the previous caching. it should be fairly simple to have a caching strategy as a singleton in the session or process level.

bentsherman:
I'm not sure that anyone is asking for that, but if we were to do anything about it, I would do something like #6353 where we add an environment variable to the `STABLE/26.04.x` branch after it is released, so that it doesn't clutter the codebase going forward

Issue #6684, opened 2025-12-23
Files emitted to topic channels clobber process outputs (strict syntax only)


stekaz:
## Bug report 

Using the new strict syntax, files emitted to a topic channel seem to clobber files declared in the output section.

```
nextflow.preview.types = true

process TEST {

    output:
    vcf: Path = file("file.vcf.gz")
    tbi: Path = file("file.vcf.gz.tbi")

    topic:
    file("foo.txt") >> 'foo'

    script:
    """
    touch file.vcf.gz{,.tbi}
    touch foo.txt
    """
}

workflow {

    TEST()

    TEST.out.vcf.view { v -> "vcf: ${v}" }
    TEST.out.tbi.view { v -> "tbi: ${v}" }
}
```

### Program output 

```
 N E X T F L O W   ~  version 25.12.0-edge

Launching `main.nf` [elegant_hopper] DSL2 - revision: 2d0edb6005

executor >  local (1)
[ae/253dde] process > TEST [100%] 1 of 1 ✔
vcf: /path/to/work/ae/253ddee30c7eb89c4c4f5a07d80720/foo.txt
tbi: /path/to/work/ae/253ddee30c7eb89c4c4f5a07d80720/file.vcf.gz.tbi
```

### Expected output

```
vcf: /path/to/work/ae/253ddee30c7eb89c4c4f5a07d80720/file.vcf.gz
tbi: /path/to/work/ae/253ddee30c7eb89c4c4f5a07d80720/file.vcf.gz.tbi
```

### Environment 

* Nextflow version: 25.12.0-edge
* Java version: openjdk 17.0.17 2025-10-21
* Operating system: Linux
* Bash version: GNU bash, version 5.3.3(1)-release (x86_64-pc-linux-gnu)

PR #6688, opened 2025-12-27
Fix strict syntax issue with files emitted to topic channels

stekaz:
Fixes [#6684](https://github.com/nextflow-io/nextflow/issues/6684)


bentsherman (approved):

bentsherman:
Thanks @stekaz for the fix, and good catch. We'll try to get this backported into 25.10

PR #6689, opened 2025-12-28
Add workflow manifest metadata to execution report

### LINKED ISSUE #3917: Include workflow and script version in execution report
## New feature

Hi! As talked with @ewels over nextflow.slack.com, the current template for the file `report.html` does not include information on which pipeline or workflow version was used to produce the data. 

The report itself offers, dates, nextflow version and depending on how the Docker images are labelled, also their version control, however not the current pipeline version. From a QA perspective, especially regarding in-house development, it would be better to add the pipeline version or workflow itself there.

## Usage scenario 

General scenario. In terms of QA and tracablity, NF version, pipeline version and Docker image versions should appear on `report.html`.

## Suggest implementation 
On `modules/nextflow/src/main/resources/nextflow/trace/ReportTemplate.html`, add `workflow.manifest.version` between `workflow.scriptName` and `workflow.scriptId`.

```html
<% if (workflow.scriptName) { %>
          <dt class="col-sm-3">Script name</dt>
          <dd class="col-sm-9"><samp>${workflow.scriptName}</samp></dd>
        <% } %>

<% if (workflow.manifest.version) { %>
          <dt class="col-sm-3">Script version</dt>
          <dd class="col-sm-9"><samp>${workflow.manifest.version}</samp></dd>
        <% } %>

        <% if (workflow.scriptId) { %>
          <dt class="col-sm-3">Script ID</dt>
          <dd class="col-sm-9"><code>${workflow.scriptId}</code></dd>
        <% } %>
```

ewels:
Include workflow manifest fields in the HTML execution report for
better traceability and QA purposes:
- Workflow name (manifest.name)
- Workflow version (manifest.version)
- Workflow description (manifest.description)
- Workflow author (manifest.author)
- Workflow home page (manifest.homePage) - as clickable link
- Workflow DOI (manifest.doi) - as doi.org link

All fields are conditionally displayed only when set in the manifest.
This complements the existing repository/revision/commitId information.

Closes #3917

Signed-off-by: Claude <noreply@anthropic.com>

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/script/parser/v2/ScriptLoaderV2Test.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/script/parser/v2/ScriptLoaderV2Test.groovy
@@ -254,4 +254,45 @@ class ScriptLoaderV2Test extends Dsl2Spec {
         noExceptionThrown()
     }
 
+    def 'should register outputs and topic emissions' () {
+
+        given:
+        def session = new Session()
+        def parser = new ScriptLoaderV2(session)
+
+        def TEXT = '''
+
+            nextflow.preview.types = true
+
+            process hello {
+
+                output:
+                out1: Path = file("out1.txt")
+                out2: Path = file("out2.txt")
+
+                topic:
+                file("version.txt") >> 'versions'
+
+                script:
+                """
+                echo "version" > version.txt
+                echo "result1" > out1.txt
+                echo "result2" > out2.txt
+                """
+            }
+            '''
+
+        when:
+        parser.setModule(true)
+        parser.parse(TEXT)
+        parser.runScript()
+        def process = ScriptMeta.get(parser.getScript()).getProcess('hello')
+        def outputs = process.getProcessConfig().getOutputs()
+
+        then:
+        outputs.getParams().size() == 2
+        outputs.getTopics().size() == 1
+        outputs.getFiles().size() == 3
+    }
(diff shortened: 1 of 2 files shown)


pditommaso (changes requested):
Oops, didn't notice tests are failing

ewels:
Oops, good catch by the tests! Now fixed..

pditommaso (approved):

modules/nextflow/src/main/resources/nextflow/trace/ReportTemplate.html
```diff
@@ -177,6 +177,36 @@ <h4>Workflow execution completed unsuccessfully!</h4>
           <dd class="col-sm-9"><code>${workflow.scriptId}</code></dd>
         <% } %>
 
+        <% if (workflow.manifest?.name) { %>
```
bentsherman: It looks like the `?.` was added to satisfy an existing test. But if the manifest is expected to be present then the test should be updated to define the manifest

ewels: But people can specify pipeline configs without a manifest section, right?

Or does Nextflow set some part of it - ie. it's always set?

ewels: Ah you're right, it's always defined. Ok, I'll update the tests.

PR #6696, opened 2025-12-31
Fix caching of task ext properties with the strict parser

### LINKED ISSUE #6695: Strict parser doesn't add ext properties to task hash
## Bug report 

Minimal example:

```groovy
// main.nf

process EXAMPLE_EXT {

    script:
    """
    echo "args = ${task.ext.args}"
    """
}

workflow {
    EXAMPLE_EXT()
}
```

```groovy
// nextflow.config

process {
    withName: 'EXAMPLE_EXT' {
        ext.args = "--some-param value"
    }
}
```

Nextflow should add the value of `task.ext.args` to the task hash, so that changing this value re-executes the task on resume, just like changing a regular input.

However, Nextflow 25.10 with the strict parser does not do this.

I believe it's because the strict parser doesn't apply the VariableVisitor, which collects all variable references in the process body. The task processor gets the task ext references from this list.

I think I originally neglected this piece because I wanted to discourage the use of external variables in the process body. The strict parser already disallows most global variables in the process body. However, since we haven't validated any good replacement for task ext, I think we should just support it for now.

bentsherman:
Close #6695 

This PR updates the strict parser to collect references to `task.ext.*` so that they are added to the task hash ([docs](https://nextflow.io/docs/latest/cache-and-resume.html#task-hash)).

I originally thought to move the `VariableVisitor` into `nf-lang` so that it is re-used by both, but I think this visitor does a lot of unnecessary work, and I didn't want to alter the old parser too much, so I just created a separate implementation for the strict parser that is much simpler. It only collects references to `task.ext.*` instead of all variable references.

_Note to self:_ once we make the strict parser the default (#6643), we should revisit the runtime layer and how it handles these variable references. I think it could be simplified quite a bit.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskHandler.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskHandler.groovy
@@ -221,6 +221,8 @@ abstract class TaskHandler {
         record.env = task.getEnvironmentStr()
         record.executorName = task.processor.executor.getName()
         record.containerMeta = task.containerMeta()
+        record.accelerator = task.config.getAccelerator()?.request
+        record.accelerator_type = task.config.getAccelerator()?.type
 
         if( isCompleted() ) {
             record.error_action = task.errorAction?.toString()
--- a/modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
@@ -103,7 +103,9 @@ class TraceRecord implements Serializable {
             vol_ctxt: 'num',        // -- /proc/$pid/status field 'voluntary_ctxt_switches'
             inv_ctxt: 'num',        // -- /proc/$pid/status field 'nonvoluntary_ctxt_switches'
             hostname: 'str',
-            cpu_model:  'str'
+            cpu_model:  'str',
+            accelerator: 'num',
+            accelerator_type: 'str'
     ]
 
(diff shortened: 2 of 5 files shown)


modules/nf-lang/src/main/java/nextflow/script/control/ScriptToGroovyHelper.java
```diff
@@ -41,6 +43,76 @@ public ScriptToGroovyHelper(SourceUnit sourceUnit) {
...
+                builder.append('.');
+                builder.append(pe.getPropertyAsString());
+                target = pe.getObjectExpression();
+            }
+
+            builder.append('.');
+            builder.append(target.getText());
+
+            return builder.reverse().toString();
+        }
```
jorgee, line 114: The reverse in last statement is reversing the full stringbuilder getting something like `ksat.txe.sgra`. I think you wanted to just reverse the order of elements of a list of strings and join with `.` later.
```suggestion
        private static String asPropertyChain(PropertyExpression node) {
            var list = new ArrayList<String>();
            list.add(node.getPropertyAsString());

            var target = node.getObjectExpression();
            while( target instanceof PropertyExpression pe ) {
                list.add(pe.getPropertyAsString());
                target = pe.getObjectExpression();
            }
            list.add(target.getText());
            Collections.reverse(list);
            return String.join(".", list);
        }
```

bentsherman: Thanks Jorge, good catch

PR #6698, opened 2026-01-02
Add missing provider credentials when updating the bare repo in multi-revision assets

jorgee:
When updating an existing local asset of a repository requiring access credentials, it is failing because the fetch command doesn't contain the provider credentials.
This pull request adds the missing repository credentials during the fetch operation to update the bare repository instance in the `MultiRevisionRepositoryStrategy` class.


pditommaso (commented):
Is there any way to capture this condition into a small test?

jorgee:
I have added the download of the Azure repo as a test. It was the only one that really required credentials to read it.

modules/nextflow/src/main/groovy/nextflow/scm/MultiRevisionRepositoryStrategy.groovy
```diff
@@ -177,7 +177,11 @@ class MultiRevisionRepositoryStrategy extends AbstractRepositoryStrategy {
         }
         final updateRevision = revision ?: getDefaultBranch(manifest)
         log.debug "Fetching (updating) bare repo for ${project} [revision: $updateRevision]"
-        getBareGit().fetch().setRefSpecs(refSpecForName(updateRevision)).call()
+        def fetch = getBareGit().fetch().setRefSpecs(refSpecForName(updateRevision))
```
pditommaso:
```suggestion
        final fetch = getBareGit().fetch().setRefSpecs(refSpecForName(updateRevision))
```

robsyme:
Heya team. Would this be a good candidate to backport to 25.10? 
Edit nevermind - this request is not as pressing as I had imagined.

jorgee:
If I am not wrong, 25.10.x does not have multi-revision. It only affects latest edge releases.

PR #6700, opened 2026-01-02
Add aws.batch.forceGlacierTransfer config option

### LINKED ISSUE #4747: Recursively downloading S3 directory restored from Glacier doesn't work
## New feature

For rnasplice pipeline, I wanted to provide my previous salmon output as an input as suggested [here](https://nf-co.re/rnasplice/1.0.2/docs/usage#salmon-quantification-files)

So I prepared input sheet like this:

sample,condition,salmon_results
xyz,case,s3://bucket/path/salmon/xyz

where xyz directory is in Glacier storage class but has been restored

xyz/
├── aux_info
│   ├── ambig_info.tsv
│   ├── expected_bias.gz
│   ├── fld.gz
│   ├── meta_info.json
│   ├── observed_bias_3p.gz
│   └── observed_bias.gz
├── cmd_info.json
├── lib_format_counts.json
├── libParams
│   └── flenDist.txt
├── logs
│   └── salmon_quant.log
├── quant.genes.sf
└── quant.sf

Each of the file inside xyz has been individually restored and has been tested for download. It does not require --force-glacier-transfer option.

But my sample-sheet above forces nxf_s3_download() function to attempt recursive download on xyz directory which is not downloadable without specifying --force-glacier-transfer option

Example:
aws s3 cp s3://bucket/path/salmon/xyz ./xyz/ --recursive does not work
Error: 
warning: Skipping file s3://bucket/path/salmon/xyz/aux_info/ambig_info.tsv. Object is of storage class GLACIER. Unable to perform download operations on GLACIER objects. You must restore the object to be able to perform the operation. See aws s3 download help for additional parameter options to ignore or force these transfers

aws s3 cp s3://bucket/path/salmon/xyz ./xyz/ --recursive  --force-glacier-transfer works

## Usage scenario 

Provide restored glacier directory paths as input

## Suggest implementation 

Similar to a change made [here](https://github.com/nextflow-io/nextflow/commit/a6a0532e3ffa20326b319455152439e1f32c52a2) , --quiet was changed to --only-show-errors , --force-glacier-transfer option could be made a default in case the s3 object is not in standard storage

Or, 
Introduce an optional setting similar to aws.client.glacierAutoRetrieval, say aws.client.glacierForceTransfer, wihch when true will add --force-glacier-transfer to nxf_s3_download() function

robsyme:
## Summary

- Add `aws.batch.forceGlacierTransfer` configuration option to enable `--force-glacier-transfer` flag for AWS CLI S3 download commands
- Clean up trailing whitespace in S3BashLib

## Problem

When Nextflow stages input files from S3 directories containing restored Glacier/Deep Archive objects, AWS CLI skips them even though they are fully accessible. This is because AWS CLI checks the storage class metadata (which remains `GLACIER` or `DEEP_ARCHIVE` even after restoration) and refuses to download unless `--force-glacier-transfer` is specified.

Error message from AWS CLI:
```
warning: Skipping file s3://bucket/path/file.tif. Object is of storage class GLACIER.
Unable to perform download operations on GLACIER objects.
```

## Solution

Add a new config option:
```groovy
aws {
    batch {
        forceGlacierTransfer = true
    }
}
```

When enabled, the `nxf_s3_download` bash function includes `--force-glacier-transfer` in `aws s3 cp` commands, allowing restored Glacier objects to be downloaded.

## Test plan

- [x] Added unit test for new config option in `AwsBatchConfigTest`
- [x] Added unit test for glacier transfer flag in `S3BashLibTest`
- [x] All existing tests pass

Closes #4747


robsyme:
@claude Review

robsyme:
In response to the performance consideration raided by @claude:

From what I can see, when enabling the flag for standard storage class objects, the flag is effectively a no-op - the transfer proceeds normally either way. There are:
- No extra API calls - it doesn't call HeadObject to verify restoration status
- No extra processing - just bypasses a conditional check
- No cost difference - same S3 GET requests either way

blajoie:
happy to help test as soon as you can point us to a nextflow dev version ! thanks @robsyme  !

plugins/nf-amazon/src/main/nextflow/cloud/aws/config/AwsBatchConfig.groovy
```diff
@@ -125,6 +125,12 @@ class AwsBatchConfig implements CloudTransferOptions, ConfigScope {
     """)
     final List<String> volumes
 
+    @ConfigOption
+    @Description("""
+        When `true`, the `--force-glacier-transfer` flag is added to AWS CLI S3 download commands, allowing restored Glacier/Deep Archive objects to be downloaded. Only needed when staging files that have been restored from archive storage (default: `false`).
```
pditommaso: It maybe useful to stress this flag does not initiate restore process from Glacier, which needs to be managed independently

pditommaso (approved):

plugins/nf-amazon/src/main/nextflow/cloud/aws/util/S3BashLib.groovy
```diff
@@ -126,9 +132,9 @@ class S3BashLib extends BashFunLib<S3BashLib> {
             local file_name=\$(basename \$1)
             local is_dir=\$($cli s3 ls \$source | grep -F "PRE \${file_name}/" -c)
             if [[ \$is_dir == 1 ]]; then
-                $cli s3 cp --only-show-errors --recursive "\$source" "\$target"
-            else 
-                $cli s3 cp --only-show-errors "\$source" "\$target"
+                $cli s3 cp --only-show-errors --recursive ${forceGlacierTransfer}"\$source" "\$target"
+            else
+                $cli s3 cp --only-show-errors ${forceGlacierTransfer}"\$source" "\$target"
```
bentsherman: The original issue says that this flag is only needed for directories, so we don't need to specify it here. I suspect it incurs some minor overhead / extra API call so I'm inclined to only specify it for directories.

robsyme: Great catch. Correct.

Issue #6680, opened 2025-12-19
Enrich FilePublishEvent with source context


robsyme:
## New feature

Include source context (process name or output name) and channel value in FilePublishEvent, allowing observers to filter and access metadata without manual event correlation.

## Use case

When building plugins that notify external systems about published files (e.g., triggering an API when BAM files are ready), the notification typically needs:

1. The published file path
2. A selector to filter which publications matter (process name or output name)
3. Associated metadata (sample ID, patient info, etc.)

Currently, FilePublishEvent only provides source, target, and labels. Getting the selector or metadata requires complex workarounds:

For publishDir files:
- Track task submissions/completions via onTaskSubmit/onTaskComplete
- Extract task hash by parsing the work directory path structure
- Maintain hash → metadata maps with synchronization for race conditions

For workflow output files:
- Buffer FilePublishEvents until onWorkflowOutput fires
- Retrospectively correlate files to channel values
- No direct link between the two event types

This results in ~400 lines of boilerplate for what could be a simple notification plugin.

## Suggested implementation

Both publishDir and workflow outputs have a natural selector and associated channel value:

| Source          | Selector                  | Channel Value           |
|-----------------|---------------------------|-------------------------|
| publishDir      | Process name (ALIGNMENT)  | Task output tuple       |
| Workflow output | Output name ("samples" in [docs example](https://www.nextflow.io/docs/latest/workflow.html#workflow-outputs)) | Published channel value |

Extend FilePublishEvent to include this context:
```groovy
class FilePublishEvent {
    Path source
    Path target
    List<String> labels

    // Source context:
    PublishSource sourceType   // PROCESS or WORKFLOW_OUTPUT
    String sourceName          // "ALIGNMENT" or "alignedBams"

    // The channel value containing this file:
    Object value
}

enum PublishSource {
    PROCESS,
    WORKFLOW_OUTPUT
}
```

### Multi-file channel values
Nextflow channels naturally group related items together - a BAM file with its index, paired-end reads, or a sample's outputs across multiple formats. This co-location is semantically meaningful: items in the same emission are related.

When a channel emission contains multiple files, each file currently generates its own event:                                                                                                                                                        

```nextflow
workflow {
  main:
    // science goes here
  publish:
    alignedBams = ALIGNMENT.out.map { meta, bam, bai ->
        meta + [bamFile: bam, index: bai]
    }
}
```

[...]

PR #6701, opened 2026-01-03
Add onWorkflowPublish event to TraceObserverV2

robsyme:
## Summary

Adds a new `onWorkflowPublish` event to `TraceObserverV2` that fires for each value as it is published to a workflow output.

This addresses the request in #6680 for richer context during file publishing. The new event provides:
- `name` - the workflow output name (e.g., "alignedBams")
- `value` - the complete channel value being published (which may contain files alongside metadata like sample IDs)

### Event timeline

Channel value 1 arrives → onWorkflowPublish  ← NEW (fires per value)
                        → onFilePublish (for each file)
Channel value 2 arrives → onWorkflowPublish  ← NEW
                        → onFilePublish (for each file)
...
All values complete     → onWorkflowOutput   (fires once at end)

Unlike `onWorkflowOutput` which fires after all values have been published, `onWorkflowPublish` fires in real-time as each value flows through, giving
1) access to the published values as items in a channel, preserving the channel structure (items in the same channel element are together), and
2) observers immediate access to the complete channel emission.

## Test plan

- [x] Existing `OutputDslTest` tests updated to verify `notifyWorkflowPublish` is called
- [x] Smoke tests pass


bentsherman:
I think this is the right idea

I'm wondering whether we should only fire this event for queue channels and not value channels. Since a value channel is just a single value, the `onWorkflowPublish` and `onWorkflowOutput` are identical, and if you use both events, you'll have to be careful to not do something twice. But depending on the use case, it may not matter

pditommaso (dismissed):
as a developer I wonder what's the difference of `onWorkflowOutput` vs `onWorkflowPublish` vs `onFilePublish`. 

At least there should be a very detailed docs and javadoc describing what event is meant for and when to use it

bentsherman:
I updated the java docs to be more comprehensive. I will follow up with a guide in the docs for how to use the trace observer

robsyme:
Amazing. Thanks team ❤️ 🙏

pditommaso:
Team, I've reverted this PR from master. This change introduces a breaking change with existing plugins (avoiding which was one of the main reasons observer v2 was introduced).

We could, of course, introduce this kind of change, but it would be preferable to keep as a last resort.
At the same time, it looks to me that the newly introduced event handler was very similar to onWorkflowOutput, or even that it achieved similar behaviour using onFilePublish (though not optimal).

I want to push further on thinking about how the requested use case can be achieved without introducing a new event handler.


(related 06cd26072db742b515614740db85aaf2c9c9662f)

robsyme:
Apologies. Is the incompatibility that plugins without the new method no longer adhere to the expanded interface?

pditommaso:
No problem! Yeah, It's a kind of Groovy limitation, because Java interfaces default method is designed to avoid that. 

Apart this, would not your use case fit into `WorkflowOutputEvent`? we talked several times to have some kind pluggable output backend. Would your case be an example of that ?

bentsherman:
@pditommaso the rationale for this new event is that it allows you to respond to published values more quickly. For example, when publishing a channel of BAM/BAI samples, it allows you to react to each sample as it is published, whereas `onWorkflowOutput` triggers only after all samples have been published

Can you elaborate on how this change was breaking? If adding a new event with default impl is breaking, then you are implying that we can't add to TraceObserverV2 at all...

Also, TraceObserverV2 is annotated with `@Beta`, which I would assume to mean that breaking changes are still allowed

bentsherman:
Since this event was a performance enhancement rather than enabling new functionality, we aren't going to worry about it right now. But I would like to investigate the compatibility issues as I think it should be possible to avoid them.

Notes for future investigation:

- We originally tried to fix the issue by making TraceObserverV2 an interface instead of a trait. However, it seems Groovy 4 still compiles interfaces w/ default methods as traits. Groovy 5 fixes this, and we are trying to upgrade (#6220), but there are issues with the Groovy type checker that need to be resolved. Once we upgrade to Groovy 5, we should try this again.

- Paolo also saw a ClassNotFoundException when testing the latest edge release, related to the `WorkflowPublishEvent` class that we added. However, I think this was an issue with the build system not using the locally built plugins. We should confirm whether running a plugin built with an older Nextflow version actually fail with this error. I would expect the plugin to pick up the new event class from the Nextflow runtime.

If these two issues can be resolved, then we should be able to add new trace events without breaking existing plugins. Until then, we will try to add new events only when absolutely necessary.

PR #6702, opened 2026-01-05
docs: Update migration note for plugin registry

bentsherman:
The 25.10 migration notes currently say that the legacy plugin index can still be used, but AFAIK this support was completely removed in 25.10

@pditommaso can the `NXF_PLUGINS_TEST_REPOSITORY` env var still be used? If not then we should add a deprecation note


docs/reference/env-vars.md
```diff
@@ -182,34 +180,32 @@ The following environment variables control the configuration of the Nextflow ru
...
-: :::{versionadded} 25.08.0-edge
+: :::{versionadded} 25.10.0
   :::
 : Delay multiplier used for HTTP retryable operations (default: `2.0`).
 
 `NXF_SCM_FILE`
-: :::{versionadded} 20.10.0
-  :::
-: Defines the path location of the SCM config file .
+: Defines the path location of the SCM config file.
```
christopher-hakkaart:
```suggestion
: The path location of the SCM config file.
```

docs/migrations/25-10.md
```diff
@@ -194,14 +194,8 @@ See {ref}`cli-auth` and {ref}`cli-launch` for details.
 
 Nextflow now uses the [Nextflow plugin registry](https://registry.nextflow.io/) to download plugins in a more efficient and scalable manner.
 
-The legacy plugin index can still be used by setting the `NXF_PLUGINS_REGISTRY_URL` environment variable:
-
-```bash
-export NXF_PLUGINS_REGISTRY_URL="https://raw.githubusercontent.com/nextflow-io/plugins/main/plugins.json"
-```
-
```
pditommaso, line 202: Why is this removed? 

https://github.com/nextflow-io/nextflow/blob/68d75445913ddfe3ef255dacde52096e99a27b88/modules/nf-commons/src/main/nextflow/plugin/PluginsFacade.groovy#L125-L129

pditommaso (changes requested):
> can the NXF_PLUGINS_TEST_REPOSITORY env var still be used? If not then we should add a deprecation note

it should be still in place 

https://github.com/nextflow-io/nextflow/blob/68d75445913ddfe3ef255dacde52096e99a27b88/modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy#L103-L103

docs/migrations/25-10.md
bentsherman: I thought you said that 25.10 doesn't support the legacy index? Isn't that why we don't publish plugins requiring >=25.10 to the legacy index?

pditommaso: I got your point, however technically both NXF_PLUGINS_TEST_REPOSITORY and NXF_PLUGINS_REGISTRY_URL are still there

bentsherman: Then I am confused. Does 25.10 still support the legacy index or not?

bentsherman:
Since we were able to confirm that the legacy plugin index still works in 25.10, I will just update the note about plugin publishing, since the registry is now GA.

PR #6713, opened 2026-01-12
Detect conda explicit files by content instead of extension

pditommaso:
## Summary

- Changed conda explicit file detection from extension-based (`.txt`) to content-based by checking for the `@EXPLICIT` marker
- The detection now reads the first 20 lines of a file to look for the `@EXPLICIT` marker, which is the standard identifier for conda explicit specification files
- This approach is more reliable and flexible than relying on file extensions

## Motivation

Previously, conda explicit files were identified solely by their `.txt` extension using `isTextFilePath()`. This had two issues:

1. **False negatives**: Conda explicit files with non-standard extensions (e.g., `.lock`, `.spec`) were not recognized
2. **False positives**: Regular text files with `.txt` extension could be incorrectly treated as conda explicit files

The `@EXPLICIT` marker is the canonical identifier for conda explicit specification files (see [conda documentation](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#building-identical-conda-environments)). By checking file content, we can reliably identify these files regardless of their extension.

### CondaCache.groovy
- Replaced `isTextFilePath()` with `isExplicitFile()` - checks if file exists and contains `@EXPLICIT` marker
- Added `containsExplicitMarker()` helper - efficiently reads only first 20 lines
- Updated `condaPrefixPath()` and `createLocalCondaEnv0()` to use the new method

### CondaCacheTest.groovy
- Updated tests to use actual conda explicit file format with `@EXPLICIT` marker
- Tests now create real files with proper content rather than relying on extension checks

## Test plan

- [x] All existing `CondaCacheTest` tests pass
- [x] New content-based detection correctly identifies files with `@EXPLICIT` marker
- [x] Files without `@EXPLICIT` marker are not incorrectly identified as explicit files
- [x] Non-existent files return `false` gracefully

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy
@@ -182,6 +182,7 @@ class CondaCache {
             return containsExplicitMarker(path)
         }
         catch( Exception e ) {
+            log.debug "Failed to check explicit file: $str - ${e.message}"
             return false
         }
     }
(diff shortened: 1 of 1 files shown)


pditommaso:
@ewels does this match the same expectation of https://github.com/nextflow-io/nextflow/pull/6665?

Issue #6664, opened 2025-12-16
`nextflow lint` breaks when reporting certain syntax errors


bentsherman:
## Bug report 

From this Slack thread: https://nextflow.slack.com/archives/C02TPSL19UJ/p1765881373697229

The following test cases have syntax errors:

```groovy
// nextflow.config
process {
  withLabel:
  'bambino' {
    container = "..."
  }
}
```

```nextflow
process SEQTK_SAMPLE {

  script:
  """
  ${
    params.is_paired_end
    ? "..."
    : "..."
  }
  """
}
```

### Expected behavior and actual behavior

The linter should report syntax errors for both snippets (the language sever reports them correctly).

Instead it fails with an error like this:

```
java.lang.StringIndexOutOfBoundsException: Range [12, 13) out of bounds for length 12
	at ...
	at java.base/java.lang.String.substring(String.java:2834)
	at nextflow.script.parser.v2.StandardErrorListener.printCodeBlock(ErrorListener.groovy:187)
	at nextflow.script.parser.v2.StandardErrorListener.onError(ErrorListener.groovy:99)
	at nextflow.cli.CmdLint$_printErrors_lambda7.doCall(CmdLint.groovy:210)
	at nextflow.cli.CmdLint.printErrors(CmdLint.groovy:205)
	at nextflow.cli.CmdLint.access$0(CmdLint.groovy)
	at nextflow.cli.CmdLint$_checkErrors_lambda4.doCall(CmdLint.groovy:193)
	at nextflow.cli.CmdLint.checkErrors(CmdLint.groovy:189)
	at nextflow.cli.CmdLint.run(CmdLint.groovy:149)
	at nextflow.cli.Launcher.run(Launcher.groovy:515)
	at nextflow.cli.Launcher.main(Launcher.groovy:675)
```

### Steps to reproduce the problem

Run `nextflow lint` with the given code snippets.

PR #6717, opened 2026-01-13
Fix lint error formatting for errors at end of line

ewels:
Fixed bug in linting where errors that included EOL characters couldn't be printed and triggered an error.

Before (nf-core/seqinspector):

CleanShot 2026-01-13 at 14 59 13@2x

After:

CleanShot 2026-01-13 at 14 59 32@2x

Solution:

- Fix `StringIndexOutOfBoundsException` in `nextflow lint` when errors occur at end-of-line positions
- Add proper bounds checking for start/end positions in `printCodeBlock()` method

<details>

## Problem
When running `nextflow lint` on files with syntax errors at positions beyond the actual line length (e.g., "Unexpected input: '\n'" errors at end of line), the output would show:
```
ERROR ~ begin 44, end 45, length 44
```
instead of properly formatted error descriptions with code context.

## Solution
Added bounds checking in `ErrorListener.printCodeBlock()` to ensure `start`, `end`, `adjStart`, and `adjEnd` values stay within the line length.

## Test plan
- [x] Run `nextflow lint .` in `test/seqinspector/` directory
- [x] Verify errors in `conf/modules.config` and `nf-test.config` display properly formatted messages
- [x] Verify JSON output still works correctly

</details>

### DIFF
--- a/docs/conda.md
+++ b/docs/conda.md
@@ -62,6 +62,10 @@ The usual Conda package syntax and naming conventions can be used. The version o
 
 The name of the channel where a package is located can be specified prefixing the package with the channel name as shown here `bioconda::bwa=0.7.15`.
 
+:::{versionchanged} 26.04.0
+By default, Nextflow uses the `conda-forge` and `bioconda` channels to resolve Conda packages. You can override this using the `conda.channels` configuration option.
+:::
+
 (conda-env-files)=
 ### Use Conda environment files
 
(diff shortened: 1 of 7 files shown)


bentsherman:
The error shown in the PR description is from nf-core/seqinspector:

https://github.com/nf-core/seqinspector/blob/a4ba47c71a2f15f92b340b37b2dd2358df54d0cc/conf/modules.config#L123

It is the same kind of error reported in #6664 , which this PR fixes. I will add a unit test for that and then merge.

bentsherman (approved):

PR #6719, opened 2026-01-13
Move shared Platform code to separate modules

jorgee:
This pull request introduces modularization improvements to the Nextflow project and decouples some classes based on the usage of Nexflow classes in Platform

 **Enhancements**

- Decouple AssetManager and ConfigBuilder from cli package:  
  - Define interfaces `ConfigNodeOption`, `ConfigRunOptions` and `ConfigCliOptions` to access the required options from `ConfigBuilder` without requiring the rest of `cli` classes  
  - Same for `HubOptions`, it is now an interface at `scm` package and implement in `cli` as `CliHubOptions`  
- Remove `NF` class dependency from `ConfigBuilder`  
  - DSL parser version can be moved to `Const` as it is just a Environment Variable inspection.  
  - `ConfigBuilder` includes a `strict` flag to indicate when to use strict mode when merging maps.   
- Remove `HistoryFile` dependency from `ConfigBuilder`  
  - Move `normalizeResumeId` to `CmdRun`. It is only used when the `cmdRun` field is set.  Use ConfigRunOptions interface to get the normalized resume id value.
- Move the classes to the new modules based on the functionality, we propose to create 3 new modules:  
    - **nf-secrets**: contains the current `secret` package in the nextflow module and `nextflow.util.SecretHelper` class. Depends on nf-commons  
    - **nf-config**: contains the current `config` package in the nextflow module and the `nextflow.util.ConfigHelper` class. Depends on `nf-secrets` and `nf-commons`  
    - **nf-scm**: contains the current `scm` package in the nextflow module and the `nextflow.script.ScriptFile` class. Depends on `nf-config` and `nf-commons`  
- Move classes to nf-commons:  
    - Some exceptions, `nexflow.executor.BashTemplate` and `nextflow.util.MustacheTemplateEngine`: used by Platform  
    - `nextflow.ast.NextflowXForm`: required by config parsers  
    - `nextflow.parser.v2.ErrorListener`: required by `ConfigParserV2`


modules/nextflow/src/main/groovy/nextflow/cli/CmdConfig.groovy
```diff
@@ -114,14 +116,15 @@ class CmdConfig extends CmdBase {
...
                 .setBaseDir(base)
-                .setCmdConfig(this)
+                .setShowAllProfiles(this.showAllProfiles)
+                .setProfile(this.profile)
 
         final config = builder.buildConfigObject()
 
         // -- validate config options
-        if( NF.isSyntaxParserV2() ) {
+        if( Const.isSyntaxParserV2() ) {
```
pditommaso: Please avoid unrequired changes to minimise the impact of this PR

bentsherman:
Thanks @jorgee for looking into this. Let's focus on your codebase analysis for now, as we need to agree on a plan first

I suspect if we finalize #5971 and make a few judgement calls, we can make this work with fewer changes. But I need to study your analysis first.

Once we agree on a plan, we can add it here as an ADR.

modules/nextflow/src/main/groovy/nextflow/cli/CmdConfig.groovy
jorgee: You are right,this is not required. Changed

modules/nextflow/src/main/groovy/nextflow/cli/CmdFs.groovy
```diff
@@ -203,7 +203,7 @@ class CmdFs extends CmdBase implements UsageAware {
     private Session createSession() {
         // create the config
         final config = new ConfigBuilder()
-                .setOptions(getLauncher().getOptions())
+                .setCliOptions(getLauncher().getOptions())
```
pditommaso, line 206: Please avoid all not strictly required changes. We can make this kind of refactor in a followup pr

pditommaso (changes requested):
Apart the previous comment, massive PR, well done!

PR #6720, opened 2026-01-13
Support multiple config option types in config spec

### LINKED ISSUE #6718: v2 config spec check: false positive around `process.publishDir` attributes
## Bug report 

False warnings are given around unrecognised config options for `process.publishDir` options.

```console
$ export NXF_SYNTAX_PARSER=v2
$ nextflow run nf-core/testpipeline -profile docker,test --outdir ./results
...
WARN: Unrecognized config option 'process.publishDir.path'
WARN: Unrecognized config option 'process.publishDir.mode'
WARN: Unrecognized config option 'process.publishDir.saveAs'
...
```

These warnings should not be triggered.

Running `25.12.0-edge`

bentsherman:
Fix #6718 

This PR updates the config spec to support multiple types for config options.

For example, `process.publishDir` can be supplied with a String, Map, or List, as shown in `ProcessDsl`. The config spec needs to be able to allow for all of these types, especially as we extend type checking to config files.

For now, this PR fixes a bug in the runtime config validation. The ConfigValidator needs to know whether a config option has type Map in order to distinguish between "map" options and scopes in the resolved config.

As a bonus, this PR adds support for the `types` field in the `ConfigOption` annotation, which allows config options to specify "additional types". See for example the `k8s.pod` option defined in K8sConfig. The config spec JSON has been updated to include these additional types in a non-breaking way. These will eventually be used by the type checker.

### DIFF
--- a/modules/nf-lang/src/main/java/nextflow/config/control/ConfigResolveVisitor.java
+++ b/modules/nf-lang/src/main/java/nextflow/config/control/ConfigResolveVisitor.java
@@ -17,6 +17,7 @@ package nextflow.config.control;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.regex.Pattern;
 
 import nextflow.config.ast.ConfigAssignNode;
 import nextflow.config.ast.ConfigIncludeNode;
@@ -78,6 +79,8 @@ public class ConfigResolveVisitor extends ConfigVisitorSupport {
 
     private class DynamicVariablesVisitor extends ConfigVisitorSupport {
 
+        private static final Pattern ENV_VAR_NAME = Pattern.compile("[A-Z_]+[A-Z0-9_]*");
+
         @Override
         protected SourceUnit getSourceUnit() {
             return sourceUnit;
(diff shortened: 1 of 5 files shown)


modules/nf-lang/src/main/java/nextflow/config/spec/SpecNode.java
```diff
@@ -81,8 +82,15 @@ private static SpecNode processScope() {
...
-            children.put(method.getName(), new Option(desc, optionType(method)));
+            if( method.getParameters().length != 1 )
+                continue;
+            if( !children.containsKey(method.getName()) ) {
+                var desc = annotatedDescription(method, "");
+                children.put(method.getName(), new Option(desc, new ArrayList<>()));
+            }
+            var option = (Option) children.get(method.getName());
+            var paramType = method.getParameterTypes()[0];
+            option.types.add(paramType);
```
jorgee, line 102: The goal of this part of code could be difficult to understand without the PR context. A comment such as "Get process options methods declared DirectiveDsl. Methods overloaded are reflected as options with multiple supported types" could help.

bentsherman: Thanks Jorge, good idea

PR #6721, opened 2026-01-13
Improve config syntax errors

bentsherman:
This PR makes a few minor improvements to linter errors:

- Report a better error message for an if statement in a config selector block. Should fix the four pipelines currently showing "Parse error" over at [strict-syntax-health](https://github.com/ewels/strict-syntax-health).

- Add a hint about using the `env()` function when an undefined variable in the config looks like an environment variable

See the PR changes for a concrete example of each improvement


ewels (approved):
Nice!

PR #6722, opened 2026-01-14
Fix spot instance detection for AWS Batch cost estimation

robsyme:
## Summary

Fix incorrect priceModel detection for AWS Batch tasks where spot instances were always reported as `standard`, causing inflated cost estimates in Seqera Platform.

## Problem

The `getPrice()` method in `AwsBatchHelper` compared the `instanceLifecycle()` enum directly to a string:

```groovy
instance.instanceLifecycle()=='spot'  // always false
```

With `@CompileStatic`, this enum-to-string comparison always returns false, so all instances defaulted to PriceModel.standard.

## Solution

Replace the string comparison with direct enum comparison:
```groovy
instance.instanceLifecycle() == InstanceLifecycleType.SPOT
```

## Test
Added AwsBatchHelperTest covering spot, scheduled, and on-demand (null) cases.


bentsherman:
@pditommaso you think this is worth backporting?

pditommaso:
Yes, definitively

robsyme:
I did not know that `@CompileStatic` would mess up the comparison. You learn something every day :D 

Thanks team.

bentsherman:
I'm surprised as well, I would expect Groovy to catch the type mismatch. But the Groovy type checker isn't perfect, it has some small holes

PR #6723, opened 2026-01-14
ADR: Nextflow Module System (trimmed)

bentsherman:
This PR trims the Module System ADR to focus on tooling improvements.

This will allow us to provide immediate value for pipeline developers and AI agents, without altering the way that pipelines are written / executed.

In my mind these bits are:
- Run modules directly via CLI
- Publish versioned modules to registry
- Install / manage versioned modules from registry instead of monorepo
- Generate module spec from source code, with extra info like tool args

### DIFF
--- a/adr/20251114-module-system.md
+++ b/adr/20251114-module-system.md
@@ -8,6 +8,9 @@
 
 ## Updates
 
+### Version 2.5 (2026-01-14)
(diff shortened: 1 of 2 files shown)


edmundmiller:
Might be out of scope but ideas for the future:
- `nextflow download` to get all the modules and containers
- `nextflowx` and `nfx` alias for `nextflow module run`
- `nextflow.lock` file

PR #6727, opened 2026-01-16
Update unit tests to DSL2 runtime

bentsherman:
This PR removes all remaining DSL1 code paths in Nextflow.

This mostly comes down to how channel factories are initialized. For example:

```groovy
if(NF.isDsl2()) {
    session().addIgniter { ch.bind(value) }
}
else {
    ch.bind(value)
}
```

In DSL2, dataflow logic is always executed in the context of a Session. This is because DSL2 added the ability to use the same channel multiple times in dataflow logic, which requires some logic to "broadcast" such channels before igniting the dataflow.

However, many unit tests still run dataflow logic without a session, which is why the above DSL1 code path wasn't removed.

This PR adds a helper function `runDataflow` which can be used to run dataflow logic with a session.

So the following test:
```groovy
when:
def result = Channel.of('a')
then:
result.val == 'a'
result.val == Channel.STOP
```

Becomes:
```groovy
when:
def result = runDataflow {
    Channel.of('a')
}
then:
result.val == 'a'
result.val == Channel.STOP
```

This seems like the most lightweight change that allows the DSL1 code path to be removed.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/ProcessDef.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/ProcessDef.groovy
@@ -94,12 +94,6 @@ class ProcessDef extends BindableDef implements IterableDef, ChainableDef {
         str.split(Const.SCOPE_SEP).last()
     }
 
-    protected void initialize() {
-        // apply config settings to the process
-        final configProcessScope = (Map)session.config.process
-        new ProcessConfigBuilder(processConfig).applyConfig(configProcessScope, baseName, simpleName, processName)
-    }
-
     @Override
     ProcessDef clone() {
         def result = (ProcessDef)super.clone()
@@ -144,9 +138,6 @@ class ProcessDef extends BindableDef implements IterableDef, ChainableDef {
 
     @Override
     Object run(Object[] args) {
-        // initialise process config
-        initialize()
-
         // invoke process with legacy inputs/outputs
         if( processConfig instanceof ProcessConfigV1 )
             output = runV1(args, processConfig)
@@ -261,14 +252,23 @@ class ProcessDef extends BindableDef implements IterableDef, ChainableDef {
     }
 
     TaskProcessor createTaskProcessor() {
-        if( !processConfig )
-            initialize()
+        // apply process directives from config settings
+        applyConfig()
+
+        // create executor for process
         final executor = session
             .executorFactory
             .getExecutor(processName, processConfig, taskBody, session)
+
+        // create task processor for process
         return session
             .newProcessFactory(owner)
             .newTaskProcessor(processName, executor, processConfig, taskBody)
+    }
+
+    protected void applyConfig() {
+        final configProcessScope = (Map)session.config.process
+        new ProcessConfigBuilder(processConfig).applyConfig(configProcessScope, baseName, simpleName, processName)
     }
 
 }
--- a/modules/nextflow/src/test/groovy/nextflow/script/ProcessDefTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/script/ProcessDefTest.groovy
@@ -1,6 +1,9 @@
 package nextflow.script
 
 import nextflow.Session
+import nextflow.executor.Executor
+import nextflow.executor.ExecutorFactory
+import nextflow.processor.TaskProcessor
 import spock.lang.Specification
 /**
  *
@@ -58,7 +61,7 @@ class ProcessDefTest extends Specification {
 
         when:
         def copy = proc.clone()
-        copy.initialize()
+        copy.applyConfig()
         then:
         def cfg1 = copy.processConfig.createTaskConfig()
         cfg1.getCpus()==2           // taken from the generic config
@@ -66,7 +69,7 @@ class ProcessDefTest extends Specification {
 
         when:
         copy = proc.cloneWithName('flow1:bar')
-        copy.initialize()
+        copy.applyConfig()
         then:
         def cfg2 = copy.processConfig.createTaskConfig()
         cfg2.getCpus()==4           // taken from the `bar` config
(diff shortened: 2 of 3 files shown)


modules/nextflow/src/test/groovy/nextflow/extension/DataflowMergeExtensionTest.groovy
```diff
@@ -71,66 +50,53 @@ class DataflowMergeExtensionTest extends Specification {
...
         then:
-        result instanceof DataflowQueue
         result.val == ['A', 'a', [1,2,3]]
         result.val == ['B', 'b', [3,4,5]]
         result.val == ['C', 'c', [6,7,8]]
         result.val == Channel.STOP
 
     }
 
-    def 'should merge with list'() {
```
jorgee, line 108: Why is this test removed? It seems the same as 'should merge with open array with custom closure' but it has been removed too

modules/nextflow/src/test/groovy/nextflow/script/ChannelOutTest.groovy
```diff
@@ -16,8 +17,8 @@ class ChannelOutTest extends Specification {
 
     def 'should get out by name' () {
         given:
-        def ch1 = Channel.value('a')
```
jorgee: Why Channel.value can not be used here?

modules/nextflow/src/test/groovy/nextflow/splitter/BytesSplitterTest.groovy
```diff
@@ -49,17 +49,4 @@ class BytesSplitterTest extends Specification {
 
     }
 
-    def testSplitterChannel() {
```
jorgee, line 52: Why is this test removed?

modules/nextflow/src/main/groovy/nextflow/script/OutputDsl.groovy
```diff
@@ -92,11 +93,8 @@ class OutputDsl {
         return opts
     }
 
-    boolean getComplete() {
-        for( final op : ops )
-            if( !op.complete )
-                return false
-        return true
+    Map<String,Object> getOutput() {
```
jorgee, line 96: I do not understand changes in this class. Is it to facilitate the testing?

modules/nextflow/src/test/groovy/nextflow/splitter/StringSplitterTest.groovy
```diff
@@ -51,19 +50,6 @@ class StringSplitterTest extends Specification {
 
     }
 
-    def testSplitChannel() {
```
jorgee, line 54: Why is this test removed?

modules/nextflow/src/test/groovy/nextflow/splitter/TextSplitterTest.groovy
```diff
@@ -86,24 +86,11 @@ class TextSplitterTest extends Specification {
...
+        def chunks = new TextSplitter().target(file).options(by:2).list()
         then:
-        channel.val == 'line1\nline2\n'
-        channel.val == 'line3\nline4\n'
-        channel.val == 'line5\n'
-        channel.val == Channel.STOP
-
-    }
-
-    def testSplitChannel() {
```
jorgee: Why is this test removed?

modules/nextflow/src/main/groovy/nextflow/script/ScriptRunner.groovy
```diff
@@ -54,11 +54,6 @@ class ScriptRunner {
      */
     private ScriptFile scriptFile
 
-    /**
-     * The script result
-     */
-    private def result
```
jorgee, line 60: Not sure why do you remove result. Is it only used in DSL1?

modules/nextflow/src/test/groovy/nextflow/splitter/BytesSplitterTest.groovy
jorgee: I see this is the same as with list but with channel

modules/nextflow/src/test/groovy/nextflow/splitter/StringSplitterTest.groovy
jorgee: Same as list too

bentsherman:
Thanks Jorge, all good comments. I will try to address all of them

modules/nextflow/src/main/groovy/nextflow/script/OutputDsl.groovy
bentsherman: Basically yes. I've been meaning to do this anyway because it is a more idomatic way to collect the workflow outputs. This way the OutputDslTest can just call this method instead of calling `getComplete()` in a loop.

modules/nextflow/src/main/groovy/nextflow/script/ScriptRunner.groovy
bentsherman: I actually meant to remove this in #6643 . It is only used for testing, which is now handled by `ScriptHelper`, so this logic is no longer needed in `ScriptRunner`

modules/nextflow/src/test/groovy/nextflow/extension/DataflowMergeExtensionTest.groovy
bentsherman: Both of these tests fail when updated to use DSL2 mode. I tried running them in a script with Nextflow 25.10 and they both fail there as well, so it's not an issue with the test harness.

I think these are just obscure cases that stopped working in DSL2, but it never came up because the tests were still using DSL1 mode. The `merge` operator is already heavily discouraged and these cases are even more obscure variants, so I think it's fine to just remove them

bentsherman: Just confirmed, these tests fail going all the way back to 24.04, possibly further. I think these cases probably stopped working when DSL2 was introduced

modules/nextflow/src/test/groovy/nextflow/splitter/BytesSplitterTest.groovy
bentsherman: This test hangs for some reason when updated to use DSL2. `BytesSplitter` is not used at all in Nextflow and the splitter channel use case is already covered by the various `Split*Op` tests. So I think it's fine to remove this one

modules/nextflow/src/test/groovy/nextflow/splitter/StringSplitterTest.groovy
bentsherman: Similar to `BytesSplitter`, this test was being difficult and isn't worth keeping. `StringSplitter` is used by `splitFasta`, the channel mode is not used.

modules/nextflow/src/test/groovy/nextflow/splitter/BytesSplitterTest.groovy
pditommaso: We should keep an eye on this, this could hide a race condition

bentsherman: But `BytesSplitter` was not used anywhere?

PR #6728, opened 2026-01-16
Add `-project-dir` option to `lint` command

### LINKED ISSUE #6634: Nextflow linter errors on custom Groovy code in lib/ as not being defined

Running `nextflow lint ./` on a project that has Groovy classes in `lib/` and using those classes in the Nextflow script and processes should not return a "is not defined" error. Instead, an error like this appears:

```console
Error main.nf:5:5: `MyHelpers` is not defined
│   5 |     MyHelpers.SayHi
╰     |     ^^^^^^^^^

Nextflow linting complete!
 ❌ 1 file had 1 error
```

Make a little Groovy helper at `lib/MyHelpers.groovy`:

```groovy
/** Helpers for my Nextflow project. */
class MyHelpers {
    public static final String SayHi = "echo 'Hello, World!'"
}
```

And a Nextflow script to use that helper:

```nextflow
process say_something {
    output: stdout()
    script: MyHelpers.SayHi
}

workflow { say_something() | view }
```

### Program output 

Running the script works:

```console
❯ nextflow run main.nf
...
[f4/d8349f] say_something | 1 of 1 ✔
Hello, World!
```

But no matter how I used the Groovy code, I can't get it to lint successfully:

```console
❯ nextflow lint ./
Error main.nf:5:5: `MyHelpers` is not defined
│   5 |     MyHelpers.SayHi
╰     |     ^^^^^^^^^

Nextflow linting complete!
 ❌ 1 file had 1 error
```

I suspect the linter is not compiling and accessing code in `lib/`.

### Environment 

* Nextflow version: 25.10.2  
* Java version: OpenJDK Runtime Environment Homebrew (build 23.0.2)
* Operating system: MacOS

bentsherman:
Close #6634 

This PR updates the `nextflow lint` command to recognize project-level assets such as the lib directory and modules directory

- Add `-project-dir` option to `nextflow lint` (defaults to `.`)
- Pass project dir to ScriptParser so that it can resolve remote module includes against `modules` directory
- Move shared class loader logic to ClassLoaderFactory so that it can be used by both Session and CmdLint
- Update ResolveVisitor to handle Groovy classes from the lib directory
- Add unit test based on linked issue

### DIFF
--- a/modules/nf-lang/src/main/java/nextflow/script/control/GroovyCompiler.java
+++ b/modules/nf-lang/src/main/java/nextflow/script/control/GroovyCompiler.java
@@ -52,12 +52,7 @@ public class GroovyCompiler {
         compilationUnit.addSource(sourceUnit);
 
         // compile source files
-        try {
-            compilationUnit.compile(Phases.CANONICALIZATION);
-        }
-        catch( GroovyBugError | Exception e ) {
-            // ignore
-        }
+        compilationUnit.compile(Phases.CANONICALIZATION);
 
         // collect compiled classes
         var moduleNode = sourceUnit.getAST();


bentsherman:
TODO: specify `-project-dir` instead of `-lib-dir` so that `nextflow lint` can understand any project-level entities -- lib dir, modules dir, etc

modules/nf-lang/src/main/java/nextflow/script/control/GroovyCompiler.java
```diff
@@ -0,0 +1,70 @@
...
+                classLoader,
+                new LazyErrorCollector(config));
+        compilationUnit.addSource(sourceUnit);
+
+        // compile source files
+        try {
+            compilationUnit.compile(Phases.CANONICALIZATION);
+        }
+        catch( GroovyBugError | Exception e ) {
+            // ignore
```
jorgee: Why exceptions are ignored here? Shouldn't it be good to log the message?

modules/nf-lang/src/main/java/nextflow/script/control/ResolveVisitor.java
```diff
@@ -264,9 +264,16 @@ protected ClassNode resolveFromClassResolver(String name) {
...
-        if( !lookupResult.isClassNode() )
-            throw new GroovyBugError("class resolver lookup result is not a class node");
-        return lookupResult.getClassNode();
+        if( lookupResult.isClassNode() )
+            return lookupResult.getClassNode();
+        // When a Groovy class from the lib directory is used, the class
+        // loader returns the URI of the Groovy file. We only need to compile
+        // the Groovy file enough to resolve the class definition for the purpose
+        // of name checking.
+        var su = lookupResult.getSourceUnit();
```
jorgee, line 273: Claude suggests adding a cache (such as Map<URI, List<ClassNode>>) to avoid recompiling the same source several times, but I think it could be also introduce a leak. What do you think is it worth to cache this compilation?

modules/nf-lang/src/main/java/nextflow/script/control/GroovyCompiler.java
bentsherman: Good point. This code was ripped from the language server, where I actually do log an error:

https://github.com/nextflow-io/language-server/blob/ed8a2d28f4c79ed9382a863c3f0d37372431eb05/src/main/java/nextflow/lsp/services/script/GroovyLibCache.java#L121-L130

However, I don't have logging set up for nf-lang so I just removed it.

Any compile errors will get reported later when Nextflow tries to use the Groovy code, although they aren't very helpful:

```console
$ NXF_SYNTAX_PARSER=v1 nextflow run .

ERROR ~ BUG! exception in phase 'semantic analysis' in source unit '_nf_script_8539748b2027415c' The lookup for MyHelpers caused a failed compilation. There should not have been any compilation from this call.

$ NXF_SYNTAX_PARSER=v2 nextflow run .

ERROR ~ BUG! exception in phase 'conversion' in source unit 'Main' The lookup for MyHelpers caused a failed compilation. There should not have been any compilation from this call.
```

I will try to log a warning here via slf4j

modules/nf-lang/src/main/java/nextflow/script/control/ResolveVisitor.java
bentsherman: I don't think it's worth doing. We are trying to move people away from the `lib` directory anyway, so I'm not worried about making it very efficient.

modules/nf-lang/src/main/java/nextflow/script/control/GroovyCompiler.java
bentsherman: Actually I don't think we need this try-catch at all. I think I added it in the language server as a paranoid safe-guard (to not crash the language server), but I don't know if there is an error condition here

PR #6733, opened 2026-01-19
Fix fetching new remote branch in multi-revision mode

### LINKED ISSUE #6732: Fetching new remote branch fails in multi-revision mode
## Expected behavior and actual behavior

**Expected:** When running a pipeline with -revision <branch-name> where the branch exists on the remote repository, Nextflow should fetch and checkout that branch successfully.

**Actual:** When using the multi-revision repository strategy (default since ~25.x), attempting to pull a branch that exists on the remote but hasn't been fetched locally fails with the error:

```
Remote does not have <branch-name> available for fetch.
```

This only affects branches that exist on the remote but are not yet present in the local bare repository. Once a branch has been fetched at least once, subsequent pulls work correctly.

## Steps to reproduce the problem

1. Ensure you're using a recent Nextflow version with multi-revision mode enabled (default)
2. Try to run or pull a pipeline with a branch that you haven't fetched before:

Clear any cached repos to ensure fresh state
```bash
rm -rf ~/.nextflow/assets/nf-core ~/.nextflow/.repos/nf-core
```
Try to pull a branch that exists on the remote
```bash
nextflow pull nf-core/fetchngs -revision preview-26-04
```

Or with any GitHub repository that has a branch not yet fetched:

```bash
nextflow run nf-core/fetchngs -revision preview-26-04
```

Program output
```
N E X T F L O W  ~  version 25.12.0-edge

Pulling nf-core/fetchngs:preview-26-04 ...
Remote does not have preview-26-04 available for fetch.
```

## Environment

- Nextflow version: 25.12.0-edge (and likely earlier versions with multi-revision mode)
- Java version: 21
- Operating system: macOS / Linux
- Bash version: N/A (not bash-related)

## Additional context

Root cause: In MultiRevisionRepositoryStrategy.refSpecForName(), when determining the correct RefSpec for fetching, the code only checks local refs in the
bare repository:

```groovy
Ref branch = getBareGit().getRepository().findRef("refs/heads/" + revision)
```

For a new branch that doesn't exist locally, this returns null, causing the method to fall through and treat the branch name as a commit SHA:

```groovy
return new RefSpec(revision + ":refs/tags/" + revision)
```
This creates an invalid RefSpec like `preview-26-04:refs/tags/preview-26-04`, which JGit fails to fetch because there's no commit with that literal name.

Fix: The method should query remote refs via `lsRemote()` when the revision is not found locally, before falling back to treating it as a commit SHA.

robsyme:
## Summary

- Fix bug where fetching a branch that exists on the remote but not in the local bare repository fails with "Remote does not have X available for fetch"
- The `refSpecForName()` method now queries remote refs via `lsRemote()` when a revision is not found locally, before falling back to treating it as a commit SHA
- Added `@Memoized` helper method `fetchRemoteRefs()` to cache results and avoid redundant network calls

Fixes #6732

## Test plan

- [x] Added integration test `should fetch new remote branch not in local bare repo` that verifies fetching a branch not present locally
- [x] Added unit test `should create correct RefSpec for branches tags and commits` that verifies fallback behavior without network access
- [x] All existing SCM tests pass
- [x] Manual verification: `nextflow pull nf-core/fetchngs -revision preview-26-04` now works correctly

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/scm/MultiRevisionRepositoryStrategy.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/MultiRevisionRepositoryStrategy.groovy
@@ -335,14 +335,14 @@ class MultiRevisionRepositoryStrategy extends AbstractRepositoryStrategy {
 
     private RefSpec refSpecForName(String revision) {
         // First, check if it's a local branch
-        final branchName = "refs/heads/$revision"
+        final branchName = "refs/heads/$revision".toString()
         final branch = getBareGit().getRepository().findRef(branchName)
         if( branch != null ) {
             return new RefSpec("$branchName:$branchName")
         }
 
         // Check if it's a local tag
-        final tagName = "refs/tags/$revision"
+        final tagName = "refs/tags/$revision".toString()
         final tag = getBareGit().getRepository().findRef(tagName)
         if( tag != null ) {
             return new RefSpec("$tagName:$tagName")


jorgee (approved):
Thank you @robsyme, for the bug catch, the fix and the smart unit test. Changes looks fine. I have just removed the `fetchRemoteRefs()` because the implementation is the same as `lsRemote(false)`, and  I have also added tags case in the test.

robsyme:
Ah, I didn't understand the `lsRemote`, but now that I look at it closer, the duplication is obvious and you changes make this a much cleaner PR. Thanks @jorgee!

bentsherman:
@jorgee I tried to make it more Groovy but I think I broke something :disappointed: 

Can you double check my work?

robsyme:
I think you're running into a GString vs String issue, Ben (they hash differently because Groovy). I think I can fix with a `toString()` or two. Commit incoming...

bentsherman:
Thanks Rob. Weird that the GString would break there, but whatever. Tests are passing :heavy_check_mark:

robsyme:
Thanks for collab, team. Much appreciated.

Ben - the GString breaks because it's used as a HashMap lookup down the call tree, which uses the hash for identity, leading to a missed key in the HashMap.

PR #6740, opened 2026-01-20
Decouple SCM classes in Nextflow independent module

jorgee:
This pull request refactors the Nextflow asset management to decouple the SCM functionality form the rest of the Nextflow code. 

Two new packages are created 
- `nf-scm`: Contains the SCM classes decouple from the rest of nextflow classes

- `nf-utils`: Contains util clases required by SCM and the rest of nextflow

alternative for #6719


pditommaso:
I have the feeling that https://github.com/nextflow-io/nextflow/pull/6719 is a better approach. What's the advantage of this PR ?

jorgee:
In this PR, classes are not depending on nf-commons. only nf-utils and nf-scm are required to move to libseqera. In the other, you require to get nf-commons, nf-secrets, nf-config and nf-scm.

pditommaso:
I understand but in my view this refactor is going too far. I think should stay in nextflow repository, other apps still benefits avoiding to import all nextflow runtime full deps tree.

PR #6742, opened 2026-01-21
Add time-based caching for K8sConfig.getClient()

### LINKED ISSUE #6741: k8s plugin must check for valid cluster token
## Bug report 

Hello,

as of 1.34 version of Kubernetes with RKE2 engine, tokens (`/var/run/secrets/kubernetes.io/serviceaccount/token`) are refreshed each hour, so everything is failing with workflows longer than 1 hour. So for incluster configs, token should be re-read at least once an hour, or with each request or watch for the file change (which is a bit tricky with inotify watch).

pditommaso:
## Summary
- Replace `@Memoized` with a Guava time-based cache that expires after a configurable interval (default 50 minutes)
- Add new config option `k8s.clientRefreshInterval` to control the refresh interval
- Allows Kubernetes client configuration to be refreshed periodically, useful for environments where credentials or tokens expire

## Changes
- Added `clientRefreshInterval` config option (default: `50m`)
- Added `clientCache` field using Guava `Cache<String, ClientConfig>`
- Refactored `getClient()` to use cache with `getClient0()` as the loader

Closes #6741

## Test plan
- [x] Added test for `clientRefreshInterval` default and custom values
- [x] Added test verifying cache behavior and expiration


docs/reference/config.md
```diff
@@ -1057,6 +1057,10 @@ The following settings are available:
   - `clientKey`
   - `clientKeyFile`
 
+`k8s.clientRefreshInterval`
```
jorgee, line 1060:
```suggestion
`k8s.clientRefreshInterval`
: :::{versionadded} 26.01.0-edge
  :::
```

pditommaso: Great

Issue #6751, opened 2026-01-23
Allow the use of SLURM clusters which forbid --mem and --mem-per-cpu options


phupe:
## New feature

Allow the use of SLURM clusters which forbid `--mem` and `--mem-per-cpu` options.

## Use case 

We have been unable to use nextflow of some national supercomputers from tier-0 which do not allow the use of the `--mem` and `--mem-per-cpu` options. This SLURM configuration is made on purpose to make easier the accounting of the CPU time which have been used the users. As this is the policy of these supercomputer, we have to deal with this setting that they will not change.

As a consequence, any nextflow process which sets the `memory` directive just fails since the `--mem` option is added in the `.command.run`file. Therefore, there is no way to run a nextflow pipeline unless we rewrite the it without the `memory` directive, which is not a realistic solution

## Suggested implementation 

We introduced the environment variable `NXF_EXECUTOR_MEM_GIGA_PER_CORE` which sets how many memory in GB is available per core on the supercomputer.  For example, if this variable is set to 4, any nextflow process requiring 40 GB of memory will make sure that the job requests a least 10 cores. The file `modules/nextflow/src/main/groovy/nextflow/executor/SlurmExecutor.groovy` is modified to consider thi variable if defined.

bentsherman:
I think this is going a bit too far trying to accommodate every variation of SLURM cluster out there

If you need something this custom, I would just create a plugin and duplicate / extend the slurm executor. [nf-ospool](https://github.com/nrminor/nf-ospool) is a recent example of this, where someone adapted the condor executor for a particular condor cluster with specific requirements

This way you can maintain your custom executor without going through us, and we don't have to bother you to test the code every time it is changed.

phupe:
Thanks for this suggestion. We implemented a dedicated nextflow plugin as recommended and made some tests. However, the use of nextflow plugins raises other issues on the supercomputers. This means that the only solution is that nextflow can natively handle this behaviour withou `--mem` option allowed. If this PR can not be accepted, we will have to ask the sys admin to pach/compile/install a nextflow version able to work on the supercomputer.  fingers crossed.

bentsherman:
What's the issue with plugins on your supercomputer?

phupe:
[...]

PR #6752, opened 2026-01-23
Allow the use of SLURM clusters which forbid --mem and --mem-per-cpu …

phupe:
This PR provides a solution to the issue #6751.

The variable   NXF_EXECUTOR_MEM_GIGA_PER_CORE`is considered to change the number of cores which must be allocated to the job.

The docs have been modified:
- page executor.html#slurm
- page env-vars.html


bentsherman:
Closing since this was implemented in a plugin instead

PR #6753, opened 2026-01-23
Fix false error in v2 parser when including config file from http url

### LINKED ISSUE #6749: Strict syntax: HTTPS links do not work in config includes

Including a configuration file via a link throws a parsing error:

```
  Error nextflow.config:1:1: Invalid include source: '/home/nvnieuwk/Documents/nextflow/nf-test/nf-test/test-data/process/abricate/summary/tests/https:/raw.githubusercontent.com/nf-core/modules/master/tests/config/test_data.config'
  │   1 | includeConfig "https://raw.githubusercontent.com/nf-core/modules/maste
  ╰     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I could work around this by adding a boolean check before the link like this:

```
includeConfig true ? "https://raw.githubusercontent.com/nf-core/modules/ff6d46d7898b93738c2bf95d0aa9c55a47485735/tests/config/test_data.config" : ""
```

Add the following line to any config file:

```
includeConfig "https://raw.githubusercontent.com/nf-core/modules/ff6d46d7898b93738c2bf95d0aa9c55a47485735/tests/config/test_data.config"
```

Then run the pipeline with strict syntax enabled:

```
NXF_SYNTAX_PARSER=v2 nextflow run main.nf -c <path-to-config>
```

### Environment 

* Nextflow version: 25.10.0 and up (strict syntax)
* Java version: 23
* Operating system: Linux
* Bash version: 5.0

bentsherman:
Close #6749

### DIFF
--- a/modules/nf-lang/src/main/java/nextflow/config/control/ResolveIncludeVisitor.java
+++ b/modules/nf-lang/src/main/java/nextflow/config/control/ResolveIncludeVisitor.java
@@ -76,6 +76,16 @@ public class ResolveIncludeVisitor extends ConfigVisitorSupport {
     }
 
     protected static URI getIncludeUri(URI uri, String source) {
+        // return source URI if it is already an absolute URI (e.g. http URL)
+        try {
+            var sourceUri = new URI(source);
+            if( sourceUri.getScheme() != null )
+                return sourceUri;
+        }
+        catch( Exception e ) {
+            // ignore
+        }
+        // otherwise, resolve the source path against the including URI
         return Path.of(uri).getParent().resolve(source).normalize().toUri();
     }
 
--- a/modules/nf-lang/src/test/groovy/nextflow/config/control/ResolveIncludeTest.groovy
+++ b/modules/nf-lang/src/test/groovy/nextflow/config/control/ResolveIncludeTest.groovy
@@ -19,7 +19,6 @@ package nextflow.config.control
 import java.nio.file.Path
 
 import org.codehaus.groovy.syntax.SyntaxException
-import spock.lang.Ignore
 import spock.lang.Specification
 import test.TestUtils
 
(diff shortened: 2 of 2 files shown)


modules/nf-lang/src/main/java/nextflow/config/control/ResolveIncludeVisitor.java
```diff
@@ -76,6 +76,16 @@ public void visitConfigInclude(ConfigIncludeNode node) {
     }
 
     protected static URI getIncludeUri(URI uri, String source) {
+        // return source URI if it is already an absolute URI (e.g. http URL)
+        try {
+            var sourceUri = new URI(source);
+            if( sourceUri.getScheme() != null )
+                return sourceUri;
```
pditommaso, line 83: I believe this is not enough. It should be used `FileHelper.asPath` otherwise it will not be able to resolve paths against Git resources and above all uri that requires authentication (e.g platform provided)

bentsherman: This class is just a pre-flight check, and it only checks local includes to make sure they actually exist. It doesn't bother checking remote includes for precisely the reasons you mentioned

Honestly we could just move this class back to the language server. It doesn't provide much benefit here since all includes will be validated when the config is loaded:

https://github.com/nextflow-io/nextflow/blob/66b743836d5e5ec6a6cb6a36689ae45070cdbeec/modules/nextflow/src/main/groovy/nextflow/config/parser/v2/ConfigDsl.groovy#L230-L240

bentsherman: But it is nice for the language server and `nextflow lint` to report the same errors as much as possible

PR #6764, opened 2026-01-27
Allow boolean params to implicitly default to false

bentsherman:
This PR changes the behavior of boolean params in the `params` block so that they implicitly default to `false`. This way the user doesn't have to specify `... = false` for every boolean param, which seems to be the desired behavior 99.9% of the time.

@ewels let me know if you think this makes sense


pditommaso:
What value they default now? null?

bentsherman:
Right now they don't default to anything, so you have to specify a default value yourself (e.g. ` = false`) or else the run will fail.

This is reasonable for other types, but for boolean it seems like it should just default to false if you don't declare a default value.

pditommaso:
Think `false` makes sense

modules/nextflow/src/main/groovy/nextflow/script/ParamsDsl.groovy
```diff
@@ -46,6 +46,8 @@ class ParamsDsl {
     private Map<String,Param> declarations = [:]
 
     void declare(String name, Class type, boolean optional, Object defaultValue = null) {
+        if( defaultValue == null && type == Boolean )
```
pditommaso, line 49: Nit, could be a primitive `boolean` ?

bentsherman: no, it is normalized to `Boolean` before this point

Issue #6760, opened 2026-01-26
Cannot pass booleans on command-line with strict syntax


just-simon-5:
## Bug report 

When passing a parameter that is set to a boolean on the command-line the validation fails saying the value of the parameter is a string when it is expected to be a boolean whenusing strict syntax

### Expected behavior and actual behavior

Expected behaviour:
Boolean parameters passed on the command line are evaluated to the boolean type and the workflow runs

Actual Behaviour:
Boolean parameters passed on the command line are evaluated to the string type and the workflow fails with a validation error

### Steps to reproduce the problem

I have attached a toy pipeline example you can use to recreate issue.

[illustrate_boolean_commandline_issue_strict_syntax.tar.gz](https://github.com/user-attachments/files/24859530/illustrate_boolean_commandline_issue_strict_syntax.tar.gz)

### Program output 

export NXF_SYNTAX_PARSER=v1         
nextflow run main.nf                

 N E X T F L O W   ~  version 25.10.3

Launching `main.nf` [pedantic_sanger] DSL2 - revision: d2a047b7ab

executor >  local (1)
[fe/3b6c61] FOO [100%] 1 of 1 ✔
foo

nextflow run main.nf --use_bar true 

 N E X T F L O W   ~  version 25.10.3

Launching `main.nf` [suspicious_ardinghelli] DSL2 - revision: d2a047b7ab

executor >  local (1)
[59/5fa1f8] BAR [100%] 1 of 1 ✔
bar

nextflow run main.nf --use_bar false

 N E X T F L O W   ~  version 25.10.3

Launching `main.nf` [magical_hoover] DSL2 - revision: d2a047b7ab

executor >  local (1)
[35/a6efc4] FOO [100%] 1 of 1 ✔
foo

export NXF_SYNTAX_PARSER=v2         
nextflow run main.nf                

 N E X T F L O W   ~  version 25.10.3

Launching `main.nf` [pensive_gautier] DSL2 - revision: d2a047b7ab

executor >  local (1)
[bd/36579a] FOO [100%] 1 of 1 ✔
foo

nextflow run main.nf --use_bar true 

 N E X T F L O W   ~  version 25.10.3

Launching `main.nf` [adoring_mcnulty] DSL2 - revision: d2a047b7ab

ERROR ~ Validation of pipeline parameters failed!

 -- Check '.nextflow.log' file for details
The following invalid input values have been detected:

* --use_bar (true): Value is [string] but should be [boolean]

[...]

PR #6765, opened 2026-01-27
Fix type detection of CLI params in v2 config parser

bentsherman:
Spun off from https://github.com/nextflow-io/nextflow/issues/6760#issuecomment-3806146712

This PR applies the CLI type detection from the `params` block to also work with config params. When a CLI param overrides a config param, the CLI param is converted to the appropriate type based on the default param value in the config.

Type annotations can't be specified in the config like they can in the script, but we can almost always infer the intended type from the default value:

```groovy
params.igenomes_ignore = false      // Boolean
params.max_cpus = 4                 // Integer
params.publish_mode = 'copy'        // String
params.config_profile_name = null   // (assume String)
```

Without this, users would have to assume that all config params are strings and convert them appropriately. But with this PR I think we can recover all of the functionality of the legacy type detection that users rely on, and even make it a bit more robust.

@ewels let me know if this change makes sense

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/config/parser/v2/ConfigDsl.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/config/parser/v2/ConfigDsl.groovy
@@ -128,13 +128,61 @@ class ConfigDsl extends Script {
         }
     }
 
+    /**
+     * Assign a value to a config option.
+     *
+     * When assigning a param, if the param was specified
+     * on the command line, then the command line value takes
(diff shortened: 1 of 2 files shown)


bentsherman:
For posterity -- I tried to add a warning when the CLI value can't be converted to match the declared type, but this turned out to be trickier due to the variety of integer / floating-point types. So I didn't bother with this. This PR simply recovers the previous CLI typecasting behavior for the v2 config parser.

PR #6766, opened 2026-01-27
Fix params inclusion across modules in v2 parser

### LINKED ISSUE #6762: v2 parser: params defined in including script not visible in module
## Bug report

With v2 strict syntax, params defined in the workflow script are not visible to included modules. This also impacts typed params, which cannot be used with modules.

### Test Matrix

| Test Case | v1 | v2 |
|-----------|-----|-----|
| `params.x = ...` in script + module | ✅ | ❌ |
| `params.x = ...` in config + module | ✅ | ✅ |
| `params { x: Type = ... }` in script + module | N/A | ❌ |
| `params { x: Type = ... }` in config | ❌ syntax error | ❌ syntax error |

### Impact on Typed Params

Typed params cannot be used with modules because:
1. Typed params only work in scripts (config gives syntax error)
2. But with v2, script-defined params are not visible to modules
3. The workaround (move params to config) loses typed params entirely

### Minimal reproducible example

**modules/greet/main.nf:**
```groovy
process GREET {
    publishDir { params.outdir }, mode: 'copy'

    output:
    path "greeting.txt"

    script:
    """
    echo "Hello World" > greeting.txt
    """
}
```

**main.nf (fails with v2):**
```groovy
#!/usr/bin/env nextflow

params.outdir = "results"

include { GREET } from './modules/greet/main.nf'

workflow {
    GREET()
}
```

### Results

Tested with **Nextflow 25.10.2**:

```bash
# v1: works
nextflow run main.nf

# v2: fails
NXF_SYNTAX_PARSER=v2 nextflow run main.nf
```

v2 output:
```
WARN: Access to undefined parameter `outdir`
ERROR ~ Target path for directive publishDir cannot be null
```

### Typed params in config (syntax error)

```groovy
// nextflow.config - this fails
params {
    outdir: String = "results"
}
```

```
Error nextflow.config:2:20: Unexpected input: '='
│   2 |     outdir: String = "results"
╰     |                    ^
ERROR ~ Config parsing failed
```

### Documentation vs behavior

The [module documentation](https://www.nextflow.io/docs/latest/module.html#module-parameters) states:

> "The module inherits the parameters defined *before* the `include` statement"

The [strict syntax documentation](https://www.nextflow.io/docs/latest/strict-syntax.html#using-params-outside-the-entry-workflow) describes using params in modules as a best practice issue, not a breaking change.

### Workaround

Move params to `nextflow.config` using untyped syntax:

```groovy
params.outdir = "results"
```

This works but loses the typed params feature.

bentsherman:
Close #6762 

This PR fixes a bug where params defined before an include were not passed to the included module.

Even though this pattern is now discouraged, it is still technically supported until we disallow params outside the entry workflow entirely.

The problem was that the v2 parser was not preserving the order of script declarations, so param assignments were always placed after includes in the generated Groovy code. The sorting code was taken from ScriptFormattingVisitor, which also has to preserve the original declaration order.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -105,8 +105,8 @@ allprojects {
 
         // Documentation required libraries
         groovyDoc 'org.fusesource.jansi:jansi:2.4.0'
-        groovyDoc "org.apache.groovy:groovy-groovydoc:4.0.29"
-        groovyDoc "org.apache.groovy:groovy-ant:4.0.29"
+        groovyDoc "org.apache.groovy:groovy-groovydoc:4.0.30"
+        groovyDoc "org.apache.groovy:groovy-ant:4.0.30"
     }
 
     test {
--- a/modules/nextflow/build.gradle
+++ b/modules/nextflow/build.gradle
@@ -25,12 +25,12 @@ dependencies {
     api(project(':nf-commons'))
     api(project(':nf-httpfs'))
     api(project(':nf-lang'))
-    api "org.apache.groovy:groovy:4.0.29"
-    api "org.apache.groovy:groovy-nio:4.0.29"
-    api "org.apache.groovy:groovy-xml:4.0.29"
-    api "org.apache.groovy:groovy-json:4.0.29"
-    api "org.apache.groovy:groovy-templates:4.0.29"
-    api "org.apache.groovy:groovy-yaml:4.0.29"
+    api "org.apache.groovy:groovy:4.0.30"
+    api "org.apache.groovy:groovy-nio:4.0.30"
+    api "org.apache.groovy:groovy-xml:4.0.30"
+    api "org.apache.groovy:groovy-json:4.0.30"
+    api "org.apache.groovy:groovy-templates:4.0.30"
+    api "org.apache.groovy:groovy-yaml:4.0.30"
     api "org.slf4j:jcl-over-slf4j:2.0.17"
     api "org.slf4j:jul-to-slf4j:2.0.17"
     api "org.slf4j:log4j-over-slf4j:2.0.17"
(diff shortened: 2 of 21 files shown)


modules/nf-lang/src/main/java/nextflow/script/control/ScriptToGroovyVisitor.java
```diff
@@ -86,7 +88,30 @@ protected SourceUnit getSourceUnit() {
...
+            else if( decl instanceof ParamBlockNode pbn )
+                visitParams(pbn);
+            else if( decl instanceof ParamNodeV1 pn )
+                visitParamV1(pn);
+            else if( decl instanceof ProcessNode pn )
+                visitProcess(pn);
+            else if( decl instanceof WorkflowNode wn )
+                visitWorkflow(wn);
+        }
+
```
jorgee, line 114: In super.visit() it was also considering visiting enum
```
for( var classNode : script.getClasses() ) {
            if( classNode.isEnum() )
                visitEnum(classNode);
        }
```
Is not this case missing in the modified code?

bentsherman: Since the script-to-groovy visitor doesn't override the visitEnum() method, it isn't needed here

jorgee: Ok, I was just warning about a possible side effect of this modification. Related to that, I was trying to check if there was a difference of using enums with this branch and the master, but I am getting a strange error when using enums with V2 parser. 

`BUG! exception in phase 'semantic analysis' in source unit 'Main' Cannot read the array length because "<local6>" is null`

But this is another bug that have file issue #6778.

bentsherman: Well I'm glad you checked, that's a pretty serious bug! Just submitted a fix for your review. Since that PR adds a test case with enums, we can merge that one first and then this one, just to be safe

jorgee (approved):

Issue #13, opened 2014-09-09
Show a more detailed error message when hitting GitHub rate limits


pditommaso:
GitHub public repositories have a very low rate limits. See https://developer.github.com/v3/#rate-limiting

When trying to download a repository for which the user reached the access rate limit, Nextflow reports that the repository does not exists. 

```
Cannot find cbcrg/piper-nf pipeline -- Make sure exists a GitHub repository at https://github.com/cbcrg/piper-nf
```

This is confusing the user. It should provide a more detailed error message suggesting to enter its credentials. 

```
HTTP/1.1 403 Forbidden
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: ETag, Link, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval
Content-Encoding: gzip
Content-Security-Policy: default-src 'none'
Content-Type: application/json; charset=utf-8
Date: Tue, 09 Sep 2014 12:41:08 GMT
Server: GitHub.com
Status: 403 Forbidden
Strict-Transport-Security: max-age=31536000; includeSubdomains; preload
Transfer-Encoding: chunked
X-Content-Type-Options: nosniff
X-Frame-Options: deny
X-GitHub-Media-Type: github.v3
X-GitHub-Request-Id: 545842C2:592E:19837FF:540EF564
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 0
X-RateLimit-Reset: 1410267414
X-XSS-Protection: 1; mode=block

{
    "documentation_url": "https://developer.github.com/v3/#rate-limiting", 
    "message": "API rate limit exceeded for 84.88.66.194. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)"
}
```

Issue #12, opened 2014-08-27
Error resuming process with output value and command defined on multiple lines


emi80:
I have a `test.nf` file like the following:

``` groovy
process resume {

    output:
    val str into out

    script:

    str = 'test'

    command = "echo ${str}"
    command += " > test.out"

    return command
}
```

When I run it for the first time everything goes well:

```
$ nextflow run test.nf
N E X T F L O W  ~  version 0.9.0
[warm up] executor > local
[2b/aac531] Submitted process > resume (1)
```

When I run the pipeline with the `-resume` options I get an error:

```
$ nextflow run test.nf
N E X T F L O W  ~  version 0.9.0
[warm up] executor > local
ERROR: Error executing process > 'resume (1)'

Caused by:
  Class cannot be created (missing no-arg constructor): groovy.lang.GString$3

Tip: check the log file '.nextflow.log' for more details
```

Setting `return command.toString()` did not help. Here is the stacktrace:

``` java
Aug-27 15:01:03.702 [Actor Thread 3] ERROR nextflow.processor.TaskProcessor - Error executing process > 'resume (1)'

Caused by:
  Class cannot be created (missing no-arg constructor): groovy.lang.GString$3

Tip: check the log file '.nextflow.log' for more details
Aug-27 15:01:03.706 [Actor Thread 3] DEBUG nextflow.processor.TaskProcessor - Process resume raise the following exception:
com.esotericsoftware.kryo.KryoException: Class cannot be created (missing no-arg constructor): groovy.lang.GString$3
    at com.esotericsoftware.kryo.Kryo$DefaultInstantiatorStrategy.newInstantiatorOf(Kryo.java:1228) ~[kryo-2.24.0.jar:na]
    at com.esotericsoftware.kryo.Kryo.newInstantiator(Kryo.java:1049) ~[kryo-2.24.0.jar:na]
    at com.esotericsoftware.kryo.Kryo.newInstance(Kryo.java:1058) ~[kryo-2.24.0.jar:na]
    at com.esotericsoftware.kryo.serializers.FieldSerializer.create(FieldSerializer.java:547) ~[kryo-2.24.0.jar:na]
    at com.esotericsoftware.kryo.serializers.FieldSerializer.read(FieldSerializer.java:523) ~[kryo-2.24.0.jar:na]
    at com.esotericsoftware.kryo.Kryo.readClassAndObject(Kryo.java:761) ~[kryo-2.24.0.jar:na]
[...]

Issue #11, opened 2014-08-22
Documentation code example not working


samuell:
The code example in the [output files section on the process page](https://github.com/nextflow-io/nextflow/blob/master/docs/process.rst#output-files) does not work.

The code xample in question:

``` groovy
process randomNum {

   output:
   file 'result.txt' into numbers

   '''
   echo $RANDOM > result
   '''
}

numbers.subscribe { println "Received: " + it.text }
```

When executed, I get:

``` bash
[14:45:53] ~/code/test/nextflow $ nextflow run randomnum.nf 
Picked up _JAVA_OPTIONS: -Dawt.useSystemAAFontSettings=lcd_hrgb -Dswing.aatext=true -Dsun.java2d.xrender=true
N E X T F L O W  ~  version 0.9.0
[warm up] executor > local
[47/583d32] Submitted process > randomNum (1)
ERROR: Error executing process > 'randomNum (1)'

Caused by:
  Missing output file: 'result.txt' expected by process: randomNum (1)

Command executed:

  echo $RANDOM > result

Command exit status:
  0

Command output:
  (empty)

Work dir:
  /home/samuel/code/test/nextflow/work/47/583d329642c9b78d65fadaabf6637a

Tip: view the complete command output by changing to the process work dir and entering the command: 'cat .command.out'
```

PR #6768, opened 2026-01-28
Module System Implementation

jorgee:
This pull request introduces a new `module` command-line interface for managing Nextflow modules, including installing, listing, and publishing modules. It adds a new `module` command with several subcommands, updates the CLI infrastructure to support unknown options for this command, and brings in new dependencies required for module packaging and extraction.

**Major new features and CLI enhancements:**

* Added a new `CmdModule` class implementing the `nextflow module` command, with subcommands for install, list, publish, run, remove, and search. This includes full CLI argument parsing, usage/help handling, and subcommand delegation.
* `ModuleRun` extends CmdRun installing the module and calling CmdRun with the downloaded module.

**Launcher and CLI infrastructure updates:**

* Registered the new `CmdModule` command in the main `Launcher` class so it is available as a top-level command.
* Modified the CLI parser in `Launcher` to allow unknown options for the `module` command and to pass these unknown options to the command implementation. [[1]](diffhunk://#diff-9644fefc5147689df1ce3c8f68bb23f80f6373ff47a9987c31b2cc43e94e1607R133-R135) [[2]](diffhunk://#diff-9644fefc5147689df1ce3c8f68bb23f80f6373ff47a9987c31b2cc43e94e1607R161-R163)
* Updated the base `CmdBase` class to support storing and retrieving unknown CLI options, enabling more flexible argument handling for subcommands.

**Dependency management:**

* Added `org.apache.commons:commons-compress` as a dependency for handling tar.gz extraction, which is necessary for module packaging and publishing.

**Building and Testing**
Current CI is failing becuase it requires a local development registry instance with modules API implemented. You will need it to update build.gradle to updated npr-api package with the new module API classes. Either with `includeBuild `or a Maven local repository. You will also need to deploy the registry locally and configure the `registry` field in `nextflow.config` to test the new CLI commans

**Open Issue and TODOs:**
- [x] Module version temporary in `nextflow_spec.json` for simplicity.
- [x] Print module run outputs
- [x] Remote module inclusion
- [x] Add documentation
- [x] Extend unit and integration tests


modules/nextflow/src/main/groovy/nextflow/module/ModuleManifest.groovy
```diff
@@ -0,0 +1,117 @@
...
+import java.nio.file.Path
+
+/**
+ * Represents a module manifest (meta.yaml) with validation
+ *
+ * @author Jorge Ejarque <jorge.ejarque@seqera.io>
+ */
+@Slf4j
+@CompileStatic
+class ModuleManifest {
```
bentsherman:
```suggestion
class ModuleSpec {
```

Let's use "module spec" across the board instead of "module manifest"

modules/nextflow/src/main/groovy/nextflow/util/NextflowSpecFile.groovy
```diff
@@ -0,0 +1,147 @@
...
+import java.nio.file.StandardOpenOption
+
+/**
+ * Manages the nextflow_spec.json file for module version declarations
+ *
+ * @author Jorge Ejarque <jorge.ejarque@seqera.io>
+ */
+@Slf4j
+@CompileStatic
+class NextflowSpecFile {
```
bentsherman: Let's move this to `nextflow.pipeline.PipelineSpec`

modules/nextflow/src/main/groovy/nextflow/module/ModuleManifest.groovy
jorgee: Changed

jorgee:
Some updates:

1.- I have added the print of process outputs when finishing the process execution.

2.- I have investigated the remote module inclusion and a possible implemetation:
When using the syntax parser V1, it is easy to implement the module resolution at `IncludeDef`. We can check if the module is installed and install if not exists.

However, when using syntax parser V2, the problem is that visitors and resolvers used for parsing are at `nf-lang`. So, even if I decouple the classes to download modules and import at `nf-lang`, there is no way to access the session, config builder, and we will add the npr-api classes. Not sure if it would be a good solution

So, what I have done is the following:
- In the include resolver and visitor, I just convert the `@scope/module` to the expected path and check if it exists. If there is no local copy of the module, it fails at compile time and suggest the user to install it. If there is a local installation, it checks the versions and integrity at `IncludeDef.` 
I could also check the installation of the declared modules at `CmdRun`, then it will only fail if a module is used and not declared.

bentsherman:
@jorgee let's move the remote module inclusion piece out of this PR. you can put it in a separate PR if you want, either way Paolo will experiment with it separately

And for the module version config, in the interest of keeping the POC simple so that we can iterate quicker, let's keep the module versions in the JSON file. Paolo can also experiment with this separately if he wants

jorgee:
Updated the branch as @bentsherman suggested
- Modules version temporary in `nextflow_spec.json`
- Remote inclusion PoC in #6815
- Added docs and tests
- It will be ready once changes in registry are aplied and new npr-api version generated

pditommaso:
I was giving a try to this, but cannot compile it 

```
> Task :nextflow:compileGroovy
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
startup failed:
/Users/pditommaso/Projects/nextflow/modules/nextflow/src/main/groovy/nextflow/module/ModuleRegistryClient.groovy: 24: unable to resolve class io.seqera.npr.api.schema.v1.ModuleRelease
 @ line 24, column 1.
   import io.seqera.npr.api.schema.v1.ModuleRelease
   ^

/Users/pditommaso/Projects/nextflow/modules/nextflow/src/main/groovy/nextflow/module/ModuleRegistryClient.groovy: 25: unable to resolve class io.seqera.npr.api.schema.v1.PublishModuleResponse
 @ line 25, column 1.
   import io.seqera.npr.api.schema.v1.PublishModuleResponse
   ^

/Users/pditommaso/Projects/nextflow/modules/nextflow/src/main/groovy/nextflow/module/ModuleRegistryClient.groovy: 205: unable to resolve class io.seqera.npr.api.schema.v1.ModuleRelease
 @ line 205, column 37.
                return new GsonEncoder<ModuleRelease>() {}.decode(body)
                                       ^

/Users/pditommaso/Projects/nextflow/modules/nextflow/src/main/groovy/nextflow/module/ModuleRegistryClient.groovy: 472: unable to resolve class io.seqera.npr.api.schema.v1.PublishModuleResponse
 @ line 472, column 36.
               return new GsonEncoder<PublishModuleResponse>() {}.decode(body)
                                      ^

/Users/pditommaso/Projects/nextflow/modules/nextflow/src/main/groovy/nextflow/module/ModuleRegistryClient.groovy: 205: unable to resolve class io.seqera.npr.api.schema.v1.ModuleRelease
 @ line 205, column 37.
                return new GsonEncoder<ModuleRelease>() {}.decode(body)
                                       ^

/Users/pditommaso/Projects/nextflow/modules/nextflow/src/main/groovy/nextflow/module/ModuleRegistryClient.groovy: 205: unable to resolve class io.seqera.npr.api.schema.v1.ModuleRelease
 @ line 205, column 37.
                return new GsonEncoder<ModuleRelease>() {}.decode(body)
                                       ^

/Users/pditommaso/Projects/nextflow/modules/nextflow/src/main/groovy/nextflow/module/ModuleRegistryClient.groovy: 472: unable to resolve class io.seqera.npr.api.schema.v1.PublishModuleResponse
 @ line 472, column 36.
               return new GsonEncoder<PublishModuleResponse>() {}.decode(body)
                                      ^

/Users/pditommaso/Projects/nextflow/modules/nextflow/src/main/groovy/nextflow/module/ModuleRegistryClient.groovy: 472: unable to resolve class io.seqera.npr.api.schema.v1.PublishModuleResponse
 @ line 472, column 36.
               return new GsonEncoder<PublishModuleResponse>() {}.decode(body)
                                      ^

8 errors

```

jorgee:
To test this branch you need to do the following:

1. Clone the registry repo and clone branch `251215-module-binary-registry-fixes`

2. Build and publish the `npr-api` version 0.20.1 in Maven local
```
$ make build
$ ./gradlew :npr-api:publishToMavenLocal 
```

3. Deploy a local docker registry to store modules
`$ docker run -d --name docker-registry -p 5010:5000 registry:latest`

4. Run the Nextflow registry
`$ bash run.sh `

5. Clone Nextflow repo and checkout this branch

7.  Include the Maven local in the `build.gradle` file
```
         mavenCentral()
         maven { url 'https://repo.eclipse.org/content/groups/releases' }
         maven { url 'https://oss.sonatype.org/content/repositories/snapshots' }
+        mavenLocal()
         maven { url = "https://s3-eu-west-1.amazonaws.com/maven.seqera.io/releases" }
         maven { url = "https://s3-eu-west-1.amazonaws.com/maven.seqera.io/snapshots" }
```
8. Build Nextflow as usual

pditommaso:
@jorgee i've made a force push by mistake, please check I've not removed anything

jorgee:
Update: 
- Add `module info` command
- Update to registry API to v1
- /api path expected in registry config URL
- `search `only prints name and description (full for -json)

To test, you must run with registry branch `fix-module-release-metadata`

jorgee:
Updated docs and other fixes. To test in dev, add the following lines in `nextflow.config`
```
registry {
     url='https://registry-dev.nextflow.io/api'
     apiKey='YOUR_REGISTRY_TOKEN'
}
```

jorgee:
@claude review

docs/reference/cli.md
```diff
@@ -1125,6 +1125,215 @@ $ nextflow log tiny_leavitt -F 'process =~ /split_letters/'
...
+
+(cli-module-list)=
+
+`list [options]`
+
+: List all modules currently installed in your project.
+: Shows module names, versions, and integrity status (whether they've been modified locally).
+: The following options are available:
+
+  `-json`
```
pditommaso: We tend to use `-o, -output` for this, see for example `lint` or `config` commands

docs/reference/cli.md
```diff
@@ -1125,6 +1125,215 @@ $ nextflow log tiny_leavitt -F 'process =~ /split_letters/'
...
+`search [options] [query]`
+
+: Search for modules in the registry by keyword or name.
+: Returns modules matching the query with their names, versions, descriptions, and download statistics.
+: The following options are available:
+
+  `-limit`
+  : Maximum number of results to return (default: varies by registry).
+
+  `-json`
```
pditommaso: Same `-o, -output`

docs/reference/cli.md
```diff
@@ -1125,6 +1125,215 @@ $ nextflow log tiny_leavitt -F 'process =~ /split_letters/'
...
+The `module` command provides a comprehensive system for managing reusable, registry-based modules. It enables installing modules from registries, running them directly, searching for available modules, and publishing your own modules for community use.
+
+**Subcommands**
+
+(cli-module-install)=
+
+`install [options] [scope/name]`
+
+: Install a module from the registry into your project.
+: Downloaded modules are stored in the `modules/` directory and version information is tracked in `nextflow_spec.json`.
```
pditommaso: I think it's not a good idea introduce yet another config file for modules. Version should go into nextflow.config along plugins versions

docs/cli.md
```diff
@@ -262,7 +262,136 @@ $ nextflow secrets set AWS_ACCESS_KEY_ID
...
+The `module install` command downloads modules from a registry and makes them available in your workflow. Modules are stored locally in the `modules/` directory and version information is tracked in `nextflow_spec.json`.
+
+Use this to add reusable modules to your pipeline, manage module versions, or update modules to newer versions.
+
+```console
+$ nextflow module install nf-core/fastqc
+$ nextflow module install nf-core/fastqc -version 1.0.0
+```
+
+After installation, module will be available in `modules/@nf-core/fastqc` and included in `nextflow_spec.json`
```
pditommaso: What's the outcome of not using `@` prefix? using a marker file could solve the ambiguity between nf-core modules and nextflow manage modules?

docs/reference/cli.md
```diff
@@ -1125,6 +1125,215 @@ $ nextflow log tiny_leavitt -F 'process =~ /split_letters/'
...
+
+: Remove a module from your project.
+: By default, removes both local files and configuration entries. Use options to control what gets removed.
+: The following options are available:
+
+  `-keep-config`
+  : Keep the version entry in `nextflow_spec.json` but delete local files from the `modules/` directory.
+
+  `-keep-files`
+  : Remove the version entry from `nextflow_spec.json` but keep local files in the `modules/` directory.
```
pditommaso: sounds a bit overkill to me these flag

pditommaso:
# Module System Code Review

## CRITICAL (Must Fix)

### 1. Logic Bug — `CmdModuleSearch` validation (dead code)
`CmdModuleSearch.groovy:64` — `if (!args && args.size() != 1)` is impossible: if `!args` is true, `&&` short-circuits. Should be `||`.

### 2. Inverted Condition — `CmdModuleInfo` (wrong log + potential NPE)
`CmdModuleInfo.groovy:102-109` — Logs "No metadata found" when metadata **IS** present. Additionally, if `fetchRelease` throws, `release` stays null and the code falls through to `printFormattedInfo` which will NPE.

### 3. `e.printStackTrace()` in production code
`ModuleRegistryClient.groovy:138` — Writes directly to stderr, bypassing the logging framework. Replace with `log.debug` or `log.error`.

---

## IMPORTANT (Should Fix)

### 5. Checksum header name mismatch between code and tests
Production code checks `X-NF-Module-Checksum`, but tests stub `X-Checksum`. Tests pass but don't actually validate the real header.

### 6. `Files.list()` / `Files.walk()` stream resource leaks
`ModuleStorage.groovy:126,171,409`, `ModuleChecksum.groovy:54`, `CmdModulePublish.groovy:222` — These streams must be closed (try-with-resources or `.withCloseable{}`). Compare with `ModuleStorage.groovy:345` where this is done correctly.

### 7. `removeModuleEntry` always returns true
`PipelineSpec.groovy:82-96` — Doesn't check if the key actually existed before returning true. `Map.remove()` returns the previous value (or null); the return value should be checked.

### 8. Bundle sent as raw bytes in JSON
`CmdModulePublish.groovy:129-135` — `byte[]` in Gson JSON will be huge/inefficient. Consider multipart upload.

### 9. `Files.readAllBytes()` for checksum computation
`ModuleChecksum.groovy:67,131` — Loads entire files into memory. Use `DigestInputStream` or read in chunks for large files.

### 10. Missing `ModulesConfig` in `extensions.idx`
`RegistryConfig` is registered but `ModulesConfig` is not — may not be discovered by the config scope mechanism.

---

## SUGGESTIONS (Nice to Have)

- **#11** Typo: "diretory" → "directory" in `CmdModulePublish.groovy:253`
- **#12** Unicode checkmark in CLI output may not render on all terminals
- **#13** `CmdModuleRun extends CmdRun` exposes all CmdRun params — consider composition
- **#14** Duplicate config-building boilerplate across all CLI commands — extract to shared method

---

## Strengths

- Zip slip / tar slip protection is correctly implemented
- Good checksum verification with cleanup on mismatch
- Solid test coverage with WireMock integration tests
- Clean CLI subcommand pattern consistent with existing `CmdPlugin`
- Proper Apache 2.0 headers, `@CompileStatic`, `@Slf4j`, Groovy idioms throughout

pditommaso:
Apologies the spam, but it looks valid points ☝️

docs/cli.md
jorgee: `.checksum` could be used to know when the module is a nextflow managed module.

docs/reference/cli.md
jorgee: To me `-keep-config` is the redundant one and seems against the approach that files must be commited in the pipeline. I see `-keep-files` useful if you want to continue having the code in the pipeline but do not manage it as a module.

docs/reference/cli.md
jorgee: Not a strong opinion either in favor or against. 
I see Paolo's point that, in current state, it seems more appropriate to put in `nextflow.config` to do not add another file, However, it depends on the future plans and if `nextflow_spec.json` would include params, outputs and maybe data that is currently in the manifest. 

`ModulesConfig` already implements the read in `nextflow.config`. We just need to implement the write/remove when installing a module. Or remove the class, if we finally keep the `nextflow_spec.json`

This review suggestion is also related to this topic. (Note to me to do not forget)
> 10. Missing ModulesConfig in extensions.idx
RegistryConfig is registered but ModulesConfig is not — may not be discovered by the config scope mechanism.

docs/reference/cli.md
jorgee: changed

docs/reference/cli.md
jorgee: changed

jorgee:
Update:
- Addressed critical and important fixes except 8 (not required because bundle is limited to 1MB not compressed) and 10 (depends on where we decide to write module versions)
- Change `-json` by `-output`
- Add module URL in `info` command

TODO/to be decided:

- [x] `remove` flags
- [x] remove '@' (I will create another PR as it require changes also in remote inclusion)
- [ ] module versions in `nextflow_spec.json` or `nextflow.config`

docs/cli.md
pditommaso: let's use `.module-checksum` or could make sense to turn it into a property file `.module-info` with a `checksum=` attribute?

docs/reference/cli.md
bentsherman: Indeed, it's not "yet another config file", it's the pipeline spec we are already planning to use

Using nextflow config is not viable because it makes the modules runtime-configurable when in reality they are not

pditommaso: We already touched this point, and indeed the explicit intent is to make runtime-configurable

bentsherman: Maybe, in the future, but we aren't working on that right now and we haven't decided on an approach for it

If module versions in config turns out to be a good approach for that, then we can implement it at that point. Until then it needs to go in a separate JSON file

docs/cli.md
jorgee: implemented with `.module-info` with a checksum attribute

pditommaso: Nice

docs/reference/cli.md
pditommaso: Sounds good

jorgee:
I have removed `-keep-config` for `module remove.` I have also changed its behaviour:
- It fails when folder exists and `.module-info` doesn't exist or when there are modifications
- Added the ` -force` flag to remove in above cases
- When removing with `-keep-files`, it removes the entry in spec and the `.module-info` file

docs/reference/cli.md
bentsherman: Weekend shower thought -- what if we did neither? Store the module versions in the `.module-info` files instead?

Only limitation is that if you want to pin a module version, you have to commit the module to source control. But we were already willing to accept this restriction in the first iteration.

One-off scripts could still include remote modules, they would just get the latest version on first run and keep using that version on subsequent runs

Then we can add the `nextflow_spec.json` or `modules` config as a future extension which takes precedence over `.module-info`. That gives us time to get a clearer picture of these two options.

@jorgee do you think this would work?

docs/module.md
```diff
@@ -279,8 +279,179 @@ This feature requires the use of a local or shared file system for the pipeline
...
+### Installing modules from a registry
+
+Use the `module install` command to download modules from a registry:
+
+```console
+$ nextflow module install nf-core/fastqc
+$ nextflow module install nf-core/fastqc -version 1.0.0
+```
+
+Installed modules are stored in the `modules/` directory and can be included by name instead of by relative path:
```
christopher-hakkaart:
```suggestion
Installed modules are stored in the `modules/` directory and can be included by name instead of a relative path:
```

docs/module.md
```diff
@@ -279,8 +279,179 @@ This feature requires the use of a local or shared file system for the pipeline
...
+$ nextflow module publish myorg/my-module
+```
+
+The argument can be either a `scope/name` reference (for an already-installed module) or a local directory path containing the module files.
+
+Your module directory must include:
+
+- `main.nf` - The module entry point
+- `meta.yaml` - Module metadata (name, description, version, etc.)
+- `README.md` - Module documentation
```
christopher-hakkaart:
```suggestion
- `main.nf`: The module entry point
- `meta.yaml`: Module metadata (name, description, version, etc.)
- `README.md`: Module documentation
```

docs/module.md
```diff
@@ -279,8 +279,179 @@ This feature requires the use of a local or shared file system for the pipeline
...
+        ├── README.md        # Documentation (required for publishing)
+        ├── main.nf          # Module entry point (required)
+        ├── meta.yaml        # Module metadata (required for publishing)
+        ├── resources/       # Optional: module binaries and resources
+        └── templates/       # Optional: process templates
+```
+
+The `modules/` directory should be committed to your Git repository to ensure reproducibility.
+
+See the {ref}`cli-page` documentation for complete details on all module commands.
```
christopher-hakkaart:
```suggestion
See the {ref}`cli-page` documentation for more information about module commands.
```
My preference, but just a suggestion

docs/reference/cli.md
jorgee: It will work with the limitation you have mentioned. I do not see the need to store the version in `.module-info`, it is already in the `meta.yml`. Versions are mainly used by the remote module inclusion when the required module is not found in the `modules` directory or the version doesn't match. I think it is also used in `install` when the version is not specified. But it is not a big deal. It could also refer to the latest in this case.

pditommaso: I reached the same conclusion, versions pinned was added in the spec when we were planning to have multiple local version. However in the current form is mostly redundant, +1 to remove external file pinning

bentsherman: > I do not see the need to store the version in `.module-info`, it is already in the `meta.yml`.

@jorgee Fair enough, but do you think it would be worth the performance benefit to duplicate the version in the `.module-info` ?

I assume we are already reading that file every time to get the checksum, whereas I'm not sure if we are always reading the `meta.yml`, and you would have to read the whole yml file just to get the version

It's probably not a big deal either way... they would be read at most once per run and they will always be local so it doesn't need to be super fast

bentsherman:
Merging into the main PR since we've resolved all of the requirements, let's continue to test and refine there

docs/cli.md
```diff
@@ -262,7 +262,135 @@ $ nextflow secrets set AWS_ACCESS_KEY_ID
...
+The `module run` command executes a module directly from the registry without requiring a wrapper workflow. This provides immediate access to module functionality for ad-hoc tasks or testing.
+
+Use this to quickly run a module, test module functionality, or execute one-off data processing tasks.
+
+```console
+$ nextflow module run nf-core/fastqc --input 'data/*.fastq.gz'
+$ nextflow module run nf-core/fastqc --input 'data/*.fastq.gz' -version 1.0.0
+```
+
+The command accepts all standard Nextflow execution options (`-profile`, `-resume`, etc.):
```
ewels, line 304: We don't really say in the new docs where `--reads` and `--index` come from.  It'd be good to have a bit more guidance about how people should know what CLI flags to use.

Could we get `nextflow module run [name] --help` to list available flags? That'd be really useful!

bentsherman: Let's move this to #6650

PR #6769, opened 2026-01-28
ADR: module parameters

pditommaso:
## Summary

Introduces ADR for module parameters specification, defining how parameters can be declared and configured in Nextflow modules.

Key aspects covered:
- Parameter specification in `meta.yaml` for documentation, validation, and IDE support
- Two options for script-level parameter definition:
  - **Option A**: Process-level `params:` block (scoped to process)
  - **Option B**: Module-level `params {}` block (accessible in workflow)
- Parameter semantics: typing, defaults, immutability
- Configuration and CLI override mechanisms
- Comparison with legacy `ext.args` approach

## Related

- Module System ADR: `adr/20251114-module-system.md`

---

### DIFF
--- a/adr/2026-01-28-module-parameters.md
+++ b/adr/2026-01-28-module-parameters.md
@@ -1,0 +1,224 @@
+# Module Parameters
+
+- Authors: Paolo Di Tommaso
+- Status: draft
+- Date: 2025-01-28
+- Tags: modules, parameters, configuration
+- Related: [Module System ADR](20251114-module-system.md)
+
+## Context
+
+The module system introduces a structured approach to module configuration through parameters defined in `meta.yaml`. Parameters provide a documented, type-safe way to customize module behavior.
+
+This document describes two aspects of module parameters:
+1. **Specification** (`meta.yaml`) - declares parameter metadata for documentation, validation, and IDE support
+2. **Definition in script** - how parameters are declared and used in the Nextflow module script
+
+## Parameter Specification (meta.yaml)
+
+Modules declare available parameters in `meta.yaml` under the `params` section. Each parameter has a name and optional attributes for type and description.
+
+```yaml
+params:
+  - name: batch_size
+    type: integer
+    description: "Process INT input bases in each batch"
+    example: 100000000
+
+  - name: use_soft_clipping
+    type: boolean
+    description: "Use soft clipping for supplementary alignments"
+
+  - name: output_format
+    type: string
+    description: "Output format (sam, bam, or cram)"
+```
+
+**Specification Attributes:**
+
+| Attribute | Required | Description |
+|-----------|----------|-------------|
+| `name` | Yes | Parameter identifier |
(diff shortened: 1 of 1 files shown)


adr/2026-01-28-module-parameters.md
```diff
@@ -0,0 +1,224 @@
...
+}
+```
+
+## CLI Override
+
+Parameters can be overridden via CLI:
+
+```bash
+# For standard workflow execution
+nextflow run <script> -process.BWA_MEM.batch_size=100000000
```
pditommaso:
```suggestion
nextflow run <script> -process.BWA_MEM.params.batch_size=100000000
```

ewels:
~~Note to self: this replaces the previously proposed `args` syntax.~~

~~See previously proposed syntax here: https://github.com/nextflow-io/nextflow/blob/474eb758c5a7b9510d43edc87932b2dedcd846c2/adr/20251114-module-system.md#tool-arguments-configuration~~

_Edit: incorrect._ This ADR goes _in addition to_ https://github.com/nextflow-io/nextflow/pull/6770

bentsherman:
One approach we are looking at with records is to refactor the ext settings as fields in the input record.

For example, nf-core/methylseq uses the [bedtools/intersect](https://github.com/nf-core/methylseq/blob/dev/modules/nf-core/bedtools/intersect/main.nf) module, which has several ext settings:

```groovy
    def args = task.ext.args ?: ''
    def prefix = task.ext.prefix ?: "${meta.id}"
    extension = task.ext.suffix ?: "${intervals1.extension}"
```

The methylseq pipeline customizes these settings [further](https://github.com/nf-core/methylseq/blob/dev/conf/modules/bedtools_intersect.config):

```groovy
process {
    withName: BEDTOOLS_INTERSECT {
        ext.args   = ''
        ext.prefix = { "${intervals1.baseName}" }
        ext.suffix = 'targeted.bedGraph'
    }
}
```

With records, we could provide these ext settings in the input record:
```groovy
// process definition
input:
(
    id: String,
    intervals1: Path,
    intervals2: Path,
    args: String,
    prefix: String,
    suffix: String
): Record
```

Then define them in an extra map operator
```groovy
// workflow logic
ch_intersect_inputs = ch_inputs
    .map { r -> record(
        id: r.id, 
        intervals1: r.bedgraph,
        args: '',
        prefix: r.bedgraph.baseName,
        suffix: 'targeted.bedGraph'
    ) }

ch_bedgraph_filtered = BEDTOOLS_INTERSECT( ch_intersect_inputs, intervals2: target_regions )
```

If this works out, we might not need module params at all. Especially since module params are constants but these ext settings often need to be dynamic (like `prefix` in the above example). But we'll see.

pditommaso:
> With records, we could provide these ext settings in the input record

This aligns with the point that "mutable" parameters should be managed as proper inputs. It makes sense to me

pditommaso:
Superseded by #7260, which unifies the module-parameters and tool-arguments specs into a single ADR. Closing in favor of that consolidated PR.

Issue #6772, opened 2026-01-28
Change the LSFExecutor killCommand to use '-d'


cgoina:
## New feature

With LSFExecutor terminate the LSF job using 'bkill -d' not just 'bkill'. This will mark the LSF job as DONE not as FAILED, which is the correct status from the LSF scheduler's perspective. This flag was introduced to be able to terminate Dask jobs and Spark Jobs and not show up as failed jobs. Which is a very similar case when the LSF jobs are started by Nextflow.

## Use case 

The main problem we encounter is when we create a Dask cluster or Spark cluster as part of a pipeline and some other part of the pipeline fails and Nextflow now automatically kills all tasks. When it kills the Dask/Spark worker tasks using 'bkill' instead of 'bkill -d' those jobs show as failed jobs but from LSF scheduler's point of view they haven't failed. The jobs were terminated as intended.

## Suggested implementation 

In LsfExecutor change getKillCommand to
```
    @Override
    protected List<String> getKillCommand() { ['bkill', '-d'] }
```

bentsherman:
Not sure I understand. When Nextflow kills jobs, either because the run was cancelled or aborted, those jobs are failed jobs from Nextflow's perspective. They did not complete successfully, their outputs cannot be used downstream. I don't see why LSF would treat them any differently

cgoina:
The real issue for example here is that one job fails and then nextflow kills all the others as intended. All jobs show the correct termination status from Nextflow's perspective, but from the LSF scheduler perspective the job aborted by Nextflow shows as a failed job, even though they didn't fail - they were terminated as intended by Nextflow. This is very similar to killing the a spark cluster because your driver failed. The flag was added for bkill exactly for this purpose. I've been running this code now and I no longer get complaints from our cluster admins. Before, they complained a lot because I ran a lot of spark workers or dask workers in a nextflow pipeline and those jobs are cancelled when something else fails. Because of that they all show as failed on the LSF dashboard and that triggers some alarm for our admins to check if anything is wrong with the cluster. Especially when many jobs get collocated on one node and all jobs on that node fail, it's a red flag and they need to check if there's a real problem with the node itself.

bentsherman:
I see. But my concern is that this is essentially a convention used by your specific cluster. Is this a widely used convention for LSF?

Because if it isn't, then changing it for your case might disrupt other clusters following a different convention (e.g. why are my jobs showing up as completed when that run failed)

cgoina:
[...]

PR #6784, opened 2026-01-30
Use -d flag for bkill in LSF Executor

cgoina:
This is the PR for https://github.com/nextflow-io/nextflow/issues/6772 to add '-d' flag when killing the LSF job. Our cluster admins suggested that this is the correct behavior because even if Nextflow kills the LSF job from the LSF scheduler's perspective the job was terminated as intended. 
I've been running this on our LSF cluster and it looks good.

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/executor/LsfExecutorTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/executor/LsfExecutorTest.groovy
@@ -471,7 +471,7 @@ class LsfExecutorTest extends Specification {
         // executor stub object
         def executor = createExecutor()
         then:
-        executor.killTaskCommand('12345').join(' ') == 'bkill 12345'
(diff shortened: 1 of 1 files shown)


pditommaso:
This may be a breaking change 

```
  The -d flag appears to have been introduced in LSF 10.1.0 (IBM Spectrum LSF). It is not present in earlier versions:                                           
                                                                                                                                                                 
  - LSF 9.1.2: No -d flag (https://www.bsc.es/support/LSF/9.1.2/lsf_command_ref/bkill.1.html)                                                                    
  - LSF 10.1.0: -d flag is documented — "Kills the jobs, then records the jobs as DONE after the jobs exit"                                                      
  (https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=reference-bkill)       
```

bentsherman:
Looks like it was released 10 years ago: https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=new-whats-in-lsf-101

PR #6785, opened 2026-01-30
Improve error messages for workflow outputs

bentsherman:
Inspired by this [Slack thread](https://nextflow.slack.com/archives/C02T98A23U7/p1769782002120739)

This PR improves several error messages that can be thrown by the `output` block.

These errors can be tricky because they occur during pipeline execution, so it helps to know (1) which workflow output failed and (2) any invalid input values that might be relevant

### DIFF
--- a/modules/nf-lang/src/main/antlr/ScriptParser.g4
+++ b/modules/nf-lang/src/main/antlr/ScriptParser.g4
@@ -403,7 +403,7 @@ variableDeclaration
     ;
 
 nameTypePairs
-    :   LPAREN nameTypePair (COMMA nameTypePair)+ rparen
+    :   LPAREN nls nameTypePair (COMMA nls nameTypePair)+ nls rparen
     ;
 
 nameTypePair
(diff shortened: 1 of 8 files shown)


modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
```diff
@@ -66,18 +66,38 @@ class PublishOp {
...
     }
 
+    /**
+     * Perform an action. If an exception is raised, bind the
+     * excpetion to the target and don't perform any more actions.
+     *
+     * @param action
+     */
+    private void safeExecute(Runnable action) {
+        if( target.isError() )
```
jorgee, line 92: I am wondering if the management of these errors should differ if the `failOnError` is set or not. It looks like it is always applying` failOnError= true`, while if `failOnError` is set to false, it should ignore the errors, shouldn't it?

bentsherman: That flag is mainly to enable a legacy publishDir behavior -- if something goes wrong when publishing a file, it decides whether or not to fail the run.

Once publishDir is removed, this flag probably will be removed too, so I don't plan on applying it anywhere else

jorgee (approved):

PR #6799, opened 2026-02-04
Add OSC 8 hyperlinks to task hashes in progress output

ewels:
## Summary
This PR adds support for OSC 8 terminal hyperlinks to make task hashes clickable in the progress output. When a task's work directory is available and cleanup is not enabled, the task hash is rendered as a clickable hyperlink that opens the work directory. Hyperlinks are also added to the pipeline repo commit hash if possible. 

Works in most terminal emulators (tested in iTerm2 and VSCode). Gracefully ignores the ANSI codes in rubbish terminal emulators and treats as normal text (tested in MacOS Terminal). Uses native protocol handling, so iTerm2 loads work directories in Finder (MacOS) and VSCode shows them in the file explorer side bar.

Links only activate when the <kbd>cmd</kbd> key is held on hover (<kbd>alt</kbd> in VSCode).

https://github.com/user-attachments/assets/369e67f7-ed48-4bc7-8e1b-53f3f016af23

## Key Changes
- **AnsiLogObserver**: Added `hyperlink()` static method to wrap text in OSC 8 escape sequences, with automatic `file://` prefix for local paths
- **AnsiLogObserver**: Modified `line()` method to render task hashes as hyperlinks when `workDir` is available and cleanup is disabled
- **ProgressRecord**: Added `workDir` field to track the current task's work directory URI
- **WorkflowStats**: Updated `markSubmitted()`, `markStatusChange()`, and `markCached()` methods to populate the `workDir` field from task metadata
- **Tests**: Added comprehensive test coverage for hyperlink generation and conditional rendering based on cleanup settings

## Implementation Details
- Uses BEL (`\007`) as the String Terminator for better terminal compatibility
- Automatically converts local paths (starting with `/`) to `file://` URIs
- Preserves URLs with existing schemes (`s3://`, `gs://`, `az://`) as-is
- Hyperlinks are only rendered when cleanup is disabled, since work directories are removed when cleanup is enabled
- Gracefully handles null or empty URLs by returning plain text

https://claude.ai/code/session_014AStcyMzd39iegXhfQL1yK

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
@@ -599,41 +599,42 @@ class CmdRun extends CmdBase implements HubOptions {
         /*
          * try to look for a pipeline in the repository
          */
-        def manager = new AssetManager(pipelineName, this)
-        if( revision )
-            manager.setRevision(revision)
-        def repo = manager.getProjectWithRevision()
+        try (def manager = new AssetManager(pipelineName, this)) {
+            if( revision )
+                manager.setRevision(revision)
+            def repo = manager.getProjectWithRevision()
 
-        boolean checkForUpdate = true
-        if( !manager.isRunnable() || latest ) {
-            if( offline )
-                throw new AbortOperationException("Unknown project `$repo` -- NOTE: automatic download from remote repositories is disabled")
-            log.info "Pulling $repo ..."
-            def result = manager.download(revision,deep)
-            if( result )
-                log.info " $result"
-            checkForUpdate = false
-        }
-        // Warn if using legacy
-        if( manager.isUsingLegacyStrategy() ){
-            log.warn1 "This Nextflow version supports a new Multi-revision strategy for managing the SCM repositories, " +
-                "but '${repo}' is single-revision legacy strategy - Please consider to update the repository with the 'nextflow pull -migrate' command."
-        }
-        // post download operations
-        try {
-            manager.checkout(revision)
-            manager.updateModules()
-            final scriptFile = manager.getScriptFile(mainScript)
-            if( checkForUpdate && !offline )
-                manager.checkRemoteStatus(scriptFile.revisionInfo)
-            // return the script file
-            return scriptFile
-        }
-        catch( AbortOperationException e ) {
-            throw e
-        }
-        catch( Exception e ) {
-            throw new AbortOperationException("Unknown error accessing project `$repo` -- Repository may be corrupted: ${manager.localPath}", e)
+            boolean checkForUpdate = true
+            if( !manager.isRunnable() || latest ) {
+                if( offline )
+                    throw new AbortOperationException("Unknown project `$repo` -- NOTE: automatic download from remote repositories is disabled")
+                log.info "Pulling $repo ..."
+                def result = manager.download(revision,deep)
+                if( result )
+                    log.info " $result"
+                checkForUpdate = false
+            }
+            // Warn if using legacy
+            if( manager.isUsingLegacyStrategy() ){
+                log.warn1 "This Nextflow version supports a new Multi-revision strategy for managing the SCM repositories, " +
+                    "but '${repo}' is single-revision legacy strategy - Please consider to update the repository with the 'nextflow pull -migrate' command."
+            }
+            // post download operations
+            try {
+                manager.checkout(revision)
+                manager.updateModules()
+                final scriptFile = manager.getScriptFile(mainScript)
+                if( checkForUpdate && !offline )
+                    manager.checkRemoteStatus(scriptFile.revisionInfo)
+                // return the script file
+                return scriptFile
+            }
+            catch( AbortOperationException e ) {
+                throw e
+            }
+            catch( Exception e ) {
+                throw new AbortOperationException("Unknown error accessing project `$repo` -- Repository may be corrupted: ${manager.localPath}", e)
+            }
         }
 
     }
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdClone.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdClone.groovy
@@ -52,25 +52,25 @@ class CmdClone extends CmdBase implements HubOptions {
         Plugins.init()
         // the pipeline name
         String pipeline = args[0]
-        final manager = new AssetManager(pipeline, this)
+        try (final manager = new AssetManager(pipeline, this)) {
+            // the target directory is the second parameter
+            // otherwise default the current pipeline name
+            def target = new File(args.size()> 1 ? args[1] : manager.getBaseName())
+            if( target.exists() ) {
+                if( target.isFile() )
+                    throw new AbortOperationException("A file with the same name already exists: $target")
+                if( !target.empty() )
+                    throw new AbortOperationException("Clone target directory must be empty: $target")
+            }
+            else if( !target.mkdirs() ) {
+                throw new AbortOperationException("Cannot create clone target directory: $target")
+            }
 
-        // the target directory is the second parameter
-        // otherwise default the current pipeline name
-        def target = new File(args.size()> 1 ? args[1] : manager.getBaseName())
-        if( target.exists() ) {
-            if( target.isFile() )
-                throw new AbortOperationException("A file with the same name already exists: $target")
-            if( !target.empty() )
-                throw new AbortOperationException("Clone target directory must be empty: $target")
+            manager.checkValidRemoteRepo()
+            print "Cloning ${manager.getProjectWithRevision()} ..."
+            manager.clone(target, revision, deep)
+            print "\r"
+            println "${manager.getProjectWithRevision()} cloned to: $target"
         }
-        else if( !target.mkdirs() ) {
-            throw new AbortOperationException("Cannot create clone target directory: $target")
-        }
-
-        manager.checkValidRemoteRepo()
-        print "Cloning ${manager.getProjectWithRevision()} ..."
-        manager.clone(target, revision, deep)
-        print "\r"
-        println "${manager.getProjectWithRevision()} cloned to: $target"
     }
 }
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdConfig.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdConfig.groovy
@@ -233,13 +233,13 @@ class CmdConfig extends CmdBase {
             return file.parent ?: Paths.get('/')
         }
 
-        final manager = new AssetManager(path, revision)
-        if( revision && manager.isUsingLegacyStrategy() ){
-            log.warn("The local asset for ${path} does not support multi-revision - 'revision' option is ignored\n" +
-                "Consider updating the project using 'nextflow pull ${path} -r $revision -migrate'")
+        try (final manager = new AssetManager(path, revision)) {
+            if( revision && manager.isUsingLegacyStrategy() ){
+                log.warn("The local asset for ${path} does not support multi-revision - 'revision' option is ignored\n" +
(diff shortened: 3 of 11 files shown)


modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -476,7 +476,41 @@ class CmdRun extends CmdBase implements HubOptions {
...
+
+    /**
+     * Constructs a web URL for viewing the repository at a specific commit.
+     * Supports GitHub, GitLab, Bitbucket, Gitea, and Azure DevOps.
+     *
+     * @param repository The repository URL (e.g., https://github.com/nextflow-io/hello)
+     * @param commitId The commit SHA
+     * @return The repository tree URL at the commit, or null if the provider is not recognized
+     */
+    protected static String getCommitUrl(String repository, String commitId) {
```
pditommaso: Why only specific host and not any uri ?

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -498,17 +532,28 @@ class CmdRun extends CmdBase implements HubOptions {
...
             fmt.fg(Color.CYAN).a("revision: ").reset()
-            fmt.fg(Color.CYAN).a(revision).reset()
+            // Make commit hash a hyperlink if commit URL is available
+            if( commitUrl && commitId ) {
+                final shortCommit = commitId.substring(0, Math.min(10, commitId.length()))
+                fmt.fg(Color.CYAN).a("\033]8;;${commitUrl}\007${shortCommit}\033]8;;\007").reset()
+                // Append the branch/tag name if revision contains more than just the commit
+                if( revision && revision != shortCommit && revision.startsWith(shortCommit) ) {
+                    fmt.fg(Color.CYAN).a(revision.substring(shortCommit.length())).reset()
+                }
```
pditommaso: This should likely go into an helper method

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
ewels: Because we have to go to from repo URI + commit to the "browse commit" URL, and how that's constructed varies between providers.

eg. comparing a GitHub and GitLab URL:

```
https://github.com/nf-core/modules/tree/c2a093c6163d30c1806397f042faff4e49c15930
https://gitlab.com/nf-core/modules-test/-/tree/8a5cfbce0c1d5a64a868871c3f48216916df3921
```

Gitlab has the extra `/-/` in the path. Bitbucket adds a `/src/`, etc. etc. I don't think that there is any way to get this automatically.

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
ewels: Done in 4a3a20e9868da7b64ec5209ae6e5f3044b2fba15

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -476,7 +476,76 @@ class CmdRun extends CmdBase implements HubOptions {
...
+            return "${repository}/-/tree/${commitId}"
+        // Bitbucket: https://bitbucket.org/owner/repo/src/{sha}
+        if( repository.contains('bitbucket.org') || repository.contains('bitbucket.') )
+            return "${repository}/src/${commitId}"
+        // Gitea: https://gitea.example.com/owner/repo/src/commit/{sha}
+        if( repository.contains('gitea.') )
+            return "${repository}/src/commit/${commitId}"
+        // Azure DevOps: https://dev.azure.com/org/project/_git/repo?version=GC{sha}
+        if( repository.contains('dev.azure.com') )
+            return "${repository}?version=GC${commitId}"
```
jorgee: I think it should be better to place it in the RepositoryProvider Implementations and has a static method in the factory to keep scm related code in the scm package.

bentsherman:
@ewels can we also add this to the Nextflow version that is printed so that it links to the GitHub release:

https://github.com/nextflow-io/nextflow/blob/887443e77f6eb0bc0cea9dfc606da4933a884965/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy#L440

ewels:
Yeah absolutely can 👍🏻  I wondered about doing that in the initial work but was worried I was overdoing it 😅  I'll add that in, give me a second.

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
ewels: Done in cf345d38a

modules/nextflow/src/main/groovy/nextflow/trace/ProgressRecord.groovy
```diff
@@ -32,6 +32,7 @@ class ProgressRecord implements Cloneable {
     final int index
     final String name     // process name
     String hash     // current task hash
+    String workDir  // current task work directory URI
```
pditommaso, line 35: Why this needs to be added here ?

modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
```diff
@@ -455,7 +455,8 @@ class CmdRun extends CmdBase implements HubOptions {
             fmt.a("$BACKGROUND N E X T F L O W ").reset()
 
             // Show Nextflow version
-            fmt.a(Attribute.INTENSITY_FAINT).a("  ~  ").reset().a("version " + BuildInfo.version).reset()
+            final versionUrl = "https://github.com/nextflow-io/nextflow/releases/tag/v${BuildInfo.version}"
+            fmt.a(Attribute.INTENSITY_FAINT).a("  ~  ").reset().a(hyperlink("version " + BuildInfo.version, versionUrl)).reset()
```
pditommaso: I wonder how this is going to be useful ?

bentsherman: to quickly see the changelog on github

ewels: hah, I suspected you might say that which is why I didn't add it at first until @bentsherman asked for it.

Limited usefulness - links to the changelog for that release I guess. But it's also not very intrusive, seeing as you need to cmd+click it for it to behave any differently to how it is right now.

I'm not fussed about this one either way, happy to remove it again if it's a problem.

modules/nextflow/src/main/groovy/nextflow/trace/ProgressRecord.groovy
ewels: * `hash` = just `4e/486876` (the short display string)
* `workDir` = `/path/to/work/4e/486876abcdef123456...` (the full absolute path/URI)

We can't derive the full path from the hash because we don't know:
1. Where the work directory base is (e.g., `/home/user/project/work/` or `s3://mybucket/work/`)
2. The full hash (only 8 chars are shown)

pditommaso (commented):
To tell the true not sure the added complexity is worth. I need to give a better look

ewels:
> To tell the true not sure the added complexity is worth. I need to give a better look

I've added a whole bunch of links in this PR. If you're worried about complexity, I can do what @adamrtalbot suggested and make a stripped back PR that just links the task hashes to their work directories.

I can 100000% guarantee that the task hash link feature is worth it. I've pretty much had my arm bitten off by a lot of people in the past 2 days since I made that video at the top of this PR.

bentsherman:
I think I agree, let's try just the work dir path for now and revert the other links (including the one I requested...)

ewels:
Ok I reverted most of the changes and rebased. PR code change should be as clean as possible now.

bentsherman (approved):

ewels:
@pditommaso does this look ok now? 🙏🏻

pditommaso:
I'm trying this but it's not showing anything for me

ewels:
> I'm trying this but it's not showing anything for me

Yup, that's expected (and a good thing). My guess is that you're using the MacOS terminal app @pditommaso - this doesn't support these style of hyperlinks, so it's doing the "graceful ignoring" thing I mentioned:

> Works in most terminal emulators (tested in iTerm2 and VSCode). Gracefully ignores the ANSI codes in ~~rubbish~~ _classic_ terminal emulators and treats as normal text (tested in MacOS Terminal).

If you try in iTerm2, Ghostty, VSCode or basically any other terminal emulator, it should work.

modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
```diff
@@ -442,9 +454,10 @@ class AnsiLogObserver implements TraceObserverV2 {
         // eg. 1 of 1
         final numbs = " ${(int)com} of ${(int)tot}".toString()
 
-        // Task hash, eg: [fa/71091a]
+        // Task hash - make clickable hyperlink to work dir when available (and cleanup not enabled)
+        final hashDisplay = (stats.workDir && !session.config.cleanup) ? hyperlink(hh, stats.workDir) : hh
```
pditommaso, line 459: Is this condition required? I don't see an explicit set for `session.config.cleanup`. Also when `workDir` can be null?

pditommaso:
> My guess is that you're using the MacOS terminal app

Got it

modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy
ewels: This is there because if `cleanup = true` is set in the config, then the work directories are deleted when the pipeline finishes. Clicking the hash links then tries to go to a non-existent directory, which causes an error and makes no sense. So if the config option is set, we simply skip creating the link completely.

I'm not sure about `workdir` being null, that might be Claude being overly cautious 🤔

ewels:
```suggestion
        final hashDisplay = !session.config.cleanup ? hyperlink(hh, stats.workDir) : hh
```

pditommaso: ok, let's keep, does not hurt and it's always better then a NPE

pditommaso:
Then then what's going to be the link behaviour for s3:// or gs:// and similar paths ?

ewels:
> Then then what's going to be the link behaviour for s3:// or gs:// and similar paths ?

It depends on your local system and what you have configured to handle those URI schemes. The terminal emulator delegates the "open" action to your OS.

For example, on macOS if you have [Cyberduck](https://cyberduck.io/) installed (as I do, I'm oldschool 😎), it will register as a handler for `s3://` and `gs://` URIs, so clicking the link would open that bucket path in Cyberduck. On Linux, it'll depends on the `xdg-open` configuration. 

I think I'm right in saying that if nothing is registered for the scheme, the click will simply do nothing. It'll still display correctly, it's just not an active link. It's the same as if you pasted an `s3://` URI into any other context on your system.

ewels:
Note that this delegation is actually kind of cool, it's the reason that it works so well out of the box in VSCode. The VSCode terminal intercepts the action and handles itself instead of bubbling it up to the OS, meaning that it opens the file in VSCode instead of Finder.

pditommaso:
Ok, understand. Let me give a quick try I believe we can make without the need to add a new workDir field

pditommaso:
In principle the task work dir could be inferred by the session.workDir + stats.hash, however there's the edge case when different executors uses different work dir that's not covered by this. Then let's current approach

pditommaso (approved):

PeterKnealeCMRI:
❤️

Issue #6802, opened 2026-02-04
Deadlock on 25.10.3 on AWS Batch


jchorl:
## Bug report 

Hi! I hit a deadlock using the AWS Batch executor on 25.10.3. Specifically I believe the issue is related to https://github.com/nextflow-io/nextflow/pull/6729 which was backported to 25.10.3: https://github.com/nextflow-io/nextflow/commit/f59137e6c0c221b9f050b0971298e538e040f924

That PR also puts quite a bit more load on `DescribeJobs`. Specifically, it calls `describeJob` for every task: https://github.com/nextflow-io/nextflow/blob/887443e77f6eb0bc0cea9dfc606da4933a884965/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy#L932

Unlike the task polling supervisor, it does _not_ pass in a `context` so these calls are not batched: https://github.com/nextflow-io/nextflow/blob/887443e77f6eb0bc0cea9dfc606da4933a884965/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy#L194-L203

AWS doesn't have a prescribed rate limit for `DescribeJobs` but it'd be so nice if these were batched as these calls also hit the throttle-wrapped AWS client. Anyway, that's not a chief issue here.

### Expected behavior and actual behavior

I'd expect my pipeline to run

### Steps to reproduce the problem

I have a pipeline that kicks off 176 tasks. It then pairs the results, kicking off another 88. On 25.10.3, it's deadlocking before submitting any of the 88. Unfortunately that pipeline is very complex and I don't have a minimal reproducible example but I have other resources that should help.

### Program output 

I generated a stack dump with `jstack`: 

[thread_dump.txt](https://github.com/user-attachments/files/25080844/thread_dump.txt)

Admittedly I used claude to analyze it:

```
  All 10 AWSBatch-executor threads are deadlocked waiting for each other:

  AWSBatch-executor-1 through 10:
    ThrottlingExecutor$Recoverable.call()        ← Running a task submission
      → ParallelPollingMonitor$1.invoke()
      → submit0() → submit()
      → notifyTaskSubmit()
      → getTraceRecord()                          ← Called during submission!
      → getNumSpotInterruptions()
      → describeJob()
      → ClientProxyThrottler.invokeMethod()
      → doInvoke1()
      → FutureTask.get()                          ← BLOCKED waiting for result


[...]

PR #6805, opened 2026-02-05
Added isCompleted check in getNumSpotInterruptions

munishchouhan:
This Pr fixes the deadlock reported here https://github.com/nextflow-io/nextflow/issues/6802

  Prevents the deadlock because:
  - Spot interruption counts are only meaningful after task completion
  - Returns null immediately for non-completed tasks
  - Avoids calling describeJob() and triggering the batching mechanism
  - Only queries AWS Batch API when the task is completed and we actually need the spot interruption count


jchorl:
Thanks for addressing so quickly!

PR #6806, opened 2026-02-05
[BACKPORT 25.10] Added isCompleted check in getNumSpotInterruptions

munishchouhan:
This PR will backport https://github.com/nextflow-io/nextflow/pull/6805 to  25.10

  Prevents the deadlock because:
  - Spot interruption counts are only meaningful after task completion
  - Returns null immediately for non-completed tasks
  - Avoids calling describeJob() and triggering the batching mechanism
  - Only queries AWS Batch API when the task is completed and we actually need the spot interruption count

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/scm/AssetManagerTest.groovy
@@ -207,12 +207,12 @@ class AssetManagerTest extends Specification {
         def manager = new AssetManager().build('nextflow-io/hello', [providers: [github: [auth: token]]])
 
         when:
-        manager.download("6b9515aba6c7efc6a9b3f273ce116fc0c224bf68")
+        manager.download("d828daeef77f391c456896a7a6384b5f76ceea41")
         then:
         folder.resolve('nextflow-io/hello/.git').isDirectory()
 
         when:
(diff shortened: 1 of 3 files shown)


pditommaso:
What about the version in master ?

munishchouhan:
> What about the version in master ?

there is another pr for that
https://github.com/nextflow-io/nextflow/pull/6805

thalassemia:
I want to add that this was throttling task submission rate to about 1 every few seconds using the Google Batch executor. Specifically, `notifyTaskSubmit() --> getTraceRecord() --> getNumSpotInterruptions() --> describeTask() --> getTaskStatus() --> apply()` would repeatedly fail with a `NotFoundException` because Batch had not finished creating the job. This [triggerred the retry policy](https://github.com/nextflow-io/nextflow/blob/c8ff0ce27c8e71caacf944ced58e7bfa97165585/plugins/nf-google/src/main/nextflow/cloud/google/batch/client/BatchClient.groovy#L198-L199) until the job came online or [five attempts](https://github.com/nextflow-io/nextflow/blob/c8ff0ce27c8e71caacf944ced58e7bfa97165585/plugins/nf-google/src/main/nextflow/cloud/google/batch/client/BatchRetryConfig.groovy#L36) were made. Thanks for the quick fix!

PR #6808, opened 2026-02-05
Fix listFiles() to follow symbolic links

### LINKED ISSUE #6807: listFiles() does not follow symbolic links
## Description

The `listFiles()` method on Path objects does not follow symbolic links, causing it to fail to locate files when:
1. The directory itself is a symlink
2. Files within the directory are symlinks

## Root Cause

In `modules/nf-commons/src/main/nextflow/extension/FilesEx.groovy`, the `listFiles0` method uses `Files.walkFileTree()` without specifying `FileVisitOption.FOLLOW_LINKS`:

```groovy
private static Collection<Path> listFiles0(Path self, Closure<Boolean> filter=null) {
    if( !self.isDirectory() )
        return null

    final result = []
    Files.walkFileTree(self, new SimpleFileVisitor<Path>() {
        // ...
    })
    return result
}
```

By default, `walkFileTree` does **not** follow symbolic links. This is inconsistent with `isDirectory()` which **does** follow symlinks by default.

## Workaround Limitation

A suggested workaround is to use `toRealPath()` to resolve symlinks before calling `listFiles()`:

```groovy
dir.toRealPath().listFiles()
```

However, this does **not work for cloud storage** (S3, Azure, GCS) because:
- S3: `toRealPath()` throws `UnsupportedOperationException`
- Azure: `toRealPath()` just returns `toAbsolutePath()` (no symlink resolution)
- GCS: Depends on Google's library implementation

## Proposed Fix

Add `FileVisitOption.FOLLOW_LINKS` to the `walkFileTree` call:

```groovy
import java.nio.file.FileVisitOption

private static Collection<Path> listFiles0(Path self, Closure<Boolean> filter=null) {
    if( !self.isDirectory() )
        return null

    final result = []
    final walkOptions = EnumSet.of(FileVisitOption.FOLLOW_LINKS)
    
    Files.walkFileTree(self, walkOptions, 1, new SimpleFileVisitor<Path>() {
        // ... same visitor code
    })

    return result
}
```

This should be safe for cloud filesystems since:
1. Cloud storage doesn't have symlinks, so FOLLOW_LINKS is effectively a no-op
2. The S3 client already uses `EnumSet.of(FileVisitOption.FOLLOW_LINKS)` internally for downloads (`S3Client.java:405`)

## Environment

- Nextflow version: master branch
- Affects both local and cloud file systems (local due to symlink issue, cloud due to workaround not being viable)

adamrtalbot:
## Summary

- Fix `listFiles()` to properly follow symbolic links by adding `FileVisitOption.FOLLOW_LINKS` to the `walkFileTree` call
- Add test case for listing files in a symlinked directory

## Problem

The `listFiles()` method was not following symbolic links because `Files.walkFileTree()` defaults to not following them. This caused `listFiles()` to fail to locate files when:
1. The directory itself is a symlink
2. Files within the directory are symlinks

A workaround using `toRealPath()` was suggested, but this doesn't work for cloud storage (S3 throws `UnsupportedOperationException`, Azure just returns `toAbsolutePath()`).

## Solution

Added `FileVisitOption.FOLLOW_LINKS` to the `walkFileTree` call in `listFiles0()`. This is safe for cloud filesystems since:
1. Cloud storage doesn't have symlinks, so `FOLLOW_LINKS` is effectively a no-op
2. The S3 client already uses `EnumSet.of(FileVisitOption.FOLLOW_LINKS)` internally for downloads

Fixes #6807

## Test plan

- [x] Added test case `should list files in symlinked directory`
- [x] All existing `FilesExTest` tests pass

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/script/parser/v2/ScriptLoaderV2Test.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/script/parser/v2/ScriptLoaderV2Test.groovy
@@ -295,6 +295,30 @@ class ScriptLoaderV2Test extends Dsl2Spec {
         outputs.getFiles().size() == 3
     }
 
+    def 'should allow optional param' () {
+
+        given:
+        def session = new Session()
+        def parser = new ScriptLoaderV2(session)
+
+        def TEXT = '''
+            params {
+                path: Path?
+            }
+
+            workflow {
+                params.path
+            }
+            '''
+
+        when:
+        parser.parse(TEXT)
+        parser.runScript()
+
+        then:
+        parser.getResult() == null
+    }
+
     def 'should support enums' () {
 
         given:
--- a/modules/nf-lang/src/main/java/nextflow/script/control/ScriptToGroovyVisitor.java
+++ b/modules/nf-lang/src/main/java/nextflow/script/control/ScriptToGroovyVisitor.java
@@ -150,7 +150,7 @@ public class ScriptToGroovyVisitor extends ScriptVisitorSupport {
             .map((param) -> {
                 var name = constX(param.getName());
                 var type = classX(param.getType());
-                var optional = constX(type.getNodeMetaData(ASTNodeMarker.NULLABLE) != null);
+                var optional = constX(param.getType().getNodeMetaData(ASTNodeMarker.NULLABLE) != null);
                 var arguments = param.hasInitialExpression()
                     ? args(name, type, optional, param.getInitialExpression())
                     : args(name, type, optional);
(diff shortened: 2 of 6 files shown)


pditommaso:
Folks, i've reverted this PR. The change looks harmless at first glance but has a wider blast radius than expected.

## Affected production callers

- `FilesEx.createLinkImpl` — enumerates a directory to create hard links recursively.
- `LocalUpdateRepository` — lists plugin directories.
- `Session.findBinEntries` — discovers executable scripts in the pipeline `bin/` directory.

## Task hash impact

`findBinEntries` populates the bin entries that are factored into task hashing. If FOLLOW_LINKS changes which files are returned (or how symlinked entries are classified), this silently invalidates cached task results — users would see unexpected cache misses with no explanation.



This is a semantic difference that could introduce bugs both in internal code and in user pipelines calling `.listFiles()` as a public API. Tho I understand that's a desirable patch, at this point  it would be better to provide an option to control it.

bentsherman:
That's fine, we can just apply this change to `listDirectory()` since it is new in 26.04

@adamrtalbot please re-submit this PR, making `listFiles0` parameterized so that `listFiles` calls it without follow_links and `listDirectory` calls it with follow_links.

PR #6731, opened 2026-01-19
Fix: CPU usage calculation with `starttime`, relying only on `/proc/pid>/stat`

JosuaCarl:
As described in #6710, `%cpu` seems to exceed the value of 100% per core. The following update addresses this by removing `/proc/stat` as a source for the globally passed ticks, relying instead on `starttime` in `/proc/<pid>/stat` as a global clock.

- Closes #6710

PR #6809, opened 2026-02-05
Fix: Reporting the correct number of utilized CPUs

### LINKED ISSUE #6743: `cpus` trace value does not correspond to used logical cores of the task
## Bug report 

- Relates to #6710

The execution trace should report the logical CPUs the process ran on, but does only the number it should have run on from the config, which is often wrong.

1. Define a config with `process.resourceLimits = [cpus: 1]`
2. Execute `nextflow run demo -r 1.0.2 -profile test,docker -c test.config --outdir out`
3. Most of the time some task has a `%cpu` > 100 which should not be, because `cpus` is at `1`

### Program output
```tsv
	task_id	native_id	name	status	%cpu	cpus	memory
	3	3888223	NFCORE_DEMO:DEMO:SEQTK_TRIM (SAMPLE2_PE)	COMPLETED	91.1%	1	4 GB
	2	3888220	NFCORE_DEMO:DEMO:SEQTK_TRIM (SAMPLE1_PE)	COMPLETED	86.0%	1	4 GB
	4	3889134	NFCORE_DEMO:DEMO:FASTQC (SAMPLE2_PE)	COMPLETED	76.3%	1	4 GB
	1	3888837	NFCORE_DEMO:DEMO:FASTQC (SAMPLE1_PE)	COMPLETED	82.2%	1	4 GB
	6	3890070	NFCORE_DEMO:DEMO:SEQTK_TRIM (SAMPLE3_SE)	COMPLETED	85.4%	1	4 GB
>	5	3890405	NFCORE_DEMO:DEMO:FASTQC (SAMPLE3_SE)	COMPLETED	117.0%	1	4 GB
	7	3891275	NFCORE_DEMO:DEMO:MULTIQC	COMPLETED	44.3%	1	4 GB
```

### Environment 

* Nextflow version: [25.10.2]
* Operating system:  Linux & MacOS

### Additional context

When limiting the cpus via `taskset -c <CPU_LIST> nextflow run ...`, everything seems to be in order:
```tsv
task_id	native_id	name	status	%cpu	cpus	memory
2	3329972	NFCORE_DEMO:DEMO:SEQTK_TRIM (SAMPLE2_PE)	COMPLETED	81.7%	1	4 GB
1	3330576	NFCORE_DEMO:DEMO:SEQTK_TRIM (SAMPLE1_PE)	COMPLETED	92.4%	1	4 GB
3	3331266	NFCORE_DEMO:DEMO:SEQTK_TRIM (SAMPLE3_SE)	COMPLETED	90.0%	1	4 GB
4	3331898	NFCORE_DEMO:DEMO:FASTQC (SAMPLE1_PE)	COMPLETED	71.5%	1	4 GB
5	3332979	NFCORE_DEMO:DEMO:FASTQC (SAMPLE2_PE)	COMPLETED	73.9%	1	4 GB
6	3333940	NFCORE_DEMO:DEMO:FASTQC (SAMPLE3_SE)	COMPLETED	77.1%	1	4 GB
7	3334923	NFCORE_DEMO:DEMO:MULTIQC	COMPLETED	38.3%	1	4 GB
```

JosuaCarl:
As stated in #6743, a problem with flexible scheduling of tasks is that the rules for execution constraints upon logical cores is not enforced, or enforced with wiggle room.

In the instance of a Docker executor, `--cpu-shared 2048` is used instead of `--cpus 2` which prioritizes resource utilization, but is no hard cutoff (see https://docs.docker.com/engine/containers/resource_constraints/#configure-the-default-cfs-scheduler).
https://github.com/nextflow-io/nextflow/blob/887443e77f6eb0bc0cea9dfc606da4933a884965/modules/nextflow/src/main/groovy/nextflow/container/DockerBuilder.groovy#L128-L129

This behavior has many upsides, but the number of `cpus` in the trace is often reported incorrectly, as it takes the value set by the user in `process.cpus` for granted, even if it was not enforced.

The PR adds a sampling script which utilizes breadth-first search to extract the tasks and its children's last used logical CPU and record them in a list. In the end, this is summed to the total value of utilized CPUs.

The current sampling rate is at `0.1s`. I tried `1.0s`, but this missed several short processes in the `demo` pipeline, although I suspect this effect may vanish for longer processes. I could see no large changes in execution time and memory utilization for the tasks for either option. Sampling was chosen, because other ways either require root permission (tracking at kernel level with `perf`) or only reported minima (`ceil(%cpu / 100)`) and maxima (`grep Cpus_allowed_list /proc/$pid/status`) for the number of utilized CPUs. In practice I observed, that the number of actually used CPUs was most of the time close to the maximum, so if an approximation was to be used that does not rely on sampling, I would suggest `grep Cpus_allowed_list /proc/$pid/status`, which was my original solution in the first commit (have a look if you want).

- Closes #6743

Edits:
- Changed the parameter that is reported from `cpus` to a new `TraceRecord` field `used_cpus`.
- Replaced trace writing with already existing method `nxf_write_trace` to align `nxf_trace_linux` with `nxf_trace_mac` and save a few lines. This could also be put into a new PR, if deemed too much deviation from the main task.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
@@ -105,7 +105,8 @@ class TraceRecord implements Serializable {
             hostname: 'str',
             cpu_model:  'str',
             accelerator: 'num',
-            accelerator_type: 'str'
+            accelerator_type: 'str',
+            used_cpus: 'num'
     ]
 
     static public Map<String,Closure<String>> FORMATTER = [
--- a/modules/nextflow/src/main/resources/nextflow/executor/command-trace.txt
+++ b/modules/nextflow/src/main/resources/nextflow/executor/command-trace.txt
@@ -179,6 +179,7 @@ nxf_write_trace() {
         "nextflow.trace/v2" \
         "realtime=$wall_time" \
         "%cpu=$ucpu" \
+        "used_cpus=$used_cpus" \
         "cpu_model=$cpu_model" \
         "rchar=${io_stat1[0]}" \
         "wchar=${io_stat1[1]}" \
@@ -199,6 +200,7 @@ nxf_trace_mac() {
     local ucpu=''
     local cpu_model=''
     local io_stat1=('' '' '' '' '' '')
+    local used_cpus=''
     nxf_write_trace
 }
 
@@ -235,7 +237,7 @@ nxf_trace_linux() {
     local mem_proc=$!
 
     ## Report number of used CPUs via polling
-    local cpus=$(nxf_cpu_watch "$task")
+    local used_cpus=$(nxf_cpu_watch "$task")
 
     wait $task
 
@@ -254,18 +256,7 @@ nxf_trace_linux() {
     local wall_time=$((end_millis-start_millis))
     [ $NXF_DEBUG = 1 ] && echo "+++ STATS %CPU=$ucpu TIME=$wall_time I/O=${io_stat1[*]}"
 
-    printf "%s\n" \
-        "nextflow.trace/v2" \
-        "realtime=$wall_time" \
-        "%cpu=$ucpu" \
-        "cpus=$cpus" \
-        "cpu_model=$cpu_model" \
-        "rchar=${io_stat1[0]}" \
-        "wchar=${io_stat1[1]}" \
-        "syscr=${io_stat1[2]}" \
-        "syscw=${io_stat1[3]}" \
-        "read_bytes=${io_stat1[4]}" \
-        "write_bytes=${io_stat1[5]}" >| "$trace_file" || >&2 echo "Error: Failed to write to file: $trace_file"
+    nxf_write_trace
 
     ## join nxf_mem_watch
     [ -e /proc/$mem_proc ] && eval "echo 'DONE' >&$mem_fd" || true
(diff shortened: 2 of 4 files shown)


muffato:
(I'm not in the Nextflow team, I'm just a Nextflow user.)

I would like to plead to leave the current `cpus`  as it is, because together with `memory` and `time` (and `disk`) they report what the workflow had defined and _requested_ for each task. The values can then be compared with the actual usage `%cpu`, `peak_rss`, and `realtime` to then optimise resource requests within the pipeline. Making resource usage match resource requests is a very important step when deploying a pipeline in production, to reduce scheduling time and wastage.

I see the metrics you're trying to add as an "advanced" trace that gives valuable insights, alongside the existing scheduler and runtime traces, like `duration`, `%mem`, `vmem`, `write_bytes `, and [_many_ others](https://www.nextflow.io/docs/latest/reference/config.html#trace), but doesn't _replace_ `cpus`.

JosuaCarl:
@muffato So would you rather opt for an additional trace value, like `used_cpus`, so the reporting of user set `cpus` is not lost?

muffato:
Personally, yes, but again I'm just a mere (moderately advanced) user. The call should come from Seqera.

bentsherman:
Closing for the same reason as #6731 -- this seems like a lot of added complexity for not much added value, consider doing it through a plugin instead.

PR #6812, opened 2026-02-05
Add `workflow.report` config scopes

bentsherman:
This PR adds the `workflow.report` config scope as a replacement for the existing report scopes:

- `dag` -> `workflow.report.dag`
- `report` -> `workflow.report.execution`
- `timeline` -> `workflow.report.timeline`
- `trace` -> `workflow.report.trace`

The config settings are otherwise identical, except that the `file` option is resolved against the workflow output directory instead of the launch directory.

This makes the standard reports easier to use with workflow outputs.

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -1748,7 +1748,7 @@ The `run` command is used to execute a local pipeline script or remote pipeline 
   $ nextflow run main.nf -profile docker
   ```
 
-- Execute a pipeline and generate the summary HTML report. For more information on the metrics, please refer the {ref}`tracing-page` section:
+- Execute a pipeline and generate the summary HTML report. For more information on the metrics, please refer the {ref}`reports-page` section:
 
   ```console
   $ nextflow run main.nf -with-report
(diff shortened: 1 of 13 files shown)


pditommaso:
Could this impact any Platform capabilities? just wondering and better e2e testing

bentsherman:
It shouldn't break platform runs that use the existing report options, since this PR falls back to them if `workflow.report` is not used

For platform runs that use the new `workflow.report` options, you will need to make sure to set the output directory. Right now you can only do that in config, but I think they are working to add an "Output directory" field

bentsherman:
On second thought, I'm not sure if we should continue to build on these reports. I think it would be better to just expose the raw data (trace events, lineage) and let external systems (agents, seqera platform) render this information to the user however they want

Maybe we should just deprecate these built-in reports and move them into a plugin

bentsherman:
Don't think this is worth doing. The original motivation was to make the report file paths relative to `outputDir` instead of `launchDir`. I think we can find a simpler way to do that, e.g. change the behavior in a future version and provide a fallback env var (`NXF_LEGACY_REPORT_FILE_ROOT`)

PR #6822, opened 2026-02-10
Improve documentation of workflow outputs

bentsherman:
This PR recovers some extra changes from #6701 . That PR was reverted due to ongoing compatibility issues with adding new trace observer events. However, the PR also added some general improvements to comments, log messages, and variable names, which I would like to keep.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/trace/TraceObserverV2.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/trace/TraceObserverV2.groovy
@@ -121,15 +121,19 @@ interface TraceObserverV2 {
     default void onFlowError(TaskEvent event) {}
 
     /**
-     * Invoked when a workflow output is published.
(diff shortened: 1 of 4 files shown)


modules/nextflow/src/main/groovy/nextflow/trace/TraceObserverV2.groovy
```diff
@@ -121,15 +121,19 @@ interface TraceObserverV2 {
     default void onFlowError(TaskEvent event) {}
 
     /**
-     * Invoked when a workflow output is published.
+     * Invoked when a workflow output is completed.
+     *
+     * For a given workflow output, this event is guaranteed
+     * to be emitted after all values for that output have been
+     * published via onWorkflowPublish().
```
bentsherman:
```suggestion
     * For a given workflow output, this event is guaranteed
     * to be emitted after all files for that output have been
     * published via onFilePublish().
```

PR #6824, opened 2026-02-11
Don't include container in task trace when containers are not enabled

bentsherman:
This PR changes the task trace to only include the task container if containers were enabled for the run.

This way, when a run uses e.g. conda environments instead of containers, the task detail in Platform won't show a container.


pditommaso:
Worth triggering e2e tests

PR #6827, opened 2026-02-12
Abort execution when platform telemetry error

### LINKED ISSUE #6758: Pipeline completed with errors ends with correct exit-code
> **Don't know if this post is valid format, please let me know if i have to re-edit**

## New feature

I have noticed that nextflow pipeline exit-code does not match with the actual exit situation, is this a design on purpose? I personally find this rather inconvenient, if i want to have a quick look of the exit state of pipeline , i will have to look at the log thoroughly. 
I would really like that the pipeline give an exit-code that match with the exit state, for example:
    - "Workflow failed on launch: Launcher error" -> 1
    -  "Workflow failed on launch: Operation aborted" -> 1
    -  "Pipeline completed with errors" -> 1 
    - etc.

## Use case 

My use case is i have a shell script for retry a failed workflow for at most 3 times, so i have to judge the exit state, if i can't have the correct exit-code recorded, i will have to scramble the log file for key word or key phrases. which i find most unreliable - cause i will have to get to know all the key word

### LINKED ISSUE #7055: Invalid Platform endpoint with -with-tower is silently ignored

When running Nextflow with `-with-tower` pointing to an invalid Seqera Platform endpoint, the error from the Platform API is silently swallowed and the workflow execution completes successfully.

The expected behavior is that Nextflow should stop execution (fail fast) when it cannot successfully communicate with the configured Platform endpoint, so the user is made aware that monitoring/reporting is not working.

Run a workflow against an invalid (or non-Platform) endpoint:

```
nextflow run main.nf -with-tower 'https://cloud.stage-seqera.io'
```

### Program output

The following is logged at DEBUG level, but the execution proceeds and completes successfully:

```
Apr-20 21:34:26.286 [main] DEBUG io.seqera.tower.plugin.TowerClient - Creating Seqera Platform observer -- endpoint=https://cloud.stage-seqera.io; requestInterval=1s; aliveInterval=1m; maxRetries=5; backOffBase=0; backOffDelay=0
Apr-20 21:34:26.571 [main] DEBUG io.seqera.tower.plugin.TowerClient - Unable to parse error cause as JSON object: <html>
<head><title>405 Not Allowed</title></head>
<body>
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx</center>
</body>
</html>

Apr-20 21:34:26.573 [main] DEBUG io.seqera.tower.plugin.TowerClient - Unexpected HTTP response
- endpoint    : https://cloud.stage-seqera.io/trace/create
- status code : 405
- response msg: null
```

### Environment

- Nextflow version: 26.03.2-edge (build 11799)
- Java version: OpenJDK 21.0.3+9-LTS
- Operating system: macOS 26.4.1
- Bash version: n/a

jorgee:
This pull request introduces a new mechanism to control error handling behavior in the `TowerClient` class by adding an `abortOnError` flag, which can be set via the environment variable `TOWER_ABORT_ON_ERROR`. When enabled, critical errors encountered while communicating with Seqera Platform will cause the workflow to abort immediately using the `AbortRunException`. The changes also include improved error propagation and additional tests to verify this behavior.

**Error Handling Improvements:**

* Added `abortOnError` flag to `TowerClient`, defaulting to `true`, and made it configurable via the `TOWER_ABORT_ON_ERROR` environment variable. This determines whether critical errors abort the workflow or are handled as warnings. [[1]](diffhunk://#diff-4e7968503d8e56d6d5f2eff04e734382b6b79fee49dc22668d9a3803c91113e2R146-R147) [[2]](diffhunk://#diff-4e7968503d8e56d6d5f2eff04e734382b6b79fee49dc22668d9a3803c91113e2R157-R175)
* Updated error handling in `TowerClient` methods (`logHttpResponse`, `parseTowerResponse`, and others) to throw `AbortRunException` when `abortOnError` is enabled, ensuring immediate workflow termination on critical errors. [[1]](diffhunk://#diff-4e7968503d8e56d6d5f2eff04e734382b6b79fee49dc22668d9a3803c91113e2R304-R316) [[2]](diffhunk://#diff-4e7968503d8e56d6d5f2eff04e734382b6b79fee49dc22668d9a3803c91113e2R403-R408) [[3]](diffhunk://#diff-4e7968503d8e56d6d5f2eff04e734382b6b79fee49dc22668d9a3803c91113e2R756-R758) [[4]](diffhunk://#diff-4e7968503d8e56d6d5f2eff04e734382b6b79fee49dc22668d9a3803c91113e2R778-R783) [[5]](diffhunk://#diff-4e7968503d8e56d6d5f2eff04e734382b6b79fee49dc22668d9a3803c91113e2R866-R874)

**Session and Exception Propagation:**

* Modified the `Session` class to specifically catch and log `AbortRunException` during observer notification, ensuring these exceptions propagate and abort the workflow as intended. [[1]](diffhunk://#diff-bee88a999f8eecd3cec5fcad24b76736162be86626f89d746932f64928f24374R50) [[2]](diffhunk://#diff-bee88a999f8eecd3cec5fcad24b76736162be86626f89d746932f64928f24374R1113-R1117)

**Tests:**

* Added new tests in `TowerClientTest` to verify the correct detection of the `abortOnError` setting and to ensure that the workflow aborts as expected when errors occur and `abortOnError` is enabled. [[1]](diffhunk://#diff-463f0ed6e146304ec6d2d8252af59dc42ba32ef8851ae68c371708e3823157c3R33) [[2]](diffhunk://#diff-463f0ed6e146304ec6d2d8252af59dc42ba32ef8851ae68c371708e3823157c3R607-R653)


plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
```diff
@@ -286,12 +301,19 @@ class TowerClient implements TraceObserverV2 {
...
+            }
             throw new AbortOperationException(resp.message)
         }
         final ret = parseTowerResponse(resp)
         this.workflowId = ret.workflowId
-        if( !workflowId )
+        if( !workflowId ) {
+            if( abortOnError ) {
+                throw new AbortRunException("Invalid Seqera Platform API response - Missing workflow Id")
+            }
```
pditommaso: Why adding this condition if an exception was already thrown?

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
```diff
@@ -286,12 +301,19 @@ class TowerClient implements TraceObserverV2 {
                 - status code : $resp.code
                 - response msg: $resp.cause
                 """.stripIndent(true)
+            if( abortOnError ) {
+                throw new AbortRunException(resp.message)
+            }
```
pditommaso: Not getting what this is adding?

jorgee: `notifyEvent` was catching all exceptions and printing a log message, but not re-throwing the exception. So, no error was produced with normal exceptions. I changed `notifyEvent ` to catch and rethrow the `AbortRunException`. I didn't want to rethrow all exceptions or use the `AbortOperationException` because it is widely used and it could be very likely to introduce side effects. The AbortRunException was only used at `ScriptRunner` and `Launcher`. It was safe to use and matches with the meaning of the exception 

I initially implemented it by invoking `session.abort` from here, but it was not working because the execution has not started yet. Abort was executed and threw the exception, but as it was caught at `notifyEvent`, the execution continued and produced failures later when trying to run the script in the aborted session. I think it could be fixed by inspecting the session just after invoking `start`, and throwing an exception if it is aborted. 

Both are easy to implement; I just selected this one because it works in the same way for all the events, and it could be reused for other observers. If you think something is wrong or it's better to abort directly, I can try with the other approach.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
jorgee: The same as the previous comment, `notifyEvent` was catching the exceptions but not re-throwing to produce an error in the execution.

jorgee:
I have updated this branch to master. 
If I am not wrong, in one of our conversations, we agreed to just fail for platform cases instead of all exceptions. This PR implements this behaviour and allows to work also as before using an environment variable. I will close #6823 in favor of this one.

bentsherman:
@pditommaso let me know if the latest changes look good to you

I like the general principle of "plugin observer errors are logged as warnings by default, observer can throw AbortRunException to fail the run"

This PR currently just adds an env var to control whether the TowerClient throws hard/soft errors. But I wonder if we should just decide for each error case whether it should be hard or soft instead of introducing an environment var

For example, if I run with `-with-tower` but fail to authenticate, maybe that should just always be a hard error

In any case, let's wait until #6946 is merged since it refactors the tower client (harder to resolve merge conflicts there)

bentsherman:
From today's discussion -- @jorgee please update this PR making all failures related to sending data to Platform hard failures, removing the need for the environment variable.

jorgee:
I have removed the environment variable and throw `AbortRunException` for all calls except heartbeats and `onFlowComplete`. 

- For `onFlowComplete`, it is the last message and execution will be finished anyway. 
- I have some doubts about the heartbeats case. Failing at the first exception could be very strict. Therefore, I have considered that a task complete message will be sent at somepoint and it will send the `AbortRunException` if there is a consistent connection problem. Is it fine? should I throw an `AbortRunException` at first error?

bentsherman:
For the heartbeats, I think keeping them as warnings is reasonable

For the `onFlowComplete`, if it fails, wouldn't that leave the platform run in an indeterminate state? Or does platform still mark the run as completed?

jorgee:
> For the `onFlowComplete`, if it fails, wouldn't that leave the platform run in an indeterminate state? Or does platform still mark the run as completed?

I' ll try to check what it does in this case. Anyway, it will not receive the trace event. The only possibility is if it monitors the head job and use the process exit code to decide if failed or not. In that case, I think it is better to warn instead of throwing the AbortRunException

jorgee:
Confirmed! I have run an execution skipping the trace completed event. The workflow state at platform is not undetermined. As head-job finishes, the workflow execution is considered completed, but marked as FAILED. 

Last updates:
- I have updated the branch with master 
- Fixed an issue introduced in TowerClient refactor where begin trace was failing but not detected by tests because the exception was not aborting the execution (commit 1d93537)

bentsherman (approved):

PR #6831, opened 2026-02-13
Fix nf-azure plugin tier-1 bugs

adamrtalbot:
## Summary

Fixes 5 correctness bugs in the nf-azure plugin identified during a technical debt audit. All existing tests pass.

### 🔴 Critical

1. **`AzFileSystem.listContainers0()` returned empty list** — The `forEach` iterated over blob containers but discarded the result of `provider.getPath()` instead of adding it to the `containers` list. Any code path relying on container listing at the filesystem level silently received nothing.

2. **NPE in `AzBatchTaskHandler` failure handling** — `info.failureInfo.message` was accessed without null-checking `failureInfo`. When Azure returns a failed task without failure detail, this throws an NPE. Fixed with safe navigation (`?.`).

3. **SAS token permissions silently wrong in `AzHelper`** — `.setPermissions(BLOB_PERMS).setPermissions(CONTAINER_PERMS)` caused the second call to overwrite the first, making `BLOB_PERMS` dead code. Since this generates a container-level user delegation SAS, only `CONTAINER_PERMS` is needed. Removed the dead call.

### 🟡 Moderate

4. **ADLS Gen2 `hdi_isfolder` truthy string bug in `AzFileAttributes`** — `directory = meta.get("hdi_isfolder")` assigned a String to a boolean field. In Groovy, the string `"false"` is truthy (non-null, non-empty), so ADLS Gen2 paths with `hdi_isfolder=false` metadata were incorrectly treated as directories. Fixed to use `"true".equalsIgnoreCase(...)`.

5. **`AzManagedIdentityOpts` ignored documented env var fallbacks** — The `@Description` annotations document `AZURE_MANAGED_IDENTITY_USER` and `AZURE_MANAGED_IDENTITY_SYSTEM` env var fallbacks, but the constructor never read them. Wired up the env var fallbacks to match documented behavior.

## Files Changed

- `plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzFileSystem.groovy`
- `plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchTaskHandler.groovy`
- `plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzHelper.groovy`
- `plugins/nf-azure/src/main/nextflow/cloud/azure/nio/AzFileAttributes.groovy`
- `plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzManagedIdentityOpts.groovy`

## Testing

- `./gradlew :plugins:nf-azure:test` — all tests pass
- `NXF_SMOKE=1 ./gradlew :plugins:nf-azure:test` — all smoke tests pass
- Minimal, targeted fixes — no refactoring or behavioral changes beyond the bug fixes


bentsherman:
Nice cleanup. `listContainers0()`, the `failureInfo?` guard, and the dead `setPermissions(BLOB_PERMS)` all look right to me — `CONTAINER_PERMS` and `BLOB_PERMS` are defined with identical permission sets, so that removal is a strict no-op (`BLOB_PERMS` is now unused and could go too).

Two things before merge:

1. **The managed identity env fallback can flip auth precedence.** `AzBatchService:379` checks `managedIdentity().isConfigured()` before `activeDirectory()`, so on a host where `AZURE_MANAGED_IDENTITY_SYSTEM=true` is exported, a run configured with a service principal silently switches to managed identity — and config `clientId` plus that env var makes `isConfigured()` throw at startup where it previously worked. Documented behaviour, and `AzActiveDirectoryOpts` has the same exposure, but it's the one change here that can break a working pipeline. Worth a changelog note plus tests (`SysEnv.push`/`pop`) for the config-set-and-env-set precedence cases.

2. The new `Map<String,String> env = null` ctor param is never read — the body calls `SysEnv` directly. Copied from `AzActiveDirectoryOpts`, but `AzStorageOpts:64` has the better form (`Map<String,String> env = SysEnv.get()`, read from `env`), which also makes the fallback testable without `SysEnv.push`.

Minor: `info?.failureInfo?.message` would match the `info?.exitCode` guard three lines up. And the description lists 5 fixes but the diff has 4 — `hdi_isfolder` is already fixed on master with a stricter check.

bentsherman (approved):

PR #6838, opened 2026-02-17
Add OSC 9;4 terminal progress bar support

edmundmiller:
## Summary

Emit [OSC 9;4 (ConEmu-style)](https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC) escape sequences during ANSI log rendering so terminals with support show a native GUI progress bar.

**Supported terminals:** Ghostty, Windows Terminal, ConEmu, WezTerm, and others.

## What it does

- Aggregates `completedCount / totalCount` across all processes to compute overall workflow percentage
- Emits the appropriate OSC 9;4 state on each render cycle:
  - `3` (indeterminate) — processes created but no tasks submitted yet
  - `1` (normal) + percentage — during execution
  - `2` (error) + percentage — when any process has `errored=true`
  - `0` (hidden) — on completion and force termination
- Detects tmux and wraps sequences in DCS passthrough (`allow-passthrough` must be enabled)

## Configuration

Enabled by default. Terminals that don't support OSC 9;4 silently ignore the sequence.

Set `NXF_OSC_PROGRESS=false` to disable (e.g. for iTerm2 where OSC 9;4 may collide with notifications).

## Tests

- Spec tests for sequence generation (all 5 states)
- tmux DCS passthrough wrapping
- tmux detection (TERM_PROGRAM and TMUX env vars)
- Config toggle (enable/disable via NXF_OSC_PROGRESS)
- Progress computation: normal, error, indeterminate, empty, 100%, dependent processes with partial downstream counts
- Manual test pipeline: `tests/osc-progress.nf`

### DIFF
--- a/docs/reference/env-vars.md
+++ b/docs/reference/env-vars.md
@@ -144,6 +144,9 @@ The following environment variables control the configuration of the Nextflow ru
 `NXF_OPTS`
 : Provides extra options for the Java and Nextflow runtime. It must be a blank separated list of `-Dkey[=value]` properties.
 
+`NXF_OSC_PROGRESS`
+: Enables/disables the OSC 9;4 terminal progress bar, which shows workflow progress in the terminal tab/title bar. Supported by Ghostty, Windows Terminal, ConEmu, and other terminals (default: `true`). Set to `false` to disable.
+
 `NXF_ORG`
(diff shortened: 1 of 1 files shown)


bentsherman:
I think this would make more sense in a plugin

PR #6840, opened 2026-02-18
Fix ResourcesAggregator deadlock with virtual thread executor

pditommaso:
## Summary

Fix a deadlock in `ResourcesAggregator.computeSummaryMap()` that causes Nextflow to hang indefinitely during shutdown when generating the execution report.

## Problem

`ResourcesAggregator` used the shared session `ExecutorService` — a `ThreadPerTaskExecutor` backed by virtual threads — to parallelize summary stats computation via `invokeAll()`. 

When other virtual threads (e.g. `PublishDir` file transfers) saturate or pin all ForkJoinPool carrier threads, the new tasks submitted by `invokeAll` can never be scheduled. Since `invokeAll` blocks until **all** futures complete, this creates a **starvation deadlock**: the main thread waits forever at `computeSummaryMap()` during `onFlowComplete`.

**Observed in production** — thread dump shows the main thread stuck at:
```
at java.util.concurrent.ThreadPerTaskExecutor.invokeAll(ThreadPerTaskExecutor.java:365)
at nextflow.trace.ResourcesAggregator.computeSummaryMap(ResourcesAggregator.groovy:75)
at nextflow.trace.ResourcesAggregator.computeSummaryList(ResourcesAggregator.groovy:90)
at nextflow.trace.ResourcesAggregator.renderSummaryJson(ResourcesAggregator.groovy:105)
at nextflow.trace.ReportObserver.renderSummaryJson(ReportObserver.groovy:201)
...
at nextflow.Session.shutdown0(Session.groovy:777)
```

This is related to but independent of #6833 — that PR fixes one cause of carrier thread saturation (S3 delete), but `ResourcesAggregator` is vulnerable to **any** scenario where carrier threads are busy.

## Fix

Replace the shared session virtual thread executor with a dedicated `FixedThreadPool` scoped to the `computeSummaryMap()` call. This is CPU-bound computation that:
- Benefits from platform threads (no I/O waiting)
- Must not compete with virtual threads for carrier thread scheduling
- Only runs once during shutdown, so the pool creation overhead is negligible

This also removes the `Session` dependency from `ResourcesAggregator`, simplifying both the class and its tests.

## Test plan

- [x] Existing `ResourcesAggregatorTest` tests pass (summary computation, JSON rendering, insertion order)
- [x] `ReportObserverTest` passes
- [x] `nf-tower` plugin compiles cleanly

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/trace/ReportObserver.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/trace/ReportObserver.groovy
@@ -113,7 +113,7 @@ class ReportObserver implements TraceObserverV2 {
     @Override
     void onFlowCreate(Session session) {
         this.session = session
-        this.aggregator = new ResourcesAggregator(session)
+        this.aggregator = new ResourcesAggregator()
         // check if the process exists
         if( Files.exists(reportFile) && !overwrite )
             throw new AbortOperationException("Report file already exists: ${reportFile.toUriString()} -- enable the 'report.overwrite' option in your config file to overwrite existing files")
--- a/modules/nextflow/src/main/groovy/nextflow/trace/ResourcesAggregator.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/trace/ResourcesAggregator.groovy
@@ -1,17 +1,16 @@
 package nextflow.trace
 
 import java.util.concurrent.Callable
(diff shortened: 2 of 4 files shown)


pditommaso:
I think it get confused with the similar one #6833, but yes, it's correct

PR #6841, opened 2026-02-18
Add support for latest-generation Google Cloud machine families

sosoihd:
#### 1. Missing support for latest-generation machine families

Google Cloud has introduced several new general-purpose machine families that are not currently supported by Nextflow:
- **C4** series: 4th gen Intel Xeon Scalable (Sapphire Rapids) - [C4 machine series](https://cloud.google.com/compute/docs/general-purpose-machines#c4_series)
- **C4A** series: AMD EPYC Genoa processors - [C4A machine series](https://cloud.google.com/compute/docs/general-purpose-machines#c4a_series)  
- **C4D** series: AMD EPYC Genoa with local SSD - [C4D machine series](https://cloud.google.com/compute/docs/general-purpose-machines#c4d_series)
- **N4** series: 4th gen Intel Xeon Scalable (Sapphire Rapids) - [N4 machine series](https://cloud.google.com/compute/docs/general-purpose-machines#n4_series)
- **N4D** series: AMD EPYC Genoa processors - [N4D machine series](https://cloud.google.com/compute/docs/general-purpose-machines#n4d_series)
- **N4A** series: AMD EPYC Genoa processors - [N4A machine series](https://cloud.google.com/compute/docs/general-purpose-machines#n4a_series)

These families offer significant improvements in:
- Price-performance ratio (up to 10% better than previous generations)
- Memory bandwidth
- Network throughput
- Energy efficiency

**Without this support, users cannot leverage:**
- Latest Intel Sapphire Rapids processors (C4, N4)
- Latest AMD EPYC Genoa processors (C4A, C4D, N4A, N4D)
- Improved performance characteristics of these newer families

#### 2. Inability to specify boot disk type

Currently, Nextflow only allows configuring the boot disk **size** via `google.batch.bootDiskSize`, but not the disk **type**. This creates several issues:

**Compatibility problems:**
- The new C4, C4A, C4D, N4, N4A, and N4D families [do not support `pd-balanced` disks](https://cloud.google.com/compute/docs/disks#disk_types_by_machine_type) (the Google Cloud default)
- These families require alternative disk types like `hyperdisk-balanced` or `pd-ssd`
- **This makes it impossible to use these new machine families at all**

**Performance optimization:**
- High-I/O workloads may benefit from `pd-ssd` (higher IOPS)
- Cost-sensitive workflows may prefer `pd-standard` (lower cost)
- Users cannot optimize disk performance for their specific workloads

**Reference:** [Google Cloud Disk Types Documentation](https://cloud.google.com/compute/docs/disks#disk-types)

### Solution

This PR addresses both issues with a comprehensive solution:

#### 1. Add support for latest-generation machine families

**Machine type recognition:**
- Added C4, C4A, C4D, N4, N4A, N4D families
- Updated local SSD handling for C4/C4A/C4D families with `-lssd` suffix

**Testing:**
- Added comprehensive tests for all new machine families
- Verified local SSD behavior (supported/not supported) per family

### Changes

| File | Change |                                                                                                                                                             
  |------|--------|                      
  | `GoogleBatchMachineTypeSelector.groovy` | Add `HYPERDISK_ONLY_FAMILIES` list, `isHyperdiskOnly()` method, C4/C4A/C4D `-lssd` handling |                                     
  | `GoogleBatchTaskHandler.groovy` | Auto-set `hyperdisk-balanced` boot disk for Hyperdisk-only families |                                                                     
  | `GoogleBatchMachineTypeSelectorTest.groovy` | Tests for new family selection, `isHyperdiskOnly()`, C4/C4A/C4D local SSD |
  | `GoogleBatchTaskHandlerTest.groovy` | Tests for Hyperdisk boot disk auto-configuration and standard family non-application |

#### Tests

  - GoogleBatchMachineTypeSelectorTest — verifies isHyperdiskOnly() returns true for C4/C4A/C4D/N4/N4A/N4D, false for N1/N2/E2/C2
  - GoogleBatchMachineTypeSelectorTest — verifies C4/C4A/C4D -lssd types return MemoryUnit(0) (pre-attached, not configurable)
  - GoogleBatchMachineTypeSelectorTest — verifies new families are selectable when explicitly requested
  - GoogleBatchTaskHandlerTest — verifies boot disk type is set to hyperdisk-balanced for C4 family
  - GoogleBatchTaskHandlerTest — verifies boot disk type is not set for standard families (N2)

### References

- [Google Cloud C4 machine series](https://cloud.google.com/compute/docs/general-purpose-machines#c4_series)
- [Google Cloud N4 machine series](https://cloud.google.com/compute/docs/general-purpose-machines#n4_series)  
- [Google Cloud Disk Types](https://cloud.google.com/compute/docs/disks#disk-types)
- [Disk Type Compatibility Matrix](https://cloud.google.com/compute/docs/disks#disk_types_by_machine_type)
- [Google Batch VM Configuration](https://cloud.google.com/batch/docs/vm-os-environment-overview)

### DIFF
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy
@@ -343,7 +343,10 @@ class GoogleBatchMachineTypeSelector {
         if( machineType.contains('lssd') )
             return true
 
-        if( (PARTIAL_LOCAL_SSD_SUPPORT_FAMILIES + NO_LOCAL_SSD_SUPPORT_FAMILIES).any { matchType(it, machineType) } )
+        if( PARTIAL_LOCAL_SSD_SUPPORT_FAMILIES.any { matchType(it, machineType) } )
+            return false
+
+        if( NO_LOCAL_SSD_SUPPORT_FAMILIES.any { matchType(it, machineType) } )
             return false
 
         return true


jorgee:
I have added support for Fusion by modifying the management of `local-ssd` when Fusion is enabled. I have prioritized the `local-ssd` but if the user specifies a instance that do not allow local SSD, it tries `hyperdisk-balanced` except for e2 families that is the one that only allows `pd-*,` it selects `pd-balanced`.

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -423,6 +428,20 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
                 )
             }
 
+            // Configure boot disk
```
jorgee, line 427: Not sure if it is really required. I have tested with 25.10.4 without Fusion using `machineTypes ` `c4-*` and `h4-*` and it was working. Not sure if it is only required when a boot disk image is specified. @sosoihd, is it your case?

sosoihd:
> I have added support for Fusion by modifying the management of `local-ssd` when Fusion is enabled. I have prioritized the `local-ssd` but if the user specifies a instance that do not allow local SSD, it tries `hyperdisk-balanced` except for e2 families that is the one that only allows `pd-*,` it selects `pd-balanced`.

Thanks a lot for your help on this, @jorgee! Really appreciate

One thing to flag: hyperdisk-balanced disks are not compatible with all machine families. For example, N1, N2, and N2D do not support

Hyperdisk types (see https://docs.cloud.google.com/compute/docs/disks/hd-types/hyperdisk-balanced?hl=en). So the fallback to hyperdisk-balanced when local SSD is not available might fail for those families.

However, N2 and N2D do support local SSDs (see https://docs.cloud.google.com/compute/docs/disks/local-ssd?hl=en#machine-series-lssd), so the local SSD path should work fine for them.

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
sosoihd: At the time, I was using version 25.09.2-edge and we had primarily tested with N4D and C4D machines, which is where we encountered the boot disk compatibility issues.

I just re-ran tests with both 25.09.2-edge and 25.10.4, and I'm now realizing the issue only occurs with the AMD variants (N4D, C4D). When I launched jobs with the others (C4, C4A, N4, N4A), the VM was correctly created with a hyperdisk-balanced boot disk. So maybe there's a subtlety I missed with those specific AMD families...

Regarding the boot disk image: we don't specify anything — we let Google Batch choose the default image, which is typically a COS (Container-Optimized OS) image since we run a Docker container inside the job.

jorgee: Ok, I will try with n4d, c4d, if fails then it is still needed. I was just thinking on removing the is only hyperdisk list

jorgee:
> One thing to flag: hyperdisk-balanced disks are not compatible with all machine families. For example, N1, N2, and N2D do not support
> 
> Hyperdisk types (see https://docs.cloud.google.com/compute/docs/disks/hd-types/hyperdisk-balanced?hl=en). So the fallback to hyperdisk-balanced when local SSD is not available might fail for those families.
> 
> However, N2 and N2D do support local SSDs (see https://docs.cloud.google.com/compute/docs/disks/local-ssd?hl=en#machine-series-lssd), so the local SSD path should work fine for them.

As far as I saw, the only family that do not support either hyperdisk or local-ssd is E2. But this is only required for the extra disk required when Fusion is enabled.
As summary, there are two cases: 
-Boot disk: will default to `pd-balanced` (default) except for hyperdisk only families that will `hyperdisk-balanced`, as you did.
-Fusion disk: it will assign `local-ssd` for all families where ssd is supported. From the list of not supported local ssd, only E2 do not support hyperdisk. Moreover, It seems hyperdisk will be the default for future instances. Thats the reason why I fallback to hyperdisk.

sosoihd:
> > One thing to flag: hyperdisk-balanced disks are not compatible with all machine families. For example, N1, N2, and N2D do not support
> > Hyperdisk types (see https://docs.cloud.google.com/compute/docs/disks/hd-types/hyperdisk-balanced?hl=en). So the fallback to hyperdisk-balanced when local SSD is not available might fail for those families.
> > However, N2 and N2D do support local SSDs (see https://docs.cloud.google.com/compute/docs/disks/local-ssd?hl=en#machine-series-lssd), so the local SSD path should work fine for them.
> 
> As far as I saw, the only family that do not support either hyperdisk or local-ssd is E2. But this is only required for the extra disk required when Fusion is enabled. As summary, there are two cases: -Boot disk: will default to `pd-balanced` (default) except for hyperdisk only families that will `hyperdisk-balanced`, as you did. -Fusion disk: it will assign `local-ssd` for all families where ssd is supported. From the list of not supported local ssd, only E2 do not support hyperdisk. Moreover, It seems hyperdisk will be the default for future instances. Thats the reason why I fallback to hyperdisk.

Agreed, that makes sense. With this approach we should be covering all scenarios. Thanks for the detailed breakdown!

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
jorgee: With 4cd it was failing, so it is required

jorgee:
Just removed a duplicated `bootDiskImage` management. It is now treated after machineType selection.

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy
```diff
@@ -255,6 +276,7 @@ class GoogleBatchMachineTypeSelector {
 
         // These families have a local SSD already attached and is not configurable.
         if( ((machineType.family == "c3" || machineType.family == "c3d") && machineType.type.endsWith("-lssd")) ||
+            ((machineType.family == "c4" || machineType.family == "c4a" || machineType.family == "c4d") && machineType.type.endsWith("-lssd")) ||
```
pditommaso: maybe this condition can be turned into an helper method and tested

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
```diff
@@ -401,11 +401,13 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
...
-                instancePolicy.setBootDisk(AllocationPolicy.Disk.newBuilder().setImage(batchConfig.getBootDiskImage()))
-
             if( fusionEnabled() && !disk ) {
-                disk = new DiskResource(request: '375 GB', type: 'local-ssd')
+                final reqMachineType = task.config.getMachineType()
+                if ( reqMachineType ) {
+                    disk = new DiskResource(request: '375 GB', type: chooseFusionDiskType(reqMachineType) )
+                } else {
+                    disk = new DiskResource(request: '375 GB', type: 'local-ssd')
+                }
```
pditommaso: Nit, conditional expression works better here

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy
jorgee: done

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
jorgee: done

plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchMachineTypeSelector.groovy
```diff
@@ -287,6 +313,50 @@ class GoogleBatchMachineTypeSelector {
...
+     *
+     * @param machineType Machine type
+     * @return Boolean value indicating if the machine type can have local ssd disks.
+     */
+    static boolean hasLocalSsd(String machineType) {
+        if( machineType.contains('lssd') )
+            return true
+
+        if( (PARTIAL_LOCAL_SSD_SUPPORT_FAMILIES + NO_LOCAL_SSD_SUPPORT_FAMILIES).any { matchType(it, machineType) } )
+            return false
```
bentsherman:
```suggestion
        if( PARTIAL_LOCAL_SSD_SUPPORT_FAMILIES.any { matchType(it, machineType) } )
            return false

        if( NO_LOCAL_SSD_SUPPORT_FAMILIES.any { matchType(it, machineType) } )
            return false
```

PR #6845, opened 2026-02-19
Ensure main script is the first one in the lineage WorkflowRun record

jorgee:
User on nextflow slack found a bug with how script files are listed in the lineage:
https://nextflow.slack.com/archives/C02T98A23U7/p1771421589893539

We intended for the main script to always be first, but then we sort the entire list by path, which breaks that guarantee
This PR modifies this behaviour to only sort the script files before adding the main script


modules/nf-lineage/src/test/nextflow/lineage/LinObserverTest.groovy
```diff
@@ -139,7 +139,7 @@ class LinObserverTest extends Specification {
         when:
         def files = observer.collectScriptDataPaths(new PathNormalizer(metadata))
         then:
-        observer.allScriptFiles() >> [ scriptFile, module1, module2 ]
+        observer.allScriptFiles() >> [ module2, scriptFile, module1 ]
```
pditommaso, line 142: I'm confused by this, the main file should not be `scriptFile`? and therefore stating the title should not be `[ scriptFile, module1, module2 ]`

jorgee: In this statement, we are not checking the order. It mocks the `allScriptFiles` method to return an unsorted set of files including the main script. It is called at the beginning of the `collectScriptDataPaths ` method and helps to test the rest of the implementation. ` files` is the list that should include first the scriptFile and later the sorted modules.

pditommaso: Oops, you are right this is not the assertion part

modules/nf-lineage/src/test/nextflow/lineage/LinObserverTest.groovy
pelleImu, line 109: Rename to: Should collect script files with invoked script at index zero ?

PR #6847, opened 2026-02-19
Speed up submission of Google Batch array jobs

thalassemia:
I noticed that Nextflow was submitting my Google Batch array jobs very slowly, maybe about one submission every 3 seconds. That was despite no explicit submission throttling in my Nextflow config. In my investigation, I identified two sources of latency.

First, the Google Batch executor uses `TaskPollingMonitor` to submit jobs. This blocks on each submission and greatly limits the achievable throughput compared to the `ParallelPollingMonitor` used by the AWS Batch executor. To fix this, I essentially copied [the `AwsBatchExecutor` logic](https://github.com/nextflow-io/nextflow/blob/master/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy) in `GoogleBatchExecutor`. The retry exceptions were sourced from [`BatchClient`](https://github.com/nextflow-io/nextflow/blob/master/plugins/nf-google/src/main/nextflow/cloud/google/batch/client/BatchClient.groovy) and [documented here](https://docs.cloud.google.com/java/docs/reference/gax/latest/com.google.api.gax.rpc#exceptions).

Second, creation of task runs for job arrays happens sequentially in a for loop. I parallelized that.

I tested these changes using the following config and Nextflow script:

```nextflow
process {
    executor = 'google-batch'
    container = 'ubuntu:22.04'
    machineType = 'e2-medium'
    cpus = 1
    memory = '4 GB'
}

google {
    // Set your project ID
    project = 'PROJECT_ID'
    location = 'us-central1'
    batch {
        spot = true
        maxSpotAttempts = 3
    }
}
```

```nextflow
params.num_tasks = 50

process sayHello {
    tag "$task_id"

    array 5
    
    input:
    val task_id

    output:
    stdout

    script:
    """
    echo "Hello from task $task_id on \$(hostname)"
    sleep 5
    echo "Task $task_id completed"
    """
}

workflow {
    task_ids = Channel.of(1..params.num_tasks)
    sayHello(task_ids) | view
}
```
By reading `.nextflow.log`, I got the following times from starting Nextflow to submission of the final job:
| No changes | `ParallelPollingMonitor` | Parallel `createTaskArray` | Both |
| --- | ---| --- | --- | 
| 27 s | 28 s | 17 s | 8 s |

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskArrayCollector.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskArrayCollector.groovy
@@ -139,8 +139,14 @@ class TaskArrayCollector {
     protected TaskArrayRun createTaskArray(List<TaskRun> tasks) {
         // prepare child job launcher scripts
         final handlers = tasks.collect( t -> executor.createTaskHandler(t).withArrayChild(true) )
-        for( TaskHandler handler : handlers ) {
-            handler.prepareLauncher()
+        // prepare launchers in parallel to avoid sequential blocking I/O on cloud storage
+        handlers.parallelStream().forEach { handler ->
+            try {
+                handler.prepareLauncher()
+            } catch (Exception e) {
+                log.error("Failed to prepare launcher for task ${handler.task.name}", e)
+                throw e
+            }
         }
 
         // create work directory
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchExecutor.groovy
@@ -187,11 +187,8 @@ class AwsBatchExecutor extends Executor implements ExtensionPoint, TaskArrayExec
     protected TaskMonitor createTaskMonitor() {
 
         // create the throttling executor
-        // note this is invoke only the very first time a AWS Batch executor is created
-        // therefore it's safe to assign to a static attribute
-        submitter = createExecutorService('AWSBatch-executor')
-
-        reaper = createExecutorService('AWSBatch-reaper')
+        this.submitter = createExecutorService('AWSBatch-executor')
+        this.reaper = createExecutorService('AWSBatch-reaper')
 
         final pollInterval = config.getPollInterval(name, Duration.of('10 sec'))
         final dumpInterval = config.getMonitorDumpInterval(name)
@@ -230,7 +227,7 @@ class AwsBatchExecutor extends Executor implements ExtensionPoint, TaskArrayExec
      * @return Creates a {@link ThrottlingExecutor} service to throttle
      * the API requests to the AWS Batch service.
      */
-    private ThrottlingExecutor createExecutorService(String name) {
+    private ThrottlingExecutor createExecutorService(String poolName) {
 
         // queue size can be overridden by submitter options below
         final qs = 5_000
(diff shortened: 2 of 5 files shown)


bentsherman (approved):
Thanks for the contribution. I'm not sure about adding a parallel loop in the core logic, but the google batch submitter pool seems reasonable and I don't want to hold up that piece any longer

pditommaso:
**Post-merge note — worth flagging for follow-up.**

A caveat that's easy to lose after a merge: the Google Batch executor took a *long* time and many iterations to reach its current stability (job/task status reconciliation, retry policy in `BatchClient`, spot handling, cleanup semantics). Moving submission and deletion onto an async `ThrottlingExecutor` pool re-introduces concurrency into exactly the paths we spent that effort hardening, so this is a place to watch for regressions rather than assume parity with AWS.

**Unexpected issues the parallel submitter can open**
- **Orphaned jobs on shutdown.** `shutdown()` calls `submitter.shutdownNow()`, which interrupts in-flight submissions. The `session.canSubmitTasks()` guard prevents *queued* submissions from creating untracked jobs, but a submission already inside `createJob` when interrupted can leave a Batch job created on GCP that was never added to the running queue — so it's never reaped.
- **Cleanup failures can abort a healthy run.** The `reaper` reuses the same executor config, including `onFailure { session?.abort(t) }`, and deletes now go through it with the `Future` discarded. A delete that permanently fails (e.g. `PermissionDenied`, or `ResourceExhausted` after retries — neither is retried by `BatchClient` for deletes) will abort the whole session during teardown.
- **Backpressure/capacity accounting.** The monitor's semaphore is released unconditionally in `evict()`, which is called defensively from several paths; any double-evict slowly inflates the effective `capacity`. Latent (it's survived in AWS), but this exposes a second provider to it.
- **Ordering.** Submission is no longer serialized on the monitor thread, so per-process submission order is no longer preserved.

**Why this is less effective here than on AWS Batch.** On AWS the `submitter` is also wired into `AwsBatchProxy`, which routes *every* Batch API call — including the high-frequency `describeJobs` polling (priority-boosted) — through the `ThrottlingExecutor`. So the auto-throttle/back-off loop sees the whole API load and protects the account quota globally. On Google we only wrap **submit** and **delete**; all polling (`getJobStatus`/`getTaskStatus`) goes straight through `BatchClient` and bypasses the throttler. The result is that the auto-throttle — a core benefit of the AWS design — is largely inert here, since the calls that most often exhaust quota in large runs (polling) neither feed the back-off loop nor get throttled. In practice this is a *submission-concurrency* win, not the quota-aware throttling AWS gets.

**Minor effect on task arrays.** For the titled use case, the impact is small: an array is submitted as a single `createJob` per task group, so parallel submission has almost nothing to parallelize. The PR's own benchmark shows `ParallelPollingMonitor` alone at 27→28 s (no gain); the real speedup came from parallelizing `createTaskArray`, which did not land in the merged commit. The concurrency added here mainly benefits high-volume **non-array** submission. Worth confirming the array-creation optimization is tracked as a follow-up.

bentsherman:
Thanks for the extra context Paolo

@thalassemia did you do any more load testing on this PR aside from the benchmark? I'm wondering how well the parallelization holds up in production

My main concern would be the inconsistent throttling (applied to submission/deletion but not status checks). But I think we can leave it for now, see how it holds up in the next few releases, and revert if it causes problems (I'd rather revert than adding more and more machinery if it comes to that).

thalassemia:
@bentsherman I did not do any additional load testing. Looking this over, I agree that the change which landed probably benefits high-volume non-array submissions more than array ones, though I haven't tested that.

PR #6854, opened 2026-02-23
Report actual GCP zone in Google Batch trace records

### LINKED ISSUE #6646: Google Batch: Report actual zone where tasks execute in trace records
# Google Batch: Report actual zone where tasks execute in trace records

## New feature

The Google Batch executor currently reports the configured **region** (e.g., `europe-west2`) in trace records rather than the actual **zone** where tasks execute (e.g., `europe-west2-a`, `europe-west2-b`, `europe-west2-c`). This prevents accurate cost estimation for spot instances, as spot pricing varies by zone rather than region.

**Current behavior:**
- `CloudMachineInfo.zone` is populated with `config.location` (region-level setting)
- Trace records contain only the region: `cloudZone = "europe-west2"`

**Desired behavior:**
- `CloudMachineInfo.zone` should contain the actual zone where Google Batch allocated the task
- Trace records should contain the specific zone: `cloudZone = "europe-west2-a"`

## Use case

**Cost estimation for Google Batch workloads:**

When running workflows with spot instances on Google Batch, accurate cost tracking requires knowing the specific zone where each task executed. Cloud pricing databases (like those used by Seqera Platform) store spot prices per zone (e.g., `europe-west2-a`, `europe-west2-b`, `europe-west2-c`) rather than per region.

**Current limitation:**
1. Spot price lookup fails because trace records contain region (`europe-west2`) but price database is keyed by zone (`europe-west2-a`)
2. Cost estimates cannot be calculated for Google Batch workflows
3. Users lack visibility into actual resource costs

**Deployment scenarios:**
- Seqera Platform integration for cost tracking and billing
- Custom monitoring solutions that track per-task resource costs
- Audit and compliance reporting requiring accurate zone information
- Multi-zone resource optimization analysis

## Suggested implementation

**1. Retrieve zone information from Google Batch API:**

After task completion, query the Google Batch API to retrieve the actual zone where the task executed:

```groovy
// In GoogleBatchTaskHandler.groovy
def getTaskStatus() {
    final job = client.getJob(jobId)
    final taskStatus = job.getStatus()
    // Extract actual zone from task allocation metadata
    final actualZone = extractZoneFromTaskStatus(taskStatus)
    return actualZone
}
```

**2. Update CloudMachineInfo with actual zone:**

Modify `GoogleBatchTaskHandler` to populate the zone field with the actual execution zone rather than the configured region:

```groovy
// In GoogleBatchTaskHandler.groovy, line ~351
machineInfo = new CloudMachineInfo(
    type: machineType.type,
    zone: getActualExecutionZone(),  // Instead of machineType.location
    priceModel: machineType.priceModel
)
```

**3. Google Batch API reference:**

The zone information should be available from:
- Job status metadata after task allocation
- Instance policy or allocation policy fields
- Task group status details

**Alternative approach:**

If retrieving zone information adds too much API overhead, consider:
- Lazy retrieval: Only fetch zone when trace records are generated
- Cache zone information per job to minimize API calls
- Make it optional via configuration flag

**Related components:**
- `plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy` (lines 351-356, 656-658)
- `plugins/nf-google/src/main/nextflow/cloud/google/batch/client/BatchClient.groovy`
- `modules/nextflow/src/main/groovy/nextflow/cloud/types/CloudMachineInfo.groovy`
- `modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy`

**Backwards compatibility:**

This change should be backwards compatible as it improves the accuracy of existing data without changing the field structure or API contracts.

jonmarti:
## Summary

Fixes #6646

The Google Batch API does not expose the actual zone where a task executes — it only provides the configured region (e.g. `europe-west2`). This means trace records report the region rather than the specific zone (e.g. `europe-west2-a`), which is important for cost analysis and debugging placement decisions.

This PR implements the workaround suggested by @pditommaso: querying the [GCP instance metadata service](https://cloud.google.com/compute/docs/metadata/overview) from within the task to capture the real zone.

## Changes

- **`TaskRun.groovy`** — Added `CMD_ZONE = '.command.zone'` constant for the zone metadata file
- **`GoogleBatchScriptLauncher.groovy`** — Inject a `curl` call to the GCP metadata endpoint (`http://metadata.google.internal/computeMetadata/v1/instance/zone`) in `headerScript()`, writing the result to `.command.zone` in the task work directory. The call is silent and non-fatal (`2>/dev/null || true`)
- **`GoogleBatchTaskHandler.groovy`** — On task completion, read `.command.zone`, parse the zone name from the metadata format (`projects/<id>/zones/<zone>`), and update the `CloudMachineInfo` with the actual zone. Uses a `volatile boolean zoneUpdated` flag to ensure the file is read at most once (avoiding repeated remote I/O on gcsfuse)

## Design decisions

- **Zone capture in `headerScript()`** — This runs at the top of `.command.run`, which executes for both regular tasks and array task children (each child runs its own `.command.run`). The parent array launcher (`.command.sh`) does not capture zone since each child writes its own `.command.zone`
- **Graceful degradation** — If the metadata service is unreachable (e.g. non-GCP environments, Fusion-enabled tasks that bypass `GoogleBatchScriptLauncher`), the trace record falls back to the configured region as before
- **Read-once semantics** — The `zoneUpdated` flag is set in a `finally` block, so even if reading fails, we don't retry on every `getMachineInfo()` call (which is invoked frequently by the polling monitor, Tower observer, and error handlers)

## Test plan

- [x] `GoogleBatchScriptLauncherTest` — 2 new tests: zone capture present in header script; zone capture absent from array launch command
- [x] `GoogleBatchTaskHandlerTest` — 6 new tests:
  - Zone updated from file on completion
  - Original zone preserved when file is missing
  - Null machineInfo returns null
  - Zone not updated when task is not completed
  - Malformed file content handled gracefully
  - Zone file read only once across multiple `getMachineInfo()` calls
- [x] All 76 tests pass (7 launcher + 69 handler), 0 failures

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskRun.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskRun.groovy
@@ -561,6 +561,7 @@ class TaskRun implements Cloneable {
     static final public String CMD_STAGE = '.command.stage'
     static final public String CMD_TRACE = '.command.trace'
     static final public String CMD_ENV = '.command.env'
+    static final public String CMD_ZONE = '.command.zone'
 
 
     String toString( ) {
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchScriptLauncher.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchScriptLauncher.groovy
@@ -109,6 +109,8 @@ class GoogleBatchScriptLauncher extends BashWrapperBuilder implements GoogleBatc
             result += 'chmod +x $HOME/.nextflow-bin/*\n'
             result += 'export PATH=$HOME/.nextflow-bin:$PATH\n'
         }
+        // capture the actual GCP zone from the instance metadata service
+        result += "curl -sf -H 'Metadata-Flavor: Google' http://metadata.google.internal/computeMetadata/v1/instance/zone > ${Escape.path(bean.workDir)}/${TaskRun.CMD_ZONE} 2>/dev/null || true\n"
         return result
     }
 
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
@@ -119,6 +119,11 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
     private volatile long timestamp
 
     /**
+     * Flag to indicate that the zone has been read from the zone file
+     */
+    private volatile boolean zoneUpdated
+
+    /**
      * A flag to indicate that the job has failed without launching any tasks
      */
     private volatile boolean noTaskJobfailure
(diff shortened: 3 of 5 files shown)


pditommaso:
When I suggested this?! 😄

jonmarti:
> When I suggested this?! 😄

@pditommaso https://github.com/nextflow-io/nextflow/issues/6646#issuecomment-3641073035 :sweat_smile:

jonmarti:
@pditommaso @munishchouhan I checked #6646 today as part of the resolution of https://seqera.atlassian.net/browse/NF-353 and eventually this ticket in Expedite: https://seqera.atlassian.net/browse/ES-190

pditommaso:
Likely an agent, not me 😄

pditommaso:
I want to explore if it can be avoided the use of an external file and rely on the API to detected the zone. See https://github.com/nextflow-io/nextflow/pull/6855

jonmarti:
Closing in favor of #6855 — Paolo's approach of parsing the zone from `StatusEvent` descriptions is a better fit:

- No core module changes (`TaskRun`)
- No extra `.command.zone` file written to GCS per task
- Works with Fusion-enabled tasks (independent of the script launcher)
- Leverages data already available from the Batch API — the `zones/<zone>/instances/<instance-id>` pattern in state transition events is [documented](https://cloud.google.com/batch/docs/troubleshooting) and already relied upon in this codebase for spot preemption detection

The only trade-off is one extra `getTaskStatus()` API call per task, but with the read-once flag it's negligible.

PR #6855, opened 2026-02-23
Report actual GCP zone from Batch status events

### LINKED ISSUE #6646: Google Batch: Report actual zone where tasks execute in trace records

pditommaso:
## Summary

Fixes #6646

Alternative to #6854 — instead of writing a `.command.zone` file from inside the task, this approach resolves the actual execution zone by parsing Google Batch `StatusEvent` descriptions, which include the zone in the format `zones/<zone>/instances/<instance-id>`.

This avoids creating an additional bucket file and keeps all changes within the Google plugin (no core module changes).

## Changes

- **`GoogleBatchTaskHandler.groovy`** — On task completion, `getMachineInfo()` calls `updateZoneFromEvents()` which fetches the task status events and parses the zone from the event description using a regex. Updates `CloudMachineInfo` with the actual zone. Uses a `volatile boolean zoneUpdated` flag to ensure the API is called at most once.

## Design decisions

- **Graceful degradation** — If no zone is found in the events (e.g. the event descriptions don't contain zone info), the trace record falls back to the configured region as before
- **Read-once semantics** — The `zoneUpdated` flag is set in a `finally` block, so even if the API call fails, we don't retry on every `getMachineInfo()` call
- **No core module changes** — Unlike the `.command.zone` approach, this doesn't touch `TaskRun` or add new bucket files

## Test plan

- [x] Data-driven test covering: zone from succeeded event, zone from preemption event, no zone in event, empty events
- [x] Null machineInfo and not-completed guards
- [x] Zone resolved only once across multiple `getMachineInfo()` calls
- [x] API error handled gracefully with fallback to original zone
- [x] All tests pass

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -1,8 +1,8 @@
 {% extends "!layout.html" %}
 
 {% block document %}
-    <div class="admonition site-banner">
-        <p class="admonition-title">Nextflow documentation is migrating</p>
+    <div class="admonition site-banner" style="background: #E2F7F3; border-left: none;">
+        <p class="admonition-title" style="background: #0CAE8E;">Nextflow documentation is migrating</p>
         <p>Nextflow documentation is being migrated to <a href="https://docs.seqera.io/nextflow/">docs.seqera.io/nextflow</a>. This site will remain available throughout the migration.</p>
     </div>
     {{ super() }}


jonmarti:
@pditommaso awesome! closed my PR in favor to this one (no core module changes, no extra file on GCS, doesn't depend on the script launcher, ...)

Issue #5718, opened 2025-01-28
file.getName() doesn't strip the parent directory as described in the documentation.


mahesh-panchal:
### Expected behavior and actual behavior

When using the `getName()` method on TaskPath objects, it should [just output the filename](https://www.nextflow.io/docs/latest/reference/stdlib.html#getting-attributes) and not the parent directory if the file gets staged in a subdirectory of a process. The actual behaviour is that `getName` provides the full staged path, unlike `getBaseName` which does strip the parent directory.

### Steps to reproduce the problem

```nextflow
workflow {
    TASK( Channel.fromPath(['data/sampleA.ext.gz','data/sampleB.ext.gz','data/sampleC.ext.gz']).collect() )
        .view()
}

process TASK {
    input:
    path files_in, stageAs: 'zipped/*'

    script:
    """
    echo "${files_in*.name}"
    echo "${files_in*.baseName}"
    """

    output:
    stdout
}

```

### Program output 

```console
$ nextflow run main.nf

 N E X T F L O W   ~  version 24.10.4

Launching `main.nf` [big_davinci] DSL2 - revision: b223d1ff7a

executor >  local (1)
[c0/b424af] process > TASK [100%] 1 of 1 ✔
[zipped/sampleA.ext.gz, zipped/sampleB.ext.gz, zipped/sampleC.ext.gz]
[sampleA.ext, sampleB.ext, sampleC.ext]

```

### Environment 

* Nextflow version: 24.10.4
* Java version: openjdk 17.0.14-internal 2025-01-21
* Operating system: Linux
* Bash version: GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)

bentsherman:
Task paths behave differently on purpose. See the note at the end of the section [here](https://www.nextflow.io/docs/latest/process.html#input-files-path)

mahesh-panchal:
I think it would be better named `getPath` to get the full path like that, and most of the time `getName` and `getPath` return the same result.

mahesh-panchal:
I'm not sure I understand. All I'm proposing is that `getName` is renamed to `getPath` and then `getName` implements returning `x.fileName.name` (as described in the doc section you linked). If these functions are used internally, then they can still be used, but I guess everywhere `getName` is used needs to be replaced with `getPath`. It's more to make the terminology consistent with the docs.

mahesh-panchal:
[...]

PR #6869, opened 2026-02-26
docs: clarify Path.name for staged process path inputs

LarytheLord:
## Summary
Clarifies `Path.name` behavior in the stdlib reference for process task paths.

The current `name` property description says it returns only the file name, but for process `path` inputs staged into subdirectories it can include the staged relative path (for example `my-dir/file.txt`). This PR adds that caveat and points users to `fileName.name` for the leaf filename.

## Change
- Updated `docs/reference/stdlib-types.md` under `name: String`.

## Why
This keeps `stdlib-types` consistent with the behavior documented in `process.md` and avoids confusion reported in #5718.

Refs: #5718

### DIFF
--- a/docs/reference/stdlib-types.md
+++ b/docs/reference/stdlib-types.md
@@ -521,8 +521,7 @@ The following properties are available:
 
 `name: String`
 : The path name, e.g. `/some/path/file.txt` -> `file.txt`.
-  For process `path` inputs staged into a subdirectory, this includes the staged relative path
-  (e.g. `my-dir/file.txt`). Use `fileName.name` to get only the file name in that case.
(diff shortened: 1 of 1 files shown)


docs/reference/stdlib-types.md
```diff
@@ -521,6 +521,8 @@ The following properties are available:
 
 `name: String`
 : The path name, e.g. `/some/path/file.txt` -> `file.txt`.
+  For process `path` inputs staged into a subdirectory, this includes the staged relative path
+  (e.g. `my-dir/file.txt`). Use `fileName.name` to get only the file name in that case.
```
bentsherman:
```suggestion
: For files staged as a task input, the path name is the path relative to the task directory (e.g., `my-dir/file.txt`). Use `fileName.name` for task paths to get only the file name.
```

PR #6873, opened 2026-02-26
Add ResourcesResolver for pre-flight resource optimization

pditommaso:
## Summary

- Adds ADR for a plugin-based `ResourcesResolver` extension point that allows a scheduler optimization engine to dynamically adjust task resource allocation (cpus, memory, time, disk) before execution
- Follows the `ContainerResolver` plugin pattern: interface in core, no-op default, overridden by nf-seqera plugin
- Called from `TaskProcessor.submitTask()` after hash computation, before `executor.submit()`
- Optimized values stored as overrides on `TaskConfig`, preserving original declared values

## Test plan

- [ ] Implement `TaskConfig` overrides map with tests
- [ ] Create core extension point (interface, DTOs, default, provider)
- [ ] Wire into `TaskProcessor.submitTask()`
- [ ] Implement `SeqeraResourcesResolver` in nf-seqera plugin
- [ ] Verify TraceRecord and TaskBean reflect overridden values
- [ ] Integration test with mock resolver

### DIFF
--- a/adr/20260225-resources-resolver.md
+++ b/adr/20260225-resources-resolver.md
@@ -1,0 +1,134 @@
+# Pre-flight Resource Optimization via ResourcesResolver
+
+- Authors: Paolo Di Tommaso
+- Status: draft
+- Date: 2026-02-25
+- Tags: scheduler, resources, plugin
+
+## Summary
+
+Add a plugin-based `ResourcesResolver` extension point that allows a scheduler optimization engine to modify task resource allocation (cpus, memory, time, disk) before execution. The Seqera scheduler plugin provides the concrete implementation calling the optimization API.
+
+## Problem Statement
+
+Tasks currently execute with user-declared resources or static `resourceLimits` caps. There is no mechanism for an external optimization engine to dynamically adjust resource allocation based on historical execution data, prediction models, or scheduling constraints before a task runs.
+
+## Design
+
+### Pattern
+
+Follows the `ContainerResolver` plugin pattern:
+
+- **`ResourcesResolver`** — interface extending `ExtensionPoint` in core (`modules/nextflow`)
+- **`DefaultResourcesResolver`** — no-op implementation in core, returns `null`
+- **`ResourcesResolverProvider`** — loads the active resolver via `Plugins.getPriorityExtensions()`
+- **`SeqeraResourcesResolver`** — in `nf-seqera` plugin, calls the scheduler optimization API. Uses `@Priority(-10)` to override the default.
+
+### Call site
+
+`TaskProcessor.submitTask()` — after hash computation, before `executor.submit()`. This means:
+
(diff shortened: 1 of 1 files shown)


adr/20260225-resources-resolver.md
```diff
@@ -0,0 +1,134 @@
...
+- **`ResourcesResolver`** — interface extending `ExtensionPoint` in core (`modules/nextflow`)
+- **`DefaultResourcesResolver`** — no-op implementation in core, returns `null`
+- **`ResourcesResolverProvider`** — loads the active resolver via `Plugins.getPriorityExtensions()`
+- **`SeqeraResourcesResolver`** — in `nf-seqera` plugin, calls the scheduler optimization API. Uses `@Priority(-10)` to override the default.
+
+### Call site
+
+`TaskProcessor.submitTask()` — after hash computation, before `executor.submit()`. This means:
+
+- The cache hash reflects **declared** (original) resources, not optimized ones — caching stays deterministic based on user intent.
```
bentsherman, line 31: This point is moot because resource requests aren't included in the task hash anyway

adr/20260225-resources-resolver.md
```diff
@@ -0,0 +1,134 @@
...
+The resolver receives a `ResourcesRequest` DTO containing:
+
+- cpus, memory, disk, time
+- accelerator (count, type)
+- machineType
+- resourceLimits (upper bounds)
+- container image
+- process name
+- attempt number
+- task index
```
bentsherman, line 46: Is this the within-process index or within-run index? Not sure that it's relevant either way

adr/20260225-resources-resolver.md
```diff
@@ -0,0 +1,134 @@
...
+### Request/Response
+
+The resolver receives a `ResourcesRequest` DTO containing:
+
+- cpus, memory, disk, time
+- accelerator (count, type)
+- machineType
+- resourceLimits (upper bounds)
+- container image
+- process name
```
bentsherman, line 44: Base name or fully-qualified name? I assume it should be the base name

adr/20260225-resources-resolver.md
```diff
@@ -0,0 +1,134 @@
...
+- attempt number
+- task index
+- input files total size in bytes
```
bentsherman, line 47: Since this request is in-memory, I would just pass the raw data e.g. list of file names with sizes, or a map based on the process input names

adr/20260225-resources-resolver.md
pditommaso: Good point

adr/20260225-resources-resolver.md
pditommaso: Actually right now it's using FQ name, however I agree it should not

adr/20260225-resources-resolver.md
pditommaso: Within run, but not really needed

pditommaso:
Closing for now

PR #6874, opened 2026-02-27
Handle Azure Batch ActiveJobAndScheduleQuotaReached with retry

### LINKED ISSUE #5575: Wait for Azure Batch job quota to clear before submitting a new job

adamrtalbot:
## Summary

Fixes #5575

When Azure Batch hits the active job and schedule quota (HTTP 409 `ActiveJobAndScheduleQuotaReached`), Nextflow now retries with configurable backoff instead of failing immediately.

## Changes

- **`AzBatchService.groovy`**: Added `applyCreateJob()` using Failsafe `RetryPolicy` with a fixed delay to handle `ActiveJobAndScheduleQuotaReached` quota errors. This layers on top of the existing `apply()` method — `apply()` handles short-lived transient errors (408, 429, 500, etc.) while `applyCreateJob()` handles the longer-lived quota exhaustion with configurable fixed delays. Extracted `createJobRequest()` and `isJobQuotaError()` as protected methods for testability.
- **`AzBatchOpts.groovy`**: Added two new config options:
  - `azure.batch.maxJobQuotaRetries` (int, default `3`) — maximum number of retries when quota is reached
  - `azure.batch.jobQuotaRetryDelay` (Duration, default `2 min`) — delay between retries
- **`AzBatchServiceTest.groovy`**: Added 6 unit tests covering: success on first try, retry then succeed, exhaust retries, zero retries, non-quota 409 errors, and non-HTTP exceptions.

## Behavior

When job creation fails with `ActiveJobAndScheduleQuotaReached`:
1. A warning is logged: `Azure Batch active job quota reached - waiting 2m before retry (attempt 1 of 3)`
2. Failsafe waits for the configured delay
3. Job creation is retried
4. If retries are exhausted, an `IllegalStateException` is thrown with a helpful message suggesting config changes


plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -450,11 +450,46 @@ class AzBatchService implements Closeable {
...
+                    log.warn "Azure Batch active job quota reached - waiting ${retryDelay} before retry (attempt ${attempt + 1} of ${maxRetries})"
+                    sleep(retryDelay.toMillis())
+                }
+                else {
+                    throw e
+                }
+            }
+        }
+    }
+
```
pditommaso: How this improve vs `apply(() -> client.createJob(content))`

adamrtalbot: `apply()` uses Failsafe to retry on transient HTTP errors (`RETRY_CODES = [408, 429, 500, 502, 503, 504]`), IOExceptions, and TimeoutExceptions. It does **not** retry on 409 — correctly, since most 409 Conflict errors shouldn't be retried.

However, `ActiveJobAndScheduleQuotaReached` is a specific 409 that **is** transient — the quota frees up as existing jobs complete. It needs a different retry strategy than `apply()` provides:

- **Longer delays** — minutes not seconds, since jobs take time to finish
- **Fewer retries** — with 2-minute delays, we don't want dozens of attempts
- **User-visible warning** — so users know they're hitting the quota limit and can adjust config

Note that `createJobRequest()` still calls `apply()` internally, so both layers are active: `apply()` handles short-lived transient errors, while `safeCreateJob()` handles the longer-lived quota exhaustion on top.

adamrtalbot: `apply()` handles transient errors (e.g. 429, 50*). A quota error will recover but requires a longer wait.

A 409 means there are too many Jobs & Tasks running, which means you should hold off before submitting more. This means the correct response isn't a quick retry via the standard Azure API error strategy, but a more realistic backoff. Not in 30 seconds but in minutes, perhaps hours. I've been conservative here, but it's reasonable to increase the timeout to a whole day if you are running at very high throughput.

Note `safeCreateJob` uses `apply()` internally, so both layers are active.

adamrtalbot: Also, adding 409 to [the accepted error codes](https://github.com/nextflow-io/nextflow/blob/fix/azure-batch-scratch/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy#L1083-L1084) will mean we treat other 409s the same as job submission. E.g. a "pool already exists" error will be handled with the same backoff retry strategy as reaching the job quota limit which is invalid behaviour.

pditommaso: You have full control on the retry logic extending that one or using a variation

pditommaso: Sorted

PR #6880, opened 2026-03-02
Suppress lint progress logging with `-q` flag

### LINKED ISSUE #6856: `nextflow lint` ignores `--quiet` flag
## Bug report 

When running nextflow lint in CI I would like to minimize the logging. I thought `--quiet` would do that, but doesn't seem to have any effect.

Run `nextflow -q lint .` 

### Program output 

```
Linting: modules/nf-core/samtools/stats/main.nf
Linting: modules/nf-core/samtools/view/tests/nextflow.config
Linting: modules/nf-core/samtools/view/main.nf
Linting: modules/nf-core/samtools/view/main.nf
Linting: modules/nf-core/samtools/view/tests/nextflow.config
Linting: modules/nf-core/samtools/view/tests/nextflow.config
Nextflow linting complete!
 ✅ 49 files had no errors
```

### Environment 

Version: 26.01.1-edge build 11194
Created: 11-02-2026 21:07 UTC (22:07 CEST)
System: Mac OS X 26.3
Runtime: Groovy 4.0.30 on OpenJDK 64-Bit Server VM 21.0.2+13-58
Encoding: UTF-8 (UTF-8)

ewels:
nextflow lint ignores the --quiet (-q) flag. The StandardErrorListener class writes progress messages directly to AnsiConsole.out, bypassing Nextflow’s logger infrastructure which respects the -q flag.

This PR fixes that. Closes #6856


ewels:
## Test fix follow-up

This commit fixes two test failures in `CmdLintTest` that were introduced alongside the quiet flag feature:

### 1. `AnsiConsole.out` vs `System.out` (`ErrorListener.groovy`)

`StandardErrorListener` was printing all output via `AnsiConsole.out`, which is a **static field initialized once at class-load time**. The test infrastructure (`OutputCapture`) works by replacing `System.out` via `System.setOut()` at test startup — but `AnsiConsole.out` holds a reference to the *original* `System.out` and is never updated. As a result, error output in quiet mode bypassed the capture stream entirely, causing `capture.toString().contains("Error")` to return `false`.

Fix: replace all `AnsiConsole.out.print/flush()` calls with `System.out.print/flush()`. ANSI rendering is already controlled via `Ansi.setEnabled(ansiLog)`, so `AnsiConsole` is not needed for that purpose.

### 2. `getQuiet()` vs `isQuiet()` (`CmdLintTest.groovy`)

`CmdLint` is `@CompileStatic`. When it accesses `launcher.options.quiet`, the Groovy static compiler emits a call to **`isQuiet()`** (the standard Java boolean getter convention), not `getQuiet()`. The Spock test stubs were written as `getQuiet() >> true`, which was never invoked — so `quiet` was always `false` inside `StandardErrorListener`, and all progress messages were printed regardless of the flag.

Fix: change both test stubs from `getQuiet() >> true` to `isQuiet() >> true`.

ewels:
## Follow-up fix: output routing in StandardErrorListener

The previous commit introduced a regression in the `test_parser_v2` integration tests (`error-message.nf`). Here's what happened and how it's now fixed.

### Root cause

`StandardErrorListener` was printing via `AnsiConsole.out`, which is a Jansi `AnsiPrintStream`. When output is piped (non-TTY), `AnsiConsole.out` transparently strips ANSI escape codes at the stream level — this is its primary purpose. Switching to plain `System.out` bypassed that stripping.

The integration test runs `nextflow -q run error-message.nf | tee stdout` then greps `stdout` for the plain-text error message. With `System.out`, any ANSI codes that slipped through (due to `Ansi.setEnabled()` being global/racy) appeared raw in the pipe, causing the `grep` to fail.

### Fix

A private `print()` helper is introduced in `StandardErrorListener` that:

- When `ansiLog=true`: delegates to `AnsiConsole.out`, which renders ANSI on a TTY and strips codes when piped — correct production behaviour.
- When `ansiLog=false`: calls `Ansi.setEnabled(false)` **immediately before** calling `text.toString()`, ensuring the global Ansi state is correct at render time regardless of what other components may have changed it. The resulting plain-text string is then written to `System.out`.

Writing to `System.out` (not `AnsiConsole.out`) in the `ansiLog=false` path means test infrastructure that replaces `System.out` (like `OutputCapture`) captures the output correctly — which is what the unit tests need.

ewels:
## Further fix: ANSI code stripping

After more investigation, the previous approach of calling `Ansi.setEnabled(false)` in the `print()` helper just before `toString()` was also incorrect.

**Root cause**: `Ansi.setEnabled()` is **static-only** in Jansi 2.4.2 — there is no per-instance enabled flag. More importantly, `Ansi.toString()` returns the raw `StringBuilder` contents without any stripping at render time. Escape codes are baked into the buffer when `fg()`, `bold()`, etc. are called — not when `toString()` is invoked.

So calling `setEnabled(false)` just before `toString()` was already too late: the codes were already in the buffer from when `onError()` built up `term`.

**Fix**: move `Ansi.setEnabled(ansiLog)` into the `ansi()` helper, so the global flag is set correctly *before* every `Ansi` object is created and before any `fg()`/`bold()` calls run. This ensures no escape codes are appended to the buffer when `ansiLog=false` (e.g. when writing to a pipe in the integration tests).

ewels:
Ok I updated it so that there is now some slightly differing behaviour, but hopefully gives the functionality we want:

* `NXF_ANSI_LOG=0` now does not print the "Linting <filename>.." status updates
* `nextflow -q lint` is effectively identical, it disables ANSI logging at the top level so no colour
* I wanted to be able to have not-verbose-but-still-colours, so I added `nextflow lint -q` which _does_ have ANSI logging with colours, but does not have the "Linting <filename>" status updates.

We could simplify this further by simply removing the "Linting <filename>" updates entirely as they give fairly limited functionality (just saying that something is happening). Happy to do that if you prefer @bentsherman

bentsherman (approved):
Simplified so that global `-q` flag suppresses progress messages. ANSI logging can be controlled independently

mashehu:
will this only be in the edge release or also in a patch for 25.10? Asking for an action to be simplified: https://github.com/seqeralabs/nf-lint-pre-commit/

bentsherman:
I think this one is also out of scope for a backport

PR #6881, opened 2026-03-02
Typed workflows

bentsherman:
This PR implements *typed workflows*, which provide first-class support for static types and records.

When the `nextflow.preview.types` feature flag is enabled in a script, workflows are treated as typed workflows with the following semantics:

- **New operator library**: streamlined set of operators for channels and values, with first-class support for static typing and records

- **Restricted syntax**: several syntax patterns are no longer allowed:
  - Pipe (`|`) and And (`&`) dataflow operators
  - Assignment via `set` or `tap`
  - Accessing process/workflow outputs via `.out` property -- use assignments instead


bentsherman:
Summary of recent changes:
- Renamed "typed dataflow" to "typed workflows" for clarity
- Require `nextflow.preview.types` flag for type annotations in workflow takes/emits in addition to new dataflow semantics, to help distinguish typed workflows vs legacy workflows
- Move "process named arguments" to separate PR, to keep this PR focused on the most essential features

docs/reference/cli.md
```diff
@@ -1481,6 +1481,7 @@ The `run` command is used to execute a local pipeline script or remote pipeline
 
 `-dump-channels`
 : Dump channels for debugging purpose.
+: This option can specify one or more tags as a comma-separated list to only dump channels with a matching tag.
```
christopher-hakkaart:
```suggestion
: Optionally accepts a comma-separated list of tags to filter output to only channels with a matching tag.
```

docs/reference/syntax.md
```diff
@@ -196,6 +196,31 @@ In order for a script to be executable, it must either define an entry workflow
...
+    messages = greetings.map { v -> "$v world!" }
+
+    emit:
+    messages: Channel<String>
+}
+```
+
+- Each workflow input in the `take:` section has a name and a type.
+
+- Each named workflow output in the `emit:` section may specify a type.
```
christopher-hakkaart, line 224:
```suggestion
Key features of typed workflows include:

- Each workflow input in the `take:` section has a name and a type.

- Each named workflow output in the `emit:` section may specify a type.
```
Optional suggestion: makes it more digestible without reading the text above the code block.

docs/tutorials/records.md
```diff
@@ -1,280 +0,0 @@
-(migrating-records)=
-
-# Migrating to records
```
christopher-hakkaart, line 3: I'm just checking to see if this was meant to be deleted.

docs/tutorials/static-types.md
```diff
@@ -1,68 +1,55 @@
...
+Static typing is optional. All existing code will continue to work.
 :::
 
 ## Overview
 
-Static types allow you to specify the types of variables and parameters in Nextflow code. This language feature serves two purposes:
+Static typing allows you to precisely model and validate the structure of your data as it flows through your pipeline. It consists of several new language features:
 
-1. **Better documentation**: Type annotations serve as code documentation, making your code more precise and easier to understand.
+- **Type annotations** can be added to inputs and outputs at every level of a pipeline, from pipeline parameters to process inputs and outputs, using the {ref}`standard types <stdlib-types>` in the Nextflow standard library. These annotations make your code easier to understand, and they are used by the Nextflow language server to identify type-related errors during development.
```
christopher-hakkaart:
```suggestion
- **Type annotations** can be added to inputs and outputs at every level of a pipeline, i.e., from pipeline parameters to process inputs and outputs, using the {ref}`standard types <stdlib-types>` in the Nextflow standard library. These annotations make your code easier to understand and are used by the Nextflow language server to identify type-related errors during development.
```

docs/tutorials/static-types.md
```diff
@@ -1,68 +1,55 @@
...
-
-- File outputs (`file` and `path` qualifiers) are converted to `file()` or `files()` based on the `arity` option when specified. To ensure accurate conversion, specify the `arity` option in the legacy syntax and review the converted code to verify the correct output function is used.
-
 ## Example: rnaseq-nf
 
-This section demonstrates how to migrate a pipeline to static types using the [rnaseq-nf](https://github.com/nextflow-io/rnaseq-nf) pipeline as an example. The completed migration is available in the [preview-25-10](https://github.com/nextflow-io/rnaseq-nf/tree/preview-25-10) branch.
+This section demonstrates how to migrate a pipeline to static typing using the [rnaseq-nf](https://github.com/nextflow-io/rnaseq-nf) pipeline as an example. See {ref}`rnaseq-nf-page` for an introduction to the rnaseq-nf pipeline.
 
-See {ref}`rnaseq-nf-page` for an introduction to the rnaseq-nf pipeline.
+This tutorial will use the following approach to migrate the pipeline:
```
christopher-hakkaart:
```suggestion
This section walks through migrating a pipeline to static typing, using [rnaseq-nf](https://github.com/nextflow-io/rnaseq-nf) as an example. See {ref}`rnaseq-nf-page` for an introduction to the pipeline.

The approach is as follows:
```
Felt wordy; this is an attempt to streamline rather than saying 'this section' and 'this tutorial,' as they are the same thing.

docs/tutorials/static-types.md
```diff
@@ -121,153 +108,172 @@ params {
...
+read_pairs_ch = channel.fromFilePairs(params.reads, checkIfExists: true, flat: true)
 ```
 
-The `RNASEQ` workflow is called by the entry workflow with the following arguments:
+Each tuple has three elements -- the sample ID (inferred from the file names) and the two FASTQ files.
 
-```nextflow
-workflow {
-    read_pairs_ch = channel.fromFilePairs(params.reads, checkIfExists: true, flat: true)
+This approach will not work with static typing because `fromFilePairs()` does not have a well-defined return type. A more robust way to model a collection of samples is with a *samplesheet* -- a CSV file specifying samples as rows and sample fields as columns.
```
christopher-hakkaart:
```suggestion
This approach will not work with static typing because `fromFilePairs()` does not have a well-defined return type. A more robust way to model a collection of samples is with a *samplesheet*. For example, a CSV file specifying samples as rows and sample fields as columns.
```
Doesn't always have to be a CSV

docs/tutorials/static-types.md
```diff
@@ -121,153 +108,172 @@ params {
...
 
-    (fastqc_ch, quant_ch) = RNASEQ(read_pairs_ch, params.transcriptome)
+Create the following samplesheet to represent the test data:
```
christopher-hakkaart, line 137: The following steps could be numbered. E.g.:

1. Create ....
2. Refactor ....
3. ...

This is difficult to edit in GitHub, I'm happy to leave this for now but will consider a followup PR to add this after it's merged.

docs/tutorials/static-types.md
```diff
@@ -121,153 +108,172 @@ params {
...
-    // ...
+    script:
+    """
+    fastqc.sh "${id}" "${fastq_1} ${fastq_2}"
+    """
 }
 ```
 
-To migrate this process, rewrite the inputs and outputs as follows:
+To migrate this process, rewrite it as follows:
```
christopher-hakkaart:
```suggestion
To migrate the `FASTQ` process, rewrite it as follows:
```

docs/tutorials/static-types.md
```diff
@@ -276,24 +282,50 @@ process QUANT {
...
+        -i ${index} \
+        -1 ${fastq_1} \
+        -2 ${fastq_2} \
+        -o quant_${id}
+    """
 }
 ```
 
-To migrate this process, rewrite the inputs and outputs as follows:
+To migrate this process, rewrite it as follows:
```
christopher-hakkaart:
```suggestion
To migrate the `QUANT` process, rewrite it as follows:
```

docs/tutorials/static-types.md
```diff
@@ -341,38 +378,174 @@ In a typed process, file patterns for `path` inputs must be declared using a *st
...
 :::
 
 <h4>INDEX</h4>
 
 Apply the same migration principles from the previous processes to migrate `INDEX`.
 
-(preparing-static-types)=
+### Migrating workflows
+
+Once every process called by a workflow has been migrated to static typing, the workflow itself can be migrated.
```
christopher-hakkaart:
```suggestion
Once you migrate every process called by a workflow to static typing, you can migrate the workflow itself.
```

docs/tutorials/static-types.md
```diff
@@ -121,153 +108,172 @@ params {
...
+```nextflow
+read_pairs_ch = channel.of(params.reads)
+    .flatMap { csv -> csv.splitCsv() }
+    .map { row ->
+        record(id: row[0], fastq_1: file(row[1]), fastq_2: file(row[2]))
+    }
+```
 
-You can use the same approach for inputs to determine the type of each output:
+This code can be simplified further by modeling `params.reads` as a collection of records instead of a file path.
```
christopher-hakkaart:
```suggestion
You can simpliy the code further by modeling `params.reads` as a collection of records instead of a file path.
```

docs/tutorials/static-types.md
```diff
@@ -341,38 +378,174 @@ In a typed process, file patterns for `path` inputs must be declared using a *st
...
 
-(preparing-static-types)=
+### Migrating workflows
+
+Once every process called by a workflow has been migrated to static typing, the workflow itself can be migrated.
+
+See {ref}`workflow-typed-page` for an overview of typed workflows.
+
+:::{note}
+You will need to enable the `nextflow.preview.types` feature flag in your script to use typed workflows.
```
christopher-hakkaart:
```suggestion
You must enable the `nextflow.preview.types` feature flag in each script that uses typed workflows.
```

docs/tutorials/static-types.md
```diff
@@ -341,38 +378,174 @@ In a typed process, file patterns for `path` inputs must be declared using a *st
...
+    fastqc_ch = FASTQC(read_pairs_ch)
+    quant_ch = QUANT(index, read_pairs_ch)
+
+    emit:
+    fastqc = fastqc_ch
+    quant = quant_ch
+}
+```
+
+The type of each workflow input can be inferred by examining how the workflow is called. In this case, `RNASEQ` is called by the entry workflow with the following arguments:
```
christopher-hakkaart:
```suggestion
You can infer the type of each workflow input by examining how the workflow is called. In this case, `RNASEQ` is called by the entry workflow with the following arguments:
```

docs/tutorials/static-types.md
```diff
@@ -341,38 +378,174 @@ In a typed process, file patterns for `path` inputs must be declared using a *st
...
+:::{note}
+In this case, the records in `read_pairs_ch` are identical to the record inputs of `FASTQC` and `QUANT`. However, `read_pairs_ch` would still be compatible if it contained additional record fields, as long as it contains the fields required by the two processes.
+:::
+
+The `FASTQC` and `QUANT` processes produce the channels `fastqc_ch` and `quant_ch`, both of which have type `Channel<Record>`:
+
+- `fastqc_ch` contains records with the fields `id` and `fastqc`
+- `quant_ch` contains records with the fields `id` and `quant`
+
+This type information can be inferred from the respective process outputs as shown in the previous section.
```
christopher-hakkaart:
```suggestion
You can infer this type of information from the respective process outputs, as shown in the previous section.
```

docs/tutorials/static-types.md
```diff
@@ -465,58 +656,32 @@ workflow {
...
-
-Use alternatives that have well-defined return types. For example, use `flatMap` instead of `flatten`.
-
-:::{tip}
-The commonly-used operators listed under {ref}`dataflow-type-channel` have well-defined return types and are recommended for use with static types.
-:::
+### Avoid unsupported operators
 
-If you cannot avoid an ambiguously-typed function, you can use a *hard cast* to specify the expected type, provided you know what the type will be at runtime.
+Many {ref}`operators <operator-page>` are not supported in typed workflows. In many cases, an unsupported operator can be replaced by another operator that is supported and/or a standard library function.
```
christopher-hakkaart:
```suggestion
Many {ref}`operators <operator-page>` are not supported in typed workflows. However, an unsupported operator can be replaced by another operator that is supported and/or a standard library function in many cases.
```

docs/reference/operator-typed.md
```diff
@@ -0,0 +1,341 @@
...
+:language: console
+```
+
+## flatMap
+
+**`Channel<E> flatMap( transform: (E) -> Iterable<R> ) -> Channel<R>`**
+
+The `flatMap` operator applies a *mapping function* to each value from a source channel. The mapping function should return a collection, and each element in the collection is emmited separately.
+
+For example:
```
christopher-hakkaart:
```suggestion
The `flatMap` operator applies a *mapping function* to each value from a source channel. The mapping function should return a collection, and each element in the collection is emmited separately:
```

docs/tutorials/static-types-operators.md
```diff
@@ -0,0 +1,460 @@
...
+
+*This operator is non-deterministic.*
+
+Use `join` to combine channels in a way that doesn’t depend on the channel ordering.
+
+### multiMap
+
+Use `map` for each branch instead. Using records instead of tuples can eliminate much of the need for `branch`.
+
+For example:
```
christopher-hakkaart, line 320:
```suggestion
Use `map` for each branch instead. Using records instead of tuples can eliminate much of the need for `branch`:
```

docs/reference/operator-typed.md
```diff
@@ -0,0 +1,341 @@
...
+
+```{literalinclude} ../snippets/mix.nf
+:language: nextflow
+```
+
+```{literalinclude} ../snippets/mix.out
+:language: console
+```
+
+The values in the output channel may be emitted in any order, for example:
```
christopher-hakkaart, line 187:
```suggestion
The values in the output channel may be emitted in any order:
```

docs/reference/operator-typed.md
```diff
@@ -0,0 +1,341 @@
...
+
+**`Channel<E> reduce( seed: R, accumulator: (R, E) -> R ) -> Value<R>`**
+
+**`Channel<E> reduce( accumulator: (E, E) -> E ) -> Value<E>`**
+
+The `reduce` operator applies an *accumulator function* sequentially to each value in a source channel, and emits the final accumulated value.
+
+The accumulator function takes two parameters -- the accumulated value and the *i*-th emitted value -- and it should return the accumulated result, which is passed to the next invocation with the *i+1*-th value. This process is repeated for each value in the source channel.
+
+For example:
```
christopher-hakkaart, line 208:
```suggestion
The accumulator function takes two parameters -- the accumulated value and the *i*-th emitted value -- and it should return the accumulated result, which is passed to the next invocation with the *i+1*-th value. This process is repeated for each value in the source channel:
```

docs/workflow-typed.md
```diff
@@ -0,0 +1,198 @@
...
+- `summary_report`: a dataflow value containing a summary report (`Value<Path>` or `Path`)
+
+Type annotations are useful for documenting the structure of each workflow output, and they can be used by the language server to validate the type of each published output.
+
+Outputs that receive a dataflow value can be declared as `Value<V>` or `V` for short. In this example, the `summary_report` is declared with type `Path` as a shorthand for `Value<Path>`.
+
+## Named workflows
+
+:::{note}
+Typed workflows require the `nextflow.preview.types` feature flag to be enabled in each script that uses them.
```
christopher-hakkaart:
```suggestion
You must enable the `nextflow.preview.types` feature flag in each script that uses typed workflows.
```

docs/workflow-typed.md
```diff
@@ -0,0 +1,198 @@
...
+
+(workflow-typed-dataflow)=
+
+## Dataflow
+
+:::{versionadded} 26.04.0
+:::
+
+:::{note}
+Typed workflows require the `nextflow.preview.types` feature flag to be enabled in each script that uses them.
```
christopher-hakkaart:
```suggestion
You must enable the `nextflow.preview.types` feature flag in each script that uses typed workflows.
```

docs/tutorials/static-types.md
bentsherman: I always find a numbered list harder to read anyway when there are code blocks in between

docs/tutorials/records.md
bentsherman: Yes. I incorporated this tutorial into the static types tutorial, which now covers typed processes, typed workflows, and records. Since we expect most users will migrate to all three at the same time

christopher-hakkaart: Sorry - worked this out as I went through, but never deleted this comment.

docs/tutorials/static-types.md
christopher-hakkaart: Happy to leave it as is for now.

bentsherman:
Thanks for the thorough review. I believe I can address all of your concerns

> Many operators were introduced to simplify very specific, common use cases (e.g. `transpose`, `branch`, `splitCsv`). The ADR doesn't provide a clear rationale for why they should be removed rather than kept alongside the new core set.

Good point. I have added a column to the table of legacy operators to help explain why a given operator was excluded

It is true that these operators serve common use cases today, but the shift from tuples to records changes the calculus. To address your specific examples:

- `transpose` does not have a well-defined return type, and it's use case with tuples doesn't really translate to records

- `branch` and `multiMap` are used primarily to adapt tuple channels into processes. They are the main example of "operator boilerplate" that we aim to eliminate with records. The `typed-operators.md` tutorial provides several detailed examples of this

- `splitCsv` is not statically typed, and is easily replaced with `flatMap` + `Path::splitCsv`

We can always add operators in the future, but the proposal is based on my analysis of which operators are actually used in practice. Given a new operator, I would want to see a clear use case first, so that we don't end up with ~50 operators again in the future.

> Forcing users to rewrite `transpose` as `flatMap` compositions adds verbosity without clear benefit.

The benefit of `flatMap` is that it is more explicit. It is much easier to understand what is happening compared to `transpose`. But the most important difference is that `flatMap` is statically typed while `transpose` is not

> Changing `join` from tuple-index to record-field matching, or `cross` from key-matching to true cross-product, introduces silent behavior changes. These are not type errors — they produce wrong results quietly. This is the riskiest part of the proposal and the hardest to debug during migration.

Not true for `join`. The v1 `join` operates on tuples while the v2 `join` operates on records -- attempting to use the v2 `join` on tuples will produce both a type error and a runtime error.

As for `cross`, this is a fair concern since v2 `cross` has the same input/output types. I am not very concerned about this because `cross` is rarely used anyway -- there are only 18 uses of `cross` in the `nf-core` GH org, vs 100 uses of `combine` and 1,300 uses of `join`.

However, if it would assuage your concern, we could use a new name such as `crossFull` or `combine`. In fact, v2 `cross` is semantically equivalent to v1 `combine` (without `by` option) except for a slightly different output type (which would be caught by the type checker / runtime).

> A single flag controls three independent things: type annotations, operator semantics, and syntax restrictions. Users who want type checking also get operator behavior changes they didn't ask for. This coupling makes incremental adoption harder, not easier.

These changes are not independent -- all three are necessary to enable static typing. For users, there is little value in adopting any one of these changes independently. Their pipelines already work, so the only real reason to migrate is for the benefit of static typing.

The other issue is one that you originally raised. Allowing users to adopt static typing in pieces within the same file leads to many variations in code style (e.g. adopting type annotations but not typed operators, and vise versa). This can be confusing for users, even more so for agents. It goes against a very goal of this ADR to *reduce* variations.

The solution is to take the same approach for typed workflows as we did for typed processes, as described in the ADR. The feature flag, in conjunction with the new syntax for typed inputs/outputs, provides a clear distinction between legacy vs typed code.

> Preview flags tend to linger. There's no defined path to stable, which can slow community adoption and make the transition longer than necessary.

I intentionally left this decision out of the ADR so that we can decide on it separately. But my plan was to either:

1. if we feel confident about the syntax / semantics for typed processes and typed workflows, then we make it stable in 26.04
2. if we don't feel confident, then we keep it in preview for 26.04 and make it stable in 26.10

Either way, the `nextflow.preview.types` flag will be replaced by `nextflow.enable.types` once the feature is stable, and this flag will remain indefinitely as a way to opt into typed processes/workflows.

I would like to merge typed workflows and share a preview build with the beta testing group so that we can make an informed decision on this in April.

> **Axis 1: Syntax parser version** — use an env flag `NXF_SYNTAX_PARSER=v3` for the transitional period

This approach is not viable:
- The intent of `NXF_SYNTAX_PARSER` is to provide a new implementation ("parser version") of the same syntax, not to add new language features.
- `NXF_SYNTAX_PARSER` is applied globally, whereas typed processes/workflows need to be enabled per-file (`nextflow.enable.types`)

> **Axis 2: Operator opt-in via explicit include**

I considered this as well, but ultimately I see too many problems:

- The use of `channel.*` doesn't make sense because we are talking about operators, not channel factories.
- Users will still need to have `nextflow.enable.types = true` in every script, so the `include` just adds noise.
- No one is going to adopt the new syntax without also adopting the new operators
- See my previous point about creating many code variations. The best we can hope for is to have only two code styles -- legacy and typed

bentsherman:
Summary of changes since last review:

- **Added "Reason for removal" column to legacy operators.** Should help explain why each operator is no longer needed. Happy to dig into each operator as you like.

  While it is true that they could be carried over temporarily and just not support type checking, I think it will be better in the long term to cut them out now. Keeping them around even as "deprecated" can still be confusing for users and agents when reading typed code and deciding which operator to use.

  I think a smooth transition is already provided by the fact that workflows can be migrated one-by-one. If you have a problematic bit of workflow logic, you can simply move it into a separate workflow for now and migrate it later.

- **Replace `cross` with `combine` in typed operators.** This removes the "silent semantic change" with `cross`. Instead, the new `combine` is equivalent to the old `combine` (without the `by` option). This also simplifies the overall migration to typed workflows since `combine` is used much more frequently than `cross`.

See also:
- [Use of `combine` in nf-core](https://github.com/search?q=org%3Anf-core+path%3A*.nf+%22.combine%28%22&type=code)
- [Use of `cross` in nf-core](https://github.com/search?q=org%3Anf-core%20path%3A*.nf%20%22.cross%28%22&type=code)

docs/reference/operator-typed.md
```diff
@@ -0,0 +1,343 @@
...
+
+```{literalinclude} ../snippets/filter-closure.out
+:language: console
+```
+
+## flatMap
+
+**`Channel<E> flatMap( transform: (E) -> Iterable<R> ) -> Channel<R>`**
+
+The `flatMap` operator applies a *mapping function* to each value from a source channel. The mapping function should return a collection, and each element in the collection is emmited separately.
```
christopher-hakkaart:
```suggestion
The `flatMap` operator applies a *mapping function* to each value from a source channel. The mapping function should return a collection. Each element in the collection is emitted separately.
```

pditommaso:
Thanks for the detailed explanation and for addressing the review feedback so thoroughly. I fully agree that records are the future, and I appreciate the work that has gone into this. That said, any hard break is not viable for our users. We should consider how to make this a more incremental update without breaking backward compatibility. Worth continuing the discussion on this.

bentsherman:
To be clear -- this PR does preserve backwards compatibility. Legacy workflows still support all existing operators, and you can migrate one workflow at a time without having to migrate the entire pipeline. I think that's very reasonable for a new opt-in feature

Working with the beta testing group, and my own experiments with nf-core/methylseq, have made it clear that there is little value or user appetite for migrating to static typing without also migrating to records. Basically everyone is going to adopt both in one pass

So for operators that are designed for tuples and have no use case with records, there is no point in bringing them into typed workflows

Also, bringing over all ~50 operators undermines a core goal of typed workflows -- to eliminate redundant code patterns. The same logic for removing `|` and `&` apply to operators -- when there are many ways to do the same thing, it increases the risk of confusing users and agents (similar to MCP overload).

Typed workflows allow us to start fresh with a streamlined set of operators, while allowing users to migrate at their own pace.

bentsherman:
Summary of latest changes:

- **Support all existing operators in typed workflows**: Remove v1/v2 operator distinction. Static typing now only "updates" the existing operator library rather than introducing a "new" library.

  Best practices guide separates "core operators" from "legacy operators" only as a recommendation -- legacy operators can continue to be used in typed workflows.

- **Replace preview flag with stable flag**: Typed processes/workflows are now stable and can be enabled with `nextflow.enable.types`. We can continue to add new functionality in the future, but the current feature set should not need to be changed further.

pditommaso:
Great work on this iteration — it's a notable improvement over the previous version. The dual-path V1/V2 architecture is clean, the `DataflowTypeHelper` normalization bridge is elegant, and the decision to keep all legacy operators available (while promoting the 13 core ones) is pragmatic. The `combine` rename over `cross` also makes much more sense.

A few things to flag:

## Potential Issues

**Thread safety in `CombineOpV2` and `JoinOpV2`** — Both operators iterate over shared collections (`leftValues`/`rightValues`) in `onNext` while another subscription thread may be adding to them. The methods are `synchronized`, so this should be safe as long as both subscriptions synchronize on the same instance — but worth double-checking that GPars doesn't dispatch to different threads concurrently.

**`ValueImpl.flatMap()` missing tuple validation** — `ChannelImpl.flatMap()` explicitly rejects `Tuple` return values with a clear error message. `ValueImpl.flatMap()` doesn't have this check, which seems like a copy-paste oversight.

**`NextflowMeta.Features.types` not connected** — The `Features` class has a `boolean types` field but it's not used for any runtime decisions. The actual enablement goes through `BaseScript.enableTyping()`. This could cause confusion if users inspect the meta object.

## Test Coverage Gaps

- No tests for `ChannelNamespace` factory methods beyond `of()` and `fromList()` — `empty()`, `fromPath()`, `fromLineage()`, `interval()`, `topic()`, `watchPath()` are untested
- No type mismatch detection tests (e.g., passing wrong types to a typed process)
- No tests for legacy operator fallback in typed workflows — important given the decision to support all existing operators (commit 892cdd260)
- `reduce()` with seed variant is untested
- No concurrent access tests for any of the stateful operators

## Typos

1. `adr/20260310-typed-workflows.md:204` — "THese" → "These"
2. `adr/20260310-typed-workflows.md:346` — "Procsses" → "Processes"
3. `adr/20260310-typed-workflows.md:470` — "opartor" → "operator"
4. `docs/workflow-typed.md:55` — "resuling" → "resulting"
5. `docs/reference/operator.md:6` — "recommennded" → "recommended"

## ADR: Identical Examples (lines 90-115)

The "typed workflow" and "legacy workflow" examples are identical — both show `ch_names: Channel<String>`. The legacy example should drop the type annotation:

```groovy
// legacy workflow
workflow HELLO {
    take:
    ch_names

    main:
    ch_names.subscribe { name ->
        println "Hello, $name!"
    }
}
```

Otherwise the two snippets are indistinguishable and the comparison doesn't illustrate anything.

## Feature Maturity

Given the scale of this change (150 files, ~6,500 additions), I think this should still be considered **preview** in the next stable release rather than promoted to `enable`. The `preview` → `enable` promotion can happen once the feature has had a full release cycle of user feedback. It would be good to display a warning message when `nextflow.enable.types` is used, something along the lines of "Typed workflows are an early feature — behavior may change in future releases."

## Removal of `.out`

The removal of the `.out` accessor in typed workflows is a hard loss. Beyond the ergonomic cost, we were planning to extend it to support `.err` for process error handling. Dropping `.out` now would close that door. Worth reconsidering whether `.out` (and future `.err`) can coexist with the typed model.

bentsherman:
Addressing minor comments:

> **Thread safety in `CombineOpV2` and `JoinOpV2`**

Indeed, these operators are designed such that `onNext` is always called sequentially for both operands, so there is no race condition. GPars invokes `onNext` sequentially by default unless you set `maxForks`

> No tests for `ChannelNamespace` factory methods

Since `ChannelNamespace` delegates most of its methods to `Channel.groovy`, I didn't see a need to duplicate the unit tests. I added a unit test for `fromPath` just as a sanity check.

> No type mismatch detection tests (e.g., passing wrong types to a typed process)

This will be covered by the language server for the time being.

> No tests for legacy operator fallback in typed workflows

Legacy fallback is tested by `ChannelImplTest` and `ValueImplTest`. I updated them anyway to call legacy operators in a typed workflow.

> No concurrent access tests for any of the stateful operators

This is also covered by `ChannelImplTest`, which tests `combine`, `groupBy`, and `join`.

> Given the scale of this change (150 files, ~6,500 additions), I think this should still be considered **preview** in the next stable release rather than promoted to `enable`.

Fair enough. I was inclined to do the same anyway

bentsherman:
> The removal of the `.out` accessor in typed workflows is a hard loss. Worth reconsidering whether `.out` (and future `.err`) can coexist with the typed model.

I originally set out to support `.out` since it is widely used. But the problem is that `.out` has an ambiguous type

- When a process/workflow emits a single output, the `.out` can refer to the ChannelOut or the single output in the ChannelOut. The type checker cannot handle both cases at the same time.

- The interoperability between typed/legacy workflows works by normalizing dataflow types around call sites (process/workflow calls). The `.out` breaks this assumption by allowing you to access call outputs by other means which can't be normalized

I also agree with your previous review :smile:  :

> Removing pipe syntax, `.out`, set/tap — painful but justified. Standardization matters, especially for AI agent readability. These syntax variants add cost without proportional value.

Especially if we keep static typing in preview for another release cycle, we can continue to look into supporting it for the stable version. But we won't be able to support it for this release

In any case, the migration is a simple search-and-replace, should be easy for humans and agents:

```groovy
// before
MY_WORKFLOW()
MY_WORKFLOW.out.foo.view()
MY_WORKFLOW.out.bar.view()

// after
my_out = MY_WORKFLOW()
my_out.foo.view()
my_out.bar.view()
```

> We were planning to extend it to support `.err` for process error handling. Dropping `.out` now would close that door.

There are other good syntax options for this, so I'm not worried about it

pditommaso:
Thanks for the detailed explanation on why `.out` is problematic with the type checker — the ambiguity between `ChannelOut` and the single output makes sense. And I agree, removing `.out` does help clean up the syntax and make it more consistent.

My main concern isn't `.out` itself but losing the ability to extend the process call syntax with things like `.err` for error handling. That's a pattern we've been considering, and dropping `.out` could close that design space.

That said, it could be explored differently — for example, Go handles errors by returning them as values alongside the result:

```go
result, err := doSomething()
if err != nil {
    // handle error
}
```

A similar pattern in typed Nextflow could look like:

```groovy
result, err = MY_PROCESS(ch_input)
err.subscribe { log.warn("Process failed: $it") }
result.view()
```

This would actually fit naturally with the typed model — errors as explicit channels in the return value rather than an accessor on the call site.

In any case, as long as typed workflows ship as preview in the next stable release, I'm happy to move forward and re-evaluate in the next iteration whether to keep or revisit this.

pditommaso (approved):

bentsherman:
Thanks Paolo. And yes, that's essentially the syntax I was thinking too:

```groovy
(result, err) = MY_PROCESS(ch_input)
err.subscribe { log.warn("Process failed: $it") }
result.view()
```

This syntax already exists in Nextflow/Groovy, so it's familiar and works out of the box. It also forces the user to "confront" the error case when a process declares it. If I don't do anything with `err`, Nextflow will warn about it as an unused variable

PR #6891, opened 2026-03-04
Fix TraceFileObserver NPE when trace file already exists

### LINKED ISSUE #6890: TraceFileObserver NPE on every task when trace file already exists

**Expected:** When a trace file already exists (e.g. on S3) and `trace.overwrite` is not enabled, Nextflow should log a clear warning that the trace file could not be created and continue the pipeline run gracefully.

**Actual:** `TraceFileObserver.onFlowCreate()` throws `AbortOperationException` (via `TraceHelper.newFileWriter()`), which `Session.notifyEvent()` catches and logs at debug level only. Because the exception occurs before the `writer` Agent is initialized, `writer` remains `null`. Every subsequent call to `onTaskComplete()`, `onTaskCached()`, and `onFlowComplete()` then throws a `NullPointerException`, producing a stack trace for every task in the run:

```
Cannot invoke "groovyx.gpars.agent.Agent.send(Object)" because "this.writer" is null
java.lang.NullPointerException: Cannot invoke "groovyx.gpars.agent.Agent.send(Object)" because "this.writer" is null
    at nextflow.trace.TraceFileObserver.onTaskComplete(TraceFileObserver.groovy:219)
```

1. Run a Nextflow pipeline with `trace.enabled = true` and `trace.file` pointing to S3 (or any path where the file already exists).
2. Do **not** set `trace.overwrite = true`.
3. Run the pipeline a second time without removing the trace file.

```groovy
// nextflow.config
trace {
    enabled = true
    file = 's3://my-bucket/results/trace.tsv'
    // overwrite = true  // intentionally omitted
}
```

```groovy
// main.nf
process FOO {
    input: val x
    output: stdout
    script: "echo $x"
}

workflow {
    Channel.of(1, 2, 3) | FOO
}
```

Run once to create the trace file, then run again to trigger the bug.

### Program output

On the second run, the `.nextflow.log` contains an NPE for every task:

```
Cannot invoke "groovyx.gpars.agent.Agent.send(Object)" because "this.writer" is null
java.lang.NullPointerException: Cannot invoke "groovyx.gpars.agent.Agent.send(Object)" because "this.writer" is null
    at nextflow.trace.TraceFileObserver.onTaskComplete(TraceFileObserver.groovy:219)
    ...
Cannot invoke "groovyx.gpars.agent.Agent.send(Object)" because "this.writer" is null
java.lang.NullPointerException: Cannot invoke "groovyx.gpars.agent.Agent.send(Object)" because "this.writer" is null
    at nextflow.trace.TraceFileObserver.onFlowComplete(TraceFileObserver.groovy:189)
    ...
```

The original `AbortOperationException` with the helpful message ("Trace file already exists ... enable the 'trace.overwrite' option") is only logged at debug level by `Session.notifyEvent()`, so users never see it with default log settings.

### Environment

* Nextflow version: 25.10.2 (also present on current master)
* Java version: 21
* Operating system: Any (observed on Linux with S3 trace path)

### Additional context

**Root cause:** In [`TraceFileObserver.onFlowCreate()`](https://github.com/nextflow-io/nextflow/blob/master/modules/nextflow/src/main/groovy/nextflow/trace/TraceFileObserver.groovy#L174), `TraceHelper.newFileWriter()` throws when the file exists. The exception is caught by `Session.notifyEvent()` and logged at debug level. But `writer` (an `Agent<PrintWriter>`) was never assigned, so every subsequent method that calls `writer.send { ... }` or `writer.await()` throws an NPE.

The four affected call sites in `TraceFileObserver`:
- `onFlowCreate` line 178: `writer.send { ... }` (header write)
- `onFlowComplete` line 189: `writer.await()`
- `onTaskComplete` line 219: `writer.send { ... }`
- `onTaskCached` line 233: `writer.send { ... }`

**Fix:** Catch the exception in `onFlowCreate` and log at warn level so the user sees it, return early, and add null-safe (`?.`) guards on `writer` at the remaining call sites so the observer degrades gracefully.

robsyme:
## Summary

- Fix `NullPointerException` spam in `TraceFileObserver` when trace file already exists and `overwrite` is not enabled
- Add visible `log.warn` so users see the actionable "enable trace.overwrite" message instead of silent failure
- Add test covering the graceful degradation path

Closes #6890

## Problem

When `TraceHelper.newFileWriter()` throws in `onFlowCreate()` (e.g. trace file already exists on S3), the `writer` Agent is never initialized. Every subsequent call to `onTaskComplete()`, `onTaskCached()`, and `onFlowComplete()` then throws:

```
Cannot invoke "groovyx.gpars.agent.Agent.send(Object)" because "this.writer" is null
java.lang.NullPointerException
    at nextflow.trace.TraceFileObserver.onTaskComplete(TraceFileObserver.groovy:219)
```

The original `AbortOperationException` with the helpful user-facing message is swallowed at `log.debug` by `Session.notifyEvent()`, so users never see why their trace file wasn't written.

## Changes

**`TraceFileObserver.groovy`:**
- Wrap file creation in `onFlowCreate()` with try-catch; log at `warn` level (user-visible) and `debug` level (full stack trace), then return early
- Add `?.` null-safe guards on `writer` in `onFlowComplete()`, `onTaskComplete()`, `onTaskCached()`
- Guard `traceFile` access in `onFlowComplete()` with `if( traceFile )`

**`TraceFileObserverTest.groovy`:**
- Add `'should degrade gracefully when trace file already exists'` — verifies that when file creation fails, all task events and flow completion proceed without exceptions, and the original file is untouched

### DIFF
--- a/adr/20251114-module-system.md
+++ b/adr/20251114-module-system.md
@@ -1,0 +1,897 @@
+# Module System for Nextflow
+
+- Authors: Paolo Di Tommaso
+- Status: approved
+- Date: 2025-01-06
+- Tags: modules, dsl, registry, versioning, architecture
+- Version: 2.7
+
+## Updates
+
+### Version 2.7 (2026-03-09)
+- **Renamed `.checksum` to `.module-info`**: Leaves room for additional properties in the future
+- **Removed `@` prefix from module scopes**: Local modules are distinguished from remote modules by presence/absence of `./` prefix
+- **Removed version pinning from config**: Installed module versions are now inferred from the `meta.yml` of each module in the `modules/` directory instead of being declared in `nextflow.config`
+
+### Version 2.6 (2026-01-28)
+- **Removed module parameters**: Module parameters specification moved to separate spec document.
+
+### Version 2.5 (2026-01-23)
+- **Module parameters**: Replaced structured tool arguments with general module parameters defined in `meta.yml`
+- **Simplified tools section**: Removed `args` property from tools; tool arguments now configured via module parameters
+- **Simplified `requires` block**: Removed `plugins`, `modules`, and `subworkflows` sub-properties; `requires` now only contains `nextflow` version constraint
+- **Process modules focus**: Removed sub-workflow references; spec is now focused on process modules only
+
+### Version 2.4 (2026-01-15)
+- **Removed transitive dependency resolution**: Module dependencies are explicit only; no automatic transitive resolution
+- **Removed `freeze` command**: No longer needed without transitive dependency management
+- **Simplified model**: Each module explicitly declares its dependencies in `nextflow.config`
+
+### Version 2.3 (2026-01-15)
+- **Resolution Rules table**: Added clear table specifying behavior for each combination of local state and declared version
+- **Local modification protection**: Locally modified modules (checksum mismatch) are NOT overridden unless `-force` flag is used
+- **Simplified storage model**: Single version per module locally (`modules/@scope/name/` without version in path)
+- **`.checksum` file**: Registry checksum cached locally for fast integrity verification without network calls
+
+### Version 2.2 (2025-01-06) — *Superseded by v2.5*
+- **Structured tool arguments**: Added `args` property to `tools` section for type-safe argument configuration
+- **New implicit variables**: `tools.<toolname>.args.<argname>` returns formatted flag+value; `tools.<toolname>.args` returns all args concatenated
+- **Deprecation**: All `ext.*` custom directives (e.g., `ext.args`, `ext.args2`, `ext.args3`, `ext.prefix`, `ext.suffix`) deprecated in favor of structured tool arguments
+- *Note: Tool arguments replaced by module parameters in v2.5*
+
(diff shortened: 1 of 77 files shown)


modules/nextflow/src/main/groovy/nextflow/trace/TraceFileObserver.groovy
```diff
@@ -171,7 +171,14 @@ class TraceFileObserver implements TraceObserverV2 {
             Files.createDirectories(parent)
 
         // create a new trace file
-        traceFile = new PrintWriter(TraceHelper.newFileWriter(tracePath,overwrite, 'Trace'))
+        try {
+            traceFile = new PrintWriter(TraceHelper.newFileWriter(tracePath, overwrite, 'Trace'))
+        }
+        catch (Exception e) {
+            log.warn "Failed to create trace file: ${tracePath.toUriString()} -- ${e.message}"
+            log.debug "Trace file creation error details", e
```
pditommaso:
```suggestion
```

Warn should be enough

PR #6897, opened 2026-03-06
Add multi-arch support to arch process directive

pditommaso:
## Summary
- Refactor `Architecture` class to support comma-separated platform values in the `arch` directive (e.g. `arch 'linux/amd64,linux/arm64'`)
- Add inner `Arch` value type holding normalized arch values, with `platforms()` and `containerPlatform()` methods
- Keep `dockerArch`/`spackArch` getters for backward compatibility (delegate to first arch)
- Remove unused `platform`, `arch`, `target` fields from `Architecture`
- Update Wave to add fusion layers for each platform in multi-arch builds
- Update AWS Batch handler to use `dockerArch` instead of removed `arch` field

## Test plan
- [x] `ArchitectureTest` — single arch, multi-arch, map-based, normalize, toString
- [ ] Wave client tests with multi-arch platform
- [ ] Integration test with `arch 'linux/amd64,linux/arm64'`


modules/nextflow/src/main/groovy/nextflow/processor/Architecture.groovy
```diff
@@ -18,90 +18,96 @@ package nextflow.processor
...
-    static protected String getArch( String value ) {
-        // return value.minus(~'.*/') // keeping for reference
-        def chunks = value.tokenize('/')
-        if( chunks.size() == 3 )
-            return chunks[1] + '/' + chunks[2]
-        else if( chunks.size() == 2 )
-            return chunks[1]
-        else
-            return chunks[0]
+    private final List<Arch> archs
```
bentsherman:
```suggestion
    private final List<ArchEntry> entries
```

minor naming suggestions

PR #6902, opened 2026-03-09
Generate per-container SAS tokens for Azure AD/Managed Identity auth

### LINKED ISSUE #5669: Not authenticated to use blobs outside of Azure blob container working directory when using Azure Entra
Related to #5448 and #5444 but both issues refer to using Fusion, this one refers to using azcopy.

They are likely to be solved by the same method, since they have the same underlying challenge: how to pass authentication to the worker node (Batch) from Nextflow.

I seem to be able to recreate the issue _without_ Fusion. 

```console
> nextflow run seqeralabs/nf-canary -r main --remoteFile az://igenomes/atacseq_samplesheet_custom.csv --run TEST_STAGE_REMOTE -w az://scidev-useast -c azure.config
N E X T F L O W  ~  version 24.10.3
NOTE: Your local project version looks outdated - a different revision is available in the remote repository [c818260035]
Launching `https://github.com/seqeralabs/nf-canary` [magical_noyce] DSL2 - revision: 2ad4214f51 [main]
Uploading local `bin` scripts folder to az://scidev-useast/tmp/cf/bcc6a54f6a9dd33780a5251d956439/bin
[69/6f65a5] Submitted process > NF_CANARY:TEST_STAGE_REMOTE (1)
ERROR ~ Error executing process > 'NF_CANARY:TEST_STAGE_REMOTE (1)'

Caused by:
  Process `NF_CANARY:TEST_STAGE_REMOTE (1)` terminated with an error exit status (1)


Command executed:

  cat atacseq_samplesheet_custom.csv

Command exit status:
  1

Command output:
  (empty)

Work dir:
  az://scidev-useast/69/6f65a5549f7a3b2357312b12a28996

Container:
  docker.io/library/ubuntu:23.10

Tip: you can try to figure out what's wrong by changing to the process work dir and showing the script file named `.command.sh`

Execution cancelled -- Finishing pending tasks before exit
```

azure.config:

```config
process.executor = 'azurebatch'

fusion {
    enabled = false
}

azure {

    storage {
        accountName = 'seqeralabs'
    }

    batch {
        location = 'eastus'
        accountName = 'seqeralabs'
        copyToolInstallMode = 'node'
        autoPoolMode = true
        allowPoolCreation = true
        deletePoolsOnCompletion = false
    }

    activeDirectory {
        servicePrincipalId = 'redacted'
        servicePrincipalSecret = 'redacted'
        tenantId = 'redacted'
    }
}
```

And with an access key:

```console
```

To reiterate what's been said above, the error appears to stem from [generateContainerSasWithActiveDirectory](https://github.com/nextflow-io/nextflow/blob/87df7cd4be4828ac1d70f3c79c0c8db583d97ad1/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzHelper.groovy#L111-L115), which is only generating a relevant key for the working container and nothing else. Generating an account level SAS seems tricky (according to @alberto-miranda).

_Originally posted by @adamrtalbot in https://github.com/nextflow-io/nextflow/issues/5444#issuecomment-2590156438_

adamrtalbot:
## Summary

Fixes #5669

When using Azure Entra ID (Active Directory) or Managed Identity authentication with Azure Batch, Nextflow previously only generated a SAS token for the working directory container. Pipelines referencing blobs in other containers (e.g., \`az://igenomes/...\` while workdir is \`az://scidev-useast/...\`) failed with authentication errors.

This PR generates per-container SAS tokens for all containers accessed during a pipeline run.

### Core infrastructure (`nf-azure` plugin)

- **`AzFileSystemProvider`**: Owns the per-container SAS token cache (`ConcurrentHashMap<String,String>`) as runtime state; `getSasToken(containerName)` returns a container-specific token or falls back to the global `sasToken`; `setSasToken()` and `getContainerSasTokens()` allow callers to register and enumerate tokens; `generateAndRegisterContainerSas()` is called during `newFileSystem0()` for AD/MI auth, lazily generating SAS for each new container accessed
- **`AzStorageOpts`**: Config-only — per-container SAS state removed; `sasToken` remains as the global/user-supplied token
- **`AzHelper`**: Added `generateUserDelegationKey()` and a new `generateContainerSasWithActiveDirectory(BlobContainerClient, Duration, UserDelegationKey)` overload
- **`AzBatchExecutor`**: Eagerly generates SAS for the workDir container at startup; resolves `AzFileSystemProvider` from the workDir `AzPath` to register the token
- **`AzFileSystem`**: `copy()` uses per-container SAS lookup via the provider
- **`AzFileCopyStrategy`**: Holds a reference to `AzFileSystemProvider` (resolved from `bean.workDir` in the constructor); constructor pre-generates SAS for all input file containers; `getSasForPath()` looks up and lazily generates tokens via the provider; `getEnvScript()` exports per-container `AZ_SAS_<CONTAINER>` env vars from the provider's token map
- **`AzBatchService`**: `getSasForPath()` resolves the provider from the `AzPath` argument for per-container lookup
- **`AzFusionEnv`**: `getOrCreateSasToken()` and `getEnvironment()` resolve the provider from `Global.session.workDir`; exports `AZURE_STORAGE_SAS_TOKEN_<CONTAINER>` per container
- **`AzPathFactory`**: `getUploadCmd()` resolves the provider from the target `AzPath` for per-container SAS lookup

### Bash library (`AzBashLib`)

- `nxf_az_upload` and `nxf_az_download` now embed SAS tokens in URLs at script-generation time and split the URL at `?` using bash parameter expansion (`${var%%\?*}` / `${var#*\?}`) to correctly insert path components before the query string
- Removed `nxf_az_sas()` function (no longer needed)

## Backward Compatibility

- Global `sasToken` (account key auth) is unchanged — `getSasToken(containerName)` falls back to it when no per-container token exists
- Per-container logic only activates when `sasToken` is null (i.e., AD/MI auth)

## Testing

- Unit tests updated: `AzBashLibTest`, `AzFileCopyStrategyTest`, `AzStorageOptsTest`
- End-to-end tested with `nf-canary` across 3 separate Azure Blob containers:
  - `scidev-useast` (workdir)
  - `igenomes` (foreign input via `--remoteFile`)
  - `outputs` (publishDir)
  - All 14 tasks succeeded including cross-container staging, publish, folder round-trips, and bin scripts


plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzStorageOpts.groovy
```diff
@@ -60,6 +62,12 @@ class AzStorageOpts implements ConfigScope {
     @PlaceholderName("<name>")
     final Map<String,AzFileShareOpts> fileShares
 
+    /**
+     * Per-container SAS tokens, keyed by container name.
+     * Used when accessing multiple blob containers with AD/MI authentication.
+     * ConcurrentHashMap ensures safe concurrent reads and writes from multiple task threads.
+     */
+    private final Map<String,String> containerSasTokens = new ConcurrentHashMap<>()
```
adamrtalbot: This would be better as some object rather than `Map<String, String>` but I think this is sufficient to demonstrate the fix.

Skitionek:
@adamrtalbot I am a bit lost why managed identity approach uses SAS token. 

https://docs.azure.cn/en-us/entra/identity/managed-identities-azure-resources/tutorial-linux-managed-identities-vm-access?pivots=identity-linux-mi-vm-access-storage
Related azcopy docs:
https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-authorize-managed-identity?tabs=linux 

Could you help me understand the motivation of not using managed identity credentials directly?

Skitionek:
> To use Managed Identities in tasks you must:
> 1. Add a user assigned managed identity to the pool
> 2. Use Fusion
> 3. Enable the following setting:

So we agree that it is possible to access storage from tasks under condition that managed identity is added to the pool.

> The Managed Identity allows _Nextflow_ to authenticate to Azure. The _tasks_ are unaware they are in a Nextflow pipeline and can't inherit this authentication. Therefore, we generate short access tokens so Nextflow can pass authentication to each individual task.

Based on above it would be sufficient, if tasks could be given indication to use managed identity instead of SAS token (under a.m. condition).

It could be also considered as a fallback method if SAS token is not provided. 

Would love to hear your input about it, however we might be under a risk of going off topic. I believe this PR is right on conceptual level and it patches an issue currently in focus. I will try to find some time to test it as well but cannot make promise on time frame right now.

PR #6903, opened 2026-03-09
Fix cached task after abort

### LINKED ISSUE #6884: Cache fail when specific conditions are met
## Bug report 

The scenario: A pipeline step succeeds (but on a retry attempt), then the pipeline continues and fails on a later step, but then on pipeline resume, the on-retry-succeeded step is not picked up, and that step runs again. This isn’t always the case; it only happens when that step had attempted to re-run in a previous execution, at least once (i.e., having a relevant work folder for that module/step created). Relevant blog post [here](https://community.seqera.io/t/steps-for-reproducible-example-for-broken-nextflow-cache-after-successful-retries/2600)

Expected behaviour: The on-retry-succeeded step is properly cached, and pipeline executes resumes from where it left
Actual behaviour: Even though it completed successfully on a retry, it won't be cached, and re-runs.

1. clone proteinfamilies
2. alter conf/test.config here to make it fail on the 1st attempt, but pass on the 2nd:
```
withName: 'NFCORE_PROTEINFAMILIES:PROTEINFAMILIES:FAA_SEQFU_SEQKIT:SEQKIT_SEQ' {
       memory = { task.attempt == 1 ? 1.MB : 2.GB } // memory = { 1.4.GB * task.attempt }
       time   = { 2.m    * task.attempt }
   }
 ```      
3. Your executor has to be able to stop a process when it requires excess memory than what was defined (or use docker), so I used slurm with this slurm.config:
```
profiles {
   slurm {
       executor {
           name              = "slurm"
           queueSize         = 100
           queueGlobalStatus = true

       }
       workDir = "/path/to/work_proteinfamilies_cache_test/"
       process {
           queue  = 'standard'
           cache  = 'lenient'
       }
       params {
           // Boilerplate options
           outdir = "${launchDir}/results"
       }
   }
}
```
4. Run with command:
`NXF_VER=25.10.4 nextflow run proteinfamilies -c conf/slurm.config -profile singularity,test,slurm -resume`
5. As soon as the work folder for `SEQKIT_SEQ` is created, kill the pipeline (ctrl +C) (this step is improtant, otherwise if it passes on the 2nd attempt, without pre-failing at least once, the cache will work properly)
6. Resume the pipeline with the same command, and wait for `SEQKIT_SEQ` to complete successfully on its 2nd attempt.
7. Kill the pipeline again.
8. Resume the pipeline with the same command. Verdict: `SEQKIT_SEQ` will begin to run again, while `SEQFU_STATS_BEFORE`, which has run in parallel and completed on its first attempt, will have been cached properly.

### Program output 

Irrelevant

### Environment 

* Nextflow version: NXF_VER=25.10.4
* Java version:
```
openjdk 21.0.3 2024-04-16 LTS
OpenJDK Runtime Environment Temurin-21.0.3+9 (build 21.0.3+9-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.3+9 (build 21.0.3+9-LTS, mixed mode, sharing)
```
* Operating system: Linux, slurm
* Bash version: `GNU bash, version 5.1.8(1)-release (x86_64-redhat-linux-gnu)`

### Additional context

Dummy bot PR with attempt to pinpoint the bug in the codebase [here](https://github.com/nextflow-io/nextflow/pull/6882)

### PR DESCRIPTION
**Pull request by jorgee**
Close #6884
Alternative for #6882
This pull request enhances task retry logic in Nextflow by improving how task failures and aborts are tracked and incorporated into cache key calculation. The changes ensure that both failure and abort counts are considered, making task resumption and retry behavior more robust and predictable.

**Task retry and cache logic improvements:**

* The hash used for determining task cache keys now includes both `failCount` and a new `abortedCount` to ensure that retries after aborts or failures use a distinct hash.
* When a cached entry is found with a status of FAILED or ABORTED, the corresponding counters (`failCount` or `abortedCount`) on the `TaskRun` object are incremented, ensuring accurate retry attempts and cache key updates.

**Task state tracking enhancements:**

* Added a new `abortedCount` field to the `TaskRun` class to track the number of times a task execution has been aborted.
* Introduced `isAborted()` and `isFailed()` helper methods to the `TraceRecord` class for clearer and more maintainable status checks.

TODO:

- [x] Check attempts
- [x] Add test to reproduce failure and check fix

### CODE REVIEW

File: modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy [hunk 1]
```diff
@@ -837,7 +842,12 @@ class TaskProcessor {
             finally {
                 lock.release()
             }
-
+            // Issue #6884: https://github.com/nextflow-io/nextflow/issues/6884
+            // When cached tasks include failures and aborts and not cached. The task.attempt could not be sync with cached failures. We update the task.attempts and task.scripts
+            if( task.failCount > 0 && task.config.getAttempt() != task.failCount + 1 ) {
+                task.config.attempt = task.failCount + 1
+                task.resolve(taskBody)
+            }
```
jorgee: This part is not currently required to fix the case of the issue. 
However, when a task has been executed with previous failures but has not been completed, it is not cached, and it is re-executed. This execution is done with `task.attempt = 1`, with this code it is re-exceuted with `task.attempt = failedCount +1`. 

This case is happening in the added test. A process is defined to fail in the first attempt and succeed for the rest. So, it should execute twice in total. In the test, the execution is aborted after the first retry. Without this code, the task will be reexecuted again twice (first fails and second succeeds). With this code, the previous failed task will be counted as an attempt and then the task runs only once.

 I am not sure if reexecuting with attempt =1 was intended or if it should be managed as in this code and update the attempts according to cached failures. @bentsherman @pditommaso what's your opinion about it?

vagkaratzas:
Any update on this one? :'D

jorgee:
This PR is ready, just this minor [comment](https://github.com/nextflow-io/nextflow/pull/6903#discussion_r2910247227) not related to the issue.

modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
bentsherman: When a task fails and succeeds on a subsequent attempt, this is usually a transient error, so I don't think we need to go out of our way to "skip" earlier attempts on a resume.

Especially if the task still needs to executed -- we don't want to preemptively decrease the number of retries it has.

There is also the case where a task initially fails because of insufficient resources. then succeeds on retry with increased resources. I think what matters is that when the task succeeds, a resumed run should be able to "cycle" through the cached failures, arrive at the successful attempt, and resumes it

bentsherman: Removed

tests/checks/resume-retried-with-abort.nf/.checks
bentsherman, line 1: @jorgee does this test need to be updated since I removed the extra logic? it is failing for me locally

jorgee: Yes, now the third execution also retries and execution will take more time. Updating it.

jorgee: It seems there was an inconsistency with failed counts and failed counts. Removing the code, the number of attempts was not increased but failedCounts yes, and it is also used for retries. 

I have renamed the `abortedCount` to `previousFailOrAbortedCount`. It stores the aborted and failed cached execution. This is just used for increasing the tries

bentsherman: Okay, I think I understand all the changes. The only problem now is that this e2e test continues to fail locally

pditommaso (changes requested):
Please add unit tests to capture this behaviour

tests/checks/resume-retried-with-abort.nf/.checks
jorgee: Is it possible to get the log of your e2e test? I would like to know why is it failing in your case.

jorgee:
> Please add unit tests to capture this behaviour
 
 There was a validation tests checking the full fuctionality. I have added unit tests in the last commit.

bentsherman:
CORRECTION: Closing in favor of #7194

PR #6905, opened 2026-03-10
Support staging closure for input files in typed process

### LINKED ISSUE #3631: option to save/restore relative file paths
Currently, when a 'path' output of process A (e.g. param1/param2/data.csv) is passed as a 'path' input to a downstream process 'B', only the basename of the file (data.csv) is preserved; the path relative to task execution directory (param1/param2) is lost.  This causes headaches whenever we adapt for Nextflow existing scripts that use paths to represent metadata.   I often end up adding boilerplate code to manually save the full paths, pass them along as extra outputs, and restore them in subsequent steps.  It would help if Nextflow had an option to save/restore the full relative paths automatically.

bentsherman:
This PR adds the ability to stage input files with a *staging closure* instead of a glob pattern, which allows for more find-grained control over individual file names

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskInputResolver.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskInputResolver.groovy
@@ -88,7 +88,9 @@ class TaskInputResolver {
     List<FileHolder> resolve(ProcessFileInput fileInput, Object value) {
         final ctx = task.context
         final normalized = normalizeInputToFiles(value, count, true)
-        final resolved = expandWildcards( fileInput.getFilePattern(ctx), normalized )
+        final resolved = fileInput.getStagingClosure() != null
+            ? expandStagingClosure( fileInput.getStagingClosure(), normalized )
+            : expandWildcards( fileInput.getFilePattern(ctx), normalized )
 
         count += resolved.size()
 
@@ -293,6 +295,24 @@ class TaskInputResolver {
         }
         m.appendTail(result)
         result.toString()
+    }
+
+    /**
+     * When a staging closure is provided for a collection of files,
+     * stage each file by applying the staging closure.
+     *
+     * @param stager
+     * @param files
+     */
+    protected static List<FileHolder> expandStagingClosure(Closure stagingClosure, List<FileHolder> files) {
+        assert files != null
+
+        final result = new ArrayBag(files.size())
+        for( final holder : files ) {
+            final stageName = stagingClosure.call(holder.getStorePath())
+            result << holder.withName(stageName)
+        }
+        return result
     }
 
     protected static Object singleItemOrList( List<FileHolder> items, boolean single, ScriptType type ) {
--- a/docs/process-typed.md
+++ b/docs/process-typed.md
@@ -241,6 +241,28 @@ process grep {
 }
 ```
 
+:::{versionadded} 26.04.0
+:::
+
+Input files can be staged using a *staging closure* instead of a file pattern:
+
+```nextflow
+process ls {
+    input:
+    slice: Set<Path>
+
+    stage:
+    stageAs(slice) { file -> "${file.parent.name}/${file.name}" }
+
+    script:
+    """
+    ls -1 */*.txt | sort
+    """
+}
+```
+
+The staging closure should define the stage name for a given input file.
+
 See {ref}`process-reference-typed` for available stage directives.
 
 ## Outputs
--- a/docs/reference/process.md
+++ b/docs/reference/process.md
@@ -79,6 +79,9 @@ The following directives can be used in the `stage:` section of a typed process:
 `stageAs( value: Iterable<Path>, filePattern: String )`
 : Stages a collection of files into the task directory under the given alias.
 
+`stageAs( value: Iterable<Path>, transform: (Path) -> String )`
+: Stage a collection of files into the task directory with an alias determined by the given closure.
+
 `stdin( value: String )`
 : Stages the given value as the standard input (i.e., `stdin`) to the task script.
 
--- a/modules/nextflow/src/main/groovy/nextflow/script/params/v2/ProcessFileInput.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/params/v2/ProcessFileInput.groovy
@@ -34,19 +34,31 @@ class ProcessFileInput {
     private Object filePattern
 
     /**
+     * Closure that defines the stage name for a given input file.
+     */
+    private Closure stagingClosure
+
+    /**
      * Lazy expression (e.g. closure) which defines which files
      * to stage in terms of the task inputs.
      * It is evaluated for each task against the task context.
      */
     private Object value
 
-    ProcessFileInput(Object filePattern, Object value) {
-        this.filePattern = filePattern != null ? filePattern : '*'
+    ProcessFileInput(Object patternOrClosure, Object value) {
+        if( patternOrClosure instanceof Closure && patternOrClosure.getMaximumNumberOfParameters() == 1 )
+            this.stagingClosure = patternOrClosure
+        else
+            this.filePattern = patternOrClosure != null ? patternOrClosure : '*'
         this.value = value
     }
 
     String getFilePattern(Map ctx) {
-        return ctx.resolveLazy(filePattern)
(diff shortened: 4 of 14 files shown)


bentsherman:
@Lehmann-Fabian let me know if this PR meets your needs from #2622

modules/nextflow/src/main/groovy/nextflow/processor/TaskInputResolver.groovy
```diff
@@ -295,6 +297,24 @@ class TaskInputResolver {
...
+     */
+    protected static List<FileHolder> expandStagingClosure(Closure stagingClosure, List<FileHolder> files) {
+        assert files != null
+
+        final result = new ArrayBag(files.size())
+        for( final holder : files ) {
+            final stageName = stagingClosure.call(holder.getStorePath())
+            result << holder.withName(stageName)
+        }
+        return result
```
bentsherman, line 315: @Lehmann-Fabian the main difference I saw with your PR is that it seems like you allowed the staging closure to still use glob patterns

I wasn't sure if this is actually needed, and the implementation is much simpler if we say that the staging closure must return a fully-resolved file name (no glob patterns), so I left it out

But let me know if you actually needed this

pditommaso (commented):
I'm skeptical there's a strong use case for dynamic resolution, on there other hands brings back bloated closure in process definitions as we had (have) for publishDir

bentsherman:
You can see Fabian's use case in #2622 

I have come to see it as an alternative to to staging via glob pattern, which doesn't give you as much control and forces certain conventions on you (e.g. numbering files)

The closure doesn't bother me so much because (1) it will be rarely used and (2) it is a mapping function rather than a closure with no arguments (which is redundant)

It's essentially the same thing we do for workflow outputs:
```groovy
output {
    samples {
        path { sample -> "fastq/${sample.id}/" }
    }
}
```

It's the simplest way to map files from one environment to another

bentsherman:
From Fabian's example ([nf-core/rangeland](https://github.com/nf-core/rangeland)):

```groovy
    script:
    """
    # ...

    # Rename files: /trend/<Tile>/<Filename> to <Tile>_<Filename>, otherwise we can not reextract the tile name later
    results=`find trend -name '*.tif*'`
    parallel -j $task.cpus 'mv {} {//}_{/}' ::: \$results
    """
```

So the main use case is when you have a multi-dimensional file collection (e.g. grid data)

Many tools expect a particular file/directory structure, and while we are trying to get away from encoding metadata in file paths, the reality is that many tools still do it

Nextflow's pattern-based staging isn't good enough here, so you end up "re-staging" the files in your process script (like the `mv` command above). So I think a `input file -> stage name` staging closure is a better way to do this

pditommaso:
## Alternative: Declarative String Replacement Rule (sed-like)

Instead of a closure, we could use a **declarative pattern-based renaming rule** — a string that specifies how to transform the original file path into the staged name.

### Option A: `sed`-style substitution syntax

```groovy
stageAs slice, 's|.*/([^/]+)/([^/]+)$|$1/$2|'
```

### Option B: Path template with named segments

Define a mini-language that references parts of the source path:

```groovy
stageAs slice, '{parent}/{name}'
```

Where built-in tokens map to path components:
| Token | Meaning | Example for `/data/tile1/sample.tif` |
|-------|---------|--------------------------------------|
| `{name}` | File name with extension | `sample.tif` |
| `{simpleName}` | File name without extension | `sample` |
| `{extension}` | File extension | `tif` |
| `{parent}` | Immediate parent dir name | `tile1` |
| `{parent2}` | Grandparent dir name | `data` |
| `{path:N}` | Last N path segments | `tile1/sample.tif` |

Examples:
```groovy
// Preserve parent directory structure
stageAs slice, '{parent}/{name}'
// tile1/sample.tif, tile2/sample.tif

// Flatten with prefix from parent
stageAs slice, '{parent}_{name}'
// tile1_sample.tif, tile2_sample.tif

// Keep last 2 path segments (same as parent/name)
stageAs slice, '{path:2}'
// tile1/sample.tif
```

### Option C: Regex substitution with explicit syntax

```groovy
stageAs slice, from: '.*/([^/]+)/([^/]+)', to: '$1/$2'
```

## Comparison

| Criteria | Closure (PR) | Path Templates (B) | Regex Sub (A/C) |
|----------|-------------|-------------------|-----------------|
| **Readability** | Medium — Groovy knowledge needed | High — self-documenting tokens | Low — regex is hard to read |
| **Power/Flexibility** | Maximum — arbitrary Groovy code | Covers 90%+ of use cases | Full regex power |
| **Declarative** | No — imperative code | Yes | Yes |
| **Serializable** | No — closures are opaque | Yes — plain string | Yes — plain string |
| **Config override** | Hard | Easy (`process.stageAs = ...`) | Possible but ugly |
| **Learning curve** | Low for Groovy users | Very low | High |
| **Composable with existing `*`/`?` patterns** | No — separate code path | Could extend existing pattern language | No |

## Recommendation: Path Templates (Option B)

The **path template** approach (`'{parent}/{name}'`) seems the strongest alternative because:

1. **Declarative & serializable** — it's just a string, making it compatible with config overrides, caching keys, and serialization
2. **Readable** — `'{parent}/{name}'` is immediately understandable vs `{ file -> "${file.parent.name}/${file.name}" }`
3. **Consistent with Nextflow's existing pattern language** — extends the glob-based `*`/`?` patterns rather than introducing a completely different paradigm (closures)
4. **Covers the real use cases** — the rangeland case (`tile/sample.tif` preservation) and the PR's test case (`group1/sample1.txt`) are both expressible with simple token substitution
5. **No closure serialization concerns** — closures in Nextflow can cause issues with `-resume` caching and Kryo serialization

bentsherman:
@flowuenne can you describe the Hive use case you encountered?

One question I also have -- do you actually need to rename / remap the input files from their original structure, or just preserve them as they are?

Part of the issue is that `stageAs` ignores directories when staging input files. It just stages them in by their base file name. So if I provide a collection of files like this:

```
/.../group1/sample1.txt
/.../group1/sample2.txt
/.../group1/sample3.txt
/.../group2/sample1.txt
/.../group2/sample2.txt
/.../group2/sample3.txt
/.../group3/sample1.txt
/.../group3/sample2.txt
/.../group3/sample3.txt
```

They will be staged in as `sample1.txt`, `sample2.txt`, and `sample3.txt`, which means they will overwrite each other.

But the problem is that Nextflow doesn't know how many parent directories should be included when staging them into a task (i.e. which part belongs to the `...`?)

1. if the files are coming from an upstream task output, then we can simply use the relative path against the task directory

2. if the files coming from outside the work directory... ?

If we could find an elegant solution to (2), we might not need the staging closure (or any other alternative) at all. We could just make the default staging behavior do the right thing and then you don't have to think about it.

grst:
I have another use case that I think would be addressed by this PR. 

10x cellranger requires fastq files to follow a specific naming convention. All fastq files belonging to a sample need to be staged into the same process and  have the following structure to be recognized by cellranger: 

```
MySampleName_S1_L001_R1_001.fastq.gz
MySampleName_S1_L001_R2_001.fastq.gz
MySampleName_S1_L002_R1_001.fastq.gz
MySampleName_S1_L002_R2_001.fastq.gz
MySampleName_S1_L003_R1_001.fastq.gz
MySampleName_S1_L003_R2_001.fastq.gz
...
```

While this is the default naming convention of Illumina demultiplexers, files can be named differently for various reasons (e.g. because they were downloaded from SRA). When they are specified in the samplesheet.csv of the the nf-core/scrnaseq pipeline, we need to rename them to follow the cellranger conventions. This is currently handled by a  [custom python logic](https://github.com/nf-core/scrnaseq/blob/f7bf36d7c7e4bddc5302c3facd8d19ca83e22226/modules/nf-core/cellranger/count/templates/cellranger_count.py) that I would be more than happy to get rid of. 

If this proposal got accepted, if I understood this correctly, I could have some groovy logic at the beginning that figures out the target filenames and places them into e.g. `meta` and then have a process like this

```nextflow
process CELLRANGER_COUNT { 
    input: 
    tuple val(meta), path(fastq)

    stage:
    stageAs(fastq) { file -> meta["filename_map"]["${file.parent.name}/${file.name}"] } 

    script: 
    """
    cellranger count [...]
    """
}

PR #6911, opened 2026-03-11
feat: add native support for TPUs and multi-vendor hardware accelerators

Joon-Klaps:
This PR extends the PodSpecBuilder to natively support a wider range of hardware accelerators, specifically targeting Google TPUs (GKE), AMD GPUs, AWS Neuron (Inferentia/Trainium), and Intel Gaudi devices. It removes the need for manual pod directive overrides by automatically mapping common hardware shorthands to their correct Kubernetes Extended Resource strings.

Previously, the getAcceleratorType logic was heavily biased toward NVIDIA GPUs (nvidia.com/gpu). If a user requested a TPU or an AMD GPU using shorthand (e.g., accelerator: 'tpu'), the builder would either fail to map it correctly or produce an invalid resource string like tpu.com/gpu. This forced AI/ML practitioners to manually override Pod specs to run workloads on non-NVIDIA hardware.


tomiles (approved):
It looks great! I like the idea of adding the .toLowerCase() method to the type definition. 👍🏻

PR #6912, opened 2026-03-12
Use destructuring syntax for process record/tuple inputs

pditommaso:
## Summary

- Proposes using the `record()` function-call notation uniformly for both process inputs and outputs
- Replaces the asymmetric `Record { ... }` block syntax currently used only in input declarations
- Aligns with existing assignment (`=`) and type annotation (`: Type`) patterns already present in process I/O

**Current (asymmetric):**
```nextflow
input:
sample: Record {         // block syntax, unique to inputs
    id: String
    fastq_1: Path
}

output:
record(id: sample.id, html: file('*.html'))   // function call
```

**Proposed (uniform):**
```nextflow
input:
sample: Sample = record(id: String, fastq_1: Path)

output:
result: OtherType = record(id: sample.id, html: file('*.html'))
```

## Test plan

- [ ] Review ADR content for accuracy and completeness
- [ ] Discuss with team whether this direction aligns with typed syntax roadmap

### DIFF
--- a/modules/nf-lang/src/test/groovy/nextflow/script/parser/ScriptAstBuilderTest.groovy
+++ b/modules/nf-lang/src/test/groovy/nextflow/script/parser/ScriptAstBuilderTest.groovy
@@ -345,26 +345,6 @@ class ScriptAstBuilderTest extends Specification {
 
             process hello {
                 input:
-                list(id)
-
-                script:
-                ""
-            }
-            '''
-        )
-        then:
-        errors.size() == 1
-        errors[0].getStartLine() == 5
-        errors[0].getStartColumn() == 5
-        errors[0].getOriginalMessage() == "Invalid process structured input -- should be `record()` or `tuple()`"
-
-        when:
-        errors = check(
-            '''\
-            nextflow.preview.types = true
-
-            process hello {
-                input:
                 tuple(id: String)
 
                 script:


adr/20260312-record-syntax-unification.md
```diff
@@ -0,0 +1,242 @@
...
+    result = record(id: sample.id, html: file('*.html'), zip: file('*.zip'))
+}
+```
+
+With optional explicit type annotation:
+
+```nextflow
+process FASTQC {
+    input:
+    sample: Sample = record(id: String, fastq_1: Path, fastq_2: Path)
```
bentsherman: Worth mentioning that the `= record(...)` part here is redundant if you already declare the input with a record type (`Sample`)

adr/20260312-record-syntax-unification.md
```diff
@@ -0,0 +1,242 @@
...
+**Output** — a `record()` function call:
+```nextflow
+process FASTQC {
+    ...
+    output:
+    record(id: sample.id, html: file('*.html'), zip: file('*.zip'))
+}
+```
+
+This asymmetry means the same concept (a record) is expressed with two different syntactic forms depending on context. The block syntax `Record { ... }` exists only in process input declarations and has no counterpart elsewhere in the language. Meanwhile, the `record()` function call used in outputs is already a general-purpose construct usable in any expression context.
```
bentsherman: > This asymmetry means the same concept (a record) is expressed with two different syntactic forms depending on context.

I think the asymmetry is intentional. It helps distinguish between two slightly different concepts -- an input vs an output. Inputs and outputs have slightly different behaviors, especially in a process.

- Inputs are receiving values from an external source, validating them against a declared structure, and staging them into the task environment

- Outputs are collecting values from the task environment and pushing them into an output structure

So, process inputs and outputs are similar in some ways and different in others. The question is whether it is better to highlight their similarities or their differences

bentsherman: For what it's worth, the language server actually does highlight their similarities. When hovering on a process call, the process hover hint will be rendered as:

```groovy
process FASTQC {
    input:
    sample: Record {
        id: String
        fastq_1: Path
        fastq_2: Path
    }

    output:
    result: Record {
        id: String
        html: Path
        zip: Path
    }
}
```

adr/20260312-record-syntax-unification.md
```diff
@@ -0,0 +1,242 @@
...
+process FASTQC {
+    input:
+    sample: Record {
+        id: String
+        fastq_1: Path
+        fastq_2: Path
+    }
+
+    output:
+    record(id: sample.id, html: file('*.html'), zip: file('*.zip'))
```
bentsherman:
```suggestion
    record(
        id: sample.id,
        html: file('*.html'),
        zip: file('*.zip')
    )
```

This rendering of the current output syntax is a more fair comparison IMO

adr/20260312-record-syntax-unification.md
```diff
@@ -0,0 +1,242 @@
...
+    output:
+    result: QcResult = record(id: sample.id, html: file('*.html'), zip: file('*.zip'))
+}
+```
+
+- Good, because same notation on both sides — `name = record(...)`.
+- Good, because reuses existing assignment and type annotation patterns.
+- Good, because `record()` is already a general-purpose function, no new syntax needed.
+- Good, because type annotations follow standard rules — `sample: Sample = record(...)` works like any typed assignment.
+- Bad, because input `record()` arguments are types rather than values, which is a different usage of the function.
```
bentsherman: This is my main issue with Option 3. The double usage of `record()` here is subtle and more likely to confuse users and agents

Whereas the syntax I ended up using for record inputs is easy to explain as an "inline record type"

So I think with Option 3 you are ultimately trading one form of double usage for another, without much benefit

bentsherman:
Great discussion overall. I would be keen to incorporate this somehow into the record type ADR, since we did discuss some of these options throughout the process but I didn't include them as alternatives in the original ADR

Would be good to document these alternatives so that we have the clear rationale for the final syntax

bentsherman:
Give me a moment, I'm going to rearrange my comments so that they can be treated separately

adr/20260312-record-syntax-unification.md
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    output:
+    record(id: sample.id, html: file('*.html'), zip: file('*.zip'))
+}
+```
+
+- Good, because input block syntax mirrors the top-level `record` definition.
+- Bad, because two different notations for the same concept in the same process definition.
+- Bad, because `Record { ... }` block syntax only exists in input declarations — it is not a general-purpose construct.
```
bentsherman: Not entirely true -- the block syntax is identical to the block syntax in a record declaration, as well as the block syntax used by the language server to display an anonymous record type e.g. in a hover hint.

adr/20260312-record-syntax-unification.md
```diff
@@ -0,0 +1,242 @@
...
+- Bad, because output `record()` as a bare statement (no assignment) doesn't allow naming the output.
```
bentsherman: Not true, the output can be an assignment:

```groovy
input:
sample: Record {
    id: String
    fastq_1: Path
    fastq_2: Path
}

output:
result: OtherType = record(
    id: sample.id,
    html: file('*.html'),
    zip: file('*.zip')
)
```

adr/20260312-record-syntax-unification.md
```diff
@@ -0,0 +1,242 @@
...
+    ...
+}
+```
+
+**Output** — a `record()` function call:
+```nextflow
+process FASTQC {
+    ...
+    output:
+    record(id: sample.id, html: file('*.html'), zip: file('*.zip'))
```
bentsherman:
```suggestion
    record(
        id: sample.id,
        html: file('*.html'),
        zip: file('*.zip')
    )
```

adr/20260312-record-syntax-unification.md
bentsherman:
```suggestion
```

adr/20260312-record-syntax-unification.md
```diff
@@ -0,0 +1,260 @@
...
+    result: QcResult = record(id: sample.id, html: file('*.html'), zip: file('*.zip'))
+}
+```
+
+- Good, because same notation on both sides — `name = record(...)`.
+- Good, because reuses existing assignment and type annotation patterns.
+- Good, because `record()` is already a general-purpose function, no new syntax needed.
+- Good, because type annotations follow standard rules — `sample: Sample = record(...)` works like any typed assignment.
+- Bad, because input `record()` arguments are types rather than values, which is a different usage of the function.
+
```
bentsherman:
```suggestion
- Bad, because the use of an assignment in the input declaration to describe the record structure conflicts with the conventional meaning of input assignments as default values.
```

adr/20260312-record-syntax-unification.md
```diff
@@ -0,0 +1,260 @@
...
+    output:
+    result: FastqcResult = record(
+        id: sample.id,
+        html: file('*.html'),
+        zip: file('*.zip')
+    )
+```
+
+- Good, because input block syntax mirrors the top-level `record` definition.
+- Bad, because two different notations for the same concept in the same process definition.
```
bentsherman:
```suggestion
- Good, because the different `Record { ... }` and `record(...)`  reflect the slightly different semantics of staging an input vs unstaging an output.
- Bad, because two different notations for similar concepts (records) create asymmetry in the process definition.
```

adr/20260312-record-syntax-unification.md
bentsherman: From our discussion -- it could be useful to have both as a way to future-proof against changes to the `Sample` type. For example, if someone adds some Path fields to `Sample`, you don't necessarily want `FASTQC` to automatically start staging them as inputs

But at that point, there is no point to using `Sample` at all -- you might as well just use the generic `Record` type:
```nextflow
    sample: Record = record(id: String, fastq_1: Path, fastq_2: Path)
```

adr/20260312-record-syntax-unification.md
awgymer: I always parsed the difference as being the input was a type declaration (it has the same format except without a leading `record` I think?) that would be used to create an anonymous record type for input, and the output was instantiating a generic record instance, which can optionally be assigned? 

This makes sense to me I think.

adr/20260312-record-syntax-unification.md
```diff
@@ -0,0 +1,261 @@
...
+    record {
+        id: String = sample.id
+        html: Path = file('*.html')
+        zip: Path = file('*.zip')
+    }
+}
+```
+
+- Good, because symmetric — same block form on both sides.
+- Bad, because the output block mixes type declarations with value assignments (`Path = file(...)`).
```
awgymer: Is this not more akin to a type definition with some defaults? I know this may not be something that really exists in Java/Groovy but it makes perfect sense when thinking of record types as akin to something like Pydantic, where you would declare a record with a default using similar syntax. 

e.g.
```py
class DemoModel(BaseModel):
    ts: datetime = Field(default_factory=datetime.now)
```

adr/20260312-record-syntax-unification.md
prototaxites: Agreed, I think this syntax is clearer and nicely separates out typing from instantiation.

adr/20260312-record-syntax-unification.md
bentsherman: Pydantic is an interesting comparison here. I think option 3 is essentially the same as Pydantic -- using function calls to create type definitions

Pydantic is constrained by Python syntax, so for them the best option is to use assignments and function calls, even though it conflates the meaning of this syntax (creating a value vs declaring a type). But this is probably still much better than creating a custom DSL that users would have to learn alongside Python

We have no such constraint in Nextflow, so we can differentiate these type declarations with different syntax

pditommaso:
**Updated ADR: Unified constructor notation for `record()` and `tuple()`**

Key changes in this update:

1. **Broadened scope from record-only to all structured types.** The proposal now establishes `name = constructor(...)` as a uniform notation that applies to both `record()` and `tuple()`, not just records.

2. **Three-tier notation for both constructors:**
   ```nextflow
   tuple(id, file('*.bam'))                          // bare
   out = tuple(id, file('*.bam'))                    // assignment
   out: Tuple<String,Path> = tuple(id, file('*.bam'))  // typed assignment
   ```
   Same pattern for `record()` — bare, assignment, typed assignment.

3. **Migration path from classic DSL2 through tuples to records:**
   ```nextflow
   tuple val(id), path(fastq)             // classic DSL2
   in = tuple(id: String, fastq: Path)    // typed — uniform notation
   in = record(id: String, fastq: Path)   // record — just change the keyword
   ```

4. **Tuple input with inline named types** — `in = tuple(id: String, fastq: Path)` — gives tuples named components while preserving positional semantics, mirroring record input notation exactly.

5. **Coexistence example** showing both `tuple()` and `record()` in the same pipeline with identical notation shape.

The core value: **users learn one pattern and apply it everywhere**. The only choice is positional (`tuple`) vs named (`record`).

adr/20260312-record-syntax-unification.md
pditommaso: Fair, remove it

adr/20260312-record-syntax-unification.md
pditommaso: My counter argument is that's aligned with the semantic for record constructor. Also the double usage is done for `file`. Above the central point is the continuity with existing notation `val`, `path`, `tuple`, etc

adr/20260312-record-syntax-unification.md
pditommaso: I understand that's intentional, but my claim is that it's not needed cognitive load. Why as a user I should think to two different notations to express the structure of input and outputs. 

Above all, the central point is how the syntax can be evolved keeping some structural continuity with the existing syntax so that as a nextflow developer I feel comfortable with it without the need to learn new concepts.

adr/20260312-record-syntax-unification.md
bentsherman: Function overloading like with `file()` is fine if done cautiously, but what I'm talking about is syntax overloading. You are hijacking the assignment and function call *syntax* for different purposes, which creates unnecessary cognitive load

The function call by itself could make sense as a reverse constructor pattern:

```groovy
input:
tuple(id: String, fastq_1: Path, fastq_2: Path)
record(id: String, fastq_1: Path, fastq_2: Path)
```

That would also have better continuity with the legacy syntax

adr/20260312-record-syntax-unification.md
bentsherman: > I feel comfortable with it without the need to learn new concepts.

Static typing and records *are* new concepts. There is no getting around that.

The question is, given that users will have to learn the new syntax for parameters, type annotations, record types, etc, is it better for the process inputs/outputs to be consistent with the new syntax everywhere else or consistent with the legacy process syntax?

If legacy continuity gets in the way of expressing inputs/outputs in the new system, surely the latter must take precedence

At the same time, we might be able to achieve both...

The best case I can see is to ditch the assignment and just have a destructor -> constructor pattern:

**legacy**
```groovy
    input:
    tuple val(id), path(fastq_1), path(fastq_2)

    output:
    tuple val(id), path("fastqc_${id}_logs")
```

**typed (tuple)**
```groovy
    input:
    tuple(id: String, fastq_1: Path, fastq_2: Path)

    output:
    tuple(id, file("fastqc_${id}_logs"))
```

**typed (record)**
```groovy
    input:
    record(
        id: String,
        fastq_1: Path,
        fastq_2: Path
    )

    output:
    record(
        id: id,
        fastqc: file("fastqc_${id}_logs")
    )
```

This works well with static typing and the legacy continuity is decent.

bentsherman: My only concerns at that point would be:

- The commas can get ugly with type annotations, for example:

```groovy
    input:
    record(
        id: String,
        single_end: Boolean,
        reads: List<Path>, // ugly
        args: String?, // ugly
        prefix: String?
    )
```

  This is partly what led me to the block syntax which doesn't require commas.

- If the new syntax is *too* similar, that could cause it's own confusion. Might be hard to distinguish between typed vs legacy syntax. This is the concern we had with typed workflows

docs/process-typed.md
```diff
@@ -89,62 +90,62 @@ process cat_opt {
...
-    id: String
-    fastq: Path
-}
 ```
 
-This approach is useful when the record type can be re-used elsewhere in the pipeline.
+This pattern mirrors the standard `record()` function used to construct records. In this example, `fastq` is staged as an input file since the `fastq` field is declared with type `Path`.
+
+:::{tip}
+Record inputs are a useful way to select a subset of fields from a larger record. This way, the process only stages what it needs, allowing you to keep related data together in your workflow logic.
```
christopher-hakkaart:
```suggestion
Record inputs are a useful way to select a subset of fields from a larger record. This way, the process stages only what it needs, keeping related data together in your workflow logic.
```

docs/process-typed.md
```diff
@@ -89,62 +90,62 @@ process cat_opt {
...
+:::
 
 ### Tuple inputs
 
-Inputs with type `Tuple` can declare the name of each tuple component:
+Tuple inputs cen be declared as a *destructured* input:
```
christopher-hakkaart:
```suggestion
Tuple inputs can be declared as a *destructured* input:
```

PR #6920, opened 2026-03-13
Fix K8s token refresh by delegating getClient() to K8sConfig cache

### LINKED ISSUE #6918: K8s executor caches client token at wrong layer — PR #6742 token refresh never triggered
## Summary

### The caching chain

```
  ↓ called once
  ↓
  ↓
```

### Reproduction

### Error

```
Caused by:
```

### Expected behavior

### Suggested fix

### Regression note

### Environment

- nf-k8s plugin: 1.5.0
- Kubernetes: AKS 1.27.7
- Related: #6741, #6742

adamrtalbot:
## Summary

Fixes #6918

### Problem

PR #6742 added a Guava cache to `K8sConfig.getClient()` with a 50-minute expiry to refresh the service account token. However, the cache was never actually consulted after startup because:

- `K8sExecutor.register()` called `k8sConfig.getClient()` **once** and stored the result in a private `K8sClient` field
- `K8sExecutor.getClient()` returned that stored field directly (never re-calling `K8sConfig`)
- `K8sTaskHandler` stored `executor.client` in its constructor, bypassing `getClient()` entirely

This caused **401 Unauthorized errors after ~60 minutes** on clusters with short-lived projected service account tokens (e.g. AKS, RKE2 with default ~1hr token lifetime).

### Fix

- Remove the `private K8sClient client` field from `K8sExecutor`
- Change `K8sExecutor.getClient()` to delegate to `k8sConfig.getClient()` on each invocation — the Guava cache handles deduplication and refreshes the token when the 50-minute window expires
- Fix `K8sTaskHandler` constructor to use `executor.getClient()` instead of accessing `executor.client` directly (the direct field access bypassed `getClient()` entirely)

### DIFF
--- a/plugins/nf-k8s/src/main/nextflow/k8s/K8sExecutor.groovy
+++ b/plugins/nf-k8s/src/main/nextflow/k8s/K8sExecutor.groovy
@@ -41,20 +41,20 @@ import org.pf4j.ExtensionPoint
 class K8sExecutor extends Executor implements ExtensionPoint {
 
     /**
-     * The Kubernetes HTTP client
-     */
-    private K8sClient client
-
-    protected K8sClient getClient() {
-        client
-    }
-
-    /**
      * @return The `k8s` configuration scope in the nextflow configuration object
      */
     @Memoized
     protected K8sConfig getK8sConfig() {
         new K8sConfig( (Map<String,Object>)session.config.k8s )
+    }
+
+    /**
+     * @return The Kubernetes HTTP client. Delegates to {@link K8sConfig#getClient()} on each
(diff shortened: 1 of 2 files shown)


pditommaso:
Thanks for identifying the root cause @adamrtalbot! Closing in favour of #6925 which takes a slightly different approach — it moves the Guava cache from `K8sConfig` to `K8sExecutor` so that the `K8sClient` itself is cached (not just `ClientConfig`). This avoids re-creating the client (including SSL cert parsing and trust manager setup) on every API call, while still refreshing the token when the cache expires.

PR #6922, opened 2026-03-13
Support process named arguments in typed workflows

bentsherman:
Spun off from #6881 

This PR adds the ability to used named arguments when calling a process in a typed workflow.

If the process declares a single record input, and the source channel contains records, then the named arguments will be appended to each record in the source channel before it is passed to the process.

Keeping as draft until #6881 is merged


bentsherman:
Closing in favor of overloading `cross()` operator with named args in #6881

PR #6925, opened 2026-03-16
Fix K8s token refresh by caching K8sClient at executor level

### LINKED ISSUE #6918: K8s executor caches client token at wrong layer — PR #6742 token refresh never triggered

pditommaso:
## Summary

Fixes #6918 — supersedes #6920

### Problem

PR #6742 added a Guava cache to `K8sConfig.getClient()` with a 50-minute expiry to refresh the service account token. However, the cache was never consulted after startup because:

- `K8sExecutor.register()` called `k8sConfig.getClient()` once and stored the result in a private `K8sClient` field
- `K8sExecutor.getClient()` returned that stored field directly
- `K8sTaskHandler` stored `executor.client` in its constructor, bypassing `getClient()` entirely

This caused **401 Unauthorized errors after ~60 minutes** on clusters with short-lived projected service account tokens (e.g. AKS, RKE2).

### Why not #6920

PR #6920 correctly identified the root cause but had two issues:

1. **Created a new `K8sClient` on every `getClient()` call** — the Guava cache in `K8sConfig` only cached `ClientConfig`, so every poll/submit/delete call triggered `new K8sClient(...)` which re-parses SSL certificates and initializes the trust manager
2. **Cache at the wrong level** — `K8sConfig` is a configuration object; the client lifecycle belongs in the executor

### Fix

- **Move the Guava cache from `K8sConfig` to `K8sExecutor`** — cache the `K8sClient` itself (not just `ClientConfig`), avoiding redundant SSL setup on every API call
- `K8sConfig.getClient()` becomes a plain factory method that creates a fresh `ClientConfig` (re-reading the token from disk)
- `K8sTaskHandler` uses `executor.getClient()` instead of direct field access
- Added `K8sExecutorTest` covering cache hit and expiration behavior

### DIFF
--- a/plugins/nf-k8s/src/main/nextflow/k8s/K8sExecutor.groovy
+++ b/plugins/nf-k8s/src/main/nextflow/k8s/K8sExecutor.groovy
@@ -16,6 +16,10 @@
 
 package nextflow.k8s
 
+import java.util.concurrent.TimeUnit
+
+import com.google.common.cache.Cache
+import com.google.common.cache.CacheBuilder
 import groovy.transform.CompileStatic
 import groovy.transform.Memoized
 import groovy.util.logging.Slf4j
@@ -41,12 +45,17 @@ import org.pf4j.ExtensionPoint
 class K8sExecutor extends Executor implements ExtensionPoint {
 
     /**
-     * The Kubernetes HTTP client
+     * Cache for the Kubernetes HTTP client. The client is refreshed periodically
+     * so that the service account token is re-read when it expires.
      */
-    private K8sClient client
+    private Cache<String, K8sClient> clientCache
 
+    /**
+     * @return The Kubernetes HTTP client. Delegates to a Guava cache that refreshes
+     * the client (including the service account token) when the configured interval expires.
+     */
     protected K8sClient getClient() {
-        client
+        clientCache.get('client', () -> new K8sClient(k8sConfig.getClient()))
     }
 
     /**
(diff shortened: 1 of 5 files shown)


plugins/nf-k8s/src/main/nextflow/k8s/K8sExecutor.groovy
```diff
@@ -41,12 +45,17 @@ import org.pf4j.ExtensionPoint
...
-    private K8sClient client
+    private Cache<String, K8sClient> clientCache
 
+    /**
+     * @return The Kubernetes HTTP client. Delegates to a Guava cache that refreshes
+     * the client (including the service account token) when the configured interval expires.
+     */
     protected K8sClient getClient() {
-        client
+        clientCache.get('client', () -> new K8sClient(k8sConfig.getClient()))
```
adamrtalbot, line 58: Took me a sec to get this, so I'll leave a comment here for future readers. The client config (including token) is cached, but the k8s client itself is recreated on every submit task. This means on a new client is created every time but it's very lightweight. While building a client, it will use the config which is refreshed every 50 minutes (default value).

This is all handled by `getClient()` reaches into the Guava cache instead of getting a fresh client. Guava cache handles use-or-get logic.

PR #6914, opened 2026-03-12
Fix cache invalidation for module binaries

FloWuenne:
## Summary

Fix task cache invalidation when using module binaries with Wave.

Task hashing previously considered project-level `bin/` scripts but not module-level binaries under `resources/usr/bin`, which could cause stale cached tasks to be reused after a module binary changed.

This change makes task hashing resolve referenced executables from the task bin paths, including module binary directories, so module binary updates invalidate the cache as expected.

Closes #6128

## Testing

- Added a regression test for module binary hash changes in `TaskHasherTest`
- Ran:
  - `./gradlew :nextflow:test --tests 'nextflow.processor.TaskHasherTest'`

PR #6927, opened 2026-03-16
Add versioned task hasher strategy

### LINKED ISSUE #6128: caching with module binaries files is not identifying changes
## Bug report 

module binaries file changes are not identified to invalidate the cache for the corresponding task.

when running a workflow that uses a module with a defined process, if cache is defined for the process (true or deep) and you resume, then the next time you run the workflow the process will use the cache and not rerun. if you change the code for the module binary :
**the expected behavior** is that the process will rerun since the module binary file has changed. 
**the actual behavior** is that  the process will still use the cache and not rerun the process although the module binary file has changed. 

main.nf

```
#!/usr/bin/env nextflow
include { analyzeData } from './modules/module1'
workflow {
    analyzeData("hi")
}
```

modules/module1/main.nf
```
process analyzeData {

    input:
    val inp

    output:
    stdout emit: bye

    script:
    """
    script.sh
    """
}
```

modules/module1/resources/usr/bin/script.sh
```
#!/usr/bin/env sh
echo "bye1"
```

### Program output 
stdout 
```
nextflow main.nf -resume -dump-hashes

 N E X T F L O W   ~  version 25.04.2

Launching `main.nf` [sleepy_meitner] DSL2 - revision: 35d5df9396

[-        ] analyzeData -
[analyzeData] cache hash: eec876dcd0f8fe26bf907aefbf0d60d0; mode: DEEP; entries: 
  de45988e5684a98dc1020e8280d7e1ec [java.util.UUID] 68c4eadb-9035-48d2-bffb-324f505928db 
  d9fb9cbb8cc05b1d61dd1991faf327c6 [java.lang.String] analyzeData 
[8c/5c93db] analyzeData [100%] 1 of 1, cached: 1 ✔
[analyzeData] cache hash: eec876dcd0f8fe26bf907aefbf0d60d0; mode: DEEP; entries: 
  de45988e5684a98dc1020e8280d7e1ec [java.util.UUID] 68c4eadb-9035-48d2-bffb-324f505928db 
  d9fb9cbb8cc05b1d61dd1991faf327c6 [java.lang.String] analyzeData 
  edd8b2f5b2b58dd7cb4d9629c6d51273 [java.lang.String]     """
    script.sh
    """
 
  611acaa318085dbeaeb4ac5ab86f7859 [java.lang.String] inp 
  865872ab1301d03dd43dbcb57a53eeca [java.lang.String] hi 
  4f9d4b0d22865056c37fb6d9c2a04a67 [java.lang.String] $ 
  16fe7483905cce7a85670e43e4678877 [java.lang.Boolean] true 
```

after changing the script.sh(even a non valid)  it is still cached  and succeed.
fyi if putting the script in:  {workflow_dir}/bin 

### Environment 

* Nextflow version: version 25.04.2 build 5947
* Java version: OpenJDK Runtime Environment Temurin-17.0.10+7 (build 17.0.10+7)
* Operating system: [macOS, Linux]

### Additional context

like I said before putting the script in: {workflow_dir}/bin:
 ```
nextflow main.nf -resume -dump-hashes

 N E X T F L O W   ~  version 25.04.2

Launching `main.nf` [golden_kilby] DSL2 - revision: 35d5df9396

[-        ] analyzeData -
[analyzeData] cache hash: 578d8d58fce6f6f3b9125c7c276952e5; mode: DEEP; entries: 
  de45988e5684a98dc1020e8280d7e1ec [java.util.UUID] 68c4eadb-9035-48d2-bffb-324f505928db 
  d9fb9cbb8cc05b1d61dd1991faf327c6 [java.lang.String] analyzeData 
  edd8b2f5b2b58dd7cb4d9629c6d51273 [java.lang.String]     """
[af/6b1cf2] analyzeData [100%] 1 of 1, cached: 1 ✔
[analyzeData] cache hash: 578d8d58fce6f6f3b9125c7c276952e5; mode: DEEP; entries: 
  de45988e5684a98dc1020e8280d7e1ec [java.util.UUID] 68c4eadb-9035-48d2-bffb-324f505928db 
  d9fb9cbb8cc05b1d61dd1991faf327c6 [java.lang.String] analyzeData 
  edd8b2f5b2b58dd7cb4d9629c6d51273 [java.lang.String]     """
    script.sh
    """
 
  611acaa318085dbeaeb4ac5ab86f7859 [java.lang.String] inp 
  865872ab1301d03dd43dbcb57a53eeca [java.lang.String] hi 
  4f9d4b0d22865056c37fb6d9c2a04a67 [java.lang.String] $ 
  16fe7483905cce7a85670e43e4678877 [java.lang.Boolean] true 
  5e1b7d12410ec6a85ca2cfa755d776f5 [sun.nio.fs.UnixPath] /Users/ron/workspaces/cytoreason-nextflow/bin/script.sh
```

pditommaso:
## Summary

Refactor `TaskHasher` into a versioned strategy so the task hashing algorithm can be selected explicitly, decoupling *what* goes into the hash from *how* the keys are hashed. This makes it possible to preserve older hashing behavior (for cache backward compatibility) while evolving the default.

The hashing version is resolved once at session startup from the `NXF_TASK_HASH_VER` environment variable and cached on the `Session`, so it is not re-parsed per task.

## Design

- `TaskHasher` is an **interface** exposing a single `compute()` method.
- `AbstractTaskHasher` holds the shared hash-key assembly (`collectKeys()`): session id, process name, script source, inputs, eval outputs, script vars, module bundle fingerprint, `module`/`conda`/`spack` directives, stub marker, etc. Subclasses supply a configured `HashBuilder` via `createHashBuilder()` and may extend `collectKeys()`.
- `TaskHasherV1`, `TaskHasherV2`, `TaskHasherV3` are the concrete strategies (see table below).
- `TaskHasherFactory` exposes a `Version` enum, resolves the default from `NXF_TASK_HASH_VER` (falling back to `STD_V3`), and creates the strategy from `session.hashStrategy`.

## Version differences

| Aspect | V1 (`std/v1`) | V2 (`std/v2`) | V3 (`std/v3`) |
|---|---|---|---|
| `HashBuilder.orderIndependentMaps` | `false` — hashes `Map` **values** only, in iteration order | `true` — hashes `entrySet()` (keys+values), order-independent | `true` (inherited from V2) |
| `HashBuilder.cacheFunnelFirst` | `false` — `CacheFunnel` dispatched **after** `Map`/`Set`/`SerializableMarker` | `true` — `CacheFunnel` dispatched **first** | `true` (inherited from V2) |
| Hashes project `bin/` scripts referenced in the task script | ❌ no | ❌ no | ✅ yes |
| Class relationship | `extends AbstractTaskHasher` | `extends AbstractTaskHasher` | `extends TaskHasherV2` + overrides `collectKeys()` |
| Default | | | ✅ (`TaskHasherFactory.DEFAULT()`) |
| Purpose | Reproduce pre-record-types hashes for cache backward compatibility | Current record-types hashing behavior | V2 behavior **plus** bin-entry hashing; the default going forward |

Notes:
- The `cacheFunnelFirst` ordering is the crux of the "record-types" difference: a value implementing both `CacheFunnel` and (e.g.) `SerializableMarker`/`Map` is hashed via its `funnel()` under V2/V3 (checked first) but via the older branch under V1 (checked last).
- Bin-entry hashing was previously in the shared base (all versions). It now lives **only** in V3 (`collectKeys()` override on top of `super.collectKeys()`); V1 and V2 no longer hash bin entries.

## Changes

- **`HashBuilder`** (`nf-commons`): add `withOrderIndependentMaps(boolean)` and `withCacheFunnelFirst(boolean)` flags (both default `true`) controlling `Map` vs `CacheFunnel`/`SerializableMarker` handling — where the version-specific behavior actually lives.
- **`nextflow.processor.hash`** package: `TaskHasher` interface, `AbstractTaskHasher`, `TaskHasherV1`, `TaskHasherV2`, `TaskHasherV3`, `TaskHasherFactory` (the old `processor/TaskHasher.groovy` becomes `processor/hash/AbstractTaskHasher.groovy`).
- **`TaskRun`**: `getTaskGlobalVars()` (folding in `task.ext.*` directive vars) and `getTaskBinEntries()` moved here from the hasher.
- **`Session`**: new `hashStrategy` field, resolved once at startup via `TaskHasherFactory.Version.DEFAULT()`.
- Minor call-site updates in `TaskConfig`, `TaskProcessor`, `LinObserver`.

## Included fix

This branch also carries the **module-binary cache invalidation fix** (#6914, closes #6128): `AbstractTaskHasher.collectKeys()` adds the module resources bundle fingerprint to the task hash when module binaries are enabled, so module binary changes invalidate the cache even when Wave is not used.

## Test plan

- [ ] `./gradlew :nextflow:test --tests "nextflow.processor.hash.TaskHasherTest"` passes
- [ ] `./gradlew :nextflow:test --tests "nextflow.processor.hash.TaskHasherFactoryTest"` passes
- [ ] `./gradlew :nextflow:test --tests "nextflow.processor.TaskRunTest"` passes
- [ ] `./gradlew :nf-commons:test --tests "nextflow.util.HashBuilderTest"` passes
- [ ] Verify V3 (default) hashes referenced `bin/` scripts; V1/V2 do not
- [ ] Verify `NXF_TASK_HASH_VER=std/v1` reproduces pre-record-types hashes
- [ ] Verify module binary changes invalidate the cache without Wave


bentsherman (commented):
Meh, seems more complicated than it's worth. Makes it harder to see the actual task hash computation because it's split up. Every change/correction we make to the task hash gets preserved forever, making it more and more obfuscated over time

I think this is an over-correction to a single anecdote of a customer who upgraded and lost their cache. Even in that case, you don't have to upgrade. You can just run the previous version with `-resume` to recover a one-off run

What we discussed was having a "rolling legacy" hash strategy. For example, you enable an env var (e.g. `NXF_LEGACY_TASK_HASH=true`) and it uses the task hash from the previous version. This would give users some wiggle room to upgrade without breaking their cache, while keeping the runtime clean as we only maintain back-compat for one trailing version

I think the task hash stability really only becomes important once you have a global cache

PR #6932, opened 2026-03-17
Fix remote modules when base dir is not the current working dir

jorgee:
This pull request refactors how remote module resolution works at parsing and compile time. The `RemoteModuleResolver` interface is updated to remove the need to explicitly pass a base directory. 
At pipeline execution modules are resolved according to the basedir. Otherwise, it uses "./modules" as resolution base dir. This fixes the problems when invoking remote pipelines with modules or executing a pipeline with a different path.

A [pipeline](https://github.com/jorgee/nf-modules-test) including remote modules is created to test this behaviour. 

In the main branch, it uses remote modules without pinned versions. You can run with the following command

`./launch.sh run jorgee/nf-modules-test `

The first time you run this pipeline, it will install the modules in the version assets folder (execution basedir)

It also contains the `pinned-versions` branch where the remote module is pinned (commited files). So, modules shouldn't be installed at execution in this case

`./launch.sh run jorgee/nf-modules-test -r pinned-versions `

### DIFF
--- a/modules/nf-lang/src/main/java/nextflow/module/spi/FallbackRemoteModuleResolver.java
+++ b/modules/nf-lang/src/main/java/nextflow/module/spi/FallbackRemoteModuleResolver.java
@@ -32,14 +32,14 @@ public class FallbackRemoteModuleResolver implements RemoteModuleResolver {
 
     @Override
     public Path resolve(String moduleName, Path projectDir) {
-        final Path modulesDir = (projectDir == null) ? projectDir.resolve("modules") : Path.of("modules").toAbsolutePath();
-        final var resolved = modulesDir.resolve(moduleName).normalize();
-        // Prevent path traversal outside the base directory
-        if (!resolved.startsWith(modulesDir.normalize())) {
-            throw new IllegalStateException("Invalid module name '" + moduleName + "' - path escapes the modules directory");
+        var baseDir = projectDir != null ? projectDir : Path.of(".").toAbsolutePath();
+        var modulesDir = baseDir.resolve("modules").normalize();
+        var resolved = modulesDir.resolve(moduleName).normalize();
+        if( !resolved.startsWith(modulesDir) ) {
+            throw new IllegalStateException("Invalid module name '" + moduleName + "' -- path escapes the modules directory");
         }
-        if (!Files.exists(resolved)) {
-            throw new IllegalStateException("Module '" + moduleName + "' not locally found at 'modules' folder - use 'nextflow module install' to download module files");
+        if( !Files.exists(resolved) ) {
+            throw new IllegalStateException("Module '" + moduleName + "' not found in 'modules' directory -- use 'nextflow module install' to install module first");
         }
         return resolved.resolve("main.nf");
     }
--- a/modules/nextflow/src/main/groovy/nextflow/exception/IllegalModulePathException.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/exception/IllegalModulePathException.groovy
@@ -24,5 +24,5 @@ import groovy.transform.InheritConstructors
  * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
  */
 @InheritConstructors
-class IllegalModulePath extends Exception {
+class IllegalModulePathException extends Exception {
 }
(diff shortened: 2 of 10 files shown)


modules/nextflow/src/main/groovy/nextflow/module/ModuleResolver.groovy
```diff
@@ -160,7 +160,7 @@ class ModuleResolver {
             // Install to modules directory (will compute directory checksum for future integrity checks)
             InstalledModule installed = storage.installModule(reference, version, tempFile, downloadUrl)
 
-            log.info "Module ${reference}@${version} installed successfully at ${installed.mainFile.parent}"
+            log.info "Module ${reference}@${version} installed successfully at ${installed.mainFile.parent.toAbsolutePath().toString()}"
```
bentsherman:
```suggestion
            log.info "Module ${reference}@${version} installed successfully at ${installed.mainFile.parent.toAbsolutePath()}"
```

bentsherman:
Thanks Jorge. Let me look at this with the linter and language server. We might need to be able to pass in the projectDir (baseDir) in those cases

jorgee:
Addressed review comment. I skipped the suggested renaming as it collides with other existing classes

modules/nextflow/src/main/groovy/nextflow/module/ModuleResolver.groovy
jorgee: fixed in last commit

modules/nf-lang/src/main/java/nextflow/module/spi/FallbackRemoteModuleResolver.java
```diff
@@ -31,7 +31,11 @@
 public class FallbackRemoteModuleResolver implements RemoteModuleResolver {
 
     @Override
-    public Path resolve(String moduleName, Path baseDir) {
+    public Path resolve(String moduleName) {
+        // Use CWD-relative "modules" directory as a best-effort fallback when no
+        // RemoteModuleResolver SPI implementation is available (e.g. running outside
+        // a full Nextflow session).
+        final Path baseDir = Path.of("modules").toAbsolutePath();
```
bentsherman: This won't work for the language server because it can manage multiple workspace roots, so it needs to be able to provide a specific projectDir. Can we keep baseDir as a parameter here?

Also, I would prefer to call it `projectDir` instead of `baseDir`, since the latter is a deprecated alias for the former

jorgee: The problem that I had is that I am not able to get the `baseDir` from the `ModuleResolver` and `ResolveIncludeVisitor` calls at nf-lang. In Nextflow, we get from `Global` and this is why I moved it to the SPI implementation. How do we expect to get workspace roots from the language server? is it an env var, system property?

jorgee: Looking at the Language Server, nf-lang and nextflow code, it looks like the `Compiler` is the class that is shared between the `ModuleResolver`, `ResolveIncludeVisitor` and the `ScriptService` at Language Server. I think the best could be to add a `projectDir` in the `Compiler` where we set the `Session.baseDir` at nextflow execution and to the workspace root at the Language Server. I am going to try to create another PR with this change

bentsherman: My thinking was to have `ResolveIncludeVisitor` accept `projectDir` as a field, since the language server calls this visitor directly. But if it's cleaner to add it to `Compiler` then that's fine too

pditommaso: Random thought tho not strictly related, it would be useful to make the lang server aware of project dir and work dir. @robsyme (or @edmundmiller?) was reporting some limitation with the lsp integration with claude

bentsherman: It is aware of the project dir, so it will be able to provide it to the module resolver just like the runtime

jorgee: @bentsherman, I have implemented adding `projectDir` in `ResolveIncludeVisitor` and `ModuleResolver`, it was simpler than passing through the `Compiler`

Issue #6885, opened 2026-03-04
nf-tower plugin blocks JVM exit after workflow completion — missing HTTP read timeout


adamrtalbot:
### Expected behavior and actual behavior

**Expected**: After a workflow completes successfully, the Nextflow JVM should exit cleanly, allowing the Azure Batch head task to terminate and release the node.

**Actual**: The JVM hangs indefinitely after workflow completion. The `TowerClient.onFlowComplete()` method blocks the main thread on an HTTP PUT call (`sendHttpMessage(urlTraceComplete, ...)`) to the Tower API that never receives a response. The JVM has been stuck for 35+ hours with 0% CPU. The workflow shows as COMPLETE on Seqera Platform but the Azure Batch task remains in active/running state forever.

### Steps to reproduce the problem

Run any pipeline via Seqera Platform on Azure Batch. The issue is timing-dependent — it occurs when the Tower API connection becomes stale during the shutdown HTTP call. Reproduction requires a network condition where the TCP connection is established but the response is never delivered.

### Program output

**Last lines of Nextflow log** — log goes silent after `TimelineObserver`, the next observer (`TowerClient`) never produces output:
```
Mar-03 00:33:40.068 [main] DEBUG nextflow.trace.WorkflowStatsObserver - Workflow completed
Mar-03 00:33:40.068 [main] DEBUG nextflow.trace.TimelineObserver - Workflow completed -- rendering execution timeline
<EOF — no further output, no "Session destroyed", no System.exit()>
```

**Thread state from the stuck node** (via `/proc`, 35+ hours after completion):
```
TID=4474   name=java               state=S (sleeping)    ← main thread blocked
TID=4821   name=tower-logs-chec    state=S (sleeping)    ← next observer never called
TID=4645   name=HttpClient-1-Se    state=S (sleeping)    ← Java HttpClient threads alive
(Tower-thread ABSENT — sender exited cleanly, sender.join() is not the issue)
```

**Network state from the stuck node**:
```
ESTAB 0 0  10.0.0.4:59344 → 13.41.18.99:443  (pid=4474)   ← Tower API
```
No TCP keepalive timers, no retransmission timers — stale connection sitting idle for 35+ hours. Tower API health check responds `200` in `0.25s` — the issue is specific to this stale connection.

### Environment

* Nextflow version: 25.10.4 (build 11173)
* Java version: Amazon Corretto 21.0.10+7-LTS
[...]

PR #6939, opened 2026-03-18
Fix LogsCheckpoint thread interrupt race condition

### LINKED ISSUE #7096: LogsCheckpoint thread re-sets interrupt flag in await(), causing empty console output files on GCS
## Bug report 

#### Root Cause
LogsCheckpoint.await() re-sets the interrupt flag after catching InterruptedException:                                                                           
   
 ```
 catch (InterruptedException e) {                                                                                                                                 
      log.trace "Interrupted logs checkpoint thread"        
      Thread.currentThread().interrupt()  // ← re-sets flag before saveFiles() runs
  }
```

  When Thread.sleep() throws InterruptedException, the JVM automatically clears the interrupt flag. Re-setting it here means saveFiles() is called with the flag set, causing all NIO writes to immediately throw ClosedByInterruptException.


_#### Expected behavior_
                                                                                                                                                                   
After a workflow completes successfully, the nf-*.txt console output, nf-*.log debug log, and timeline-*.html files are uploaded to the GCS work directory with their full contents. The Seqera Platform UI shows the run's console output correctly.                                                                            
   
 _#### Actual behavior_                                                                                                                                              
                                                            
The nf-*.txt file is present in the GCS work directory but is 0 bytes. The Platform UI shows "No data available" on the run's console output tab. The nf-*.log and timeline-*.html uploads also fail silently (WARN-level ClosedByInterruptException stack traces visible in the Nextflow debug log).

1. Launch any pipeline from Seqera Platform using a Google Batch CE with a GCS work directory and cloudcache.enabled = true
2. Allow the pipeline to complete successfully
3. Check the nf-*.txt file in the GCS work directory — it will be 0 bytes                                                                                        
4. In the Platform UI the run's console output shows "No data available"

### Program output 

```
May-04 17:26:56.123 [tower-logs-checkpoint] DEBUG i.seqera.tower.plugin.LogsCheckpoint - Interrupted logs checkpoint thread                                      
  May-04 17:26:56.408 [tower-logs-checkpoint] WARN  io.seqera.tower.plugin.LogsHandler - Unable to upload nextflow log file: /.nextflow/cache/nf-2zptHCqNUEko7Y.log
   — reason: Unknown Error                                                                                                                                         
  com.google.cloud.storage.StorageException: Unknown Error                                                                                                         
      ...                                                                                                                                                          
      Suppressed: com.google.cloud.storage.RetryContext$RetryBudgetExhaustedComment: Unretryable error (attempts: 1, maxAttempts: 10, ...)                         
  Caused by: java.nio.channels.ClosedByInterruptException: null                                                                                                    
      at java.base/java.nio.channels.spi.AbstractInterruptibleChannel.end(AbstractInterruptibleChannel.java:200)                                                   
      ...                                                                                                                                                          
      at io.seqera.tower.plugin.LogsHandler.saveFiles(LogsHandler.groovy:86)
      at io.seqera.tower.plugin.LogsCheckpoint.run(LogsCheckpoint.groovy:74)                                                                                       
                                                                                                                                                                   
  May-04 17:26:56.565 [tower-logs-checkpoint] WARN  io.seqera.tower.plugin.LogsHandler - Unable to upload nextflow timeline file:                                  
  /.nextflow/cache/timeline-2zptHCqNUEko7Y.html — reason: Unknown Error                                                                                            
  com.google.cloud.storage.StorageException: Unknown Error                                                                                                         
      ...                                                   
  Caused by: java.nio.channels.ClosedByInterruptException: null
```
                                                                                                                                                                  

### Environment 

* Nextflow version: 25.10.4 (build 11173, 10-02-2026) 
* Java version: openjdk 21.0.1 2023-10-17
* Operating system: macOs (running on Google Batch CE)

### Additional context

Issue Investigation from Support Ticket submission

pditommaso:
## Summary
- Fix race condition in `LogsCheckpoint.run()` where `saveFiles()` was called with the thread interrupt flag already set after `await()` caught an `InterruptedException`
- This caused the AWS SDK to throw `AbortedException` ("Thread was interrupted") during S3 uploads at workflow shutdown
- Move the interrupt check to after `await()` returns and before `saveFiles()`, so the loop exits cleanly on shutdown

## Test plan
- [ ] Verify no `AbortedException` warnings in logs when running workflows with Tower/Seqera platform integration and remote work directory

### DIFF
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsCheckpoint.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsCheckpoint.groovy
@@ -68,8 +68,10 @@ class LogsCheckpoint implements TraceObserverV2 {
     protected void run() {
         log.debug "Starting logs checkpoint thread - interval: ${interval}"
         try {
-            while( !Thread.currentThread().isInterrupted() ) {
+            while( true ) {
                 await(interval)
+                if( Thread.currentThread().isInterrupted() )
+                    break
                 handler.saveFiles()
             }
         }
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsCheckpoint.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsCheckpoint.groovy
@@ -39,6 +39,7 @@ class LogsCheckpoint implements TraceObserverV2 {
     private Thread thread
     private Duration interval
     private LogsHandler handler
+    private final Object lock = new Object()
 
     @Override
     void onFlowCreate(Session session) {
@@ -55,13 +56,17 @@ class LogsCheckpoint implements TraceObserverV2 {
 
     @Override
     void onFlowComplete() {
-        thread.interrupt()
+        synchronized(lock) {
+            thread.interrupt()
+        }
         thread.join()
     }
 
     @Override
     void onFlowError(TaskEvent event) {
-        thread.interrupt()
+        synchronized(lock) {
+            thread.interrupt()
+        }
         thread.join()
     }
 
@@ -72,7 +77,11 @@ class LogsCheckpoint implements TraceObserverV2 {
                 await(interval)
                 if( Thread.currentThread().isInterrupted() )
                     break
-                handler.saveFiles()
+                synchronized(lock) {
+                    if( Thread.currentThread().isInterrupted() )
+                        break
+                    handler.saveFiles()
+                }
             }
         }
         finally {


bentsherman:
I feel like we have iterated on this code a dozen times trying to find the right logic :sweat_smile:

bentsherman:
@jorgee requesting your review since you are looking into #6885 (not exactly the same but seems related)

jorgee (changes requested):
The PR reduced the probability of the race condition but I think there is still a chance that the thread is interrupted during saving files. To fix it, we could add a `synchronized` block for `handler.savefiles()` and `thread.interrupt()` in `onFlowComplete()` and `onFlowError()`.

pditommaso:
Please do

jorgee (approved):
Added a lock between interrupt and save files to ensure log checkpoint thread is never interrupted during `handler.saveFiles()`.

PR #6545, opened 2025-11-06
Add platform-related metadata to WorkflowRun lineage record

jorgee:
close #5077

This PR retrieves the platform metadata during the `onWorkflowCreate` and stores in the `platform` property in the `WorkflowMetadata` class. The lineage module has been extended, adding a `metadata` field in the `WorkflowRun` record, including the platform metadata.

This PR includes the changes of #6069.

To test the whole integration, it is required to generate an nf-launcher image with the nextflow changes as in e2e tests. Then, use `tw launch` to run a workflow including the options to use a modified launcher image (--launcher-image) and a `nextflow.config` file (--config) with the following lineage configuration

```nextflow
lineage.enabled=true
lineage.store.location='s3://<bucket>/lineage'
```
Use a `<bucket>` associated with the Compute Environment that is running the pipeline to ensure you have the proper permissions.

Once finished, you can use the `nextflow lineage` commands to inspect the workflow lineage records.

Example of the generated metadata extracted form a test execution:
```json
"metadata" : {
    ...
    "platform": {
        "workflowId": "4mjnvcrG15bCMp",
        "user": {
          "id": "13898",
          "userName": "jorge-ejarque",
          "organization": null
        },
        "workspace": {
          "id": "20121123931769",
          "name": "jorge-test-workspace",
          "fullName": "Jorge test workspace",
          "organization": "test-org-jorge"
        },
        "computeEnv": {
          "id": "3xjr2I8RofsG5Kx0pjZQoQ",
          "name": "test-ce",
          "platform": "aws-cloud"
        },
        "pipeline": {
          "id": "51175491284294",
          "name": "https://github.com/nextflow-io/hello",
          "revision": "master",
          "commitId": null
        },
      },
}
```

TODO:

- [x] Check if version upgrade is required

PR #6946, opened 2026-03-19
Add Seqera NIO filesystem for datasets and refactor TowerClient/TowerObserver

jorgee:
## Summary

- Implements a `seqera://` NIO `FileSystemProvider` in `nf-tower`, enabling Nextflow pipelines to reference Seqera Platform datasets as standard file paths (e.g. `seqera://org/workspace/datasets/name`)
- Path hierarchy: root → org → workspace → resource type → dataset file (with optional `@version` pinning)
- Refactors `TowerClient` into two classes: `TowerClient` (pure HTTP API client) and `TowerObserver` (workflow telemetry via `TraceObserverV2`), so the API client can be reused by the new filesystem without pulling in observer lifecycle. **This is done to allow using the FS without requiring to create an observer.**
- Merges `TowerCommonApi` into `TowerClient`, which is now the natural home for shared API methods
- Registers via `META-INF/services/java.nio.file.spi.FileSystemProvider`

## New files

| File | Purpose |
|------|---------|
| `TowerObserver` | Extracted `TraceObserverV2` implementation (task events, heartbeats, workflow lifecycle) formerly in `TowerClient` |
| `dataset/SeqeraDatasetClient` | API calls: list orgs/workspaces/datasets/versions, download |
| `fs/SeqeraPath` | `Path` implementation with 0–4 depth hierarchy |
| `fs/SeqeraFileSystem` | `FileSystem` with lazy org/workspace/dataset caches |
| `fs/SeqeraFileSystemProvider` | `FileSystemProvider` SPI: read, write, list, attributes, copy |
| `fs/SeqeraFileAttributes` | `BasicFileAttributes` backed by dataset metadata |
| `fs/SeqeraPathFactory` | Nextflow `PathFactory` integration |
| `fs/ResourceTypeHandler`, `fs/DatasetsResourceHandler` | Extensibility interface for future resource types |
| `fs/DatasetInputStream`, `fs/DatasetOutputStream` | Stream wrappers for dataset read/write |
| `exception/ForbiddenException`, `exception/NotFoundException` | HTTP error types for API responses |

## Changes to existing files

| File | Change |
|------|--------|
| `TowerClient` | Stripped of observer logic; now a pure API client. Added public `sendApiRequest()` + GET support in `makeRequest()`. Absorbed `TowerCommonApi` methods. |
| `TowerCommonApi` (deleted) | Methods merged into `TowerClient` |
| `TowerFactory` | Creates `TowerObserver` and `TowerClient` separately. `client()` now also activates when `accessToken` is present, so `seqera://` paths work without `tower.enabled` |
| `TowerPlugin` | Registers `SeqeraPathFactory` |
| `BaseCommandImpl`, `AuthCommandImpl`, `LaunchCommandImpl` | Updated to use the refactored `TowerClient` API |
| Tests | Split accordingly: `TowerClientTest` for API client, `TowerObserverTest` for observer; new tests for all `fs/` and `dataset/` classes |

## Test plan

- [x] `SeqeraPathTest` — path parsing, URI round-trips, relativize/resolve, `getFileName`, `asUri`
- [x] `SeqeraFileSystemTest` — cache loading, workspace/dataset resolution, thread safety
- [x] `SeqeraFileSystemProviderTest` — newInputStream (latest + pinned version), readAttributes, newDirectoryStream, error propagation
- [x] `SeqeraDatasetClientTest` — API URL construction, response mapping, error handling
- [x] `TowerObserverTest` — extracted observer logic works identically
- [x] `TowerClientTest` — API client methods after refactor

```
./gradlew :plugins:nf-tower:test
```


jorgee:
Some comments about current implementation:
- Some refactoring is needed to decouple the Tower client (API calls) from the Observer. A client initialization must be done at filesystem initialization and another at observer. Due to token refresh, HxClient must be shared to avoid authentication problems when using different clients.  I would like to make it after merging #6545
- The Dataset API does not allow streaming the content, so read and write are done through temporary files. 
- Only `csv `and `tsv` extensions are allowed; the format is recognized by the extension.
- Due to the above comments, I am considering making it read-only
- Every change in the dataset creates a new version `seqera://org/workspace/datasets/name`  accesses the latest version and `seqera://org/workspace/datasets/name@version`

tests/seqera-dataset.nf
pditommaso, line 1: Let's use io.seqera:tower-api instead of creating all of these

jorgee: Done

pditommaso: Thanks!

jorgee:
Updated to the latest changes in master. Ready for review. It is implemented as read-only FS
- The dataset path: `seqera://<org>/<workspace>/datasets/<name>`
- You can use `nexflow fs` command to browse datasets in Seqera Platform 
```
# List orgs
$ nextflow fs ls seqera://*
seqera-academy
nf-core
seqeralabs
community

#List worspaces
$ nextflow fs ls seqera://nf-core/*
AWSmegatests

# List avaialable items in the workspace (currently just datasets)
$ nextflow fs ls seqera://nf-core/AWSmegatests/*
datasets

# List available datasets
$ nextflow fs ls seqera://nf-core/AWSmegatests/datasets/*
GM_nascent
feb-16-test-6
methylseq_test_full
node-red-tests
proteinfamilies
rnaseq_samplesheet_full
test_rnaseq

# show content of a dataset
$ nextflow fs cat seqera://nf-core/AWSmegatests/datasets/test_rnaseq
sample,fastq_1,fastq_2,strandedness
WT_REP1,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357070_1.fastq.gz,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357070_2.fastq.gz,reverse
WT_REP1,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357071_1.fastq.gz,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357071_2.fastq.gz,reverse
WT_REP2,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357072_1.fastq.gz,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357072_2.fastq.gz,reverse
RAP1_UNINDUCED_REP1,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357073_1.fastq.gz,,reverse
RAP1_UNINDUCED_REP2,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357074_1.fastq.gz,,reverse
RAP1_UNINDUCED_REP2,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357075_1.fastq.gz,,reverse
RAP1_IAA_30M_REP1,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357076_1.fastq.gz,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357076_2.fastq.gz,reverse

# Download a dataset
$ nextflow fs cp seqera://nf-core/AWSmegatests/datasets/test_rnaseq test_rnaseq.csv
$ cat test_rnaseq.csv 
sample,fastq_1,fastq_2,strandedness
WT_REP1,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357070_1.fastq.gz,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357070_2.fastq.gz,reverse
WT_REP1,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357071_1.fastq.gz,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357071_2.fastq.gz,reverse
WT_REP2,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357072_1.fastq.gz,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357072_2.fastq.gz,reverse
RAP1_UNINDUCED_REP1,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357073_1.fastq.gz,,reverse
RAP1_UNINDUCED_REP2,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357074_1.fastq.gz,,reverse
RAP1_UNINDUCED_REP2,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357075_1.fastq.gz,,reverse
RAP1_IAA_30M_REP1,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357076_1.fastq.gz,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357076_2.fastq.gz,reverse
```

- A data set can be used in a pipeline as the following (only for read access at this moment)

```
params.dataset = 'seqera://seqeralabs/showcase/datasets/sarek_samples'
process TEST {
        input:
                path(file)
        output:
                stdout
        script:
        """
        cat $file
        """
}

workflow {
        TEST(file(params.dataset)).view()
}
```

bentsherman (approved):

pditommaso:
Pulling @jordeu to be sure this is aligned with Fusion

bentsherman:
@jorgee can you write a small ADR describing the seqera filesystem hierarchy? that way we can make sure Fusion and Nextflow are aligned more easily

pditommaso (changes requested):
⚠️ **Agent-generated review** — worth looking into these points.

---

## Critical

**1. Dead `getAccessToken()` in TowerObserver** (`TowerObserver.groovy:251-255`)
No `accessToken` field exists in `TowerObserver` — Groovy will resolve it as a null dynamic property, so this always throws. It's dead code left over from the split. Remove it.

**2. Dead `loadSchema()` in TowerObserver** (`TowerObserver.groovy:537-546`)
Identical to `TowerClient.loadSchema()` and never called from TowerObserver. Remove.

**3. `newFileSystem()` violates NIO contract** (`SeqeraFileSystemProvider.groovy:76-88`)
Per `FileSystemProvider.newFileSystem()`, if a FS already exists for the URI, it must throw `FileSystemAlreadyExistsException`. Currently it silently returns the existing instance. The `getOrCreateFileSystem()` method already handles the "get or create" pattern correctly.

---

## Important

**4. `getFileSystem()` is not synchronized** (`SeqeraFileSystemProvider.groovy:91`)
`newFileSystem()` and `getOrCreateFileSystem()` are `synchronized`, but `getFileSystem()` reads from the `LinkedHashMap` without synchronization — data race. Either synchronize it or switch to `ConcurrentHashMap`.

**5. `resolveDataset()` thread safety gap** (`SeqeraFileSystem.groovy:186`)
Calls `synchronized resolveDatasets()` which returns the mutable list from `datasetCache`, then iterates it outside the lock. Concurrent `invalidateDatasetCache()` could cause `ConcurrentModificationException`. Either synchronize `resolveDataset()` or return a defensive copy from `resolveDatasets()`.

**6. `SeqeraPath.iterator()` returns absolute paths instead of name components** (`SeqeraPath.groovy:386-394`)
NIO contract says `Path.iterator()` yields relative name elements (`acme`, `research`, `datasets`, `samples`). This implementation returns increasingly deeper absolute paths. Will break code relying on standard `Path.iterator()` semantics.

**7. `startsWith()`/`endsWith()` use string comparison** (`SeqeraPath.groovy:243-259`)
`seqera://acme-corp`.startsWith(`seqera://acme`) → `true`, which is wrong. NIO expects component-wise comparison.

**8. Error message copy-paste bug in `AuthCommandImpl.deleteTokenViaApi`**
```groovy
final error = response.message ?: "HTTP ${response.message}"
```
The fallback should use `response.code`, not `response.message` again.

---

## Suggestions

**9. `DatasetInputStream.size()` returns -1** (`DatasetInputStream.groovy:60`)
Not a valid `SeekableByteChannel.size()` return value. Also, `SeqeraFileAttributes.size()` always returns 0 — `Files.size()` will report 0 for all datasets, which could break Nextflow's file staging/progress logic.

**10. `fsKey()` always returns just the scheme** (`SeqeraFileSystemProvider.groovy:292-295`)
One FS per JVM is the intent, but if different endpoints/tokens are ever needed this will silently share the wrong FS. At minimum document the limitation.

**11. `ResourceTypeHandler` / `DatasetsResourceHandler` are dead code**
`SeqeraFileSystemProvider` handles datasets inline and doesn't use these strategy classes. Either wire them in or remove until needed.

**12. `DatasetOutputStream` references unimplemented `uploadDataset()`**
`SeqeraDatasetClient.uploadDataset()` throws `UnsupportedOperationException`. Since the FS is read-only, `DatasetOutputStream` is unreachable but should be removed or clearly gated.

**13. Unrelated changes bundled**
`ScriptResolveVisitor.java` (Record/Tuple input resolution) and `CmdModuleCreate.groovy` (namespace validation) are unrelated to this feature — ideally separate PRs for bisectability.

**14. Full memory buffering on download** (`SeqeraDatasetClient.groovy:138`)
Entire dataset loaded as String then converted to `ByteArrayInputStream` — ~40 MB heap per concurrent download for large datasets. Known limitation per PR description, but worth a TODO.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
```diff
@@ -0,0 +1,328 @@
...
+        if (d >= 1 && !fs.listOrgNames().contains(sp.org))
+            throw new NoSuchFileException("seqera://${sp.org}", null, "Organisation not found")
+        if (d >= 2)
+            fs.resolveWorkspaceId(sp.org, sp.workspace)
+        if (d >= 3 && sp.resourceType != 'datasets')
+            throw new NoSuchFileException("seqera://${sp.org}/${sp.workspace}/${sp.resourceType}", null, "Unsupported resource type")
+    }
+
+    private static DatasetVersionDto resolveVersion(SeqeraFileSystem fs, DatasetDto dataset, String pinnedVersion) throws IOException {
+        final versions = fs.client.listVersions(dataset.id, dataset.workspaceId)
```
jordeu: `resolveVersion()` calls `fs.client.listVersions()` on every `newInputStream()` invocation without any caching. Unlike orgs/workspaces and datasets, which are cached in `SeqeraFileSystem`, dataset versions are fetched from the API each time. In a pipeline that reads the same dataset across multiple processes, this results in N redundant `GET /datasets/{id}/versions` calls. Consider adding a version cache in `SeqeraFileSystem` (similar to `datasetCache`) or at least a short-lived/per-dataset memoization.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/DatasetsResourceHandler.groovy
```diff
@@ -0,0 +1,58 @@
...
+import io.seqera.tower.model.DatasetDto
+
+/**
+ * {@link ResourceTypeHandler} implementation for Seqera Platform datasets.
+ * Lists datasets in a workspace and delegates I/O to {@link SeqeraFileSystemProvider}.
+ *
+ * @author Seqera Labs
+ */
+@CompileStatic
+class DatasetsResourceHandler implements ResourceTypeHandler {
```
jordeu: `ResourceTypeHandler` and `DatasetsResourceHandler` are defined but never imported, instantiated, or referenced outside their own files. The provider hardcodes `'datasets'` directly in `newDirectoryStream()` (depth 2 listing), `validateDirectoryExists()`, and the depth-3/4 logic in `newInputStream()`. If the intent is to support future resource types, the dispatching through handlers should be wired in now. Otherwise, consider removing these classes and adding the abstraction when a second resource type actually arrives.

jordeu (commented):
Overall functionality looks good to me

jorgee:
All review points have been addressed in commits `3520d6631` (ADR) and `9b5d99aeb` (code changes):

### Critical (1–3)
- **1 & 2**: Removed dead `getAccessToken()` and `loadSchema()` from `TowerObserver`
- **3**: `newFileSystem()` now throws `FileSystemAlreadyExistsException` per NIO contract (+ test)

### Important (4–8)
- **4**: `getFileSystem()` is now `synchronized`
- **5**: `resolveDataset()` is now `synchronized`
- **6**: `SeqeraPath.iterator()` returns relative name components via `getName(i)` (+ tests)
- **7**: `startsWith()`/`endsWith()` rewritten with component-wise comparison (+ regression tests for the `acme-corp` vs `acme` case)
- **8**: Fixed copy-paste bug: `response.message` → `response.code` in `deleteTokenViaApi`

### Suggestions (9–14)
- **9**: `DatasetInputStream.size()` now throws `UnsupportedOperationException` (+ updated test)
- **10**: Documented `fsKey()` single-FS-per-JVM limitation
- **11 & 12**: Removed dead code — `DatasetOutputStream`, `ResourceTypeHandler`, `DatasetsResourceHandler`, `uploadDataset()` stub
- **13**: Unrelated changes noted — will consider separating in future
- **14**: Replaced full memory buffering with streaming download via new `TowerClient.sendStreamingRequest()` → `HxClient.sendAsStream()`. Same retry/auth chain, same exception types. `SeqeraDatasetClient.downloadDataset()` catches `UnauthorizedException`/`ForbiddenException`/`NotFoundException` and converts to NIO exceptions (`AbortOperationException`/`AccessDeniedException`/`NoSuchFileException`)

### Additional improvements
- **URI encoding**: Fixed filename encoding — replaced `URLEncoder.encode()` (form encoding, `+` for spaces) with `new URI(null, null, fileName, null).rawPath` (RFC 2396 path encoding, `%20`) (Dataset filename with spaces were failing because URL encode translate to + and API was expecting %20
- **Version caching**: Added `versionCache` in `SeqeraFileSystem` so `resolveVersions()` doesn't hit the API on every `newInputStream()` call. Cache invalidated together with dataset cache
- **ADR**: Added `adr/20260310-seqera-dataset-filesystem.md` documenting the feature architecture
- **Stale comment**: Updated `SeqeraPathFactory` comment referencing removed `newOutputStream`

jordeu (approved):

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/DatasetsResourceHandler.groovy
jorgee: Yeah, it was though for extension. Removed.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
jorgee: cache added in 9b5d99a

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
```diff
@@ -0,0 +1,331 @@
...
+ * Follows the {@code LinFileSystemProvider} pattern for structure.
+ * Write support follows the {@code AzFileSystemProvider} buffered-upload pattern.
+ *
+ * @author Seqera Labs
+ */
+@Slf4j
+@CompileStatic
+class SeqeraFileSystemProvider extends FileSystemProvider {
+
+    static final String SCHEME = 'seqera'
```
pditommaso: use visibility accessor `private` or `public` otherwise groovy makes a getter for it

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
```diff
@@ -0,0 +1,331 @@
...
+
+    /** One filesystem per (endpoint + accessToken) key */
+    private final Map<String, SeqeraFileSystem> fileSystems = new LinkedHashMap<>()
```
pditommaso: Any potential race condition with this? may ConcurrentHashMap be better

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
```diff
@@ -0,0 +1,331 @@
...
+    String getScheme() { SCHEME }
+
+    // ---- FileSystem lifecycle ----
+
+    @Override
+    synchronized FileSystem newFileSystem(URI uri, Map<String, ?> env) throws IOException {
+        checkScheme(uri)
+        final key = fsKey(uri, env)
+        if (fileSystems.containsKey(key))
+            throw new FileSystemAlreadyExistsException("seqera:// filesystem already exists")
```
pditommaso:
```suggestion
            throw new FileSystemAlreadyExistsException("File system `seqera://` already exists")
```

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
```diff
@@ -0,0 +1,331 @@
...
+        final TowerClient towerClient = TowerFactory.client()
+        if (!towerClient)
+            throw new IllegalStateException("seqera:// paths require tower.accessToken to be configured")
```
pditommaso:
```suggestion
            throw new IllegalStateException("File system `seqera://` requires the Seqera Platform access token to be provider - use `tower.accessToken config option or TOWER_ACCESS_TOKEN env variable")
```

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
```diff
@@ -0,0 +1,331 @@
...
+        return new SeqeraPath(fs, uri.toString())
+    }
+
+    // ---- Read operations ----
+
+    @Override
+    InputStream newInputStream(Path path, OpenOption... options) throws IOException {
+        final sp = toSeqeraPath(path)
+        if (sp.depth() != 4)
+            throw new IllegalArgumentException("newInputStream requires a dataset path (depth 4): $path")
```
pditommaso:
```suggestion
            throw new IllegalArgumentException("Operation `newInputStream` requires a dataset path (depth 4): $path")
```

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
```diff
@@ -0,0 +1,331 @@
...
+        final dataset = fs.resolveDataset(workspaceId, sp.datasetName)
+        final version = resolveVersion(fs, dataset, sp)
+        log.debug "Downloading dataset '${sp.datasetName}' version ${version.version} (${version.fileName}) from workspace $workspaceId"
+        return fs.client.downloadDataset(dataset.id, String.valueOf(version.version), version.fileName, dataset.workspaceId)
+    }
+
+    @Override
+    SeekableByteChannel newByteChannel(Path path, Set<? extends OpenOption> options, FileAttribute<?>... attrs) throws IOException {
+        if (options?.contains(StandardOpenOption.WRITE) || options?.contains(StandardOpenOption.APPEND))
+            throw new UnsupportedOperationException("seqera:// filesystem is read-only")
```
pditommaso:
```suggestion
            throw new UnsupportedOperationException("File system `seqera://` is read-only")
```

Let's use this pattern consistently

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
```diff
@@ -0,0 +1,331 @@
...
+        }
+        // Dataset file
+        final workspaceId = fs.resolveWorkspaceId(sp.org, sp.workspace)
+        final dataset = fs.resolveDataset(workspaceId, sp.datasetName)
+        return (A) new SeqeraFileAttributes(dataset)
+    }
+
+    @Override
+    Map<String, Object> readAttributes(Path path, String attributes, LinkOption... options) throws IOException {
+        throw new UnsupportedOperationException("readAttributes(String) not supported by seqera:// filesystem")
```
pditommaso:
```suggestion
        throw new UnsupportedOperationException("Operation `readAttributes(String)` not supported by `seqera://` file system")
```


Let's use `Operation` to avoid beginning the sentence with an identifier

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
```diff
@@ -0,0 +1,331 @@
...
+
+    @Override
+    void copy(Path source, Path target, CopyOption... options) throws IOException {
+        toSeqeraPath(source)
+        if (target instanceof SeqeraPath)
+            throw new UnsupportedOperationException("seqera:// filesystem is read-only")
+        // cross-provider (seqera → local): stream to target
+        newInputStream(source).withCloseable { InputStream is ->
+            Files.copy(is, target, StandardCopyOption.REPLACE_EXISTING)
+        }
```
pditommaso: it would be preferable the use `try` with auto closable resources  e.g. 

```
try (InputStream = newInputStream(source)) {
  .. 
}
```

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
```diff
@@ -80,139 +57,44 @@ class TowerClient implements TraceObserverV2 {
...
-
-    protected TowerCommonApi commonApi
-
     private Duration readTimeout = TowerConfig.DEFAULT_READ_TIMEOUT
 
     private Duration connectTimeout = TowerConfig.DEFAULT_CONNECT_TIMEOUT
 
-    TowerClient(Session session, TowerConfig config) {
-        this.session = session
+    TowerClient(TowerConfig config, Map env) {
```
pditommaso: Is the `env` required by tests? It would be preferable to avoid passing as argument

plugins/nf-tower/src/main/io/seqera/tower/plugin/BaseCommandImpl.groovy
```diff
@@ -19,44 +19,27 @@ package io.seqera.tower.plugin
...
      */
     @Memoized
-    protected HxClient createHttpClient(String accessToken = null) {
-        return HxClient.newBuilder()
-            .connectTimeout(Duration.ofMillis(API_TIMEOUT_MS))
-            .bearerToken(accessToken)
-            .build()
+    protected TowerClient createTowerClient(String apiUrl, String accessToken) {
+        final env = SysEnv.get()
+        return new TowerClient( new TowerConfig( [accessToken: accessToken, endpoint: apiUrl, httpConnectTimeout: Duration.of(API_TIMEOUT_MS)], env), env)
```
pditommaso: Feel this can be improved adding `TowerClient` constructor or a factory method

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraPath.groovy
```diff
@@ -0,0 +1,483 @@
...
+    }
+
+    @Override
+    Path resolveSibling(String other) {
+        return getParent().resolve(other)
+    }
+
+    @Override
+    Path relativize(Path other) {
+        if (!(other instanceof SeqeraPath))
```
pditommaso:
```suggestion
        if (other !instanceof SeqeraPath)
```

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraPath.groovy
```diff
@@ -0,0 +1,483 @@
...
+            final sb = new StringBuilder('/').append(workspace)
+            if (resourceType) {
+                sb.append('/').append(resourceType)
+                if (datasetName) {
+                    sb.append('/').append(datasetName)
+                    if (version) sb.append('@').append(version)
+                }
+            }
+            uriPath = sb.toString()
+        }
```
pditommaso, line 405: Bit too many nested `if`

pditommaso (changes requested):
It looks like a great progress. I left a bunch of minor notes. 

I'll post also a few ones reported by Claude worth reviewing or just documenting as the first one (in the following comment)

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraPath.groovy
```diff
@@ -0,0 +1,483 @@
...
+        return getParent().resolve(other)
+    }
+
+    @Override
+    Path resolveSibling(String other) {
+        return getParent().resolve(other)
+    }
+
+    @Override
+    Path relativize(Path other) {
```
pditommaso, line 374: **Critical** — `relativize()` only works when `other` is a descendant of `this`. Per the NIO contract, `base.relativize(other)` must produce `..` segments for upward traversal (e.g. relativizing `seqera://org/ws1` against `seqera://org/ws2` should yield `../ws2`). Currently it throws `IllegalArgumentException`. Given the read-only virtual 4-level hierarchy this is unlikely to be hit in practice, but it should at minimum be documented as a known limitation.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
```diff
@@ -0,0 +1,331 @@
...
+    // ---- Copy ----
+
+    @Override
+    void copy(Path source, Path target, CopyOption... options) throws IOException {
+        toSeqeraPath(source)
+        if (target instanceof SeqeraPath)
+            throw new UnsupportedOperationException("seqera:// filesystem is read-only")
+        // cross-provider (seqera → local): stream to target
+        newInputStream(source).withCloseable { InputStream is ->
+            Files.copy(is, target, StandardCopyOption.REPLACE_EXISTING)
```
pditommaso: **Important** — `copy()` unconditionally passes `StandardCopyOption.REPLACE_EXISTING` to `Files.copy`, ignoring the caller's `CopyOption[]` arguments. If the caller does not pass `REPLACE_EXISTING`, the method should not add it.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/DatasetInputStream.groovy
```diff
@@ -0,0 +1,74 @@
...
+    private long position0 = 0L
+    private boolean open = true
+
+    DatasetInputStream(InputStream inputStream) {
+        this.inputStream = inputStream
+    }
+
+    @Override
+    int read(ByteBuffer dst) throws IOException {
+        final bytes = new byte[dst.remaining()]
```
pditommaso: **Important** — Each `read(ByteBuffer dst)` invocation allocates `new byte[dst.remaining()]`. For large datasets read in small chunks this creates excessive GC pressure. Consider a reusable buffer field.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
```diff
@@ -0,0 +1,331 @@
...
+        if (!towerClient)
+            throw new IllegalStateException("seqera:// paths require tower.accessToken to be configured")
+        final client = new SeqeraDatasetClient(towerClient)
+        final seqeraFs = new SeqeraFileSystem(this, client)
+        fileSystems.put(key, seqeraFs)
+        return seqeraFs
+    }
+
+    @Override
+    synchronized FileSystem getFileSystem(URI uri) {
```
pditommaso, line 90: **Important** — `getFileSystem()` does `fileSystems.values().find { true }` — it always returns the first entry regardless of the URI. Combined with `fsKey()` always returning `"seqera"`, the `Map` is effectively a single-entry container. Consider simplifying to a single `SeqeraFileSystem` field instead of a map, or key on the endpoint URL if multi-endpoint support is intended.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraPath.groovy
```diff
@@ -0,0 +1,483 @@
...
+    String toString() {
+        if (!isAbsolute()) return relPath
+        // Return the canonical human-readable representation
+        final d = depth()
+        if (d == 0) return "${SCHEME}://"
+        return toUri().toString()
+    }
+
+    @Override
+    Path toAbsolutePath() { this }
```
pditommaso: **Important** — `toAbsolutePath()` on a relative `SeqeraPath` (where `fs == null`) silently returns `this`. Per the NIO contract it should throw when called on a relative path with no default directory context.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
```diff
@@ -80,139 +57,44 @@ class TowerClient implements TraceObserverV2 {
...
 
-    private String workflowId
-
-    private String watchUrl
-
     private String endpoint
 
-    private ResourcesAggregator aggregator
-
     protected Map<String,String> env = System.getenv()
```
pditommaso: **Important** — `TowerClient` is now shared across threads (observer + filesystem). The `env` field is `protected` and non-final, meaning a subclass or test could reassign it after construction. Consider making it `private final`.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
```diff
@@ -414,156 +244,18 @@ class TowerClient implements TraceObserverV2 {
...
-     * @param handler A {@link TaskHandler} object representing the task submitted
-     * @param trace A {@link TraceRecord} object holding the task metadata and runtime info
-     */
-    @Override
-    void onTaskSubmit(TaskEvent event) {
-        events << new ProcessEvent(trace: event.trace)
-    }
-
     /**
      * Send an HTTP message, when a process has started
```
pditommaso: **Important** — Orphaned Javadoc block referencing `@param handler` / `@param trace` from the old observer code, plus unused imports for `TaskHandler` and `TraceRecord`. These should be cleaned up.

plugins/nf-tower/src/main/io/seqera/tower/plugin/dataset/SeqeraDatasetClient.groovy
```diff
@@ -0,0 +1,199 @@
...
+    SeqeraDatasetClient(TowerClient towerClient) {
+        this.towerClient = towerClient
+    }
+
+    private String getEndpoint() { towerClient.endpoint }
+
+    /**
+     * @return current user info (id, userName, etc.) from GET /user-info
+     */
+    Map<String, Object> getUserInfo() {
```
pditommaso: **Suggestion** — `TowerClient.getUserInfo()` and `SeqeraDatasetClient.getUserInfo()` do the same thing, with the latter adding exception mapping. `AuthCommandImpl` calls the former directly, bypassing the wrapper. Consider standardizing on one path for consistency.

pditommaso (commented):
Review of the Seqera NIO filesystem and TowerClient/Observer refactor. 8 inline comments — 1 critical, 6 important, 1 suggestion.

plugins/nf-tower/src/main/io/seqera/tower/plugin/BaseCommandImpl.groovy
jorgee: Doing this change would imply an important refactor of auth and launch commands. This method is called in several places and the arguments can change depending on the code. For instance, in auth command, it is first called with the login token and later with the configured token. I prefer to leave it out of the scope of the PR.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
jorgee: No, I added because it was provided by the old withEnvironment(). I have reviewed and now this is only used to get the TOWER_REFRESH_TOKEN from the environment, I can just get it from `SysEnv.get` . No need to store the env.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
jorgee: Applied

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
jorgee: Applied (also fixed the typo: "to be provider" → "to be provided")

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
jorgee: Applied

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
jorgee: Applied

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
jorgee: Applied

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraPath.groovy
jorgee: Applied

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
jorgee: Done — added `public` modifier.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
jorgee: Replaced the map with a single `volatile SeqeraFileSystem fileSystem` field — `fsKey()` was always returning the same key, so the map was effectively a single-entry container. The `synchronized` lifecycle methods + `volatile` field handle thread safety.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
jorgee: Done — changed to `try (InputStream is = newInputStream(source)) { ... }` pattern.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraPath.groovy
jorgee: Refactored to a flat list-based segment builder — no nested `if`s anymore.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraPath.groovy
jorgee: Implemented full NIO-compliant `relativize()` with `..` segment support for upward traversal (e.g. `seqera://org/ws1` relativized against `seqera://org/ws2` yields `../ws2`).

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
jorgee: Fixed — `copy()` now passes through the caller's `CopyOption[]` arguments instead of hardcoding `REPLACE_EXISTING`.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/DatasetInputStream.groovy
jorgee: Done — added a reusable `byte[]` buffer field that grows to the max chunk size seen. No per-call allocation.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystemProvider.groovy
jorgee: Replaced the map with a single `volatile SeqeraFileSystem fileSystem` field and removed `fsKey()`. Same as the ConcurrentHashMap comment — the map was always a single-entry container.

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraPath.groovy
jorgee: Fixed — now throws `IllegalStateException` for relative paths per the NIO contract.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
jorgee: Removed the `env` field entirely. It was only used in `setupClientAuth()` for `TOWER_REFRESH_TOKEN` — now reads directly via `SysEnv.get('TOWER_REFRESH_TOKEN')`. Constructor parameter also removed.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
jorgee: Cleaned up — removed the stale `@param handler` / `@param trace` Javadoc block and the unused `TaskHandler` and `TraceRecord` imports.

plugins/nf-tower/src/main/io/seqera/tower/plugin/dataset/SeqeraDatasetClient.groovy
jorgee: Refactored `getUserInfo()` into `getUserId()` — calls `towerClient.getUserInfo()` internally and returns only the user ID. This removes the full method duplication while keeping the exception mapping in one place for the filesystem use case.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
jorgee: Removed `env` field from `TowerClient`

jorgee:
Addressed all the comments, except one, which implies a refactor on `auth` and `launch` commands. I think it is out of the scope of the PR.

plugins/nf-tower/build.gradle
```diff
@@ -57,6 +58,9 @@ dependencies {
     compileOnly 'io.seqera:lib-httpx:2.1.0'
     api 'io.seqera:lib-platform-oidc:0.1.0'
 
+    api('io.seqera:tower-api:1.121.0') {
+        exclude group: 'io.micronaut.servlet'
```
pditommaso, line 62: This is good, it would be nice to patch also on Platform to prevent the leak of unneeded deps

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileAttributes.groovy
```diff
@@ -0,0 +1,87 @@
...
+    boolean isDirectory() { directory }
+
+    @Override
+    boolean isSymbolicLink() { false }
+
+    @Override
+    boolean isOther() { false }
+
+    @Override
+    long size() { 0L }
```
pditommaso, line 83: Could not be returned the real size for files? Is not Platform exposing it?

jorgee: I can't see any `size` field in `DatasetDto` nor `DatasetVersionDto`

plugins/nf-tower/build.gradle
jorgee: Do you refer to `lib-platform-oidc`? I can't see deps in the runtimeClass comming from this lib.

pditommaso: it should tower-api module in platform project

plugins/nf-tower/src/main/io/seqera/tower/plugin/fs/SeqeraFileSystem.groovy
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * Resolve a DatasetDto by name within a workspace.
+     * @throws NoSuchFileException if no dataset with the given name exists
+     */
+    synchronized DatasetDto resolveDataset(long workspaceId, String name) throws NoSuchFileException {
+        final datasets = resolveDatasets(workspaceId)
+        final found = datasets.find { DatasetDto d -> d.name == name }
+        if (!found)
+            throw new NoSuchFileException("seqera:///datasets/${name}", null, "Dataset '${name}' not found in workspace")
```
pditommaso: Why it's using `seqera:///` (3 slashes). Also, is not uri composed as `{orgName}/{workspaceName}/etc` ?

jorgee: Fixed in [da8c511](https://github.com/nextflow-io/nextflow/pull/6946/commits/da8c5118044fdac8a8e16fb4d825e19a05dd5112). I have moved this check and exception throw to caller methods where the can access the whole SeqeraPath

bentsherman:
@jorgee can you fix the DCO ?

jorgee:
DCO failure is in a @pditommaso commit.

pditommaso (approved):

bentsherman:
@jorgee can you just squash the PR to a single commit and rebase?

jorgee:
> @jorgee can you just squash the PR to a single commit and rebase?

Done

PR #6949, opened 2026-03-20
Fix typo in TowerRetryPolicy: maxAttemps -> maxAttempts

### LINKED ISSUE #6948: Typo in TowerRetryPolicy: maxAttemps prevents tower.retryPolicy.maxAttempts from taking effect

Setting `tower.retryPolicy.maxAttempts` in `nextflow.config` should control the number of retry attempts for Platform API requests. Instead, the value is silently ignored and the default (5) is always used.

The cause is a typo in `TowerRetryPolicy.groovy` [line 78](https://github.com/nextflow-io/nextflow/blob/321c57f88ae20ab533cf9c18cb11d641f437a4d5/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerRetryPolicy.groovy#L78):

```groovy
this.maxAttempts = opts.maxAttemps as Integer ?: legacy.maxRetries as Integer ?: RetryConfig.DEFAULT_MAX_ATTEMPTS
```

The map lookup reads `opts.maxAttemps` (missing trailing `t`), but the Nextflow config parser produces a map key `maxAttempts` (correctly spelled). The lookup returns `null`, so the value always falls through to `legacy.maxRetries` or `RetryConfig.DEFAULT_MAX_ATTEMPTS`.

The test at `TowerRetryPolicyTest.groovy:47` also uses the misspelled key (`maxAttemps: 3`), so it passes despite the bug.

Add this to `nextflow.config`:

```groovy
tower {
    enabled = true
    accessToken = '...'
    retryPolicy {
        maxAttempts = 10
    }
}
```

Run any pipeline with `-with-tower`. The retry policy will still use 5 attempts (the default), not 10.

The legacy config key `tower.maxRetries` works correctly as a workaround because it's read from the `legacy` map parameter, not from `opts`.

### Program output

No error is produced — the misconfiguration is silent. The value is simply ignored.

### Environment

* Nextflow version: master (introduced in #6354)
* Java version: any
* Operating system: any

### Additional context

The fix is a one-character change on line 78 of `TowerRetryPolicy.groovy`:

```diff
-        this.maxAttempts = opts.maxAttemps as Integer ?: legacy.maxRetries as Integer ?: RetryConfig.DEFAULT_MAX_ATTEMPTS
+        this.maxAttempts = opts.maxAttempts as Integer ?: legacy.maxRetries as Integer ?: RetryConfig.DEFAULT_MAX_ATTEMPTS
```

The test at `TowerRetryPolicyTest.groovy:47` should also be updated to use the correctly spelled key to verify the fix:

```diff
-                maxAttemps: 3,
+                maxAttempts: 3,
```

robsyme:
## Summary
- Fix misspelled map key `maxAttemps` → `maxAttempts` in `TowerRetryPolicy`, which caused `tower.retryPolicy.maxAttempts` to be silently ignored
- Update corresponding test to use the correct spelling

Fixes #6948

## Test plan
- [x] Existing unit test updated and passes with correct key spelling

### DIFF
--- a/plugins/nf-tower/src/test/io/seqera/tower/plugin/TowerRetryPolicyTest.groovy
+++ b/plugins/nf-tower/src/test/io/seqera/tower/plugin/TowerRetryPolicyTest.groovy
@@ -44,7 +44,7 @@ class TowerRetryPolicyTest extends Specification {
         def customOptions = [
                 delay: '1s' as nextflow.util.Duration,
                 maxDelay: '60s' as nextflow.util.Duration,
-                maxAttemps: 3,
+                maxAttempts: 3,
                 jitter: 0.5,
                 multiplier: 1.5
         ]
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerRetryPolicy.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerRetryPolicy.groovy
@@ -75,7 +75,7 @@ class TowerRetryPolicy implements Retryable.Config, ConfigScope {
     TowerRetryPolicy(Map opts, Map legacy=Map.of()) {
         this.delay = opts.delay as Duration ?: legacy.backOffDelay as Duration ?: RetryConfig.DEFAULT_DELAY
         this.maxDelay = opts.maxDelay as Duration ?: RetryConfig.DEFAULT_MAX_DELAY
-        this.maxAttempts = opts.maxAttemps as Integer ?: legacy.maxRetries as Integer ?: RetryConfig.DEFAULT_MAX_ATTEMPTS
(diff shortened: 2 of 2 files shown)


plugins/nf-tower/src/test/io/seqera/tower/plugin/TowerRetryPolicyTest.groovy
```diff
@@ -44,7 +44,7 @@ class TowerRetryPolicyTest extends Specification {
         def customOptions = [
                 delay: '1s' as nextflow.util.Duration,
                 maxDelay: '60s' as nextflow.util.Duration,
-                maxAttemps: 3,
```
bentsherman, line 47: damn even the test was wrong 😆

robsyme: I had a little chuckle at this one, too :D

PR #6952, opened 2026-03-20
Fix AWS Batch machine type trace for new instance families

FloWuenne:
## Summary

This PR fixes AWS Batch trace machine type propagation for newer EC2 instance families whose names are not yet recognized by the pinned AWS SDK enum list.

Nextflow was resolving the machine type through `instance.instanceType().toString()`, which depends on the generated SDK enum. For newer families such as `r8id.xlarge`, `m8id.xlarge`, and `c8id.xlarge`, the current SDK maps the typed accessor to `UNKNOWN_TO_SDK_VERSION` instead of preserving the raw instance type string.

This change switches the AWS Batch helper to prefer `instance.instanceTypeAsString()` and only fall back to the enum value if needed. That makes the trace path resilient to newly released EC2 instance families and ensures the machine type value can be propagated correctly downstream.

## Changes

- use the raw EC2 instance type string in `AwsBatchHelper`
- keep the existing enum-based value only as a fallback
- add regression coverage for SDK-unknown instance types
- add explicit tests showing the current pinned SDK maps `r8id.xlarge`, `m8id.xlarge`, and `c8id.xlarge` to `UNKNOWN_TO_SDK_VERSION`

## Testing

- `./gradlew :plugins:nf-amazon:test --tests 'nextflow.cloud.aws.batch.AwsBatchHelperTest'`

## Notes

- No AWS SDK version bump is included in this PR. The code change is intentional so that Nextflow remains robust even when AWS releases instance families before they appear in a future SDK enum update.
- This change was implemented with Codex GPT-5.4 assistance.


pditommaso:
I believe the proper solution is to update the aws sdk

FloWuenne:
@pditommaso From codex checking the AWS SDK, these instance types are not yet in the official enum from the newest version (might be wrong but double check).

pditommaso:
Screenshot 2026-03-20 at 21 31 42

pditommaso:
But I agree this prevent updating to latest sdk to have access to latest types.

plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchHelper.groovy
```diff
@@ -166,11 +166,15 @@ class AwsBatchHelper {
...
 
         new CloudMachineInfo(
-                instance.instanceType().toString(),
+                getInstanceType(instance),
                 instance.placement().availabilityZone(),
                 getPrice(instance))
     }
 
+    protected String getInstanceType(Instance instance) {
+        instance.instanceTypeAsString() ?: instance.instanceType()?.toString()
```
pditommaso: May not make more sense ?

```
instance.instanceType()?.toString() ?: instance.instanceTypeAsString()
```

FloWuenne: I don't think so because instance.instanceType()?.toString()  will be UNKNOWN_TO_SDK_VERSION, not null and therefore it's not invalidated and `instance.instanceTypeAsString()` would never run?

Not 100% sure here though!

pditommaso: Got it. i've simplified a bit. 👍

plugins/nf-amazon/src/test/nextflow/cloud/aws/batch/AwsBatchHelperTest.groovy
```diff
@@ -35,19 +36,41 @@ class AwsBatchHelperTest extends Specification {
...
+        and:
+        INSTANCE = Instance.builder().instanceType(InstanceType.fromValue(TYPE)).instanceType(TYPE).build()
+    }
+
+    def 'should map unknown generation 8 instance types to sdk sentinel'() {
+        expect:
+        InstanceType.fromValue('r8id.xlarge') == InstanceType.UNKNOWN_TO_SDK_VERSION
+        InstanceType.fromValue('m8id.xlarge') == InstanceType.UNKNOWN_TO_SDK_VERSION
+        InstanceType.fromValue('c8id.xlarge') == InstanceType.UNKNOWN_TO_SDK_VERSION
+    }
```
bentsherman:
```suggestion
```

This test will break when we update the AWS SDK and the new instance types are defined

The most important thing to test is: when `getInstanceType()` is called with an unknown type, it should return the raw string value (covered by the previous test)

So this test can be removed

PR #6957, opened 2026-03-23
Add consumableResources process directive for AWS Batch

robsyme:
## Summary

- Add `consumableResources` process directive that maps to [AWS Batch resource-aware scheduling](https://docs.aws.amazon.com/batch/latest/userguide/resource-aware-scheduling.html)
- Enables license-seat-aware job scheduling — AWS Batch holds jobs in RUNNABLE until consumable resources are available
- Supports string shorthand (`consumableResources 'my-license'` implies quantity 1), map syntax (`consumableResources 'lic': 2`), and is repeatable across multiple calls

### Usage

```nextflow
process runDragen {
    consumableResources 'my-dragen-license'
    cpus 4
    memory '16 GB'

    script:
    """
    dragen --ref-dir /ref ...
    """
}
```

### Implementation

- **Core module**: Directive registered in `ProcessBuilder.DIRECTIVES`, handler methods for string/map syntax with `ConfigList` accumulation, `TaskConfig.getConsumableResources()` getter normalizing all input forms
- **nf-lang**: `ProcessDsl.DirectiveDsl` updated for config validation
- **nf-amazon plugin**: `AwsBatchTaskHandler.configJobDefRequest()` maps directive to `ConsumableResourceProperties` on `RegisterJobDefinitionRequest` via `RegisterJobDefinitionModel`; included in job definition hash
- **Docs**: Full directive documentation in `docs/reference/process.md` with FIFO queue ordering caveat; added to AWS Batch executor's supported directives list

Closes #5917

## Test plan

- [x] `ProcessBuilderTest` — string shorthand, map syntax, multi-resource, repeatable accumulation
- [x] `TaskConfigTest` — Map normalization, List passthrough, String shorthand, null handling
- [x] `AwsBatchTaskHandlerTest` — single resource, multiple resources, absent directive, hashing, no regression on existing tests
- [x] Full nf-amazon plugin test suite passes
- [x] Core module smoke tests pass (3 pre-existing failures unrelated to this change)
- [x] Live validation with AWS Batch consumable resource

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #5917, opened 2025-03-24
Support for AWS Batch Consumable Resources


exai-sukh:
## New Feature

AWS has introduced a new feature for AWS Batch, called **"Consumable Resources"**. It would be great if Nextflow's AWS executor could support this. You may see additional details in the AWS blog post [How to use rate-limited resources in AWS Batch jobs with resource-aware scheduling](https://aws.amazon.com/blogs/hpc/how-to-use-rate-limited-resources-in-aws-batch-jobs-with-resource-aware-scheduling/).

## Use Case

I have workflows which need staggered submissions to reduce the likelihood of throttling errors. This feature would resolve such issues.

## Suggested Implementation

Specify **ConsumableResourceProperties** within the [process directive]. This could be added to the [SubmitJobRequest](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/batch/model/SubmitJobRequest.html) in https://github.com/nextflow-io/nextflow/blob/7d3cb2b7e094862fe6d37843284951d401372f77/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy#L713

bentsherman:
If it's a new feature then it's likely only available in AWS SDK v2, which we haven't upgraded to yet. We plan to do that this year.

In the meantime, we do have some retry-with-backoff logic to handle rate limit errors. Is that not working for you?

gbillings-profound:
Would love to see this implemented! We have a pipeline using commercially-licensed software in AWS Batch, and our license is for a set number of concurrent processes; implementing consumable resources seems like the cleanest way to manage the license seats efficiently. Other possible strategies are using `process.maxForks` or just capping the number of vCPUs for the compute environment, but they both have significant limitations.

bentsherman:
Looking at the API docs, I guess we could have a process directive like this:

```nextflow
process dragen {
  consumableResources 'my-dragen-license': 1, 'other-license': 2

  // ...
}
```

Nextflow uses AWS SDK v2 now so it should be feasible

gbillings-profound:
Something like that would be ideal for us!

bentsherman:
Resolved in Nextflow 26.04 using the process `hints` directive

PR #6960, opened 2026-03-23
ADR: hints process directive for executor-specific scheduling hints

robsyme:
## Summary

Architecture Decision Record proposing a new `hints` process directive with namespaced keys as a generic extension point for executor-specific scheduling hints.

**First use case:** AWS Batch [consumable resources](https://docs.aws.amazon.com/batch/latest/userguide/resource-aware-scheduling.html) for license-seat-aware scheduling.

## Options Evaluated

1. **Dedicated `consumableResources` directive** — too narrow, each executor feature would need its own directive
2. **Overload `resourceLabels`** — conflates metadata tags with scheduling behavior; `resourceLabels` uses replacement semantics in config overrides, so a sysadmin's cost-tracking labels would be wiped out if a user sets consumable resources (and vice versa)
3. **New `hints` directive** (recommended) — generic, extensible, namespaced keys, executors consume what they understand

## Proposed Syntax

```nextflow
process runDragen {
    hints 'consumable-resource:my-dragen-license': 1
}
```

Ref: #5917, #6957

### DIFF
--- a/adr/20260323-hints-process-directive.md
+++ b/adr/20260323-hints-process-directive.md
@@ -1,136 +1,180 @@
 # `hints` process directive for executor-specific scheduling hints
 
 - Authors: Rob Syme
-- Status: draft
+- Status: accepted
 - Deciders: Paolo Di Tommaso, Ben Sherman, Rob Syme
 - Date: 2026-03-23
 - Tags: directive, executor, scheduling
 
-Technical Story: [nextflow-io/nextflow#5917](https://github.com/nextflow-io/nextflow/issues/5917), [PR #6957](https://github.com/nextflow-io/nextflow/pull/6957)
-
 ## Summary
 
-Introduce a `hints` process directive for executor-specific scheduling hints that don't map to existing resource directives. The first use case is AWS Batch consumable resources for license-seat-aware scheduling.
+Introduce a `hints` process directive for executor-specific scheduling hints that don't map to existing directives.
 
 ## Problem Statement
 
-Users running commercially licensed software (e.g. DRAGEN, Schrodinger) need to limit concurrent job execution based on available license seats — not just within a single pipeline run (`maxForks`), but across multiple concurrent runs. AWS Batch introduced [resource-aware scheduling](https://docs.aws.amazon.com/batch/latest/userguide/resource-aware-scheduling.html) in February 2025, which models this natively via consumable resources. Nextflow has no way to pass this information to the executor.
+Many executors can be configured in various ways on a per-task basis. For example:
 
-More broadly, cloud batch systems expose scheduling knobs that don't fit neatly into Nextflow's existing resource directives (`cpus`, `memory`, `disk`, `accelerator`, `queue`). Today users must resort to `clusterOptions` (string-based, fragile) or custom `ext` attributes (no standard semantics). A structured, namespaced directive would provide a clean extension point.
+- AWS Batch jobs can use *consumable resources* to limit concurrent job execution based on non-standard resources such as software license seats.
 
-## Goals or Decision Drivers
+- Google Batch jobs can specify a *provisioning model* to control the use of spot vs on-demand VMs on a per-task basis.
 
-- Avoid a proliferation of narrow, executor-specific directives (e.g. `consumableResources`, `schedulingPolicy`, etc.)
+- Seqera Scheduler supports a variety of resource and scheduling settings, including spot/on-demand provisioning.
+
+These settings can be exposed by Nextflow as executor-specific config options, such as `google.batch.spot`, but config options are applied globally. In order to apply a setting to specific processes or tasks, it must be exposed as a process directive.
+
+Process directives in Nextflow aim to provide a common vocabulary for executing tasks in many different environments. Directives such as `cpus`, `memory`, and `time` have broadly the same meaning across most executors, making it easier for users to write portable pipelines.
+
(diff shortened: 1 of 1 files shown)


bentsherman:
Related: [WDL requirements and hints](https://openwdl.org/wdl/bioinformatics/workflows/announcing-wdl-1-2-0/#requirements-and-hints)

bentsherman:
Another use case: scheduling priority for AWS Batch: #6998

pditommaso:
Thanks @robsyme for the thorough ADR. The `hints` directive is the right direction — we're all aligned on that. Here's a refined proposal for the naming, syntax, and validation semantics.

## Key format: camelCase with optional dot-scoped categories

Keys use **camelCase** and support an optional **dot-separated scope** to group related hints into logical categories:

```nextflow
process runDragen {
    hints consumableResource: 'my-dragen-license'
    hints 'scheduling.priority': 10
    hints 'scheduling.provisioningModel': 'spot'
}
```

```groovy
// nextflow.config
process {
    withName: 'runDragen' {
        hints = [
            consumableResource: 'my-dragen-license',
            'scheduling.provisioningModel': 'spot'
        ]
    }
}
```

This keeps simple hints flat (`consumableResource`) while letting related hints cluster under a shared scope (`scheduling.priority`, `scheduling.provisioningModel`) — readable without imposing structure where it's not needed.

## Optional executor prefix for targeting

When a hint is specific to a given executor or platform, it can be prefixed with an executor/vendor namespace using **slash notation**:

```nextflow
hints 'aws/consumableResource': 'my-dragen-license'
hints 'seqera/scheduling.provisioningModel': 'spot'
hints 'k8s/scheduling.nodeSelector': 'gpu=true'
```

Full key format: `[executor/][scope.]hintName`

Unprefixed hints are "universal" — any executor that understands them can consume them. Prefixed hints are only consumed by the matching executor. This gives pipeline authors a way to be explicit when portability matters and specific when it doesn't.

## Value parsing

Hint values are always strings or integers at the Nextflow level. **Interpretation and parsing of values is delegated to the implementing executor.** A value could be a simple literal:

```nextflow
hints 'scheduling.provisioningModel': 'spot'
```

Or it could use `=` as a conventional key-value separator when the executor needs structured content:

```nextflow
hints 'k8s/scheduling.nodeSelector': 'gpu=true'
```

This keeps the core directive simple (a `Map<String, Object>`) while giving executors the flexibility to define richer value semantics without burdening the Nextflow config parser.

## Validation

Instead of silently ignoring unknown hints (as the current ADR proposes), we should **validate against known hints** to catch typos and misconfiguration:

- **Unprefixed hints**: Nextflow checks against a registry of known hint keys. An unknown key (e.g. `scheduling.provisioningModle`) produces a **warning** suggesting the closest match — similar to how we handle unknown directives today.
- **Prefixed hints** (e.g. `aws/consumableResource`): The target executor performs a **strict check** — an unrecognized key is an **error**, since the user explicitly declared the executor scope and a typo or unsupported hint should not pass silently.

This two-tier approach balances portability (a pipeline using unprefixed hints can run on executors that don't support them, with a warning) against correctness (a pipeline using executor-prefixed hints gets fast failure if the hint is wrong).

## Accumulation vs. replacement semantics

Config overrides use **replacement by default**, consistent with how other map-type directives (`resourceLabels`, `ext`) work:

```groovy
// This replaces all hints from the process definition
process {
    withName: 'runDragen' {
        hints = ['scheduling.provisioningModel': 'spot']
    }
}
```

For hints that support **accumulation** (i.e. adding to rather than replacing), use the `+` prefix on the key name:

```groovy
// This adds a consumable resource while preserving existing hints
process {
    withName: 'runDragen' {
        hints = ['+consumableResource': 'another-license']
    }
}
```

Not all hints support accumulation — it only makes sense for hints that are inherently multi-valued (like `consumableResource`, where a process might need multiple license types). Scalar hints like `scheduling.provisioningModel` are single-valued by nature — using `+` on them should produce an error.

## Initial hint catalog

For the first implementation, I'd suggest supporting at least:

| Hint key | Value type | Accumulation | Executors | Use case |
|----------|-----------|--------------|-----------|----------|
| `consumableResource` | `String` or `Map<String,Integer>` | Yes | AWS Batch | License-seat-aware scheduling ([#5917](https://github.com/nextflow-io/nextflow/issues/5917)) |
| `scheduling.provisioningModel` | `String` (`'spot'`, `'standard'`) | No | AWS Batch, Google Batch, Azure Batch | Spot/preemptible instance selection ([#4262](https://github.com/nextflow-io/nextflow/issues/4262), [#5850](https://github.com/nextflow-io/nextflow/issues/5850)) |
| `scheduling.priority` | `Integer` | No | AWS Batch | Job scheduling priority ([#6998](https://github.com/nextflow-io/nextflow/issues/6998)) |

As @adamrtalbot noted, K8s node selectors/affinity ([#5570](https://github.com/nextflow-io/nextflow/issues/5570)) are a natural future candidate (e.g. `scheduling.nodeSelector`).

## Summary of changes vs. current ADR

| Aspect | Current ADR | This proposal |
|--------|------------|---------------|
| Key format | Colon-separated (`consumable-resource:name`) | camelCase with optional dot scope (`scheduling.provisioningModel`) |
| Executor scoping | Implicit (executor picks what it knows) | Explicit optional prefix (`aws/hint`) |
| Unknown hints | Silently ignored | Warning (unprefixed) or error (prefixed) |
| Config override | Accumulation by default | Replacement by default, `+` prefix for accumulation |

PR #6966, opened 2026-03-24
Harden ProcessEntryHandler types and tests

pditommaso:
## Summary
- Harden types in ProcessEntryHandler.parseComplexParameters: use explicit Map typing for the traversal variable and a for loop instead of a closure
- Fix potential ClassCastException when a plain param key conflicts with a dotted key (e.g. --foo bar and --foo.id 1 passed together): the instanceof Map check now properly overwrites non-map values instead of crashing
- Add unit test covering the dot-notation conflict scenario

## Rationale
The parseComplexParameters method used untyped def variables and containsKey checks that did not guard against a non-map value already occupying an intermediate key. This hardens the code with proper types and a defensive instanceof check, along with a test to prevent regressions.

## Test plan
- [x] New test: dotted key overwrites conflicting plain value
- [x] Existing tests pass

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java_version: [17, 24]
+        java_version: [17, 25]
 
     steps:
       - name: Checkout
@@ -116,7 +116,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java_version: [17, 24]
+        java_version: [17, 25]
         test_mode: ["test_integration", "test_parser_v2", "test_docs", "test_aws", "test_azure", "test_google", "test_wave"]
     steps:
       - name: Checkout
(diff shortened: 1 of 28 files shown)


modules/nextflow/src/test/groovy/nextflow/script/ProcessEntryHandlerTest.groovy
```diff
@@ -100,6 +100,25 @@ class ProcessEntryHandlerTest extends Specification {
...
+        def script = Mock(BaseScript)
+        def meta = Mock(ScriptMeta)
+        def handler = new ProcessEntryHandler(script, session, meta)
+
+        when:
+        'both --foo bar and --foo.id 1 are passed, the dotted key wins'
+        def result = handler.parseComplexParameters([
+            'foo': 'bar',
+            'foo.id': '1'
+        ])
```
bentsherman, line 115: Should we log a warning then that `--foo` was ignored in favor of `--foo.id` ?

pditommaso: Make sense

PR #6968, opened 2026-03-24
Fix: cast module run CLI params to declared meta.yml input types

### LINKED ISSUE #6964: `nextflow module run` does not coerce boolean-like string values for `val` inputs
### Description

When using `nextflow module run` to run a module that has a `val` input expecting a boolean, passing `false` (or `0`, or an empty string `''`) via the CLI always results in a truthy value inside the process script. This is because CLI arguments are passed as strings, and in Groovy any non-empty string (including `"false"` and `"0"`) evaluates to `true` in a boolean context.

### How to Reproduce

1. Use a module that has a boolean `val` input, e.g. `nf-core/salmon/quant` which declares:

```groovy
input:
val alignment_mode
```

and uses it as:

```groovy
if (alignment_mode) {
    reference = "-t ${transcript_fasta}"
    input_reads = "-a ${reads}"
}
```

2. Configure the module registry in `nextflow.config`:

```groovy
registry {
    url = 'https://registry-dev.nextflow.io/api'
}
docker.enabled = true
wave.enabled = true
wave.strategy = 'conda'
```

3. Run the module with `--alignment_mode false`:

```bash
nextflow module run nf-core/salmon/quant \
    --meta.id test_sample \
    --reads "reads_1.fastq.gz,reads_2.fastq.gz" \
    --index /path/to/salmon/index \
    --gtf /path/to/genes.gtf \
    --transcript_fasta /path/to/transcriptome.fasta \
    --alignment_mode false \
    --lib_type A \
    --outdir results
```

### Observed Behavior

The value `"false"` is passed as a string to the process. Since Groovy's `"false".asBoolean()` returns `true` (non-empty string), the `if (alignment_mode)` block is entered, causing salmon to run in alignment mode (`-a` flag) instead of mapping mode (`--index` flag).

The same happens with `--alignment_mode 0` and `--alignment_mode ''` — all are treated as truthy.

### Expected Behavior

`nextflow module run` should coerce string values `"false"`, `"0"`, and `""` to their boolean/falsy equivalents when feeding them into `val` inputs, so that `--alignment_mode false` results in a falsy value inside the process script.

### Environment

- Nextflow edge (latest)
- macOS

jorgee:
## Summary

Fixes #6964 — `nextflow module run` does not coerce boolean-like string values for `val` inputs.

When running a module via CLI (e.g. `--alignment_mode false`), all parameters arrive as strings. This PR reads the module's `meta.yml` `input` section to find the declared type for each parameter, then validates and casts the value before passing it to the process.

- **`ModuleSpec.loadInputTypes(Path)`** — new static method that parses the `input` section of a `meta.yml` file into a flat `Map<String, String>` (parameter name → type). Supports the new paramSpec format (`{name, type, description}`) and the old nf-core format (`{paramName: {type, description}}`). Tuple inputs are flattened recursively, including the `{tuple: [...]}` map-value variant.
- **`ProcessEntryHandler.loadInputTypesFromMetaYml()`** — resolves `meta.yml` via the script's binding path and delegates to `ModuleSpec.loadInputTypes`. Silently skips if the file is absent or unreadable.
- **`ProcessEntryHandler.validateAndCastParam()`** — validates and casts each top-level parameter against its declared type: `boolean` (`"false"` → `Boolean.FALSE`), `integer`, `float`, `string`, `map`, `file`, `directory`. Throws `IllegalArgumentException` with a clear message on type mismatch.

## Test plan

- [ ] `ModuleSpecTest` — 7 new tests: `loadInputTypes` with missing file, no input section, paramSpec format, old nf-core format, tuple flattening in both formats, malformed YAML
- [ ] `ProcessEntryHandlerTest` — 18 new tests: `validateAndCastParam` for all 6 type cases, passthrough for unknown type
- [ ] Existing tests pass: `./gradlew :nextflow:test --tests "nextflow.module.ModuleSpecTest" --tests "nextflow.script.ProcessEntryHandlerTest"`

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
@@ -240,15 +240,14 @@ class ProcessEntryHandler {
     private Map<String, Class> getModuleSpecInputTypes() {
         try {
             final scriptPath = script.getBinding().getScriptPath()
-            if( scriptPath == null ) 
-                return null
-            final specPath = scriptPath.resolveSibling(ModuleStorage.MODULE_MANIFEST_FILE)
-            return ModuleSpec.loadInputTypes(specPath)
+            final specPath = scriptPath?.resolveSibling(ModuleStorage.MODULE_MANIFEST_FILE)
+            if( specPath )
+                return ModuleSpec.loadInputTypes(specPath)
         }
         catch( Exception e ) {
             log.debug "Could not load input types from module spec: ${e.message}"
-            return null
         }
+        return Collections.emptyMap()
     }
 
     /**
--- a/modules/nextflow/src/main/groovy/nextflow/module/ModuleSpec.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/module/ModuleSpec.groovy
@@ -83,7 +83,7 @@ class ModuleSpec {
      */
     static Map<String, Class> loadInputTypes(Path path) {
         if( !Files.exists(path) )
-            return null
+            return Collections.emptyMap()
 
         Map<String, Object> data
         try( final stream = Files.newInputStream(path) ) {
@@ -91,19 +91,18 @@ class ModuleSpec {
         }
         catch( Exception e ) {
             log.warn "Failed to parse module spec at ${path}: ${e.message}"
-            return null
+            return Collections.emptyMap()
         }
 
         final inputSection = data?.get('input') as List
         if( !inputSection )
-            return null
+            return Collections.emptyMap()
 
         final Map<String, Class> inputTypes = new LinkedHashMap<>()
-        for( final item : inputSection ) {
+        for( final item : inputSection )
             extractInputTypes(item, inputTypes)
-        }
 
-        return inputTypes.isEmpty() ? null : inputTypes
+        return inputTypes
     }
 
     /**
--- a/modules/nextflow/src/test/groovy/nextflow/module/ModuleSpecTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/module/ModuleSpecTest.groovy
@@ -133,12 +133,12 @@ requires:
         '1.0.0.0'       | false
     }
 
-    def 'should return null from loadInputTypes when file does not exist' () {
+    def 'should return empty input types when file does not exist' () {
         expect:
-        ModuleSpec.loadInputTypes(tempDir.resolve('nonexistent.yml')) == null
+        ModuleSpec.loadInputTypes(tempDir.resolve('nonexistent.yml')) == [:]
     }
 
-    def 'should return null from loadInputTypes when no input section' () {
+    def 'should return empty input types when no input section' () {
         given:
         def metaYaml = tempDir.resolve('meta.yml')
         metaYaml.text = '''\
@@ -146,7 +146,7 @@ requires:
             description: FastQC quality control
             '''.stripIndent()
         expect:
-        ModuleSpec.loadInputTypes(metaYaml) == null
+        ModuleSpec.loadInputTypes(metaYaml) == [:]
     }
 
     def 'should load input types in new module spec format' () {
@@ -197,13 +197,13 @@ requires:
         types == [meta: Map, reads: Path, index: Path]
     }
 
-    def 'should return null from loadInputTypes for malformed YAML' () {
+    def 'should return empty input types for malformed YAML' () {
         given:
         def metaYaml = tempDir.resolve('meta.yml')
         metaYaml.text = ': invalid: yaml: {'
 
         expect:
-        ModuleSpec.loadInputTypes(metaYaml) == null
+        ModuleSpec.loadInputTypes(metaYaml) == [:]
     }
 
     def 'should validate module name format' () {
--- a/modules/nextflow/src/test/groovy/nextflow/script/ScriptProcessRunTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/script/ScriptProcessRunTest.groovy
@@ -18,6 +18,7 @@ package nextflow.script
 
 import java.nio.file.Files
 
+import nextflow.Global
 import test.Dsl2Spec
 
 import static test.ScriptHelper.*
@@ -125,5 +126,36 @@ class ScriptProcessRunTest extends Dsl2Spec {
         then:
         def e = thrown(Exception)
         e.message.contains('Missing required parameter: --requiredParam')
+    }
+
+    def 'should cast boolean parameter to boolean' () {
+        given:
+        def folder = Files.createTempDirectory('test')
+
+        def module = folder.resolve('module.nf')
+        module.text = '''\
+            process quant {
+                input:
+                val alignment_mode
+
+                exec:
+                assert alignment_mode instanceof Boolean
+            }
+            '''
+
+        def spec = folder.resolve('meta.yml')
+        spec.text = '''\
+            input:
+            - name: alignment_mode
+              type: boolean
+            '''
+
+        when:
+        runScript(module, config: [params: [alignment_mode: 'false']])
+        then:
+        Global.session.isSuccess()
+
+        cleanup:
+        folder?.deleteDir()
     }
 }


modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
```diff
@@ -198,9 +204,122 @@ class ProcessEntryHandler {
             }
         }
 
+        // Validate and cast based on meta.yml input types if available
+        final typeMap = loadInputTypesFromMetaYml()
```
bentsherman: Can we apply the `meta.yml` types only for v1 inputs?

For v2 inputs it's not needed since the process inputs are already typed

pditommaso: Do you mean using record types?

bentsherman: Not even that, just regular types like Integer, String, Boolean, Path. It is already handled for typed processes, see `getTypedValueForInput()`

The `meta.yml` enrichment is only needed for `val` inputs in v1 processes

pditommaso: For example the `alignment_mode` is declared as Boolean in the spec. Would v2 be aware of that? 

https://github.com/nf-core/modules/blob/6fbc98fa4f4f5976a311f8cf46332dcf38eb1bfc/modules/nf-core/salmon/quant/main.nf#L15

bentsherman: yes, because in a typed process it would be declared as:
```groovy
// legacy process
val alignment_mode

// typed process
alignment_mode: Boolean
```

jorgee: sometimes when referring to v2 is a bit confusing. You mean the typed processes, not to syntax parser V2 which supports both typed and not types processes

pditommaso: > yes, because in a typed process it would be declared

understand but it should be taken into consideration existing modules. I guess right now it's an Object, therefore the meta spec should be applied

modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
```diff
@@ -198,9 +204,122 @@ class ProcessEntryHandler {
...
+                    throw new IllegalArgumentException(
+                        "Parameter '--${name}' expects a string value but got type ${value?.class?.simpleName}: '${value}'.")
+                return value
+
+            case 'file':
+            case 'directory':
+                if( !(value instanceof CharSequence) )
+                    throw new IllegalArgumentException(
+                        "Parameter '--${name}' expects a file/directory path but got type ${value?.class?.simpleName}: '${value}'.")
+                return value
```
pditommaso: The `file`/`directory` type handling returns raw strings as-is, while `getTypedValueForInput` (line 440-442) resolves `Path`-typed inputs using `Nextflow.file(str)`. For processes where the DSL type is `Object` (the common case), file/directory params from meta.yml will remain as raw strings rather than resolved `Path` objects. Consider whether `validateAndCastParam` should call `Nextflow.file()` for these types to be consistent.

modules/nextflow/src/test/groovy/nextflow/script/ProcessEntryHandlerTest.groovy
pditommaso, line 1: There are good unit tests for `validateAndCastParam` and `loadInputTypes` individually, but no test verifies the full path where `parseComplexParameters` loads meta.yml and applies casting. Consider adding a test with a stubbed `ScriptBinding` whose `scriptPath` points to a directory containing a real meta.yml, verifying that `parseComplexParameters` returns properly typed values end-to-end.

modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
```diff
@@ -198,9 +204,122 @@ class ProcessEntryHandler {
...
+                        "Parameter '--${name}' expects a boolean value (true/false) but got: '${value}'")
+                }
+                throw new IllegalArgumentException(
+                    "Parameter '--${name}' expects a boolean value but got type ${value?.class?.simpleName}: '${value}'")
+
+            case 'string':
+                if( !(value instanceof CharSequence) )
+                    throw new IllegalArgumentException(
+                        "Parameter '--${name}' expects a string value but got type ${value?.class?.simpleName}: '${value}'.")
+                return value
```
pditommaso: Minor: the `string` type validation throws if the value is not a `CharSequence`. Low risk since CLI params arrive as strings, but if any upstream processing ever converts a value to a non-string type before meta.yml casting is applied, this would throw unnecessarily.

modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
```diff
@@ -198,9 +204,122 @@ class ProcessEntryHandler {
...
+
+            final metaYml = scriptPath.parent.resolve(ModuleStorage.MODULE_MANIFEST_FILE)
+            final typeMap = ModuleSpec.loadInputTypes(metaYml)
+            return typeMap
+        }
+        catch( Exception e ) {
+            log.debug "Could not load input types from meta.yml: ${e.message}"
+            return null
+        }
+    }
```
pditommaso, line 251: Nit: `loadInputTypesFromMetaYml` re-parses the meta.yml on every call. Currently called once per process invocation so not a real issue, but caching the result in an instance field would be a simple improvement if this path is ever hit more than once.

modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
```diff
@@ -198,9 +204,122 @@ class ProcessEntryHandler {
...
+
+            case 'file':
+            case 'directory':
+                if( !(value instanceof CharSequence) )
+                    throw new IllegalArgumentException(
+                        "Parameter '--${name}' expects a file/directory path but got type ${value?.class?.simpleName}: '${value}'.")
+                return value
+
+            default:
+                log.debug "Unknown spec type '${type}' for parameter '--${name}' — skipping cast"
```
pditommaso: Nit: this log message uses an em-dash (`—`) character. The rest of the codebase uses ASCII `--` in log messages. Consider replacing for consistency.

modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
jorgee: Not sure if we should convert to `Path` at this point. At least, I think it is not currently needed. Later, there is a conversion to Path in `getTypedValueForInput` for typed processes or in `parseFileInput` for v1 `FileInParams`. `parseFileInput` is also taking into account lists.

jorgee: In `nf-core/fastqc`, the `reads` parameter is defined as "file" but could be also a `List`. Better to keep them as `String`.

modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
jorgee: Replaced in 74311c8

modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
jorgee: Instance field added to cache meta.yml load [74311c8](https://github.com/nextflow-io/nextflow/pull/6968/commits/74311c8a6b87f37cae01ba50da5b59e762cd23e9)

modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
jorgee: Removing the `CharSequence` check and returning `value.toString()` (74311c8)

modules/nextflow/src/test/groovy/nextflow/script/ProcessEntryHandlerTest.groovy
jorgee: Extended the tests to validate different argument parsing types for legacy (V1) with and without meta.yml and for typed processes (V2) (74311c8)

modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
jorgee: Finally, I have modified the `validateAndCastParam` and `getValueForInput` to manage file/directory cases when receiving either strings, Paths or List<Path>

modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
jorgee: I have applied the `meta.yml` cast in `getProcessArgumentsV1`. In the other case, it is not needed. The same is done in `getTypedValueForInput()`

bentsherman:
There is a problem with the module spec schema:

https://github.com/nextflow-io/nextflow/blob/5b97b272b67ff68f918789a28c5ecf93248c19b6/adr/module-spec-schema.json#L220-L242

It implies the following structure for inputs/outputs:

```yaml
input:
  - name: multiqc_files
    type: file
  - name: multiqc_config
    type: file
  - name: extra_multiqc_config
    type: file
  - name: multiqc_logo
    type: file
```

When in reality all `meta.yml` files use the following structure:

```yaml
input:
  - multiqc_files:
      type: file
  - multiqc_config:
      type: file
  - extra_multiqc_config:
      type: file
  - multiqc_logo:
      type: file
```

So the module spec parser in this PR is trying to support three formats when in fact there are only two, and the third one in the module spec schema is incorrect.

So we need to fix the module spec schema. I will push some changes here and pass it back to @jorgee

bentsherman:
On further review, I see that the module spec schema is correct, but some of the examples hadn't been updated correctly. So the PR is correct to handle both the new module spec format and legacy nf-core format

I will fix the discrepancy and update here

PR #6972, opened 2026-03-25
Document workflow output lineage types

jorgee:
This pull request updates the logic for determining the type name of parameters in the `LinObserver` class, resolving collections as List


pditommaso:
Why exposing such low level details, would not be better to keep it more abstract from the impl?

jorgee:
I added just to be coherent with standard library types. See Rob's comment in slack about `Collection` type in data lineage.

modules/nf-lineage/src/main/nextflow/lineage/LinObserver.groovy
```diff
@@ -408,10 +409,16 @@ class LinObserver implements TraceObserverV2 {
...
+        if( param instanceof Bag )
+            return Bag.simpleName
+        if( param instanceof Set )
+            return Set.simpleName
+        if( param instanceof List )
+            return List.simpleName
+        if( param instanceof Map )
             return Map.simpleName
+        if( param instanceof Iterable )
+            return Iterable.simpleName
```
bentsherman:
```suggestion
        if( param instanceof Collection )
            return List.simpleName
        if( param instanceof Map )
            return Map.simpleName
```

Let's just keep it simple

jorgee: Ok, applied your suggestion and modified the tests accordingly

jorgee:
Update: 
- Resolve all Collections to List
- Update doc
- Update PR description and name

modules/nf-lineage/src/main/nextflow/lineage/LinObserver.groovy
```diff
@@ -19,6 +19,7 @@ package nextflow.lineage
 import nextflow.NextflowMeta
 import nextflow.extension.FilesEx
 import nextflow.lineage.exception.OutputRelativePathException
+import nextflow.script.types.Bag
```
bentsherman:
```suggestion
```

modules/nf-lineage/src/test/nextflow/lineage/LinObserverTest.groovy
```diff
@@ -47,6 +47,7 @@ import nextflow.script.ScriptMeta
 import nextflow.script.TokenVar
 import nextflow.script.WorkflowMetadata
 import nextflow.script.params.EnvOutParam
+import nextflow.util.ArrayBag
```
bentsherman:
```suggestion
```

pditommaso (dismissed):
Not sure to understand about the rationale of this ?

bentsherman:
To keep the types in lineage records consistent with Nextflow standard types. These lists can be correctly annotated as `List` or `Collection`, but only the former type exists in the Nextflow types. Rob N was confused by the inconsistency

pditommaso:
but list means that it will guarantee an ordering,  I think collection was chosen because that assumption cannot be made

bentsherman:
Yes I think that is why we called it `Collection`. On the other hand, it is serialized into a List (JSON array) at the end of the day with a particular order.

I don't have a strong opinion on this. I would be fine with keeping it as a weird edge

pditommaso:
Think it's fine to keep and just document the semantic

PR #6973, opened 2026-03-25
Add resourceAllocation field to trace record

pditommaso:
## Summary
- Add `resourceAllocation` transient field to `TraceRecord` exposing scheduler-allocated resources (cpuShares, memoryMiB, accelerators, time)
- Source the value from the last `TaskAttempt.resources`, falling back to `TaskState.resourceAllocation` when no attempts exist
- Forward the field through `TowerClient` to the platform
- Bump sched-client to 0.46.0-SNAPSHOT (renames `TaskState.resourceRequirement` → `resourceAllocation`)

## Test plan
- [x] Unit tests for `getResourceAllocation()` in `SeqeraTaskHandlerTest` (null state, single attempt, multiple attempts, fallback scenarios)
- [x] Unit test for `resourceAllocation` propagation in `TowerClientTest.makeTasksReq()`
- [x] Existing trace metadata test updated to verify `resourceAllocation` included in trace record


pditommaso:
@bentsherman have a look at this. I think we can remove accelerator for "classic" trace and only keep as platform trace via `resourceAllocation`

modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
```diff
@@ -125,6 +123,7 @@ class TraceRecord implements Serializable {
     transient private ContainerMeta containerMeta
     transient private Integer numSpotInterruptions
     transient private String logStreamId
+    transient private Map<String,Object> resourceAllocation
```
bentsherman, line 128: not blocking, but the naming is a bit awkward. how about `resourcesAllocated` ?

modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
```diff
@@ -103,9 +103,7 @@ class TraceRecord implements Serializable {
             vol_ctxt: 'num',        // -- /proc/$pid/status field 'voluntary_ctxt_switches'
             inv_ctxt: 'num',        // -- /proc/$pid/status field 'nonvoluntary_ctxt_switches'
             hostname: 'str',
-            cpu_model:  'str',
-            accelerator: 'num',
-            accelerator_type: 'str'
```
bentsherman, line 108: These fields denoted the *requested* accelerators. Now that you are moving them into the allocated resources, will they denote the accelerators *allocated* by sched?

plugins/nf-seqera/src/test/io/seqera/executor/SeqeraTaskHandlerTest.groovy
```diff
@@ -579,11 +582,92 @@ class SeqeraTaskHandlerTest extends Specification {
...
+        result.cpuShares == 512
+        result.memoryMiB == 1024
+    }
+
     def 'should return granted time from resource requirement'() {
         given:
         def handler = createHandler()
         handler.cachedTaskState = new SchedTaskState()
-            .resourceRequirement(new ResourceRequirement().time('2h'))
+            .resourceAllocation(new ResourceRequirement().time('2h'))
```
bentsherman, line 670: Should `ResourceRequirement` be renamed to something like `ResourceRequest` ? Seems like here it is being used to model things other than hard requirements

modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
pditommaso: Eventually, it's prefer to keep aligned to sched naming for now

plugins/nf-seqera/src/test/io/seqera/executor/SeqeraTaskHandlerTest.groovy
pditommaso: Let's move this into Sched API naming, because it's inherited from there

modules/nextflow/src/main/groovy/nextflow/trace/TraceRecord.groovy
pditommaso: Good point. reverted

PR #6926, opened 2026-03-16
Fix Google Batch exit code when spot claim is successfully retried

jorgee:
close #6779
This pull request addresses the handling of exit codes for Google Batch spot instance retries and improves test coverage for these scenarios. The main change ensures that tasks retried on spot instances correctly update their exit status by reading the `.exitcode` file, and the associated tests are updated to cover this logic.

Improvements to spot instance exit code handling:

* [`GoogleBatchTaskHandler.groovy`](diffhunk://#diff-b5373e32074aac589c96f4ce9de7a10a7621721006052186873d610388acaf08R691-R694): Added logic to check if a task has an exit status >= 50000 (indicative of a spot instance retry) and, if so, update the exit status by reading the `.exitcode` file.

Enhancements to test coverage:

* [`GoogleBatchTaskHandlerTest.groovy`](diffhunk://#diff-2351c0cfbd6ced5d60124834a8b49d8b995ffdb15033870be2f3c4283778c5b4L799-R799): Updated mock job status handling to use the `JOB_STATE` variable, allowing more flexible test scenarios.
* [`GoogleBatchTaskHandlerTest.groovy`](diffhunk://#diff-2351c0cfbd6ced5d60124834a8b49d8b995ffdb15033870be2f3c4283778c5b4L811-R822): Modified test cases to include scenarios where the exit code is re-read for retried spot instances, and expanded the test matrix to cover both failed and succeeded job states, as well as the number of times `readExitFile()` is called.

PR #6977, opened 2026-03-25
Fix Google Batch exit code after spot preemption retry

ejseqera:
When using `google.batch.maxSpotAttempts > 0`, Google Batch will automatically retry the task with the same job ID right after a 50001 exit is returned. When a Google Batch job succeeds after an internal spot preemption retry (via `lifecyclePolicy` with `RETRY_TASK` on exit code 50001), `GoogleBatchTaskHandler` reports the exit code from the **first failed attempt** (50001) instead of the **final successful attempt** (0). This causes Nextflow to terminate the pipeline even though the Batch job completed successfully. 

This can be mitigated by including something like:
```
    process.errorStrategy = {
        if (task.exitStatus == 50001) {
            return 'ignore'
        }
 }
```
But if a user forgets, jobs will appear to have failed even when a preemption has been retried and the task succeeds.         

### Root cause

`getExitCode()` filters task status events for those with a `TaskExecution` and picks the latest by timestamp. The problem: Google Batch's `SUCCEEDED` status event does **not** carry a `TaskExecution` object; only the `FAILED` preemption event does. So the method always returns 50001 from the failed attempt, even when the job state is `SUCCEEDED`.

### Fix

In `checkIfCompleted()`, when the Batch job state is `SUCCEEDED` but the Batch API-derived exit code is non-zero, fall back to the `.exitcode` file written by the final successful attempt.

- This affects every pipeline using `google.batch.maxSpotAttempts` on spot VMs
- Without fix: any spot preemption that Batch successfully retries still terminates the pipeline
- With fix: exit code correctly resolves to 0 from the `.exitcode` file

### Test plan

- [x] Added Spock test for spot preemption retry scenario (array and non-array tasks)
- [x] All existing `GoogleBatchTaskHandlerTest` tests pass

### DIFF
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchTaskHandler.groovy
@@ -559,6 +559,13 @@ class GoogleBatchTaskHandler extends TaskHandler implements FusionAwareTask {
             log.debug "[GOOGLE BATCH] Process `${task.lazyName()}` - terminated job=$jobId; task=$taskId; state=$state"
             // finalize the task
             task.exitStatus = getExitCode()
+            // When the Batch job succeeded but the API-reported exit code reflects a failed
+            // intermediate attempt (e.g. spot preemption with exit code 50001 that was retried),
+            // fall back to the .exitcode file written by the final successful attempt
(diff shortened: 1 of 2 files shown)


bentsherman:
I thought we just fixed this in #6926 ... is this issue still happening in latest edge?

joshiesh:
I was unfortunately testing with 26.02.0-edge, the latest edge version does resolve the issue. Closing this. Thanks Ben!

PR #6979, opened 2026-03-26
Fix: handle empty publishedValues in PublishOp.onComplete for optional outputs

### LINKED ISSUE #6978: `module run` crashes with "Cannot access first() element from an empty List" when process has empty optional outputs
### Description

`nextflow module run` crashes with `NoSuchElementException: Cannot access first() element from an empty List` when running a module whose process has optional output channels that produce no files. The underlying process task completes successfully (exit code 0), but the error occurs during the output publishing phase in `PublishOp.onComplete`.

### Version

- Nextflow: `26.03.0-edge build 0`
- System: Mac OS X 26.3.1
- Runtime: Groovy 4.0.30 on OpenJDK 64-Bit Server VM 21+35

### Command used

```bash
nextflow module run nf-core/trimgalore \
    --meta.id test_sample \
    --reads "reads_1.fastq.gz,reads_2.fastq.gz" \
    --outdir results_trimgalore
```

### Error output

```
[PIPELINE] main.nf | profile=standard
[WORKDIR] /Users/pditommaso/Projects/nextflow/test-modules/work
[PROCESS 13/39f78a] TRIMGALORE (test_sample)
[ERROR] ERROR ~ Cannot access first() element from an empty List

 -- Check '.module.log' file for details

[FAILED] completed=1 failed=0 cached=0
```

### Debug log (`.module.log`)

The relevant lines from the debug log:

```
[TaskFinalizer] DEBUG nextflow.processor.TaskProcessor - Process TRIMGALORE > Skipping output binding because one or more optional files are missing: fileoutparam<2:1>
[TaskFinalizer] DEBUG nextflow.processor.TaskProcessor - Process TRIMGALORE > Skipping output binding because one or more optional files are missing: fileoutparam<3:1>
[TaskFinalizer] DEBUG nextflow.processor.TaskProcessor - Process TRIMGALORE > Skipping output binding because one or more optional files are missing: fileoutparam<4:1>
[main] ERROR nextflow.Session - Cannot access first() element from an empty List
java.util.NoSuchElementException: Cannot access first() element from an empty List
	at org.codehaus.groovy.runtime.DefaultGroovyMethods.first(DefaultGroovyMethods.java:10343)
	at nextflow.extension.PublishOp.onComplete(PublishOp.groovy:221)
	at nextflow.extension.PublishOp$_apply_closure2$_closure7.doCall(PublishOp.groovy:78)
	at nextflow.extension.PublishOp.safeExecute(PublishOp.groovy:94)
	at nextflow.extension.PublishOp$_apply_closure2.doCall(PublishOp.groovy:78)
	at nextflow.extension.DataflowHelper$3.afterStop(DataflowHelper.groovy:257)
	...
```

### Analysis

The `nf-core/trimgalore` module defines several optional output channels (`zip`, `html`, `unpaired`) that may not produce files depending on the run parameters. When these optional outputs emit no files:

1. The task itself completes successfully (exit 0, trimmed reads are produced)
2. The output binding step correctly skips the missing optional files
3. But `PublishOp.onComplete()` at line 221 calls `.first()` on an empty list, causing the crash

### Expected behavior

`nextflow module run` should handle empty optional output channels gracefully and complete successfully, publishing only the outputs that were produced (in this case: `reads`, `log`, and `versions_trimgalore`).

jorgee:
## Summary

- Fixes `NoSuchElementException: Cannot access first() element from an empty List` in `PublishOp.onComplete` when a process has optional output channels that produce no files
- Guards the `.first()` call with an empty-list check, returning `null` when no values were published on a value channel

Fixes #6978

## Test plan

- [ ] Run `nextflow module run` on a module with optional outputs that produce no files (e.g. `nf-core/trimgalore` with parameters that skip optional outputs)
- [ ] Verify the run completes successfully and publishes only the non-empty outputs
- [ ] Verify existing publish behaviour is unchanged when optional outputs do produce files

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/extension/PublishOp.groovy
@@ -217,6 +217,7 @@ class PublishOp {
      */
     protected void onComplete() {
         // publish individual record if source is a value channel
+        // NOTE: handle (invalid) empty dataflow value from legacy process, legacy collect(), etc
         final outputValue = CH.isValue(source)
             ? (publishedValues ? publishedValues.first() : null)
             : publishedValues


jorgee:
Current implementation is printing the optional outputs with `null`. It is easy not to print them, but the print feature is the `OutputDsl `. It will also apply to workflows. Moreover, not sure if it could be better to print with nulls for completeness

bentsherman:
Thanks Jorge. This issue has actually been reported in other places as well.

It comes from the fact that some operations can produce an empty dataflow value, which is invalid since a dataflow value should always contain a value (even if null).

Most of these cases are addressed by typed processes and typed workflows, and could be avoided in legacy code using `ifEmpty`.

In this case, we could add a post-processing step to ProcessEntryHandler to impute empty dataflow values with `null`, but we might as well just solve the issue for everyone

Printing `null` is intentional for optional outputs, to show that they exist but are optional

PR #6982, opened 2026-03-27
Send module resources as build context for container builds

pditommaso:
## Summary
- When a `containerFile` (Dockerfile) is provided alongside module resources, send the module resources as `buildContext` instead of `containerConfig.layers`
- This makes module `bin/` and `lib/` files available to `COPY`/`ADD` instructions during the image build, rather than overlaying them as opaque layers at runtime
- When no `containerFile` is present (image-only requests), existing behavior is preserved — module resources remain as `containerConfig.layers`

## Motivation
Enable support for Dockerfile `COPY` and `ADD` commands using module resources as build context. Currently module resources are sent as `containerConfig.layers`, which are overlaid at runtime and not accessible during the image build. By sending them as `buildContext` when a `containerFile` is present, module-provided files become available to `COPY`/`ADD` instructions in the Dockerfile.

## Test plan
- [x] Existing test: module resources as layers when no containerFile (unchanged behavior)
- [x] New test: module resources sent as `buildContext` when containerFile is present
- [x] New test: module resources as `buildContext` + project resources as layer with containerFile
- [x] All nf-wave plugin tests pass

### DIFF
--- a/plugins/nf-wave/src/main/io/seqera/wave/plugin/SubmitContainerTokenRequest.groovy
+++ b/plugins/nf-wave/src/main/io/seqera/wave/plugin/SubmitContainerTokenRequest.groovy
@@ -158,4 +158,12 @@ class SubmitContainerTokenRequest {
      */
     String buildTemplate
 
+    /**
+     * A layer holding the build context for this container request.
+     * When a container file (Dockerfile) is provided, module resources
+     * are sent as build context so they are available to COPY/ADD instructions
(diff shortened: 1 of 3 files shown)


pditommaso:
Tagging @adamrtalbot for visibility

pditommaso:
It turns out it's not needed. Closing in favour of https://github.com/nextflow-io/nextflow/pull/6984

PR #7004, opened 2026-04-04
[codex] Clarify self-update guidance for sudo installs

llewellyn-sl:
## Summary
This draft adds a targeted warning for `nextflow self-update` installs that were done with `sudo` or into root-owned locations, plus recovery steps if that leaves the installation unusable.

## Source context
- Slack source: https://seqera.slack.com/archives/C0APYR2A3A8/p1775254519687639
- Jira: https://seqera.atlassian.net/browse/EDU-1119
- Related product discussion: https://seqera.slack.com/archives/CUTDS6JE9/p1770300590508689

## Doc target
- `docs/install.md`
- `docs/updating-nextflow.md`

## Rationale
The EDU ticket called out a docs gap around users breaking local Nextflow installs after running `self-update` with elevated permissions. The existing docs already warn that restricted permissions can break self-install and self-update, so this draft keeps the change narrow: it makes the `sudo` / root-owned path explicit, recommends a user-writable install location, and adds a short recovery sequence for the broken-install case.

## Validation
- Reviewed the edited docs diff directly
- Attempted `python3 -m sphinx -b dummy . /tmp/nextflow-docs-dummy` from `docs/`, but this environment does not have `sphinx` installed

## Follow-up
If we see more cases like this, the install and update docs may also benefit from a short troubleshooting section for common local permission mistakes.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -94,6 +94,8 @@ To install Nextflow with the self-installing package:
 
     :::{warning}
     Nextflow updates its executable during the self-install process, therefore the update can fail if the executable is placed in a directory with restricted permissions.
+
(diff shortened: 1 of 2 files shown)


llewellyn-sl:
Tracking EDU-1119 docs automation recovery run.

PR #7007, opened 2026-04-06
Fix NPE in GitLab/Gitea getGitCredentials when token set but password null

### LINKED ISSUE #7006: GitlabRepositoryProvider: NullPointerException on git clone when password is null but token is set

**Expected:** When GitLab credentials are configured with `user` and `token` but no `password`, Nextflow should either:
- Use the token for git clone authentication (as `BitbucketRepositoryProvider` already does), or
- Fail with a clear error message indicating that the `password` field is required

**Actual:** Nextflow throws an unhandled `NullPointerException: Cannot invoke "String.toCharArray()" because "password" is null` during `git clone`. This is confusing because API calls to GitLab succeed (the token is used via `PRIVATE-TOKEN` header), and `hasCredentials()` returns `true` — only the clone step fails.

1. Create an SCM config with GitLab credentials that have a `user` and `token` but no `password`:

```groovy
providers {
    gitlab {
        platform = 'gitlab'
        server = 'https://gitlab.example.com'
        user = 'myuser'
        token = 'glpat-xxxxxxxxxxxx'
    }
}
```

2. Run a pipeline hosted on that GitLab instance:

```bash
NXF_SCM_FILE=scm.config nextflow run org/pipeline -r main
```

### Program output 

```
N E X T F L O W  ~  version 25.10.2
Pulling org/pipeline ...
ERROR ~ Cannot invoke "String.toCharArray()" because "password" is null

```

Stack trace from the log:

```
java.lang.NullPointerException: Cannot invoke "String.toCharArray()" because "password" is null
    at org.eclipse.jgit.transport.UsernamePasswordCredentialsProvider.<init>(UsernamePasswordCredentialsProvider.java:62)
    at nextflow.scm.RepositoryProvider.getGitCredentials(RepositoryProvider.groovy:218)
```

### Environment 

* Nextflow version: 25.10.2
* Java version: 21 (Corretto)
* Operating system: Linux (AWS Batch)

### Additional context

There is an inconsistency in `GitlabRepositoryProvider`:

- `hasCredentials()` is overridden to return `true` when a token is present (even without a password)
- `getAuth()` is overridden to use the token via `PRIVATE-TOKEN` header for API calls
- But `getGitCredentials()` is **not** overridden, so git clone falls through to `RepositoryProvider.getGitCredentials()` which passes the null password to JGit's `UsernamePasswordCredentialsProvider`

`BitbucketRepositoryProvider` already handles this by overriding `getGitCredentials()` to fall back to the token:

```groovy
CredentialsProvider getGitCredentials() {
    return new UsernamePasswordCredentialsProvider(getUser(), getToken() ?: getPassword())
}
```

Adding the same override to `GitlabRepositoryProvider` would fix the inconsistency and allow token-only GitLab credentials to work for clone operations.

Note: the [docs](https://www.nextflow.io/docs/latest/git.html) show `password` in the GitLab example and mention using the token as the password value, so providing `password` is a valid workaround. The issue is that the NPE is unhelpful and the behavior is inconsistent with `hasCredentials()` returning `true`.

robsyme:
## Summary
- Override `getGitCredentials()` in `GitlabRepositoryProvider` and `GiteaRepositoryProvider` to fall back to the token when password is not provided, using `getToken() ?: getPassword()`
- Matches the existing pattern in `BitbucketRepositoryProvider` (line 64-66)
- Fixes the `NullPointerException: Cannot invoke "String.toCharArray()" because "password" is null` when SCM config has `user` and `token` but no `password`

Fixes #7006

## Test plan
- [x] Added parameterized `'should return git credentials'` tests for both GitlabRepositoryProvider and GiteaRepositoryProvider covering password-only, token-only, and both configurations
- [x] All SCM unit tests pass (`NXF_SMOKE=1 ./gradlew :nextflow:test --tests "nextflow.scm.*"`)

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/scm/GiteaRepositoryProvider.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/GiteaRepositoryProvider.groovy
@@ -20,6 +20,8 @@ package nextflow.scm
 import groovy.transform.CompileDynamic
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
+import org.eclipse.jgit.transport.CredentialsProvider
+import org.eclipse.jgit.transport.UsernamePasswordCredentialsProvider
 
 import java.nio.charset.StandardCharsets
 
@@ -63,6 +65,12 @@ final class GiteaRepositoryProvider extends RepositoryProvider {
         return getToken()
             ? true
             : super.hasCredentials()
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    CredentialsProvider getGitCredentials() {
+        return new UsernamePasswordCredentialsProvider(getUser(), getToken() ?: getPassword())
     }
 
     @Override
--- a/modules/nextflow/src/main/groovy/nextflow/scm/GitlabRepositoryProvider.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/GitlabRepositoryProvider.groovy
@@ -18,6 +18,8 @@ package nextflow.scm
 
 import groovy.json.JsonSlurper
 import groovy.util.logging.Slf4j
+import org.eclipse.jgit.transport.CredentialsProvider
+import org.eclipse.jgit.transport.UsernamePasswordCredentialsProvider
 
 import java.nio.charset.StandardCharsets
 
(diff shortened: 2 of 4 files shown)


pditommaso:
Thanks Rob! ✅

PR #7017, opened 2026-04-08
Rename .module-info to .module_info

jorgee:
Suggested by @ewels change to Snake case to be inline with nextflow_schema.json, etc.

### DIFF
--- a/adr/20251114-module-system.md
+++ b/adr/20251114-module-system.md
@@ -9,7 +9,7 @@
 ## Updates
 
 ### Version 2.7 (2026-03-09)
-- **Renamed `.checksum` to `.module-info`**: Leaves room for additional properties in the future
+- **Renamed `.checksum` to `.module_info`**: Leaves room for additional properties in the future
 - **Removed `@` prefix from module scopes**: Local modules are distinguished from remote modules by presence/absence of `./` prefix
 - **Removed version pinning from config**: Installed module versions are now inferred from the `meta.yml` of each module in the `modules/` directory instead of being declared in `nextflow.config`
 
@@ -78,7 +78,7 @@ include { MY_PROCESS } from './modules/my-process.nf'
 
 **Resolution Order**:
 1. Check local `modules/scope/name/` exists
-2. Verify integrity against `.module-info` file
+2. Verify integrity against `.module_info` file
 3. Apply resolution rules (see below)
 
 **Resolution Rules**:
@@ -90,7 +90,7 @@ include { MY_PROCESS } from './modules/my-process.nf'
 | Exists, checksum mismatch | **Warn**: locally modified, will NOT replace unless `-force` is used |
 
 **Key Behaviors**:
-- **Local modification**: When the local module content was manually changed (checksum mismatch with `.module-info`), Nextflow warns and does NOT override to prevent accidental loss of local changes
+- **Local modification**: When the local module content was manually changed (checksum mismatch with `.module_info`), Nextflow warns and does NOT override to prevent accidental loss of local changes
 - **Force flag**: Use `-force` with `nextflow module install` to override locally modified modules
 
 **Resolution Timing**: Modules resolved at workflow parse time (after plugin resolution at startup).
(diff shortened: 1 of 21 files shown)


pditommaso (changes requested):
Disagree, we tend to use Kebab-case over snakecase

bentsherman (approved):

ewels:
> Disagree, we tend to use Kebab-case over snakecase

Really? I thought it was the opposite.. Looking now there are relatively few files that have multiple words in at all, but those that do seem to use snake_case:

- `execution_report.html`
- `execution_timeline.html`
- `execution_trace.txt`
- `pipeline_dag.dot` / `pipeline_dag.html`
- (`nextflow_schema.json`)

Which kebab-case files are you thinking of @pditommaso ?

pditommaso:
**Analysis: kebab-case vs snake_case naming conventions across the Nextflow codebase**

Prompted by: _"make an analysis of using kebab-case vs snake_case in this codebase"_

### Summary by area

| Area | Convention | Examples |
|------|-----------|---------|
| **Resource files** | **kebab-case** | `command-run.txt`, `cloud-boot.txt`, `build-info.properties`, `az-locations.json`, `vm-list-size-*.json` |
| **CLI flags** | **kebab-case** | `-with-docker`, `-pool-size`, `-work-dir`, `-dump-hashes`, `-params-file` |
| **Output filenames** | **kebab-case** | `trace-<ts>.txt`, `report-<ts>.html`, `timeline-<ts>.html`, `dag-<ts>.html` |
| **Trace record fields** | **snake_case** | `task_id`, `native_id`, `peak_rss`, `peak_vmem`, `read_bytes`, `error_action` |
| **Config properties** | **camelCase** | `publishDir`, `errorStrategy`, `maxRetries`, `storeDir`, `cacheDir` |
| **Process directives** | **camelCase** | `beforeScript`, `afterScript`, `stageInMode`, `resourceLabels` |
| **Channel operators** | **camelCase** | `flatMap`, `groupTuple`, `splitCsv`, `collectFile`, `toSortedList` |
| **Environment variables** | **SCREAMING_SNAKE** | `NXF_LOG_FILE`, `NXF_EXECUTOR`, `NXF_DOCKER_LEGACY` |

### Key findings

1. **No snake_case in resource filenames.** Every resource file under `src/main/resources/` and `src/resources/` across all modules and plugins uses either kebab-case or single-word names. Zero files use underscores.

2. **snake_case is limited to trace record fields** — the column names in trace CSV/TSV output (`task_id`, `peak_rss`, `read_bytes`, etc.). This is the only place snake_case appears as a deliberate convention.

3. **kebab-case dominates the "external/user-facing" layer:**
   - CLI flags (what users type)
   - Resource filenames (what gets bundled in JARs)
   - Generated output filenames (what users see on disk)

4. **camelCase dominates the "API/DSL layer":**
   - Config keys, process directives, channel operators — everything users write in `.nf` and `.config` files follows Groovy/Java camelCase conventions.

### Convention map

```
User types CLI flag     →  kebab-case      (-with-trace)
User writes config      →  camelCase       (trace.fields)
Nextflow reads resource →  kebab-case      (command-run.txt)
Nextflow writes output  →  kebab-case      (trace-20260408.txt)
Trace file columns      →  snake_case      (task_id, peak_rss)
Env vars                →  SCREAMING_SNAKE (NXF_EXECUTOR)
```

### Conclusion

The codebase is consistent within each layer. **kebab-case** is the convention for filenames and CLI flags (external/OS-level identifiers). **snake_case** is only used for trace record field names (tabular data output). There is no mixing — resource files never use snake_case, and trace fields never use kebab-case.

pditommaso:
I wanted to provide a comprehensive analyses and as you can see kebab-cases in the dominant convention, with a few exception, as you can see. 

Apart this, I'd argue the the `.module-info` it's an internal implementation detail that's not supposed to managed directly by user. As such I don't this worth caring too much about and spent time on refactoring something has been already agreed and released

jorgee:
I thought it was easier to modify. Tests are failing because in the last changes, we moved some code to the `npr-client` in the registry codebase and it is also using the .module-file. So we should also update it. Minor change, but just to reconsider if the change is really needed.

ewels:
Yes I was only thinking about user-facing files, of which there aren't many. I wonder why my one said that it uses underscores? Is that an old pattern before we added timestamps to those default file names I wonder?

Anyway, this is not a thing I care about much at all. I had in my head that we'd discussed moving towards snake case for output files but if that's not the case then it's fine as it stands 👍🏻 Happy for this to be closed and apologies for the noise @jorgee

mashehu:
>  wonder why my one said that it uses underscores? 

Figured that part out now, @ewels. We give these files a snake_case name in the nf-core template 🤦🏻 https://github.com/nf-core/tools/blob/main/nf_core/pipeline-template/nextflow.config#L275-L290

PR #7018, opened 2026-04-08
Add option to disable printing workflow outputs

bentsherman:
This PR adds the ability to disable printing workflow outputs by specifying `-output-format none`

It also improves the workflow output printing logic by (1) using an onComplete handler instead of an igniter and (2) not printing in the event of a pipeline error

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -1614,7 +1614,7 @@ The `run` command is used to execute a local pipeline script or remote pipeline 
 `-output-format`
 : :::{versionadded} 26.04.0
   :::
-: Output format for printing workflow outputs. Options: `text` (default), `json`.
+: Output format for printing workflow outputs. Options: `text` (default), `json`, `none`.
 
 `-params-file`
 : Load script parameters from a JSON/YAML file.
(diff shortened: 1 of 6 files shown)


rpetit3:
Playing around with the latest edge release, I wonder if an option to output to a json file might also be useful. (I may have also missed that!)

For a single sample, using `-output-format json` 2894 lines are the outputs in JSON format, and 113 lines are related to the the pipeline. `text` produces less, but I think users will still be confused (e.g. _"was some debug text left in?_).

I think my usage would be to default to `none` then when debugging switch to `json` or `json file` if available.

I do recognize this as super useful information to have on hand to review/testing/debugging.

bentsherman:
For that the idea is to use `-q` with `-output-format json` to suppress all other output and just print the JSON. Then you can redirect stdout to a file. This is similar to how `nextflow inspect` works, except `nextflow inspect` sets `-q` automatically

jorgee (approved):
Changes look fine to me. Just a minor comment, shouldn't be better to set default to `none` for the `CmdRun` case to work as in the previous version?

bentsherman:
Thanks Jorge. Since workflow outputs are still new and we want to promote it, I think it's fine to have it enabled by default. Also for consistency with `nextflow module run` which prints the human-readble outputs by default.

Issue #6988, opened 2026-03-30
`nextflow run -offline` should not affect file-system local repositories


MatthiasZepper:
## New feature

I am proposing a new feature that implements a finer grained check in the AssetManager to distinguish between file-system local and truly remote SCM repositories. 

## Use case 

This feature is useful to resolve a chicken-and-egg conundrum with offline / air-gapped HPC environments and pipelines managed by an on-premise Seqera Platform Enterprise installation:

- The compute environment needs to be set up with `tw compute-envs add slurm ... --env "head:NXF_OFFLINE=true"` to ensure that Nextflow doesn't try connecting to the web. Only setting `--env "head:NXF_DISABLE_CHECK_LATEST=true"` falls short, since it disables the check for the latest Nextflow version, but not attempts to connect to the plugin registry.

- At the same time, setting `NXF_OFFLINE` to _true_ prevents pipelines from launching for the first time with an _Unknown project local/pipeline.git -- NOTE: automatic download from remote repositories is disabled_ error.

This can be resolved either by pulling from file-system local paths even in offline mode (this proposal) or alternatively by introducing an `NXF_DISABLE_CHECK_PLUGINS` for the plugin registry analogous to `NXF_DISABLE_CHECK_LATEST` for the Nextflow version.

## Steps to reproduce

1. Create a local bare clone of a Git repository, e.g. with `git clone --bare git@github.com:nextflow-io/socks.git`. (_This is equivalent to using [`nf-core pipelines download --platform -r "main" nextflow-io/socks`](https://nf-co.re/docs/nf-core-tools/pipelines/download#adapting-downloads-to-seqera-platform) but avoids the additional dependency_)
2. Try launching a run in `-offline` mode with an absolute path and the `file:/` prefix: `nextflow run -offline file:/$(readlink -f socks.git)`. (_This is the setup that would be chosen in an on-premise Seqera Platform Enterprise installation_):
>  N E X T F L O W   ~  version 25.10.4
> Unknown project `local/socks` -- NOTE: automatic download from remote repositories is disabled
3. Optionally, repeat without `-offline`. `nextflow run file:/$(readlink -f socks.git)`:
>  N E X T F L O W   ~  version 25.10.4
> Pulling local/socks ...
> downloaded from file:/Users/matthias.zepper/Documents/Coding/Nextflow/download_pipelines/socks
> Launching `/Users/matthias.zepper/Documents/Coding/Nextflow/download_pipelines/socks` 
[...]

PR #7019, opened 2026-04-09
Make plugin metadata prefetch failure non-fatal

### LINKED ISSUE #7013: Plugin metadata prefetch fails hard even when result is unused (pre-installed plugins, exact versions)

When all requested plugins are pinned to exact versions and already installed in `$NXF_HOME/plugins/`, Nextflow should start them from the local directory without requiring network access.

Instead, Nextflow unconditionally calls the plugin registry API at startup (`PluginsFacade.groovy:415`), and if the registry is unreachable, the run aborts with a `PluginRuntimeException` even though the fetched metadata is never used.

This affects users in airlocked/firewalled environments who pre-install plugins into their container images. They have done everything right (pinned versions, pre-populated the plugins directory), but Nextflow still requires network access to a service whose response it will discard.

1. Install a plugin locally:
```bash
curl -L -o nf-schema-2.4.2.zip \
  "https://registry.nextflow.io/api/v1/plugins/nf-schema/2.4.2/download/nf-schema-2.4.2.zip"
unzip nf-schema-2.4.2.zip -d $NXF_HOME/plugins/
```

2. Create a workflow that pins the plugin version:
```groovy
plugins {
    id 'nf-schema@2.4.2'
}

workflow {
    log.info "Plugin loaded successfully"
}
```

3. Block access to `registry.nextflow.io` (e.g. via firewall rule or `/etc/hosts`) and run:
```bash
nextflow run main.nf
```

4. Nextflow fails at the metadata prefetch, before it ever checks whether the plugin is already installed locally.

Note: `NXF_OFFLINE=true` works around the plugin issue but also blocks remote pipeline cloning (`CmdRun.groovy:620-621`), making it unusable for environments that need to fetch pipelines from a Git server (e.g. Seqera Platform pulling from GitLab).

### Program output

```
Unable to connect to https://registry.nextflow.io/api/v1/plugins/dependencies?plugins=nf-schema%402.4.2&nextflowVersion=25.10.2 - cause: <connection error>
```

### Environment

* Nextflow version: 25.10.2+ (any version using the registry-based plugin system)
* Java version: any
* Operating system: any
* Bash version: any

### Additional context

The root cause is in the startup sequence in `PluginsFacade.start()`:

```groovy
// prefetch the plugins meta
updater.prefetchMetadata(startable)       // <-- always makes HTTP call, throws on failure
// finally start the plugins
for( PluginRef plugin : startable ) {
    updater.prepareAndStart(plugin.id, plugin.version)  // <-- would succeed from local dir
}
```

In `PluginUpdater.load0()`, when a plugin has an exact version and is already installed, no metadata is consulted:
- Line 351-359: version resolution is skipped (exact semver provided)
- Line 372-375: directory exists, download is skipped
- Line 386-406: transitive deps are read from the plugin's own MANIFEST, not registry metadata

The prefetched metadata is write-only for this code path.

A minimal fix would be to make the `prefetchMetadata` failure non-fatal (log a warning instead of throwing), so that the actual failure surfaces later only if something genuinely needs the metadata (version resolution for unpinned plugins, download URLs for missing plugins). If nothing needs it, the run succeeds.

See also #6988, which describes a related problem with `NXF_OFFLINE` being too coarse-grained.

jorgee:
## Summary

Fixes #7013

- Make `HttpPluginRepository.prefetch()` catch exceptions and log a warning instead of aborting startup
- Initialize the plugins map to an empty `HashMap` on failure so that `getPlugin()` can still attempt on-demand lazy fetches later
- Refactor tests to verify prefetch error tolerance while still testing `fetchMetadata()` error behavior directly

## Context

When all plugins are pinned to exact versions and pre-installed locally, Nextflow unconditionally calls the plugin registry API at startup. If the registry is unreachable, the run aborts even though the fetched metadata would never be used.

This affects users in airlocked/firewalled environments who pre-install plugins into container images. `NXF_OFFLINE=true` works around it but also blocks remote pipeline cloning, making it unusable for environments that need to fetch pipelines from a Git server.

With this fix, the prefetch error is deferred — it only surfaces later if something genuinely needs the metadata (version resolution for unpinned plugins, download URLs for missing plugins).

## Test plan

- [x] `HttpPluginRepositoryTest` — all 9 tests pass
- [x] All `nextflow.plugin.*` tests pass
- [x] Manual: pre-install a plugin, block registry access, verify `nextflow run` succeeds

### DIFF
--- a/modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
+++ b/modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
@@ -88,7 +88,6 @@ class HttpPluginRepository implements PrefetchUpdateRepository {
             }
             catch (Exception e) {
                 log.warn "Failed to prefetch plugin metadata - ${e.message}"
-                log.debug "Plugin metadata prefetch error details", e
                 this.plugins = new HashMap<>()
             }
         }
--- a/modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
+++ b/modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
@@ -83,7 +83,14 @@ class HttpPluginRepository implements PrefetchUpdateRepository {
     @Override
     void prefetch(List<PluginRef> plugins) {
         if (plugins && !plugins.isEmpty()) {
-            this.plugins = fetchMetadata(plugins)
+            try {
+                this.plugins = fetchMetadata(plugins)
+            }
+            catch (Exception e) {
+                log.warn "Failed to prefetch plugin metadata - ${e.message}"
+                log.debug "Plugin metadata prefetch error details", e
+                this.plugins = new HashMap<>()
+            }
         }
     }
 
--- a/modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
+++ b/modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
@@ -36,6 +36,7 @@ import nextflow.extension.FilesEx
 import nextflow.file.FileHelper
 import nextflow.file.FileMutex
 import org.pf4j.InvalidPluginDescriptorException
+import org.pf4j.ManifestPluginDescriptorFinder
 import org.pf4j.PluginDependency
 import org.pf4j.PluginRuntimeException
 import org.pf4j.PluginState
@@ -155,6 +156,11 @@ class PluginUpdater extends UpdateManager {
                 log.trace "Prefetching plugin metadata from repository: ${repo.getClass().getSimpleName()} [${repo.id}]; plugins=${plugins}"
                 repo.prefetch(plugins)
             }
+        }
+        // When remote registry an no plugins detected after prefetch, add local repo to work with downloaded plugins
+        if( !offline && getPluginsMap().isEmpty() ){
+            final localRepo = new LocalUpdateRepository('downloaded', pluginsStore)
+            this.addRepository(localRepo)
         }
     }
 
--- a/modules/nf-commons/src/test/nextflow/plugin/HttpPluginRepositoryTest.groovy
+++ b/modules/nf-commons/src/test/nextflow/plugin/HttpPluginRepositoryTest.groovy
@@ -104,14 +104,26 @@ class HttpPluginRepositoryTest extends Specification {
         r1.requires == ">=24.10.0"
     }
 
-    // ------------------------------------------------------------------------
-
-    def 'handle prefetch error when metadata service is unavailable'() {
+    def 'catch prefetch exceptions'() {
         setup:
         wiremock.stop()
 
         when:
         unit.prefetch([new PluginRef("nf-fake")])
+
+        then:
+        noExceptionThrown()
+        unit.getPlugins() == [:]
+    }
+
+    // ------------------------------------------------------------------------
+
+    def 'handle fetchMetadata error when metadata service is unavailable'() {
+        setup:
+        wiremock.stop()
+
+        when:
+        unit.fetchMetadata([new PluginRef("nf-fake")])
 
         then:
         def err = thrown PluginRuntimeException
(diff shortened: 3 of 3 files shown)


jorgee:
Tested without connectivity to registry. 

- When no plugin available: Print warning and fail because it can't download the plugin
```
 N E X T F L O W   ~  version 26.03.2-edge

WARN: Failed to prefetch plugin metadata - Unable to connect to https://registry.nextflow.io/api/v1/plugins/dependencies?plugins=nf-schema%402.4.2&nextflowVersion=26.03.2-edge - cause: null
Downloading plugin nf-schema@2.4.2
ERROR ~ Plugin with id nf-schema not found in any repository
```

- When plugin version is pinned and exists: Print the warning message at prefetch and succesfuly run the pipeline
```
 N E X T F L O W   ~  version 26.03.2-edge

WARN: Failed to prefetch plugin metadata - Unable to connect to https://registry.nextflow.io/api/v1/plugins/dependencies?plugins=nf-schema%402.4.2&nextflowVersion=26.03.2-edge - cause: null
Launching `main.nf` [nostalgic_almeida] revision: 6b4f49b787

Plugin loaded successfully
```

- When no pinned version: Print the warning and fail because can't check latest version.
```
 N E X T F L O W   ~  version 26.03.2-edge

WARN: Failed to prefetch plugin metadata - Unable to connect to https://registry.nextflow.io/api/v1/plugins/dependencies?plugins=nf-schema&nextflowVersion=26.03.2-edge - cause: null
ERROR ~ Cannot find latest version of nf-schema plugin

```

modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
```diff
@@ -83,7 +83,14 @@ class HttpPluginRepository implements PrefetchUpdateRepository {
     @Override
     void prefetch(List<PluginRef> plugins) {
         if (plugins && !plugins.isEmpty()) {
-            this.plugins = fetchMetadata(plugins)
+            try {
+                this.plugins = fetchMetadata(plugins)
+            }
+            catch (Exception e) {
+                log.warn "Failed to prefetch plugin metadata - ${e.message}"
+                log.debug "Plugin metadata prefetch error details", e
```
pditommaso: Let's keep only warn

MatthiasZepper:
Thanks for addressing this problem, it was indeed causing an issue for us at NGI (me) when setting up a recent Nextflow version on our HPC. 

Already in its current form, this will likely work perfectly for us. However, since there is already an environment variable `NXF_DISABLE_CHECK_LATEST`, you could optionally extent its reach to not only disable checking for the latest Nextflow version, but also for the latest plugin versions? 

That would introduce a 4th case:  _When no pinned version, but `NXF_DISABLE_CHECK_LATEST=true` : Print a warning and try using the most recent locally available version of the plugin._

I see why allowing this may cause hard to debug issues (incompatible Nextflow/Plugin combinations?), but also reckon that few developers and almost no users will actually track the minimally required plugin version. If somebody is already keeping their `NXF_HOME` around for a while and adding to it, it will likely contain several older versions of a plugin and the most recent of them may actually work. Giving the user the power to launch a "_Try with what is locally available_" run seems helpful in certain edge cases such as a read-only shared plugin directory on an HPC?

pditommaso:
I wonder if NXF_OFFLINE should be honoured and therefore the call to the registry skipped at all

bentsherman:
> I wonder if NXF_OFFLINE should be honoured and therefore the call to the registry skipped at all

It is honored, but NXF_OFFLINE also causes local git repositories to not work: #6988

In fact it might be simpler to fix that issue. Then you could just run with `NXF_OFFLINE=true`

bentsherman (changes requested):
I'm starting to think this PR is the wrong approach. I think we could standardize on the following pattern for plugins, modules, and pipelines:

- If `NXF_OFFLINE=true`, use only the local cache and fail if something is not available locally
- Otherwise, use the local cache but always check remote source for latest version

This already works for plugins, we just need to fix #6988 for pipelines. I think modules also effectively work this way even though they don't explicitly check NXF_OFFLINE

bentsherman:
Closing in favor of #7035

PR #7020, opened 2026-04-09
Fix resolution of config profiles with strict syntax parser

### LINKED ISSUE #4960: Order of 'includeConfig' and 'profiles' in nextflow.config seems to affect nested property overrides
Hi Nextflow team,

I expected that the order in which 'includeConfig' and 'profiles' are declared in nextflow.config would have no effect on property values.
However, I think I'm seeing that 'profiles' must be declared _after_ any 'includeConfig' statements, in order for profile overrides to work correctly. In particular, this seems to affect what I'd describe as nested properties. 

Given the following two files:

nextflow.config:

```nextflow
profiles {
    hpc {
        process {

         process_nested.value = "hpc.config"
         process_nonest       = "hpc.config"
       
         withLabel: "labelExample" {
             memory = 400
        }
       }
    }
}

includeConfig "memory.config"

```

and memory.config:

```nextflow
process {
    process_nested.value = "memory.config"
    process_nonest       = "memory.config"

    withLabel: "labelExample" {
        memory = 4
    }
}

```

and using the following command line invocation:

`nextflow config -profile hpc`

The nested properties 'process_nested.value' and 'labelExample' fail to be overridden by the hpc profile, while 'process_nonest' is correctly overridden.

(Of note, the problematic overrides work as expected once the 'includeConfig' declaration is moved _above_ the 'profiles' declaration)

### Program output 

the unexpected output is as follows:

```nextflow
process {
   process_nested {
      value = 'memory.config'
   }
   process_nonest = 'hpc.config'
   withLabel:labelExample {
      memory = 4
   }
}
```

(No .nextflow.log is generated by nextflow config) 

### Environment 

* Nextflow version: Both 23.10.1 and 23.04.4
* Java version: Java/17.0.4
* Operating system: Linux
* Bash version: GNU bash, version 4.2.46(2)-release

### Additional context

The example provided is for the nextflow config command, but I'm also seeing similar behavior for nextflow run

bentsherman:
Fix #4960 

This PR fixes the v2 config parser to apply profiles only after the entire config file is parsed, so that profiles are not overridden by subsequent config blocks. This behavior is consistent with the v1 config parser.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/config/ConfigParser.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/config/ConfigParser.groovy
@@ -33,13 +33,6 @@ interface ConfigParser {
     ConfigParser setIgnoreIncludes(boolean value)
 
     /**
-     * Toggle whether to strip secrets when rendering the config.
-     *
-     * @param value
-     */
-    ConfigParser setStripSecrets(boolean value)
-
-    /**
      * Toggle whether to render the source code of closures.
      *
      * @param value
@@ -52,6 +45,13 @@ interface ConfigParser {
      * @param value
      */
     ConfigParser setStrict(boolean value)
+
+    /**
+     * Toggle whether to strip secrets when rendering the config.
+     *
+     * @param value
+     */
+    ConfigParser setStripSecrets(boolean value)
 
     /**
      * Define variables which will be made available to the config script.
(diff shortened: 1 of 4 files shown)


bentsherman:
This fix is not actually needed. The v1 config parser applies profiles after the `profiles` block, so the v2 config parser already behaves correctly in this regard.

PR #7024, opened 2026-04-10
Apply socket timeout to S3 CRT connections

jorgee:
This pull request updates how the `socket_timeout` configuration is handled for AWS S3 transfers using the CRT client in the SDK v2 integration. The main improvement is that when `socket_timeout` is set, it is now translated into a CRT connection health check, allowing the detection and closure of stalled S3 connections. The documentation and tests have also been updated to reflect this new behavior.

**AWS S3 CRT Client Timeout Handling:**
- The `socket_timeout` property for S3 transfers using the CRT client is now mapped to a connection health check that enforces a minimum throughput of 1 byte per second within the specified timeout period, instead of being ignored. This helps detect and close stalled CRT connections.

**Documentation Updates:**
- The user guide (`aws-java-sdk-v2.md`) and configuration reference (`config.md`) have been updated to clarify the new behavior of the `aws.client.socketTimeout` option for CRT-based S3 transfers, including the lack of a default and the new health check mechanism. [[1]](diffhunk://#diff-8dc76da53116aee8da6debfb02777d7ca4c4281195cf1a07ba1c60b343b482e7R19-R24) [[2]](diffhunk://#diff-9ae244d0598c1cc7efc9875baf25a5f8b4bc13aabad4fbe17dd4bc5cfe856909L248-R248)

**Testing Enhancements:**
- The test suite for S3 client configuration now verifies that the CRT health configuration is set with the correct minimum throughput and timeout values when `socket_timeout` is specified.

**Dependency and Import Adjustments:**
- Added the necessary import for `S3CrtConnectionHealthConfiguration` to support the new health check configuration.
- Cleaned up unused imports in the S3 client configuration test file.

### DIFF
--- a/docs/guides/aws-java-sdk-v2.md
+++ b/docs/guides/aws-java-sdk-v2.md
@@ -16,6 +16,12 @@ The HTTP client in SDK v2 does not support overriding certain advanced HTTP opti
 - `aws.client.socketSendBufferSizeHint`
 - `aws.client.userAgent`
 
+## Connection timeouts
+
+The `aws.client.connectionTimeout` option continues to work as before, controlling the timeout for establishing new connections.
+
+The `aws.client.socketTimeout` option behaves differently depending on the HTTP client. For synchronous connections, it applies directly as a socket timeout with a default of `30000` ms. For S3 transfers using the CRT client, no default is applied; when explicitly set, this value is translated into a CRT connection health check that enforces a minimum throughput of 1 byte per second within the specified timeout period. This allows Nextflow to detect and close stalled CRT connections that are no longer making progress.
+
 ## Parallel S3 operations
 
 Nextflow manages S3 transfers, including uploads, downloads, and S3-to-S3 copies, separately from other S3 API calls such as listing a directory or retrieving object metadata.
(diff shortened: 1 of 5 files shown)


docs/guides/aws-java-sdk-v2.md
```diff
@@ -16,6 +16,12 @@ The HTTP client in SDK v2 does not support overriding certain advanced HTTP opti
 - `aws.client.socketSendBufferSizeHint`
 - `aws.client.userAgent`
 
+## Connection timeouts
+
+The `aws.client.connectionTimeout` option continues to work as before, controlling the timeout for establishing new connections.
+
+The `aws.client.socketTimeout` option behaves differently depending on the HTTP client. For synchronous connections, it applies directly as a socket timeout with a default of `30000` ms. For S3 transfers using the CRT client, no default is applied; when explicitly set, this value is translated into a CRT connection health check that enforces a minimum throughput of 1 byte per second within the specified timeout period. This allows Nextflow to detect and close stalled CRT connections that are no longer making progress.
```
bentsherman: If we know the default for the sync client is 30s, can we just apply that default manually to the CRT client as well? Then we could simply say that the default is 30s across the board

jorgee: I have done this in the latest commit and updated docs

bentsherman (approved):

PR #7026, opened 2026-04-13
Exclude params from resolved config text

bentsherman:
This PR removes the `params` scope from the resolved config text that is sent to Seqera Platform

The params do not need to be shown in the resolved config because they are already sent separately in `makeBeginReq()`.

Also, the params in the resolved config can differ from the actual params if the script defines params with default values, which will not be reflected in the resolved config. Removing the params from the resolved config text removes this potential source of confusion


robsyme:
I think this is a great move to remove the (potentially misleading) parameter configuration from the package that Nextflow sends to Platform, but is there scope here to also send the _actual_ resolved parameters (after -params-file etc)?

PR #7030, opened 2026-04-14
docs: Restructure modules as a section and add registry steps

christopher-hakkaart:
- Create a top-level section for modules in the sidebar
- Split into a similar structure to plugins
- Create clear guided pages for shared modules
- Add page about registry


docs/reference/cli.md
```diff
@@ -1377,45 +1281,183 @@ The `module` command provides a comprehensive system for managing registry-based
...
-
 (cli-module-validate)=
 
 `validate [options] <namespace/name or path>`
 
 : Validate a module before publishing to the registry.
 : Verifies that all required files are present (`main.nf`, `meta.yml`, `README.md`) and that the module spec contains all required fields (name, version, description, license).
 
-: **Examples:**
+**Examples**
```
bentsherman, line 1291: I'm divided on whether it's better to show all the module CLI examples together or separately per subcommand. I'm wondering if we should just not show them here at all and incorporate them into the reference docs. Or, maybe your current approach would work better if we just trim down the examples

What do you think?

docs/index.md
```diff
@@ -87,6 +87,17 @@ sharing
...
+```{toctree}
+:hidden:
+:caption: Modules
+:maxdepth: 1
+
+modules/modules
+modules/using-modules
+modules/developing-modules
+modules/module-registry
+```
```
bentsherman, line 99: I think this is the right direction, since modules have effectively become first-class entities like pipelines via `nextflow module run`

To avoid confusion, I think I will try to remove or significantly trim down the existing Modules page. The Scripts page already has a blurb with links to workflows / processes / modules under "Script definitions", so we could always expand that section if needed. Let me see how it works out

docs/reference/cli.md
christopher-hakkaart: I wasn't particularly happy with this either. Examples shown under the subcommands break up the long list, but they also disrupt the page's flow. However, modules were the only command with examples under subcommands, too, so I moved them to the bottom to keep every section on the page the same structure.

It's another discussion, but I've been thinking about a shift to each CLI option having its own page, similar to https://cli.github.com/manual/gh

I like being able to command f things on a page, but this could break up the page, and we could do something interesting with lists or cards, and eventually automations to update.

In the meantime, we can limit the examples to avoid overload.

docs/index.md
christopher-hakkaart: Sounds good 👍

docs/reference/cli.md
bentsherman: I think having a separate page per command would be smart. For now I will just trim the examples

christopher-hakkaart: Sounds good - can airmark for after BOS Summit

bentsherman: Trimming the examples was easy since most of them are already covered in the usage docs

docs/index.md
bentsherman: I manage to incorporate the old Modules page into the new docs

The new module docs are much more focused on the `module` CLI, whereas the old docs were more a generic description of using modules. Likely could be improved but I think it's a good start for now

A lot of that old content has been deleted or condensed in the new docs, so it's more focused on how modules are typically used rather than what is theoretically possible

I also deleted most of the module docs in the CLI usage page because it just duplicates the "Using modules" page with some unnecessary fluff

docs/modules/developing-modules.md
```diff
@@ -0,0 +1,234 @@
...
+```
+modules/
+└── myorg/
+    └── my-module/
+        ├── .module-info     # Integrity checksum (generated at install/creation)
+        ├── README.md        # Documentation (required for publishing)
+        ├── main.nf          # Module script (required)
+        ├── meta.yml         # Module spec (required for publishing)
+        ├── resources/       # Optional: module resources
+        └── templates/       # Optional: process script templates
```
christopher-hakkaart:
```suggestion
└── myorg/
    └── my-module/
        ├── .module-info     # Integrity checksum
        ├── README.md        # Documentation
        ├── main.nf          # Module script
        ├── meta.yml         # Module spec
        ├── resources/       # Optional: Module resources
        └── templates/       # Optional: Process script templates
```
The sentence above describes this as the registry structure, and the sentence below says it does not need to be followed for not published modules. So I'm not sure if we need the required after these points, as everything is framed as for publishing already.

docs/index.md
christopher-hakkaart: This looks really good to me.

Focusing on the CLI makes sense to me. I think the categorization is clear. A lot of the generic descriptions are probably legacy DSL1 to DSL2 content, which I think we are past now.

I would add back the module aliases and multiple inclusions on the using modules page. I think they are common enough use cases. Unless I missed those somewhere else?

bentsherman: Added links to the include syntax

bentsherman (approved):

Issue #5653, opened 2025-01-07
Poor File Staging/Unstaging Performance with Google Batch Executor


KatieAtGordian:
## Bug report 

I am documenting this problem as a known issue with [discussion here in slack](https://nextflow.slack.com/archives/C02TLU7U57F/p1735328326105149).

There is **no plan to support the fix to this issue** (per slack discussion, the solution is "too clunky"), but I would like to document it so that others encountering this may chime in and/or save some time debugging why their google batch pipelines are taking so long to complete in comparison to other executors.

### Expected behavior and actual behavior

**Expected behavior**:
File staging and unstaging using the default, built-in support for the Google Batch executor should complete in a reasonable amount of time. "Reasonable" means transfer speeds comparable to standard GCP file transfer achieved by using, e.g., `gcloud storage` or `gsutil`.

**Actual behavior**:
File unstaging/delocalization using the built-in support in the Google Batch executor is extremely slow, e.g., taking ~11 hours for 600GB of data. This is because it's using `gcsfuse`, which is significantly slower than `gsutil` which is used by the Google LifeSciences executor (via the `nxf_gs_upload` function in `.command.run`).

The officially recommended solution is to use Wave/Fusion; however, this solution is not appropriate for all users, especially those who cannot support the injection of 3rd party services into their workflows.

This is problematic especially because (1) poor performance is essentially the default behavior for the executor (as in, wave/fusion are the recommended solution but are not enabled by default), and (2) this reflects a disparity with the lifesciences executor, which is deprecated by GCP and will not be available after July 8, 2025.

### Steps to reproduce the problem

* Config
  * `process.executor = 'google-batch'`
  * `fusion.enabled = false` or omit from config
  * `wave.enabled = false` or omit from config
* Run any process that generates at least a few GB of data.
  * See the example process below. Running this with 20GB of data takes ~27 mins with defaults and ~12 mins with a `gsutil`-enabled solution I wrote in a fork (~3-4 mins of this is generating the file).) 

Here's an example process for testing purposes that just writes a file of the given size.

```
process DUMMY_WRITE {
    label 'process_single'
    disk { 2 * file_size_gb as nextflow.util.MemoryUnit * 1024**3 }
    

    input:
    val file_size_gb

    output:
    path 'dummy_dir/dummy.txt', emit: ch_dummy

    publishDir (
        path: "${params.publish_dir}/",
        mode: 'copy',
        )

    script:
    """
    # Write a file of size file_size_gb
[...]

PR #7031, opened 2026-04-14
feat(nf-google): add configurable CLI staging transports and safer default unstaging for Google Batch

tomiles:
This PR addresses [#5653](https://github.com/nextflow-io/nextflow/issues/5653) by adding optional Google Batch stage-in/stage-out copy transports (`posix`, `gcloud`, `gsutil`) and aligning transfer behavior with other cloud executors while preserving backward-compatible defaults.

This design allows granular optimization of staging behavior at the process level (via `stageInMode` / `stageOutMode`), rather than forcing a single workflow-wide override. In practice, only processes that use `copy` for stage-in or stage-out (stage-out defaults to `copy` in Google Batch when unset) and have the CLI tools in their runtime environment will use the new transport path, while other modes keep existing behavior.

## What changed
- Added `google.batch.stageInCopyTransport` and `google.batch.stageOutCopyTransport` to select override for transfer transport in `copy` mode.
- Added Google Batch transfer options:
  - `google.batch.gcloudCli`
  - `google.batch.gsutilCli`
  - `google.batch.maxParallelTransfers`
  - `google.batch.maxTransferAttempts`
  - `google.batch.delayBetweenAttempts`
- Added/updated Google Batch bash staging helpers with CLI tool first behavior and POSIX fallback.
- Folded in the #6917 behavior: default Google Batch unstaging to `copy` when `stageOutMode` is unset to avoid `move` failures in gcsfuse-mounted work-dir workflows (e.g. overlapping outputs/symlinked paths).

## Docs
- Updated `docs/google.md` with the new transport options and default unstaging rationale.
- Updated `docs/reference/config.md` with all new `google.batch.*` config entries and defaults.

## Acknowledgements
Thanks to [@dmvanbeek](https://github.com/dmvanbeek/) and [@rhassaine](https://github.com/rhassaine) for input and hackathon discussions that helped shape this approach.

### DIFF
--- a/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchScriptLauncher.groovy
+++ b/plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchScriptLauncher.groovy
@@ -58,11 +58,7 @@ class GoogleBatchScriptLauncher extends BashWrapperBuilder implements GoogleBatc
     protected GoogleBatchScriptLauncher() {}
 
     GoogleBatchScriptLauncher(TaskBean bean, Path remoteBinDir, BatchConfig batchConfig) {
-        // Unstaging is cross-device on Google Batch (gcsfuse-mounted work dir).
-        // `move` can fail with overlapping outputs or symlinked paths.
-        if( bean.stageOutMode == null )
-            bean.stageOutMode = 'copy'
-        super(bean, copyStrategyFor(bean, batchConfig))
+        super(defaultStageOutMode(bean), copyStrategyFor(bean, batchConfig))
         // keep track the google storage work dir
         this.remoteWorkDir = (CloudStoragePath) bean.workDir
         this.remoteBinDir = toContainerMount(remoteBinDir)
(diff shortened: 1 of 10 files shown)


plugins/nf-google/src/main/nextflow/cloud/google/batch/GoogleBatchScriptLauncher.groovy
```diff
@@ -101,6 +104,20 @@ class GoogleBatchScriptLauncher extends BashWrapperBuilder implements GoogleBatc
...
+    private static ScriptFileCopyStrategy copyStrategyFor(TaskBean bean, BatchConfig batchConfig) {
+        batchConfig.usesGoogleBatchStaging()
+                ? new GoogleBatchFileCopyStrategy(bean, batchConfig)
+                : new SimpleFileCopyStrategy(bean)
+    }
+
+    private static TaskBean defaultStageOutMode(TaskBean bean) {
+        // Unstaging is cross-device on Google Batch (gcsfuse-mounted work dir).
+        // `move` can fail with overlapping outputs or symlinked paths.
+        if( bean.stageOutMode == null )
```
dchaley: Does it make sense to throw an error here – or somewhere – if the stageOutMode is set to `move` ? Are there cases where `move` to cloud storage makes sense if you know you don't have overlaps or symlinks?

PR #7035, opened 2026-04-15
Feature: Allow cloning from local Git repositories when `--offline`

### LINKED ISSUE #6006: Offline mode (NXF_OFFLINE='true') prevents running pipelines from local file:/ bare Git repositories

### LINKED ISSUE #6988: `nextflow run -offline` should not affect file-system local repositories

MatthiasZepper:
close #6988

This pull requests suggests a change to the `AssetManager` to allow cloning from local SCM repositories even when `NXF_OFFLINE=true`. 

Feel free to choose a different implementation, since I am unfamiliar with the codebase of Nextflow and Groovy in general and relied heavily on AI to propose this contribution. 

Furthermore, I requested to generate extended tests for this functionality, but ultimately refrained from including them in this PR, because without Groovy knowledge, I was unsure if those tests are meaningful at all. You can find the extended tests in [this commit](https://github.com/nextflow-io/nextflow/commit/f800abd847df39f78618cb76b22e1619899873d6). Let me know if I should cherry-pick that commit and add it to this  PRs branch as well.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
@@ -532,15 +532,6 @@ class AssetManager implements Closeable {
         return strategy?.getGitRepositoryUrl() ?: provider.getCloneUrl()
     }
 
-    /**
-     * @return {@code true} when the SCM source points to a local file-system repository.
-     */
-    boolean isLocalScmSource() {
-        if( provider instanceof LocalRepositoryProvider )
-            return true
-        return false
-    }
-
     File getLocalPath() {
         return strategy?.getLocalPath()
     }
@@ -655,6 +646,13 @@ class AssetManager implements Closeable {
 
     boolean isLocal() {
         return localPath && localPath.exists()
+    }
+
+    /**
+     * @return {@code true} when the SCM source points to a local file-system repository.
+     */
+    boolean isLocalScmSource() {
+        return provider instanceof LocalRepositoryProvider
     }
 
     /**
--- a/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
@@ -539,7 +539,7 @@ class AssetManager implements Closeable {
         if( provider instanceof LocalRepositoryProvider )
             return true
 
-        final repoUrl = provider?.getRepositoryUrl()
+        final repoUrl = provider?.getRepositoryUrl() ?: provider?.getCloneUrl()
         if( !repoUrl )
             return false
 
--- a/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
@@ -536,15 +536,17 @@ class AssetManager implements Closeable {
      * @return {@code true} when the SCM source points to a local file-system repository.
      */
     boolean isLocalScmSource() {
-        final repoUrl = getGitRepositoryUrl()
+        if( provider instanceof LocalRepositoryProvider )
+            return true
+
+        final repoUrl = provider?.getRepositoryUrl()
         if( !repoUrl )
             return false
-        try {
-            return new GitUrl(repoUrl).protocol == 'file'
-        }
-        catch( IllegalArgumentException e ) {
-            return false
-        }
+
+        if( repoUrl.startsWith('file:') )
+            return true
+
+        return new File(repoUrl).isAbsolute()
     }
 
     File getLocalPath() {
--- a/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
@@ -532,6 +532,21 @@ class AssetManager implements Closeable {
         return strategy?.getGitRepositoryUrl() ?: provider.getCloneUrl()
     }
 
+    /**
+     * @return {@code true} when the SCM source points to a local file-system repository.
+     */
+    boolean isLocalScmSource() {
+        final repoUrl = getGitRepositoryUrl()
+        if( !repoUrl )
+            return false
+        try {
+            return new GitUrl(repoUrl).protocol == 'file'
+        }
+        catch( IllegalArgumentException e ) {
+            return false
+        }
+    }
+
     File getLocalPath() {
         return strategy?.getLocalPath()
     }
--- a/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
@@ -538,15 +538,7 @@ class AssetManager implements Closeable {
     boolean isLocalScmSource() {
         if( provider instanceof LocalRepositoryProvider )
             return true
-
-        final repoUrl = provider?.getRepositoryUrl() ?: provider?.getCloneUrl()
-        if( !repoUrl )
-            return false
-
-        if( repoUrl.startsWith('file:') )
-            return true
-
-        return new File(repoUrl).isAbsolute()
+        return false
     }
 
     File getLocalPath() {
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy
@@ -620,10 +620,11 @@ class CmdRun extends CmdBase implements HubOptions {
             if( revision )
                 manager.setRevision(revision)
             def repo = manager.getProjectWithRevision()
+            final localScm = manager.isLocalScmSource()
 
             boolean checkForUpdate = true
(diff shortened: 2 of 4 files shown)


bentsherman:
@jorgee I think this might be a better alternative to #7019 . Let me know what you think

jorgee:
This PR looks fine to me. I will review the proposed test and maybe add a couple of them to test the changes in the CmdRun part.

Regarding the alternative with #7019, I think they aren't mutually exclusive. #7019 allows to continue running pipelines when there is a problem fetching metadata but all plugins are pinned and already in the local cache. Maybe, I would just merge the initial commit (I tried to add a local plugin fallback in the last commit that break some test). It could make runs more reliable without adding complex code. Anyway I am also OK merging just this one to keep things simple and to avoid possible side effects of not fetching plugin metadata.

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
```diff
@@ -532,6 +532,23 @@ class AssetManager implements Closeable {
...
+            return true
+
+        final repoUrl = provider?.getRepositoryUrl() ?: provider?.getCloneUrl()
+        if( !repoUrl )
+            return false
+
+        if( repoUrl.startsWith('file:') )
+            return true
+
+        return new File(repoUrl).isAbsolute()
```
jorgee: Not sure if all this checks are really needed. It should return true if provider is local, otherwise false

jorgee: Looking at the code flow in CmdRun.getScriptFile0():                                                                                                                                                      
                                                                                                                                                                                                                          
  1. new AssetManager(pipelineName, this) is called (line 619)                                                                                                                                                            
  2. This calls resolveName(pipelineName) which handles file:/ URLs via resolveNameFromGitUrl() (AssetManager.groovy:459-494)
  3. For file: URLs, it creates a ProviderConfig with path:url.domain (line 471), which triggers attr.platform = 'file' (ProviderConfig.groovy:97-98)                                                                     
  4. createHubProvider(hub) then creates a LocalRepositoryProvider via RepositoryFactory                                                                                                                                  
                                                                                                                                                                                                                          
  So by the time isLocalScmSource() is called, the provider is always a LocalRepositoryProvider for any local filesystem source. The URL-based fallback checks (file: prefix, File.isAbsolute()) are dead code — there's  
  no real scenario where the provider is not a LocalRepositoryProvider but the URL still points to a local path.

jorgee: Simplified and update tests

jorgee (approved):
Solution is good! 
I have just cleaned some unnecessary checks. See [this comment](https://github.com/nextflow-io/nextflow/pull/7035#discussion_r3092429549). I have also updated tests including two new test for CmdRun.getScriptFile0()

bentsherman:
Thanks Jorge. I will review and merge later

> Regarding the alternative with #7019, I think they aren't mutually exclusive. #7019 allows to continue running pipelines when there is a problem fetching metadata but all plugins are pinned and already in the local cache.

I think the alternative to #7019 is to run with `NXF_OFFLINE=true`. It just so happens that Matthias also uses "remote" pipelines that are actually local via `file://...`. Fixing that problem should will make the `NXF_OFFLINE` usage more robust across the board -- modules, pipelines, and plugins.

modules/nextflow/src/main/groovy/nextflow/scm/AssetManager.groovy
MatthiasZepper: Thanks for the detailed explanation. That makes total sense and is exactly why I felt uneasy with relying so heavily on AI without actual understanding of the codebase.

PR #7045, opened 2026-04-17
Fix resolution of `-with-tower` with `TOWER_API_ENDPOINT`

### LINKED ISSUE #7044: The TOWER_API_ENDPOINT environment variable is not used when running a pipeline with the -with-tower flag
## Bug report 

We have an on-premises installation of the Seqera Platform and want to use it to monitor pipeline runs using `nextflow` on the CLI with the `-with-tower` flag. 

The docs say that the environment variable `TOWER_API_ENDPOINT` can be used to configure a custom endpoint but it does not seem to work. Setting the endpoint directly on the command line or via a `nextflow.config` file does the job.

We run a test pipeline like:

```
export TOWER_ACCESS_TOKEN=myT0k3n
export TOWER_API_ENDPOINT=https://platform.my.org/api

nextflow run hello -with-tower
```

and would expect `nextflow` to use the configured endpoint but it uses the default endpoint from Seqera. The `.nextflow.log` file says:
```
Apr-17 15:07:37.050 [main] DEBUG io.seqera.tower.plugin.TowerClient - Creating Seqera Platform observer -- endpoint=https://api.cloud.seqera.io; requestInterval=1s; aliveInterval=1m; maxRetries=5; backOffBase=0; backOffDelay=0
``` 

while it should be `endpoint=https://platform.my.org/api`

```
export TOWER_ACCESS_TOKEN=myT0k3n
export TOWER_API_ENDPOINT=https://platform.my.org/api

nextflow run hello -with-tower
```

### Program output 

Got:
```
 N E X T F L O W   ~  version 25.10.4

Launching `https://github.com/nextflow-io/hello` [confident_gautier] DSL2 - revision: d828daeef7 [master]

executor >  local (4)
[5b/d3e138] process > sayHello (1) [100%] 4 of 4 ✔
Bonjour world!

Hola world!

Hello world!

Ciao world!
```

Expected:
```
 N E X T F L O W   ~  version 25.10.4

Launching `https://github.com/nextflow-io/hello` [big_liskov] DSL2 - revision: d828daeef7 [master]

Monitor the execution with Seqera Platform using this URL: https://platform.my.org/user/myuser/watch/foo
executor >  local (4)
[9b/ddf9ac] process > sayHello (1) [100%] 4 of 4 ✔
Hola world!

Hello world!

Ciao world!

Bonjour world!
```

### Environment 

* Nextflow version: 25.10.4 build 11173
* Java version: OpenJDK 64-Bit Server VM 21.0.10+7-Ubuntu-124.04
* Operating system: Linux

bentsherman:
Close #7044 

Previously, the `ConfigBuilder` assigned defaults for all of the `-with-*` CLI options. Now that these config settings are accessed through config classes like `TowerConfig` and `WaveConfig`, default values should only be assigned there instead of `ConfigBuilder`

The report options like `-with-report` were already corrected. This PR applies the same refactor to `-with-tower` and `-with-wave`

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/config/ConfigBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/config/ConfigBuilder.groovy
@@ -718,8 +718,6 @@ class ConfigBuilder {
             config.tower.enabled = true
             if( cmdRun.withTower != '-' )
                 config.tower.endpoint = cmdRun.withTower
-            else if( !config.tower.endpoint )
-                config.tower.endpoint = 'https://api.cloud.seqera.io'
         }
 
         // -- set wave options
@@ -729,8 +727,6 @@ class ConfigBuilder {
             config.wave.enabled = true
             if( cmdRun.withWave != '-' )
                 config.wave.endpoint = cmdRun.withWave
-            else if( !config.wave.endpoint )
-                config.wave.endpoint = 'https://wave.seqera.io'
         }
 
         // -- set fusion options
--- a/modules/nextflow/src/test/groovy/nextflow/config/ConfigBuilderTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/config/ConfigBuilderTest.groovy
@@ -854,6 +854,7 @@ class ConfigBuilderTest extends Specification {
         then: // command line should override the config file
         config.trace instanceof Map
         config.trace.enabled
+        !config.trace.file
     }
 
     def 'should set session report options' () {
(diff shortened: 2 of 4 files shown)


pditommaso:
Which versions are affected by this issue?

bentsherman:
The linked issue confirmed in for 25.10

However, I think this has always been the behavior, since the `ConfigBuilder` does not check the environment var when applying the default value. So I don't think it's a regression

pditommaso (approved):

PR #7046, opened 2026-04-17
Fix S3FileSystemProvider.newInputStream() draining full object on close

### LINKED ISSUE #7043: S3FileSystemProvider.newInputStream() drains the full object body on close, blocking for minutes on large files
#### Expected behavior
Calling `path.withInputStream { readLine() }` (or any partial read followed by `close()`) on an S3 NIO path should complete in seconds — the time needed to open the connection, read the requested bytes, and release the HTTP resource.

#### Actual behavior
When the caller does not consume the full S3 object, `close()` blocks for as long as it takes to download the remaining bytes of the object over the network. For a multi-GB file, this can block a GPars actor thread for tens of minutes. On a small head node (e.g. 2 vCPUs), this serializes the pipeline and is visible as a completely silent gap in the log (no log entries except the periodic `TaskPollingMonitor` dump every 5 minutes).

#### Root cause
`S3FileSystemProvider.newInputStream()` (`plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3FileSystemProvider.java:184`) returns the raw `ResponseInputStream<GetObjectResponse>` from `s3Client.getObject()`. When the caller calls `close()` on the NIO `InputStream`, the close propagates through:

```
BufferedInputStream.close()
  → SdkFilterInputStream.close()
    → S3ChecksumValidatingInputStream.close()
      → EofSensorInputStream.close()
        → ResponseEntityProxy.streamClosed()
          → ContentLengthInputStream.close()   ← reads remaining bytes to EOF
            → NioSocketImpl.timedRead()        ← blocks on the socket
```

Apache HTTP client's `ContentLengthInputStream.close()` drains the response body so the connection can be returned to the pool. For a small HTTP response this is unnoticeable; for a multi-GB S3 object it blocks the caller for the full download time.

The AWS SDK v2 provides `ResponseInputStream.abort()` specifically to close the HTTP connection without draining, but because `newInputStream()` returns a plain `InputStream`, the caller can only call `close()`.

Related AWS Java SDK v2 issue: https://github.com/aws/aws-sdk-java-v2/issues/2117

#### How to reproduce
Use any S3 NIO path with a large object (e.g. a multi-GB FASTQ) and call any pattern that reads partial content and closes — the most common form is:

```groovy
def path = file('s3://<bucket>/<multi-GB-object>')
path.withInputStream { is ->
    new BufferedReader(new InputStreamReader(new GZIPInputStream(is))).readLine()
}
// This withInputStream call blocks for the full download time of the object
// instead of returning after reading the first line.
```

This is exactly what nf-core/sarek does in `addReadgroupToMeta()`/`readFirstLineOfFastq()` to extract the flowcell ID from the first FASTQ line. On a resumed run from a GCP head node against S3 source files, this produces a ~30 minute blocking call per file.

#### Stack trace captured during the hang

```
Thread[#107,Actor Thread 27,5,main]
  sun.nio.ch.Net.poll(Native Method)
  sun.nio.ch.NioSocketImpl.timedRead(NioSocketImpl.java:280)
  sun.nio.ch.NioSocketImpl.read(NioSocketImpl.java:346)
  sun.security.ssl.SSLSocketInputRecord.read(SSLSocketInputRecord.java:489)
  ...
  org.apache.http.impl.io.ContentLengthInputStream.read(ContentLengthInputStream.java:176)
  org.apache.http.impl.io.ContentLengthInputStream.close(ContentLengthInputStream.java:101)
  org.apache.http.impl.execchain.ResponseEntityProxy.streamClosed(ResponseEntityProxy.java:142)
  org.apache.http.conn.EofSensorInputStream.close(EofSensorInputStream.java:172)
  software.amazon.awssdk.services.s3.internal.checksums.S3ChecksumValidatingInputStream.close(S3ChecksumValidatingInputStream.java:163)
  software.amazon.awssdk.core.io.SdkFilterInputStream.close(SdkFilterInputStream.java:83)
  java.io.BufferedInputStream.close(BufferedInputStream.java:618)
  org.codehaus.groovy.runtime.IOGroovyMethods.withStream(IOGroovyMethods.java:1195)
  org.apache.groovy.nio.extensions.NioExtensions.withInputStream(NioExtensions.java:1539)
  ...
  <user-script>.readFirstLineOfFastq(script:699)
  <user-script>.addReadgroupToMeta(script:648)
```

#### Proposed fix
In `S3FileSystemProvider.newInputStream()`, wrap the returned `ResponseInputStream` so its `close()` calls `abort()` on the underlying HTTP stream. Callers who fully consume the stream are unaffected (reads work normally via the wrapper); callers who only partially consume the stream get a fast abort instead of a multi-minute drain.

```java
final ResponseInputStream<GetObjectResponse> result = s3Path
    .getFileSystem()
    .getClient()
    .getObject(s3Path.getBucket(), s3Path.getKey());

if (result == null)
    throw new IOException(...);

return new FilterInputStream(result) {
    @Override
    public void close() {
        result.abort();
    }
};
```

#### Environment
- Nextflow version: 26.03.1-edge
- nf-amazon: 3.8.1
- Java version: 21.0.5
- Operating system: Linux (GCP head node)

jorgee:
Fixes #7043.

### Summary
`S3FileSystemProvider.newInputStream()` previously returned the raw `ResponseInputStream<GetObjectResponse>` from the AWS SDK. When a caller closed the stream without having read it to EOF (e.g. `path.withInputStream { readLine() }`), Apache HTTP client's `ContentLengthInputStream.close()` would drain the remaining response body to release the connection back to the pool. For a multi-GB S3 object this blocked the caller for many minutes. On a small head node (e.g. 2 vCPUs) the blocked actor thread serialised the pipeline, producing a completely silent gap of tens of minutes in the log (only `TaskPollingMonitor` dumps every 5 minutes).

### Fix
Wrap the returned stream so that `close()` calls `ResponseInputStream.abort()` on the underlying stream instead of triggering the Apache drain. The AWS SDK v2 `ResponseInputStream` exposes `abort()` exactly for this case (close the HTTP connection without draining).

### Test
Adds `S3InputStreamAbortTest` — a regression test against the public ~1GB FASTQ at `s3://ngi-igenomes/test-data/sarek/SRR7890919_WES_HCC1395BL-EA_normal_1.fastq.gz` (anonymous, eu-west-1). The test reads the first line and closes on a background thread, bounded by `Future.get(30s)` (Spock's `@Timeout`/`Thread.interrupt()` cannot unblock a native SSL read, so the bound has to be on the caller side). Without the fix the close() drain blows the 30s timeout; with the fix the test completes in seconds.

### Notes
This is a potentially user-visible semantic change: a caller that relied on `close()` to drain and discard the rest of the stream for connection reuse would now get an abort. In practice:
- Callers that consume the whole stream before closing are unaffected (reads still flow through the wrapper normally).
- Connection reuse still works — `abort()` closes the pooled connection instead of reusing it, which for an S3 NIO provider accessing arbitrary large objects is the correct trade-off.

### DIFF
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3FileSystemProvider.java
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3FileSystemProvider.java
@@ -215,6 +215,10 @@ public class S3FileSystemProvider extends FileSystemProvider implements FileSyst
 			public void close() {
 				result.abort();
 			}
+            // Just-used for testing
+            void abort(){
+                result.abort();
+            }
 		};
 	}
 
--- a/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3FileSystemProvider.java
+++ b/plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3FileSystemProvider.java
@@ -17,6 +17,7 @@
 package nextflow.cloud.aws.nio;
 
 import java.io.ByteArrayInputStream;
+import java.io.FilterInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -59,6 +60,7 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
+import software.amazon.awssdk.core.ResponseInputStream;
 import software.amazon.awssdk.services.s3.model.*;
 import software.amazon.awssdk.services.s3.model.S3Object;
 import com.google.common.base.Preconditions;
@@ -194,7 +196,7 @@ public class S3FileSystemProvider extends FileSystemProvider implements FileSyst
 		Preconditions.checkArgument(!s3Path.getKey().equals(""),
 				"cannot create InputStream for root directory: %s", FilesEx.toUriString(s3Path));
 
-        InputStream result = s3Path
+        final ResponseInputStream<GetObjectResponse> result = s3Path
             .getFileSystem()
             .getClient()
             .getObject(s3Path.getBucket(), s3Path.getKey());
@@ -202,7 +204,18 @@ public class S3FileSystemProvider extends FileSystemProvider implements FileSyst
         if (result == null)
 			throw new IOException(String.format("The specified path is a directory: %s", FilesEx.toUriString(s3Path)));
 
-		return result;
+		// Wrap the response stream so that close() aborts the underlying HTTP connection
+		// instead of draining the remaining bytes. Apache HTTP client's ContentLengthInputStream.close()
+		// reads to end-of-stream to release the connection back to the pool, which for a large S3
+		// object (e.g. a multi-GB FASTQ) can block the caller for many minutes. Callers of
+		// newInputStream() typically do not consume the whole object, so abort() is the correct
+		// semantics here.
+		return new FilterInputStream(result) {
+			@Override
+			public void close() {
+				result.abort();
+			}
+		};
 	}
 
 	@Override
--- a/plugins/nf-amazon/src/test/nextflow/cloud/aws/nio/S3InputStreamAbortTest.groovy
+++ b/plugins/nf-amazon/src/test/nextflow/cloud/aws/nio/S3InputStreamAbortTest.groovy
@@ -1,0 +1,135 @@
+/*
+ * Copyright 2013-2026, Seqera Labs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


bentsherman:
> This is a potentially user-visible semantic change: a caller that relied on close() to drain and discard the rest of the stream for connection reuse would now get an abort.

The only documented way to stream files in the Nextflow std lib is [withReader](https://docs.seqera.io/nextflow/reference/stdlib-types#withreader-action-bufferedreader----), which provides a BufferedReader in a closure and automatically closes it

So if anyone is calling `close()`, they are operating outside the standard library anyway

bentsherman (approved):

PR #7052, opened 2026-04-20
Rename `module info` subcommand to `module view`

pditommaso:
## Summary
- Rename the `nextflow module info` subcommand to `nextflow module view`
- Keep `info` registered as a backward-compatible alias so existing usage keeps working
- Introduce a generic `CmdBase.getAliases()` hook and thread it through `CmdModule` so both JCommander parsing and the usage/help lookup resolve aliases
- Update the CLI reference, module guide and 26.04 migration notes to use the new name

## Rationale
All other `nextflow module` subcommands are named after a verb describing the action they perform: `create`, `install`, `run`, `list`, `remove`, `search`, `publish`, `spec`, `validate`. `info` was the odd one out — a noun describing the output rather than the action. Renaming it to `view` aligns it with the rest of the command surface and makes the CLI read consistently (`nextflow module <verb> ...`). The legacy `info` alias is retained to avoid breaking users and scripts already relying on the existing name.

## Test plan
- [ ] `./gradlew :nextflow:test --tests "nextflow.cli.module.CmdModuleInfoTest"`
- [ ] Manual: `nextflow module view <scope>/<name>` works
- [ ] Manual: `nextflow module info <scope>/<name>` still works (alias)
- [ ] Manual: `nextflow module -h` lists the command under the new name


jorgee:
Docs are still with `info`. Shouldn't be updated to `inspect`?

pditommaso:
Updated to `view` (including docs)

bentsherman (approved):

PR #7056, opened 2026-04-20
Add $schema ref to generated module spec

bentsherman:
This PR adds a `$schema` property to the module spec which denotes the module schema

The generated module spec always uses the new schema: https://github.com/nextflow-io/schemas/blob/main/module/v1/schema.json

So if the `$schema` isn't specified, you can assume it is the nf-core schema:
https://github.com/nf-core/modules/blob/master/modules/meta-schema.json


pditommaso (commented):
Just realised, there's an implicit version on the path however the schema version is not defined explicitly anywhere

bentsherman:
It's part of the `$id` property in the schema:

https://github.com/nextflow-io/schemas/blob/d0d2e56863327e430008693e2ce7bea9ad1d1466/module/v1/schema.json#L3

I don't think there is a separate version property in JSON schema

pditommaso (approved):
Fair enough

PR #7060, opened 2026-04-21
Update workflow outputs tutorial

bentsherman:
- Update workflow outputs tutorial to use Nextflow 25.10 and `preview-25-10` branch of rnaseq-nf
- Add recommendation to migrate to static typing + records before migrating to workflow outputs

### DIFF
--- a/docs/tutorials/workflow-outputs.md
+++ b/docs/tutorials/workflow-outputs.md
@@ -2,7 +2,11 @@
 
 # Migrating to workflow outputs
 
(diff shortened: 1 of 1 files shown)


bentsherman:
The reason for the recommendation is that using workflow outputs with tuples turns out to be quite verbose for large pipelines (see nf-core/rnaseq as an example)

So I think it's easier to do static typing + records -> workflow outputs rather than the other way around, even though workflow outputs were added first.

christopher-hakkaart (approved):
LGTM - makes sense to provide this tip up front

PR #7065, opened 2026-04-23
Replace current cloud info url call by Cloud Info client

jorgee:
This pull request updates the Google Batch machine type selection logic to use the new `lib-cloudinfo` client library instead of manual HTTP requests and JSON parsing. This contains default retries as well as connection and request timeouts.

**Dependency update:**

* Added `io.seqera:lib-cloudinfo:1.0.0` as an API dependency in `build.gradle` to enable use of the CloudInfo client library.

**Refactoring to use CloudInfo client:**

* Replaced manual HTTP requests and JSON parsing in `GoogleBatchMachineTypeSelector.groovy` with the `CloudInfoClient` API for retrieving product and pricing information.
* Imported `CloudPrice` and `CloudInfoClient` from the new library and initialized a `CloudInfoClient` instance in the selector class. [[1]](diffhunk://#diff-0f651a8bc13f65b635fae5eaca24bac94174df5abbc836fa4952b1c5ee6aa489R18-L20) [[2]](diffhunk://#diff-0f651a8bc13f65b635fae5eaca24bac94174df5abbc836fa4952b1c5ee6aa489R104-R109)
* Removed the hardcoded `CLOUD_INFO_API` URL, as the client library now handles API endpoint management.

### DIFF
--- a/plugins/nf-google/build.gradle
+++ b/plugins/nf-google/build.gradle
@@ -58,7 +58,7 @@ dependencies {
     api 'com.google.cloud:google-cloud-logging:3.23.5'
     api 'com.google.cloud:google-cloud-nio:0.128.5'
     api 'com.google.cloud:google-cloud-storage:2.58.0'
-
+    api 'io.seqera:lib-cloudinfo:1.0.0'
     // Force patched version to address CVE-2025-55163 (MadeYouReset HTTP/2 DDoS vulnerability)
     runtimeOnly 'io.grpc:grpc-netty-shaded:1.75.0'
     // Force patched version to address GHSA-72hv-8253-57qq (jackson-core Number Length Constraint Bypass DoS)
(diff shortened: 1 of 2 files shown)


pditommaso (approved):
Well done

plugins/nf-google/build.gradle
```diff
@@ -58,7 +58,7 @@ dependencies {
     api 'com.google.cloud:google-cloud-logging:3.23.5'
     api 'com.google.cloud:google-cloud-nio:0.128.5'
     api 'com.google.cloud:google-cloud-storage:2.58.0'
-
+    api 'io.seqera:lib-cloudinfo:1.0.0'
```
pditommaso, line 61: Worth checking if it pulls undesired deps

jorgee: Nothing strange. It just has` io.seqera:lib-httpx` and `io.seqera:lib-serde-jackson`

pditommaso: nice

Issue #7064, opened 2026-04-23
Add support for Apple's native container runtime on macOS


Joon-Klaps:
## New feature

Hi devs, loving nextflow! In the summer of 2025 Apple announced [`container`](https://github.com/apple/container) (Apache 2.0 license), allowing containers to run natively on macOS instead of using a hidden Linux VM.

Unlike Docker Desktop, Apple's runtime integrates directly with macOS via the Virtualization framework instead of booting a Linux VM. It uses the same OCI image format as Docker/Podman, pulls from the same registries, and supports both native `linux/arm64` images and `linux/amd64` images via Rosetta translation. The result is faster container startup, lower memory and CPU overhead, and no commercial license requirement.

## Use case 

A lot of pipeline devs and users are using Apple Silicon Macs (M1/M2/M3/M4) for running nextflow either for local development or actual runs. I believe, the only supported container engine for MacOS M* computers is Docker, which requires Docker Desktop for Apple Silicon and is not always as lightweight as we would want it to be.

With Apple's native `container` CLI available, these users could run Nextflow pipelines directly on their Mac with lower overhead and no external tooling. 

## Suggested implementation 

I believe, being able to run it similar to other executors ie `executor = 'apple_container'`,  `-with-apple-container`.

Joon-Klaps:
https://github.com/nextflow-io/nextflow/pull/7073

PR #7066, opened 2026-04-23
Adding `apple-container` as a new executor

Joon-Klaps:
(Attemps to) Adress #7064 

Claude* and I have tried to implement the [`container`](https://github.com/apple/container) as a new executor and as a mini project to learn more about apple containers and nextflow.

## Changes

- New `apple` container engine (`apple { enabled = true }`) and `apple_container` executor
- CLI flag `-with-apple-container` (analogous to `-with-docker`)
- Cross-architecture support: when `process.arch = 'x86_64'`, emits `--platform linux/amd64 --rosetta` so x86_64 images run via Rosetta
- Wave compatibility: `apple` added to the `DOCKER_LIKE` engines list in `WaveContainerResolver`, since Apple uses the same OCI format as Docker

## Testing

Tested on Apple Silicon (macOS) with the Apple `container` CLI installed.

**nf-core/rnaseq with Wave (recommended):**
```bash
NXF_PLUGINS_MODE=prod NXF_PLUGINS_DIR=$HOME/.nextflow/plugins \
  ./launch.sh run nf-core/rnaseq \
    -profile 'test',wave,arm64 \
    -with-apple-container \
    --outdir ./test-output
```
Pipeline ran end-to-end successfully (51+ processes, `containerEngine: apple` shown in run summary).

**Simple pipeline without Wave:**
```bash
./launch.sh run nextflow-io/rnaseq-nf -with-apple-container -c custom.config
```
Where `custom.config` contained `process.arch = 'x86_64'` as `container` assumes the container is `linux/arm64`. 

>*I mention claude as a declaration of the use of AI and that it has heavily assisted me in the development of the code & docs of this PR


pditommaso:
Thanks for the contribution but nor sure we need this

Joon-Klaps:
Hi @pditommaso, could you elaborate a bit more on this?

I would assume that reducing overhead would still be beneficial, even if small relative to task execution time. Is it because it doesn't translate into meaningful gains for large analyses (which would get run on  HPC/cloud), or that macOS is mostly used for development, where such optimisations don’t matter as much?

bentsherman:
I would try to implement this in a plugin instead of adding it to core. Since you just need a local executor that uses the apple container runtime, it should be possible.

You already have the executor, you just need to use the apple container runtime "under the hood" instead of exposing it is a user-facing container runtime.

https://docs.seqera.io/nextflow/plugins/developing-plugins

pditommaso:
I mean, as developer, what's benefits of this approach considering docker is mainstream ?

pditommaso:
I've thought a bit more about this, and I agree that could make sense supporting apple container, however as container engine not as an executor. Made an implementation here https://github.com/nextflow-io/nextflow/pull/7073. Thanks for opening this discussion. 

Closing in favour of https://github.com/nextflow-io/nextflow/pull/7073

PR #7070, opened 2026-04-24
Add seqera:// data-links support to nf-tower filesystem

jorgee:
## Summary

Extends the `seqera://` NIO filesystem in `nf-tower` with a second resource type, `data-links`. Paths of the form `seqera://<org>/<ws>/data-links/<provider>/<name>/<sub-path>` resolve to files and directories inside Platform-managed data-links (S3/GCS/Azure buckets or prefixes).

Listings and attribute queries go through the Platform's `/data-links/{id}/browse[/path]` endpoints; byte reads go through pre-signed URLs returned by `/data-links/{id}/generate-download-url` and fetched with a plain JDK `HttpClient`. Only the Seqera access token is required — no AWS/GCP/Azure credentials, no cloud SDK dependency is introduced.

As part of this change, the existing dataset-specific logic in `SeqeraFileSystemProvider`, `SeqeraFileSystem`, and `SeqeraPath` is extracted into a real `ResourceTypeHandler` abstraction; `DatasetsResourceHandler` and `DataLinksResourceHandler` are the two implementations. The generic `fs/` classes become resource-type-agnostic for depth ≥ 3 (enforced by `ResourceTypeAbstractionTest`).

Design artifacts: [spec.md](https://github.com/nextflow-io/nextflow/blob/260422-seqera-datalinks-fs/specs/260422-seqera-datalinks-fs/spec.md), [plan.md](https://github.com/nextflow-io/nextflow/blob/260422-seqera-datalinks-fs/specs/260422-seqera-datalinks-fs/plan.md), [ADR](https://github.com/nextflow-io/nextflow/blob/260422-seqera-datalinks-fs/adr/20260422-seqera-datalinks-filesystem.md).

## Highlights

- **Path shape**: `seqera://<org>/<ws>/data-links/<provider>/<name>/<sub-path>`. Provider segments are the lowercase `DataLinkProvider.toString()` value (`aws`, `google`, `azure`, …).
- **Generic lazy pagination via `PagedIterable<T>`**: a single shared abstraction backs both the workspace data-link list (offset paginated) and data-link content browse (token paginated). The first page is fetched eagerly so `IOException` surfaces at the call site, not at the first `Iterator.hasNext()`. Two named static fetchers (`DataLinkListFetcher`, `DataLinkContentFetcher`) own their own cursor state.
- **Reliable file-vs-directory detection**: `readAttributes` on a sub-path lists the path's **parent** directory and finds the entry by name; the entry's `type` (FILE/FOLDER) is the authoritative signal, and a missing entry → `NoSuchFileException`. The `/browse/{path}` response shape alone does not reliably distinguish file/directory/missing paths.
- **Per-path attribute caching**: listings attach `SeqeraFileAttributes` to each emitted `SeqeraPath`; the provider also writes resolved attributes back onto the path after a fresh read. Subsequent `readAttributes` calls on the same path instance hit the cache (zero API calls).
- **Single-call data-link resolution**: `getDataLink(ws, provider, name)` issues a combined keyword search (`<name> provider:<provider>`) so the server returns at most one match. `@Memoized`, including `null` misses.
- **Cached user-id on the filesystem**: `SeqeraFileSystem` holds the `TowerClient` directly and exposes `getUserId()` cached for the lifetime of the FS — the token doesn't change during a pipeline run. User/workspace lookup is shared infrastructure across resource types, not a dataset-client method.
- **`credentialsId` forwarding**: when `DataLinkDto.credentials` is non-empty, the first credential's `id` is forwarded as the `credentialsId` query parameter on browse and download-URL requests.
- **Error mapping**: 401 → `AbortOperationException`; 403 → `AccessDeniedException`; 404 → `NoSuchFileException`. Consistent with the dataset client.
- **369 unit tests** pass (Spock + `Mock(TowerClient)`). The pre-existing dataset tests are unchanged and continue to pass.

## Requirements / prerequisites

> ⚠️ **Platform permission**: the Seqera Platform user whose access token is used to run the pipeline must have a **Maintain** role (or higher) on the workspace. Lower roles (e.g. View) cannot list/browse data-links through the Platform API and will see `AccessDeniedException` on any `seqera://<org>/<ws>/data-links/...` path.

- `nf-tower` plugin must be enabled with `tower.accessToken` / `TOWER_ACCESS_TOKEN`.

## Known limitations

- **Signed URL expiration is not handled transparently**. Very long reads that outlive the URL's validity window surface as `IOException`; Nextflow task retry handles recovery.
- **No per-item last-modified exposed by the Platform browse API**. `SeqeraFileAttributes.lastModifiedTime()` returns `Instant.EPOCH` for data-link entries.
- **Read-only in this iteration**. Write operations raise `UnsupportedOperationException`. The Platform's `/data-links/{id}/upload` endpoints are a natural future extension point.
- **No data-link write, rename, delete, or management operations** (create/update/delete the data-link entity itself).
- **Single Platform endpoint per JVM** (unchanged from the dataset feature).

## Test plan

- [x] `./gradlew :plugins:nf-tower:test` — all 369 tests pass (verified locally)
- [x] `./gradlew :plugins:nf-tower:dependencies --configuration runtimeClasspath` shows no new cloud-SDK artifacts (no `aws-sdk`, `google-cloud-storage`, `azure-*`)
- [x] Manual: `nextflow fs ls seqera://<org>/<ws>/data-links/*` lists providers
- [x] Manual: `nextflow fs ls seqera://<org>/<ws>/data-links/<provider>/*` lists data-link names
- [x] Manual: `nextflow fs ls seqera://<org>/<ws>/data-links/<provider>/<name>/*` lists top-level bucket entries
- [x] Manual: `nextflow fs stat seqera://<org>/<ws>/data-links/<provider>/<name>/<file>` reports `is directory: false` and the correct `size`
- [x] Manual: `nextflow fs stat seqera://<org>/<ws>/data-links/<provider>/<name>/<dir>` reports `is directory: true`
- [x] Manual: `nextflow fs stat seqera://<org>/<ws>/data-links/<provider>/<name>/<missing>` raises `NoSuchFileException`
- [x] Integration test: pipeline reads a file inside a data-link via `file('seqera://…/data-links/<provider>/<name>/path/to/file')` using only `TOWER_ACCESS_TOKEN`
- [ ] Manual: verify that a Platform user with a **View** role (below Maintain) receives a clear `AccessDeniedException`


jorgee:
## Follow-up changes (pushed in `f260d00`)

Resolved the merge conflicts with `master` and some changes after internal review. Summary of what changed since initial commits:

### Correctness / robustness
- **Datasets listing no longer aborts on a bad entry.** `DatasetsResourceHandler.list()` previously let a `NoSuchFileException` (dataset with no versions, or only disabled versions) propagate out of the `collect`, killing the entire `datasets/` listing. It now skips-and-logs such datasets at debug, while real I/O errors still propagate. Added two tests.
- **Signed URLs are redacted in errors/logs.** `fetchSignedUrl` failures previously put the full pre-signed cloud URL (with `X-Amz-Signature` / SAS token / GCS `Signature` in the query) into the exception message. Added `redactUrl()` to strip the query before it's surfaced.
- **`PagedIterable` is now explicitly single-use.** `iterator()` throws `IllegalStateException` on a second call instead of silently skipping/duplicating pages (the page fetcher carries mutable cursor state). `getFirstPage()` remains safe. Added `PagedIterableTest` (10 cases).

### Code style / consistency
- Removed unused imports (`AccessDeniedException`, `AccessMode`) from both handlers.
- Replaced compound one-liners (`if (...) { found = it; break }`) and the empty inline `catch (...) {}` with the plugin's house style (multi-line blocks).
- Split inline annotations (`@Override`) and single-statement `if` bodies onto their own lines in the new files, to match the surrounding Nextflow style.
- Reverted purely-cosmetic reflows in `SeqeraPath`/`SeqeraFileSystemProvider` (one-line `@Override`s, brace style, variable renames) so the diff against `master` shows only substantive changes. Restored a few "why" comments and javadocs that were dropped from methods that still exist (`loadOrgWorkspaceCache`, `listOrgNames`, `listWorkspaceNames`, `resolveWorkspaceId`, `close`, and the `URI.create()`-pitfalls / URI-normalization notes).

All `:plugins:nf-tower:test` pass (400 tests, 0 failures).

jorgee:
## Provider-side filtering for data-links listing (pushed in `9b8ad21`)

Implemented the optional server-side filter noted in the previous comment.

- Added `SeqeraDataLinkClient.listDataLinksByProvider(workspaceId, provider)`, which queries the Platform with `search=provider:<provider>` (the same keyword `getDataLink` already relies on) so only that provider's data-links are paged.
- `DataLinksResourceHandler.list()` now uses it for the `data-links/<provider>/` listing instead of scanning the entire workspace list, keeping a client-side equality check as a guard against a non-exact keyword match.
- Tests: 2 new client tests assert the exact `search=provider%3A...` URL and the empty case; the handler tests now mock `listDataLinksByProvider` (the "returns names" test keeps a stray non-matching entry to prove the guard still drops it).

`:plugins:nf-tower:test` → 402 tests, 0 failures.

Note: `getDataLinkProviders` (the `data-links/` root listing) still does a full scan by design — it must enumerate all distinct providers — but it is `@Memoized`, so it's a one-time cost per run.

jorgee:
@pditommaso fixed conflicts with master and run a claude review. Also I included fixes for detected issues and improvements. It is ready for your review.

pditommaso (approved):
Thanks @jorgee for the thorough follow-up work! I reviewed the thread and confirmed all the self-documented fixes have landed:

- Datasets listing now skips-and-logs bad entries instead of aborting the whole listing
- Signed URLs are redacted in errors/logs via `redactUrl()`
- `PagedIterable` is explicitly single-use (fail-fast on second `iterator()`)
- Provider-side filtering for `data-links/<provider>/` via `listDataLinksByProvider`, with the client-side equality guard retained

Code and tests look good. One minor leftover: the "View role → clear `AccessDeniedException`" item in the test plan is still unchecked — the 403 mapping is in place, so just worth confirming manually or deferring explicitly. Approving. 🚀

PR #7071, opened 2026-04-24
Default NXF_FUSION_TRACE to false

pditommaso:
## Summary

- Change the default for `NXF_FUSION_TRACE` from `true` to `false` so that Fusion trace-based resource metrics (CPU, memory, I/O) are opt-in rather than on by default.
- Document the default in `docs/reference/env-vars.md`.

## Test plan

- [ ] Existing `TaskHandlerTest` / `BashWrapperBuilderTest` suites (which set the variable explicitly) still pass.
- [ ] With the variable unset, the bash command-trace wrapper is used as before.

### DIFF
--- a/docs/reference/env-vars.md
+++ b/docs/reference/env-vars.md
@@ -126,7 +126,7 @@ The following environment variables control the configuration of the Nextflow ru
 `NXF_FUSION_TRACE`
 : :::{versionadded} 26.04.0
   :::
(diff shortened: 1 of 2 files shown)


bentsherman (approved):
I don't have a strong a opinion on this, so I'm fine with either default

pditommaso:
Still not 100% sure with new metrics, safer to keep opt-in

PR #7075, opened 2026-04-24
Use toUriString for paths in work-dir and FilesEx error messages

pditommaso:
## Summary

Paths backed by non-default filesystems (S3, Azure Blob, etc.) render as opaque provider objects when interpolated directly into a GString, producing unhelpful messages like `Cannot get read permission on file: nextflow.file.http.XPath@...`.

Wrap the offending paths with `toUriString(...)` in:
- `Session.init` — the "Cannot create work-dir" abort message now shows the work-dir URI.
- `FilesEx` — the `mkdirs`/`canRead`/`canWrite`/`canExecute`/`lastModified` debug/trace logs now include a usable URI.

Also splits the long single-line `if(!workDir.mkdirs()) throw ...` in `Session.init` onto two lines for readability.

## Test plan

- [x] `./gradlew :nextflow:compileGroovy :nf-commons:compileGroovy` passes
- [ ] Trigger a non-writable work-dir and confirm the error message shows the URI

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/Session.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/Session.groovy
@@ -54,6 +54,7 @@ import nextflow.exception.MissingLibraryException
 import nextflow.exception.ScriptCompilationException
 import nextflow.executor.ExecutorFactory
 import nextflow.extension.CH
+import nextflow.extension.FilesEx
 import nextflow.file.FileHelper
 import nextflow.file.FilePorter
 import nextflow.plugin.Plugins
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/Session.groovy
```diff
@@ -445,7 +445,8 @@ class Session implements ISession {
      */
     Session init( ScriptFile scriptFile, List<String> args=null, Map<String,?> cliParams=null, Map<String,?> configParams=null ) {
 
-        if(!workDir.mkdirs()) throw new AbortOperationException("Cannot create work-dir: $workDir -- Make sure you have write permissions or specify a different directory by using the `-w` command line option")
+        if(!workDir.mkdirs())
+            throw new AbortOperationException("Cannot create work-dir '${FilesEx.toUriString(workDir)}' -- Make sure you have write permissions or specify a different directory by using the `-w` command line option")
```
bentsherman, line 449: Nit, you should be able to use `workDir.toUriString()` here and eliminate the extra import

Issue #5589, opened 2024-12-09
Checkpoints for 'resuming' a pipeline at a fixed point without cacheing


adamrtalbot:
## New feature

One of the earliest features of Nextflow is a resume feature, allowing users to relaunch pipelines with the same or similar settings and skipping completed work. This feature works very well at using cached results while also ensuring if a pipeline has changed or is deterministic it is re-ran properly.

However, there are many scenarios where resume does not work. Typically, when launching these pipelines it becomes apparent the user expects resume to work differently. A lot of the time, people expect `-resume` to allow them to skip chunks of their pipeline, saving them costs and time, but they are disappointed when it turns out their updated input files or non-deterministic channels have caused the entire pipeline to restart.

Rather than overloading the resume technique, we could imagine a system where a user could code in specific checkpoints into their pipeline for achieving this. The user could configure known workflow sections and re-start their pipeline from the checkpoint, skipping the previous step with few (if any) checks at all.

## Usage scenario 

I'm going to imagine a checkpoint operator, which serialises the state of a channel and can use it as input for subsequent runs:

```
workflow {
    sw1_out = subworkflow_1(inputs)
    sw1_out.checkpoint("checkpoint1")

    sw2_out = subworkflow_2(sw1_out)
    sw2_out.checkpoint("checkpoint2")

    sw3_out = subworkflow_3(sw2_out)
    // outputs
}
```

This is what the pipeline looks like:
```mermaid
flowchart TD
    A[input] --> C(subworkflow 1)
    subgraph workflow
    C --> D(subworkflow 2)
    D --> E(subworkflow 3)
    end
    checkpoint1 -.-> D
    D -.-> checkpoint2([checkpoint 2])
    C -.-> checkpoint1([checkpoint 1])
    checkpoint2 -.-> E
    E --> outputs[outputs]
```

The first time you run the pipeline would be normal:
```
nextflow run main.nf
```

The second time, you could re-start from checkpoint 1:
```
nextflow run main.nf -checkpoint checkpoint1
```

This would bypass the first subworkflow.

etc...
```
nextflow run main.nf -checkpoint checkpoint2
```

The main challenge would be how to save the global state of the pipeline, including all the auxiliary files etc. If the pipeline has dramatically diverged, it might be too difficult to resume. We would need something to catch that scenarioo

[...]

PR #7076, opened 2026-04-24
Add WorkflowInterceptor extension point for named-workflow plugins

huangzhibo:
## Context

Relates to #5589 (checkpoint / skip-forward discussion).

Adds a minimal `pf4j` `ExtensionPoint` — `nextflow.plugin.WorkflowInterceptor` — that lets a plugin wrap the execution of *named* workflows. The core change is tiny; the actual stage-archiving / reuse logic lives in a separate plugin, [`nf-stage`](https://github.com/huangzhibo/nf-stage), which is the first consumer.

## What changes

- New interface `WorkflowInterceptor` in `nf-commons` (32 lines)
- `WorkflowDef.run()` looks up an interceptor via `Plugins.getExtension(WorkflowInterceptor)`; if one is registered, delegates to it with a `proceed` closure that runs the original body (13 lines, extracted into `runDefault()`)
- Entry workflows (`name == null`) are never intercepted
- Behavior is unchanged when no plugin is installed

Total: +45 / -0 across 2 files.

## API

```groovy
interface WorkflowInterceptor extends ExtensionPoint {
    Object intercept(Object workflow, Object[] args, Closure proceed)
}
```

The interceptor owns the invocation: it may mutate `args` (e.g. clone channel inputs in place) before calling `proceed()`, and post-process the returned `ChannelOut`. Whether and when `proceed()` is invoked, and how the result channels are driven, is the plugin's responsibility.

## Why an extension point

Issue #5589 surfaces two positions: users want stage-level skip-forward semantics; core maintainers want `-resume` left alone and prefer pipeline authors prototype the pattern first. An extension point sidesteps the tension — it doesn't add any user-facing semantics to core, but unblocks plugin-side experimentation. If a pattern proves out in a plugin, promoting it to core remains an option later.

## Consumer: `nf-stage`

`nf-stage` uses this hook to archive each named workflow's `emit:` outputs on first run and, on reruns, replay them from a content-addressed archive when a SHA-256 digest over `(stage name, static args, channel input contents)` matches — no `work/` dependency, no per-task hash. The mechanics (clone channel args into `args[]`, let `proceed()` wire processes onto the clones, decide after digest compute whether to feed clones or STOP them) rely on the interceptor controlling args mutation and post-proceed channel flow, which is why the hook signature includes `args[]` and a `proceed` closure rather than simple before/after callbacks.

## Test plan

- [x] Existing `WorkflowDef` tests pass (no behavior change without a plugin)
- [x] Validated in `nf-stage` end-to-end: named sub-workflows intercepted, entry workflow not intercepted, `proceed()` round-trips `ChannelOut` correctly
- [ ] Happy to add a core-side test with a mock interceptor if reviewers prefer

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/script/WorkflowDef.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/script/WorkflowDef.groovy
@@ -24,6 +24,8 @@ import nextflow.exception.MissingProcessException
 import nextflow.exception.MissingValueException
 import nextflow.exception.ScriptRuntimeException
 import nextflow.extension.CH
+import nextflow.plugin.Plugins
+import nextflow.plugin.WorkflowInterceptor
 import nextflow.util.TestOnly
 /**
  * Models a script workflow component
@@ -181,6 +183,17 @@ class WorkflowDef extends BindableDef implements ChainableDef, IterableDef, Exec
     }
 
     Object run(Object[] args) {
+        final interceptor = name != null ? Plugins.getExtension(WorkflowInterceptor) : null
+        if( interceptor != null ) {
+            final result = interceptor.intercept(this, args, { runDefault(args) })
+            if( result instanceof ChannelOut )
+                this.output = result
+            return result
+        }
+        return runDefault(args)
+    }
+
+    private Object runDefault(Object[] args) {
         binding = new WorkflowBinding(owner)
         ExecutionStack.push(this)
         try {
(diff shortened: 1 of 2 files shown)


huangzhibo:
Closing — we're going to maintain this change in an internal Nextflow fork for now. May revisit upstreaming once the patterns prove out, possibly tied to a concrete proposal under #5589. Thanks.

PR #7077, opened 2026-04-26
Fix runtime type reflection in nf-lang

bentsherman:
Spun out from #7025 to make that PR more focused

This PR fixes the runtime type reflection in nf-lang to handle parameterized types correctly, using `Type` and `ParameterizedType` in addition to `Class`.

It improves both the rendering of type names (e.g. in error messages) and the extraction of types in plugin specs (e.g. config option types, function parameter/return types).

This fix is also needed by the language server to accurately type-check plugin definitions.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/plugin/spec/ConfigSpec.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/plugin/spec/ConfigSpec.groovy
@@ -15,10 +15,12 @@
  */
 package nextflow.plugin.spec
 
+import java.lang.reflect.ParameterizedType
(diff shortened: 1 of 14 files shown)


jorgee (approved):
Changes looks fine. Just a comment I noted here and I missed in the other PR. The plugin spec type can have a `String` or a  `Map` with `name` and `typeArguments`. It is only used to write the plugin spec. Not sure who is using this spec. Could it create incompatibilities with other tools using specs generated with this PR.

bentsherman:
It is currently only used by the language server, which already supports both string and map

PR #7078, opened 2026-04-26
Error messages: Fix ANSI colours

ewels:
Fix missing ANSI colours in v2 parser compilation errors


pditommaso:
Let's make sure ansi escape chars are not leaked with "agent" mode is enabled

ewels:
It shouldn't - when agent mode is running it's not a tty, so it won't have colours.

bentsherman (approved):
YOLO

PR #7079, opened 2026-04-27
feat: add two Indian scientists to NameGenerator

deepmind11:
Add Gopalasamudram Narayana Ramachandran (Ramachandran plot, structural biology)
and Jagadish Chandra Bose (pioneer of radio/microwave optics and plant biology)
to the list of scientists used for random mnemonic name generation.

[ci skip]

Signed-off-by: HGZ <harshitghosh@gmail.com>

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/util/NameGenerator.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/util/NameGenerator.groovy
@@ -430,6 +430,9 @@ class NameGenerator {
             // Hypatia - Greek Alexandrine Neoplatonist philosopher in Egypt who was one of the earliest mothers of mathematics. https://en.wikipedia.org/wiki/Hypatia
             "hypatia",
 
+            // Jagadish Chandra Bose - Indian physicist, biologist, and biophysicist who pioneered the investigation of radio and microwave optics and made seminal contributions to plant biology. https://en.wikipedia.org/wiki/Jagadish_Chandra_Bose
+            "jagadish",
+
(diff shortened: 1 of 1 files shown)


modules/nextflow/src/main/groovy/nextflow/util/NameGenerator.groovy
```diff
@@ -430,6 +430,9 @@ class NameGenerator {
             // Hypatia - Greek Alexandrine Neoplatonist philosopher in Egypt who was one of the earliest mothers of mathematics. https://en.wikipedia.org/wiki/Hypatia
             "hypatia",
 
+            // Jagadish Chandra Bose - Indian physicist, biologist, and biophysicist who pioneered the investigation of radio and microwave optics and made seminal contributions to plant biology. https://en.wikipedia.org/wiki/Jagadish_Chandra_Bose
+            "jagadish",
```
pditommaso: Thanks for contributing this, however we do usually use surnames. In this case I guess it should be "bose" that's already used right?

pditommaso:
Thanks for the contribution! I've kept only the **Ramachandran** entry and dropped the **Jagadish Chandra Bose** one via a follow-up commit.

The reason: name entries are matched by their short key, and `"jagadish"` collides conceptually with the existing crowded `j-` cluster while `"ramachandran"` is a clean, unambiguous addition. Keeping the single well-scoped entry keeps the change minimal and easy to merge.

The `ramachandran` addition is unchanged and good to go. 🙏

Issue #7032, opened 2026-04-14
Unused files in process inputs are of `nextflow.processor.TaskPath` type which makes invalid input paths for other processes


nvnieuwk:
### Expected behavior and actual behavior

Files in records that are not part of a defined record type in a process input are converted to `nextflow.processor.TaskPath`. This class is an implementation of the `Path` type but returns the relative path to the file in the work directory of the first process, which is invalid in the second process. This causes an error when trying to run another process on the output of the first process (see below). I fixed this by running `.toRealPath()` on all `TaskPath` files but this adds a lot of redundant code. It would help a lot if either all output files are properly converted back to a `Path` types using `.toRealPath()` or if `TaskPath` could be made valid for process inputs.

### Steps to reproduce the problem

1. Put the following code in a `main.nf` file:
```
nextflow.preview.types = true
workflow {
    def p1 = P1(channel.of(record(id: 'test', file: file('test.txt'))))
    P2(p1)
}

process P1 {
    input:
    input: Input

    output:
    input + record(file2: file('test2.txt'))

    script:
    """
    echo "test" > test2.txt
    """
}

process P2 {
    input:
    input: Input

    script:
    """
    cat ${input.file} ${input.file2}
    """
}

record Input {
    id: String
    file: Path
    file2: Path?
}
```

2. Create a `test.txt` file: `echo 'test' > test.txt`
3. Run Nextflow with the latest edge version: `NXF_VER=26.03.2-edge nextflow run main.nf`

### Program output 

```
$ NXF_VER=26.03.2-edge nextflow run main.nf 

 N E X T F L O W   ~  version 26.03.2-edge

Launching `main.nf` [hopeful_wilson] revision: 5b69a8f056

WARN: Static typing is a preview feature -- syntax and behavior may change in future releases
executor >  local (1)
[8d/8f5dd7] process > P1 (1) [100%] 1 of 1 ✔
[-        ] process > P2     -
WARN: File system is unable to get file attributes file: test.txt -- Cause: java.nio.file.ProviderMismatchException
ERROR ~ Error executing process > 'P2 (1)'

Caused by:
  Something went wrong while creating task hash for process 'P2' -- Offending keys: [
   - type=java.util.UUID value=2439a953-c814-43dd-b9c1-358bdb5d8b48, 
   - type=java.lang.String value=P2, 
   - type=java.lang.String value=    """
      cat ${input.file} ${input.file2}
      """
  , 
   - type=java.lang.String value=input, 
[...]

PR #7083, opened 2026-04-28
Fix staging and resume for typed record inputs with nested Path fields

stekaz:
Typed process inputs using record types with nested `Path` fields are (currently) not always handled like normal file inputs.

For example, given an input such as:

```nextflow
input:
sample: SampleRecord
```

where `SampleRecord.path` is a `Path`, the compiler should implicitly stage `sample.path` into the task work directory and rewrite references such as:

```nextflow
${resource.path}
```

to the process working directory, for example:

```
input.txt
```

But in some cases this does not happen. The compiler can miss the implicit file stager for nested Path fields. When that happens, task scripts either render the original source path instead of the staged task-local path, or the expected file could be missing from the working directory. And `-resume` could behave incorrectly.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskContext.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskContext.groovy
@@ -19,6 +19,7 @@ package nextflow.processor
 import nextflow.NF
 import nextflow.script.ScriptMeta
 
+import java.lang.reflect.Modifier
 import java.nio.file.Path
 import java.nio.file.Paths
 import java.nio.file.Files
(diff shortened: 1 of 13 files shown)


bentsherman:
Is this related to #7032 ?

stekaz:
Yes - I think they're both in the same problem area, but the failure modes are different. This PR fixes the input-side staging/localization (and resume) for typed records with nested Path fields. The other issue occurs when a record input is reused in an output record causing the downstream process to receive paths that are only valid in the upstream context.

bentsherman:
Note: even when a process input is a record type (e.g. `Sample`), the incoming value is always a `RecordMap`, not an actual instance of the record type. So you don't need the extra logic for named record types because they aren't used for runtime values

PR #7084, opened 2026-04-28
Allow collection params to be loaded from a file

### LINKED ISSUE #7082: List parameters improperly deserialized

The typed behavior is not being respected when deserializing csv files from a parameter. 

```
params {
	samplesheet: List<Sample>
}

record Sample {
	id: String
	fastq: Path
}
```

The expectation is that this would create a record object (of type Sample), however it generates a LinkedHashMap.

```
workflow {
	channel
		.fromList(params.samplesheet)
		.map{ row ->
			def new_record = record(id: row.id, fastq: file(row.fastq))
			println "Original class: ${row.getClass()}"
			println "New record class: ${new_record.getClass()}"
		}
}
```

That returns:

```
Original class: class java.util.LinkedHashMap
New record class: class nextflow.util.RecordMap
```

A  basic workflow is provided here.

```
nextflow.preview.types = true

params {
	samplesheet: List<Sample>
}

record Sample {
	id: String
	fastq: Path
}

workflow {

	params.samplesheet.each{ x -> println "Original row class: ${x.getClass()}" }

	channel
		.fromList(params.samplesheet)
		.map{ row ->
			def new_record = record(id: row.id, fastq: file(row.fastq))
			println "Original class: ${row.getClass()}"
			println "New record class: ${new_record.getClass()}"
		}
}
```

### Program output 

```
 N E X T F L O W   ~  version 26.03.4-edge

Launching `ex.nf` [desperate_poisson] revision: b9b7df3e81

WARN: Static typing is a preview feature -- syntax and behavior may change in future releases
Original class: class java.util.LinkedHashMap
Original class: class java.util.LinkedHashMap
New record class: class nextflow.util.RecordMap
```

### Environment 

* Nextflow version: 26.03.4-edge
* Java version: openjdk 17 2021-09-14
* Operating system: linux
* Bash version: zsh 5.8.1 (x86_64-ubuntu-linux-gnu)

### Additional context

Having to re-specify the record when reading the rows in seems to defeat the point of creating `record Sample`. It would be beneficial if the `List<Sample>` output rows in record format, as specified by `Sample`.

(Thanks for all of the hard work!)

bentsherman:
This PR is a redo of #6675 which was removed from the 26.04 release in #7025 

See the ADR changes for discussion and rationale

### DIFF
--- a/adr/20250825-workflow-params.md
+++ b/adr/20250825-workflow-params.md
@@ -39,6 +39,8 @@ This approach has several limitations:
 
 - Coerce CLI parameter values based on declared types, rather than relying on heuristics.
 
+- Support collection-type parameters that can be loaded from structured files (CSV, JSON, YAML).
+
 ## Non-goals
 
 - Removing the legacy `params.foo = bar` syntax -- legacy parameters must continue to work without modification.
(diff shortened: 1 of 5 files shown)


bentsherman:
1. #7185 fixed type casting to record collections, so samplesheets can now be loaded and cast to the appropriate type
2. #7208 implements automatic samplesheet loading when executing a named workflow directly, which is a better home for this behavior than the `params` block
3. follow-up to (2) should expose the file-loading behavior as a function (e.g. `fromData()`) and expose an extension point for plugins to support other file formats

PR #7088, opened 2026-04-29
Fix module binaries PATH not injected on `nextflow module run` (#7087)

### LINKED ISSUE #7087: Module binaries are not added to the PATH when running nextflow module run

Expected behavior:
Running the same process as a module with `nextflow module run` should resolve module binary scripts from `resources/usr/bin` when `nextflow.enable.moduleBinaries = true` is set (via `-c test.config`), just like `nextflow run failtest.nf`.

Actual behavior:
`nextflow module run cellgeni/failexample` fails with exit status `127` because `echo.sh` is not found (`command not found`), while `nextflow run failtest.nf` succeeds with the same config.

Self-contained repro files:
- `test.config`
- `failtest.nf`
- `modules/cellgeni/failexample/main.nf`

`test.config`:
```groovy
```

`modules/cellgeni/failexample/main.nf`:
```nextflow
/*
 * Module: cellgeni/failexample
 */

process FAILEXAMPLE {
    input:
    val greeting

    output:
    stdout

    script:
    """
    echo '${greeting}'
    echo.sh
    """
}
```

`failtest.nf`:
```nextflow
include { FAILEXAMPLE } from './modules/cellgeni/failexample'

workflow  {
    FAILEXAMPLE("Hello, World!")
}
```

Run 1 (fails):
```bash
nextflow -c test.config module run cellgeni/failexample --greeting 'Hello world!'
```

Run 2 (works):
```bash
nextflow -c test.config run failtest.nf
```

### Program output

Failing execution (`nextflow -c test.config module run cellgeni/failexample --greeting 'Hello world!'`):

```text
 N E X T F L O W   ~  version 26.04.0

Launching `/lustre/scratch124/cellgen/cellgeni/aljes/nf-metacells/modules/cellgeni/failexample/main.nf` [compassionate_cori] revision: bd30b4f67c

executor >  lsf (2)
[ae/d58586] process > FAILEXAMPLE [  0%] 0 of 1, retries: 1
[4c/22316b] NOTE: Process `FAILEXAMPLE` terminated with an error exit status (127) -- Execution is retried (1)
WARN: Killing running tasks (1)

WARN: Failed to render execution report -- see the log file for details
WARN: Failed to render execution timeline -- see the log file for details
```

Relevant task error (`nf-work/4c/22316b6fec6bcedbca7a33ee0a1a08/.command.log`):

```text
Hello world!
/lustre/scratch124/cellgen/cellgeni/aljes/nf-metacells/nf-work/4c/22316b6fec6bcedbca7a33ee0a1a08/.command.sh: line 3: echo.sh: command not found
...
Exited with exit code 127.
```

Successful comparison run (`nextflow -c test.config run failtest.nf`):

```text
 N E X T F L O W   ~  version 26.04.0

Launching `failtest.nf` [mighty_kilby] revision: 3877f72024

executor >  lsf (1)
[89/d44147] process > FAILEXAMPLE [100%] 1 of 1 ✔
```

Attached logs in archive include:
- `.nextflow.log` files from both runs
- full task logs and wrappers for failed and successful tasks

### Environment

* Nextflow version: 26.04.0
* Java version: `openjdk version "17.0.8.1" 2023-08-24` (Temurin 17.0.8.1+1)
* Operating system: Linux (`Linux 5.15.0-174-generic x86_64 GNU/Linux`)
* Bash version: `GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)`

### Additional context

- The failing module-run task wrapper (`nf-work/4c/.../.command.run`) does not include `nxf_container_env`/PATH export for module binaries.
- The successful workflow-run task wrapper (`nf-work/89/.../.command.run`) does include module-binary PATH injection.
- This suggests `nextflow module run` is not applying module binary path injection the same way as regular `nextflow run` for this setup, despite `nextflow.enable.moduleBinaries = true` in `test.config`.

[issue_repro_bundle_v2.tar.gz](https://github.com/user-attachments/files/27213587/issue_repro_bundle_v2.tar.gz)

pditommaso:
## Summary

Fixes #7087.

When a module is launched directly via `nextflow module run <scope>/<name>`, the module's `main.nf` is loaded as the **entry** script, not as an *included* module. As a result `ScriptMeta.isModule()` returns `false` for the script that owns the process, and `TaskProcessor.getModuleBundle()` returned `null`, so the `resources/usr/bin` (and `resources/bin`, `resources/usr/local/bin`) directory was never injected into the task `PATH` -- even with `nextflow.enable.moduleBinaries = true`.

This was inconsistent with `nextflow run` + `include { … } from './modules/...'`, which marks the included script as a module and gets the PATH injection correctly.

### Change

Single-line change in `TaskProcessor.getModuleBundle()` -- switch the gate from `isModule()` to `getScriptPath()`:

```groovy
return meta?.getScriptPath() ? meta.getModuleBundle() : null
```

The bundle is now resolved whenever the owner script has a known path, regardless of whether the script was loaded as the main entry or as an included module. The feature remains opt-in via `nextflow.enable.moduleBinaries`, and `ResourcesBundle.scan()` still returns an empty bundle when no `resources/` directory exists, so this is a no-op for scripts without one.

## Why this approach

I considered three options (described in the linked issue investigation):

1. **Drop the `isModule()` gate** -- this PR. Smallest, most direct fix.
2. **Mark the entry script as a module via `setModule(true)`** -- not viable: in `ScriptLoaderV2` `setModule(true)` also flips `skipEntryFlow=true`, which would prevent the implicit-workflow execution that `nextflow module run` relies on. Would require a new flag (e.g. `isEntryModule`) -- more invasive.
3. **Have `CmdModuleRun` synthesize a wrapper script that `include`s the module** -- larger refactor.

Option 1 was chosen as the minimal and semantically correct fix.

## Possible side-effects and implications

The change widens the set of scripts that are eligible to contribute a `resources/` bundle to the task environment. Concretely:

- **Behavior change (intended).** A standalone pipeline script `pipeline.nf` that has a sibling `resources/usr/bin/` (or `resources/bin/`, `resources/usr/local/bin/`) directory will now also get those directories prepended to the task `PATH`, *only* when the user explicitly opts in with `nextflow.enable.moduleBinaries = true`. Previously this was silently ignored unless the script was loaded via `include`. This is arguably the correct semantics of the feature flag and matches how an end user would expect it to work.

- **No effect when the feature flag is off.** `TaskProcessor.getBinDirs()` still gates the call on `session.enableModuleBinaries()`, so behaviour is unchanged for the default configuration.

- **No effect when there is no `resources/` directory.** `ResourcesBundle.scan()` returns an empty bundle when the directory does not exist (`bundle/ResourcesBundle.groovy:147-148`), so `getBinDirs()` produces an empty list and `PATH` is not modified.

- **No effect on the `include`-based path.** Scripts loaded as included modules continue to have `getScriptPath()` set, so they keep working exactly as before.

- **Bundle staging / fingerprinting.** When the bundle is non-empty, `ResourcesBundle.fingerprint()` is part of cache-key inputs in some code paths (e.g. wave/container-bundle scenarios). For a previously "non-module" script that *now* has a sibling `resources/`, the fingerprint of cached tasks could change *only* the first time the user opts into `moduleBinaries`. Since opting in is an explicit user action this is acceptable; existing module-run / include-based pipelines are unaffected.

- **Defensive guard kept.** The check now uses `getScriptPath()` rather than dropping the guard entirely, because `ScriptMeta.getModuleBundle()` throws `IllegalStateException` when `scriptPath` is null. This preserves robustness for any code path that happens to register a script without setting its path.

## Test plan

- [x] `./gradlew :nextflow:test --tests "nextflow.processor.TaskProcessorTest"` passes (includes a new test case `should resolve module bundle when script path is set regardless of isModule` that exercises the fix).
- [x] `./gradlew :nextflow:test --tests "nextflow.script.ScriptMetaTest" --tests "nextflow.SessionTest" --tests "nextflow.cli.CmdRunTest"` -- existing tests for related areas still pass.
- [ ] End-to-end repro from the issue (`nextflow -c test.config module run cellgeni/failexample --greeting 'Hello world!'`) succeeds with this change applied -- recommended manual verification before merge.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -1569,7 +1569,11 @@ class TaskProcessor {
     ResourcesBundle getModuleBundle() {
         final script = this.getOwnerScript()
         final meta = ScriptMeta.get(script)
-        return meta?.isModule() ? meta.getModuleBundle() : null
+        // Resolve the resources bundle whenever the owner script has a known path,
+        // not only when it was loaded as an included module. This allows module
+        // binaries to be picked up also when a module is launched directly as the
+        // entry script via `nextflow module run` -- see #7087
+        return meta?.getScriptPath() ? meta.getModuleBundle() : null
     }
 
     @Memoized
--- a/modules/nextflow/src/test/groovy/nextflow/processor/TaskProcessorTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/processor/TaskProcessorTest.groovy
@@ -35,6 +35,7 @@ import nextflow.script.BaseScript
 import nextflow.script.BodyDef
 import nextflow.script.ProcessConfig
 import nextflow.script.ProcessConfigV1
+import nextflow.script.ScriptMeta
 import nextflow.script.ScriptType
 import nextflow.script.bundle.ResourcesBundle
 import nextflow.script.params.FileInParam
@@ -98,6 +99,41 @@ class TaskProcessorTest extends Specification {
         cleanup:
         home.deleteDir()
 
+    }
+
+    def 'should resolve module bundle when script path is set regardless of isModule' () {
+        given:
+        def folder = Files.createTempDirectory('test')
+        def mod = folder.resolve('mod1'); mod.mkdir()
+        def bin = mod.resolve('resources/usr/bin'); bin.mkdirs()
+        def scriptPath = mod.resolve('main.nf'); Files.createFile(scriptPath)
+        Files.createFile(bin.resolve('echo.sh'))
+        and:
+        def script = Mock(BaseScript)
+        def meta = Mock(ScriptMeta) {
+            getScriptPath() >> scriptPath
+            // Simulate the failing case: script is loaded as the entry, not as an included module
+            isModule() >> false
+            getModuleBundle() >> ResourcesBundle.scan(mod.resolve('resources'))
+        }
+        and:
+        def session = Mock(Session) { getConfig() >> [:] }
+        def executor = Mock(Executor) {}
+        def processor = Spy(TaskProcessor, constructorArgs: [[session:session, executor:executor]])
+        processor.getOwnerScript() >> script
+
+        when:
+        ResourcesBundle bundle
+        GroovyMock(ScriptMeta, global: true)
+        ScriptMeta.get(script) >> meta
+        bundle = processor.getModuleBundle()
+
+        then:
+        bundle != null
+        bundle.getBinDirs() == [bin]
+
+        cleanup:
+        folder?.deleteDir()
     }
 
     @Unroll


pditommaso:
@bentsherman what do you think about this? it may be worth adding a specific flag e.g. moduleEntry ?

PR #7093, opened 2026-04-30
Fix K8s 401 after token expiry (#6918)

### LINKED ISSUE #6918: K8s executor caches client token at wrong layer — PR #6742 token refresh never triggered

pditommaso:
## Summary

Closes #6918.

Fixes the remaining 401 Unauthorized failure modes on the K8s executor after token expiry, left open after #6742 and #6925.

Three caching layers contributed to the bug. After this PR, all three are self-healing:

| Layer | Before | After |
|---|---|---|
| `K8sConfig.getClient()` | Guava 50-min cache (#6742) | unchanged |
| `K8sExecutor.getClient()` | Guava cache wrapping the above (#6925) | unchanged |
| `K8sTaskHandler.client` | captured at construction, never refreshed | replaced by `getClient()` delegating to executor |
| `K8sClient` request loop | 401 propagated immediately | retries with token re-read from disk when `tokenPath` is set |

### Why a retry alone wasn't enough

K8s service-account tokens aren't refreshed via an API call — the kubelet rotates them by overwriting the file at the mount path (`/var/run/secrets/kubernetes.io/serviceaccount/token`). The client just needs to re-read the file from disk on 401. This PR records the token file path on `ClientConfig` (`tokenPath`) for the three discovery paths (in-cluster, kubeconfig `tokenFile`, Nextflow-config `tokenFile`) and uses it in the retry handler.

For inline `token:` strings in kubeconfig there's no file to re-read; those still fail fast on 401, unchanged.

### Files

- `K8sTaskHandler.groovy` — drop cached `client` field, add `getClient()` delegating to executor
- `ClientConfig.groovy` — add `Path tokenPath`; populate it from `fromNextflowConfig`/`fromUserAndCluster` when a `tokenFile` is given
- `ConfigDiscovery.groovy` — populate `tokenPath` in `fromCluster`
- `K8sClient.groovy` — `apply()` retries on 401 when `tokenPath` is set; `onRetry` listener calls new `refreshToken()` to re-read the file and update `config.token`

## Test plan

- [x] New `ClientConfigTest` cases assert `tokenPath` is preserved for `tokenFile` (Nextflow + kubeconfig) and not set for inline tokens
- [x] `ConfigDiscoveryTest.should load from cluster env` extended to assert `config.tokenPath == TOKEN_FILE`
- [x] New `K8sClientTest.should re-read token from disk and retry on 401 when tokenPath is set` — verifies file is re-read and second request uses the fresh token
- [x] New `K8sClientTest.should not retry 401 when tokenPath is not set` — preserves existing fail-fast behavior for inline tokens
- [x] All 247 nf-k8s tests pass (`./gradlew :plugins:nf-k8s:test`)

### DIFF
--- a/plugins/nf-k8s/src/main/nextflow/k8s/K8sTaskHandler.groovy
+++ b/plugins/nf-k8s/src/main/nextflow/k8s/K8sTaskHandler.groovy
@@ -70,8 +70,6 @@ class K8sTaskHandler extends TaskHandler implements FusionAwareTask {
 
     private ResourceType resourceType = ResourceType.Pod
 
-    private K8sClient client
-
     private String podName
 
     private BashWrapperBuilder builder
@@ -93,7 +91,6 @@ class K8sTaskHandler extends TaskHandler implements FusionAwareTask {
     K8sTaskHandler( TaskRun task, K8sExecutor executor ) {
         super(task)
         this.executor = executor
-        this.client = executor.getClient()
         this.outputFile = task.workDir.resolve(TaskRun.CMD_OUTFILE)
         this.errorFile = task.workDir.resolve(TaskRun.CMD_ERRFILE)
         this.exitFile = task.workDir.resolve(TaskRun.CMD_EXIT)
@@ -115,6 +112,8 @@ class K8sTaskHandler extends TaskHandler implements FusionAwareTask {
     }
 
     protected K8sConfig getK8sConfig() { executor.getK8sConfig() }
+
+    protected K8sClient getClient() { executor.getClient() }
 
     protected boolean useJobResource() { resourceType==ResourceType.Job }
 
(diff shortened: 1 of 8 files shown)


adamrtalbot (approved):
it looks good to me: 
- if 401,
- refresh token
- to maxAttempts

Will need some real world verification.

pditommaso:
Have you a way to replicate and test a dev build?

PR #7094, opened 2026-04-30
Validate module spec against JSON schema

pditommaso:
## Summary

- `nextflow module validate` now formally validates `meta.yml` against the upstream module JSON schema (`nextflow-io/schemas/main/module/v1/schema.json`) using `com.networknt:json-schema-validator`.
- Schema validation runs between the structural check and the Nextflow-specific spec check; if it fails, validation stops there.
- Hand-coded spec checks that the schema already covers (`name`/`description` required, per-param `type`/`description` required, `TODO_TYPE` placeholder) are removed — the schema is the single source of truth for those rules. Checks the current schema does not cover stay in `ModuleSpec.validate()`: `version`/`license` required, semver pattern, `namespace/name` regex, and the `TODO_DESCRIPTION` placeholder.
- New `--schema <url-or-path>` flag on `nextflow module validate` to override the default schema location. Accepts an `http(s)://` URL, a `file:` URI, or a local file path.
- Schema load failures abort with a clear message pointing the user at `--schema` as the escape hatch.

A follow-up PR against `nextflow-io/schemas` will tighten the schema (require `version`/`license`, add the semver and namespace/name patterns, reject `TODO_DESCRIPTION`) so the remaining hand-coded checks can also be removed.

## Test plan

- [x] `./gradlew :nextflow:test --tests 'nextflow.module.ModuleSpecTest' --tests 'nextflow.module.ModuleSpecFactoryTest' --tests 'nextflow.module.ModuleSchemaValidatorTest' --tests 'nextflow.cli.module.CmdModuleValidateTest'` — all green
- [x] New `ModuleSchemaValidatorTest`: pass, missing required field, invalid enum value, `file:` URI schema, hard-fail on missing schema
- [x] `CmdModuleValidateTest` rewritten to use a hermetic local schema fixture (no network in CI) and to split the missing-fields case into schema-level vs. nextflow-only scenarios
- [ ] Manual: `nextflow module validate <module>` against a real module
- [ ] Manual: `nextflow module validate <module> --schema ./local-schema.json`


modules/nextflow/src/main/groovy/nextflow/module/ModuleSpec.groovy
jorgee: default schema is also hardcoded here
```suggestion
        result['$schema'] = ModuleSchemaValidator.DEFAULT_SCHEMA_URL
```

modules/nextflow/src/main/groovy/nextflow/module/ModuleSchemaValidator.groovy
```diff
@@ -0,0 +1,110 @@
...
+     * Validate a meta.yml file against the JSON schema located at the given
+     * URL or local file path.
+     *
+     * @param metaYaml       Path to the meta.yml file to validate
+     * @param schemaLocation URL (http/https), file: URI, or local file path of the schema
+     * @return List of validation error messages, empty if the spec is valid
+     */
+    static List<String> validate(Path metaYaml, String schemaLocation) {
+        final schemaText = loadSchema(schemaLocation)
+        final factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V202012)
```
jorgee: Version is hardcoded by `SpecVersion.VersionFlag.V202012`. Shouldn't be getting from the loaded schema?

jorgee (commented):
In #7056, we added the `$schema` in `meta.yml`, but it is not taken into account in this PR as it is just considering using the `DEFAULT_SCHEMA_URL` or the URL provided by the user. Not a big deal now because both will be the same value, but thinking in the future, we could have `meta.yml` files pointing to a version which is not the default.

I was initially thinking the implementation on the PR is wrong, and we should use the schema in `$schema`. However, this could also point to a wrong schema and it will break the validation. @pditommaso have you considered it?  What should we do in the case $schema is not the same as the `DEFAULT_SCHEMA_URL`. Should the validation fail, or should we ignore/warn and always validate through the  `DEFAULT_SCHEMA_URL` (or the user specified one)?

modules/nextflow/src/main/groovy/nextflow/module/ModuleSpec.groovy
pditommaso: Make sense

modules/nextflow/src/main/groovy/nextflow/module/ModuleSchemaValidator.groovy
pditommaso: Agree

bentsherman (approved):

pditommaso:
Allow me to review again this, not fully sure on some details

bentsherman:
Marking as draft until you review

pditommaso:
Ok, made a few minor changes

PR #7097, opened 2026-05-04
fix(nf-tower): ensure final log checkpoint runs after thread interrupt

### LINKED ISSUE #7096: LogsCheckpoint thread re-sets interrupt flag in await(), causing empty console output files on GCS

rnaidu-seqera:
## Summary                                                                                                                                                       
                                                            
  - `LogsCheckpoint.await()` was re-setting the thread interrupt flag after catching `InterruptedException`, causing subsequent NIO writes in `saveFiles()` to     
  throw `ClosedByInterruptException` and leave empty `nf-*.txt`, `nf-*.log`, and `timeline-*.html` files in the GCS work directory
  - Fixed by changing `await()` to return a boolean instead of re-setting the flag, and calling `Thread.interrupted()` in `run()` to clear the flag before         
  `saveFiles()` executes — ensuring the final upload always succeeds regardless of interrupt source                                                                
  - Added `synchronized(lock)` guard so `onFlowComplete()` cannot interrupt the thread mid-upload
                                                                                                                                                                   
## Test plan                                                                                                                                                     
   
  - [x] Existing `LogsCheckpointTest` tests pass (interval configuration via default, env var, and config file)                                                    
  - [x] New unit test `'should perform final saveFiles when interrupted mid-sleep'` passes — verifies `handler.saveFiles()` is called exactly once when`onFlowComplete()` interrupts the thread mid-sleep, and that the thread interrupt flag is clear at the time of the call                                          
  - [x] Manually verified fix on Google Batch CE with GCS work directory and `cloudcache.enabled = true`: `nf-*.txt` file uploaded to GCS with full content after workflow completion (previously 0 bytes)

### DIFF
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsCheckpoint.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsCheckpoint.groovy
@@ -74,14 +74,13 @@ class LogsCheckpoint implements TraceObserverV2 {
         log.debug "Starting logs checkpoint thread - interval: ${interval}"
         try {
             while( true ) {
-                await(interval)
-                if( Thread.currentThread().isInterrupted() )
-                    break
+                final interrupted = await(interval)
+                Thread.interrupted()  // clear flag so NIO writes in saveFiles() succeed
                 synchronized(lock) {
-                    if( Thread.currentThread().isInterrupted() )
-                        break
                     handler.saveFiles()
                 }
+                if( interrupted )
+                    break
             }
         }
         finally {
@@ -89,13 +88,14 @@ class LogsCheckpoint implements TraceObserverV2 {
         }
     }
 
-    protected void await(Duration interval) {
+    protected boolean await(Duration interval) {
         try {
             Thread.sleep(interval.toMillis())
+            return Thread.currentThread().isInterrupted()
         }
         catch (InterruptedException e) {
             log.debug "Interrupted logs checkpoint thread"
-            Thread.currentThread().interrupt()
+            return true
         }
     }
 }
--- a/plugins/nf-tower/src/test/io/seqera/tower/plugin/LogsCheckpointTest.groovy
+++ b/plugins/nf-tower/src/test/io/seqera/tower/plugin/LogsCheckpointTest.groovy
@@ -80,4 +80,26 @@ class LogsCheckpointTest extends Specification {
         cleanup:
         SysEnv.pop()
     }
+
+    def 'should perform final saveFiles when interrupted mid-sleep' () {
+        given:
+        SysEnv.push(TOWER_LOGS_CHECKPOINT_INTERVAL: '60s')
+        def session = Mock(Session) {
+            getWorkDir() >> TestHelper.createInMemTempDir()
+            getConfig() >> [:]
+        }
+        def handler = Mock(LogsHandler)
+        def checkpoint = new LogsCheckpoint()
+
+        when:
+        checkpoint.onFlowCreate(session)
+        checkpoint.@handler = handler   // inject mock before thread wakes up
+        checkpoint.onFlowComplete()
+
+        then:
+        1 * handler.saveFiles()
+
+        cleanup:
+        SysEnv.pop()
+    }
 }
--- a/plugins/nf-tower/src/test/io/seqera/tower/plugin/LogsCheckpointTest.groovy
+++ b/plugins/nf-tower/src/test/io/seqera/tower/plugin/LogsCheckpointTest.groovy
@@ -97,7 +97,7 @@ class LogsCheckpointTest extends Specification {
         checkpoint.onFlowComplete()
 
         then:
-        1 * handler.saveFiles()
+        1 * handler.saveFiles() >> { assert !Thread.currentThread().isInterrupted() }
 
         cleanup:
         SysEnv.pop()


plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsCheckpoint.groovy
```diff
@@ -74,28 +74,28 @@ class LogsCheckpoint implements TraceObserverV2 {
         log.debug "Starting logs checkpoint thread - interval: ${interval}"
         try {
             while( true ) {
-                await(interval)
-                if( Thread.currentThread().isInterrupted() )
-                    break
+                final interrupted = await(interval)
+                Thread.interrupted()  // clear flag so NIO writes in saveFiles() succeed
```
jorgee, line 78: Reading the description of the PR I have not clear what was causing the issue and why you set `Thread.interrupted()` in all the cases. Is it not causing issues in the normal execution?

pditommaso: I think the main mistake is to use interrupt to signal the termination that was introduced in previous iteration. 

https://github.com/nextflow-io/nextflow/blob/60273273b716eaeb3a0fc0395c02320f7d6a4511/plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsCheckpoint.groovy#L59-L61

`interrupt` should be use to force the thread interruption (that's not the case). 

I'd suggest going in the change history, revert that implementation and proper exiting the thread

jorgee: I think it was added in #6787. I see it is in stable 25.10.x. But this branch doesn't include #6939. So, maybe the reported problem is caused by the race condition and interrupt is happening during upload. Because, if the thread is interrupted, it should break without calling saveFiles.

@rnaidu-seqera, Is the issue happening in 25.10.4?

bentsherman:
Closing in favor of #6939 , see linked issue

Issue #6974, opened 2026-03-25
`ConfigScope`of plugin not found


JosuaCarl:
### Expected behavior and actual behavior

We are currently trying to address the warnings of the v2 syntax parser:
```
...
WARN: Unrecognized config option 'co2footprint.dataFile.enabled'
WARN: Unrecognized config option 'co2footprint.location'
...
```

But a proposed PR which should align all `ConfigScope` classes with the [template in the documentation](https://nextflow.io/docs/latest/plugins/developing-plugins.html) (see https://github.com/nextflow-io/nf-co2footprint/pull/362) does not fix the issue. During a deeper dive I found that the `ConfigValidator` validates against all `ConfigScope` classes, but our classes are never recognized.

### Steps to reproduce the problem

Use v2 Parser with https://github.com/nextflow-io/nf-co2footprint/pull/362 of the nf-co2footprint plugin.

### Program output 

I collected the following info during runtime:
- `Plugins.isStarted('nf-co2footprint')` -> `true` indicates that the plugin is correctly started
- `Plugins.getExtensions(ConfigScope)` lists a bunch of `ConfigScope` classes, but none from the plugin are among them
- `Plugins.manager.getExtensionClasses('nf-co2footprint')` -> `[CO2FootprintFactory, CO2FooprintExtension]` this lists the `PluginExtensionPoint` and `TraceObserverFactory`, which is to be expected, but not the `ConfigScope` 
- Consequently `Plugins.getExtensionsInPluginId(ConfigScope, 'nf-co2footprint')` is empty

### Environment 

* Nextflow version: 25.10.4

### Sneak Peak into the main config class

```Groovy
@ScopeName('co2footprint')
@Description('The `co2footprint` scope allows you to configure the `nf-co2footprint` plugin.')
@Slf4j
class CO2FootprintConfig implements ConfigScope {

    ...

    @ConfigOption(types=[GString])
    @Description('Location GeoCode from Electricity maps.')
    final String location
    
    ...

    /**
     * No-arg constructor required by Nextflow's v2 config parser to
     * discover and validate config options at parse time.
     */
    CO2FootprintConfig() {}

    /**
     * Loads configuration from a map and sets up defaults and fallbacks.
     * Also sets up CPU and CI data sources and assigns machine type and PUE.
     *
     * @param configMap   Map of configuration values (from Nextflow config)
     * @param cpuData     TDPDataMatrix with CPU power draw data
[...]

PR #7098, opened 2026-05-05
docs: document ConfigScope discovery rules

pinin4fjords:
## Summary

The plugin developer guide currently shows a single-option `ConfigScope` example with `@ScopeName` and a no-arg constructor, but does not mention registering the scope in `build.gradle` `extensionPoints`. Following the example as written does not eliminate `Unrecognized config option` warnings under the v2 syntax parser, because the scope is never discovered.

This PR fills four related gaps:

- Documents the `extensionPoints` registration step required for the top-level scope to be discovered
- Adds a nested-scope example, showing that fields whose declared type implements `ConfigScope` are auto-discovered and do not need a `@ConfigOption` annotation
- Notes that `@ConfigOption` fields are not inherited (the validator uses `Class.getDeclaredFields()`), so each `ConfigScope` class must declare its options directly
- Notes that the `types` attribute on `@ConfigOption` should only list [standard Nextflow types](https://docs.seqera.io/nextflow/reference/stdlib-types), not runtime classes like `GString`, `Number` or `Map`

These rules are derivable from `SpecNode.java` and `ConfigValidator.groovy`, but were not surfaced for plugin authors. Context for the gap: nextflow-io/nf-co2footprint#362 and the related upstream issue #6974.

## Style

The new content is additive; the existing single-option example is left intact. Triple-quoted `@Description` strings, `:::{note}` admonitions and prose voice match the surrounding section.

### DIFF
--- a/docs/developer/config-scopes.mdx
+++ b/docs/developer/config-scopes.mdx
@@ -43,16 +43,16 @@ The fully-qualified class name (in this case, `nextflow.hello.HelloConfig`) must
 
 A *config option* is defined as a field with the `@ConfigOption` annotation. The field name determines the name of the config option.
 
-:::note
-Config options are collected with `Class.getDeclaredFields()`, which does not include inherited fields. Each scope class must declare its options directly, even if a shared parent class provides common logic.
-:::
-
 For example:
 
 ```groovy
     @ConfigOption
     String createMessage
 ```
+
+:::note
+Inherited fields are not included. Each scope class must declare its options directly.
+:::
 
 The `@ConfigOption` annotation can specify an optional set of types that are valid in addition to the field type. For example, the `fusion.tags` option, which accepts either a String or Boolean, is declared as follows:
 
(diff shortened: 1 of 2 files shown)


docs/plugins/developing-plugins.md
```diff
@@ -237,6 +237,69 @@ class MyPluginConfig implements ConfigScope {
...
+nextflowPlugin {
+    // ...
+    extensionPoints = [
+        'nextflow.myplugin.MyPluginConfig',
+        // other extension points...
+    ]
+}
+```
+
+Configuration options can be grouped by adding a nested `ConfigScope` as a field on the parent scope. Nextflow discovers nested scopes by walking fields whose declared type implements `ConfigScope`, so the field itself does not need a `@ConfigOption` annotation. Only the top-level scope needs `@ScopeName`, a no-arg constructor, and an entry in `extensionPoints`.
```
christopher-hakkaart:
```suggestion
Configuration options can be grouped by adding a nested `ConfigScope` as a field on the parent scope. Nextflow discovers nested scopes by walking fields whose declared type implements `ConfigScope`. The field itself does not need a `@ConfigOption` annotation. Only the top-level scope needs `@ScopeName`, a no-arg constructor, and an entry in `extensionPoints`.
```

christopher-hakkaart: Very minor suggestion. I'll let one of the developers confirm the content.

bentsherman (approved):

PR #7103, opened 2026-05-06
Pick up force-pushed branches in multi-revision repo strategy

### LINKED ISSUE #7102: Force-pushed branch not picked up by multi-revision repo (returns "Already-up-to-date")

When the upstream branch is force-pushed (history rewritten), Nextflow's multi-revision repository strategy does **not** detect the change. The bare repo silently keeps the old SHA and `download()` short-circuits with `Already-up-to-date`, so the pipeline continues to run the stale commit.

### Steps to reproduce

1. Run a pipeline pointing at a remote branch (e.g. `main`) so Nextflow caches it under `~/.nextflow/assets/.repos/<project>/bare/`.
2. On the remote, rewrite history of that branch and push with `--force`.
3. Run Nextflow again with the same revision.

Observed: local `refs/heads/<branch>` in the bare repo still points at the old commit; the corresponding `clones/<oldSha>/` still exists; `download()` returns `Already-up-to-date`.

### Root cause

`MultiRevisionRepositoryStrategy.refSpecForName(...)` constructs `RefSpec`s without the leading `+` that allows non-fast-forward updates. After a force-push, JGit's `FetchCommand` records `REJECTED` for the local ref update (visible via `FetchResult.getTrackingRefUpdates()`) but does **not** throw, so the bare repo silently keeps the old SHA. Then `revisionToCommitWithBareRepo()` resolves to the stale commit, whose clone directory still exists, and `download()` short-circuits.

### Verified

Confirmed on a real reproduction with diagnostic logging:

```
DEBUG n.s.MultiRevisionRepositoryStrategy - Bare repo fetch update for jorgee/nf-sleep [revision: main]: refs/heads/main <- refs/heads/main => REJECTED (old=3851da8d..., new=d17b807d...)
```

JGit fetched the new object but refused to move the local branch ref because the update was non-fast-forward.

### Affected file

`modules/nextflow/src/main/groovy/nextflow/scm/MultiRevisionRepositoryStrategy.groovy`

jorgee:
## Summary
- Force-update local bare-repo branches by prepending `+` to the branch `RefSpec`s in `refSpecForName(...)`. After a remote force-push the local ref now moves to the new commit instead of being silently rejected as non-fast-forward. Tag and commit refspecs are unchanged.
- Inspect `FetchResult.getTrackingRefUpdates()` after every fetch via a new `verifyFetchResult` helper, used by both `checkBareRepo()` and `createSharedClone()`:
  - `NEW` / `FAST_FORWARD` / `NO_CHANGE` → debug log
  - `FORCED` → **WARN** with old → new SHAs (upstream history was rewritten)
  - any other result (`REJECTED`, `LOCK_FAILURE`, `IO_FAILURE`, `REJECTED_OTHER_REASON`, `REJECTED_MISSING_OBJECT`) → `AbortOperationException`, so the failure is no longer silent.

Fixes #7102

## Test plan
- [x] Unit tests for `verifyFetchResult` policy:
  - happy results (NEW / FAST_FORWARD / NO_CHANGE) pass silently
  - `FORCED` does not throw
  - rejected/failure results throw `AbortOperationException` with the revision and repo label in the message
- [x] End-to-end test using a local bare upstream: initial fetch lands on commit A, the upstream is force-pushed to a divergent commit B, second fetch advances the local bare repo's `refs/heads/main` to B. Exercises the `+` refspec and the WARN log path together — no GitHub credentials required.
- [x] Existing local tests in `MultiRevisionRepositoryStrategyTest` still pass (`should list commits`, `should create correct RefSpec for branches tags and commits`).
- [x] Manual reproduction against a real GitHub remote: cached pipeline against `main`, force-pushed `main`, re-ran Nextflow — bare repo updated and WARN logged:
  ```
  WARN n.s.MultiRevisionRepositoryStrategy - Non-fast-forward (force) update on bare repo for jorgee/nf-sleep [revision: main]: refs/heads/main <- refs/heads/main => FORCED (old=3851da8d..., new=d17b807d...)
  ```


modules/nextflow/src/main/groovy/nextflow/scm/MultiRevisionRepositoryStrategy.groovy
```diff
@@ -368,6 +372,22 @@ class MultiRevisionRepositoryStrategy extends AbstractRepositoryStrategy {
         return new RefSpec("$revision:$tagName")
     }
 
+    private void verifyFetchResult(FetchResult result, String revision, String repoLabel) {
+        for( TrackingRefUpdate update : result.getTrackingRefUpdates() ) {
+            final r = update.result
+            final detail = "${project} [revision: $revision]: ${update.localName} <- ${update.remoteName} => $r (old=${update.oldObjectId?.name()}, new=${update.newObjectId?.name()})"
+            if( r == RefUpdate.Result.FORCED ) {
+                log.warn "Non-fast-forward (force) update on $repoLabel for $detail"
```
ewels: Nice! How about a slight rephrasing? Just so that folks who see this out of the blue are more likely to understand..

```suggestion
                log.warn "Force-push detected on $repoLabel ($detail); resetting local copy to match remote"
```

'cc wordsmith @christopher-hakkaart

ewels (commented):
Nice - thanks @jorgee!

I think this should go into a 26.04 patch release if possible. It's already tripped up quite a few people.

modules/nextflow/src/main/groovy/nextflow/scm/MultiRevisionRepositoryStrategy.groovy
jorgee: Suggestion looks fine to me

jorgee:
> I think this should go into a 26.04 patch release if possible. It's already tripped up quite a few people.

I think patch is possible. Changes are minor, located in one class and this feature has been released for the first time in 26.04.

pditommaso (approved):
LGTM — the diagnosis matches the symptom, the `+` refspec is the right JGit mechanism, and `verifyFetchResult` correctly turns silent rejections into hard errors. The end-to-end test with a local bare upstream + force-push is exactly the regression coverage this needed.

A few nits, all optional / non-blocking:

**1. Make `verifyFetchResult` `@PackageScope` instead of `private`**

The tests currently reach it via `strategy.invokeMethod('verifyFetchResult', ...)` because the method is `private` on a `@CompileStatic` class. Marking it `@PackageScope` would let same-package Spock tests call it directly and keep static dispatch — no reflection in tests.

**2. Tighten the FORCED warn message**

```
Non-fast-forward (force) update on bare repo for jorgee/nf-sleep [revision: main]: refs/heads/main <- refs/heads/main => FORCED (old=3851da8d..., new=d17b807d...)
```
is information-dense and uses `localName <- remoteName`, which is the opposite of git's conventional `oldSha..newSha` ordering. Something like:
```groovy
log.warn "Force-push detected on ${project}@${revision} — resetting ${repoLabel} ref ${update.localName} (${shortOld}..${shortNew})"
```
would read closer to `git fetch`'s own output. Same information, just easier to scan.

**3. Drop the `repoLabel.capitalize()` trick**

`${repoLabel.capitalize()} fetch update for ...` works, but inlining the two call sites with hard-coded "Bare repo" / "Shared clone" strings is simpler and avoids the implicit dependency on the exact `repoLabel` values passed in.

PR #7108, opened 2026-05-06
Fall back to `nextflow_schema.json` types for CLI param coercion

ewels:
## Summary

Under syntax parser v2 (default in 26.04.0), CLI params arrive as strings. Unless the pipeline declares typed params in `main.nf` or non-`null` defaults in `nextflow.config`, those values stay as strings — which breaks param logic that expects e.g. numeric comparison on `params.max_cpus`.

This change reads property types from `nextflow_schema.json` (the nf-core convention) when it sits next to `main.nf`, and uses them to coerce CLI string values before they reach `ConfigBuilder`. Pipelines that already had a schema get typed params for free with no `main.nf` changes.

- New helper: `nextflow.config.SchemaParamsHelper`
- One call site: `CmdRun.run()`, immediately after `parsedParams()`
- Walks `properties` blocks at the top level and under `definitions` / `\$defs` / `allOf` / `oneOf` / `anyOf`
- Coerces `string` / `integer` / `number` / `boolean` (anything else left untouched)
- Best-effort: missing or malformed schema is silently ignored; un-coercible values stay as strings
- Gated on `NF.isSyntaxParserV2()` and respects `NXF_DISABLE_PARAMS_TYPE_DETECTION`

| Pipeline state | Before | After |
|---|---|---|
| Untyped main.nf, no schema | params are strings | params are strings (unchanged) |
| Untyped main.nf, **with schema** | params are strings (broken) | **typed via schema ✅** |
| Typed `params { }` in main.nf | typed via `ParamsDsl` | typed (no conflict) |

## Test plan

- [x] Unit tests for `SchemaParamsHelper` (12 cases: flat schema, nested `definitions`/`\$defs`/`allOf`, coercion of all four JSON Schema types, malformed schema, missing schema, mixed-case booleans, duplicate-name first-wins, `NXF_DISABLE_PARAMS_TYPE_DETECTION` opt-out, syntax parser v1 opt-out, pre-typed values from params files, typeless property)
- [x] End-to-end manual test with a minimal pipeline + `nextflow_schema.json` confirms `--max_cpus 8 --skip_qc TRUE --ratio 0.5` arrive as `Integer` / `Boolean` / `Float`
- [x] Existing `CmdRunTest`, `ParamsDslTest`, and config parser v2 tests still pass

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/config/SchemaParamsHelper.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/config/SchemaParamsHelper.groovy
@@ -1,0 +1,167 @@
+/*
+ * Copyright 2013-2026, Seqera Labs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package nextflow.config
+
+import java.nio.file.Files
+import java.nio.file.Path
+
+import groovy.json.JsonSlurper
+import groovy.transform.CompileStatic
+import groovy.util.logging.Slf4j
+import nextflow.NF
+import nextflow.SysEnv
+
+/**
+ * Coerces CLI param string values to typed values using a
+ * {@code nextflow_schema.json} file (JSON Schema, as used by nf-core
+ * pipelines) as a fallback type source.
+ *
+ * Under syntax parser v2, CLI params arrive as strings. When the pipeline
+ * has not declared typed params in main.nf or non-null defaults in
+ * nextflow.config, those strings stay strings -- and may break param logic
+ * that expects e.g. numeric comparison. If a {@code nextflow_schema.json}
+ * lives next to main.nf, this helper reads property types from it and
+ * coerces CLI values accordingly, giving the pipeline typed params for
+ * free without requiring main.nf changes.
+ *
+ * Coercion is best-effort and non-destructive: values that don't match a
+ * declared type are left as strings, and a missing or malformed schema
+ * is silently ignored.
+ *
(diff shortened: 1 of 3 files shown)


ewels:
## Note: duplication with the language server

Note that the Nextflow language server already has equivalent schema-walking code: [`ParameterSchemaVisitor.java`](https://github.com/nextflow-io/language-server/blob/main/src/main/java/nextflow/lsp/services/script/ParameterSchemaVisitor.java#L93-L160). It walks `nextflow_schema.json` for IDE type-checking on `params.foo` accesses, handling the same shape (`$defs` / `defs` / `definitions` / top-level `properties`) and the same `boolean`/`integer`/`number`/`string` mapping.

`ParameterSchemaVisitor` cannot be imported back into Nextflow core — the language server depends on `nf-lang`, not the other way around — and it returns Groovy AST `ClassNode` for IDE typing, not runtime values for CLI coercion.

The right consolidation is to lift the schema-walking into the shared `nf-lang` module as a small utility returning raw param attribute maps (`Map<String, Map<String, Object>>` — name → `{type, default, description, ...}`). Then:

- `SchemaParamsHelper` (Nextflow core) reads `type` and coerces
- `ParameterSchemaVisitor` (language-server) reads `type` + `default` + `description` and builds `ClassNode`

Both repos would agree on what counts as a property and which keys the schema is searched under.

If you like, I can re-shape this PR to put the schema reader in `nf-lang`, then open a parallel language-server PR that switches `ParameterSchemaVisitor` to consume it. This removes the duplication, but expands scope and is cross-repo (needs an `nf-lang` version bump on the language-server side once the new Nextflow ships).

modules/nextflow/src/main/groovy/nextflow/config/SchemaParamsHelper.groovy
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    /**
+     * Parse a JSON schema file and return a map of {@code paramName -> jsonType}
+     * (e.g. {@code "integer"}, {@code "number"}, {@code "boolean"}).
+     */
+    static Map<String,String> readSchemaTypes(Path schemaFile) {
+        try {
+            final root = new JsonSlurper().parse(schemaFile)
+            final types = new LinkedHashMap<String,String>()
```
nvnieuwk, line 86: The types keyword can also be a list of multiple types, how would you handle that case?

ewels:
@bentsherman can you remind me what you thought of this PR? I was waiting for feedback from you really, as to whether I should port it elsewhere.

Maybe you said that you didn't want it at all but if so then I forget the reasoning.

PR #7120, opened 2026-05-08
Honor token field in BitbucketServerRepositoryProvider

jorgee:
## Summary

- `BitbucketServerRepositoryProvider` previously inherited the default `RepositoryProvider` auth, which only reads `user`/`password` and ignored `token`. Configs supplying an HTTP access token hit BBS without an Authorization header (or with null Git credentials), producing 401 "Not authorized" at clone/branch lookup.
- Override `hasCredentials`, `getAuth`, and `getGitCredentials` so a token yields `Authorization: Bearer <token>` for REST and is used as the password for Git over HTTPS. `user`+`password` still works via Basic auth; when both `token` and `password` are set, the token wins (matches `BitbucketRepositoryProvider`).
- Add unit tests covering the three auth branches and an change a test using  `NXF_BITBUCKET_SERVER_ACCESS_TOKEN` to use the setToken() instead of setAuth()

## Test plan

- [x] `./gradlew :nextflow:test --tests "nextflow.scm.BitbucketServerRepositoryProviderTest"` (23 tests, 0 failures, 3 skipped — env-gated integration tests)

- [ ] Verify end-to-end pipeline launch from Seqera Platform against a BBS repository configured with token-only credentials.


pditommaso:
A few notes from a self-review pass:

### Important — bug in `'should list branches'` test

`modules/nextflow/src/test/groovy/nextflow/scm/BitbucketServerRepositoryProviderTest.groovy:113`

```groovy
def items = token.tokenize(':')
def config = items > 1
    ? new ProviderConfig(...).setUser(items[0]).setToken(items[1])
    : new ProviderConfig(...).setToken(token)
```

`items` is a `List<String>`, so `items > 1` compares a List to an Integer — it should be `items.size() > 1`. The test is `@Ignore`d so CI doesn't catch it, but anyone unignoring it to validate the fix against a real BBS endpoint (the deferred manual-test checkbox in the PR description) will hit a `ClassCastException` or silently take the wrong branch.

### Minor

1. **Repeated `getToken()` calls** in `BitbucketServerRepositoryProvider.getAuth()` and `getGitCredentials()`. Stylistic only — caching to a local would read a bit cleaner:
    ```groovy
    final tkn = getToken()
    if( tkn ) return new String[] { "Authorization", "Bearer " + tkn }
    ```

2. **Pre-existing typo** `platform:'bitbucketsever'` (missing `r`) at test lines 114, 115, 130. Not introduced by this PR — flagging in case you want to clean it up while in the area.

pditommaso:
Umm, new azure tests?! 😕

jorgee:
I haven't changed anything on azure in thia PR. @bentsherman, have you updated rnaseq-nf?

pditommaso (approved):

PR #7121, opened 2026-05-08
Change `stdout` outputs to `stderr`, to not break piped output

### LINKED ISSUE #5942: log.error and error does not print to sderr
## New feature

I would love an easy way to print to stderr.
log.error print to stdout, but prints in red with the ERROR keyword before.
I understand that it's for logging purpose first, but I would love for an easy way print to stderr in Nextflow .
While we're at it, if we could use the same method to print similar messages in the plugins that would be perfection.
If I look at this for example: https://github.com/nextflow-io/nf-schema/blob/ecf159f53d45200ef70920c03e75077b5689a386/plugins/nf-schema/src/main/nextflow/validation/ValidationExtension.groovy#L319-L323
the log.error is printed to stdout while the other message is indeed printed to both stdout and stderr.

## Use case 

I would want to be able to print my error message from the pipeline side, error and exit right away without catching other unrelated errors.

## Suggested implementation 

I think the error method should the right way to do it, if it was printing to stderr.
Still I did notice some unrelated messages that appears and throws some red herrings that confuses users (and me too)

### LINKED ISSUE #7116: `nextflow inspect|config -o json` doesn't return valid json

I would like to directly run the output of these commands through a json parser. 
Unfortunately, the output is not pure JSON, e.g.

```bash
nextflow inspect . -o json
Monitor the execution with Seqera Platform using this URL: https://cloud.seqera.io/user/***
{    
    "processes": [
....

```

or 

```bash
nextflow config -o json 
WARN: Unrecognized config option 'validation.defaultIgnoreParams'
WARN: Unrecognized config option 'validation.monochromeLogs'
{
    "params": {
```

Create a minimal nextflow config e.g

```groovy
params {
    input = null
}
plugins {
    id 'nf-schema@2.5.1' // Validation of pipeline parameters and creation of an input channel from a sample sheet
}
validation {
    defaultIgnoreParams = ["genomes"]
    monochromeLogs = params.monochrome_logs
}

```

and run ` nextflow config -o json` on it.

### Program output 

```console
WARN: Unrecognized config option 'validation.defaultIgnoreParams'
WARN: Unrecognized config option 'validation.monochromeLogs'
{
    "params": {
        "input": null
    },
    "plugins": [
        "nf-schema@2.5.1"
    ],
    "validation": {
        "defaultIgnoreParams": [
            "genomes"
        ],
        "monochromeLogs": null
    }
}
```

### Environment 

Version: 26.04.0 build 12031
System: Mac OS X 26.4.1
Runtime: Groovy 4.0.31 on OpenJDK 64-Bit Server VM 21.0.2+13-58

### Additional context

I guess the source for both is different, but the general assumption I had is that i get pure json output in stdout when running `-o json`.

### LINKED ISSUE #7340: `nextflow config` writes warnings to stdout, corrupting redirected config output
### Expected behaviour and actual behaviour

Running:

```bash
nextflow config nf-core/rnaseq > test.rnaseq.config
```

should write a valid resolved Nextflow configuration to `test.rnaseq.config`.

Instead, warning messages are included at the beginning of the redirected output:

```text
WARN: Unrecognized config option 'validation.defaultIgnoreParams'
WARN: Unrecognized config option 'validation.monochromeLogs'
```

As a result, the generated file is not valid Nextflow configuration syntax and cannot subsequently be supplied with `-c`. Nextflow fails while parsing the first warning line:

```text
Error test.rnaseq.config:1:5: Unexpected input: ':'
```

The warnings should be written to standard error rather than standard output, so that redirecting the resolved configuration produces a valid file.

```bash
nextflow config nf-core/rnaseq > test.rnaseq.config
head test.rnaseq.config
```

The resulting file begins with:

```text
WARN: Unrecognized config option 'validation.defaultIgnoreParams'
WARN: Unrecognized config option 'validation.monochromeLogs'
params {
   input = null
   genome = null
   splicesites = null
   gtf_extra_attributes = 'gene_name'
   gtf_group_features = 'gene_id'
   skip_gtf_filter = false
   skip_gtf_transcript_filter = false
```

Trying to run the pipeline with this file:

```bash
nextflow run nf-core/rnaseq \
    -r 3.26.0 \
    -profile apptainer \
    --input sycon_fetchngs/samplesheet/samplesheet.csv \
    --fasta GCF_964019385.1_ocSycCili1.1_genomic.fna \
    --gtf GCF_964019385.1_ocSycCili1.1_genomic.gtf \
    --aligner star_salmon \
    --use_rustqc \
    --save_reference \
    --outdir sycon_rnaseq \
    -c test.rnaseq.config \
    -preview \
    -resume
```

### Program output

```text
Nextflow 26.04.6 is available - Please consider updating your version to it

 N E X T F L O W   ~  version 26.04.4

WARN: This Nextflow version supports a new Multi-revision strategy for managing the SCM repositories, but 'nf-core/rnaseq:3.26.0' is single-revision legacy strategy - Please consider to update the repository with the 'nextflow pull -migrate' command.
Error test.rnaseq.config:1:5: Unexpected input: ':'
│   1 | WARN: Unrecognized config option 'validation.defaultIgnoreParams'
╰     |     ^

ERROR ~ Config parsing failed

```

### Environment

* Nextflow version: `26.04.4`
* Pipeline: `nf-core/rnaseq 3.26.0`
* Java version: `openjdk version "17.0.10" 2024-01-16`
* Operating system: `Linux`
* Distribution/kernel: `Ubuntu 24.04.4 LTS (GNU/Linux 6.8.0-106-generic x86_64)`
* Shell version: `GNU bash, version 5.2.21(1)-release (x86_64-pc-linux-gnu)`
* Execution profile: `apptainer`

### Additional context

The issue may be connected to validation configuration options present in `nf-core/rnaseq 3.26.0`, but regardless of whether those options are recognised, their warnings should not become part of the exported configuration.

ewels:
Prevent messages such as `Monitor the execution with Seqera Platform using this URL:` from breaking the JSON output, eg. from `nextflow inspect`. Allows piping into `jq` and other tools.

See https://github.com/nextflow-io/nextflow/issues/7116

<details>

# Route progress, banners and log fallbacks to stderr

Some Nextflow commands now emit machine-readable output on stdout (`nextflow inspect -o json`, `nextflow run` outputs, etc.). Today, log messages, progress UI, banners and spinners are also written to stdout, which makes that output difficult to pipe into downstream tools (`jq`, etc.).

This change applies a simple rule: **a command's main output stays on stdout; everything else (progress, status, banners, log fallbacks) moves to stderr**. Human-readable command output (e.g. `--help`, `--version`, `nextflow info`, `nextflow view`, `nextflow list`) is unaffected — it remains on stdout because it *is* the main output of those commands.

## Files changed

- **`modules/nextflow/src/main/groovy/nextflow/util/LoggerHelper.groovy`** — the global logback appender's `System.out` fallback (used when no `AnsiLogObserver` is registered, or the observer isn't started yet) now writes to `System.err`. Highest leverage change: affects log output from every command, including the "Monitor the execution with Seqera Platform" sticky message that breaks `nextflow inspect | jq` today.
- **`modules/nextflow/src/main/groovy/nextflow/trace/AnsiLogObserver.groovy`** — progress rendering, summary block, and info/warning/error UI panels now use `AnsiConsole.err` instead of `AnsiConsole.out`. `AnsiConsole.systemInstall()` is unchanged.
- **`modules/nextflow/src/main/groovy/nextflow/cli/CmdRun.groovy`** — the `N E X T F L O W ~ version …` banner and the `Launching … [run-name] - revision: …` line now use `AnsiConsole.err`.
- **`modules/nextflow/src/main/groovy/nextflow/util/SpinnerUtil.groovy`** — animated spinners (and the non-ANSI plaintext fallback in `stopWithMessage`) now write to stderr.
- **`modules/nextflow/src/main/groovy/nextflow/trace/AgentLogObserver.groovy`** — class-level docstring corrected to say "standard output" instead of "standard error". The implementation, the integration tests, and the design intent all agree that the structured `[PIPELINE]`/`[WORKDIR]`/`[SUCCESS]` lines emitted in `NXF_AGENT_MODE` *are* the meaningful programmatic output of agent mode and belong on stdout; only the docstring was wrong.

## Deliberately unchanged

- **`Session.printConsole()`** stays on stdout. It is the data path used by the `view()` channel operator, `echo true` task stdout forwarding, and the workflow `output {}` DSL block — i.e. exactly the kind of programmatic output stdout is for.
- **`AgentLogObserver` output streams** — agent-mode structured output (`[PIPELINE]`, `[WORKDIR]`, `[SUCCESS]`, `[ERROR]`, `[FAILED]`, etc.) stays on stdout. In `NXF_AGENT_MODE` these lines *are* the main output, designed to be parsed by AI tooling.
- **`--help`, `--version`, `nextflow info`, `nextflow view`, `nextflow list`, `nextflow clean`, `nextflow clone`, `nextflow plugin create`, `nextflow log`, `nextflow lineage`, `nextflow inspect`, `nextflow config`, `nextflow lint`, `nextflow fs`, `nextflow secrets`, `wave`, etc.** — these all already write their main output on stdout and are unaffected.
- **`nf-tower` `LaunchCommandImpl`** — `nextflow launch` currently mixes its banner and the workflow ID on stdout. Left for a follow-up that can introduce a JSON output mode.
- **`nf-tower` `AuthCommandImpl`** — interactive prompts and confirmations stay on stdout (these commands aren't piped).

## Net effect

`nextflow inspect | jq`, `nextflow run … | …` (and equivalents for other pipeable commands) no longer have stdout polluted by progress UI, banners, spinners, or log fallbacks. The "Monitor the execution with Seqera Platform using this URL: …" sticky message in particular — which is emitted via `log.info(STICKY, …)` from the Tower observer — now correctly lands on stderr in commands like `nextflow inspect` because it falls through the no-observer path in `LoggerHelper`.

## Verification

- `./gradlew :nextflow:compileGroovy` passes.
- `agent-output.nf` and `agent-output-error.nf` integration tests pass: `[PIPELINE]`/`[WORKDIR]`/`[SUCCESS]` lines from `AgentLogObserver` and `World received: Hello from agent mode` from `Session.printConsole()` continue to land on stdout, as the tests assert.

</details>


jorgee (commented):
Overall, it looks fine with me.

Some concerns:

- It could be a visible change for the user; shouldn't we warn about it? Is the merge commit message included in the changelog enough, or is anything else needed in the docs?

- I think it could also be good to run e2e test just in case they parse one of the changed messages from stdout to avoid strange failures in the future.

- Missing regression test checking the issue case: A small integration test that runs `nextflow  inspect -o json` on a trivial pipeline and asserts stdout parses as valid JSON would lock this in and prevent regressions on the no-observer log-fallback path (which is the easiest one to accidentally re-break).

ewels:
Thanks @jorgee!

* I leave the docs question to @christopher-hakkaart but for me, changelog is enough. This is basically reverting to what I suspect is expected behaviour for most people? But I don't feel strongly about this.
* Sure - how do we do this on a PR?
* Pushed in 66667f07a

christopher-hakkaart:
I think Changelog is probably enough here. At least from my understanding, this isn't a breaking change, but rather a visual fix.

ewels:
It's only breaking if people have automations / tests etc. set up that rely on checking for specific strings within `STDOUT` Nextflow output (which will now become `STDERR`). I'm sure that this will affect some people, but I'm hoping not many. I think much more severe breaking changes have gone out in the past with no more than a changelog entry 😅  This is what the changelog is for, after all.

Can always try to highlight it at the top of the changelog if we want?

ewels:
@jorgee - how do I trigger an e2e test / get this PR moving?

bentsherman:
Make a commit with `[e2e prod]`

ewels:
I think it worked..? :)

bentsherman:
You have to look up the platform runs to see if they actually succeeded. The CI check only verifies the launch. You can find the runs through the CI logs

ewels:
It seems that they run under Paolo's user space? https://cloud.seqera.io/user/pditommaso/watch/4IPHkGZiiLv4d2

I don't have access to that.

I think, anyway.. Not entirely clear to me how this works - trying to figure it out now.

ewels:
I found the GitHub actions logic pretty confusing to follow, so have put in a separate PR to try to clean it up a bit: https://github.com/nextflow-io/nextflow/pull/7192

Also adds the ability to trigger tests with a comment and a manual dispatch (GitHub UI actions page dropdown), in addition to the existing commit message. Also prints the e2e platform URL and next steps on the CI actions summary.

bentsherman (approved):

pditommaso:
Let me double check this change 🙏

pditommaso:
LGTM! ✅

PR #7137, opened 2026-05-13
[DOC]: add `flat: false` when replacing toSortedList

DriesSchaumont:


docs/tutorials/static-types-operators.md
```diff
@@ -481,7 +481,7 @@ channel.of(3, 2, 1, 4)
 
 // after
 channel.of(3, 2, 1, 4)
-    .collect()
+    .collect(flat: false)
```
bentsherman, line 484: This option is not supported when static typing is enabled

PR #7139, opened 2026-05-13
Tighten CI security and remove dead code in workflows

ewels:
Follow-up to `#gha-updates` (#7138) after a deeper pass through `build.yml`, `claude.yml`, `docs.yml`. Driven by a manual security review on top of [zizmor](https://docs.zizmor.sh) — no zizmor findings before or after, but the pass surfaced real issues zizmor doesn't catch: secret-dumping debug lines, dead permissions, a multi-line `$GITHUB_ENV` bug, and a trigger that fires the runner with no possible match.

## `build.yml`

- **Remove `env | sort` (×2) and `cat $HOME/.nextflow/scm`.** GitHub's log masking is a fragile exact-string replace; dumping every env var or `cat`ing a file that contains `NXF_GITHUB_ACCESS_TOKEN` is the pattern that leaked secrets in the March 2025 `tj-actions/changed-files` incident. Both looked like leftover debug.
- **Rewrite `Get the commit message`.** `${{ secrets.GITHUB_TOKEN }}`, `${{ github.event.pull_request.head.sha }}` and `${{ github.repository }}` moved into `env:`; `curl + jq` swapped for `gh api` (pre-installed, picks up `GH_TOKEN` automatically). Same shape as several recent supply-chain incidents — safe today only because GitHub controls every value.
- **GCP credentials → `$RUNNER_TEMP`.** Were being written to `$GITHUB_WORKSPACE`, which an over-broad `upload-artifact` glob could exfiltrate by accident. `$RUNNER_TEMP` is outside the workspace and auto-cleaned.
- **`test-e2e` → `permissions: {}`.** Job dispatches a workflow in `seqeralabs/showcase-automation` via `AUTOMATION_GITHUB_TOKEN` (separate PAT) and touches nothing here. Previous `actions: write` + `contents: write` on the default token were dead.
- **Heredoc for `COMMIT_MESSAGE` in `$GITHUB_ENV`.** Old form embedded literal quotes into the value and broke on multi-line / `"`-containing messages. Latent rather than blocking (run.sh just greps for `[e2e prod]`) but worth fixing.
- **`release` permissions → `contents: write` only.** Was `actions/contents/packages/pull-requests/issues: write`. Verified by grepping `*.gradle`/`*.sh`/`Makefile` for `ghcr.io`, `gh issue`, `gh pr`, `gh workflow` — only hit is `gh release create/upload`. Docker pushes go to Docker Hub and `public.cr.seqera.io`, not `ghcr.io`.
- **Workflow-level `concurrency:`.** Cancels in-flight `pull_request` runs on new pushes; `master` / `STABLE-*` / `test*` / `dev*` are never cancelled so a partial release can't be aborted.

## `claude.yml`

- **Drop `assigned` from `issues:` trigger.** The `if` clause only checks `@claude` in `issue.body`, so every assignment was starting a runner just to skip.
- **`concurrency:` keyed on issue/PR number.** Multiple `@claude` mentions on the same thread queue rather than race.

## `docs.yml`

- **Python `3.9.25` → `3.14`.** Renovate had us on a near-EOL branch; no docs-specific reason to stay.
- **`cache: 'pip'`** with `cache-dependency-path: docs/requirements.txt`. ~30s/PR.
- **Workflow-level `concurrency:`.**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 1
           submodules: true
@@ -62,7 +62,7 @@ jobs:
           NXF_GITHUB_ACCESS_TOKEN: ${{ secrets.NXF_GITHUB_ACCESS_TOKEN }}
 
       - name: Setup Java ${{ matrix.java_version }}
-        uses: actions/setup-java@v4
+        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
         with:
           java-version: ${{matrix.java_version}}
           distribution: 'temurin'
@@ -99,7 +99,7 @@ jobs:
 
       - name: Publish tests report
         if: always()
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: report-unit-tests-jdk-${{ matrix.java_version }}
           path: |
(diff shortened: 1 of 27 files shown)


pditommaso (approved):

bentsherman:
@ewels can you fix the merge conflicts? then I will merge it

ewels:
@copilot resolve the merge conflicts in this pull request

ewels:
@copilot careful - I think you added "persist-credentials: false" back to the build.yml workflow. But the credentials are needed for the script that pushes the release. This edit was reverted in the upstream PR and comments explaining why added. You just removed this in the merge conflict resolution I think.

ewels:
Should be good now I think @bentsherman

PR #7141, opened 2026-05-13
Expose dataset file size in seqera:// file attributes

jorgee:
## Summary

- Bump `io.seqera:tower-api` to `1.158.0`, which adds the new `fileSize` field on `DatasetVersionDto`.
- Map `fileSize` in `SeqeraDatasetClient.mapVersion` when parsing the versions response.
- Use the resolved `DatasetVersionDto` (not `DatasetDto`) to back `SeqeraFileAttributes` for dataset files: `size()` returns `version.fileSize`; `creationTime()`/`lastModifiedTime()` use the version's own timestamps; `fileKey()` is now `${datasetId}@${version}` so it stays unique per file.
- `SeqeraFileSystemProvider.readAttributes` resolves the dataset version (reusing `resolveVersion`, whose result is memoized in `SeqeraFileSystem.versionCache`) and passes it to `SeqeraFileAttributes`.

Previously `Files.size(path)` on a `seqera://<org>/<workspace>/datasets/<name>` path always returned `0`; it now returns the real size reported by the platform.

## Test plan

- [x] `./gradlew :plugins:nf-tower:test` — full plugin suite passes
- [x] Updated `SeqeraDatasetClientTest` to assert `fileSize` parsing in `listVersions`
- [x] Updated `SeqeraFileSystemProviderTest` so the dataset-file `readAttributes` case stubs `/datasets/{id}/versions` and asserts `attrs.size() == 4096L`


jorgee:
Require the new tower-api package once changes in Platform API are pushed.

plugins/nf-tower/src/main/io/seqera/tower/plugin/dataset/SeqeraDatasetClient.groovy
```diff
@@ -97,6 +97,7 @@ class SeqeraDatasetClient {
         log.debug "SeqeraDatasetClient GET $url"
         final resp = towerClient.sendApiRequest(url)
         checkFsResponse(resp, url)
+        log.debug("RESPONSE: $resp")
```
bentsherman: Did you mean to keep these logs? My sense is that they are not needed

pditommaso (changes requested):
## Review

Thanks for the PR. The model change (size as a per-version property, backed by `DatasetVersionDto`) is the right direction, and the caching reuse via `resolveVersion` keeps the extra round-trip amortized. However, there is a silent regression and some debug instrumentation that should be addressed before merge.

---

### 🔴 Critical

**`lastModifiedTime()` silently regresses to `Instant.EPOCH`**

`SeqeraDatasetClient.mapVersion` (`SeqeraDatasetClient.groovy:194-205`) never sets `lastUpdated` on the constructed `DatasetVersionDto`, while `SeqeraFileAttributes.lastModifiedTime()` now reads `version.lastUpdated`. As a result, every dataset-file path will report `Instant.EPOCH` for `lastModifiedTime()` — and `lastAccessTime()` too, since it delegates to it. The bug is masked because the existing test fixtures (`versionsJson()`) omit `lastUpdated` and no test asserts the timestamp.

Suggested fix in `mapVersion`:

```groovy
dto.lastUpdated = m.lastUpdated ? OffsetDateTime.parse(m.lastUpdated as String) : null
```

Plus a test in `SeqeraFileSystemProviderTest` that asserts `attrs.lastModifiedTime()` matches the version's `lastUpdated`.

---

### 🟠 Important

**1. Debug logging dumps full API response bodies and parsed maps**

Four new statements in `SeqeraDatasetClient.groovy`:
- Line 100 — `log.debug("RESPONSE: $resp")` after `listDatasets`
- Line 127 — `log.debug("RESPONSE: $resp")` after `listVersions`
- Line 181 — `log.debug("Mapping dataset $m")` in `mapDataset`
- Line 195 — `log.debug("Mapping version $m")` in `mapVersion`

`resp.toString()` includes the entire response body. With `DEBUG` enabled (common in support bundles / CI), this dumps full dataset and workspace inventories per call. They look like leftover instrumentation from developing the PR and are also stylistically inconsistent with the rest of the file (parens form vs. the existing `log.debug "..."` style).

Suggestion: remove all four. If any are retained for diagnostics, log only the URL and `resp.code`, never the body.

**2. PR description / diff mismatch on `tower-api` version**

The PR description says "Bump `io.seqera:tower-api` to `1.154.0`", but `build.gradle` actually bumps to `1.158.0`. Please confirm `1.158.0` is the intended version and update the description so reviewers aren't reading the wrong delta (the `1.154.0` → `1.158.0` interval may include other DTO changes worth a sanity check).

**3. `(m.fileSize as Long) ?: 0L` collapses null/missing into 0**

`SeqeraDatasetClient.groovy:200`. A genuinely empty file and a response from an older Tower deployment that omits `fileSize` become indistinguishable. This preserves the prior behavior of `Files.size() == 0` for the legacy case, so it's acceptable, but worth a short comment so future readers know "size unknown" is not a distinct state.

---

### 🟢 Minor (nice-to-have)

- **Missing trailing newline** in `SeqeraFileAttributes.groovy` — visible as `\ No newline at end of file` in the diff.
- **`fileKey()` format change**: from `dataset.id` (String) to `"${datasetId}@${version}"`. `SeqeraFileSystemProvider.isSameFile` uses path equality so this is safe in-tree, but any external caller that cached the previous fileKey will see different values. Worth a one-line changelog note.
- **Extra HTTP round-trip on first `readAttributes`** per dataset (subsequent calls hit `versionCache`). Acceptable; just note that externally-published new versions remain invisible until cache invalidation — already a known property of this filesystem.
- **Test coverage gaps**, in order of value:
  - Pinned-version size (`seqera://acme/research/datasets/samples@1` → expect `100L`, not `4096L`) — would catch any regression in version pinning.
  - Timestamp assertion on `attrs.lastModifiedTime()` / `attrs.creationTime()` — would have caught the Critical issue above.
  - Legacy response without `fileSize` → expect `attrs.size() == 0L`.

---

### Strengths

- Routing `size`, `creationTime`, `lastModifiedTime` through the version DTO is the right model.
- `resolveVersion` reuse + the existing `versionCache` keeps the cost amortized.
- Test fixtures were updated consistently across both the client and provider tests.
- Javadoc on `SeqeraFileAttributes` was updated to reflect the new source of truth.

plugins/nf-tower/src/main/io/seqera/tower/plugin/dataset/SeqeraDatasetClient.groovy
jorgee: Removed — these were leftover instrumentation from development. All four debug statements (the two RESPONSE dumps plus the two `Mapping ...` lines) are gone.

jorgee:
Thanks for the thorough review — addressed in 769a59234.

**🔴 Critical — `lastModifiedTime()` → `Instant.EPOCH`**
Confirmed and fixed. `mapVersion` now parses `lastUpdated`:
```groovy
dto.lastUpdated = m.lastUpdated ? OffsetDateTime.parse(m.lastUpdated as String) : null
```
`SeqeraFileSystemProviderTest` now asserts `creationTime()`/`lastModifiedTime()`/`lastAccessTime()` against the resolved version's timestamps.

**🟠 Important**
1. Removed all four debug statements (lines 100, 127, 181, 195).
2. PR description updated to `1.158.0`.
3. Added a comment at the `fileSize ?: 0L` line noting "size unknown" is not distinguishable from an empty file.

**🟢 Minor**
- Trailing newline added to `SeqeraFileAttributes.groovy`.
- Test coverage: added pinned-version size (`samples@1` → `100L`), the timestamp assertions above, and a legacy response without `fileSize` → `size() == 0L`.
- `fileKey()` changelog note: per our convention the plugin changelog is updated at release time, so I left it for the release commit rather than this feature commit.

`./gradlew :plugins:nf-tower:test` passes.

pditommaso (approved):
Well done

pditommaso:
Reviewed the follow-up commit `769a59234`. All prior review points are addressed:

- **`lastModifiedTime()` regression fixed** — `mapVersion` now parses `lastUpdated`, so `SeqeraFileAttributes.lastModifiedTime()` returns the real timestamp instead of `Instant.EPOCH`.
- **Stray debug logs removed** — all 4 sites (`RESPONSE: $resp` ×2, `Mapping dataset`, `Mapping version`) cleaned up.
- **`fileSize ?: 0L` semantics documented** — comment explains that "size unknown" (legacy Tower) is indistinguishable from a genuinely empty file.
- **Trailing newline added** to `SeqeraFileAttributes.groovy`.
- **Test matrix expanded**:
  - `SeqeraDatasetClientTest` asserts both `dateCreated` and `lastUpdated` parsing.
  - `SeqeraFileSystemProviderTest` now covers (a) timestamps on the default-version path, (b) pinned-version `samples@1` size + timestamps, and (c) legacy response with no `fileSize` → `size() == 0L`.

All required CI checks green. Merging.

PR #5738, opened 2025-02-03
feat(wave client): Support local manifests in `FUSION_CONTAINER_CONFIG_URL`

alberto-miranda:
## Description
This PR enables Nextflow to accept local files as an additional source for `FUSION_CONTAINER_CONFIG_URL`. This can be useful for testing `fusion` and also for offline deployments. Local files can be specified using the `file://` schema or with no schema at all.

PR #7142, opened 2026-05-13
feat(wave): Support fetching local files in `FUSION_CONTAINER_CONFIG_URL`

alberto-miranda:
## Description

This change allows `FUSION_CONTAINER_CONFIG_URL` (and the corresponding Nextflow config option `fusion.containerConfigUrl`) to accept a local filesystem path as an alternative to an http(s) URL. The path must be expressed as a `file:` URI.

>[!note]
>   This PR resurrects #5738, adapted to the current changes in `master`.

### Rationale
Today, `FUSION_CONTAINER_CONFIG_URL` only accepts `http(s)` URLs. The Wave client fetches the JSON manifest from the provider URL and applies the declared layers to the task container. This works fine in production, but makes iterating on manifest changes painful: any change (e.g. bumping a Fusion layer, swapping in a development build, or testing a new Fusion Snapshots variant) means publishing a new JSON to a reachable URL before every run, then chasing down stale artefacts in object storage afterwards.

By accepting a local manifest, it is possible to decouple the development loop from the production pipeline: engineers can edit a JSON on disk and re-run a pipeline immediately, instead of round-tripping every change through object storage and an externally reachable URL.

### Changes

1. Renamed `FusionConfig.containerConfigUrl()` to `FusionConfig.containerConfigURI()` (return type `URL` → `URI`) so `file:` URIs travel cleanly through the resolution path.
2. Relaxed `FusionConfig.validProtocol()` to also accept `file:` URIs.
3. Updated the validation error message in `FusionConfig.getConfig()`.
4. `WaveClient.resolveContainerConfig()` now collects `List<URI>` (previously `List<URL>`).
5. Ported `replaceFusionArch()` to accept `URI` (no-op on local manifests, since the arch-suffix regex doesn't match).
6. Routing in new `fetchContainerConfig(URI)`:
   - `http`/`https` to the existing HTTP path,
   - `file:` to a new `fetchContainerConfig(Path)` that reads from disk,
   - Anything else throws.
7. Extra tests for the new logic in `FusionConfigTest` and `WaveClientTest`.


alberto-miranda:
@pditommaso @bentsherman 

I double checked the CI failures and I believe the `test (25, test_azure)` and `test (17, test_azure)` are unrelated to this PR: 

If I understood the `validation/azure.sh` test correctly, it relies on the defaults shipped by [`nextflow-io/rnaseq-nf`](https://github.com/nextflow-io/rnaseq-nf), right? If so, a couple of direct-to-`master` commits in [`nextflow-io/rnaseq-nf`](https://github.com/nextflow-io/rnaseq-nf) earlier today seem to have replaced those defaults with `null`:

  ```diff
  - params.reads = "${baseDir}/data/ggal/ggal_gut_{1,2}.fq"
  - params.transcriptome = "${baseDir}/data/ggal/ggal_1_48850000_49020000.Ggal71.500bpflank.fa"
  + params.reads = null
  + params.transcriptome = null
```

I'm happy to make changes to this PR if my diagnosis is wrong, just say the word!

**Comment by pditommaso** [comment]
Apologies the silly question, are not local path already supported via `file:/` prefix?

**Comment by alberto-miranda** [comment]
> Apologies the silly question, are not local path already supported via `file:/` prefix?

@ Paolo Not silly at all, that was my read too at first, and it took some digging to see the gap:

When trying to pass a file through `FUSION_CONTAINER_CONFIG_URL` the pipeline fails with an `invalid URI scheme file`:

```bash
$ curl https://fusionfs.seqera.io/releases/v2.5-amd64.json > foo.json
$ export FUSION_CONTAINER_CONFIG_URL=file://$PWD/foo.json
$ nextflow run hello -profile aws
Nextflow 26.04.1 is available - Please consider updating your version to it

 N E X T F L O W   ~  version 25.10.5

NOTE: Your local project version looks outdated - a different revision is available in the remote repository [d828daeef7]
Launching `https://github.com/nextflow-io/hello` [backstabbing_lamarr] DSL2 - revision: afff16a9b4 [master]

[-        ] sayHello -
ERROR ~ Error executing process > 'sayHello (3)'

Caused by:
  invalid URI scheme file



```

A `file:` URI does pass the `FusionConfig.validProtocol()` check, but `WaveClient.fetchContainerConfig(URL)` relies on  `java.net.http.HttpRequest.Builder.uri(URI)` which accepts http/https. So `file:///` was effectively never wired through. 

This PR closes the gap by dispatching local-file URIs through `Paths.get(...).text` instead of the HTTP client. The http/https path is unchanged.

PS: Bare paths (no `file:///`) get the same treatment.

pditommaso:
It maybe a glitch in the uri resolution, however it would be preferable to stick using uri prefix (not adding bar file paths)

alberto-miranda:
> It maybe a glitch in the uri resolution, however it would be preferable to stick using uri prefix (not adding bar file paths)

@pditommaso No problem from me, dropped bare paths with 21f8364

pditommaso (changes requested):
Thanks for resurrecting #5738. The core change is small and well-scoped, and the tests cover the happy path with real temp files. A few items worth addressing before merge — the dev-loop UX this PR is enabling is undermined by some rough edges in validation, error reporting, and the arch-rewrite path.

## Important

### 1. `FusionConfig.validProtocol()` admits forms that crash later
[`FusionConfig.groovy:172-174`](https://github.com/nextflow-io/nextflow/blob/comp-869/local-fusion-manifest/modules/nextflow/src/main/groovy/nextflow/fusion/FusionConfig.groovy#L172-L174)

The check is `url.startsWith('file:')`, which accepts inputs that later blow up in `Paths.get(URI)`:
- `file://host/tmp/manifest.json` → `IllegalArgumentException: URI has an authority component`
- `file:relative.json` / `file:./x.json` → `IllegalArgumentException: URI is not hierarchical`

A user with `file://path/to/json` (a very common mistake) gets a stack trace from deep inside `WaveClient`, far from the offending config line.

Suggest tightening to `file:/` (matching the intent of 21f8364) — or, more robustly, parse to a `URI` and validate `scheme == 'file' && authority == null && path?.startsWith('/')`.

### 2. Local-file fetch has no user-friendly error messages
[`WaveClient.groovy:450-454`](https://github.com/nextflow-io/nextflow/blob/comp-869/local-fusion-manifest/plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy#L450-L454)

`fetchContainerConfig(Path)` does `configPath.text` directly. Failure modes surface as raw NIO exceptions with no Fusion-specific context:
- missing file → `NoSuchFileException`
- directory → `IOException: Is a directory`
- permission denied → `AccessDeniedException`

The HTTP path wraps non-2xx responses in `BadResponseException` with the configured URL — the file path should do the same, e.g. `throw new IllegalStateException("Cannot read Fusion container config from $configPath", e)`. Optionally pre-validate with `Files.isRegularFile` / `Files.isReadable` for a cleaner message.

### 3. `replaceFusionArch()` silently rewrites local file paths
[`WaveClient.groovy:407-413`](https://github.com/nextflow-io/nextflow/blob/comp-869/local-fusion-manifest/plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy#L407-L413)

The PR description says the function is a "no-op on local manifests", but the regex `/(?<=[-_])(amd64|arm64)(?=\.)/` applies to the entire URI string regardless of scheme. So `file:///home/me/fusion-amd64.json` on a `linux/arm64` task is silently rewritten to `fusion-arm64.json`, and Wave then tries to read a file the developer didn't intend.

This is plausible in practice — developers debugging arch-specific manifests will name files this way.

Two options:
- Skip the rewrite entirely when `uri.scheme == 'file'` (matches PR description, cleanest).
- Or keep current behaviour and document it + add a test.

### 4. No docs / `@Description` update
[`docs/reference/config.md:883-884`](https://github.com/nextflow-io/nextflow/blob/master/docs/reference/config.md#L883) and the `@Description` in [`FusionConfig.groovy:73-77`](https://github.com/nextflow-io/nextflow/blob/comp-869/local-fusion-manifest/modules/nextflow/src/main/groovy/nextflow/fusion/FusionConfig.groovy#L73-L77) still describe `fusion.containerConfigUrl` as an HTTP-only "URL". Without doc updates, the local-loop workflow this PR enables is undocumented. A sentence noting that `http(s)://` and `file:/` schemes are supported (with a short example) would do it.

## Minor

- **Parameterised test for URI shapes.** The local-file test uses `manifest.toUri().toString()` which produces `file:///...`. Given URI parsing's quirks, a `@Unroll` table covering both `file:/...` and `file:///...` (and rejecting `file://host/...` / `file:relative`) would be more reassuring than a single happy-path case. ([`WaveClientTest.groovy:957-971`](https://github.com/nextflow-io/nextflow/blob/comp-869/local-fusion-manifest/plugins/nf-wave/src/test/io/seqera/wave/plugin/WaveClientTest.groovy#L957-L971))
- **Redundant `@Memoized` on three overloads.** A second call with the same URI hits the dispatcher's cache anyway, so memoizing the URL/Path inner methods just duplicates cache keys. Consider keeping `@Memoized` only on the URI dispatcher. ([`WaveClient.groovy:439-457`](https://github.com/nextflow-io/nextflow/blob/comp-869/local-fusion-manifest/plugins/nf-wave/src/main/io/seqera/wave/plugin/WaveClient.groovy#L439-L457))
- **`synchronized` + `@Memoized`.** `@Memoized` already produces thread-safe lazy caches; the `synchronized` keyword on top serializes calls across *different* URIs/paths on the same instance. Pre-existing pattern, but worth a follow-up — per-key sync would give better throughput when many distinct manifests are fetched.
- **`Path.of(uri)` instead of `Paths.get(uri)`.** Lets you drop the new `java.nio.file.Paths` import. Stylistic only.
- **Security note.** Accepting `file:` URIs from `fusion.containerConfigUrl` / `FUSION_CONTAINER_CONFIG_URL` widens the local-file-read surface driven by config. Likely fine because the Wave client runs in the same trust domain as the user submitting the workflow, but worth confirming with the Wave/Platform team that there's no context where this config could come from a less-trusted source. No code change needed if the threat model is OK with it.

alberto-miranda:
@pditommaso I addressed your comments but the docs workflow seems to be failing because:
1. `imghdr` was removed from the Python standard library in `3.13`
2. the repo pins `sphinx==5.3.0` which still imports it
3. the workflow pins `python-version: '3.14'`

I'm happy to fix it in this PR or a follow up, but don't want to muck up the existing setup, so I would appreciate some guidance on the preferred approach :pray:

pditommaso:
Sorry for the long delay. All set here!

pditommaso (approved):

PR #7143, opened 2026-05-13
Fix azure rnaseq-nf test

bentsherman:
rnaseq-nf was recently updated to define test input data only in config and not in the main script

However, the azure e2e test currently runs rnaseq-nf with a full config override (`-C`) instead of a config append (`-c`).  As a result, the e2e fails because no input data is specified

This PR updates the azure e2e test to use `-c` instead of `-C` for rnaseq-nf so that the default project config is still used


jorgee (approved):

jorgee (commented):
I approved by mistake, Tests still failing there is still a missing -C with the rnaseq-nf run

validation/azure.sh
jorgee:
```suggestion
$NXF_CMD -c ./azure.config \
```

jorgee (approved):
Fixed now. You can merge

PR #7119, opened 2026-05-08
docs: Migrate docs to new build

christopher-hakkaart:
- Squashed from dev branch to help with noise

PR #7144, opened 2026-05-14
Honor COLUMNS when determining ANSI output width

### LINKED ISSUE #6274: COLUMNS environment variable not honored
## Bug report 

See discussion at the forum: https://community.seqera.io/t/output-width/2170

ANSI output do not honor the `COLUMNS` variable set by bash. That causes issues in the expected width when the user want to change it for whatever reason, either to make it smaller than the actual available space or just to set a desired width in non-interactive executions.

Run any pipeline with long process names under SLURM and look at the standard output.
Alternatively, try to change the width when running the process:
`COLUMNS=50 nextflow run ...`

### Program output 

The output will be 80 characters wide in the SLURM execution and whatever your terminal sizes in the interactive execution.
The expectation would be that using the `COLUMNS` environment variable, the width would be the one stored there.

### Environment 

* Nextflow version: 25.04.2
* Java version: 17.0.3.1 2022-04-22 LTS
* Operating system: GNU/Linux (3.10.0-1160.88.1.el7.x86_64)
* Bash version: 5.1.0(1)-release

Eldergenix:
Fixes #6274

## Summary

- Use `COLUMNS` as the ANSI log width fallback when `TERMINAL_WIDTH` is not set to a valid positive integer.
- Preserve the existing `TERMINAL_WIDTH` override and existing terminal-width fallback behavior for absent or invalid values.
- Add isolated coverage for valid, absent, malformed, zero, negative, and overflow env values.
- Document the `COLUMNS` behavior in the environment variable reference.

## Validation

- `JAVA_HOME=/opt/homebrew/opt/openjdk@21/libexec/openjdk.jdk/Contents/Home PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH ./gradlew :nextflow:test --tests nextflow.trace.AnsiLogObserverTest --no-daemon` - passed after code changes; build successful in 5m 25s.
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21/libexec/openjdk.jdk/Contents/Home PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH ./gradlew :nextflow:test --tests nextflow.trace.AnsiLogObserverTest --no-daemon` - passed after the docs edit; Gradle reported tasks up-to-date and build successful in 13s.
- `git diff --check` - passed.

## Notes

- Docs HTML build was not run locally because Sphinx/MyST are not installed in this Python environment (`ModuleNotFoundError: No module named 'sphinx'`).
- Full `make test` / full CI matrix was not run locally.
- Manual `COLUMNS=50 nextflow run ...` smoke was not run; the env-width fallback is covered by `AnsiLogObserverTest`.

### DIFF
--- a/docs/reference/env-vars.mdx
+++ b/docs/reference/env-vars.mdx
@@ -359,6 +359,16 @@ The Seqera Platform workspace ID. Can also be configured using the `tower.worksp
 
 Defines a proxy server used as a fallback for any protocol (HTTP, HTTPS, FTP) that does not have a scheme-specific proxy variable set. A scheme-specific variable (e.g. `HTTPS_PROXY`) takes precedence over `ALL_PROXY`. Proxy authentication is supported by providing the credentials in the proxy URL.
 
+##### `COLUMNS`
+
+<AddedInVersion version="26.10" />
+
+Sets the terminal width of ANSI-formatted log output when `TERMINAL_WIDTH` is not set to a positive integer.
+
+:::note
+Bash sets `COLUMNS` in interactive shells but does not export it, so it is not visible to Nextflow unless you export it explicitly, e.g. `export COLUMNS=100`, or set it for the command, e.g. `COLUMNS=100 nextflow run <pipeline>`. In non-interactive contexts such as a batch job script, `COLUMNS` is not set at all — use `TERMINAL_WIDTH` to force a width there.
+:::
+
 ##### `FTP_PROXY`
 
 Defines the FTP proxy server. Proxy authentication is supported by providing the credentials in the proxy URL, e.g. `ftp://user:password@proxy-host.com:port`. Credentials containing special characters must be URL-encoded (percent-encoded).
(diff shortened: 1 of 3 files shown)


bentsherman:
Thanks for the PR @Eldergenix — the fix is in the right place and behaves correctly. I verified it end to end rather than just from the unit tests: with the change, `COLUMNS=50 nextflow run ... -ansi-log true` renders the narrow layout, `TERMINAL_WIDTH=200` still wins over `COLUMNS=50`, and `TERMINAL_WIDTH=-1` correctly falls back to `COLUMNS`. Reverting only `AnsiLogObserver.groovy` reproduces the original 80-column output, so this does fix #6274. I also confirmed `AnsiLogObserver` is the only remaining consumer of the jline terminal width — `Launcher` already reads `COLUMNS` for the CLI help width — so the two are now consistent.

I've rebased the branch onto current `master` and pushed a small follow-up commit (previous head was `5402d3b7` if you need it).

The rebase was not a clean one, which is worth flagging: `docs/reference/env-vars.md` was **deleted** on `master` by the docs migration in #7119, which moved the reference docs to MDX. Your docs hunk no longer applied at all, so I ported the `COLUMNS` entry to `docs/reference/env-vars.mdx`.

The follow-up commit also addresses a few review comments:

- Moved the `COLUMNS` entry to its alphabetical position in the "Other settings" section (it had landed after `HTTPS_PROXY`).
- Added `<AddedInVersion version="26.10" />`, per the convention in the new docs build.
- Added a note that bash sets `COLUMNS` in interactive shells but does **not** export it, so it isn't visible to Nextflow unless exported explicitly or set per-command. This matters for the SLURM case in #6274: in a batch script `COLUMNS` isn't set at all, so `TERMINAL_WIDTH` is still the answer there. Without that note I think people will read this as fixing the non-interactive case outright.
- Reworded the `TERMINAL_WIDTH` entry, which still said "when set to an integer value" and now contradicted the new positive-only behavior. It now documents the `TERMINAL_WIDTH` → `COLUMNS` → auto-detection chain.
- Realigned the `where:` block columns and added a `TERMINAL_WIDTH`-only row, so the pre-existing path is covered too.

One thing worth calling out in the PR description: rejecting non-positive values also changes behavior for `TERMINAL_WIDTH` on its own — `TERMINAL_WIDTH=-5` previously passed straight through and set a negative width, which produced nonsense output rather than falling back. That's a genuine improvement, but it isn't obvious from the title, so it should be visible to anyone reading the changelog.

Nice catch on covering the integer-overflow case in the tests. LGTM once CI is green on the rebase.

bentsherman (approved):

PR #7150, opened 2026-05-16
ADR: Strict syntax parser

bentsherman:
This PR adds an ADR for the strict syntax parser, back-dated to the release date of Nextflow 26.04 when the `nf-lang` module was added

It describes the key design goals and decisions that we made, as well as some discussion of alternatives

### DIFF
--- a/adr/20250508-strict-syntax-parser.md
+++ b/adr/20250508-strict-syntax-parser.md
@@ -1,0 +1,315 @@
+# Strict syntax parser
+
+- Authors: Ben Sherman
+- Status: accepted
+- Deciders: Ben Sherman, Paolo Di Tommaso
+- Date: 2025-05-08
+- Tags: lang, parser
+
+## Summary
+
+Implement a formal grammar and parser for Nextflow scripts and config files.
+
+## Problem Statement
+
+Nextflow was originally developed as a Groovy DSL. Nextflow scripts and config files are parsed and executed as Groovy scripts with AST transforms for Nextflow-specific concepts such as process definitions, include statements, and config blocks.
+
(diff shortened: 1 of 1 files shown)


adr/20250508-strict-syntax-parser.md
```diff
@@ -0,0 +1,315 @@
...
+With a Nextflow SDK, this example might be written as follows:
+
+```groovy
+import nextflow.Nf
+
+@Nf.Process(
+    cpus = 2,
+    memory = '4 GB'
+)
+def echo(String greeting) {
```
jorgee: should it be `sayHello` instead of `echo` (workflow call is s`ayHello`)

adr/20250508-strict-syntax-parser.md
```diff
@@ -0,0 +1,315 @@
...
+}
+```
+
+A Nextflow SDK has many potential advantages:
+
+- SDKs have a much more "normal" developer experience than DSLs -- errors can still happen, but they are the same "kinds" of errors encountered with any other library.
+
+- SDKs are just libraries in an existing language, so they can be used with existing developer tools for that language.
+
+- SDks can be exposed to multiple programming languages via language bindings, so Nextflow could be used directly in the user's preferred language (e.g. Python).
```
jorgee:
```suggestion
- SDKs can be exposed to multiple programming languages via language bindings, so Nextflow could be used directly in the user's preferred language (e.g. Python).
```

jorgee (approved):
Looks fine just small changes to do. One typo and I think an example has a different name in the definition than the workflow execution. Look at the comments.

PR #7151, opened 2026-05-16
ADR: TaskReadinessGate plugin extension point

robsyme:
## Summary

Adds an architecture decision record proposing a new `TaskReadinessGate` plugin extension point that defers task submission until external preconditions are met (e.g. restoring S3 objects from Glacier before an AWS Batch job tries to stage them).

The motivation is that plugins needing this capability today must subclass an executor and its task handler purely to override `TaskHandler.isReady()`, which:
- forces users to opt in via `process.executor = '<plugin-specific-name>'` instead of a drop-in `plugins { id '...' }`;
- couples plugins to executor internals (e.g. `@CompileStatic` is incompatible with `AwsBatchTaskHandler` subclassing due to its proxy dispatch);
- requires a new executor subclass per (executor × cold-storage backend) combination.

The proposed SPI is consulted by `TaskPollingMonitor` before submitting any task, works uniformly across every executor, and is a small additive change to core (one interface + a handful of lines at the call site). Behavior is bit-identical when no plugin registers a gate.

The ADR documents the contract (polling-based, must return promptly, exceptions signal permanent failure), the call-site integration, considered alternatives (channel operator, S3 NIO interception — both rejected with reasoning), explicit non-goals, and follow-ups left for later (`AbstractAsyncReadinessGate` helper, per-process scoping, gate ordering).

Status in the ADR is `draft` — opening for review and discussion before implementation.

## Test plan

- [ ] Reviewers (Paolo, Ben) read the ADR and confirm the SPI shape, contract, and call-site integration.
- [ ] Confirm the non-goals (no helper class, no scoping, no ordering) are acceptable for v1.
- [ ] Confirm the placement under `nextflow.processor` rather than `nf-commons`.
- [ ] Once accepted, follow-up PR implements the interface, the three edits to `TaskPollingMonitor`, Spock specs, and developer-docs section.

### DIFF
--- a/adr/20260516-task-readiness-gate.md
+++ b/adr/20260516-task-readiness-gate.md
@@ -1,14 +1,14 @@
 # `TaskReadinessGate` plugin extension point for deferred task submission
 
 - Authors: Rob Syme
-- Status: draft
+- Status: accepted
 - Deciders: Paolo Di Tommaso, Ben Sherman, Rob Syme
-- Date: 2026-05-16
+- Date: 2026-05-20
 - Tags: plugin, extension-point, scheduler, task-submission
 
 ## Summary
 
-Introduce a `TaskReadinessGate` plugin extension point that lets a plugin defer task submission until an external precondition is met (e.g. an S3 object has been restored from Glacier), removing the need for plugins to subclass an executor and its task handler.
+Introduce a `TaskReadinessGate` plugin extension point that lets a plugin defer task submission until an external precondition is met (e.g. an S3 object has been restored from Glacier), removing the need for plugins to subclass an executor and its task handler. Plugins implement a blocking `void prepare(TaskHandler)` method; core runs each gate on a managed virtual-thread executor via a new `TaskGateManager`, and the standard `TaskPollingMonitor` delegates to the manager from `schedule`, `canSubmit`, and `evict`.
 
 ## Problem Statement
 
@@ -30,19 +30,21 @@ A generic extension point — "consult these plugins before submitting any task"
 
 - Work uniformly across every executor that submits tasks via `TaskPollingMonitor` (i.e. all existing executors).
 
-- Be safe-by-default: a plugin that ships a gate cannot accidentally stall the scheduler thread for the rest of the run.
+- Keep the plugin-author surface minimal — gate authors write straightforward blocking code, not an async state machine.
 
-- Allow gates to kick off async preparation work the moment a task is queued, not when an executor slot frees up — cold-storage restores can take hours and must not wait for capacity.
+- Allow gates to kick off preparation work the moment a task is queued, not when an executor slot frees up — cold-storage restores can take hours and must not wait for capacity.
 
-- Allow plugins to signal permanent task failure cleanly, integrating with the standard `errorStrategy` machinery.
+- Allow plugins to signal permanent task failure cleanly, integrating with the standard `errorStrategy` machinery (including retry routing through `ProcessRetryableException`).
 
 ## Non-goals
 
-- **Per-process gate scoping.** A gate is consulted for every task. Selectors that restrict a gate to specific processes are a config-layer concern that can be added later.
+- **Per-process gate scoping as a new directive.** A gate is consulted for every task. Plugins implement per-process opt-out by reading the existing `hints` directive (e.g. `hints 'glacier/skip': true`).
 
-- **Gate ordering or priority.** Evaluation order across gates is unspecified. An `int order()` default method can be added non-breakingly if a real use case appears.
+- **An executor-level timeout.** Core does not enforce a wall-clock deadline on `prepare`. The right value depends entirely on which plugin is registered and on the workload (Glacier Standard is hours; Deep Archive Bulk is days). Plugins own timeout policy, typically reading a per-process hint (e.g. `hints 'glacier/maxWait': '5h'`) and throwing when the deadline elapses.
 
-- **An `AbstractAsyncReadinessGate` helper class.** A base class that runs blocking work on a dedicated executor and exposes results via per-task `CompletableFuture`s would lower the cliff for plugin authors. We are deliberately not shipping it in v1: the contract is documented in javadoc, and consumers with their own cross-task state (dedup, rate limiting, batched API calls) typically would not use a per-task-future helper. It can be extracted later when a clear use case appears.
+- **Gate ordering or priority.** Evaluation order across gates is unspecified; all gates run in parallel.
+
+- **An `AbstractAsyncReadinessGate` helper class.** Since the SPI is blocking and core owns the executor, there is no async runtime for a helper to wrap. Plugin authors write blocking code directly.
 
 - **Replacing `TaskHandler.isReady()`.** The existing method stays and is still consulted; gates are AND-combined with it. No flag day for existing subclasses.
 
@@ -51,6 +53,13 @@ A generic extension point — "consult these plugins before submitting any task"
 - **Option A — `TaskReadinessGate` SPI in core (this proposal).**
 - **Option B — Channel operator `glacierRestore()`.**
 - **Option C — `FileSystemProvider` interception in nf-amazon's S3 NIO.**
+
+Within Option A, there was a sub-decision between two contract shapes:
+
+- **A.1 polling** — `boolean isReady(TaskRun)`, called repeatedly on the scheduler thread; plugin manages its own async state.
+- **A.2 blocking** — `void prepare(TaskHandler) throws InterruptedException`, called once per task on a core-managed virtual-thread executor; plugin writes straightforward blocking code.
+
+The blocking variant (A.2) was chosen — see "Why blocking, not polling" below.
 
 ## Pros and Cons of the Options
 
@@ -61,7 +70,7 @@ A new extension point consulted by `TaskPollingMonitor` before admitting a task 
 - Good, because it matches the actual problem ("defer task submission until external state is ready"), which is generic across executors and backends.
 - Good, because users get drop-in behavior — `plugins { id '...' }` is enough; no `process.executor` override.
 - Good, because the plugin stops being tied to a specific executor implementation and works with every executor in a single class.
-- Good, because the change to core is small (one interface plus a few lines in `TaskPollingMonitor`) and behavior is bit-identical when no plugin registers a gate.
+- Good, because the change to core is small (one interface, one manager class, ~14 lines in `TaskPollingMonitor`) and behavior is bit-identical when no plugin registers a gate.
 - Bad, because it requires a coordinated upstream change in Nextflow before the consumer plugin can be refactored.
 
 ### Option B — Channel operator `glacierRestore()`
@@ -84,7 +93,7 @@ Rejected. Documented here to avoid revisiting.
 
 ## Decision
 
-Adopt Option A. Introduce a `TaskReadinessGate` interface in `nextflow.processor` and consult registered implementations from `TaskPollingMonitor`. Option B may be pursued separately by individual plugins as a complementary surface.
+Adopt Option A with the blocking-`prepare` contract (A.2). Introduce a `TaskReadinessGate` interface in `nextflow.processor` and a `TaskGateManager` that orchestrates registered implementations on behalf of `TaskPollingMonitor`. Option B may be pursued separately by individual plugins as a complementary surface.
 
 ## Core capabilities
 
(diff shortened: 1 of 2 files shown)


pditommaso:
## Proposal: collapse the SPI to a single blocking `prepare()`, run on a core-managed executor

Re-read the ADR a few times. The cleanest factoring I can see eliminates the polling contract entirely and lets the plugin author write the obvious blocking code. Core absorbs the async management, plugin authors don't see `CompletableFuture`, the scheduler thread is never blocked, and the SPI shrinks to one method.

### The interface

```groovy
package nextflow.processor

interface TaskReadinessGate extends ExtensionPoint {

    /**
     * Blocking preparation. Invoked once per task on a managed background thread
     * (virtual threads by default). The task is admitted for submission when this
     * method returns; throw to mark it as permanently failed.
     *
     * Implementations may block freely — Thread.sleep, network calls, long polling.
     * The scheduler thread is never blocked by this call.
     *
     * Implementations must honor Thread.interrupt() so that task eviction or
     * workflow abort can unblock the prepare call promptly.
     */
    void prepare(TaskHandler handler) throws InterruptedException
}
```

That's the whole SPI. No `isReady()`. No fire-once-on-enqueue/poll-on-`canSubmit` split. No `CompletableFuture` in the plugin's vocabulary. "Done" = method returned. "Failed" = method threw.

### Core integration in `TaskPollingMonitor`

A shared executor — one pool governs total prepare-time concurrency across all gates. Easier to split into per-gate pools later than to merge.

```groovy
private List<TaskReadinessGate> readinessGates = Collections.emptyList()
private ExecutorService gateExecutor

private final ConcurrentMap<TaskHandler, List<Future<?>>> gateFutures = new ConcurrentHashMap<>()

@Override
TaskMonitor start() {
    readinessGates = Plugins.getExtensions(TaskReadinessGate)
    gateExecutor = Threads.useVirtual()
        ? Executors.newVirtualThreadPerTaskExecutor()
        : Executors.newCachedThreadPool(new CustomThreadFactory('ReadinessGate'))
    session.onShutdown { gateExecutor.shutdownNow() }
    // ... existing start logic
    return this
}

@Override
void schedule(TaskHandler handler) {
    pendingLock.lock()
    try {
        pendingQueue << handler
        if( readinessGates ) {
            final futures = new ArrayList<Future<?>>(readinessGates.size())
            for( gate in readinessGates ) {
                futures << gateExecutor.submit { gate.prepare(handler) }
            }
            gateFutures.put(handler, futures)
        }
        taskAvail.signal()
        notifyTaskPending(handler)
    }
    finally {
        pendingLock.unlock()
    }
}

protected boolean canSubmit(TaskHandler handler) {
    // Gates evaluated before maxForks/capacity short-circuit, so a task waiting
    // behind a full forks limit still has its prepare() running in the background.
    allGatesReady(handler) \
        && handler.isReady() \
        && handler.canForkProcess() \
        && (capacity > 0 ? checkQueueCapacity(handler) : true)
}

private boolean allGatesReady(TaskHandler handler) {
    final futures = gateFutures.get(handler)
    if( !futures ) return true
    for( f in futures ) {
        if( !f.isDone() ) return false
        if( f.isCancelled() ) { handleGateFailure(handler, new CancellationException()); return false }
        try { f.get() }
        catch( ExecutionException e ) { handleGateFailure(handler, e.cause); return false }
        catch( InterruptedException e ) { Thread.currentThread().interrupt(); return false }
    }
    gateFutures.remove(handler)
    return true
}

@Override
boolean evict(TaskHandler handler) {
    gateFutures.remove(handler)?.each { it.cancel(true) }
    // ... existing remove + slotAvail.signal() logic
}
```

The scheduler thread's only gate-related work per tick is `f.isDone()` per gate per pending task — a volatile read on the `CompletableFuture` state machine, nanoseconds. All blocking work lives on the gate executor. The future cache is bounded by `pendingQueue.size() + runningQueue.size()` and is cleaned up on admission or eviction.

Two notable differences from the ADR's call-site design:

1. **No fire-once-on-`schedule` + poll-on-`canSubmit` split.** `schedule()` submits once to the executor; `canSubmit()` polls the future. The two-call-site contortion in the current ADR was working around a constraint (the polling contract) that the new design doesn't have.
2. **Reorder `canSubmit` so readiness evaluates before `canForkProcess()`/`checkQueueCapacity()`.** Doesn't affect the new design's correctness — the preparation is already running asynchronously — but keeps the AND chain semantically ordered (cheapest fixed-shape checks first, then capacity gates).

### What a Glacier gate looks like

```groovy
class GlacierReadinessGate implements TaskReadinessGate {

    @Override
    void prepare(TaskHandler handler) throws InterruptedException {
        for( S3Path p : extractS3Inputs(handler.task) ) {
            issueRestoreIfNeeded(p)              // RestoreObject — idempotent, ~10ms
            while( !isRestored(p) ) {            // HeadObject poll
                if( isPermanentlyFailed(p) ) {
                    throw new ProcessException("Glacier restore failed for $p")
                }
                Thread.sleep(60_000)             // virtual thread parks, ~zero cost
            }
        }
    }
}
```

That's the entire plugin. No executor wiring, no future cache, no idempotency dance, no "must return promptly" reasoning. Cross-task dedup, if wanted, stays a plugin-internal optimization (`ConcurrentMap<S3Key, CompletableFuture<Void>>` inside `issueRestoreIfNeeded`).

### Why blocking, not `CompletableFuture<Void> prepare()`

The future-returning variant is more flexible — plugins with native async SDKs could return their SDK's future directly, skipping a parked thread. We considered it and chose against:

- Plugin authors are domain experts (bioinformatics, cloud infra), not async-Java experts. `void` + `throws` is unambiguous; `CompletableFuture` chains, cancellation propagation, and `CompletionException` unwrapping are not.
- Virtual threads make the "parked blocking thread" cost negligible (~1 KB/task). The async-SDK win is real but small.
- `FilePorter` is already the in-tree precedent: `FileTransfer` is a blocking `Runnable` submitted to a managed pool. New extension points should follow.
- Uniform threading model = uniform observability. One named pool, predictable thread dumps. Per-plugin executors fragment that.

The future variant is a non-breaking future extension if a real need emerges (e.g. an overload `default CompletableFuture<Void> prepareAsync(TaskHandler) { ... wrap prepare in core executor ... }`).

### Why `TaskHandler`, not `TaskRun`

Every other method on the scheduler boundary takes `TaskHandler` (`schedule`, `canSubmit`, `submit`, `evict`). Making this one method take `TaskRun` is a small but unnecessary asymmetry. The base class is portable (lives in `nextflow.processor`, not in any executor module), and `handler.status` / `submitTimeMillis` are genuinely useful for gates that want to self-impose a "preparing too long, give up" timeout.

The cost is wider surface area — `submit()`, `kill()`, `setStatus()` are reachable from the handle. The javadoc needs an explicit "read-only; do not invoke lifecycle methods" line. Worth it for the uniformity.

### What we consciously trade away

These are real and worth naming:

- **Plugin authors with native async SDKs pay a small wrap cost** (~1 KB per parked virtual thread).
- **Pure synchronous predicates pay one executor submission** (microseconds).
- **Cross-task dedup is plugin-internal** — same as it would be in the future-returning variant, just at a slightly different layer.

None of these are blocking issues; they're the price of a smaller SPI.

### Safety net worth shipping with v1

To avoid silent hangs from gates that never complete:

```groovy
// in allGatesReady, before the futures loop:
final maxWait = config.getGateMaxWait()   // default: e.g. 24h, configurable
if( maxWait && System.currentTimeMillis() - handler.scheduleTimeMillis > maxWait.millis ) {
    futures.each { it.cancel(true) }
    handleGateFailure(handler, new ProcessException("Task readiness gate timed out after $maxWait"))
    return false
}
```

Surfaces stuck preparations as task failures, where `errorStrategy 'retry'` can pick them up naturally.

### Backward compatibility

Identical to the ADR's claim: with no gates registered, `canSubmit` evaluates one empty-list check per pending task per tick; `schedule` adds one untaken branch. `TaskHandler.isReady()` is unchanged. The SPI is purely additive.

### Testing

Same matrix as the ADR proposes, simpler to express:

- No gates registered → `canSubmit` matches current behavior.
- One gate that returns from `prepare()` immediately → task admitted as soon as capacity/forks allow.
- One gate that sleeps then returns → task waits in pending across ticks; admitted on the tick after `prepare` returns.
- Gate throws → task fails with the thrown cause; routed through `errorStrategy`.
- Multiple gates → all-must-complete semantics; one failure aborts the task.
- Eviction during prepare → future cancelled, gate's `InterruptedException` propagates as cancellation.
- Gate exceeds `gateMaxWait` → task fails with timeout cause.

Gates injected through a test double instead of a real PF4J context, same as the ADR proposes.

### Open question: per-process enable/disable?

The ADR's non-goals already list "per-process gate scoping" as out of scope, but with the runtime cost moved into core (executor submission + future tracking per task per gate), it's worth revisiting whether pipelines should be able to opt out at the process level. Something like:

```nextflow
process FAST_PATH {
    readinessGate false   // skip all registered gates for this process
    // ...
}

process GLACIER_DEPENDENT {
    readinessGate 'glacier'   // enable only the named gate
    // ...
}
```

The case **for** adding it:

- A pipeline with 50 processes, only 3 of which read Glacier-stored inputs, pays one executor submission per task for the other 47 even when no gate would do meaningful work. Cheap (microseconds) but non-zero.
- It gives users an escape hatch without uninstalling the plugin — useful when a gate has a bug or when one specific process needs to bypass it for a hotfix.
- Composes with the existing `withName` / `withLabel` selector machinery, so the config-layer cost is small.

The case **against**:

- The plugin itself can short-circuit cheaply (`if (extractS3Inputs(task).isEmpty()) return`). For most gates this makes the per-process toggle redundant.
- Adds a public directive whose semantics across multiple gates need design (`readinessGate false` vs `readinessGate ['glacier']` vs `readinessGate.glacier false` — there are several reasonable shapes).
- Better deferred until a real "I want this gate off for this process" report appears.

I lean toward **deferring this to a follow-up**, matching the ADR's existing posture, but flagging it explicitly here because moving the runtime cost into core changes the cost-benefit slightly. Worth a sentence in the ADR's non-goals section confirming the position either way.

---

### TL;DR vs the current ADR

| | Current ADR | This proposal |
|---|---|---|
| Methods on SPI | `boolean isReady(TaskRun)` | `void prepare(TaskHandler)` |
| Plugin must be non-blocking | Yes (contract) | No (runs on managed executor) |
| Call sites in `TaskPollingMonitor` | Two (`schedule` fire-once + `canSubmit` poll) | One (`schedule` submit + `canSubmit` future-poll) |
| Plugin-author cognitive load | "Must return promptly; first call kicks off, subsequent calls report; must be idempotent" | "Write blocking code; throw if hopeless; honor interrupts" |
| Async runtime ownership | Plugin's problem | Core's problem |
| Helper class needed | Yes (deferred to "later") | No |
| Foot-guns | Multiple (idempotency, blocking, polling cost) | One (gate must honor interrupts) |

Net: smaller SPI, fewer call sites, no foot-guns about scheduler-thread blocking, and the resulting plugin code is ~5 lines for the Glacier case instead of ~40. Worth the bounded extra work in core (the gate executor + future cache, ~50 lines).

Happy to draft the actual diff against `TaskPollingMonitor.groovy` if there's interest in moving this direction.

robsyme:
Yeah, I like this suggestion. simpler interface is much nicer. Complexity in core and much smaller effort for plugin authors. @pditommaso - when we come to the actual plugin, are you ok with the a plugin-specific process directive as suggested in the comment?

```
process FAST_PATH {
    readinessGate false   // skip all registered gates for this process
    // ...
}
```

pditommaso:
Think this should go through via the new `hints` directive. Wdyt?

bentsherman:
FYI, I updated the implementation PR to merge into this branch

Please try to keep the ADR up to date with the latest decisions, as it looks like the impl has diverged a good bit

robsyme:
Updated the ADR to match the design that landed in #7158:

- SPI is blocking `void prepare(TaskHandler)`, not polling `boolean isReady(TaskRun)`.
- Orchestration lives in a new `TaskGateManager` — `TaskPollingMonitor` collapses to one field + four delegation lines.
- No `executor.gateMaxWait` config option — plugins own timeout policy via the existing `hints` directive (`glacier/maxWait`, `glacier/skip`).
- Cause-preservation in the exception unwrap so `ProcessRetryableException` markers reach `resumeOrDie` via `error.cause`.
- `AbstractAsyncReadinessGate` helper is now a permanent non-goal (no async runtime for it to wrap).

Status moved from draft to accepted.

pditommaso:
Worth moving this fwd, is there anything blocking?

robsyme:
In the current form, works well for cloud, but I did some more thorough testing and discovered it doesn't work for staging files into work/stage. I'll add some more details today and try to find a solution that works for all possible glacier inputs - I'd hate to contribute to a fragmentation of features across run types.

hilljessica8787-dot (approved):

hilljessica8787-dot (approved):

robsyme:
Promised details on the staging problem:

What our first approach doesn't cover: The gate assumes the first read of a remote input happens after task scheduling. This is true on AWS Batch with an S3 work dir (worker runs nxf_s3_download at task start) and under Fusion (lazy FUSE reads). It's false whenever the input's scheme differs from the work dir's scheme. Minimal repro:
```nextflow
// workDir = /scratch (local or NFS), executor = local, slurm, pbs, ...
process ALIGN {
    input:
    path bam   // resolved from s3://bucket/sample.bam, storage class GLACIER
    ...
}
```
Because s3 != file, Executor.isForeignFile() classifies the input as foreign, and TaskProcessor.invokeTask() stages it eagerly on the head node. I'd rather not us ship a fix that only covers the obvious use-cases (cloud executor with cloud workdir) but falls over for those on HPC but still interact with S3.

**Proposed fix.** A second, path-keyed extension point alongside the task-keyed one:

```nextflow
package nextflow.file

interface FileReadinessGate extends ExtensionPoint {
    void prepare(Collection<Path> paths) throws InterruptedException
}
```

Called synchronously at the top of FilePorter.transfer(Batch) with the batch's source paths before any download is submitted, and again before TaskHasher.compute() when the hash mode reads content. Same contract language as TaskReadinessGate: block freely, honor interrupts, throw for permanent failure (wrapped as ProcessStageException so errorStrategy applies), plugins own timeout policy.

**Deferred:** operator-level reads (splitCsv on a Glacier-stored samplesheet, file(...).text, etc.). Those never involve a task and would need per-operator or filesystem-provider interception - the ADR keeps them out of scope but shapes the interface so those call sites can be added without changes. To be clear, this doesn't reopen the Option C rejection in the existing ADR: that was about gating Batch submission, which a head-node hook can't see; head-node reads are the one thing it can.

PR #7160, opened 2026-05-20
Add Nextflow module info to lineage TaskRun entry

jorgee:
## Summary

- Add a new `module` field to the lineage `TaskRun` model (`modules/nf-lineage/.../model/v1beta1/TaskRun.groovy`) that records the remote Nextflow module defining the process executed by a task, encoded as `name@version` (e.g. `nf-core/fastqc@1.0.0`).
- `LinObserver.getTaskModule` walks `task.processor.getOwnerScript()` → `ScriptMeta` to locate the script and `LinObserver.extractModuleInfo` reads the module's `meta.yml` next to the `.module-info` marker (the latter identifies the directory as a Nextflow-managed remote module rather than any directory that happens to contain a `meta.yml`).
- Manifest parsing is wrapped in `try/catch` and logged as a warning — a corrupt `meta.yml` will not fail the task complete handler. The field is `null` for local includes and for processes defined in the main script.
- Field is appended at the end of `TaskRun`'s `@Canonical` constructor, preserving backward compatibility with existing positional constructor callers.

## Test plan

- [x] `./gradlew :nf-lineage:test` — full nf-lineage suite green
- [x] New unit tests cover the happy path (`meta.yml` + `.module-info` → `name@version`), no owner script, script not marked as module, script marked as module but no manifest (local include), and malformed `meta.yml` (warn + null)
- [x] Manual: run a pipeline that uses a remote module (e.g. `nf-core` module) with lineage enabled and confirm the `module` field is populated in the persisted `TaskRun` entry
- [x] Manual: run a pipeline with only local includes / main script and confirm the `module` field is `null`


pditommaso (changes requested):
## Code review

Solid feature overall — happy/edge paths are covered and the `.module-info` + `meta.yml` AND-check correctly disambiguates remote modules from local includes (since `setModule(true)` is called by `IncludeDef`/`ScriptLoaderV1`/`ScriptLoaderV2` for *any* include). Backward compatibility of the `TaskRun` positional constructor is preserved via Groovy's `@Canonical` `defaults=true` semantics.

Requesting changes for the following:

### Important

1. **Risk of persisting `"null@null"`** — `LinObserver.groovy:315`
   ```groovy
   return "${spec.name}@${spec.version}"
   ```
   `ModuleSpecFactory.fromYaml` does `spec.name = data.name as String` (and same for `version`). A valid-YAML manifest that happens to be missing those keys silently persists the GString `"null@null"` into the lineage record. Suggest:
   ```groovy
   if( !spec.name || !spec.version ) {
       log.warn "Incomplete module manifest at '${manifestPath.toUriString()}'"
       return null
   }
   return "${spec.name}@${spec.version}".toString()
   ```
   The explicit `.toString()` is safer for downstream serializers (Kryo) than letting a `GString` flow into a `String` field.

2. **Per-task overhead with no caching** — `storeTaskRun` runs on every `onTaskComplete`. Each call does 2 `Files.exists` + a YAML parse on potentially remote FS. For workflows with thousands of tasks reusing the same module, this is O(tasks) instead of O(modules). Modules are immutable for the duration of a run, so caching is safe. Easiest fix: annotate `extractModuleInfo(Path scriptPath)` with `@groovy.transform.Memoized` — keyed on the script path, this turns it into a one-shot lookup per module.

3. **`catch( Throwable e )` is too broad** — `LinObserver.groovy:317`. Catches `OutOfMemoryError`, `StackOverflowError`, `ThreadDeath`, etc. Use `Exception` (or `IOException | RuntimeException` if you want explicitness).

### Minor

4. **Test coverage gap on serialization** — All 5 new tests exercise `getTaskModule(task)` in isolation with mocks; they do not cover `storeTaskRun` end-to-end nor the new field's encode/decode roundtrip. Consider extending `LinEncoderTest`'s `should encode and decode TaskRun` to assert the `module` field survives the roundtrip.

5. **Malformed-YAML test doesn't assert the warning log** — Acceptable, but a regression that silently drops the warning would go unnoticed. A `MemoryAppender` assertion would lock it in.

6. **Field naming** — `module` is fine but slightly ambiguous given the rich `nextflow.module.*` package. `moduleRef` / `moduleId` would be more self-documenting and align with `workflowRun` (a reference id) rather than the noun "module". Not blocking.

7. **Style nit** — `LinObserver.groovy:301`: missing space before `{` — `extractModuleInfo(Path scriptPath){`.

8. **Uncovered edge cases** — symlinked module paths and subworkflow includes work with the current implementation (verified by reading `loadModuleV2` + `ScriptMeta`), but a brief integration-style test would lock that behavior in.

### Verification done

- Read `LinObserver.groovy`, `TaskRun.groovy`, `IncludeDef.groovy`, `ScriptMeta.groovy`, `ModuleSpecFactory.groovy`.
- `grep -rn 'setModule\s*('` — confirmed local + remote includes both set `module=true`.
- `grep -rn 'new TaskRun('` in test sources — confirmed pre-existing positional callers still compile under `defaults=true`.
- `./gradlew :nf-lineage:test` — green.

Overall: changes requested, non-blocking. Once 1–3 are addressed, this is ready to merge.

jorgee:
Thanks for the review — all points addressed in the latest pushes:

**Important**
1. **`"null@null"` risk** — `extractModuleInfo` now returns `null` (with a `log.debug`) when the manifest is missing `name`/`version`, and the success path returns `"${spec.name}@${spec.version}".toString()` so a `GString` never reaches the `String` field / Kryo.
2. **Per-task overhead** — `extractModuleInfo(Path)` is now `@Memoized`, so the `Files.exists` checks + YAML parse run once per module instead of once per task.
3. **`catch(Throwable)`** — narrowed to `catch(Exception)`.

**Minor**
4. **Serialization coverage** — extended `LinEncoderTest`'s TaskRun roundtrip to set and assert `workflowRun` + `moduleId`.
5. **Warning assertion** — added a logback `ListAppender` helper; the malformed-manifest test now asserts the `WARN` is emitted. (The incomplete-manifest case is `debug`-level only, so its test just asserts `null`.)
6. **Field naming** — renamed `module` → `moduleId` (and `getTaskModule` → `getTaskModuleId`) to align with the `workflowRun` reference convention.
7. **Style nit** — added the missing space before the method body brace.

pditommaso (approved):
Thanks @jorgee for addressing all the review points — the null-safety fix, memoization, and the extra test coverage look great. LGTM 👍

PR #7163, opened 2026-05-20
Fix `module run` to allow optional path inputs

### LINKED ISSUE #7161: "nextflow module run" with multiple optional path inputs ?
## Bug report 
Cannot find how to specify (multiple) optional `path` input with `nextflow module run`

I do not know exactly which is the "expected" behaviour here, but actual is impossibilty to run a module with multiple optional `path` inputs

```bash
nextflow-26.04.1-dist module run -with-singularity nf-core/deepvariant/rundeepvariant \
    --meta.id test \
    --input https://github.com/nf-core/test-datasets/raw/refs/heads/modules/data/genomics/homo_sapiens/illumina/bam/test.paired_end.sorted.bam \
    --index https://github.com/nf-core/test-datasets/raw/refs/heads/modules/data/genomics/homo_sapiens/illumina/bam/test.paired_end.sorted.bam.bai \
    --intervals "" \
    --meta2.id fasta \
    --fasta https://github.com/nf-core/test-datasets/raw/refs/heads/modules/data/genomics/homo_sapiens/genome/genome.fasta \
    --meta3.id fai \
    --fai https://github.com/nf-core/test-datasets/raw/refs/heads/modules/data/genomics/homo_sapiens/genome/genome.fasta.fai \
    --meta4.id gzi \
    --gzi "" \
    --meta5.id par \
    --par_bed "" \
    --outdir test_out
```

And `dv.config` containing
```bash
process {
  withName: DEEPVARIANT_RUNDEEPVARIANT {
    ext.args = "--model_type WGS"
  }
}
```

### Program output 

Multiple possible errors depending on what I tried :
* All optional inputs to <empty_string> or <empty_list> -> `input file name collision`
* Put different fake filenames (not existing) -> tool error when it tries to handle non-existing file
* Excluding optional inputs from command-line -> `Missing required parameter`

### Environment 

* Nextflow version: 26.04.1
* Java version: openjdk 23.0.2-internal 2025-01-21
* Operating system: Linux
* Bash version: GNU bash, version 5.2.37(1)-release (x86_64-pc-linux-gnu)

### Additional context

In above command, multiple `path` inputs are optional, these are for sure (based on [module code](https://github.com/nf-core/modules/blob/master/modules/nf-core/deepvariant/rundeepvariant/main.nf)) :
* `intervals` (from 1st `meta` tuple)
* `tuple val(meta4), path(gzi)`
* `tuple val(meta5), path(par_bed)`

Attached `.module.log` : [module.log.txt](https://github.com/user-attachments/files/28056117/module.log.txt)

bentsherman:
Close #7161 

This PR adds support for optional file inputs to `nextflow module run`.

When a `path` input is not supplied a value at runtime, the ProcessEntryHandler defaults to `[]` instead of throwing an error. This is needed to allow for optional file inputs which are a common use case.

For legacy processes, one consequence is that missing required files will not be reported until task execution. This is because legacy processes have no way to declare whether a `path` input is optional.

For typed processes, optional file inputs are declared as `Path?` and so can be validated prior to task execution. With this PR, a `null` param value is allowed as long as the input is declared as optional.


modules/nextflow/src/test/groovy/nextflow/script/ProcessEntryHandlerTest.groovy
```diff
@@ -353,4 +353,66 @@ class ProcessEntryHandlerTest extends Specification {
...
+        def pathParam = Mock(FileInParam) { getName() >> 'intervals' }
+
+        when: 'the input is not present in params for a path input'
+        def result = handler.getValueForInputV1(pathParam, [:], [:])
+
+        then: 'returns an empty list instead of throwing'
+        result == []
+    }
+
+    def 'should return empty list for missing optional input (v2)' () {
```
jorgee:
```suggestion
    def 'should return null for missing optional input (v2)' () {
```

modules/nextflow/src/main/groovy/nextflow/script/ProcessEntryHandler.groovy
```diff
@@ -292,6 +292,8 @@ class ProcessEntryHandler {
         final value = namedArgs.get(name)
 
         if( value == null ) {
+            if( param instanceof FileInParam )
+                return []
```
jorgee, line 299: Shouldn't we add a log for this conversion?

jorgee (approved):
Looks fine and fixes the issue case. Nothing is blocking from my side just minor comments.

jorgee:
One comment that come into my mind when reviewing the PR, but out of the scope of the PR. Can we have the same issue with input variables (val) in V1? Nothing in the documentation says that optional parameters are supported. I see a strange case, but not sure if it is done in nf-core modules or if optional parameters are just used for files?

pditommaso:
Took a closer look — overall this is a clean, surgical change. The downstream wiring is correct: I traced both the `[]` (v1) and `null` (v2) returns through `TaskProcessor.applyInputContextV2` (lines 1801–1884) and confirmed that `null` is skipped and re-validated against `!param.optional` with a clear error message. The `Path?` → `isOptional()=true` model is correctly leveraged.

One thing worth considering, plus a few nice-to-haves:

### Important to consider

**The original repro from #7161 (`--intervals ""`) is not actually fixed — only omission is.**

The new guard in `ProcessEntryHandler.getValueForInputV1` only handles `value == null`. The reporter's command used `--intervals ""` (empty string), which means `namedArgs.get('intervals') == ""` (non-null) → falls through to `parseFileInput("")` → `Nextflow.file("")` → still errors.

Users naturally write `--foo ""` on the CLI to signal "no value", so this gap will likely come back as a follow-up bug report. One-liner fix:

```groovy
if( value == null || (param instanceof FileInParam && value instanceof CharSequence && !value) ) {
    if( param instanceof FileInParam )
        return []
    throw new IllegalArgumentException("Missing required parameter: --${name}")
}
```

Or, document explicitly in the changelog / docs that the user must *omit* the flag rather than pass an empty string.

### Nice to have

1. **No log signal when v1 silently swallows a missing path input.** A typo like `--read` vs `--reads` will now surface as a far-downstream task-execution failure with no hint that the root cause was a misspelled flag. Even a `log.debug` (or throttled `log.warn1`) noting "path input `--${name}` not provided, defaulting to empty list" would help post-mortem debugging.

2. **No v1 negative test for non-`FileInParam` inputs.** The new test covers the early-return for `FileInParam`, but there's no explicit assertion that e.g. `ValueInParam` still throws. The pre-existing `should throw exception for missing required parameter` test catches the regression by accident, but doesn't pin the contract of the `instanceof FileInParam` guard. Strengthening it to assert the exception message would make the contract explicit.

3. **`ProcessInput` is mocked rather than constructed.** It's a simple value object with a 3-arg constructor (`name`, `type`, `optional`) — using a real instance instead of `Mock(ProcessInput) { ... }` would make the v2 tests more refactor-resistant.

4. **No changelog / docs update.** This changes the failure semantics for `nextflow module run` with missing path inputs — worth a line in `changelog.txt` and a sentence in the `nextflow module run` docs.

### Follow-up idea (not for this PR)

The PR notes that legacy processes have no way to declare optionality. The recently-merged `meta.yml` schema validation in #7094 makes `ModuleSpec.ModuleParam` a natural place to add an `optional: true` field — so `getValueForInputV1` could honor that metadata via `paramTypes` (extended with an optional flag) and reject truly-required missing inputs while permitting truly-optional ones. Worth a follow-up issue.

bentsherman:
> Can we have the same issue with input variables (val) in V1? Nothing in the documentation says that optional parameters are supported. I see a strange case, but not sure if it is done in nf-core modules or if optional parameters are just used for files?

In theory yes, `val` inputs can also be optional. But in practice they tend to have a default value (`0`, `false`, `''`, etc) which is easy to supply on the command line.

This fix is needed for `path` inputs because of how the path string is being converted to a file path under the hood, so users need a way to circumvent that. Omitting the option is the simplest solution and also consistent with how optional file params work at the pipeline level

PR #7167, opened 2026-05-21
ADR: lineage validate semantic equivalence check

edmundmiller:
## Summary

Adds an ADR for `nextflow lineage validate` — design doc that precedes the implementation PR (#TBD, stacked on this one).

The ADR pins the design before the surface stabilises:

- **Equivalence model**: outputs + key inputs (commitId + repository), checksum-only file comparison, flattened to terminal outputs
- **Baselines**: peer LID *and* snapshot file from day one; `tower://` resolvable via `nf-tower` plugin (SPI seam)
- **Architecture**: shared `LineageValidator` core used by both the CLI and `LineageSnapshotter` (no test-vs-CI drift)
- **Diff categories**: outputs / params / workflow-identity / status (fail) + resources / runtime (informational, `--strict-runtime` opt-in)
- **CI ergonomics**: auto-detect `CI`, `GITHUB_ACTIONS`, `AGENT`/`CLAUDECODE`; `--json` output; exit codes `0/1/2`
- **Causality in reports**: one-hop upstream context for diverging outputs
- **Failed runs**: compared, with status divergence surfaced

Explicit non-goals: not bit-for-bit content checking, not a reproducibility guarantee, not a full DAG engine, not Platform-aware by itself.

Also identifies five lineage-model gaps (`ValueOutput`, runtime fingerprint, `-outputDir` accessor, named aliases, container digest stability) as named follow-ups.

The implementation PR stacked on top of this exercises the v1 design so we can iterate on feedback in practice.

## Test plan

- [ ] Read `adr/20260521-lineage-validate.md` and verify the design matches expectations
- [ ] Identify any missing non-goals or model gaps before approval

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### [Stack](https://github.com/kitlangton/stack)

1. **#7167** 👈 current
2. #7168

PR #7168, opened 2026-05-21
Add nextflow lineage validate command (first pass)

edmundmiller:
## Summary

First-pass implementation of `nextflow lineage validate` per the design in #7167 (stacked on that PR). Ships the v1 surface so we can use it in practice and feed back into the design.

What's in:

- **`LinNormalizer`** — encodes lineage records, strips ephemeral fields, supports `--output-base` relativization and `--ignore-fields`; provides recursive structural compare with diff paths
- **`nextflow lineage validate <lid> --against <lid>`** — CLI command with `-against`, `-ignore-fields`, `-output-base` options; emits JGit-style unified diff and exits non-zero on divergence
- **`LineageSnapshotter`** — Spock helper for pipeline tests; supports save-or-compare with `UPDATE_SNAPSHOTS` env var, plus direct `assertEquivalent(lidA, lidB)`
- **Integration tests** — seven scenarios covering equivalent runs, output checksum drift, parameter drift, output count mismatch, nested directories, `--ignore-fields`, and diff rendering
- **Regression test for JCommander wiring** — catches the option-parsing bug fixed in the last commit

Known gaps vs the ADR (deferred to follow-ups):

- No shared `LineageValidator` core yet; CLI and `LineageSnapshotter` still have parallel logic
- No snapshot-file baseline (`--against <file.json>`) or `--save-snapshot`
- No `--json` / `--summary` / CI auto-detection
- No diff categorisation, runtime-fingerprint handling, or one-hop causality enrichment
- No `LineageResolver` SPI; only `lid://` works today

Treat this as a working spike to validate the design against real lineage data — feedback expected to inform the v1.1 refactor.

## Test plan

- [ ] Run `./gradlew :nf-lineage:test --tests LinValidateIntegrationTest`
- [ ] Run `./gradlew :nextflow:test --tests LauncherTest` (regression for JCommander wiring)
- [ ] Manual: launch two pipeline runs against the local lineage store, `nextflow lineage validate lid://<runA> -against lid://<runB>`, confirm divergence on params/outputs and equivalence on identical runs
- [ ] Try `-ignore-fields commitId` to verify field exclusion

### [Stack](https://github.com/kitlangton/stack)

1. #7167
2. **#7168** 👈 current

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdLineage.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdLineage.groovy
@@ -46,6 +46,7 @@ class CmdLineage extends CmdBase implements UsageAware {
         void diff(ConfigMap config, List<String> args)
         void find(ConfigMap config, List<String> args)
         void check(ConfigMap config, List<String> args)
+        void validate(ConfigMap config, List<String> args)
     }
 
     interface SubCmd {
@@ -68,6 +69,7 @@ class CmdLineage extends CmdBase implements UsageAware {
         commands << new CmdDiff()
         commands << new CmdFind()
         commands << new CmdCheck()
+        commands << new CmdValidate()
     }
 
     @Parameter(hidden = true)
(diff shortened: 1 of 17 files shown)


bentsherman:
Squashing into the main PR since it's just a POC. Feel free to experiment on the same PR -- as long as there's an ADR at the top I don't care

PR #7172, opened 2026-05-22
Fix bucket cleanup in S3 test

bentsherman:
Fixes an S3 unit test that creates a bucket without deleting it on cleanup

### DIFF
--- a/plugins/nf-amazon/src/test/nextflow/cloud/aws/nio/AwsS3NioTest.groovy
+++ b/plugins/nf-amazon/src/test/nextflow/cloud/aws/nio/AwsS3NioTest.groovy
@@ -448,6 +448,8 @@ class AwsS3NioTest extends Specification implements AwsS3BaseSpec {
         then:
         thrown(UnsupportedOperationException)
 
+        cleanup:
+        deleteBucket(bucketName)
     }
 
     @Ignore // FIXME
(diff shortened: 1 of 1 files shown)


pditommaso (approved):

jorgee (approved):

jorgee:
I have fixed the test and checked that all buckets have been removed.

bentsherman:
Thanks Jorge

Interesting explanation from claudio about why `def` vs `final` matters here:

> Yes — it's a consequence of how Spock restructures the method bytecode to support `cleanup:`.
> 
> Spock's AST transformation wraps the test body in a `try/finally` so that `cleanup:` always runs, even when the test throws. To make a variable declared in `given:` reachable inside that `finally` block, Spock has to **hoist the declaration** to the top of the method (outside the try), separate from the assignment:
> 
> ```groovy
> // What Spock generates (roughly)
> def bucketName = null          // hoisted declaration
> try {
>     bucketName = createBucket() // given: assignment
>     Files.delete(...)           // when:
>     ...                         // then:
> } finally {
>     deleteBucket(bucketName)    // cleanup:
> }
> ```
> 
> That two-step pattern — declare-then-assign — is illegal for `final` variables. A `final` local must be assigned exactly once, at its declaration site. So when Spock tries to emit `final bucketName = null` at the top of the method and `bucketName = createBucket()` inside the try, the compiler rejects it with "might not have been initialized" (or a similar definite-assignment error).
> 
> With `def`, the same split is fine because the variable is just a nullable reference with no assignment-once constraint.
> 
> The reason this was invisible before your change: the original test had no `cleanup:` block, so Spock never needed to hoist `bucketName` outside the try — it stayed a normal local variable in the `given:` scope and `final` was harmless there.

PR #6771, opened 2026-01-28
Add spec for multiple Nextflow registry support

pditommaso:
## Summary

- Adds feature specification for supporting multiple Nextflow registries with ordered fallback
- Enables enterprise/air-gapped environments to use private registries while optionally falling back to public
- Supports API key authentication for private registries
- Maintains backward compatibility with existing single-registry configuration

## Key Design Decisions

| Aspect | Decision |
|--------|----------|
| Config syntax | `registry { url = ['url1', 'url2']; apiKey = 'key' }` |
| Env var syntax | `NXF_PLUGINS_REGISTRY_URL=url1,url2` (comma-separated) |
| Authentication | Single `apiKey` in registry block for authenticated access |
| Precedence | Env var completely overrides config (Nextflow convention) |
| Default registry | Explicit only - no implicit fallback |
| Backward compat | Single string `url` still works |

## Example Configuration

```groovy
registry {
    url = ['https://internal.company.com/plugins', 'https://registry.nextflow.io/api']
    apiKey = 'your-api-key'
}
```

## Spec Contents

- 6 user stories (P1: config, env var, backward compat; P2: air-gapped, authenticated access, error reporting)
- 17 functional requirements covering config syntax, resolution, authentication, precedence, and logging
- Edge cases and success criteria defined

## Out of Scope

- Per-registry credentials (different API keys for different registries)
- Registry health checks or automatic failover based on latency
- Caching of "plugin not found" responses

## Test plan

- [ ] Review spec for completeness
- [ ] Validate design decisions align with Nextflow conventions
- [ ] Check edge cases coverage

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #7180, opened 2026-05-27
Use registry config for plugin repos

jorgee:
Related to #6771

## Problem

The `registry` config scope and the plugin/module resolution paths handled the
default Nextflow registry inconsistently:

- **Modules**: a `registry` block with a custom `url` fully replaced the default
  registry — the configured list was authoritative and the public registry was
  dropped.
- **Plugins**: the `registry` scope was not wired into plugin resolution at all;
  only a *single* plugin repository could be set, via `indexUrl` /
  `NXF_PLUGINS_REGISTRY_URL`.

## Summary

This PR wires the `registry` config scope into **plugin** resolution and makes
both modules and plugins treat the configured registries as **authoritative**:

- When no registry is configured, the default public registry
  (`https://registry.nextflow.io/api`) is used.
- When one or more `registry.url` entries are set, Nextflow queries **exactly
  those**, in the order listed. The default is **not** added implicitly — to
  keep using the public registry alongside your own, list it explicitly.

This restores the pre-existing module behaviour and applies the same rule to
plugins.

**Plugins** (`nf-commons`)
- `PluginUpdater.addRegistryRepos(RegistryConfig)` *replaces* the default
  registry repository (`registry`/`nextflow.io`, derived from `indexUrl` /
  `NXF_PLUGINS_REGISTRY_URL`) with an `HttpPluginRepository` for each configured
  URL. Repo ids are generated as `registry-N` with a collision guard; test repos
  and the offline local repo are preserved.
- `PluginsFacade.applyRegistryConfig(config)` extracts the `registry` block and
  calls `addRegistryRepos` on the existing updater (no updater recreation), but
  only when `registry.url` is explicitly set — so the `NXF_PLUGINS_REGISTRY_URL`
  default and API authentication are left untouched.
- `HttpPluginRepository.refresh()` skips the registry round-trip when no
  metadata has been prefetched (its plugin map is empty or null, via Groovy
  truthiness) — required because `UpdateManager.addRepository()` calls
  `refresh()` immediately as each configured registry is added, before
  `prefetch()` populates the metadata map.
- `PluginUpdater.getPluginsMap()` overrides pf4j's implementation, which merges
  repositories with `Map.putAll` keyed by plugin id — so with multiple
  registries the last repo serving a plugin id would fully overwrite the
  earlier ones, discarding releases that exist only in an earlier-listed
  registry (and computing "latest" within a single registry). The override
  unions each plugin's releases across all repositories, de-duplicated by
  version; on a version clash the earlier-listed registry wins (first listed =
  highest priority). Releases are copied into fresh `PluginInfo` objects so the
  per-repository caches are not mutated.

  > **`NXF_PLUGINS_TEST_REPOSITORY` behaviour note.** This union applies to *all*
  > multi-repository setups, not just `registry.url`. In the test-repository path
  > `wrap()` lists the default registry first and the test repo second, so under
  > the new *first-listed-wins* rule the default registry now wins on a **version
  > clash**, whereas the previous `Map.putAll` let the test repo (listed last)
  > override it. The test repo still contributes any versions the default lacks.
  > This is unlikely to bite in practice (test plugins use unpublished versions),
  > but it is an intentional semantic change; covered by the
  > `should not let a later repository override the default on a version clash`
  > unit test.

**Modules** (`nextflow`)
- `RegistryClientFactory.forConfig()` builds the client from `cfg.allUrls`, which
  falls back to the default only when nothing is configured. No implicit default
  is prepended.

`RegistryConfig` itself is unchanged — it stays a faithful representation of the
user's config block, returning the default only when no `url` is configured.

## `NXF_PLUGINS_REGISTRY_URL` vs `registry.url` (behaviour after this PR)

| `NXF_PLUGINS_REGISTRY_URL` | `registry.url` | Plugin registry | Module registry |
|---|---|---|---|
| unset | unset | built-in default | built-in default |
| set   | unset | env var URL | built-in default |
| unset | set   | `registry.url` | `registry.url` |
| set   | set   | `registry.url` (env var overridden) | `registry.url` |

- `registry.url` is applied at config-load time, after the env-var-based default
  repo is created, so **`registry.url` takes precedence** over
  `NXF_PLUGINS_REGISTRY_URL` for plugins.
- `NXF_PLUGINS_REGISTRY_URL` affects **plugins only**; it never influences module
  resolution. Modules fall back to the built-in public registry whenever
  `registry.url` is unset.
- The env var continues to work unchanged when `registry.url` is not set
  (including pointing at the legacy `plugins.json` index).

This behaviour is also documented in `docs/plugins/plugin-registry.md` and the
`registry` config scope reference.

## Out of scope

- **API authentication** for additional registries — `apiKey` still authenticates
  only the primary registry.
- **Deprecating `NXF_PLUGINS_REGISTRY_URL`** — tracked separately. It is not a
  1:1 replacement for `registry.url`: the env var is plugin-only and supports the
  legacy `plugins.json` index, which `registry.url` does not.

## Test plan

- [x] `./gradlew :nf-commons:test` (936 tests, 0 failures) — `PluginUpdaterTest`:
      configured registries replace the default, replacement even when a
      configured URL matches the default, releases unioned across registries
      (first-listed wins on a version clash), offline no-op, null-config no-op
- [x] `./gradlew :nextflow:test` (3772 tests, 0 failures) — `RegistryClientFactoryTest`:
      configured registries queried in order, only the configured registry
      queried (no default fallback)
- [x] Manual test: include registry-dev in config and run with a plugin only
      available in registry-dev

## Implementation note

We also considered routing the plugin calls through `RegistryClient` — it takes
a list of URLs and performs the calls until one replies successfully. This works
well for modules, but it doesn't fit pf4j's `PluginUpdater` and how plugins are
managed, where it expects to manage them separately.

Moreover, there is a chicken-and-egg problem between plugin loading and config
resolution: `PluginUpdater` is initialized *before* the config is resolved,
because some SCM and filesystem extensions are themselves provided by plugins and
are required at that stage. At that point only the default plugin registry is
available. Later, at load time — once the config is resolved — the registries
from the `registry` scope replace the default, and plugins are resolved using
the configured registries.

### Bootstrap-window caveat

Because `applyRegistryConfig()` runs inside `Plugins.load(config)`, the
`registry` scope is **not** in effect during the steps that precede it in
`CmdRun.run()` — resolving the main script (`getScriptFile`) and parsing the
config (`ConfigBuilder.build`). Plugins lazily started in that window via
`Plugins.startIfMissing(...)` — e.g. a filesystem provider for a remote script
or an `includeConfig` location (`s3://`, `gs://`, `az://`, triggered from
`FileHelper.getOrCreateFileSystemFor`), or the `nf-codecommit` SCM provider —
are resolved against the **default** plugin registry (or
`NXF_PLUGINS_REGISTRY_URL`) only.

Consequences:
- Plugins already installed/cached load fine, regardless of registry config.
- A plugin that must be downloaded in this window resolves from the default
  registry — so **the default registry may still be fetched even when it is not
  listed in `registry.url`**. A plugin that lives *only* in a custom
  `registry.url` registry cannot be resolved at this stage; it must be available
  in the default registry, set via `NXF_PLUGINS_REGISTRY_URL`, or pre-installed.
- This window is unchanged by this PR (the plugin path never consulted
  `registry.url` before config load), and it's also why `NXF_PLUGINS_REGISTRY_URL`
  remains useful — it is the only setting that influences pre-config plugin
  resolution. Documented in `docs/plugins/plugin-registry.md`.


jorgee:
Ready for review.

jorgee:
Pushed `ff876f371` addressing internal AI review feedback on dev-mode handling and empty-url semantics.

**Dev mode** — In dev mode plugins are resolved from the development classpath and never downloaded from a registry (`installPlugin`/`updatePlugin` are unsupported, `shouldUpdate` returns `false` for `RuntimeMode.DEVELOPMENT`). `PluginsFacade.applyRegistryConfig` now warns and skips when the `registry` scope is set in dev mode, rather than calling `addRegistryRepos` on a `DevPluginUpdater` — whose protected constructor never wires the pf4j `repositories` list and would otherwise NPE.

**Empty-url semantics** — Clarified that an empty or unset `registry.url` counts as "not configured" and leaves the default registry in place, consistent with module resolution in `RegistryClientFactory`. Since `RegistryConfig.getAllUrls()` always falls back to the default, the `if(!urls) return` guard in `addRegistryRepos` was dead code; removed it and documented the contract.

**Tests** — Added `PluginsFacadeTest` coverage for the config-wiring seam: `registry.url` set is applied to the updater, empty/unset url is a no-op, and dev mode is ignored.

`./gradlew :nf-commons:test --tests PluginsFacadeTest --tests PluginUpdaterTest` — passing.

modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
```diff
@@ -100,6 +101,108 @@ class PluginUpdater extends UpdateManager {
...
+     * each {@link HttpPluginRepository} with the metadata it actually needs.
+     */
+    void addRegistryRepos(RegistryConfig registryConfig) {
+        if( offline || !registryConfig )
+            return
+        final urls = registryConfig.getAllUrls()
+        // the configured registries take over: drop the default registry repository so that
+        // only the user-provided endpoints are queried
+        for( String id : DEFAULT_REGISTRY_REPO_IDS )
+            removeRepository(id)
```
pditommaso: Both default ids are removed unconditionally, but `wrap()` only ever creates **one** of them (`registry` for http, or `nextflow.io` for a `.json` url). On top of that, pf4j 2.3.0's `removeRepository()` logs `"Repository with id X not found, doing nothing"` **unconditionally** — the `log.warn` sits outside the loop and fires even on a successful removal. Net effect: two misleading `WARN` lines on every run where `registry.url` is set.

Suggest removing only ids that actually exist:

```groovy
final present = this.@repositories*.id as Set
for( String id : DEFAULT_REGISTRY_REPO_IDS )
    if( id in present ) removeRepository(id)
```

modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
```diff
@@ -100,6 +101,108 @@ class PluginUpdater extends UpdateManager {
...
+     * repositories does not mutate the metadata cached by each repository.
+     */
+    private static PluginInfo copyPluginInfo(PluginInfo src) {
+        final copy = new PluginInfo()
+        copy.id = src.id
+        copy.name = src.name
+        copy.description = src.description
+        copy.provider = src.provider
+        copy.projectUrl = src.projectUrl
+        copy.releases = src.releases != null
```
pditommaso, line 189: Minor: `PluginInfo.repositoryId` isn't copied here. It's not used in the download path (`getFileDownloader`/`getFileVerifier` resolve via `getPlugin(id)`), so no functional impact — but the copies will render `repositoryId='null'` in `toString()`/debug logs. Consider copying it for cleaner diagnostics.

modules/nf-commons/src/main/nextflow/plugin/PluginsFacade.groovy
```diff
@@ -305,9 +307,28 @@ class PluginsFacade implements PluginStateListener {
...
+        if( !registryMap?.url )
+            return
+        // in dev mode plugins are resolved from the development classpath, not downloaded from
+        // any registry, so the `registry` scope has no effect on plugin resolution
+        if( mode==DEV_MODE ) {
+            log.warn "Plugin registry config is ignored in development mode -- plugins are resolved from the development classpath"
+            return
+        }
+        this.registryConfig = new RegistryConfig(registryMap)
+        updater?.addRegistryRepos(registryConfig)
```
pditommaso: `applyRegistryConfig` isn't idempotent: a second `load(config)` would re-run `addRegistryRepos`, and since the defaults are already gone the collision guard would mint `registry-2`/`registry-3`… — duplicate repos pointing at the same URLs. There's a single call site today (`CmdRun`), so the risk is low, but a cheap guard (e.g. skip when `registryConfig != null`, or when a `registry-0` repo already exists) would make re-entry safe.

pditommaso (commented):
Reviewed the plugin-registry wiring — the core logic is correct against pf4j 2.3.0: the `getPluginsMap()` override properly fixes the `putAll` last-wins release shadowing, the `refresh()` guard is justified by `addRepository()`'s eager refresh, and `getRepositories()` doesn't auto-refresh so prefetched metadata survives. Good docs and thorough tests.

A few minor, non-blocking findings left inline. The main behavioral change worth calling out in release notes: configured registries are now **authoritative** for plugins too (the default is no longer implicitly appended), which is a breaking change for anyone relying on the old fallback.

modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
jorgee: Fixed in 90272adca. `addRegistryRepos` now builds the present-id set via direct field access and calls `removeRepository(id)` only for ids that actually exist, so pf4j-update's unconditional "Repository with id X not found" warning no longer fires. Consolidated with the collision-guard set that followed.

modules/nf-commons/src/main/nextflow/plugin/PluginUpdater.groovy
jorgee: Fixed in 90272adca — `copyPluginInfo` now copies `repositoryId` so the merged copies no longer render `repositoryId='null'` in `toString()`/debug logs.

modules/nf-commons/src/main/nextflow/plugin/PluginsFacade.groovy
jorgee: Fixed in 90272adca — `applyRegistryConfig` now returns early when `registryConfig` is already set, so a repeated `load(config)` won't mint duplicate `registry-N` repos. A no-op first call (no `registry.url`) leaves `registryConfig` null, so a later call with a URL still applies. Added a `should apply registry config only once when invoked repeatedly` test.

jorgee:
Fixed review comments and also fixed merge conflicts due to documentation changes

pditommaso (commented):
Overall this looks solid and well-tested — the `getPluginsMap()` union override is correct (I traced it against `pf4j-update:2.3.0`: merged releases download via the absolute `release.url`, so keeping a foreign `repositoryId` is safe), and the `HttpPluginRepository.refresh()` short-circuit is genuinely needed since `addRepository()` calls `refresh()` synchronously before `prefetch()` populates the map. A few suggestions below.

### 1. Drop the `PluginsFacade.registryConfig` field — move the apply-once guard onto the updater

The `registryConfig` field is read only inside `applyRegistryConfig`: as the re-entrancy sentinel and to pass the config to the updater. That state really belongs to the updater, which owns `repositories`. Moving the guard there removes the field and makes `addRegistryRepos` safe by construction, rather than relying on every caller to guard first:

```groovy
// PluginsFacade
protected void applyRegistryConfig(Map config) {
    final registryMap = Bolts.navigate(config, 'registry') as Map
    if( !registryMap?.url )
        return
    if( mode==DEV_MODE ) {
        log.warn "Plugin registry config is ignored in development mode -- plugins are resolved from the development classpath"
        return
    }
    updater?.addRegistryRepos(new RegistryConfig(registryMap))
}
```

```groovy
// PluginUpdater
private boolean registryReposApplied

void addRegistryRepos(RegistryConfig registryConfig) {
    if( offline || !registryConfig || registryReposApplied )
        return
    registryReposApplied = true
    final urls = registryConfig.getAllUrls()
    ...
}
```

This also removes a latent ordering fragility: today `this.registryConfig` is set *before* `updater?.addRegistryRepos(...)`, so if `updater` were ever null at that point the config would be marked applied but never wired in. With the flag on the updater that can't happen. The two `PluginsFacadeTest` cases asserting `facade.@registryConfig` would assert on the updater interaction instead (`1 *` first call, `0 *` second), which is what they're really verifying.

### 2. Priority inversion for `NXF_PLUGINS_TEST_REPOSITORY`

The new union `getPluginsMap()` applies to **all** multi-repo setups, not just `registry.url`. In the test path `wrap()` builds `[registry, uri]` (public first, test repo second). Before this PR, `putAll` let the test repo (`uri`) override the public entry on a colliding version; now the public registry is first-listed and wins on a version clash, with the test repo only contributing versions the public one lacks. So a test that republishes an *existing* public version under `NXF_PLUGINS_TEST_REPOSITORY` would silently resolve the public URL instead of the test one. Low real-world likelihood (test plugins usually use unpublished versions), but it's an unadvertised semantic change — worth a note in the PR description and ideally an integration-test check.

### 3. `AddedInVersion` stamps disagree

`docs/plugins/plugin-registry.mdx` says `26.10` while `docs/reference/config.mdx` says `26.04` for the same `registry` scope landing. Current `VERSION` is `26.07.0-edge`, so `26.04` is already in the past. Pick one real target release and use it in both docs.

jorgee:
Thanks for the follow-up review. Addressed in d328c10f4:

**1. Drop `registryConfig` field / move the guard onto the updater** — done. The field is gone; `PluginUpdater` now carries a `registryReposApplied` flag and `addRegistryRepos` no-ops on re-entry, so it's idempotent by construction and the pre-updater-call ordering fragility is removed. The `PluginsFacadeTest` cases now assert on the updater interaction (`1 *` / `0 *`), and the once-only test moved to `PluginUpdaterTest` (`should apply the registry config only once when invoked repeatedly`).

**2. `NXF_PLUGINS_TEST_REPOSITORY` priority inversion** — documented in the PR description (behaviour note under the `getPluginsMap()` bullet) and pinned down with a unit test, `should not let a later repository override the default on a version clash`, which mirrors the `[default, extra]` layout and asserts the first-listed default wins the clash while the extra repo still contributes its unique versions.

**3. `AddedInVersion` stamps** — kept as `26.04` in `config.mdx` and `26.10` in `plugin-registry.mdx` deliberately: they mark two different landings. The `registry` config scope itself shipped with the module system (first tagged `v26.03.0-edge`, i.e. the `26.04` stable line), so `26.04` is the accurate "added in" for the scope reference. The plugin-side wiring is new in this PR and targets the next stable, `26.10`. Happy to revisit if you'd prefer both to point at a single release regardless.

pditommaso (approved):
All three review points addressed — LGTM 👍

- `copyPluginInfo` now copies `repositoryId` ✅
- Idempotency guard moved onto the updater (`registryReposApplied`), which is cleaner than a facade-level check — idempotent by construction, sentinel field dropped ✅

Minor / nice-to-have on `addRegistryRepos` — the removal now skips the non-existent default id, so one of the two misleading warnings is gone. But pf4j-update 2.3.0's `UpdateManager.removeRepository()` logs `"Repository with id X not found, doing nothing"` **unconditionally** — the `log.warn` is reached even after a successful removal (verified in the bytecode). So removing the one default id that *does* exist still emits a single misleading WARN at startup.

Not blocking. If you want it fully silent, bypass `removeRepository` and drop the entry directly, e.g.:

```groovy
// pf4j 2.3.0 removeRepository() warns even on success — mutate the list directly
this.@repositories.removeIf { it.id in DEFAULT_REGISTRY_REPO_IDS }
```

jorgee:
Thanks — addressed the `removeRepository()` warning nit in 8d3fe611a. `addRegistryRepos` now drops the default registry repositories by mutating the list directly (`this.@repositories.removeIf { it.id in DEFAULT_REGISTRY_REPO_IDS }`) instead of calling `removeRepository(id)`, so pf4j-update 2.3.0's unconditional "Repository with id X not found" WARN is gone entirely. Existing `PluginUpdaterTest` cases covering both default-id branches (`registry` for http, `nextflow.io` for the `.json` index) still pass.

bentsherman (approved):

PR #7181, opened 2026-05-27
Make plugin metadata prefetch resilient to registry rate limiting [#7176]

### LINKED ISSUE #7176: HttpPluginRepository.prefetch() crashes startup on HTTP 429 — registry.nextflow.io/api/v1/plugins/dependencies called unconditionally since 26.04.2

**Expected:** When `nextflow plugin install <plugin>@<version>` is run and the registry returns HTTP 429 (rate limit), Nextflow should retry with backoff and/or degrade gracefully — it should not abort startup. Additionally, if all required plugins are already installed locally, the metadata prefetch network call should be skipped entirely.

**Actual:** `HttpPluginRepository.sendAndParse()` throws a hard `PluginRuntimeException` on any non-200 response including 429, crashing Nextflow startup unconditionally. With ~40 parallel CI jobs, `registry.nextflow.io/api/v1/plugins/dependencies` is reliably rate-limited and every job fails at startup.

1. Set up a CI system with 40+ parallel jobs, each running:
```bash
nextflow plugin install nf-schema@2.5.1
```
2. All jobs start within a short window (typical in matrix CI)
3. Every job calls `registry.nextflow.io/api/v1/plugins/dependencies` at startup via `HttpPluginRepository.prefetch()` — even if the plugin is already cached locally
4. The registry returns 429; Nextflow aborts with `PluginRuntimeException`

This is a **regression from 25.04.3**, which used `DefaultUpdateRepository` (static `plugins.json`) instead of `HttpPluginRepository` and did not make per-startup `/api/v1/plugins/dependencies` calls. Pinning back to 25.04.3 eliminates the failures.

### Program output

```
PluginRuntimeException: Invalid response while fetching plugin metadata from:
  https://registry.nextflow.io/api/v1/plugins/dependencies?plugins=nf-schema%402.5.1&nextflowVersion=26.04.2
- http status: 429
- response   : <rate limit response body>
```

### Root cause (specific code pointers, v26.04.2)

The call chain with no 429 handling and no skip-if-installed logic:

**1. `PluginsFacade.groovy:415`** — unconditional call, no `if (!offline)` guard:
```groovy
updater.prefetchMetadata(startable)
```
`offline` is correctly read from `NXF_OFFLINE` at line 66 and passed to `PluginUpdater` at line 254, but never used to guard this call.

**2. `PluginUpdater.groovy:150–163`** — iterates repos, calls `prefetch()` on each `PrefetchUpdateRepository`. `this.offline` is stored on the instance but never consulted here.

**3. `HttpPluginRepository.groovy:142–157`** — builds and fires the HTTP request on every startup, regardless of whether the plugin is already installed:
```groovy
def uri = url.resolve("v1/plugins/dependencies?plugins=...&nextflowVersion=...")
return sendAndParse(HttpRequest.newBuilder().uri(uri).GET().build())
```

**4. `HttpPluginRepository.groovy:165–167` ← THE BUG** — hard-aborts on any non-200 including 429, with no `Retry-After` handling:
```groovy
if( resp.statusCode() != 200 ) {
    throw new PluginRuntimeException("Invalid response ... http status: ${resp.statusCode()} ...")
}
```

### Environment

* Nextflow version: 26.04.2 (regression from 25.04.3)
* Java version: 21
* Operating system: Linux (self-hosted CI runners)
* Bash version: 5.x

### Additional context

- **Related issue #7013** "Plugin metadata prefetch fails hard even when result is unused" — closed with suggestion to use `NXF_OFFLINE=true`. That workaround is incomplete: it requires plugins to be pre-cached locally *and* prevents any network access, making it incompatible with steps that legitimately need to install plugins.
- **Related PR #7019** "Make plugin metadata prefetch failure non-fatal" — addresses the hard-abort symptom but was not merged.
- **Suggested fixes:**
  1. In `sendAndParse()`: handle 429 specifically — respect `Retry-After` header, retry with exponential backoff, don't hard-abort.
  2. In `prefetchMetadata` / `prefetch()`: skip the network call if all requested plugins are already present locally.
  3. Defence-in-depth: add `if (!offline)` guard at `PluginsFacade.groovy:415`.

pditommaso:
## Summary

Fixes [#7176](https://github.com/nextflow-io/nextflow/issues/7176) — under burst CI load the plugin registry returns HTTP 429 with `Retry-After`, but the HTTP client ignored the hint and every attempt landed inside the same rate-limit window, aborting Nextflow startup.

Four layered changes:

1. **Skip registry round-trip when plugins are already cached** — `PluginUpdater.prefetchMetadata` filters out plugins that are pinned to a specific version and already installed locally. When every requested plugin matches, the registry call is skipped entirely. Unpinned specs still go through so latest-release resolution keeps working.

2. **Non-fatal prefetch** — `HttpPluginRepository.prefetch` catches `PluginRuntimeException`, logs a `warn`, and leaves an empty plugins map. Prefetch is an optimization, not a hard requirement; a sustained rate-limit window or true outage must not abort startup. The lazy `getPlugin(id)` path still surfaces per-plugin errors when an actually-missing plugin is needed downstream. The `plugins` field is initialised eagerly so `getPlugin` and `refresh` are safe to call when prefetch was skipped or failed.

3. **Honor `Retry-After` in the HTTP layer** — picks up `lib-httpx@2.3.0` + `lib-retry@2.1.0` ([libseqera#72](https://github.com/seqeralabs/libseqera/pull/72)) so `HxClient` retries on 429/503 honor the server-supplied delay as a lower bound, capped by the configured `maxDelay`.

4. **Explicit `NXF_OFFLINE` guard** — `PluginsFacade.start()` skips `prefetchMetadata` entirely when `NXF_OFFLINE=true`. Currently handled transitively (no `PrefetchUpdateRepository` is wired in offline mode, so the iteration is a no-op), but an explicit `if (!offline)` at the call site makes the no-network guarantee unmissable in the code and protects against regressions if a remote `PrefetchUpdateRepository` is later added to the offline-mode repo list by mistake.

### Why the Failsafe bump

`lib-retry@2.1.0` targets Failsafe ≥3.2.1 because `RetryPolicyBuilder.handleIf` now requires `CheckedPredicate` and `ExecutionAttemptedEvent` renamed `getLastFailure` → `getLastException`. Bumping Nextflow's direct Failsafe dependency from `3.1.0` → `3.3.2` aligns with the lib and surfaces some pre-existing API drift that is migrated mechanically (no behavior change):

- `event.lastFailure` → `event.lastException`
- `Predicate<? extends Throwable>` → `CheckedPredicate<? extends Throwable>` on `RetryPolicyBuilder.handleIf`

Sites touched: `PluginUpdater`, `PublishDir`, `GridTaskHandler`, `K8sClient`, `BatchClient` (nf-google), `AzBatchService`, `AzFileSystem` (nf-azure), and the matching `AzBatchServiceTest` `Mock` declaration.

### End-to-end behavior

- **CI common case** (pinned + already cached): zero registry calls.
- **CI cold case** (first install of a pinned plugin): registry returns 429 → `HxClient` waits `min(maxDelay=90s, max(backoff, Retry-After))` → retries up to 5x → usually succeeds.
- **Sustained outage / hard rate-limit**: retries exhaust → `prefetch()` degrades non-fatally → startup continues; downstream operations that genuinely need the metadata fail with a specific error.

## Test plan

- [x] `:nf-commons:test --tests HttpPluginRepositoryTest --tests PluginUpdaterTest` — green
- [x] `:plugins:nf-k8s:test`, `:plugins:nf-google:test`, `:plugins:nf-azure:test --tests AzBatchServiceTest --tests AzFileSystemTest` — green
- [x] All affected plugins compile cleanly under Failsafe 3.3.2 (`./gradlew compileGroovy` on each)
- [x] New WireMock test `'prefetch degrades gracefully when the registry keeps returning 429'` in `HttpPluginRepositoryTest`
- [ ] CI to verify end-to-end on a real Nextflow invocation

### DIFF
--- a/modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
+++ b/modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
@@ -52,7 +52,9 @@ class HttpPluginRepository implements PrefetchUpdateRepository {
     private final URI url
     private final HxClient httpClient
 
-    private Map<String, PluginInfo> plugins
+    // Initialised eagerly so that getPlugin()/refresh() are safe to call even when prefetch()
+    // was skipped (e.g. PluginUpdater skip-if-installed) or failed (non-fatal degradation).
+    private Map<String, PluginInfo> plugins = new HashMap<>()
 
     HttpPluginRepository(String id, URI url) {
         this.id = id
@@ -83,7 +85,16 @@ class HttpPluginRepository implements PrefetchUpdateRepository {
     @Override
     void prefetch(List<PluginRef> plugins) {
         if (plugins && !plugins.isEmpty()) {
-            this.plugins = fetchMetadata(plugins)
+            try {
+                this.plugins = fetchMetadata(plugins)
+            }
+            catch (PluginRuntimeException e) {
+                // Don't abort Nextflow startup if the registry is unreachable or rate-limited:
+                // downstream code will either resolve the plugins from the local store or fail
+                // with a more specific error when the metadata is actually needed.
+                log.warn "Failed to prefetch plugin metadata from ${url} - cause: ${e.message}"
+                this.plugins = new HashMap<>()
+            }
         }
     }
 
@@ -99,10 +110,6 @@ class HttpPluginRepository implements PrefetchUpdateRepository {
 
     @Override
     Map<String, PluginInfo> getPlugins() {
-        if (plugins==null) {
-            log.warn "getPlugins() called before prefetch() - plugins map will be empty"
-            return Map.of()
-        }
         return Collections.unmodifiableMap(plugins)
     }
 
--- a/modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
+++ b/modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
@@ -84,17 +84,17 @@ class HttpPluginRepository implements PrefetchUpdateRepository {
     // like any other implementation of UpdateRepository.
     @Override
     void prefetch(List<PluginRef> plugins) {
-        if (plugins && !plugins.isEmpty()) {
-            try {
-                this.plugins = fetchMetadata(plugins)
-            }
-            catch (PluginRuntimeException e) {
-                // Don't abort Nextflow startup if the registry is unreachable or rate-limited:
-                // downstream code will either resolve the plugins from the local store or fail
-                // with a more specific error when the metadata is actually needed.
-                log.warn "Failed to prefetch plugin metadata from ${url} - cause: ${e.message}"
-                this.plugins = new HashMap<>()
-            }
+        if (!plugins)
+            return
+        try {
+            this.plugins = fetchMetadata(plugins)
+        }
+        catch (PluginRuntimeException e) {
+            // Don't abort Nextflow startup if the registry is unreachable or rate-limited:
+            // downstream code will either resolve the plugins from the local store or fail
+            // with a more specific error when the metadata is actually needed.
+            log.warn "Failed to prefetch plugin metadata from ${url} - cause: ${e.message}"
+            this.plugins = new HashMap<>()
         }
     }
 
--- a/modules/nf-commons/src/main/nextflow/plugin/PluginsFacade.groovy
+++ b/modules/nf-commons/src/main/nextflow/plugin/PluginsFacade.groovy
@@ -411,8 +411,10 @@ class PluginsFacade implements PluginStateListener {
             final skippedIds = skippable.collect{ plugin -> plugin.id }
             log.debug "Plugin 'start' is not required in embedded mode -- ignoring for plugins: $skippedIds"
         }
-        // prefetch the plugins meta
-        updater.prefetchMetadata(startable)
+        // prefetch the plugins meta - skipped in offline mode to make the no-network guarantee
+        // explicit (in offline mode no remote repo is wired in, so this is also a no-op transitively)
+        if( !offline )
+            updater.prefetchMetadata(startable)
         // finally start the plugins
         for( PluginRef plugin : startable ) {
             updater.prepareAndStart(plugin.id, plugin.version)
--- a/modules/nextflow/build.gradle
+++ b/modules/nextflow/build.gradle
@@ -67,7 +67,7 @@ dependencies {
     api ('org.jsoup:jsoup:1.15.4')
     api 'jline:jline:2.9'
     api 'org.pf4j:pf4j:3.14.1'
-    api 'dev.failsafe:failsafe:3.1.0'
+    api 'dev.failsafe:failsafe:3.3.2'
     api 'io.seqera:lib-trace:0.1.0'
     api 'com.fasterxml.woodstox:woodstox-core:7.1.1'
     api 'org.apache.commons:commons-compress:1.27.1'  // For tar.gz extraction
--- a/modules/nextflow/src/main/groovy/nextflow/executor/GridTaskHandler.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/GridTaskHandler.groovy
@@ -21,13 +21,13 @@ import static nextflow.processor.TaskStatus.*
 import java.nio.file.Path
 import java.nio.file.attribute.BasicFileAttributes
 import java.time.temporal.ChronoUnit
-import java.util.function.Predicate
 import java.util.regex.Pattern
 
 import dev.failsafe.Failsafe
 import dev.failsafe.RetryPolicy
 import dev.failsafe.event.EventListener
 import dev.failsafe.event.ExecutionAttemptedEvent
+import dev.failsafe.function.CheckedPredicate
 import dev.failsafe.function.CheckedSupplier
 import groovy.transform.CompileStatic
 import groovy.transform.Memoized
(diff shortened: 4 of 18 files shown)


modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
```diff
@@ -83,7 +85,16 @@ class HttpPluginRepository implements PrefetchUpdateRepository {
...
+            try {
+                this.plugins = fetchMetadata(plugins)
+            }
+            catch (PluginRuntimeException e) {
+                // Don't abort Nextflow startup if the registry is unreachable or rate-limited:
+                // downstream code will either resolve the plugins from the local store or fail
+                // with a more specific error when the metadata is actually needed.
+                log.warn "Failed to prefetch plugin metadata from ${url} - cause: ${e.message}"
+                this.plugins = new HashMap<>()
+            }
```
bentsherman: nit: use if-guard pattern instead of nesting try-catch in the if statement

modules/nf-commons/src/main/nextflow/plugin/PluginsFacade.groovy
```diff
@@ -411,8 +411,10 @@ class PluginsFacade implements PluginStateListener {
             final skippedIds = skippable.collect{ plugin -> plugin.id }
             log.debug "Plugin 'start' is not required in embedded mode -- ignoring for plugins: $skippedIds"
         }
-        // prefetch the plugins meta
-        updater.prefetchMetadata(startable)
+        // prefetch the plugins meta - skipped in offline mode to make the no-network guarantee
+        // explicit (in offline mode no remote repo is wired in, so this is also a no-op transitively)
+        if( !offline )
+            updater.prefetchMetadata(startable)
```
bentsherman, line 417: do we actually need this? we considered it in a previous PR but my sense was that this would only delay an inevitable failure, and likely with a worse error message

pditommaso: Fair point — but I think the new `isAlreadyInstalled` filter changes the dynamics.

In offline mode with cached plugins, the guard makes the no-network contract explicit at the call site. Without it, `prefetchMetadata` is still a no-op (no `PrefetchUpdateRepository` is wired in offline mode) — but that's an implicit invariant in `wrap()`, whereas the guard documents it locally and protects against a regression where someone adds a `PrefetchUpdateRepository` to the offline path.

For the missing-plugin case you're right — error path is identical either way.

modules/nf-commons/src/main/nextflow/plugin/HttpPluginRepository.groovy
pditommaso: Make sense! 👉  ebe72864

jorgee:
Pushed `e4cc899d4` — fixes a gap in the skip-if-installed optimization.

### The problem
`prefetchMetadata` filtered out already-installed plugins via `isAlreadyInstalled`, which checked `pluginManager.getPlugin(ref.id)`. But at prefetch time (`PluginsFacade.start()` → `prefetchMetadata`, before `prepareAndStart`) the local plugins have **not been loaded into the manager yet** — its per-run plugins root (`.nextflow/plr/<uuid>/`) is still empty, and pf4j logs `No plugins`. So `getPlugin()` always returned `null`, `isAlreadyInstalled` always returned `false`, and the `/v1/plugins/dependencies` round-trip fired on **every** startup even when the plugin was cached — exactly the condition that triggers the 429 storm under parallel CI load.

Verified empirically before the fix: a `run` (and `plugin install`) of an **already-installed** `nf-schema@2.5.1` still made **1 registry call** (and reused the cached zip without downloading).

### The change
`isAlreadyInstalled` now checks the **on-disk plugin store** instead of the runtime manager, mirroring `installPlugin()` which reuses the cached copy whenever the `<id>-<version>` store directory exists (it only downloads when missing — the `classes/META-INF/MANIFEST.MF` check there is a corruption *warning*, not a gate):

```groovy
return pluginsStore != null && FilesEx.exists(pluginsStore.resolve("${ref.id}-${ref.version}"))
```

Directory presence is the authoritative "no remote metadata needed" signal and works for both real installs and the test fixtures. (`private` → `protected` for direct unit testing.)

### New unit test
`PluginUpdaterTest › 'should detect already-installed pinned plugin from the on-disk store'` — confirms `getPlugin()` is null at this point (the trap the old code fell into), that a pinned + cached plugin is detected, that a different version is not, and that an unpinned spec still needs metadata. Verified **red against the old code, green after the fix**.

### Validation
- `:nf-commons:test --tests nextflow.plugin.*` — green (incl. `HttpPluginRepositoryTest`, `PluginUpdaterTest`).
- End-to-end with a `make pack` binary (`NXF_PLUGINS_DIR` per process):
  - **run / `plugin install` of cached `nf-schema@2.5.1`** → `skipping registry metadata prefetch`, **0 registry calls**, exit 0.
  - **cold install** (empty dir) → **1 registry call**, plugin installed, exit 0 (normal path intact).
  - **100 parallel jobs, plugin cached (the CI scenario)** → **100/100 success, 0 total registry calls** (before this fix all 100 would hit the registry).

jorgee (approved):
Looks fine with me.

modules/nf-commons/src/main/nextflow/plugin/PluginsFacade.groovy
pditommaso: I assume this is solved, let me know otherwise

PR #7184, opened 2026-05-28
Fix cache hash for user-defined record types

jorgee:
## Summary

Typed records — those declared via `record … { … }` and instantiated through their generated constructor — fell through the type dispatch in `HashBuilder` to the catch-all that uses `value.hashCode()`. Because the generated record classes don't override `hashCode`, every instance produced a different identity-based hash, and `-resume` always re-ran the task even when inputs were equal.

The user-visible symptom in `-dump-hashes` is a record value rendered as `_nf_script_xxx.Dataset@20d692ba` (the JVM default `Object.toString()`), with a different `@<hex>` suffix on every run.

This PR adds a `Record` branch to `HashBuilder.with(Object)` that reflectively reads the record's public, non-static instance fields into a `LinkedHashMap<String,Object>` and hashes it through the same `hashUnorderedCollection` path used for `Map`/`RecordMap`. Net effects:

- A typed record whose fields haven't changed hashes the same across runs → `-resume` works.
- A record built with the `record(...)` stdlib helper and a typed-record constructor with the same field values produce **the same** hash.
- The hash is independent of named-argument / field-declaration order (sum of per-entry hashes is commutative).
- The new branch is placed **after** the `Map` branch so `RecordMap` (which is both `Record` and `Map`) keeps its existing hashing path.

## Test plan

New unit tests in `HashBuilderTest`:

- [x] `should hash typed record by field values, independent of JVM identity` — regression test for the `-resume` bug.
- [x] `typed record and equivalent RecordMap should hash to the same value`
- [x] `record built with the record(...) built-in should hash the same as the typed constructor`
- [x] `record(...) built-in hash should not depend on the order of named arguments`
- [x] `nested typed records should hash by value` — recursion through the new branch.
- [x] `nested typed record and nested RecordMap should hash equally`

`./gradlew :nf-commons:test --tests "nextflow.util.HashBuilderTest"` — 17/17 pass.
`CacheHelperTest`, `RecordMapTest`, `TypeHelperTest` — unaffected.

### DIFF
--- a/modules/nf-commons/src/main/nextflow/util/HashBuilder.java
+++ b/modules/nf-commons/src/main/nextflow/util/HashBuilder.java
@@ -19,6 +19,8 @@ package nextflow.util;
 import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -27,6 +29,7 @@ import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
@@ -47,6 +50,7 @@ import nextflow.extension.Bolts;
 import nextflow.extension.FilesEx;
 import nextflow.io.SerializableMarker;
 import nextflow.script.types.Bag;
+import nextflow.script.types.Record;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import static nextflow.Const.DEFAULT_ROOT;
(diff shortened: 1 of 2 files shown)


modules/nf-commons/src/main/nextflow/util/HashBuilder.java
```diff
@@ -152,6 +156,9 @@ else if( value instanceof CacheFunnel )
         else if( value instanceof Map )
             hashUnorderedCollection(hasher, ((Map) value).entrySet(), mode);
 
+        else if( value instanceof Record )
+            hashUnorderedCollection(hasher, recordToMap((Record) value).entrySet(), mode);
```
bentsherman, line 160: Every record instance is a RecordMap, so it should be caught by the `Map` case above

modules/nf-commons/src/test/nextflow/util/HashBuilderTest.groovy
```diff
@@ -154,6 +173,78 @@ class HashBuilderTest extends Specification {
         hash1.hash() == hash2.hash()
     }
 
+    def 'should hash typed record by field values, independent of JVM identity'() {
+        given:
+        def r1 = new SampleRecord('alpha', 1)
+        def r2 = new SampleRecord('alpha', 1)
+        def r3 = new SampleRecord('beta', 1)
+        def r4 = new SampleRecord('alpha', 2)
```
bentsherman, line 181: Records cannot be created via constructor. But it may be that this happens to work and is only disallowed by the type checker

bentsherman:
Closing in favor of #7185

PR #7185, opened 2026-05-28
Fix cast and instanceof with record types

bentsherman:
This PR fixes the use of casting records to a named record type, for example:

```groovy
workflow {
    record(id: '1', fastq: file('1.fastq')) as Sample
}

record Sample {
    id: String
    fastq: Path
}
```

Casting a record to a named record type should be a no-op at runtime. It is only used by the type checker to validate that the record satisfies the requirements of the record type.

If the cast is is not removed at runtime, the record would be cast to an instance of `Sample`, since record types are compiled to actual classes at runtime with an implicit constructor. This would break the runtime's assumption that all records are `RecordMap`s.

### DIFF
--- a/modules/nf-lang/src/main/java/nextflow/script/control/StripTypesVisitor.java
+++ b/modules/nf-lang/src/main/java/nextflow/script/control/StripTypesVisitor.java
@@ -70,7 +70,7 @@ public class StripTypesVisitor extends ClassCodeExpressionTransformer {
         // Erase record type parameters so that records (with type Record)
         // can be dispatched to these methods at runtime
         for( var param : node.getParameters() ) {
-            if( param.getType().redirect() instanceof RecordNode )
+            if( isNamedRecordType(param.getType()) )
                 param.setType(ClassHelper.dynamicType());
         }
         super.visitMethod(node);
@@ -95,17 +95,25 @@ public class StripTypesVisitor extends ClassCodeExpressionTransformer {
     }
 
     private Expression stripTypeAnnotation(CastExpression node) {
-        return STRIP_TYPES.contains(node.getType())
+        return shouldStripType(node.getType())
             ? node.getExpression()
             : node;
     }
 
     private Expression stripTypeAnnotation(DeclarationExpression node) {
         if( node.getLeftExpression() instanceof VariableExpression ve ) {
-            if( STRIP_TYPES.contains(ve.getType()) )
+            if( shouldStripType(ve.getType()) )
                 node.setLeftExpression(new VariableExpression(ve.getName()));
         }
         return node;
+    }
+
+    private boolean shouldStripType(ClassNode type) {
+        return STRIP_TYPES.contains(type) || isNamedRecordType(type);
+    }
+
+    private boolean isNamedRecordType(ClassNode type) {
+        return type.redirect() instanceof RecordNode;
     }
 
 }
--- a/modules/nf-lang/src/main/java/nextflow/script/control/StripTypesVisitor.java
+++ b/modules/nf-lang/src/main/java/nextflow/script/control/StripTypesVisitor.java
@@ -113,13 +113,17 @@ public class StripTypesVisitor extends ClassCodeExpressionTransformer {
 
     private Expression transformCast(CastExpression node) {
         var type = node.getType();
+        if( STRIP_TYPES.contains(node.getType()) ) {
+            return node.getExpression();
+        }
         if( type.getGenericsTypes() != null ) {
             var fn = parameterizedType(type);
             return callThisX("_as_type", args(transform(node.getExpression()), classX(fn.getDeclaringClass()), constX(fn.getName())));
         }
-        else {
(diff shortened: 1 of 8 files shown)


jorgee:
In my opinion, this behaviour is strange or at least incomplete; a user could do something like this:

```
record Sample{
	id:String
	num: Integer
}

workflow{
	def a = record(id: "hello", num: 1) as Sample
	if (a instanceof Sample){
		println("It is a sample")
	}else{
		println("not a sample")
	}
}
```
 
You create a record, cast it to `Sample` and then check the type with `instanceof` and see that it is not really a `Sample`. Not sure, if it is possible to replace the `instanceof` by a method that checks the defined fields instead of the class. At least we should add a note describing it in the docs. (I haven't found anything about it)

bentsherman:
Good catch. I actually meant to implement a compiler transform for that as well, so that `sample instanceof Sample` becomes something like `_instanceof_record(sample, Sample)` which performs a runtime check. I guess that fell through the cracks during development

I will try to implement that in this PR

bentsherman:
@jorgee added the runtime check for `instanceof`, ready for review

modules/nf-lang/src/main/java/nextflow/script/control/StripTypesVisitor.java
```diff
@@ -94,18 +111,78 @@ public Expression transform(Expression node) {
         return super.transform(node);
     }
 
-    private Expression stripTypeAnnotation(CastExpression node) {
-        return STRIP_TYPES.contains(node.getType())
-            ? node.getExpression()
-            : node;
+    private Expression transformCast(CastExpression node) {
```
jorgee, line 114: Previously, only `Channel/Record/Tuple/Value` casts were intercepted; everything else stayed a native Groovy cast. Now every CastExpression is routed through _as_type → TypeHelper.asType, and asType has this branch:

 ```
 if( type == Path )
      return TypeHelper.asPathType(value.toString()) 
```  
It throws `AbortOperationException` if `!Files.exists()`

So in the new language, `'foo.txt' as Path` now throws if the file doesn't exist, whereas a native cast just constructs the `Path`. This is a meaningful, user-visible semantic change that's a side effect of broadening cast interception.

**Comment by jorgee** [comment]
The new cast that redirects to `asType` is breaking a test that is casting a list to a Tuple.

```
Expected no exception to be thrown, but got 'nextflow.exception.ScriptCompilationException'
Expected no exception to be thrown, but got 'nextflow.exception.ScriptCompilationException'
	at spock.lang.Specification.noExceptionThrown(Specification.java:136)
	at nextflow.script.parser.v2.ScriptLoaderV2Test.should strip unsupported type annotations(ScriptLoaderV2Test.groovy:302)
Caused by: nextflow.exception.ScriptCompilationException: Script compilation failed
	at nextflow.script.parser.v2.ScriptLoaderV2.parse0(ScriptLoaderV2.groovy:129)
	at nextflow.script.parser.v2.ScriptLoaderV2.parse(ScriptLoaderV2.groovy:90)
	at nextflow.script.parser.v2.ScriptLoaderV2Test.should strip unsupported type annotations(ScriptLoaderV2Test.groovy:298)
Caused by: org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
Main: 3: The class nextflow.script.types.Tuple<java.lang.String, java.lang.String, java.lang.String> (supplied with 3 type parameters) refers to the class nextflow.script.types.Tuple which takes no parameters
 @ line 3, column 44.
   '1', '1.fastq', '2.fastq'] as Tuple<Stri
```
                                 ^

bentsherman:
Good catches. I was wondering about this too, whether to override all cast expressions or not. I updated the cast visitor to preserve the existing behavior and only inject the runtime check for things like `as Sample` and `as List<Sample>`

jorgee:
I think the check of the existing `Path` is still happening in the case of generics. `['no-existing-file.txt'] as List<Path>` is throwing the exeception. Maybe is out of the scope of the PR, but I am wondering why the cast of `TypeHelper `must check the existence of the file.

bentsherman:
Some users have asked to make `checkIfExists: true` the default in the `file()` function, because this is almost always what you want (i.e. reading a file), whereas you would only want `checkIfExists: false` when writing a new file. But we can't realistically do that because `file()` is used everywhere. It would be a big breaking change

Instead, I introduced a new default with the `params` block, where `Path` params verify the file existence, and new files must be declared as `String`. So the purpose of checking file existence is to verify the existence of input files before they are read. An optional file should be `null` rather than a reference to a missing file.

The `as` operator should have the same behavior as the `params` block since it has the same purpose (validating/converting external data to Nextflow types). The only difference being that `params` can cast command-line params from string to number/boolean/path.

Since this PR is meant to be a bug fix for 26.04, I restricted the changes to only affect new use cases (parameterized types and record types). But we should probably also apply this behavior to `as Path` in the future. I would consider it an aspect of the strict parser since it is just a stricter validation of what users are already doing.

jorgee (approved):

PR #7190, opened 2026-05-29
Improve Tower telemetry error handling on transient gateway failures

pditommaso:
## Problem

A pipeline run aborted because the Seqera Platform **progress/telemetry** endpoint returned a transient **HTTP 502 Bad Gateway**. From the user's perspective the run "stopped quietly": the failure reason was hard to find and the run gave up after only a few seconds of retries.

Investigation of the run log surfaced three distinct issues:

1. **The reason was an unreadable HTML dump.** `TowerClient` could not parse the gateway's HTML error page as JSON, so it surfaced the entire `<html>…502 Bad Gateway…</html>` markup as the `AbortRunException` message.
2. **The retry window was too short.** The retry policy gave up in ~5 seconds (5 attempts) on a transient 502 that is usually short-lived.
3. *(Discovered while testing)* **`AuthCommandImplTest` was hitting the real network** and could run for >6 minutes.

### 1. Surface a concise reason for HTTP gateway errors
`TowerClient.parseCause` now reduces an HTML error body to a concise reason by extracting the `<title>` text (gateways/proxies put the status there, e.g. `502 Bad Gateway`); the match tolerates attributes and collapses internal whitespace. Non-HTML bodies fall back to the plain text as-is, and JSON error objects are unchanged.

Resulting abort message:
```
Unexpected HTTP response
- endpoint    : https://cloud.seqera.io/api/trace/<id>/progress?workspaceId=<id>
- status code : 502
- response msg: 502 Bad Gateway
```

### 2. Extend the retry window to ~3 minutes
Raise the Tower retry policy default `maxAttempts` from 5 to 10. Combined with the existing `delay` (350ms), `multiplier` (2.0) and `maxDelay` (90s), the 9 exponential backoff gaps span a retry window of ~3 minutes (`0.35·(2^9−1) ≈ 178.9s`), so transient gateway errors are ridden out before the run fails.

### 3. Make `AuthCommandImplTest` hermetic
Several `collectStatus` test cases stubbed `checkApiConnection` but did not isolate `SysEnv` nor stub `createTowerClient`. Because `collectStatus` calls `createTowerClient(endpoint, accessToken).getUserInfo()` when a token is present, these tests picked up a real `TOWER_ACCESS_TOKEN` from the developer's environment and made live network calls — one against `https://unreachable.example.com`, which (after extending the retry window) hung ~3 minutes. Isolating the environment with `SysEnv.push([:])`/`pop()` drops the class runtime from **>6.5 min to ~8s**.

## Testing
- `TowerClientTest` — 30 tests, incl. new cases for HTML→reason extraction (attributes, multiline title, plain-text fallback) and the `traceProgress` 502 abort path.
- `TowerRetryPolicyTest` — asserts the new default and that the computed backoff window is ~3 min.
- `AuthCommandImplTest` — 58 tests, now ~8s.

### DIFF
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerRetryPolicy.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerRetryPolicy.groovy
@@ -42,6 +42,14 @@ import nextflow.util.RetryConfig
  */
 class TowerRetryPolicy implements Retryable.Config, ConfigScope {
 
+    /**
+     * Default max number of attempts. Combined with the default {@code delay} (350ms),
+     * {@code multiplier} (2.0) and {@code maxDelay} (90s), the 9 exponential backoff gaps
+     * span a retry window of about 3 minutes, so that transient gateway errors (e.g. a
+     * `502 Bad Gateway`) can be ridden out before aborting the run.
+     */
+    static final int DEFAULT_MAX_ATTEMPTS = 10
+
     @ConfigOption
     @Description("""
         Initial delay before retrying a failed Tower operation (default: `350ms`).
@@ -56,7 +64,7 @@ class TowerRetryPolicy implements Retryable.Config, ConfigScope {
 
     @ConfigOption
     @Description("""
-        Maximum number of retry attempts for Tower operations (default: `5`).
+        Maximum number of retry attempts for Tower operations (default: `10`).
     """)
     int maxAttempts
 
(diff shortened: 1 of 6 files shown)


plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerRetryPolicy.groovy
```diff
@@ -42,6 +42,14 @@ import nextflow.util.RetryConfig
  */
 class TowerRetryPolicy implements Retryable.Config, ConfigScope {
 
+    /**
+     * Default max number of attempts. Combined with the default {@code delay} (350ms),
+     * {@code multiplier} (2.0) and {@code maxDelay} (90s), the 9 exponential backoff gaps
+     * span a retry window of about 3 minutes, so that transient gateway errors (e.g. a
+     * `502 Bad Gateway`) can be ridden out before aborting the run.
+     */
+    static final int DEFAULT_MAX_ATTEMPTS = 10
```
bentsherman, line 51: The description says that the 5 attempts took only 5 seconds... this seems like a really short window for 5 attempts. I wonder if it would be better to increase the initial delay instead

pditommaso: That's good to recover quickly a temporary glitch. Network errors can have very different nature

bentsherman (approved):

PR #7191, opened 2026-05-30
Fix issue with workflow completion handler being called twice

bentsherman:
This PR fixes an issue (reported internally) where the `onComplete` handler is called twice when trace logging is enabled

When a closure is rendered in a GString, it is actually invoked:

```groovy
hook = { -> println 'hello!' }
println "$hook"
```

```
$ groovy test.groovy
hello!
null
```

As a result, we need to be careful about printing closures directly. This PR fixes the log messages in `Session::shutdown0()` and adds a test to confirm it

### DIFF
--- a/modules/nextflow/src/testFixtures/groovy/test/TestHelper.groovy
+++ b/modules/nextflow/src/testFixtures/groovy/test/TestHelper.groovy
@@ -132,7 +132,7 @@ class TestHelper {
         output.readLines().findAll { line ->
             !line.contains('DEBUG') &&
             !line.contains('INFO') &&
-            (includeWarn || !line.contains('WARN')) &&
+            (!includeWarn || !line.contains('WARN')) &&
             !line.contains('plugin') &&
             !line.contains('-> [')  // pf4j dependency graph lines
         }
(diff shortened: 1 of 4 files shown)


pditommaso:
Groovy craziness !

modules/nextflow/src/testFixtures/groovy/test/TestHelper.groovy
```diff
@@ -132,7 +132,7 @@ class TestHelper {
         output.readLines().findAll { line ->
             !line.contains('DEBUG') &&
             !line.contains('INFO') &&
-            (!includeWarn || !line.contains('WARN')) &&
+            (includeWarn || !line.contains('WARN')) &&
```
jorgee: This change now contradicts the comment. It still says: Whether to **also filter out** WARN lines (default: false)

It is also the cause of the failing test LinPathTest:91; It should include `includeWarn=true`. Two tests where filterLogNoise is call with` includeWarn=true` now do the opposite of before — CmdLineageTest:106 and LinCommandImplTest:118, both asserting stdout.size() == 1. Previously true suppressed WARN; now it retains WARN. These tests presumably pass because those runs emit no WARN lines.

jorgee (changes requested):

modules/nextflow/src/testFixtures/groovy/test/TestHelper.groovy
bentsherman: okay, I must have misread "includeWarn" to mean "include in the result" vs "include in the filtering". I will juts rename the variable to be more clear

jorgee (approved):

PR #7193, opened 2026-06-01
Use a stop flag instead of thread interrupt to shut down the logs checkpoint

pditommaso:
## Rationale

`LogsCheckpoint` runs a background thread that periodically uploads the
`.nextflow.log`, timeline and report files to Seqera Platform. The thread was
stopped by calling `thread.interrupt()` (guarded by a lock) and using the
**interrupt status as the loop-exit signal**.

Using interrupt this way is the wrong tool for graceful shutdown:

- **Wrong semantics.** `Thread.interrupt()` is the JVM's *cancellation* primitive
  for unblocking a thread stuck in a blocking call — not a "please wind down when
  convenient" signal. Here it was overloaded to also mean "exit the loop", mixing
  the wake-up mechanism with the control decision.
- **Fragile signal.** Interrupt is a single shared status bit. `saveFiles()`
  ultimately calls into provider/cloud SDK code (`FileHelper.copyPath`); any such
  code that catches `InterruptedException` and clears the flag without re-asserting
  it could **swallow the stop signal**, leaving the thread running.
- **Implicit coupling.** Correctness relied on a non-obvious invariant: interrupt
  was only ever called while holding `lock`, and `saveFiles()` also ran under
  `lock`, so the interrupt could never land mid-upload. Easy to break in a later edit.

## What changed

Replace interrupt-as-control-signal with an explicit `volatile boolean stopped`
flag coordinated through the **existing intrinsic monitor** (no new concurrency
primitives):

- **`stop()`** sets `stopped = true`, `notifyAll()`s to wake the parked thread,
  then `thread.join()`s. The join is deliberately **outside** the `synchronized`
  block — joining while holding the monitor would deadlock, because the woken
  thread must re-acquire the lock to return from `wait()` and exit.
- **`run()`** parks in `lock.wait(interval)` and the loop condition is the flag
  (`while(!stopped)` + post-wait `if(stopped) break`). The `stopped` check is
  co-located with `wait()` under the same lock, so there is **no lost-wakeup race**.
- **Interrupt is now only handled defensively.** It is no longer a control signal;
  a single `catch(InterruptedException)` around the loop ensures that if the JVM or
  external code interrupts the thread (e.g. on abrupt shutdown) it still terminates
  cleanly, logs at debug, and re-asserts the interrupt flag — instead of leaking an
  uncaught exception out of the thread.
- **`stop()` hardening:** guarded against a `null` thread (if `onFlowCreate` failed
  before starting it) and against repeated invocation (both `onFlowError` and
  `onFlowComplete` can fire), making shutdown idempotent.
- Removed the now-redundant `await()` helper.

## Behaviour

Unchanged from the caller's perspective: shutdown still waits for any in-flight
`saveFiles()` cycle to finish (`stop()` blocks on the monitor / `join()`), and the
loop still breaks before starting a new save — no new final flush was introduced.

## Tests

Added lifecycle coverage to `LogsCheckpointTest`:
- start the thread then `onFlowComplete()` → thread terminates;
- `onFlowComplete()` with no thread started → no exception.

### DIFF
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsCheckpoint.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsCheckpoint.groovy
@@ -39,6 +39,7 @@ class LogsCheckpoint implements TraceObserverV2 {
     private Thread thread
     private Duration interval
     private LogsHandler handler
+    private volatile boolean stopped
     private final Object lock = new Object()
 
     @Override
@@ -56,16 +57,23 @@ class LogsCheckpoint implements TraceObserverV2 {
 
     @Override
     void onFlowComplete() {
-        synchronized(lock) {
-            thread.interrupt()
-        }
-        thread.join()
+        stop()
     }
 
     @Override
     void onFlowError(TaskEvent event) {
+        stop()
+    }
+
+    protected void stop() {
+        if( thread == null )
+            return
         synchronized(lock) {
-            thread.interrupt()
+            if( stopped )
+                return
+            stopped = true
+            // wake up the checkpoint thread without relying on thread interruption
+            lock.notifyAll()
         }
         thread.join()
     }
@@ -73,29 +81,22 @@ class LogsCheckpoint implements TraceObserverV2 {
     protected void run() {
         log.debug "Starting logs checkpoint thread - interval: ${interval}"
         try {
-            while( true ) {
-                await(interval)
-                if( Thread.currentThread().isInterrupted() )
-                    break
-                synchronized(lock) {
-                    if( Thread.currentThread().isInterrupted() )
+            synchronized(lock) {
+                while( !stopped ) {
+                    // releases the lock and waits until the timeout elapses or stop() notifies
+                    lock.wait(interval.toMillis())
+                    if( stopped )
                         break
                     handler.saveFiles()
                 }
             }
         }
+        catch( InterruptedException e ) {
+            log.debug "Interrupted logs checkpoint thread - cause: ${e.message}"
+            Thread.currentThread().interrupt()
+        }
         finally {
             log.debug "Terminating logs checkpoint thread"
-        }
-    }
-
-    protected void await(Duration interval) {
-        try {
-            Thread.sleep(interval.toMillis())
-        }
-        catch (InterruptedException e) {
-            log.debug "Interrupted logs checkpoint thread"
-            Thread.currentThread().interrupt()
         }
     }
 }
--- a/plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsCheckpoint.groovy
+++ b/plugins/nf-tower/src/main/io/seqera/tower/plugin/LogsCheckpoint.groovy
@@ -92,14 +92,21 @@ class LogsCheckpoint implements TraceObserverV2 {
     protected void run() {
         log.debug "Starting logs checkpoint thread - interval: ${interval}"
         try {
-            synchronized(lock) {
-                while( !stopped ) {
-                    // releases the lock and waits until the timeout elapses or stop() notifies
-                    lock.wait(interval.toMillis())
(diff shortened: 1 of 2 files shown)


jorgee:
Flagging an overlap with #7188, which touches the same class for what turns out to be a related but distinct goal.

This PR cleanly fixes the *signaling* mechanism — dropping interrupt-as-control-signal in favor of a `stopped` flag is the right call, and the lost-wakeup reasoning is sound. But it preserves the two properties that cause the zombie-head-job hang reported in #7188:

1. **`saveFiles()` still runs while holding `lock`:**

   ```groovy
   synchronized(lock) {
       while( !stopped ) {
           lock.wait(interval.toMillis())
           if( stopped ) break
           handler.saveFiles()   // holds lock for the whole upload
       }
   }
   ```

2. **`thread.join()` is still unbounded**, and `stop()` must acquire `lock` to signal:

   ```groovy
   synchronized(lock) {          // blocks here if the worker is hung in saveFiles()
       if( stopped ) return
       stopped = true
       lock.notifyAll()
   }
   thread.join()                 // no timeout
   ```

So if a cloud upload stalls on a half-open socket inside `saveFiles()`, the worker holds `lock`, `stop()` blocks trying to acquire it, and even past that the unbounded `join()` blocks forever. The worker being a daemon doesn't help, because it's the non-daemon shutdown thread parked in `onFlowComplete()` → `join()` that keeps the JVM alive. The description's note that "shutdown still waits for any in-flight `saveFiles()` cycle to finish" is exactly this failure mode.

#7188 addresses the liveness side: it moves `saveFiles()` out of any lock and bounds the join with a configurable `terminateTimeout` (default 120s), abandoning a genuinely-hung daemon worker so the head job exits. It also includes a regression test where `saveFiles()` blocks forever and shutdown still returns within the timeout — the scenario that triggered the incident, which isn't covered here.

Suggest reconciling the two: either land #7188 for the hang fix, or, if the monitor-based design is preferred here, fold in moving `saveFiles()` outside the lock and adding a bounded join — otherwise the shutdown deadlock remains.

Minor: `stopped` is read/written only under `synchronized(lock)`, so the `volatile` modifier is redundant.

pditommaso:
I'd argue that very hard to have a sensible timeout for thead.join since it takes the overall run executions that could span from mins to days

jorgee:
But, in this case the join is just waiting for the copy of the logs, not the whole execution.

pditommaso:
Got your point, think timeout makes sense

pditommaso:
Thanks for the thorough review @jorgee — all three points were spot on. I've folded the liveness fix into this PR rather than landing #7188 separately, keeping the monitor-based design. Pushed in 96a954de3, 07880287f, 480614c2d:

1. **`saveFiles()` out of the lock** (07880287f) — the monitor is now held only for the `wait()` handshake; the upload runs outside it. So `stop()` can always acquire `lock` to signal, even while an upload is in flight. The `if(stopped) break` guard stays co-located with `wait()` under the same lock acquisition, so the lost-wakeup safety is preserved.
2. **Bounded join** (96a954de3) — `stop()` now `join`s for a configurable `terminateTimeout` (config `tower.logs.checkpoint.terminateTimeout`, env `TOWER_LOGS_CHECKPOINT_TERMINATE_TIMEOUT`, default `120s`, matching #7188's keys). On timeout the daemon worker is abandoned with a warning, so it can't keep the JVM alive.
3. **Regression test** (480614c2d) — `saveFiles()` blocks forever; asserts `onFlowComplete()` returns within the timeout and the daemon worker is abandoned.

On the `volatile` note: it's no longer redundant after (1) — `stopped` is now read outside the lock (the loop condition and the post-`wait` check), so the modifier is required. Kept intentionally.

jorgee (approved):
Looks fine to me. It also replaces #7188

pditommaso:
Thanks!

PR #7196, opened 2026-06-02
ADR: nextflow tree command

edmundmiller:
## Summary
- add a draft ADR for an agent-friendly `nextflow tree` command
- define the concise workflow/subworkflow/module outline format
- cover aliases, unresolved includes/calls, duplicates, partial parse failures, and JSON/DOT output

## Testing
- not run; documentation-only change

### DIFF
--- a/adr/20260601-nextflow-tree.md
+++ b/adr/20260601-nextflow-tree.md
@@ -1,0 +1,398 @@
+# Agent-friendly project outline with `nextflow tree`
+
+- Authors: Edmund Miller, OpenAI
+- Status: draft
+- Date: 2026-06-01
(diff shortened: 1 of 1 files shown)


bentsherman:
This could also be a plugin command:

```bash
nextflow plugin nf-tree:run [options]
```

A plugin command can do anything, so just re-use the parsing infra from the lint command.

See also: https://github.com/nextflow-io/language-server/blob/main/src/main/java/nextflow/lsp/services/script/WorkspacePreviewProvider.java

(the LSP backend for rendering the project tree in VS Code)

bentsherman:
Make it a plugin command

PR #7197, opened 2026-06-02
`nextflow plugins -help`: show subcommands

ewels:
Show the available subcommands for `nextflow plugins` in the CLI help.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdPlugin.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdPlugin.groovy
@@ -35,7 +35,7 @@ import org.eclipse.jgit.api.Git
  */
 @CompileStatic
 @Parameters(commandDescription = "Execute plugin-specific commands")
-class CmdPlugin extends CmdBase {
+class CmdPlugin extends CmdBase implements UsageAware {
 
     @Override
     String getName() {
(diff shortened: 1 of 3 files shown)


modules/nextflow/src/main/groovy/nextflow/cli/CmdPlugin.groovy
```diff
@@ -51,10 +51,61 @@ class CmdPlugin extends CmdBase {
...
+            }
+        }
+        println result.join('\n').toString()
+    }
+
     @Override
     void run() {
-        if( !args )
-            throw new AbortOperationException("Missing plugin command - usage: nextflow plugin install <pluginId,..>")
+        if( !args ) {
```
jorgee, line 121: I think we should print the reason of the failure before printing the usage

modules/nextflow/src/main/groovy/nextflow/cli/CmdPlugin.groovy
```diff
@@ -51,10 +51,61 @@ class CmdPlugin extends CmdBase {
...
+                    break
+                case 'create':
+                    result << 'Create a new plugin project from the official template'
+                    result << 'Usage: nextflow plugin create [<plugin name> <provider name> [project path]]'
+                    result << ''
+                    result << 'When no arguments are provided, the command prompts interactively for the required values.'
+                    result << ''
+                    break
+                default:
+                    throw new AbortOperationException("Unknown plugin sub-command: ${args[0]}")
```
pditommaso: The `default` branch aborts on the `<plugin-name>:<command>` form, which is the third invocation advertised in the top-level help just above (and arguably the primary purpose of `nextflow plugin`).

For example `nextflow plugin nf-hello:greet -help` parses `args = ['nf-hello:greet']`, `command.help` is set, so `Launcher.checkForHelp()` calls `usage()` → `usage(args)`. `args[0]` is neither `install` nor `create`, so this `default` throws `Unknown plugin sub-command: nf-hello:greet` — i.e. asking for help on a documented, valid invocation produces an error claiming it does not exist. (Execution in `run()` is fine — it handles the `:` form via `args[0].contains(CMD_SEP)`; only the usage path aborts.)

Suggest falling back to general usage for the `:` form rather than aborting:

```groovy
default:
    if( args[0].contains(CMD_SEP) ) {
        result << 'Execute a plugin-specific command'
        result << 'Usage: nextflow plugin <plugin-name>:<command> [args]'
        result << ''
        result << 'See the documentation of the individual plugin for its available commands.'
        result << ''
    }
    else {
        throw new AbortOperationException("Unknown plugin sub-command: ${args[0]}")
    }
```

This keeps the genuine typo error (e.g. `nextflow plugin instal -help`) while not blowing up on the legitimate plugin-command form. Worth a test pinning the behavior down.

ewels:
Addressed the review feedback in 81db3ae3d:

- `usage()` no longer aborts on the `<plugin-name>:<command>` form — it now shows a plugin-specific usage block, so `nextflow plugin nf-hello:greet -help` works as documented (@pditommaso).
- For a genuine unknown sub-command, the failure reason is now printed *before* the general usage instead of a bare abort (@jorgee).
- Added `CmdPluginUsageTest` pinning the output for each case.

pditommaso (approved):
LGTM. The no-arg `usage(); return` pattern is consistent with the other UsageAware commands (CmdFs, CmdSecret, CmdAuth, CmdModule), and both earlier review comments are addressed.

PR #7203, opened 2026-06-08
Add module directive to lineage TaskRun record

### LINKED ISSUE #7202: Lineage TaskRun record omits cache-determining inputs (module, eval, stub), so 'lineage diff' can't explain those cache misses
### Bug report

The data-lineage docs present `nextflow lineage diff` as a way to understand why `-resume` re-ran a task. However, the lineage `TaskRun` record stores only a subset of the inputs that go into the task cache hash, so for changes in the omitted inputs the diff shows no meaningful difference even though the task was correctly re-executed.

The cache hash key (`TaskHasher.compute()`) includes, among others:

- environment `module` directives (`task.config.getModule()`)
- `eval` output commands (`getOutputEvals()`, added in #5470)
- the stub-run flag

None of these has a corresponding field in the lineage `TaskRun` model (`modules/nf-lineage/src/main/nextflow/lineage/model/v1beta1/TaskRun.groovy`), which stores: `sessionId, name, codeChecksum, script, input, container, conda, spack, architecture, globalVars, binEntries, workflowRun`.

The `module` omission is the clearest: the record captures `conda`, `spack`, `container` and `architecture` but not `module`, even though all are environment inputs to the same hash.

### Steps to reproduce

`main.nf`:

```nextflow
process GREET {
    module 'samtools/1.9'
    input:  val name
    output: path 'out.txt'
    script: """ echo "Hello ${name}" > out.txt """
}
workflow { GREET(Channel.of('world')) }
```

`nextflow.config`:

```groovy
lineage.enabled = true
```

1. `nextflow run main.nf -dump-hashes`
2. Change only `samtools/1.9` to `samtools/1.17`
3. `nextflow run main.nf -resume -dump-hashes` — the task correctly re-runs

`-dump-hashes` pinpoints the change:

```
0a5461eb… [java.lang.String] samtools/1.9     (run 1)
bd743eeb… [java.lang.String] samtools/1.17    (run 2)
```

`nextflow lineage diff <run1-taskrun> <run2-taskrun>` shows only the parent run id changing — nothing about the module:

```
-    "workflowRun": "lid://f54deed4…"
+    "workflowRun": "lid://4c89b0ed…"
```

The task LID *is* the task hash, so the two records necessarily have different LIDs, yet the diff surfaces no field that explains it.

### Expected behaviour

The `TaskRun` record should capture the cache-hash inputs it currently omits, so that comparing two task runs reflects what actually invalidated the cache:

- `module` (sits naturally beside `conda` / `spack` / `container` / `architecture`)
- `eval` output commands
- a stub-run indicator (partially covered today, since `codeChecksum` switches to the stub source under stub runs)

Note this would make `lineage diff` consistent with the cache hash, but not a full substitute for `-dump-hashes`: the record stores resolved values rather than the ordered hash-key list, and lineage `DataPath` checksums are always computed with the default hash mode, so under `deep`/`lenient` `cacheMode` the record can still diverge from the cache decision. `-dump-hashes` remains the authoritative tool.

### Environment

- Nextflow version: confirmed on 26.04.3 (release) and reproduced in `master` source (`TaskHasher` keys on `module`/`eval`/`stub`; the `TaskRun` v1beta1 model has fields for none of them)
- Java version: 21
- Lineage: `lineage.enabled = true`

### Related

- #6250 (documenting the data structure / JSON schema — the model should be complete before a schema is pinned)
- #7159 (generate lineage schema from source — same reason)
- #6545 (precedent: enriching lineage records with extra metadata)
- #4894 (broader "print cache invalidation reason" UX)

A PR adding the `module` field follows; `eval` and `stub` are left for discussion since their representation needs a design decision.

pinin4fjords:
Closes part of #7202.

The task cache hash keys on environment `module` directives (`TaskHasher` adds `task.config.getModule()` to the hash), but the lineage `TaskRun` record had no field for them. As a result, comparing two task runs with `nextflow lineage diff` could not surface a `module` change that invalidated the cache, even though the record already captures the other environment inputs (`conda`, `spack`, `container`, `architecture`).

This adds a `module` field to the `TaskRun` v1beta1 model and populates it in `LinObserver`, alongside the existing environment fields.

### Example

```nextflow
process GREET {
    module 'samtools/1.9'
    input:  val name
    output: path 'out.txt'
    script: """ echo "Hello ${name}" > out.txt """
}
workflow { GREET(Channel.of('world')) }
```

Changing `samtools/1.9` to `samtools/1.17` and re-running with `-resume` correctly re-runs the task. Before this change, `lineage diff` of the two task runs showed only the parent run id changing; now it shows the `module` field changing.

### Scope

This covers the `module` directive only. The `eval` output commands and the stub-run flag (the other cache-hash inputs missing from the record, per #7202) are left out: representing `eval` commands needs a design decision, and the stub flag is partially covered already via `codeChecksum` switching to the stub source. Those remain tracked in #7202.

This makes the record consistent with that cache-hash input; it does not make `lineage diff` a full substitute for `-dump-hashes`, since the record stores resolved values rather than the ordered hash-key list and `DataPath` checksums always use the default hash mode.

### Tests

- `LinEncoderTest` now round-trips a populated `module` field.
- `LinObserverTest`, `LinCommandImplTest` updated for the new field.
- `LinTypeAdapterFactoryTest` confirms records without a `module` field still decode (backward compatible).

### DIFF
--- a/modules/nf-lineage/src/main/nextflow/lineage/model/v1beta1/TaskRun.groovy
+++ b/modules/nf-lineage/src/main/nextflow/lineage/model/v1beta1/TaskRun.groovy
@@ -65,6 +65,10 @@ class TaskRun implements LinSerializable {
      */
     String architecture
     /**
+     * Environment modules used for the task run
+     */
+    List<String> module
+    /**
      * Global variables defined in the task run
      */
     Map globalVars
--- a/modules/nf-lineage/src/main/nextflow/lineage/model/v1beta1/TaskRun.groovy
+++ b/modules/nf-lineage/src/main/nextflow/lineage/model/v1beta1/TaskRun.groovy
@@ -67,7 +67,7 @@ class TaskRun implements LinSerializable {
     /**
      * Environment modules used for the task run
      */
-    List<String> module
+    List<String> envModule
     /**
      * Global variables defined in the task run
      */
--- a/modules/nf-lineage/src/test/nextflow/lineage/serde/LinEncoderTest.groovy
+++ b/modules/nf-lineage/src/test/nextflow/lineage/serde/LinEncoderTest.groovy
@@ -204,7 +204,7 @@ class LinEncoderTest extends Specification{
         def taskRun = new TaskRun(
             uniqueId.toString(),"name", new Checksum("78910", "nextflow", "standard"), 'this is a script',
             [new Parameter("String", "param1", "value1")], "container:version", "conda", "spack", "amd64",
-            [a: "A", b: "B"], [new DataPath("path/to/file", new Checksum("78910", "nextflow", "standard"))]
+            ["samtools/1.9"], [a: "A", b: "B"], [new DataPath("path/to/file", new Checksum("78910", "nextflow", "standard"))]
         )
         when:
         def encoded = encoder.encode(taskRun)
@@ -222,6 +222,7 @@ class LinEncoderTest extends Specification{
         result.conda == "conda"
         result.spack == "spack"
         result.architecture == "amd64"
+        result.module == ["samtools/1.9"]
         result.globalVars == [a: "A", b: "B"]
         result.binEntries.size() == 1
         result.binEntries.get(0).path == "path/to/file"
--- a/modules/nf-lineage/src/test/nextflow/lineage/serde/LinEncoderTest.groovy
+++ b/modules/nf-lineage/src/test/nextflow/lineage/serde/LinEncoderTest.groovy
@@ -222,7 +222,7 @@ class LinEncoderTest extends Specification{
         result.conda == "conda"
         result.spack == "spack"
         result.architecture == "amd64"
-        result.module == ["samtools/1.9"]
+        result.envModule == ["samtools/1.9"]
         result.globalVars == [a: "A", b: "B"]
         result.binEntries.size() == 1
         result.binEntries.get(0).path == "path/to/file"
--- a/modules/nextflow/src/test/groovy/nextflow/cli/CmdLineageTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/cli/CmdLineageTest.groovy
@@ -197,7 +197,7 @@ class CmdLineageTest extends Specification {
                 'this is a script',
                 [new Parameter( "val", "sample_id","ggal_gut"),
                 new Parameter("path","reads",["lid://45678/output.txt"])],
-                null, null, null, null, [:],[], null)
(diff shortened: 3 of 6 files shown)


jorgee:
Uses same name as Nextflow modules in #7160

modules/nf-lineage/src/main/nextflow/lineage/model/v1beta1/TaskRun.groovy
```diff
@@ -64,6 +64,10 @@ class TaskRun implements LinSerializable {
      * Architecture defined in the Spack environment used for the task run
      */
     String architecture
+    /**
+     * Environment modules used for the task run
+     */
+    List<String> module
```
jorgee: Maybe better to rename to `envModule` to do not collide with #7160
```suggestion
    List<String> envModule
```

modules/nf-lineage/src/test/nextflow/lineage/serde/LinEncoderTest.groovy
```diff
@@ -222,6 +222,7 @@ class LinEncoderTest extends Specification{
         result.conda == "conda"
         result.spack == "spack"
         result.architecture == "amd64"
+        result.module == ["samtools/1.9"]
```
jorgee:
```suggestion
        result.envModule == ["samtools/1.9"]
```

jorgee (changes requested):
Rename to `envModule`

pditommaso:
Let's keep `envModule` rename in its own PR, likely this https://github.com/nextflow-io/nextflow/pull/7205

pditommaso:
Is this PR not duplicate of https://github.com/nextflow-io/nextflow/pull/7160?

jorgee:
No, this refers to environment modules, and #7160 is about adding the Nextflow module id in the lineage record.

pditommaso:
Umm, not sure it should be included. Lineage is not a replacement for canonical task info

bentsherman:
@pditommaso actually we did say that the TaskRun record should contain all of the task hash components so that `nextflow lineage diff` can show the difference when a task is re-executed

By that logic we should also include environment modules... even though I would hope they were extinct by now...

bentsherman:
Closing as not planned. I think there is little appetite to carry environment modules forward into new features. They have little to no reproducibility guarantees compared to conda / containers / spack. If anything I would rather deprecate the `modules` directive and encourage users to move to one of the other options (and encourage their sys admins to support the same)

PR #7208, opened 2026-06-08
Direct execution for named workflows

bentsherman:
This PR adds the ability to execute named workflows directly via `nextflow run`, similar to module execution via `nextflow module run`.

- Command-line params are mapped to worklfow takes in a similar manner as for modules
- A channel input can be loaded from a CSV/JSON/YAML samplesheet by supplying the file path
- Workflow emits are published as output files, but no output directory is created. Instead, the workflow output JSON refers to output files by their work directory path


bentsherman:
Closing in favor of #7379

PR #7210, opened 2026-06-09
Propagate Seqera Intelligent Compute scheduler run identifier to Platform

jorgee:
## Description

When a workflow runs on **Seqera Intelligent Compute** (the `seqera` executor), the scheduler assigns a **run identifier**. Platform currently has no way to associate that id with its workflow record.

This is the **Nextflow side** of that propagation where it adds two pieces of information to the trace requests Nextflow already sends to Platform:

1. `workflow.scheduler.enabled` — whether the run is scheduler‑managed (config‑derived, mirrors the existing `wave`/`fusion` flags), on **begin/complete**.
2. `schedulerRunId` — the scheduler run id, as a top‑level field on **progress** requests.

**Platform does not consume these yet** — that's a separate workstream. Until then Platform ignores the new fields (it ignores unknown JSON properties), so this change is safe to ship ahead of the Platform side.

## How the run id is obtained and propagated

- `SeqeraExecutor.createRun()` assigns the run id on the **first task submission** and publishes it to `WorkflowMetadata.scheduler` (a new lazily‑created field, cross‑thread `volatile`).
- The `enabled` flag is **config‑derived** (`process.executor` → `executor.name` → `NXF_EXECUTOR`, compared to `seqera`) because the executor is not yet registered at create/begin time. It rides on the `workflow` object via `WorkflowMetadata.toMap()`, exactly like `wave`/`fusion`.
- The run id ships on **progress only**: the first task submission both assigns the id and triggers the first progress request, so a progress record always carries it. Heartbeats only fire when there are no pending tasks (before the first task — no id yet; or during a lull — already delivered by progress), so a heartbeat copy would be redundant.
- A new `SchedulerMetadata.toMap()` exposes only `enabled`; `runId` is excluded at the source via a `TowerJsonGenerator` converter (so it never leaks into the begin/complete `workflow` object). `nf-lineage` uses the same `toMap()`, keeping the unset run id out of the lineage record/hash.

## Example: JSON requests sent to Platform

For a scheduler‑managed run (`process.executor = 'seqera'`), with Wave and Fusion enabled.

### `PUT /trace/{workflowId}/begin` — new `scheduler` block on the workflow object

```diff
 {
   "workflow": {
     "runName": "scheduler-test",
     "sessionId": "…",
     "wave":   { "enabled": true },
     "fusion": { "enabled": true, "version": "2.6" },
+    "scheduler": { "enabled": true },
     "...": "..."
   },
   "processNames": ["RNASEQ:FASTQC", "RNASEQ:QUANT", "MULTIQC"],
   "towerLaunch": false,
   "instant": 1750000000000
 }
```

### `PUT /trace/{workflowId}/progress` — new top-level `schedulerRunId`

```diff
 {
   "tasks": [
     { "taskId": 1, "process": "RNASEQ:FASTQC", "status": "RUNNING", "...": "..." }
   ],
   "progress": { "running": 1, "succeeded": 0, "failed": 0, "...": "..." },
   "containers": [],
+  "schedulerRunId": "run-7Hk2…",
   "instant": 1750000000123
 }
```

### `PUT /trace/{workflowId}/complete` — `scheduler.enabled` present, **no** run id

```diff
 {
   "workflow": {
     "runName": "scheduler-test",
     "success": true,
+    "scheduler": { "enabled": true,  runId": "run-7Hk2…" },
     "...": "..."
   },
   "metrics": [ "..." ],
   "progress": { "...": "..." },
   "instant": 1750000009999
 }
```

Notes:
- `workflow.scheduler.runId` is **never** serialized (excluded by `SchedulerMetadata.toMap()` / the `TowerJsonGenerator` converter) — the run id travels only as the top-level `schedulerRunId` on progress.
- For **non‑scheduler** runs, `scheduler` serializes as `{ "enabled": false }` (same as `wave`/`fusion` today) and `schedulerRunId` is omitted from progress entirely.

## Components changed

| Module | Change |
|---|---|
| `modules/nextflow` | new `SchedulerMetadata`; `WorkflowMetadata.scheduler` field + lazy `getScheduler()` |
| `nf-seqera` | `SeqeraExecutor.createRun()` publishes the run id to the scheduler metadata |
| `nf-tower` | `TowerObserver` emits top‑level `schedulerRunId` on progress; `TowerJsonGenerator` registers a `SchedulerMetadata → toMap()` converter |
| `nf-lineage` | `LinObserver.collectWorkflowMetadata()` converts `scheduler` via `toMap()` |
| `adr/` | `20260609-scheduler-run-identifier-propagation.md` |

## Testing

- Unit tests added/updated: `SchedulerMetadataTest`, `TowerObserverTest`, `TowerJsonGeneratorTest`, `SeqeraExecutorTest`.
- Full suites pass: `:nextflow`, `:nf-lineage`, `:plugins:nf-tower`, `:plugins:nf-seqera`.
- Manual stage validation (rnaseq‑nf on Intelligent Compute, scheduler/Platform/Wave on stage) pending.

### DIFF
--- a/adr/20260609-scheduler-run-identifier-propagation.md
+++ b/adr/20260609-scheduler-run-identifier-propagation.md
@@ -6,8 +6,6 @@
 - Date: 2026-06-09
 - Tags: nf-seqera, nf-tower, intelligent-compute, scheduler, platform, trace, accounting
 
-Technical Story: COMP-1817 (dependency of COMP-1705 cost-alignment)
-
 ## Summary
 
 When a workflow runs on the Seqera Intelligent Compute scheduler (the `seqera` executor),
@@ -41,8 +39,7 @@ after completion. Today neither value reaches Platform:
 ## Non-goals
 
 - The Platform-side changes (new `WfSchedulerMeta` embedded entity, DTO fields, persistence in
-  the progress handler, and the cost-retrieval cron). Those are the Platform half of
-  COMP-1817 / COMP-1705.
+  the progress handler, and the cost-retrieval cron). Those are the Platform half of this work.
 - Per-task cost or VM attribution and workspace-level rollups (out of scope for the consumer
   feature's v1).
 
@@ -146,5 +143,5 @@ consuming it.
 
 ## Links
 
-- Consumer feature spec: `seqeralabs/platform` PR #11345 — `specs/260529-cost-alignment/spec.md`
-  (COMP-1705), whose FR-001 requires the scheduler run identifier on the Platform side.
+- The Platform-side cost-alignment feature is the consumer of this change; it requires the
+  scheduler run identifier on the Platform side.
--- a/adr/20260609-scheduler-run-identifier-propagation.md
+++ b/adr/20260609-scheduler-run-identifier-propagation.md
@@ -1,0 +1,150 @@
+# Propagate the Seqera Intelligent Compute scheduler run identifier to Platform
+
+- Authors: Jorge Ejarque
+- Status: draft
+- Deciders: Jorge Ejarque, Paolo Di Tommaso
+- Date: 2026-06-09
+- Tags: nf-seqera, nf-tower, intelligent-compute, scheduler, platform, trace, accounting
+
+Technical Story: COMP-1817 (dependency of COMP-1705 cost-alignment)
+
+## Summary
+
+When a workflow runs on the Seqera Intelligent Compute scheduler (the `seqera` executor),
+the scheduler assigns a run identifier that Platform currently has no way to associate with
+its workflow record. This ADR covers the **Nextflow side**: capturing that run identifier and
+propagating it to Platform on trace requests, plus exposing a config-derived flag that marks a
+run as scheduler-managed.
+
+## Problem Statement
+
+Platform needs to know, per workflow, (a) whether the run was executed via the Intelligent
+Compute scheduler and (b) the scheduler's run identifier. These are required to retrieve
+authoritative cost and resource-usage metrics (settled cost, VM types, etc.) from the scheduler
+after completion. Today neither value reaches Platform:
+
+- The scheduler `runId` is created inside `SeqeraExecutor` (lazily, on the first task
+  submission) and never leaves the executor.
+- There is no "scheduler enabled" marker analogous to the existing Wave/Fusion flags.
+
+## Goals or Decision Drivers
+
+- Make the run identifier available to Platform **as early as possible** — the first task
+  submission both assigns the id and triggers the first progress request, so it is reported at
+  the earliest moment it exists, well before completion.
+- Mark scheduler-managed runs with an `enabled` flag available at workflow create/begin time,
+  mirroring the existing Wave and Fusion metadata.
+- Keep `nextflow` core free of any dependency on the `nf-seqera` plugin.
+- Reuse existing patterns (metadata value classes, the `platform`/`scheduler` shared metadata,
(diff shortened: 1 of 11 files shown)


jorgee:
Some minor comments:

- Regarding naming: I have used `scheduler` and 'schedulerRunId' just for coherence with the `nf-seqera` plugin in the Nextflow codebase. Another option is to rename to a product name such as `intelligentCompute` and `intelligentComputeRunId`
- `SchedulerMetadata` is added directly to `Session.workflowMetadata` for simplicity and to mimic the `wave` and `fusion` behaviour; however, it could also be possible to add it inside `PlatformMetadata`.

Tagging @stefanoboriero

adr/20260609-scheduler-run-identifier-propagation.md
```diff
@@ -0,0 +1,150 @@
...
+  mirroring the existing Wave and Fusion metadata.
+- Keep `nextflow` core free of any dependency on the `nf-seqera` plugin.
+- Reuse existing patterns (metadata value classes, the `platform`/`scheduler` shared metadata,
+  the `TowerJsonGenerator` field-stripping hook) rather than introducing new mechanisms.
+
+## Non-goals
+
+- The Platform-side changes (new `WfSchedulerMeta` embedded entity, DTO fields, persistence in
+  the progress handler, and the cost-retrieval cron). Those are the Platform half of
+  COMP-1817 / COMP-1705.
```
jonmarti: nit: I'd avoid references to Jira tickets since they can expire, be changed, etc.

jonmarti (approved):

adr/20260609-scheduler-run-identifier-propagation.md
jorgee: Thank you. I removed from other places; I missed this part.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerObserver.groovy
```diff
@@ -491,6 +505,9 @@ class TowerObserver implements TraceObserverV2 {
         result.put('tasks', payload)
         result.put('progress', getWorkflowProgress(true))
         result.put('containers', getNewContainers(tasks))
+        final schedulerRunId = getSchedulerRunId()
+        if( schedulerRunId )
+            result.put('schedulerRunId', schedulerRunId)
```
pditommaso, line 510: By doing this, that's sent for all tasks. Therefore the backend needs to duplicate it I guess ?

jorgee: Yes, the`schedulerRunId` is included in all trace records produced after the first task is produced. Not sure how difficult it is to ignore it at the backend once the workflow's scheduler runId is updated. I can't see any existing API call to update the workflow; so, a new update method must be included in the Platform trace API with similar behaviour update behaviour as implemented in the begin/complete calls.

stefanoboriero:
Hey @jorgee there's a proposal to pivot to a different approach by adding a PATCH /workflow/{workflowId} endpoint, where Nextflow could send once the runId as soon as it has it available: https://github.com/seqeralabs/platform/pull/11399#issuecomment-4672206952. This would decouple it from the rest of the workflow lifecycle, and could be applicable for lineageID as well: do you think it's feasible on the Nextflow side?

jorgee:
Yes, this is feasible.

pditommaso:
Closing in favour of new design mentioned above

PR #7211, opened 2026-06-09
Fix lineage records for stub-run and eval outputs

jorgee:
## Problem

Two gaps break lineage tracking, whose goal is to capture everything that distinguishes one task execution from another (ideally everything that feeds the task hash) so differences can be diffed from the lineage records.

1. **Stub-run produces no task lineage records.** Running with `-stub-run` writes only the `WorkflowRun` record; every `TaskRun`, `TaskOutput` and `FileOutput` record is silently lost.

2. **Output eval commands are absent from the lineage record.** Output `eval(...)` commands contribute to the task hash, but the command strings are not stored in the `TaskRun` lineage record. Two executions that differ only in an eval command get different hashes but indistinguishable lineage records.

### Stub-run

`LinObserver.onTaskComplete` does fire for stub tasks, but `storeTaskRun` throws and the exception is swallowed at debug level by `Session.notifyEvent`:

```
groovy.lang.MissingPropertyException: No such property: source for class: nextflow.script.dsl.ProcessDslV1
    at nextflow.processor.TaskRun.getStubSource(TaskRun.groovy:1023)
    at nextflow.lineage.LinObserver.storeTaskRun(LinObserver.groovy:258)
```

`storeTaskRun` computes the code checksum from `task.stubSource` when `session.stubRun` is set. `getStubSource()` accessed `config?.getStubBlock()?.source`. `getStubBlock()` returns a `TaskClosure` (a Groovy `Closure`); accessing `.source` as a **property** routes through `Closure.getProperty`, which resolves the name against the closure's delegate — the `ProcessDslV1` script object — which has no `source` property. `resolveStub` is unaffected because it calls `block.getSource()` as a **method**. Because `WorkflowRun` is written in `onFlowBegin` (before any task completes), it survives while everything from `onTaskComplete` is lost.

### Eval outputs

A normal run already records the eval *result value* in `TaskOutput` (`{"type":"eval","name":"nxf_out_eval_1","value":"hello"}`), but the eval *command* (`echo hello`) — the part that feeds the task hash via `TaskHasher.computeEvalOutputCommands` — is not stored anywhere in the `TaskRun` record. Moreover the hash keys on **both** the eval output name and its command, so the name must be recorded too: otherwise two executions that differ only by eval output name get different hashes but indistinguishable lineage records.

## Solution

- **`TaskRun.getStubSource()`**: call `getStubBlock()?.getSource()` (method) instead of `?.source` (property).
- **Lineage `TaskRun` model**: add a `Map<String,String> eval` field (output name → command, in output-declaration order) populated in `LinObserver.storeTaskRun` from `task.getOutputEvals()`; `null` when the task has no eval outputs. Both the name and the command feed the task hash (`TaskHasher.computeEvalOutputCommands`), so both are recorded. Declaration order is preserved, which is deterministic across executions of the same script, so eval records remain directly comparable.

### Backward compatibility

The lineage models are (de)serialized by Gson via name-based field reflection, not by positional constructor. Records written before this change have no `eval` field and deserialize with `eval == null`; new records add it. The field is appended last, matching the convention for additive fields (e.g. `labels` in `TaskOutput`/`FileOutput`). `v1beta1` is a beta model; an additive optional field does not warrant a version bump.

## Tests

- `TaskRunTest`: `getStubSource()` regression (via a real `TaskClosure` whose owner has no `source` property — verified to fail on the old `.source` and pass on `.getSource()`) plus a null-stub-block case.
- `LinObserverTest`: eval output names and commands are recorded as a map; `null` when there are no evals.
- `tests/lineage-stub-eval.nf` + `.checks`: end-to-end under both `-stub-run` and normal mode, asserting the `TaskRun` record exists, carries `"eval":{"nxf_out_eval_N":"echo hello"}`, and records the correct source code (stub vs. main script).


jorgee:
This PR partially fixes #7202

modules/nf-lineage/src/main/nextflow/lineage/model/v1beta1/TaskRun.groovy
```diff
@@ -76,4 +76,9 @@ class TaskRun implements LinSerializable {
      * Workflow run associated to the task run
      */
     String workflowRun
+    /**
+     * Output eval commands executed by the task run, mapped by output name.
+     * Both the name and the command feed the task hash, so both are recorded.
+     */
+    Map<String,String> eval
```
bentsherman: can you place this field after `script` instead?

modules/nf-lineage/src/test/nextflow/lineage/LinObserverTest.groovy
```diff
@@ -715,6 +715,100 @@ class LinObserverTest extends Specification {
...
+        observer.storeTaskRun(task, observer.normalizer)
+        def result = store.load(hash.toString()) as nextflow.lineage.model.v1beta1.TaskRun
+        then:
+        result.eval == [eval1: 'echo one', eval2: 'echo two']
+
+        cleanup:
+        folder?.deleteDir()
+    }
+
+    def 'should store null eval when task has no output evals' () {
```
bentsherman: this test seems unnecessary, I assume there is already a test for a minimal task run that covers this

tests/checks/lineage-stub-eval.nf/.checks
bentsherman, line 1: don't think this case is worth an e2e test

bentsherman (changes requested):
Looks good overall, just a few minor change requests

modules/nf-lineage/src/main/nextflow/lineage/model/v1beta1/TaskRun.groovy
jorgee: Moved the field after `script` and updated the positional constructor accordingly.

modules/nf-lineage/src/test/nextflow/lineage/LinObserverTest.groovy
jorgee: Removed. The 'should store task run with...' test builds the expected record without an eval value, so eval == null is already covered for a minimal task run.

tests/checks/lineage-stub-eval.nf/.checks
jorgee: Removed the e2e test. The stub-source fix is covered by TaskRunTest and the eval recording by LinObserverTest.

bentsherman (approved):

modules/nf-lineage/src/main/nextflow/lineage/model/v1beta1/TaskRun.groovy
```diff
@@ -44,6 +44,11 @@ class TaskRun implements LinSerializable {
      * Resolved task script
      */
     String script
+    /**
+     * Output eval commands executed by the task run, mapped by output name.
+     * Both the name and the command feed the task hash, so both are recorded.
+     */
+    Map<String,String> eval
```
bentsherman, line 51: @jorgee Please make a PR to update the lineage schema as well

PR #7213, opened 2026-06-10
Pipeline composition

bentsherman:
This PR adds an ADR for remote pipeline inclusion, aka "meta-pipelines".

It describes an approach for including remote pipelines into a meta-pipeline in a way that preserves dataflow concurrency between pipeline inputs/outputs.

It discusses alternative approaches such as pipeline chaining / nf-cascade and why they don't satisfy certain use cases (preserving dataflow concurrency).

It also walks through a basic example of fetchngs -> rnaseq.


ewels:
Great write up, thanks for this Ben!

As you might expect, I'm most concerned about the params. You characterise it as a one-off cost which is mitigated by LLMs, however that doesn't take into account _updates_ to included pipelines (a core functionality with included modules). The `params` drift with updates would be dangerous and a constant source of dev work.

I'd still love to look into how we could bulk import nested config and apply it at root level. Even if it is a separate import + apply mechanism (eg. like config profiles in a sense?). I think without it, the use of the meta pipeline functionality is substantially limited.

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,283 @@
...
+### Best practices for included pipeline
+
+To be importable in practice, a pipeline's core workflow (and its dependent modules/workflows) should be free of external context:
+
+1. No `params` usage outside the entry workflow -- pass values as explicit process/workflow inputs.
+2. No `publishDir` -- use the `output` block.
+3. No use of project-level assets (`projectDir`, `bin`, `lib`) within the core workflow. Module-level assets can be used through the module `resources/` bundle and `moduleDir`.
+4. Declare software dependencies (`container`, `conda`) in the process definition, not in config.
+5. No default `ext` settings in config -- specify these defaults in the process definition or use explicit process inputs. Otherwise, any default `ext` settings must be replicated manually in the meta-pipeline.
+6. No plugin functions within the core workflow.
```
jorgee: No clear about some of these best practices and what's the issue of not following them; maybe could be good to add an example.

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,283 @@
...
+    > results/output-fetchngs-rnaseq.json
+
+# perform RNAseq analysis
+nextflow -q run nf-core/rnaseq \
+    -params-file results/output-fetchngs-rnaseq.json \
+    -output-format json \
+    > results/output-rnaseq.json
+```
+
+While pipeline chaining has always been possible in theory, new language features such as [workflow outputs](20251020-workflow-outputs.md) and [record types](20260306-record-types.md) make it much more practical. Each pipeline can define a structured output which can be passed to the next pipeline via JSON. Mismatches between an upstream output and downstream input (e.g. missing columns, different column names) can be resolved by a small adapter pipeline.
```
pinin4fjords: I remain to be convinced of the point of pipeline chaining _if_ we can trivially make meta pipelines.

adr/20260608-pipeline-composition.md
```diff
@@ -0,0 +1,283 @@
...
+    ch_samples = NFCORE_FETCHNGS( ids )
+
+    // adapt fetchngs output to rnaseq input (add strandedness)
+    ch_samples = ch_samples.map { r ->
+        r + record(strandedness: params.strandedness)
+    }
+
+    // perform RNAseq analysis
+    multiqc_report = NFCORE_RNASEQ( ch_samples )
+
```
jorgee, line 321: Some parts are not clear or can confuse the users and I think it is better to clarify

1- If I am not wrong, when executing a process, an instance of the process is executed per channel element, in the case of named workflow, just a workflow instance is executed passing the channel. What is the model in pipelines? Is it the same as processes, one pipeline instance execution per channel element?

2- Regarding the pipeline channels, is there only one input and one output channel? I suppose these channels require elements that match the `params` and `outputs` format. Are the input channel elements records whose properties match with `param` names? Are the output channel elements auto-generated records similar to what is printed in outputs?

pinin4fjords:
> As you might expect, I'm most concerned about the params. 

Agreed. Feel like we need some sort of auto-import of the params of child workflows, so e.g. they appear automatically in Platform, and I could say e.g. `meta.rnaseq.pseudoaligner = 'kallisto'` in the meta pipeline's nextflow.config to override. 

Then some auto-assembly of docs as well. 

Basically we need to standardise at the nextflow level where a bunch of the non-nextflow pieces need to live.

adr/20260608-remote-pipeline-inclusion.md
adamrtalbot: Plugin support feels like a requirement, functions like a webhook or logging statement could be critical for the workflow. The main challenge might be supporting multiple versions (e.g. WORKFLOW1 uses plugin@1.2.3 and WORKFLOW2 uses plugin@2.4.1), but maybe we can just say "ONE PLUGIN ONLY"

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,283 @@
...
+
+### Best practices for included pipeline
+
+To be importable in practice, a pipeline's core workflow (and its dependent modules/workflows) should be free of external context:
+
+1. No `params` usage outside the entry workflow -- pass values as explicit process/workflow inputs.
+2. No `publishDir` -- use the `output` block.
+3. No use of project-level assets (`projectDir`, `bin`, `lib`) within the core workflow. Module-level assets can be used through the module `resources/` bundle and `moduleDir`.
+4. Declare software dependencies (`container`, `conda`) in the process definition, not in config.
+5. No default `ext` settings in config -- specify these defaults in the process definition or use explicit process inputs. Otherwise, any default `ext` settings must be replicated manually in the meta-pipeline.
```
adamrtalbot: `ext.args` is soooooo powerful, yet clearly breaks the interface contract for processes.

I still think we should promote args to a directive and it will solve a number of these issues (`process.args`) 😉 .

```nextflow
process {
    args "--concise"
    // etc...
}

// main.nf
my_process(ch_inputs, args: "--verbose")

// nextflow.config
process.withName 'my_process' {
    args = "--verbose"
}
```

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,283 @@
...
+6. No plugin functions within the core workflow.
+
+For process directives, it is helpful to distinguish *what* is executed vs *how* it is executed. Directives that affect the *what* (`container`, `ext` settings) should be owned by the process definition. Directives that affect the *how* (`cpus`, `memory`, `executor`, `queue`, `errorStrategy`) should be owned by the meta-pipeline.
```
adamrtalbot: I don't understand the distinction here.

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,283 @@
...
+
+Sourcing remote pipelines from the Nextflow registry implies a `nextflow pipeline` command group for publishing and installing pipelines, similar to `nextflow module`. Even with a Git-based approach, a CLI is likely still needed to install and update remote pipelines.
+
+### Using plugin functions in included pipeline
+
+If an included pipeline uses plugin functions in the core workflow, these plugins must be explicitly declared in the meta-pipeline config, since the included pipeline config is not inherited.
+
+Alternatively, these core plugin dependencies could be specified in the pipeline spec under `requires.plugins`. When installing a pipeline, Nextflow could copy these plugin declarations into the meta-pipeline config and/or spec.
+
+Since this use case is rare -- plugin functions are typically used in the entry workflow outside the core workflow -- it can be deferred in the first iteration.
```
adamrtalbot: With more private plugin registries, I expect more utility methods in plugins (e.g. `updateLims(sampleId, status)`), but maybe this is premature optimization.

adr/20260608-remote-pipeline-inclusion.md
adamrtalbot: Agreed, I think pipeline chaining is used because metapipelines don't work right now. If they did, the number of pipeline chains drops.

That's not to say they're _never_ useful, but it's much less common.

Two main use cases:
- Run major pipeline (sarek, rnaseq) and add a few auxiliary processes
- Daisy chain two pipelines (fetchngs -> rnaseq)

Both are solved better by metapipelines than pipeline chaining.

The main use case for daisy chaining is actually wiring nextflow up to _non_  Nextflow tools, e.g. Nextflow into an ETL system. In this case structured inputs and outputs are still very useful.

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,283 @@
...
+
+The Nextflow-in-Nextflow approach treats the included pipeline as a *black box* -- it preserves the exact pipeline behavior (core workflow + entry workflow + config) while forfeiting dataflow composition (separate dataflow graphs).
+
+An ideal solution might combine the best of both: compose pipelines into a single dataflow graph (white box) while inheriting each pipeline's params, outputs, and config so they need not be replicated (black box). We considered such a model, where an included pipeline contributes its shell as namespaced, overridable defaults, but rejected it. Dataflow composition fundamentally requires exposing the core workflow as a set of channel ports, so the white-box mechanism is unavoidable; inheritance would only layer implicit behavior on top of it. That behavior comes at a steep cost: it relocates a one-time *write* cost (boilerplate) into a recurring *read* cost (hidden defaults, auto-bound arguments, auto-published outputs), burdens every tool that must now understand it (linter, type checker, config resolution, resume), and conflicts with the frozen-island philosophy that otherwise governs vendored code.
```
adamrtalbot: I agree with this. The added complexity is enormous.

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,283 @@
...
+    ids = channel.fromPath(params.input).splitCsv()
+    ch_samples = NFCORE_FETCHNGS( ids )
+
+    // adapt fetchngs output to rnaseq input (add strandedness)
+    ch_samples = ch_samples.map { r ->
+        r + record(strandedness: params.strandedness)
+    }
+
+    // perform RNAseq analysis
+    multiqc_report = NFCORE_RNASEQ( ch_samples )
```
adamrtalbot: Side note - I would remove MultiQC from all nf-core pipelines and put them in the metapipelines, i.e. no MultiQC repeats, but that's a matter of opinion.

```nextflow
FETCHNGS(ch_inputs)
RNASEQ(fetchngs.out)
MULTIQC(RNASEQ.out.qc_files)
```

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,283 @@
...
+
+- **The handoff is a channel, not a file.** The black-box approaches block until fetchngs finishes before rnaseq starts. Here, `ch_samples` is a live channel: rnaseq begins aligning each sample the moment fetchngs emits it. This is the dataflow composition that motivates the meta-pipeline.
+
+- **The adapter is an operator, not a pipeline.** The strandedness gap that required a separate `fetchngs-rnaseq.nf` adapter in the chaining example collapses to a single `map` operator in the meta-pipeline.
+
+- **Params and outputs are replicated, not inherited.** `--input` and `--strandedness` are declared in the meta-pipeline's own `params` block and passed explicitly into the core workflows. Similarly, any outputs must be declared as such in the meta-pipeline's `output` block. The included pipelines do not contribute any of their own params, entry workflows, or output blocks.
+
+**Configuration**
+
+Resource directives -- the *how* -- live in the meta-pipeline config. Because each included pipeline is part of the dataflow graph, its processes can be targeted like normal:
```
adamrtalbot: This sort of thing is really a convention and might not be massively relevant for the ADR. I might make the document focus on what is or isn't possible instead of best practices:

 - If I specify directives in the process, are they respected in the workflow?
 - What if the same process is used in two imported workflows?
 - What if the same process is different in the two workflows?
 - What if a metapipeline author wants to change them?
 - What if a metapipeline user wants to change them?

adr/20260608-remote-pipeline-inclusion.md
bentsherman: Following these guidelines makes it so that when you include the core workflow and its dependent modules/subworkflows, it is self-contained

For example:
- if the core workflow uses project-level assets like `bin` or `lib`, I have to remember to copy them into the meta-pipeline
- if the core workflow uses a param directly and I import that into the meta-pipeline, I have to remember to define the same param (with the same meaning) in the meta-pipeline
- and so on

adr/20260608-remote-pipeline-inclusion.md
bentsherman: at this point the value prop of pipeline chaining appears to be low development overhead (just plug A into B)

adr/20260608-pipeline-composition.md
bentsherman: `NFCORE_FETCHNGS` and `NFCORE_RNASEQ` in this example are just named workflows. they have no special behavior.

the idea is that each included pipeline should define its "core workflow" separate from the params/entry/output trio. then, only the core workflows are imported, the entrypoints are discarded

so, this code runs exactly as you would expect. no hidden behavior. this is partly why I see it as a long-term benefit over the "best of both" approach. that would save you some typing but be harder to read

adr/20260608-remote-pipeline-inclusion.md
bentsherman: plugins for webhooks / logging typically live outside the core workflow. so the meta-pipeline would just import the core workflow logic and decide whether to include those plugins in its own shell

I have yet to see a plugin that is actually used in a workflow's core logic, although it's certainly possible. Most plugins provide third-party integrations at the pipeline boundary

adr/20260608-remote-pipeline-inclusion.md
bentsherman: both `ext.args` and `process.args` can work, as long as the default value for the arg is defined in the process definition rather than in config

the core problem is that when I import a workflow, Nextflow doesn't know which config is "tied" to that workflow

adr/20260608-remote-pipeline-inclusion.md
bentsherman: in other words, some directives affect the task result while others don't

adr/20260608-remote-pipeline-inclusion.md
bentsherman: I was wondering about that. Wasn't sure if you would want a meta-pipeline to produce one multiqc report per pipeline or just one for the whole thing

adr/20260608-remote-pipeline-inclusion.md
bentsherman: I think all of the answers to these questions flow naturally from the fact that the included pipeline is just a workflow + modules/subworkflows in the meta-pipeline repository. Every process from the included pipeline is just another process in the meta-pipeline DAG. So, config works like normal.

bentsherman: Claude did really hook on to this what/how distinction though... probably talked about it more than I would have :sweat_smile:

adr/20260608-pipeline-composition.md
jorgee: Ok, defining a "core workflow" with takes and emtis is a requirement for "importable" pipelines?

adr/20260608-remote-pipeline-inclusion.md
bentsherman: A LIMS integration sounds like something that could live outside the core workflow

adr/20260608-pipeline-composition.md
bentsherman: I updated the appendix example to also define the fetchngs and rnaseq workflows, so that it's clear what is being imported into the meta-pipeline. Hopefully it's clear now that it's just regular pipeline code

adr/20260608-remote-pipeline-inclusion.md
adamrtalbot: I agree but this might become more popular with the plugin registry + vibe coding. 

Sounds like premature optimization by me, easier to just tell people to be careful and deal with it if it's a problem.

adr/20260608-remote-pipeline-inclusion.md
bentsherman: @ewels @pinin4fjords @adamrtalbot

Pulling everyone into this thread to talk about auto-inheritance

> As you might expect, I'm most concerned about the params. You characterise it as a one-off cost which is mitigated by LLMs, however that doesn't take into account updates to included pipelines (a core functionality with included modules). The params drift with updates would be dangerous and a constant source of dev work.

That's fair, but not my main point. The core problem is this -- if you want to preserve dataflow concurrency between pipelines, then you can't really just auto-import params into the meta-pipeline. You have to define which params are replaced with inter-pipeline wiring vs exposed to the top-level. That amounts to just writing the meta-workflow.

The development overhead is what it is. I suggest the AI skill just as an idea. I'm sure it could also handle updates. All of that is better than having loads of hidden behavior that makes the meta-pipeline impossible to reason about

> I'd still love to look into how we could bulk import nested config and apply it at root level. Even if it is a separate import + apply mechanism (eg. like config profiles in a sense?). I think without it, the use of the meta pipeline functionality is substantially limited.

Not sure I understand this point. Most of the config is just standard boilerplate, so it doesn't make sense to auto-import it because you will just get lots of duplicate config

Unless you are talking about `ext` config. That will depend on whether we can move the default `ext` settings into the process definition

bentsherman: Building on what Adam said:

> In a scenario where I update my workflow from v1.1 to v1.2, an update to params should be explicit in the input block, not implicit and I hope it doesn't change too much.

The nice thing about an explicit meta-pipeline definition is that when I update the included pipeline, the linter / language server will immediately pick up on any inconsistencies, because it's just regular code. I'm not sure the tooling would be able to do that if there was a lot of implicit behavior

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,355 @@
...
+
+// pipeline
+include { NFCORE_RNASEQ } from 'nf-core/rnaseq'
+```
+
+Including a remote pipeline is equivalent to including the top-level `main.nf` of that pipeline; any named workflow defined there can be included by name. By convention, the main script defines only the entry workflow and the core workflow (e.g. `NFCORE_RNASEQ` in nf-core/rnaseq). From this point, the included workflow can be called like any other workflow.
+
+When a pipeline is included, it is vendored into the meta-pipeline project under `workflows/<scope>/<name>/`. Included pipelines are isolated -- each included pipeline has its own `modules/` and `workflows/` directories. This way, two pipelines can use different versions of the same module without compromising reproducibility.
+
+Included pipelines should be committed to the meta-pipeline repository. The pipeline should have a *pipeline spec* (`nextflow_spec.json`) which specifies the pipeline version, so that Nextflow can track local changes.
```
bentsherman: @adamrtalbot 

> My main concern here is versioning of imported workflows. Do we include a lock file or something to ensure consistency or just trust in the files that are copied into the workflow code?

See here. Like modules, we will likely want to have some sort of checksum verification (e.g. `.pipeline-info`)

I guess the simplest way would be to commit the entire pipeline, even though only the core workflow will be used. Then you can have a single checksum for the entire pipeline directory

It's probably still useful to keep the pipeline shell in the meta-pipeline repo, since e.g. your agent will want to refer to it when updating the meta-pipeline

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,355 @@
...
+    bams: Channel<Path>
+    counts: Channel<Path>
+}
+```
+
+The meta-pipeline includes the core workflow from each pipeline and composes them into an entry workflow with params and outputs:
+
+```groovy
+include { NFCORE_FETCHNGS } from 'nf-core/fetchngs'
+include { NFCORE_RNASEQ } from 'nf-core/rnaseq'
```
bentsherman: For anyone feeling adventurous, here is what Claude and I came up with while exploring auto-inheritance:

```groovy
include { NFCORE_FETCHNGS } from 'nf-core/fetchngs'
include { NFCORE_RNASEQ } from 'nf-core/rnaseq'

params {
    input: Path // meta entry point
    strandedness: String = 'auto' // one new knob
    // aligner / fasta / ... inherited from rnaseq's params, override on CLI as --rnaseq.fasta=...
}

workflow {
    main:
    ch_ids = channel.fromPath(params.input).splitCsv()
    ch_samples = NFCORE_FETCHNGS( ch_ids )

    ch_samples = samples.map { r -> r + record(strandedness: params.strandedness) }

    // rnaseq.* params automagically passed to rnaseq workflow via named arguments
    NFCORE_RNASEQ( samples: ch_samples )

    // no publish/output blocks: each pipeline's outputs publish under <output-dir>/<pipeline>/
    // question: what if I don't want to publish something (e.g. fetchngs output)?
}
```

Feel free to take it and run with it...

adr/20260608-remote-pipeline-inclusion.md
bentsherman: sure, that's why I call them out as best practices instead of hard rules. you can use a plugin function as long as you remember to declare it in the meta-pipeline config

adr/20260608-remote-pipeline-inclusion.md
pinin4fjords: Well, chaining has development overhead, it's quite a faff, all we have to do is bring meta-pipeline dev under that faff level

adr/20260608-remote-pipeline-inclusion.md
pinin4fjords: Exactly the way I was thinking. We just namespace the children's params

pditommaso (changes requested):
Thanks for putting this together, Ben — the dataflow-composition motivation and the rejection of the runtime-inheritance hybrid are both nicely argued. A few thoughts to share before this moves past `draft`:

**1. The key technical challenge could be expanded.** At its core this proposes a mechanism to include a fully-fledged Nextflow workflow into another, mimicking how we already include modules and sub-workflows. The part I'd love to see fleshed out is *how channels and values get bound into the included workflow's inputs*. The ADR sets out the policy (params live at the top level, the core workflow consumes everything via `take:`) but doesn't yet describe the binding mechanics: how a scalar value vs. a streaming channel is bound at the call site, the value-channel/queue-channel broadcast semantics, and whether a typed `take:` can accept a bare value type like `String`/`Path`. The example here (`take: aligner: String`) also reads a bit differently from the typed-workflows ADR, where every `take:` input is a channel type. Since this binding question largely determines feasibility, it'd be great to work it out explicitly.

**2. The nomenclature can be better shaped.** The document moves between "meta-pipeline" and "remote pipeline", and I think the framing could be sharpened. Terms like **workflow modularisation / workflow inclusion / workflow composition** might describe what's happening (composing one workflow into another) more directly than introducing a new "meta-pipeline" category.

**3. There's some overlap with existing sub-workflow inclusion.** Once you discard the entry workflow, params, and `output` block and import only the core workflow, what's left looks a lot like a sub-workflow. It'd be helpful to clarify how this differs from including a remote sub-workflow, and what the main benefit is that justifies a separate mechanism (separate storage layout, a new `nextflow_spec.json`, a separate CLI, etc.).

**4. A possible framing.** I'd lean toward framing the next step as **enabling remote *sub-workflows*** — the natural progression after remote modules (processes). Module (process) → sub-workflow → composition feels like a clean, incremental story that reuses the conventions we already have, rather than introducing a "pipeline" as a new top-level artifact with its own resolution rules, storage path, and spec file. If we get remote sub-workflow inclusion right, "meta-pipelines" might largely fall out of it as a usage pattern rather than a new concept.

adr/20260608-remote-pipeline-inclusion.md
adamrtalbot: nf-core copy+pastes modules for subworkflows and it works well!

bentsherman:
@pditommaso thanks for the review

> The part I'd love to see fleshed out is _how channels and values get bound into the included workflow's inputs_.

There isn't much to say here because it just works like normal. In the appendix example, `NFCORE_RNASEQ` is just a named workflow. The meta-pipeline calls it the same way that rnaseq would call it. The only difference is that some inputs might come from upstream outputs instead of params.

> ... how a scalar value vs. a streaming channel is bound at the call site, the value-channel/queue-channel broadcast semantics, and whether a typed `take:` can accept a bare value type like `String`/`Path`.

A workflow take can be a channel, a dataflow value, or a regular value. This is how it has always worked

> The document moves between "meta-pipeline" and "remote pipeline", and I think the framing could be sharpened. Terms like **workflow modularisation / workflow inclusion / workflow composition** might describe what's happening (composing one workflow into another) more directly than introducing a new "meta-pipeline" category.

"Meta-pipeline" is the top-line feature that everyone is after, but the only actual new feature proposed by the ADR is "remote pipeline inclusion" -- how to install a pipeline as a component and keep it in sync with the source. This is why the ADR is titled "Remote pipeline inclusion". Once you have that, everything else is just normal workflow composition and convention.

They are distinct concepts -- the ADR does not treat them as interchangeable.

> Once you discard the entry workflow, params, and `output` block and import only the core workflow, what's left looks a lot like a sub-workflow. It'd be helpful to clarify how this differs from including a remote sub-workflow, and what the main benefit is that justifies a separate mechanism (separate storage layout, a new `nextflow_spec.json`, a separate CLI, etc.).

The core workflow looks like a subworkflow because it is a subworkflow :smile: 

The only new thing that we introduce here is installing a pipeline into a project as a component and keeping it in sync with the remote source (either from Git or the registry). For that you likely need a pipeline spec (version, checksum) and a CLI (installing, updating). I just haven't spelled all that out yet because the bigger question right now is how to minimize developer overhead

> I'd lean toward framing the next step as **enabling remote _sub-workflows_** — the natural progression after remote modules (processes). Module (process) → sub-workflow → composition feels like a clean, incremental story that reuses the conventions we already have, rather than introducing a "pipeline" as a new top-level artifact with its own resolution rules, storage path, and spec file. If we get remote sub-workflow inclusion right, "meta-pipelines" might largely fall out of it as a usage pattern rather than a new concept.

Looks like you arrived at the same place as me. Remote workflows are the real feature, meta-pipelines emerge naturally as a convention on top.

I'm not sure whether it's worth trying to distinguish between pipelines / workflows / subworkflows. They're all basically the same thing. Especially if we add the ability to execute named workflows directly (#7208). The difference boils down to boilerplate, which we want to minimize anyway

This is why I just talk about "remote pipeline inclusion", because when I import a workflow, I don't really care whether that workflow is a "pipeline" like rnaseq or a "subworkflow" like `BAM_STATS_SAMTOOLS`. Workflow composition works the same way either way.

Happy to rename the ADR to "remote workflow inclusion" to align with the `workflow` keyword.

ewels:
> My main concern here is versioning of imported workflows. Do we include a lock file or something to ensure consistency or just trust in the files that are copied into the workflow code?

Modules have a `.moduleinfo` file with a hash to allow checking that stuff wasn't modified. I think I saw something similar mentioned here for pipelines / workflows?

ewels:
> Having unpredictable global scope params blocks is just weird and if we were designed Nextflow today we would never include this behaviour. In other languages, globals need to be used with caution and are generally not advised.

@adamrtalbot agreed, I never said _global_. I would love it if the pipeline config is imported within a dedicated scope and treated as a baseline default. Then the import-ing pipeline can override anything, but doesn't need to duplicate config that isn't being changed.

Doing this would not be trivial. The only way I can think of is to do something fairly radical like rendering the config at import time and saving that to a locked config file somewhere. Or some other crazy mechanism.

ewels:
> Config or params? In my mind they are very different concepts, I was referring to parameters here.

Ideally params, but might need to be config for all the `ext` stuff..?

> Happy to rename the ADR to "remote workflow inclusion" to align with the workflow keyword.

Yeah as it stands I think this basically boils down to the functionality we already have with `nf-core subworkflows`, right? Which is quite far from what I think of as meta-pipelines. Still good to have and useful..

bentsherman:
> Yeah as it stands I think this basically boils down to the functionality we already have with `nf-core subworkflows`, right?

Can the nf-core tooling install a workflow from a pipeline repo? e.g. NFCORE_RNASEQ from nf-core/rnaseq? I think that is the main thing that this ADR adds

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,351 @@
...
+### Pipeline inclusion and storage
+
+Modules and pipelines share the same include syntax and naming conventions:
+
+```groovy
+// module
+include { BWA_MEM } from 'nf-core/bwa/mem'
+
+// pipeline
+include { NFCORE_RNASEQ } from 'nf-core/rnaseq'
```
bentsherman: One point that makes me hesitant to reframe the ADR as "remote workflow inclusion" -- here we are referencing the *pipeline* by name (`nf-core/rnaseq`)

It could be the GitHub repo or an entity in the Nextflow registry, but either way, the pipeline itself plays a role in facilitating the inclusion. Even if we only include the core workflow (`NFCORE_RNASEQ`), we likely need to store the entire pipeline code in the meta-pipeline repo, because that is the thing that is versioned

As a user, I will want to know that my meta-pipeline is using a specific pipeline version (e.g. nf-core/rnaseq 3.3.0), so in effect we have to say that we are including the entire pipeline

bentsherman:
From our discussion today:

- define a *workflow registry* as a distinct concept from modules, but follows similar patterns (spec file, publish/install mechanism)

- workflow registry encapsulates both "core workflows" and "subworkflows"

- pipeline developer should isolate core workflow and publish it to the workflow registry, then other pipelines can import it

- for users who want to daisy-chain pipelines with minimal effort -- just make a pipeline chain (e.g. nf-cascade, platform actions) and sacrifice dataflow concurrency

adr/20260608-remote-pipeline-inclusion.md
maxulysse: could we force shared modules if we wanted too?

edmundmiller (commented):
left a few thoughts on scope/clarity — overall the direction makes sense to me.

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Summary
+
+Add the ability to include a remote pipeline into a *meta-pipeline*.
+
+## Problem Statement
+
+Nextflow supports reusing process definitions via remote *module* inclusion (e.g. `include { BWA_MEM } from 'nf-core/bwa/mem'`), but there is no standard mechanism to reuse an entire *pipeline* as a building block. Users must either fork and copy code, or compose/chain multiple `nextflow run` sessions which forfeits dataflow composition.
+
+The natural unit of reuse for a pipeline is its *core workflow* -- the named workflow that takes and emits channels (e.g. `NFCORE_RNASEQ` in `nf-core/rnaseq`) -- as distinct from the deployment shell around it (the `params`, entry workflow, `output` block, and config). The nf-core community has already structured their pipelines around this split in anticipation of meta-pipelines.
```
edmundmiller: might be worth clarifying whether this means nf-core pipelines are already generally import-ready today, or that they are moving toward this split. the rest of the ADR depends pretty heavily on a clean core-workflow / shell boundary.

adr/20260608-pipeline-composition.md
```diff
@@ -0,0 +1,351 @@
...
+The natural unit of reuse for a pipeline is its *core workflow* -- the named workflow that takes and emits channels (e.g. `NFCORE_RNASEQ` in `nf-core/rnaseq`) -- as distinct from the deployment shell around it (the `params`, entry workflow, `output` block, and config). The nf-core community has already structured their pipelines around this split in anticipation of meta-pipelines.
+
+The decision to make: how should a remote pipeline be distributed, resolved, stored, and included so that it can be composed into a meta-pipeline while preserving dataflow composition.
+
+## Goals
+
+- **Preserve dataflow composition**: the included pipeline participates in the meta-pipeline's dataflow graph (same session, same DAG, same work dir), enabling incremental reaction to emitted outputs.
+
+- **Preserve reproducibility**: an included pipeline should produce the exact same results as it would when executed directly. Transitive dependencies should not be silently altered to reduce duplication.
```
edmundmiller, line 33: maybe qualify “exact same results” a bit here? since params, config, outputs, plugins, and publishing are not inherited, direct-run equivalence seems to require the meta-pipeline to explicitly reproduce the relevant shell behavior.

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Non-goals
+
+- **Nested pipeline execution**: avoid Nextflow-in-Nextflow execution, which forfeits dataflow composition.
+
+- **Pipeline execution via registry**: out of scope for first iteration. Registry-based execution (e.g. `nextflow pipeline run nf-core/rnaseq@3.0.0`) may be investigated in the future.
+
+## Decision
+
+Allow remote pipelines to be included into meta-pipelines, using the same namespacing conventions and include syntax as modules. Store the included pipeline in the meta-pipeline repository under `workflows/<scope>/<name>/` with its own subdirectories for modules and subworkflows. The meta-pipeline owns all top-level concerns (entry workflow, params, outputs, config). Pipelines should be written with a self-contained core workflow to make importing as easy as possible.
```
edmundmiller: this boundary makes sense to me. might be worth spelling out the implication that this is a contract for pipelines that want to be composable, rather than automatic reuse of arbitrary existing pipelines.

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,351 @@
...
+
+```groovy
+// module
+include { BWA_MEM } from 'nf-core/bwa/mem'
+
+// pipeline
+include { NFCORE_RNASEQ } from 'nf-core/rnaseq'
+```
+
+Including a remote pipeline is equivalent to including the top-level `main.nf` of that pipeline; any named workflow defined there can be included by name. By convention, the main script defines only the entry workflow and the core workflow (e.g. `NFCORE_RNASEQ` in nf-core/rnaseq). From this point, the included workflow can be called like any other workflow.
```
edmundmiller: could this clarify what happens to the unnamed entry workflow / top-level params / output block when including the top-level `main.nf`? “equivalent to including main.nf” sounds a little like ordinary script inclusion, but the model here discards the shell.

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,351 @@
...
+
+When a pipeline is included, it is vendored into the meta-pipeline project under `workflows/<scope>/<name>/`. Included pipelines are isolated -- each included pipeline has its own `modules/` and `workflows/` directories. This way, two pipelines can use different versions of the same module without compromising reproducibility.
+
+Included pipelines should be committed to the meta-pipeline repository. The pipeline should have a *pipeline spec* (`nextflow_spec.json`) which specifies the pipeline version, so that Nextflow can track local changes.
```
edmundmiller: should `nextflow_spec.json` align with the module metadata shape from the module-system ADR, or is this intentionally a separate pipeline-level spec? a short note would help avoid seeming like two parallel metadata conventions.

adr/20260608-remote-pipeline-inclusion.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Open Questions
+
+### Sourcing from Nextflow registry vs Git repositories
+
+Pipelines could be stored in the Nextflow registry (as a new artifact type) or fetched directly from Git repositories. The pipeline registry is a potential long-term goal with other use cases, but it likely introduces additional scope that is not strictly related to meta-pipelines. Using existing Git repositories would be an expedient solution for the first iteration.
```
edmundmiller: since distribution/resolution/storage are part of the decision, could this ADR define the minimal first slice here? for example, git/tag install into `workflows/<scope>/<name>/`, with registry support left as a later extension.

pditommaso:
> - define a *workflow registry* as a distinct concept from modules, but follows similar patterns (spec file, publish/install mechanism)

I'm still convinced that this isn't a distinct concept — it's the same entity at a different granularity, and the framing matters because it drives how much we end up building.

**Logical view.** A module is just a component with a well-defined input/output interface that hides its internals — it's self-contained. By that definition a process, a subworkflow, and a pipeline's core workflow are all *modules*; they differ only in granularity, not in kind. We already agreed upthread that "they're all basically the same thing." If the abstraction is identical, introducing a second top-level entity ("workflow") next to "module" creates a distinction the type system and the user's mental model don't actually need.

**Tactical view.** Making it a distinct concept means re-implementing the entire surface twice — for `module` and for `workflow` — for operations that are literally identical:

```
create      Create a new skeleton
install     Install from the registry
run         Run directly from the registry
list        List installed
remove      Remove an installed entry
search      Search the registry
view        Show info and usage template
publish     Publish to the registry
spec        Generate a meta.yml spec
validate    Validate structure and metadata
```

That's two CLIs, two sets of registry APIs, two spec formats, two sets of Platform integrations, two docs trees — to express what is, behaviorally, one operation set over one kind of artifact. The cost is recurring (every new capability has to be added in both places and kept in sync), not one-off.

**Proposal.** Keep a single registry and a single command group over one artifact: the *module*. If we need to differentiate a coarse-grained "core workflow" from a fine-grained process, do it with a `type`/`kind` field in the spec (and resolution conventions), not with a parallel system. The publish/install/spec mechanism is shared by construction; the differences (e.g. a core workflow vendoring its own `modules/`) become metadata and layout conventions on the same entity rather than a second concept.

This also keeps the door open for #7208 (running named workflows directly) without forking the model — "run a module" and "run a workflow" stay the same command.

adr/20260608-remote-pipeline-inclusion.md
bentsherman: Idea from today's dicussion: import the inner pipeline params as a record type:

```groovy
include { NFCORE_FETCHNGS } from 'nf-core/fetchngs'
include { NFCORE_RNASEQ } from 'nf-core/rnaseq'
include { params as RnaseqParams } from 'nf-core/rnaseq'

params {
    input: Path // meta entry point
    strandedness: String = 'auto' // one new knob
    rnaseq: RnaseqParams
}

// ...
```

bentsherman: Related idea: include params/entry/output trio as a fake named workflow so that you can just call it:

```groovy
include { workflow as FETCHNGS } from 'nf-core/fetchngs'
include { workflow as RNASEQ } from 'nf-core/rnaseq'

workflow {
    fetchngs_out = FETCHNGS(params.fetchngs)

    // glue logic...
    rnaseq_in = params.rnaseq + fetchngs_out

    rnaseq_out = RNASEQ(rnaseq_in)
}
```

This doesn't require the inner pipelines to isolate a core workflow, it just includes the entire pipeline (params and publishing, not config)

If we allow channels in the params block (we already do it for the output block), we can even preserve the dataflow concurrency

pditommaso:
Note, my previous comment is based on the assumption is focused on sub-workflow only, not plain pipelines.

adr/20260608-remote-pipeline-inclusion.md
bentsherman: Removed, no longer relevant

adr/20260608-pipeline-composition.md
bentsherman: replaced "core workflow" distinction with just including the entire pipeline, so reproducibility is easier to guarantee

adr/20260608-remote-pipeline-inclusion.md
bentsherman: pipeline spec left as an open question. we might be able to enable meta-pipelines without it

adr/20260608-remote-pipeline-inclusion.md
bentsherman: open question resolved to source pipelines from registry

adr/20260608-remote-pipeline-inclusion.md
bentsherman: Resolved by latest updates

adr/20260608-remote-pipeline-inclusion.md
bentsherman: updated ADR to focus on native dataflow composition as an important factor (writing meta-pipelines with regular dataflow logic)

adr/20260608-remote-pipeline-inclusion.md
bentsherman: relaxed language here -- the important bit is that any settings in config need to be moved to process definition or copied/included into meta-pipeline config

bentsherman:
## Version 1.1

- Added separate ADR for **workflow modules** based on Paolo's [comment](https://github.com/nextflow-io/nextflow/pull/7213#issuecomment-4743140306) above. Keeping it in this PR for now so that we can discuss them together

- Workflows / subworkflows are just modules -- they use the same infrastructure (spec, registry API, CLI) as process modules, with some extensions for things like transitive deps

- Clarify main ADR to focus on remote pipeline inclusion via Nextflow registry -- pipelines will be treated as a separate concept from modules

- Replace the **core workflow distinction** with **pipeline inclusion** -- instead of isolating and importing only the core workflow of a pipeline, we extend the include syntax to support including an entire pipeline (the `params` / `workflow` / `output` trio) like a named workflow. See "Pipeline composition" in the ADR for an example

## Next steps

Gonna keep exploring ways to minimize overhead for params and publishing ([thread](https://github.com/nextflow-io/nextflow/pull/7213#discussion_r3388925957))

adr/20260608-remote-pipeline-inclusion.md
bentsherman: Added example to ADR, see latest comment

adr/20260608-pipeline-composition.md
```diff
@@ -0,0 +1,420 @@
...
+- Config params (`outdir`, `publish_dir_mode`, `max_cpus`, etc)
+- Resource settings (`cpus`, `memory`, `time`, etc)
+- Environment profiles (executors, software dependencies, test profiles)
+- Reports (execution, timeline, trace)
+- Manifest (name, authors, description, etc)
+- Plugins
+
+### Reducing params/output boilerplate
+
+In the example above, the meta-pipeline re-declares the params and outputs from each included pipeline. This boilerplate can be avoided by importing each pipeline's `params` block and `output` block as *record types*:
```
bentsherman, line 371: @pinin4fjords @adamrtalbot @ewels 

Here is my latest attempt at a "best of both worlds" where the meta-pipeline is a unified Nextflow pipeline but doesn't require tons of developer overhead

Basically we allow the `params` block and `output` block to be imported like record types so that you only have to redeclare one param per pipeline, and you get all the defaults baked in

pinin4fjords: YES. Sweet sweet sugary goodness.

adr/20260608-pipeline-composition.md
```diff
@@ -0,0 +1,420 @@
...
+params {
+    fetchngs: FetchngsParams        // input
+    strandedness: String = 'auto'   // unique to meta-pipeline
+    rnaseq: RnaseqParams            // input, aligner, fasta
+}
+
+workflow {
+    main:
+    // fetch FASTQ samples from NCBI SRA
+    fetchngs = NFCORE_FETCHNGS( params.fetchngs )
```
adamrtalbot, line 394: Is it worth naming this argument to avoid confusion?
```suggestion
    fetchngs = NFCORE_FETCHNGS( params: params.fetchngs )
```

adr/20260608-pipeline-composition.md
adamrtalbot: So `params` become a special type of input? Syntactic sugar over this:

```nextflow
take:
    RnaseqParams: Map = params
    fastqs: List[path]
    // etc
```

adr/20260608-pipeline-composition.md
bentsherman: that would be ambiguous with a param called `params` (which is dumb but valid nonetheless)

the contract is: if you import an entry workflow as a named workflow, it expects a single record argument with all the params

adr/20260608-pipeline-composition.md
bentsherman: given a params block like this:

```groovy
params {
    input: Channel<Sample>
    aligner: String = 'star_salmon'
    fasta: Path
}
```

importing it as `RnaseqParams` essentially gives you this:
```groovy
record RnaseqParams {
    input: Channel<Sample>
    aligner: String = 'star_salmon'
    fasta: Path
}
```

importing the entry workflow gives you something like this:
```groovy
workflow NFCORE_RNASEQ {
    take:
    params: RnaseqParams

    // ...

    emit:
    output: RnaseqOutput
}
```

pditommaso:
_This comment focuses only on the **workflow-modules** ADR — I'll follow up separately on the rest of the PR._

Overall I think the direction makes good sense — treating a (sub)workflow as just another *kind* of module feels right, and reusing the registry/CLI/spec rather than standing up a parallel system seems like the way to go.

A few things it'd be worth clarifying or firming up before this moves out of draft:

**1. The `workflows/` prefix seems used a little inconsistently.** It shows up in the storage path (`workflows/<scope>/<name>/`) but not in how a workflow module is *referenced* — the include is just `from 'nf-core/fastq_align_star'`, the same as a process. It'd probably be worth picking one model and applying it consistently everywhere a workflow module is named: the `include`, the `module run` command, the meta definition, and the registry. Either the `workflows/` qualifier is part of the identifier throughout, or it stays a purely internal storage detail — mixing the two could make resolution ambiguous (the resolver would have to probe both `modules/` and `workflows/`) and be a bit confusing for users.

**2. Worth clarifying that a workflow module can include *any* module.** The ADR reads a little as though only process modules get included, but a workflow naturally composes both processes and other workflows — the example's own `requires.modules` already lists a subworkflow (`bam_sort_stats_samtools`). It'd help to state explicitly that composition is recursive: a workflow module can include/depend on any other module, process or workflow.

**3. The workflow interface is probably best defined by `take:`/`emit:` rather than duplicated in the spec.** This is the part I'd most want to get right. In the typed model a process module's inputs and outputs live in its `input:`/`output:` blocks — the workflow equivalent is the `take:`/`emit:` definitions. So a workflow module's interface (input/output names + types) could be defined primarily by `take:`/`emit:` in the source, with the spec/schema just *annotating* it (per-input/output `description`, etc.) rather than re-declaring a parallel interface. That would keep a single source of truth and avoid the meta.yml-vs-source drift nf-core lives with today. What does still seem to need a defined schema for `kind: Workflow` is the surrounding metadata (`description`, `keywords` — I think `topics` should be `keywords` — `license`, `requires`), not the interface itself.

**4. Dependencies + versioning — there's a bit of tension here.** This reintroduces `requires.modules` and transitive resolution, which the module-system ADR had removed. For workflows that may be unavoidable (a workflow inherently depends on its processes), but it does reopen the diamond-dependency question. One option might be to avoid pinning exact versions: a punctual version like `@0.0.0-4e3e10e` would tend to conflict as soon as two workflows share a dependency, whereas a minimum version or a range would let the resolver flatten to a single compatible version. Might be worth spelling out the dependency version model and how conflicts are resolved.

**5. The params/output *mapping* might be better as its own ADR.** Once `take:`/`emit:` is the canonical interface, the remaining question is how to *project* it onto the CLI surface — turning simple values (CLI args / samplesheet rows) into `Channel`s on the way in, and channels back into published outputs on the way out. That's a fairly novel concept, and as written it sits a little awkwardly with the params ADR (which excludes `Channel` types from params). The rules for it (which `take:` types load-from-file vs. bind as scalars, how a `Channel<Record>` desugars into a collection param + `channel.fromList(...)`) feel substantial enough that they might be cleaner as a dedicated ADR, keeping this one focused on workflow modules as a registry/packaging concept.

pditommaso:
Nice work on this — the design is coherent and the doc is refreshingly honest about its own sharp edges (the alternatives section and the best-practices caveats do a lot of the reviewer's work). Reusing `include { workflow as X }` instead of inventing a new keyword, per-pipeline `modules/` isolation, and driving config through name-prefixed selectors are all the right calls. A few things I'd like to see resolved before this leaves draft.

**1. The config gap needs real exploration, not a footnote.**
"Inclusion captures the main script and modules but not config or `lib`" is currently a best-practices aside, but for the existing ecosystem this is the whole game — config is where behavior is actually *governed and controlled* (`ext.args`, resource labels, publish rules, profiles, plugins). The current best practices essentially ask authors to rewrite pipelines so config no longer carries logic, which almost no shipping pipeline does today. I don't think we should hand-wave this; it deserves a dedicated section that explores how much of the config shell can be captured/vendored vs. reconstructed, and what the minimum viable story is for a real pipeline. This is the make-or-break for adoption.

**2. `.pipeline-info` vs `nextflow_spec.json` is inconsistent — and the choice matters.**
The storage section pins version + checksum in `.pipeline-info`, but the appendix layout shows `nextflow_spec.json` per pipeline and no `.pipeline-info`, while Open Questions still treats the spec as undecided. These aren't interchangeable: one is an internal lockfile Nextflow manages, the other is a user-facing pipeline contract (params/outputs/requires). We should decide which is authoritative for versioning and make the examples consistent, otherwise it reads as two competing mechanisms.

**3. The params→channel binding needs a general model.**
The examples show `input: Path` becoming `input: Channel<Sample>` so a live channel can be passed in, but it's not clear *how* the mapping/binding is actually managed — it looks like it works for a few specific shapes (a samplesheet-like channel of records) rather than as a general rule. What's the binding semantics when a param is scalar vs. channel, single vs. queue, typed record vs. arbitrary? I'd want the general approach spelled out rather than inferred from one idealized case.

**4. Consider a dedicated `input` block instead of overloading `params`.**
Rather than overloading `params` to double as the composition interface (and carrying the type-conversion behavior above), it may be cleaner to introduce an explicit `input` definition — symmetric with `output`. `params` stays the CLI/config surface; `input`/`output` become the composition contract (take/emit). That symmetry would make the params→take / output→emit mapping explicit in the language rather than implicit, and gives the binding rules in (3) a natural home.

**General suggestions**

- **Split out "workflow modules."** The channel-typed-input behavior this ADR leans on really belongs to the workflow-modules ADR. I'd move it fully there and use that ADR to strengthen the input/output contract (the spec / `meta.yml` definition), so this ADR can depend on a stable interface rather than co-defining it.
- **Descope params from "workflow modules."** A workflow module *is* a module, so its `params` definition is already covered by the Module Parameters ADR (#7260) — no need to re-specify params handling here or in the workflow-modules ADR. Lean on that unified `params {}` model instead.
- **Reframe as "pipeline composition."** "Meta-pipeline" describes the artifact but not the intent. Framing the whole feature as *pipeline composition* makes the goal clearer and reads more naturally against "module inclusion" and "workflow modules."

Overall I'm in favor of the direction — dataflow composition over nested execution is the right choice. The above are mostly about making the contract explicit and being honest that config/environment is the hard part.

bentsherman:
Thanks @pditommaso for the reviews. I think we have converged on an approach, and I mostly just need to clarify some details in the ADRs

At this point I will move the workflow modules ADR into a separate PR, refine it based on your feedback, and begin implementation

I will also keep the workflow execution piece separate in a third PR (#7208) since it is not essential to meta-pipelines and deserves more dedicated treatment on the params -> take mapping

The *syntax* for pipeline composition can also be implemented independently of everything else, since a user could just copy+paste a pipeline into a meta-pipeline. But the full solution will require (1) workflow modules in the registry and (2) pipelines in the registry

adr/20260608-workflow-modules.md
```diff
@@ -0,0 +1,199 @@
...
+
+```groovy
+// process
+include { BWA_MEM } from 'nf-core/bwa/mem'
+
+// workflow
+include { FASTQ_ALIGN_STAR } from 'nf-core/fastq_align_star'
+```
+
+When a workflow is included, it is vendored into the including project under `workflows/<scope>/<name>/`. Included workflows should be committed to the including repository.
```
bentsherman: From @pditommaso 's review and our discussion today, I think we're generally agreed to just use the `modules` directory for both processes and workflows. It's not what nf-core has done, but it aligns better with our paradigm -- processes and workflows are both "modules", they exist in the same namespace, they have the same include syntax, so it makes sense to store them in the same place

Tooling can use the `kind` field in the module spec to easily distinguish between processes and workflows

adr/20260608-workflow-modules.md
```diff
@@ -0,0 +1,199 @@
...
+
+Transitive dependencies (specified by `requires.modules`) should also be installed in the `modules/` and `workflows/` directories alongside the included workflow. Since modules are flattened, it is not possible for a pipeline to use two different versions of the same process or workflow.
```
bentsherman: From our discussion today, this part is a can of worms:

- nf-core modules are currently versioned by commit hash and workflows pin exact versions. There is no way to resolve conflicts between workflows. Instead, all nf-core workflows are kept in sync at publish time. This makes it harder to update modules because you also have to update all consuming workflows

- We could make nf-core modules use minimal semantic versioning, e.g. `0.N.0` (always bump the minor version). The workflows could specify a range of versions for each module dependency, enabling version conflict resolution. This shifts the burden from process authors to workflow authors, who have to go back and update the version range when a process update is breaking

- We can avoid all of this by just allowing each workflow to have its own `modules` directory. It's just more duplication in the pipeline project. For what it's worth, this is the approach we plan to take for remote pipelines anyway.

jorgee: In the third point, do you mean workflow modules would be stored in  `workflows/<scope>/<name>/modules/<scope>/<name>`?

adr/20260608-workflow-modules.md
```diff
@@ -0,0 +1,199 @@
...
+    id: String
+    bam: Path
+    bai: Path
+}
+```
+
+The user can run the workflow directly as follows:
+
+```bash
+nextflow module run rnaseq.nf \
```
jorgee: Shouldn't be the workflow module reference such as `nf-core/rnaseq` and it should expect a process or workflow called `RNASEQ`?

adr/20260608-workflow-modules.md
jorgee: Regarding installing separately in `modules/` and `workflows/`. If they are both managed as modules and name is unique, is this distinction really needed? It is a minor issue, but currently we check if a module exist in `modules/`, so now we should look at `modules/` and `workflows/`. For run or module inclusion, we don't know in advance if the module reference is a process or workflow, we need to inspect the meta.yml either locally or remotely.

bentsherman: yes, well it would be `modules/<scope>/<name>/modules/<scope>/<name>` based on the other point about storing both processes and workflows under `modules`

adr/20260608-workflow-modules.md
bentsherman: it should work just like process modules:
```bash
# registry module
nextflow module run nf-core/rnaseq

# local module
nextflow module run ./modules/nf-core/rnaseq/main.nf
```

and the module should define exactly one process or workflow, regardless of the name

adr/20260608-workflow-modules.md
bentsherman: yes, see my [previous comment](https://github.com/nextflow-io/nextflow/pull/7213#discussion_r3513746136). I think we will just store both under `modules`

bentsherman:
Moved workflow modules ADR to #7342 

Will update pipeline composition ADR soon based on latest feedback

bentsherman:
ADR updated to focus on "pipeline composition" as the core feature. Remote pipeline inclusion is treated as a secondary goal.

Responding to Paolo's feedback in detail:

> **1. The config gap needs real exploration, not a footnote.**

I added a blurb about which pieces of config likely need to be reconstructed. I can try to lay out a more concrete example later.

The main prerequisite is that a pipeline will need to adopt the `params` and `output` block. Otherwise the params/publishing logic is just too scattered for Nextflow to reason about it here. That migration alone should remove a lot of the burden around config.

Even `ext` doesn't need to be removed entirely -- the defaults just need to be moved into the process definition.

The rest of the config -- anything outside of `process` -- simply has to be reconstructed, because all other config settings are global. So you can't simply compose them like modules, you have to decide what you want the global settings to be and define them.

> **2. `.pipeline-info` vs `nextflow_spec.json` is inconsistent — and the choice matters.**

Updated the appendix example to use `.pipeline-info` for now.

> **3. The params→channel binding needs a general model.**

As the ADR states, this behavior is described in the workflow modules ADR. I will make sure to expand on the binding rules there. But it is the same behavior as used for executing a named workflow.

> **4. Consider a dedicated `input` block instead of overloading `params`.**

I don't think users will want to have to define their params twice. But let's see how this evolves.

mahesh-panchal:
What are your thoughts on this kind of structure where the container is dictated by the input.

https://github.com/mahesh-panchal/nextflow-quarto-website/blob/main/main.nf

In this example the module is the same, but one aliased module requires a container with R, while the other requires a container with Julia. The default container can't be set in the process definition and must be via the config in this case.

adr/20260608-pipeline-composition.md
```diff
@@ -0,0 +1,427 @@
...
+    withName: 'NFCORE_FETCHNGS:.*:SRATOOLS_FASTERQDUMP' {
+        cpus   = 6
+        memory = 24.GB
+    }
+    withName: 'NFCORE_RNASEQ:.*:STAR_ALIGN' {
+        cpus   = 12
+        memory = 72.GB
+    }
+}
+```
```
adamrtalbot, line 356: If we copy+paste the pipeline configuration into the repo on import and add an `includeConfig "pipelines/nf-core/rnaseq/nextflow.config` statement, this would automatically include the config from nf-core/rnaseq and make a lot of people happy?

mahesh-panchal: One thing that may cause issues here is then the priority of patterns and simple names potentially causing issues applying to other processes (and where those `includeConfig`s are placed will also determine priority), and there's also that issue where the actual process name configuration applies too all processes even if they've been aliased. It's a can of worms.

mahesh-panchal:
> It seems like we always end up battling two conflicting things:
> 
>     * workflows should be fully encapsulated so they can be imported into another workflow cleanly
> 
>     * but also, I'd like it to include all the extra config and cruft not encapsulated in the workflow please

But this is basically what we're allowed to do. We use what's given to us, and implement what's simplest/makes sense to us.

adr/20260608-pipeline-composition.md
bentsherman: I think Adam basically covered it. You can either provide the container as a process input, or you can copy/include the config manually in the meta-pipeline (and make sure the selectors are specific enough, etc). Pick your poison

Either way, I don't think we can make config travel automatically with a pipeline/module because the config is a parallel system

muffato: > Either way, I don't think we can make config travel automatically with a pipeline/module because the config is a parallel system

Would it be possible to restrict `withName` and `withLabel` directives to a certain workflow prefix when doing composition ?
A (sub/)workflow could ship its own `nextflow.config` with all the `process {}` config it needs for itself, possibly with very loose `withName: SAMTOOLS_SORT` rules.
When imported, even via `include {worflow as IMPORTED_WF} ...`, its `nextflow.config` would be considered and only applied to the subset of the entire pipeline.

bentsherman:
> What are your thoughts on this kind of structure where the container is dictated by the input.

@mahesh-panchal you can set the container based on a process input and supply it from a param:

```groovy
process HELLO {
    container container_image

    input:
    val container_image

    // ...
}
```

```groovy
include { HELLO as HELLO_1 } from './hello.nf'
include { HELLO as HELLO_2 } from './hello.nf'

workflow {
    HELLO_1( params.container_image_1 )
    HELLO_2( params.container_image_2 )
}
```

bentsherman:
POC is now implemented. Moving back to draft until I have finalized the code.

This PR now includes an end-to-end example: `examples/pipeline-composition`. It's basically the ADR example in an executable form. Feel free to try it out and share feedback!

docs/reference/syntax.mdx
```diff
@@ -111,6 +111,20 @@ include { hello } from './some/module'
 include { bye as goodbye } from './some/module'
 ```
 
+<AddedInVersion version="26.10" />
+
+The `workflow` name refers to the entire pipeline defined by the included script, i.e. its params block, entry workflow, and output block. The `params` and `output` names refer to the params block and output block as record types. Each of them must be given an alias:
```
christopher-hakkaart:
```suggestion
The `workflow` name refers to the entire pipeline defined by the included script, i.e., its params block, entry workflow, and output block. The `params` and `output` names refer to the params block and output block as record types and must be given an alias:
```

adr/20260608-remote-pipeline-inclusion.md
bentsherman: This can be handled at the pipeline level. Pipelines can make the multiqc step optional via param if they want to support meta-pipelines with only a top-level multiqc report

PR #7216, opened 2026-06-10
Remove support for `when:` section in typed processes

bentsherman:
This PR removes support for `when:` in typed processes.

While it was preserved during the preview period to ease migration, the stable version of typed processes should not support it

Aside from being an anti-pattern, `when:` can actually break semantic guarantees of dataflow logic. If a process is called with a dataflow value and returns a dataflow value, and I use `when:` to disable that process via config, it will return an "empty" dataflow value which is invalid

Rather than deprecating it across the board, it will remain allowed in legacy processes and removed in typed processes. Migrating to static typing is the best point at which to also migrate away from `when:` because records and type checking makes it easier to do

### DIFF
--- a/modules/nf-lang/src/main/java/nextflow/script/control/VariableScopeVisitor.java
+++ b/modules/nf-lang/src/main/java/nextflow/script/control/VariableScopeVisitor.java
@@ -342,17 +342,13 @@ class VariableScopeVisitor extends ScriptVisitorSupport {
         for( var input : asFlatParams(node.inputs) ) {
             vsc.declare(input, input);
 
-            // suppress "unused variable" warnings since Path inputs are implicity staged
+            // suppress "unused variable" warnings since Path inputs are implicitly staged
             vsc.findVariableDeclaration(input.getName(), input);
         }
 
         vsc.pushScope(ProcessDsl.StageDsl.class);
         visitDirectives(node.stagers, "stage directive", false);
         vsc.popScope();
-
-        if( !(node.when instanceof EmptyExpression) )
-            vsc.addWarning("Process `when` section will not be supported in a future version", "", node.when);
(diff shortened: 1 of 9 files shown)


modules/nf-lang/src/main/java/nextflow/script/parser/ScriptAstBuilder.java
```diff
@@ -492,13 +492,16 @@ private ProcessNode processDef(ProcessDefContext ctx) {
         if( !typingEnabled && !topics.isEmpty() )
             collectSyntaxError(new SyntaxException("The `topic:` section is not supported in a legacy process", topics));
 
+        if( typingEnabled && !(when instanceof EmptyExpression) )
+            collectSyntaxError(new SyntaxException("The `when:` section is not supported in a typed process", when));
```
jorgee, line 496: Nice to have a test to validate that the error is raised when `when:` is used in a typed process.

modules/nf-lang/src/main/java/nextflow/script/control/VariableScopeVisitor.java
jorgee, line 382: Just a question, not blocking. Is this warning for processV1 is still intended?

bentsherman: yes. it's a "paranoid warning" so it's only shown by the language server as an "info" diagnostic if you have it enabled. it won't show up in the linter or runtime

pditommaso (changes requested):
I'm not totally convinced about removing this. I was even considering that this could implement a similar pattern as until/recursion

bentsherman:
`when:` is not compatible with static typing. I don't see how recursion is relevant here

PR #7217, opened 2026-06-10
docs: Clarify distinction between modules and script inclusion

bentsherman:
When DSL2 was introduced, "modules" referred to both "a process in a script" and "any script included by another script". Over time, I ended up emphasizing the latter definition in the docs.

However, with the emergence of registry modules, subworkflows, and meta-pipelines, the meaning of "module" has become muddled. This shows up in several places:

- the disconnect between Nextflow and nf-core's definition of a module (nf-core subworkflow != nf-core module)
- the explanation of local vs registry modules in the docs

I think it has become clear that a "module" is really just a "process in a script". A workflow in a script is not a module.

This PR clarifies the definition of "module", separating it from the more general concept of "script inclusion". It aligns the use of "module" with the module registry, module CLI, and nf-core convention.

Related: #6650 , #7213 

TODO:
- [x] introduce script inclusion in the "Scripts" page instead of the modules overview?


docs/script.md
```diff
@@ -11,7 +11,7 @@ Nextflow uses UTF-8 as the default character encoding for source files. Make sur
 :::
 
 :::{warning}
-Nextflow scripts have a maximum size of 64 KiB. To avoid this limit for large pipelines, consider moving pipeline components into separate files and including them as modules.
+Nextflow scripts have a maximum size of 64 KiB. To avoid this limit for large pipelines, consider moving pipeline components into separate files and including them into the main script.
```
christopher-hakkaart:
```suggestion
Nextflow scripts have a maximum size of 64 KiB. To avoid this limit for large pipelines, consider moving pipeline components into separate files and including them in the main script.
```

christopher-hakkaart:
I like the direction to limit the definition.

> A **module** is a script that contains a single process definition.

With this definition, it's really just the `main.nf` with only one process definitions. Where does this leave associated files? E.g., `meta.yml`. Do we refer to those as module files?

christopher-hakkaart:
We should also look at the training to make sure this is echoed there

docs/modules/modules.md
```diff
@@ -3,9 +3,9 @@
 # Overview
 
 Nextflow scripts can include **definitions** (workflows, processes, and functions) from other scripts.
-When a script is included in this way, it is referred to as a **module**.
+A **module** is a script that contains a single process definition.
```
pditommaso:
```suggestion
A **module** is self-contained process definition providing a well-defined module specification file [link to module spec].
```

Think it should be stressed a module defined the process script along with companion metadata spec file

bentsherman: @christopher-hakkaart I think this answers your question. A module is primarily the script + spec

christopher-hakkaart: Yup - thanks!

bentsherman:
I notified the training team about this PR, they will double-check their material

bentsherman:
@christopher-hakkaart I think this PR is ready now

christopher-hakkaart (approved):
Looks good!

PR #7226, opened 2026-06-11
Fix staging of Path fields of included record types

### LINKED ISSUE #7225: Record `Path` fields are not staged when the record type is included from another module
## Summary

With the typed/static-types syntax (`nextflow.enable.types = true`), a process input that is a `Path` field of a **record type included from another module file** (e.g. `include { Pair } from './types.nf'`) is **not staged**. The implicit `stageAs` directives that should be generated for the record's `Path` fields are missing, so the field value is interpolated into `.command.sh` as a raw source path instead of a staged `TaskPath`.

- On a local filesystem the bug is masked: the raw absolute path still resolves, so the task succeeds (but `rec.r1.getClass()` reveals `sun.nio.fs.UnixPath` instead of `nextflow.processor.TaskPath`).
- On an object-store work dir (AWS Batch + Fusion/S3) the raw path is `/bucket/key`, which never passes through the Fusion mount mapping, and the task fails with `No such file or directory`.

Defining the same record **inline** in the script that declares the process stages the fields correctly. Direct `Path` inputs are also unaffected.

## Reproducer

https://github.com/robsyme/nf-record-fusion-repro

```bash
export NXF_SYNTAX_PARSER=v2
nextflow run robsyme/nf-record-fusion-repro
```

`MAKE` emits `Pair(id, r1: Path, r2: Path)` where `Pair` is included from `types.nf`. `CONSUME_A` reads `rec.r1`/`rec.r2` (record fields) and fails on Fusion/S3; `VIA_PATH` reads the same files as direct `Path` inputs and succeeds.

Isolating the variables:

| Record definition | Field staged? |
|---|---|
| included from `types.nf` | no — raw source path |
| inline in `main.nf` | yes — `TaskPath` |

## Evidence

`getClass()` evaluated at script-render time, same run:

```
CONSUME_A s02: r1class=class nextflow.cloud.aws.nio.S3Path  r1=/bucket/.../s02_1.txt   # record field -> raw, UNSTAGED
VIA_PATH  s02: r1class=class nextflow.processor.TaskPath     r1=s02_1.txt               # direct Path -> staged
```

Failed `CONSUME_A` `.command.sh`:
```
cat /bucket/.../s02_1.txt /bucket/.../s02_2.txt > s02.a.out
# -> cat: /bucket/.../s02_1.txt: No such file or directory
```

## Root cause

The v2 script compiler (`ScriptCompiler`) resolves and converts each source file in a single per-source pass, and the entry script is analyzed **before** the modules it includes. When `ProcessToGroovyVisitorV2.visitProcessInputType` inspects a record-typed input to generate the implicit `stageAs` directives, the included record node is found (`isRecordType` is true) but its field types are still unresolved ClassNodes — `isPathType` returns false, so no stager is emitted for the `Path` fields.

At runtime the unstaged field then misses the staged-path substitution in `TaskInputResolver.normalizeValue` (the path is absent from the staged-file `holders` map) and reaches the task script as a raw source path.

## Environment

- Nextflow 26.04.2 / 26.04.3, also reproduced on master; strict v2 parser (`NXF_SYNTAX_PARSER=v2`)
- Any executor (staging bug visible locally); failure surfaces on AWS Batch + Fusion v2 + Wave

---

*Edited: the original report attributed the failure to `nextflow.enable.moduleBinaries`. That was a misdiagnosis — the flag is unrelated; the trigger is the record type being included from a separate module file.*

robsyme:
Fixes #7225

## Problem

When a typed process (`nextflow.enable.types = true`) declares an input whose type is a record **included from another module file**, the implicit `stageAs` directives for the record's `Path` fields are not generated. The fields are interpolated into `.command.sh` as raw source paths instead of staged paths. On a shared local filesystem the raw path still resolves, which masks the bug, but on an object-store work dir (e.g. Fusion/S3) the task fails with `No such file or directory`.

Isolating the variables with the reproducer shows the trigger is the record type being included from a separate file (the `nextflow.enable.moduleBinaries` flag suspected in the original report is unrelated):

| Record definition | Field staged? |
|---|---|
| included from `types.nf` | no — raw source path |
| inline in `main.nf` | yes — `TaskPath` |

## Cause

`ScriptCompiler` performs resolution and Groovy conversion per source file, and the entry script is analyzed **before** the modules it includes. When `ProcessToGroovyVisitorV2` inspects a record input to infer the implicit file inputs, the included record node is found (`isRecordType` is true) but its field types are still unresolved ClassNodes, so `isPathType` returns false and no stager is generated for the `Path` fields. The unstaged field then bypasses the staged-path substitution in `TaskInputResolver.normalizeValue` and reaches the task script as a raw source path.

## Solution

- `ModuleResolver` now records the include dependencies of each source file, returning existing source units on repeat encounters so the dependency graph is complete.
- `ScriptCompiler` analyzes each source file only after the modules it includes from, so included types are fully resolved before they are used. Guards handle include cycles and unparseable modules.

## Verification

- New unit test `ScriptLoaderV2Test."should stage path fields of an included record type"` fails before the change (0 file inputs generated) and passes after (2 file inputs).
- The issue reproducer (https://github.com/robsyme/nf-record-fusion-repro) now renders the record fields as staged `TaskPath` names instead of raw paths.
- Transitive includes (process in module A consuming a record included from module B) also stage correctly.
- Circular includes still fail with a `StackOverflowError`, but this is pre-existing behavior at runtime module loading (`IncludeDef.loadModuleV2`), unchanged by this PR.

### DIFF
--- a/docs/executor.md
+++ b/docs/executor.md
@@ -486,14 +486,28 @@ The following {ref}`hints <process-hints>` are supported:
 - `machineRequirement.machineTypes`
 - `machineRequirement.maxSpotAttempts`
 - `machineRequirement.provisioning`
+- `predictionModel`
 
-Each hint overrides the corresponding field of the `seqera.executor.machineRequirement` config scope on a per-process basis. Keys may be used as-is or with the `seqera/` prefix to restrict them to this executor.
+The `machineRequirement.*` hints override the corresponding field of the `seqera.executor.machineRequirement` config scope on a per-process basis. The `predictionModel` hint overrides the run-level `seqera.executor.predictionModel` for a single process: set it to `qr/v1` or `qr/v2` to select a model, or to `none` to disable resource prediction for that process even when the run enables a model. When omitted, the process inherits the run-level prediction model. Keys may be used as-is or with the `seqera/` prefix to restrict them to this executor.
 
 For example, to override the provisioning mode for a single process:
 
 ```nextflow
 process hello {
     hints 'seqera/machineRequirement.provisioning': 'spotFirst'
+
+    script:
+    """
+    your_command --here
+    """
+}
+```
+
+To disable resource prediction for a single process when the run sets a model:
+
+```nextflow
+process hello {
+    hints 'seqera/predictionModel': 'none'
 
     script:
     """
--- a/docs/migrations/dsl1.md
+++ b/docs/migrations/dsl1.md
@@ -96,14 +96,14 @@ cheers
 
 Similarly, in DSL2, process outputs can be consumed by multiple consumers automatically, which makes workflow scripts much easier to read and write.
 
-## Modules
+## Includes
 
 In DSL1, the entire Nextflow pipeline must be defined in a single file. For example, `main.nf`. This restriction becomes cumbersome as a pipeline grows and hinders the sharing and reuse of pipeline components.
 
-DSL2 introduces the concept of "module scripts" (or "modules" for short), which are Nextflow scripts that can be "included" by other scripts. While modules are not essential to migrating to DSL2, nor are they mandatory in DSL2, modules can help you organize a large pipeline into multiple smaller files and take advantage of modules created by others. See {ref}`modules-page` to learn more about modules.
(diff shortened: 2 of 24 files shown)


bentsherman:
Thanks for the issue and PR

I kept your test case and tried to find a simpler and more general solution -- all scripts should be "analyzed" before any script is converted to Groovy

It was enough to fix your test case, let's see if doesn't break anything else...

bentsherman (approved):

PR #7228, opened 2026-06-12
Propagate Seqera Intelligent Compute scheduler run identifier to Platform

jorgee:
## Description

When a workflow runs on **Seqera Intelligent Compute** (the `seqera` executor), the scheduler assigns a **run identifier**. Platform currently has no way to associate that id with its workflow record. This is the **Nextflow side** of that propagation: it delivers the scheduler run id to Platform via a dedicated **`PATCH /workflow/{workflowId}`** request with body `{ "schedRunId": "…" }`.

> Supersedes #7210, which carried the run id as a top‑level `schedulerRunId` field on the trace **progress** request. Per review, that mechanism was replaced with the dedicated `PATCH /workflow/{workflowId}` endpoint.

## How the run id is obtained and propagated

- `SeqeraExecutor.createRun()` assigns the run id on the **first task submission** and publishes it to `WorkflowMetadata.schedRunId` (`session.workflowMetadata?.setSchedRunId(runId)`).
- `schedRunId` is a single `volatile String` on `WorkflowMetadata` — an internal executor→observer hand‑off. It is **excluded from `WorkflowMetadata.toMap()`**, so it is not serialized on the begin/complete workflow object or in the lineage record; Platform receives it only through the dedicated PATCH.
- `TowerObserver` sends the id via a **one‑off `PATCH /workflow/{workflowId}`** (`TowerClient.updateWorkflow`). The trace sender thread fires it the first time the id is non‑null and then latches, so it is sent **exactly once** — at the earliest moment the id exists (first task submission), well before completion.
- Delivery is **best‑effort**: `updateWorkflow` returns the HTTP `Response` (it does not throw), and the caller logs a failed update and lets the run carry on instead of aborting. Transient failures are already retried at the HTTP‑client layer (`retryConfig`), so the single attempt latches only once that retry policy is exhausted.

## What changed vs the earlier iteration

An earlier version of this PR also carried a config‑derived `sched.enabled` boolean (peer of `wave`/`fusion`) on the begin/complete workflow object, via a `SchedMetadata` value class. **That has been removed**, because:

- it was never consumed inside Nextflow;
- deriving it from run‑level config is unreliable — it misses per‑process `withName:{ executor='seqera' }` overrides, so it could report `false` for a run that does use the scheduler and sends a `runId`;
- the presence of `schedRunId` already tells Platform the run is scheduler‑managed.

So `SchedMetadata`, the `TowerJsonGenerator` converter, and the `LinObserver` handling are all gone; the only scheduler data on the wire is `schedRunId` via the PATCH.

## Components changed

| Module | Change |
|---|---|
| `modules/nextflow` | `WorkflowMetadata.schedRunId` (`volatile`) field, excluded from `toMap()` |
| `nf-seqera` | `SeqeraExecutor.createRun()` publishes the run id to `session.workflowMetadata.schedRunId` |
| `nf-tower` | `TowerObserver` sends the run id once via `PATCH /workflow/{id}` (best‑effort); `TowerClient.updateWorkflow` returns the `Response` + `PATCH` support |
| `adr/` | `20260609-scheduler-run-identifier-propagation.md` |

## Testing

- Unit tests added/updated: `TowerObserverTest`, `TowerClientTest`, `SeqeraExecutorTest`.
- Suites pass: `:nextflow`, `:nf-lineage`, `:plugins:nf-tower`, `:plugins:nf-seqera`.

## TODO

- [ ] End‑to‑end validation against the Platform implementation (the workflow‑extension satellite table and the `PATCH /workflow/{workflowId}` endpoint) once it lands — confirm `schedRunId` is persisted and exposed as expected. This Platform side is a separate workstream; since delivery is best‑effort, it no longer creates a hard Nextflow ↔ Platform version dependency.


plugins/nf-seqera/src/main/io/seqera/executor/SeqeraExecutor.groovy
```diff
@@ -142,6 +142,11 @@ class SeqeraExecutor extends Executor implements ExtensionPoint {
         log.debug "[SEQERA] Creating run: ${request}"
         final response = client.createRun(request)
         this.runId = response.getRunId()
+        // publish the scheduler run id so the Tower observer can propagate it to Platform
+        // on trace progress/heartbeat requests (used for cost and resource accounting)
```
stefanoboriero: Outdated comment, this is no longer passed through progess/heartbeat

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
```diff
@@ -152,6 +152,24 @@ class TowerClient {
...
+        if( resp.error ) {
+            final message =  """\
+                Unexpected HTTP response
+                - endpoint    : $url
+                - status code : $resp.code
+                - response msg: $resp.message
+                """.stripIndent(true)
+            throw new AbortRunException(message)
+        }
+    }
```
stefanoboriero, line 165: Can we make this not fail the run? I think it's an acceptable degradation if platform cannot store the runId, it would just show less information in Platform, but the workflow could carry on.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerJsonGenerator.groovy
```diff
@@ -47,6 +48,7 @@ class TowerJsonGenerator extends DefaultJsonGenerator {
                 .addConverter(Path) { Path p, String key -> p.toUriString() }
                 .addConverter(Duration) { Duration d, String key -> d.durationInMillis }
                 .addConverter(NextflowMeta) { NextflowMeta m, String key -> m.toJsonMap() }
+                .addConverter(SchedulerMetadata) { SchedulerMetadata m, String key -> m.toMap() }
```
stefanoboriero: Could the paylaod sent be a simple boolean `"schedEnabled": true/false` instead of the nested object `"scheduler": {"enabled": true}`?

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerObserver.groovy
```diff
@@ -428,6 +430,31 @@ class TowerObserver implements TraceObserverV2 {
...
+     * life of the run, so it is sent exactly once; Platform stores it as a workflow-extension field
+     * used for cost and resource accounting.
+     */
+    protected void sendSchedulerRunId() {
+        if( schedulerRunIdSent )
+            return
+        final id = getSchedulerRunId()
+        if( !id )
+            return
+        client.updateWorkflow([schedulerRunId: id], workspaceId, workflowId)
```
stefanoboriero: Can this payload be sent as shorthand `"schedRunId": "xxxx"` instead of `schedulerRunId`?

stefanoboriero (commented):
Added some comments in order to align with the API shape from https://github.com/seqeralabs/platform/pull/11399 after addressing comments on that pr

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
jorgee: In past releases, failed platform telemetry messages never stop the run but we recently change failures stop the execution except the heartbeats. In this case, I do not have a strong opinion on it. I understand your view. @bentsherman, @pditommaso  do you agree on not aborting the run if the `schedulerRunId` update message fails?

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerJsonGenerator.groovy
jorgee: I was following the same format as Wave and Fusion.  Flattening to `schedEnabled `makes
  scheduler the odd one out and diverges from that established convention — that's the concern, not implementation difficulty.

plugins/nf-seqera/src/main/io/seqera/executor/SeqeraExecutor.groovy
jorgee: Fixed — removed the stale reference; the comment now just notes the runId is added to the workflow metadata to propagate it to Platform.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerObserver.groovy
jorgee: Done — the payload now sends `schedRunId` instead of `schedulerRunId`.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
stefanoboriero: My main concern is that since this uses a new Platform endpoint, hard failure would create a dependency between Nextflow and Platform version since runs against an "older" Platform version would fail to call this endpoint with status 405, at least for IntelligentCompute runs. And I wouldn't be comfortable introducing such hard dependency on a patch release, but I leave the final judgment and call to you guys

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerJsonGenerator.groovy
stefanoboriero: I know, but it feels wasteful to store an object on Platform side when it's really going to hold only one value. We can however accept the consistent payload and deserialise it to a single field platform side if you prefer

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
jorgee: Ok, you are right. I think it makes sense to change to just log.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerJsonGenerator.groovy
stefanoboriero: Ok let's be consistent on the Nextflow and API side, storage layout should not dictate API shape. I'll resolve this conversation once I make the changes in Platfrom

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerClient.groovy
jorgee: Agreed — changed to best-effort: `sendSchedulerRunId` now logs the error and lets the run carry on instead of aborting, attempting the update just once. Added a test covering the non-abort behavior.

pditommaso (changes requested):
### Naming: apply the `sched` prefix consistently, and align the ADR to it

The change reads cleanly and follows the `wave`/`fusion` pattern well, but the `Sched`/`scheduler` naming is split: the wire field already uses the Seqera `Sched` convention (`schedRunId`, matching `SchedClient`/`SchedRunId`/`SchedTaskState` in `nf-seqera`), while every other new identifier uses the full word `Scheduler`. And the ADR specifies the wire field as `schedulerRunId`, which **doesn't match the code's `schedRunId`** — since Platform is built against the ADR and the receiver ignores unknown JSON keys (Micronaut default, as the ADR itself notes), a mismatch would fail silently and only surface in e2e.

Let's go fully abbreviated (`Sched`/`sched`) everywhere and make the ADR match the implementation. All cases:

**Code identifiers**

| Current | → |
|---|---|
| class `SchedulerMetadata` + file `SchedulerMetadata.groovy` | `SchedMetadata` / `SchedMetadata.groovy` |
| test `SchedulerMetadataTest` + file | `SchedMetadataTest` |
| `WorkflowMetadata.scheduler` field | `sched` |
| `WorkflowMetadata.getScheduler()` | `getSched()` |
| `TowerObserver.getSchedulerRunId()` | `getSchedRunId()` |
| `TowerObserver.sendSchedulerRunId()` | `sendSchedRunId()` |
| `TowerObserver.schedulerRunIdSent` | `schedRunIdSent` |
| `TowerJsonGenerator` converter type `SchedulerMetadata` | `SchedMetadata` |
| `LinObserver`: cast `SchedulerMetadata` + key `"scheduler"` | `SchedMetadata` + `"sched"` |
| `SeqeraExecutor` locals `scheduler` / `?.scheduler` | `sched` / `?.sched` |
| javadoc ref `{@code schedulerRunId}` in `SchedMetadata` | `{@code schedRunId}` |

**Wire / JSON contract** — ⚠️ confirm with the Platform endpoint owner before merge

| Current | → |
|---|---|
| PATCH body `schedRunId` | already correct ✓ |
| workflow block `scheduler: {…}` (begin/complete) | `sched: {…}` |
| nested `workflow.scheduler.runId` | `workflow.sched.runId` |

Note the block currently serializes as `scheduler` (a sibling of `wave`/`fusion`, which are *not* abbreviated). Renaming it to `sched` is a wire-contract decision — fine to lock now since the Platform side isn't built yet, but it needs sign-off from whoever owns `PATCH /workflow/{id}` and the begin/complete consumers.

**ADR (`20260609-scheduler-run-identifier-propagation.md`) — align to the above**

- `schedulerRunId` → `schedRunId` (lines 94, 103, 123)
- `SchedulerMetadata` → `SchedMetadata` (49, 77, 82, 84, 113, 131, 137, 140)
- `getScheduler()` → `getSched()` (132); `sendSchedulerRunId()` → `sendSchedRunId()` (96)
- `WorkflowMetadata.scheduler` / `workflow.scheduler[.runId]` and JSON examples `scheduler: {…}` → `.sched` (49, 56, 105, 120, 126, 134, 147)
- "the `scheduler` entry" in the lineage section → "the `sched` entry" (112, 139)
- Leave as English prose: the word "scheduler", and the Platform-side `scheduler_enabled` column name (that's Platform's, out of our control — flag it to them separately).
- Test `def`-blocks / description strings (`'should serialise scheduler metadata…'`, etc.) are prose — optional to touch.

---

### Other (minor, non-blocking)

1. **`sendSchedulerRunId()` latches on failure.** `schedulerRunIdSent = true` is set *outside* the try/catch, so one transient PATCH failure permanently abandons the early signal for the whole run. The complete request still carries `runId` as a fallback, so it's not lost — but since the sender loop runs many times, consider latching only on success to preserve the "earliest moment" goal.
2. **`TowerClient.updateWorkflow` throws `AbortRunException`** while its sole caller swallows it as best-effort — slightly misleading naming, though it mirrors the existing `sendAndProcessRequest` style, so acceptable.
3. **`getScheduler()` double-checked locking on a non-`volatile` field** — same as the existing `getPlatform()`, so not a regression (and `runId` itself is correctly `volatile`). Leave unless you're revisiting `getPlatform()` too.
4. **Missing trailing newline** in `SchedulerMetadata.groovy`.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerJsonGenerator.groovy
stefanoboriero: Changes made on Platform side, the workflow being now accepts the object `'"sched": {"enabled": true/false}'`, notice the change from scheduler to sched for consistency

jorgee:
Pushed `7935c1efb` addressing @stefanoboriero's and @pditommaso's review:

**Naming — adopt the `Sched`/`sched` convention consistently** (matching `SchedClient`/`SchedRunId` in `nf-seqera`, and the API shape agreed on the Platform side):
- `SchedulerMetadata` → `SchedMetadata` (class + file)
- `WorkflowMetadata.scheduler` → `sched`; `getScheduler()` → auto getter
- `TowerObserver`: `getSchedulerRunId`/`sendSchedulerRunId`/`schedulerRunIdSent` → `getSchedRunId`/`sendSchedRunId`/`schedRunIdSent`
- wire/JSON keys `scheduler: {…}` → `sched: {…}` on begin/complete, plus the `nf-lineage` entry
- ADR aligned to the implementation (`schedRunId`, `SchedMetadata`, `sched: {…}`)

**Best-effort run-id update (Minor 2 in the review):** `TowerClient.updateWorkflow` now returns the `Response` instead of throwing `AbortRunException`; the caller logs a failed update and the run carries on. No more throw-then-catch.

**`sched` lifecycle ( Minor 3):** initialized eagerly in the `WorkflowMetadata` constructor like `wave`/`fusion` (instead of the lazy double-checked getter), so the double-checked-locking-on-non-volatile concern goes away and it's consistent with its sibling metadata.

**Trailing newline ( Minor 4):** fixed.

**On the `sendSchedRunId()` latch (review minor 1):** I've left it latching after one attempt. The HTTP client is built with `.retryConfig(retryPolicy)`, so transient failures (5xx / connection errors) are already retried at the client layer — we only latch once that retry policy is exhausted, not on a single transient blip. An alternative would be to latch only on success for non-4xx failures (i.e. keep retrying transient/server errors across sender-loop iterations, but give up immediately on a 4xx like the 405 from an older Platform). Happy to switch to that if preferred.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerJsonGenerator.groovy
jorgee: Changed to `sched`

modules/nextflow/src/main/groovy/nextflow/script/SchedMetadata.groovy
```diff
@@ -0,0 +1,103 @@
...
+     * are not considered here; the flag reflects the run-level executor selection.
+     */
+    private static String resolveExecutor(Session session) {
+        final process = session?.config?.process as Map
+        if( process?.executor )
+            return process.executor.toString()
+        final executor = session?.config?.executor as Map
+        if( executor?.name )
+            return executor.name.toString()
+        return SysEnv.get('NXF_EXECUTOR')
```
pditommaso: This is expected to be `seqera` by definition. However, tasks can, in principle, use its their own provider, each task can have it's sub-category 

https://github.com/nextflow-io/nextflow/blob/b5ac62e0ca84f30e5c4c799318dab4ccc35992a9/plugins/nf-seqera/src/test/io/seqera/executor/SeqeraTaskHandlerTest.groovy#L261-L261

Above all, why this is needed?

jorgee: I'm not sure I'm fully getting your comment.
  
 In this code, I was trying to infer if the run is using Intelligent Compute checking the config. However, I think the flag is even weaker than it looks: a process can select the scheduler purely via `process.withName:foo { executor='seqera' }` with nothing set at run level. `ExecutorFactory.getExecutorName()`
  honors that per-process value, so the task really runs on the scheduler and gets a `runId` — but `SchedMetadata.resolveExecutor()` only looks at run-level config (`process.executor` / `executor.name` /
  `NXF_EXECUTOR`), so it would report `enabled=false`. We'd then PATCH a `runId` for a run marked `sched.enabled=false`, which is contradictory.
  
  Beyond that, `enabled` isn't used anywhere in Nextflow — it's only serialized to Platform. It was added in the first approach at @stefanoboriero's request. Since the `runId` (via `PATCH` ) already
  implies the run is scheduler-managed, we could drop the flag and the `resolveExecutor` heuristic entirely and rely on `schedRunId` (plus what Platform already knows from the launch / compute environment).

  @stefanoboriero — do you still need `sched.enabled` on the Platform side, or can we remove it and rely on `schedRunId`?

stefanoboriero: I see, it's unfortunate but let's drop `schedEnabled` from the begin request: it was a nice to have for us, but we can work around this in Platform looking at the computeEnv config or at the schedRunId.

pditommaso: Even if the `enabled` flag is needed is could/should set along the runId (or even better as a side effect of setting of runId via `setRunId`)

jorgee: If I am not wrong, it is nice to have at the beginning, isn't it? This is what is happening with the current code in platform runs where `process.executor = 'seqera'`.  Nextflow is not using the `enabled` flag, so either we drop it and just use `schedRunId` or keep it as it is. I can't see any benefit of setting` enabled` along `runId `

pditommaso: Agree

jorgee: Then, what do you prefer to keep or drop?

pditommaso: Drop

stefanoboriero: Could something be set in the Session when SeqeraExecutor#register executes, then we set the SchedMetadata#enabled based on that? I guess there might be the edge case where the plugin is registered but not used still :thinking:

jorgee: Yes, I think we can use `Plugins.isStarted('nf-seqera')` to infer the `schedEnabled ` for all the cases. Looking at the code, I see the `nf-seqera` plugin is auto-started if the config contains` process.executor='seqera'`. In other cases, such as setting the seqera executor using `process.withName`, they must add the plugin in the config.

pditommaso: Useless! if the runId is set it's enabled by definition

jorgee:
Pushed `a0d7e34c6` (rebased on the latest merge from master).

Following @pditommaso's question on `SchedMetadata` and the discussion with @stefanoboriero, **we decided to drop the `sched.enabled` flag entirely and propagate only `schedRunId`**. Reasoning:

- `enabled` was never consumed inside Nextflow — it was pure telemetry for Platform.
- Deriving it from run-level config is unreliable: a process can select the scheduler via `process.withName:foo { executor='seqera' }` with nothing set at run level, so `resolveExecutor()` would report `enabled=false` while the run actually uses the scheduler and sends a `runId` — a contradiction.
- The presence of `schedRunId` already tells Platform the run is scheduler-managed, so the flag was redundant.

What changed in this commit:
- Removed the `SchedMetadata` value class (and its test), the `TowerJsonGenerator` converter, and the `LinObserver` handling.
- The run id is now a single `volatile String schedRunId` on `WorkflowMetadata`, used purely as the executor→observer hand-off and **excluded from `toMap()`** — so the begin/complete workflow object (and the lineage record) no longer carry any scheduler data.
- `schedRunId` is delivered solely via the best-effort `PATCH /workflow/{id}`.

Net ~−270 lines. The PR description is updated to match. @stefanoboriero — this means Platform no longer receives `sched: {enabled}` on begin/complete; it should rely on the `schedRunId` PATCH.

modules/nextflow/src/main/groovy/nextflow/script/SchedMetadata.groovy
jorgee: Enabled drop in last commit, I just store the `schedRunId` in the workflowMetadata.

stefanoboriero (approved):

pditommaso (commented):
Good simplification. Only a minor concern about exposing this attribute via WorkflowMeta. Using PlatformMeta sounds a better fit. Look into https://github.com/nextflow-io/nextflow/pull/7251

pditommaso (approved):

PR #7233, opened 2026-06-15
ADR: Process `error:` section

bentsherman:
This PR adds an ADR for the proposed process `error:` section, which provides a way to capture and emit *domain errors*, allowing users to handle them with workflow logic rather than the error strategy.

### DIFF
--- a/adr/20260615-process-error-section.md
+++ b/adr/20260615-process-error-section.md
@@ -1,0 +1,176 @@
+# Process error section
+
+- Authors: Ben Sherman
+- Status: proposed
+- Deciders: Ben Sherman, Paolo Di Tommaso
+- Date: 2026-06-15
+- Tags: lang, static-types, processes, error-handling
+
+## Summary
+
+Add an `error:` section to typed processes, mirroring the `output:` section, that lets a process emit a *domain error* as a value instead of aborting the run.
+
+## Problem Statement
+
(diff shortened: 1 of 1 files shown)


adr/20260615-process-error-section.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Non-goals
+
+- Support for legacy processes. The `error:` section is introduced for typed processes only; legacy support may follow later.
+
+- New syntax for *classifying* errors (no exit-code lists, no guard expressions).
+
+## Solution
+
+Introduce an **`error:` section** for typed processes, with the same syntax as the `output:` section. A domain error is detected structurally -- **a task that succeeds (exit 0) but does not fulfill its declared `output:`** -- and is emitted as an error value instead of triggering the error strategy.
```
adamrtalbot, line 47: I'm not sure this is sufficient for separating the domain vs infrastructure errors. 

See https://nfcore.slack.com/archives/C043FMKUNLB/p1782301366504629 for a relevant discussion.

bentsherman: Left some comments there. That discussion seems to be more about distinguishing between different kinds of execution errors

adr/20260615-process-error-section.md
```diff
@@ -0,0 +1,176 @@
...
+
+    error:
+    record(id: id, log: file('aligner.log'))
+
+    script:
+    """
+    aligner ${reads} ${index} > aligned.bam 2> aligner.log || {
+        rm -f aligned.bam
+        exit 0
+    }
```
bentsherman, line 83: TODO: wrap exit code wrangling in an `error()` Bash function for re-use

PR #7235, opened 2026-06-15
docs: Clarify storeDir caveats with env, eval, stdout outputs

bentsherman:
This PR clarifies that `storeDir` does not work with env, eval, or stdout outputs.

Refactored warnings as bulleted list to reduce warning overload. It isn't really warranted here

Related: #5785


docs/reference/process.mdx
```diff
@@ -1713,13 +1713,11 @@ process make_blast_db {
 }
 ```
 
-:::warning
-If a process uses `storeDir` and all of its outputs are optional, the process will always be skipped, even if the store directory is empty. This issue can be avoided by specifying at least one required file output.
-:::
+- The `env`, `eval`, and `stdout` output qualifiers cannot be used with `storeDir` because they rely on helper files in the task directory. Use `path` outputs instead.
```
bentsherman, line 1718:
```suggestion
Caveats:

- The `env`, `eval`, and `stdout` output qualifiers cannot be used with `storeDir` because they rely on helper files in the task directory. Use `path` outputs instead.
```

christopher-hakkaart (approved):

PR #7239, opened 2026-06-17
Add hidden plural/singular aliases for noun subcommands

ewels:
## Summary

Nextflow's noun subcommands are singular (`plugin`, `module`), which makes them easy to mistype in the plural. This adds hidden command aliases so common typos resolve to the correct subcommand:

| Command | Hidden alias |
|---------|--------------|
| `plugin` | `plugins` |
| `module` | `modules` |
| `secrets` | `secret` (inverse — `secrets` is already plural) |

## Implementation

Aliases are declared via JCommander's `commandNames` field on each command's `@Parameters` annotation, matching the existing `lineage` / `li` alias pattern. `Launcher` already wires these through `addCommand(cmd.name, cmd, aliases(cmd))`, and `getParsedCommand()` returns the canonical name when a command is invoked via an alias, so command resolution is unaffected.

The aliases are **not** shown in `nextflow help` because `Launcher.printCommands()` renders only each command's canonical `getName()` and never enumerates the JCommander aliases.

## Testing

Added a `LauncherTest` case asserting that both the canonical names and the plural aliases (`plugin`/`plugins`, `module`/`modules`) resolve to their command classes. `LauncherTest` passes.


pditommaso (approved):
Fair

PR #6429, opened 2025-09-25
enhance: Upgrade Claude Code hooks with advanced JSON reporting

edmundmiller:
## Summary

This PR enhances the Claude Code development hooks with advanced JSON reporting capabilities, transforming them from simple pass/fail indicators into intelligent development assistants.

#### 🔧 EditorConfig Hook (`format-editorconfig.py`)
- **File Change Detection**: SHA256 hash comparison to identify specific formatting changes
- **Performance Metrics**: Execution timing and eclint version reporting
- **Auto-Installation Progress**: Real-time feedback during eclint installation
- **Structured Error Handling**: JSON errors with actionable suggestions
- **Rich Success Messages**: Detailed reporting of actual changes applied

#### 🧪 Test Runner Hook (`run-tests.py`)
- **Gradle Output Parsing**: Structured analysis of build results and timing
- **XML Test Result Parsing**: Detailed test statistics with method names and execution times
- **Intelligent Failure Analysis**: Pattern matching for NPEs, assertions, timeouts
- **Context-Aware Suggestions**: File-specific recommendations based on failure patterns
- **Comprehensive Success Reporting**: Test method details and build metrics

#### 📊 Advanced JSON Output
- **Strategic Control Fields**: Proper use of `decision: "block"`, `suppressOutput: true`, `systemMessage`
- **Structured Metadata**: Rich data for Claude's intelligent decision-making
- **Performance Insights**: Timing metrics and progress indication
- **Enhanced Error Context**: Detailed diagnostics with actionable fix suggestions

### Benefits

1. **Better Claude Decision-Making**: Structured data enables intelligent responses to test failures
2. **Actionable Feedback**: Specific suggestions help Claude fix issues automatically  
3. **Rich Context**: Detailed results provide better understanding of changes and impacts
4. **Performance Insights**: Timing and metrics help optimize development workflow
5. **User Experience**: Clean, informative messages with progress indication

## Test Plan

- [x] Test EditorConfig hook with various file types (Groovy, Java, Markdown)
- [x] Test formatting change detection and performance metrics
- [x] Test test runner hook with source files and direct test files
- [x] Test XML result parsing and intelligent error suggestions
- [x] Test enhanced JSON output and structured error handling
- [x] Test both hooks running in parallel with proper timeout handling

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #7240, opened 2026-06-18
Claude Code hooks (1/4): EditorConfig formatting

edmundmiller:
First of a 5-PR stack that adds [Claude Code hooks](https://docs.claude.com/en/docs/claude-code/hooks)
for **developing the Nextflow engine** (not for authoring Nextflow scripts — that tooling
lives in `nextflow-io/agent-skills`). This was previously one large PR (#6429); it's split
into one hook per PR so each can be reviewed and understood on its own.

All hooks are small, self-contained **Bash** scripts under `.claude/hooks/`, wired up in
`.claude/settings.json`. They parse the Claude Code JSON payload with `jq` (the only added
dependency; hooks no-op silently if `jq` is absent).

## This PR — EditorConfig enforcement (PostToolUse, async)
- `hooks/format-editorconfig.sh`: after every `Edit`/`Write`/`MultiEdit` on a source file
  (`.groovy`/`.java`/`.gradle`/`.md`/`.txt`/`.yml`/`.yaml`/`.json`), apply the project
  `.editorconfig` rules via `eclint fix`.
- `eclint` is an **optional** dependency — the hook skips silently if it isn't installed
  (no global `npm install` side effect).
- Runs with **`async: true`** so formatting happens in the background and never blocks the
  conversation.
- Also adds the `.claude/settings.json` + `.claude/README.md` scaffolding the later PRs build on.

## The stack (review front-to-back)
1. **(this PR)** EditorConfig formatting (async)
2. Build check — `make compile` on Stop/SubagentStop — #7241
3. Test runner — run the edited file's gradle test class — #7242
4. IntelliJ IDEA formatter (async, optional) — #7243
5. Integration test runner — run the edited `tests/*.nf` case (`asyncRewake`) — #7244

Each later PR targets the previous branch, so `settings.json`/`README.md` grow cleanly with
no cross-PR conflicts.

### DIFF
--- a/.claude/README.md
+++ b/.claude/README.md
@@ -1,0 +1,35 @@
+# Claude Code Hooks for Nextflow Development
+
+This directory contains [Claude Code hooks](https://docs.claude.com/en/docs/claude-code/hooks)
+configured to improve the Nextflow development experience. Each hook is a small,
+self-contained Bash script under `hooks/`, wired up in `settings.json`.
+
+Hooks are added incrementally — one per pull request — so each can be reviewed and
+understood on its own. This file documents the hooks that are currently wired up.
+
+## Requirements
+
+- [`jq`](https://jqlang.github.io/jq/) — hooks parse the Claude Code JSON payload with it.
+  If `jq` is not on `PATH`, the hooks no-op silently.
+
+## Hooks
+
+### 1. EditorConfig enforcement (PostToolUse, async)
+- **Script**: `hooks/format-editorconfig.sh`
+- **Trigger**: after every `Edit`/`Write`/`MultiEdit` on a source file
+  (`.groovy`, `.java`, `.gradle`, `.md`, `.txt`, `.yml`, `.yaml`, `.json`)
+- **Action**: applies `.editorconfig` rules with `eclint fix`
+- **Dependencies**: `eclint` (optional — install with `npm install -g eclint`; the hook
+  skips silently if it is not present)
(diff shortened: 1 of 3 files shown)


bentsherman:
Since I've been using agents more over the past month, I'm just not convinced that these hooks are needed. I've found that Claude is already quite good at verifying its own changes by running tests (which also builds the code so you don't really need a separate build hook). And the agents are only going to get better, so I suspect hooks like these will only get in their way or cause unnecessary extra noise

The editorconfig hook is the only one I'm still considering, since it's very basic formatting rules. Here also, Claude is already quite good at writing correctly-formatted code

edmundmiller:
I agree that hooks should not add noise or repeat work. If running the tests already builds the code, we may not need a separate build hook.

But basic checks should not depend on the agent remembering to run them. Agents usually do. Sometimes they do not, and then the reviewer pays the cost. Code should not come back if it does not build or pass the relevant tests.

A hook makes that rule consistent. It runs the same check every time without using tokens to decide what to run. We can keep successful runs quiet and make the checks faster with targeted tests, incremental builds, and caching.

Agents will keep getting better. That does not make automatic checks less useful. Mechanical work should be handled by tools.

bentsherman:
I agree in principle. I just haven't found a case in my own work where any of these hooks would make a meaningful difference. Let's revisit when that changes

Sometimes [worse is better](https://dreamsongs.com/RiseOfWorseIsBetter.html) :trollface:

PR #7249, opened 2026-06-22
Fix issues with v2 config parser

robsyme:
### Problem

On the 25.10.x line, the strict (v2) config parser aborts a run at config-build time with `Unknown configuration profile: '<name>'` whenever the `profiles {}` block lives in a config file that is **not the last one parsed**. The legacy (v1) parser handles the same configuration fine.

This surfaces on Seqera Platform launches in particular: Platform writes a config into the launch directory (a `tower {}` scope, no profiles) which is parsed after the pipeline's own `nextflow.config`, so every profile-using pipeline fails under the v2 parser. It reproduces with plain `nextflow config`/`nextflow run` against two config files; `includeConfig` is not involved.

### Root cause

`ConfigBuilder` creates a single `ConfigParser` instance and calls `parse()` for each config file in order, then validates the requested `-profile` against`parser.getDeclaredProfiles()` once at the end.

`ConfigParserV2` **reassigned** its `declaredProfiles` field on every `parse()`call (`declaredProfiles = script.getDeclaredProfiles()`), so the field ended up holding only the profiles from the last file parsed. The v1 parser keeps a `final` set and adds to it, accumulating across all files.

### Fix

Initialise `declaredProfiles` to an empty set and accumulate with `addAll(...)` across `parse()` calls, matching the v1 behaviour. Two lines.

This is already fixed on the default branch as part of the larger #6643 ("enable v2 syntax parser by default"), which can't be cherry-picked cleanly onto 25.10.x; this PR is the minimal extraction of just the profile-accumulation fix.

### Verification

- New regression test `should accumulate declared profiles across multiple config files` reuses one parser across two `parse()` calls (the second declaring no profiles) and asserts the earlier file's profiles are retained.
- Confirmed the test **fails on the unpatched 25.10.x branch** and **passes with the fix**; the rest of `ConfigParserV2Test` is unchanged and green.

### Scope / risk

The only non-test consumer of `getDeclaredProfiles()` is `ConfigBuilder`'s `checkValidProfile` gate, so the blast radius is limited to profile validation. The v2 parser is opt-in on 25.10.x (`NXF_SYNTAX_PARSER=v2`).

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/config/parser/v2/ConfigDsl.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/config/parser/v2/ConfigDsl.groovy
@@ -44,11 +44,11 @@ class ConfigDsl extends Script {
 
     private Path configPath
 
-    private Map paramOverrides
+    private Map cliParams
 
     private List<String> profiles
 
-    private Map target = [:]
+    private Map target = [params: [:]]
 
     private Set<String> declaredProfiles = []
 
@@ -70,9 +70,13 @@ class ConfigDsl extends Script {
         this.configPath = path
     }
 
-    void setParams(Map paramOverrides) {
-        this.paramOverrides = paramOverrides
-        target.params = paramOverrides
+    void setParams(Map params) {
+        this.cliParams = params
+        (target.params as Map).putAll(params)
+    }
+
+    void setConfigParams(Map params) {
+        (target.params as Map).putAll(params)
     }
 
     void setProfiles(List<String> profiles) {
@@ -120,7 +124,7 @@ class ConfigDsl extends Script {
     void assign(List<String> names, Object value) {
         if( names.size() == 2 && names.first() == 'params' ) {
             declareParam(names.last(), value)
-            if( paramOverrides.containsKey(names.last()) )
+            if( cliParams.containsKey(names.last()) )
                 return
         }
         navigate(names.init()).put(names.last(), value)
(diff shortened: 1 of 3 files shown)


bentsherman:
@robsyme does this also fix #6623 ? or is that a different bug? it was also fixed by #6643 but was never backported to 25.10

that bug is also a likely blocker so we might as well make sure it is fixed as well

robsyme:
Heya Ben. 

No, this PR does not fix #6623. They're separate bugs with separate fixes.

Both bugs were fixed together in #6643, but by two independent code changes - my PR here only picks out one of those fixes.

The two fixes are:

**Profile accumulation (the one this PR backports) - my customer bug**
ConfigParserV2: declaredProfiles = script.getDeclaredProfiles() → declaredProfiles.addAll(...). Fixes the loud Unknown configuration profile failure.

**Params from includeConfig (issue #6623)**
In 25.10.x, ConfigDsl.setParams() aliases target.params = paramOverrides, and assign() skips any param already in paramOverrides. When an active profile does includeConfig, the include is parsed with .setParams(target.params)  - so the main config's declared params get passed in as if they were CLI overrides, and the included file's assignment is silently skipped. #6643 fixed this by splitting cliParams (true --param overrides, which should win) from configParams (accumulated config values, for reference only). This PR currently does not touch this path.

How about I adjust this PR to backport both fixes?

bentsherman (approved):

PR #7251, opened 2026-06-23
Hold scheduler run id under PlatformMetadata (workflow.platform.schedRunId)

pditommaso:
## Description

Stacked on top of #7228. Moves the Seqera Intelligent Compute scheduler run id from a top-level `WorkflowMetadata.schedRunId` field to `PlatformMetadata.schedRunId` — i.e. `workflow.platform.schedRunId`, grouped with the other Platform identifiers (`workflowId`, `workflowUrl`).

## Rationale

**Better not to expose the run id as a top-level `workflow` attribute.** The run id is a Platform-assigned identifier that is meaningless outside a Seqera Platform / Intelligent Compute run — conceptually a sibling of `platform.workflowId`, not portable run metadata like `workflow.runName` or `workflow.sessionId`. A public field on `WorkflowMetadata` is reachable as `workflow.schedRunId` in any pipeline, advertising it to every run as though it were general workflow metadata. Nesting it under `platform` keeps it where it belongs: **more coherent Platform-related metadata.**

**It also makes the serialization behaviour fall out for free.** The original PR needed a special-case exclusion in `WorkflowMetadata.toMap()` so the id wasn't serialized on the begin/complete/lineage workflow object. Under `platform` that's unnecessary:

- `TowerObserver.makeCompleteReq()` already strips the whole `platform` sub-object before sending.
- At `onFlowBegin` — when the begin request and the lineage record capture `toMap()` — the id is always still `null` (it is assigned later, on the first task submission).

So the begin/complete/lineage workflow object never carries the real value, **without** the special-case `toMap()` filter. The dedicated `PATCH /workflow/{workflowId}` remains the sole carrier — unchanged from #7228.

## Changes

| Module | Change |
|---|---|
| `modules/nextflow` | Remove `WorkflowMetadata.schedRunId` (+ its `toMap()` exclusion); add `PlatformMetadata.schedRunId` (`volatile`) |
| `nf-seqera` | `SeqeraExecutor.createRun()` → `session.workflowMetadata?.platform?.setSchedRunId(runId)` (consistent with the existing `platform?.workflowId` read) |
| `nf-tower` | `TowerObserver.getSchedRunId()` reads `workflowMetadata?.platform?.schedRunId` |
| `adr/` | Updated to reflect the `PlatformMetadata` location and the rationale above |

## Testing

`SeqeraExecutorTest`, `TowerObserverTest`, `WorkflowMetadataTest` updated and green (12 / 31 / 8 tests, 0 failures). Wire contract and PATCH delivery are unchanged from #7228.

### DIFF
--- a/adr/20260609-scheduler-run-identifier-propagation.md
+++ b/adr/20260609-scheduler-run-identifier-propagation.md
@@ -44,23 +44,38 @@ never leaves the executor.
 
 ## Solution or decision outcome
 
-The scheduler run id is held on `WorkflowMetadata` as a single `volatile String schedRunId` field
-and delivered to Platform via a one-off `PATCH /workflow/{workflowId}` request with body
-`{ "schedRunId": "…" }`.
+The scheduler run id is held on `PlatformMetadata` as a single `volatile String schedRunId` field —
+i.e. `workflow.platform.schedRunId`, grouped with the other Platform identifiers (`workflowId`,
+`workflowUrl`) rather than promoted to a top-level `workflow.schedRunId` attribute. It is delivered to
+Platform via a one-off `PATCH /workflow/{workflowId}` request with body `{ "schedRunId": "…" }`.
 
 - `SeqeraExecutor.createRun()` assigns the run id on the first task submission and publishes it to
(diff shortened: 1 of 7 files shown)


jorgee (approved):
Looks like a better approach.

PR #7256, opened 2026-06-24
docs: Normalize version callouts

bentsherman:
1. Remove old version-added / version-changed callouts (<=22.10)
2. Round up edge version callouts (prior to 26.04) to next stable version
3. Remove patch number from stable version callouts (26.04.0 -> 26.04)

I kept (3) in a separate commit since it is a more substantial change. I think we should do it because the major version (e.g. 26.04) is the only relevant number for documenting features. Similar to how we say "Java 17" or "Groovy 4", not "Java 17.0.19" or "Groovy 4.0.31"


.docusaurus_site/package-lock.json
bentsherman, line 1: The package-lock was out of sync (name, indentation) so `npm install` kept updating it. Fixing it here

christopher-hakkaart (approved):
Looks good. Nice edit.

PR #7259, opened 2026-06-25
Add ephemeral OS disk support to Azure Batch pools

adamrtalbot:
## Why

Azure Batch supports ephemeral OS disks, which place the node OS disk on local/cache storage instead of remote Azure storage. This improves node performance and lowers cost. nf-azure had no way to enable it.

## What

- Add `azure.batch.pools.<name>.ephemeralOsDisk` config option (boolean, default `false`)
- `poolVmConfig()` attaches an `OSDisk` with ephemeral settings on the cache disk (`DiffDiskPlacement.CACHE_DISK`) when enabled
- Include the flag in the pool cache funnel, so autopool pool IDs account for it (toggling provisions a new pool rather than reusing an existing one)
- Add unit tests for both config parsing and VM config generation
- Document the new option in `config.mdx`

## Notes

- The selected VM size must support an ephemeral OS disk with a cache large enough for the OS image, otherwise pool creation fails.
- Adding the flag to the funnel changes autopool pool IDs; on upgrade, existing autopools are not reused.

### DIFF
--- a/docs/reference/config.mdx
+++ b/docs/reference/config.mdx
@@ -513,6 +513,12 @@ The client ID for an Azure [managed identity](https://learn.microsoft.com/en-us/
 
 Enable autoscaling feature for the pool identified with `<name>`.
 
+##### `azure.batch.pools.<name>.ephemeralOsDisk`
+
+<AddedInVersion version="26.10.0" />
+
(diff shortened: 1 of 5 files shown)


bentsherman:
do you still want to add the config option? seems like a simple addition, but I'd rather not add another lever if it is strictly worse

PR #7260, opened 2026-06-25
ADR: Module Parameters and Tool Arguments

pditommaso:
## Summary

Consolidates two separate ADR drafts — #6769 (*module parameters*) and #6770 (*module tool arguments*) — into a single ADR, on the premise that **a tool flag is just a parameter whose string form is `-K <value>`**.

Highlights:

- **One `params {}` model** for both module parameters and tool arguments — no separate `tools.*.args` namespace.
- Replaces opaque, scattered `task.ext.args` strings with **first-class, documented, schema-backed parameters** accessible across CLI, config, and APIs.
- Override binding resolved by the **included process name** (config / CLI / params-file), so the same module included twice configures independently.
- Presents and compares **two candidate grouping designs**:
  - **Option A** — a `group` attribute + a single `ToolArg` type (nested blocks in `params {}`).
  - **Option B** — record-typed params (`ToolOpt` / `ToolArgs`), flat `params {}` — **author's recommendation**.
- Programmatic access (`.name` / `.value`), tool-option serialization rules, validation/errors, and migration from `ext.args`.

**Supersedes and replaces #6769 and #6770**, which are being closed in favor of this unified ADR.

### DIFF
--- a/adr/20260624-module-parameters.md
+++ b/adr/20260624-module-parameters.md
@@ -1,0 +1,416 @@
+# Module Parameters and Tool Arguments
+
+- Authors: Paolo Di Tommaso
+- Status: proposed — pending decision between Option A and Option B (author recommends B)
+- Date: 2026-06-24
+- Tags: modules, parameters, tools, arguments, configuration, records
+- Related: [Module System](20251114-module-system.md), [Workflow params](20250825-workflow-params.md), [Record types](20260306-record-types.md)
+
+## Summary
+
+Unify two previously separate drafts — *Module Parameters* ([PR #6769](https://github.com/nextflow-io/nextflow/pull/6769)) and *Tools Arguments* (an unpublished draft, not committed to `adr/`) — into a single mechanism: a tool flag is just a parameter whose string form happens to be `-K <value>`. Both drafts agree on the host (`params {}` block), the override binding, and the custom-type machinery; they differ only in **how grouping/nesting is expressed**. This ADR presents the two candidate designs as **Option A** (a `group` attribute + a single `ToolArg` type) and **Option B** (record-typed params with `ToolOpt`/`ToolArgs`), then compares them. This ADR supersedes both drafts.
+
+## Problem Statement
+
+Two adjacent problems were being solved by two separate designs:
+
+1. **Module parameters** — modules need a documented, type-safe way to declare and override configurable values, replacing scattered `params.x` assignments.
+
+2. **Tool arguments** — modules wrap one or more command-line tools (e.g. `bwa`, `samtools`), and their CLI flags are today passed through `task.ext.args`/`ext.args2` as raw strings. This mechanism has fundamental shortcomings:
+   - **Buried and scattered.** The values live in `ext.args` assignments spread across several config files (`nextflow.config`, profiles, institutional configs, `conf/modules.config`), with no single place that declares what a module accepts.
+   - **Undocumented.** A raw string like `-K 100000000 -Y` carries no description of what each flag does, its type, or its allowed values.
+   - **No schema.** There is no machine-readable definition of a module's arguments, so they cannot be validated, introspected, or surfaced by tooling and IDEs.
+   - **Opaque semantics.** Arguments are concatenated strings — easy to mistype, impossible to type-check, and meaningless to anything but the underlying tool.
+
+Keeping module params and tool args as two parallel concepts (`params` vs `tools.<tool>.args`) compounds this: two specification schemas, two override syntaxes, and two mental models for the same thing — *a named, typed, overridable input to a module*.
+
+**Goal:** expose tool arguments (and module settings generally) as **first-class parameters** — declared in one place, documented and backed by a schema for validation, and accessible programmatically and consistently across the CLI, config, and APIs.
+
+## Decision Drivers
+
+- **Replace `task.ext.args`.** Superseding the opaque `ext.args`/`ext.args2` strings with a documented, typed, introspectable mechanism is the primary goal — not merely offering an alternative.
+- **One concept, one syntax.** A tool argument is a parameter with a particular serialization. Don't invent a second namespace for it.
+- **Reuse the `params {}` block.** The typed `params {}` block already exists (workflow params ADR). Module params should be the same block.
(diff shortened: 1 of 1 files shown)


adr/20260624-module-parameters.md
```diff
@@ -0,0 +1,416 @@
...
+
+# Common foundations (shared by both options)
+
+These elements are identical regardless of which grouping design is chosen.
+
+### Host: the `params {}` block
+
+Module parameters are declared in a module-level `params {}` block, reusing the workflow-params syntax (`name: Type` / `name: Type = default`). The two options below differ only in how grouped/tool params are declared *inside* this block.
+
+> The original *Module Parameters* draft (PR #6769) weighed a **process-level** `params:` block against this **module-level** `params {}` block, and the module-level block is adopted here for continuity with workflow params. Note that the "Option A / Option B" labels in *this* ADR denote a **different axis** — how grouping is expressed — than the A/B options in PR #6769 (which were about host placement).
```
bentsherman, line 56: This ADR doesn't really explain why the process `params:` option was abandoned. That question was never resolved, and if anything, it seems like people favored the process `params:` option

Using the `params` block creates an ambiguity when there is a process and an entry workflow, or when there are multiple processes

bentsherman:
I saw module params and tool arguments as two alternative approaches... I think trying to combine them has made the ADR less coherent

Might be better to throw out the references to previous ADRs and just frame this ADR around the problem you are trying to solve, whether that be replacing `ext.args` or something else

The main challenge with `ext.args` is that it often depends on the task inputs. So any approach that is not dynamic per-task is not a feasible replacement for `ext.args`. It may have some other use case, but it doesn't cover `ext.args`

Some ideas we discussed on the previous ADRs:

- a `tools` map namespaced by command/subcommand (e.g. `tools.samtools_view.*`)

- for each tool, storing both an `args` list and `kwargs` map to support arbitrary CLI options

- declaring args / prefix / suffix in the input record so that they are both typed and dynamic per-task (likely adds operator boilerplate to workflow logic)

adr/20260624-module-parameters.md
```diff
@@ -0,0 +1,416 @@
...
+A single tool-option value, whatever the type is named, renders to its command-line form. The **name is the tool option name**, and the group/record name is the **tool name** (`bwa`, `samtools`) — which is why `params.bwa` renders that tool's full argument string.
+
+| Value | Renders as |
+|-------|------------|
+| `K = 100000000` | `-K 100000000` |
+| `Y = true` | `-Y` |
+| `output_fmt = "cram"` | `--output_fmt cram` |
+| unset / null / false | `` (omitted) |
+
+**Prefix inference** (overridable via `<Type>:prefix`): a single-character name uses `-`; a name of two or more characters uses `--`. **Boolean** `true` emits only `prefix+name`. An `enum` constrains the allowed values; a `default` supplies the value when unset. Components are accessible via `.name` and `.value`.
```
jorgee: It looks like the serialization imposes space between the prefix+name and value. It disables options like `-02`,` -j4` or `--name=value`. I think we should add a kind of `<Type>:separator` to cover this case

pditommaso: Good point, thanks — the space-only form is too rigid and can't express `-O2`, `-j4` or `--name=value`.

I've added a `<Type>:separator` attribute (`ToolArg:separator` / `ToolOpt:separator`) next to the existing `:prefix`, defaulting to a single space. Set it to `''` for glued forms (`-O2`, `-j4`) or `'='` for `--name=value`; booleans ignore it. Updated the serialization section to reflect this.

adr/20260624-module-parameters.md
pditommaso: That's a fair point — I should spell out the rationale in the ADR.

The idea is to keep a single, consistent way to define parameters rather than spread the concept across pipeline, process and sub-workflow levels. Underlying that, it's also about what a param *is*: an immutable value resolved once and overridable at launch (CLI + config) — essentially the counterpart to a process/workflow input. A process-level `params:` block tends to blur that boundary, whereas anything that needs to vary per task is really an input.

On the ambiguity you mention (process + entry workflow, or multiple processes): agreed it needs care. The override path is scoped to the included process name (`-process.<NAME>.params.…`, `withName:`), so the same module included twice can be configured independently — I'll make that link clearer in the text.

pditommaso:
@bentsherman I think the main thing driving this is how unwieldy `ext.args` has become: the values are scattered across many config files, hidden, and neither documented nor validated. The aim is to normalise them as regular params — scoped to the target module/process, documented, typed and schema-backed — and to avoid overloading the syntax with yet another layer dedicated only to tool params.

On the dynamic-per-task point: that's intentionally out of scope. As noted on the other thread, a param is an immutable launch-time value, so anything that has to vary per task belongs in a process input rather than a param. This isn't meant as a drop-in for those per-task uses of `ext.args`.

It's admittedly a bit of a painful change, but I think it's worth it to bring module settings into the open as well-documented, actionable parameters.

bentsherman:
> On the dynamic-per-task point: that's intentionally out of scope. As noted on the other thread, a param is an immutable launch-time value, so anything that has to vary per task belongs in a process input rather than a param. This isn't meant as a drop-in for those per-task uses of `ext.args`.

Thanks for the clarification. It is tricky though -- as a module author, I don't know how other users will use my module in their pipeline. They might provide a static setting or a per-sample setting, depending on what they're doing. So if my options as the module author are (1) a process input (which can be dynamic if needed) or (2) a module param (which can't), I will always choose a process input to be safe.

In other words, the per-task use case comes from *users* of a module, not the module itself

It might be useful to take a mid-sized nf-core pipeline like [methylseq](https://github.com/nf-core/methylseq) -- maybe even isolate just a subworkflow -- and have an agent reason about how to replace the ext settings with a type-safe approach

bentsherman:
From our discussion today, there are two core problems:

1. Documenting available CLI options for a tool so that an agent knows what can be used
2. Providing a way to supply tool args via `nextflow module run` instead of `ext.args` in a companion config file

Related requirements:
- Users don't want to be forced to document/maintain every single tool option
- Ideal interface should support both direct usage via `nextflow module run` and pipeline composition (dynamic per-task tool args from a channel)

PR #7266, opened 2026-06-29
fix: lower trace-access.nf memory to fit CI runner capacity

pditommaso:
## Problem

The `trace-access.nf` integration test (`test_integration` and `test_parser_v2` jobs) doubles the task memory on each retry — `1 → 2 → 4 → 8 GB` — expecting **3 re-submissions** before the 4th attempt (8 GB) succeeds and prints the trace-meta assertion.

GitHub's Azure-hosted runners now report only **7.8 GB** of total memory. The local executor rejects the 8 GB attempt with:

```
ProcessUnrecoverableException: Process requirement exceeds available memory -- req: 8 GB; avail: 7.8 GB
```

This abort happens *before* the 4th re-submission, so only **2** re-submissions occur and the check fails:

```
+ [[ 2 == 3 ]]
```

The failure is purely environmental (runner downsized below 8 GB) and reproduces on `master`.

## Fix

Scale the memory sequence down to `256.MB → 512.MB → 1.GB → 2.GB`. The peak (2 GB) stays well under the runner capacity while still exercising the `task.previousTrace.memory` doubling and `task.previousException` exposure that this test covers. The check's expected line is updated to `mem: 2 GB (previous: 1073741824)` accordingly.

## Verification

Ran the check locally against the dev build: 3 re-submissions, final attempt prints `mem: 2 GB (previous: 1073741824) (error: ... exit status (137))`, check exits 0.


bentsherman:
Azure credentials expired :smile:

bentsherman (approved):

PR #7270, opened 2026-06-30
docs: Move Static Typing pages to dedicated section

christopher-hakkaart:
Proposal to make Static typing a top-level section.

This treats Static typing more like a feature, with the concept and reference pages living there. Guides/tutorials about migrating are still under those sections.


bentsherman (approved):
Updates:
- flattened "Typed workflows" subsection, now "Static typing" just has a flat list for processes, workflows, and parameters
- Removed new "Typed outputs" page, linked the overview table to the syntax reference instead. Might re-add it in the future but not sure it's worth a new section just to describe the type annotation syntax
- Scaled back some duplicated content
- Fixed a few gaps throughout the static typing docs

PR #7273, opened 2026-06-30
Use scheduler API exit code, fall back to .exitcode file

jordeu:
## Summary

The Seqera (SIC) executor determined a task's exit status only from the `.exitcode` file in the task work directory (`SeqeraTaskHandler.readExitFile()`), returning the `Integer.MAX_VALUE` sentinel when the file could not be read.

The `.exitcode` file is written by the task wrapper's `on_exit` trap (`command-run.txt`), which `printf`s the status to the file and then exits with it. If the container is killed **before** `on_exit` runs — OOM (137), spot reclaim, node loss, timeout — the file is never written; and there is a brief window where the work-dir object is not yet visible. In those cases the real exit code was lost: the task failed with a generic `ProcessException` carrying no numeric exit status, which Nextflow reports as *"terminated for an unknown reason"* and which prevents exit-code-based `errorStrategy`/`maxRetries` (e.g. retry-on-OOM) from working.

This PR makes the scheduler API exit code (`TaskState.getExitCode()`) the source of truth and keeps the `.exitcode` file as the fallback — the **API-first, file-fallback** pattern already used by the AWS Batch, Azure Batch and Google Batch executors, whose exit-status logic is line-for-line equivalent. The API value is already present in the cached task state from the status poll, so **no extra API call** is made; the `.exitcode` file is read only when the API reports no exit code.

## Changes

- `SeqeraTaskHandler.checkIfCompleted()`: use `cachedTaskState.getExitCode()` for the exit status when it is non-null, falling back to `readExitFile()` otherwise. This mirrors `AwsBatchTaskHandler` and `AzBatchTaskHandler` (`apiExitCode != null ? apiExitCode : readExitFile()`). The `Integer.MAX_VALUE` sentinel — and the generic "unknown reason" error it triggers — now occurs only when both the API code is null and the `.exitcode` file is unreadable.
- Added unit tests: API exit code used on a failed task and on a succeeded task when the file is unavailable, API-over-file precedence (the file is not read when the API reports a code), and the file fallback when the API reports no exit code.

## Test plan

- `./gradlew :plugins:nf-seqera:test` — passes.
- Existing `SeqeraTaskHandlerTest` cases (including the no-exit-code error-message path) are unchanged and remain green.


bentsherman:
In recent versions we actually updated most executors to prefer the API exit code over the exitcode file, for example K8s:

https://github.com/nextflow-io/nextflow/blob/17ef32d8b662137e58e890efbb761e8bda86b2a3/plugins/nf-k8s/src/main/nextflow/k8s/K8sTaskHandler.groovy#L434-L442

So it would be more consistent to do the same here, unless you have some specific reason to prefer the exitcode file

jordeu:
Good call, @bentsherman — flipped it in 710f48ace. The SIC executor now prefers the scheduler API exit code and falls back to the `.exitcode` file only when the API doesn't report one, matching the `apiExitCode != null ? apiExitCode : readExitFile()` pattern used by K8s, AWS Batch and Azure Batch. No specific reason to prefer the file — this is more consistent and correct for the OOM/spot-reclaim case where the container dies before the wrapper writes the file. Tests updated to assert the new precedence.

bentsherman (approved):

pditommaso (approved):
Verified the change against the AWS Batch, Azure Batch and Google Batch handlers: this is the same API-first, .exitcode-file-fallback pattern (line-for-line equivalent to AwsBatchTaskHandler:313 and AzBatchTaskHandler:119), using the same Integer.MAX_VALUE sentinel and error-gating. Good fix — recovers the real exit code on OOM/spot-reclaim/timeout so exit-code-based errorStrategy/maxRetries works. Well covered by unit tests.

PR #7278, opened 2026-06-30
docs: deslop semantics reference

christopher-hakkaart:
I pulled this down and ran the Seqera deslop skill we are developing over this PR.

In this PR, it mostly removes prose and shortens/shortens sentences. I've run it leniently, so it looks at the pages as a whole and aligns changes with the Nextflow in-house style (so it's a similar tone/style to other pages in this docs set). It was already pretty good, so changes are minimal.

Otherwise, it reads well and looks good.

### DIFF
--- a/docs/reference/semantics.mdx
+++ b/docs/reference/semantics.mdx
@@ -5,7 +5,7 @@ description: Reference for the semantics of Nextflow language constructs.
 
 # Semantics
 
-This page describes the *semantics* of the Nextflow language -- that is, the meaning and behavior of each language construct. It is a companion to the [Syntax][syntax-page] reference, which describes the *grammar* of the language (how constructs are written). See the [Scripts][script-page] page for a practical introduction to the language.
+The *semantics* of the Nextflow language are the meaning and behavior of each language construct. This page is a companion to the [Syntax][syntax-page] reference, which describes the *grammar* of the language (how constructs are written). See the [Scripts][script-page] page for a practical introduction to the language.
 
 ## Values
 
@@ -105,7 +105,7 @@ See [String][stdlib-types-string] for the available operations.
 
 ### Null
 
-The special value `null` represents the absence of a value. Accessing a property or calling a method on `null` raises a *null reference* error, so `null` is best avoided where possible. The [safe navigation operator](#navigation-operators) can be used to access a property that may be `null`.
+The special value `null` represents the absence of a value. Accessing a property or calling a method on `null` raises a *null reference* error. Avoid `null` where possible. The [safe navigation operator](#navigation-operators) can be used to access a property that may be `null`.
 
 ### Collections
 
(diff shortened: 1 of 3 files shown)


docs/reference/semantics.mdx
```diff
@@ -627,7 +627,7 @@ numbers.inject(0) { acc, v -> acc + v }     // -> 15
 numbers.every { v -> v > 0 }                // -> true
 ```
 
-A higher-order function returns a value rather than mutating external state, which makes it easier to reason about and avoids common sources of error. For example, `inject` accumulates a result without an external accumulator variable:
+A higher-order function returns a value rather than mutating external state. This makes it easier to reason about and avoids common errors. For example, `inject` accumulates a result without an external accumulator variable:
```
bentsherman:
```suggestion
A higher-order function returns a value rather than mutating external state, which makes it easier to reason about and avoids common errors. For example, `inject` accumulates a result without an external accumulator variable:
```

docs/reference/semantics.mdx
```diff
@@ -648,7 +648,7 @@ An [assert statement](./syntax#assert) evaluates a boolean condition and raises
 assert params.input != null : 'The --input parameter is required'
 ```
 
-If the condition is false, an error is raised. The optional message after the colon is included in the error. If no message is given, Nextflow reports the values of the sub-expressions in the condition (a *power assertion*), which makes it easy to see why the assertion failed:
+If the condition is false, an error is raised. The optional message after the colon is included in the error. If no message is given, Nextflow reports the values of the sub-expressions in the condition (a *power assertion*). This makes it easy to see why the assertion failed:
```
bentsherman:
```suggestion
If the condition is false, an error is raised. The optional message after the colon is included in the error. If no message is given, Nextflow reports the values of the sub-expressions in the condition:
```

docs/reference/semantics.mdx
```diff
@@ -734,7 +734,7 @@ println items   // -> [1, 2, 3]
 
 To avoid this, create a copy before mutating, or use operations that return a new value (e.g. `list + [item]` instead of `list.add(item)`).
 
-*Reassigning* a parameter, on the other hand, does not affect the caller. The parameter is a separate reference to the same object, so assigning a new value to it changes only the function's local reference:
+*Reassigning* a parameter, on the other hand, does not affect the caller. The parameter is a separate reference to the same object. Assigning a new value to it changes only the function's local reference:
```
bentsherman:
```suggestion
*Reassigning* a parameter, on the other hand, does not affect the caller. The parameter is a separate reference to the same object, so assigning a new value to it changes only the function's local reference:
```

docs/script.mdx
```diff
@@ -298,19 +298,19 @@ See [Closures][semantics-closures] and [Iteration][semantics-iteration] for more
 
 ## Script definitions
 
-So far, we have been focusing on the basic building blocks of Nextflow code, like variables, strings, collections, and closures.
+The previous sections covered the basic building blocks of Nextflow code: variables, strings, collections, and closures.
 
-In practice, however, Nextflow scripts are composed of *workflows*, *processes*, and *functions* (collectively known as *definitions*), and can *include* definitions from other scripts.
+In practice, Nextflow scripts are composed of *workflows*, *processes*, and *functions* (collectively known as *definitions*), and can *include* definitions from other scripts.
```
bentsherman:
```suggestion
In practice, Nextflow scripts are composed of *workflows*, *processes*, *functions*, and *types* (collectively known as *definitions*), and can *include* definitions from other scripts.
```

PR #7283, opened 2026-07-01
Skip nf-tower when superseded by nf-seqera

jorgee:
## Problem

Since `nf-seqera@1.0.0` the former standalone `nf-tower` plugin is integrated into `nf-seqera`. Nextflow no longer builds or autostarts `nf-tower` (the autostart path in `PluginsFacade` already points at `nf-seqera`), but many existing pipelines / institutional configs still pin it explicitly via `plugins { id 'nf-tower' }`.

When such a config also enables Platform (any of `tower.enabled`, `fusion.enabled`, `TOWER_ACCESS_TOKEN`, or `process.executor = 'seqera'`), `nf-seqera` is auto-loaded **in addition** to the explicitly declared `nf-tower`. Both register a Platform observer, so the run is reported to Seqera Platform **twice**.

## Change

In `PluginsFacade.pluginsRequirement()`, after the plugin requirements are assembled, drop `nf-tower` whenever `nf-seqera` (>=1.0.0) is also requested. This single chokepoint covers every combination of explicit declaration + the four autostart triggers, since they all feed the same `specs` list.

- `dropSupersededTowerPlugin(specs)` — removes `nf-tower` and logs a `warn` when superseded.
- `isSeqeraSupersedingTower(version)` — true for a null/blank (default bundled) version or `>=1.0.0`; a hypothetical `nf-seqera@0.x` pin leaves `nf-tower` untouched.

## Tests

`PluginsFacadeTest`:
- `nf-tower` + `tower.enabled=true` → `[nf-seqera]`
- `nf-tower` + `TOWER_ACCESS_TOKEN` set → `[nf-seqera]`
- both plugins declared explicitly → `[nf-seqera]`
- `nf-tower` alone → untouched
- `nf-seqera@0.1.0` + `nf-tower` → both kept (version does not supersede)
- data-driven version matrix for `isSeqeraSupersedingTower`

### DIFF
--- a/modules/nf-commons/src/main/nextflow/plugin/PluginsFacade.groovy
+++ b/modules/nf-commons/src/main/nextflow/plugin/PluginsFacade.groovy
@@ -27,6 +27,7 @@ import nextflow.SysEnv
 import nextflow.exception.AbortOperationException
 import nextflow.extension.Bolts
 import nextflow.extension.FilesEx
+import nextflow.util.VersionNumber
 import org.pf4j.DefaultPluginManager
 import org.pf4j.PluginManager
 import org.pf4j.PluginState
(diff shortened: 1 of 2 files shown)


jorgee:
I have run e2e in stage an the duplicated platform telemetry issue is not reproduced

bentsherman (approved):

PR #7289, opened 2026-07-01
Add docker.cpuLimits config option

### LINKED ISSUE #5323: Change behavior of `process.cpus` back to `--cpus` when using Docker with a local executor
## New feature

Recently Nextflow changed the behavior of `process.cpus` to use `--cpu-shares` instead of `--cpus` when using Docker. I understand the benefits of this when using a cloud or HPC environment, but this doesn't make much sense for a local executor when the use of this parameter is to _limit_ the number of CPUs this process will use.

This change effectively removed any ability to limit the impact on the system cpu-wise when running on a local machine. This change results in degradation of the functionality of Nextflow in this case.

## Suggest implementation 

I suggest changing it back to `--cpus` passed to Docker for the local executor.

bentsherman:
Fix #5323 

This PR adds the `docker.cpuLimits` config option, which uses `--cpus` to limit CPU usage of Docker containers instead of `--cpu-shares`. This approach is preferred when using the local executor, as noted in the linked issue

Also removes the `NXF_DOCKER_LEGACY` env var, which no longer works since 25.10 and is undocumented anyway

### DIFF
--- a/docs/reference/config.mdx
+++ b/docs/reference/config.mdx
@@ -824,6 +824,8 @@ The following settings are available:
 
 ##### `docker.cpuLimits`
 
+<AddedInVersion version="26.10" />
+
 Use the `--cpus` flag to limit CPU usage instead of `--cpu-shares` (default`false`).
 
 Unlike `--cpu-shares`, which only affects the relative scheduling weight of a container, `--cpus` enforces a hard limit on the number of CPUs available to it. See the [Docker documentation](https://docs.docker.com/engine/containers/resource_constraints/#cpu) for details.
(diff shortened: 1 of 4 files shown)


christopher-hakkaart (approved):
Docs look good

pditommaso (approved):

docs/reference/config.mdx
```diff
@@ -822,6 +822,12 @@ The `docker` scope controls how [Docker](https://www.docker.com) containers are
 
 The following settings are available:
 
+##### `docker.cpuLimits`
```
bentsherman, line 825:
```suggestion
##### `docker.cpuLimits`

<AddedInVersion version="26.10" />
```

PR #7290, opened 2026-07-01
Document exit logic in task wrapper script

### LINKED ISSUE #5556: Improve docs and rationale about exit logic in Bash unstage function
The unstage logic introduced by #5345 is not trivial, see here:

[#5345](https://github.com/nextflow-io/nextflow/pull/5345/files#diff-45cb34df99cf9da6b8e6e4200508360f046fc91277c094df28ec634d3af91ccbR102-R108)

The goal of this task is to improve the description about the different of `last_err` vs `nxf_main_ret` vs `nxf_unstage_ret` and the general logic.

bentsherman:
Close #5556 

Runtime behavior is unchanged since comments are stripped from the rendered wrapper script

### DIFF
--- a/modules/nextflow/build.gradle
+++ b/modules/nextflow/build.gradle
@@ -93,6 +93,15 @@ dependencies {
 test {
     minHeapSize = "512m"
     maxHeapSize = "4096m"
+    // Dev-mode plugin discovery reads plugins/*/build/resources/main/META-INF/MANIFEST.MF at
+    // runtime. Each plugin's copyPluginManifest task (re)writes that file by truncating and
+    // copying; when a full `gradle test` run schedules both, order this task after them so a
+    // manifest is never read mid-write, which otherwise surfaces intermittently as the pf4j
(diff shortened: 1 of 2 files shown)


pditommaso (approved):
LGTM — verified the comments against the actual bash and the rendering path:

- `BashTemplateEngine` strips all `## `-prefixed lines, so the rendered wrapper is unchanged. ✓
- `on_exit` precedence, the `0`-of-last-command note, and the `set -e`/`set -u` fallback path all match the code. ✓
- `nxf_unstage` conditional (outputs only when `nxf_main_ret==0`, controls unconditional) is accurate. ✓

Nice-to-have (non-blocking): the `nxf_unstage_controls` comment says its exit status "becomes the generic `last_err` fallback in on_exit", but that only holds when no after script is defined — with `{{after_script}}` present, it runs after `nxf_unstage`, so `$?` reflects the after script and a controls failure is lost rather than surfaced as `last_err`. Could tighten to "…unless an after script runs later."

PR #7291, opened 2026-07-01
docs: Add note about IAM roles for google.batch.serviceAccountEmail

### LINKED ISSUE #4603: PERMISSION_DENIED: caller does not have access to act as the specified service account:
## Bug report 

Prior to running my Nextflow pipeline, I'm setting my GCP service account credentials via `$GOOGLE_APPLICATION_CREDENTIALS`, as stated in the [Nextflow docs](https://www.nextflow.io/docs/latest/google.html).

For the sake of this issue, my service account is `nextflow@XXX.iam.gserviceaccount.com`. 

However, when I run my Nextflow pipeline, I get the following error:

```
Caused by:
  PERMISSION_DENIED: caller does not have access to act as the specified service account: "default@XXX.gserviceaccount.com"
```

For the sake of this issue, `default@XXX.gserviceaccount.com` refers to the "default" service account that I was previously using, but I switched to `nextflow@XXX.iam.gserviceaccount.com`, which has more restricted permissions. 

I've double checked, and `$GOOGLE_APPLICATION_CREDENTIALS` is set to `nextflow@XXX.iam.gserviceaccount.com`.  I've also tried on multiple machines (local and Github codespace). In both cases, setting `$GOOGLE_APPLICATION_CREDENTIALS` to `nextflow@XXX.iam.gserviceaccount.com` still results in an error message pointing to `default@XXX.gserviceaccount.com`. 

I've also tried using `gcloud auth activate-service-account` to set the service account to `nextflow@XXX.iam.gserviceaccount.com` (along with `$GOOGLE_APPLICATION_CREDENTIALS`), but I'm still getting the same permissions error.

Changing the GCP service account via the `$GOOGLE_APPLICATION_CREDENTIALS` environmental variable should actually change the service account, and not result in a `PERMISSION_DENIED:` error for a different service account. 

### Environment 

* Nextflow version: 23.10.0
* Java version: openjdk version "21-internal" 2023-09-19
* Operating system: Ubuntu 22.04.3 
* Bash version: 5.1.16(1)-release (x86_64-pc-linux-gnu)

### LINKED ISSUE #5622: Docs: permissions required for service account if using google.batch.serviceAccountEmail
## New feature

The docs at https://www.nextflow.io/docs/latest/google.html#google-cloud and https://seqera.io/blog/nextflow-with-gbatch/ do not state the permissions required if using `google.batch.serviceAccountEmail` to change the service account used for each GCP Batch job, versus each batch job using the default account. `google.batch.serviceAccountEmail`  is useful when one needs to access GCP resources from batch jobs, and has specific IAM roles set for the target service account to be used for the batch jobs (instead of using the default service account).

Without the correct IAM roles set for the account use via `google.batch.serviceAccountEmail`, the GCP Batch jobs just stay in the `scheduled` state forever. 

## Usage scenario 

Any time one wants to use `google.batch.serviceAccountEmail`. 

## Suggest implementation 

More docs at https://www.nextflow.io/docs/latest/google.html#google-cloud and/or https://seqera.io/blog/nextflow-with-gbatch/. 

Maybe also more info at:

> google.batch.serviceAccountEmail
> Define the Google service account email to use for the pipeline execution. If not specified, the default Compute Engine service account for the project will be used.

> Note that the google.batch.serviceAccountEmail service account will only be used for spawned jobs, not for the Nextflow process itself. See the [Google Cloud](https://www.nextflow.io/docs/latest/google.html#credentials) documentation for more information on credentials.

bentsherman:
Close #5622

### DIFF
--- a/docs/reference/config.mdx
+++ b/docs/reference/config.mdx
@@ -1262,6 +1262,9 @@ Network tags are ignored when using instance templates.
 ##### `google.batch.serviceAccountEmail`
 
 The Google service account email to use for the pipeline execution. If not specified, the default Compute Engine service account for the project will be used.
+
+The service account must have the IAM roles listed in the [Google Cloud Batch documentation for Nextflow](https://cloud.google.com/batch/docs/nextflow), otherwise Cloud Batch jobs can remain stuck in the `SCHEDULED` state indefinitely.
+
 This service account will only be used for tasks submitted by Nextflow, not for Nextflow itself. See [Credentials][google-credentials] for more information on Google Cloud credentials.
 
 ##### `google.batch.spot`
--- a/docs/reference/config.mdx
+++ b/docs/reference/config.mdx
@@ -1263,7 +1263,7 @@ Network tags are ignored when using instance templates.
 
 The Google service account email to use for the pipeline execution. If not specified, the default Compute Engine service account for the project will be used.
 
-The service account must have the IAM roles listed in the [Google Cloud Batch documentation for Nextflow](https://cloud.google.com/batch/docs/nextflow), otherwise Cloud Batch jobs can remain stuck in the `SCHEDULED` state indefinitely.
+The service account must have the IAM roles listed in the [Google Cloud Batch documentation for Nextflow](https://cloud.google.com/batch/docs/nextflow). Without these roles, Cloud Batch jobs stay in the `SCHEDULED` state and never run.
 
 This service account will only be used for tasks submitted by Nextflow, not for Nextflow itself. See [Credentials][google-credentials] for more information on Google Cloud credentials.
 


docs/reference/config.mdx
```diff
@@ -1262,6 +1262,9 @@ Network tags are ignored when using instance templates.
 ##### `google.batch.serviceAccountEmail`
 
 The Google service account email to use for the pipeline execution. If not specified, the default Compute Engine service account for the project will be used.
+
+The service account must have the IAM roles listed in the [Google Cloud Batch documentation for Nextflow](https://cloud.google.com/batch/docs/nextflow), otherwise Cloud Batch jobs can remain stuck in the `SCHEDULED` state indefinitely.
```
```suggestion
The service account must have the IAM roles listed in the [Google Cloud Batch documentation for Nextflow](https://cloud.google.com/batch/docs/nextflow). Without these roles, Cloud Batch jobs stay in the `SCHEDULED` state and never run.
```

christopher-hakkaart (approved):
Suggestion - otherwise looks good.

PR #7294, opened 2026-07-02
Exclude null config env entries from task environment

### LINKED ISSUE #5722: env with non-existent host environment variable leads to warnings

If I reference a host env var in config like:

```
env {
  OPENAI_API_KEY = env(OPENAI_API_KEY)
}
```

, and if the the host variable does not actually exist, `env()` returns null. Then, when we run the workflow, we get messages like:

```
WARN: Environment variable `OPENAI_API_KEY` evaluates to an empty value
```

Given that this variable is not set on the host, I would expect it to end up not being set at all. 

- Trigger any workflow with the above config without setting the env var.
- Observe the error. 

### Program output 

```
WARN: Environment variable `OPENAI_API_KEY` evaluates to an empty value
```

### Environment 

* Nextflow version: 24.11.0-edge

### Additional context

Related to propagation of keys for MulitQC AI summary functionality

bentsherman:
Fix #5722 

When an `env` config setting references an unset host environment var via `env('HOST_VAR')`, the resolved value is null.

This PR updates the TaskProcessor to exclude these null values from the task environment instead of adding them with a null value. The current behavior causes a spurious "evaluates to an empty value" warning.

Explicit empty-string values (e.g. `env.FOO = ''`) are unaffected.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -1606,7 +1606,11 @@ class TaskProcessor {
         // add the taskConfig environment entries
         if( session.config.env instanceof Map ) {
             session.config.env.each { name, value ->
-                result.put( name, value?.toString() )
+                // skip entries with a null value (e.g. `env(HOST_VAR)` referencing a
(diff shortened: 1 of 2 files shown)


pditommaso:
Worth adding a unit test to capture this logic

pditommaso (approved):

PR #7295, opened 2026-07-02
Round up CPUs to nearest valid value for AWS Batch Fargate

### LINKED ISSUE #5781: AWS Batch Fargate requires exact CPU values?
When running wf-human-variation from ONT on AWSBatch Fargate, I get this error

```
ERROR ~ Error executing process > 'ingress:minimap2_alignment (1)'

Caused by:
  Requirement of 12 CPUs is not allowed by Fargate -- Check process with name 'ingress:minimap2_alignment (1)'
```

From what I can tell it's because the requested CPU value does not match one from the list found here:

https://github.com/nextflow-io/nextflow/blob/944f48f9b52d47032bf57e16cfe11fe9410972b2/plugins/nf-amazon/src/main/nextflow/cloud/aws/batch/AwsBatchTaskHandler.groovy#L923-L943

It seems a bit restrictive to require specific CPU values, am I doing something wrong or is that in fact the case?

bentsherman:
Fix #5781

This PR normalizes the CPU request for tasks executed via AWS Fargate. It rounds the request up to the nearest accepted value (1, 2, 4, 8, 16) instead of failing the run.

It only fails when the request exceeds the max accepted value (16).


pditommaso (approved):
LGTM — clean, well-scoped fix with good test coverage. The rounding logic is correct and the normalized `cpus` value is now consistently used for both the memory-slot lookup and the VCPU requirement.

One minor, non-blocking observation: `normaliseFargateMem` emits a `log.warn` when it clamps memory, but `normaliseFargateCpus` rounds the request up silently. A user asking for `cpus 12` gets 16 vCPUs (and a knock-on 32 GB memory floor) with no log line explaining why. Consider adding a `log.warn` mirroring the memory one — purely a UX nicety, the behavior is correct as-is.

bentsherman:
Thanks for the review! Addressed the suggestion in e16f36b4f — `normaliseFargateCpus` now emits a `log.warn` when it rounds the request up, mirroring the memory-clamping warning:

```
Process '<name>' cpus requirement of 12 is not allowed by Fargate and will be rounded up to 16
```

PR #7296, opened 2026-07-02
Add graceful fallback for array directive with unsupported executors

### LINKED ISSUE #5924: Graceful fallback when `array` is used on unsupported executors

bentsherman:
Fix #5924 

This PR allows the `array` directive to gracefully degrade when applied to unsupported executors (e.g. `local`)

This makes it easier to enable array jobs across the board with a single setting, instead of having to specify it for each process that can use it. Especially useful when e,g, a pipeline has a mix of local jobs and SLURM jobs.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
@@ -62,6 +62,7 @@ import nextflow.exception.ProcessUnrecoverableException
 import nextflow.executor.CachedTaskHandler
 import nextflow.executor.Executor
 import nextflow.executor.StoredTaskHandler
+import nextflow.executor.TaskArrayExecutor
 import nextflow.extension.CH
 import nextflow.extension.DataflowHelper
 import nextflow.file.FileHelper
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/main/groovy/nextflow/processor/TaskProcessor.groovy
```diff
@@ -298,7 +299,14 @@ class TaskProcessor {
...
         final arraySize = config.getArray()
-        this.arrayCollector = arraySize > 0 ? new TaskArrayCollector(this, executor, arraySize) : null
+        if( arraySize > 0 && executor instanceof TaskArrayExecutor ) {
+            this.arrayCollector = new TaskArrayCollector(this, executor, arraySize)
+        }
+        else {
+            if( arraySize > 0 )
+                log.warn "Executor '${executor.name}' does not support job arrays -- the array directive will be ignored for process '$name'"
+            this.arrayCollector = null
+        }
```
pditommaso: Nit, consider wrapping into a helper to make more readable

pditommaso (approved):

PR #7300, opened 2026-07-03
Revert the stream swap after the redirect to .command.out and .command.err

### LINKED ISSUE #7299: stdout and stderr are swapped by .command.run in SLURM
## Bug report 

When nextflow execute a process in SLURM the `.command.run` redirects the stdout and stderr to `.command.out` and `.command.err` respectively. But while doing so it also swaps the two streams  which mean they are not sent correctly to any downstream capture.

`.command.run` should execute the `.command.sh` capture copies of the output and error streams and leave them in their assigned stream.

```nextflow
params.outdir = "logs"
def outdir = file(params.outdir).toAbsolutePath().normalize()

workflow {
    // clusterOptions writes directly to this path when the job starts,
    // so it must exist before the job is submitted.
    outdir.mkdirs()

    ECHO_STREAMS()
}

process ECHO_STREAMS {
    clusterOptions {
        [
            "--output=${outdir}/stdout.log".toString(),
            "--error=${outdir}/stderr.log".toString()
        ]
    }

    script:
    """
    echo "STDOUT_MESSAGE: this line was written to stdout"
    echo "STDERR_MESSAGE: this line was written to stderr" 1>&2
    """
}
```

```bash
NXF_SYNTAX_PARSER=v1  nextflow run reproduce_nextflow_stdout_stderr_swap.nf --outdir log
cat log/stdout.log
cat log/stderr.log
```
### Program output 
Note the order of the two message compared to the order of the two command above.
```
STDERR_MESSAGE: this line was written to stderr
STDOUT_MESSAGE: this line was written to stdout
```

### Environment 

* Nextflow version: 26.04.4
* Java version: 21.0.3
* Operating system: Linux
* Bash version: GNU bash, version 5.1.8

### Additional context

I ran this on SLURM to capture the output of `.command.run` but this probably applies elsewhere. 
I believe that the culprit is this line in the .command.run: 
```bash 
(set -o pipefail; (nxf_launch | tee .command.out) 3>&1 1>&2 2>&3 | tee .command.err)
```

tcezard:
fixes #7299 
This PR attenpt to fix the stream swap when executing .command.sh in .command.run
It does not change the similar stream swap in `nxf_unstage` but that might be needed as well.


modules/nextflow/src/test/resources/nextflow/executor/test-bash-wrapper-with-trace.txt
```diff
@@ -151,7 +151,7 @@ nxf_trace_linux() {
     local io_stat0=($(2> /dev/null < /proc/$pid/io sed 's/^.*:\s*//' | head -n 6 | tr '\n' ' ' || echo -n '0 0 0 0 0 0'))
     local start_millis=$(nxf_date)
     trap 'kill $mem_proc' ERR
-    
+
```
jordeu: This change is unnecessary, and it will break a test

tcezard: Added the spaces back

jordeu (approved):

Issue #2, opened 2014-02-28
Add git commit hash number to version information


pditommaso:
Modify the build script in such a way to include the git commit number in version information number

pditommaso:
Outdated

PR #7308, opened 2026-07-08
feat: add plugin directory trust guard to harden shared plugin caches

aknownuser:
## Summary

Nextflow loads and executes plugin code from the local plugin cache (`$id-$version`) **without re-verifying its origin** when the directory already exists — `PluginUpdater.load0()` short-circuits the download (and therefore the `sha512sum`/`CompoundVerifier` path) whenever the cached directory is present, and `PluginsFacade.isAllowed()` only matches plugin IDs, not content.

On a **shared, writable** plugin cache (`NXF_PLUGINS_DIR` or `$NXF_HOME/plugins`), a lower-trust local user can pre-populate an official pinned plugin coordinate (e.g. `nf-amazon@3.10.0`) with attacker-controlled code that then runs inside a victim's Nextflow process.

The plugin cache is expected to be a **per-user, private trust boundary** (default `$HOME/.nextflow/plugins`). This PR adds a defense-in-depth guard that verifies a plugin directory is trusted before loading it, addressing the exact precondition of the issue — write access to a trusted cache — with no network calls, no checksums, and no persisted state (works offline).

## Changes and Reasoning

**New `PluginSecurity` helper** (`modules/nf-commons/.../plugin/PluginSecurity.groovy`)
A directory is considered trusted only when it is **owned by the current user (or root)** *and* **not writable by group or others**. This accepts the two legitimate deployment shapes — a private user cache (e.g. `0700`) and an admin-managed **read-only** shared cache (e.g. `root:root 0755`) — while rejecting an attacker-owned directory or a group/world-writable cache.

> The both-conditions check matters: a permissions-only check would miss the reported scenario, where the poisoned directory was `attacker:attacker 755` (not group/other-writable) — ownership is what made it dangerous.

**Call sites**
- `PluginsFacade.init(...)` — guards the plugins **root** at setup (both `init` overloads), so startup-loaded plugins are covered.
- `PluginUpdater.load0()` — guards each `$id-$version` directory immediately before `loadPluginFromPath`, covering cache reuse and the attacker-owned-subdir case.

**New env var `NXF_PLUGINS_STRICT_MODE`** (`warn` | `strict` | `off`)
- `warn` (default) — logs a warning and continues (zero breakage for existing setups).
- `strict` — aborts the run with `AbortOperationException`.
- `off` — disables the check.

The guard is a no-op on filesystems without POSIX permissions and in **dev mode** (where plugins load from the source tree, not the cache).

**Compatibility / rollout**
Default `warn` means no existing deployment breaks; the default `$HOME/.nextflow/plugins` (user-owned) is silent. `strict` is opt-in now; a future stable release can promote the default.

## Testing

New `PluginSecurityTest` (16 cases) and existing plugin tests pass:

- ✅ private `0700` / read-only `0755` → accepted in all modes
- ✅ group/world-writable → warns in `warn`, throws in `strict`
- ✅ `off` mode → never throws
- ✅ null dir / non-POSIX FS → no-op
- ✅ env-var mode resolution (`warn` default, `strict`, `off`)
- ✅ `:nf-commons:test` full suite — no regression in `PluginsFacadeTest` / `PluginUpdaterTest`

End-to-end (dev launcher, `NXF_PLUGINS_MODE=prod`):

| Scenario | Result |
|---|---|
| `strict` + `0777` cache | Aborts — workflow does not run |
| `warn` + `0777` cache | Warns, workflow proceeds |
| `strict` + private `0700` cache | Silent, workflow proceeds |

## Documentation

- `docs/reference/env-vars.mdx` — new `NXF_PLUGINS_STRICT_MODE` entry.
- `docs/plugins/using-plugins.mdx` — security note in the **Caching** section (keep the cache private; supported shared pattern is admin-owned read-only).

### DIFF
--- a/docs/plugins/using-plugins.mdx
+++ b/docs/plugins/using-plugins.mdx
@@ -80,6 +80,7 @@ To use Nextflow plugins in an offline environment:
     Nextflow will attempt to download newer versions of plugins if their versions are not set. See [Identifiers][using-plugins-identifiers] for more information.
     :::
 
+[env-vars]: ../reference/env-vars
 [install-standalone]: ../install#standalone-distribution
 [using-plugins-config]: ./using-plugins#configuration
 [using-plugins-identifiers]: ./using-plugins#identifiers
--- a/docs/plugins/using-plugins.mdx
+++ b/docs/plugins/using-plugins.mdx
@@ -57,9 +57,9 @@ Plugin declarations in Nextflow configuration files are ignored when specifying 
 When Nextflow downloads plugins, it caches them in the directory specified by `NXF_PLUGINS_DIR` (`$HOME/.nextflow/plugins` by default).
 
 :::warning
-The plugin cache should be a private, per-user directory that is not writable by other users. Nextflow loads and executes cached plugin code without re-verifying its origin, so a shared plugin cache that is writable by other users (or owned by another user) allows a lower-trust user to replace cached plugin code with their own, which would then run inside your Nextflow process.
+The plugin cache should be a private, per-user directory that other users cannot write to. Nextflow loads and executes cached plugin code without re-verifying its origin. If the cache is writable by other users (or owned by another user), a lower-trust user can replace the cached code with their own, and that code then runs inside your Nextflow process.
 
-Keep `NXF_PLUGINS_DIR` owned by you and not world-writable (a group-writable directory owned by you, as created under the common umask 002, is fine). If a shared cache is required, use a directory owned by an administrator or a dedicated service account that regular users cannot modify, and add that owner to [`NXF_PLUGINS_TRUSTED_OWNERS`](../reference/env-vars.mdx). Use [`NXF_PLUGINS_STRICT_MODE`](../reference/env-vars.mdx) to control whether Nextflow warns about or refuses to load plugins from an untrusted directory.
+Keep `NXF_PLUGINS_DIR` owned by you and not world-writable. A group-writable directory owned by you, as created under the common umask 002, is fine. If you need a shared cache, use a directory owned by an administrator or a dedicated service account that regular users cannot modify, and add that owner to [`NXF_PLUGINS_TRUSTED_OWNERS`][env-vars]. Use [`NXF_PLUGINS_STRICT_MODE`][env-vars] to control whether Nextflow warns about or refuses to load plugins from an untrusted directory.
 :::
 
 ## Offline usage
--- a/docs/plugins/using-plugins.mdx
+++ b/docs/plugins/using-plugins.mdx
@@ -56,6 +56,12 @@ Plugin declarations in Nextflow configuration files are ignored when specifying 
 
 When Nextflow downloads plugins, it caches them in the directory specified by `NXF_PLUGINS_DIR` (`$HOME/.nextflow/plugins` by default).
 
+:::warning
+The plugin cache should be a private, per-user directory that is not writable by other users. Nextflow loads and executes cached plugin code without re-verifying its origin, so a shared plugin cache that is writable by other users (or owned by another user) allows a lower-trust user to replace cached plugin code with their own, which would then run inside your Nextflow process.
+
+Keep `NXF_PLUGINS_DIR` private (for example `chmod 700`) and owned by you. If a shared cache is required, use an administrator-owned, **read-only** directory that regular users cannot modify. Use [`NXF_PLUGINS_STRICT_MODE`](../reference/env-vars.mdx) to control whether Nextflow warns about or refuses to load plugins from an untrusted directory.
+:::
+
 ## Offline usage
 
 When running Nextflow in an offline environment, any required plugins must be downloaded and moved into the offline environment prior to any runs.
--- a/docs/plugins/using-plugins.mdx
+++ b/docs/plugins/using-plugins.mdx
@@ -59,7 +59,7 @@ When Nextflow downloads plugins, it caches them in the directory specified by `N
 :::warning
 The plugin cache should be a private, per-user directory that is not writable by other users. Nextflow loads and executes cached plugin code without re-verifying its origin, so a shared plugin cache that is writable by other users (or owned by another user) allows a lower-trust user to replace cached plugin code with their own, which would then run inside your Nextflow process.
 
-Keep `NXF_PLUGINS_DIR` private (for example `chmod 700`) and owned by you. If a shared cache is required, use an administrator-owned, **read-only** directory that regular users cannot modify. Use [`NXF_PLUGINS_STRICT_MODE`](../reference/env-vars.mdx) to control whether Nextflow warns about or refuses to load plugins from an untrusted directory.
+Keep `NXF_PLUGINS_DIR` owned by you and not world-writable (a group-writable directory owned by you, as created under the common umask 002, is fine). If a shared cache is required, use a directory owned by an administrator or a dedicated service account that regular users cannot modify, and add that owner to [`NXF_PLUGINS_TRUSTED_OWNERS`](../reference/env-vars.mdx). Use [`NXF_PLUGINS_STRICT_MODE`](../reference/env-vars.mdx) to control whether Nextflow warns about or refuses to load plugins from an untrusted directory.
 :::
 
 ## Offline usage
--- a/modules/nf-commons/src/main/nextflow/plugin/PluginSecurity.groovy
+++ b/modules/nf-commons/src/main/nextflow/plugin/PluginSecurity.groovy
@@ -56,13 +56,27 @@ class PluginSecurity {
     static final String MODE_STRICT = 'strict'
     static final String MODE_OFF = 'off'
 
+    static final Set<String> MODES = [MODE_WARN, MODE_STRICT, MODE_OFF] as Set
+
     /**
      * Resolve the strict mode from the {@code NXF_PLUGINS_STRICT_MODE} environment variable.
+     *
+     * The value is normalised (lower-cased and trimmed) and validated against the known modes.
+     * An unset or blank value falls back to the default silently, while an unrecognised value
+     * (e.g. a typo) falls back to the default and logs a warning so it is not silently ignored.
      *
      * @return one of {@code warn} (default), {@code strict} or {@code off}
      */
     static String getMode() {
-        return SysEnv.get('NXF_PLUGINS_STRICT_MODE', MODE_WARN)
+        final raw = SysEnv.get('NXF_PLUGINS_STRICT_MODE')
+        final mode = raw?.toLowerCase()?.trim()
+        if( !mode )
+            return MODE_WARN
+        if( !MODES.contains(mode) ) {
+            log.warn "Invalid NXF_PLUGINS_STRICT_MODE value '${raw}' - expected one of ${MODES.join(', ')}; using default '${MODE_WARN}'"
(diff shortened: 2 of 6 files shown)


modules/nf-commons/src/main/nextflow/plugin/PluginSecurity.groovy
```diff
@@ -0,0 +1,116 @@
...
+    static final String MODE_STRICT = 'strict'
+    static final String MODE_OFF = 'off'
+
+    /**
+     * Resolve the strict mode from the {@code NXF_PLUGINS_STRICT_MODE} environment variable.
+     *
+     * @return one of {@code warn} (default), {@code strict} or {@code off}
+     */
+    static String getMode() {
+        return SysEnv.get('NXF_PLUGINS_STRICT_MODE', MODE_WARN)
```
jorgee: Looking at the code, it seems that typos, different case and empty string will silently behave as warn mode.  It would be better having a `.toLowerCase()?.trim()` and validate against the known set, logging a warning on an unrecognized value.

modules/nf-commons/src/test/nextflow/plugin/PluginSecurityTest.groovy
```diff
@@ -0,0 +1,141 @@
...
+class PluginSecurityTest extends Specification {
+
+    private Path tempDirWithPerms(String perms) {
+        final dir = Files.createTempDirectory('test-plugin-sec')
+        Files.setPosixFilePermissions(dir, PosixFilePermissions.fromString(perms))
+        return dir
+    }
+
+    @Unroll
+    def 'should accept a private dir owned by current user - mode=#MODE perms=#PERMS' () {
```
jorgee: Current test are checking the permision branch but no test is checking the untrustedOwner branch, maybe good to add a test for this case.

aknownuser: Done. now its checking the untrustedowner

pditommaso (changes requested):
The trust-guard's intent is sound and the `warn`/`strict`/`off` plumbing is clean. My concern is the **trust predicate itself: it produces false positives in the exact deployment shapes Nextflow is most used in** — spurious warnings on every run in the default `warn` mode, and *aborted* runs under `NXF_PLUGINS_STRICT_MODE=strict`. Requesting changes to tighten the predicate, plus a note on a stronger long-term direction.

## Blocking — false positives

1. **Group-writable check breaks the umask-002 default** — `PluginSecurity.isUntrusted` (`PluginSecurity.groovy:94`). On stock Ubuntu/Debian/RHEL/Fedora regular-user sessions (umask `002`, user-private-group), `FilesEx.mkdirs(root)` creates `~/.nextflow/plugins` as `rwxrwxr-x`. The guard flags the directory Nextflow *just created itself* — every `run`/`config`/`info`/`log`/`clean`, and every freshly-downloaded plugin dir (`PluginUpdater.groovy:406`). A dir the user owns is theirs to make group-writable; the threat is a *foreign* writer.

2. **Owner compared by name, not UID** — `PluginSecurity.groovy:126`. When the running UID has no `/etc/passwd` entry (containers, OpenShift, K8s, arbitrary-UID HPC — very common for Nextflow), `System.getProperty('user.name')` returns `?` while `attrs.owner().name` resolves to the numeric UID string. Mismatch → the user's own private cache is flagged. Same issue under NFS/LDAP/AD name resolution.

3. **Root running against a user-owned cache** — `PluginSecurity.groovy:95`. `sudo nextflow …` with the user's own `0700` cache: `me` = `root`, owner = `alice`, so `alice != me && alice != 'root'` → flagged.

4. **Only literal `root` is exempted, but the javadoc promises admin-managed shared caches** — `PluginSecurity.groovy:95`. A read-only shared cache owned by a service account (e.g. `nextflow:nextflow 0755`) is rejected for every non-admin user. Under site-wide `strict`, every `nextflow run` aborts with "refusing to load plugins" — the exact deployment the class doc says is supported.

5. **Per-plugin invocation floods duplicate warnings** — `PluginUpdater.groovy:406`. `checkTrustedDir` runs once per plugin/dependency in `load0`, so one underlying condition emits N identical warnings per run, and `getMode()` re-reads the env and re-emits the "Invalid NXF_PLUGINS_STRICT_MODE" warning N times too.

### Suggested fixes (cheap, keep it offline-safe)
- Compare owner by **UID**, not name → fixes #2 and #3.
- Treat group/other-write as untrusted **only when the owner isn't you** → kills the umask-002 storm (#1).
- Make the trusted-owner set configurable, or trust a read-only dir regardless of owner → #4.
- Warn once per run, not per plugin → #5.

## Better long-term strategy

Worth noting for direction (not required in this PR): download-time integrity is **already** covered — the registry serves `sha512sum` (`HttpPluginRepository.groovy:220`) and pf4j's `CompoundVerifier` checks the archive on download. This guard's real job is the *post-download cache-poisoning* window.

A load-time "checksum vs registry" would defend that window but requires re-fetching a trusted baseline from the registry on every run — which **breaks offline/air-gapped use** (a documented, supported mode) and makes registry availability a hard runtime dependency.

The genuinely stronger answer to supply-chain integrity is a **committed lockfile with pinned plugin hashes**, verified offline at load (the `go.sum` / `package-lock.json` model): the trusted baseline lives in the pipeline repo under version control, so it survives cache poisoning *and* registry compromise *and* silent drift, with no network. That's a separate, ADR-worthy feature — but it's the direction I'd aim for. This directory guard is a fine cheap complement in the meantime, once the false positives above are fixed.

aknownuser:
Yes the long term strategy you suggested is the ideal situation, an ADR for that would be perfect! 

I have addressed the false positives you detected. Let me know if you think this still needs work.

You may have seen that this setup is opt-in, which is made this way to ovoid any breaking change to previous usage of nextflow.

docs/plugins/using-plugins.mdx
```diff
@@ -56,6 +56,12 @@ Plugin declarations in Nextflow configuration files are ignored when specifying
 
 When Nextflow downloads plugins, it caches them in the directory specified by `NXF_PLUGINS_DIR` (`$HOME/.nextflow/plugins` by default).
 
+:::warning
+The plugin cache should be a private, per-user directory that is not writable by other users. Nextflow loads and executes cached plugin code without re-verifying its origin, so a shared plugin cache that is writable by other users (or owned by another user) allows a lower-trust user to replace cached plugin code with their own, which would then run inside your Nextflow process.
+
+Keep `NXF_PLUGINS_DIR` owned by you and not world-writable (a group-writable directory owned by you, as created under the common umask 002, is fine). If a shared cache is required, use a directory owned by an administrator or a dedicated service account that regular users cannot modify, and add that owner to [`NXF_PLUGINS_TRUSTED_OWNERS`](../reference/env-vars.mdx). Use [`NXF_PLUGINS_STRICT_MODE`](../reference/env-vars.mdx) to control whether Nextflow warns about or refuses to load plugins from an untrusted directory.
+:::
```
christopher-hakkaart, line 63:
```suggestion
:::warning
The plugin cache should be a private, per-user directory that other users cannot write to. Nextflow loads and executes cached plugin code without re-verifying its origin. If the cache is writable by other users (or owned by another user), a lower-trust user can replace the cached code with their own, and that code then runs inside your Nextflow process.

Keep `NXF_PLUGINS_DIR` owned by you and not world-writable. A group-writable directory owned by you, as created under the common umask 002, is fine. If you need a shared cache, use a directory owned by an administrator or a dedicated service account that regular users cannot modify, and add that owner to [`NXF_PLUGINS_TRUSTED_OWNERS`][env-vars]. Use [`NXF_PLUGINS_STRICT_MODE`][env-vars] to control whether Nextflow warns about or refuses to load plugins from an untrusted directory.
:::
```

docs/plugins/using-plugins.mdx
christopher-hakkaart, line 82:
```suggestion

[env-vars]: ../reference/env-vars
```

christopher-hakkaart (commented):
Adding some style alignment suggestions.

pditommaso:
Opened #7394 implementing the committed-lockfile direction raised in [this review](https://github.com/nextflow-io/nextflow/pull/7308#pullrequestreview-4654989911) — a `plugins.lock` pinning per-archive SHA-512, verified offline, populated trust-on-first-use (go.sum model, no dedicated command). It's a **complement** to this PR's directory guard: #7394 covers the archive (supply-chain / registry compromise / drift), while this PR covers the extracted tree that actually executes. Neither replaces the other.

pditommaso:
Following up on the long-term direction from [my review above](https://github.com/nextflow-io/nextflow/pull/7308#pullrequestreview-4654989911): I've opened #7394, which implements the committed-lockfile approach (`plugins.lock`, per-archive SHA-512, verified offline, populated trust-on-first-use — the go.sum / package-lock.json model).

Before treating this as "superseded", one precision — the two PRs defend **different surfaces**, so it isn't a straight replacement:

| Threat | Defended by |
|---|---|
| Poisoning of the **extracted** `$id-$version/` tree — the code that actually executes — in a shared/writable cache | **this PR** (`PluginSecurity` directory guard) |
| A tampered/compromised **archive**, registry/mirror compromise, silent version drift | **#7394** (`plugins.lock`) |

The archive-hash lockfile in #7394 intentionally does **not** re-verify the extracted tree (re-hashing the retained zip can't detect edits to already-unpacked files), so #7394 does not cover what this PR covers.

That makes the choice a scoping decision rather than "lockfile replaces guard":

- **Land both** — the lockfile handles supply-chain/drift; this guard stays as cheap defense-in-depth for the extracted tree on shared caches.
- **Close this in favour of #7394 + the private-cache guidance** — if we're comfortable that keeping the plugin cache per-user-private (documented in #7394) is a sufficient answer to the extracted-tree surface, this guard becomes optional and its false-positive/complexity surface may not be worth carrying.

My lean is to close this in favour of #7394 and rely on the private-cache guidance, keeping the plugin-integrity story to a single, clean, offline-verifiable mechanism — but flagging the trade-off explicitly so it's a deliberate call.

pditommaso:
Sorry for the churn on this one 🙏 — thanks for bearing with the back-and-forth.

Stepping back, my main worry with the ownership/permission approach is **noise on shared filesystems**. The check flags a directory as untrusted when it's foreign-owned or world-writable — but the standard HPC/CI setup is an admin- or service-account-owned shared cache that every user reads. For every user who isn't that owner, each plugin dir looks "untrusted", so in the default `warn` mode we'd emit a warning per plugin, on every run — on a perfectly legitimate, read-only cache. That's exactly the population this is meant to help, and it'd train them to ignore the warnings.

So I took a different iteration in #7394: instead of checking *who owns* the plugin directory, it pins and verifies a **hash of the extracted plugin contents** (a committed `plugins.lock`, go.sum-style, opt-in). Because it checks the bytes that actually get executed, one mechanism covers both cache poisoning *and* supply-chain/registry tampering — and it uses no ownership heuristic, so it's silent on healthy shared caches and only speaks up on a genuine content mismatch.

I think #7394 supersedes this PR, so I'm inclined to close this in its favour — but happy to hear if you see a gap. Really appreciate the work that got us here; it's what surfaced the whole problem.

aknownuser:
Yes, the implementation in https://github.com/nextflow-io/nextflow/pull/7394 is complete and fixes the issue we have here and others. thanks for taking the time

PR #7311, opened 2026-07-08
ADR: nextflow doctor command

edmundmiller:
## Summary
- Add ADR for a diagnostic-only `nextflow doctor` command
- Cover bootstrap fallback plus JVM install checks
- Define stable text output, failure semantics, and offline behavior

## Verification
- Read adr/20260708-nextflow-doctor.md
- Verified branch diff against origin/master contains only the ADR


bentsherman:
I think this will work better as a skill, e.g. as troubleshooting notes for the `install-nextflow` skill

https://github.com/nextflow-io/agent-skills

PR #7312, opened 2026-07-08
Tolerate stale/transient resize errors on autoscale Azure Batch pools

jorgee:
## Problem

Task submission against Azure Batch fails with:

```
Azure Batch pool 'nextflow-ci' has resize errors and no agents are available
```

on pools that are actually healthy — autoscale enabled, parked at 0 nodes (no pending work), quota fine, VM SKU offered and unrestricted. The pool's `resizeErrors` holds a single **stale** `AllocationTimedout` from days earlier; autoscale has since re-evaluated to target=0 repeatedly without ever running a new resize, so the error was never overwritten.

## Root cause

`AzBatchService.checkPool()` threw whenever an existing pool carried *any* resize error while at 0 nodes — regardless of error code, age, or whether autoscale was enabled:

```groovy
else if ( pool.resizeErrors && pool.currentDedicatedNodes==0 && pool.currentLowPriorityNodes==0 ) {
    throw new IllegalStateException("Azure Batch pool '${pool.id}' has resize errors and no agents are available")
}
```

This is a self-reinforcing deadlock:

- Azure Batch persists the last resize error and only clears it on the next resize.
- `checkPool()` throws *before* the task is enqueued, so autoscale never sees `$PendingTasks > 0`, never resizes, and never clears the stale error.
- `getOrCreatePool()` throws before caching the pool, so every subsequent task re-runs the same failing check.

A pool that has already recovered (or would recover on the next allocation) is treated as permanently broken.

## Fix

Keep the existing gate, but distinguish recoverable from fatal via a new `isRecoverableResizeError(pool)`:

- **Warn and proceed** when *all* hold: autoscale enabled, `allocationState == STEADY`, and every resize error code is transient (`AllocationTimedout` / `AllocationFailed`, case-insensitive). On an autoscale pool, submitting the task is exactly what triggers the fresh resize that clears the stale error.
- **Throw** (unchanged) otherwise: fixed-size pools (target won't move), non-steady pools, or any non-transient error code (quota, invalid config).

Transient codes use an **allowlist**, so unknown codes still fail fast — no regression vs. current behavior.

Since `ResizeError` carries no timestamp, error *age* can't be measured directly; autoscale + error-code is the semantically correct discriminator.

## Tests

8 new Spock cases in `AzBatchServiceTest`: stale-transient-autoscale-steady → allowed; fixed-size → throws; non-transient code → throws; RESIZING → throws; nodes-present → allowed; plus a data-driven `isRecoverableResizeError` matrix. Full `nf-azure` suite green (105 tests, 0 failures).

### DIFF
--- a/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
+++ b/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
@@ -763,7 +763,7 @@ class AzBatchService implements Closeable {
      * errors can recover on the next resize, so submission should not be blocked.
      * Compared case-insensitively.
      */
-    private static final List<String> TRANSIENT_RESIZE_ERROR_CODES = ['allocationtimedout', 'allocationfailed']
+    private static final List<String> TRANSIENT_RESIZE_ERROR_CODES = ['AllocationTimedout', 'AllocationFailed']
 
     protected void checkPool(BatchPool pool, AzVmPoolSpec spec) {
         if( pool.state != BatchPoolState.ACTIVE ) {
@@ -791,9 +791,9 @@ class AzBatchService implements Closeable {
      * and every recorded resize error is a transient/retryable code.
      */
     protected boolean isRecoverableResizeError(BatchPool pool) {
-        return pool.isEnableAutoScale() \
-            && pool.allocationState == AllocationState.STEADY \
-            && pool.resizeErrors?.every { err -> err.code?.toLowerCase() in TRANSIENT_RESIZE_ERROR_CODES }
+        return pool.isEnableAutoScale() &&
+                pool.allocationState == AllocationState.STEADY &&
+                pool.resizeErrors?.every { err -> TRANSIENT_RESIZE_ERROR_CODES.any { it.equalsIgnoreCase(err.code) } }
     }
 
     protected void checkPoolId(String poolId) {
--- a/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
+++ b/plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
@@ -28,6 +28,7 @@ import dev.failsafe.function.CheckedPredicate
 
 import com.azure.compute.batch.BatchClient
 import com.azure.compute.batch.BatchClientBuilder
+import com.azure.compute.batch.models.AllocationState
 import com.azure.compute.batch.models.AutoUserScope
 import com.azure.compute.batch.models.AutoUserSpecification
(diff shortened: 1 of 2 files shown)


pditommaso:
## Review

**Verdict: 👍 Approve.** Tight, well-scoped fix for a real self-reinforcing deadlock. Sound reasoning, good test coverage, no correctness bugs. A few minor nits below, none blocking.

### Correctness
- **Logic is right.** The three-part discriminator (autoscale / STEADY / allowlisted code) correctly identifies pools where submitting the task is exactly what triggers the clearing resize. The `else if` → inner `if/else` preserves the `taskSlotsPerNode` check below (still a separate `if`). ✅
- **Null-safety leans the safe way.** `isEnableAutoScale()` null, `code` null, or `resizeErrors` null all collapse the `&&` chain to `false` → throws. Fail-closed is the correct default. ✅
- **Allowlist, not denylist.** Unknown codes still fail fast → no regression vs. current behavior. ✅
- **Behavioral trade-off worth naming (not a bug):** for a *genuine, persistent* capacity shortage that keeps re-emitting `AllocationFailed`/`AllocationTimedout`, the task now gets submitted and sits **pending** instead of failing fast. That's the intended trade (better than false-failing a healthy pool), but it converts a loud failure into a silent wait — the `log.warn` is the only signal. Fine as designed; just make sure ops know to watch that warn.

### Minor nits (optional)
- **Locale on `toLowerCase()`** in `isRecoverableResizeError`: use `toLowerCase(Locale.ROOT)`, or better `TRANSIENT_RESIZE_ERROR_CODES.any { it.equalsIgnoreCase(err.code) }`. The specific codes have no dotted-I so there's no actual bug today, but locale-independent matching is the correct habit for a code-comparison path.
- **Backslash line continuation** in `isRecoverableResizeError` works but is fragile (a trailing space after `\` silently breaks it). Groovy convention is usually operator-at-EOL or wrapping in `( … )`. Cosmetic.
- `TRANSIENT_RESIZE_ERROR_CODES` could be a `Set`, but at 2 elements it's irrelevant — leave it.

### Conventions
Matches the file: `@CompileStatic`, `protected` for testability, `log.warn` via `@Slf4j`, Javadoc on both additions. Constant naming/placement consistent with the module. ✅

### Tests
Strong. 8 cases cover happy path, fixed-size, non-transient code, RESIZING, nodes-present, plus a data-driven matrix including case-insensitivity and mixed transient+fatal codes. `Spy` calling real methods is the right tool here.
- Tiny gap: no assertion that `log.warn` fires on the recoverable path, and no case for a `null` error `code`. Neither is necessary — the matrix already pins the branch logic.

### Security
Nothing sensitive. The warn logs only pool id and error codes — no secrets. No new attack surface.

pditommaso (approved):

jorgee:
Addressed the review nits.

PR #7313, opened 2026-07-08
feat(nf-seqera): send pipeline-secret references to the Seqera Scheduler

pditommaso:
## What this does

Makes the `seqera` executor deliver Seqera Platform pipeline secrets (`secret 'FOO'` directive) to tasks running on the Seqera Scheduler. Companion to the scheduler-side work in seqeralabs/sched#653 (PR seqeralabs/sched#658).

## How it works

- `SeqeraTaskHandler.submit()` maps each `secret` directive name to the cloud secret-store reference `tower-<workflowId>/<name>` and sends it in the task's new `secrets` map. The executor sends **references only** — never the value; the scheduler resolves it at the edge (ECS `valueFrom` / VM agent).
- `SeqeraExecutor` now declares `isSecretNative()=true`, so `BashWrapperBuilder` skips its local-store `source` snippet — which pointed at a `$HOME/.nextflow/secrets/…` file on the head node that does not exist in the remote scheduler task container (the reason `secret` was silently broken here before).

## ⚠️ Temporary — must revert before merge

To compile against the new `Task.secrets` field, this PR:
- adds `mavenLocal()` to the root `build.gradle`, and
- bumps `nf-seqera`'s `sched-client` to `0.69.0`.

The new field currently exists only in Maven-local (from the sched branch). **Once a sched release publishes `sched-client` with the `secrets` field, drop the `mavenLocal()` line.** This is the cross-repo release-ordering dependency; the PR is a draft until then.

## Verification

- ✅ Full `:plugins:nf-seqera:test` passes (added: directive-name → reference mapping, empty-directive no-op, `isSecretNative`).
- ⏳ End-to-end run of a `secret 'FOO'` pipeline pending the sched release.

Ref seqeralabs/sched#653.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -88,6 +88,7 @@ allprojects {
     }
 
     repositories {
+        mavenLocal()
         mavenCentral()
         maven { url 'https://repo.eclipse.org/content/groups/releases' }
         maven { url = "https://s3-eu-west-1.amazonaws.com/maven.seqera.io/releases" }
--- a/docs/plugins/using-plugins.mdx
+++ b/docs/plugins/using-plugins.mdx
@@ -56,6 +56,12 @@ Plugin declarations in Nextflow configuration files are ignored when specifying 
 
 When Nextflow downloads plugins, it caches them in the directory specified by `NXF_PLUGINS_DIR` (`$HOME/.nextflow/plugins` by default).
 
+:::warning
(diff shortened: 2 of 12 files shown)


pditommaso:
Superseded by #7314. This branch was accidentally cut from `security-plugin-dir-trust-guard` instead of `master`, so it carried three unrelated plugin-dir-trust-guard commits. #7314 contains only the nf-seqera change, cherry-picked cleanly onto `master`. Closing this one.

PR #7316, opened 2026-07-09
Add managed identity support for Azure Batch container registries

adamrtalbot:
## Overview

Adds support for authenticating to private container registries (e.g. Azure Container Registry) using a **user-assigned managed identity**, as an alternative to `userName`/`password`.

```groovy
azure {
    registry {
        server = 'myregistry.azurecr.io'
        managedIdentityResourceId = '/subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<name>'
    }
}
```

## Details

- New `azure.registry.managedIdentityResourceId` config option. Value is the **full ARM resource ID** of the user-assigned identity (what the Batch SDK's `BatchNodeIdentityReference.setResourceId` expects), not a client-id GUID.
- Env fallback `AZURE_REGISTRY_MANAGED_RESOURCE_ID`; config value wins over env.
- When set, the managed identity takes **precedence** over `userName`/`password`.
- Wires the identity into the pool container config via `ContainerRegistryReference.setIdentityReference(...)`.

## Limitation

The managed identity must already be attached to the pool nodes. Nextflow's data-plane pool creation (`BatchPoolCreateContent`) has no identity setter, so this requires a pre-existing pool (or an identity attached out-of-band). Documented in the guide and reference.

## Tests

- `AzRegistryOptsTest`: config/env/precedence for the resource id, full `isConfigured()` matrix.
- `AzBatchServiceTest`: `poolVmConfig` cases for the username/password path and the managed-identity-wins path.

Generated by Claude Code

### DIFF
--- a/docs/azure.mdx
+++ b/docs/azure.mdx
@@ -550,6 +550,23 @@ Nextflow uses the following environment variables if the registry credentials ar
 - `AZURE_REGISTRY_PASSWORD`: the password for Azure Container Registry authentication
 :::
 
+Alternatively, you can authenticate to a private registry with a user-assigned [managed identity](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview) instead of a username and password:
+
+```groovy
+azure {
+    registry {
+        server = '<registry-server>'     // e.g., 'myregistry.azurecr.io'
+        managedIdentityResourceId = '<managed-identity-resource-id>' // e.g., '/subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<name>'
+    }
+}
+```
+
+The value is the full ARM resource ID of the managed identity, not its client ID. When set, it takes precedence over `userName` and `password`, and can also be provided via the `AZURE_REGISTRY_MANAGED_RESOURCE_ID` environment variable.
+
+:::note
+The managed identity must already be attached to the pool nodes. Nextflow cannot attach an identity when auto-creating a pool, so this option requires a pre-existing pool (or an identity attached out-of-band).
+:::
+
 Container images from public registries such as Docker Hub can be used without additional configuration, even when a private registry is specified in the `azure.registry` scope. The Docker runtime on Azure Batch VMs automatically uses the appropriate registry based on the container image specified for the task.
 
 ### VM images
--- a/docs/azure.mdx
+++ b/docs/azure.mdx
@@ -550,22 +550,8 @@ Nextflow uses the following environment variables if the registry credentials ar
 - `AZURE_REGISTRY_PASSWORD`: the password for Azure Container Registry authentication
 :::
 
-Alternatively, you can authenticate to a private registry with a user-assigned [managed identity](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview) instead of a username and password:
+To authenticate to a private registry with a user-assigned [managed identity](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview) instead of a username and password:
 
-```groovy
-azure {
-    registry {
-        server = '<registry-server>'     // e.g., 'myregistry.azurecr.io'
-        managedIdentityResourceId = '<managed-identity-resource-id>' // e.g., '/subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<name>'
-    }
-}
-```
-
-The value is the full ARM resource ID of the managed identity, not its client ID. When set, it takes precedence over `userName` and `password`, and can also be provided via the `AZURE_REGISTRY_MANAGED_RESOURCE_ID` environment variable.
-
-:::note
-The managed identity must already be attached to the pool nodes. Nextflow cannot attach an identity when auto-creating a pool, so this option requires a pre-existing pool (or an identity attached out-of-band).
-:::
 
 Container images from public registries such as Docker Hub can be used without additional configuration, even when a private registry is specified in the `azure.registry` scope. The Docker runtime on Azure Batch VMs automatically uses the appropriate registry based on the container image specified for the task.
 
--- a/docs/reference/config.mdx
+++ b/docs/reference/config.mdx
@@ -605,6 +605,14 @@ The client ID for an Azure [managed identity](https://learn.microsoft.com/en-us/
 
 When `true`, use the system-assigned [managed identity](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview) to authenticate Azure resources. Defaults to environment variable `AZURE_MANAGED_IDENTITY_SYSTEM`.
 
+##### `azure.registry.managedIdentityResourceId`
+
+The ARM resource ID of a user-assigned managed identity used to authenticate to a private container registry, e.g. `/subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<name>`. When set, it takes precedence over `azure.registry.userName` and `azure.registry.password`. Defaults to environment variable `AZURE_REGISTRY_MANAGED_RESOURCE_ID`.
+
+:::note
+The managed identity must already be attached to the pool nodes. Nextflow cannot attach an identity when auto-creating a pool, so this option requires a pre-existing pool (or an identity attached out-of-band).
+:::
+
 ##### `azure.registry.password`
 
 The password to connect to a private container registry.
--- a/docs/reference/config.mdx
+++ b/docs/reference/config.mdx
@@ -607,10 +607,10 @@ When `true`, use the system-assigned [managed identity](https://learn.microsoft.
 
 ##### `azure.registry.managedIdentityResourceId`
 
-The ARM resource ID of a user-assigned managed identity used to authenticate to a private container registry, e.g. `/subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<name>`. When set, it takes precedence over `azure.registry.userName` and `azure.registry.password`. Defaults to environment variable `AZURE_REGISTRY_MANAGED_RESOURCE_ID`.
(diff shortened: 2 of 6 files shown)


docs/reference/config.mdx
```diff
@@ -605,6 +605,14 @@ The client ID for an Azure [managed identity](https://learn.microsoft.com/en-us/
 
 When `true`, use the system-assigned [managed identity](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview) to authenticate Azure resources. Defaults to environment variable `AZURE_MANAGED_IDENTITY_SYSTEM`.
 
+##### `azure.registry.managedIdentityResourceId`
+
+The ARM resource ID of a user-assigned managed identity used to authenticate to a private container registry, e.g. `/subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<name>`. When set, it takes precedence over `azure.registry.userName` and `azure.registry.password`. Defaults to environment variable `AZURE_REGISTRY_MANAGED_RESOURCE_ID`.
```
```suggestion
The ARM resource ID of a user-assigned managed identity for authenticating to a private container registry. For example, `/subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<name>`. When set, it takes precedence over `azure.registry.userName` and `azure.registry.password`. Defaults to the `AZURE_REGISTRY_MANAGED_RESOURCE_ID` environment variable.
```

docs/reference/config.mdx
```diff
@@ -605,6 +605,14 @@ The client ID for an Azure [managed identity](https://learn.microsoft.com/en-us/
...
+
+:::note
+The managed identity must already be attached to the pool nodes. Nextflow cannot attach an identity when auto-creating a pool, so this option requires a pre-existing pool (or an identity attached out-of-band).
```
```suggestion
The managed identity must already be attached to the pool nodes. Nextflow cannot attach an identity when auto-creating a pool. This option requires a pre-existing pool with the identity attached.
```

docs/azure.mdx
```diff
@@ -550,6 +550,23 @@ Nextflow uses the following environment variables if the registry credentials ar
...
+        managedIdentityResourceId = '<managed-identity-resource-id>' // e.g., '/subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<name>'
+    }
+}
+```
+
+The value is the full ARM resource ID of the managed identity, not its client ID. When set, it takes precedence over `userName` and `password`, and can also be provided via the `AZURE_REGISTRY_MANAGED_RESOURCE_ID` environment variable.
+
+:::note
+The managed identity must already be attached to the pool nodes. Nextflow cannot attach an identity when auto-creating a pool, so this option requires a pre-existing pool (or an identity attached out-of-band).
+:::
```
christopher-hakkaart:
````suggestion
To authenticate to a private registry with a user-assigned [managed identity](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview) instead of a username and password:

```groovy
azure {
    registry {
        server = ''     // e.g., 'myregistry.azurecr.io'
        managedIdentityResourceId = '' // e.g., '/subscriptions//resourceGroups//providers/Microsoft.ManagedIdentity/userAssignedIdentities/'
    }
}
```

The value is the full ARM resource ID of the managed identity, not its client ID. When set, it takes precedence over `userName` and `password`. You can also provide it through the `AZURE_REGISTRY_MANAGED_RESOURCE_ID` environment variable.

:::note
The managed identity must already be attached to the pool nodes. Nextflow cannot attach an identity when auto-creating a pool. This option requires a pre-existing pool with the identity attached.
:::
````

**Review by christopher-hakkaart — COMMENTED** [review]
Style suggestions.

**Review by bentsherman — COMMENTED** [review]
Reviewed at `1cea23fcf975cb990e3c7bfadfe383a2103e3cff` against `master`.

The design is clean and the tests pass, but I have one blocking question about whether the feature can actually work, plus a docs file that got mangled when the suggestions were applied.

## Intent

The problem is real. Registry passwords in `nextflow.config` are a recurring complaint on Azure, ACR supports token exchange through managed identity, and the Batch SDK already models it via `ContainerRegistryReference.setIdentityReference(BatchNodeIdentityReference)`. We were not using that setter. Good thing to fix.

## Verification

Checked out the branch and ran `./gradlew :plugins:nf-azure:test --tests '*AzRegistryOptsTest*' --tests '*AzBatchServiceTest*'`. Build successful, `AzBatchServiceTest` 110 tests and `AzRegistryOptsTest` 3 tests, zero failures. Both new `poolVmConfig` cases show up in the JUnit XML.

I did not test against a live Batch account, so the finding below comes from reading the call graph rather than from a failed pull.

I did check the SDK claim in the description by decompiling `azure-compute-batch:1.0.0-beta.6`, and it holds up: `BatchPool` exposes `getIdentity()`, but `BatchPoolCreateParameters` has no identity setter at all. Nextflow genuinely cannot attach a managed identity to a pool it creates.

## The feature looks inoperative in the configuration it documents (blocking)

This is the one I would like an answer on before merge.

The identity reference reaches Azure through `poolVmConfig()`, and `poolVmConfig()` has exactly one caller: `createPool()` at `AzBatchService.groovy:937`. `getOrCreatePool()` (line 831) calls `getPool()` first, and when the pool already exists it calls `checkPool()`, which validates state, resize errors and slots per node. It sends nothing back to Azure. So the container registry configuration, identity or password, is only ever transmitted when Nextflow creates the pool.

Now put that next to the limitation note in the PR description, that the identity "must already be attached to the pool nodes" and this "requires a pre-existing pool". The two halves cancel out:

- Nextflow creates the pool. The registry config is sent, but no identity is attached to the nodes, so there is nothing for the reference to resolve to.
- The pool already exists with the identity attached. Nextflow never sends the registry config, so `setIdentityReference` is never called.

I cannot find a configuration where the new code path does useful work. The first case may also be an active regression for anyone who follows the docs: pool creation with an `identityReference` naming an identity the pool does not have is likely to fail the create call outright, where previously it would have fallen back to username and password. I have not confirmed that failure against live Azure, so treat it as a risk rather than a fact.

What would make this work is patching the container configuration of an existing pool, or moving pool creation to the management plane where `BatchPoolIdentity` is settable. Both are much bigger changes than this PR. Note that `azure.batch.poolIdentityClientId` sidesteps the problem entirely because Fusion reads it at task runtime, so nothing has to be sent at pool create time. That mechanism is not available here.

If I am wrong and it does work, one successful pull against a real ACR would settle it, and I would like to know how.

## The `docs/azure.mdx` addition is broken (must fix)

e4ccff2b applied the review suggestions, and the `docs/azure.mdx` one lost its body. The file now reads, at line 553:

> To authenticate to a private registry with a user-assigned managed identity instead of a username and password:

followed by two blank lines and then the unrelated paragraph about public registries. The groovy example, the note that the value is an ARM resource ID and not a client ID, the precedence rule, the `AZURE_REGISTRY_MANAGED_RESOURCE_ID` fallback and the admonition about pre-attached identities are all gone. The suggestion in the review thread has all of it, the applied commit does not. A fenced block inside a suggestion block is easy to mangle this way.

Net effect: the guide promises an example and delivers a blank line, and the new env var is documented nowhere outside the config reference.

## Smaller things

**Two managed identity options, two identifier formats.** `azure.batch.poolIdentityClientId` takes a client ID GUID, `azure.registry.managedIdentityResourceId` takes a full ARM resource ID. Both configure a user-assigned identity on the same pool, and people will reasonably expect to paste the same string into both. The difference is forced by the SDK since `BatchNodeIdentityReference` only has `resourceId`, so the code is right, but the docs should say so and cross-reference the two options.

**Env var name.** The config key is `managedIdentityResourceId` and the existing var is `AZURE_MANAGED_IDENTITY_USER`, so `AZURE_REGISTRY_MANAGED_IDENTITY_RESOURCE_ID` matches both conventions. `AZURE_REGISTRY_MANAGED_RESOURCE_ID` drops the word "identity". Nothing has shipped, so renaming is free now and a breaking change later.

**Missing `<AddedInVersion />`.** Seven other options in `docs/reference/config/azure.mdx` carry one, including `azure.batch.poolIdentityClientId`.

**Silent precedence.** Setting both the identity and username/password ignores the credentials. It is documented and tested, but a `log.warn` when both are present would save someone an afternoon.

**Default server.** Passing only `managedIdentityResourceId` with no `server` leaves the default `docker.io` and attaches an Azure identity reference to a Docker Hub entry. This pre-dates the PR since username/password behave the same way, so not a blocker.

**Tests.** Coverage of what the code does is good and the `isConfigured()` matrix is thorough. The gap is that nothing exercises the pre-existing pool path, which is exactly where the blocking finding lives. Minor nit: lines 479 and 498 spell out `com.azure.json.JsonProviders.createReader(...)` although `JsonProviders` is imported at line 35 and used bare at line 733.

Backward compatibility, style and security all look fine to me. The username/password path is untouched when the new option is absent, the only behavior change for existing users is the wording of the `IllegalArgumentException`, and the `log.debug` prints an ARM path rather than a credential.

## Summary

Must fix before merge:

1. Demonstrate the feature working against a live ACR and a real pool, or rework it so the identity reference reaches Azure on the pre-existing pool path.
2. Restore the lost content in `docs/azure.mdx`, the section currently ends mid-thought.

Worth doing while the option is unreleased:

3. Rename the env var to `AZURE_REGISTRY_MANAGED_IDENTITY_RESOURCE_ID`.
4. Add `<AddedInVersion />` and cross-reference `azure.batch.poolIdentityClientId`.
5. Warn when both auth methods are configured.

**Comment by bentsherman** [comment]
@adamrtalbot I don't have a local azure setup at the moment so I think claude just looked at the code. But if you can confirm that you tested the cases claude mentioned that will be good enough for me

File: docs/azure.mdx [hunk 4]
```diff
@@ -550,6 +550,9 @@ Nextflow uses the following environment variables if the registry credentials ar
 - `AZURE_REGISTRY_PASSWORD`: the password for Azure Container Registry authentication
 :::
 
+To authenticate to a private registry with a user-assigned [managed identity](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview) instead of a username and password:
+
+
```
christopher-hakkaart, line 555: @adamrtalbot Was there suppose to be an example here?

PR #7023, opened 2026-04-10
Add uv process directive for Python package management

evanroyrees:
## Summary

Add a new `uv` process directive for managing Python dependencies via the [uv](https://docs.astral.sh/uv/) package manager, following the existing `conda` and `spack` directive patterns.

Supports inline packages and requirements files:

```nextflow
process FOO {
    uv 'numpy pandas'
}

process BAR {
    uv "${moduleDir}/requirements.txt"
}
```

## Changes

- `UvConfig` / `UvCache` — config scope and environment creation/caching
- Process directive registration, task execution integration (`TaskBean`, `TaskRun`, `BashWrapperBuilder`)
- Task hashing, lineage tracking, job array support
- CLI options (`-with-uv` / `-without-uv`), config builder, launcher arg normalization
- Environment variables (`NXF_UV_ENABLED`, `NXF_UV_CACHEDIR`)
- Documentation (process directive, config scope, env vars, `docs/uv.md`)
- Tests for all new and modified components

Signed-off-by: Evan Rees <evanroyrees@gmail.com>

PR #7317, opened 2026-07-09
Package directive

evanroyrees:
# Generic `package` directive with pluggable package providers

## Summary

Continues the `package-directive` proof-of-concept from #6342 into a complete pluggable
package-provider system: a generic `package` process directive backed by per-manager
`nf-<provider>` plugins.

```nextflow
process hello {
    package 'numpy pandas', provider: 'uv'

    script:
    """
    python -c 'import numpy, pandas'
    """
}
```

**Lineage:** picks up the conda/pixi PoC from #6342 (thanks @edmundmiller) and follows the
maintainer guidance there: one generic directive with pluggable providers, aligned with Wave,
instead of per-manager directives (#7023). New providers address long-standing requests: nix and
guix (#1526) and the R ecosystem via pak and install2.r.

**Base:** rebased onto `master`. The 11 PoC commits from `nextflow-io:package-directive` are
squashed into the first commit of this branch (original authorship credited in the message), so
the PR now targets `master` directly.

## What's included

- **Core enablement**: `package` directive under the strict (v2) parser, the
  `nextflow.preview.package` feature flag (script header or config), a `packages` config scope
  (`packages.provider`, `packages.autoDetect`), and manifest auto-detection for processes defined
  in included module scripts.
- **Five new provider plugins**: `nf-uv`, `nf-nix`, `nf-guix`, `nf-pak`, `nf-install2r`, each
  implementing the `PackageProvider` SPI with its own environment cache, config scope, and tests.
- **conda/pixi providers extended and fixed**: per-process `options:` override, pixi manifest
  fixes (channel-qualified specs, builds inside the cache prefix instead of the module directory),
  new `pixi.channels` option, removal of the PoC's duplicated `CondaCache`/`CondaConfig` and of
  the never-read `<provider>.enabled` options.
- **Wave hand-off**: conda-based providers go to Wave; local-only providers are built locally.
- **Plugins use the `io.nextflow.nextflow-plugin` Gradle DSL** like every other plugin on
  `master` (generated manifest and `extensions.idx`, `nextflowVersion = 26.08.0-edge`,
  `VERSION` files, entries in `plugins-info.txt` regenerated via `releaseInfo`).
- **nf-test CI matrix**, **docs** (MDX pages, directive/config/feature-flag reference, sidebar),
  **ADR** `adr/20260707-package-provider-plugins.md` (marked `proposed`).

## Changes to core (outside `plugins/`) and why

Each of these touches shared code, so the reason and a way to reproduce the original problem on
`master` + the PoC branch are given.

### `nf-commons`: `Escape.shell(String)` / `Escape.shell(String[])`

New helper that wraps a value in POSIX single quotes. Needed because every provider cache builds a
`bash -c` command string and the existing `Escape.cli` does not quote all shell metacharacters.

**Reproduce (bug):** on the PoC branch, run a process with `package "cowsay>=6", provider: "uv"`.
The task succeeds, but `uv pip install ... cowsay>=6` was parsed by bash as `cowsay` with stdout
redirected to a file literally named `=6` in the launch directory, so the constraint was dropped.
After this change the command is `uv pip install ... 'cowsay>=6'`.

### `nextflow/conda/CondaCache.groovy`: quote plain package tokens

The `nf-conda` provider delegates to this class, so the `package` directive path needs it too;
the legacy `conda` directive shares the code. Each token is passed through `Escape.shell`, with
one level of pre-existing single or double quotes honored so specs written for the unquoted
command (`conda 'bwa "samtools>=1.0"'`) keep working. YAML/explicit-file branches are unchanged.

**Reproduce (bug):** `conda 'bwa>=0.7'` creates a file named `=0.7` in the launch directory and
installs an unconstrained `bwa`.

### `processor/TaskRun.groovy`, `TaskBean`, `TaskHasher`, `BashWrapperBuilder`: environment resolved at hash time

The PoC created the environment inside `BashWrapperBuilder`, i.e. during `handler.submit()` on
the single task-submission thread, and swallowed failures. Now `TaskRun.getPackageEnv()` resolves
it on the per-process thread at hash time, exactly like `getCondaEnv()`; the path travels in
`TaskBean.packageEnv` and the wrapper only emits the activation snippet.

**Reproduce (bug, resume):** run a module process whose `requirements.txt` is auto-detected, then
append a line to the file and run with `-resume`. Before: `cached=1`. After: the task re-executes.
Same for editing the directive itself (`cowsay>=6` to `cowsay>=5`).

**Reproduce (bug, swallowed failure):** `package "no-such-package-xyz", provider: "uv"` with a
script that only `echo`es. Before: `WARN: Failed to create package environment` and the task
succeeds without the env. After: the run fails at the task with the provider's error. An
unparsable directive value (e.g. `package 123`) now fails the same way.

**Reproduce (bug, container):** `package 'cowsay', provider: 'uv'` plus `container 'ubuntu:22.04'`
with `docker.enabled = true`. Before: a uv env was built on the head node and a host path was
`source`d inside the container. After: no local env is built and the wrapper has no activation.

**Reproduce (perf):** several processes with distinct `package` specs. Before: every env build
blocked the submission of all other tasks. After: builds run on each process's own thread.

### `processor/TaskRun.groovy`: auto-detection scope

Auto-detection now (1) skips processes that already declare `conda`, `spack` or `container`, and
(2) only applies to processes defined in included module scripts, not the entry script.

**Reproduce (bug):** put a `requirements.txt` in the project root next to `main.nf` and enable the
feature. Before: every process in `main.nf` got a uv environment, including one with a
`container` directive, and a process with `conda 'python'` next to an `environment.yml` had two
conda environments activated in its wrapper. After: none of those get an auto-detected env.

### `nf-wave`: hand-off receives the resolved spec

Wave now gets `task.getPackageSpec()` (so auto-detected manifests are built by Wave like explicit
directives), sends conda environment *files* as base64 content instead of a path, and skips pixi
manifests (`pixi.toml`/`pixi.lock`) which are not conda environment files. Pixi package lists still
go to Wave's conda builder. Local-only providers (uv, nix, guix, pak, install2r) still skip Wave.

**Reproduce (bug):** `package "/path/env.yml", provider: "conda"` with `wave.enabled = true`.
Before: Wave received the *path string* as a package name. After: the file content, base64, same
as the legacy `conda` directive path.

### `script/dsl/ProcessBuilder.groovy`: bare `pixi` directive removed

The PoC whitelisted a standalone `pixi 'pkg'` directive that no runtime code consumed and the
strict parser rejected as unrecognized. Pixi is used via `package 'pkg', provider: 'pixi'`.

### `Session.groovy`: memoized `getPackageManager()`

`TaskRun` and `BashWrapperBuilder` each constructed a new `PackageManager` per task, and the
constructor probes every provider plugin's binary (`conda --version`, `uv --version`,
`Rscript -e ...`). With six plugins loaded that was roughly a dozen subprocess spawns per task on
the submitter thread. One instance per session now.

### `packages/PackagesConfig.groovy` and plugin config scopes: `nextflow.config.spec` imports

`master` renamed `nextflow.config.schema.*` to `nextflow.config.spec.*`. The PoC-era config
classes still imported the old package, so the strict config validator did not discover them.

**Reproduce (bug):** with the PoC imports, any run with `packages.provider = 'conda'` in the
config logs `WARN: Unrecognized config option 'packages.provider'`.

### `nf-lang` parser

`package` is a Groovy keyword, so it cannot be declared on `DirectiveDsl`. The AST builder stops
rejecting it as an identifier and the resolver allows it only in the process-directive section;
it is still rejected as an input/output qualifier. v1 parser untouched.

## Testing

- Unit: `nf-commons` (`EscapeTest`), `nextflow` (`packages.*`, `TaskRunTest`, `TaskHasherTest`,
  `BashWrapperBuilderTest`, `CondaCacheTest`), `nf-lang` (`ScriptResolveTest`), `nf-wave`
  (`WaveClientTest`), and all seven provider plugins. Green locally.
- End to end with the dev build and `NXF_SYNTAX_PARSER=v2`: uv package list, header and config
  feature flag, version-constraint quoting, resume invalidation on spec and on manifest content,
  failure propagation, auto-detect scoping (entry script, module script, `container`, legacy
  `conda`), no local env build for a containerized task.
- Full local suite (`./gradlew test`, 669 suites) green except 19 nf-amazon tests that fail
  identically on pristine `master` on this machine (they read `~/.aws/config` directly).
- `.github/workflows/package-providers.yml` runs an nf-test job per provider.

## Known follow-ups (not in this PR)

- All caches treat "prefix directory exists" as "environment complete"; a JVM killed mid-install
  leaves a partial env that is reused. Same limitation as core `CondaCache`. A completion marker
  or build-then-rename would fix it across the board.
- Config-level `installOptions`/`createOptions` and `install2r.repos` are not part of the env
  hash (only per-process overrides are), matching `CondaCache` behavior.
- guix `guix.scm` is a package definition, not a manifest; should be dropped from auto-detect.
- uv treats any spec containing `/` as an existing venv directory, which rejects
  `git+https://...` and local path specs.
- CI: conda job likely needs `conda-remove-defaults: true`; the path filter should include
  `TaskRun.groovy`, `nextflow/conda/**`, `nf-lang/**`, `nf-wave/**`.
- Spack provider, conda-lock support, and full extraction of conda into `nf-conda`.

All commits are DCO signed-off.


edmundmiller:
#6342 for reference

PR #7320, opened 2026-07-09
Fix Fusion license token cache key not being interpolated

### LINKED ISSUE #7319: Fusion license token cache key is a non-interpolated single-quoted literal in TowerFusionToken

`TowerFusionToken` caches Fusion license tokens in a Guava cache (`tokenCache`) that is meant to be keyed by product and version. The key is built as:

```groovy
final key = '${product}-${version}'
```

Because the string is **single-quoted**, Groovy does not interpolate it — `key` is always the constant literal `${product}-${version}`, identical for every call regardless of the `product` / `version` arguments. The intent was clearly a double-quoted GString.

- **Expected:** distinct `(product, version)` pairs produce distinct cache keys/entries.
- **Actual:** all pairs collide on a single shared cache entry. The first token cached is returned for any subsequent product/version, and the cached error response (`tokenCache.put(key, ...)`) is shared the same way.

The outgoing request object is built with the correct `product`/`version`, so the HTTP request itself is fine — only the cache key is wrong.

### Impact

Currently **latent**. Within a single run's head-job JVM, `product = config.sku()` and `version = config.version()` are constant, so only one `(product, version)` pair is ever requested and the single cache slot always corresponds to the right token. The bug would surface if more than one product or version were requested in the same JVM (a second licensed SKU, or a version change mid-session): the later request would receive the earlier token from the cache rather than fetching its own.

### Location

https://github.com/nextflow-io/nextflow/blob/37984dd39ac1b50785b8cc0eda034891ab904baa/plugins/nf-seqera/src/main/io/seqera/tower/plugin/TowerFusionToken.groovy#L142

`key` is used only in `tokenCache.get` / `invalidate` / `put`, and nothing downstream re-interpolates the literal (Guava keys by `equals`/`hashCode`), so the constant key takes full effect.

### Suggested fix

Use a double-quoted GString so the value interpolates:

```groovy
final key = "${product}-${version}"
```

Found by code inspection. A unit test that invokes the token lookup with two different `(product, version)` pairs and asserts two distinct cache entries (or two `sendRequest` calls) would fail on the current code and pass after the one-character fix.

### Program output

N/A — this is a caching-correctness bug, not a runtime failure or crash.

### Environment

* Nextflow version: `master` (observed at commit `37984dd`)
* Java version: n/a (source-level bug)
* Operating system: n/a
* Bash version: n/a

### Additional context

This is the Groovy single-quoted `String` vs double-quoted `GString` distinction: single quotes yield a plain `java.lang.String` fixed at compile time, with no interpolation.

robsyme:
Closes #7319.

### Problem

The Fusion license-token cache key in `TowerFusionToken.getLicenseToken` was written with **single quotes**:

```groovy
final key = '${product}-${version}'
```

In Groovy a single-quoted literal is not interpolated, so `key` was always the constant string `${product}-${version}`, identical for every call. All `(product, version)` pairs collided on one Guava cache entry, so the first token cached was returned for any subsequent product/version, and the cached error response was shared the same way. The outgoing request was built correctly; only the cache key was wrong.

It is currently latent because `product`/`version` are constant within a run, but it defeats the intended per-product/version caching and would misfire if a second product or version were ever requested in the same session.

### Fix

Use a double-quoted GString so the value interpolates, typed as `String` so it still satisfies the `Cache<String, ...>` key type under `@CompileStatic`:

```groovy
final String key = "${product}-${version}"
```

### Test

Added `should cache license tokens per product and version` to `TowerFusionEnvTest`, using the existing WireMock harness to return a distinct token per product. It requests two products and asserts each receives its own token and both reach the endpoint. The test fails on the old key (the second product gets the first's cached token) and passes with the fix.

`./gradlew :plugins:nf-seqera:test --tests "*TowerFusionEnvTest"` → 14 passed, 0 failed.


pditommaso (approved):
Reviewed the merge-conflict resolution and the fix.

**Fix is correct:** `'${product}-${version}'` (single-quoted, literal) → `"${product}-${version}"` (double-quoted GString). Verified CompileStatic-safe — `final String key` triggers the implicit GString→String coercion at assignment, so the `Cache<String, ...>` key type is satisfied with no runtime CCE. Lookup/invalidate/store all use the same `key` (lines 147/155/164), so the single-line change propagates to all three cache operations.

**Test is a true regression guard:** two products with distinct WireMock stubs keyed on `$.product`; on the old literal key the second product collides on the shared cache entry (returns `token-a`, and the `product-b` stub sees 0 requests) → fails; on the fix each product keys independently → passes.

**Landed in the correct path** for current master (`plugins/nf-tower/`, after the nf-seqera consolidation revert); no stale duplicate in nf-seqera.

Minor non-blocking notes: cache key uses raw `version` while the request body normalizes `version ?: 'unknown'` — a null vs 'unknown' version for the same product would produce two keys for identical requests (harmless cache-miss edge). Optional: normalize the key to match, and add an inline comment noting the double-quote requirement to prevent regression.

LGTM 👍

PR #7321, opened 2026-07-10
Update nf-azure: node communication mode setting, allow Azure Compute Gallery and allow configuration of verification state

LennyBEL:
### **Overview**

**Node Communication Mode**
In our environment, Azure Batch is configured within a virtual network. However, Nextflow does not expose a configuration item for the Node Communication Mode. The Classic mode has been deprecated for a while, but Nextflow pools still default to it because they use an older API version. Without a configuration item, we cannot explictly tell it to use "simplified" mode.

This PR adds `azure.batch.pools.<name>.targetCommunicationMode` (named to match the Azure Batch API and SDKs). Accepted values are `simplified` and `classic`, defaulting to **`simplified`** since classic is deprecated and should no longer be used.

This only affects users running Nextflow on Azure in a networked environment.

Reference: https://learn.microsoft.com/en-us/azure/batch/simplified-compute-node-communication

**Azure Compute Gallery**
Last year we had a major outage when an update to the Microsoft Marketplace Ubuntu 22.04 HPC image flipped its verification state to `unverified`. Because Nextflow only targets `verified` images, new pools could no longer be created. Our pipeline tooling did not support Ubuntu 24.04 yet, so we had to scramble to restore service.

In order to avoid such an event again, Microsoft suggested that we switch to Azure Compute Gallery images. Within Azure Compute Gallery, we have full control over the images. However, Nextflow did not support the usage of them yet.

Two new options (virtualMachineImageId and osType) have been introduced to support usage of the Azure Compute Gallery. 

**Image verification state**
For non-production and testing purposes we occasionally need to use unverified Marketplace images, but Nextflow always forced `verified`.

This PR adds `azure.batch.pools.<name>.verification` with values `verified` (default), `unverified`, or `any` - where `any` ignores the verification state entirely. This setting is ignored when `virtualMachineImageId` is set.

### Expected Impact

We have been running these changes as a custom `nf-azure` fork (based on the upstream plugin) in production for a while, so we are confident they behave as intended.

- **Networked Azure users:** pools created by Nextflow will now use `simplified` communication mode by default instead of `classic`. The effect is generally positive, but users should be aware of the change.
- **Everyone else:** the new options are no-ops unless explicitly set. Verification still defaults to `verified` (unchanged behavior), and the Compute Gallery options are ignored when not configured.

### Tests

- **`AzPoolOptsTest`**
  - Defaults: `verification=verified`, `osType=linux`, `targetCommunicationMode=simplified`, `virtualMachineImageId=null`.
  - Compute Gallery image configuration (`virtualMachineImageId` + `sku`).
  - Parsing/validation of `verification` (`verified`/`unverified`/`any`, case-insensitive), with an error on invalid values.
  - Parsing/validation of `targetCommunicationMode` and `osType`, with errors on invalid values.
  - Cache-key (`funnel`) sensitivity: the pool hash changes when the image config or communication mode changes.
- **`AzBatchServiceTest`**

Docs updated: `docs/azure.mdx` and `docs/reference/config.mdx`.

### DIFF
--- a/docs/azure.mdx
+++ b/docs/azure.mdx
@@ -597,6 +597,43 @@ azure {
 }
 ```
 
+**Custom images from an Azure Compute Gallery**
+
+You can provision pool nodes from a custom VM image published in an [Azure Compute Gallery](https://learn.microsoft.com/en-us/azure/virtual-machines/azure-compute-gallery) by setting `virtualMachineImageId` to the image version resource ID. When set, `publisher` and `offer` are ignored, and `sku` must be set to the Batch node agent SKU id that matches the image operating system.
+
+```groovy
+azure {
+    batch {
+        pools {
+            <pool-name> {
+                virtualMachineImageId = '/subscriptions/<subscription>/resourceGroups/<group>/providers/Microsoft.Compute/galleries/<gallery>/images/<definition>/versions/<version>'
+                sku = 'batch.node.ubuntu 22.04'
+            }
+        }
+    }
+}
+```
+
+:::note
+Custom images require Microsoft Entra authentication (service principal or managed identity). The Batch account identity must have read access to the gallery image, and the pool typically requires `virtualNetwork` to be set.
+:::
+
+**Image verification**
+
+By default, Nextflow only selects images that Azure Batch has formally verified. Set `verification` to `unverified` to also allow images that Azure Batch lists but has not verified, or to `any` to accept both. This setting is ignored when `virtualMachineImageId` is set.
+
+```groovy
+azure {
+    batch {
+        pools {
+            <pool-name> {
+                verification = 'unverified'  // 'verified' (default), 'unverified' or 'any'
+            }
+        }
+    }
+}
+```
+
 ### Advanced features
 
 **Virtual networks**
@@ -618,6 +655,14 @@ Nextflow can submit tasks to existing pools that are already attached to a virtu
 :::warning
 Virtual networks require Microsoft Entra authentication (service principal or managed identity).
 :::
+
+**Node communication mode**
+
+Pools can be configured to use [simplified compute node communication](https://learn.microsoft.com/en-us/azure/batch/simplified-compute-node-communication), which changes how the Batch service communicates with the compute nodes and reduces the outbound network access required by the nodes. The default is `simplified`; set it to `classic` to restore the legacy behavior.
+
+```groovy
+azure.batch.pools.<pool-name>.targetCommunicationMode = 'simplified'  // 'simplified' (default) or 'classic'
+```
 
 **Start tasks**
 
--- a/docs/reference/config.mdx
+++ b/docs/reference/config.mdx
@@ -537,6 +537,10 @@ The mount options for mounting the file shares (default`-o vers=3.0,dir_mode=077
 
 The offer type of the virtual machine type used by the pool identified with `<name>` (default`centos-container`).
 
+##### `azure.batch.pools.<name>.osType`
+
+The OS type of the pool nodes. Can be either `linux` or `windows` (default`linux`).
+
 ##### `azure.batch.pools.<name>.privileged`
 
 Enable the task to run with elevated access. Ignored if `runAs` is set (default`false`).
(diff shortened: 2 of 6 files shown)


adamrtalbot (commented):
Can we split into two PRs? I think it will be quicker to merge the networking and image PRs separately.

docs/azure.mdx
```diff
@@ -619,6 +656,14 @@ Nextflow can submit tasks to existing pools that are already attached to a virtu
...
 :::
 
+**Node communication mode**
+
+Pools can be configured to use [simplified compute node communication](https://learn.microsoft.com/en-us/azure/batch/simplified-compute-node-communication), which changes how the Batch service communicates with the compute nodes and reduces the outbound network access required by the nodes. The default is `simplified`; set it to `classic` to restore the legacy behavior.
+
+```groovy
+azure.batch.pools.<pool-name>.targetCommunicationMode = 'simplified'  // 'simplified' (default) or 'classic'
+```
+
```
adamrtalbot, line 666: Currently, nextflow doesn't set `targetCommunicationBehaviour` which leaves is treated as `Default` by Azure Batch. This meant it was `Classic` until 30th September 2024.

After that, Azure didn't guarantee `Simplified` (classic Azure 🙄) but in our hands it ended up as `Simplified` even on vnets.  If you see something different, that's disappointing but not surprising.

Rather than defaulting to `simplified`, I would allow `null`, which preserves old behaviour, then add `classic` and `simplified` as explicit. 

Otherwise we might end up in the same situation in 6 months time when Azure deprecate `simplified` and require us to use the next networking technology. Having `null` = `default` becomes "trust the Azure Batch API".

docs/azure.mdx
```diff
@@ -597,6 +597,43 @@ azure {
...
+You can provision pool nodes from a custom VM image published in an [Azure Compute Gallery](https://learn.microsoft.com/en-us/azure/virtual-machines/azure-compute-gallery) by setting `virtualMachineImageId` to the image version resource ID. When set, `publisher` and `offer` are ignored, and `sku` must be set to the Batch node agent SKU id that matches the image operating system.
+
+```groovy
+azure {
+    batch {
+        pools {
+            <pool-name> {
+                virtualMachineImageId = '/subscriptions/<subscription>/resourceGroups/<group>/providers/Microsoft.Compute/galleries/<gallery>/images/<definition>/versions/<version>'
+                sku = 'batch.node.ubuntu 22.04'
```
adamrtalbot, line 610:
```suggestion
                sku = 'batch.node.ubuntu 24.04'
```

docs/azure.mdx
```diff
@@ -597,6 +597,43 @@ azure {
...
+            }
+        }
+    }
+}
+```
+
+:::note
+Custom images require Microsoft Entra authentication (service principal or managed identity). The Batch account identity must have read access to the gallery image, and the pool typically requires `virtualNetwork` to be set.
+:::
```
adamrtalbot, line 619: We should make this a stronger message because users will skip over it:

```suggestion
:::warning
Custom images require Microsoft Entra authentication (service principal or managed identity). The **Azure Batch account identity** must have read access to the gallery image, and the pool typically requires `virtualNetwork` to be set.
:::
```

docs/reference/config.mdx
```diff
@@ -537,6 +537,10 @@ The mount options for mounting the file shares (default`-o vers=3.0,dir_mode=077
 
 The offer type of the virtual machine type used by the pool identified with `<name>` (default`centos-container`).
 
+##### `azure.batch.pools.<name>.osType`
+
+The OS type of the pool nodes. Can be either `linux` or `windows` (default`linux`).
+
```
adamrtalbot, line 543: 😨 

This one line may prevent this PR from being merged 😅

adamrtalbot: to be fair, this isn't just "I don't like Windoze", it's increasing the support area _dramatically_ and Nextflow is very explicit about only supporting POSIX systems.

adamrtalbot (commented):
Thank you for the contribution!

Definitely split into multiple PRs, this is 3 separate features:

- networking (a quick fix and needed!)
- custom images (a necessary feature, straightforward)
- Windows support 😆 

And I don't want _one_ of those features to block the others.

docs/azure.mdx
LennyBEL: First: my apologies for the late response. I had two days off this week :).

Then... the reason we had to introduce this as an option is because despite Microsoft's deprecation of the 'Classic' mode, it still defaults to 'Classic'. MS support clarified this was because Nextflow uses an outdated API version to create the pools. This API version is fixed inside the Batch SDK, so there's that...

MS Support response: _"PG has determined that the issue is that the pool is being created with an old API version: https://blablabla.westeurope.batch.azure.com/pools?api-version=2024-02-01.19.0 - This version doesn't support the "simplified by default", upgrading to 2025-06-01 or newer should fix the default for scenarios where you cannot explicitly set the communication mode to simplified."_

So, the reason I decided to force it on 'simplified' by default is simply because you can't trust the Azure Batch API to take the correct option. That said, I also follow in your reasoning and for me the most important is that the option is there at all. If you feel that "null" is the safest default, then I'll follow :).

docs/reference/config.mdx
LennyBEL: Very fair remark. Windows support isn't needed at all really, it was just a logical by-product of adding Compute Gallery support.

I'll remove this option ;).

LennyBEL:
Individual comments addressed. I'll split this into multiple PRs now.

LennyBEL:
Split into #7337 and #7338. Windows support has been dropped from the changes. Feedback has been incorporated into the respective PRs.

Closed.

docs/azure.mdx
adamrtalbot: > MS support clarified this was because Nextflow uses an outdated API version to create the pools.

From some digging, MS are wrong, which doesn't surprise me, Azure Support probably just typed it into ChatGPT 🙄 

We use the Azure Batch Java SDK `com.azure:azure-compute-batch:1.0.0-beta.3` which uses get latest which _does_ resolve to 2024-07-01.20.0 as of right now. The [API changelog](https://learn.microsoft.com/en-us/rest/api/batchservice/batch-service-rest-api-versioning#version-2024-02-01190) never claims a change between simplified and classic networking, but the Batch documentation _clearly_ claims the default changes on 30th September 2024. So the claim from support directly contradicts their own documentation.

None of this matter of course, we should (a) add the field and (b) update the SDK to beta6 (there's still no GA release!)

PR #7100, opened 2026-05-05
Address Netty CVEs by bumping AWS SDK and excluding micronaut-core-bom from test classpath

jorgee:
## Summary

Address two HIGH-severity Dependabot alerts on transitively-pulled Netty:

- **GHSA-pwqr-wmgm-9rr8** — `netty-codec-http` HTTP Request Smuggling (alerts [#155](https://github.com/nextflow-io/nextflow/security/dependabot/155), [#156](https://github.com/nextflow-io/nextflow/security/dependabot/156))
- **GHSA-w9fj-cfpg-grvv** — `netty-codec-http2` CONTINUATION Frame Flood DoS (alerts [#157](https://github.com/nextflow-io/nextflow/security/dependabot/157), [#158](https://github.com/nextflow-io/nextflow/security/dependabot/158))

Rather than pinning Netty versions explicitly, this PR fixes the *root causes* of how the vulnerable Netty was reaching Nextflow:

1. **Bump AWS SDK** in `nf-amazon` (2.33.2 → 2.44.3), `nf-codecommit` (2.31.64 → 2.44.3). This SDK version already pull patched Netty 4.1.133.Final natively in their POMs.
The Azure SDK in `nf-azure` (`azure-storage-blob` 12.33.2 → 12.33.3, `azure-identity` 1.18.2 → 1.18.3). The newer SDKs already pull patched Netty 4.1.132.Final natively in their POMs.
2. **Exclude `io.micronaut:micronaut-core-bom` from `testRuntimeClasspath`** in the three plugins. That BOM (transitively reached through `testImplementation project(':nextflow')` → `npr-api` → `micronaut-serde-api`) was declaring a constraint forcing Netty to 4.2.5.Final on the test classpath only — overriding whatever the AWS / Azure SDKs ship and re-introducing the vulnerable line. The exclusion lets the test classpath inherit the same Netty as runtime.
3. **`nf-tower`**: bumped WireMock 3.13.1 → 3.13.2 (latest patch; ships Jetty 11.0.26).

## Why this is better than version pinning

- **No magic version numbers in the build files.** The Netty version simply tracks whatever the AWS / Azure SDKs ship. Future SDK bumps automatically pick up new Netty patches without any manual intervention.
- **Consistent runtime/test classpaths.** Without the BOM exclusion, the test classpath would pin Netty to a different version than runtime — masking real ABI issues. With the exclusion, both classpaths resolve to the same Netty.
[...]

PR #7330, opened 2026-07-13
Add ast-grep Groovy lint foundations

edmundmiller:
## What

Adds a small [ast-grep](https://ast-grep.github.io/) setup for **structural (AST-based) linting of Groovy sources**, under `.config/astgrep/`. This is the foundation PR: it wires up the tooling and ships a **single, advisory (warning-severity) starter rule** so the mechanism can be reviewed in isolation. Further rules are intended to stack as follow-up PRs.

## Why lead with `no-wildcard-import`

It maps directly to an **already-documented house style**: the [Nextflow developer docs](https://docs.seqera.io/nextflow/developer/) mandate single-class imports (the recommended IntelliJ setup raises the "class count to use import with `*`" thresholds to 99, effectively disabling star imports). So this rule encodes an existing written convention rather than introducing a new opinion.

As a secondary motivation, wildcard imports invite name collisions as the codebase evolves — a recent *illustration* (in Java, not Groovy) was #7100, where the AWS SDK's and the JDK's `AccessDeniedException` clashed because both packages were star-imported. Note this Groovy-only linter would **not** have caught that `.java` case; it's cited only to show the class of problem wildcards create.

Scoped to plain `import foo.bar.*` — **2 findings** today:
- `plugins/nf-console/src/main/nextflow/ui/console/Nextflow.groovy:19` — `import javax.swing.*`
- `plugins/nf-tower/src/main/io/seqera/tower/plugin/auth/AuthCommandImpl.groovy:24` — `import java.awt.*`

(Static wildcard imports — `import static ...*` — are idiomatic for enums/constants and Spock tests, ~106 occurrences; intentionally **not** flagged here. A separate rule for those could be a follow-up if desired.)

## How it works

- `.config/astgrep/build.sh` fetches [dekobon/tree-sitter-groovy](https://github.com/dekobon/tree-sitter-groovy) at a **pinned immutable commit** (`8c70dc6`, v0.2.2) into a gitignored `.grammar-build/` (not vendored), asserts the resolved SHA matches the pin, and compiles `groovy.dylib` locally. The library is native code built on each machine, so it works on macOS (Mach-O) and Linux (ELF) — nothing binary is committed.
- `.config/astgrep/sgconfig.yml` registers `groovy` as an ast-grep custom language.
- `.config/astgrep/rules/` holds one YAML lint rule per file.

## Usage

```bash
cd .config/astgrep && ./build.sh          # once
ast-grep scan -c .config/astgrep/sgconfig.yml
```

Requires `ast-grep` (>= 0.44) and a C compiler.

## Notes

- **Advisory only** — `severity: warning`; not wired into CI, nothing blocks.
- Intended follow-up rules (separate PRs): `no-print-stack-trace`, `no-system-out-println` (both tie to the `@Slf4j` logger convention), `no-thread-sleep`, and optionally static-wildcard imports.

### DIFF
--- a/.config/astgrep/.gitignore
+++ b/.config/astgrep/.gitignore
@@ -1,0 +1,3 @@
+# ast-grep build artifacts (rebuilt locally by build.sh)
+.grammar-build/
+groovy.dylib
--- a/.config/astgrep/README.md
+++ b/.config/astgrep/README.md
@@ -16,6 +16,17 @@ ast-grep scan -c .config/astgrep/sgconfig.yml # scan the repo
 Requires [`ast-grep`](https://ast-grep.github.io/guide/quick-start.html) (>= 0.44) and a
 C compiler (`gcc`/`clang`).
 
+## Testing
+
+Each rule has a test file in `rule-tests/` with `valid` (should not match) and
+`invalid` (should match) cases. Run them with:
+
+```bash
+cd .config/astgrep && ast-grep test
+```
+
+After changing a rule, refresh the captured output with `ast-grep test -U`.
+
 ## How it works
 
 - **`build.sh`** fetches [dekobon/tree-sitter-groovy](https://github.com/dekobon/tree-sitter-groovy)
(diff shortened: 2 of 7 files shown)


bentsherman (commented):
I can see the vision. Still, detecting wildcard imports isn't exactly life-changing for us. I wonder if you could find something more valuable to detect? Not sure what though

Issue #5634, opened 2025-01-03
NF does not perform proxy authentification and cannot access github


kpetrov:
## Bug report 

Doesn't authenticate at the proxy server, no internet access

### Expected behavior and actual behavior

We have a computing cluster with internet access provided via proxy.  Most programs take the credentials from the environmental variables and work just fine. Nextflow doesn't try to authenticate.

### Steps to reproduce the problem

install nextflow, install graalvm, run
)->nextflow run nf-core/demo    -profile maestro    --input samplesheet.csv 

### Program output 

 N E X T F L O W   ~  version 24.10.3

Pulling nf-core/demo ...
WARN: Cannot read project manifest -- Cause: Unable to tunnel through proxy. Proxy returns "HTTP/1.1 407 Proxy Authentication Required"
Cannot find `nf-core/demo` -- Make sure exists a GitHub repository at this address `https://github.com/nf-core/demo`

### Environment 

* Nextflow version: 24.10.3
* Java version: graalvm 22.3.1
* Operating system: Linux RHEL 8.6
* Bash version: 4.4.20

http_proxy=http://user:mysecretkeyhere@squid:3128
### Additional context

Proxy is required for security and audit.

bentsherman:
Can you share your Nextflow log? You should see these two lines if `http_proxy` is set:

```
Setting http proxy: ProxyConfig[...]
Setting http proxy authenticator
```

kpetrov:
[nextflow.log](https://github.com/user-attachments/files/18333059/nextflow.log)


thanks for a quick reply, here is the log with some debug info

menzel:
I have the same issue, password/user and proxy server url are shown fine the log, when Nextflow tries to use it, it fails with  "Authentication required".

Is there a workaround the download the file manually and place it somewhere for Nextflow to find?

bentsherman:
If it's the git repo for the pipeline, you can just clone the repo via `git` or some other way that works, then run the pipeline locally

braffes:
After a lot of compilation, I finally found the issue.

Firstly, I had to add the following line to the launch.sh script:

`JVM_ARGS+=" -Djdk.http.auth.tunneling.disabledSchemes=''"`

This is due to a behavior change in Java 8u111, as described in the [release notes](https://www.oracle.com/java/technologies/javase/8u111-relnotes.html):
"Disable Basic authentication for HTTPS tunneling

[...]

PR #7331, opened 2026-07-13
Support authenticated forward HTTP/HTTPS proxies across the client stack

### LINKED ISSUE #7305: Support authenticated forward HTTP proxies across Nextflow's HTTP stack

pditommaso:
## Summary

Behind an authenticating corporate proxy, Nextflow's HTTP paths (plugin downloads, registry calls, SCM, Tower, Wave, data-links) failed with an immediate `407`: proxy routing and credentials from `HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY` / `NO_PROXY` were never propagated to the `java.net.http.HttpClient` clients Nextflow builds.

This PR threads the parsed proxy configuration into every HTTP client construction site, adds the missing environment-variable handling, and fixes the JDK's HTTPS-tunnelling credential quirk.

## Changes

**Proxy plumbing (`HxClient` stack)**
- Move `ProxyConfig` from `nextflow` to `nf-commons` so plugins can reach it.
- Add a static registry + `HxProxyConfig` view (`ProxyConfig.proxyConfig()`) resolved from the launch environment.
- `Launcher.setProxy` registers the parsed proxy plus `NO_PROXY` hosts, so `HttpPluginRepository`, `RegistryClientFactory`, `RepositoryProvider`, `SraExplorer`, `TowerClient`, `AuthCommandImpl` and `WaveClient` (both the main and token-refresh clients) honour the same routing and credentials.
- Clients are configured on `HxClient.Builder` directly (not via `.httpClient(...)`), so `withProxyConfig(...)` isn't a silent no-op and the internal token-refresh clients inherit the proxy too. The now-unused `ProxyConfig.configure(HttpClient.Builder)` helper is removed.
- Bump `lib-httpx` to `2.4.0` (carries `withProxyConfig` / `HxProxyConfig`).

**Remaining raw-`HttpClient` sites in nf-tower**
- `DataLinksResourceHandler`: converted from a plain JDK `HttpClient` to `HxClient` (no bearer token, so the pre-signed-URL fetch still does **not** carry the Seqera `Authorization` header) with `withProxyConfig`.
- `TowerXAuth`: **kept** on a plain JDK `HttpClient` — its token-refresh flow owns a `CookieManager` and reads the `JWT` / `JWT_REFRESH_TOKEN` cookies back out of it, and `HxClient.Builder` has no `cookieHandler()` hook. The proxy selector + proxy authenticator are wired explicitly instead (rationale documented inline).

**Environment-variable handling (`Launcher` / `ProxyConfig`)**
- `ALL_PROXY` / `all_proxy` is used as a fallback when no scheme-specific proxy variable is set (scheme-specific still takes precedence).
- `ProxyConfig.parse` percent-decodes the userinfo `user:pass` so credentials with special characters (`@`, `:`, …) work; a literal `+` is preserved (userinfo is not form-encoded).

**HTTPS CONNECT tunnelling (the core #7305 fix)**
- The JDK default `jdk.http.auth.tunneling.disabledSchemes=Basic` strips proxy credentials from the HTTPS `CONNECT` request, so HTTPS targets behind an authenticating proxy fail with `407` even when the authenticator is wired. `Launcher` now clears the property when the proxy carries credentials, without ever overriding a value the user set (e.g. via `NXF_OPTS`).

## Testing

Each change lands as a separate commit with unit tests (RED→GREEN): proxy resolution + `withProxyConfig` wiring, a WireMock integration test proving a request routes through an authenticating proxy, `ALL_PROXY` fallback, credential URL-decoding, `disabledSchemes` clearing (including the no-credentials and user-set guardrails), and per-site tests for `DataLinksResourceHandler` and `TowerXAuth`.

## Not included (follow-ups)

- Docs for `ALL_PROXY` and the `NXF_OPTS='-Djdk.http.auth.tunneling.disabledSchemes='` fallback in `env-vars.mdx`.
- An end-to-end HTTPS `CONNECT`-tunnelling integration test — the current WireMock test deliberately covers only plain-HTTP, so the `disabledSchemes` fix is unit-covered but not exercised end-to-end.

## Related

- Fixes #7305
- Replaces #7306 (consolidates its scope; see the discussion there)


bentsherman:
Does this PR also resolve #5634 ?

ewels (approved):
I'm away from a keyboard so can't test right now, but skimming the code it seems sensible from what I can understand. My buddy Claudio is happy too.

bentsherman (approved):
Compared with #5634 and it looks like this PR fixes part of it. So I will just merge this one and submit a follow-up PR with any remaining fixes needed for the older issue

PR #7332, opened 2026-07-13
Fix concurrent modification with job arrays

### LINKED ISSUE #6108: java.util.ConcurrentModificationException when using array jobs in Slurm
Hi,
Sometimes I am getting **java.util.ConcurrentModificationException** errors by Nextflow when using array on jobs submitted to Slurm. It hapens quite randomly, and usually is successful upon re-run.
```

May-21 05:20:24.920 [Actor Thread 121] DEBUG nextflow.processor.TaskProcessor - Handling unexpected condition for
  task: name=Analyze:JoinPQarms (F145125 for loss using 1 CPUs and 64 MB memory); work-dir=/analysis/work/79/f38ad66152498f507edd28f58467fb
  error [java.util.ConcurrentModificationException]: java.util.ConcurrentModificationException
May-21 05:20:24.919 [Actor Thread 122] ERROR nextflow.processor.TaskProcessor - Error executing process > 'Analyze:Aneuploidies:Wisecondor:SplitLossesGains (F146225 using 1 CPUs and 64 MB memory)'

Caused by:
  java.util.ConcurrentModificationException


java.util.ConcurrentModificationException: null
	at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1095)
	at java.base/java.util.ArrayList$Itr.next(ArrayList.java:1049)
	at nextflow.executor.BashWrapperBuilder.createContainerBuilder(BashWrapperBuilder.groovy:706)
	at nextflow.executor.BashWrapperBuilder.makeBinding(BashWrapperBuilder.groovy:286)
	at nextflow.executor.BashWrapperBuilder.buildNew0(BashWrapperBuilder.groovy:410)
	at nextflow.executor.BashWrapperBuilder.buildNew0(BashWrapperBuilder.groovy:172)
	at nextflow.executor.BashWrapperBuilder.build(BashWrapperBuilder.groovy:425)
	at nextflow.executor.GridTaskHandler.prepareLauncher(GridTaskHandler.groovy:107)
	at nextflow.processor.TaskArrayCollector.createTaskArray(TaskArrayCollector.groovy:141)
	at nextflow.processor.TaskArrayCollector.collect(TaskArrayCollector.groovy:103)
	at nextflow.processor.TaskProcessor.submitTask(TaskProcessor.groovy:2339)
	at nextflow.processor.TaskProcessor.checkCachedOrLaunchTask(TaskProcessor.groovy:839)
	at nextflow.processor.TaskProcessor.invokeTask(TaskProcessor.groovy:655)
	at nextflow.processor.InvokeTaskAdapter.call(InvokeTaskAdapter.groovy:52)
	at groovyx.gpars.dataflow.operator.DataflowOperatorActor.startTask(DataflowOperatorActor.java:120)
	at groovyx.gpars.dataflow.operator.ForkingDataflowOperatorActor.access$001(ForkingDataflowOperatorActor.java:35)
	at groovyx.gpars.dataflow.operator.ForkingDataflowOperatorActor$1.run(ForkingDataflowOperatorActor.java:58)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

[https://stackoverflow.com/questions/14948971/why-does-arraylist-not-throw-concurrentmodificationexception-when-modified-from](https://stackoverflow.com/questions/14948971/why-does-arraylist-not-throw-concurrentmodificationexception-when-modified-from)

**_nextflow version 24.10.4.5934
openjdk 21.0.7 2025-04-15
OpenJDK Runtime Environment (build 21.0.7+6-Ubuntu-0ubuntu124.04)
OpenJDK 64-Bit Server VM (build 21.0.7+6-Ubuntu-0ubuntu124.04, mixed mode, sharing)_**

bentsherman:
Fix #6108

`TaskArrayRun.getContainerConfig()` appended the executor array-index variable (e.g. `SLURM_ARRAY_TASK_ID`) to the container env whitelist by mutating the list in place. Because `ContainerHelper.parseEnvWhitelist()` returns the session config list by reference, that list is shared across every `ContainerConfig` rebuilt from the session config, so one actor thread mutating it while another iterated it in `BashWrapperBuilder`/`FusionHelper` threw a `ConcurrentModificationException`

This PR removes the shared container config mutation. The array index variable is already carried on `TaskBean`, so instead add it to the container environment where the whitelist is applied:
- `BashWrapperBuilder.createContainerBuilder()` for the standard container path
- `FusionHelper.runWithContainer()` for the Fusion path

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -742,6 +742,11 @@ class BashWrapperBuilder {
             builder.addEnv(var)
         }
 
+        // for job arrays, propagate the array index variable (e.g. SLURM_ARRAY_TASK_ID)
+        // into the container environment
+        if( arrayIndexName )
+            builder.addEnv(arrayIndexName)
+
         // when secret are not managed by the execution platform natively
         // the secret names are added to the container env var white list
         if( !isSecretNative() && secretNames )  {
--- a/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/BashWrapperBuilder.groovy
@@ -504,7 +504,9 @@ class BashWrapperBuilder {
         final lines = new StringBuilder()
         lines << '### ---\n'
         lines << "### name: '${bean.name}'\n"
-        if( bean.arrayIndexName ) {
+        // the array metadata block only applies to the array dispatcher task (which carries the
+        // child work-dirs); array children only carry the index name to expose it in the container
+        if( bean.arrayIndexName && bean.arrayWorkDirs ) {
             lines << '### array:\n'
             lines << "###   index-name: ${bean.arrayIndexName}\n"
             lines << "###   index-start: ${bean.arrayIndexStart}\n"
--- a/modules/nextflow/src/main/groovy/nextflow/executor/GridTaskHandler.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/executor/GridTaskHandler.groovy
@@ -209,9 +209,15 @@ class GridTaskHandler extends TaskHandler implements FusionAwareTask {
     }
 
     protected BashWrapperBuilder createTaskWrapper(TaskRun task) {
-        return fusionEnabled()
-            ? fusionLauncher()
-            : executor.createBashWrapperBuilder(task)
+        if( fusionEnabled() )
+            return fusionLauncher()
+        final builder = executor.createBashWrapperBuilder(task)
+        // for a containerised array child, the container is built from the child task bean
+        // (which does not carry the array index variable); expose the scheduler array-index
+        // variable (e.g. SLURM_ARRAY_TASK_ID) inside the task container
+        if( isArrayChild && executor instanceof TaskArrayExecutor )
+            builder.arrayIndexName = ((TaskArrayExecutor) executor).getArrayIndexName()
+        return builder
     }
 
     protected String stdinLauncherScript() {
--- a/modules/nextflow/src/main/groovy/nextflow/fusion/FusionHelper.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/fusion/FusionHelper.groovy
@@ -63,6 +63,11 @@ class FusionHelper {
         for( String env : containerConfig.getEnvWhitelist())
             containerBuilder.addEnv(env)
 
+        // for job arrays, propagate the array index variable (e.g. SLURM_ARRAY_TASK_ID)
+        // into the container environment
+        if( launcher.arrayIndexName )
+            containerBuilder.addEnv(launcher.arrayIndexName)
+
         // patch the cmd wrapping the last item in quotes
         final patchCmd = new ArrayList(runCmd)
         patchCmd[-1] = "'${patchCmd[-1]}'"
--- a/modules/nextflow/src/main/groovy/nextflow/processor/TaskArrayRun.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/processor/TaskArrayRun.groovy
@@ -18,8 +18,6 @@ package nextflow.processor
 
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
-import nextflow.container.ContainerConfig
-import nextflow.executor.TaskArrayExecutor
 
 /**
  * Models a task array, which submits a collection of independent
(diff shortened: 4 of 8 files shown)


jorgee:
Reviewed this before approving and found a gap that needs to be closed first: the `BashWrapperBuilder` half of the fix never fires for **non-Fusion grid arrays** (SLURM/SGE/PBS/PBS Pro/LSF/TCS combined with Docker/Singularity/Podman/etc.), so the array-index variable is silently dropped from the task container on those platforms.

**Why**

`arrayIndexName` is populated on `TaskBean` only when `task instanceof TaskArrayRun` (`TaskBean.groovy:180-183`). But:

- the `TaskArrayRun` has `isContainerEnabled() == false`, so its own container builder is never invoked;
- for a non-Fusion grid array the actual `docker run`/`singularity exec` is emitted into each **child's** `.command.run`, built from a plain child `TaskRun` via `TaskArrayCollector.createTaskArray()` → `handler.prepareLauncher()` → `GridTaskHandler.createTaskWrapper(childTask)` → `new TaskBean(childTask)`, where `arrayIndexName` is `null`.

`withArrayChild(true)` only flags the *handler* (`TaskHandler.isArrayChild`), never the bean, so `if( arrayIndexName )` in `createContainerBuilder()` is skipped and the container launches without `-e SLURM_ARRAY_TASK_ID`.

The Fusion path works because Fusion is container-native and wraps the container at *array-submit* time via `FusionHelper.runWithContainer()` using the `TaskArrayRun` bean (which does carry `arrayIndexName`). AWS/Google Batch are unaffected — they inject the array index natively and never read `getEnvWhitelist()`.

Note that the added `BashWrapperBuilderTest` passes because it stubs `getArrayIndexName()` on the builder, which bypasses exactly this broken wiring. I'll push a fix + a handler-level test that exercises the real child path.

jorgee:
Pushed 21aef4fd to close the gap. Three parts:

**1. Fix — propagate the index name to the child launcher** (`GridTaskHandler.createTaskWrapper()`)

The handler is the one place that knows both that it's building an array child (`isArrayChild`) and which `TaskArrayExecutor` is in play, so it now injects the executor's array index name onto the (non-Fusion) child builder:

```groovy
protected BashWrapperBuilder createTaskWrapper(TaskRun task) {
    if( fusionEnabled() )
        return fusionLauncher()
    final builder = executor.createBashWrapperBuilder(task)
    // for a containerised array child, the container is built from the child task bean
    // (which does not carry the array index variable); expose the scheduler array-index
    // variable (e.g. SLURM_ARRAY_TASK_ID) inside the task container
    if( isArrayChild && executor instanceof TaskArrayExecutor )
        builder.arrayIndexName = ((TaskArrayExecutor) executor).getArrayIndexName()
    return builder
}
```

This reuses the whitelist branch already added in `BashWrapperBuilder.createContainerBuilder()` — no new mechanism. Scoped to non-Fusion because the Fusion path already handles it at array-submit time via `FusionHelper.runWithContainer()` with the `TaskArrayRun` bean. All grid `TaskArrayExecutor`s (SLURM/SGE/PBS/PBS Pro/LSF/TCS) route through this method; AWS/Google Batch inject the index natively.

**2. Guard — the `### array:` metadata block** (`BashWrapperBuilder.getTaskMetadata()`)

Now that array children carry `arrayIndexName`, the metadata block is gated on `arrayWorkDirs` too:

```groovy
if( bean.arrayIndexName && bean.arrayWorkDirs ) {
```

Only the dispatcher task (`TaskArrayRun`) carries `arrayWorkDirs`; without this guard a child would render a bogus `### array:` block and NPE on `for( Path it : bean.arrayWorkDirs )`.

**3. Test — exercise the real child path** (`GridTaskHandlerTest`)

Added `should propagate the array index variable to a containerised array child launcher`: it builds the child launcher from a plain `TaskRun` (asserting `childBuilder.arrayIndexName == null` as the precondition — the actual broken state), runs it through the real `createTaskWrapper()`, and asserts the handler injects `SLURM_ARRAY_TASK_ID`. A companion test confirms a non-array task is untouched and never queries `getArrayIndexName()`. I verified the positive test fails against the pre-fix `createTaskWrapper()`.

I left the existing `BashWrapperBuilderTest` case as-is — it's a valid unit test of the whitelist/`addEnv` consumption side; the new handler test covers the propagation side that was missing.

jorgee (approved):
I have found an issue in the non-fusion path and pushed a fix and extra tests to cover this case. Ready to merge from my side.

bentsherman:
Thanks for catching this edge case. However I need to review this PR as a whole again. I'm concerned about polluting the code too much with all of this array logic. Need to see if we can do it in a more isolated manner

PR #7335, opened 2026-07-14
Add seqera.executor.providerConfig option

pditommaso:
## Summary

Adds a `seqera.executor.providerConfig` configuration option that forwards backend-specific provider settings to the Seqera scheduler when creating a run, and fixes two config-validation gaps that otherwise flag valid sub-keys as unrecognized.

`CreateRunRequest.providerConfig` already exists in the sched API and `SchedClient.createRun()` already forwards it — but the Seqera executor never populated it and there was no config surface for it.

## Changes

**providerConfig option**
- `ExecutorOpts`: new `Map<String,String> providerConfig` config option (parsed like the existing `taskEnvironment`).
- `SeqeraExecutor.createRun()`: sets `.providerConfig(seqeraConfig.providerConfig)` on the `CreateRunRequest`.
- Docs: document `seqera.executor.providerConfig` in the config reference.

**Config validation fixes** (both surfaced while testing the option above)
- `ConfigValidator`: recognize parameterized `Map<K,V>` options. `isMapOption` only matched raw `Map.class`, so a `Map<String,String>` option (which reflects as a `ParameterizedType`) was treated as a scope and its sub-keys warned as *Unrecognized config option* — e.g. `seqera.executor.providerConfig.lazyPull` and the pre-existing `seqera.executor.taskEnvironment.*`. Now the raw type is resolved.
- `WaveConfig`: declare `wave.build.compression` as a map option so `wave.build.compression.mode` (and `level`/`force`) are recognized (the field previously had no `@ConfigOption`, so `SpecNode` dropped it).

## Usage

```groovy
// nextflow.config
seqera.executor.providerConfig = [ lazyPull: 'true' ]

wave.build.compression.mode = 'estargz'
```

For example, `lazyPull` enables experimental lazy eStargz container pulling on the `aws-vm` backend (see seqeralabs/sched#669), which pairs with Wave-built eStargz images.

## Tests

- `ExecutorOptsTest` (populated + null `providerConfig`), `SeqeraExecutorTest` (`createRun` forwards `providerConfig`).
- `ConfigValidatorTest` (`isMapType` recognizes raw + parameterized map types).

### DIFF
--- a/plugins/nf-wave/src/main/io/seqera/wave/plugin/config/WaveConfig.groovy
+++ b/plugins/nf-wave/src/main/io/seqera/wave/plugin/config/WaveConfig.groovy
@@ -286,6 +286,11 @@ class BuildOpts implements ConfigScope {
 
     final CondaOpts conda
 
+    @ConfigOption(types=[Map])
+    @Description("""
+        Build image compression settings, e.g. `[mode: 'estargz']`. Supported keys: `mode`
+        (`gzip`, `estargz`, `zstd`), `level` and `force`.
+    """)
     final BuildCompression compression
 
     @ConfigOption
--- a/plugins/nf-wave/src/main/io/seqera/wave/plugin/config/WaveConfig.groovy
+++ b/plugins/nf-wave/src/main/io/seqera/wave/plugin/config/WaveConfig.groovy
@@ -286,7 +286,6 @@ class BuildOpts implements ConfigScope {
 
     final CondaOpts conda
 
-    @ConfigOption(types=[Map])
     @Description("""
         Build image compression settings, e.g. `[mode: 'estargz']`. Supported keys: `mode`
         (`gzip`, `estargz`, `zstd`), `level` and `force`.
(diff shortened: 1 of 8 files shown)


plugins/nf-wave/src/main/io/seqera/wave/plugin/config/WaveConfig.groovy
```diff
@@ -286,6 +286,11 @@ class BuildOpts implements ConfigScope {
 
     final CondaOpts conda
 
+    @ConfigOption(types=[Map])
```
bentsherman: `@ConfigOption` should not be used for nested config scopes. `BuildCompression` is already a `ConfigScope` so it will be handled correctly

bentsherman:
```suggestion
```

bentsherman:
PR looks good, aside from my one comment. That line can be deleted

plugins/nf-wave/src/main/io/seqera/wave/plugin/config/WaveConfig.groovy
pditommaso: Great

bentsherman (approved):

PR #7337, opened 2026-07-15
Update nf-azure: Add Azure Batch node communication mode option

LennyBEL:
Split from #7321

### Feature: Node Communication Mode
In our environment, Azure Batch is configured within a virtual network. However, Nextflow does not expose a configuration item for the Node Communication Mode. The Classic mode has been deprecated for a while, but Nextflow pools still default to it because they use an older API version. Without a configuration item, we cannot explictly tell it to use "simplified" mode.

This PR adds azure.batch.pools.<name>.targetCommunicationMode (named to match the Azure Batch API and SDKs). Accepted values are simplified and classic; defaulting to null to let the Azure Batch API choose.

Reference: https://learn.microsoft.com/en-us/azure/batch/simplified-compute-node-communication

### Expected Impact

None, as the default behavior does not change at all.

### Tests

- **`AzPoolOptsTest`**
  - Parsing/validation of `targetCommunicationMode`, with errors on invalid values.
  - Cache-key (`funnel`) sensitivity: the pool hash changes when the communication mode changes.
- **`AzBatchServiceTest`**

**Docs updated: `docs/azure.mdx` and `docs/reference/config.mdx`.**


plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
```diff
@@ -175,6 +182,18 @@ class AzPoolOpts implements CacheFunnel, ConfigScope {
...
+
+    protected static BatchNodeCommunicationMode parseCommunicationMode(value) {
+        if( value == null )
+            return null
+        if( value instanceof BatchNodeCommunicationMode )
+            return value
+        final str = value.toString().toLowerCase()
+        if( str == 'simplified' ) return BatchNodeCommunicationMode.SIMPLIFIED
+        if( str == 'classic' ) return BatchNodeCommunicationMode.CLASSIC
+        throw new IllegalArgumentException("Invalid azure.batch.pools.<name>.targetCommunicationMode value: '$value' - expected 'simplified' or 'classic'")
```
adamrtalbot, line 196: We could include an explicit "default" here:

I've also removed the using BatchNodeCommunication mode because I think it makes the method confusing.

```suggestion
    protected static BatchNodeCommunicationMode parseCommunicationMode(value) {
        final str = value.toString().toLowerCase()
        if( str == 'simplified' ) return BatchNodeCommunicationMode.SIMPLIFIED
        if( str == 'classic' ) return BatchNodeCommunicationMode.CLASSIC
        if( str == 'default' ) return BatchNodeCommunicationMode.DEFAULT
        throw new IllegalArgumentException("Invalid azure.batch.pools.<name>.targetCommunicationMode value: '$value' - expected 'simplified', 'classic' or 'default'")
    }
```

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
```diff
@@ -175,6 +182,18 @@ class AzPoolOpts implements CacheFunnel, ConfigScope {
         this.password = opts.password
         this.virtualNetwork = opts.virtualNetwork
         this.lowPriority = opts.lowPriority as boolean
+        this.targetCommunicationMode = parseCommunicationMode(opts.targetCommunicationMode)
```
adamrtalbot, line 185: The Nextflow team have the final say on style, but it may be more disciplined to only use parseCommunicationMode when targetCommunicationMode is populated:

```suggestion
        this.targetCommunicationMode = opts.targetCommunicationMode != null ?parseCommunicationMode(opts.targetCommunicationMode) : null
```

Style choice, it's been a while since I've contributed so I want one of the core developers to look at it!

adamrtalbot (approved):
Nice feature, I agree it's worth having.

Suggestion: expose `default` as an explicit value alongside `simplified`/`classic`. The SDK defines `BatchNodeCommunicationMode.DEFAULT`, so the enum set would be complete and users get an explicit way to request the Batch service default rather than only reaching it by omitting the option.

Note: `default` is likely functionally equivalent to leaving the option unset but I'm not certain.

I want one of the core Nextflow devs to check it for style.

docs/reference/config.mdx
```diff
@@ -577,6 +577,10 @@ Enable the `startTask` to run with elevated access (default`false`).
 
 The `startTask` that is executed as the node joins the Azure Batch node pool.
 
+##### `azure.batch.pools.<name>.targetCommunicationMode`
+
+The node communication mode used by the pool. Can be either `classic` or `simplified`. When not set, the Azure Batch service selects the mode. In `simplified` mode the Batch service initiates communication with the compute nodes, reducing the outbound network access required by the nodes.
```
christopher-hakkaart, line 582:
```suggestion
The node communication mode used by the pool. One of `classic` or `simplified`. When not set, the Azure Batch service selects the mode. In `simplified` mode, the Batch service initiates communication with the compute nodes, which reduces the outbound network access the nodes require.

```

docs/azure.mdx
```diff
@@ -619,6 +619,16 @@ Nextflow can submit tasks to existing pools that are already attached to a virtu
...
 
+**Node communication mode**
+
+Pools can be configured to use [simplified compute node communication](https://learn.microsoft.com/en-us/azure/batch/simplified-compute-node-communication), which changes how the Batch service communicates with the compute nodes and reduces the outbound network access required by the nodes.
+
+```groovy
+azure.batch.pools.<pool-name>.targetCommunicationMode = 'simplified'  // 'simplified' or 'classic'
+```
+
+When this option is not set, the Azure Batch service selects the communication mode.
```
christopher-hakkaart, line 630:
````suggestion
To use [simplified compute node communication](https://learn.microsoft.com/en-us/azure/batch/simplified-compute-node-communication), set `targetCommunicationMode` to `simplified`. This changes how the Batch service communicates with the compute nodes and reduces the outbound network access the nodes require.

```groovy
azure.batch.pools.<pool-name>.targetCommunicationMode = 'simplified'  // 'simplified' or 'classic'
```

When this option is not set, the Azure Batch service selects the communication mode.
````

christopher-hakkaart (commented):
Editorial suggestions to make the text more task orientated.

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
LennyBEL: > Note: `default` is likely functionally equivalent to leaving the option unset but I'm not certain.

I'm about 99% certain that that's the case. However, I can't find any clear documentation on that...

However, given that this option was never explicitly set up until now, we can safely conclude that leaving it empty means it takes the 'Default' option.

LennyBEL:
This PR can be closed. Turns out the communication mode property has been removed all together since beta.6 of the SDK (which Nextflow upgraded to literally 2 weeks ago). This property was available up until beta.5.

In beta.6, looks like they updated to an API version that doesn't expose the property anymore and defaults to simplified.

https://raw.githubusercontent.com/Azure/azure-sdk-for-java/main/sdk/batch/azure-compute-batch/CHANGELOG.md

image

PR #7338, opened 2026-07-15
Add Azure Compute Gallery image and verification options

LennyBEL:
Split from #7321

### **Overview**

**Azure Compute Gallery**
Last year we had a major outage when an update to the Microsoft Marketplace Ubuntu 22.04 HPC image flipped its verification state to `unverified`. Because Nextflow only targets `verified` images, new pools could no longer be created. Our pipeline tooling did not support Ubuntu 24.04 yet, so we had to scramble to restore service.

In order to avoid such an event again, Microsoft suggested that we switch to Azure Compute Gallery images. Within Azure Compute Gallery, we have full control over the images. However, Nextflow did not support the usage of them yet.

A new option `virtualMachineImageId` has been introduced to support usage of the Azure Compute Gallery. 

**Image verification state**
For non-production and testing purposes we occasionally need to use unverified Marketplace images, but Nextflow always forced `verified`.

This PR adds `azure.batch.pools.<name>.verification` with values `verified` (default), `unverified`, or `any` - where `any` ignores the verification state entirely. This setting is ignored when `virtualMachineImageId` is set.

**Note: Windows support was left out compared to the original PR. OsType is no longer configurable and remains fixed on Linux.**

### Expected Impact
None, as the default behavior does not change at all.

We have been running these changes as a custom `nf-azure` fork (based on the upstream plugin) in production for a while, so we are confident they behave as intended.

### Tests

- **`AzPoolOptsTest`**
  - Compute Gallery image configuration (`virtualMachineImageId` + `sku`).
  - Parsing/validation of `verification` (`verified`/`unverified`/`any`, case-insensitive), with an error on invalid values.
  - Cache-key (`funnel`) sensitivity: the pool hash changes when the image config changes.
- **`AzBatchServiceTest`**

Docs updated: `docs/azure.mdx` and `docs/reference/config.mdx`.


plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
```diff
@@ -144,8 +144,21 @@ class AzPoolOpts implements CacheFunnel, ConfigScope {
...
+    @Description("""
+        The resource ID of a custom VM image from an Azure Compute Gallery to use for the pool nodes
+        (e.g. `/subscriptions/<id>/resourceGroups/<group>/providers/Microsoft.Compute/galleries/<gallery>/images/<definition>/versions/<version>`).
+        When set, `publisher` and `offer` are ignored, and `sku` must be set to the Batch node agent SKU id that matches the image OS (e.g. `batch.node.ubuntu 24.04`).
+    """)
+    final String virtualMachineImageId
+
+    @ConfigOption
+    @Description("""
+        The image verification type to match when resolving the VM image from the Batch supported-images list. Can be `verified`, `unverified`, or `any` (default: `verified`). Ignored when `virtualMachineImageId` is set.
```
adamrtalbot: I feel like verified should be included in unverified? It seems weird to reject verified images when you're allowing unverified images 🤔 this would remove the need for "any".

adamrtalbot: In fact, `unverified` could be a boolean:

```nextflow
unverified = false // Only verified images
unverified = true // allow unverified images, here be monsters
```

docs/azure.mdx
```diff
@@ -597,6 +597,43 @@ azure {
...
+azure {
+    batch {
+        pools {
+            <pool-name> {
+                verification = 'unverified'  // 'verified' (default), 'unverified' or 'any'
+            }
+        }
+    }
+}
+```
```
adamrtalbot, line 637: This might be bordering on too much information for this doc and should go to the advanced config options.

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
```diff
@@ -175,6 +188,20 @@ class AzPoolOpts implements CacheFunnel, ConfigScope {
         this.password = opts.password
         this.virtualNetwork = opts.virtualNetwork
         this.lowPriority = opts.lowPriority as boolean
+        this.virtualMachineImageId = opts.virtualMachineImageId ?: null
+        this.verification = parseVerification(opts.verification)
+    }
```
adamrtalbot: A bit of code tidying, let's put them with their logical groupings.

```suggestion
AzPoolOpts(Map opts) {
this.runAs = opts.runAs ?: ''
this.privileged = opts.privileged ?: false
this.publisher = opts.publisher ?: DEFAULT_PUBLISHER
this.offer = opts.offer ?: DEFAULT_OFFER
this.virtualMachineImageId = opts.virtualMachineImageId ?: null
this.verification = parseVerification(opts.verification)
this.sku = opts.sku ?: DEFAULT_SKU
this.vmType = opts.vmType ?: DEFAULT_VM_TYPE
this.fileShareRootPath = opts.fileShareRootPath ?: buildFileShareRootPath()
this.vmCount = opts.vmCount as Integer ?: 1
this.autoScale = opts.autoScale as boolean
this.scaleFormula = opts.scaleFormula
this.schedulePolicy = opts.schedulePolicy
this.scaleInterval = opts.scaleInterval as Duration ?: DEFAULT_SCALE_INTERVAL
this.maxVmCount = opts.maxVmCount as Integer ?: vmCount *3
this.startTask = new AzStartTaskOpts( opts.startTask ? opts.startTask as Map : Map.of() )
this.registry = opts.registry
this.userName = opts.userName
this.password = opts.password
this.virtualNetwork = opts.virtualNetwork
this.lowPriority = opts.lowPriority as boolean
    }
```

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
```diff
@@ -175,6 +188,20 @@ class AzPoolOpts implements CacheFunnel, ConfigScope {
         this.password = opts.password
         this.virtualNetwork = opts.virtualNetwork
         this.lowPriority = opts.lowPriority as boolean
+        this.virtualMachineImageId = opts.virtualMachineImageId ?: null
+        this.verification = parseVerification(opts.verification)
```
adamrtalbot: Yeah, I think I like unverified = true/false more:

```suggestion
        this.virtualMachineImageId = opts.virtualMachineImageId ?: null
        this.unverified = opts.verification as Boolean
```

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
adamrtalbot, line 179: There are enough fields here that it may require it's own config scope.

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -704,16 +706,22 @@ class AzBatchService implements Closeable {
                 continue
             if( it.osType != opts.osType )
                 continue
-            if( it.verificationType != opts.verification )
+            if( opts.verification != null && it.verificationType != opts.verification )
```
adamrtalbot: If unverified was just a boolean, we might use falsey here to catch `null` and false together:

Or something, not sure, just spitballing 😆 

```suggestion
            if( opts.unverified && it.verificationType != opts.verification )
```

adamrtalbot (approved):
This would also help when Azure deprecates a Batch image without a replacement image lined up. It would improve the supply chain security of our Azure users. So worth it overall.

adamrtalbot (commented):
(Claude review here, but raised some good points)

Focused, sensible change — the `verification == null` = "any" handling in `getImage`, the gallery/marketplace branch in `poolVmConfig`, and the `verified` default all look correct. A few points below, most-important first.

**Cache-key note (whole-PR):** the pool funnel now hashes `verification`, which was a field but was previously *not* hashed. On upgrade every existing auto-pool id changes (hence the test hash change `42f3635f…`→`7483c5b1…`), so existing pools won't be reused and will idle out. Worth calling out in the description — "Expected Impact: None" isn't quite accurate for users relying on pool reuse across the upgrade.

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -704,16 +706,22 @@ class AzBatchService implements Closeable {
...
         }
 
-        log.debug "[AZURE BATCH] No VM image matching sku=$opts.sku; publisher=$opts.publisher; offer=$opts.offer; OS type=$opts.osType; verification type=$opts.verification - supported images: $available"
-        throw new IllegalStateException("Cannot find a matching VM image with publisher=$opts.publisher; offer=$opts.offer; OS type=$opts.osType; verification type=$opts.verification")
+        log.debug "[AZURE BATCH] No VM image matching sku=$opts.sku; publisher=$opts.publisher; offer=$opts.offer; OS type=$opts.osType; verification type=${opts.verification ?: 'any'} - supported images: $available"
+        throw new IllegalStateException("Cannot find a matching VM image with publisher=$opts.publisher; offer=$opts.offer; OS type=$opts.osType; verification type=${opts.verification ?: 'any'}")
+    }
+
+    protected ImageReference customImageReference(AzPoolOpts opts) {
+        if( !opts.sku )
```
adamrtalbot: **Medium — this guard is unreachable.** In the `AzPoolOpts` constructor `sku` is `opts.sku ?: DEFAULT_SKU` (`"batch.node.ubuntu 24.04"`), so `opts.sku` is never null/empty and `if( !opts.sku )` never fires.

The docs promise that for a gallery image "`sku` must be set to the Batch node agent SKU id that matches the image OS", but in practice a user who omits `sku` silently gets `batch.node.ubuntu 24.04`. If their custom image is a different OS the pool fails to provision at Azure with a much more confusing error than the intended one.

Either drop the dead guard, or detect "sku not explicitly provided" (check before the `?: DEFAULT_SKU` fallback in the constructor) and enforce it for the gallery path.

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
```diff
@@ -51,6 +51,8 @@ import com.azure.compute.batch.models.BatchContainerConfiguration
 import com.azure.compute.batch.models.ContainerRegistryReference
 import com.azure.compute.batch.models.ContainerType
 import com.azure.compute.batch.models.ElevationLevel
+import com.azure.compute.batch.models.ImageReference
+import com.azure.compute.batch.models.MetadataItem
```
adamrtalbot: Unused import — only `BatchMetadataItem` (below) is used; `MetadataItem` isn't referenced. Drop this line.

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
```diff
@@ -195,6 +222,8 @@ class AzPoolOpts implements CacheFunnel, ConfigScope {
         hasher.putUnencodedChars(schedulePolicy ?: '')
         hasher.putUnencodedChars(virtualNetwork ?: '')
         hasher.putBoolean(lowPriority)
+        hasher.putUnencodedChars(virtualMachineImageId ?: '')
+        hasher.putUnencodedChars(verification?.toString() ?: 'any')
```
adamrtalbot: Minor: `verification` is hashed even when `virtualMachineImageId` is set, where it's documented/coded as ignored — two gallery configs differing only in a no-op `verification` get different pool ids. Cosmetic.

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
```diff
@@ -175,6 +188,20 @@ class AzPoolOpts implements CacheFunnel, ConfigScope {
...
+        if( value == null )
+            return ImageVerificationType.VERIFIED
+        if( value instanceof ImageVerificationType )
+            return value
+        final str = value.toString().toLowerCase()
+        if( str == 'verified' ) return ImageVerificationType.VERIFIED
+        if( str == 'unverified' ) return ImageVerificationType.UNVERIFIED
+        if( str == 'any' ) return null
+        throw new IllegalArgumentException("Invalid azure.batch.pools.<name>.verification value: '$value' - expected 'verified', 'unverified' or 'any'")
     }
```
adamrtalbot, line 194: Optional: `null` here encodes a third state ("any") on an enum-typed field — a one-line comment would help future readers, since the type alone doesn't hint that null is meaningful.

docs/reference/config.mdx
```diff
@@ -577,6 +577,14 @@ Enable the `startTask` to run with elevated access (default`false`).
 
 The `startTask` that is executed as the node joins the Azure Batch node pool.
 
+##### `azure.batch.pools.<name>.verification`
+
+The image verification type to match when resolving the VM image from the Batch supported-images list. Can be `verified`, `unverified`, or `any` (default`verified`). Ignored when `virtualMachineImageId` is set.
```
```suggestion
The image verification type to match when resolving the VM image from the Batch supported-images list. One of `verified`, `unverified`, or `any` (default `verified`). Ignored when `virtualMachineImageId` is set.
```

docs/reference/config.mdx
```diff
@@ -577,6 +577,14 @@ Enable the `startTask` to run with elevated access (default`false`).
...
+
+##### `azure.batch.pools.<name>.virtualMachineImageId`
+
+The resource ID of a custom VM image from an Azure Compute Gallery to use for the pool nodes (e.g. `/subscriptions/<id>/resourceGroups/<group>/providers/Microsoft.Compute/galleries/<gallery>/images/<definition>/versions/<version>`). When set, `publisher` and `offer` are ignored, and `sku` must be set to the Batch node agent SKU id that matches the image OS (e.g. `batch.node.ubuntu 24.04`).
```
```suggestion
The resource ID of a custom VM image from an Azure Compute Gallery for the pool nodes (e.g., `/subscriptions/<id>/resourceGroups/<group>/providers/Microsoft.Compute/galleries/<gallery>/images/<definition>/versions/<version>`). When set, `publisher` and `offer` are ignored, and `sku` must be the Batch node agent SKU id matching the image OS (for example, `batch.node.ubuntu 24.04`).

```

docs/azure.mdx
```diff
@@ -597,6 +597,43 @@ azure {
...
+    batch {
+        pools {
+            <pool-name> {
+                virtualMachineImageId = '/subscriptions/<subscription>/resourceGroups/<group>/providers/Microsoft.Compute/galleries/<gallery>/images/<definition>/versions/<version>'
+                sku = 'batch.node.ubuntu 24.04'
+            }
+        }
+    }
+}
+```
```
christopher-hakkaart, line 615:
````suggestion
**Custom images from an Azure Compute Gallery**

To provision pool nodes from a custom VM image published in an [Azure Compute Gallery](https://learn.microsoft.com/en-us/azure/virtual-machines/azure-compute-gallery), set `virtualMachineImageId` to the image version resource ID. When set, `publisher` and `offer` are ignored, and `sku` must be set to the Batch node agent SKU ID that matches the image operating system.

```groovy
azure {
    batch {
        pools {
             {
                virtualMachineImageId = '/subscriptions//resourceGroups//providers/Microsoft.Compute/galleries//images//versions/'
                sku = 'batch.node.ubuntu 24.04'
            }
        }
    }
}
```
````

Comment (christopher-hakkaart) [reply docs/azure.mdx:615]: I think this is slightly more task than concept. Suggestion only.

File: docs/azure.mdx [hunk 13]
```diff
@@ -597,6 +597,43 @@ azure {
...
+
+:::warning
+Custom images require Microsoft Entra authentication (service principal or managed identity). The **Azure Batch account identity** must have read access to the gallery image, and the pool typically requires `virtualNetwork` to be set.
```
christopher-hakkaart:
```suggestion
Custom images require Microsoft Entra authentication (service principal or managed identity). The **Azure Batch account identity** must have read access to the gallery image. The pool typically requires `virtualNetwork` to be set.
```

docs/azure.mdx
```diff
@@ -597,6 +597,43 @@ azure {
...
+:::
+
+**Image verification**
+
+By default, Nextflow only selects images that Azure Batch has formally verified. Set `verification` to `unverified` to also allow images that Azure Batch lists but has not verified, or to `any` to accept both. This setting is ignored when `virtualMachineImageId` is set.
```
christopher-hakkaart:
```suggestion
By default, Nextflow selects only images that Azure Batch has verified. Set `verification` to `unverified` to also allow images that Azure Batch lists but has not verified, or to `any` to accept both. Ignored when `virtualMachineImageId` is set.
```

christopher-hakkaart (commented):
I've added some editorial suggestions. I'll defer to Adam for the exact location of the more advanced content.

LennyBEL:
@adamrtalbot Would you like me to incorporate your suggested changes, or leave it as-is? You approved the PR so I'm a bit confused ;)

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
LennyBEL: Agreed. I've updated it to a property 'allowUnverifiedImages' which should explain itself. Defaults to false.

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
LennyBEL: No longer relevant with the change to the boolean

docs/reference/config.mdx
LennyBEL: No longer relevant

plugins/nf-azure/src/main/nextflow/cloud/azure/batch/AzBatchService.groovy
LennyBEL: Added an IllegalArgumentException before the default is set

Not sure if AzPoolOpts is the appropriate place here, let me know if not

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
LennyBEL: Whoops, should've done this together with the other comment

Anyway, fair remark, and it's adjusted accordingly

docs/azure.mdx
```diff
@@ -597,8 +597,45 @@ azure {
...
+                virtualMachineImageId = '/subscriptions//resourceGroups//providers/Microsoft.Compute/galleries//images//versions/'
+                sku = 'batch.node.ubuntu 24.04'
+            }
+        }
+    }
+}
+```
+
+:::warning
+Custom images require Microsoft Entra authentication (service principal or managed identity). The **Azure Batch account identity** must have read access to the gallery image. The pool typically requires `virtualNetwork` to be set.
```
adamrtalbot:
```suggestion
Custom images require Microsoft Entra authentication (service principal or managed identity). The **Azure Batch account identity** must have read access to the gallery image.
```

That's not true, is it? Even if it is, it's not relevant for this warning.

docs/azure.mdx
```diff
@@ -597,8 +597,45 @@ azure {
...
+
+```groovy
+azure {
+    batch {
+        pools {
+             {
+                virtualMachineImageId = '/subscriptions//resourceGroups//providers/Microsoft.Compute/galleries//images//versions/'
+                sku = 'batch.node.ubuntu 24.04'
+            }
+        }
```
adamrtalbot, line 612: The placeholders have been stripped from this example — the pool name is missing (`pools { { ... } }`) and the resource ID reads `/subscriptions//resourceGroups//...`. Looks like the `<...>` tokens were eaten as HTML, possibly by the GitHub web editor. The `allowUnverifiedImages` example below kept its `<pool-name>` intact.

```suggestion
        pools {
            <pool-name> {
                virtualMachineImageId = '/subscriptions/<subscription-id>/resourceGroups/<resource-group>/providers/Microsoft.Compute/galleries/<gallery>/images/<image-definition>/versions/<version>'
                sku = 'batch.node.ubuntu 24.04'
            }
        }
```

docs/azure.mdx
adamrtalbot: > The pool typically requires `virtualNetwork` to be set

Is this actually a requirement of gallery images, or just how your deployment happens to be set up? The Entra auth requirement is real, but I don't believe gallery images inherently require a virtual network. Suggest either confirming this against the Azure Batch docs or softening the wording.

docs/reference/config.mdx
```diff
@@ -509,6 +509,10 @@ The name of the batch service region, e.g. `westeurope` or `eastus2`. Not needed
 
 The client ID for an Azure [managed identity](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview) that is available on all Azure Batch node pools. This identity is used by Fusion to authenticate to Azure storage. If set to `'auto'`, Fusion will use the first available managed identity.
 
+##### `azure.batch.pools.<name>.allowUnverifiedImages`
```
adamrtalbot, line 518: New config options should carry an `<AddedInVersion version="..." />` tag for the target release, as the other recent options in this file do — here and on `virtualMachineImageId` below.

docs/reference/config.mdx
```diff
@@ -577,6 +581,11 @@ Enable the `startTask` to run with elevated access (default`false`).
 
 The `startTask` that is executed as the node joins the Azure Batch node pool.
 
+##### `azure.batch.pools.<name>.virtualMachineImageId`
```
adamrtalbot, line 592: Missing `<AddedInVersion>` tag here too. Also there's a stray double blank line after this entry's description.

plugins/nf-azure/src/main/nextflow/cloud/azure/config/AzPoolOpts.groovy
```diff
@@ -195,6 +211,10 @@ class AzPoolOpts implements CacheFunnel, ConfigScope {
         hasher.putUnencodedChars(schedulePolicy ?: '')
         hasher.putUnencodedChars(virtualNetwork ?: '')
         hasher.putBoolean(lowPriority)
+        hasher.putUnencodedChars(virtualMachineImageId ?: '')
+        // 'allowUnverifiedImages' only affects marketplace image resolution; it's ignored when a gallery image is set
```
adamrtalbot: The conditional hashing here is deliberately shaped so that default configs produce byte-identical hashes to pre-PR Nextflow (empty `putUnencodedChars` is a no-op, and the boolean is only added when set) — meaning existing users see no auto-pool-id churn on upgrade. That's the right call, but the comment only mentions the gallery-image case. Worth spelling out the back-compat reason so a future refactor doesn't "simplify" this to an unconditional `putBoolean` and invalidate every user's auto pool id:

```suggestion
        // 'allowUnverifiedImages' only affects marketplace image resolution; it's ignored when a gallery image is set.
        // NOTE: only hashed when set, so that default configs keep the same hash as previous Nextflow versions
        // (no auto-pool-id churn on upgrade) - do not change to an unconditional putBoolean
```

docs/azure.mdx
LennyBEL: Another fair remark, yeah

Guess that slipped in from a colleague and I didn't catch on it. I'll quickly investigate if it actually is needed or not, but I agree that it should be a separate warning even if it were true

adamrtalbot (commented):
Nice, this looks really good.

Just minor docs and comments to add.

docs/azure.mdx
LennyBEL: This was tackled in your earlier comment :) I've removed it, as it's indeed not true.

docs/reference/config.mdx
LennyBEL: I was thinking about that, but I don't know in which version this PR will be merged, so I found it a bit hard to include that here already

I assume that this tag would be added by whomever eventually merges the PR?

Of course, I can add it if you'd like

adamrtalbot: Assuming this was merged, it would be 26.10

adamrtalbot (approved):

bentsherman (approved):

PR #7342, opened 2026-07-15
Workflow modules

bentsherman:
Spun out from #7213 

This PR proposes an ADR for workflow modules -- extending the module system to support both processes and workflows.

Updated based on latest round of feedback

### DIFF
--- a/adr/20260608-workflow-modules.md
+++ b/adr/20260608-workflow-modules.md
@@ -1,0 +1,199 @@
+# Workflow modules
+
+- Authors: Ben Sherman
+- Status: draft
+- Date: 2026-06-08
+- Tags: workflows, modules, registry
+
+## Summary
+
+Add the ability to include a workflow as a module from the Nextflow registry.
+
+## Problem Statement
+
+A Nextflow *module* is currently defined as a standalone process definition (with corresponding spec file). The [module system ADR](20251114-module-system.md) defines how these modules are published, distributed, and executed through the Nextflow registry.
+
+There is a similar need to share and re-use workflows / subworkflows. The nf-core community has curated a collection of re-usable subworkflows in the [nf-core/modules](https://github.com/nf-core/modules) repository.
+
+The module system should be extended to include both standalone *processes* and *workflows*.
+
+## Goals
+
+- **Workflow inclusion**: allow workflows to be included from the registry using the same include syntax as for modules.
+
+- **Workflow execution**: allow workflows to be executed directly via `nextflow module run`.
+
+- **Reuse existing infrastructure**: the module system already defines a registry API and CLI for publishing and installing modules. Treating workflows as a separate concept would require duplicating much of this infrastructure.
+
+## Non-goals
+
+- **Pipeline inclusion**: *pipelines* are distinct from *workflows* -- they specify params, publishing, and config, not just workflow logic.
+
+## Decision
+
+Extend the definition of *module* to include both standalone *processes* and standalone *workflows*. Allow workflow modules to be published, installed, included, and executed via the Nextflow registry. Store the included workflow in the including repository under `workflows/<scope>/<name>/`.
+
+## Core Capabilities
+
+### Workflow modules vs process modules
+
+A module can refer to a *workflow module* or *process module*, depending on whether it defines a workflow or process.
+
+Workflow modules can be published to and queried from the Nextflow registry using the same modules API and `nextflow module` CLI. Workflow modules also have the same directory structure as process modules.
+
+Since workflow modules are just modules, a process module and workflow module cannot have the same name in a module namespace.
+
+### Module spec
+
+The module spec is extended as follows in order to support standalone workflows:
+
+- A `kind` field to distinguish between workflow modules (`kind: Workflow`) and process modules (`kind: Process`).
+
+- A `requires.modules` field to specify *transitive dependencies*, since a workflow can depend on other processes and workflows.
+
+- Certain fields (`topics`, `tools`) cannot be specified for workflow modules because they are not applicable.
+
+For example:
+
+```yaml
+name: nf-core/fastq_align_star
+kind: Workflow
+version: 0.0.0-4e3e10e
+description: Align reads to a reference genome using bowtie2 then sort with samtools
+authors:
+  - "@JoseEspinosa"
+license: MIT
+requires:
+  nextflow: ">=24.04.0"
+  modules:
+    - nf-core/star/align@0.0.0-4e3e10e
+    - nf-core/samtools/sort@0.0.0-4e3e10e
+    - nf-core/samtools/index@0.0.0-4e3e10e
+    - nf-core/samtools/stats@0.0.0-4e3e10e
+    - nf-core/samtools/idxstats@0.0.0-4e3e10e
+    - nf-core/samtools/flagstat@0.0.0-4e3e10e
+    - nf-core/bam_sort_stats_samtools@0.0.0-4e3e10e
+```
+
+### Workflow inclusion and storage
+
+Workflow modules use the same include syntax and naming conventions as process modules:
+
+```groovy
+// process
+include { BWA_MEM } from 'nf-core/bwa/mem'
+
+// workflow
+include { FASTQ_ALIGN_STAR } from 'nf-core/fastq_align_star'
+```
+
+When a workflow is included, it is vendored into the including project under `workflows/<scope>/<name>/`. Included workflows should be committed to the including repository.
+
+Transitive dependencies (specified by `requires.modules`) should also be installed in the `modules/` and `workflows/` directories alongside the included workflow. Since modules are flattened, it is not possible for a pipeline to use two different versions of the same process or workflow.
+
+### Workflow execution
+
+Typed workflows can be executed directly by inferring the `params` and `output` blocks from the `take:` and `emit:` sections.
+
+For example, given the following workflow:
+
+```groovy
+workflow RNASEQ {
+    take:
+    samples: Channel<Sample>
+    index: Path
+
+    main:
+    ch_aligned = ALIGN(samples, index)
+    multiqc_report = MULTIQC(ch_aligned.collect())
+
+    emit:
+    aligned: Channel<AlignedSample> = ch_aligned
+    multiqc_report: Path = multiqc_report
+}
+
+record Sample {
+    id: String
+    fastq_1: Path
+    fastq_2: Path
+}
+
+record AlignedSample {
+    id: String
+    bam: Path
+    bai: Path
+}
+```
+
+The user can run the workflow directly as follows:
+
+```bash
+nextflow module run rnaseq.nf \
+    --samples input.csv \
+    --index index.fasta
+```
+
+Nextflow executes the `RNASEQ` workflow as if it were wrapped in the following entry workflow:
+
+```groovy
+params {
+    samples: Channel<Sample>
+    index: Path
+}
+
+workflow {
+    main:
+    rnaseq = RNASEQ(params.samples, params.index)
+
+    publish:
+    aligned = rnaseq.aligned
+    multiqc_report = rnaseq.multiqc_report
+}
+
+output {
+    aligned: Channel<AlignedSample> {}
+    multiqc_report: Path {}
+}
+```
+
+This way, the user can run a workflow directly without having to write an entry workflow for it.
+
+Direct execution requires the ability to load an input channel directly from an index file (samplesheet). This can be done by loading the samplesheet data based on the file extension (CSV, JSON, YAML), casting each record to the given record type, and loading the collection as a channel. The data-loading function may be extended via plugin to support additional formats (e.g. Parquet).
+
+The channel input can use a generic type such as `Map` or `Record`, or a custom record type to enable further validation. In the above example, using the `Sample` type ensures that each samplesheet row is validated against the record fields and the `fastq_1` and `fastq_2` columns are treated as file paths.
+
+When executing a named workflow directly, output files are not published to an output directory. Instead, the workflow output printed by Nextflow simply refers to output files by their work directory path.
+
+## Alternatives
+
+### Workflows vs processes
+
+One alternative is to treat workflows as a separate concept from modules, restricting the definition of *module* to only include standalone processes.
+
+However, the broader meaning of *module* is a re-usable component, and both processes and workflows are re-usable components, so it makes more sense to extend the module system rather than introduce a parallel system for workflows. A parallel system would also require duplicating a lot of existing code (registry API, CLI, etc).
+
+Process modules and workflow modules can be distinguished by different specializations of the module spec (`kind: Process` vs `kind: Workflow`) and different storage locations (`modules/` vs `workflows/`).
+
+### Workflows vs subworkflows
+
+The nf-core community makes a distinction between *workflows* and *subworkflows*:
+
+- Subworkflows are stored in the `subworkflows/` directory and can be shared across pipelines.
+
+- Workflows are stored in the `workflows/` directory and are owned by a specific pipeline.
+
+Nextflow makes no such distinction, as there is no functional difference between a workflow and a subworkflow. Both are stored in the `workflows/` directory and both can be published and installed through the Nextflow registry.
+
+The Nextflow registry makes it easier for pipelines to share and reuse workflows, since the pipeline can be the source-of-truth rather than having to surrender ownership to the `nf-core/modules` repository.
+
+### Workflows vs pipelines
+
+One alternative is to treat all workflows as pipelines, restricting the definition of *module* to only include standalone processes.
+
+Aside from the problems mentioned under [Workflows vs processes](#workflows-vs-processes), this approach conflates two commonly-understood concepts: a *pipeline* which is an end-to-end analysis, and a *workflow* or *subworkflow* which is a component in a larger analysis. It is more congruent with conventional understanding to treat workflows (subworkflows) as a separate concept from pipelines.
+
+## Links
+
+- [nf-core terminology](https://nf-co.re/docs/usage/getting_started/terminology)
+- Refines: [Module system](20251114-module-system.md)
+- Related: [Remote pipeline inclusion](20260608-remote-pipeline-inclusion.md)
--- a/adr/20260608-workflow-modules.md
+++ b/adr/20260608-workflow-modules.md
@@ -1,7 +1,7 @@
 # Workflow modules
 
 - Authors: Ben Sherman
-- Status: draft
+- Status: proposed
 - Date: 2026-06-08
 - Tags: workflows, modules, registry
 
--- a/adr/20260608-workflow-modules.md
+++ b/adr/20260608-workflow-modules.md
@@ -23,7 +23,7 @@ The module system should be extended to include both standalone *processes* and 
 
 - **Workflow execution**: allow workflows to be executed directly via `nextflow module run`.
 
-- **Reuse existing infrastructure**: the module system already defines a registry API and CLI for publishing and installing modules. Treating workflows as a separate concept would require duplicating much of this infrastructure.
+- **Reuse existing infrastructure**: the module system already defines a registry API and CLI for publishing and installing modules. Extend this system to include workflows instead of treating workflows as a separate concept.
 
 ## Non-goals
 
@@ -31,7 +31,7 @@ The module system should be extended to include both standalone *processes* and 
 
 ## Decision
 
-Extend the definition of *module* to include both standalone *processes* and standalone *workflows*. Allow workflow modules to be published, installed, included, and executed via the Nextflow registry. Store the included workflow in the including repository under `workflows/<scope>/<name>/`.
+Extend the definition of *module* to include both standalone *processes* and standalone *workflows*. Allow workflow modules to be published, installed, included, and executed via the Nextflow registry. Store installed workflows under `modules/<scope>/<name>/` alongside process modules. Allow each workflow module to have its own `modules` directory for transitive dependencies.
 
 ## Core Capabilities
 
@@ -51,15 +51,20 @@ The module spec is extended as follows in order to support standalone workflows:
 
 - A `requires.modules` field to specify *transitive dependencies*, since a workflow can depend on other processes and workflows.
 
-- Certain fields (`topics`, `tools`) cannot be specified for workflow modules because they are not applicable.
+- Fields that are relevant only to process modules (i.e. `topics`, `tools`) cannot be specified for workflow modules.
 
-For example:
+For example, [nf-core/fastq_align_star](https://github.com/nf-core/modules/tree/master/subworkflows/nf-core/fastq_align_star):
 
 ```yaml
 name: nf-core/fastq_align_star
 kind: Workflow
 version: 0.0.0-4e3e10e
-description: Align reads to a reference genome using bowtie2 then sort with samtools
+description: Align reads to a reference genome using STAR then sort, index, and compute statistics with samtools
+keywords:
+  - align
+  - fasta
+  - genome
+  - reference
 authors:
   - "@JoseEspinosa"
 license: MIT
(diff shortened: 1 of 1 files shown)


bentsherman:
Addressing @pditommaso 's feedback in more detail:

> **1. The `workflows/` prefix seems used a little inconsistently.**

ADR updated -- both process modules and workflow modules are now stored in the `modules` directory

> **2. Worth clarifying that a workflow module can include *any* module.**

The ADR does specify this, but I added an example of a workflow module's directory tree to make it more clear

> **3. The workflow interface is probably best defined by `take:`/`emit:` rather than duplicated in the spec.**

Agreed. I added an example of inputs/outputs in the module spec. At the very least we need `name` and `description`. Not sure if `type` is really needed for v1.

> **4. Dependencies + versioning — there's a bit of tension here.**

ADR updated -- workflow modules now get their own `modules` directory to store nested dependencies. It will make for some long file paths but it seems like the right trade-off.

> **5. The params/output *mapping* might be better as its own ADR.**

I kept this section here for now so that I can continue to refine the mapping rules. I will probably just move it back to #7208

adr/20260608-workflow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+                                │   └── meta.yml
+                                └── samtools/flagstat/
+                                    ├── .module-info
+                                    ├── main.nf
+                                    └── meta.yml
+```
+
+Note that `fastq_align_star` declares only its *direct* dependencies (`star/align` and `bam_sort_stats_samtools`) in `requires.modules`. The `samtools` processes are transitive dependencies of `bam_sort_stats_samtools` and are vendored within *its* own `modules` directory, not directly under `fastq_align_star`. Because each workflow module vendors its own dependencies, the same module may appear more than once in the tree when it is used by multiple workflows.
+
+### Workflow execution
```
bentsherman: TODO: clarify mapping rules for CLI params -> workflow takes, beyond the one example

Consider moving back to #7208

adr/20260608-workflow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Transitive dependencies
+
+Workflow modules, unlike process modules, can include other modules (both processes and workflows). As a result, we must address the *diamond dependency problem* -- what happens when two workflows depend on different versions of the same module?
+
+Currently, nf-core enforces a flat module structure. nf-core modules are currently versioned by commit hash and workflows pin exact versions. There is no way to resolve conflicts between workflows. Instead, all nf-core workflows are kept in sync at publish time. This approach places the maintenance burden on *module developers* -- when updating a module, you must also update all consuming workflows.
+
+Instead, the ADR avoids this problem entirely by allowing each workflow to have its own `modules` directory. The downside is that modules will be duplicated in the project, making the project larger and increasing the number of scripts to be parsed.
```
bentsherman, line 269: TODO: evaluate the impact on project size

It should be minimal if we install only the script+spec and exclude tests. In principle you don't need the tests in your pipeline if they were verified when the module was published

bentsherman: We might also consider the impact on compile time, since Nextflow will end up parsing more scripts when duplication is allowed. But this should be minimal compared to overall runtime

adr/20260608-workflow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+### Workflow inclusion and storage
+
+Workflow modules use the same include syntax and naming conventions as process modules:
+
+```groovy
+// process
+include { BWA_MEM } from 'nf-core/bwa/mem'
+
+// workflow
+include { FASTQ_ALIGN_STAR } from 'nf-core/fastq_align_star'
```
bentsherman, line 141: Transitive dependencies also complicates how we sync nf-core subworkflows to the registry. They have includes based on nf-core's directory structure (flat, separate modules vs subworkflows) whereas ideally they should use the canonical module name as shown here

Perhaps we can be clever and re-write these include statements when importing to the registry? Should be possible with some regex, or an AST transform is always an option

jorgee:
I have some differences about how modules are managed in workflows and pipelines. In modules, we propose to use the `module.requires` and they will be stored in the workflow `modules` folder during the installation. However, the current module version pinning in pipelines is done by committing the module to the `modules` folder. If the module is not committed, Nextflow uses the latest.
 
My doubt is why we do not apply the pipelines way for workflows? So, the version is pinned if the module is included in the folder; otherwise, use the latest. In this way, we will not need the `module.requires` and looking at the repo, it seems to match the current behaviour in the nf-core subworkflows repo. They use relative paths to the modules folder in the repo, so it is pointing to the latest version. Adopting it, the nf-core subworkflows import will be basically uploading `main.nf, changing the relative path in imports by the module references.  

The problem that I could see is if there are a big amount modules pinned that could pass the 1MB limit for the upload compressed artifact.

bentsherman:
Are you talking about bundling the transitive modules in the workflow module? i suppose we could do that with just a directory tree of `.module-info` files. But it just doesn't seem necessary when we already have the module spec

The version pinning approach we took for modules was mostly a way to avoid depending on the pipeline spec since we don't have it yet. Once we have the pipeline spec, I could see us moving the version pins to there. So I wouldn't worry too much about consistency there

jorgee:
> Are you talking about bundling the transitive modules in the workflow module? 

Yes, I was thinking it also to simplify the import of nf-core subworkflow. But it can make publisehd subworkflows unreliable. I think we can extract the nf-core modules versions from the git commit.

awgymer:
> it seems to match the current behaviour in the nf-core subworkflows repo. They use relative paths to the modules folder in the repo, so it is pointing to the latest version

Yeah... my experience of this from working closely with `nf-core` components is that it was a design mistake and any new system should avoid it because it causes so many headaches. 

If it's about concordance with the way modules operate currently I'd argue it would be better to just make people have to pin modules too. 

> The problem that I could see is if there are a big amount modules pinned that could pass the 1MB limit for the upload compressed artifact.

FWIW I tested running a simple `zip` on the entire `nf-core/modules` folder (only the `.nf` and meta yml files no test code) and that would be `4.4M` for about ~2.5k modules. I think a 1MB limit on compressed size would be quite difficult to hit for even large subworkflows. All the equivalent files in one of my personal nf-core like pipelines was `153K` . 

**EDIT**: 
Just thinking about this further, I think that it's fairly imperative that users are able to update the version of modules which are pinned within a workflow module type locally. i.e. I might install `nf-core/complex_subworkflow` and for some specific reason I may want to upgrade or downgrade process module within that workflow. With that in mind it may make sense to have the registry just carry the manifest and then the default install pull the versions in that manifest into the nested module tree and you would have some e.g. `nextflow module update-deps` command which would allow the user to manually update the deps in said manifest and then refresh the install. 

Of course this would probably require the top-level manifest to carry the full nested module tree of dependencies because otherwise it would become tricky to update a nested workflow module.

jorgee:
> Just thinking about this further, I think that it's fairly imperative that users are able to update the version of modules which are pinned within a workflow module type locally. i.e. I might install `nf-core/complex_subworkflow` and for some specific reason I may want to upgrade or downgrade process module within that workflow. With that in mind it may make sense to have the registry just carry the manifest and then the default install pull the versions in that manifest into the nested module tree and you would have some e.g. `nextflow module update-deps` command which would allow the user to manually update the deps in said manifest and then refresh the install.

Not sure if I have really get your point. Do you want to update the dependencies of an already installed workflow (locally) or update the dependencies of a published workflow in the remote server? For the first one, I think you can do it with the current `module install` commands. If you go to `modules/nf-core/complex_subworkflow` and you run `nextflow module install <module-to-update>@<version>`, it will override the module in the current workflow installation. I am not sure if this command should also update the `module.requires` in the local workflow manifest or not. For the second case, you can publish a new version of the workflow with the updated manifest.

adr/20260608-workflow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+  - fasta
+  - genome
+  - reference
+authors:
+  - "@JoseEspinosa"
+license: MIT
+requires:
+  nextflow: ">=24.04.0"
+  modules:
+    - nf-core/star/align@0.0.0-4e3e10e
```
jorgee, line 72: is `requires.modules` just allowing to pin a version or are we going to support ranges or constraints (< 2.0.0)

awgymer:
The first. 

One issue right now is that `nf-core` does not have "real" versioning of modules. Its arbitrary git commits and a different "version" might mean a logic change or it might mean formatting, or a container which can be overridden in config anyway - or any combination of these. 

My issue with needing to do `modules install` in arbitrarily nested locations is:
- highly manual per-module 
- not traceable
- prone to human error

adr/20260608-workflow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+  - align
+  - fasta
+  - genome
+  - reference
+authors:
+  - "@JoseEspinosa"
+license: MIT
+requires:
+  nextflow: ">=24.04.0"
+  modules:
```
jorgee, line 71: Regarding this, is it required to validate if the required modules have been published before, or is it a workflow developer's responsibility? Ideally, it we will not publish incomplete workflows or with mistakes in the dependencies. However, it could be tricky because there are workflows that can be published in a registry depending on modules stored in other registries, and the dependency resolution will depend on the local configuration os the user (repositories set in `registry.urls`). For instance, an organization that is hosting its own registry used to store workflows that depend on modules in the public registry.

awgymer: What does eg cargo/pypi do for this?

jorgee: Asking Claude, seems both options are used and depends on the registries architecture. It seems cargo checks because they have the central (crates.io). However, Maven does not check dependencies and the error is discovered at install time.

awgymer: As a user I think it would be a much worse experience if the validation didn't happen at publish time. I think it's also worth considering that nextflow has so far been a primarily bioinformatics platform and that many users are not necessarily coming from engineering backgrounds and the friction of failing installs and/or not providing publish-time checks could carry more risk than in some languages

adr/20260608-workflow-modules.md
bentsherman: My inclination is to not implement version ranges for v1, since we can always add it later

It depends on whether other module namespaces will want to use version ranges or if they will want to be like nf-core

adr/20260608-workflow-modules.md
bentsherman: I figured we would validate this at publish time. Publishing a workflow module should require that all module dependencies are already published in the same registry

Cross-registry deps can be handled for now by mirroring the public registry into the private one

adr/20260608-workflow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+// process
+include { BWA_MEM } from 'nf-core/bwa/mem'
+
+// workflow
+include { FASTQ_ALIGN_STAR } from 'nf-core/fastq_align_star'
+```
+
+When a workflow is included, it is vendored into the including project under `modules/<scope>/<name>/`. Included workflows should be committed to the including repository.
+
+Each workflow module should store its dependencies within its own `modules` directory. This way, two workflows can use different versions of the same module without introducing a version conflict. Dependencies are not bundled with the workflow module -- they are installed when the workflow module is installed.
```
bentsherman, line 146: @awgymer pulling your question into a thread:

> One issue right now is that `nf-core` does not have "real" versioning of modules. Its arbitrary git commits and a different "version" might mean a logic change or it might mean formatting, or a container which can be overridden in config anyway - or any combination of these. 

> My issue with needing to do `modules install` in arbitrarily nested locations is:
> - highly manual per-module 
> - not traceable
> - prone to human error 

In theory, I think you would be able to edit the `requires.modules` section of an installed workflow module, changing versions to your liking, then run `nextflow modules install/update` once in the project root and Nextflow would recursively install modules to align it with the module specs. Nextflow would just warn about locally-modified modules for certain operations

This is different from how we currently manage modules though, since we don't have a pipeline spec to pin module versions right now but workflow modules will pin their dependencies. So we'll have to think about this more

adr/20260608-workflow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+
+Aside from the problems mentioned under [Workflows vs processes](#workflows-vs-processes), this approach conflates two commonly-understood concepts: a *pipeline* which is an end-to-end analysis, and a *workflow* or *subworkflow* which is a component in a larger analysis. It is more congruent with conventional understanding to treat workflows (subworkflows) as a separate concept from pipelines.
+
+### Transitive dependencies
+
+Workflow modules, unlike process modules, can include other modules (both processes and workflows). As a result, we must address the *diamond dependency problem* -- what happens when two workflows depend on different versions of the same module?
+
+Currently, nf-core enforces a flat module structure. nf-core modules are currently versioned by commit hash and workflows pin exact versions. There is no way to resolve conflicts between workflows. Instead, all nf-core workflows are kept in sync at publish time. This approach places the maintenance burden on *module developers* -- when updating a module, you must also update all consuming workflows.
```
bentsherman, line 267: @awgymer this is the main assumption I wanted to check from the nf-core side

My sense is that module developers will not want to deal with semantic versioning and version ranges across thousands of modules, which is why we are going down this route of just duplicating the modules in every workflow

I'm curious which trade-off you think folks would prefer?

adr/20260608-workflow-modules.md
awgymer: Yeah a pipeline really needs a way to pin module versions I think (other than at the include point)

adr/20260608-workflow-modules.md
awgymer: It will be a split I expect, but personally I would rather we moved to semantic versioning - although a transition period within nf-core would probably be needed

muffato: I think introducing another versioning on the modules/nf-core side will be confusing and a pain to maintain. Upstream software often don't follow semantic versioning. We shouldn't burden module developers with finding out how backwards-compatible each software update is to come up with a parallel semantic version.
If we want a version number, I'd stick to the version of the wrapped software itself, with a "build" number attached to it like in Conda. Every time the software is updated, the build number goes back to 0. Every time the module itself is updated, the build number is incremented.

awgymer: I don't think the version has anything to do with the internal tool really? 

Semantic versioning is reasonably clear on what the different levels mean but it's also easy enough to define one for nextflow. Something like: 
- major: i/o channel interface changes 
- minor: non-breaking functional change (ie container bump) 
- patch: formatting, meta yml etc 

The current system of git sha gives the user absolutely 0 indication and also has no way to tell at a glance if a version is newer or older than another one. 

You could pin it to the internal tool version *but*:
- some modules use multiple tools 
- build versions again give no indication if the nextflow module has breaking changes or is purely formatting

jorgee:
Apart from the `install`, `publish` and `run`, there are other implications of adding workflows in other `module` command.
- list: Include the module kind
- spec: Generate the spec for a workflow
- validate: Validate a workflow module (here we could check if dependencies exists according to the configuration)
- search: filter by 'kind'
- create: Add an option to generate a basic workflow module?
- view: It should render workflow modules

jorgee:
In `module run` for processes, we support running with no static types, relying on the process input definitions (`path`, `val`, `tuple`) and the `meta.yml` to convert CLI to the input channels. 

In workflows, I see the ADR focuses on typed workflows. Are we only going to support `module run` executions for static typed workflows, or both? 

If it is just for typed workflows, nf-core subworkflows are not currently typed, so if we try to automatically import nf-core subworkflows in the current state, they will not be used in module run. So, I wonder if this automatic migration makes sense. 

To support non-typed workflows, we should rely on a good `meta.yml` description as processes. Current, the ADR examples and the nf-core workflows schema only set a type per channel; however, it is not possible to describe the internal structure of the channel. Looking at the nf-core subworkflow meta.ymls, it looks like some of them have tried to define the channel structure. See below some examples:

  1. structure: property — used but absent from the schema. `mafft_align`:
  - ch_fasta:
      type: file
      pattern: "*.{bam/cram/sam}"        # ← stale copy-paste: it's a FASTA subworkflow
      structure:
        - - meta: {type: map, ...}
          - fasta: {type: file, pattern: "*.{fa,fasta}", ontologies: [...]}
  structure nests the tuple's element definitions — the schema doesn't define it (0 occurrences), so it's unvalidated.

  2. Structure encoded in free-text description. `fastq_align_star`:
  - ch_reads:
      description: |
        List of input FastQ files ...
        Structure: [ val(meta), [ path(reads) ] ]   # ← tuple shape as prose
  (note ch_reads here has no type at all.)

  3. Input keys = flattened internal channel elements, NOT the take channels. `bam_sort_stats_samtools`:
  take:  ch_bam, ch_fasta_fai            # 2 channels
  input: meta, bam, fasta                # 3 entries — the elements INSIDE the channels
 
 So, in case we want to support the execution of non-typed workflow we should include a structure or payload property to describe the internal structure of the channel.

adr/20260608-workflow-modules.md
bentsherman: After discussing with the team, I am less worried about this question. I think as long as workflow modules depend only on other registry modules, Nextflow should be able to duplicate or deduplicate the dependency tree at install-time

So I think the current proposal makes sense as a first step, and we can add support for version ranges / resolution in the future

bentsherman:
@jorgee I just don't think there is enough information in legacy workflows, even in the `meta.yml`, to support direct execution. That's why I de-scoped it, although we can always try in the future

I don't think it's a blocker for registry auto-import though. The `module run` command should just fail for legacy workflows. This functionality doesn't exist yet anyway so there's no regression. The core feature is the publish / install / inclusion

modules/nextflow/src/main/groovy/nextflow/script/IncludeDef.groovy
```diff
@@ -181,7 +181,11 @@ class IncludeDef {
     Path resolveRemoteModulePath(String moduleName) {
         // Use SPI to get the remote module resolver implementation
         def resolver = RemoteModuleResolverProvider.getInstance()
-        return resolver.resolve(moduleName, session.baseDir)
+        // Resolve relative to the including file's directory (context-relative) so that a
+        // workflow module's own dependencies are found under its nested `modules/` directory
+        // (nested vendoring). For the top-level script this is the project base dir.
+        final base = getOwnerPath()?.getParent() ?: session.baseDir
```
pditommaso: Really like the nested-vendoring direction here 👍 — one edge case on the top-level fallback worth a look.

The comment says *"For the top-level script this is the project base dir"*, but `getOwnerPath()?.getParent()` is the **entry script's** directory, which only equals `session.baseDir` when the entry script sits at the project root. With `manifest.mainScript = 'workflows/main.nf'`, `baseDir` stays `<root>` while `getOwnerPath().getParent()` becomes `<root>/workflows` — so a top-level remote include would resolve against `<root>/workflows/modules/` instead of `<root>/modules/` where `module install` vendors it, and fail to find it.

Might be worth keying the top-level case off `baseDir` explicitly (e.g. only use the owner's parent for nested/module contexts), or confirming that a subdir `mainScript` is out of scope for now.

modules/nf-lang/src/main/java/nextflow/script/control/ResolveIncludeVisitor.java
```diff
@@ -137,14 +137,18 @@ private static void setPlaceholderTargets(IncludeNode node) {
     }
 
     private URI getIncludeUri(String source) {
+        var parent = Path.of(uri).getParent();
```
pditommaso: Minor robustness nit (not a blocker): hoisting `Path.of(uri)` to the top means it now runs for the remote-module branch too — previously that branch only used `projectDir` and never touched `uri`.

For `nextflow run` this is a non-issue since source URIs are real `file:` paths. But in the language server an unsaved/virtual document has a non-`file:` URI (e.g. `untitled:`), and `Path.of(uri)` throws for those — so a remote include that used to resolve via `projectDir` would now surface a spurious diagnostic. The surrounding try/catch keeps it from crashing, and local includes already had this limitation, so it's minor — but if you'd like remote includes to keep working on virtual docs, computing `parent` lazily (only in the branches that need it) would preserve the old behavior. Same pattern appears to be in `ModuleResolver.java` ~line 94, if so worth the same treatment.

pditommaso (commented):
Took a focused pass on the include-resolution changes — the nested-vendoring approach reads well. Two spots worth a second glance before merge (details inline):

- **`IncludeDef.resolveRemoteModulePath`** — the top-level fallback assumes the entry script's parent equals `baseDir`, which breaks for a subdir `manifest.mainScript`.
- **`ResolveIncludeVisitor.getIncludeUri`** — a minor language-server robustness nit from hoisting `Path.of(uri)` above the remote branch.

Neither is a blocker if those configs are considered out of scope — just flagging so it's a deliberate call. Nice work on the feature overall! 🙌

modules/nf-cli-v1/src/main/groovy/nextflow/cli/module/CmdModuleCreate.groovy
```diff
@@ -137,13 +162,84 @@ class CmdModuleCreate extends CmdBase {
...
+
+            nextflow.enable.types = true
+
+            workflow ${defName} {
+                take:
+                greeting: String
+
+                main:
+                // TODO: implement the workflow logic
+                message = greeting
```
pditommaso, line 183: Maybe we should public an `Hello` basic process, and use in the workflow template

pditommaso:
What's the current limitation on `module run` for workflow, using the the basic example it return the error 

```
Workflow `HELLO` cannot be executed directly because it is not typed -- static typing is required to map pipeline parameters to workflow inputs

```

christopher-hakkaart (approved):
Docs look good

jorgee:
> What's the current limitation on `module run` for workflow, using the the basic example it return the error
> 
> ```
> Workflow `HELLO` cannot be executed directly because it is not typed -- static typing is required to map pipeline parameters to workflow inputs
> ```

The `nextflow module run` for workflows only works if the workflow is typed. 

I added a `typed` flag in `module run create`. When this is defined, it generates typed code, and you can run with `module run`. Moreover, in the create command answer,  it says how to run the module when`typed` is enabled; otherwise, it says how to include the module in the workflow without typed. 

Rethinking it, one option to improve it is creating typed workflows by default, but typed workflows are currently in preview (not sure it will be the same in 26.10). At least, I could add a more visible note when the user tries to create a non-typed workflow to warn that `nextflow module run` will not work in this case.

pditommaso:
> Rethinking it, one option to improve it is creating typed workflows by default

Let's do this

bentsherman:
Updates:
- Incorporated discussions from this PR into the ADR
- Added emergent architectural decisions from implementation to the ADR (where appropriate)
- Fixed several bugs/gaps found during review
- Changed `module create` to create typed modules by default so that scaffolded workflow modules are executable by default (static typing will be stable in 26.10)

modules/nf-cli-v1/src/main/groovy/nextflow/cli/module/CmdModuleCreate.groovy
bentsherman: I'm good with that. I can set up a `nextflow-io` namespace and upload a `hello` module to it (likely just the one generated by `module create`)

We could even create a `nextflow-io/modules` repo and maintain the module there, but that might be overkill

pditommaso: Let's do that! (hello module)

bentsherman: Done. Scaffolded workflow module uses this module dep: https://registry.nextflow.io/modules/nextflow-io/hello@1.0.0

(which is just the process module created by `module create`)

adr/20260608-workflow-modules.md
bentsherman: Updated ADR with analysis of impact on nf-core/rnaseq. Updated `module install` to not install `tests` directory from registry bundle

pditommaso:
Given a try to this, however I'm getting a little bump 

```
» ../nextflow/launch.sh module create -kind Workflow
Enter module namespace: nf-foo
Enter module name: Hello

  Module namespace : nf-foo
  Module name      : Hello
  Directory        : ./modules/nf-foo/Hello

Are you OK to continue [y/N]? y
Module created successfully at path: modules/nf-foo/Hello

To run the module:

  nextflow module run nf-foo/Hello --greeting 'Hello world!'

» ../nextflow/launch.sh module run nf-foo/Hello --greeting 'Hello world!'
Invalid module reference: nf-foo/Hello
```

Am I doing something wrong?

pditommaso:
It's a casing mismatch between the two subcommands, not a usage error.

`module create` validates the name with `[a-zA-Z0-9][a-zA-Z0-9._\-]*` (uppercase allowed), so `Hello` is accepted and it even prints the run command. But `module run` parses the reference through `ModuleReference.parse`, whose pattern requires a **lowercase** name:

```
^([a-z0-9][a-z0-9._\-]*)\/([a-z][a-z0-9._\-]*(?:\/[a-z][a-z0-9._\-]*)*)$
```

So `nf-foo/Hello` fails on the uppercase `H`; `nf-foo/hello` works. `module create` should enforce the same rule as `ModuleReference` (ideally reuse `ModuleReference.parse`) so it can't create un-runnable modules.

awgymer:
What is the distinction or interplay here with module "name" and the directory path "name". 

For example in nf-core `nf-core/hello` would be the required path name but the module name itself in the file or for usage in a workflow would still be `HELLO`.

bentsherman:
@pditommaso fixed the issue with `module create`

@awgymer I guess modules do have two different kind of names, the "registry name" (`nf-core/hello`) and the "definition name" (`HELLO`). We'll have to clarify this in the docs. They are largely independent of each other since they are used in different places

pditommaso (approved):
Thanks Ben, re-tested and the `module create` fix works nicely.

Reusing `ModuleReference.parse` is exactly right — the naming rules now live in one place, and validation happens before the confirmation prompt so nothing gets scaffolded on a bad name. Verified end to end:

- `module create nf-foo/Hello` (and the interactive equivalent) now fails fast with a clear message and creates no directory
- `module create` → `module run nf-foo/hello --greeting 'Hello world!'` runs through to `[SUCCESS]`, so the command printed by `create` actually works now

LGTM 👍

PR #7344, opened 2026-07-16
Drop transient object tags when publishing files to S3

### LINKED ISSUE #7339: Nextflow preserves the temporary tag when publishing files
## Bug report 

By default, Fusion tags all task outputs (via `fusion.tags`) as `nextflow.io/temporary=true`, so that the user can delete intermediates automatically using a retention policy

However, when Nextflow publishes task outputs to the output directory, it preserves this temporary tag, which means that a retention policy on the output directory will still delete "permanent" outputs

Nextflow should drop this tag when publishing outputs to object storage

pditommaso:
## Problem

When Nextflow copies an S3 object from one bucket to another (e.g. `publishDir`), it was propagating the source object's tags to the destination.

By default Fusion tags all task outputs with `nextflow.io/temporary=true` (via `fusion.tags`) so intermediates can be reaped by a retention policy. Because these tags were carried over on publish, a retention policy on the output directory would also delete "permanent" published outputs.

## Root cause

`S3Client.copyFile()` only set the `REPLACE` tagging directive when the publish target had explicit tags:

```java
if( tags != null && !tags.isEmpty()) {
    reqBuilder.taggingDirective(TaggingDirective.REPLACE);
    reqBuilder.tagging(...);
}
```

With no tags the directive was left unset, so the S3 `CopyObject` API defaulted to `COPY` — inheriting the source object's tags on the destination.

## Fix

Always set the `REPLACE` tagging directive so the destination gets exactly the requested tag set (empty = no tags) rather than inheriting source tags. Extracted into a testable `applyTagging` helper.

## Test

Added a unit test in `S3ClientTest` verifying `REPLACE` is always used and tags are set only when provided.

Closes #7339


jorgee (commented):
I think it is the simplest case to solve the issue, and it is the correct behaviour for `PublishDir`. What I don't like about the solution is that it will apply to all s3-to-s3 copies done by Nextflow.  Basically, it will only affect `nextflow fs cp/mv` commands because the file porter will not perform any s3-to-s3 copy because they are not considered foreign files. 

So I see two options: we can keep the current code that removes tags in` fs `command and document this change; or add a new field in `TagAwareFile` to set the tagging directive (replace or not) and set it just in the `PublishDir` case.

pditommaso:
`nextflow fs cp/mv` is mostly a debugging/testing and secondary tooling command. It was never meant to be used in production pipelines, and there's no spec for how it should behave with tags. I think it's fine to ignore that case for this issue.

plugins/nf-amazon/src/main/nextflow/cloud/aws/nio/S3Client.java
```diff
@@ -591,12 +591,23 @@ public void uploadDirectory(File source, S3Path target) throws IOException {
...
+        reqBuilder.taggingDirective(TaggingDirective.REPLACE);
+        if( tags != null && !tags.isEmpty() ) {
             log.debug("Setting tags: {}", tags);
-            reqBuilder.taggingDirective(TaggingDirective.REPLACE);
             reqBuilder.tagging(Tagging.builder().tagSet(tags).build());
         }
+    }
+
+    public void copyFile(CopyObjectRequest.Builder reqBuilder, List<Tag> tags, String contentType, String storageClass) throws IOException {
+        applyTagging(reqBuilder, tags);
```
bentsherman, line 610: This covers file copies but what about directory copies? Fusion applies the temporary tag recursively to directory outputs so we need to do the same

jorgee: When publishing a directory, Nextflow calls `CopyMoveHelper.copyDirectory` and it is walking the files in the directory and performing a `FileSystemProvider.copy` per file.

bentsherman (approved):
Looks good. We can always add a CLI option to `nextflow fs cp` for specifying tags if someone asks for it

bentsherman:
@pditommaso do you want to backport to 26.04 ?

pditommaso:
Only if there's active demand

PR #7345, opened 2026-07-16
Allow trace observers to be loaded from the pipeline lib/ directory

adamrtalbot:
## Summary

Adds a `trace.observers` config option that lists observer class names to load from the pipeline `lib/` directory classloader. Each class may implement `TraceObserver`/`TraceObserverV2` (or a matching factory) and is routed to the V1 or V2 observer list by the interface it implements.

Previously, trace observers (e.g. `onFilePublish` hooks) could only be provided by installed pf4j plugins via extension points — the `lib/` classloader was never consulted by the plugin system. This change lets a pipeline ship a lightweight, pipeline-local observer without building and installing a plugin.

## Changes

- `Session.groovy`: `createObserversV1`/`createObserversV2` now also instantiate classes named in `trace.observers`, loaded from the lib classloader via a new memoized `loadLibObservers()`. Unknown classes fail fast with `AbortOperationException`.
- `TraceConfig.groovy`: register the new `trace.observers` config option (validation + docs).

## Example

`nextflow.config`:
```groovy
trace.observers = ['HelloObserver']
```

`lib/HelloObserver.groovy`:
```groovy
import nextflow.trace.TraceObserverV2
import nextflow.trace.event.FilePublishEvent

class HelloObserver implements TraceObserverV2 {
    @Override
    void onFilePublish(FilePublishEvent event) {
        println "[HelloObserver] published -> ${event.target}"
    }
}
```

Running a pipeline that publishes a file prints:
```
[HelloObserver] published -> .../results/hello.txt
```

## Test plan

- [x] `make compile` clean
- [x] Manually verified end-to-end: observer in `lib/` fires `onFilePublish` on a published file
- [ ] Add a Spock unit test for `loadLibObservers` (factory + observer, V1/V2, bad-name abort)


pditommaso (changes requested):
Observer and more in general plugins should not depend on un-versioned files. No go for this 👎

bentsherman:
That thread is talking about using functions from `lib` in config. What does that have to do with this?

PR #7347, opened 2026-07-16
Guard trace log against string formatting errors

### LINKED ISSUE #6600: Unexpected error when running pipeline with trace log level
## Bug report 

An unexpected error is raised when running the nf-core/oncoanalizer with trace log `NXF_TRACE=nextflow`. It is working without setting the NXF_TRACE environment. 

I have created an MRE that produces the error in the pipeline.
main.nf:

```
process TEST {
    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
        'https://depot.galaxyproject.org/singularity/ubuntu:20.04' :
        'quay.io/nf-core/ubuntu:20.04' }"

        input:
                val(id)
        output:
                path('output.txt')
        script:
        """
        echo run $id > output.txt
        """
}

workflow{
        Channel.of(1) | TEST | view
}
```

nextflow.config file:
```
singularity.enabled     = true
singularity.autoMounts  = true

process{
        withName:TEST {
              ext {
                 log_level = { "${params.log_level}" }
              }
              publishDir = [path:{ "${params.outdir}" }, mode:'copy', saveAs:{ filename -> get_saveas_path(meta, task, filename, "${meta.key}/alignments/dna/${filename}") }]
           }
}

def get_saveas_path(meta, task, filename, path, panel_resource_creation=false) {
    if (filename.equals('versions.yml')) {
        return null
    } else if (filename.contains('.command.')) {
        if (filename ==~ /.*\.command\.(sh|out|err|log|run)/) {
            def process_name = task.process.toLowerCase().replaceFirst(/^.+:/, '')
            return "${meta.key}/logs/${meta.id}.${process_name}${filename}"
        } else {
            return null
        }
    } else {
        return path
    }
}
```

### Program output 

In  version 25.04.8, it fails with the following error:

```
ERROR ~ Unexpected error while finalizing task 'TEST (1)' - cause: Cannot get property 'containerEngine' on null object

```

In version 25.10.0, it returns this error:

```
ERROR ~ Unexpected error while finalizing task 'TEST (1)' - cause: No such property: task for class: nextflow.script.dsl.ProcessDslV1
Possible solutions: class

```
It doesn't provide any stack trace in `.nextflow.log`

### Environment 

* Nextflow version: 25.04.9 and 25.10.0
* Java version: 21
* Operating system: Linux

bentsherman:
Fix #6600 

This PR fixes a trace log that can cause the pipeline to fail. It logs every method call invoked by WorkflowBinding, which serializes arbitrary method call arguments that can fail

Rather than trying to patch specific cases, we simply guard the trace log so that it doesn't affect the run

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/script/WorkflowBindingTest.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/script/WorkflowBindingTest.groovy
@@ -16,9 +16,9 @@
 
 package nextflow.script
 
-import spock.lang.Specification
-
+import nextflow.NF
 import nextflow.extension.OpCall
+import spock.lang.Specification
 
 /**
  *
@@ -27,7 +27,7 @@ import nextflow.extension.OpCall
 class WorkflowBindingTest extends Specification {
 
     def setupSpec() {
-        WorkflowBinding.init()
+        NF.init()
     }
 
     def 'should lookup variables' () {
@@ -76,6 +76,21 @@ class WorkflowBindingTest extends Specification {
         1 * binding.getComponent0('foo') >> null
         thrown(MissingMethodException)
 
+    }
+
+    def 'should not fail when tracing a method call' () {
+        given:
+        def BOOM = new Object() {
+            @Override
+            String toString() { throw new MissingPropertyException('task', Object) }
+        }
+        def ARGS = [BOOM] as Object[]
+
+        when:
+        def result = WorkflowBinding.renderArgs(ARGS)
+        then:
+        noExceptionThrown()
+        result == '<unable to render args: MissingPropertyException>'
     }
 
     def 'should get an extension method as variable' () {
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/test/groovy/nextflow/script/WorkflowBindingTest.groovy
```diff
@@ -27,7 +27,7 @@ import nextflow.extension.OpCall
 class WorkflowBindingTest extends Specification {
 
     def setupSpec() {
-        WorkflowBinding.init()
+        NF.init()
```
jorgee, line 30: Is this change really needed for the new test?

jorgee (approved):
Changes are fine, just a confirm the NF.init() change in the test is intentional

modules/nextflow/src/test/groovy/nextflow/script/WorkflowBindingTest.groovy
bentsherman: It was needed to run the test in isolation. There are several test files that only pass when you run the whole test suite, because they depend on some global state which happens to be initialized by a previous test

So I try to fix these issues as I find them so that each test class passes in isolation

PR #7350, opened 2026-07-16
Fix plugin blocks incorrectly merged in v2 config parser

### LINKED ISSUE #6754: Plugins configuration is not merged when passing configs via the CLI

When using the strict syntax and passing configs via the CLI, plugins blocks (at the least) do not get merged like with `includeConfig`. I would think that the behavior should be the same for CLI config inclusion or within the config definition itself.

#### nextflow.config
```groovy
includeConfig 'other.config'

plugins {
    id 'nf-schema'
}
```

#### other.config
```groovy
plugins {
    id 'nf-boost'
}
```

Run `nextflow config` in a few different ways

#### No args

```console
➜  NXF_SYNTAX_PARSER=v2 NXF_VER=25.10.3 nextflow config
plugins = ['nf-boost', 'nf-schema']
➜  NXF_SYNTAX_PARSER=v2 NXF_VER=25.12.0-edge nextflow config
plugins = ['nf-boost', 'nf-schema']
```

#### Passing other.config

```console
➜  NXF_SYNTAX_PARSER=v2 NXF_VER=25.10.3 nextflow -c other.config config
plugins = ['nf-boost']
➜  NXF_SYNTAX_PARSER=v2 NXF_VER=25.12.0-edge nextflow -c other.config config
plugins = ['nf-boost']
```

#### Passing other.config, then nextflow.config

```console
➜  NXF_SYNTAX_PARSER=v2 NXF_VER=25.10.3 nextflow -c other.config -c nextflow.config config
plugins = ['nf-boost', 'nf-schema']
➜  NXF_SYNTAX_PARSER=v2 NXF_VER=25.12.0-edge nextflow -c other.config -c nextflow.config config
plugins = ['nf-boost', 'nf-schema']
```

#### Passing nextflow.config, then other.config

```console
➜  NXF_SYNTAX_PARSER=v2 NXF_VER=25.10.3 nextflow -c nextflow.config -c other.config config
plugins = ['nf-boost']
➜  NXF_SYNTAX_PARSER=v2 NXF_VER=25.12.0-edge nextflow -c nextflow.config -c other.config config
plugins = ['nf-boost']
```

### Environment 

* Nextflow version: 25.10.3, 25.12.0-edge, syntax parser v2
* Java version: openjdk version "22.0.1" 2024-04-16
* Operating system: macOS

### LINKED ISSUE #7214: Remove plugins through configuration
## New feature

A config option or some way to change which plugins are loaded from a lower-level config.

## Use case 

When a plugin is loaded from a higher-level config (e.g. institutional config) this can introduce overhead and possibly lead to crashed pipelines. It would be handy to have an option to remove selected plugins, or options from higher level configs in general. This would also help in pushing new plugins to users, because they still have a way to remove it, even though it is defined at another point.

Currently you could only remove all plugins:
`institution.config`
```config
plugins {
    id 'nf-co2footprint@1.3.0'
}
```
`local.config`
```config
plugins = []
```

## Suggested implementation 
`local.config`
```
plugins {
    exclude 'nf-co2footprint@1.3.0'
}
```

bentsherman:
Fix #6754

The v2 config parser currently merges subsequent `plugins` blocks when it should simply use the last defined block

This is already how the v1 parser behaves and the v2 parser in most scenarios, except the case of a main config including another config and both defining a `plugins` block

Regardless of whether we allow plugins merging in the future, this fix ensures that the current behavior is consistent

### DIFF
--- a/modules/nextflow/src/test/groovy/nextflow/config/parser/v2/ConfigParserV2Test.groovy
+++ b/modules/nextflow/src/test/groovy/nextflow/config/parser/v2/ConfigParserV2Test.groovy
@@ -87,14 +87,6 @@ class ConfigParserV2Test extends Specification {
         def main = folder.resolve('nextflow.config')
         def snippet = folder.resolve('other.config')
 
-        main.text = """
-            includeConfig 'other.config'
-
-            plugins {
-                id 'nf-schema@2.6.1'
-            }
-            """
-
         snippet.text = '''
             plugins {
                 id 'nf-boost'
@@ -103,11 +95,28 @@ class ConfigParserV2Test extends Specification {
             '''
 
         when:
-        def config = new ConfigParserV2().parse(main)
+        main.text = """
+            includeConfig 'other.config'
 
+            plugins {
+                id 'nf-schema@2.6.1'
+            }
+            """
+        def config = new ConfigParserV2().parse(main)
         then:
-        // the last `plugins` block wins
         config.plugins == ['nf-schema@2.6.1'] as Set
+
+        when:
+        main.text = """
+            plugins {
+                id 'nf-schema@2.6.1'
+            }
+
+            includeConfig 'other.config'
+            """
+        config = new ConfigParserV2().parse(main)
+        then:
+        config.plugins == ['nf-boost', 'nf-schema@2.5.1'] as Set
 
         cleanup:
         folder?.deleteDir()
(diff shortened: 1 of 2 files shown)


modules/nextflow/src/test/groovy/nextflow/config/parser/v2/ConfigParserV2Test.groovy
```diff
@@ -81,6 +81,38 @@ class ConfigParserV2Test extends Specification {
         config.process.cpus == 1
     }
 
+    def 'should resolve multiple plugins block by replacement' () {
```
jorgee, line 84: In the issue examples, the reporter showed a different behaviour depending on the order, shouldn't we add a test with the reverse order (`plugins` block before the include)?

bentsherman: Done

jorgee (approved):

PR #7352, opened 2026-07-16
Add `singularity.resourceLimits` config option

### LINKED ISSUE #6797: Add memory and cpu limits to processes launched through Apptainer / Singularity with a local executor
## New feature

Hi,

As discussed in [nf-core's Slack](https://nfcore.slack.com/archives/CE6SDBX2A/p1770032989654369), it would be helpful to be able to limit the resources used by a process launched with Apptainer or Singularity when using the local executor.

## Use case 

Currently, there is a discrepancy between Docker and Apptainer / Singularity engine behaviours when running a pipeline locally (already explained in [Nextflow documentation](https://www.nextflow.io/docs/latest/executor.html#local)). Although docker containers are limited in terms of cpu and memory (using the task's cpu and memory), apptainer / singularity containers can take up all available cpus and memory available on the host machine.

This can also be a problem during CI testing with apptainer, since the executor is local.

## Suggested implementation 

In line with what was proposed by @bentsherman , a solution could be to add flags like:
```
executor.$apptainer.memoryLimits
executor.$apptainer.cpuLimits
```
to pass task.cpus and task.memory values down to the container parameters.

Thanks a lot for all the hard work!

bentsherman:
Close #6797

This PR adds the ability to enforce resource limits on Singularity/Apptainer containers when using the local executor

When `singularity.resourceLimits = true`, Singularity containers are launched with `--cpus` and `--memory`, which prevents them from exceeding their allocated resources. Similar to what Docker does automatically

This feature needs to be opt-in because it requires cgroups in the host environment. Enabling it automatically could break existing code


christopher-hakkaart (commented):
Docs look good

pditommaso:
Naming suggestion: this adds `resourceLimits` (covering both `--cpus` and `--memory`), while Docker uses `docker.cpuLimits` for essentially the same feature. Since `docker.cpuLimits` was only just added in #7289 (2026-07-13) and has not shipped in a stable release yet, it should be safe to rename it to `docker.resourceLimits` here so the option name is unified across all container engines. `resourceLimits` is also the more accurate name since it now governs memory too.

bentsherman:
Well, there is a difference here:

- `docker.cpuLimits` only controls whether to use the *legacy* CPU limits via `--cpus`. Otherwise it always uses `--cpu-shares`. And it always specifies `--memory` limits regardless of this option

- For Singularity and Apptainer, `resourceLimits` controls whether to specify cpu/memory limits at all

So the different names actually reflect the different behavior of each container runtime. Forcing them to be the same for consistency's sake would just blur meaning

Although I agree it would be nice to have consistent naming. I think these settings are largely a historical accident, as each container runtime adds support for proper cgroups-based resource limits over time in different ways

pditommaso (approved):

PR #7353, opened 2026-07-16
Report syntax error for unexpected character in gstring

### LINKED ISSUE #6883: `$/` in scripts shifts the inputs to the right
## Bug report 

Minimal example:

```

process test {
input:
val val1
val val2
val val3
output:
path "*", includeInputs: false, emit: to_publish
script:
"""
echo "This is a test process. Received values: ${val1}, ${val2}, ${val3}" > test_output.txt
echo "$/"
echo "This is a test process. Received values: ${val1}, ${val2}, ${val3}" > test_output2.txt

"""
}

workflow {
main:
test("val1", "val2", "val3")

}
```

Output:
```
cat test_output.txt
This is a test process. Received values: val1, val2, val3
cat test_output2.txt
This is a test process. Received values: val2, val3, 
```

### Environment 

* Nextflow version: 26.01.1-edge

### Additional context

bentsherman:
Fix #6883 

Long ago, I simplified the lexer rules for GStrings to remove a hack inherited from the Groovy grammar. However, that caused certain syntax errors to not be caught, leading to weird runtime behavior

This PR adds the lexer rules necessary to complete the GString handling, so that cases like `"$/"` are reported as invalid syntax rather than leading to silently incorrect behavior

### DIFF
--- a/modules/nf-lang/src/main/antlr/ConfigLexer.g4
+++ b/modules/nf-lang/src/main/antlr/ConfigLexer.g4
@@ -170,6 +170,10 @@ GStringExprStart
     :   '${' -> pushMode(DEFAULT_MODE)
     ;
 
+GStringUnexpectedChar
+    :   . { require(errorIgnored, "Unexpected character: '" + getText().replace("'", "\\'") + "'", -1, false); }
+    ;
+
 mode TDQ_GSTRING_MODE;
 TdqGStringEnd
     :   TdqStringQuotationMark -> popMode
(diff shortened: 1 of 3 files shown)


jorgee (approved):
Looks fine. Just a minor comment not blocking. Despite the issue was in the script, the fixes have been also apply in the `ConfigLexer` but there are no test in `ConfigAstBuilderTest` covering this part, I think for coverage purposes it would be good to add similar test for the config case.

bentsherman:
As a convention I decided not to duplicate these tests initially, and instead just be disciplined about keeping the config parser in sync. One day I hope to find a way to deduplicate this shared logic but I don't know if that will ever happen...

PR #7354, opened 2026-07-17
Fix lint reporting of excluded paths

### LINKED ISSUE #6945: nextflow lint -exclude misses some cases when absolute paths are reported
## Bug report 

I'm using `nextflow lint` to check syntax in my repo. What's strange is when I exclude nf-core it sometimes reports absolute paths which include `nf-core`.

```
✨ Pixi task (lint): nextflow lint -exclude .pixi .                                                                                 
Warn  modules/local/dnadotplot/main.nf:20:9: Variable was declared but not used
│  20 |     def prefix = task.ext.prefix ?: "${meta.id}"
╰     |         ^^^^^^

...

Warn  subworkflows/local/prepare_input/main.nf:167:28: Implicit closure parameter is deprecated, declare an explicit parameter instead
│ 167 |     input.hic_ch.filter { !it.isEmpty() }
╰     |                            ^^

...

Error nf-test.config:3:18: Unexpected input: ':'
│   3 |         "nf-test": "0.9.0"
╰     |                  ^

Nextflow linting complete!
 ❌ 2 files had 2 errors
 ✅ 137 files had no errors

```

In this version it reports only relative paths. But when I add `exclude nf-core`:

```
✨ Pixi task (lint): nextflow lint -exclude .pixi -exclude nf-core .                                                                
Warn  modules/local/dnadotplot/main.nf:20:9: Variable was declared but not used
│  20 |     def prefix = task.ext.prefix ?: "${meta.id}"
╰     |         ^^^^^^

...

Warn  modules/local/functions.nf:84:42: Implicit closure parameter is deprecated, declare an explicit parameter instead
│  84 |             .join( join_args, rhs.map{ [ it[0].subMap(args.keySet) ] +
╰     |                                          ^^

...

Warn  /Users/mahpa906/Documents/Projects/Earth-Biogenome-Project-pilot/modules/nf-core/tabix/tabix/main.nf:33:9: Variable was declared but not used
│  33 |     def prefix = task.ext.prefix ?: "${meta.id}"
╰     |         ^^^^^^

Warn  /Users/mahpa906/Documents/Projects/Earth-Biogenome-Project-pilot/modules/nf-core/tidk/plot/main.nf:37:9: Variable was declared but not used
│  37 |     def args = task.ext.args ?: ''
╰     |         ^^^^

...

Error nf-test.config:3:18: Unexpected input: ':'
│   3 |         "nf-test": "0.9.0"
╰     |                  ^

Nextflow linting complete!
 ❌ 2 files had 2 errors
 ✅ 106 files had no errors

```

The paths are now a mix of relative and absolute, and still includes some with `nf-core` in them.

```
git clone https://github.com/NBISweden/Earth-Biogenome-Project-pilot.git
cd Earth-Biogenome-Project-pilot
nextflow run lint -exclude .pixi -exclude nf-core .
```

### Program output 

It's included above to show actual output.

### Environment 

* Nextflow version: 25.10.4
* Java version: openjdk 17.0.11 2024-04-16 LTS
* Operating system: macOS

### Additional context

~~I wonder if this has anything to do with the `-` in `nf-core`.~~ I tested by replacing `nf-core` with `local` and it still printed out absolute paths with local in them.

bentsherman:
Fix #6945 

The `lint` command `-exclude` option did not account for the fact that the linting process can pick up additional files via includes. These included files were not filtered against the exclude patterns. Additionally, they were reported by absolute path instead of relative path

This PR fixes both issues. Source files are filtered again after linting and prior to reporting, and they are reported by relative path

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdLint.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdLint.groovy
@@ -42,6 +42,7 @@ import nextflow.script.parser.v2.ErrorSummary
 import nextflow.script.parser.v2.StandardErrorListener
 import nextflow.util.ClassLoaderFactory
 import nextflow.util.PathUtils
+import nextflow.util.TestOnly
 import org.codehaus.groovy.control.SourceUnit
 import org.codehaus.groovy.control.messages.SyntaxErrorMessage
 import org.codehaus.groovy.control.messages.WarningMessage
@@ -120,6 +121,9 @@ class CmdLint extends CmdBase {
 
     private ErrorSummary summary = new ErrorSummary()
 
+    @TestOnly
+    protected Path root
+
     @Override
     String getName() { 'lint' }
 
@@ -156,8 +160,9 @@ class CmdLint extends CmdBase {
         final List<File> files = []
 
         for( final input : inputs ) {
+            final start = root != null ? root.resolve(input) : Path.of(input)
             PathUtils.visitFiles(
-                Path.of(input),
+                start,
                 (path) -> !PathUtils.isExcluded(path, excludePatterns),
                 (path) -> files.add(path.toFile()))
         }
@@ -229,6 +234,7 @@ class CmdLint extends CmdBase {
         compiler.getSources()
             .values()
             .stream()
+            .filter((source) -> !isExcludedSource(source))
             .sorted(Comparator.comparing((SourceUnit source) -> source.getSource().getURI()))
             .forEach((source) -> {
                 final errorCollector = source.getErrorCollector()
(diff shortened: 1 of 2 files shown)


jorgee (changes requested):
It seems there is a regression with PR changes. I looks like the exclude pattern is evaluated on the absolute path. So having `work `or other default exclude pattern in the project path is incorrectly excluding files. Find below a small example about this problem.  

I have created a `main.nf` with an error and the file is excluded becasue the parent of the project is "work/"
```
/tmp/tmp.9QR9INj4dy/work/myproject$  nextflow lint .
Nextflow linting complete!
 No files found to process
```
This what claude finds:

 **isExcludedSource** walks the absolute path past the project root and over-matches
  
` CmdLint.groovy:265-273` — the ancestor walk starts from the absolute source URI and climbs all the way to the filesystem root:

```
  for( Path path = Path.of(uri); path != null; path = path.getParent() ) {
      if( PathUtils.isExcluded(path, excludePatterns) )
          return true
  }
```
  
`PathUtils.isExcluded` matches when the path string `endsWith("/" + pattern)`. Because the walk goes above the linted project into the absolute prefix, any directory component anywhere in the absolute path that coincides with an exclude pattern excludes the file — including directories the user never intended to lint against.

The default exclude list contains 'work', and work is an extremely common ancestor-directory name.

jorgee:
I have open #7368 with the fix of the issue I found in my review.

jorgee (approved):
Just fixed DCO in the merge commit.

PR #7355, opened 2026-07-17
Fix spurious 'Unrecognized config option' warning for wave.build.compression and wave.build.conda

pditommaso:
## Problem

Setting `wave.build.compression.mode` (or any `wave.build.conda.*` key) produced a spurious warning:

```
WARN: Unrecognized config option 'wave.build.compression.mode'
```

## Cause

The `compression` and `conda` fields in `BuildOpts` were neither annotated with `@ConfigOption` nor backed by a `ConfigScope` (both are wave-api Java classes). They were therefore absent from the config spec entirely, so the validator flagged their nested keys as unrecognized.

## Fix

Annotate both fields as map options (`@ConfigOption(types=[Map])`), matching how their docstrings already describe them. The validator then recognizes `wave.build.compression` and `wave.build.conda` as map-valued options and skips validating their subkeys.

## Test

Added a `WaveConfigTest` case asserting the config spec exposes both `build.compression` and `build.conda` as map options. Full `WaveConfigTest` passes (46 tests, 0 failures).


bentsherman:
The problem is that there is a difference between a map-type config option and a config scope:

```groovy
// map-type option (incorrect)
wave.build.compression = [
  mambaImage: '...',
  basePackages: '...'
]

// scope (correct)
wave.build.compression.mambaImage = '...'
wave.build.compression.basePackages = '...'
```

It doesn't make a difference for runtime validation but it does matter for type checking. The type checker will tell users to use the wrong syntax.

I think we need a way to mark a field as a nested scope without using a `ConfigScope` type. Like a `@NestedScope` marker class

pditommaso:
Since these are deps into other project, likely we need a tiny shared lib for these annotations. do you agree?

bentsherman:
Let me try the `@NestedScope` idea first, if it works it would be much simpler

docs/developer/config-scopes.mdx
```diff
@@ -69,6 +69,26 @@ The scope class referenced by the field type defines config options and scopes i
 
 See `ExecutorConfig` and `ExecutorRetryConfig` for an example of how a nested scope is defined and constructed.
 
+### Nested scopes from external types
+
+Sometimes a nested scope is backed by a type that cannot implement `ConfigScope`, such as a class from an external library. In this case, the field can be annotated with `@NestedScope` to mark it as a nested scope without requiring its type to implement `ConfigScope`.
```
christopher-hakkaart:
```suggestion
Some nested scopes are backed by a type that cannot implement `ConfigScope`, such as a class from an external library. Annotate the field with `@NestedScope` to mark it as a nested scope without requiring its type to implement `ConfigScope`.
```

docs/developer/config-scopes.mdx
```diff
@@ -69,6 +69,26 @@ The scope class referenced by the field type defines config options and scopes i
...
+
+Each non-static instance field of the annotated field's type becomes a config option of the resulting scope. The field name determines the name of the nested scope, and the description (if any) is taken from the `@Description` annotation on the field.
```
christopher-hakkaart:
```suggestion
Each non-static instance field of the annotated field's type becomes a config option of the resulting scope. The field name determines the name of the nested scope, and the description (if any) comes from the `@Description` annotation on the field.
```

docs/developer/config-scopes.mdx
```diff
@@ -69,6 +69,26 @@ The scope class referenced by the field type defines config options and scopes i
...
+
+For example, in the `wave` scope the `build.conda` and `build.compression` options are backed by wave-api types (`CondaOpts` and `BuildCompression`) that live in an external library:
```
christopher-hakkaart:
```suggestion
For example, in the `wave` scope, the `build.conda` and `build.compression` options are backed by wave-api types (`CondaOpts` and `BuildCompression`) defined in an external library:
```

docs/developer/config-scopes.mdx
```diff
@@ -69,6 +69,26 @@ The scope class referenced by the field type defines config options and scopes i
...
+    @NestedScope
+    @Description('''
+        The `wave.build.compression` scope controls build image compression, e.g. `wave.build.compression.mode = 'estargz'`.
+    ''')
+    final BuildCompression compression
+```
+
+This exposes `wave.build.compression.mode`, `wave.build.compression.level`, and `wave.build.compression.force` as recognized config options.
+
+Because the config options are derived from the external type's fields, inner field descriptions and inner nested scopes are not supported with `@NestedScope`. Prefer a dedicated `ConfigScope` class where possible.
```
christopher-hakkaart:
```suggestion
This exposes `wave.build.compression.mode`, `wave.build.compression.level`, and `wave.build.compression.force` as recognized config options.

Because config options are derived from the external type's fields, `@NestedScope` does not support inner field descriptions or inner nested scopes. Prefer a dedicated `ConfigScope` class where possible.
```

christopher-hakkaart (commented):
Suggestions to tighten docs

bentsherman (approved):
I tried my NestedScope idea, and then I tried a third idea -- mirror the external Wave types as ConfigScope classes

I committed both so that you can see, but I think I like the current state. The NestedScope idea is still a bit hacky; it doesn't support things like inner field descriptions and inner nested scopes

Mirroring the classes with ConfigScope seems like the idiomatic way to handle this. The overall WaveConfig interface is unchanged -- it still returns the wave-api types

I also added a unit test to ensure that these mirrored classes stay aligned with the external types

bentsherman:
(sorry Chris, didn't mean to trigger your review here)

christopher-hakkaart:
> (sorry Chris, didn't mean to trigger your review here)

No worries 😅

PR #7357, opened 2026-07-17
Add note about AI co-authorship to CLAUDE.md [ci fast]

bentsherman:
This PR adds a note to CLAUDE.md specifying that agents should not co-author commits. Only humans should be authors since only they are accountable for committed code


pditommaso:
Not sure about this, implicitly the person submitting the pr the owner and responsible for the code with or without agent Co-authorship tag, however I think it's useful as marker and for transparency purposes to better understand the source of the code.

bentsherman:
Groovy's [AGENTS.md](https://github.com/apache/groovy/blob/master/AGENTS.md) handles that trade-off like so:

> - **Attribute AI assistance in commits.** When AI tooling assisted on
>   a change, consider adding an `Assisted-by:` trailer naming the
>   tool(s) — for example:
> 
>   ```
>   Assisted-by: <tool name and version>
>   ```
> 
>   `Assisted-by:` is the default and reflects the ASF's stance that a
>   human contributor performs the final check on every change.
>   `Co-authored-by:` is conventionally used for human co-authors.

Would that be acceptable?

pditommaso:
It may be fine, I have the suspect that Claude do not care much about that and it will use anyway its own signature

bentsherman:
It has always remembered to add the DCO sign-off for me so I expect it will follow this guidance as well. Even if it forgets 1% of the time I think it's worth having

Updated, ready to review

pditommaso (approved):

PR #7364, opened 2026-07-20
Fix publish mode override when re-publishing an existing file

### LINKED ISSUE #7303: Different output mode for different published outputs doesn't work
So I am publishing multiple processes outputs in the output block in the main wf. In `nextflow.config`, I specify a general output mode, then for some outputs I wanna change that, in my case I wanna change from relink to copy. This doesn't work, what happens is all outputs gets relinked

I think this should be easily reproduced without a script, I have this in my main wf
```groovy
output {
    merged_reads {
        path { meta, _reads -> "merged_reads/${meta.donor}" }
    }
    qc_html {
        path { meta, _html -> "fastqc/html/${meta.donor}" }
        mode 'copy'
    }
    multiqc_report {
        path "multiqc/report"
        mode 'copy'
    }
    trimmed_reads {
        path { meta, _reads -> "trimmed_reads/${meta.donor}" }
    }
    qc_html_trimmed {
        path { meta, _html -> "fastqc/html_trimmed/${meta.donor}" }
        mode 'copy'
    }
}

```

and this in my `nextflow.config`:
```groovy
includeConfig 'conf/params.config'
includeConfig 'conf/singularity.config'
includeConfig 'conf/process.config'

outputDir = "${baseDir}/NorwagianWoods"
workflow.output.mode = 'rellink'

```

### Program output 
 Didn't fail, the but it didn't copy the outputs

### Environment 

* Nextflow version: [version 26.04.4 build 12445] 
* Java version: [openjdk 23.0.2-internal 2025-01-21]
* Operating system: [Linux]
* Bash version: [5.0.17(1)-release (x86_64-pc-linux-gnu)

bentsherman:
Fix #7303

When a workflow output is re-published to a destination that already exists (e.g. on a -resume run after changing the output mode), PublishDir decided whether to overwrite the existing file using a content-only comparison. A symbolic link resolves to the same content as its source, so the check always concluded the file was identical and skipped it, leaving the stale link in place even though the publish mode had changed from a link mode to copy/move (or vice versa).

Add checkPublishModeMismatch() so an existing destination whose type (symlink vs real file) no longer matches the requested mode is always overwritten, regardless of content equality. The check is limited to the default filesystem, so cloud/Fusion publishing is unaffected.


modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
```diff
@@ -489,6 +489,11 @@ class PublishDir {
     }
 
     protected boolean shouldOverwrite(Path source, Path target) {
+        // always overwrite when the existing file was published with a different mode
+        // (e.g. a symlink but now `copy`, or vice versa), even if its content matches
+        if( checkPublishModeMismatch(target) )
```
jorgee: This is checked before the `overwrite instanceof Boolean` block (PublishDir.groovy:492). A user who explicitly set `overwrite` to  `false` will now have their published file overwritten anyway whenever the mode type changed. Is it the intended behavior after this change? If not, I suggest moving the `checkPublishModeMismatch(target)` block after the `overwrite instanceof Boolean` part. With this change, it will behave as follows:
 - Unset (null) — the main bug case — still falls through to the mismatch check, then hash comparison.
 - overwrite: true — returns true immediately (unchanged, no need for the mismatch check anyway).
 - overwrite: false — now returns false and honors the user's explicit "never overwrite", instead of the mismatch check silently overriding it.

jorgee (commented):
Found an edge case when user set override: false.

modules/nextflow/src/main/groovy/nextflow/processor/PublishDir.groovy
bentsherman: Fixed

jorgee (approved):

PR #7367, opened 2026-07-21
Add seqera.executor.maxCpusPerUser option

jonmarti:
## Summary

Adds a `seqera.executor.maxCpusPerUser` configuration option so the Seqera scheduler can enforce a per-user vCPU cap. `CreateRunRequest.schedulingRequirement` exists in the sched API and `SchedClient.createRun()` forwards it — the executor just didn't populate it and there was no config surface for it.

## Changes

- `ExecutorOpts`: new `Integer maxCpusPerUser` option.
- `SeqeraExecutor.createRun()`: when set, builds `schedulingRequirement.maxCpusPerUser` on the request; when unset, sends no `schedulingRequirement`.
- Bumps `sched-client` `0.71.0` → `0.73.2` to pick up `CreateRunRequest.schedulingRequirement`.
- Tests: `ExecutorOptsTest` (parse + default-null), `SeqeraExecutorTest` (set → request field; unset → null).
- Docs: `seqera.executor.maxCpusPerUser` in the config reference.

Run-level only by design: the cap applies per user across the compute environment, so there is no per-process `hints` override (unlike `predictionModel`).

Companion to the scheduler-side enforcement (`UserVcpuCapacityEnforcer`), which reads `schedulingRequirement.maxCpusPerUser` and defers a user's over-budget tasks to `PENDING`.


docs/reference/config.mdx
```diff
@@ -1802,6 +1802,10 @@ The Seqera scheduler service endpoint URL (required).
 
 The EC2 key pair name for SSH access to instances.
 
+###### `seqera.executor.maxCpusPerUser`
+
+The maximum number of concurrent vCPUs a single user may consume across all of their runs on this compute environment. The limit is applied independently to each user sharing the compute environment. When not set, no CPU limit is enforced (unlimited).
```
```suggestion
The maximum number of concurrent vCPUs a single user may consume across all of their runs on this compute environment. The limit is applied independently to each user sharing the compute environment. When not set, no CPU limit is enforced.
```
To match `seqera.executor.predictionModel`

christopher-hakkaart (commented):
Minor suggestion and small fix near this change.

docs/reference/config.mdx
christopher-hakkaart, line 1785:
```suggestion
```

christopher-hakkaart, line 1786:
```suggestion
```

jonmarti: resolved: https://github.com/nextflow-io/nextflow/commit/4981011bf9f0f30c0e3db77fa9cb86388fbc12b2

jonmarti: resolved: https://github.com/nextflow-io/nextflow/pull/7367/commits/4981011bf9f0f30c0e3db77fa9cb86388fbc12b2

pditommaso:
## Review

Overall clean and well-tested, and it follows the surrounding patterns. Two things to fix, one structural change, one question.

### 1. Structure: this should be a nested `schedulingRequirement` scope, not a flat option

The plugin already has the pattern for exactly this: `machineRequirement` is a nested `ConfigScope` (`MachineRequirementOpts`) mapped to the API's `MachineRequirement` by `SchemaMapperUtil.toMachineRequirement`. The new option maps to `CreateRunRequest.schedulingRequirement.maxCpusPerUser` — same shape — so it should mirror the same structure instead of flattening one nested API field into a top-level config key.

```groovy
// io/seqera/config/SchedulingRequirementOpts.groovy
@CompileStatic
class SchedulingRequirementOpts implements ConfigScope {

    @ConfigOption
    @Description("""...""")
    final Integer maxCpusPerUser

    /* required by config scope -- do not remove */
    SchedulingRequirementOpts() {}

    SchedulingRequirementOpts(Map opts) {
        this.maxCpusPerUser = opts.maxCpusPerUser as Integer
        if( maxCpusPerUser != null && maxCpusPerUser <= 0 )
            throw new IllegalArgumentException("Invalid 'seqera.executor.schedulingRequirement.maxCpusPerUser' value '${maxCpusPerUser}' - expected a positive integer")
    }

    Integer getMaxCpusPerUser() { maxCpusPerUser }
}
```

```groovy
// ExecutorOpts
final SchedulingRequirementOpts schedulingRequirement
...
this.schedulingRequirement = new SchedulingRequirementOpts(opts.schedulingRequirement as Map ?: Map.of())
```

```groovy
// SchemaMapperUtil — mirrors toMachineRequirement, returns null when no settings
static SchedulingRequirement toSchedulingRequirement(SchedulingRequirementOpts opts) {
    if( !opts || opts.maxCpusPerUser == null )
        return null
    new SchedulingRequirement().maxCpusPerUser(opts.maxCpusPerUser)
}
```

```groovy
// SeqeraExecutor.createRun() — back to a one-liner, no inline ternary
.schedulingRequirement(SchemaMapperUtil.toSchedulingRequirement(seqeraConfig.schedulingRequirement))
```

Why:

- **Config mirrors the API**, same as `machineRequirement` → `MachineRequirement`. Someone reading `CreateRunRequest` can find the config key and vice versa.
- **The next scheduling knob has a home.** `SchedulingRequirement` is a generated schema class that will grow (queue priority, max tasks per user, …). Otherwise each one becomes another top-level `seqera.executor.*` key with no hint that they are one API object.
- **Mapping leaves the executor.** The `!= null ? … : null` ternary in `createRun()` becomes a mapper call, unit-testable in isolation like the machine-requirement mappers already are, and the "omit when unset" behaviour comes free from the mapper's null return.
- **Validation gets an obvious home** — the `<= 0` check from point 2 lives in the new scope class, next to the field.

No back-compat cost: the option is new in this PR, so there is no flat form to keep accepting. Doing it now is free; after a release it means either a deprecated alias or a breaking rename. Docs then become `seqera.executor.schedulingRequirement.maxCpusPerUser`, slotting in after the `machineRequirement.*` block.

### 2. `0` silently means "unlimited" (correctness)

`ExecutorOpts.groovy`:

```groovy
this.maxCpusPerUser = opts.maxCpusPerUser as Integer ?: null
```

Verified in Groovy:

| config value | resulting field |
|---|---|
| `0` | **`null` → no cap sent → unlimited** |
| `-4` | `-4` → sent to the server → HTTP 400 at run creation |
| `'16'` | `16` ✓ |

The `?: null` idiom is copied from the `as String ?: null` lines above, where it is harmless. On a number it changes semantics: a misconfigured or computed-to-zero value **disables the cap instead of failing**. Meanwhile the scheduler (`RunServiceImpl.validateSchedulingRequirement`) explicitly rejects `<= 0` with *"must be greater than zero"* — so the client swallows exactly the input the server was careful to reject, and forwards the one it will reject remotely.

`ExecutorOpts` already has the convention for this (see `endpoint` and `autoLabels`): throw `IllegalArgumentException` with the fully-qualified config path. See the snippet in point 1. Please add test cases for `0` and for a negative value.

### 3. Docs omit the user-visible failure mode

The config reference describes the cap but not what happens when it bites. From the scheduler side:

- Over-budget tasks stay **PENDING** and retry on a later cycle (fine, invisible).
- A task whose own request exceeds the cap is **failed outright** — `REJECT_OVERSIZED` → `updateTaskFailed` with `"Task requires N vCPUs but the per-user maximum … is M"`.

So `maxCpusPerUser = 16` with a `cpus 32` process is a hard pipeline failure, not throttling. That belongs in the docs, ideally with the pairing hint: keep the cap ≥ the largest per-task `cpus`, or clamp with `process.resourceLimits.cpus`. Worth a sentence in the `@Description` too, since that is what the config LSP surfaces.

### 4. Question: client-set cap vs. CE policy

The sched API docs for this field say: *"The value is configured on the compute environment and carried on each run."* This PR makes the pipeline's own `nextflow.config` the source, i.e. **the user being capped sets their own cap** — they can omit it or raise it freely. As a self-throttle ("don't let my run hog the CE") that is coherent; as the enforcement mechanism the description implies ("so the Seqera scheduler can enforce a per-user vCPU cap") it is not, since it is trivially bypassed.

Not a blocker — just confirm it is deliberate, and if Platform is later going to populate `schedulingRequirement` from the CE record, decide now which side wins (today `RunServiceImpl` reads only the request, so a client value would silently override an admin policy). The docs could say plainly that this is a per-run setting applied to the submitting user, not an admin-enforced quota.

### 5. Dependency bump — verify the transitive graph

`0.71.0` → `0.73.2` spans two minor versions, so unrelated API changes ride along. The comment three lines below the bump is a standing warning that `sched-client` drags in an older `lib-httpx`/`failsafe` and that the `api` pin exists to prevent a runtime `NoSuchMethodError`. That pin is version-sensitive, so please confirm the resolved versions did not shift:

```
./gradlew :plugins:nf-seqera:dependencies --configuration runtimeClasspath
```

Runtime classpath breakage here would not show up in unit tests.

### Nits (no action needed)

- `getMaxCpusPerUser()` is redundant — Groovy generates it for a `final` property — but every field in the file has one, so consistency wins.
- `Integer` over `int` is the right call: null is the meaningful "unset".
- No per-process `hints` override is correct — the scheduler denormalizes the cap from the run onto the task spec.
- Removing the duplicated `The following settings are available:` line is a good drive-by fix.
- Tests match the file's conventions (`SysEnv.push` + spec-level `cleanup()`, `executor.@field` injection) and cover both the set and unset paths.

---

Verdict: looks good once points 1–3 are addressed. 4 is a design question for you, 5 is a quick check.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

jonmarti:
**Re point 4 (client-set cap vs CE policy):** Deliberate, and yes it's a per-run self-throttle as written — the pipeline's own config sets it, so a user can omit or raise it. That matches this PR's scope: give a run a way to say "don't let me hog the CE."

The admin-enforced direction is exactly where this is heading on the Platform side. The corresponding Platform PR (held as draft) adds `maxCpusPerUser` to the compute-environment `SchedConfig` and **emits it into the generated `nextflow.config`** for the CE — so for Platform-launched runs the CE's value is the source, not something the user hand-writes. That's why nesting it under `schedulingRequirement` now matters: Platform already emits `seqera.executor.machineRequirement.*`, and this slots in beside it.

The open question you're really pointing at — *what wins when both a client value and a CE policy exist* — is live: today `RunServiceImpl` reads only the request, so a client value would override an admin policy. Since Platform is the one populating it from the CE record, the clean answer is for the enforcement side to treat the CE value as authoritative (client can only lower, not raise). I'll capture that as a follow-up on the Scheduler side rather than fold it into this PR. I'll also add a sentence to the docs making explicit that, outside Platform-managed CEs, this is a per-run setting applied to the submitting user — not an admin quota.

jonmarti:
@pditommaso thanks for the thorough review. Follow-up on the rest of the points (commits `d95b7e95d`, `f5dd9e3e3`):

**1. Nested `schedulingRequirement` scope** — done. New `SchedulingRequirementOpts` `ConfigScope` mirroring `MachineRequirementOpts`; `SchemaMapperUtil.toSchedulingRequirement` mirrors `toMachineRequirement` (null when unset); `createRun()` is back to a one-liner. Config key is now `seqera.executor.schedulingRequirement.maxCpusPerUser`.

**2. `0` / negative** — fixed. The `as Integer ?: null` idiom is gone; the new scope's constructor throws `IllegalArgumentException` on a non-positive value with the fully-qualified config path, matching the `endpoint`/`autoLabels` convention. Added `0` and negative test cases.

**3. Docs failure mode** — added. The reference now notes that a task whose own `cpus` exceeds the cap fails outright (not throttled), with the pairing hint (keep the cap ≥ largest per-task `cpus`, or clamp with `process.resourceLimits.cpus`), plus the aggregate-queueing behaviour. Mirrored a short version into the `@Description`.

**4. Client-set cap vs CE policy** — replied separately: https://github.com/nextflow-io/nextflow/pull/7367#issuecomment-5106904350

**5. Dependency graph** — I'll confirm `:plugins:nf-seqera:dependencies --configuration runtimeClasspath` shows `lib-httpx` still at `2.4.0` / `failsafe` at `3.3.2` (the `api`-pin's purpose) before this merges, and report back here.

**Nits** — left the `getMaxCpusPerUser()`-style getters for file consistency as you noted; `Integer`-over-`int` kept.

jonmarti:
@pditommaso **Point 5 confirmed.** `:plugins:nf-seqera:dependencies --configuration runtimeClasspath` after the `sched-client 0.71 -> 0.73.2` bump:

```
+--- io.seqera:lib-httpx:2.2.0 -> 2.4.0
|         \--- dev.failsafe:failsafe:3.3.2
+--- io.seqera:lib-httpx:2.4.0 (*)
```

`sched-client` still pulls `lib-httpx 2.2.0` transitively, the `api` pin overrides it to `2.4.0`, and `failsafe` resolves to `3.3.2` (not `3.1.0`). The pin is doing its job — no runtime `NoSuchMethodError` risk. Graph unchanged by the bump.

pditommaso (approved):
Validated — all five points hold.

1–2 match the surrounding patterns (`SchedulingRequirementOpts` mirrors `MachineRequirementOpts`, non-positive values throw with the fully-qualified path). 3 is factually correct against the scheduler: `REJECT_OVERSIZED` → `updateTaskFailed`, `BLOCK` keeps the task `PENDING`. 5 re-checked here: `lib-httpx 2.2.0 -> 2.4.0`, `failsafe 3.3.2` — the pin is intact. `:plugins:nf-seqera:test` green (108 tests).

Pushed 727d71d3 to trim the docs: the reference entry had grown to three paragraphs, and the last one promised that a Platform-launched run uses the CE value, which nothing implements yet. Now one sentence for the cap plus the fail-not-throttle warning, matching the density of the neighbouring entries. `@Description` mirrored.

On point 4: agreed that CE-authoritative (client can only lower) is the right end state — happy for that to be a scheduler-side follow-up.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

jonmarti:
@pditommaso @bentsherman — question on the CI state here.

**`test_aws` and `test_wave` are timing out — is this expected on PR builds?**

The `Nextflow CI` run on this branch shows `test (17/25, test_aws)` and `test (17/25, test_wave)` as **cancelled**, while every other job (both `Build`s, `test_integration`, `test_parser_v2`, `test_docs`, `test_azure`, `test_google`) passes.

From what I can tell it's a timeout, not a code failure:
- All four jobs ran ~90 min (13:09 → 14:40 UTC) then terminated — a clean hit on `timeout-minutes: 90`, no failing test / stack trace.
- `test_aws`/`test_wave` run the real-cloud E2E suites (`validation/awsbatch.sh`, `validation/wave.sh`), which need AWS/Tower credentials.
- This branch's changes are confined to the `nf-seqera` plugin config (`seqera.executor.schedulingRequirement.maxCpusPerUser`); `awsbatch.sh`/`wave.sh` don't reference any of it, and the `nf-seqera` unit tests pass in the `Build` job.
- The latest **master** `Nextflow CI` run cancels the same two jobs the same way, so it doesn't appear specific to this PR.

Questions:
1. Is `test_aws`/`test_wave` timing out a known state on PR / fork builds (e.g. missing or throttled cloud credentials), and is it safe to disregard for merge?
2. If they're expected to pass, is there anything on my side needed to run them, or would a maintainer re-run help?

Happy to rebase or adjust if there's something I'm missing.

pditommaso:
Expired keys, trying again

jonmarti:
@pditommaso I'm not authorized to merge it. Can you do it?

PR #7369, opened 2026-07-21
Accept http requesting-protocol for https proxy CONNECT tunnel auth

### LINKED ISSUE #5634: NF does not perform proxy authentification and cannot access github

bentsherman:
Fix #5634

For Basic proxy authentication the JDK hard-codes "http" as the
requesting protocol regardless of the destination scheme,
which passes the string literal "http". ProxyConfig.authenticator()'s guard therefore never matched for an https_proxy-derived config authenticating an HTTPS CONNECT tunnel (protocol="https" != requesting "http"), silently discarding valid proxy credentials and causing a 407.

Relax the guard to also accept an http requesting-protocol when the
configured protocol is https (the CONNECT-tunnel case), keeping the
host/port/requestor-type checks intact. Adds a regression test driving
the real authenticator with the exact field values the JDK supplies for
an HTTPS tunnel.

This complements #7331 (which cleared `jdk.http.auth.tunneling.disabledSchemes` so the authenticator is consulted at all); without this fix the consulted authenticator still rejected the credentials.

### DIFF
--- a/modules/nf-commons/src/test/nextflow/util/ProxyConfigTest.groovy
+++ b/modules/nf-commons/src/test/nextflow/util/ProxyConfigTest.groovy
@@ -102,15 +102,6 @@ class ProxyConfigTest extends Specification {
         invokeAuth(authenticator, Authenticator.RequestorType.PROXY, 'proxy.example.com', 8080, 'https') == null
     }
 
-    def 'authenticator is null-safe when the requesting protocol is missing'() {
-        given: 'an https_proxy-derived config'
-        def proxy = new ProxyConfig(protocol: 'https', host: 'proxy.example.com', port: '8080', username: 'foo', password: 'bar')
-        def authenticator = proxy.authenticator()
-
-        expect: 'a null requesting protocol does not throw and yields no credentials'
-        invokeAuth(authenticator, Authenticator.RequestorType.PROXY, 'proxy.example.com', 8080, null) == null
-    }
-
     def 'proxyConfig should return null when no proxy is registered'() {
         expect:
         ProxyConfig.proxyConfig() == null
--- a/modules/nf-commons/src/test/nextflow/util/ProxyConfigTest.groovy
+++ b/modules/nf-commons/src/test/nextflow/util/ProxyConfigTest.groovy
@@ -79,6 +79,20 @@ class ProxyConfigTest extends Specification {
         invokeAuth(authenticator, Authenticator.RequestorType.PROXY, 'other.example.com', 8080, 'https') == null
     }
 
+    def 'authenticator should release credentials for an https proxy over an http CONNECT tunnel'() {
+        given: 'an https_proxy-derived config (Launcher sets protocol=https from HTTPS_PROXY)'
+        def proxy = new ProxyConfig(protocol: 'https', host: 'proxy.example.com', port: '8080', username: 'foo', password: 'bar')
+        def authenticator = proxy.authenticator()
+
+        when: 'the JDK authenticates an HTTPS CONNECT tunnel — it reports requestingProtocol="http", not "https" (see #5634)'
+        def auth = invokeAuth(authenticator, Authenticator.RequestorType.PROXY, 'proxy.example.com', 8080, 'http')
+
+        then: 'credentials must still be released, otherwise HTTPS targets 407 behind an authenticating proxy'
+        auth != null
+        auth.userName == 'foo'
+        new String(auth.password) == 'bar'
+    }
+
     def 'proxyConfig should return null when no proxy is registered'() {
         expect:
         ProxyConfig.proxyConfig() == null
--- a/modules/nf-commons/src/test/nextflow/util/ProxyConfigTest.groovy
+++ b/modules/nf-commons/src/test/nextflow/util/ProxyConfigTest.groovy
@@ -93,6 +93,24 @@ class ProxyConfigTest extends Specification {
         new String(auth.password) == 'bar'
     }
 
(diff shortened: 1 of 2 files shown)


bentsherman:
The JDK hard-coding was confirmed against JDK 17 / 21 / 25 source code, so it appears to be a stable behavior

pditommaso:
## Review — Accept http requesting-protocol for https proxy CONNECT tunnel auth

**Overview.** The diagnosis is accurate and the fix is minimal and well-scoped. When a proxy comes from `HTTPS_PROXY`, `ProxyConfig` has `protocol="https"`, but for Basic proxy auth the JDK hard-codes the requesting protocol as `"http"` when authenticating an HTTPS CONNECT tunnel, so the old guard `protocol.equalsIgnoreCase(getRequestingProtocol())` never matched and credentials were silently dropped (407). Relaxing that single guard for the `https`-config/`http`-request case, while keeping host/port/requestor-type checks intact, is the right approach. Good complement to #7331.

### Correctness — one real concern (latent NPE)

The relaxation flips a null-safety property of the original line:

- **Old:** `protocol.equalsIgnoreCase(getRequestingProtocol())` — receiver is the non-null `protocol` field, so a `null` requesting protocol is null-safe (returns `false`).
- **New:** `getRequestingProtocol().equalsIgnoreCase('http')` — receiver is now the JDK-supplied value. If `getRequestingProtocol()` is ever `null`, this **NPEs** instead of falling through to `return null`.

`java.net.Authenticator.getRequestingProtocol()` is not contractually non-null. For Basic proxy auth it's practically always set, so this is low-probability, but it's a genuine robustness regression for a one-character cost. Suggest flipping the receiver:

```groovy
final isHttpsTunnel = protocol.equalsIgnoreCase('https') && 'http'.equalsIgnoreCase(getRequestingProtocol())
```

Everything else is sound:
- The relaxation is **one-directional** (`https` config accepts `http` request only) — an `http` config still rejects an `https` request.
- Host/port checks are untouched, so credentials are never released to a non-matching proxy endpoint.

### Code quality & conventions

- The comment clearly explains the non-obvious JDK behavior — appropriate, since this is exactly the kind of thing a future reader might "fix" back.
- Style matches surrounding code; the narrow inline guard is preferable to extracting a helper for one use.

### Test coverage

- The regression test is precise: it drives the real authenticator with `requestingProtocol="http"` and asserts both username and password. It fails without the fix.
- **Minor gap:** no explicit test asserts the reverse stays rejected (config `protocol=http`, requesting `https` -> `null`). Cheap to add and locks in the one-directional intent.

### Security

- No credential-leak risk introduced: host + port + requestor-type gating is preserved; only the http/https protocol string is relaxed, and only for the matching proxy host.

### Verdict

Approve with one small change: guard against a `null` requesting protocol by flipping the `equalsIgnoreCase` receiver (`'http'.equalsIgnoreCase(...)`). Optionally add the reverse-direction negative test.

_🤖 Generated with [Claude Code](https://claude.com/claude-code)_

bentsherman:
Thanks for the review! Addressed both points in b0e8402:

- **Latent NPE:** flipped the receiver to `'http'.equalsIgnoreCase(getRequestingProtocol())` so a null requesting protocol falls through to `return null` instead of throwing, restoring the null-safety the original single-expression guard had.
- **Test coverage:** added the reverse-direction negative test (http proxy config still rejects an https request) to lock in the one-directional relaxation. Also added a null requesting-protocol test — I verified it fails with an NPE at the exact call site without the receiver flip, so it guards against regressing the fix above.

All `ProxyConfigTest` tests green (10/10).

modules/nf-commons/src/test/nextflow/util/ProxyConfigTest.groovy
```diff
@@ -79,6 +79,38 @@ class ProxyConfigTest extends Specification {
...
+
+    def 'authenticator is null-safe when the requesting protocol is missing'() {
+        given: 'an https_proxy-derived config'
+        def proxy = new ProxyConfig(protocol: 'https', host: 'proxy.example.com', port: '8080', username: 'foo', password: 'bar')
+        def authenticator = proxy.authenticator()
+
+        expect: 'a null requesting protocol does not throw and yields no credentials'
+        invokeAuth(authenticator, Authenticator.RequestorType.PROXY, 'proxy.example.com', 8080, null) == null
+    }
+
```
bentsherman:
```suggestion
```

pditommaso (approved):
LGTM — both review points fully addressed in b0e8402 (NPE-safe receiver flip + reverse-direction and null-protocol regression tests). Fix is correct, minimal, and well-tested.

PR #7372, opened 2026-07-21
Refactor HubOptions from a trait into a record

bentsherman:
Spun out from #5971 

Split the `HubOptions` trait, which mixed CLI parameter declaration and the credentials data model, into two focused pieces:

- `nextflow.scm.HubOptions`: an immutable record holding the git provider and user credentials, i.e. the data model consumed by `AssetManager`
- `nextflow.cli.HubAware`: carries the `-hub`/`-user` command line parameters and builds a `HubOptions` via `toHubOptions()`

Commands (`CmdClone`, `CmdPull`, `CmdRun`) now implement `HubAware` and pass `toHubOptions()` to `AssetManager` rather than mixing in the trait, decoupling the runtime data model from the CLI parameter declaration.

`HubOptionsTest` is updated to exercise the record API.


pditommaso:
Nice cleanup — splitting the CLI param-carrier (`HubAware`) from the data model (`scm.HubOptions`) removes the `scm → cli` dependency direction, which is the real win here. Verified the branch builds and `LauncherTest`/`AssetManagerTest` pass. A few suggestions:

### Prefer a plain immutable class over the `record`

The record forces a public canonical accessor per component (`user()`), and we also override `getUser()` to return the *parsed* username. So `user()` returns the raw `user:password` string while `getUser()` strips the password — two accessors for one field returning different values. That's an easy-to-trip-over footgun (someone reaches for `.user()` expecting the username and gets the password too), and it's baked in because a record can't suppress its canonical accessors.

A plain `@CompileStatic` class with final fields exposes only the getters we write, so the raw credential string never becomes a first-class accessor:

```groovy
@CompileStatic
class HubOptions {
    private final String provider
    private final String credentials

    HubOptions(String provider, String credentials) {
        this.provider = provider
        this.credentials = credentials
    }

    String getProvider() { provider }

    String getUser() {
        if( !credentials ) return credentials
        final p = credentials.indexOf(':')
        return p != -1 ? credentials.substring(0, p) : credentials
    }

    String getPassword() {
        if( !credentials ) return null
        final p = credentials.indexOf(':')
        if( p != -1 ) return credentials.substring(p + 1)
        final console = System.console()
        if( !console ) return null
        print "Enter your $provider password: "
        return new String(console.readPassword())
    }
}
```

The only thing the record gave for free was `equals`/`hashCode`/`toString`, and nothing here uses them on `HubOptions`. (Avoid `@Immutable` as a shortcut — it auto-generates `getCredentials()`, which re-opens the same leak.)

### Consider splitting the `CmdPull` behavior change into a separate PR

In `CmdPull`, the `--all` branch went from `new AssetManager(proj)` to `new AssetManager(proj, toHubOptions())`, so bulk pulls now receive hub credentials where they previously didn't. That looks like an intentional (and reasonable) fix, but it's a behavior change riding along in a pure refactor — worth isolating so it's easy to spot in the history and revert independently if needed.

### Minor

- Stale KDoc in `AssetManager`: a few `@param`/doc lines still say *"See `{@link HubOptions}` trait"* and *"credentials provided on the command line"* — `HubOptions` is now a data model, not a trait, and no longer command-line-specific.
- `setupCredentials` rebuilds the object (`new HubOptions(hub, hubOpts.user())`) solely to swap in the resolved provider for `getPassword()`'s prompt string. Correct and faithful to the old `cliOpts.hubProvider = hub` mutation, just slightly obscure — a short comment would help.

bentsherman:
Thanks Paolo — all addressed in `Apply suggestions from review`:

1. **Record → plain immutable class.** Done, adopted your class almost verbatim. The one addition is a `withProvider(String)` helper so `setupCredentials` can rebind to the resolved provider without re-exposing the raw `user:password` string (which is exactly the leak we wanted to close). The raw `user()`/`provider()` canonical accessors are gone; `AssetManager` now uses `getUser()`/`getProvider()`.

2. **`CmdPull --all` behavior change.** Reverted to `new AssetManager(proj)` so this PR stays a pure refactor. Passing hub credentials to bulk pull can come back as its own intentional change if we want it.

3. **Stale `AssetManager` KDoc.** Fixed the three `@param hubOpts` lines that still called it a trait / command-line options.

4. **`setupCredentials` obscurity.** The rebuild is now `hubOpts.withProvider(hub)` (self-documenting) plus a short comment on why.

No test changes were needed — the existing constructions use the positional `HubOptions(provider, credentials)` form and read `getUser()`/`getPassword()`. `HubOptionsTest`, `AssetManagerTest`, and `LauncherTest` all pass.

pditommaso (approved):
Thanks Ben — all four points addressed cleanly. Validated locally: `LauncherTest` (34/34) and `HubOptionsTest` (3/3) pass, and the raw `credentials` string is no longer exposed as a public accessor. The remaining `AssetManagerTest` failures on my machine are just the `NXF_GITHUB_ACCESS_TOKEN` network tests, unrelated to this change. LGTM 🚀

PR #7373, opened 2026-07-21
Split CLI config handling out of ConfigBuilder

bentsherman:
Spun out from #5971 

Extract the CLI-specific configuration logic (option/command handling and config-file resolution: home, base dir, local, and user config files) from `ConfigBuilder` into a new `ConfigCmdAdapter` that wraps it. `ConfigBuilder` is left as a lean, `@CompileStatic` runtime builder that consumes an explicit list of config entries, and all call sites are migrated to `ConfigCmdAdapter` (CmdRun, CmdConfig, CmdClean, CmdNode, CmdFs, CmdLineage, CmdAuth, the module commands, PluginAbstractExec, K8sDriverLauncher, and the nf-console/nf-tower plugins). `DefaultRemoteModuleResolver` is updated for the new build() API.

This prepares the runtime config builder to be decoupled from the CLI, and keeps the forthcoming CLI/runtime module split a mechanical move.

Two intentional behavior changes come with the rewrite (both covered by new tests in ConfigBuilderTest):

- `session` is no longer created as a guaranteed top-level config scope. It is not a config scope, so code must not rely on it existing when unset.
- The v2 config parser no longer exposes environment variables to the config binding (they remain available to the v1 parser). The v2 parser does not support implicit environment variables.

### DIFF
--- a/modules/nextflow/src/main/groovy/nextflow/cli/CmdAuth.groovy
+++ b/modules/nextflow/src/main/groovy/nextflow/cli/CmdAuth.groovy
@@ -22,7 +22,7 @@ import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
 import nextflow.Const
 import nextflow.SysEnv
-import nextflow.config.ConfigBuilder
+import nextflow.config.ConfigCmdAdapter
 import nextflow.exception.AbortOperationException
 import nextflow.platform.PlatformHelper
 import nextflow.plugin.Plugins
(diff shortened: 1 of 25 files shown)


pditommaso (approved):
Ship-worthy refactor — the CLI/runtime layering is clean and test coverage is solid (behavior changes both have dedicated tests). Approving.

A few minor nice-to-haves, none blocking:

- **`ConfigCmdAdapter.setOptions(null)` NPE guard.** Under `@CompileStatic`, `builder.setIgnoreIncludes(options?.ignoreConfigIncludes)` boxes to `Boolean`, so a null `options` unboxes null into the `boolean` param → NPE. The old code guarded this. No current call site passes null, so it's latent, but a cheap `?: false` restores the old robustness.

- **Confirm the tower/auth `NXF_CONFIG_FILE` drop.** `readConfig()` previously used `setCurrentDir(HOME)`, which also honored `NXF_CONFIG_FILE` and `$HOME/.nextflow/nextflow.config`; it now reads only `$HOME/.nextflow/config`. Almost certainly intended (the added comment says so), just flagging in case anyone relies on the override for `nextflow auth`.

- **v2 env-binding change deserves changelog visibility.** The v2 parser no longer exposing env vars to the config binding is a user-visible semantic change — worth an explicit changelog/release-note line, not just the PR body.

Everything else is cosmetic (e.g. the now-redundant third `.exists()` + warn branch in `buildGivenFiles` is unreachable once `validateConfigFiles` has run).

bentsherman:
- Confirmed locally that `ConfigCmdAdapter.setOptions(null)` will not NPE, even with CompileStatic

- Removed the comment about `nextflow.config` / `config`, which was a holdover from debugging. This change does preserve the intent of `auth` / `launch` which is to load the tower credentials from the user-level config

- The remove of implicit env vars from the v2 config parser is already documented [here](https://docs.seqera.io/nextflow/strict-syntax#implicit-environment-variables). This change is just removing already-dead code

PR #7363, opened 2026-07-20
Support workflow modules (kind: Workflow) in the module system

jorgee:
> **Draft** — stacked on #7342 (the workflow-modules ADR). Base is `adr-workflow-modules` so the diff shows only the implementation; retarget to `master` once #7342 lands.

Implements the **workflow modules** feature from the ADR (#7342): the module system is extended to support standalone *workflow* modules (`kind: Workflow`) alongside process modules — published, installed, resolved, included, and (via a follow-up PR) executed through the existing registry / `nextflow module` CLI infrastructure.

### Spec model & schema
- `ModuleSpec`/`ModuleSpecFactory` parse a top-level `kind` (`Process`|`Workflow`, default `Process`) and a nested `requires.modules` list of direct dependencies.
- `adr/module-spec-schema.json` allows `kind`, `requires.modules` (exact-pin references), nested `scope/name` names, and adds `channel` / `custom-record` to the parameter `type` enum as documentation tags for statically-typed declarations.

### Install / include
- **Nested vendoring** — `ModuleResolver` installs a workflow module together with its transitive dependencies, each vendored under its own nested `modules/` directory (`modules/<scope>/<name>/modules/...`). No cross-module flattening — duplication accepted per the ADR's diamond-dependency decision — plus dependency-cycle detection.
- **Context-relative include resolution** — the runtime path (`IncludeDef`) and both compile-time paths (nf-lang `ResolveIncludeVisitor` and the source-discovery `ModuleResolver`) resolve a module's remote includes relative to the *including file's* directory, so a workflow module's own includes find its nested dependencies.
- **Integrity checksum** covers the module's full installed subtree (own files + vendored deps), computed after vendoring — so a freshly installed workflow-with-deps reads `VALID`, and edits to either the module or a vendored dependency are detected as `MODIFIED`.
- **`module install -update-deps`** — for an already-installed module, re-vendors its dependencies to match the (possibly locally-edited) `meta.yml` without reinstalling the module itself: installs new deps, updates changed versions, prunes removed ones (a locally-modified dependency is never silently overwritten or pruned). Ignored when the module is not installed; mutually exclusive with `-force`.

### Publish / validate
- `kind`-aware validation: a workflow module must define exactly one `workflow`; when its `meta.yml` declares `input`/`output`, the counts must match the workflow's `take:`/`emit:` arity.
- **`requires.modules` are validated against the registry** at publish time (they exist at their pinned version), since dependencies are re-resolved from the registry by consumers at install time.
[...]

PR #7379, opened 2026-07-24
Direct execution of named workflows

jorgee:
> **Draft.** Ports PR #7208 ("Direct execution for named workflows"). Complements the workflow-modules PR (#7363) but is independent of it — this only touches the script/runtime layer, no module-system changes.

Enable running a single named, statically-typed workflow directly, without writing an explicit entry workflow. This is the execution primitive behind `nextflow module run scope/name` for workflow modules, and it also applies to a local `nextflow run script.nf` that defines exactly one named workflow.

## What it does

Given a script with a single named, typed workflow:

```groovy
nextflow.enable.types = true

workflow HELLO_SHEET {
    take:
    greetings: Channel<Map>   // samplesheet
    prefix: String            // scalar

    main:
    ...
    emit:
    result: Channel<String> = ...
}
```

```bash
nextflow run script.nf --greetings sheet.csv --prefix Hi
```

Nextflow synthesizes an entry workflow that maps `--<take>` params to the workflow inputs, invokes it, and publishes its `emit:` channels as outputs.

## Changes
- **nf-lang `ScriptToGroovyVisitor`** — lower each workflow take with its declared type (`_take_(name, type)`) so take types are available at runtime.
- **`WorkflowDef` / `WorkflowParamsDsl`** — record take types and expose `getDeclaredInputTypes()`; untyped takes (`Object`) are omitted.
- **`ScriptMeta.hasExecutableWorkflows()`** — a non-module script with exactly one named workflow.
- **`BaseScript.run0()`** — when no entry is selected, synthesize an entry via `WorkflowEntryHandler` for a single named workflow, else fall back to the process entry handler; a multi-workflow script still reports "No entry workflow specified".
- **`WorkflowEntryHandler`** (new) — maps `--<take>` params to inputs using the declared types:
  - `Channel<T>` → load a samplesheet file (CSV / JSON / YAML) or wrap a collection with `channel.fromList`;
  - `Value<T>` → value channel;
  - scalar (`String`/`Integer`/`Path`/…) → coerced value.
  Then invokes the workflow and publishes its emits. **Typed workflows only** (guarded by `isTypingEnabled()`).

## Not in this PR (follow-ups)
- **Outputs** reuse the existing `OutputDsl` publishing (as the process entry handler does); the ADR's no-output-directory / work-dir-path index-file behaviour is a follow-up.
- **Typed-record samplesheets** — rows currently load as `Map`s; casting each row to a custom `record` type (`Channel<Sample>`) is a follow-up (the `__Params`-class TODO from #7208).

## Testing
- **Unit**: `WorkflowEntryHandlerTest` (samplesheet loading for CSV/JSON/YAML, param→take resolution by type, guards). Full `nextflow.script.*` and `nf-lang` suites pass.
- **Manual e2e**: ran a typed workflow with a `Channel<Map>` samplesheet take + a scalar `String` take via `nextflow module run ./mod --greetings sheet.csv --prefix Hi` → outputs published; verified the scalar is passed as a value and the CSV becomes a channel.

Co-authored from PR #7208 by @bentshermann.

### DIFF
--- a/adr/20260601-named-workflow-execution.md
+++ b/adr/20260601-named-workflow-execution.md
@@ -130,7 +130,7 @@ The following coercions apply:
 | `Map<K,V>`, `Record` | not supported | used as-is |
 | `Tuple` | not supported | not supported |
 | record types | not supported | map converted to record; fields are validated and converted to declared type |
-| `Channel<E>` | a samplesheet path, loaded as described below | a samplesheet path |
+| `Channel<E>` | a samplesheet path, loaded as described below; `E` must be a `Map`, `Record`, or record type | same as the command line |
 | `Value<V>` | parsed as `V`, then wrapped in a value channel | converted to `V`, then wrapped in a value channel |
 
 Composite types cannot be expressed as a single command-line string, so they must be supplied through the config. Nextflow does not attempt to split a command-line string into a collection, because there is no separator that is safe for every element type.
@@ -176,7 +176,7 @@ Where `loadData()` is a generic data-loading function that supports multiple fil
 
 *Future work:* the supported formats are currently a fixed set (CSV, JSON, YAML) implemented internally, and `loadData()` is not available to pipeline code. Exposing it as a function and allowing it to be extended via plugin to support additional formats (e.g. Parquet) is deferred.
 
-The channel input can use a generic type such as `Map` or `Record`, or a custom record type to enable further validation. In the above example, using the `Sample` type ensures that each samplesheet row is validated against the record fields and the `fastq_1` and `fastq_2` columns are treated as file paths.
+The channel input can use a generic type such as `Map` or `Record`, or a custom record type to enable further validation. It cannot use a scalar element type such as `Path` or `String`, because a samplesheet record is a set of named fields, not a single value. In the above example, using the `Sample` type ensures that each samplesheet row is validated against the record fields and the `fastq_1` and `fastq_2` columns are treated as file paths.
 
 ### Saving output channels to index files
 
--- a/adr/20260601-named-workflow-execution.md
+++ b/adr/20260601-named-workflow-execution.md
@@ -172,7 +172,9 @@ workflow {
 }
 ```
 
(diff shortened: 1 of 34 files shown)


adr/20260601-named-workflow-execution.md
```diff
@@ -0,0 +1,185 @@
...
+
+The following coercions apply:
+
+| Declared type | Command line / params file | Config |
+| ------------- | -------------------------- | ------ |
+| `Boolean`, `Integer`, `Float`, `String` | parsed from the string | used as-is |
+| `Duration`, `MemoryUnit`, `VersionNumber` | parsed from the string | parsed if given as a string |
+| `Path` | resolved to a path, which must exist | resolved to a path, which must exist |
+| `List<E>`, `Set<E>`, `Bag<E>` | not supported | each element converted to `E` |
+| `Map<K,V>`, `Record` | not supported | used as-is |
```
jorgee, line 130: I think we use dot notation to express maps/records in the case of processes. Maybe we could use it to support CLI maps

modules/nextflow/src/main/groovy/nextflow/script/BaseScript.groovy
```diff
@@ -273,15 +273,24 @@ abstract class BaseScript extends Script implements ExecutionContext {
...
-                // Create a workflow to execute the process (single process or first of multiple)
+            // a process or named workflow can be executed directly only
+            // when the script was launched via `nextflow module run`
+            final moduleRun = session.isModuleRun()
+            if( moduleRun && meta.hasExecutableWorkflows() ) {
+                // Execute a single named workflow directly
+                final handler = new WorkflowEntryHandler(this, session, meta)
+                this.entryFlow = handler.createEntryWorkflow()
+            }
+            else if( moduleRun && meta.hasExecutableProcesses() ) {
```
jorgee, line 289: This change is restricting the execution of process executions from module run. I agree but it is causing the failure in the integration test `test/process-entry.nf` introduced in #6381. From the error message, it looks like you can specify a process as entry point  but entryName is only for workflows.

bentsherman (approved):

christopher-hakkaart (approved):

bentsherman:
I took the opportunity here to align the parameter handling logic across pipeline params, process inputs, and workflow inputs. I also went through several rounds of agent review to catch various bugs and edge cases.

Let's continue to refine as we implement workflow modules.

PR #7380, opened 2026-07-24
Fix flaky CmdLogTest/CmdLineageTest (filter plugin-loading stack-trace noise)

jorgee:
## Problem

`CmdLogTest` and `CmdLineageTest` intermittently fail on CI with `stdout.size() == N` assertion failures — the same commit passed and failed across consecutive CI runs with **no code change**, and they pass locally in isolation.

## Root cause

These tests run the real CLI (`CmdLog.run()` / `CmdLineage.run()`), which calls `Plugins.init()` → pf4j's `loadPlugins()`, scanning the plugins directory. When the build environment contains a plugin with an **invalid descriptor** (an empty `Plugin-Id`), pf4j raises `InvalidPluginDescriptorException` and its ~500-line stack trace is written to the captured stdout:

```
org.pf4j.InvalidPluginDescriptorException: Field 'id' cannot be empty
	at org.pf4j.AbstractPluginManager.validatePluginDescriptor(...)
	at org.pf4j.AbstractPluginManager.loadPlugins(...)
	...
```

The tests use `TestHelper.filterLogNoise(...)` to strip log noise before asserting on the command's real output, but that helper stripped only `DEBUG`/`INFO`/`WARN`/plugin lines — not raw stack-trace/exception lines — so the flood inflated `stdout.size()` (observed: 513 / 643 lines instead of 1 / 3).

## Fix

Harden `test.TestHelper.filterLogNoise` (used only by CLI/lineage output tests) to also drop exception stack traces:
- stack-trace frames (`at …`),
- `Caused by:`,
- `... N more`,
- fully-qualified exception header lines (a dotted class name ending in `Exception`/`Error`).

The exception-header match is deliberately strict so ordinary command output such as `Error loading lid://… - Lineage record … not found` is **not** treated as noise.

Adds `TestHelperTest` verifying the filter drops a pf4j-style stack trace, keeps real output, and does not mistake ordinary error messages for stack traces.

## Notes
- This makes the tests robust to plugin-loading diagnostics regardless of build environment. The separate, deeper question of *why* the CI build sometimes stages a plugin with an empty `id` descriptor (a build-cache / plugin-manifest issue) is out of scope here.

## Testing
`TestHelperTest`, `CmdLogTest`, `CmdLineageTest`, `WorkflowMetadataTest`, `LinPathTest`, `LinCommandImplTest` (all users of `filterLogNoise`) pass.

### DIFF
--- a/modules/nextflow/src/test/groovy/test/TestHelperTest.groovy
+++ b/modules/nextflow/src/test/groovy/test/TestHelperTest.groovy
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2013-2026, Seqera Labs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package test
+
+import spock.lang.Specification
+
+/**
+ * Tests for {@link TestHelper#filterLogNoise}.
+ *
+ * @author Jorge Ejarque <jorge.ejarque@seqera.io>
+ */
+class TestHelperTest extends Specification {
+
+    def 'should keep command output and drop log noise'() {
+        given:
+        def output = '''\
+            10:00:00.000 [main] DEBUG nextflow.Session - some debug
+            10:00:00.001 [main] INFO  nextflow.Session - some info
+            real output line 1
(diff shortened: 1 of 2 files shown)


bentsherman:
NOTE: I also encountered this issue and tried to fix it in two different ways:

- https://github.com/nextflow-io/nextflow/commit/b61eda21ead89cf647e725a1dfc84dfb4ad34493 : includes an edit to `build.gradle` to force tests to run after the plugin manifest is created

- https://github.com/nextflow-io/nextflow/commit/98e8573ed996eabd9b75c059b48c8b70c2870365 : reverts the previous fix and includes an edit to the DevPluginManager

I piggy-backed these fixes on other PRs because I was having a hard time reproducing the error and I wanted to make sure the fix worked for an actual PR. The first fix was a false positive, but the second fix seems to work

I think the DevPluginManager is the right place to fix this error so that tests don't have to deal with it. It is also a natural continuation of #7264

bentsherman:
Closing for now. We can revisit if the flaky test failures pop up again

PR #7382, opened 2026-07-24
Fix parameterized type cast in process body

### LINKED ISSUE #7381: Static typing: parameterized `as` cast fails at runtime in process script/exec
### Bug report

Under static typing (`nextflow.enable.types = true`), a cast to a **parameterized** type (e.g. `as Map<String,String>`, `as List<String>`) inside a process `script:` or `exec:` block compiles fine and passes the type checker, but throws at runtime:

```
No signature of method: nextflow.script.ScriptBinding._as_type() is applicable for argument types: () values: []
Possible solutions: asType(java.lang.Class), hashCode()
```

It looks like `expr as SomeType<...>` is lowered to a call to `_as_type()` on the script binding (which doesn't exist), whereas the same cast **without** type arguments (`as Map`) is lowered to `asType(Class)` correctly.

### Minimal reproduction

```nextflow
nextflow.enable.types = true

process FOO {
    output:
    stdout()

    script:
    def x = ([a: '1'] + [b: '2']) as Map<String,String>
    "echo done"
}

workflow { FOO() }
```

```
nextflow run main.nf
```

Fails with the `_as_type()` `MissingMethodException` shown above (script `main.nf` line 6).

### Characterization

| Cast | Location | Result |
|------|----------|--------|
| `x as Map<String,String>` | process `script:`/`exec:` | ❌ runtime `_as_type()` |
| `x as List<String>` | process `script:`/`exec:` | ❌ runtime `_as_type()` |
| `x as Map` (raw, no type args) | process `script:`/`exec:` | ✅ works |
| `x as String` | process `script:`/`exec:` | ✅ works |
| `x as Map<String,String>` | workflow body / operator closure | ✅ works |

So the failure is specific to **parameterized-type casts inside a process body**. The identical parameterized cast works in a workflow-body closure, e.g. this runs fine and prints `[a:1, b:2]`:

```nextflow
nextflow.enable.types = true
workflow {
    channel.of([a: '1']).map { m -> (m + [b: '2']) as Map<String,String> }.view()
}
```

### Expected behavior

`expr as SomeType<...>` inside a process `script:`/`exec:` block should behave like `expr as SomeType` (a runtime coercion), or the parser/type checker should reject the parameterized cast statically rather than emitting an unresolved `_as_type()` call.

### Environment

- Nextflow version: `26.04.6` build `12646`
- Discovered while migrating [nf-core/fetchngs](https://github.com/nf-core/fetchngs) to static typing.

bentsherman:
Fix #7381 

This PR refactors the `_as_type` calls for parameterized type casts into static calls on `TypeHelper` directly (same pattern used by `TaskCmdXformVisitor` -> `LangHelpers`)

This fixes an issue where parameterized type casts worked everywhere in a script except for the process body. It is also cleaner overall and consistent with existing patterns


pditommaso (approved):
## Review

Correct root-cause fix. `StripTypesVisitor` previously emitted `callThisX("_as_type", …)` — a `this`-scoped call resolving to `protected` shims on `BaseScript`. In a process body (`exec:`/output blocks) `this` is not the `BaseScript` instance, so those methods were unresolvable and the cast failed, while working fine in the script/workflow body. Switching to a direct static call `TypeHelper.asType(…)` resolves regardless of `this`, so it now works everywhere.

### Correctness (verified against source)
- **Behavior-preserving:** the deleted `BaseScript` methods held no instance state — they only forwarded to `TypeHelper`. Static dispatch is identical.
- **No overload ambiguity:** the record branch emits a 2-arg `asType(value, Class)`; only `asType(Object, Type)` exists at 2 args and `Class implements Type`, so it binds to the same target the old shim hit. The new `asType(Object, Class, String)` is the only 3-arg overload and its body is the old shim verbatim.
- **No dangling references:** `_as_type`/`_instanceof_record_type` were generated only by `StripTypesVisitor` and consumed only by `BaseScript`; removing generator and definition together leaves nothing dangling. Scripts recompile each run, so no stale-bytecode risk.
- **Classpath:** `TypeHelper` was already reachable via the old shims — no new requirement.

### Strengths
- Real regression test covering the exact failing path (`as Sample` inside `exec:`).
- Net simplification that removes an indirection layer and aligns with the existing `TaskCmdXformVisitor → LangHelpers` pattern, including hoisting the `ClassNode` to a `static final` field.
- Doc comment kept in sync.

No Critical/Important/Minor issues found.

**Ready to merge** — just rebase, the branch is currently behind `master`.

PR #7394, opened 2026-07-27
feat: add plugin lockfile integrity (plugins.lock)

pditommaso:
## Summary

Adds an opt-in, offline-verifiable integrity check for plugins: a committed `plugins.lock` file that pins a canonical SHA-512 of each plugin's **extracted `$id-$version/` directory** — the code Nextflow actually loads and executes — keyed by `id@version`, following the `go.sum` / `package-lock.json` model.

Because it hashes the code that runs (not the archive it came from), **one mechanism covers both integrity surfaces**:

- **Supply chain / registry compromise / silent drift** — a tampered archive extracts to a different tree → hash mismatch.
- **Cache poisoning of the executed code** — a lower-trust user editing already-extracted files on a shared cache → hash mismatch, detected **regardless of directory ownership or permissions**.

Full design rationale in `adr/20260727-plugin-lockfile-integrity.md`.

## Relationship to #7308 (supersedes)

This started as the "committed lockfile" long-term direction raised in the review of #7308. In its final form it **supersedes** #7308's `PluginSecurity` directory-ownership guard rather than complementing it:

| | Extracted-tree poisoning | Supply chain / drift | Noise on a healthy shared cache |
|---|---|---|---|
| #7308 (ownership guard) | heuristic (misses group-writable-trusted-owner) | ✗ | **warns per plugin, every run** |
| this PR (tree-hash lock) | **✓ — any modification, any owner** | **✓** | **silent** (output only on a real mismatch) |

The content hash is both **stronger** (catches modifications an ownership check misses) and **quieter** (no ownership heuristic, so no false-positive warnings on legitimate shared/admin caches). I'd suggest closing #7308 in favour of this. The zero-config baseline for the tree surface remains the operational guidance to keep the plugin cache per-user private; this is the committed, enforceable layer on top.

## How it works

- **No dedicated command.** Populated automatically on first load (trust-on-first-use). Enable by creating an empty `plugins.lock` (`touch plugins.lock`) and running once, then review the diff and commit.
- **Never rewrites silently.** An existing entry is only verified; a plugin whose tree differs from a committed entry is a *failure*, gated by `NXF_PLUGINS_LOCK_MODE` (`warn` default / `strict` / `off`). Re-pin explicitly: delete the entry and re-run.
- **Offline, no false positives.** Verification re-hashes the local extracted directory in `PluginUpdater.load0()` before loading (cold and warm caches, same path); never re-fetches from the registry. No retained archive, no "archive absent" case.

## Performance

Full re-hash on every load is fine for small plugins; for large cloud plugins it's material. The ADR documents a **follow-up** optimisation (a local, non-committed `(path,size,mtime)` fingerprint cache to skip the re-hash when nothing changed, trusted only on a private non-writable cache — ownership selects the *strategy*, never emits a warning), deliberately left as measure-first rather than built up front.

## Testing

- `PluginLockFileTest` — round-trip, blank/`touch`ed file, malformed file, lookup, add/update.
- `PluginLockVerifierTest` — mode resolution; `sha512Tree` determinism & content-sensitivity; dormant vs enabled-by-presence; TOFU auto-pin writes the tree hash; match passes; mismatch aborts (`strict`) / warns once (`warn`) / ignored (`off`); **cache-poisoning caught regardless of ownership**.
- `PluginUpdaterTest`, `PluginsFacadeTest`, `CmdPluginTest` — unchanged, still green.

Docs: `NXF_PLUGINS_LOCK_MODE` in `env-vars.mdx`; a Lockfile section + shared-cache note in `using-plugins.mdx`.

Supersedes #7308.


docs/plugins/using-plugins.mdx
```diff
@@ -56,6 +56,30 @@ Plugin declarations in Nextflow configuration files are ignored when specifying
 
 When Nextflow downloads plugins, it caches them in the directory specified by `NXF_PLUGINS_DIR` (`$HOME/.nextflow/plugins` by default).
 
+:::note
+The plugin cache is shared across pipelines and is not access-controlled. On multi-tenant or shared systems, use a private cache directory per user (set `NXF_PLUGINS_DIR` to a location only you can write) to avoid loading plugin artifacts populated by another user.
```
christopher-hakkaart:
```suggestion
The plugin cache is shared across pipelines and has no access controls. On multi-tenant or shared systems, set `NXF_PLUGINS_DIR` to a directory only you can write to. This prevents Nextflow from loading plugin artifacts from other users.
```

docs/plugins/using-plugins.mdx
```diff
@@ -56,6 +56,30 @@ Plugin declarations in Nextflow configuration files are ignored when specifying
...
+:::
+
+### Lockfile
+
+<AddedInVersion version="26.07" />
+
+A `plugins.lock` file pins the exact plugin artifacts a pipeline expects. For each plugin it records the `sha512` checksum of the plugin archive, keyed by `id@version`. The file is meant to be committed to the pipeline repository so that everyone running the pipeline resolves the same plugin artifacts.
```
christopher-hakkaart:
```suggestion
A `plugins.lock` file pins the exact plugin artifacts a pipeline expects. For each plugin, it records the `sha512` checksum of the plugin archive, keyed by `id@version`. Commit the file to the pipeline repository so that everyone who runs the pipeline resolves the same plugin artifacts.
```

docs/plugins/using-plugins.mdx
```diff
@@ -56,6 +56,30 @@ Plugin declarations in Nextflow configuration files are ignored when specifying
...
+
+The lockfile is populated automatically, like `go.sum` or `package-lock.json` — there is no separate command. To enable it, create an empty file in the pipeline directory and run the pipeline once:
+
+```bash
+touch plugins.lock
+```
+
+The first time each plugin is downloaded, its archive checksum is added to `plugins.lock`. Review the resulting file and commit it. On subsequent runs Nextflow verifies each plugin against the committed checksum. When no `plugins.lock` file is present, the feature is dormant and has no effect.
```
christopher-hakkaart:
````suggestion
To enable the lockfile:

1. Create an empty `plugins.lock` file in the pipeline directory:

   ```bash
   touch plugins.lock
   ```

2. Run the pipeline once. The first time Nextflow downloads a plugin, it adds that archive's checksum to `plugins.lock`.
3. Review the resulting file and commit it.

On later runs, Nextflow verifies each plugin against the committed checksum. Without a `plugins.lock` file, the feature does nothing.
````

File: docs/plugins/using-plugins.mdx [hunk 4]
```diff
@@ -56,6 +56,30 @@ Plugin declarations in Nextflow configuration files are ignored when specifying
...
+
+Verification is fully offline: Nextflow re-computes the checksum of the plugin archive from a copy retained in the local cache and compares it to the lock entry, without contacting the plugin registry. An existing entry is never rewritten automatically — if a plugin archive legitimately changes, delete its entry and run again to re-pin it.
+
+Use [`NXF_PLUGINS_LOCK_MODE`][using-plugins-env-vars] to control what happens on a checksum mismatch: `warn` (default) logs a warning and continues, `strict` aborts the run, and `off` skips verification. A plugin whose retained archive is missing (for example, a cache populated before this feature existed) cannot be verified offline; it is reported but never aborts the run, and is never re-downloaded just to verify it.
```
christopher-hakkaart:
```suggestion
Verification runs offline. Nextflow re-computes the checksum from the copy of the archive retained in the local cache and compares it to the lock entry, without contacting the plugin registry.

Nextflow never rewrites an existing entry. If a plugin archive changes legitimately, delete its entry and run the pipeline again to re-pin it.

Use [`NXF_PLUGINS_LOCK_MODE`][using-plugins-env-vars] to control what happens when a checksum does not match:

| Mode             | Behavior                      |
| ---------------- | ----------------------------- |
| `warn` (default) | Logs a warning and continues. |
| `strict`         | Aborts the run.               |
| `off`            | Skips verification.           |

Nextflow cannot verify a plugin whose retained archive is missing. For example, in a cache populated before this feature existed. It reports the plugin, does not abort the run, and does not re-download the archive to verify it.
```

docs/plugins/using-plugins.mdx
```diff
@@ -56,6 +56,30 @@ Plugin declarations in Nextflow configuration files are ignored when specifying
...
+
+The lockfile complements, but does not replace, the private-cache guidance above: the cache isolation prevents untrusted artifacts from being loaded, while the lockfile ensures the artifacts that are loaded match what the pipeline pinned.
```
christopher-hakkaart:
```suggestion
The lockfile does not replace a private cache directory. A private cache prevents Nextflow from loading untrusted artifacts. The lockfile confirms only that the artifacts it loads match what the pipeline pinned.
```

christopher-hakkaart (commented):
Suggestions for docs

pditommaso:
@christopher-hakkaart for some reason I'm not able to apply your changes. Can you submit them again ?

pditommaso:
@claude review

pditommaso:
Addressed in 787c96f:

1. `plugins.lock` is now resolved from the pipeline project directory (wired via `Plugins.setLockFileDir` from `CmdRun`), not the launch directory.
2. `NXF_PLUGINS_LOCK_MODE=off` is now a full escape hatch — no hashing, no pinning, the file is not even read.
3. A failed TOFU pin warns instead of aborting the run.
4. The lockfile is written via a temp file + atomic move.
5. A malformed (or future-version) lockfile aborts with `AbortOperationException` and a fix-or-delete hint.

Also: `version` is validated, `sha512Tree()` skips non-regular files, and the docs cover the location, the scope, and stale entries. The full re-hash per load stays a measure-first follow-up in the ADR.

docs/plugins/using-plugins.mdx
pditommaso: Applied, along with the rest of your suggestions in this file — folded into the rewrite in 787c96f (numbered enable steps, the mode table, and the closing paragraph on the private cache). The wording that referred to the plugin *archive* checksum was reworded to the extracted-directory hash, since that is what the final version pins.

jorgee (changes requested):
I have look at the PR and overall sounds good, but I think there is a gap in the integrity mechanism when resolved version is different from the lock and already in the cache. 

**1. Version resolution gap**

The lock is keyed by `id@version` and never participates in version *resolution* — `load0()` resolves the coordinate first (`PluginUpdater.groovy:508-527`) and only then calls the verifier with the already-final `id@version` (`:541`).

`plugins.lock` pins to certain plugin version; but if an upgrade is produced and the plugin is declared unpinned, the run would resolve to a high version which can be already extracted in the shared cache. What happens:

- The directory exists, so the download is short-circuited and pf4j's `CompoundVerifier` never runs. Nothing checks the tree against the registry.
- `PluginLockVerifier.groovy:122` — `getEntry("nf-amazon@2.1.0")` returns `null`.
- `:125-128` — `if( entry == null ) { pin(fqid, actual); return }`. **`getMode()` is never consulted on this branch.**
- The lockfile is rewritten, one `log.info` line is emitted, and `2.1.0` executes.

Therefore a `strict` run can be executed without the proper plugin verification. 
Smallest fix I'd suggest — make `strict` mean "everything loaded must already be pinned", replacing `PluginLockVerifier.groovy:125-128`:

```groovy
if( entry == null ) {
    if( getMode() == Mode.STRICT )
        throw new AbortOperationException(
            "Plugin '$fqid' is not present in the plugins lock file: $lockPath\n" +
            "- pin the plugin version in the `plugins` config scope, or re-run with NXF_PLUGINS_LOCK_MODE=warn to record it")
    pin(fqid, actual)
    return
}
```
Independently of that, I'd `log.warn` (not `log.info`) when the lock already holds a *different* version of the same plugin

One side effect of any auto-pin, related to the above: for `nextflow run org/pipeline` the lock dir is the managed clone under `$NXF_HOME/assets` (`CmdRun.groovy:341,346`). Appending an entry there dirties that clone.

Apart from this version gap, I run agent review that found other issues worth to have a look as I think they are gaps in the model:

**2. An *added* symlink under `<plugin>/lib/` is loaded by pf4j but invisible to the hash** 

`PluginLockVerifier.groovy:177-183` only digests entries where `attrs.isRegularFile()`, and `Files.walkFileTree` without `FOLLOW_LINKS` reports a symlink as non-regular — so links are dropped from the digest.

pf4j picks them up regardless. In 3.12.0, `FileUtils.getJars` uses `File.listFiles(new JarFileFilter())`, and `ExtensionFileFilter.accept` is **name-only** — `getName().toUpperCase().endsWith(".JAR")`, with no `isFile()` check. `DirectoryFileFilter` recursion uses `Files.isDirectory(path)`, which follows links.

I confirmed this: pinned a plugin tree, then added `nf-amazon-x.y.z/lib/aaa-evil.jar` as a symlink to an attacker-owned file. The tree hash was unchanged and `strict` mode passed silently, while the jar is on the plugin classloader.

**3. Plugins started before `setLockFileDir()` are neither verified nor pinned**

`Plugins.setLockFileDir(baseDir)` is called at `CmdRun.groovy:346`, but `Plugins.init()` runs at `:335` and two things in between can start a plugin while `lockFile` is still `null` — at which point `getLockVerifier()` (`PluginUpdater.groovy:381-385`) builds a *dormant* verifier and the plugin loads with no hash check and no pin, in every mode including `strict`:

- `:338` `getScriptFile(pipeline)` → `AssetManager` → `RepositoryFactory.groovy:105,123,137` → `Plugins.startIfMissing('nf-codecommit')`
- `:342` `parsedParams(...)` → `validateParamsFile` (`:815`) → `FileHelper.asPath` → `FileHelper.groovy:377` → `Plugins.startIfMissing(...)`. So a `-params-file s3://…` starts **nf-amazon** four lines before the lock dir is set.

The second one is a one-line ordering fix: move `Plugins.setLockFileDir(baseDir)` up to immediately after `final baseDir = scriptFile.parent` at `:341`, ahead of `parsedParams`.

The first is structurally unavoidable — you need the project directory to locate the lock file, and finding the project directory is what starts the SCM plugin — so I'd just note it in the docs' scope paragraph next to the existing dev/embedded caveat. Same paragraph should mention `nextflow plugin install`/`pull`: neither calls `setLockFileDir` (only `CmdRun` does), so those paths neither verify nor pin, which makes `nextflow plugin install nf-amazon@2.1.0` a supported way to seed the cache with a tree a later `strict` run then trusts on sight (see #1).

Note also that `FileHelper.SCHEME_CHECKED` (`FileHelper.groovy:367,387`) memoises per scheme, so a plugin started on the pre-hook path is never revisited once the lock *is* armed.

**4. No test covers the integration, so the feature can be silently unwired**

`PluginLockFileTest` and `PluginLockVerifierTest` are good unit tests of the verifier in isolation, but nothing exercises the wiring. In particular nothing pins that `load0()` actually calls `verify()` *before* `loadPluginFromPath()` — a refactor could reorder or drop that call and the whole suite would stay green. Also untested: that `setLockFile()` invalidates the cached verifier, and that `PluginsFacade.setLockFileDir(dir)` resolves `plugins.lock` under `dir`.

pditommaso:
Thanks @jorgee, all four addressed and pushed (up to b33b135):

**1. Version resolution gap.** Took your suggested fix: strict mode now aborts when a loaded plugin has no lock entry instead of auto-pinning it, so an unpinned coordinate that resolves past the pin (and is already extracted in a shared cache) can no longer slip through unverified. `warn`/default still pins on first use, and now also warns when the lock already holds a *different* version of the same plugin.

**2. Symlink invisible to the hash.** Nice catch, that was a real bypass. `sha512Tree` now includes symbolic links — it records the link's path + target (domain-separated from regular files, never dereferenced), so an added or retargeted `lib/*.jar` link changes the tree hash and fails strict verification.

**3. Ordering.** Moved `setLockFileDir` ahead of `parsedParams`, so a plugin started by a remote `-params-file` is covered. The SCM-provider start and `plugin install`/`pull` are structural (you need the project dir to locate the lock), so I documented them in the scope note rather than trying to wire them.

**4. Integration tests.** Added tests that pin the wiring, not just the verifier in isolation: `installPlugin` verifies+pins on the load path, strict aborts *before* `loadPluginFromPath` on a mismatch, `setLockFile` invalidates the cached verifier, and `PluginsFacade.setLockFileDir` resolves `plugins.lock` under the given dir.

Also folded in a small no-behaviour-change refactor to keep `verify()`/`sha512Tree()` readable after the extra branches. Ready for another look whenever you have a minute 🙏

jorgee (approved):

bentsherman (changes requested):
@aknownuser @pditommaso Before we move forward with this PR, I think we need to define Nextflow's threat model

I have been reflecting on this ever since I encountered [Groovy's threat model](https://github.com/apache/groovy/blob/master/THREAT_MODEL.md). Groovy assumes you are running trusted code, on trusted input data, in a trusted environment -- otherwise Groovy would be vulnerable by construction. I think Groovy's threat model pretty much applies wholesale to Nextflow as well

Many of the recent security-related PRs assume an untrusted environment: a shared HPC system -> poorly-managed directory permissions -> a malicious user reading your secrets or poisoning your cache. This kind of situation strikes me as being out-of-model

Even if it were in-model, the current PR doesn't really close the vulnerability:
- checksum verification is opt-in and only warns by default -- that doesn't prevent a cache-poisoning attack if you only notice the warning after the run completes
- transitive plugins don't seem to be covered
- plugins with SCM providers aren't covered (because they must be loaded before the lock file is resolved)

All of these gaps seem to undermine the whole purpose of the lock file, which makes me think it's not worth doing

PR #7405, opened 2026-07-28
Fix `params` block breaking piped operators in the entry workflow

### LINKED ISSUE #7399: A `params` block makes pipe-form operators undefined in the entry workflow
## Problem

In a script without `nextflow.enable.types`, declaring a `params { }` block causes the **entry**
workflow to lose the DSL1 operators (`view`, `map`, `filter`, …) from its scope. Any operator called
via the pipe form then fails to compile with "`x` is not defined". Named workflows in the same file
are unaffected, and so is the dotted call form.

The script does not run — this is not limited to `nextflow lint`.

## Minimal reproducible example

```nextflow
params {
    greeting: String = 'hello'
}

workflow {
    channel.of(1, 2) | view()
}
```

**Observed**:

```
Error mre.nf:6:24: `view` is not defined
│   6 |     channel.of(1, 2) | view()
╰     |                        ^^^^

ERROR ~ Script compilation failed
```

**Expected**: the pipeline runs and prints `1` and `2`.

Deleting the `params` block — the only change — makes it run:

```nextflow
workflow {
    channel.of(1, 2) | view()
}
```
```
1
2
```

An **empty** `params { }` block triggers it too, so it is the block's presence rather than its
contents.

## Scope

One file showing what is and is not affected:

```nextflow
params {
    greeting = 'hello'
}

workflow SUB {
    take:
    ch

    main:
    ch | view()        // OK — named workflow
}

workflow {
    def a = channel.of(1, 2)
    a.view()           // OK — dotted form
    a | view()         // Error: `view` is not defined
}
```

The `params` block is otherwise fully functional — the same script with the dotted form runs and
reads the param:

```nextflow
params {
    greeting = 'hello'
}

workflow {
    channel.of(1, 2).view()
    println "greeting is ${params.greeting}"
}
```
```
greeting is hello
1
2
```

## Summary

| | no `params` block | `params` block |
|---|---|---|
| untyped (no flag) | `a \| view()` works | **`a \| view()` fails** |
| `nextflow.enable.types = true` | fails | fails |

Per https://docs.seqera.io/nextflow/workflow-typed I don't think there should be a problem with using the `params` block without typing:

> Set `nextflow.enable.types = true` in every script that uses typed workflows. The `params` block
> and `output` block can be used without this feature flag.

And per https://docs.seqera.io/nextflow/workflow#special-operators--and- I don't think `|` is disallowed either?:

> These operators are not supported when [static typing](https://docs.seqera.io/nextflow/tutorials/static-types) is enabled

## Environment

- Nextflow 26.04.4
- Reproduced with both `nextflow run` and `nextflow lint`

bentsherman:
Fix #7399

This PR fixes a bug in the strict parser where a `params {}` block causes piped operators like `ch | view` to not be recognized in the entry workflow

This is because the `WorkflowDsl` class used by the entry workflow must be augmented both with the typed `params` variable (when `params` block is defined) and the piped operators (when static typing is not enabled)

Due to code entropy (and lack of a unit test) these two behaviors accidentally clobbered each other. This PR fixes the issue and adds a regression test keep it fixed

### DIFF
--- a/modules/nf-lang/src/main/java/nextflow/script/control/VariableScopeVisitor.java
+++ b/modules/nf-lang/src/main/java/nextflow/script/control/VariableScopeVisitor.java
@@ -260,7 +260,6 @@ class VariableScopeVisitor extends ScriptVisitorSupport {
     public void visitWorkflow(WorkflowNode node) {
         var classScope = workflowDsl(node.isEntry());
         if( node.isEntry() && paramsType != null ) {
-            classScope = new ClassNode(classScope.getTypeClass());
             var paramsMethod = classScope.getDeclaredMethods("getParams").get(0);
             paramsMethod.setReturnType(paramsType);
         }
--- a/modules/nf-lang/src/test/groovy/nextflow/script/control/ScriptResolveTest.groovy
+++ b/modules/nf-lang/src/test/groovy/nextflow/script/control/ScriptResolveTest.groovy
@@ -534,4 +534,21 @@ class ScriptResolveTest extends Specification {
         deleteDir(root)
     }
 
+    def 'should resolve piped operators in the entry workflow with a params block' () {
+        when:
+        def errors = check(
+            '''\
+            params {
+                greeting: String = 'hello'
+            }
+
+            workflow {
+                channel.of(1, 2) | view
+            }
+            '''
+        )
+        then:
+        errors.size() == 0
+    }
+
 }


pditommaso (commented):
Verified locally: without the fix the new test fails with `` `view` is not defined ``, with it the full `:nf-lang:test` suite is green. The root cause diagnosis looks right — `workflowDsl()` already returns a fresh node, so the removed copy was silently dropping the operator methods it had just added.

Two follow-up suggestions and a couple of minor notes.

### The remaining asymmetry is a trap

`visitOutputs` (`VariableScopeVisitor.java:506-513`) has a near-identical block, but it starts from `ClassHelper.makeCached(OutputDsl.class)` — a *shared cached* node — so there the defensive `new ClassNode(...)` copy is load-bearing and must **not** be removed. After this PR the two blocks look almost the same while having opposite requirements, and neither states why. That's the same entropy that produced this bug.

Suggestion: remove the convention rather than document it — e.g. let `workflowDsl()` accept the `paramsType` and apply the override itself, so "must be a private copy" is enforced by construction. Failing that, a one-line comment at each site.

### The test pins only half of the interaction

The new test pins the operator half, but nothing pins the `params` type override that was clobbering it: `TypeCheckingTest.groovy` has no `params` references, and the only other `params {}` blocks in nf-lang tests are in `ScriptAstBuilderTest` (parser-level, no scope resolution). If this change had broken the `paramsType` override instead of fixing operators, the suite would still be green.

Cheap to close in the same test:

- declared param resolves — add `println(params.greeting)` to the workflow body
- undeclared one still errors — `params.missing`, expecting 1 error

Also uncovered: `params {}` with typing *enabled*, where operators must not be added but params must still resolve.

### Minor

- `getDeclaredMethods("getParams").get(0)` is unguarded at both sites. Pre-existing and fine in practice, but `getDeclaredMethods` ignores supertypes, so it would throw `IndexOutOfBoundsException` if `getParams` ever moved to a parent interface.
- Unverified and out of scope here, but in the same method: `ClassNode.addMethod` in `workflowDsl()` may call `setDeclaringClass` on the `MethodNode`s taken from the cached `WorkflowDslV1` node, which would mutate shared AST across compilations. Could matter for the long-running language server — worth a separate look.

PR #7409, opened 2026-07-29
docs: split stdlib docs into separate pages

bentsherman:
This PR splits the stdlib docs into a separate page per type/namespace.

I also deleted some index pages but we might need to keep them due to cross-links

It seems like there is a limitation on the number of indents:

image

I don't know if this is a docusaurus limitation or just a config setting we can change


christopher-hakkaart:
The indentation is because we try not to go above 3 levels. However, I think this makes sense here. I also like index pages as a triage and intro to a set of pages like this. You can see what I would do here: #7410

christopher-hakkaart (approved):
Approved, as I think the splitting out looks good. Please see comment and #7410 to see improvements to indentation and suggested index pages. I'm not commited to these if you disagree.

PR #7412, opened 2026-07-30
Align Seqera executor maxSpotAttempts and Fusion handling

pditommaso:
Make the Seqera executor follow the same pattern used by AWS Batch and Google Batch.

- The spot attempts policy now lives in a `maxSpotAttempts` method in `SeqeraTaskHandler`, mirroring `AwsBatchTaskHandler` and `GoogleBatchTaskHandler`: the config value wins, otherwise fall back to `FusionConfig.DEFAULT_SNAPSHOT_MAX_SPOT_ATTEMPTS` when Fusion snapshots are enabled. `SchemaMapperUtil` no longer owns that policy, it just maps the resolved value. The method takes the hint-overlaid `MachineRequirementOpts` since Seqera resolves `seqera/machineRequirement.*` hints per task.
- The Fusion snapshot flag passed to the scheduler is guarded by `fusionEnabled()` at the call site, as done by the other two executors.
- `SeqeraExecutor.isFusionEnabled` now simply delegates to `FusionHelper.isFusionEnabled` instead of throwing from a boolean getter; the "requires Fusion" abort moved to `register`.

Behavior is unchanged in practice — Fusion is mandatory for this executor, so the new guard is always true. This aligns the structure and adds test coverage for the spot attempts resolution.


plugins/nf-seqera/src/main/io/seqera/executor/SeqeraTaskHandler.groovy
```diff
@@ -125,7 +127,8 @@ class SeqeraTaskHandler extends TaskHandler implements FusionAwareTask {
             baseMachineOpts,
             task.getContainerPlatform(),
             task.config.getDisk(),
-            fusionConfig().snapshotsEnabled()
+            fusionEnabled() && fusionConfig().snapshotsEnabled(),
```
bentsherman: are these `fusionEnabled()` checks needed? I thought Fusion is always enabled for SIC

plugins/nf-seqera/src/main/io/seqera/executor/SeqeraTaskHandler.groovy
```diff
@@ -165,6 +168,15 @@ class SeqeraTaskHandler extends TaskHandler implements FusionAwareTask {
         executor.getBatchSubmitter().submit(this, schedTask)
     }
 
+    protected int maxSpotAttempts(MachineRequirementOpts opts) {
+        final result = opts?.maxSpotAttempts
+        if( result )
```
stefanoboriero, line 175: Should we guard against negative values and fail early in the case?

plugins/nf-seqera/src/main/io/seqera/executor/SeqeraTaskHandler.groovy
pditommaso: You're right — Fusion is mandatory here (`register` aborts when it's not enabled), so `fusionEnabled()` is always true by the time a task is submitted. Dropped both occurrences in ff33e7db5's parent (87e7949a7); the snapshot flag is now tested directly at the `toMachineRequirement` call site and in `maxSpotAttempts`.

This does make the Seqera handler diverge slightly from `AwsBatchTaskHandler`/`GoogleBatchTaskHandler`, which keep the `fusionEnabled() &&` guard — but for those two Fusion is genuinely optional, so it's load-bearing there and dead weight here.

The test's `FUSION` column became meaningless as a result, so I collapsed it to just `SNAPSHOTS`.

plugins/nf-seqera/src/main/io/seqera/executor/SeqeraTaskHandler.groovy
pditommaso: Good catch — added in ff33e7db5.

A negative value was being silently dropped rather than misbehaving: `maxSpotAttempts` returned it as-is, then `SchemaMapperUtil` mapped any non-positive value to `null`. So no wrong request went out, but an invalid config was ignored without a word. It now throws:

```groovy
if( result != null && result < 0 )
    throw new IllegalArgumentException("Invalid maxSpotAttempts value: ${result} -- the value must be zero or a positive number")
```

Used `IllegalArgumentException` to match the neighbouring machine-requirement validation in `SchemaMapperUtil` (e.g. the invalid disk type / EBS-with-node-allocation checks), which is raised from this same `submit()` path.

Worth noting neither sibling executor validates this: AWS uses a plain truthiness check so a negative value passes straight through, and Google's `result > 0` silently treats negative as unset. Left those alone here, but they're arguably worth a follow-up.

stefanoboriero (approved):

bentsherman (approved):
CI failure looks like a flaky cloudinfo test

pditommaso:
Weird, twice in a row on the Java 25 suite

PR #7413, opened 2026-07-30
Use Platform user identity for the Seqera executor userName label

pditommaso:
The `nextflow.io/userName` auto label emitted by the Seqera executor was taken from `workflow.userName` — the OS user running the launcher — so the resource tags that end up on scheduler-provisioned cloud resources reported `ec2-user`/`root` (or whoever runs the launcher pod) instead of the Seqera Platform user that submitted the run. Cost reports grouped by that tag therefore attribute spend to the launcher environment, not to a person.

The label now reports `workflow.platform.user.userName`, which the Platform observer populates from the trace-create response (`onFlowCreate`) before the first task submission — so it is available when the executor builds the run labels. When there is no Platform identity (runs launched without a Platform token, plain CLI usage) it falls back to the OS user as before, so behaviour outside Platform is unchanged.

Only the value changes: the key stays `nextflow.io/userName`, so existing cost reports and dashboards built on it keep working and start showing the submitting Platform principal. Note that the identity is the owner of the token Nextflow was given — a user PAT resolves to that user, a workspace automation token resolves to that service account.

Docs: the `seqera.executor.autoLabels` reference now states which identity `userName` carries.

Related: seqeralabs/sched#883

### DIFF
--- a/docs/reference/config.mdx
+++ b/docs/reference/config.mdx
@@ -1786,7 +1786,7 @@ The following settings are available:
 
 ###### `seqera.executor.autoLabels`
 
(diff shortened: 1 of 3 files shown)


pditommaso:
Tested end to end

PR #7414, opened 2026-07-30
docs: Move channel factories and operators to stdlib reference

bentsherman:
This PR builds on #7409 by consolidating the channel factories and typed operators pages into the stdlib reference:

- Channel factories -> `channel` namespace
- Typed operators -> `Channel` type

Also adds a tutorial for topic channels, since the `channel.topic()` reference was trying to give a mini-tutorial. I based it loosely on nf-core/rnaseq since rnaseq-nf doesn't collect tool versions and nf-core/rnaseq is the most well-known pipeline that uses this pattern


bentsherman:
If the tutorial feels weird, another option would be to add a section on topic channels to the "Workflows" page under "Dataflow". It would be much shorter and introductory, similar to the current `channel.topic()` docs. I figured it deserved an end-to-end example, hence the tutorial

christopher-hakkaart (approved):
Looks really good already. I made very minor suggestions in #7451

PR #7417, opened 2026-07-30
docs: Move config scopes into separate pages

bentsherman:
This PR moves each config scope in the reference docs to its own page.

Minor additional edits:
- fixed default values missing a colon
- fixed various section heading issues
- added missing version callouts


christopher-hakkaart (approved):
Overall this looks really good. I've made #7436 to apply the style guide. I've noted that it might have made some words sentence case that might have been title case on purpose.

This PR is good as is, but some of the PR suggestions would bring it in line with our current style guide. Take or leave what you do or don't like. Approving to keep moving.

PR #7433, opened 2026-08-03
Use `command -v` instead of `which` in the launcher Java lookup

### LINKED ISSUE #7423: Launcher prints "Illegal option --" on Debian 12+ images: `which java` without stderr redirect
## Symptom

On a Debian 12/13-based image, every `nextflow` invocation prints an error before its real
output — including with no arguments at all:

```
$ nextflow
Illegal option --
Usage: nextflow [options] COMMAND [arg...]

$ nextflow -v
Illegal option --
nextflow version 26.04.6.12646
```

The run itself is unaffected, so it is easy to miss. It is not easy to miss if anything
parses the output: it cost me a version-reporting script that took the first line of
`nextflow -v` and recorded `Illegal option --` as the version.

## Cause

`bash -x` traces it to the launcher's Java lookup:

```
++ /usr/bin/which --tty-only --read-alias --read-functions --show-tilde --show-dot java
Illegal option --
```

nextflow line 319:

```bash
JAVA_CMD="$(which java)" || JAVA_CMD=java
```

Debian and Ubuntu have long shipped a `which` **shell function** that pipes aliases and
functions into `/usr/bin/which` with GNU-which flags:

```bash
which () { ( alias; declare -f ) | /usr/bin/which --tty-only --read-alias --read-functions --show-tilde --show-dot $@ }
```

Debian 12 replaced GNU which with debianutils' implementation, which accepts only `[-as]`
and rejects those flags. The function was not updated, so any script calling `which`
through it gets the error on stderr. `$(...)` captures stdout, the error escapes to the
terminal, and the launcher continues.

This affects any Debian 12+ base — including `condaforge/miniforge3`, which is where I
hit it.

## Suggested fix

The launcher already uses `command -v` in seven other places (lines 72, 102, 104, 128,
130, 147, 149). Line 319 is the odd one out:

```diff
-        JAVA_CMD="$(which java)" || JAVA_CMD=java
+        JAVA_CMD="$(command -v java)" || JAVA_CMD=java
```

`command -v` is POSIX, is a shell builtin, cannot be shadowed by a distro function, and
matches the rest of the file.

Line 298 has the same call but is already immune, because it redirects:

```bash
NXF_BIN=$(which "$0" 2>/dev/null)
```

so `2>/dev/null` on line 319 would also suppress it — though `command -v` seems the
better fix, since it removes the dependency on an external `which` entirely rather than
hiding its output.

## Environment

- nextflow 26.04.6 build 12646
- `condaforge/miniforge3` (Debian trixie), running under apptainer
- `/usr/bin/which` from debianutils: `Usage: /usr/bin/which [-as] args`

pditommaso:
Fixes #7423

On Debian 12+ based images (`condaforge/miniforge3`, etc.) `which` is a shell function that pipes aliases and functions into `/usr/bin/which` with GNU-which flags:

```bash
which () { ( alias; declare -f ) | /usr/bin/which --tty-only --read-alias --read-functions --show-tilde --show-dot $@ }
```

Debian 12 replaced GNU which with the debianutils implementation, which accepts only `[-as]` and rejects those flags, so it prints `Illegal option --` on stderr. Bash exports shell functions through the environment, so the launcher inherits it. `$(...)` captures only stdout, so the error leaks to the terminal on every invocation:

```
$ nextflow -v
Illegal option --
nextflow version 26.04.6.12646
```

Harmless for the run itself, but it breaks anything parsing the output — e.g. a script taking the first line of `nextflow -v` as the version.

`command -v` is POSIX, is a shell builtin, cannot be shadowed by a distro function, and matches the seven other lookups already in this file (lines 72, 102, 104, 128, 130, 147, 149). Exit-status semantics are unchanged, so the `|| JAVA_CMD=java` fallback still fires when no `java` is on `PATH`.

The other `which` call, at line 298, already redirects stderr and is unaffected.

Credit to @rec3141 for the report and the diagnosis.

### DIFF
--- a/nextflow
+++ b/nextflow
@@ -295,7 +295,7 @@ NXF_CLASSPATH=${NXF_CLASSPATH:-''}
 
 # Determine the path to this file
 if [[ $NXF_PACK = dist ]]; then
-    NXF_BIN=$(which "$0" 2>/dev/null)
+    NXF_BIN=$(command -v "$0")
     [ $? -gt 0 -a -f "$0" ] && NXF_BIN="./$0"
 fi
 
--- a/nextflow
+++ b/nextflow
@@ -316,7 +316,7 @@ if [ ! -x "$JAVA_CMD" ] ; then
     elif [ -x /usr/libexec/java_home ]; then
         JAVA_CMD="$(/usr/libexec/java_home -v 17+ 2>/dev/null)/bin/java" || JAVA_CMD=java
     else
-        JAVA_CMD="$(which java)" || JAVA_CMD=java
+        JAVA_CMD="$(command -v java)" || JAVA_CMD=java
     fi
 fi
 


bentsherman (approved):

pditommaso:
One semantic difference worth recording, not a blocker for this PR.

`command -v` resolves shell functions, aliases and builtins *before* searching `PATH`; `which` only searches `PATH`. So it is the same class of thing as the bug being fixed here: if a `java` shell function were exported into the environment (rare, but module/conda-style wrappers do export functions, and bash propagates them through the environment), `command -v java` returns the bare string `java` instead of a binary path.

The consequence is mild. `resolve_link java` fails its `[[ ! -f $1 ]]` test and `exit 1` stays confined to the `$( )` subshell, so `JAVA_KEY` ends up as a bare directory path, `[ -f "$JAVA_KEY" ]` fails, and the launcher just re-probes the Java version instead of using the cache. That is exactly what already happens whenever the `|| JAVA_CMD=java` fallback fires, so there is no new failure mode — only a lost cache hit in an already-unusual environment.

If we ever want to close that gap, `type -P java` forces a `PATH`-only lookup and cannot be shadowed by a function, alias or builtin. It is a bashism, but the launcher is `#!/usr/bin/env bash` and already uses `[[ ]]` and `&>`, so it costs nothing portability-wise.

Also worth noting for the record: the seven existing `command -v` calls cited in the description are existence checks (`&>/dev/null`), not path captures, so the consistency argument applies a little less directly to these two lines than to those. Still a good reason to prefer it — I'd keep the current change as-is.

PR #7439, opened 2026-08-04
docs: Move containuer runtimes into separate pages

bentsherman:
This PR moves the container runtimes into separate pages.

While I'm trying to stick to just the mechanical refactor, in this case I had to do some cleanup because these container docs were so out of date

I'm happy to improve these pages further if you see any low-hanging fruit. Either way, these pages probably deserve a deeper review in the future, now that they have more room to breathe.


christopher-hakkaart (approved):
Added suggestions #7450

PR #7440, opened 2026-08-04
Update CI bucket to s3://nextflow-oss-ci

bentsherman:
Test infra is moving to a new AWS account with a new bucket. This replaces every `s3://nextflow-ci` reference with `s3://nextflow-ci-oss` across the integration (`tests/` and `tests-v1/`) and validation (`validation/`) suites, plus two nf-k8s unit-test fixture strings.

## Follow-ups (not in this PR)

These are separate from the bucket rename and need infra decisions:

- **Batch queue names** — `process.queue = 'nextflow-ci'` in `validation/awsbatch.config` and `validation/awsbatch-unstage-fail.config`. These are compute queue names, not buckets. The new AWS account likely needs a different queue name.
- **KMS test** — `tests/checks/publish-s3-kms.nf` (and the `tests-v1` counterpart) hardcode bucket `nf-kms-xyz` and a KMS key ARN in account `195996028523`. This test will fail on the new account unless an equivalent bucket + key are provisioned and the ARN updated.
- **GitHub secrets** — `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` repo secrets must be repointed at the new account.
- **Merge ordering** — the new bucket must exist and be writable by the new credentials before this merges, otherwise the AWS/Fusion/Wave integration jobs break.

Public/external buckets (`s3://rnaseq-nf`, `s3://ngi-igenomes`) are untouched.


pditommaso (approved):

plugins/nf-k8s/src/test/nextflow/k8s/K8sConfigTest.groovy
```diff
@@ -393,13 +393,13 @@ class K8sConfigTest extends Specification {
     def 'should set env and sec context' () {
         given:
         def ctx = [
-                [env: 'FUSION_BUCKETS', value: 's3://nextflow-ci'],
+                [env: 'FUSION_BUCKETS', value: 's3://nextflow-oss-ci'],
```
bebosudo: Let's use `nextflow-ci-oss` please

bebosudo (approved):

plugins/nf-k8s/src/test/nextflow/k8s/K8sConfigTest.groovy
bentsherman: weird that claude didn't update this one. must be because it's only a placeholder here

bentsherman: nevermind, I see what you mean

validation/awsbatch.sh
```diff
@@ -43,17 +43,19 @@ $NXF_CMD run test-complexpaths.nf -resume -c awsbatch.config
 
 $NXF_CMD run test-subdirs.nf -c awsbatch.config
 
-NXF_CLOUDCACHE_PATH=s3://nextflow-ci/cache \
+NXF_CLOUDCACHE_PATH=s3://nextflow-ci-oss/cache \
 $NXF_CMD run nextflow-io/rnaseq-nf \
     -profile batch \
+    -w s3://nextflow-ci-oss/work \
```
bentsherman, line 49: TODO: update rnaseq-nf to use new bucket or remove

in any case it's better to have it pinned here

PR #7441, opened 2026-08-04
docs: Move executors into separate pages

bentsherman:
This PR moves the executors into separate pages. Mostly a mechanical refactor, but I also trimmed and aligned the descriptions a bit

Possible future enhancements (out of scope):
- move executor dropdown under "Runtime reference" (leave an introductory page under "Running pipelines" ?)
- move the AWS Batch / Azure Batch / Google Batch content into the executor pages, reduce "Compute and storage" to just "Storage"
- (related) spin off separate pages for Azure Blob, Google Cloud Storage


christopher-hakkaart (approved):
Looks good - I applied the style guide in #7443 

Take or leave what you agree or disagree with.

PR #7442, opened 2026-08-04
Fix description for process config options with method overloads

### LINKED ISSUE #173: Config options with multiple overloads lose their description, or gain it only intermittently
### Summary

Follow-up to #172, which fixed the *type* shown for config options with several accepted types. The **description** for those options has the same order-dependence and is still wrong: it is taken from whichever overload is enumerated first, and only one overload per directive is annotated. Some process directives lose their documentation entirely, and others gain or lose it between runs of the same build.

### Mechanism

`SpecNode.processScope()` (in `nf-lang`) builds the process scope by walking `ProcessDsl$DirectiveDsl.getDeclaredMethods()`, keeping the one-parameter overloads. Decompiled, it does:

```
if( !map.containsKey(method.getName()) )
    map.put(method.getName(), new Option(annotatedDescription(method, ""), types));
map.get(method.getName()).types.add(method.getParameterTypes()[0]);
```

The type list accumulates across every overload — which is what #172 was able to fix by joining them. The **description is only read from the first overload encountered**, and later overloads contribute nothing but their type.

Exactly one overload per directive carries `@Description`:

```
HAS @Description <- void arch(Map<String, ?>);
no @Description  <- void arch(Map<String, ?>, String);
no @Description  <- void arch(String);

HAS @Description <- void cache(String);
no @Description  <- void cache(Boolean);
```

So the description survives only when the annotated overload happens to be enumerated first.

### Impact

Two distinct symptoms, both from that one line.

**1. Descriptions permanently lost.** Where the annotated overload is reliably not first, the option never has documentation. `arch`, `disk` and `publishDir` returned an empty description in 6 out of 6 runs. `arch` is annotated on `arch(Map)` but `arch(String)` is enumerated first, so the text is never used.

**2. Descriptions appear and disappear between runs.** For the rest, enumeration order is not stable, so the same build gives different answers. Over 6 runs of the same `language-server-all.jar`, requesting completion inside a `process { }` block:

| option | description present | description empty |
|---|---|---|
| `accelerator` | 1 | 5 |
| `cache` | 1 | 5 |
| `clusterOptions` | 1 | 5 |
| `pod` | 1 | 5 |
| `shell` | 1 | 5 |
| `arch` | 0 | 6 |
| `disk` | 0 | 6 |
| `publishDir` | 0 | 6 |

Non-overloaded directives are unaffected — `cpus` has a single annotated overload and returned its description in 6 of 6 runs.

Hover is affected the same way, since `ConfigHoverProvider` reads the same `option.description()`.

Note that `Class.getDeclaredMethods()` order was **stable** across 5 runs when called in isolation on a bare JVM, so the instability appears to depend on the surrounding load rather than on reflection order alone. The `synchronized` added to `ConfigSpecFactory.defaultScopes()` in #172 did not remove it.

### Reproduction

Workspace with `main.nf` (any valid script) and:

```groovy
process {
    cpus = 4

}
```

Session: `initialize` → `initialized` → `didChangeConfiguration` → `didOpen main.nf` → `didOpen nextflow.config` → completion on the blank line inside `process {`. Read the `documentation` field of the returned items and repeat with a fresh server process. Opening the `.nf` file first matters — config-only sessions were stable in my earlier testing on #172.

### Suggested fix

When merging overloads, do not let the first one decide the description — prefer a non-empty one. Mirroring #172's shape, something like keeping the existing description only if the incoming one is empty:

```java
var existing = map.get(name);
var description = annotatedDescription(method, "");
if( existing.description().isEmpty() && !description.isEmpty() )
    // adopt the annotated description
```

Making the enumeration order deterministic would fix symptom 2 but not symptom 1, since `arch`, `disk` and `publishDir` would still deterministically pick the unannotated overload.

### Environment

- language-server `main` @ fd0be65 (i.e. including the #172 fix), `nf-lang` 26.04.6
- JDK 17 (Corretto 17.0.19), Linux x86_64

Found by the JAR-versus-native-image response equivalence check added in #146: with the #172 fix in place the type comparison is now stable, and this is what the diff surfaced next.

bentsherman:
Fix https://github.com/nextflow-io/language-server/issues/173

This PR updates SpecNode to handle method overloads when loading process config from ProcessDsl

A process directive can have multiple overloads, but only one overload is expected to specify a description. So SpecNode needs to extract the description from whichever method defines it.

### DIFF
--- a/modules/nf-lang/src/main/java/nextflow/config/spec/SpecNode.java
+++ b/modules/nf-lang/src/main/java/nextflow/config/spec/SpecNode.java
@@ -80,9 +80,9 @@ public sealed interface SpecNode {
      * process directives.
      *
      * Directives with multiple method overloads are treated as
-     * options with multiple supported types. Method overloads with
-     * multiple parameters are ignored because they are not supported
-     * in the configuration.
+     * options with multiple supported types. Only one overload needs
+     * to provide a description. Method overloads with multiple parameters
+     * are ignored because they are not supported in the configuration.
      */
     private static SpecNode processScope() {
         var description = """
@@ -94,13 +94,14 @@ public sealed interface SpecNode {
         for( var method : ProcessDsl.DirectiveDsl.class.getDeclaredMethods() ) {
             if( method.getParameters().length != 1 )
                 continue;
-            if( !children.containsKey(method.getName()) ) {
-                var desc = annotatedDescription(method, "");
-                children.put(method.getName(), new Option(desc, new ArrayList<>()));
-            }
-            var option = (Option) children.get(method.getName());
-            var paramType = method.getParameterTypes()[0];
-            option.types.add(paramType);
+            var name = method.getName();
+            var desc = annotatedDescription(method, "");
+            var option = (Option) children.get(name);
+            if( option == null )
+                children.put(name, option = new Option(desc, new ArrayList<>()));
+            else if( option.description().isEmpty() && !desc.isEmpty() )
+                children.put(name, option = new Option(desc, option.types()));
+            option.types.add(method.getParameterTypes()[0]);
         }
         return new Scope(description, children);
     }
--- a/modules/nf-lang/src/main/java/nextflow/config/spec/SpecNode.java
+++ b/modules/nf-lang/src/main/java/nextflow/config/spec/SpecNode.java
@@ -92,9 +92,9 @@ public sealed interface SpecNode {
      * set of process directives.
      *
      * Directives with multiple method overloads are treated as
-     * options with multiple supported types. Method overloads with
-     * multiple parameters are ignored because they are not supported
-     * in the configuration.
+     * options with multiple supported types. Only one overload needs
+     * to provide a description. Method overloads with multiple parameters
+     * are ignored because they are not supported in the configuration.
      *
      * @param description
      */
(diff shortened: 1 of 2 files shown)


bentsherman:
Related: https://github.com/nextflow-io/language-server/pull/146

I am ironing out any non-deterministic LSP behavior so that I can reliably verify native vs JVM language server behavior

pditommaso (approved):
Nice fix — thanks Ben! The logic is spot on: because only one overload carries the `@Description` and `getDeclaredMethods()` has no guaranteed ordering, an overloaded directive could end up with an empty description depending on which overload was seen first. Upgrading the `Option` when a non-empty description shows up (while reusing the same `types` list so nothing is lost) handles both iteration orders correctly. The added test on `arch`/`cache`/`clusterOptions`/`disk`/`publishDir` is a good regression guard. 👍

Two minor, non-blocking notes:

1. **Needs a rebase** — `master` has since refactored `processScope()` into `directiveScope(String description)` (now shared by both the `process` and `agent` scopes), so the PR currently shows as conflicting. The loop body is essentially unchanged, so the fix should port over verbatim into `directiveScope`.

2. **Optional readability nit** — the assignment-as-expression in `children.put(name, option = new Option(...))` is a touch dense. Totally fine to leave as-is, but splitting it into two statements might read a little more clearly. Your call.

Approving on the merits — just needs the rebase before merge. 🙏

PR #7452, opened 2026-08-06
Treat a missing `k8s` config scope as an empty one

pditommaso:
A pipeline that selects the Kubernetes executor without declaring a `k8s` block aborts before running anything:

```
Cannot invoke &#34;java.util.Map.get(Object)&#34; because &#34;opts&#34; is null
	at nextflow.k8s.K8sConfig.(K8sConfig.groovy:220)
	at nextflow.k8s.K8sExecutor.memoizedMethodPriv$getK8sConfig(K8sExecutor.groovy:66)
```

`K8sExecutor.getK8sConfig` passes `session.config.k8s` straight to `K8sConfig(Map)`; an absent scope navigates to `null`, and the constructor dereferences it on its first line.

## Why this is a bug, not a missing requirement

Every `k8s` option either has a default or is legitimately unset, so a config with no `k8s` block is a valid configuration. The codebase already says so in two places:

- `K8sConfig()` (no-arg) delegates to `this(Collections.emptyMap())`;
- `K8sDriverLauncher:286` guards the same construction with `config.k8s instanceof Map ? new K8sConfig(config.k8s as Map) : new K8sConfig()`.

The executor path is the one caller that does neither.

## Fix

Guard the scope lookup at the call site, matching how every other config scope is resolved (`Session`, `DefaultObserverFactory`, `ExecutorFactory`, `AwsOptions`, `WaveClient`, …):

```groovy
new K8sConfig( session.config.k8s as Map ?: Collections.emptyMap() )
```

## Tests

`K8sExecutorTest` — `getK8sConfig()` with a session whose config has no `k8s` key, which is the reported path. It fails with the original NPE when the fix is reverted. Full `nf-k8s` suite passes.

### DIFF
--- a/plugins/nf-k8s/src/main/nextflow/k8s/K8sConfig.groovy
+++ b/plugins/nf-k8s/src/main/nextflow/k8s/K8sConfig.groovy
@@ -216,7 +216,11 @@ class K8sConfig implements ConfigScope {
         this(Collections.emptyMap())
     }
 
-    K8sConfig(Map opts) {
+    K8sConfig(Map options) {
+        // a missing `k8s` scope navigates to null, which is the same thing as an empty one -- the
+        // no-arg constructor above already says so. Tolerate it here rather than at each call site,
+        // so a config with no `k8s` block cannot NPE its way out of an executor.
+        final opts = options ?: Collections.emptyMap()
         autoMountHostPaths = opts.autoMountHostPaths as boolean
         cleanup = opts.cleanup as Boolean
         client = opts.client as Map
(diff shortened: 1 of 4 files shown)


bentsherman:
The established pattern for config classes is to guard at the call site (see for example `DefaultObserverFactory`). So I would rather stay consistent with that.

pditommaso:
Agreed — `as Map ?: Collections.emptyMap()` at the call site is what `Session`, `ExecutorFactory`, `AwsOptions` and `WaveClient` all do, and the constructor-side null tolerance had no precedent. Moved the guard into `K8sExecutor.getK8sConfig()` and restored `K8sConfig(Map opts)` unchanged.

The `K8sConfig(null)` spec went away with it; the `K8sExecutorTest` case covering the reported path (session config with no `k8s` key) remains and still fails with the NPE when the fix is reverted.

PR #7453, opened 2026-08-07
Use the Seqera Platform default workspace when none is set locally

ewels:
## Why

Nextflow's local workspace selection — `nextflow auth config`, `tower.workspaceId`, the `-workspace` flag — exists largely because Seqera Platform historically had **no** concept of a default workspace. Every user had to tell Nextflow, on every machine, something Platform could not tell it.

Platform now has that concept: since API v1.173.0 the `GET /user-info` response carries a top-level `defaultWorkspaceId` (a per-user preference, falling back to a global system default, and access-validated server-side).

This PR teaches Nextflow to pick it up. **Local configuration is untouched and still wins.** The only behaviour that changes is the fallback: where Nextflow previously jumped straight to the personal workspace, it now asks Platform first.

```
-workspace <name>  →  tower.workspaceId  →  TOWER_WORKSPACE_ID  →  Platform account default  →  personal
```

The practical effect: an org user who never ran `nextflow auth config` gets their runs in the right workspace instead of silently in Personal.

## Discoveries along the way

Three things surfaced that were not part of the original ask, and two of them mattered more than the feature:

**The workspace was resolved in more places than expected.** Making the observer aware of the default is not enough — `TowerFusionToken`, nf-wave's `TowerConfig` and `SeqeraExecutor` each resolve the workspace independently. Left alone, a run would be *reported* into the Platform default while its **Fusion licence** was requested against Personal (a hard abort, surfacing as "Unable to validate Fusion license"), its **Wave registry credentials** looked up in Personal, and its scheduler run created in Personal. This hits precisely the users the feature targets. Fixed by resolving once and publishing the result, so every subsystem reads the same value.

**`nextflow launch` never honoured `TOWER_WORKSPACE_ID`.** `readConfig()` reads the config file only. Harmless before — the env var was ignored and you got Personal — but once a Platform default existed, ignoring it would have sent runs to a *different real workspace*. Fixed by routing the launch path through the same precedence ladder as everything else.

**A best-effort lookup was inheriting a run-critical retry budget.** The new call sits in `Session.init()`, before the pipeline script is parsed, and initially used the telemetry retry policy: 10 attempts with exponential backoff. An unreachable Platform could have stalled the start of *every* run for minutes before falling back to Personal anyway. Now bounded to a single attempt with a short timeout.

<details>
<summary><b>Implementation and decisions</b></summary>

### Resolution policy lives in one place

`PlatformHelper.getEffectiveWorkspaceId(opts, env, Supplier<String>)` is the single definition of "local wins, otherwise use the Platform default". The Platform lookup is passed in as a supplier so `PlatformHelper` stays free of I/O and of any dependency on the API client — nf-wave consumes it and has no `TowerClient`. A sibling `isPlatformRun(env)` gives the `TOWER_WORKFLOW_ID` rule one definition instead of five.

A Platform-driven run is authoritative: when `TOWER_WORKFLOW_ID` is set, the environment decides and the account default is never consulted.

### Publishing the resolved value

`TowerFactory.create()` writes the resolved workspace into `session.config.tower.workspaceId`. All four consumers already read through `PlatformHelper.getWorkspaceId(session.config.tower, env)`, so they are fixed with no edits at the consumers. There is precedent: `WaveFactory` already writes resolved facts back into `session.config`.

It publishes **only** when the user set nothing — it never overwrites a configured value, and never invents a config key. A `log.info` names the workspace, so a run landing somewhere the user did not configure is explained rather than mysterious.

Resolution stays in `TowerFactory` rather than moving to `Session.init`: it is correctly gated behind `isEnabled()`, so a plain local run that merely has a token in `seqera-auth.config` pays nothing.

### Bounding the lookup

`TowerConfig.forLookup()` produces a config with one attempt and a 10s timeout, used for the session-init lookup. The CLI commands deliberately do **not** use it: `nextflow launch` and `nextflow auth status` make other calls through the shared client in the same flow that keep the default retry policy, so bounding just this one call buys no real guarantee while costing an extra `/user-info` fetch and TLS handshake (a second client has its own memo and connection pool).

### Other changes

- `TowerClient.describeUser()` is memoized so `getUserInfo()` and `getDefaultWorkspaceId()` share one round-trip — `nextflow auth status` drops from 3 identical `/user-info` calls to 1. `listUserWorkspacesAndOrgs` likewise (`launch -workspace NAME` fetched it twice).
- `nextflow auth status` reports source `platform` when the value came from the account default; `nextflow auth config` no longer labels the no-selection case "Personal workspace" when a default exists.
- Older/Enterprise Platform versions that predate the field return `null`, which degrades cleanly to the personal workspace. The lookup swallows failures — it must never abort a run.

### Behaviour change worth flagging

`-workspace <name>` now takes priority over `tower.workspaceId`, where previously the config setting won. An explicit CLI flag beating a config file is the conventional order, and the full precedence is now documented and covered by a test — but it is a user-visible change.

### Notes for reviewers

- `PlatformHelper.getWorkspaceId`'s javadoc documents the session-config coupling, since that is the function every affected consumer calls.
- Also fixed a pre-existing trap in `TowerRetryPolicy`: defaults were resolved with the elvis operator, and `0` is falsy in Groovy, so `maxAttempts = 0` ("don't retry") silently became 10 and `jitter = 0` became `0.25`. Invalid values now warn and fall back rather than aborting a run, since configs relying on the old silent behaviour work today.

</details>

## Follow-ups (deliberately not in this PR)

- The same falsy-elvis trap exists in six sibling retry classes (nf-wave, nf-seqera, nf-k8s, nf-google, nf-azure, and the SRA datasource). Routing all seven through a shared helper is its own change.
- Longer term the resolved workspace arguably belongs on `session.workflowMetadata.platform`, which already carries `workflowId` and `workflowUrl` — a typed channel instead of a config-map key. Larger blast radius across three plugins.

## Testing

Unit tests cover the precedence ladder, the publication (including that it is a no-op when the user configured a workspace), the bounded lookup config, and the `auth status` display. Full suites pass for nf-tower, nf-wave, nf-seqera and the affected core tests.

Not yet exercised against a live Platform account with a default workspace set — worth a manual check before merge that Wave and Fusion land in the same workspace as the run.

---
_Generated by [Claude Code](https://claude.ai/code/session_01P9R8okDLEDkubhnMrVsWdJ)_


ewels:
Ran the [skill](https://gist.github.com/geoffreylitt/a29df1b5f9865506e8952488eac3d524) mentioned by @bentsherman in Slack to generate this explainer HTML:

[2026-08-07-explanation-seqera-default-workspace.zip](https://github.com/user-attachments/files/30840932/2026-08-07-explanation-seqera-default-workspace.zip)

ewels:
## Blocked: `/user-info.defaultWorkspaceId` doesn't mean what this PR assumes

CI is red on all 12 cloud/integration suites, and the cause is not a flake or a misconfigured CI account — it's the premise of the PR. Flagging here for a Platform-side decision before this goes further.

### What this PR tried to do

When no workspace is set locally (`tower.workspaceId` / `TOWER_WORKSPACE_ID`), fall back to the user's default workspace configured in Seqera Platform, so `nextflow run -with-tower` reports into the workspace the user expects.

### Why it can't work against today's API

`GET /user-info` populates `defaultWorkspaceId` from `UserServiceImpl.getUserDefaultWorkspaceId(userId)`, which resolves in two steps:

1. the per-user preference (`user.options.defaultWorkspaceId`), if the user still has access to it;
2. **otherwise the deployment-wide `tower.admin.default-workspace-id`** — documented in `ENVIRONMENT-VARIABLES.md` as *"ID of the Tower workspace ID in which to land users on logging in"*, and mapped to `${TOWER_EVAL_WORKSPACE_ID}` in `application-seqera.yml`.

Both meanings arrive in the same field. A client cannot distinguish *"this user chose this workspace"* from *"this is where the UI lands users after login"*, so adopting the value for run reporting is wrong for every user who has not set a personal default.

There is a second issue independent of the first: the fallback is gated on `hasWorkspaceAccess(userId, id)` — **visibility, not launch permission**. A user who can see a workspace but not launch in it still gets it returned as their default.

### What CI actually shows

The Nextflow CI service account resolves to `40230138858677 [community / showcase]`, and `POST /trace/create?workspaceId=40230138858677` then returns 403:

```
Using default workspace configured in your Seqera Platform account: 40230138858677 [community / showcase]
ERROR ~ Abort exception produced when notifying an event -
  Cannot run in Seqera Platform workspace 40230138858677 [community / showcase]:
  your access token does not have permission to launch runs there.
  This is the default workspace configured in your Seqera Platform account,
  which is used when no workspace is set locally.
```

That is the *default* case for a cloud user, not a broken account — which is why fixing CI alone would not make this PR correct.

### What would unblock the client side

Either (or both):

1. **Expose the per-user preference distinctly** — a separate field, or a flag on the existing one marking the value as user-set rather than the system landing workspace.
2. **Gate the fallback on launch capability** rather than visibility.

Until one of those lands, there is no way for Nextflow to implement adopt-the-default correctly.

### Aside, unrelated to the API shape

On cloud, `TOWER_EVAL_WORKSPACE_ID` appears to point at the public community showcase workspace, so any user without a personal default currently "lands" there. Probably worth a separate sanity check on whether that is intended.

### Status of this branch

No further commits are planned pending that decision. Unit tests are green (452 `nf-tower`, 27 `nextflow.platform`); the cloud suites stay red for the reason above.

Worth noting the diagnostics half of this branch stands on its own regardless of what happens to the workspace-adoption behaviour — the error explanation and the `id [org / workspace]` naming are what made this diagnosable at all. The log line above used to read `40230138858677` followed by a raw HTTP status dump.

<details>
<summary>Source references</summary>

- `modules/platform-user/impl/src/main/groovy/io/seqera/tower/service/UserServiceImpl.groovy` — `getUserDefaultWorkspaceId`
- `modules/platform-user/impl/src/main/groovy/io/seqera/tower/controller/UserController.groovy` — `/user-info` handler
- `modules/platform-infra/impl/src/main/groovy/io/seqera/tower/service/property/PropertyServiceImpl.groovy` — `@Value('${tower.admin.default-workspace-id}')`
- `tower-config/src/main/resources/application-seqera.yml`, `ENVIRONMENT-VARIABLES.md`

Caveat on confidence: the method body was reconstructed from GitHub code-search fragments rather than read verbatim, and the deployed value of `TOWER_EVAL_WORKSPACE_ID` is not visible from here. The 403 against `community / showcase` is directly observed in the CI logs, so the behaviour is confirmed regardless of the exact source text.

</details>

---
_Generated by [Claude Code](https://claude.ai/code)_

ewels:
Now discussing internally on slack.

PR #7466, opened 2026-08-10
Bound Seqera executor scheduler requests with a request timeout

### LINKED ISSUE #7435: nf-seqera: missing HTTP request timeout lets TaskPollingMonitor's poll thread hang forever on a stalled scheduler response
## Summary
Running `nf-core/sarek` through the `seqera` executor (`nf-seqera` plugin) against a `sched` VM backend cluster, the run stalls indefinitely shortly after a large batch of tasks is submitted. The Nextflow debug log shows the `[Task monitor]` thread — driven by `TaskPollingMonitor` — simply stops emitting any log line at all partway through the run, while other independent components in the same JVM keep running and logging normally. Tracing the call path, this looks like `TaskPollingMonitor`'s single poll thread getting stuck forever inside a blocking HTTP call that has no timeout configured, rather than a network/auth/backend outage.

**Environment:** Nextflow `26.04.6`, plugin `nf-seqera@0.21.1`, pipeline `nf-core/sarek`, `-profile test_full`, `seqera` executor against a `sched` VM backend (Azure).

## Symptom in the log

`TaskPollingMonitor` starts with a fixed poll interval:
```
DEBUG n.processor.TaskPollingMonitor - Creating task monitor for executor 'seqera' > capacity: 1000; pollInterval: 10s; dumpInterval: 5m
```

For the whole run, every ~10s, it produces a repeating pair of `[Task monitor]` lines per tracked task, e.g. for a normally-running task:
```
14:07:51.663 [Task monitor] DEBUG io.seqera.sched.client.SchedClient - Sched status response: statusCode=200; body={"taskState":{"id":"tsk-<id>:azr-vm","status":"RUNNING","name":"NFCORE_SAREK:SAREK:FASTQC (HCC1395N-1)", ...}}
14:07:51.665 [Task monitor] DEBUG i.seqera.executor.SeqeraTaskHandler - [SEQERA] checkIfCompleted status=RUNNING
```

This is stable until a batch of ~24 CPU-heavy alignment tasks (`BWAMEM1_MEM`, ~24 cores / 30 GB each) is submitted in one shot. The monitor keeps confirming them `SUBMITTED` on schedule for a couple more cycles — the **last** such pair recorded is:
```
14:11:52.374 [Task monitor] DEBUG io.seqera.sched.client.SchedClient - Sched status response: statusCode=200; body={"taskState":{"id":"tsk-<id>:azr-vm","status":"SUBMITTED","name":"NFCORE_SAREK:SAREK:FASTQ_PREPROCESS_GATK:FASTQ_ALIGN:BWAMEM1_MEM (HCC1395T)", ...}}
14:11:52.375 [Task monitor] DEBUG i.seqera.executor.SeqeraTaskHandler - [SEQERA] checkIfRunning taskId=tsk-<id>:azr-vm; status=SUBMITTED
```

**From that point on, no `[Task monitor]` log line of any kind appears again** for the rest of the captured run — not even the `"<<< barrier arrives ... terminating tasks monitor poll loop"` line that would be logged if the poll loop had exited (cleanly, or via an uncaught exception escaping its outer try/catch). Meanwhile a completely separate thread — the keep-alive heartbeat — keeps firing on its own ~60s cadence with clean responses throughout the same window:
```
14:12:04.319 [Seqera-batch-submitter] DEBUG i.s.executor.SeqeraBatchSubmitter - [SEQERA] Sending keep-alive for run run-<id>:azr-vm
14:12:04.576 [Seqera-batch-submitter] DEBUG io.seqera.sched.client.SchedClient - Sched status response: statusCode=200; body={}
```
... (repeats every ~60s, unbroken, for the next ~43 minutes) ...
```
14:55:05.410 [Tower-thread] DEBUG io.seqera.http.HxClient - Received 401 status for auth key <key>, attempting token refresh
14:55:05.675 [ForkJoinPool.commonPool-worker-2] DEBUG io.seqera.http.HxTokenManager - JWT token refresh completed for key <key>
14:55:18.944 [Seqera-batch-submitter] DEBUG i.s.executor.SeqeraBatchSubmitter - [SEQERA] Sending keep-alive for run run-<id>:azr-vm
14:55:19.943 [Seqera-batch-submitter] DEBUG io.seqera.http.HxClient - Received 401 status for auth key <key>, attempting token refresh
14:55:20.203 [ForkJoinPool.commonPool-worker-2] DEBUG io.seqera.http.HxTokenManager - JWT token refresh completed for key <key>
14:55:20.290 [Seqera-batch-submitter] DEBUG io.seqera.sched.client.SchedClient - Sched status response: statusCode=200; body={}
14:56:20.301 [Seqera-batch-submitter] DEBUG i.s.executor.SeqeraBatchSubmitter - [SEQERA] Sending keep-alive for run run-<id>:azr-vm
14:56:20.587 [Seqera-batch-submitter] DEBUG io.seqera.sched.client.SchedClient - Sched status response: statusCode=200; body={}
```
(log capture ends here)

So over a ~44 minute window: the keep-alive heartbeat fires cleanly every ~60s (~44 times), and even the access token expiring mid-stall and being refreshed goes through without issue — but the `[Task monitor]` thread, whose job is to poll every 10s, produces **zero** log output the entire time. That isolates the freeze specifically to the monitor thread / its blocking call, not to the HTTP client, auth, or scheduler connection in general.

## Root cause (from reading `TaskPollingMonitor` + the `nf-seqera`/`sched-client` call chain)

`TaskPollingMonitor.pollLoop()` runs on a single dedicated thread (`"Task monitor"`). Each iteration calls `checkAllTasks(tasks)`, which iterates the running-task queue **sequentially** and calls `checkTaskStatus(handler)` for each one:
```groovy
for (int i=0; i<queue.size(); i++) {
    final handler = queue.get(i)
    try {
        checkTaskStatus(handler)   // -> handler.checkIfRunning() / handler.checkIfCompleted()
    }
    catch (Throwable error) {
        handler.kill()
        handleException(handler, error)
    }
}
```
The `try/catch` only protects against a *thrown* exception — there is no timeout around the call itself. `pollLoop()`'s own bounded wait (`await()`, a timed `Condition.await(delta, ...)`) is only reached *after* `checkAllTasks()` returns, so it offers no protection here either.

For the `seqera` executor, `checkIfRunning()`/`checkIfCompleted()` (in `SeqeraTaskHandler`) delegate to `SchedClient.describeTask()`, which calls into `HxClient.sendWithRetry()` — a **single synchronous call** to JDK 11+'s built-in `java.net.http.HttpClient`, executed directly on the calling ("Task monitor") thread.

JDK's `HttpClient` has **no default read/response timeout** — only a connect timeout, which `SchedClient` does set (`connectTimeout(Duration.ofSeconds(10))`). A per-request response timeout is applied only when `HttpRequest.timeout(...)` is explicitly set, which `HxClient`/`SchedClient` only does when `SchedClientConfig.requestTimeout` is non-null:
```java
if (config.getRequestTimeout() != null) {
    builder.timeout(config.getRequestTimeout());
}
```
`nf-seqera`'s executor builds its `SchedClientConfig` with `.endpoint/.platformUrl/.accessToken/.refreshToken/.retryConfig` only — **`requestTimeout` is never set**, so it stays `null` and no per-request timeout is ever applied. (This is directly visible in the log: `SchedClientConfig[endpoint=..., requestTimeout=null, ...]` at client creation.)

Net effect: if the scheduler backend ever accepts the connection but stalls before finishing a response for a single status check, `httpClient.send()` blocks **indefinitely** on that one call. It never throws, so the surrounding `try/catch` never fires, `checkAllTasks()` never returns, and the single-threaded poll loop is stuck forever — silently, with no error logged, no further status checks for *any* task (not just the stalled one), and no way to recover without restarting the run. This matches the observed symptom exactly, and also explains why unrelated threads (the keep-alive heartbeat, JWT refresh) kept working fine: they don't go through this stuck call.

## Suggested fix
- `nf-seqera` should always set a sane `requestTimeout` when constructing `SchedClientConfig`, so a stalled server response can't block the poll thread forever (a JDK `HttpClient` call with `HttpRequest.timeout(...)` set throws an `HttpTimeoutException`, which the existing per-task `try/catch` in `TaskPollingMonitor` would then handle correctly).
- Longer term, `TaskPollingMonitor.checkAllTasks()` could bound each `checkTaskStatus()` call with its own watchdog/timeout independent of the executor implementation, so a single misbehaving `TaskHandler` (in any executor, not just `seqera`) can't freeze the entire poll loop for every task.

## Logs
A full Nextflow debug log covering the stall is available and can be shared privately on request.

pditommaso:
Closes #7435.

## What this fixes

The `Task monitor` thread can block indefinitely inside `SchedClient.describeTask()` on a stalled scheduler response. Only the connect phase was bounded — a fixed 10s inside `sched-client` — and nothing bounded the wait for the response body once the connection was established. Because `TaskPollingMonitor` polls every task from that single thread, one hung poll stalls `checkAllTasks()` for the whole session: no task in the run advances.

## What changes

A new config scope holding both transport bounds:

```groovy
seqera.executor.httpClient.connectTimeout = '10 sec'   // reaching the scheduler
seqera.executor.httpClient.requestTimeout = '45 sec'   // waiting for its answer
```

`requestTimeout` is the one that fixes #7435 — `connectTimeout` only ever bounded reaching the service, and it is exposed here because it was previously a hardcoded literal with no knob at all. `0 sec` restores unbounded responses. Defaults preserve today's behaviour on the connect side and bound the response side for the first time.

The scope follows `wave.httpClient`, which exists for exactly this reason; `nf-tower` reaches the same shape with a flat `httpConnectTimeout`/`httpReadTimeout` pair. Both keep transport settings out of `retryPolicy`, and so does this.

## How the two interact with retries

`sched-client` decides retries per request (seqeralabs/sched#937), so the two timeouts are retried differently:

| | bounds | on expiry | retried? |
|---|---|---|---|
| `connectTimeout` | establishing the TCP/TLS connection | `HttpConnectTimeoutException` | **every** request — it proves nothing was delivered, so state-creating POSTs too |
| `requestTimeout` | waiting for the response, once connected | `HttpTimeoutException` | idempotent requests only |

`describeTask` is a GET and `cancelTask` a DELETE, so both take the idempotent branch: a timed-out poll is re-sent under `seqera.executor.retryPolicy` rather than propagating into `checkAllTasks()`, where it would fail the task. `createRun` and `createTasks` are POSTs and take the narrow branch, so a timed-out submit fails cleanly instead of minting a duplicate set of task ids that nothing polls or cancels.

**What this costs.** The retries run on the `Task monitor` thread, inside the blocking call — not on the next poll cycle. With the defaults (`45 sec` × `retryPolicy.maxAttempts` of 10, plus ~205s of accumulated backoff) a scheduler that stalls on every attempt holds that thread for roughly 11 minutes on a single task. That is bounded where it previously was not, but it is not free: lower `requestTimeout` or `retryPolicy.maxAttempts` to tighten it.

## Dependencies

`sched-client` 0.73.2 → **0.79.3**, and `lib-httpx` 2.4.0 → **2.6.0**.

The client bump is required for correctness, not for the API. `requestTimeout` existed as far back as 0.73.2, but setting it there silently forced `maxAttempts(1)` and discarded `retryPolicy` for *every* call, so a single timed-out poll would fail the task — the opposite of what this PR is for. The decoupling landed in seqeralabs/sched#937, which is **not** in 0.78.0 or 0.79.1; it merged after 0.79.1 was cut. `connectTimeout` needs seqeralabs/sched#1095 on top. `v0.79.3` is the first release carrying both — verified by checking that both merge commits are ancestors of the tag.

The lib-httpx bump is not cosmetic. `SchedHxClient` expresses the per-request policy above by overriding `shouldRetryOnException(HttpRequest, Throwable)`, an overload added in 2.6.0. Held at 2.4.0 the override binds to nothing and every request falls back to the library default of retrying any `IOException` — which re-sends a timed-out batch submit and duplicates the tasks it creates. Gradle's highest-wins would resolve 2.6.0 transitively anyway, but the declaration and its comment are updated so the constraint is stated rather than relied on by accident.

## Tests

`ExecutorOptsTest` — the `45 sec`/`10 sec` defaults, an explicit value for each, and the two bounds coexisting without displacing one another. A `where:` block covers `0 sec` mapping to unbounded in both the String and `Duration` spellings: `nextflow.util.Duration.asBoolean()` is false at zero, so a truthiness test would quietly turn `requestTimeout = 0.sec` back into the default.

`SeqeraExecutorTest` — both timeouts reach `SchedClientConfig`.

`:plugins:nf-seqera:test` — **257 tests, 0 failures**, against the published `sched-client:0.79.3`. The graph resolves `lib-httpx:2.6.0` and `lib-retry:2.1.0`, and so `failsafe:3.3.2` — the version the plugin compiles against, which is what the old 2.4.0 pin was there to guarantee.


pditommaso:
@claude review

pditommaso:
Addressed in 8e796945a (`[ci fast]`). Suite now 261 tests, 0 failures.

**1 — `connectTimeout` asymmetry at zero: fixed.** `HttpClientOpts` now rejects a non-positive value in the constructor with a message naming `seqera.executor.httpClient.connectTimeout`, and the constraint is documented in both the option description and `seqera.mdx`. One correction: the exception came from `SchedClientConfig.Builder` (added in seqeralabs/sched#1095), not the JDK `HttpClient.Builder` — same symptom, different source.

**3 — wiring tests that don't test the wiring: fixed, and the first attempt wasn't enough.** Extracted `clientConfig()` from `createClient()` so the test calls production code, then deleted `.connectTimeout(...)` to check the fix bit — the suite stayed green. Extraction alone did nothing, because no spec asserted the `connectTimeout` on the built `SchedClientConfig`. Added one (plus a defaults spec), re-ran the deletion, confirmed it now fails. Note the defaults spec cannot detect deletion by itself, since `SchedClientConfig` falls back to 10s regardless.

**2 — 45s ceiling on a 100-task submit: not changed, needs data.** Facts confirmed. Choosing a different default without p99 submit latency for the largest batch would be guessing. Worth adding that the batch size is already tunable via `NXF_SEQERA_TASK_PER_REQUEST`, so there is an escape hatch if 45s proves tight in practice.

**4 — docs ordering: declined, the premise does not hold.** The nested scopes are not alphabetical today: the order is `machineRequirement`, `schedulingRequirement`, `httpClient`, `retryPolicy`, and `schedulingRequirement` preceding `retryPolicy` predates this PR. Moving `httpClient` first still would not make it alphabetical, so there is no convention to restore.

On what could not be verified from the review checkout: the 261 tests ran against the published `sched-client:0.79.3`, and the resolved graph is `lib-httpx:2.6.0` → `lib-retry:2.1.0` → `failsafe:3.3.2` — the version the plugin compiles against, which is what the old 2.4.0 pin existed to guarantee.

PR #7476, opened 2026-08-13
Improve URL staging resilience with configurable retry cooldown and regression coverage

### LINKED ISSUE #7475: Add configurable retry with cooldown for failed URL file staging
## Summary:
When input files are staged from HTTP or HTTPS sources, transient network failures can cause otherwise valid `nf-test` tests to fail. This is particularly visible in module tests that stage inputs from GitHub raw links. A stage-in retry policy would improve reliability without rerunning the full task. This could be generally useful beyond testing frameworks as well.

## Observed failures:

```
 Caused by:
    Can't stage file https://raw.githubusercontent.com/TheJacksonLaboratory/jds-nf-test/main/wgs/gvcf/unzip/example_101.gvcf -- reason: raw.
githubusercontent.com
```

```
ERROR ~ No such file or directory: https://raw.githubusercontent.com/TheJacksonLaboratory/jds-nf-test/main/lcwgs/mouse/lcwgs_HR_input.csv
```

```
curl: (6) Could not resolve host: www.nextflow.io
```

## Problem:
These errors are often transient DNS, host resolution, or temporary connectivity issues. Today they surface as hard stage-in failures, which fails the test even when a short wait and retry would succeed.

## Proposal:
Add configurable retry behavior for stage-in of remote URL inputs.

## Requested behavior:

1. When URL stage-in fails with retry-eligible errors, wait for cooldown and retry.
1. Retry up to N attempts, then fail normally if still unsuccessful.
1. Log each attempt with reason, wait duration, and attempts remaining.
1. Keep default behavior unchanged unless enabled.

## Configuration:

- stageIn.retry.enabled: false by default
- stageIn.retry.maxAttempts: N, for example 3
- stageIn.retry.cooldown: X duration, for example 10s
- stageIn.retry.backoff: fixed or exponential
- stageIn.retry.jitter: optional to reduce herd effects
- stageIn.retry.retryableErrors: optional override list or patterns

## Suggested default retryable conditions:

- DNS/host resolution errors, including curl error 6
- connection timeout, connection reset, temporary network unreachable
- HTTP 429 and HTTP 5xx
- temporary TLS handshake or socket transport failures

## Suggested default non-retryable conditions:

- confirmed HTTP 404 not found
- confirmed 401 or 403 auth and permission errors
- malformed URL or invalid scheme

## Note on ambiguous errors:
The message no such file or directory for URL sources may hide transport-level failures depending on how exceptions are wrapped. It would help if staging errors preserved the underlying cause type so retry classification is accurate.

## Why this matters:

- Reduces flaky test failures in CI for module pipelines that use remote fixture files
- Avoids expensive full task reruns when only stage-in was transiently broken
- Improves robustness for users depending on raw GitHub and other remote assets

## Acceptance criteria:

- Transient URL stage-in failures recover within configured attempts and task proceeds.
- Permanent URL failures fail fast or after policy as expected, with clear final message.
- Retry metadata appears in logs and trace output.
- Existing behavior remains unchanged when feature is disabled.

## Suggested implementation 

Add new features to `modules/nextflow/src/main/groovy/nextflow/file/FilePorter.groovy` and expand testing in `modules/nextflow/src/test/groovy/nextflow/file/FilePorterTest.groovy` to cover.

MikeWLloyd:
## Summary
This PR improves reliability for remote input staging (especially HTTP/HTTPS URLs such as raw GitHub links) by making retry cooldown configurable and refining retry classification for transient URL failures. It directly addresses #7475

## Motivation
Module and CI tests that stage remote files can fail due to transient DNS/network issues (for example host resolution failures). While staging retries already existed, retry delay was hardcoded and some URL-related error shapes were treated too aggressively as non-retryable.

## What changed
1.	Added configurable staging retry cooldown
•	New config key: filePorter.retryDelay
•	Default: 250ms
•	Existing filePorter.maxRetries behavior is preserved
2.	Updated staging retry wait behavior
•	Replaced small hardcoded/random sleep with configured retry delay
•	Retry log now includes the configured wait duration
3.	Refined retry classification for URL staging
•	Keeps fast-fail behavior for clearly non-retryable conditions
•	Allows retry for HTTP/HTTPS NoSuchFileException patterns that appear transient/wrapped
•	Avoids retries for clear missing-resource signals (for example 404/not found)
4.	Added targeted regression tests
•	Retries transient HTTP NoSuchFileException and succeeds
•	Retries transient DNS-like URL failure and succeeds
•	Does not retry clear missing HTTP resource (fails immediately as expected)

## Files touched
1.	FilePorter.groovy
2.	FilePorterTest.groovy

### Passed:
1.	JAVA_HOME=$(/usr/libexec/java_home -v 17) ./gradlew :nextflow:test --tests nextflow.file.FilePorterTest
2.	JAVA_HOME=$(/usr/libexec/java_home -v 17) ./gradlew :nextflow:test --tests 'nextflow.file.*'

## Backward compatibility
1.	Existing defaults remain compatible
2.	Retry count behavior is unchanged unless users tune settings
3.	New setting is optional

## Suggested reviewer focus
1.	Retry classification logic in FilePorter for HTTP/HTTPS NoSuchFileException paths
2.	Config naming/default for filePorter.retryDelay
3.	Coverage and intent of new FilePorter regression tests


bentsherman:
If we're going to update the FilePorter then I would just use a standard retry with exponential backoff + jitter like we do throughout the codebase (see SraRetryConfig for an example)

MikeWLloyd:
@bentsherman thanks, I can look at that as an example and adjust. Before digging into that, I just want to confirm, is this something that would be of interest to implement in the first place? I want to avoid putting in effort if this feature isn't something that would be implemented.

bentsherman:
Yes I think it makes sense to do this. We've been adding standard retry mechanisms across the codebase, and remote file staging clearly deserves the same treatment

PR #7488, opened 2026-08-15
Stop fabricating a memory request for the Seqera executor

pditommaso:
## Summary

The `memory` directive is optional, but the Seqera executor substituted `1024 MiB` whenever a process left it out — unchanged since the executor was introduced:

```groovy
int memoryMiB = task.config.getMemory() ? (int) (task.config.getMemory().toBytes() / (1024 * 1024)) : 1024
```

The scheduler therefore received a figure indistinguishable from a declared request, so it could neither apply its own default nor report that nothing had been asked for — and a run of unspecified tasks looked deliberately sized at 1 GiB.

## What changes

- **Absent directive** → omit `memoryMiB`, so the scheduler resolves and surfaces its own default.
- **Zero-valued directive** → omit too, with `log.warn1(…, firstOnly: true)` once per process, since unlike an absent directive it is an accident rather than a choice. Exactly zero is the only invalid figure reachable: `MemoryUnit(long)` asserts non-negative and `TaskConfig.getMemory0()` maps a falsy directive to `null`, so only the string form (`memory '0 GB'`) survives.
- **cpus untouched.** `TaskConfig.getCpus0()` already guarantees at least 1 and throws on a negative, so there is nothing to handle. The `?: 1` fallback it was wrapped in could never fire and only implied otherwise, so it is gone.

## A positive sub-MiB directive is forwarded on purpose

The conversion can still yield zero for a *positive* value under 1 MiB, and that is the case seen in production: `nf-core/bwa/index` — behind `NFCORE_SAREK:PREPARE_GENOME:BWAMEM1_INDEX` — declares `memory { 6.B * fasta.size() }`, which falls under 1 MiB for any reference below ~170 KB.

That zero is deliberately not caught here. Normalising a non-positive request is the scheduler's job, it already does it, and it reports the substituted default. Restating the rule client-side would split one normalisation across two codebases that then have to be kept in agreement — the failure this change is undoing. Recorded in the javadoc and pinned by a test.

## Ordering

Requires **seqeralabs/sched#1094** (closes seqeralabs/sched#1086), which resolves an unspecified request to a single default at admission and surfaces it on the task detail page. It is also what normalises the forwarded zero above.

**Must land after it** — the scheduler has to handle an absent axis before the plugin starts sending one. The default there is `1024 MiB`, deliberately the same figure this code substitutes today, so once both are in, behaviour is unchanged for existing pipelines; the scheduler simply knows the number is its own and says so.

## Test evidence

`where:`-driven cases on `memoryMiB()` (declared / absent / exactly 1 MiB / string-form zero / positive sub-MiB), plus a `submit()`-level test asserting an absent directive reaches the scheduler as an unset axis.

```
./gradlew :plugins:nf-seqera:test --tests '*SeqeraTaskHandlerTest*'   BUILD SUCCESSFUL
```

`SeqeraTaskHandlerTest` — 65 tests, 0 failures. `log.warn1(…, firstOnly: true)` is new to this file; the precedent is elsewhere (`FusionEnvProvider`, `TowerObserver`).

### DIFF
--- a/plugins/nf-seqera/src/main/io/seqera/executor/SeqeraTaskHandler.groovy
+++ b/plugins/nf-seqera/src/main/io/seqera/executor/SeqeraTaskHandler.groovy
@@ -245,11 +245,21 @@ class SeqeraTaskHandler extends TaskHandler implements FusionAwareTask {
      * whole run of unspecified tasks looked deliberately sized. Omitting the field instead lets the
      * scheduler resolve and surface its default (seqeralabs/sched#1086).
      * <p>
-     * A non-positive figure gets the same treatment for the same reason — it carries no usable
-     * intent — but it is worth telling the user about, because unlike an absent directive it is
-     * almost always a config accident: a {@code memory} closure that evaluated to zero, or a
-     * {@code params.max_memory} that was never set. Warned once per process rather than once per
-     * task, since the mistake belongs to the process definition.
+     * A zero-valued directive is omitted for the same reason, and warned about once per process,
+     * since unlike an absent directive it is a config accident rather than a choice. Exactly zero
+     * is the only invalid figure reachable here: {@code MemoryUnit(long)} asserts a non-negative
+     * value, and {@code TaskConfig.getMemory0} maps a falsy directive to null via
+     * {@code MemoryUnit.asBoolean}, so what survives is the string form — {@code memory '0 GB'} —
+     * whose string is truthy.
+     * <p>
+     * The conversion below can still yield zero for a <em>positive</em> sub-MiB directive: the
+     * size-derived idiom nf-core uses for index builds, {@code memory { 6.B * fasta.size() }} in
+     * {@code nf-core/bwa/index}, falls under 1 MiB for any reference below ~170 KB — which is what
+     * a test or CI profile supplies, and is why {@code BWAMEM1_INDEX} submits a zero today. That
+     * zero is forwarded deliberately rather than caught here. The scheduler already resolves any
+     * non-positive request to its default and reports having done so; restating that rule
+     * client-side would split one normalisation across two codebases that then have to be kept in
+     * agreement, which is the failure this whole change is undoing.
      *
      * @return the requested memory in MiB, or null to leave the axis unspecified
      */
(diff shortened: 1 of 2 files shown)


pditommaso:
@claude review

pditommaso:
Verified all four points against `origin/master`. Pushed `be1d02100`.

**The sub-MiB analysis is correct, and it is the production case.** `nf-core/bwa/index` — the module behind `NFCORE_SAREK:PREPARE_GENOME:BWAMEM1_INDEX` — declares `memory { 6.B * fasta.size() }`, which falls under 1 MiB for any reference below ~170 KB, i.e. every test/CI profile. So the daily `memoryMiB: 0` is a positive value truncating in the division, not a literal zero directive, and the `toBytes() <= 0` guard never sees it.

**Guard left where it is, deliberately.** Normalising a non-positive request is the scheduler's job and it already does it, surfacing the substituted default on the task detail page. Restating that rule client-side would split one normalisation across two codebases that then have to be kept in agreement — which is exactly the failure this change is undoing. The javadoc and a test case now record that the zero is forwarded on purpose and say where the rule lives.

Also taken:
- **Safe-navigation dropped.** Confirmed at `TaskRun.groovy:387` — `lazyName()` dereferences `processor` unconditionally in its fallback branch, so the elvis arm would NPE in the case `?.` appears to guard. Now plain `task.processor.name`.
- **Warning wording.** `MemoryUnit(long)` asserts `value>=0`, so "non-positive" could only ever mean exactly zero. Reworded, with a test case pinning that the string form is the only reachable zero.
- **`submit()`-level test added**, asserting an absent directive reaches the scheduler as an unset axis — the call site rather than the helper. 65 tests, 0 failures.

**And the description was wrong, my error.** There is no other `warn1` in this file on master; I read it on an unrelated in-flight branch. `firstOnly` is new here and the precedent is elsewhere (`FusionEnvProvider`, `TowerObserver`). Description corrected.

On the serialization question: it does not affect behaviour either way. `ResourceRequirement.memoryMiB` is a boxed `Integer`, and the scheduler keys its resolution on `getMemoryMiB() != null` after deserialization — an omitted field and an explicit `null` both arrive as null and both hit the default. The only difference is cosmetic, whether the stored request JSON carries a `"memoryMiB": null` entry.

PR #7503, opened 2026-08-18
Fix independently-verified findings from a security audit

ewels:
A security audit of this repo surfaced a handful of latent issues. I reproduced each one independently before touching anything, and only fixed what I could confirm was real.

Why these matter, not just what changed:

- Nextflow's Docker entrypoint builds a shell command from arbitrary caller-supplied arguments. If that reconstruction isn't airtight, anything able to influence those args (a container invoker, an orchestration layer) can break out of the intended command and run its own — the difference between "runs the pipeline task" and "runs whatever it wants as root inside the container."
- Debug/trace logging exists so people can troubleshoot connectivity problems, and troubleshooting logs get shared — pasted into tickets, Slack, screen shares — far more casually than anyone treats a credential. A JWT sitting in a trace line means a support workflow doubles as an exfiltration path.
- Zip extraction that trusts entry names is a well-known class of bug (Zip Slip): a crafted archive can write outside its intended target directory. The code path isn't wired up to anything today, but "nothing calls it yet" is a property of the current codebase, not a property of the function — it's exploitable the moment something does.
- The Gradle wrapper fetches a distribution over HTTPS but never checked what it got matched what was intended. That's the whole point of pinning a build tool's checksum: it turns "trust the network and the upstream host forever" into "trust it once, verify it every time after."

None of these needed new dependencies or behavior changes — each was a small, targeted correction once confirmed.

### DIFF
--- a/docker/entry.sh
+++ b/docker/entry.sh
@@ -36,9 +36,6 @@
 # enable debugging
 [[ "$NXF_DEBUG_ENTRY" ]] && set -x
 
-# wrap cli args with single quote to avoid wildcard expansion
-cli=''; for x in "$@"; do cli+="'$x' "; done
-
 # the NXF_USRMAP hold the user ID in the host environment
 if [[ "$NXF_USRMAP" ]]; then
 # create a `nextflow` user with the provided ID
@@ -50,13 +47,19 @@ useradd -u "$NXF_USRMAP" -G docker -s /bin/bash nextflow
 chown nextflow /var/run/docker.sock
 chown -R nextflow /.nextflow
 
+# `su` only accepts a command string, so re-quote each arg with `%q`
+# (not a hand-rolled `'$x'` wrap, which breaks — and lets the arg escape
+# the quoting — the moment an arg contains a single quote).
+cli=''; for x in "$@"; do printf -v esc '%q' "$x"; cli+="$esc "; done
+
 # finally run the target command with `nextflow` user
 su nextflow << EOF
 [[ "$NXF_DEBUG_ENTRY" ]] && set -x
 exec bash -c "$cli"
 EOF
 
-# otherwise just execute the command
+# otherwise just execute the command directly — no string round-trip, no
+# quoting to get wrong
 else
-exec bash -c "$cli"
+exec "$@"
 fi
--- a/docker/entry.sh
+++ b/docker/entry.sh
@@ -25,8 +25,8 @@
 # with such ID and adds it to the `docker` group, then assigns the docker
 # socket file ownership to that user.
 #
-# Finally it switches the `nextflow` user using the `su` command and
-# executes the original target command line.
+# Finally it switches to the `nextflow` user and executes the original
+# target command line.
 #
 # authors:
 #  Paolo Di Tommaso
@@ -47,16 +47,8 @@ useradd -u "$NXF_USRMAP" -G docker -s /bin/bash nextflow
 chown nextflow /var/run/docker.sock
 chown -R nextflow /.nextflow
 
-# `su` only accepts a command string, so re-quote each arg with `%q`
-# (not a hand-rolled `'$x'` wrap, which breaks — and lets the arg escape
-# the quoting — the moment an arg contains a single quote).
-cli=''; for x in "$@"; do printf -v esc '%q' "$x"; cli+="$esc "; done
-
-# finally run the target command with `nextflow` user
-su nextflow << EOF
-[[ "$NXF_DEBUG_ENTRY" ]] && set -x
-exec bash -c "$cli"
-EOF
+# run the target command as `nextflow`, passing argv through (no shell)
+exec runuser -u nextflow -- "$@"
 
 # otherwise just execute the command
 else
--- a/docker/entry.sh
+++ b/docker/entry.sh
@@ -58,8 +58,7 @@ su nextflow << EOF
 exec bash -c "$cli"
 EOF
 
-# otherwise just execute the command directly — no string round-trip, no
-# quoting to get wrong
+# otherwise just execute the command
 else
 exec "$@"
 fi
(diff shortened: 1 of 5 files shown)


docker/entry.sh
```diff
@@ -50,13 +47,19 @@ useradd -u "$NXF_USRMAP" -G docker -s /bin/bash nextflow
...
+
 # finally run the target command with `nextflow` user
 su nextflow << EOF
 [[ "$NXF_DEBUG_ENTRY" ]] && set -x
 exec bash -c "$cli"
 EOF
 
-# otherwise just execute the command
+# otherwise just execute the command directly — no string round-trip, no
+# quoting to get wrong
```
ewels:
```suggestion
# otherwise just execute the command
```

docker/entry.sh
```diff
@@ -50,6 +47,11 @@ useradd -u "$NXF_USRMAP" -G docker -s /bin/bash nextflow
 chown nextflow /var/run/docker.sock
 chown -R nextflow /.nextflow
 
+# `su` only accepts a command string, so re-quote each arg with `%q`
+# (not a hand-rolled `'$x'` wrap, which breaks — and lets the arg escape
+# the quoting — the moment an arg contains a single quote).
+cli=''; for x in "$@"; do printf -v esc '%q' "$x"; cli+="$esc "; done
```
pditommaso: The `%q` escaping is neutralized here, so the injection this PR aims to close is still exploitable.

`$cli` is interpolated into `exec bash -c "$cli"` on line 58 — i.e. inside **double quotes**, which are then re-parsed by a third shell. Inside double quotes `\`` and `\$` collapse back to a literal backtick and `$`, and `bash -c` then parses the result. Reproduced with exactly this code:

```console
$ bash t.sh echo 'X`id`Y' 'a$HOME'
--- script handed to su ---
exec bash -c "echo X\`id\`Y a\$HOME "
--- running it ---
Xuid=501(...) gid=20(staff) ...Y a/Users/...
```

So `docker run -e NXF_USRMAP=1000 img nextflow run '`touch /pwned`'` still runs attacker-chosen commands as the `nextflow` user. Two independent issues: (a) the double-quote wrapper strips the escaping, (b) `bash -c` adds a third parsing round. `%q` also emits `$'...\n...'` for args containing newlines or non-ASCII, which is inert inside double quotes and silently mangles the argument.

Two ways to fix. Minimal — drop the wrapper so `%q` output lands unquoted (verified safe for backticks, `$`, `'` and `*`):

```diff
 su nextflow << EOF
 [[ "$NXF_DEBUG_ENTRY" ]] && set -x
-exec bash -c "$cli"
+exec $cli
 EOF
```

Better — skip the shell entirely and pass argv straight through, matching the shape of the new `exec "$@"` in the `else` branch:

```sh
exec setpriv --reuid nextflow --regid docker --init-groups -- "$@"
# or: exec runuser -u nextflow -- "$@"
```

Minor, on the comment itself: "`su` only accepts a command string" isn't what the code does — there's no `-c`, the script is piped in on stdin. Worth rewording so the next reader doesn't keep the `bash -c` construct on the strength of that justification.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerXAuth.groovy
```diff
@@ -96,7 +97,7 @@ class TowerXAuth implements XAuthProvider {
                 .build()
 
         final resp = httpClient.send(req, HttpResponse.BodyHandlers.ofString())
-        log.debug "Refresh cookie response: [${resp.statusCode()}] ${resp.body()}"
+        log.debug "Refresh cookie response: [${resp.statusCode()}]"
```
pditommaso: Non-blocking: dropping `resp.body()` removes the only diagnostic for refresh failures — on a non-200 the method returns `false` having logged just the status code, so a 400 carrying `{"error":"invalid_grant"}` becomes undebuggable. The body is only sensitive in the success case (token arrives in the cookies), so it could be kept on the failure branch:

```groovy
if( resp.statusCode() != 200 ) {
    log.debug "Refresh cookie response: [${resp.statusCode()}] ${resp.body()}"
    return false
}
log.debug "Refresh cookie response: [${resp.statusCode()}]"
```

pditommaso (changes requested):
Thanks for picking these up — three of the four changes look correct to me. I verified that the pinned `distributionSha256Sum` matches the published checksum for `gradle-9.6.1-bin.zip`, and the Zip-Slip guard in `RemoteSession` is sound (`Path.startsWith` is element-wise so the `/tmp/foo` vs `/tmp/foobar` prefix trap doesn't apply, both sides are normalized, and absolute entry names like `/etc/passwd` are caught because `resolve` returns them verbatim). Token redaction in `TowerXAuth` is good too.

The blocker is `docker/entry.sh`: as written the `%q` re-quoting does **not** close the injection in the `NXF_USRMAP` branch — I reproduced it locally. Details inline.

docker/entry.sh
ewels: Good catch — reproduced locally. Dropped the `%q` / `bash -c` round-trip entirely and switched to `exec runuser -u nextflow -- "$@"` so argv passes through with no third shell parse (same shape as the `else` branch). Also added `util-linux` to the Dockerfile — the AL2023 base image does not ship `su`/`runuser`/`setpriv` with the current package set, so the old `su` heredoc path would not have worked anyway. Verified the injection payload (`X\`id\`Y`, `$HOME`, quotes, glob) stays literal and runs as the `nextflow` user.

plugins/nf-tower/src/main/io/seqera/tower/plugin/TowerXAuth.groovy
ewels: Done — response body is logged on non-200 only; success path logs status code without the JWT-bearing body.

docker/entry.sh
ewels: Applied in 2d07bcd.

pditommaso (approved):

PR #7504, opened 2026-08-18
Authorize Claude PR reviews against live collaborator permission

ewels:
The `@claude` review trigger is gated on `author_association`, which describes the comment, not the commenter's access to this repo — an org `MEMBER` can carry that association with no write access here.

`claude-code-action` already rejects non-write actors before Claude starts, so that gap does not let someone post an approving review. What it does allow is burning a runner and leaving a red X on the PR, and it only fails after checkout.

This adds a live collaborator-permission check (admin/write) that runs before checkout. Unauthorized mentions skip cleanly and get a one-line reply. Same pattern as the e2e PR-comment trigger in #7192.


ewels:
@claude review

seems appropriate somehow...? 😅

pditommaso (approved):
Approving — the change is sound and the structural win is real: the gate runs *before* the untrusted PR head is checked out, it fails closed (default `bash -e`, so a failed `gh api` kills the job rather than falling through to an empty `authorized`), `ACTOR` is passed via `env` rather than inlined so there's no injection surface, and both downstream steps carry the guard. The `author_association` pre-filter also can't produce a false negative: write access always implies `OWNER`/`MEMBER`/`COLLABORATOR`, so nobody authorized gets dropped before the new step.

One correction for the record, since the comment in the file is the thing future readers will trust. The description says an unauthorized commenter "could still get a run to fire and post an approval" — the approval half isn't reachable. `claude-code-action` v1.0.191 already runs this same check in agent mode: `src/entrypoints/run.ts:193` calls `checkWritePermissions` for entity contexts (`issue_comment`, `pull_request_review*`), and `src/github/validation/permissions.ts:124-139` hits the identical `getCollaboratorPermissionLevel` endpoint accepting only `admin`/`write`, then throws before Claude starts. What an org `MEMBER` without write access could actually do is burn a runner and leave a red X — exactly what the replaced comment described. This PR is still worth having (defence in depth that doesn't depend on a third party's internals, gate ahead of checkout, red X becomes a clean skip), but the framing overstates it.

Nice-to-haves, none blocking:

- **`"maintain"` is unreachable.** `.permission` is the legacy field — `admin`/`write`/`read`/`none`, with maintain mapped to write and triage to read. Behaviour is correct; the literal is dead. Worth dropping or annotating. Don't switch to `.role_name` to "fix" it: that returns custom org role names the allowlist wouldn't cover, so `.permission` is the safer field here.
- **Say something when rejected.** The job currently goes green with every step skipped, which reads as "I mentioned @claude and nothing happened". Everyone reaching the Authorize step has already passed the MEMBER/COLLABORATOR pre-filter, so there's no spam vector — a one-line `gh pr comment` (as the e2e workflow in #7192 does) or a `$GITHUB_STEP_SUMMARY` note would cover it.
- **Authorization lives in a repeated step-level `if:`.** A step added later without the guard runs unauthorized, silently. A separate `authorize` job with a job output plus `needs:` and one job-level `if:` makes that impossible to get wrong. #7192 uses the same step shape, so consistency argues the other way — a comment noting the invariant would be enough.
- **Bot actors diverge from the action.** The action treats any `[bot]` actor as authorized; this gate rejects them (`dependabot[bot]` returns `none`). Probably the behaviour you want, worth a line so it isn't rediscovered later.

Two things to be aware of rather than change:

- **This can't be exercised before merge.** `issue_comment` and `pull_request_review*` run the workflow definition from the base branch, so the Authorize step only goes live once this lands (the run on this PR skipped at the job `if:` using master's file). The one runtime unknown is whether `secrets.GITHUB_TOKEN` under this job's `permissions:` block reads `/collaborators/{user}/permission` — the docs state no push-access requirement and the action makes the same call today, so it should be fine, and the failure mode is loud rather than silently-authorize. Worth watching the first mention after merge.
- **#7192 conflicts.** It deletes `.github/workflows/claude.yml` and folds it into `pr-commands.yml` with its own Authorize step. Whichever merges second needs a rebase — cleanest is this one first, then #7192 rebases onto it.

ewels:
Picked up the review notes:

- Dropped the unreachable `maintain` comparison (`.permission` is admin/write/read/none; maintain maps to write)
- Added a one-line PR reply when the authorize step denies
- Noted that every step after Authorize must keep the guard
- Tightened the framing: this is defence in depth ahead of checkout / a clean skip instead of a red X — the action already blocks non-write actors before Claude runs, so an unauthorized org MEMBER could not have posted an approval

